PR #23, opened 2019-03-19
Issue #6 Add initial preflight checks for linux host

### LINKED ISSUE #6: Add basic pre-flight checks and setup (Linux / Libvirt)
Depends on: #9 

Which checks are needed for Linux; libvirt?

### LINKED ISSUE #28: Add config abstraction
For use with #6

### LINKED ISSUE #35: Logging framework for CRC
We need a logging framework for CRC which will enable us to collect the logs in a log file. This will be beneficial for many reasons, such as
 * It should help system tray users when CRC needs to notify about some error and user should be able to able to find the details logs. 
* It would help issue tracking and debugging for failures easier.
* It will simplify the code in log run i.e. we do not need to add print statements for printing it to console. 

Here are some examples of logging framework

[1] https://github.com/sirupsen/logrus
[2] https://github.com/uber-go/zap

### LINKED ISSUE #40: Add crc base dir as ~/.crc

anjannath:
Network related checks not there yet...

What should we check regarding the network for libvirt/linux setup?

/cc @praveenkumar 

Fixes #6 
Fixes #35 
Fixes #28 
Fixes #40


gbraad:
We have never really tested if a single-node cluster can run on a different network (IP range) yet.

But at the moment we assume there is a specific network, let's call it `crc`, and it this does not exist, we can use `virsh net-define [network-def.xml]`. the problem we might have is that the same MAC has to be used for generating the image, or that we have to change the mac address each time when we import a new image. :-/

cmd/crc/cmd/setup.go
```diff
@@ -19,4 +20,6 @@ var setupCmd = &cobra.Command{
 }
 
 func runSetup(arguments []string) {
+	// check if user is root if not exit
```
gbraad: a `TODO`, right? If so, add this clearly:

```
    // TODO: 
```

pkg/crc/preflight/preflight.go
```diff
@@ -0,0 +1,115 @@
...
+	preflightCheckSucceedsOrFails("",
+		checkDefaultPoolHasSufficientSpace,
+		"Checking if default pool has sufficient free space",
+		"",
+	)
+}
+
+// SetupHost performs the prerequisite checks and setups the host to run the cluster
+func SetupHost() {
+	preflightCheckAndFix(checkVirtualizationEnabled, fixVirtualizationEnabled, "Setting up virtualization")
```
gbraad: format both blocks in the same way... else it looks VERY different and unclear what happens (difference is).

pkg/crc/preflight/preflight.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/code-ready/crc/pkg/crc/output"
+	"github.com/spf13/viper"
+	"os"
+)
+
+type PreflightCheckFixFuncType func() (bool, error)
+
+func preflightCheckSucceedsOrFails(configNameOverrideIfSkipped string, check PreflightCheckFixFuncType, message string, configNameOverrideIfWarning string) {
+	output.Out("-- %s ... ", message)
+	isConfiguredToSkip := viper.GetBool(configNameOverrideIfSkipped)
```
gbraad: This has to be provided as a config object/struct to the checks...
... we should not have to use `viper` in the package level.

gbraad: we can hardcode this now, but have a dynamic solution later.

gbraad: you could just pass an array of `strings:bool`. you could populate this with all the viper boolean values (for now).

pkg/crc/preflight/preflight_checks.go
```diff
@@ -0,0 +1,325 @@
...
+		path, err = os.Readlink(path)
+		if err != nil {
+			return false, errors.New("Libvirt cli virsh was not found in path")
+		}
+	}
+	return true, nil
+}
+
+func fixLibvirtInstalled() (bool, error) {
+	cmd := exec.Command("sudo", "dnf", "install", "-y", "libvirt", "libvirt-devel", "libvirt-daemon-kvm", "qemu-kvm")
```
gbraad: why `libvirt-devel`?

pkg/crc/preflight/preflight_checks.go
```diff
@@ -0,0 +1,325 @@
...
+
+	if size, _ := units.FromHumanSize(defaultPoolSize); int64(freeSpace.Available) >= size {
+		return true, nil
+	}
+	return false, fmt.Errorf("Not enough space in default pool, available free space: %s", units.HumanSize(freeSpace.Available))
+}
+
+func fixDefaultPoolHasSufficientSpace() (bool, error) {
+	// edit default pool to increase available space
+	return true, nil
```
gbraad: suggestion might be needed... as it would be hard to fix this automatically.

pkg/crc/preflight/preflight_checks.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+	err = xml.Unmarshal(stdOut, &freeSpace)
+	if err != nil {
+		return false, err
+	}
+
+	if size, _ := units.FromHumanSize(defaultPoolSize); int64(freeSpace.Available) >= size {
+		return true, nil
+	}
+	return false, fmt.Errorf("Not enough space in default pool, available free space: %s", units.HumanSize(freeSpace.Available))
```
gbraad: perhaps we should actually not use the default pool. as the KVM driver uses user session (at least, the images are located in the home folder)

pkg/crc/preflight/preflight_checks.go
anjannath: IIRC I saw this requirement in the installer readme and added it here.. i'll recheck and verfiy

pkg/crc/preflight/preflight_checks.go
anjannath: Hmm... that means we could define a new pool just for crc purpose (probably have to use the system session) and then the driver can use that

pkg/crc/preflight/preflight.go
anjannath: Using viper here also preserves the priority of the various config values, ENV_VARIABLE vs FLAGS etc..

gbraad: let's create a new task for this.

gbraad: #28

pkg/crc/preflight/preflight_checks.go
```diff
@@ -0,0 +1,430 @@
...
+	<dns>
+	  <srv service='etcd-server-ssl' protocol='tcp' domain='test1.tt.testing' target='etcd-0.test1.tt.testing' port='2380' weight='10'/>
+	  <host ip='192.168.126.11'>
+		<hostname>api.test1.tt.testing</hostname>
+		<hostname>etcd-0.test1.tt.testing</hostname>
+	  </host>
+	</dns>
+	<ip family='ipv4' address='192.168.126.1' prefix='24'>
+	  <dhcp>
+		<host mac='52:fd:fc:07:21:82' name='test1-nfxdm-master-0' ip='192.168.126.11'/>
```
gbraad: This is risky... as the MAC is not static.
this needs to be part of a template perhaps...

gbraad: or this should be read from the archive with the disk images.

@praveenkumar how to handle a possible static mac?

pkg/crc/preflight/preflight_checks.go
```diff
@@ -0,0 +1,430 @@
...
+
+	cputype := re.FindString(flags)
+	if cputype == "" {
+		return false, errors.New("Virtualization is not available for you CPU")
+	}
+	return true, nil
+}
+
+func fixVirtualizationEnabled() (bool, error) {
+	return false, errors.New("You need to enable virtualization in bios")
```
gbraad: `s/bios/BIOS`

anjannath:
The mac address can be specified in the domain xml, and we can have a
static mac address as constant.

but the kvm machine driver's xml template doesn't have it as of now.

pkg/crc/preflight/preflight_checks.go
anjannath: The mac address can be specified in the domain xml, and we can have a static mac address as constant, and use template to substitute it, in the network xml definition 

but the kvm machine driver's xml template doesn't have the field for it as of now.

pkg/crc/preflight/preflight.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	output.Out("FAIL")
+	errorMessage := fmt.Sprintf("   %s", err.Error())
+	output.Out(errorMessage)
+	if isConfiguredToWarn {
+		output.Out(errorMessage)
+	} else {
+		os.Exit(1)
+	}
```
praveenkumar: @anjannath shouldn't this be like?

```golang
        errorMessage := fmt.Sprintf("   %s", err.Error())
	if isConfiguredToWarn {
                output.Out("WARN")
		output.Out(errorMessage)
                 return
	}
        output.Out("FAIL")
	output.Out(errorMessage)
        os.Exit(1)
```

anjannath: @praveenkumar Thanks.. Fixed it!

cmd/crc/cmd/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+	SkipCheckCrcNetwork         = createSetting("skip-check-crc-network", false)
+	WarnCheckCrcNetwork         = createSetting("warn-check-crc-network", false)
+	SkipCheckCrcNetworkActive   = createSetting("skip-check-crc-network-active", false)
+	WarnCheckCrcNetworkActive   = createSetting("warn-check-crc-network-active", false)
+)
+
+// CreateSetting returns a filled struct of ConfigSetting
+// takes the config name and default value as arguments
+func createSetting(name string, defalut interface{}) *setting {
+	s := setting{Name: name, DefaultValue: defalut}
```
gbraad: default

cmd/crc/cmd/root.go
```diff
@@ -38,3 +47,20 @@ func Execute() {
 		os.Exit(1)
 	}
 }
+
+func initViper() {
+	v := viper.New()
+	v.SetConfigFile(constants.ConfigPath)
```
gbraad: where is configpath?

pkg/crc/constants/constants.go
```diff
@@ -1,9 +1,39 @@
...
+	NodeMac      = "52:fd:fc:07:21:82"
+	NodeIP       = "192.168.126.11"
+	PoolName     = "crc"
+	PoolDir      = "/var/lib/libvirt/images"
+	CrcEnvPrefix = "CRC"
 )
+
+var (
+	ConfigFile = ".crc.json"
+	ConfigPath = filepath.Join(GetHomeDir(), ConfigFile)
```
gbraad: ah, OK.
so this lives in `~/file` ?

pkg/crc/constants/constants.go
```diff
@@ -1,9 +1,39 @@
...
+	DomName      = "crc"
+	NodeMac      = "52:fd:fc:07:21:82"
+	NodeIP       = "192.168.126.11"
+	PoolName     = "crc"
+	PoolDir      = "/var/lib/libvirt/images"
+	CrcEnvPrefix = "CRC"
 )
+
+var (
+	ConfigFile = ".crc.json"
```
gbraad: hmmm... have to create the directory structure soon.

pkg/crc/preflight/preflight.go
```diff
@@ -56,60 +54,60 @@ func preflightCheckAndFix(check, fix PreflightCheckFixFuncType, message string)
 
 // StartPreflightChecks performs the preflight checks before starting the cluster
 func StartPreflightChecks() {
-	preflightCheckSucceedsOrFails("",
+	preflightCheckSucceedsOrFails(config.GetViperBool(cmdConfig.SkipCheckVirtEnabled.Name),
 		checkVirtualizationEnabled,
 		"Checking if Virtualization is enabled",
-		"",
+		config.GetViperBool(cmdConfig.WarnCheckVirtEnabled.Name),
```
gbraad: we might want to consider this to change down the line...
we still use config that is typed by naming... and located in `cmd`.
for now this is better than before...

pkg/crc/config/viper_config.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+)
+
+var (
+	GlobalViper *viper.Viper
+	ViperConfig map[string]interface{}
+)
+
+// GetViperBool returns the value of a boolean config key
+func GetViperBool(key string) bool {
```
gbraad: Ah, this is in pkg... so in that case do not use the name `Viper` but just `GetBool`...
the same for `ViperConfig`...

in this level we should have abstracted most of it away.

cmd/crc/cmd/root.go
```diff
@@ -3,9 +3,14 @@ package cmd
...
 	"github.com/spf13/cobra"
 	"os"
+	"strings"
 
+	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
+	"github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/output"
+	"github.com/spf13/viper"
```
gbraad: what is still `viper` that we use in this file?
part of the initi... right? what actually happens here:

  * we set the configfile
  * we set a prefix for the replacer
  * load the config

can this we moved perhaps?

anjannath: Yes, i was actually thinking of putting it in `pkg/crc/config/viper_config.go`

pkg/crc/preflight/preflight_checks.go
gbraad: we should test if the mac can be changed from the image... as it is only the DNS overlay that might need it

pkg/crc/constants/constants.go
anjannath: Yes

cmd/crc/cmd/root.go
anjannath: `$HOME/.crc.json` for now, till we have the directory structure in place.

pkg/crc/preflight/preflight.go
```diff
@@ -5,13 +5,13 @@ import (
...
 	"github.com/code-ready/crc/pkg/crc/output"
-	"os"
+	log "github.com/sirupsen/logrus"
 )
 
 type PreflightCheckFixFuncType func() (bool, error)
 
 func preflightCheckSucceedsOrFails(isConfiguredToSkip bool, check PreflightCheckFixFuncType, message string, isConfiguredToWarn bool) {
-	output.Out("-- %s ... ", message)
+	log.Infof("%s ... ", message)
```
anjannath: So, this is how i am planning us to use the logrus pkg, whenever we need to log to conosle/file
Everything is logged to file by, and all info msgs are logged to console by default which can be changed using the `log-level` global flag

cmd/crc/cmd/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+	WarnCheckKvmDriver          = createSetting("warn-check-kvm-driver", false)
+	SkipCheckDefaultPool        = createSetting("skip-check-default-pool", false)
+	WarnCheckDefaultPool        = createSetting("warn-check-default-pool", false)
+	SkipCheckDefaultPoolSpace   = createSetting("skip-check-default-pool", false)
+	WarnCheckDefaultPoolSpace   = createSetting("warn-check-default-pool", false)
+	SkipCheckCrcNetwork         = createSetting("skip-check-crc-network", false)
+	WarnCheckCrcNetwork         = createSetting("warn-check-crc-network", false)
+	SkipCheckCrcNetworkActive   = createSetting("skip-check-crc-network-active", false)
+	WarnCheckCrcNetworkActive   = createSetting("warn-check-crc-network-active", false)
+)
```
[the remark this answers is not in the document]
anjannath: For now just checking if the user is providing a boolean value or not (`true | false`) as only those values can be set as config now, as we add more config parameters, we can add new validations.

pkg/crc/config/viper_config.go
```diff
@@ -0,0 +1,78 @@
+package config
```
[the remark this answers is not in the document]
anjannath: Yes, this way we can import the config packag in `cmd` as well `pkg` namespace.

pkg/crc/preflight/preflight_checks.go
```diff
@@ -0,0 +1,474 @@
...
+	stdOut, err := cmd.Output()
+	if err != nil {
+		return false, err
+	}
+	err = xml.Unmarshal(stdOut, &freeSpace)
+	if err != nil {
+		return false, err
+	}
+
+	if size, _ := units.FromHumanSize(defaultPoolSize); freeSpace.Available >= float64(size) {
```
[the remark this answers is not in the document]
anjannath: Done

pkg/crc/preflight/preflight.go
```diff
@@ -0,0 +1,159 @@
...
+		"Checking if default pool has sufficient free space",
+		config.GetBool(cmdConfig.WarnCheckDefaultPoolSpace.Name),
+	)
+	preflightCheckSucceedsOrFails(config.GetBool(cmdConfig.SkipCheckCrcNetwork.Name),
+		checkLibvirtCrcNetworkAvailable,
+		"Checking if Libvirt crc network is available",
+		config.GetBool(cmdConfig.WarnCheckCrcNetwork.Name),
+	)
+	preflightCheckSucceedsOrFails(config.GetBool(cmdConfig.SkipCheckCrcNetworkActive.Name),
+		checkLibvirtCrcNetworkAvailable,
```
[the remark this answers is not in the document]
anjannath: Fixed the typo

pkg/crc/preflight/preflight_checks.go
```diff
@@ -0,0 +1,474 @@
...
+		path, err = os.Readlink(path)
+		if err != nil {
+			return false, errors.New("Libvirt cli virsh was not found in path")
+		}
+	}
+	return true, nil
+}
+
+func fixLibvirtInstalled() (bool, error) {
+	cmd := exec.Command("sudo", "dnf", "install", "-y", "libvirt", "libvirt-daemon-kvm", "qemu-kvm")
```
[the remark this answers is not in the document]
anjannath: Done

cmd/crc/cmd/config/config.go
```diff
@@ -0,0 +1,75 @@
+package config
+
+import (
+	validations "github.com/code-ready/crc/pkg/crc/config"
```
gbraad: :+1:

cmd/crc/cmd/config/config.go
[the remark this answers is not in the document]
gbraad: understandable. Subject to refactoring anyway

praveenkumar:
@anjannath `config view` command is showing the default values of the config instead it should just show what user set without the default one.

cmd/crc/cmd/config/config.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/spf13/cobra"
+	"strings"
+)
+
+type validationFnType func(interface{}, interface{}) bool
```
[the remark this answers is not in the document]
anjannath: Shoot... first argument should be a `string`, forgot to change it here.
@praveenkumar :)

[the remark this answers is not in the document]
anjannath: Also, we need two because sometime we might need the key also to validate.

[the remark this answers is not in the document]
anjannath: @praveenkumar the change is in here https://github.com/code-ready/crc/pull/23/commits/e5700274cabbd62392c44b84ad14e3ef15fd6b0d

anjannath:
@praveenkumar Pushed the modifications

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -0,0 +1,479 @@
...
+		return false, err
+	}
+	var netXMLDef strings.Builder
+	err = t.Execute(&netXMLDef, config)
+	if err != nil {
+		return false, err
+	}
+	cmd := exec.Command("virsh", "--connect", "qemu:///system", "net-define", "/dev/stdin")
+	cmd.Stdin = strings.NewReader(netXMLDef.String())
+	err = cmd.Run()
```
praveenkumar: Each `cmd` which is executed using the `exec` should also have `stderr`defined otherwise in case of failure it will just give the fatal but not the actual error.

Assume you have the `tt0` bridge already available and then try to create the same.

```
$ sudo ./crc setup
[sudo] password for prkumar: 
crc - Local OpenShift 4.0 clusters
INFO -- [Setting up virtualization] ...           
INFO -- [Setting up KVM] ...                      
INFO -- [Installing Libvirt] ...                  
INFO -- [Adding user to libvirt group] ...        
INFO -- [Enabling libvirt] ...                    
INFO -- [Starting Libvirt service] ...            
INFO -- [Installing docker-machine-driver-kvm] ...  
INFO -- [Creating default storage pool] ...       
INFO -- [Setting up default pool] ...             
INFO -- [Setting up Libvirt crc network] ...      
FATAL [exit status 1]  ===> This doesn't give any idea why it is failed :(
```

```golang
	cmd := exec.Command("virsh", "--connect", "qemu:///system", "net-define", "/dev/stdin")
	cmd.Stdin = strings.NewReader(netXMLDef.String())
	buf := new(bytes.Buffer)
	cmd.Stderr = buf
	err = cmd.Run()
	if err != nil {
		return false, fmt.Errorf("%v, %s", err, buf.String())
	}
```

```
$ sudo ./crc setup
[sudo] password for prkumar: 
crc - Local OpenShift 4.0 clusters
INFO -- [Setting up virtualization] ...           
INFO -- [Setting up KVM] ...                      
INFO -- [Installing Libvirt] ...                  
INFO -- [Adding user to libvirt group] ...        
INFO -- [Enabling libvirt] ...                    
INFO -- [Starting Libvirt service] ...            
INFO -- [Installing docker-machine-driver-kvm] ...  
INFO -- [Creating default storage pool] ...       
INFO -- [Setting up default pool] ...             
INFO -- [Setting up Libvirt crc network] ...      
FATAL [exit status 1, error: Failed to define network from /dev/stdin
error: internal error: bridge name 'tt0' already in use.   => this now gives why it failed.
]
```

anjannath: @praveenkumar Updated the PR, have a look

cmd/crc/cmd/config/view.go
```diff
@@ -0,0 +1,73 @@
...
+	"text/template"
+)
+
+const (
+	DefaultConfigViewFormat = "- {{.ConfigKey | printf \"%-35s\"}}: {{.ConfigValue}}"
+)
+
+var configViewFormat string
+
+type ConfigViewTemplate struct {
```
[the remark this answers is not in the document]
anjannath: Done!

PR #41, opened 2019-04-12
Issue #13 Add code for --output json for commands

### LINKED ISSUE #13: Commands can be configured to print output in json
CRC should be able to give command output in json  if required.

LalatenduMohanty:
Fixes #13 

Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

### DIFF
--- a/cmd/crc/cmd/version.go
+++ b/cmd/crc/cmd/version.go
@@ -1,13 +1,36 @@
+/*
+Copyright (C) 2019 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 
 	crcPkg "github.com/code-ready/crc/pkg/crc"
+	outputPkg "github.com/code-ready/crc/pkg/crc/output"
 	"github.com/spf13/cobra"
 )
 
+var (
+	outputFormat string
+)
+
 func init() {
+	versionCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "json, for output in json format")
 	rootCmd.AddCommand(versionCmd)
 }
 
@@ -21,5 +44,15 @@ var versionCmd = &cobra.Command{
 }
 
 func runPrintVersion(arguments []string) {
-	fmt.Printf("version: %s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetCommitSha())
+	output := fmt.Sprintf("%s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetCommitSha())
+	if outputFormat == "json" {
+		strOutput := outputPkg.VersionOutputStruct{output}
+		jsonOutput, err := json.Marshal(strOutput)
+		if err != nil {
+			fmt.Println(err)
+		}
+		fmt.Println(string(jsonOutput))
+	} else {
+		fmt.Println("Version: ", output)
+	}
 }
--- a/pkg/crc/output/types.go
+++ b/pkg/crc/output/types.go
@@ -1,0 +1,25 @@
+/*
+Copyright (C) 2019 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


cmd/crc/cmd/version.go
```diff
@@ -1,13 +1,36 @@
...
+*/
+
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 
 	crcPkg "github.com/code-ready/crc/pkg/crc"
+	outputPkg "github.com/code-ready/crc/pkg/crc/output"
```
gbraad: no need to add `pkg`

cmd/crc/cmd/version.go
```diff
@@ -1,13 +1,36 @@
...
+limitations under the License.
+*/
+
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 
 	crcPkg "github.com/code-ready/crc/pkg/crc"
```
gbraad: in this case calling it crc would be good enough, or alias as `libcrc`

cmd/crc/cmd/version.go
```diff
@@ -21,5 +44,15 @@ var versionCmd = &cobra.Command{
 }
 
 func runPrintVersion(arguments []string) {
-	fmt.Printf("version: %s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetCommitSha())
+	output := fmt.Sprintf("%s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetCommitSha())
+	if outputFormat == "json" {
+		strOutput := outputPkg.VersionOutputStruct{output}
```
gbraad: the output struct is not needed inside the `output` package. the command in this case can handle. however, the `output` library can be instructed what needs to be printed in a normal case FROM this struct if needed.

pkg/crc/output/types.go
```diff
@@ -0,0 +1,25 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package output
+
+type VersionOutputStruct struct {
```
gbraad: I'm not totally convinced this to be part of output, as at the moment the pkg itself is not reponsible for filling the struct.

I'll have to 'sleep' about this...

gbraad:
I do not think we need to make the structs part of the Output package.

rather, each command is repsonsible for returning a struct (without json annotated field), and on command level we just have a general compare for `-o json` that calls appropriately

```
if json {
   output.Json(commandOutput)
} else {
   output.Out("Status: %s", commandOutput.Status)
}

// and the output being
output.OutJson(input interface{}) {
    encoded, err := json.Marshal(emp)
    if err != nil {
        output.Error(err)
        return
    }
    output.Out(encoded)
}
```

praveenkumar:
> Note; each functionality inside pkg/ is responsible for their own struct for output. As you can see in the Start function for machine I call machine.Start from the command. This Start function would now return a simple bool, errorbut this should becomeMachineStartStruct, error. But it would be unwise to define all of these inside the output` packages.

My thinking is to have the structures as part of command itself and then just pass it to generate the json encoding, but we need to define what all we need for a specific command json output. So let's say our `crc start` command require the output as json then what would be content we are returning that structure should be part of the command. when you say functionality inside pkg/ does that mean now our json command output now content multiple json or we are going to merge them?

gbraad:
@praveenkumar @LalatenduMohanty I was thinking along the lines of. WDYT?

cmd/crc/cmd/version.go
LalatenduMohanty: I think `libcrc` is better.

cmd/crc/cmd/version.go
LalatenduMohanty: You mean `outputPkg`? Usually it is good practice to add `Pkg` as we might be having similar names in `cmd` or `pkg`. I do not see a downside while using `outputPkg` as the name.

pkg/crc/output/types.go
LalatenduMohanty: Right. I will also give it another thought. We might be able to find a better place for it.

LalatenduMohanty:
> `But it would be unwise to define all of these inside the `output` package.

I am fine to define it in respective commad.go file. That will be simple I think. 

> but we need to define what all we need for a specific command json output. So let's say our `crc start` command require the output as json then what would be content we are returning that structure should be part of the command.

+1

gbraad:
Since #63 got merged, some of the expected structure can be seen in the `start`, `stop` and `delete` command.

cmd/crc/cmd/version.go
gbraad: for consistency with other code we do not add `Pkg`; alias is not needed.

gbraad:
For the moment we have postponed the JSON output, but we will keep this PR open. We have decided that each external function, ued by commands, have to accept a struct and return a struct when used.

PR #92, opened 2019-04-25
Issue #78 Add preflight code for dnsmasq crc.conf

gbraad:
Fixes #78 
Fixes #94 
Closes #85

PR #96, opened 2019-04-25
[WIP] Issue #94 Add systemd interaction

### LINKED ISSUE #94: Add functionality to interact with systemd services
Needed for #92

gbraad:
Work in progress.

Fixes #94 
Needed by #92

### DIFF
--- a/pkg/crc/systemd/systemd.go
+++ b/pkg/crc/systemd/systemd.go
@@ -1,0 +1,94 @@
+package systemd
+
+import (
+	"fmt"
+
+
+	crcos "github.com/code-ready/crc/pkg/os"
+
+	"github.com/code-ready/crc/pkg/crc/systemd/actions"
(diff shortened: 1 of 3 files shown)


pkg/crc/systemd/systemd.go
```diff
@@ -0,0 +1,90 @@
+package systemd
+
+import (
+	"fmt"
+
+
+	. "github.com/code-ready/crc/pkg/os"
```
praveenkumar: Do explicit mention the pkg here like for this one you can use `crcos`, it will help to track which functions coming from which package.

gbraad: I hate this too. Noticed the pre-flights did the same and this causes "Goto Definition" not to work correctly.

gbraad: Will also fix this in the preflight code...

gbraad:
This got added to the PR #92

Issue #111, opened 2019-05-06
Add tests for setup config settings


jsliacan:
Write tests for `crc config set`, `crc config get`, `crc config view`. The settings should be reflected in the behavior of `crc setup` and `crc start`. 

For instance, after `crc config set skip-check-crc-network true` one should see 
```
"skip-check-crc-network": "true"
```
in `~/.crc/crc.json` file. Hence, `crc config view` should display this property as set to `true` and `crc config get skip-check-crc-network` should return the same. Consequently, `crc setup` should skip the relevant check and `crc start` should not perform the check either. 

Ideally, this will work for all configurable properties (see output of `crc config --help`).

gbraad:
Make sure the check fails once without the network existing... and
when skipped it should not error out of the command.

On Mon, May 6, 2019 at 5:20 PM jakub sliacan <notifications@github.com> wrote:
>
> Write tests for crc config set, crc config get, crc config view. The settings should be reflected in the behavior of crc setup and crc start.
>
> For instance, after crc config set skip-check-crc-network true one should see
>
> "skip-check-crc-network": "true"
>
> in ~/.crc/crc.json file. Hence, crc config view should display this property as set to true and crc config get skip-check-crc-network should return the same. Consequently, crc setup should skip the relevant check and crc start should not perform the check either.
>
> Ideally, this will work for all configurable properties (see output of crc config --help).
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.



-- 

   Gerard Braad | http://gbraad.nl
   [ Doing Open Source Matters ]

praveenkumar:
Fixed by #121

PR #121, opened 2019-05-09
Issue 111 Initial tests for setup config

jsliacan:
As far as I can tell this addresses all items mentioned in Issue #111 and comments below. I am happy to modify & add further checks.

### DIFF
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -15,43 +15,182 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/*
-
-This file is not needed yet.
-
-*/
-
 package crcsuite
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/DATA-DOG/godog"
-	"github.com/code-ready/clicumber/testsuite"
+	"github.com/DATA-DOG/godog/gherkin"
+
+	clicumber "github.com/code-ready/clicumber/testsuite"
 )
 
-var sh testsuite.ShellInstance
+var (
+	CRCHome    string
+	bundleURL  string
+	bundleName string
+)
 
-func crcStateIs(expected string) error {
+// FeatureContext defines godog.Suite steps for the test suite.
+func FeatureContext(s *godog.Suite) {
 
-	fmt.Println("checking state of CRC...")
-	err := testsuite.ExecuteCommand("crc state")
+	// CRC related steps
+	s.Step(`^removing CRC home directory succeeds$`,
+		RemoveCRCHome)
+	s.Step(`^starting CRC with default bundle and default hypervisor (succeeds|fails)$`,
+		StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails)
+	s.Step(`^starting CRC with default bundle and hypervisor "(.*)" (succeeds|fails)$`,
+		StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails)
+	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds$|fails)$`,
+		SetConfigPropertyToValueSucceedsOrFails)
 
+	// CRC file operations
+	s.Step(`^file "([^"]*)" exists in CRC home folder$`,
+		FileExistsInCRCHome)
+	s.Step(`"(JSON|YAML)" config file "(.*)" in CRC home folder (contains|does not contain) key "(.*)" with value matching "(.*)"$`,
+		ConfigFileInCRCHomeContainsKeyMatchingValue)
+	s.Step(`"(JSON|YAML)" config file "(.*)" in CRC home folder (contains|does not contain) key "(.*)"$`,
+		ConfigFileInCRCHomeContainsKey)
+	s.Step(`removing file "(.*)" from CRC home folder succeeds$`,
+		DeleteFileFromCRCHome)
+
+	s.BeforeSuite(func() {
+		// set CRC home var
+		CRCHome = SetCRCHome()
+
+		// remove $HOME/.crc
+		err := RemoveCRCHome()
+		if err != nil {
+			fmt.Println(err)
+		}
+
+	})
+
+	s.AfterSuite(func() {
+		err := DeleteCRC()
+		if err != nil {
+			fmt.Println(err)
+		}
+	})
+
+	s.BeforeFeature(func(this *gherkin.Feature) {
+
+		if _, err := os.Stat(bundleName); os.IsNotExist(err) {
+			// Obtain the bundle to current dir
+			fmt.Println("Obtaining bundle...")
+			bundle, err := DownloadBundle(bundleURL, ".")
+			if err != nil {
+				fmt.Errorf("Failed to obtain CRC bundle, %v\n", err)
+			}
+			fmt.Println("Using bundle:", bundle)
+		} else if err != nil {
+			fmt.Errorf("Unknown error obtaining the bundle %v.\n", bundleName)
+		} else {
+			fmt.Println("Using existing bundle:", bundleName)
+		}
+
+	})
+}
+
+func DeleteFileFromCRCHome(fileName string) error {
+
+	theFile := filepath.Join(CRCHome, fileName)
+
+	if _, err := os.Stat(theFile); os.IsNotExist(err) {
+		return nil
+	}
+
+	err := clicumber.DeleteFile(theFile)
+	if err != nil {
+		fmt.Errorf("Error deleting file %v", theFile)
+	}
+	return nil
+}
+
+func FileExistsInCRCHome(fileName string) error {
+
+	theFile := filepath.Join(CRCHome, fileName)
+
+	_, err := os.Stat(theFile)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("file %s does not exists, error: %v ", theFile, err)
+	}
+
+	return err
+}
+
+func ConfigFileInCRCHomeContainsKeyMatchingValue(format string, configFile string, condition string, keyPath string, expectedValue string) error {
+
+	configPath := filepath.Join(CRCHome, configFile)
+
+	config, err := clicumber.GetFileContent(configPath)
 	if err != nil {
 		return err
 	}
 
-	actual := sh.GetLastCmdOutput("stdout")
-	if actual != expected {
-		return fmt.Errorf("CRC state does not match %s. Expected: %s, Actual: %s.\n", expected, expected, actual)
+	keyValue, err := clicumber.GetConfigKeyValue([]byte(config), format, keyPath)
+	if err != nil {
+		return err
+	}
+
+	matches, err := clicumber.PerformRegexMatch(expectedValue, keyValue)
+	if err != nil {
+		return err
+	} else if (condition == "contains") && !matches {
+		return fmt.Errorf("For key '%s' config contains unexpected value '%s'", keyPath, keyValue)
+	} else if (condition == "does not contain") && matches {
+		return fmt.Errorf("For key '%s' config contains value '%s', which it should not contain", keyPath, keyValue)
 	}
 
 	return nil
 }
 
-// FeatureContext defines godog.Suite steps for the test suite.
-func FeatureContext(s *godog.Suite) {
-	// Status verification
-	s.Step(`^CRC (should be|is) in state "([^"]*)"$`,
-		crcStateIs)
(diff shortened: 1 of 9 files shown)


jsliacan:
@gbraad the only reason is that `testsuite` is a package name that we later might want to have in `crc` itself when tests are consolidated. Plus seeing `clicumber` in code tells me where a function/var comes from without looking at imports. But if it's bad practice, I'll happily change it to default `testsuite`!

gbraad:
@jsliacan We usually change it when we have a conflict names... but clarification, to read an improved readability, can also be a reason. I would wonder why we would have a `testsuite` inside CRC, but understandable now... thanks.

jsliacan:
Sorry, I wasn't clear! I didn't mean to have the clicumber's `testsuite` inside CRC, but a test package called `testsuite` inside CRC. Right now I called it `crcsuite`. But name `crcsuite` doesn't suggest that it's a _test_ package, which is not great. Having a `testsuite` instead of `crcsuite` would probably be better. But I wasn't sure given that clicumber has a `testsuite`... It was more or less an arbitrary choice in the end.

gbraad:
OK, no worries

test/integration/crcsuite/crcsuite.go
```diff
@@ -25,33 +25,149 @@ package crcsuite
...
+	clicumber "github.com/code-ready/clicumber/testsuite"
 )
 
-var sh testsuite.ShellInstance
+// FeatureContext defines godog.Suite steps for the test suite.
+func FeatureContext(s *godog.Suite) {
+
+	// CRC related steps
+	s.Step(`^removing CRC home directory succeeds$`,
+		CRCRemoveHome)
```
gbraad: Can we not use CRC in every step? This is unnecessary as all steps are related to CRC, as this is the Application Under Test.

test/integration/crcsuite/crcsuite.go
```diff
@@ -25,33 +25,149 @@ package crcsuite
...
+
+	// CRC file operations
+	s.Step(`^CRC file "([^"]*)" exists$`,
```
gbraad: What is a 'CRC file' ? from the test step it is not clear what it is... but it sounds like a special type of file...

or is it just a file existence check fro the CRC home folder?

test/integration/crcsuite/crcsuite.go
jsliacan: Good point. I'll update it. Although in this particular step we don't want to remove `$HOME` but `$HOME/.crc`, so I'd probably keep the CRC bit here. Or?

test/integration/crcsuite/crcsuite.go
jsliacan: > or is it just a file existence check fro the CRC home folder?
yes, the corresponding function is analogous to the FileExists from clicumber but prepends the location with `/path/to/.crc/`. Relative paths aren't processed well for some reason and I thought they look quite bad :) , e.g. `../../../../.crc` (in this case). If there's a better way to do it I'll change it right away.

test/integration/crcsuite/crcsuite.go
jsliacan: @gbraad I removed 'CRC' from `CRCDelete` (it deletes CRC instance; used in e.g. BeforeFreature/AfterFeature) and from `CRCRemoveHome` (it removes `$HOME/.crc`; used in e.g. BeforeFeature)

I kept it in `CRCConfigFileContainsKey` - analogue of clicumber's `ConfigFileContainsKey` but for files placed inside `$HOME/.crc`. Same goes for `CRCFileExists` and `CRCConfigFileContainsKeyMatchingValue`.

test/integration/crcsuite/crcsuite.go
gbraad: The naming is weird, as we generally call it the `CRC home folder`, etc. and therefore CRC file does not mean much, but rather 'File exists in the CRC home folder'

jsliacan: @gbraad it's updated now. I also changed function names so they correspond to the steps.

test/integration/features/basic.feature
```diff
@@ -28,17 +30,14 @@ Checks whether CRC top-level commands behave correctly.
...
+    When executing "crc start -b ~/Downloads/crc_libvirt_4.1.0-rc.5.tar.xz" succeeds
+    Then stdout should contain "Running"
     
-  Scenario: CRC stop
-    When executing "crc stop" succeeds
-    Then stdout should contain """Stopping "crc" """
-    And stdout should contain "true"
-    And stdout should not contain "VM Failed to gracefully shutdown, try the kill command"
+  Scenario: CRC kill
+    When executing "crc stop -f"
```
anjannath: :+1:

anjannath:
Console logs from test run in CI: https://ci.centos.org/job/codeready-crc-pr/121/console

test/integration/crcsuite/crcsuite.go
```diff
@@ -25,33 +25,176 @@ package crcsuite
...
 		return err
 	}
+	theFile := filepath.Join(usr.HomeDir, ".crc", fileName)
+
+	_, err = os.Stat(theFile)
+
+	if os.IsNotExist(err) {
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("file %s neither exists nor doesn't exist, error: %v", theFile, err)
```
praveenkumar: what does that even mean? `file <> neither exists nor doesn't exist` ?

test/integration/crcsuite/crcsuite.go
```diff
@@ -25,33 +25,176 @@ package crcsuite
...
+		return err
+	}
+	theFile := filepath.Join(usr.HomeDir, ".crc", fileName)
+
+	_, err = os.Stat(theFile)
+
+	if os.IsNotExist(err) {
+		return fmt.Errorf("file %s does not exists, error: %v ", theFile, err)
+	} else if err != nil {
+		return fmt.Errorf("file %s neither exists nor doesn't exist, error: %v", theFile, err)
```
praveenkumar: same here ?

praveenkumar:
@jsliacan Can you rebase this PR and push again?

test/integration/crcsuite/crcsuite.go
jsliacan: @praveenkumar It was a placeholder for an error message that I was supposed to think through later - we don't know what the error is, so the message has to be meaningless. But I agree it can be less annoying than that :). What should I change it to? And thanks for catching this!

anjannath:
@jsliacan So the test failed due to missing `out/` dir, which is created only when we cross compile, so can you please add `make cross` to line 104 in the `centos_ci.sh` script.

```diff
diff --git a/centos_ci.sh b/centos_ci.sh
index 1cc8e97..8e49f7a 100644
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -101,6 +101,7 @@ get_bundle
 
 # setup and run integration tests
 make
+make cross
 crc setup
 make integration
```

test/integration/features/basic.feature
```diff
@@ -13,12 +13,14 @@ Checks whether CRC top-level commands behave correctly.
...
-    """Use "crc [command] --help" for more information about a command.
+    """
+    Use "crc [command] --help" for more information about a command.
     """
 
   Scenario: CRC setup
     When executing "crc setup" succeeds
-    Then stdout should contain "Starting Libvirt crc network"
+    Then stdout should contain "Caching oc binary"
+    And stdout should contain "Starting Libvirt crc network"
```
praveenkumar: This complete scenario will only going to work with linux, what we should do it to make our test platform independent from starting since in long run we will not struggle to patch it later. For me `crc setup` is a command and it can finish with an exist code. Adding stdout just to see the test string doesn't make much sense to me.

@anjannath @gbraad wdyt?

test/integration/features/basic.feature
```diff
@@ -28,17 +30,14 @@ Checks whether CRC top-level commands behave correctly.
     And stdout should contain "Installing crc-driver-libvirt"
     And stdout should contain "Setting up Libvirt crc network"
     And stdout should contain "Starting Libvirt crc network"
-    
+
   Scenario: CRC start
-    When executing "crc start -b ~/Downloads/crc_libvirt_v4.1.0.rc0.tar.xz" succeeds
-    Then stdout should contain "Creating VM"
-    And stdout should contain "Running"
+    When executing "crc start -b ~/Downloads/crc_libvirt_4.1.0-rc.5.tar.xz" succeeds
```
praveenkumar: Instead of that always download it on the current crc binary directory and use `crc start -b crc_libvirt_4.1.0-rc.5.tar.xz` but again this will be differ on the host. so you might have to put a logic around it.

test/integration/features/config.feature
```diff
@@ -0,0 +1,90 @@
...
+            | crc config set skip-check-libvirt-running true       | crc config set skip-check-libvirt-running false       | skip-check-libvirt-running       |
+            | crc config set skip-check-crc-network-active true    | crc config set skip-check-crc-network-active false    | skip-check-crc-network-active    |
+            | crc config set skip-check-libvirt-installed true     | crc config set skip-check-libvirt-installed false     | skip-check-libvirt-installed     |
+            | crc config set skip-check-libvirt-driver true        | crc config set skip-check-libvirt-driver false        | skip-check-libvirt-driver        |
+            | crc config set skip-check-crc-network true           | crc config set skip-check-crc-network false           | skip-check-crc-network           |
+            | crc config set skip-check-crc-dnsmasq-file true      | crc config set skip-check-crc-dnsmasq-file false      | skip-check-crc-dnsmasq-file      |
+            | crc config set skip-check-user-in-libvirt-group true | crc config set skip-check-user-in-libvirt-group false | skip-check-user-in-libvirt-group |
```
praveenkumar: Again these all will work with linux so use a build tag for different operating system.

test/integration/crcsuite/crcsuite.go
praveenkumar: I don't think you need that `else` part at all, you need to catch if file is exist or not.

```golang
if _, err := os.Stat("/path/to/whatever"); os.IsNotExist(err) {
	// path/to/whatever does not exist
}
```

praveenkumar: I don't think you need that `else` part at all, you need to catch if file is exist or not.

```golang
if _, err := os.Stat("/path/to/whatever"); os.IsNotExist(err) {
	// path/to/whatever does not exist
}
```

tsedmik:
@praveenkumar @gbraad could you please look at this issue. It is ready for merge and we (me and @jsliacan) are unable to merge it (lack of permissions)

praveenkumar:
@jsliacan This need to be rebased after #239 in once, we are planning to merge it today and then you should rebase it and do another push.

praveenkumar:
retest this please.

praveenkumar:
From the CentOS CI side below doesn't look good and need to be fixed.
```
16:01:14 + make integration
16:01:14 go test --timeout=60m github.com/code-ready/crc/test/integration -v --tags=integration --godog.tags=linux 
16:01:15 panic: runtime error: index out of range
16:01:15 
16:01:15 goroutine 1 [running]:
16:01:15 github.com/code-ready/crc/test/integration/crcsuite.ParseFlags()
16:01:15 	/root/payload/test/integration/crcsuite/prepare.go:64 +0xb0
16:01:15 github.com/code-ready/crc/test/integration.parseFlags()
16:01:15 	/root/payload/test/integration/integration_test.go:62 +0x25
16:01:15 github.com/code-ready/crc/test/integration.TestMain(0xc0000e3f10)
16:01:15 	/root/payload/test/integration/integration_test.go:33 +0x34
16:01:15 main.main()
16:01:15 	_testmain.go:38 +0x10e
16:01:15 FAIL	github.com/code-ready/crc/test/integration	0.032s
```

test/integration/crcsuite/prepare.go
```diff
@@ -0,0 +1,73 @@
...
+	"os/user"
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/download"
+)
+
+// Download bundle for testing
+func DownloadBundle(bundleLocation string, bundleDestination string) (string, error) {
+
+	if bundleLocation[:4] != "http" {
```
praveenkumar: This I don't like but we need to expand the use of `Download` function to support `file` URI also, which is by default is not present.

jsliacan: I agree it's not nice. It's a temp solution as you write. I made a note.

PR #128, opened 2019-05-14
Capture machine logs to crc log file

praveenkumar:
- Update the logic of getting the machine logs to stdout also.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -10,8 +10,6 @@ import (
 	log "github.com/code-ready/crc/pkg/crc/logging"
 )
 
-var logLevel string
-
 var rootCmd = &cobra.Command{
 	Use:   commandName,
 	Short: descriptionShort,
@@ -37,14 +35,14 @@ func init() {
 	config.InitViper()
 	setConfigDefaults()
 	rootCmd.AddCommand(cmdConfig.ConfigCmd)
-	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "log level (e.g. \"debug | info | warn | error\")")
+	rootCmd.PersistentFlags().StringVar(&log.LogLevel, "log-level", constants.DefaultLogLevel, "log level (e.g. \"debug | info | warn | error\")")
 }
 
 func runPrerun() {
 	output.OutF("%s - %s\n", commandName, descriptionShort)
 
 	// Setting up logrus
-	log.InitLogrus(logLevel)
+	log.InitLogrus(log.LogLevel)
 }
 
 func runPostrun() {
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -21,15 +21,15 @@ const (
 	DefaultHostname      = "crc-jtskh-master-0"
 	DefaultWebConsoleURL = "https://console-openshift-console.apps.tt.testing"
 	DefaultDiskImage     = "crc.disk"
+	DefaultLogLevel      = "info"
+	ConfigFile           = "crc.json"
+	LogFile              = "crc.log"
 )
 
 var (
-	CrcBaseDir  = filepath.Join(GetHomeDir(), ".crc")
-	ConfigFile  = "crc.json"
-	ConfigPath  = filepath.Join(CrcBaseDir, ConfigFile)
-	LogFile     = "crc.log"
-	LogFilePath = filepath.Join(CrcBaseDir, LogFile)
-
+	CrcBaseDir      = filepath.Join(GetHomeDir(), ".crc")
+	ConfigPath      = filepath.Join(CrcBaseDir, ConfigFile)
+	LogFilePath     = filepath.Join(CrcBaseDir, LogFile)
 	MachineBaseDir  = CrcBaseDir
 	MachineCertsDir = filepath.Join(MachineBaseDir, "certs")
 	MachineCacheDir = filepath.Join(MachineBaseDir, "cache")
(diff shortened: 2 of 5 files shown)


pkg/crc/machine/machine.go
```diff
@@ -212,11 +217,23 @@ func getDriverOptions(machineConfig config.MachineConfig) interface{} {
...
+		log.SetDebug(true)
+		logging.CloseLogFile()
+		logfile, err := logging.OpenLogfile()
+		if err != nil {
+			return err
+		}
+		log.SetOutWriter(logfile)
+		log.SetErrWriter(logfile)
 	} else {
 		log.SetDebug(true)
```
anjannath: @praveenkumar So if `--log-level=debug` we'll only log to console and not to the log file?

pkg/crc/machine/machine.go
```diff
@@ -212,11 +217,23 @@ func getDriverOptions(machineConfig config.MachineConfig) interface{} {
...
 	}
+	return nil
+}
+
+func unsetMachineLogging() {
+	logging.CloseLogFile()
+	logging.InitLogrus(constants.LogLevel)
```
anjannath: No need to initialize logrus again here..

pkg/crc/machine/machine.go
gbraad: what about the discard statements?

anjannath: We don't need them any more. before if libmachine logging was disabled, they were discarded, now it is logged to the log file `~/.crc/crc.log`, and if `--log-level debug` is passed on the command line, logs are written only to console.

pkg/crc/machine/machine.go
praveenkumar: If you don't initialize it again then after you close the logging file it will not have other log info after the start.

pkg/crc/machine/machine.go
praveenkumar: Yes so we don't need the discard since we are going to capture everything in the log file.

gbraad: OK :+1:

pkg/crc/machine/machine.go
anjannath: We initialize logrus in pre run of `root.go` so it should have..

gbraad: > again

I am wondering about this also. This sounds like a problem or else an easy way to mess things up.

cmd/crc/cmd/root.go
```diff
@@ -37,14 +35,14 @@ func init() {
 	config.InitViper()
 	setConfigDefaults()
 	rootCmd.AddCommand(cmdConfig.ConfigCmd)
-	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "log level (e.g. \"debug | info | warn | error\")")
+	rootCmd.PersistentFlags().StringVar(&constants.LogLevel, "log-level", "info", "log level (e.g. \"debug | info | warn | error\")")
```
gbraad: I think we should be clearer: `constants.DefaultLogLevel` would make more sense here.

gbraad: ... oh wait... you are setting the value here INTO the constant?!????

cmd/crc/cmd/root.go
```diff
@@ -37,14 +35,14 @@ func init() {
...
 }
 
 func runPrerun() {
 	output.OutF("%s - %s\n", commandName, descriptionShort)
 
 	// Setting up logrus
-	log.InitLogrus(logLevel)
+	log.InitLogrus(constants.LogLevel)
```
gbraad: ... this is used as a shared value. This is not right. 

The purpose of the machine dir as-is can be interpreted as a const, as during use of the applciation this value will NOT change. But the `LogLevel` gets changed from a default to the actual value. If so, this is not the same behaviour.

gbraad: Shouldn't such a value make more sense to live in `logging.LogLevel` and use `constants.DefaultLogLevel` for an initial value in the `cmd` ?

cmd/crc/cmd/root.go
gbraad:
```
rootCmd.PersistentFlags().StringVar(&logging.LogLevel, "log-level", constants.DefaultLogLevel, "log level (e.g. \"debug | info | warn | error\")")
```

praveenkumar:
@gbraad updated, have a look now.

cmd/crc/cmd/root.go
```diff
@@ -37,14 +35,14 @@ func init() {
 	config.InitViper()
 	setConfigDefaults()
 	rootCmd.AddCommand(cmdConfig.ConfigCmd)
-	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "log level (e.g. \"debug | info | warn | error\")")
+	rootCmd.PersistentFlags().StringVar(&log.LogLevel, "log-level", constants.DefaultLogLevel, "log level (e.g. \"debug | info | warn | error\")")
```
gbraad: :+1:

pkg/crc/constants/constants.go
```diff
@@ -21,15 +21,15 @@ const (
 	DefaultHostname      = "crc-jtskh-master-0"
 	DefaultWebConsoleURL = "https://console-openshift-console.apps.tt.testing"
 	DefaultDiskImage     = "crc.disk"
+	DefaultLogLevel      = "info"
+	ConfigFile           = "crc.json"
+	LogFile              = "crc.log"
```
gbraad: :=1

pkg/crc/logging/logging.go
```diff
@@ -12,7 +12,10 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
-var logfile *os.File
+var (
+	logfile  *os.File
+	LogLevel string
```
gbraad: an `iota` in future would be nicer

PR #135, opened 2019-05-15
Issue #134 Add BindFlag func for binding flags to config properties

### LINKED ISSUE #134: Bind config properties with start flags
As of now, viper config properties such as `vm-driver` is not linked to flag `--vm-driver`

anjannath:
Fixes #134

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -11,13 +11,14 @@ import (
 	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/preflight"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 func init() {
 	rootCmd.AddCommand(startCmd)
+	startCmd.Flags().AddFlagSet(startCmdFlagSet)
 
-	startCmd.Flags().StringVarP(&bundlePath, "Bundle", "b", constants.DefaultBundle, "The system bundle used for deployment of the OpenShift cluster.")
-	startCmd.Flags().StringP(config.VMDriver.Name, "d", constants.DefaultVMDriver, fmt.Sprintf("The driver to use for the CRC VM. Possible values: %v", constants.SupportedVMDrivers))
+	crcConfig.BindFlagSet(startCmd.Flags())
 }
 
 var startCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


cmd/crc/cmd/start.go
```diff
@@ -18,6 +18,7 @@ func init() {
 
 	startCmd.Flags().StringVarP(&bundlePath, "Bundle", "b", constants.DefaultBundle, "The system bundle used for deployment of the OpenShift cluster.")
 	startCmd.Flags().StringP(config.VMDriver.Name, "d", constants.DefaultVMDriver, fmt.Sprintf("The driver to use for the CRC VM. Possible values: %v", constants.SupportedVMDrivers))
+	crcConfig.BindFlag(config.VMDriver.Name, startCmd.Flags().Lookup("vm-driver"))
```
praveenkumar: This means we are going to do that for each flag which we take from the config side, it is not scaleable, I would like to have it like we had in minishift.

gbraad: Agree

gbraad: However, this is simplistic for now and works... right?

anjannath: We do not have any other flags now, so i didn't create a wrapper around the `BindPFlags` where we can pass in a flagset!

cmd/crc/cmd/start.go
```diff
@@ -55,3 +57,11 @@ func runStart(arguments []string) {
 		logging.ErrorF(err.Error())
 	}
 }
+
+func initStartCmdFlagSet() *pflag.FlagSet {
+	flagSet := pflag.NewFlagSet("start", pflag.ExitOnError)
+	flagSet.StringP(config.Bundle.Name, "b", constants.DefaultBundle, "The system bundle used for deployment of the OpenShift cluster.")
+	flagSet.StringP(config.VMDriver.Name, "d", constants.DefaultVMDriver, fmt.Sprintf("The driver to use for the CRC VM. Possible values: %v", constants.SupportedVMDrivers))
+
+	return flagSet
```
praveenkumar: 👍

pkg/crc/config/viper_config.go
```diff
@@ -81,3 +81,13 @@ func IsSet(key string) bool {
...
+
+// BindFlags binds flags to config properties
+func BindFlag(key string, flag *pflag.Flag) error {
+	return globalViper.BindPFlag(key, flag)
+}
+
+// BindFlagset binds a flagset to their repective config properties
+func BindFlagSet(flagSet *pflag.FlagSet) error {
+	return globalViper.BindPFlags(flagSet)
+}
```
praveenkumar: 👍

Issue #139, opened 2019-05-17
Allow memory and CPU from the config and also as part of start flags.


praveenkumar:
We need to allow cpu and memory from the config and also as flag options, also need to put a validation function for same.

jmazzitelli:
I mention this over in https://github.com/code-ready/crc/issues/127#issuecomment-493429805 but since memory and CPU instructions are more relevant here, I'll document here as well for completeness (see that link if you want to know how to increase disk space).

Until this issue is implemented, you can increase memory and CPU for the CRC VM using these instructions (these instructions assume the CRC VM is currently running).

**For MEMORY:**

```
# increase memory to 16GB
virsh -c qemu:///system setmaxmem crc 16000000 --config
virsh -c qemu:///system setmem crc 16000000 --config
crc stop
crc start
```

**For CPUS:**

```
# increase virtual CPUs to 5
virsh -c qemu:///system setvcpus crc 5 --maximum --config
virsh -c qemu:///system setvcpus crc 5 --config
crc stop
crc start
```

robin-owen:
For the memory, can't we use shortened values for the maximum? As an example, `16G` as opposed to `16000000`?

jmazzitelli:
> For the memory, can't we use shortened values for the maximum? As an example, `16G` as opposed to `16000000`

Probably, I didn't try it.

robin-owen:
These commands worked for me:

```
$ sudo virsh setmaxmem crc 16G --config
$ sudo virsh setmem crc 16G --config
$ crc stop && crc start
```

anjannath:
Fixed in #143

PR #143, opened 2019-05-20
Issue #139 Add memory and cpu options to config

praveenkumar:


cmd/crc/cmd/start.go
```diff
@@ -36,7 +37,7 @@ var (
 )
 
 func runStart(arguments []string) {
-	if err := constants.ValidateDriver(crcConfig.GetString(config.VMDriver.Name)); err != nil {
+	if err := validateStartFlags(); err != nil {
```
anjannath: :+1:

cmd/crc/cmd/config/config.go
```diff
@@ -22,6 +23,8 @@ var (
 	// Start command settings in config
 	VMDriver = createSetting("vm-driver", nil, []validationFnType{validations.ValidateDriver})
 	Bundle   = createSetting("bundle", nil, nil)
+	CPUs     = createSetting("cpus", constants.DefaultCPUs, []validationFnType{validations.ValidateCPUs})
+	Memory   = createSetting("memory", nil, []validationFnType{validations.ValidateMemory})
```
gbraad: We have a default for memory...

we currently take the default from the bundle...  but that would be the same for the CPUs, so why different here?

cmd/crc/cmd/start.go
```diff
@@ -46,8 +47,8 @@ func runStart(arguments []string) {
 		Name:       constants.DefaultName,
 		BundlePath: crcConfig.GetString(config.Bundle.Name),
 		VMDriver:   crcConfig.GetString(config.VMDriver.Name),
-		Memory:     constants.DefaultMemory,
```
gbraad: DefaultMemory has been defined, but not used anymore?

pkg/crc/validation/validation.go
```diff
@@ -0,0 +1,32 @@
...
+func ValidateCPUs(value int) error {
+	if value < constants.DefaultCPUs {
+		return fmt.Errorf("CPUs required >=%d", constants.DefaultCPUs)
+	}
+	return nil
+}
+
+// ValidateMemory is check if provided Memory count is valid
+func ValidateMemory(value int) error {
+	if value < constants.DefaultMemory {
```
gbraad: Used in the validation, but not as default...

cmd/crc/cmd/config/config.go
```diff
@@ -20,8 +21,10 @@ var SettingsList = make(map[string]*setting)
 
 var (
 	// Start command settings in config
-	VMDriver = createSetting("vm-driver", nil, []validationFnType{validations.ValidateDriver})
-	Bundle   = createSetting("bundle", nil, nil)
+	VMDriver = createSetting("vm-driver", constants.DefaultVMDriver, []validationFnType{validations.ValidateDriver})
+	Bundle   = createSetting("bundle", nil, []validationFnType{validations.ValidateBundle})
+	CPUs     = createSetting("cpus", constants.DefaultCPUs, []validationFnType{validations.ValidateCPUs})
+	Memory   = createSetting("memory", constants.DefaultMemory, []validationFnType{validations.ValidateMemory})
```
gbraad: :+1:

PR #163, opened 2019-05-24
Issue #110 Prevent segfault because of calling err.Error() on empty err

### LINKED ISSUE #110: Pre-flight check fails when checking for group membership of the user
Tried `crc start` without the bundle downloaded yet and got a panic error:

```
jeffmaury@localhost crc-0.85.0-linux-amd64]$ ./crc start
INFO  Checking if Virtualization is enabled       
INFO  Checking if KVM is enabled                  
INFO  Checking if Libvirt is installed            
INFO  Checking if user is part of libvirt group   
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x884e68]

goroutine 1 [running]:
github.com/code-ready/crc/pkg/crc/preflight.preflightCheckSucceedsOrFails(0x9a6300, 0x9bb8f8, 0x9ab201, 0x29, 0x0)
	/home/prkumar/work/github/crc/pkg/crc/preflight/preflight_common.go:26 +0x188
github.com/code-ready/crc/pkg/crc/preflight.StartPreflightChecks()
	/home/prkumar/work/github/crc/pkg/crc/preflight/preflight_linux.go:25 +0x201
github.com/code-ready/crc/cmd/crc/cmd.runStart(0xddadb8, 0x0, 0x0)
	/home/prkumar/work/github/crc/cmd/crc/cmd/start.go:41 +0xce
github.com/code-ready/crc/cmd/crc/cmd.glob..func6(0xdb3b00, 0xddadb8, 0x0, 0x0)
	/home/prkumar/work/github/crc/cmd/crc/cmd/start.go:25 +0x3f
github.com/spf13/cobra.(*Command).execute(0xdb3b00, 0xddadb8, 0x0, 0x0, 0xdb3b00, 0xddadb8)
	/home/prkumar/work/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:766 +0x2cc
github.com/spf13/cobra.(*Command).ExecuteC(0xdb3640, 0xdb3fc0, 0xc00017df68, 0xc00017df88)
	/home/prkumar/work/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:852 +0x2fd
github.com/spf13/cobra.(*Command).Execute(0xdb3640, 0x1, 0xdb3fc0)
	/home/prkumar/work/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:800 +0x2b
github.com/code-ready/crc/cmd/crc/cmd.Execute()
	/home/prkumar/work/github/crc/cmd/crc/cmd/root.go:59 +0x2d
main.main()
	/home/prkumar/work/github/crc/cmd/crc/main.go:8 +0x20
[jeffmaury@localhost crc-0.85.0-linux-amd64]$ 

```

### LINKED ISSUE #159: Handle the errors effectively to prevent segfault/panic
Looks like some of the errors are either we ignored or not handled correctly and that need to address properly.

- If a member is not part of `libvirt` group and user just use `crc start` before `crc setup` it will give panic.

- If user use `crc start` on already running instance that also give the panic.

anjannath:
Fixes #110
Fixes #159

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -103,15 +103,19 @@ func Start(startConfig StartConfig) (StartResult, error) {
 
 		result.Status = vmState.String()
 	} else {
-		logging.InfoF(" Starting stopped VM ...")
 		host, err := libMachineAPIClient.Load(machineConfig.Name)
 		s, err := host.Driver.GetState()
 		if err != nil {
 			logging.ErrorF("Error getting the state for host: %v", err)
 			result.Error = err.Error()
 		}
+		if s == state.Running {
+			result.Status = s.String()
+			return *result, nil
+		}
 
 		if s != state.Running {
+			logging.InfoF(" Starting stopped VM ...")
 			if err := host.Driver.Start(); err != nil {
 				logging.ErrorF("Error starting stopped VM: %v", err)
 				result.Error = err.Error()
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -2,7 +2,6 @@ package preflight
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -16,6 +15,7 @@ import (
 	"text/template"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine/libvirt"
 	"github.com/code-ready/crc/pkg/crc/oc"
@@ -172,10 +172,10 @@ func checkUserPartOfLibvirtGroup() (bool, error) {
 		return false, fmt.Errorf("%+v : %s", err, stdErr)
 	}
 	if strings.Contains(stdOut, "libvirt") {
-	        logging.Debug("Current user is already in the libvirt group")
+		logging.Debug("Current user is already in the libvirt group")
 		return true, nil
 	}
-	return false, err
+	return false, errors.NewF("%s not part of libvirtd group", currentUser.Username)
 }
 
 func fixUserPartOfLibvirtGroup() (bool, error) {


pkg/crc/machine/machine.go
```diff
@@ -93,15 +93,19 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 		result.Status = vmState.String()
 	} else {
-		logging.InfoF(" Starting stopped VM ...")
 		host, err := libMachineAPIClient.Load(machineConfig.Name)
 		s, err := host.Driver.GetState()
 		if err != nil {
 			logging.ErrorF("Error getting the state for host: %s", err)
 			result.Error = err.Error()
 		}
+		if s == state.Running {
```
anjannath: If we have this check early on in the start command, then we can prevent performing the preflight checks. I am for putting it there.. but since all the machine stuff is here added it here, would like to know what @gbraad @praveenkumar thinks!

pkg/crc/preflight/preflight_common.go
```diff
@@ -23,7 +23,7 @@ func preflightCheckSucceedsOrFails(configuredToSkip bool, check PreflightCheckFi
 		return
 	}
 
-	log.Fatal(err.Error())
+	log.FatalF(" Above Check Failed: %+v", err)
```
praveenkumar: :+1:

pkg/crc/machine/machine.go
praveenkumar: I don't know, this is like early feedback that machine already running ?

anjannath: Yes, i am for doing this check in the start command.. even before calling `machine.Start()` and starting the preflight checks..

gbraad: This is a different issue... valuable, but no need to address here

pkg/crc/machine/machine.go
```diff
@@ -93,15 +93,19 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+			result.Status = s.String()
+			return *result, err
```
gbraad: what valuse has `err` here?

This is to shortcircuit the `start` when the VM is already in a running state

anjannath: Any error that might have occurred during getting the state of the VM :P 
Yup does not make sense.. if we already get into the if block.. will just return `nil` instead!

gbraad: In that case line 102 should be a return of the error... right?
as originally it just continues and would check the running state and try again, but if the previous call failed, that would fail also.

same actually here. it misses the return of the error

anjannath: @gbraad Updated return value to `nil`.

pkg/crc/preflight/preflight_common.go
praveenkumar: So our error should be correctly returned I think current solution doesn't going well in long term so better fix the function which return that error as `nil` even result is false.

praveenkumar:
here https://github.com/code-ready/crc/blob/master/pkg/crc/preflight/preflight_checks_linux.go#L161 this should have been.

```
return false, errors.NewF("%s not part of libvirtd group", currentUser.Username)
```

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -172,10 +172,10 @@ func checkUserPartOfLibvirtGroup() (bool, error) {
 		return false, fmt.Errorf("%+v : %s", err, stdErr)
 	}
 	if strings.Contains(stdOut, "libvirt") {
-	        logging.Debug("Current user is already in the libvirt group")
+		logging.Debug("Current user is already in the libvirt group")
 		return true, nil
 	}
-	return false, err
+	return false, errors.NewF("%s not part of libvirtd group", currentUser.Username)
```
praveenkumar: 👍

PR #164, opened 2019-05-24
Issue 162 Add user interaction for stop command

praveenkumar:


anjannath:
So this means in ci we should always use `crc stop -f`

cmd/crc/cmd/stop.go
```diff
@@ -22,15 +26,46 @@ var stopCmd = &cobra.Command{
...
+func runStop(arguments []string) {
 	stopConfig := machine.StopConfig{
 		Name: constants.DefaultName,
 	}
 
+	killConfig := machine.KillConfig{
+		Name: constants.DefaultName,
+	}
+
+	if force {
```
anjannath: :+1:

cmd/crc/cmd/stop.go
```diff
@@ -22,15 +26,46 @@ var stopCmd = &cobra.Command{
...
+		killVM(killConfig)
+		errors.Exit(0)
+	}
+
 	commandResult, err := machine.Stop(stopConfig)
+	if err != nil {
+		// Here we are catching the error "VM Failed to gracefully shutdown, try the kill command"
+		// from the driver side and if it exist then we ask user input for power off operation.
+		if strings.Contains(err.Error(), "kill") {
```
anjannath: :+1:

cmd/crc/cmd/stop.go
```diff
@@ -22,15 +26,46 @@ var stopCmd = &cobra.Command{
...
+				errors.Exit(0)
+			}
+		}
+		errors.ExitWithMessage(1, err.Error())
+	}
+	output.Out(commandResult.Success)
+}
+
+func killVM(killConfig machine.KillConfig) {
+	commandResult, err := machine.Kill(killConfig)
```
gbraad: Isn't this a `PowerOff` ? Kill sounds more like a `kill -9` of the VM process.

pkg/crc/machine/machine.go
```diff
@@ -152,6 +152,28 @@ func Stop(stopConfig StopConfig) (StopResult, error) {
...
+	libMachineAPIClient := libmachine.NewClient(constants.MachineBaseDir, constants.MachineCertsDir)
+	host, err := libMachineAPIClient.Load(killConfig.Name)
+
+	if err != nil {
+		result.Success = false
+		result.Error = err.Error()
+		return *result, err
+	}
+
+	if err := host.Kill(); err != nil {
```
gbraad: this naming is not given by us, so I do not care much.

gbraad:
it is not really 'user interaction' but introducing a 'force' option. title/description does not match here with expectation

cmd/crc/cmd/stop.go
gbraad: do we need to do this? the result is a `false`. can't we use this instead?
t he state is running and returned false on stop.,,

besides, this error message needs to be modified in the driver, which means tis check is fragile.

cmd/crc/cmd/stop.go
jsliacan: Also, if the user chooses `n`, then it defaults to the message `[VM Failed to gracefully shutdown, try the kill command]` even though in the meantime the VM did shut down gracefully.

cmd/crc/cmd/stop.go
praveenkumar: Problem is the state, if your VM is in `stop` or `delete` state then error will be returned and user have to prompted for it which doesn't make sense :(

cmd/crc/cmd/stop.go
praveenkumar: That is what driver is returning `VM Failed to gracefully shutdown, try the kill command`, I am still think that just have `-f` for stop no user interaction at all, we will increase the timelimit around `120sec` on driver side.

praveenkumar:
@gbraad updated as per your comments.

anjannath:
It'd bee good to have this in as it unblocks #121

cmd/crc/cmd/stop.go
gbraad: I saw you made those changes

cmd/crc/cmd/stop.go
gbraad: This got changed by additionally checking for state == running.

Issue #165, opened 2019-05-24
Add capability to download the latest oc binary to cache


praveenkumar:
We need to add the download capability for oc binary which should be the latest one and can be downloaded from http://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/ and can be added to cache folder, it will helpful for us to perform some of the post cluster operations which we require like making sure all the operator are up and serving correctly.

anjannath:
Closed via #166

PR #166, opened 2019-05-27
Issue #165 Cache oc binary

praveenkumar:
Cached the oc binary `~/.crc/cache/oc` location and part of preflight check.

### DIFF
--- a/pkg/crc/oc/oc_cache.go
+++ b/pkg/crc/oc/oc_cache.go
@@ -1,0 +1,170 @@
+package oc
+
+import (
+	"fmt"
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/download"
+	"github.com/code-ready/crc/pkg/extract"
+	"github.com/pkg/errors"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+const (
+	OC_CACHE_DIR = "oc"
+	TAR          = "tar.gz"
+	ZIP          = "zip"
+)
+
+// Oc is a struct with methods designed for dealing with the oc binary
+type OcCached struct{}
+
+func (oc *OcCached) EnsureIsCached() error {
+	if !oc.IsCached() {
+		err := oc.cacheOc()
+		if err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
+
+func (oc *OcCached) GetCacheFilepath() string {
+	return filepath.Join(constants.MachineCacheDir, OC_CACHE_DIR)
+}
+
+func (oc *OcCached) IsCached() bool {
+	if _, err := os.Stat(filepath.Join(oc.GetCacheFilepath(), constants.OcBinaryName)); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+// cacheOc downloads and caches the oc binary into the minishift directory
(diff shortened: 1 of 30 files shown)


pkg/crc/oc/oc_cache.go
```diff
@@ -0,0 +1,170 @@
+package oc
+
+import (
+	"fmt"
```
gbraad: is this file formatted? looks like it does not use tab, but a spaced indent

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -7,6 +7,11 @@ import (
 
 // StartPreflightChecks performs the preflight checks before starting the cluster
 func StartPreflightChecks() {
+	preflightCheckSucceedsOrFails(false,
+		checkOcBinaryCached,
+		"Checking if oc binary is cached",
+		false,
```
gbraad: Shouldn't this warn?

at the moment thjis is not an essential part of the deploymen.

praveenkumar: I am going to send a follow up PR which will make it essential, want to make sure to check the cluster operators are running and state is as expected.

pkg/crc/oc/oc_cache.go
praveenkumar: I did use `gofmt`, will check again.

praveenkumar: @gbraad I opened this file on vim and even goLand and both place I can see it as using tab not space, how did you open it?

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -16,6 +21,12 @@ func StartPreflightChecks() {
 
 // SetupHost performs the prerequisite checks and setups the host to run the cluster
 func SetupHost() {
+	preflightCheckAndFix(false,
```
jsliacan: Should we also have some `skip-check-oc-binary-cached` config property?

pkg/crc/oc/oc_cache.go
```diff
@@ -0,0 +1,170 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "Cannot make '%s' executable", finalBinaryPath)
+		}
+
+		return nil
+	}
+	return nil
+}
+
+func copy(src, dest string) error {
```
jsliacan: Why is caching not done straight into `/home/user/.crc/cache/oc` but it's split into two steps instead? (more of a question rather than a review comment...)

pkg/crc/oc/oc_cache.go
gbraad: github interface

pkg/crc/preflight/preflight_darwin.go
praveenkumar: @jsliacan no as I said this is required for post vm part so we are always going to need it.

pkg/crc/oc/oc_cache.go
praveenkumar: @jsliacan first step is to download in the temp location so in case it is no complete then we can start over and also there we extract the payload, second part is mostly just copy the extract binary to location.

gbraad: this is a common pattern. alternatively it could have used a suffix wlike `.part`. we need this to prevent incomplete downloads (at the moment we do support resuming downloads).

pkg/crc/preflight/preflight_darwin.go
gbraad: we can not skip this...

PR #167, opened 2019-05-28
Add some debug logs to preflight checks

cfergeau:
This adds some logs during preflight on macos

### DIFF
--- a/cmd/crc/cmd/config/view.go
+++ b/cmd/crc/cmd/config/view.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"github.com/code-ready/crc/pkg/crc/config"
-	log "github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/spf13/cobra"
 	"io"
 	"os"
@@ -37,10 +37,10 @@ to the options that you set when you run the 'crc start' command.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		tmpl, err := determineTemplate(configViewFormat)
 		if err != nil {
-			log.Fatal(err)
+			logging.Fatal(err)
 		}
 		if err := runConfigView(config.AllConfigs(), tmpl, os.Stdout); err != nil {
-			log.Fatal(err)
+			logging.Fatal(err)
 		}
 	},
 }
--- a/pkg/crc/oc/oc_cache.go
+++ b/pkg/crc/oc/oc_cache.go
@@ -49,6 +49,7 @@ func (oc *OcCached) IsCached() bool {
 // cacheOc downloads and caches the oc binary into the minishift directory
 func (oc *OcCached) cacheOc() error {
 	if !oc.IsCached() {
+		logging.Debug("Downloading oc")
 		// Create tmp dir to download the oc tarball
 		tmpDir, err := ioutil.TempDir("", "crc")
 		if err != nil {
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 
 	"github.com/code-ready/crc/pkg/crc/oc"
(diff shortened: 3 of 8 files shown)


pkg/crc/oc/oc_cache.go
```diff
@@ -13,6 +13,8 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	log "github.com/code-ready/crc/pkg/crc/logging"
```
gbraad: llet's be consistent... we either use `log` or `logging`... or if this conflicts, `crclog` as alias.

anjannath:
:+1:  Can we also add those logging for the Linux preflight checks too.. we need logging from those checks also.

cfergeau:
Oh I missed that the PR was approved, and added some more logs anyways. I did not fully test the logs though as my box is not fully setup yet, and I don't want to modify it.

gbraad:
I see:

* caa5f1e use stdlib functions to get user dirs
* 4d14b63 Revert "use stdlib functions to get user dirs"

in that case, wouldn't it be better to force push and remove this commit with a `rebase` ?

gbraad:
Commit titles are not consistent with our current pattern. they should use `Issue #[no] [description]`

cfergeau:
> in



> I see:
> 
>     * [caa5f1e](https://github.com/code-ready/crc/commit/caa5f1ea82e91cae372c615e09b813e22996b104) use stdlib functions to get user dirs
> 
>     * [4d14b63](https://github.com/code-ready/crc/commit/4d14b63eddce538bb23bb8eae629bc34d4012697) Revert "use stdlib functions to get user dirs"
> 
> 
> in that case, wouldn't it be better to force push and remove this commit with a `rebase` ?

My bad, they never were meant to be in this PR.

cfergeau:
> Commit titles are not consistent with our current pattern. they should use `Issue #[no] [description]`

I can change them, though in my opinion the issue # is not so important that it should be using up precious shortlog space. It fits better further down the log where you don't have any space limitations.

gbraad:
> precious shortlog space

I totally agree, but this was decided early on and was part of the modus operandi of Minishift. We haven't added the commit checks and I personally found them to work against getting contribs. However, we enforce the need to work on tasks with an issue, so a commit needs to reference it... this way the automation can kick in... BUT, this automation seems to only work on commit messages like `Fix #[no]` or `Fixes #[no]` ? @code-ready/crc-devel WDYT?

cfergeau:
> > precious shortlog space
> 
> I totally agree, but this was decided early on and was part of the modus operandi of Minishift. We haven't added the commit checks and I personally found them to work against getting contribs. However, we enforce the need to work on tasks with an issue, so a commit needs to reference it... this way the automation can kick in... BUT, this automation seems to only work on commit messages like `Fix #[no]` or `Fixes #[no]` ? @code-ready/crc-devel WDYT?

Fix/Fixes #[no] will automatically close the corresponding issues when the PR is pushed ( https://help.github.com/en/articles/closing-issues-using-keywords ) Dunno if it makes sense to have it on multiple commits? (really no strong feeling here, I'll adjust the full commit log to whatever is the right way)

cmd/crc/cmd/config/view.go
```diff
@@ -37,10 +37,10 @@ to the options that you set when you run the 'crc start' command.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		tmpl, err := determineTemplate(configViewFormat)
 		if err != nil {
-			log.Fatal(err)
+			logging.Fatal(err)
 		}
 		if err := runConfigView(config.AllConfigs(), tmpl, os.Stdout); err != nil {
-			log.Fatal(err)
+			logging.Fatal(err)
```
praveenkumar: 👍

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -78,6 +85,7 @@ func checkOcBinaryCached() (bool, error) {
 	if !oc.IsCached() {
 		return false, errors.New("oc binary is not cached.")
 	}
+	logging.Debug("oc binary already cached")
```
praveenkumar: 👍

PR #173, opened 2019-05-30
update readme now that alphas are available

finp:


robin-owen:
I see that there are two commits here, and that there's no issue number referenced in the commit message.

Would you mind combining these commits into a single commit and changing the commit message so that it reads "Issue #<issue_number> update README now that alphas are available"?

finp:
gotcha, closing

PR #186, opened 2019-06-05
Issue #144 Add clear-cache flag to crc delete command

### LINKED ISSUE #144: Handling cache cleanup
At the moment all the bundles are extracted in the `~/.crc/cache` folder, but aren't cleaned up. A single bundle extracted takes up over 8+ GB of storage. We would need to include a command to remove these files; interactively or forced? Perhaps even report about this with a status command (#68) ?

anjannath:
Fixes #144

### DIFF
--- a/cmd/crc/cmd/delete.go
+++ b/cmd/crc/cmd/delete.go
@@ -1,15 +1,20 @@
 package cmd
 
 import (
+	"fmt"
+	"os"
+
 	"github.com/spf13/cobra"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/output"
-
+	"github.com/code-ready/crc/pkg/crc/input"
 	"github.com/code-ready/crc/pkg/crc/machine"
+	"github.com/code-ready/crc/pkg/crc/output"
 )
 
 func init() {
+	deleteCmd.Flags().BoolVarP(&clearCache, "clear-cache", "", false,
+		fmt.Sprintf("Clear the cache directory: %s", constants.MachineCacheDir))
 	rootCmd.AddCommand(deleteCmd)
 }
 
@@ -22,15 +27,26 @@ var deleteCmd = &cobra.Command{
 	},
 }
 
-func runDelete(arguments []string) {
+var clearCache bool
 
+func runDelete(arguments []string) {
 	deleteConfig := machine.DeleteConfig{
 		Name: constants.DefaultName,
+	}
+	if clearCache {
+		deleteCache()
 	}
 
 	commandResult, err := machine.Delete(deleteConfig)
 	output.Out(commandResult.Success)
 	if err != nil {
 		output.Out(err.Error())
+	}
+}
+
+func deleteCache() {
+	yes := input.PromptUserForYesOrNo("Do you want to delete cache", globalForce)
+	if yes {
+		os.RemoveAll(constants.MachineCacheDir)
 	}
 }
--- a/cmd/crc/cmd/stop.go
+++ b/cmd/crc/cmd/stop.go
@@ -15,7 +15,6 @@ import (
 
 func init() {
 	rootCmd.AddCommand(stopCmd)
-	stopCmd.Flags().BoolVarP(&force, "force", "f", false, "Stop the VM forcefully")
 }
 
 var stopCmd = &cobra.Command{
@@ -26,10 +25,6 @@ var stopCmd = &cobra.Command{
 		runStop(args)
 	},
 }
-
-var (
-	force bool
-)
 
 func runStop(arguments []string) {
 	stopConfig := machine.StopConfig{
@@ -49,7 +44,7 @@ func runStop(arguments []string) {
 			// Most of the time force kill don't work and libvirt throw
 			// Device or resource busy error. To make sure we give some
 			// graceful time to cluster before kill it.
-			if force {
+			if globalForce {
 				killVM(killConfig)
 				errors.Exit(0)
 			}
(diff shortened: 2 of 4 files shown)


pkg/util/util.go
```diff
@@ -0,0 +1,17 @@
...
+
+func PromptUserForYesOrNo(message string) bool {
+	var response string
+	output.OutF(message + "? [Y/N]: ")
+	fmt.Scanf("%s", response)
+	if strings.ToLower(response) == "y" {
+		return true
+	}
+	return false
+}
```
praveenkumar: I don't want to have util package, in minishift we had so much util part that was very confusing and most of time created our cyclic dep issues :(

Not sure at this point what should we use. @gbraad wdyt?

gbraad: same here... I dislike the idea of a 'util  as we end up putting it in each namespace.

gbraad: it is better to call this `input` as we have an output, right?

cmd/crc/cmd/delete.go
```diff
@@ -22,15 +25,31 @@ var deleteCmd = &cobra.Command{
...
 	commandResult, err := machine.Delete(deleteConfig)
 	output.Out(commandResult.Success)
 	if err != nil {
 		output.Out(err.Error())
 	}
 }
+
+func deleteCache() {
+	if globalForce {
+		os.RemoveAll(constants.MachineCacheDir)
```
gbraad: can we have a function for this?

pkg/util/util.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"fmt"
+	"github.com/code-ready/crc/pkg/crc/output"
+	"strings"
+)
+
+func PromptUserForYesOrNo(message string) bool {
+	var response string
+	output.OutF(message + "? [Y/N]: ")
```
cfergeau: I would use "[y/N]" as the default behaviour is to consider the user answered 'no'.

gbraad: right, I agree. This was also the behaviour as in Minishift.

anjannath: Will use `[y/N]`

cmd/crc/cmd/delete.go
cfergeau: This could also be written as
```
if globalForce {
    yes := true
else {
    yes := input.PromptUserForYesOrNo("Do you want to delete cache")
}
if yes {
    os.RemoveAll(constants.MachineCacheDir)
}
```
Or even be factored into PromptUserForYesOrNo (with a 'force' argument), as the other user of this method has exactly the same requirements, assume yes if 'force' is true, ask if not.

gbraad:
Would it be possible to put the `input` code in a separate commit ? this way we could in future identify functionality easier, and rebase more targeted if needed.

cmd/crc/cmd/delete.go
```diff
@@ -2,14 +2,17 @@ package cmd
...
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/input"
 	"github.com/code-ready/crc/pkg/crc/output"
 
 	"github.com/code-ready/crc/pkg/crc/machine"
 )
 
 func init() {
+	deleteCmd.Flags().BoolVarP(&clearCache, "clear-cache", "", false, "Clear the cache directory: $HOME/.crc")
```
praveenkumar: As per message it should be `$HOME/.crc/cache` or `fmt.Sprintf("Clear the cache directory: %s", constants.MachineCacheDir)`

cfergeau:
> Would it be possible to put the `input` code in a separate commit ? this way we could in future identify functionality easier, and rebase more targeted if needed.

I'd probably also do the 'stop -f' -> 'global -f' change in a separate commit so that it stands out more that there was a slight change to our cmdline interface. The commit log could detail some more that it's backward compatible, ...

anjannath:
Split the input code and `gloabalForce` flag changes into two commits. And factored as suggested in https://github.com/code-ready/crc/pull/186#discussion_r294291337

cmd/crc/cmd/stop.go
```diff
@@ -49,14 +42,12 @@ func runStop(arguments []string) {
...
+			if globalForce {
 				killVM(killConfig)
 				errors.Exit(0)
 			}
-			var userInput string
-			output.OutF("Do you want to force power off [y/n]: ")
-			fmt.Scan(&userInput)
-			if strings.ToLower(userInput) == "y" {
+			yes := input.PromptUserForYesOrNo("Do you want to force power off")
+			if yes {
```
cfergeau: This hunk belongs to the previous commit

cmd/crc/cmd/stop.go
```diff
@@ -42,11 +42,7 @@ func runStop(arguments []string) {
 			// Most of the time force kill don't work and libvirt throw
 			// Device or resource busy error. To make sure we give some
 			// graceful time to cluster before kill it.
-			if globalForce {
-				killVM(killConfig)
-				errors.Exit(0)
-			}
-			yes := input.PromptUserForYesOrNo("Do you want to force power off")
+			yes := input.PromptUserForYesOrNo("Do you want to force power off", globalForce)
```
cfergeau: Might have been more natural to have a first commit introducing globalForce, second one introduces PromptUserForYesOrNo(), and 3rd one adds clear-cache.

anjannath: Okay, that makes sense, let me reorder those!

anjannath: Fixed up the commits, have a look at your leisure!

Issue #184, opened 2019-06-05
Post-Flight checks before kubelet start


gbraad:
As part of the VM bringup, we need to make sure the environment is able to run the OpenShift cluster. However, AFTER the VM bringup, we could still experience issues that are related to other resources, like network connectitivity, disk space, etc. These results can indicate and therefore stop the start of kubelet.

For the current timeframe we need to consider two tests; DNS connectivity:

  * is `crc.testing` and `apps-crc.testing` returning the expected result? If not, this is an `ERROR` and prevents further execution
  * is an external resource, like `...` available? If not, a `WARN` should be given but the kubelet could get started.

If `WARN` we will start the `kubelet` service. Any startup issues have to be caught in that case with a possible status check, like #68. This is not going to be solved in this issue.

praveenkumar:
Done and merged #187

PR #187, opened 2019-06-05
Issue #184 Add dns check for local and public domain

praveenkumar:
Add post dns checks.


gbraad:
@anjannath We need confirmation on some behaviour. I get error code `0` to domains that do not exist with `dig short lklfdsakldsklfhsa.com` for instance.

anjannath:
Right, i see the same behaviour, and found this old bug files for the same issue:  https://bugs.launchpad.net/ubuntu/+source/bind9/+bug/854705

Issue #97, opened 2019-04-25
Create separate pkg for different Linux distro specific pre-flight and setup 


praveenkumar:
As of now we are only targeting `Fedora/CentOS/RHEL` as part of our setup command and even some of the preflight test cases. Moving forward we should group it then distro specific so it would be easy. Might be just create a common interface and the each distro can implement those functions.

gbraad:
See: https://github.com/code-ready/crc/issues/185#issuecomment-499122791

gbraad:
We likely need to use `lsb_release -i` to determine the distro we are running on.

gbraad:
#190

guillaumerose:
With Ubuntu support we have this now, kind of.

PR #190, opened 2019-06-06
[WIP] Issue #97 Distro specific pre-flight checks for Ubuntu

gbraad:


gbraad:
@code-ready/crc-devel would like feedback

praveenkumar:
My idea was not having it this way but have a top level `Linux` interface which have those functions and then we have preflight_checker for RHEL/Ubuntu ... etc. which implement those as per the OS. The way we were doing for the minishift provisioner side, it will make it more easy for outside folks to contribute as per their distro specific for preflight checks.

gbraad:
@praveenkumar that is good for future use, but also needs more implementation and a lot of duplication.

In that case you would still perform the same check, as the provisioner directly did this on the `os_release` to select a provisioner. In my opinion that is an improvement we can make.

A refactor is needed anyway as at the moment the PreStart of services and the pre-flight checks are completely independent. I would suggest we actually pass a config in.

gbraad:
CentOS CI fails because of `WARN exec: "lsb_release": executable file not found in $PATH `
Interesting ... what kind of CentOS image is used?

praveenkumar:
retest this please.

pkg/os/distro_linux.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type DistroID string
+
+func (t DistroID) String() string {
+	return string(t)
+}
+
+func getLSBRelease() (string, error) {
+	stdOut, _, err := RunWithDefaultLocale("lsb_release", "-is")
```
praveenkumar: I would say use `/etc/os-release` and get the ID from there it there for every distro might be the reason why docker-machine doing it that way.

praveenkumar:
> but also needs more implementation and a lot of duplication.

No duplication just different implementations for different functions. I think it will be more cleaner and done quickly also. 

> A refactor is needed anyway as at the moment the PreStart of services and the pre-flight checks are completely independent. I would suggest we actually pass a config in.

Right but that will be just function parameters change, shouldn't have much impact about this part.

cfergeau:
> My idea was not having it this way but have a top level `Linux` interface which have those functions and then we have preflight_checker for RHEL/Ubuntu ... etc. which implement those as per the OS. The way we were doing for the minishift provisioner side, it will make it more easy for outside folks to contribute as per their distro specific for preflight checks.

I was about to make a similar suggestion, top level LinuxPreflightChecker interface with the common checks, and then RHELPreflightChecker with the RHEL specific checks, which starts by calling the common preflight checks, and then does the distro specific ones (or are there some interdependencies between the common/specific preflight checks which would make this tricky?)

pkg/crc/preflight/preflight_linux.go
```diff
@@ -89,18 +99,40 @@ func SetupHost() {
...
+		)
+		*/
+		/*
+		preflightCheckAndFix(config.GetBool(cmdConfig.SkipCheckUserInLibvirtGroup.Name),
+			checkUserPartOfLibvirtdGroup,
+			fixUserPartOfLibvirdtGroup,
+			"Adding user to libvirt group",
+			config.GetBool(cmdConfig.WarnCheckUserInLibvirtGroup.Name),
+		)
+		*/
```
cfergeau: This probably belongs to the next commit?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -157,6 +169,14 @@ func fixLibvirtEnabled() (bool, error) {
 }
 
 func checkUserPartOfLibvirtGroup() (bool, error) {
+	return checkUserPartOfGroup("libvirt")
+}
+
+func checkUserPartOfLibvirtdGroup() (bool, error) {
+	return checkUserPartOfGroup("libvirtd")
+}
+
```
cfergeau: I suspect these one liner helpers could be avoided by using lambdas in the corresponding `preflightCheckAndFix` calls (but I'm really not familiar enough with go to be fully sure this would work ;)

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -123,6 +124,17 @@ func fixLibvirtInstalled() (bool, error) {
 	return true, nil
 }
 
+// TODO: needs refactor
+func fixLibvirtInstalledUbuntu() (bool, error) {
+	logging.Debug("Trying to install libvirt")
+	stdOut, stdErr, err := crcos.RunWithPrivilege("apt-get", "install", "-y", "qemu-kvm", "libvirt-daemon", "libvirt-daemon-system")
+	if err != nil {
```
cfergeau: There's a `PackageManager` variable defined in LinuxDistribution, but it's not being used (and I'm not sure it should, as the command syntax is going to be different, the package names are also going to be different)

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -113,6 +113,7 @@ func checkLibvirtInstalled() (bool, error) {
 	return true, nil
 }
 
+// TODO: needs refactor
```
cfergeau: With this comment alone, it's not clear what needs to be refactored exactly. Maybe better to leave it out of this commit?

pkg/crc/preflight/preflight_linux.go
```diff
@@ -100,7 +100,7 @@ func SetupHost() {
 		config.GetBool(cmdConfig.WarnCheckKvmEnabled.Name),
 	)
 
-	if distro.ID == crcos.FEDORA {
+	if distro.ID == crcos.FEDORA || distro.ID == crcos.RHEL {
```
cfergeau: Is this something that should have been part of the previous commit?

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -84,7 +84,7 @@ func fixVirtualBoxInstallation() (bool, error) {
 }
 
 func checkResolverFilePermissions() (bool, error) {
-	err := unix.Access(resolverFile, unix.R_OK | unix.W_OK)
+	err := unix.Access(resolverFile, unix.R_OK|unix.W_OK)
 	if err != nil {
```
cfergeau: These changes are unrelated to this commit/this patch series.

pkg/crc/services/dns/dns.go
```diff
@@ -63,7 +63,7 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) (services.Servi
 	}
 
 	// Remove the dnsmasq container if exist during the VM stop cycle
-	drivers.RunSSHCommandFromDriver(serviceConfig.Driver,"sudo podman rm -f dnsmasq")
+	drivers.RunSSHCommandFromDriver(serviceConfig.Driver, "sudo podman rm -f dnsmasq")
```
cfergeau: Same for this change.

pkg/crc/preflight/preflight_checks_darwin.go
gbraad: there are due to missed `gofmt` by a previous commit. this is something that needs to be fixed... just happened to be in my commit as I always run this befor e`git commit`.

pkg/crc/preflight/preflight_checks_linux.go
gbraad: part of WIP

pkg/crc/preflight/preflight_checks_linux.go
gbraad: This was for possible solution we might were able to give for a known distro when we havent added all steps. such `Missing ... Use your the package manager %s to install these missing ... blah bla`, just didnt come to this (yet). However, we will solve this differently now.

pkg/os/distro_linux.go
cfergeau: As a data point, `redhat-lsb-core` which provides `lsb_release` is not installed on my fedora 30 machine, so maybe it's not part of the default install.

gbraad:
@sleshchenko does this provide the expected result?

Note: a very old release. This is the artifact from the PR: https://194-165193530-gh.circle-artifacts.com/0/crc/linux-amd64/crc Just use it for the setup steps (do not run `crc start` as this will not work with the bundle you used)

pkg/crc/preflight/preflight_checks_linux.go
gbraad: @cfergeau not mentioning the possible packagemanager leads people to use `yum`. This even happened in: https://github.com/code-ready/crc/issues/378#issue-476863030

pkg/crc/preflight/preflight_checks_linux.go
gbraad: I prefered to use `func` as lambda-like approach, but it is not easy to understand by everybody. Keeping it readable is more important, plus it is now understandable:

`check user part of group libvirtd` 

`checkUserPartOfGroup("libvirtd")`

gbraad:
started with a refactor of the pre-flights

rgov:
For what it's worth, I tried following the Getting Started Guide on Ubuntu 18.04 and even after installing packages, the `libvirtd` group was not created by default. So perhaps you should check for this case.

cfergeau:
> For what it's worth, I tried following the Getting Started Guide on Ubuntu 18.04 and even after installing packages, the `libvirtd` group was not created by default. So perhaps you should check for this case.

See https://github.com/code-ready/crc/issues/622 , the `libvirtd` group is actually not needed, the correct name is `libvirt`

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -157,6 +168,14 @@ func fixLibvirtEnabled() (bool, error) {
 }
 
 func checkUserPartOfLibvirtGroup() (bool, error) {
+	return checkUserPartOfGroup("libvirt")
+}
+
+func checkUserPartOfLibvirtdGroup() (bool, error) {
+	return checkUserPartOfGroup("libvirtd")
```
cfergeau: For the record, see https://github.com/code-ready/crc/issues/622 , the `libvirtd` group is not needed, `libvirt` group is fine on recent ubuntus.

Issue #188, opened 2019-06-05
Setup shows the ERRO even fix function do needful


praveenkumar:
```
19:53:36 + crc setup
19:53:36 INFO[0000] Error loading global state: open /root/.crc/globalstate.json: no such file or directory 
19:53:36 crc - Local OpenShift 4.x cluster
19:53:36 INFO  Caching oc binary                           
19:53:36 ERRO Error occured: oc binary is not cached.      
19:53:38 INFO  Copying '/tmp/crc103903385/oc' to '/root/.crc/cache/oc/oc' 
19:53:38 INFO  Setting up virtualization                   
19:53:38 INFO  Setting up KVM                              
19:53:38 INFO  Installing Libvirt                          
19:53:38 INFO  Adding user to libvirt group                
19:53:38 ERRO Error occured: [root] not part of libvirtd group 
19:53:38 INFO  Enabling libvirt                            
19:53:38 INFO  Starting Libvirt service                    
19:53:38 INFO  Installing crc-driver-libvirt               
19:53:39 INFO  Setting up Libvirt crc network              
19:53:40 ERRO Error occured: Libvirt network crc not found 
19:53:40 INFO  Starting Libvirt crc network                
19:53:40 ERRO Error occured: Libvirt crc network is not active 
19:53:41 INFO  Writing Network Manager config for crc      
19:53:41 INFO  Writing dnsmasq config for crc        
```

This is because of recent PR merge from @cfergeau, this is not the good UX for `crc setup`.

cfergeau:
Nope, it's caused by commit f70e40f45ff6a, something like this fixes the issue:
```
diff --git a/pkg/crc/preflight/preflight_checks_linux.go b/pkg/crc/preflight/preflight_checks_linux.go
index 601af77..e7de588 100644
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -2,6 +2,7 @@ package preflight
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -15,7 +16,7 @@ import (
 	"text/template"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/errors"
+	crcerr "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
[...]

Issue #179, opened 2019-06-03
Error loading global state


jsliacan:
When starting `crc` the first time (without `~/.crc` directory peresent), one gets the following error at the `setup` stage.
```
INFO[0000] Error loading global state: open /home/jsliacan/.crc/globalstate.json: no such file or directory 
```
As far as functionality is concerned, everything seems to work fine afterwards.

The steps to replicate:
```
$ rm -rf ~/.crc/
$ crc setup
```
Output:
```
INFO[0000] Error loading global state: open /home/jsliacan/.crc/globalstate.json: no such file or directory 
crc - Local OpenShift 4.x cluster
INFO  Caching oc binary                           
INFO  Copying '/tmp/crc677218342/oc' to '/home/jsliacan/.crc/cache/oc/oc' 
INFO  Setting up virtualization                   
INFO  Setting up KVM                              
INFO  Installing Libvirt                          
INFO  Adding user to libvirt group                
INFO  Enabling libvirt                            
INFO  Starting Libvirt service                    
INFO  Installing crc-driver-libvirt               
INFO  Setting up Libvirt crc network              
INFO  Starting Libvirt crc network                
INFO  Writing Network Manager config for crc      
INFO  Writing dnsmasq config for crc 
```

gbraad:
this happens on every command as the globalstate gets loaded as part of the library's `init`.

praveenkumar:
fixed by #191 since we have those as part of Debug instead info messages.

PR #191, opened 2019-06-06
Issue #188 Change crc/errors with errors package

praveenkumar:
Function/methods which need to return an error without
putting in the logs should be using fmt.ErrorF or error standard
package.

### DIFF
--- a/pkg/crc/init.go
+++ b/pkg/crc/init.go
@@ -10,6 +10,6 @@ func init() {
 	var err error
 	state.GlobalState, err = state.NewGlobalState(constants.GlobalStatePath)
 	if err != nil {
-		log.InfoF("Error loading global state: %s", err.Error())
+		log.DebugF("Error loading global state: %s", err.Error())
 	}
 }
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -2,6 +2,7 @@ package preflight
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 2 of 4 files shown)


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -175,7 +175,7 @@ func checkUserPartOfLibvirtGroup() (bool, error) {
 		logging.Debug("Current user is already in the libvirt group")
 		return true, nil
 	}
-	return false, errors.NewF("%s not part of libvirtd group", currentUser.Username)
+	return false, fmt.Errorf("%s not part of libvirtd group", currentUser.Username)
```
praveenkumar: `errors.NewF` implementation is like below.

```
func NewF(text string, args ...interface{}) error {
	logging.ErrorF(fmt.Sprintf("Error occured: %s", text), args)
	return fmt.Errorf(text, args...)
}
```

cfergeau:
Looks good to me!

cfergeau:
Hmm, though I think @gbraad wanted to discourage the use of fmt.New?

praveenkumar:
> though I think @gbraad wanted to discourage the use of fmt.New?

You mean `fmt.ErrorF` but `errors` std package doesn't have `ErrorF` functionality so either you use the `fmt.ErrorF` or `errors.New(fmt.Sprintf("...", var)`.

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -1,7 +1,8 @@
 package preflight
 
 import (
-	"github.com/code-ready/crc/pkg/crc/errors"
+	"fmt"
+	"errors"
```
cfergeau: I think we usually keep this alphabetically sorted?

pkg/crc/init.go
```diff
@@ -10,6 +10,6 @@ func init() {
 	var err error
 	state.GlobalState, err = state.NewGlobalState(constants.GlobalStatePath)
 	if err != nil {
-		log.InfoF("Error loading global state: %s", err.Error())
+		log.DebugF("Error loading global state: %s", err.Error())
```
cfergeau: I believe this is "fixing" https://github.com/code-ready/crc/issues/179 ?

PR #195, opened 2019-06-10
Issue #194 Add missing platform specific config field back to config cmd

### LINKED ISSUE #194: crc config --help not showing all the configurable options for different platform.
In linux if we try to see what all the options available as part of config it is output as below. which doesn't consist of all the options for the platform.

```
$ ./crc config -h
Modifies crc configuration properties. Some of the configuration properties are equivalent
to the options that you set when you run the 'crc start' command.
Configurable properties (enter as SUBCOMMAND): 

 * vm-driver
 * bundle
 * cpus
 * memory
 * skip-check-virtualbox-installed
 * warn-check-virtualbox-installed

Usage:
  crc config SUBCOMMAND [flags]
  crc config [command]

Available Commands:
  get         Gets a crc configuration property.
  set         Sets a crc configuration property.
  unset       Unsets a crc configuration property.
  view        Display complete crc configurations.

Flags:
  -h, --help   help for config

Global Flags:

Use "crc config [command] --help" for more information about a command.
```

praveenkumar:
Fixes #194

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -45,14 +45,17 @@ var (
 	ConfigCmd = &cobra.Command{
 		Use:   "config SUBCOMMAND [flags]",
 		Short: "Modifies crc configuration properties.",
-		Long: `Modifies crc configuration properties. Some of the configuration properties are equivalent
-to the options that you set when you run the 'crc start' command.
-Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields(),
 		Run: func(cmd *cobra.Command, args []string) {
 			cmd.Help()
 		},
 	}
 )
+
+func init() {
+	ConfigCmd.Long = `Modifies crc configuration properties. Some of the configuration properties are equivalent
+to the options that you set when you run the 'crc start' command.
+Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields()
+}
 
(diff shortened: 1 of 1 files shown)


cmd/crc/cmd/config/config.go
```diff
@@ -45,15 +45,18 @@ var (
...
 		},
 	}
 )
 
+func init() {
+	ConfigCmd.Long = `Modifies crc configuration properties. Some of the configuration properties are equivalent
+to the options that you set when you run the 'crc start' command.
+Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields()
+}
+
```
praveenkumar: Since `configurableFields` are not able to fully defined when it is part of `cmd` option, this will make sure it is have all the platform specific values also.

cmd/crc/cmd/config/config.go
```diff
@@ -45,15 +45,18 @@ var (
...
-		Long: `Modifies crc configuration properties. Some of the configuration properties are equivalent
-to the options that you set when you run the 'crc start' command.
-Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields(),
 		Run: func(cmd *cobra.Command, args []string) {
 			cmd.Help()
 		},
 	}
 )
 
+func init() {
```
anjannath: :+1:

PR #221, opened 2019-06-12
Remove unwanted dependencies from vendor directory

praveenkumar:
Most of those deps are from the coredns side and since we removed the dependencies now this should also need to be removed.


praveenkumar:
retest this please.

Issue #146, opened 2019-05-21
Implement environment command


gbraad:
Like we had before, a command like `oc-env` is much appreciated when using several terminals, or when the terminal closed.

PR #222, opened 2019-06-12
Issue #146 Add oc-env command

praveenkumar:


pkg/os/shell/shell.go
```diff
@@ -0,0 +1,146 @@
...
+	"fmt"
+	"os"
+	"runtime"
+	"strings"
+
+	"github.com/code-ready/machine/libmachine/shell"
+)
+
+var (
+	supportedShell = []string{"bash", "fish", "powershell", "cmd", "emacs", "tcsh", "zsh"}
```
gbraad: do we really want to support 'fish' again?

gbraad: I would also place this in an `_os.go` file, as powershell is windows only, bash is on both linux and macos... but we do not want to support hthis on the windows bash implementation, right? (we had issues with this before). same for `cmd`, not available on anything else.

vendor/github.com/code-ready/machine/libmachine/shell/shell_windows.go
```diff
@@ -0,0 +1,84 @@
...
+			} else if strings.Contains(strings.ToLower(shell), "cmd") {
+				return "cmd", nil
+			} else {
+				fmt.Printf("You can further specify your shell with either 'cmd' or 'powershell' with the --shell flag.\n\n")
+				return "cmd", nil // this could be either powershell or cmd, defaulting to cmd
+			}
+		}
+	}
+
+	if os.Getenv("__fish_bin_dir") != "" {
```
gbraad: fish on windows? ever used???

praveenkumar:
retest this please.

pkg/os/shell/shell_windows.go
```diff
@@ -0,0 +1,5 @@
+package shell
+
+var (
+	supportedShell = []string{"bash", "powershell", "cmd", "zsh"}
```
gbraad: I would not support bash and zsh on Windows.
This means people run it from Cygwin or WSL?

praveenkumar: Right in case of Cygwin or WSL side

gbraad: but on WSL you can not communicate with the hypervisor as the command is recognized/using the Linux executable. So this would in that case not work as expected.

Cygwin would work, but I think it is better to use a force option here instead.

praveenkumar: @gbraad why we need the force option, it does autodetect the shell env and print out as expected, as long as user using it on powershell/cmd it will only give that outputs.

gbraad: I do not think you understand me. Why would you want to support cygwin on Windows, when at times this didn't work as expected? At this point I do not think it is a ood choice.

so, this is why I mention `crc.exe oc-env -s bash` to be an option to FORCE if you insist on trying this on Cygwin's bash.

pkg/os/shell/shell_windows.go
```diff
@@ -0,0 +1,5 @@
+package shell
+
+var (
+	supportedShell = []string{"bash", "powershell"}
```
gbraad: ? I guess you made a small mistake.
what about `cmd`, and instead we have `bash`?

Issue #228, opened 2019-06-13
podman fails sometime with requested IP address 10.88.0.8 is not available for dnsmasq


praveenkumar:
Sometime when VM is stopped non-gracefully/aborted and we start it again we hit the podman dnsmasq issue where it says failed with following error:

```
Error while adding pod to CNI network "podman": failed to allocate for range 0: requested IP address 10.88.0.8 is not available in range set 10.88.0.1-10.88.255.254" 
error configuring network namespace for container 5e8b119f1b153e9b23a770838f632bef9d41c699592639ab1fff445458ea3ba1: failed to allocate for range 0: requested IP address 10.88.0.8 is not available in range set 10.88.0.1-10.88.255.254 
```

- https://github.com/containers/libpod/issues/2767

praveenkumar:
fixed by #229

Issue #225, opened 2019-06-12
CRC 4.1.0 not able to start on OSX with enabled WIFI and wired network


skabashnyuk:
I have this error on my OSX 10.14.5 with Virtualbox 6, crc -version: 0.87.0-alpha-4.1.0+3a5033a
```
[14:07:20]sj:code-ready#: crc start -b ~/Downloads/crc_vbox_4.1.0.tar.xz
crc - Local OpenShift 4.x cluster
INFO Checking if oc binary is cached
INFO Checking if VirtualBox is Installed
INFO Checking file permissions for resolver
INFO Extracting the Bundle tarball ...
INFO Creating VM ...
INFO Bridge IP on the host: 192.168.130.1
INFO Restarting the network
ERRO Failed internal dns query: ssh command error:
command : host -R 3 foo.apps-crc.testing
err     : exit status 1
output  : Host foo.apps-crc.testing not found: 3(NXDOMAIN)
 :
INFO Running
[ssh command error:
command : host -R 3 foo.apps-crc.testing
err     : exit status 1
output  : Host foo.apps-crc.testing not found: 3(NXDOMAIN)
]
```
```
[root@crc-4gdnp-master-0 core]# cat /etc/resolv.conf
# Generated by NetworkManager
search ckc.redhat.com
nameserver 10.0.2.3
```
```
[root@crc-4gdnp-master-0 core]# cat /var/srv/dnsmasq.conf
user=root
port= 53
bind-interfaces
expand-hosts
log-queries
srv-host=_etcd-server-ssl._tcp.crc.testing,etcd-0.crc.testing,2380,10
local=/crc.testing/
domain=crc.testing
address=/apps-crc.testing/192.168.130.111
address=/etcd-0.crc.testing/192.168.130.111
address=/api.crc.testing/192.168.130.111
address=/api-int.crc.testing/192.168.130.111
address=/crc-4gdnp-master-0.crc.testing/192.168.130.111
```
I've noticed that during start wired network is switched on/off

![Знімок екрана  о 14 09 10](https://user-images.githubusercontent.com/1614429/59347141-8fa3fb00-8d1c-11e9-90a3-aeb5bcf59d09.png)
![Знімок екрана  о 14 08 59](https://user-images.githubusercontent.com/1614429/59347149-916dbe80-8d1c-11e9-95f2-ddc9f82918ef.png)

May be that is the reason why  /etc/resolv.conf has the wrong content.

praveenkumar:
[...]

PR #234, opened 2019-06-13
Issue #228 Add debug info for Network device restart in Mac

praveenkumar:
- Add logic to wait for device is connected to internet.
- Host bridge info should be debug message.

### DIFF
--- a/pkg/crc/network/nameservers.go
+++ b/pkg/crc/network/nameservers.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	crcos "github.com/code-ready/crc/pkg/os"
 	"github.com/code-ready/machine/libmachine/drivers"
 )
 
@@ -25,45 +26,15 @@ func HasNameserversConfigured(driver drivers.Driver) bool {
 	return i != 0
 }
 
-func GetResolvValuesFromInstance(driver drivers.Driver) *ResolvFileValues {
+func GetResolvValuesFromInstance(driver drivers.Driver) (*ResolvFileValues, error) {
 	cmd := "cat /etc/resolv.conf"
 	out, err := drivers.RunSSHCommandFromDriver(driver, cmd)
 
 	if err != nil {
-		return nil
+		return nil, err
 	}
 
-	searchdomains := []SearchDomain{}
-	nameservers := []NameServer{}
-
-	for _, line := range strings.Split(strings.TrimSuffix(out, "\n"), "\n") {
-		if len(line) > 0 && (line[0] == ';' || line[0] == '#') {
-			continue
-		}
-
-		f := strings.Fields(line)
-		if len(f) < 1 {
-			continue
-		}
-
-		switch f[0] {
-		case "nameserver":
-			nameservers = append(nameservers, NameServer{IPAddress: f[1]})
-		case "search":
-			for i := 0; i < len(f)-1; i++ {
-				searchdomains = append(searchdomains, SearchDomain{Domain: f[i+1]})
-			}
-		default:
-			// ignore
-		}
-	}
-
-	resolvFileValues := ResolvFileValues{
-		SearchDomains: searchdomains,
-		NameServers:   nameservers,
-	}
-
-	return &resolvFileValues
+	return parseResolveConfFile(out)
 }
 
 func CreateResolvFileOnInstance(driver drivers.Driver, resolvFileValues ResolvFileValues) {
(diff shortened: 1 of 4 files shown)


jsliacan:
Tried 5 times on 10.14.5 and no warnings printed during `crc start`. Seems to work. Was on WiFi each time.

jsliacan:
Worked another 3 times. Not sure what you changed though.

skabashnyuk:
issue https://github.com/code-ready/crc/issues/225 reproduced. Also reproduced on yet another OSX in the same network.
logs https://pastebin.com/QpyW1RAa

praveenkumar:
@skabashnyuk `echo dddddd= | base64 --decode | sudo tee /etc/resolv.conf > /dev/null` L#1687 doesn't look like it is correct.

skabashnyuk:
I thought that was some private information and I've replaced it myself before publishing. Do you want me to restore original values?

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -60,20 +60,36 @@ func createResolverFile(InstanceIP string, path string) (bool, error) {
...
 			return false, fmt.Errorf("%s: %v", stderr, err)
 		}
 	}
+	waitForNetwork()
 	return true, err
 }
+
+// Wait for Network wait till the network is up, since it is required to resolve external dnsquery
+func waitForNetwork() {
+	cmd := "ping -c4 1.1.1.1"
```
gbraad: why we use this????!!!??? this is not guaranteed to work.
We should be checking the users gateway.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -60,20 +60,36 @@ func createResolverFile(InstanceIP string, path string) (bool, error) {
...
+	// retry for 5 times
+	for i := 0; i < 5; i++ {
+		_, _, err := crcos.RunWithDefaultLocale("ping", "-c4", "1.1.1.1")
```
gbraad: why do we not use the `cmd` here?
I would first extract the test endpoint, but it would be best to use the network gateway that is set to the network interface.

praveenkumar:
retest this please.

pkg/crc/network/nameservers.go
```diff
@@ -100,3 +101,47 @@ func HasNameserverConfiguredLocally(nameserver NameServer) (bool, error) {
 
 	return strings.Contains(string(file), nameserver.IPAddress), nil
 }
+
+func GetResolvValuesFromHost() (*ResolvFileValues, error) {
```
gbraad: Why do you not refactor `func GetResolvValuesFromInstance` to reuse the same code?
This is unnecessary duplication as you could create the previous function to accept the multiline as input and return the filled struct.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -67,23 +68,41 @@ func createResolverFile(InstanceIP string, path string) (bool, error) {
...
+
+// Wait for Network wait till the network is up, since it is required to resolve external dnsquery
+func waitForNetwork() {
+	hostResolv, err := network.GetResolvValuesFromHost()
+	if err != nil {
+		logging.WarnF("Not able to read host resolv file (%v)", err)
+	}
+	// retry for 5 times
+	for i := 0; i < 5; i++ {
+		for _, ns := range hostResolv.NameServers {
```
gbraad: nameservers here is OK too, as this means the gateway allows routing.
👍

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -67,23 +68,41 @@ func createResolverFile(InstanceIP string, path string) (bool, error) {
...
 	}
+	waitForNetwork()
 	return true, err
 }
+
+// Wait for Network wait till the network is up, since it is required to resolve external dnsquery
+func waitForNetwork() {
+	hostResolv, err := network.GetResolvValuesFromHost()
+	if err != nil {
+		logging.WarnF("Not able to read host resolv file (%v)", err)
```
gbraad: s/Not able/Unable/

pkg/crc/network/nameservers.go
```diff
@@ -100,3 +71,50 @@ func HasNameserverConfiguredLocally(nameserver NameServer) (bool, error) {
...
+	// TODO: we need to add runtime OS in case of windows.
+	out, stderr, err := crcos.RunWithDefaultLocale("cat", "/etc/resolv.conf")
+	if crcos.CurrentOS() == crcos.WINDOWS {
+		// TODO: we need to add logic in case of windows.
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("%s : %v", stderr, err)
+	}
+	return parseResolveConfFile(out)
```
gbraad: :+1

pkg/crc/network/nameservers.go
```diff
@@ -25,45 +26,15 @@ func HasNameserversConfigured(driver drivers.Driver) bool {
...
-		}
-	}
-
-	resolvFileValues := ResolvFileValues{
-		SearchDomains: searchdomains,
-		NameServers:   nameservers,
-	}
-
-	return &resolvFileValues
+	return parseResolveConfFile(out)
```
gbraad: re-use for the win

👍

pkg/crc/services/dns/dns.go
```diff
@@ -55,7 +55,11 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) (services.Servi
 		return *result, err
 	}
 
-	orgResolvValues := network.GetResolvValuesFromInstance(serviceConfig.Driver)
+	// We need to restart the Host Network before updating
+	// the VM's /etc/resolv.conf file.
+	res, err := runPostStartForOS(serviceConfig, result)
```
gbraad: what if an `err` occurs here?

pkg/crc/services/dns/dns.go
```diff
@@ -55,7 +55,11 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) (services.Servi
...
+
+	orgResolvValues, err := network.GetResolvValuesFromInstance(serviceConfig.Driver)
```
gbraad: because this `err` will do make this unclear to what might have happened.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -67,23 +68,41 @@ func createResolverFile(InstanceIP string, path string) (bool, error) {
...
+		time.Sleep(1 * time.Second)
+		stdout, stderr, err := crcos.RunWithDefaultLocale("networksetup", "-setnetworkserviceenabled", netdevice, "off")
+		logging.DebugF("Disabling the %s Device (stdout: %s), (stderr: %s)", netdevice, stdout, stderr)
+		stdout, stderr, err = crcos.RunWithDefaultLocale("networksetup", "-setnetworkserviceenabled", netdevice, "on")
+		logging.DebugF("Enabling the %s Device (stdout: %s), (stderr: %s)", netdevice, stdout, stderr)
 		if err != nil {
 			return false, fmt.Errorf("%s: %v", stderr, err)
 		}
 	}
+	waitForNetwork()
```
gbraad: surround with whitelines to make this stand out a little bit more...

pkg/crc/services/dns/dns.go
praveenkumar: We are returning it https://github.com/code-ready/crc/pull/234/files#diff-77b86e271b726efdfa282a22b5c5fd27R75 here.

gbraad:
compilation passed, and does not effect libvirt, so safe to merge.

Issue #235, opened 2019-06-14
Copy kubeconfig under machine instead using from cache.


praveenkumar:
Right now we are directly using the kubeconfig  from the bundle which we need to put under `machine` so that when we try to use oc operation as part of post cluster we should be using that kubeconfig which will be unmodified by users. 

Another issue is most of the time user export this kubeconfig and then also login to different cluster provider which make it flooded.

praveenkumar:
fixed by #236

PR #236, opened 2019-06-14
Issue #235 Copy kubeconfig file from cache to instance dir

praveenkumar:
This kubeconfig file is system:admin role and this should
be used for cluster admin operations like resource creation
on different namespace or getting the information. Sometime
User just export it from cache dir just after installation
and this will be bloated if they start login to different
clusters without unsetting it.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -3,8 +3,6 @@ package machine
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/code-ready/crc/pkg/crc/systemd"
-	"github.com/code-ready/crc/pkg/os"
 	"io/ioutil"
 	"path/filepath"
 	"time"
@@ -15,6 +13,8 @@ import (
 
 	// host and instance related
 	"github.com/code-ready/crc/pkg/crc/network"
+	"github.com/code-ready/crc/pkg/crc/systemd"
+	crcos "github.com/code-ready/crc/pkg/os"
 	// cluster services
 	"github.com/code-ready/crc/pkg/crc/services"
 	"github.com/code-ready/crc/pkg/crc/services/dns"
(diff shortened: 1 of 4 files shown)


pkg/crc/machine/machine.go
```diff
@@ -15,6 +13,8 @@ import (
 
 	// host and instance related
 	"github.com/code-ready/crc/pkg/crc/network"
+	"github.com/code-ready/crc/pkg/crc/systemd"
```
gbraad: :+1

pkg/crc/machine/machine.go
```diff
@@ -184,6 +184,18 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		logging.WarnF("Failed Public dns query: %v : %s", err, queryOutput)
 	}
 
+	// Copy Kubeconfig file from bundle extract path to machine directory.
+	// In our case it would be ~/machine/crc
+	logging.InfoF("Copying kubeconfig file to instance dir ...")
+	if err := crcos.CopyFileContents(
+		filepath.Join(extractedPath, "kubeconfig"),
+		filepath.Join(constants.MachineInstanceDir, constants.DefaultName, "kubeconfig"),
```
gbraad: if we use DefaultName here, we assume that there will always be a single instance. We should be using MachineConfig.Name here.

pkg/crc/oc/oc_cache.go
```diff
@@ -2,17 +2,18 @@ package oc
...
-	"github.com/code-ready/crc/pkg/extract"
-	"github.com/pkg/errors"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
```
gbraad: :+1

pkg/os/util.go
```diff
@@ -55,3 +58,30 @@ func CurrentExecutable() (string, error) {
 	}
 	return currentExec, nil
 }
+
+func CopyFileContents(src, dst string, permission os.FileMode) error {
```
gbraad: why is src not typed as string?

praveenkumar: In golang if 2 variable same type then each variable doesn't need explicit type definition, but I update.

pkg/crc/machine/machine.go
```diff
@@ -190,6 +190,18 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 	}
 
+	// Copy Kubeconfig file from bundle extract path to machine directory.
+	// In our case it would be ~/machine/crc
+	logging.InfoF("Copying kubeconfig file to instance dir ...")
+	if err := crcos.CopyFileContents(
+		filepath.Join(extractedPath, "kubeconfig"),
+		filepath.Join(constants.MachineInstanceDir, machineConfig.Name, "kubeconfig"),
+		0644); err != nil {
+		logging.ErrorF("Error to copy kubeconfig content %v", err)
```
anjannath: You could change this error message to: `Error while copying kubeconfig content`

Issue #237, opened 2019-06-14
Wait till cluster operators are in healthy state after 3m.


praveenkumar:
Right now we are just putting 3 mins wait to check if cluster operators are come up in healthy state, after that 3 mins we should wait 2 more mins (with polling) if all operator are in healthy state or not and let user know which operator still are in that state.

guillaumerose:
Fixed by https://github.com/code-ready/crc/commit/d6cc7a194ea1bc8fa7987fdad576bfa4f0279245

Issue #68, opened 2019-04-18
Implement status command


gbraad:

anjannath:
What are the things that should be checked to get the status?
We could check if 
- All the cluster operators have reached their expected state.
- Web-console pods are in running state.

will these be sufficient to infer that the cluster is running or more parameters needs to be checked?

anjannath:
Closed via  #256

PR #242, opened 2019-06-17
[Hold] Issue #237 Wait for cluster operators to be available

praveenkumar:
User should know after 3 mins if any operator is still not completely
in the *Available* state then they should wait before putting workload.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -212,13 +212,39 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		result.Error = err.Error()
 	}
 	if kubeletStarted {
-		logging.InfoF("Starting OpenShift cluster ... [waiting 3m]")
+		logging.InfoF("Starting OpenShift cluster ... [waiting 5m]")
 	}
 	result.KubeletStarted = kubeletStarted
 
 	// If no error, return usage message
 	if result.Error == "" {
+		// This initial sleep time is really required to make sure the cluster
+		// operators able to register their state to apiserver from last
+		// time since we are using the disk image which have it older
+		// state as part of etcd  data.
 		time.Sleep(time.Minute * 3)
+		// Checks after 3 mins wait all the operators are in 'Available' state.
+		// If any of the operator still not available then wait for 2 more mins and tell users
+		// about which operator is still not available and need time.
+		var status bool
+		crcoc := oc.GetOC(machineConfig.Name)
+		for i := 0; i < 2; i++ {
+			status, err = oc.GetClusterOperatorStatus(crcoc)
+			if err != nil {
+				logging.ErrorF("Failed to get cluster status: %v", err)
+				result.Error = err.Error()
+				return *result, err
+			}
+			if status {
+				continue
+			}
+			time.Sleep(time.Minute)
+		}
+
+		if !status {
+			logging.Warn("Some cluster operators are still not in completely 'Available' state. Check with 'crc status' command")
+		}
+
 		logging.InfoF("To access the cluster using 'oc', run 'oc login -u kubeadmin -p %s %s'", result.ClusterConfig.KubeAdminPass, result.ClusterConfig.ClusterAPI)
 		logging.InfoF("Access the OpenShift web-console here: %s", result.ClusterConfig.WebConsoleURL)
 		logging.InfoF("Login to the console with user: kubeadmin, password: %s", result.ClusterConfig.KubeAdminPass)
--- a/pkg/crc/oc/clusteroperator.go
+++ b/pkg/crc/oc/clusteroperator.go
@@ -1,0 +1,78 @@
+package oc
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"time"
+)
+
+var ignoreClusterOperators = []string{"monitoring", "machine-config", "marketplace"}
+
+type ClusterOperator struct {
+	Items []struct {
+		Metadata struct {
+			Name string `json:"name"`
+		} `json:"metadata"`
+		Status struct {
+			Conditions []struct {
+				LastTransitionTime time.Time `json:"lastTransitionTime"`
+				Reason             string    `json:"reason"`
+				Status             string    `json:"status"`
+				Type               string    `json:"type"`
+			} `json:"conditions"`
+		} `json:"status,omitempty"`
+	} `json:"items"`
+}
+
+func GetClusterOperatorStatus(oc OC) (bool, error) {
+	allAvailable := true
+	data, stderr, err := oc.RunOcCommand("get", "co", "-ojson")
+	if err != nil {
+		return false, fmt.Errorf("%s - %v", stderr, err)
+	}
+
+	var co ClusterOperator
+
+	err = json.Unmarshal([]byte(data), &co)
+	if err != nil {
+		return false, err
+	}
+	for _, c := range co.Items {
+		if contains(c.Metadata.Name, ignoreClusterOperators) {
+			continue
+		}
+		for _, con := range c.Status.Conditions {
+			switch con.Type {
+			case "Available":
+				if con.Status != "True" {
+					logging.Info(c.Metadata.Name, "operator not available, retrying ...")
+					logging.Debug(c.Metadata.Name, "operator not available, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			case "Degraded":
+				if con.Status != "False" {
+					logging.Info(c.Metadata.Name, "operator is degraded, retrying ...")
+					logging.Debug(c.Metadata.Name, "operator is degraded, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			case "Progressing":
+				if con.Status != "False" {
+					logging.Info(c.Metadata.Name, "operator is still progressing ...")
+					logging.Debug(c.Metadata.Name, "operator is still progressing, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			}
+		}
+	}
+	return allAvailable, nil
+}
+
+func contains(value string, list []string) bool {
+	for _, v := range list {
+		if v == value {
+			return true
+		}
+	}
+	return false
+}


pkg/crc/machine/machine.go
```diff
@@ -195,11 +196,32 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 
 	// If no error, return usage message
 	if result.Error == "" {
 		time.Sleep(time.Minute * 3)
+		// Checks after 3 mins wait all the operators are in 'Available' state.
+		// If any of the operator still not available then wait for 2 more mins and tell users
+		// about which operator is still not available and need time.
+		var status bool
+		for i := 0; i < 2; i++ {
+			status, err = oc.GetClusterOperatorStatus(filepath.Join(constants.MachineCacheDir, "oc", "oc"), filepath.Join(extractedPath, crcBundleMetadata.ClusterInfo.KubeConfig))
```
gbraad: can we create the command as a separate variable?
and same for the kubeconfig path?

These can be reused, and as varibale it makes the code more readable.

gbraad: can we create the command as a separate variable?
and same for the kubeconfig path?

These can be reused, and as varibale it makes the code more readable.

pkg/crc/machine/machine.go
```diff
@@ -195,11 +196,32 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+				return *result, err
+			}
+			if status {
+				continue
+			}
+			time.Sleep(time.Minute)
+		}
+
+		if !status {
+			logging.Warn("Mentioned cluster operators are still not in completely 'Available' state. Wait for some more time before using this cluster")
```
gbraad: this suggestion does not provide enough info to what the user should look for with this command.

it is better to expose this as a possible status command (to concise the info) and allow for a verbose option to get all details.

praveenkumar:
retest this please

anjannath:
@praveenkumar Is this in WIP because of #236?

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,79 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	crcos "github.com/code-ready/crc/pkg/os"
+	"time"
+)
+
+var ignoreClusterOperators = []string{"monitoring", "machine-config", "marketplace"}
+
+type ClusterOperator struct {
```
anjannath: This can be named as `PodStatus`

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,79 @@
...
+				LastTransitionTime time.Time `json:"lastTransitionTime"`
+				Reason             string    `json:"reason"`
+				Status             string    `json:"status"`
+				Type               string    `json:"type"`
+			} `json:"conditions"`
+		} `json:"status,omitempty"`
+	} `json:"items"`
+}
+
+func GetClusterOperatorStatus(ocPath string, kubeconfigPath string) (bool, error) {
```
anjannath: and we can make this more generic, then it can be used to get the status of all the pods from any namespace, for e.g status of the pods from `openshift-console` namespace

anjannath: maybe pass in a ignore list as an argument

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	for _, c := range co.Items {
+		if isContain(c.Metadata.Name, ignoreClusterOperators) {
+			continue
+		}
+		for _, con := range c.Status.Conditions {
+			switch con.Type {
+			case "Available":
+				if con.Status != "True" {
+					logging.InfoF(c.Metadata.Name, "operator not available, retrying ...")
```
anjannath: Use format specifiers here.

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,79 @@
...
+					logging.DebugF(c.Metadata.Name, "operator not available, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			case "Degraded":
+				if con.Status != "False" {
+					logging.InfoF(c.Metadata.Name, "operator is degraded, retrying ...")
```
anjannath: Same here.. `logging.InfoF("%s operator is degraded, retrying ...", c.Metadata.Name)`

praveenkumar: I used `Info` only instead `InfoF` now.

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,78 @@
...
+			case "Available":
+				if con.Status != "True" {
+					logging.Info(c.Metadata.Name, "operator not available, retrying ...")
+					logging.Debug(c.Metadata.Name, "operator not available, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			case "Degraded":
+				if con.Status != "False" {
+					logging.Info(c.Metadata.Name, "operator is degraded, retrying ...")
+					logging.Debug(c.Metadata.Name, "operator is degraded, Reason: ", con.Reason)
```
cfergeau: This is quite redundant, maybe the Info output is not really useful here (too verbose), and we can only print an Info summary at the end ("still in progress, N operators are degraded, N oerators are progressing"?

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,78 @@
...
+					logging.Debug(c.Metadata.Name, "operator is still progressing, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			}
+		}
+	}
+	return allAvailable, nil
+}
+
+func isContain(value string, list []string) bool {
```
cfergeau: Can it be only `contains` rather than `isContain`?

pkg/crc/oc/oc.go
```diff
@@ -0,0 +1,30 @@
...
+	return oc
+}
+
+func (oc OC) RunOcCommand(args ...string) (string, string, error) {
+	args = append(args, "--kubeconfig", oc.KubeconfigPath)
+	stdout, stderr, err := crcos.RunWithDefaultLocale(oc.OcBinaryPath, args...)
+	if err != nil {
+		return stdout, stderr, err
+	}
+	return stdout, stderr, nil
```
cfergeau: This can be summed up as
```
args = append(args, "--kubeconfig", oc.KubeconfigPath)
return crcos.RunWithDefaultLocale(oc.OcBinaryPath, args...)
```

pkg/crc/oc/oc.go
```diff
@@ -0,0 +1,30 @@
+package oc
+
+import (
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	crcos "github.com/code-ready/crc/pkg/os"
+)
+
+type OC struct {
```
cfergeau: Not familiar with go conventions, should be this `type OC` or `type Oc`?

pkg/crc/oc/oc.go
```diff
@@ -0,0 +1,30 @@
...
+	OcBinaryPath   string
+	KubeconfigPath string
+}
+
+func GetOC(machineName string) OC {
+	oc := OC{
+		OcBinaryPath:   filepath.Join(constants.OcCacheDir, constants.OcBinaryName),
+		KubeconfigPath: filepath.Join(constants.MachineInstanceDir, machineName, "kubeconfig"),
```
cfergeau: Fwiw, we should have some helpers to find where the kubeconfig file and the oc file were stored by crc, I don't think it's good to duplicate their exact location each time we need it. However, the problem is preexisting with eg the `machines` dir, so this can be solved at a later time.

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, v := range list {
+		if v == value {
```
cfergeau: No golang builtin for this? (honest question, this is a part of go I haven't looked at much yet :)

pkg/crc/machine/machine.go
```diff
@@ -214,11 +216,32 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		logging.InfoF("Starting OpenShift cluster ... [waiting 3m]")
 	}
 	result.KubeletStarted = kubeletStarted
-	//
 
 	// If no error, return usage message
 	if result.Error == "" {
 		time.Sleep(time.Minute * 3)
+		// Checks after 3 mins wait all the operators are in 'Available' state.
+		// If any of the operator still not available then wait for 2 more mins and tell users
```
cfergeau: Shouldn't we be doing this for 6 minutes total, and check every minute (or even more often than that?)? If all operators are done in 1 minute, I think we can tell the user everything is fine right away rather than waiting for 2 more minutes?

pkg/crc/oc/clusteroperator.go
praveenkumar: @cfergeau no it is not buildin part :)

pkg/crc/machine/machine.go
praveenkumar: @cfergeau there is no way operator are done within the min. this check we are adding because those are not even done in 3 mins.

pkg/crc/oc/oc.go
praveenkumar: @cfergeau we can have it `Oc`, shouldn't be problem, I put it both capital since both are initial and doesn't look good if one is small and one is capital one :)

pkg/crc/machine/machine.go
cfergeau: It would still be more consistent to do the check every minute even if it's not going to be ready after 3 minutes. Maybe in the future they will make improvements and the cluster wil be readymuch faster :)

pkg/crc/oc/clusteroperator.go
praveenkumar: @cfergeau I am using the same way the installer used to do it, Debug is useful to figure out what is the reason without manually checking grepping it for clusteroperator.

praveenkumar:
retest this please.

pkg/crc/oc/clusteroperator.go
gbraad: `inArray` or `contains`

and this is generic code. consider a better location for this kind of functionality.

pkg/crc/machine/machine.go
gbraad: but wihtin a minute you could at least check some status and possibly report progress.

pkg/crc/oc/clusteroperator.go
gbraad: agree...

`Contains` makes more sense... but I also think this is more generic.

pkg/crc/machine/machine.go
praveenkumar: @gbraad within the minute there is no `co` resource created.

pkg/crc/oc/oc.go
gbraad: if no further error processing is needed, return the function result directly.

pkg/crc/oc/oc.go
gbraad: 'were' stored? I do not think we need to figure this out, but have consistency about what we will do.

It would be best to ALWAYS use the version from the running instance, which therefore means we need to copy the kubeconfig

pkg/crc/machine/machine.go
gbraad: we do not talk about created, but status changes to be 'ready/available'... right? and 'each minute' to test and report about this sounds to me a reasonable expectation. The user needs feedback... but rather have him wait for 3 minutes before anything? By that time he might conclude the process hangs ...

praveenkumar: @gbraad they way we are now getting the operator response back using `oc get co` and within the 1-2 mins time `co => clusteroperator` resources is not even present so the unmarshal will fail then and there. For waiting time feedback for every operation we need to introduce a timedots like we have in minishift but that would be a separate issue/PR.

gbraad:
If unmarshaling can fail, we should handle that in code, as a slow(er)
system might also have this happen after 3 mins... Or whenever. Failing to
unmarshal indicates that the api server is unavailable, so inform the user
that the state is undetermined or api server is unavailable. It is still
state/info we can use.

praveenkumar:
@gbraad It is not only unmarshaling, which we can handle but also the false result due to store as last state in the disk image. What might happen if we have only 1 mins of interval from beginning then it might be goes to first cycle and as soon as `co` available it put it last state which was true for every operator and that gets succeed which is false. 

So I am still in favour to have 3 mins grace waiting time, but I do agree to put an info if a system is so slow that within 3 mins `co` is not available from api side.

Following is 2 different run when I go with 1 min time interval.

1. This something we can easily handle in the codebase.
```
$ ./crc start -b ~/Downloads/crc_vbox_4.1.2.tar.xz -d virtualbox
INFO Checking if VirtualBox is Installed          
INFO Checking file permissions for /etc/resolv.conf 
INFO Extracting the Bundle tarball ...            
INFO Creating VM ...                              
INFO Restarting the host network                  
INFO Check internal and public dns query ...      
INFO Copying kubeconfig file to instance dir ...  
INFO Starting OpenShift cluster ... [waiting 5m]  
INFO Waiting for initial 1 min                    
INFO Getting cluster operator status              
ERRO Failed to get cluster status: error: the server doesn't have a resource type "co"
 - exit status 1 
```

2.  This give us false result since it took the state from the time disk created.
```
$ ./crc start -b ~/Downloads/crc_vbox_4.1.2.tar.xz -d virtualbox
INFO Checking if VirtualBox is Installed          
INFO Checking file permissions for /etc/resolv.conf 
INFO Extracting the Bundle tarball ...            
INFO Creating VM ...                              
INFO Restarting the host network                  
INFO Check internal and public dns query ...      
INFO Copying kubeconfig file to instance dir ...  
INFO Starting OpenShift cluster ... [waiting 5m]  
INFO Waiting for initial 1 min                    
INFO Getting cluster operator status              
INFO To access the cluster using 'oc', run 'oc login -u kubeadmin -p 2K9Nb-ATSIE-pyAAK-p9UJN https://api.crc.testing:6443' 
INFO Access the OpenShift web-console here: https://console-openshift-console.apps-crc.testing 
INFO Login to the console with user: kubeadmin, password: 2K9Nb-ATSIE-pyAAK-p9UJN 
CodeReady Containers instance is running
```

gbraad:
Disk state when created (and therefore previous run state) has to be handle
anyway. This sounds like a clean action need to happen...

However, this also sounds like something that needs to be resolved from
openshift. Inconsisteng/previous state is not wished for

praveenkumar:
> However, this also sounds like something that needs to be resolved from openshift. Inconsisteng/previous state is not wished for

@gbraad What is our course of action for this PR to be in? Openshift side of inconsistent we will follow either on ML or file a bug on respective repo.

gbraad:
I am of the opinion that giving it 3 minutes is working around an issue.
This is technical debt that has to be dealt with on our end anyways. It is
unlikely it would be fixed in due course, so... If you want it now to wait
for 3 minutes instead, I am not satisfied, as the previous wait state was a
temporary fix. Now it would become the norm? If this option is chosen o
would want the progress dots ASAP... And this is still just a workarond.

praveenkumar:
@gbraad Ack, I don't know right now the best way so I am going to put this as hold till we have a better understanding.

gbraad:
Let's discuss this 'soon-ish' with the whole @code-ready/crc-devel team ?

anjannath:
@praveenkumar Can you please split the `oc` part of changes (the oc.ClusterOperatorStatus function and related changes) into a separate commit? That way i can use it in #68 :)

cfergeau:
I've rebased/updated this PR in https://github.com/cfergeau/crc/commits/issue_237
In my testing with 4.2.10, I haven't noticed the state of the cluster being stale ("ready") right after boot before switching to something different, so maybe this issue is gone.
I haven't necessarily addressed all the comments from this PR, I'll have to go through them.

gbraad:
Leaving this PR closed as an updated version is being worked on by @cfergeau (is this still the case?)

praveenkumar:
yes some of the part is already worked by @cfergeau but we still not wait until all the operators are up and running. Our source of truth in that case is only `crc status` which shows `running` in case all the operators are running.

PR #246, opened 2019-06-17
Error out when using `crc start --vm-driver xxx` when `xxx` is not the driver which is going to be used

cfergeau:


pkg/crc/machine/machine.go
```diff
@@ -105,6 +105,12 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		result.Status = vmState.String()
 	} else {
 		host, err := libMachineAPIClient.Load(machineConfig.Name)
+		if host.Driver.DriverName() != startConfig.VMDriver {
+			err := errors.NewF("VM driver '%s' was requested, but loaded VM is using '%s' instead",
+				startConfig.VMDriver, host.Driver.DriverName())
+			result.Error = err.Error()
+			return *result, err
+		}
```
praveenkumar: This is occur when `crc start => crc stop => crc start -d <different_what_used_initially>`

pkg/crc/errors/error.go
```diff
@@ -13,6 +13,6 @@ func New(text string) error {
 }
 
 func NewF(text string, args ...interface{}) error {
-	logging.ErrorF(fmt.Sprintf("Error occured: %s", text), args)
+	logging.ErrorF(fmt.Sprintf("Error occured: %s", text), args...)
```
praveenkumar: 👍

praveenkumar:
ok to test

Issue #249, opened 2019-06-18
Change /etc/resolv.conf ownership for Mac


praveenkumar:
As of now `/etc/resolv.conf` is non writeable by user and need sudo access after start which we want to avoid so as a result we are going to provide the ownership of that file to current user.

praveenkumar:
fixed by #254

PR #254, opened 2019-06-20
Issue #249 Add Instance IP to host /etc/resolv.conf

praveenkumar:
This will allow us to add instance IP on fly so user doesn't
need to do it manually.

As preflight check we are also changing the permission for
host /etc/resolv.conf file to provide RW access to current user.


praveenkumar:
retest this please.

praveenkumar:
retest this please

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -23,6 +23,12 @@ func StartPreflightChecks() {
 		"Checking file permissions for resolver",
 		config.GetBool(cmdConfig.WarnCheckResolverFilePermissions.Name),
 	)
+
+	preflightCheckSucceedsOrFails(config.GetBool(cmdConfig.SkipCheckResolvConfFilePermissions.Name),
+		checkResolvConfFilePermissions,
+		"Checking file permissions for resolv.conf",
```
anjannath: We should put the full path of the file `/etc/resolv.conf` and `/etc/resolver/testing`

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -28,8 +28,15 @@ type resolverFileValues struct {
...
-	success, err := createResolverFile(serviceConfig.IP, filepath.Join("/", "etc", "resolver", serviceConfig.BundleMetadata.ClusterInfo.BaseDomain))
+	// Update resolv.conf file for host
+	success, err := updateResolvConfFile(serviceConfig.IP, filepath.Join("/", "etc", "resolv.conf"))
+	if !success {
+		result.Success = success
+		return *result, err
+	}
+
+	// Write resolver config to host
+	success, err = createResolverFile(serviceConfig.IP, filepath.Join("/", "etc", "resolver", serviceConfig.BundleMetadata.ClusterInfo.BaseDomain))
```
anjannath: We can update the `resolverFile` variable with this path

gbraad:
Will review ASAP

pkg/crc/network/template.go
```diff
@@ -7,13 +7,13 @@ import (
 
 const (
 	resolvFileTemplate = `# Generated by CRC
-search {{ range .SearchDomains }}{{ .Domain }}{{ end }}
+{{ range .SearchDomains }}search {{ .Domain }}{{ end }}
 {{ range .NameServers }}nameserver {{ .IPAddress }}
```
cfergeau: Why is this changed?

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -146,3 +147,27 @@ func fixOcBinaryCached() (bool, error) {
...
+		return false, err
+	}
+
+	stdOut, stdErr, err := crcos.RunWithPrivilege("chown", currentUser.Username, resolvFile)
+	if err != nil {
+		return false, fmt.Errorf("Unable to change ownership of the resolver file: %s %v: %s", stdOut, err, stdErr)
+	}
+
+	return true, nil
+}
```
cfergeau: You could introduce 2 helpers to share the code with `checkResolverFilePermissions` and `fixResolverFilePermissions`

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -28,8 +28,15 @@ type resolverFileValues struct {
 }
 
 func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *services.ServicePostStartResult) (services.ServicePostStartResult, error) {
-	// Write resolver file
-	success, err := createResolverFile(serviceConfig.IP, filepath.Join("/", "etc", "resolver", serviceConfig.BundleMetadata.ClusterInfo.BaseDomain))
+	// Update resolv.conf file for host
+	success, err := updateResolvConfFile(serviceConfig.IP, filepath.Join("/", "etc", "resolv.conf"))
```
cfergeau: Is there a bug filed against `oc` to ask them to use the native macos APIs? (though I suspect this comes from go stdlib).

pkg/crc/network/template.go
praveenkumar: @cfergeau this is changed because if there is no search in the `/etc/resolv.conf` then end result after processing the template will be `search ` without anything and that will cause issue for command like dig to parse it.

pkg/crc/preflight/preflight_checks_darwin.go
praveenkumar: Yes, this refactor I can do 👍

pkg/crc/services/dns/dns_darwin.go
praveenkumar: No, I didn't file it but as you said it is going to come from go stdlib, not sure if filling the issue there will make any sense.

cfergeau: On the other hand, adding a workaround in our code without noticing other parties of the bug means this will never get fixed in the right place :(

pkg/crc/network/template.go
cfergeau: Ah I see. This could almost go in a separate commit as this is an unrelated bugfix.. Is this also behaving as expected when multiple search domains are in use?

praveenkumar: yes now it behave as expected even multiple search domains are in use.

pkg/crc/services/dns/dns_darwin.go
praveenkumar: @cfergeau it is not only the `oc` even dig doesn't honour the resolver for mac, but I do get your point will try to create an issue on origin side.

cfergeau: On my machine, it's only `oc` and `dig` which are using `resolv.conf`, `ping` for example does not seem to rely on it.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -146,3 +125,40 @@ func fixOcBinaryCached() (bool, error) {
...
+func addFileWritePermissionToUser(filename string) (bool, error) {
+	logging.DebugF("Making %s readable/writable by the current user", filename)
+	currentUser, err := user.Current()
+	if err != nil {
+		return false, err
+	}
+
+	stdOut, stdErr, err := crcos.RunWithPrivilege("chown", currentUser.Username, filename)
+	if err != nil {
+		return false, fmt.Errorf("Unable to change ownership of the resolver file: %s %v: %s", stdOut, err, stdErr)
```
cfergeau: `resolver` should be removed or replaced with `filename`

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -146,3 +125,40 @@ func fixOcBinaryCached() (bool, error) {
...
+	}
+
+	err = os.Chmod(filename, 0644)
+	if err != nil {
+		return false, fmt.Errorf("Unable to change permissions of the resolver file: %s %v: %s", stdOut, err, stdErr)
```
cfergeau: Same comment here

PR #256, opened 2019-06-21
Add `crc status` command

anjannath:
```
$ crc status
CRC VM:          Stopped
OpenShift:       Stopped
Cache Usage:     9.835GB
Cache Directory: /home/anjan/.crc/cache
```

```
$ crc status
CRC VM:          Running
OpenShift:       Running (v4.x)
Disk Usage:      10.09GB of 32.2GB (Inside the CRC VM)
Cache Usage:     9.835GB
Cache Directory: /home/anjan/.crc/cache
```

### DIFF
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -1,0 +1,78 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/docker/go-units"
+	"github.com/spf13/cobra"
+	"os"
+	"path/filepath"
+	"text/template"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/machine"
+)
+
+func init() {
+	rootCmd.AddCommand(statusCmd)
+}
+
+var statusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Display status of the cluster",
+	Long:  "Show details about OpenShift cluster and crc vm",
+	Run: func(cmd *cobra.Command, args []string) {
+		runStatus()
+	},
+}
+
+var statusFormat = `CRC VM:          {{.CrcStatus}}
+OpenShift:       {{.OpenShiftStatus}}
+Disk Usage:      {{.DiskUsage}}
+Cache Usage:     {{.CacheUsage}}
+Cache Directory: {{.CacheDir}}
+`
+
+type Status struct {
+	CrcStatus       string
+	OpenShiftStatus string
+	DiskUsage       string
+	CacheUsage      string
+	CacheDir        string
+}
+
+func runStatus() {
+	statusConfig := machine.ClusterStatusConfig{Name: constants.DefaultName}
+	clusterStatus, err := machine.Status(statusConfig)
+	if err != nil {
+		errors.ExitWithMessage(1, "Error while getting cluster status: %v", err)
+	}
+	cacheDir := constants.MachineCacheDir
+	var size int64
+	err = filepath.Walk(cacheDir, func(_ string, info os.FileInfo, err error) error {
+		if !info.IsDir() {
+			size += info.Size()
+		}
+		return err
+	})
+	if err != nil {
+		errors.ExitWithMessage(1, fmt.Sprintf("Error finding size of cache: %s", err.Error()))
+	}
+	cacheUsage := units.HumanSize(float64(size))
+	diskUse := units.HumanSize(float64(clusterStatus.DiskUse))
+	diskSize := units.HumanSize(float64(clusterStatus.DiskSize))
+	diskUsage := fmt.Sprintf("%s of %s (Inside the CRC VM)", diskUse, diskSize)
+	status := Status{clusterStatus.CrcStatus, clusterStatus.OpenshiftStatus, diskUsage, cacheUsage, cacheDir}
+	printStatus(status, statusFormat)
+}
+
+func printStatus(status interface{}, statusFormat string) {
+	tmpl, err := template.New("status").Parse(statusFormat)
+	if err != nil {
+		errors.ExitWithMessage(1, fmt.Sprintf("Error creating status template: %s", err.Error()))
+	}
+	err = tmpl.Execute(os.Stdout, status)
+	if err != nil {
+		errors.ExitWithMessage(1, fmt.Sprintf("Error executing status template:: %s", err.Error()))
+	}
+}
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -355,13 +355,65 @@ func Ip(ipConfig IpConfig) (IpResult, error) {
 		result.Error = err.Error()
 		return *result, err
 	}
-
 	if result.IP, err = host.Driver.GetIP(); err != nil {
 		result.Success = false
 		result.Error = err.Error()
 		return *result, err
 	}
+	return *result, nil
+}
 
+func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
+	result := &ClusterStatusResult{Name: statusConfig.Name, Success: true}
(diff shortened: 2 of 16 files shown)


cmd/crc/cmd/status.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			errors.ExitWithMessage(1, err.Error())
+		}
+		state, err := host.Driver.GetState()
+		if err != nil {
+			errors.ExitWithMessage(1, fmt.Sprintf("Error fetching cluster status: %s", err.Error()))
+		}
+	}*/
+
+	openshiftStatus := "Stopped"
```
gbraad: I do not need to look further. I already see there is too much logic in wrong level,

anjannath: Right, i need to push the new changes, i was waiting for #237 to get in, since i am using the `oc.ClusterOperatorsAvailable` function from that PR.

gbraad: unfortunately this one was postponed due to an issue with the operator state that the cluster returns after the API server is just up; this state is unreliable.

We would look into the situation and try to clean this state... if not, something else needs to be done :-s

anjannath: I have requested to split out the `oc.ClusterOperatorStatus` related changes in a separate commit in that PR (https://github.com/code-ready/crc/pull/242#issuecomment-509517560), the wait period related discussion is independent of those changes, as that is only getting the status of the cluster operators and returning that.

cmd/crc/cmd/status.go
```diff
@@ -0,0 +1,73 @@
...
+	Short: "Display status of the cluster",
+	Long:  "Show details about OpenShift cluster and crc vm",
+	Run: func(cmd *cobra.Command, args []string) {
+		runStatus()
+	},
+}
+
+var statusFormat = `CRC VM:     {{.CrcStatus}}
+OpenShift:  {{.OpenShiftStatus}}
+DiskUsage:  {{.DiskUsage}}
```
cfergeau: Make it explicit that this is the Disk Usage in the VM? Is this going to be printed on stdout? If yes, this should contain spaces ("Disk Usage", "Cache Usage")

cmd/crc/cmd/status.go
```diff
@@ -0,0 +1,73 @@
...
+CacheUsage: {{.CacheUsage}} (used by oc binary, Disk image)
```
cfergeau: Is the output of the status command a good place to indicate where the crc cache directory is? 
"used by oc binary, Disk image" belongs in the documentation I think, as this needs more details for better understanding.

pkg/crc/machine/machine.go
```diff
@@ -328,13 +330,59 @@ func Ip(ipConfig IpConfig) (IpResult, error) {
...
+	if err != nil {
+		result.Success = false
+		result.Error = err.Error()
+		return *result, err
+	}
+
+	if vmStatus == state.Running {
+		// check if all the clusteroperators are running
+		ocConfig := oc.UseOCWithConfig(statusConfig.Name)
+		cosAvailable, err := oc.GetClusterOperatorStatus(ocConfig)
```
cfergeau: imo the "cosAvailable" name is a bit misleading. `oc.GetClusterOperatorStatus(ocConfig)` returns true when all operators are ready. I read `cosAvailable` as something meaning "could we run `oc get co` on the cluster?" (but maybe that's me who is misreading it).

pkg/crc/machine/machine.go
```diff
@@ -328,13 +330,59 @@ func Ip(ipConfig IpConfig) (IpResult, error) {
...
+			result.Error = err.Error()
+			return *result, err
+		}
+		if cosAvailable {
+			// TODO:get openshift version as well and add to status
+			openshiftStatus = fmt.Sprintf("Running (v4.x)")
+		}
+
+		diskSize, diskUse, mountpoint := getDiskUsage(host.Driver, "/dev/vda3")
+		diskUsage = fmt.Sprintf("%s of %s (Mounted On: %s)", diskUse, diskSize, mountpoint)
```
cfergeau: Do we really need the mountpoint?
I'd prefer that machine.Status returns the raw data, and then `cmd/crc/cmd/status.go` will do the string formatting. This means changing `ClusterStatusResult` to have `DiskUse int64; DiskSize int64; MountPoint string;` (pseudo code)

pkg/crc/machine/machine.go
```diff
@@ -401,3 +449,20 @@ func unsetMachineLogging() {
...
+
+func getDiskUsage(driver drivers.Driver, mountpoint string) (string, string, string) {
+	cmd := fmt.Sprintf(
+		"df -h %s | awk 'FNR > 1 {print $2,$5,$6}'",
+		mountpoint)
+
+	out, err := drivers.RunSSHCommandFromDriver(driver, cmd)
+
+	if err != nil {
+		return "", "ERR", ""
```
cfergeau: This should return a normal go error

cmd/crc/cmd/status.go
anjannath: Yes, this is printed to stdout, the output looks like the following, which are separated by new lines. So we don't need the commas.
```
$ crc status                          
CRC VM:     Running
OpenShift:  Stopped
DiskUsage:  31% of 30G (Mounted On: /)
CacheUsage: 9.835GB (used by oc binary, Disk image)
```
Will mention that the disk usage is in the VM.

cmd/crc/cmd/status.go
anjannath: Yes we can add that information here..

pkg/crc/machine/machine.go
anjannath: I did not modify this function, this is taken from the minishift code base, will change it accordingly

pkg/crc/machine/machine.go
anjannath: Hmm.. would `operatorsRunning` be more meaningful?

jsliacan:
@anjannath If I understand correctly, the status command will either return "Running" + info, or "Stopped". If I followed through the code correctly, then if some COs aren't available, the status will be "Stopped" - please correct me if I made a mistake. If I'm right, then I'm not sure that this is desirable. It would be inconsistent with, say, using `crc stop` at that point whose output wouldn't be that the instance is already stopped.

Also, what happens if `crc status` is issued but CRC has been deleted? Just the usual `Machine "crc" does not exist. Use "crc start" to add a new one.`?

anjannath:
@jsliacan So if the desired cluster operators are not available it'd show OpenShift status as stopped, but the crc vm status would be running, Example o/p is here https://github.com/code-ready/crc/pull/256#discussion_r303720357

> Also, what happens if crc status is issued but CRC has been deleted? Just the usual Machine "crc" does not exist. Use "crc start" to add a new one.?

Yes, Correct!

pkg/crc/machine/machine.go
cfergeau: yes, looks good to me!

pkg/crc/machine/machine.go
```diff
@@ -355,13 +355,65 @@ func Ip(ipConfig IpConfig) (IpResult, error) {
...
-
 	if result.IP, err = host.Driver.GetIP(); err != nil {
 		result.Success = false
 		result.Error = err.Error()
 		return *result, err
 	}
+	return *result, nil
+}
+
+func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
```

pkg/crc/machine/machine.go
```diff
@@ -355,13 +355,65 @@ func Ip(ipConfig IpConfig) (IpResult, error) {
...
+	if vmStatus == state.Running {
+		// check if all the clusteroperators are running
+		ocConfig := oc.UseOCWithConfig(statusConfig.Name)
+		operatorsRunning, err := oc.GetClusterOperatorStatus(ocConfig)
+		if err != nil {
+			result.Success = false
+			result.Error = err.Error()
+			return *result, err
+		}
+		if operatorsRunning {
```

Issue #260, opened 2019-06-25
Implement the `crc IP` command.


praveenkumar:
We need to implement `crc IP` command.

cfergeau:
Why do we want such a command? In general, we don't want the user to directly access the VM, and if they want to do it, it should have DNS entries (api.crc.testing comes to mind). If needed, we can easily add other names.

praveenkumar:
@cfergeau most of the time when user try to deploy an application and use `NodePort` service then they always require the node IP (in our case it's VM IP), I got few messages from user about it so thought of having it also part of crc. This is trivial command and if it provide some value to user then having it would be good.

- https://kubernetes.io/docs/concepts/services-networking/service/#nodeport

praveenkumar:
This is fixed by #261

PR #261, opened 2019-06-25
Issue #260 Add IP command for crc

praveenkumar:


pkg/crc/machine/machine.go
```diff
@@ -304,6 +304,32 @@ func Delete(deleteConfig DeleteConfig) (DeleteResult, error) {
...
+	libMachineAPIClient := libmachine.NewClient(constants.MachineBaseDir, constants.MachineCertsDir)
+	host, err := libMachineAPIClient.Load(ipConfig.Name)
+
+	if err != nil {
+		result.Success = false
+		result.Error = err.Error()
+		return *result, err
+	}
+
+	if result.IP, err = host.Driver.GetIP(); err != nil {
```
cfergeau: I'd put `result.IP, err = host.Driver.GetIP()` on a separate line outside of the `if` block, but maybe that's a matter of preference.

Issue #248, opened 2019-06-18
libvirt: Revise host dnsmasq configuration


cfergeau:
Now that we are running dnsmasq in the `crc` VM, we can have a simpler dnsmasq configuration on the host ( `server=/.testing/192.168.130.11` is probably enough), or we could directly add the dns server to `/etc/resolv.conf`

gbraad:
@zeenix This is actually the issue we referenced. For this to work, we can first just use the DNSmasq server as running inside the VM. Instead of having to rely on NetworkManager to use a dnsmasq overlay, we can just use a simple entry to add the VM as a nameserver.

gbraad:
I will create follow-up tasks for dealing with `systemd-resolv`, but we have also concluded that this is ONLY the default on headless setups; Ubuntu desktops should still use NetworkManager as the default.

zeenix:
> I will create follow-up tasks for dealing with `systemd-resolv`, but we have also concluded that this is ONLY the default on headless setups; Ubuntu desktops should still use NetworkManager as the default.

Yeah, i don't think we need to even think about supporting Ubuntu server/headless right now so I'd suggest not creating a task for that one but up to you.

PR #264, opened 2019-06-25
Issue #248 Use dns server from VM to resolve domain in linux.

praveenkumar:
Now that we are running dnsmasq in the crc VM, we can have a simpler dnsmasq configuration on the host to directly add the dns server to /etc/resolv.conf

Don't spend time on review this change right now, I still have to figure out some of the missing parts.

### DIFF
--- a/pkg/crc/machine/driver_linux.go
+++ b/pkg/crc/machine/driver_linux.go
@@ -7,11 +7,10 @@ import (
 
 func init() {
 	LibvirtDriver := MachineDriver{
-		Name:          "Libvirt",
-		Platform:      crcos.LINUX,
-		Driver:        "libvirt",
-		UseDNSService: false,
-		DriverPath:    constants.CrcBinDir,
+		Name:       "Libvirt",
+		Platform:   crcos.LINUX,
+		Driver:     "libvirt",
+		DriverPath: constants.CrcBinDir,
 	}
 
 	SupportedDrivers = []MachineDriver{
--- a/pkg/crc/machine/libvirt/templates.go
+++ b/pkg/crc/machine/libvirt/templates.go
@@ -13,7 +13,7 @@ const (
 	<mac address='52:54:00:fd:be:d0'/>
 	<ip family='ipv4' address='192.168.130.1' prefix='24'>
 	  <dhcp>
-		<host mac='{{ .MAC }}' name='{{ .HostName }}' ip='{{ .IP }}'/>
+		<host mac='{{ .MAC }}' ip='{{ .IP }}'/>
 	  </dhcp>
 	</ip>
   </network>`
@@ -21,7 +21,6 @@ const (
 
 type NetworkConfig struct {
 	NetworkName string
-	HostName    string
 	MAC         string
 	IP          string
 }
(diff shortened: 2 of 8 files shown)


pkg/crc/machine/driver_linux.go
```diff
@@ -9,7 +9,7 @@ func init() {
 		Name:          "Libvirt",
 		Platform:      crcos.LINUX,
 		Driver:        "libvirt",
-		UseDNSService: false,
+		UseDNSService: true,
```
gbraad: since this is true, we might as well drop this option as it is always true now.

pkg/crc/machine/libvirt/templates.go
```diff
@@ -12,25 +12,9 @@ const (
...
-	  <srv service='etcd-server-ssl' protocol='tcp' domain='crc.testing' target='etcd-0.crc.testing' port='2380' weight='10'/>
-	  <host ip='{{ .IP }}'>
-		<hostname>api.crc.testing</hostname>
-		<hostname>api-int.crc.testing</hostname>
-		<hostname>etcd-0.crc.testing</hostname>
-	  </host>
-	</dns>
-	<ip family='ipv4' address='192.168.130.1' prefix='24'>
-	  <dhcp>
-		<host mac='{{ .MAC }}' name='{{ .HostName }}' ip='{{ .IP }}'/>
```
gbraad: This is necessary to make sure we assign a 'static' IP address.

pkg/crc/machine/libvirt/templates.go
```diff
@@ -12,25 +12,9 @@ const (
...
-	  </dhcp>
-	</ip>
   </network>`
 )
 
 type NetworkConfig struct {
 	NetworkName string
-	HostName    string
-	MAC         string
-	IP          string
```
gbraad: So you want to drop the preditable 'static' (or predefined) IP address?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"golang.org/x/sys/unix"
```
gbraad: should be defined below the golang framework imports, but before our own. right?

pkg/crc/services/dns/dns_linux.go
```diff
@@ -1,12 +1,57 @@
...
+	hostResolv, err := network.GetResolvValuesFromHost()
+	if err != nil {
+		return false, fmt.Errorf("Unable to read host resolv file (%v)", err)
+	}
+
+	foundExistingInstanceIP := false
+
+	for i, ns := range hostResolv.NameServers {
+		// Update the nameserver IP with instance IP if already exist
+		if strings.Contains(ns.IPAddress, "192.168.130") {
```
gbraad: contains or startswith?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -296,6 +296,8 @@ func checkLibvirtCrcNetworkAvailable() (bool, error) {
 			}
 			logging.Debug("libvirt 'crc' network definition contains ", constants.DefaultHostname)
 			return true, nil
+
+			return true, nil
```
cfergeau: This change seems unneeded here

pkg/crc/machine/driver_linux.go
cfergeau: Agreed.

zeenix:
I've an alternative solution that still takes the first 3 patches from here but uses `nmcli` to add the dns server to NM instead. I haven't managed to fully test it yet but hopefully I'll be able to create a PR in a few hours.

Issue #272, opened 2019-06-27
Put info about bundle version in the output.


praveenkumar:
Right now most of the time folks are using new release but an older version of the bundle which might have cert expired, we need to put the bundle version in the output as info so when they share the logs with us, we know for sure that an older version is used.

praveenkumar:
This is fixed by #27

PR #273, opened 2019-06-27
Issue #272 Verify the user provide bundle against released one

praveenkumar:
Most of the time user using older version of crc bundle with new
release and this give an unexpected error to the user. We always
need to use the bundle version which released as part of the release.
This check make sure user is going to use the updated one.

Minor refactor around getting bundle info as part of version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 BUNDLE_VERSION = 4.1.0
-CRC_VERSION = 0.87.0-alpha-$(BUNDLE_VERSION)
+CRC_VERSION = 0.87.0-alpha
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
 
 # Go and compilation related variables
@@ -37,6 +37,7 @@ PACKAGES := go list ./... | grep -v /out
 
 # Linker flags
 VERSION_VARIABLES := -X $(REPOPATH)/pkg/crc.crcVersion=$(CRC_VERSION) \
+    -X $(REPOPATH)/pkg/crc.bundleVersion=$(BUNDLE_VERSION) \
 	-X $(REPOPATH)/pkg/crc.commitSha=$(COMMIT_SHA)
 
 # https://golang.org/cmd/link/
--- a/cmd/crc/cmd/version.go
+++ b/cmd/crc/cmd/version.go
@@ -37,5 +37,5 @@ var versionCmd = &cobra.Command{
 }
 
 func runPrintVersion(arguments []string) {
-	fmt.Printf("version: %s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetCommitSha())
+	fmt.Printf("version: %s-%s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetBundleVersion(), crcPkg.GetCommitSha())
 }
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -11,7 +11,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 
-	// host and instance related
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/systemd"
 	crcos "github.com/code-ready/crc/pkg/os"
@@ -53,7 +52,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		Memory:     startConfig.Memory,
 	}
 
-	logging.InfoF("Extracting the Bundle tarball ...")
+	logging.InfoF("Extracting the %s Bundle tarball ...", filepath.Base(machineConfig.BundlePath))
 	crcBundleMetadata, extractedPath, err := bundle.GetCrcBundleInfo(machineConfig)
 	if err != nil {
 		logging.ErrorF("Error to get bundle Metadata %v", err)
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -1,8 +1,12 @@
 package validation
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 
+	"github.com/code-ready/crc/pkg/crc"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/machine"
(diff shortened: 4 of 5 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 BUNDLE_VERSION = 4.1.0
-CRC_VERSION = 0.87.0-alpha-$(BUNDLE_VERSION)
+CRC_VERSION = 0.87.0-alpha
```
gbraad: CRC releases will somehow always be tied to a version, as we need to make sure a bundle can ebe brought up with the associated CRC version. Else we would end up having to test a multitude of version combinations. What is the idea here?

cmd/crc/cmd/version.go
```diff
@@ -37,5 +37,5 @@ var versionCmd = &cobra.Command{
 }
 
 func runPrintVersion(arguments []string) {
-	fmt.Printf("version: %s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetCommitSha())
+	fmt.Printf("version: %s-%s+%s\n", crcPkg.GetCRCVersion(), crcPkg.GetBundleVersion(), crcPkg.GetCommitSha())
```
gbraad: you want to be able to debug the version of the bundle used.... but this indicates another issue too,.. the CRC binary is not independent of the actual bundle it can support.

pkg/crc/machine/machine.go
```diff
@@ -53,14 +54,26 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 		return *result, err
 	}
 
+	// Verify, user is using latest bundle with released version of crc.
+	if constants.DefaultHostname != crcBundleMetadata.Nodes[0].Hostname {
+		hypervisor := machineConfig.VMDriver
+		if hypervisor == "virtualbox" {
+			hypervisor = "vbox"
+		}
+		errStr := fmt.Sprintf("%s bundle is not supported for this release use update one (crc_%s_%s.tar.xz)", filepath.Base(machineConfig.BundlePath), hypervisor, crc.GetBundleVersion())
```
gbraad: confirmed here...

s/update/updated

pkg/crc/machine/machine.go
```diff
@@ -53,14 +54,26 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+	logging.InfoF("Extracting the %s Bundle tarball ...", filepath.Base(machineConfig.BundlePath))
 	crcBundleMetadata, extractedPath, err := bundle.GetCrcBundleInfo(machineConfig)
 	if err != nil {
 		logging.ErrorF("Error to get bundle Metadata %v", err)
 		result.Error = err.Error()
 		return *result, err
 	}
 
+	// Verify, user is using latest bundle with released version of crc.
+	if constants.DefaultHostname != crcBundleMetadata.Nodes[0].Hostname {
```
gbraad: this is not a valid check, as it assume a single situation.

pkg/crc/machine/machine.go
```diff
@@ -53,14 +54,26 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+		hypervisor := machineConfig.VMDriver
+		if hypervisor == "virtualbox" {
+			hypervisor = "vbox"
```
gbraad: why is this necessary?

pkg/crc/machine/machine.go
gbraad: feels more like a semantic version scheme is needed...

pkg/crc/version.go
```diff
@@ -23,6 +23,9 @@ var (
 
 	// The SHA-1 of the commit this binary is build off
 	commitSha = "sha-unset"
+
+	// Bundle version which used for the release.
+	bundleVersion = "0.0.0-unset"
```
gbraad: OK, here we set the value of the bundle statically for the binary

cmd/crc/cmd/version.go
gbraad: So a static version is used... but it is not enforced in the code?

pkg/crc/machine/machine.go
gbraad: OK, as I understand... you check the cached/running instance node hostname to the one from the bundle... right? If so, make this more obvious by extracting this in a named function,

gbraad:
the intention of what is described in the first message is not clear from the code. I would suggest to extract some of the logic in more clerly named functions; `ValidateVersionInstanceWithImageBundle`, or `ValidateBundleVersions() { return GetInstanceBundleVersion() == GetImageBundleVersion() }` or something similar.

gbraad:
Also not sure why the virtualdriver is overriden here from `virtualbox` to `vbox` ?

Makefile
praveenkumar: @gbraad idea is every release support the released version of bundle only (not older or newer). we still are struggling with the 30 days cert rotation which is not yet resolved for us. Most of the user are trying a older bundle version with latest version of crc and getting issue around the cert expire.

cmd/crc/cmd/version.go
praveenkumar: @gbraad I didn't get that comment, we had it part of Makefile till now and this is just I made it part of codebase.

pkg/crc/machine/machine.go
praveenkumar: @gbraad because our bundle contain `vbox` instead `virtualbox` in the tarball, we can either change it there, make part of snc script or here.

gbraad: this is not the place to change/override it... this feels like a workaround value.

@code-ready/crc-devel WDYT?

cmd/crc/cmd/version.go
gbraad: See https://github.com/code-ready/crc/pull/273#issuecomment-507158507

pkg/crc/machine/machine.go
praveenkumar: @gbraad yes this is workaround, I will create an issue on snc side so we can generated the tarball with hypervisor next time, Will put Note here meanwhile.

praveenkumar: @gbraad https://github.com/code-ready/snc/issues/56 here

gbraad: If this is approved and merged, will we eventually 'remove' this with the next release?

praveenkumar: @gbraad yes we will remove that part from next release.

pkg/crc/validation/validation.go
```diff
@@ -39,5 +43,11 @@ func ValidateBundle(bundle string) error {
 	if _, err := os.Stat(bundle); os.IsNotExist(err) {
 		return errors.NewF("Expected file %s does not exist", bundle)
 	}
+	// Check if the version of the bundle provided by user is same as what is released with crc.
+	releaseBundleVersion := crc.GetBundleVersion()
+	userProvidedBundleVersion := filepath.Base(bundle)
+	if !strings.Contains(userProvidedBundleVersion, fmt.Sprintf("%s.tar.xz", releaseBundleVersion)) {
```
gbraad: slightly harder to read, but you are constructing the bundle path to compare.

PR #274, opened 2019-06-28
Issue #218 Approve node csr after cluster is up

### LINKED ISSUE #218: Alpha-2 (0.87.0): oc logs: remote error: tls: internal error
log can not be viewed...

How to reproduce:
```
$ oc new-project demo
$ oc process openshift//eap72-mongodb-s2i|oc apply -f -
$ oc get pod
NAME                       READY     STATUS      RESTARTS   AGE
eap-app-1-build            1/1       Running     0          3m34s
eap-app-mongodb-1-deploy   0/1       Completed   0          3m33s
eap-app-mongodb-1-r7827    1/1       Running     0          3m21s
$ oc logs eap-app-1-build
Error from server: Get https://crc-4gdnp-master-0:10250/containerLogs/demo/eap-app-1-build/sti-build: remote error: tls: internal error
$ oc logs eap-app-mongodb-1-r7827
Error from server: Get https://crc-4gdnp-master-0:10250/containerLogs/demo/eap-app-mongodb-1-r7827/eap-app-mongodb: remote error: tls: internal error
```

praveenkumar:
Fixes #218

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -15,7 +15,9 @@ import (
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/systemd"
 	crcos "github.com/code-ready/crc/pkg/os"
+
 	// cluster services
+	"github.com/code-ready/crc/pkg/crc/oc"
 	"github.com/code-ready/crc/pkg/crc/services"
 	"github.com/code-ready/crc/pkg/crc/services/dns"
 
@@ -214,7 +216,6 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		logging.InfoF("Starting OpenShift cluster ... [waiting 3m]")
 	}
 	result.KubeletStarted = kubeletStarted
-	//
 
 	// If no error, return usage message
 	if result.Error == "" {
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/machine.go
```diff
@@ -224,6 +225,14 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		logging.InfoF("Login to the console with user: kubeadmin, password: %s", result.ClusterConfig.KubeAdminPass)
 	}
 
+	// Approve the node certificate.
+	crcOc := oc.GetOC(machineConfig.Name)
```
gbraad: just reading this fragment of code makes me wonder what it does:

`get OC for an named instance of CRC` ?... why is `oc` associated with an instance?

pkg/crc/oc/oc.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/code-ready/crc/pkg/crc/constants"
+	crcos "github.com/code-ready/crc/pkg/os"
+)
+
+type OC struct {
+	OcBinaryPath   string
+	KubeconfigPath string
+}
+
+func GetOC(machineName string) OC {
```
gbraad: so, the function name is too ambigious. it should be something like `UseOCWithConfig`, right?

pkg/crc/oc/oc.go
```diff
@@ -0,0 +1,48 @@
...
+	oc := OC{
+		OcBinaryPath:   filepath.Join(constants.OcCacheDir, constants.OcBinaryName),
+		KubeconfigPath: filepath.Join(constants.MachineInstanceDir, machineName, "kubeconfig"),
```
gbraad: so the `name` is used for the kubeconfig path... this means you do not get oc, but get a configuration.

pkg/crc/oc/oc.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	return oc
+}
+
+func (oc OC) RunOcCommand(args ...string) (string, string, error) {
+	args = append(args, "--kubeconfig", oc.KubeconfigPath)
+	return crcos.RunWithDefaultLocale(oc.OcBinaryPath, args...)
+}
+
+func ApproveCSR(oc OC) error {
```
gbraad: `ApproveCSR` for an `OC Config`? ?

gbraad: When we talk about `oc` it feels like the command, but here we talk about an `environment` or `configuration`. I would prefer to see that in the naming to avoid this confusion

cfergeau:
Can you explain in the commit log why this is needed, and if this is the right fix long-term, or just a workaround while the root cause is found/fixed?

pkg/crc/oc/oc.go
praveenkumar: Ah it should be `ApproveNodeCSR` since we are approving the node csr value.

gbraad: also... but it is more how `OC` is passed in that sounds merely like a `oc` command wrapper, while it is a configed instance of the `oc` command, right? What aboud `OCConfig`.

Issue #278, opened 2019-06-28
Document install dependency before run `make build_docs`


eduardocerqueira:
In Building the documentation section from current README.md is not explicit the required dependency of `podman` in the system before generate the local documentation through:

```
make build_docs
```

It is a simple and easy issue to be fixed and can safe few minutes of some new developer or someone now familiar with podman.

robin-owen:
Fixed by #278.

PR #279, opened 2019-06-28
Issue #278 install podman before make build_docs

eduardocerqueira:
For new contributors and people not familiar with podman this small change in the README will make thigs a little clear when generating the locall documentation


praveenkumar:
ok to test

PR #280, opened 2019-06-29
[WIP] Add hyperkit support

praveenkumar:


pkg/crc/machine/hyperkit/driver.go
```diff
@@ -0,0 +1,47 @@
...
+		CPU:         config.CPUs,
+		UUID:        uuid.NewUUID().String(),
+		Cmdline:     config.KernelCmdLine,
+		VmlinuzPath: config.Kernel,
+		InitrdPath:  config.Initramfs,
+		// This force to add entry of DiskPath under crc machine config.json
+		DiskImage:   filepath.Join(constants.MachineBaseDir, "machines", config.Name, constants.DefaultName),
+		DiskPathUrl: config.DiskPathURL,
+		SSHKeyPath:  config.SSHKeyPath,
+		Hyperkit:    "/Users/prkumar/bin/hyperkit",
```
praveenkumar: This need to be changed instead of hard coded.

gbraad: can not merge or approve this... what is the idea to do with this?

cfergeau: This needs to be handled at crc setup time, hyperkit will be downloaded to the cache directory

praveenkumar: @cfergeau That's right.

gbraad: we do not need to install any of the drivers to a system path...

... we did with libvirt, but we can force them to be started from the `crc` homedir from our code anyways. This would avoid the need to use admin rights for this function and also simplifies a possible cleaning step.

praveenkumar: @gbraad yes, I think this can be done, we need to register the driver path I think, need to take a look but I hope it is possible. But this one here is not about the driver it actually the `hyperkit` binary since brew one doesn't support what we want even binary which comes with `docker-for-mac` which is the older version. I am not sure why they are not updating it to brew.

gbraad: It is possible... in one of the earliest commits I changed the binary names from `machine` to `crc`. this also includes the path lookups. We would just include or change this to `$CRCHOME/driver/` for instance and be done. also, we should avoid any clashes this way, as it would avoid people using different versions all over the place.

brew? I would like to avoid htis as much as possible... this should be for just the developer, and not for our runtime, right? else we are dependent of them updating our components (or the user to do this)... rather have this under our control from start to finish. And it seems docker-for-mac also does, as they do not bother with their uploads to brew either. all comes from a dmg.

pkg/crc/machine/hyperkit/driver.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/machine/config"
+	"github.com/code-ready/machine/libmachine/drivers"
+	"github.com/pborman/uuid"
+	"path/filepath"
+)
+
+type hyperkitDriver struct {
+	*drivers.BaseDriver
+	DiskImage     string
```
cfergeau: Currently, the hyperkit driver is not using DiskImage, I think it's better if the exact location of the disk image is internal to the driver, so it can be removed.

pkg/crc/machine/hyperkit/driver.go
```diff
@@ -0,0 +1,47 @@
...
+	Cmdline       string
+	UUID          string
+	VpnKitSock    string
+	VSockPorts    []string
+	VmlinuzPath   string
+	InitrdPath    string
+	KernelCmdLine string
+	DiskPathUrl   string
+	SSHKeyPath    string
+	Hyperkit      string
```
cfergeau: I'd prefer to name this HyperKitPath (I've changed the driver accordingly)

cfergeau:
These changes are now part of PR #302, this one can be closed I think?

praveenkumar:
@cfergeau ack, closing this.

Issue #281, opened 2019-07-01
Add driver option to setup command


praveenkumar:
Right now we are only have setup command for just one hypervisor but for Mac/Linux we have virtualbox and hyperkit/HyperV so we do need to expose a flag on setup command so we can install requisites as per hypervisor.

anjannath:
Fixed in https://github.com/code-ready/crc/commit/56eb5cf94ba1439ff095681570c384f20aa6c239

PR #282, opened 2019-07-01
Issue #281 Add vm driver flag for setup command

praveenkumar:
Right now we are doing setup for single hypervisor but in
case of mac we have virtualbox and hyperkit supported so
we need to add setup functionality for both hypervisor and
to do we need a driver flag using that we can make decision
about which hypervisor we need to set.

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -42,7 +42,7 @@ func runStart(arguments []string) {
 		errors.Exit(1)
 	}
 
-	preflight.StartPreflightChecks()
+	preflight.StartPreflightChecks(crcConfig.GetString(config.VMDriver.Name))
 
 	startConfig := machine.StartConfig{
 		Name:       constants.DefaultName,
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -1,11 +1,19 @@
 package cmd
 
 import (
-	"github.com/code-ready/crc/pkg/crc/preflight"
+	"fmt"
 	"github.com/spf13/cobra"
+
+	"github.com/code-ready/crc/cmd/crc/cmd/config"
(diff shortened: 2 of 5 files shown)


cmd/crc/cmd/start.go
```diff
@@ -42,7 +42,7 @@ func runStart(arguments []string) {
 		errors.Exit(1)
 	}
 
-	preflight.StartPreflightChecks()
+	preflight.StartPreflightChecks(crcConfig.GetString(config.VMDriver.Name))
```
gbraad: a `struct` to be passed in is more appropriate, as this would be in line with the rest of the functionality.

we already do this for the start command and therefore the same start struct COULD be mapped or even used.

cfergeau:
> I'd rather see a better integration with the rest of the code.
> 
> map against a struct that can be passed in from the cmd level, just like this is done for the start command.
> 
> this struct can then also be populated from the start command.

For what it's worth, I still don't think passing struct arounds (instead of more usual function calls with multiple arguments) is very good from a code readability perspective. I would go the opposite way, and try to remove these structs ;) So I would prefer not to introduce a struct here, and keep things as they are (or pass config.VMDriver rather than a string to avoid a string based API).

gbraad:
The interaction between cmd and pkg is almost always string based, as they
use the configuration from viper. although the intention of the struct is
at the moment less obvious and perhaps even unnecessary/unwanted it always
to prvent using config inside the pokg level and consistency... until we
decide to remove this.

cfergeau:
Forgot to mention it before, but this PR is part of the hyperkit PR #302 (it's a prerequisite of some of the work there).

anjannath:
This we already have in master 56eb5cf94ba1439ff095681570c384f20aa6c239, can be closed!

PR #285, opened 2019-07-02
cut 0.88.0

praveenkumar:


praveenkumar:
Wait for all CI before merge.

test/integration/features/config.feature
```diff
@@ -13,15 +13,15 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
         @darwin
         Examples: Config settings on Mac 
             | property  | value1                | value2                                     |
-            | bundle    | crc_vbox_4.1.0.tar.xz | /path/to/fake/bundle/crc_vbox_4.1.0.tar.xz |
+            | bundle    | crc_vbox_4.1.3.tar.xz | /path/to/fake/bundle/crc_vbox_4.1.3.tar.xz |
```
cfergeau: Fwiw, it would be great if we could substitude this 4.1.3 from the Makefile somehow, rather than having to change it in multiple places.

praveenkumar: @cfergeau, yes that we will do for sure, will create an follow up issue.

praveenkumar:
> Looks good to me.
> I think there was some confusion on osx with the last release, where the default virt driver was hyperkit rather than vbox, maybe it would be better to change it before releasing?

@cfergeau that was what we thought before but @gbraad was having a strong opinion not to do this way since we want to have default hypervisor from beginning to set user exceptions.

Issue #286, opened 2019-07-02
Bundle version for different place should be comes from a function or Makefile


praveenkumar:
Right now we have different places hard coded bundle version info which should be taken from a function or from makefile.

reference https://github.com/code-ready/crc/pull/285 changed we did here.

praveenkumar:
This is fixed by #289

PR #275, opened 2019-06-28
Add more integration tests

jsliacan:
This PR contains new integration tests. This is wip and I'll be glad to hear what to add/replace/remove/rework. 

- interaction with oc command (in `oc.feature`)
- end-to-end story creating an app (in `story.feature`)

The `oc` command interaction is only basic as we'll adopt OpenShift extended tests and we do not want to duplicate much work. The story is a very simple health check of the cluster (if an app can be deployed then at least something has to be working :) ). 

Thanks for feedback!

PR #289, opened 2019-07-02
Issue #286 Bundle version taken from Makefile

jsliacan:
This removes the hardcoded bundle version from a feature file. Instead, the bundle is now taken to be the one that CRC was started with.

### DIFF
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -44,7 +44,7 @@ func FeatureContext(s *godog.Suite) {
 		StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails)
 	s.Step(`^starting CRC with default bundle and hypervisor "(.*)" (succeeds|fails)$`,
 		StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails)
-	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds$|fails)$`,
+	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds|fails)$`,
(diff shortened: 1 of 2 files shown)


praveenkumar:
@jsliacan Changes looks good but then it not what I had in mind so might be change the commit log since your changes are not depend on the Makefile at all.

jsliacan:
> not depend on the Makefile at all

The bundle name is taken from the `BUNDLE_LOCATION` variable that's passed to the `make integration`. In #275 I also define a default `BUNDLE_LOCATION` inside the Makefile. So it will be directly from the Makefile then. Should I still change the commit message?

praveenkumar:
@jsliacan what I meant is the current changes in the PR, you might change the commit message like `Issue #286 Integration: Remove hard coded bundle value` and then put a bit explanation.

jsliacan:
@praveenkumar OK, it's updated now. Thanks!

test/integration/crcsuite/crcsuite.go
```diff
@@ -44,7 +44,7 @@ func FeatureContext(s *godog.Suite) {
 		StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails)
 	s.Step(`^starting CRC with default bundle and hypervisor "(.*)" (succeeds|fails)$`,
 		StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails)
-	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds$|fails)$`,
+	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds|fails)$`,
```
praveenkumar: 👍

PR #297, opened 2019-07-05
Issue #286 Add GetDefaultBundle function

praveenkumar:
Right now, all the bundle version are hardcoded to platform specific constants, this makes update to bundle version error prone.

### DIFF
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -2,11 +2,11 @@ package validation
 
 import (
 	"fmt"
+	"github.com/code-ready/crc/pkg/crc/version"
 	"os"
 	"path/filepath"
 	"strings"
 
-	"github.com/code-ready/crc/pkg/crc"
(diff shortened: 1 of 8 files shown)


pkg/crc/validation/validation.go
```diff
@@ -2,11 +2,11 @@ package validation
 
 import (
 	"fmt"
+	"github.com/code-ready/crc/pkg/crc/version"
```
cfergeau: shouldn't this go after the `crc/machine` import?

praveenkumar: @cfergeau No, we want to use `crc/machine` only for machine related stuff, this is independent of it.

cfergeau: I'm only saying that the `"github.com/code-ready/crc/pkg/crc/version"` line should be moved down, I'm not suggesting moving code in other packages.

PR #301, opened 2019-07-08
Create passwordless sudoer `crc_ci` for CentosCI

jsliacan:
To avoid running tests under `root`, create a `crc_ci` user with passwordless sudo privileges.

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# bundle location
+BUNDLE=crc_libvirt_4.1.3.tar.xz
+
 # Output command before executing
 set -x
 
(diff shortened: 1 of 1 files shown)


anjannath:
We can remove the previous unzip in: https://github.com/code-ready/crc/pull/301/files#diff-15c128a70ef5dd7aaf264bce566298fdL91

centos_ci.sh
```diff
@@ -88,7 +110,6 @@ function perform_artifacts_upload() {
 function get_bundle() {
   mkdir $HOME/Downloads
   curl -L 'https://github.com/code-ready/crc-ci-jobs/releases/download/4.0.1/crc_bundle.zip' -o $HOME/Downloads/bundle.zip
```
gbraad: Do we use this? If not, this is conflicting with what is in https://github.com/code-ready/crc/pull/301/files#diff-15c128a70ef5dd7aaf264bce566298fdR128 ?

jsliacan: We download `bundle.zip` to `/root/Downloads`, then copy it to `home/crc_ci/Downloads` after `crc_ci` is created, then unzip it there - this results in `/home/crc_ci/crc_libvirt_4.1.3.tar.xz`.

centos_ci.sh
```diff
@@ -104,28 +120,36 @@ function upload_logs() {
 
 function run_tests() {
   set +e
-  make integration BUNDLE_LOCATION="/root/Downloads/crc_libvirt_4.1.3.tar.xz"
+  make integration BUNDLE_LOCATION=$HOME/Downloads/crc_libvirt_4.1.3.tar.xz
```
gbraad: would prefer if we can extract this as a variable (global)

centos_ci.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# bundle location
+BUNDLE=$HOME/Downloads/crc_libvirt_4.1.3.tar.xz
```
gbraad: maybe JUST the file name?, ome and downloads is defined by the extract location

centos_ci.sh
```diff
@@ -104,28 +123,36 @@ function upload_logs() {
 
 function run_tests() {
   set +e
-  make integration BUNDLE_LOCATION="/root/Downloads/crc_libvirt_4.1.3.tar.xz"
+  make integration BUNDLE_LOCATION=$BUNDLE
```
gbraad: `/home/.../Downloads/$BUNDLEFILENAME`

PR #302, opened 2019-07-08
Add hyperkit support

cfergeau:
This supersedes PR#280. It's missing changes for `crc setup` in order to download the `crc-driver-hyperkit` and `hyperkit` binaries. If `hyperkit` is in /usr/local/bin and `crc-driver-hyperkit` is in the PATH, `crc start`/`crc stop`/`crc delete` should work fine.
The `Move generation of .crc-exist file to generic code` commit does not belong in this PR, but it was more convenient to have it here for now.


gbraad:
Draft and deployments? Something changed recently ;-)

cfergeau:
What I force-pushed still needs work, stop/... is not working properly yet.

pkg/crc/machine/machine.go
```diff
@@ -157,13 +157,22 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+	for i := 0; i < 30; i++ {
+		hostIP, err = network.DetermineHostIP(instanceIP)
+		if err == nil {
+			logging.DebugF("Bridge IP on the host: %s", hostIP)
+			break
+		} else {
+			logging.DebugF("Error finding host IP (%v) - retrying", err)
+		}
+		time.Sleep(2 * time.Second)
+	}
```
praveenkumar: @cfergeau `host.Driver.GetIP()` suppose to wait till the VM is up and then only provide the IP since the mac address which it looks suppose to be different for each run until we are passing a specific UUID => https://github.com/code-ready/machine-driver-hyperkit/blob/master/pkg/hyperkit/driver.go#L220

pkg/crc/machine/machine.go
```diff
@@ -91,6 +92,13 @@ func Start(startConfig StartConfig) (StartResult, error) {
 
 	// Pre-VM start
 	driverInfo, _ := getDriverInfo(startConfig.VMDriver)
+	if driverInfo.DriverPath != "" {
+		/* We need to add the directory containing crc-driver-xxx to the PATH
+		* so that libmachine can find it
+		*/
+		os.Setenv("PATH", fmt.Sprintf("%s:%s", driverInfo.DriverPath, os.Getenv("PATH")))
+		logging.InfoF("machine.Start(): driverInfo.DriverPath=%s PATH=%s", driverInfo.DriverPath, os.Getenv("PATH"))
```
praveenkumar: So here we are explicit setting the driver path to env. I though we make changes in libmachine side to take this driver path as argument and run it.

pkg/crc/machine/hyperkit/driver.go
```diff
@@ -31,8 +32,7 @@ func CreateHost(config config.MachineConfig) *hyperkitDriver {
 		},
 		Memory:      config.Memory,
 		CPU:         config.CPUs,
-		/* UUID will not be hardcoded in subsequent commits */
-		UUID:        "63655f17-7992-4f8e-8274-4aad9f7cf742",
+		UUID:        uuid.NewUUID().String(),
```
praveenkumar: so here we are generating random UUID which means it will also going to produce different mac address.

pkg/crc/machine/machine.go
gbraad: Have a look at: https://github.com/code-ready/machine/commit/59879005e2d35019745338a9d8e364ad6dd86bba

pkg/crc/constants/constants.go
```diff
@@ -34,6 +34,7 @@ var (
 	MachineCertsDir    = filepath.Join(MachineBaseDir, "certs")
 	MachineCacheDir    = filepath.Join(MachineBaseDir, "cache")
 	OcCacheDir         = filepath.Join(MachineCacheDir, "oc")
+	HyperKitCacheDir   = filepath.Join(MachineCacheDir, "hyperkit")
```
gbraad: ? What is this used for ?

If this is for the driver, I would rather see a generic directory ... `drivers`, but thi is not really 'cache'

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -85,6 +97,146 @@ func fixVirtualBoxInstallation() (bool, error) {
...
+	if fi.Sys().(*syscall.Stat_t).Uid != 0 {
+		return false, fmt.Errorf("%s is not owned by root", path)
+	}
+
+	return true, nil
+}
+
+func checkHyperKitInstalled() (bool, error) {
+	logging.DebugF("Checking if hyperkit is installed")
+	hyperkitPath := filepath.Join(hyperkitCacheDir, "hyperkit")
```
gbraad: Not suggested as the solution...

I would expect something like `.crc/drivers/`

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -85,6 +97,146 @@ func fixVirtualBoxInstallation() (bool, error) {
...
+	if err != nil {
+		logging.DebugF("%s not executable", hyperkitPath)
+		return false, err
+	}
+
+	return checkSuid(hyperkitPath)
+}
+
+func fixHyperKitInstallation() (bool, error) {
+	hyperkitFile, err := download(hyperkitDownloadURL, hyperkitCacheDir, 0755)
```
gbraad: shouldn't download happen to a temporary directory, verified against a hash, and then confirm it was downloaded correctly.

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -33,19 +48,36 @@ func StartPreflightChecks() {
...
-		"Setting up virtualization",
-		config.GetBool(cmdConfig.WarnCheckVirtualBoxInstalled.Name),
-	)
+
+	switch vmDriver {
+	case "virtualbox":
+		preflightCheckAndFix(config.GetBool(cmdConfig.SkipCheckVirtualBoxInstalled.Name),
+			checkVirtualBoxInstalled,
+			fixVirtualBoxInstallation,
+			"Setting up virtualization",
```
gbraad: I suggest to use the hypervisor name here too, to avoid ambiguity

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -33,19 +48,36 @@ func StartPreflightChecks() {
...
+			config.GetBool(cmdConfig.WarnCheckVirtualBoxInstalled.Name),
+		)
+	case "hyperkit":
+		preflightCheckAndFix(config.GetBool(cmdConfig.SkipCheckHyperKitInstalled.Name),
+			checkHyperKitInstalled,
+			fixHyperKitInstallation,
+			"Setting up virtualization",
```
gbraad: I suggest to use the hypervisor name here too, to avoid ambiguity

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -131,7 +131,7 @@ func updateResolvConfFile(instanceIP string, resolvConfFile string) (bool, error
 
 	for i, ns := range hostResolv.NameServers {
 		// Update the nameserver IP with instance IP if already exist
-		if strings.Contains(ns.IPAddress, "192.168.130") {
+		if strings.Contains(ns.IPAddress, "192.168.130") || strings.Contains(ns.IPAddress, "192.168.64") {
```
gbraad: can we be more specific about the IP address range and create constants to refer to?

pkg/crc/machine/machine.go
cfergeau: changing this on libmachine side is quite invasive as for now its API is based on machine names only. So I went for this approach for now as it's probably simpler. I'll look at changing libmachine API instead.

pkg/crc/machine/hyperkit/driver.go
cfergeau: We are generating a random UUID when `CreateHost` is called. This only happens on first `crc start` run, or after calling `crc delete`. If you only use `crc start`/`crc stop`, then we do not recreate the VM/regenerate a new UUID each run https://github.com/code-ready/crc/blob/master/pkg/crc/machine/machine.go#L88

pkg/crc/constants/constants.go
cfergeau: It's something which we can download from the net when needed, so it's some form of local binary cache. ACK for moving it some place else. `~/.crc/bin` ?

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: Yes most likely, though we currently don't do that for any binary that we download (virtualbox.dmg, oc, ...). Or maybe we could ship all of this in a .dmg?

pkg/crc/services/dns/dns_darwin.go
praveenkumar: @gbraad will be obsolete by #315

pkg/crc/constants/constants.go
cfergeau: Just remembered that I only did the same as what is done for OcCacheDir, the `oc` binary goes there. So maybe keep it for consistency, but open a bug to get this fixed in both cases.

pkg/crc/preflight/preflight_darwin.go
cfergeau: Fixed locally, this will be in the next update to this PR

pkg/crc/preflight/preflight_darwin.go
cfergeau: Fixed locally, this will be in the next update to this PR

pkg/crc/machine/machine.go
cfergeau: See below for the explanation why we do not generate a new UUID each run (if the VM already exists, we do not recreate it with a different UUID)

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -6,27 +6,39 @@ import (
...
 	"github.com/code-ready/crc/pkg/crc/oc"
+	dl "github.com/code-ready/crc/pkg/download"
 	crcos "github.com/code-ready/crc/pkg/os"
 )
 
 const (
 	virtualBoxDownloadURL   = "https://download.virtualbox.org/virtualbox/6.0.4/VirtualBox-6.0.4-128413-OSX.dmg"
 	virtualBoxMountLocation = "/Volumes/VirtualBox"
 
+	hyperkitDownloadURL       = "https://657-55985023-gh.circle-artifacts.com/0/hyperkit"
```
praveenkumar: @cfergeau I am just worry about how long this artifact going to stay on circle ci. According to the policy https://circleci.com/docs/2.0/artifacts/#artifacts-overview they also suggest not to rely on it for distribution side. Can we have it as part of hyperkit binary release or some other place which we own?

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -6,27 +6,39 @@ import (
...
+	hyperkitDriverDownloadURL = "https://teuf.fedorapeople.org/crc-driver-hyperkit"
```
praveenkumar: @cfergeau this will be replaced with proper release we do on the driver side right? Anything stoping to do that release?

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: I don't know what would be a good location to get the binary from? We probably should update it to something newer anyway. At this point, I'm thinking maybe it would make sense to include it in the crc_hyperkit_xxx.tar.xz tarball?
Do you have a suggestion for a place where we could upload the file?

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: Oh thanks for pointing this out, I had totally forgotten about it. Yep, definitely should be replaced by something more official. I have no idea what the workflow is to make a release, so that would be the main blocker ;) Apart from this, I think the current crc-driver-hyperkit git master is good for release.

pkg/crc/constants/constants.go
cfergeau: It turns out this constant was not even used ;) However there were similar variables elsewhere. Latest iteration of the code puts these in ~/.crc/bin/

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: I changed it to ~/.crc/bin/

pkg/crc/machine/machine.go
cfergeau: This is now done, see https://github.com/code-ready/machine/pull/13 
And the end result is actually simpler than what I was doing here with PATH changes.

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: I filed https://github.com/code-ready/crc/issues/321 for that issue, I'd suggest not blocking the merging of this PR on this.

gbraad:
> [WIP] add hyperkit support

Still WIP ?

cfergeau:
> Draft and deployments? Something changed recently ;-)

Now I'm not sure how to remove the "draft" tag :(

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: I've now made a crc-driver-hyperkit release, attached a crc-driver-hyperkit binary to it, and I've also added the hyperkit binary there  for now.

anjannath:
Tested locally, works for me :+1: 

The CI failure is due to `gofmt` error, need to run `make fmt` and push again.

pkg/crc/preflight/preflight_checks_darwin.go
gbraad: we had already agreed on this on slack

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: For now, the hyperkit binary is put next to the crc-machine-driver binary on github. Can be changed later if we have a better place.

pkg/crc/constants/constants.go
```diff
@@ -28,6 +28,7 @@ const (
 
 var (
 	CrcBaseDir         = filepath.Join(GetHomeDir(), ".crc")
+	CrcBinDir          = filepath.Join(CrcBaseDir, "bin")
```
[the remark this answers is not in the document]
cfergeau: I have a patch moving `oc` to ~/.crc/bin, I'll add it to the PR.

pkg/crc/machine/machine.go
```diff
@@ -171,13 +171,22 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, err
 	}
 
-	hostIP, err := network.DetermineHostIP(instanceIP)
+	var hostIP string
+	for i := 0; i < 30; i++ {
```
praveenkumar: This complete logic can move to network package only.

```patch
diff --git a/pkg/crc/machine/machine.go b/pkg/crc/machine/machine.go
index a0b3403..a995b07 100644
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -3,6 +3,7 @@ package machine
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/code-ready/crc/pkg/crc/network"
 	"io/ioutil"
 	"path/filepath"
 	"time"
@@ -11,7 +12,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/errors"
 
-	"github.com/code-ready/crc/pkg/crc/network"
 
@@ -171,17 +171,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, err
 	}
 
-	var hostIP string
-	for i := 0; i < 30; i++ {
-		hostIP, err = network.DetermineHostIP(instanceIP)
-		if err == nil {
-			logging.DebugF("Bridge IP on the host: %s", hostIP)
-			break
-		} else {
-			logging.DebugF("Error finding host IP (%v) - retrying", err)
-		}
-		time.Sleep(2 * time.Second)
-	}
+	hostIP, err := network.DetermineHostIPWithRetry(instanceIP, 30)
 	if err != nil {
 		logging.ErrorF("Error determining host IP: %v", err)
 		result.Error = err.Error()
diff --git a/pkg/crc/network/utils.go b/pkg/crc/network/utils.go
index abf8725..d4cb88c 100644
--- a/pkg/crc/network/utils.go
+++ b/pkg/crc/network/utils.go
@@ -3,6 +3,7 @@ package network
 import (
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/code-ready/crc/pkg/crc/errors"
 
@@ -81,3 +82,15 @@ func DetermineHostIP(instanceIP string) (string, error) {
 
 	return "", errors.New("unknown error occurred")
 }
+
+
+func DetermineHostIPWithRetry(instanceIP string, retry int) (string, error) {
+	for i := 0; i < retry; i++ {
+		hostIP, err := DetermineHostIP(instanceIP)
+		if err == nil {
+			return hostIP, nil
+		}
+		time.Sleep(2 * time.Second)
+	}
+	return "", fmt.Errorf("Not able to find host IP after %d retry", retry)
+}
```

gbraad: agree

PR #304, opened 2019-07-09
Issue #15 Add user pull secret feature.

praveenkumar:


praveenkumar:
retest this please

anjannath:
retest this please

cmd/crc/cmd/config/view.go
```diff
@@ -56,6 +56,10 @@ func determineTemplate(tempFormat string) (*template.Template, error) {
 func runConfigView(cfg map[string]interface{}, tmpl *template.Template, writer io.Writer) error {
 	var lines []string
 	for k, v := range cfg {
+		// In case of pull secret don't show the value.
+		if k == "pullsecret" {
```
gbraad: should be a const value to make this more obvious and ability to lookup references

cmd/crc/cmd/start.go
```diff
@@ -68,6 +70,7 @@ func initStartCmdFlagSet() *pflag.FlagSet {
 	flagSet := pflag.NewFlagSet("start", pflag.ExitOnError)
 	flagSet.StringP(config.Bundle.Name, "b", constants.GetDefaultBundle(), "The system bundle used for deployment of the OpenShift cluster.")
 	flagSet.StringP(config.VMDriver.Name, "d", machine.DefaultDriver.Driver, fmt.Sprintf("The driver to use for the CRC VM. Possible values: %v", machine.SupportedDriverValues()))
+	flagSet.StringP(config.PullSecret.Name, "p", "", "Pull secret for User (Get it from https://cloud.redhat.com/openshift/install/metal/user-provisioned")
```
gbraad: we use the URL twice.

can we extract this? as I am also sure this 'might' change...

pkg/crc/input/input.go
```diff
@@ -18,3 +23,15 @@ func PromptUserForYesOrNo(message string, force bool) bool {
...
+	output.OutF(message + ": ")
+	pullsecret, err := terminal.ReadPassword(int(syscall.Stdin))
+	if err != nil {
+		return "", err
+	}
+	if err := validation.ImagePullSecret(string(pullsecret)); err != nil {
+		return "", err
+	}
+	return string(pullsecret), nil
+}
```
anjannath: This function can be generic, used for getting any kind of secret from user, also the validation is already done in https://github.com/code-ready/crc/pull/304/files#diff-3cdf39384e7d570617455a54805a594aR109

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,75 @@
...
+	base64OfPullSec := base64.StdEncoding.EncodeToString([]byte(pullSec))
+	output, err := drivers.RunSSHCommandFromDriver(driver, fmt.Sprintf("cat <<EOF | tee /tmp/pull-secret.yaml\n%s\nEOF", fmt.Sprintf(pullSecret, base64OfPullSec)))
+	if err != nil {
+		return err
+	}
+	logging.DebugF("Output is : %s", output)
+	return nil
+}
+
+func addpullSecretToIntanceDisk(driver drivers.Driver, pullSec string) error {
```
anjannath: Is this to easily make sure that we are adding the correct pull secret? or it used by the openshift cluster?

cmd/crc/cmd/start.go
```diff
@@ -96,3 +106,26 @@ func validateStartFlags() error {
...
+func getPullSecretFileContent() (string, error) {
+	// In case user doesn't provide a file in start command or in config then ask for it.
+	if crcConfig.GetString(config.PullSecretFile.Name) == "" {
+		pullsecret, err := input.PromptUserForSecret(fmt.Sprintf("Image pull secret for User (Copy it from %s)", constants.DefaultPullSecretURL))
+		// This is just to provide a new line after user enter the pull secret.
+		fmt.Println()
+		if err != nil {
+			return "", err
+		}
+		return pullsecret, nil
```
cfergeau: Should we save this pull secret somewhere? (ideally once we know for sure it's working)

pkg/crc/pullsecret/pullsecret.go
cfergeau: 'Intance' -> Instance

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,75 @@
...
+	output, err := drivers.RunSSHCommandFromDriver(driver, fmt.Sprintf("cat <<EOF | sudo tee /var/lib/kubelet/config.json\n%s\nEOF", pullSec))
+	if err != nil {
+		return err
+	}
+	logging.DebugF("Output is : %s", output)
+	return nil
+}
+
+func addKubeconfigFileToInstance(driver drivers.Driver, kubeconfig string) error {
+	_, err := drivers.RunSSHCommandFromDriver(driver, fmt.Sprintf("cat <<EOF | sudo tee /tmp/kubeconfig\n%s\nEOF", kubeconfig))
```
cfergeau: Why the sudo since it's writing to /tmp?

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if err := addpullSecretToIntanceDisk(driver, pullSec); err != nil {
+		return err
+	}
+	return nil
+}
+
+func addpullSecretFileToInstance(driver drivers.Driver, pullSec string) error {
+	base64OfPullSec := base64.StdEncoding.EncodeToString([]byte(pullSec))
+	output, err := drivers.RunSSHCommandFromDriver(driver, fmt.Sprintf("cat <<EOF | tee /tmp/pull-secret.yaml\n%s\nEOF", fmt.Sprintf(pullSecret, base64OfPullSec)))
```
cfergeau: Nothing is cleaning up that `/tmp/pull-secret.yaml` file? Or is `/tmp` cleaned up on reboot?

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return err
+	}
+	logging.DebugF("Output is : %s", output)
+	return nil
+}
+
+func addpullSecretToIntanceDisk(driver drivers.Driver, pullSec string) error {
+	output, err := drivers.RunSSHCommandFromDriver(driver, fmt.Sprintf("cat <<EOF | sudo tee /var/lib/kubelet/config.json\n%s\nEOF", pullSec))
```
cfergeau: The `oc replace` call is not triggering the recreation of this file?

pkg/crc/machine/machine.go
```diff
@@ -204,6 +205,21 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 
+	// kubeconfig file is also needed inside the VM to perform pull secret replace operation.
+	kubeConfigFile, err := ioutil.ReadFile(filepath.Join(constants.MachineInstanceDir, machineConfig.Name, "kubeconfig"))
+	if err != nil {
+		logging.ErrorF("Failed to read %s: %v", filepath.Join(constants.MachineInstanceDir, machineConfig.Name, "kubeconfig"), err)
+		result.Error = err.Error()
+		return *result, err
+	}
+	//	Update the user pull secret before kubelet start.
+	logging.Info("Adding user's pull secret ...")
```
cfergeau: We only need to do this on first start, it seems you are doing it on every start?

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// This command timeout in 80 secs if not able to replace the pull
+	// secret. We first start the kubelet and wait till the api server is
+	// up and running by hitting it with replace command.
+	replacePullSecretCmd = `timeout 80 bash -c 'sudo systemctl start kubelet && \
+until oc --config /tmp/kubeconfig replace -f /tmp/pull-secret.yaml 2>/dev/null 1>&2; \
+do echo "Waiting for recovery apiserver to come up."; sleep 1; done && \
+sudo systemctl stop kubelet && \
+sudo crictl stopp $(sudo crictl pods -q) && \
+sudo crictl rmp $(sudo crictl pods -q)'`
```
cfergeau: I understand you want to keep the pull secret update self-contained, however starting the kubelet, and then stopping it and removing the misc pods seems quite heavyweight. Since the only time we are going to update the pull secret is going to be during the `start` cmd, can we let `machine.start` start the kubelet process, and only call `oc replace`? Or do we need to restart the kubelet anyway because it can't deal with changes to the pull secret dynamically?

pkg/crc/validation/validation.go
```diff
@@ -51,3 +52,36 @@ func ValidateBundle(bundle string) error {
 	}
 	return nil
 }
+
+// ValidatePath check if provide path is exist
+func ValidatePath(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau this is the file downloaded by the user or created by him/her, this is not the file which will be present already on system so I choose not to check, also when you try to read it using `ioutil` it will through the error if permission is not correct.

pkg/crc/machine/machine.go
```diff
@@ -204,6 +205,21 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, err
 	}
 
+	// kubeconfig file is also needed inside the VM to perform pull secret replace operation.
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau no, I wished we have one :(

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau on reboot it is cleaned also that in the VM so we don't care since there is other place this file available.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau `oc replace` call is only replace the kubernetes resource, it is not going to affect this file since this is on disk and not mounted with any kubenetes resources.

pkg/crc/machine/machine.go
praveenkumar: yes, if a user do stop/start that is not required but atm we are just forcefully doing for every start.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau didn't get you, to replace the pull secret we need to have an apiserver which is why we are starting the kubelet and as soon as we have that then we replace it and again stop it.

cfergeau: What is happening is that in cmd.Start(), we start kubelet, replace the pull secret, stop kubelet, cleanup pods, and then start again kubelet (later in Start()). I was wondering if the restart is mandatory for the changes to take effect, or if we could optimize this to delay start up as little as possible?

pkg/crc/machine/machine.go
cfergeau: How long does this take approximately when you tested this?

pkg/crc/pullsecret/pullsecret.go
cfergeau: Yes, I was hoping that the process 'owning' this file would notice the kubernetes resource change, and rewrite it. Do you know which process is creating it on a fresh openshift install?

pkg/crc/machine/machine.go
praveenkumar: @cfergeau ~30-40 sec.

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau i am bit confuse should we make it part of `snc` script?

pkg/crc/machine/machine.go
cfergeau: This is quite a long time to wait for each `crc start`, I think it's important to only do it when it's needed.

cmd/crc/cmd/start.go
praveenkumar: @cfergeau then we have to implement a way to check if it is stored somewhere and use it from there. So as per our current flow we ask user to download the pullsecret file and use it as argument or part of config set command so it's already in the filesystem and we don't have to store it.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau that's just mistake, first I thought some other location but then thought of using this one.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau after replace the pull secret restart is mandatory as per conversion I had on aos-devel ML but we can try to start the kubelet replace it and keep it running and see if that also work.

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
cfergeau: I would say no. Actually, now I'm wondering, since we are starting the kubelet in the VM, can't we run the `oc replace` from the host?

pkg/crc/pullsecret/pullsecret.go
cfergeau: oh, if you were told that this is mandatory, I'm fine with it, I'd document it either in a comment or in the commit log.

cmd/crc/cmd/start.go
cfergeau: If we only ask it after `crc delete` and on initial start, I would say it's less important to cache it ourselves.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau the way we are right now doing it in the vm using below command. which make sure as soon as it get the secret resource just replace it and the soon stop the kublet service. 

```
$ timeout 80 bash -c 'sudo systemctl start kubelet && \
until oc --config /tmp/kubeconfig replace -f /tmp/pull-secret.yaml 2>/dev/null 1>&2; \
do echo "Waiting for recovery apiserver to come up."; sleep 1; done'
```

we can do the refactor and start using from the host but I am bit spectacle about it since on the host there are still some of unknown (like firewall or dns working etc.) once we do it from guest it is much reliable.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,98 @@
...
+kind: Secret
+metadata:
+  name: pull-secret
+  namespace: openshift-config
+type: kubernetes.io/dockerconfigjson`
+
+	// This command timeout in 80 secs if not able to replace the pull
+	// secret. We first start the kubelet and wait till the api server is
+	// up and running by hitting it with replace command.
+	replacePullSecretCmd = `timeout 80 bash -c 'sudo systemctl start kubelet && \
```
cfergeau: Hiding `systemctl start kubelet` in `replacePullSecretCmd` is a bit opaque, as this means we can't move `updateClusterIdCmd` first, and then call `replacePullSecretCmd`. I'd start the kubelet explicitly using `systemd.NewInstanceSystemdCommander`

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return err
+	}
+	logging.DebugF("Output of %s: %s", replacePullSecretCmd, output)
+
+	clusterID := uuid.New()
+	output, err = drivers.RunSSHCommandFromDriver(driver, fmt.Sprintf(updateClusterIdCmd, clusterID))
+	if err != nil {
+		return err
+	}
```
cfergeau: This can probably be rearranged in:
- start kubelet
- call addPullSecret method which will copy the file and run the needed command
- call addClusterId method which will copy the file and ru the needed command
- stop kubelet (maybe in defer)

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+func AddPullSecretAndClusterID(driver drivers.Driver, pullSec string, kubeconfig string) (rerr error) {
+	defer func() {
+		output, err := drivers.RunSSHCommandFromDriver(driver, stopKubeletAndRemovePodsCmd)
+		logging.DebugF("Output of %s: %s", stopKubeletAndRemovePodsCmd, output)
+		if err != nil {
+			rerr = err
+		}
+	}()
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau No this is needed even any of those fail those will return the error and defer error will be return later so all those will be there.

pkg/crc/machine/machine.go
```diff
@@ -218,6 +219,25 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+		result.Error = err.Error()
+		return *result, err
+	}
+
+	// If it is restart then don't again add user pull secret.
+	if !exists {
+		// Update the user pull secret before kubelet start.
+		logging.Info("Adding user's pull secret and cluster ID ...")
+		if err := pullsecret.AddPullSecretAndClusterID(host.Driver, startConfig.PullSecret, string(kubeConfigFile)); err != nil {
+			logging.ErrorF("Failed to update user pull secret: %v", err)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau it is the `err` variable which will give idea about what actually which command failed.

cmd/crc/cmd/start.go
```diff
@@ -96,3 +106,30 @@ func validateStartFlags() error {
...
+
+func getPullSecretFileContent() (string, error) {
+	var (
+		pullsecret string
+		err        error
+	)
+	// In case user doesn't provide a file in start command or in config then ask for it.
+	if crcConfig.GetString(config.PullSecretFile.Name) == "" {
+		pullsecret, err = input.PromptUserForSecret("Image pull secret", fmt.Sprintf("Copy it from %s", constants.DefaultPullSecretURL))
+		// This is just to provide a new line after user enter the pull secret.
```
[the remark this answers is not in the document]
praveenkumar: > Or does this belong to documentation?

Yes that will go as part of documentation.

> And maybe the logic should be changed to:

I thought of that but it shouldn't be `~/.crc/machines/crc/pull-secret.json` but `~/.crc/cache` since pull secret is not specific to machine, it is not going to change for every start.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau why we want move `updateClusterIdCmd` first, Yes we can have this command separate this command.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau doesn't this we are doing as of now, just kubelet start and add pull secret is in same command.

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
cfergeau: The advantage of `~/.crc/machines/crc/pull-secret.json` is that `crc delete` would get rid of it in case an invalid secret was given. But you are right that asking for a secret after each `crc delete` may be too much.

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
cfergeau: Sorry, was a bit cryptic, but the error message is `Failed to update user pull secret` which is not correct as it could be the cluster id which we could not update.

pkg/crc/pullsecret/pullsecret.go
cfergeau: No need to move it, this was just an example to explain why I think it's better to separate kubelet startup from pull secret update.

pkg/crc/pullsecret/pullsecret.go
cfergeau: yes, kubelet start would be better in a separate command, and the subfunctions/methods did not exist in the version I reviewed.

praveenkumar: @cfergeau yes I updated those now.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau got you, now it is separated.

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau what do you suggest should be something like `Failed to update user pull secret or Cluster id` as message ?

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau we do check the validation of pull secret and also it's not only delete but also stop start use case we shouldn't ask user the pull secret if he provided during the start.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,169 @@
...
+func addKubeconfigFileToInstance(driver drivers.Driver, kubeconfigFilePath string) error {
+	kubeconfig, err := ioutil.ReadFile(kubeconfigFilePath)
+	if err != nil {
+		return err
+	}
+	_, err = drivers.RunSSHCommandFromDriver(driver, fmt.Sprintf("cat <<EOF | tee /tmp/kubeconfig\n%s\nEOF", string(kubeconfig)))
+	if err != nil {
+		return err
+	}
+	return nil
```
cfergeau: I'm wondering if we should introduce a copyFileToInstance()/copyDataToInstance helper as we duplicate this a few times.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+	logging.DebugF("Output of %s: %s", updateClusterIdCmd, output)
+	return nil
+}
+
+func setPullSecretAndClusterID(driver drivers.Driver) (rerr error) {
+	sd := systemd.NewInstanceSystemdCommander(driver)
+	defer func() {
+		// Stop the kublet service.
```
cfergeau: "kubelet"

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,169 @@
...
+			rerr = err
+		}
+	}()
+
+	// Start the kubelet service
+	if _, err := sd.Start("kubelet"); err != nil {
+		return err
+	}
+
+	// Replace existing pull secret to the user pull secret
```
cfergeau: "to the user" -> "with the user"

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+func setPullSecretAndClusterID(driver drivers.Driver) (rerr error) {
+	sd := systemd.NewInstanceSystemdCommander(driver)
+	defer func() {
+		// Stop the kublet service.
+		if _, err := sd.Stop("kubelet"); err != nil {
+			rerr = err
+		}
```
cfergeau: If `sd.Start("kubelet");` fails, it's likely `sd.Stop("kubelet")` would fail too, and if I'm not mistaken, we will overwrite `setPullSecretAndCLusterId` return value with this error, which I don't think is desirable. Maybe change `if err != nil { rerr = err }` to `if rerr == nil { rerr = err }`

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,169 @@
...
+		return fmt.Errorf("Cannot write to %s: %v", pullSecretFilePath, err)
+	}
+	return nil
+}
+
+func GetPullSecretFromCache(pullSecretFilePath string) (string, error, bool) {
+	// Check if file exist
+	if _, err := os.Stat(pullSecretFilePath); os.IsNotExist(err) {
+		return "", nil, false
+	}
```
cfergeau: Isn't this ValidatePath(pullSecretFilePath) ?

cmd/crc/cmd/start.go
```diff
@@ -68,6 +79,7 @@ func initStartCmdFlagSet() *pflag.FlagSet {
 	flagSet := pflag.NewFlagSet("start", pflag.ExitOnError)
 	flagSet.StringP(config.Bundle.Name, "b", constants.GetDefaultBundle(), "The system bundle used for deployment of the OpenShift cluster.")
 	flagSet.StringP(config.VMDriver.Name, "d", machine.DefaultDriver.Driver, fmt.Sprintf("The driver to use for the CRC VM. Possible values: %v", machine.SupportedDriverValues()))
+	flagSet.StringP(config.PullSecretFile.Name, "p", "", fmt.Sprintf("File path of Image pull secret for User (Download it from %s)", constants.DefaultPullSecretURL))
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau no, we don't have any default value. we persist it only once user provide us. Default values are something there and user doesn't have to provide even once.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,169 @@
...
+  name: pull-secret
+  namespace: openshift-config
+type: kubernetes.io/dockerconfigjson`
+
+	// This command timeout in 80 secs if not able to replace the pull secret.
+	replacePullSecretCmd = `timeout 80 bash -c 'until oc --config /tmp/kubeconfig replace -f /tmp/pull-secret.yaml 2>/dev/null 1>&2; \
+do echo "Waiting for recovery apiserver to come up."; sleep 1; done'`
+
+	// This command is used to update the clusterID using random uuid.
+	updateClusterIdCmd = `oc --config /tmp/kubeconfig patch clusterversion version -p '{"spec":{"clusterID":"%s"}}' --type merge`
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I didn't get this, the first command is there to check the server readiness why all subsequent command follow the same. I am neutral about it.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// This command make sure we stop the kubelet and clean up the pods
+	// We also providing a 2 sec sleep so that stop pods get settled and
+	// ready for remove. Without this 2 sec time sometime it happen some of
+	// the pod are not completely stopped and when remove happen it will throw
+	// an error like below.
+	// remove /var/run/containers/storage/overlay-containers/97e5858e610afc9f71d145b1a7bd5ad930e537ccae79969ae256636f7fb7e77c/userdata/shm: device or resource busy
+	stopAndRemovePodsCmd = `bash -c 'sudo crictl stopp $(sudo crictl pods -q) && sleep 2 &&\
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I don't know, this is something I didn't get in my system but CI failed once so just added in safer side.

cmd/crc/cmd/start.go
```diff
@@ -35,21 +39,28 @@ var startCmd = &cobra.Command{
 
 var (
 	startCmdFlagSet = initStartCmdFlagSet()
+	pullSecretFile  string
 )
 
 func runStart(arguments []string) {
 	if err := validateStartFlags(); err != nil {
-		errors.Exit(1)
+		errors.ExitWithMessage(1, err.Error())
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau and it is bad, because if you just exist user doesn't even know why it exit and no error message :(

cmd/crc/cmd/start.go
```diff
@@ -96,3 +108,44 @@ func validateStartFlags() error {
...
+			}
+			pullsecret = string(data)
+		}
+	}
+	if err := validation.ImagePullSecret(pullsecret); err != nil {
+		return "", err
+	}
+
+	// Add pull secret to cache.
+	if err := ps.StorePullSecretToCache(pullsecret, constants.MachineCacheDir, constants.PullSecretFile, 0755); err != nil {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau there is a reason since `ps.StorePullSecretToCache` is actually check the if parent directory available or not and create if not available but later one just need to read the file with a given file path. `0755` is there if the directory not present and we are creating all the directory with this permission.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau then what if the start happen successful but stop failed then how would we know?

cfergeau: If everything was successful until the `defer`, then `rerr` will be `nil`. If then something fails in the `defer`, then we'll overwrite the `nil` `rerr` with the error that happened. If something failed before, then we will not be overwriting it.

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
cfergeau: If all the errors we return come from crc errors package, then there will be an error message:
https://github.com/code-ready/crc/blob/master/pkg/crc/errors/error.go#L11
At the time I  wrote 7d2d016 this was the case, there were no standard go errors. As soon as we have a mix of crc errors and errors from outside crc, then yes, it's getting bad, and sometimes we will have an error message, and sometimes not.
Overall, I think we should remove the logging from our error implementation, and add back that ExitWithMessage. But that's a totally different discussion than what this pull request is about ;)

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau this create cyclic dependency which is not allowed in go :(.

```
package github.com/code-ready/crc/cmd/crc
	imports github.com/code-ready/crc/cmd/crc/cmd
	imports github.com/code-ready/crc/cmd/crc/cmd/config
	imports github.com/code-ready/crc/pkg/crc/config
	imports github.com/code-ready/crc/pkg/crc/validation
	imports github.com/code-ready/crc/pkg/crc/machine
	imports github.com/code-ready/crc/pkg/crc/pullsecret
	imports github.com/code-ready/crc/pkg/crc/validation
```

cfergeau: Ok :(

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
cfergeau: Ah I think I'd favour passing 0644, and using `err := os.MkdirAll(destDir, mode | 0111)` (or just hardcode 0755/0644 in that function).
However, now that we talk about it, I don't think the pull secret should be readable by anyone with an account on the system, I'd use 0700 and 0600 to keep it secret from other users.

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
cfergeau: yes something like this :)

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau Ah, I got it now, so I am going to use the `multierror` functionality.

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
cfergeau: I'd prefer that we keep this change out of this PR for now.

pkg/crc/pullsecret/pullsecret.go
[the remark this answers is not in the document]
cfergeau: For better readability/better maintainability:
```
        // Add the Pull secret kubernetes resource definition to Instance
	if err := addpullSecretSpecToInstance(driver, pullSec); err != nil {
		return err
	}

	// Replace user pull secret and add cluster ID
	if err := setPullSecretAndClusterID(driver); err != nil {
		return err
	}
```
If I'm working on this code/refactoring it/..., I have no way of knowing that the first method call *must* come before the second one.

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
cfergeau: This does not seem updated in the latest iteration.

pkg/crc/pullsecret/pullsecret.go
cfergeau: Hmm, why not, but I think once the first error happened, we can ignore the ones which happen on cleanup.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,173 @@
...
+	// add random cluster id
+	if err := addClusterID(driver); err != nil {
+		m.Collect(err)
+	}
+
+	return nil
+}
+
+func StorePullSecretToFile(pullsecret string, destDir string, fileName string, mode os.FileMode) error {
+	err := os.MkdirAll(destDir, mode)
```
cfergeau: I'd hardcode 0700 here too  if you hardcode the 0600 below.

praveenkumar: @cfergeau no, all other directory are 0755 so I didn't hard code it.

praveenkumar:
> Hmm, why not, but I think once the first error happened, we can ignore the ones which happen on cleanup.

@cfergeau this I changed to used `multiError` struct so now it should not ignore any of the error https://play.golang.org/p/XYcvuCjQfvh

cmd/crc/cmd/start.go
```diff
@@ -96,3 +108,46 @@ func validateStartFlags() error {
...
+			pullsecret = string(data)
+		}
+	}
+	if err := validation.ImagePullSecret(pullsecret); err != nil {
+		return "", err
+	}
+
+	// Add pull secret to cache.
+	if !pullSecretFilePresent {
+		if err := ps.StorePullSecretToFile(pullsecret, constants.MachineCacheDir, constants.PullSecretFile, 0755); err != nil {
```
cfergeau: I'd prefer to get rid of this 0755 altogether and hardcode it.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,176 @@
...
+		return err
+	}
+
+	// Replace user pull secret and add cluster ID
+	if err := setPullSecretAndClusterID(driver); err != nil {
+		return err
+	}
+
+	// Add user pull secret to the instance
+	if err := addpullSecretToIntanceDisk(driver, pullSec); err != nil {
```
cfergeau: Still "Intance" typo.

cfergeau: Any reason for tying `setPullSecretAndClusterID` together?
I would write this as:
```
if err := addKubeconfigFileToInstance(driver, kubeconfigFilePath); err != nil {
        return err
}

if err := setPullSecret(driver, pullSec); err != nil {
//...
}

if err := setClusterID(driver); err != nil {
//...
}
```

pkg/crc/pullsecret/pullsecret.go
cfergeau: Use 0755 if you think it's better, but I would not add a `mode` argument.

pkg/crc/pullsecret/pullsecret.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	pullSecretFilePath := filepath.Join(destDir, fileName)
+	// Use 0600 mode to create this so it is readable only by current user.
+	if err := ioutil.WriteFile(pullSecretFilePath, []byte(pullsecret), 0600); err != nil {
+		return fmt.Errorf("Cannot write to %s: %v", pullSecretFilePath, err)
+	}
+	return nil
+}
+
+func GetPullSecretFromFile(pullSecretFilePath string) (string, bool, error) {
```
cfergeau: `GetPullSecretFromFile` and `StorePullSecretToFile` are not consistent, one takes dir + filename, the other takes full path.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: > Any reason for tying setPullSecretAndClusterID together?

@cfergeau yes so I don't have to start the kubelet service 2 times and the clean up again :(

cfergeau: I had something like this in mind:
https://github.com/cfergeau/crc/commit/6355e185bd9c27b87e705c733ecc5d7d4e210a38 (only compile tested)

pkg/crc/pullsecret/pullsecret.go
cfergeau: Maybe https://github.com/cfergeau/crc/commit/b5fd6743e20e71b6766adb7c91835bf11e7c8f41 ?

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau Initially I had same codebase then when I tried to run it and check the how many pods run in a given span of time is way too much.  When I was using a single script which was doing `kubelet start` => changing all the bits => `kubelet stop` => removing the pods that time only 4 pods which got created during that timeframe now after we made it moduler now around 8 pods are running in that time frame and if we again putting the defer on top of `AddPullSecretAndClusterID` then we might hit 10-12 pods and each time we delay we are adding more delay in clean up step so I didn't consider this idea.

pkg/crc/pullsecret/pullsecret.go
praveenkumar: @cfergeau that just matter of taste :)

pkg/crc/pullsecret/pullsecret.go
cfergeau: hmm ok, I did not realize this before now. The whole point of making the command modular was to be able to group related things together. If we can not move things around freely because otherwise the cleanup becomes too costly, then we could probably have kept the 2 bash commands together.

Issue #308, opened 2019-07-11
Add functionality to get the cluster operator status, programatically.


praveenkumar:
Right now we don't have any way to get the cluster operator status programmatically, we should have it so command like `status` can consume it.

gbraad:
you mean, the status of a single operator?

praveenkumar:
@gbraad nope, I meant all the required operators.

gbraad:
check... as selective checking would sound 'wasteful' to me.

anjannath:
Fixed in #309

PR #309, opened 2019-07-11
Issue #308 Add function to get get operator status

praveenkumar:
We need to check cluster operator status at different places, like
checking the status of the cluster or even implementing the status
command.

### DIFF
--- a/pkg/crc/oc/clusteroperator.go
+++ b/pkg/crc/oc/clusteroperator.go
@@ -1,0 +1,78 @@
+package oc
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"time"
+)
+
+var ignoreClusterOperators = []string{"monitoring", "machine-config", "marketplace"}
+
+type ClusterOperator struct {
+	Items []struct {
+		Metadata struct {
+			Name string `json:"name"`
+		} `json:"metadata"`
+		Status struct {
+			Conditions []struct {
+				LastTransitionTime time.Time `json:"lastTransitionTime"`
+				Reason             string    `json:"reason"`
+				Status             string    `json:"status"`
+				Type               string    `json:"type"`
+			} `json:"conditions"`
+		} `json:"status,omitempty"`
+	} `json:"items"`
+}
+
+func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
+	allAvailable := true
+	data, stderr, err := oc.RunOcCommand("get", "co", "-ojson")
+	if err != nil {
+		return false, fmt.Errorf("%s - %v", stderr, err)
+	}
+
+	var co ClusterOperator
+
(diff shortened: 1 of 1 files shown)


jsliacan:
I like it, it's going to be useful. I was wondering, if an operator is not available, it could just be slow. But your data is fixed once you unmarshal it, so you can't just try that one operator a few times and proceed onto the next one after it's available. Now it's all pretty fast, but when I wrote the tests I used to wait a long time before "authentication" operator became available. Would you just call the whole `GetClusterOperatorStatus` over and over if one of the operators was slow to come on?

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,75 @@
...
+				LastTransitionTime time.Time `json:"lastTransitionTime"`
+				Reason             string    `json:"reason"`
+				Status             string    `json:"status"`
+				Type               string    `json:"type"`
+			} `json:"conditions"`
+		} `json:"status,omitempty"`
+	} `json:"items"`
+}
+
+func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
```
jsliacan: `GetClusterOperatorStatus` sounds like it takes an argument - the cluster operator whose status I want to know. Maybe using plural would be OK: `GetClusterOperatorsStatus`. Or `CheckClusterOperatorsAvailable` - because we're not really checking status, just conjunctive availability.

praveenkumar:
> `Would you just call the whole GetClusterOperatorStatus over and over if one of the operators was slow to come on?` 

@jsliacan for us if one operator which suppose to work and not working, means breaking the cluster experience. What I hoping from consumer of this function to use it in a loop with a sleep time.

pkg/crc/oc/clusteroperator.go
praveenkumar: @jsliacan I just used what `oc` have, they use `oc get clusteroperator`

anjannath:
@gbraad Are we good to merge this?

pkg/crc/oc/clusteroperator.go
```diff
@@ -0,0 +1,75 @@
...
+				if con.Status != "False" {
+					logging.Debug(c.Metadata.Name, " operator is degraded, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			case "Progressing":
+				if con.Status != "False" {
+					logging.Debug(c.Metadata.Name, " operator is still progressing, Reason: ", con.Reason)
+					allAvailable = false
+				}
+			}
```
cfergeau: You could add a
```
default:
  logging.Infof("Unexpected operator status for %s: %s", c.Metadata.Name, con.Type")
  allAvailable = false
```
to make this more future proof? Or are we guaranteed that there will never be more `con.Type` names?

praveenkumar: @cfergeau yes we can make that change just to be sure in the future if openshift side add any additional `con.Type`.

PR #310, opened 2019-07-11
Issue #306 Add pre-cluster start certs validity check

### LINKED ISSUE #306: Post VM start, pre cluster startup check: validate certificate time
Since the issue with the certificates hasn't been solved yet, we have to provide an intermediate solution that can help the user/us to identify issues related to the expired certificates.

We need to add a post-flight check, right before the cluster startup, that checks the certificates for validity.

anjannath:
Fixes #306

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -55,7 +55,7 @@ func runStart(arguments []string) {
 
 	commandResult, err := machine.Start(startConfig)
 	if err != nil {
-		errors.Exit(1)
+		errors.ExitWithMessage(1, err.Error())
 	}
 	if commandResult.Status == "Running" {
 		output.Out("CodeReady Containers instance is running")
--- a/pkg/crc/input/input.go
+++ b/pkg/crc/input/input.go
@@ -12,7 +12,7 @@ func PromptUserForYesOrNo(message string, force bool) bool {
 	}
 	var response string
 	output.OutF(message + "? [y/N]: ")
-	fmt.Scanf("%s", response)
+	fmt.Scanf("%s", &response)
 	if strings.ToLower(response) == "y" {
 		return true
 	}
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -30,6 +30,9 @@ import (
 	"github.com/code-ready/machine/libmachine/host"
 	"github.com/code-ready/machine/libmachine/log"
 	"github.com/code-ready/machine/libmachine/state"
+
+	// cluster related import
+	"github.com/code-ready/crc/pkg/crc/cluster"
 )
 
 func Start(startConfig StartConfig) (StartResult, error) {
@@ -146,6 +149,17 @@ func Start(startConfig StartConfig) (StartResult, error) {
 
 	// Post-VM start
 	host, err := libMachineAPIClient.Load(machineConfig.Name)
+	if err != nil {
+		logging.ErrorF("Error loading %s vm: %v", machineConfig.Name, err)
+		result.Error = err.Error()
+		return *result, err
+	}
+	// Check the certs validity inside the vm
+	logging.Info("Verifying validity of the cluster certificates ...")
+	if err := cluster.CheckCertsValidity(host.Driver); err != nil {
+		result.Error = err.Error()
+		return *result, err
+	}
 	instanceIP, err := host.Driver.GetIP()
 	if err != nil {
 		logging.ErrorF("Error getting the IP: %v", err)
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -1,0 +1,25 @@
+package cluster
+
+import (
+	"fmt"
+	"github.com/code-ready/machine/libmachine/drivers"
+	"strings"
+	"time"
+)
+
+// CheckCertsValidity checks if the cluster certs have expired
+func CheckCertsValidity(driver drivers.Driver) error {
+	certExpiryDateCmd := `date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-client-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds`
+	output, err := drivers.RunSSHCommandFromDriver(driver, certExpiryDateCmd)
+	if err != nil {
+		return err
+	}
+	certExpiryDate, err := time.Parse(time.RFC3339, strings.TrimSpace(output))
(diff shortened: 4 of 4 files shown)


pkg/crc/input/input.go
```diff
@@ -12,7 +12,7 @@ func PromptUserForYesOrNo(message string, force bool) bool {
 	}
 	var response string
 	output.OutF(message + "? [y/N]: ")
-	fmt.Scanf("%s", response)
+	fmt.Scanf("%s", &response)
```
praveenkumar: As per `Scanf` function definition it take `func Scanf(format string, a ...interface{})` any value then why you want to pass the address of string instead directly this string, does't it evaluate? looks like type should be the pointer https://play.golang.org/p/VSc_P1QC77G

cmd/crc/cmd/start.go
```diff
@@ -55,7 +55,7 @@ func runStart(arguments []string) {
 
 	commandResult, err := machine.Start(startConfig)
 	if err != nil {
-		errors.Exit(1)
+		logging.Fatal(err)
```
praveenkumar: why we want this change, does it show the error message and have exit code non-zero? cc @cfergeau

pkg/crc/machine/machine.go
```diff
@@ -401,3 +405,16 @@ func unsetMachineLogging() {
 	logging.CloseLogFile()
 	logging.SetupFileHook()
 }
+
+func checkBundleCertsValidity(bundleTime string) error {
```
praveenkumar: Since this method doesn't have any relation to `machine` can we move it to `bundle` package?

pkg/crc/input/input.go
anjannath: Right, i found out about it today..

cmd/crc/cmd/start.go
anjannath: This just exits, and doesnot print the err message, and user does not have any clue as to why it exited, and yes it throws exit code 1

gbraad: Wouldn't it then be a good idea to modify the Exit method to also accept a message?

pkg/crc/machine/machine.go
```diff
@@ -401,3 +409,19 @@ func unsetMachineLogging() {
...
+	certExpiryDateCmd := `date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-client-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds`
+	output, err := drivers.RunSSHCommandFromDriver(driver, certExpiryDateCmd)
+	if err != nil {
+		return err
+	}
+	certExpiryDate, err := time.Parse(time.RFC3339, strings.TrimSpace(output))
+	if err != nil {
+		return err
+	}
+	if time.Now().After(certExpiryDate) {
```
gbraad: 👍

pkg/crc/machine/machine.go
anjannath: Thanks, your comment reminded me that i had to change the function, which i did.. and now i think it belongs here, as its getting the cert expiry date from the vm.

cmd/crc/cmd/start.go
anjannath: Yes, we already have `ExitWithMessage`, will use that!

gbraad: logging exiting when fatal is used sounds very unintuitive. I would not expect logging to exit (feels like a side-effect)... even errors performing an exit is weird, but at least the function name makes you expect it to happen.

praveenkumar:
@anjannath @gbraad shouldn't we take it from metadata since it have build time `"buildTime": "2019-06-28T09:30:28+00:00",` which we can compare and return this way we can tell user very early before even starting the VM?

gbraad:
buildtime is not conclusive and ONLY an indicator, as when certificates have rotated, the in VM certificates are newer.

You can, and that was the idea, have an inidication this this might NOT be the latest bundle, as it is older than 30 days.

anjannath:
> @anjannath @gbraad shouldn't we take it from metadata since it have build time "buildTime": "2019-06-28T09:30:28+00:00", which we can compare and return this way we can tell user very early before even starting the VM?

`buildTime` is set as part of the snc script and after that we wait addtional 24hrs for the cert rotation, we can probably have a preflight check for the bundle (and warn) using the `buildTime`, but checking the certs from inside the VM (pre cluster-start) seemed more robust..

gbraad:
> seemed more robust..

and the only way. The buildtime is informational, but it can never error out the start, merely warn.

praveenkumar:
@anjannath Can you add that info in commit message (like @gbraad said when cert rotation happen then buildtime is not conclusive) ? I am OK with that patch will merge after.

anjannath:
@praveenkumar Done!

cmd/crc/cmd/start.go
```diff
@@ -55,7 +55,7 @@ func runStart(arguments []string) {
 
 	commandResult, err := machine.Start(startConfig)
 	if err != nil {
-		errors.Exit(1)
+		errors.ExitWithMessage(1, err.Error())
```
praveenkumar: 👍

pkg/crc/machine/machine.go
```diff
@@ -204,6 +206,12 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, err
 	}
 
+	// Check the certs validity inside the vm
+	logging.Info("Verifying validity of the cluster certificates ...")
+	if err := checkClusterCertsValidity(host.Driver); err != nil {
+		result.Error = err.Error()
+		return *result, err
+	}
```
praveenkumar: @anjannath this block should go https://github.com/code-ready/crc/pull/310/files#diff-06d0e75b8f65eb58c0f83515f2a7217dL146 here, we don't want to execute anything on the VM if certs are not valid.

pkg/crc/machine/machine.go
```diff
@@ -401,3 +414,19 @@ func unsetMachineLogging() {
 	logging.CloseLogFile()
 	logging.SetupFileHook()
 }
+
+func checkClusterCertsValidity(driver drivers.Driver) error {
```
praveenkumar: @anjannath can we have this function away from machine side? may be create an another package call `cluster` and have all the cluster related checks there? this kind of function will explode the `machine.go` file and would be become something like we had in `minishift/start.go` side.

anjannath: @praveenkumar Done!

PR #312, opened 2019-07-12
Add function to read symlink file.

praveenkumar:
Most of users on mac have `/etc/resolv.conf` file symlink
to `/var/run/resolv.conf`. Also `waitforNetwork` function
try to read the file which is in updating state during network
reload so reading this file will fail. We need to get the host
nameservers before reloading the network.

### DIFF
--- a/pkg/crc/network/nameservers.go
+++ b/pkg/crc/network/nameservers.go
@@ -74,12 +74,17 @@ func HasNameserverConfiguredLocally(nameserver NameServer) (bool, error) {
 
 func GetResolvValuesFromHost() (*ResolvFileValues, error) {
 	// TODO: we need to add runtime OS in case of windows.
-	out, err := ioutil.ReadFile("/etc/resolv.conf")
+	fInfo, err := crcos.GetFilePath("/etc/resolv.conf")
+	if err != nil {
+		return nil, err
+	}
+	out, err := ioutil.ReadFile(fInfo)
 	if crcos.CurrentOS() == crcos.WINDOWS {
 		// TODO: we need to add logic in case of windows.
 	}
 
 	if err != nil {
+		fmt.Println("This is the error here")
 		return nil, fmt.Errorf("Failed to read resolv.conf: %v", err)
 	}
 	return parseResolveConfFile(string(out))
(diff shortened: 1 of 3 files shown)


pkg/os/util.go
```diff
@@ -85,3 +85,15 @@ func CopyFileContents(src string, dst string, permission os.FileMode) error {
 
 	return nil
 }
+
+// GetFilePath returns the file path even it is symlink
+func GetFilePath(path string) (string, error) {
+	fInfo, err := os.Lstat(path)
```
gbraad: just like in Minishift ;-)

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -93,7 +101,7 @@ func restartNetwork() (bool, error) {
 		}
 	}
 
-	err = waitForNetwork()
+	err = waitForNetwork(hostResolv)
```
gbraad: is actually a very generic network function, but we can refactor this later.

pkg/crc/network/nameservers.go
```diff
@@ -74,12 +74,17 @@ func HasNameserverConfiguredLocally(nameserver NameServer) (bool, error) {
 
 func GetResolvValuesFromHost() (*ResolvFileValues, error) {
 	// TODO: we need to add runtime OS in case of windows.
-	out, err := ioutil.ReadFile("/etc/resolv.conf")
+	fInfo, err := crcos.GetFilePath("/etc/resolv.conf")
```
cfergeau: I must be missing something, as I far as I can tell, ioutil.ReadFile already follows symlinks:
```
package main

import "bytes"
import "io/ioutil"
import "log"
import "os"


        
func main() {
	const content = "test file content"
        var byte_content = []byte(content)

	err := ioutil.WriteFile("/tmp/realfile", byte_content, 0644)
	if err != nil {
		log.Fatal(err)
	}
	err = os.Symlink("/tmp/realfile", "/tmp/symlink")
	if err != nil {
		log.Fatal(err)
	}
			        
	out, err := ioutil.ReadFile("/tmp/symlink")
	if err != nil {
		log.Fatal(err)
	}
	if !bytes.Equal(out, byte_content) {
		log.Fatal("symlink content does not match expected content")
	}
	log.Println("Symlink test worked as expected")
}
```

praveenkumar: @cfergeau, you are not missing anything, `ioutil.ReadFile` able to read the file which is even symlink but I added if in future we required some symlink path, also I hope this is not causing any issue.

Issue #314, opened 2019-07-12
Use `/etc/hosts` instead `/etc/resolv.conf` for OSX


praveenkumar:
At this moment we are using the `/etc/resolv.conf` to add the nameserver as vm ip to resolve the cluster specific domain name. It is problematic since most of users have `/etc/resolv.conf` symlinked to `/var/run/resolv.conf` and that also change whenever network restart happen. `oc` also respect `/etc/hosts` entry so atleast we can have apiserver and auth URI as part of this.

gbraad:
This issue is created as `oc` does not understand the macOS resolver when compiled without CGO; see https://github.com/openshift/origin/issues/23266#issuecomment-508390514

Note: It is not that `oc` respects `/etc/hosts` but gracefully falls back to libc resolving when the Go binary is not compiled using CGO. This is more of a side-effect than a choice.

gbraad:
Ref: https://github.com/code-ready/crc/issues/293

praveenkumar:
resolved by #315

PR #315, opened 2019-07-12
Issue #314 Use /etc/hosts instead /etc/resolv.conf

praveenkumar:
OSx have `/etc/resolv.conf` symlink to `/var/run/resolv.conf` which
reloads when network restart happen. `oc` does respect /etc/hosts
entry and it is less error prone.


praveenkumar:
retest this please.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -123,37 +125,37 @@ func waitForNetwork(hostResolv *network.ResolvFileValues) error {
 	return fmt.Errorf("Host is not connected to internet.")
 }
 
-// updateResolvConfFile updates the host's /etc/resolv.conf file with Instance IP.
-func updateResolvConfFile(instanceIP string, resolvConfFile string) (bool, error) {
```
cfergeau: The 'bool' return value does not seem to be needed as it's the same as (error == nil)

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -123,37 +125,37 @@ func waitForNetwork(hostResolv *network.ResolvFileValues) error {
...
-	foundExistingInstanceIP := false
-
-	for i, ns := range hostResolv.NameServers {
-		// Update the nameserver IP with instance IP if already exist
-		if strings.Contains(ns.IPAddress, "192.168.130") {
-			hostResolv.NameServers[i].IPAddress = instanceIP
-			foundExistingInstanceIP = true
-			break
+	// Check if the Domain already present if yes then remove those
+	if hosts.HasHostname(constants.DefaultAPIDomain) {
```
cfergeau: `constants.DefaultAPIDomain` can be used if a fallback is needed, but the value should be built from `baseDomain`, `clusterName` and `appsDomain` from `crc-bundle-info.json`

pkg/crc/services/dns/dns_darwin.go
praveenkumar: @cfergeau it is required by calling function, where it is used as part of the `result` struct. or you mean just return `error == nil` and it will evaluate as bool ?

cfergeau: The calling function can generate the 'success' value itself, for example:
```
if err != nil {
	result.Success = false
	return *result, err
}
```

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -123,37 +125,37 @@ func waitForNetwork(hostResolv *network.ResolvFileValues) error {
 	return fmt.Errorf("Host is not connected to internet.")
 }
 
-// updateResolvConfFile updates the host's /etc/resolv.conf file with Instance IP.
-func updateResolvConfFile(instanceIP string, resolvConfFile string) (bool, error) {
-	// Get the current value of resolv.conf file
-	hostResolv, err := network.GetResolvValuesFromHost()
+// updateHostsConfFile updates the host's /etc/hosts file with Instance IP.
+func updateHostsConfFile(instanceIP string, clusterName string, baseDomain string, appsDomain string) (bool, error) {
```
cfergeau: I think this function can be made both shorter and more generic, see https://github.com/cfergeau/crc/commit/173751dde0c550f5eb77974735bffbd9a5e84f79

pkg/crc/services/dns/dns_darwin.go
praveenkumar: @cfergeau then I have to change it to more places, since all the other functions signature is same which called from the calling function.

cfergeau: ah this is indeed a common pattern in this file :-/

praveenkumar: @cfergeau I changed all the functions signature as of now, don't want to put another commit message for this refactor.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -123,37 +124,39 @@ func waitForNetwork(hostResolv *network.ResolvFileValues) error {
...
-	// Write to the resolv.conf file.
-	resolvFile, _ := network.CreateResolvFile(*hostResolv)
-	err = ioutil.WriteFile(resolvConfFile, []byte(resolvFile), 0644)
-	if err != nil {
-		return false, err
+	return true, nil
+}
+
+func replaceHostname(hosts goodhosts.Hosts, ipAddress string, hostname string) error {
+	logging.DebugF("Updating %s to %s /etc/hosts file", hostname, ipAddress)
```
cfergeau: 'in /etc/hosts file'

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -123,37 +124,39 @@ func waitForNetwork(hostResolv *network.ResolvFileValues) error {
...
 
-	// Write to the resolv.conf file.
-	resolvFile, _ := network.CreateResolvFile(*hostResolv)
-	err = ioutil.WriteFile(resolvConfFile, []byte(resolvFile), 0644)
-	if err != nil {
-		return false, err
+	return true, nil
+}
+
+func replaceHostname(hosts goodhosts.Hosts, ipAddress string, hostname string) error {
```
cfergeau: My bad, I hadn't tested these proposed changes besides compilation. This has to be `hosts *goodhosts.Hosts` otherwise /etc/hosts will not be modified

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -123,37 +124,39 @@ func waitForNetwork(hostResolv *network.ResolvFileValues) error {
...
-	foundExistingInstanceIP := false
-
-	for i, ns := range hostResolv.NameServers {
-		// Update the nameserver IP with instance IP if already exist
-		if strings.Contains(ns.IPAddress, "192.168.130") {
-			hostResolv.NameServers[i].IPAddress = instanceIP
-			foundExistingInstanceIP = true
-			break
+	for _, hostname := range hostnames {
+		if err := replaceHostname(hosts, instanceIP, hostname); err != nil {
```
cfergeau: This needs to be `&hosts`

pkg/crc/services/dns/dns_darwin.go
praveenkumar: @cfergeau it is modified, what actually happened in your proposed changes you didn't try to use `host.Flush()` which actually write the file to disk, I already do that so this is not required.

Issue #319, opened 2019-07-16
Add user defined nameserver to the instance.


praveenkumar:
Sometime the default nameserver which is added by the hypervisor doesn't work as expected so we should provide a way to add a custom nameserver if user defined as part of start or from the config.

praveenkumar:
Fixed by #320

PR #320, opened 2019-07-16
Issue #319 Add a start option to pass namesever.

praveenkumar:
Sometime default nameserver provided by hypervisor doesn't
work so we need a way for user to specify it.

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -23,10 +23,11 @@ var SettingsList = make(map[string]*setting)
 
 var (
 	// Start command settings in config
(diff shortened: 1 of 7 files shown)


cfergeau:
Do you have more details about "sometimes"? I've seen the nameserver in /etc/resolv.conf become stale on macos, but we avoid using /etc/resolv.conf on that platform, and we should also avoid as much as possible to restart the network. I'd prefer that we keep this as a very last resort (ie do not add this option until it's clear we have no other choice)

praveenkumar:
@cfergeau this change is not on the host side this is on the VM side. I will add more detail in the commit message.

praveenkumar:
I need to add more things to it and remove some of unused part, will do it soon.

Issue #322, opened 2019-07-16
Remove network reload on macos


cfergeau:
On macos, when modifying a file in /etc/resolver/ as a regular user, one has to restart networking for the change to be taken into account by the rest of the system. For some reason, this does not occur when modifying the file with sudo, but using sudo is not an option for `crc`.

It would be nice to figure out how the system is notified of the change when done as root, and see if we can do the same thing as a regular user (is there a specific command to call?), so that we don't need to restart the network (this causes VPN connections to drop, this can cause IP addresses to change and SSH connection to be lost, ...).

In the mean time, we can minimize the number of network restarts by only triggering one when the file actually changed.

cfergeau:
> On macos, when modifying a file in /etc/resolver/ as a regular user, one has to restart networking for the change to be taken into account by the rest of the system. For some reason, this does not occur when modifying the file with sudo, but using sudo is not an option for `crc`.
> 
> It would be nice to figure out how the system is notified of the change when done as root, and see if we can do the same thing as a regular user (is there a specific command to call?), so that we don't need to restart the network (this causes VPN connections to drop, this can cause IP addresses to change and SSH connection to be lost, ...).

Actually, https://unix.stackexchange.com/questions/400395/what-is-causing-the-changes-in-etc-resolver-on-macos-sierra-to-be-noticed has a good explanation for that :
> Running it with my regular user had no effect. My theory is that whatever is watching /etc/resolver watches for changes to the directory, and not to changes to files within it, and the reloads everything in it when it sees a change to the directory. Vim creates a swap file, by default in the same directory as the file being edited. This is a change to the directory, and is picked up by the watcher.

PR #323, opened 2019-07-16
darwin: Only restart network when needed

cfergeau:
We currently have to restart macos network in order so that the changes we
make in /etc/resolver as a regular are used by the OS.
However, this also has the side-effect of killing VPN connections, it
can lead to IP address changes and loss of SSH connectivity, ...
As long as the VM IP did not change, we do not need to modify anything
in /etc/resolver/. This commit takes advantage of that to minimize
network restarts, they only get triggered when the content of
/etc/resolver/ changed.

This is related to https://github.com/code-ready/crc/issues/322

### DIFF
--- a/pkg/crc/services/dns/dns_darwin.go
+++ b/pkg/crc/services/dns/dns_darwin.go
@@ -3,7 +3,6 @@ package dns
 import (
 	"bytes"
 	"fmt"
-	"io/ioutil"
 	"path/filepath"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


pkg/crc/services/dns/dns_darwin.go
```diff
@@ -37,22 +37,41 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
...
+		logging.InfoF("Network restart not needed")
 	}
+
 	// we pass the result and error on
 	result.Success = true
 	return *result, nil
 }
 
-func createResolverFile(InstanceIP string, path string) error {
+func writeIfContentChanged(path string, new_content []byte) (bool, error) {
```
praveenkumar: Can we extract this function and have it part of `pkg/os/util.go` so we can reuse it in future if we required that kind of functionality and also name as `WriteIfFileContentChanged`

cfergeau: Done, I named it `WriteFileIfContentChanged` and added a `perm` argument to be similar to ioutil.WriteFile.

praveenkumar:
@cfergeau looks like CI failed during the `fmt` check execute `make fmt` once to have it all green and then we can have it in.

PR #328, opened 2019-07-19
Retry reading the content of the resolv.conf until time-out reached

praveenkumar:
Right now, we are blocking the start in case of a mac host is not able to reach to the network
after network restart. If a host offline then it will block start forever which we don't want.

In current logic if a host is initially connected to vpn and then network restart happen, that means host
now have a different nameserver than before and this test will fail.

Making it non blocking allow us to overcome both the situation without much compromise.

In case of a host connected to vpn restart of network going to change the host nameserver and that will surely fail to ping the previous stored nameservers. If we are changing it to a well known nameserver at least 90% case it will succeed, going to fail only if a organisation block icmp from this nameserver.

### DIFF
--- a/pkg/crc/services/dns/dns_darwin.go
+++ b/pkg/crc/services/dns/dns_darwin.go
@@ -101,12 +101,20 @@ func restartNetwork() error {
 
 // Wait for Network wait till the network is up, since it is required to resolve external dnsquery
 func waitForNetwork() error {
-	hostResolv, err := network.GetResolvValuesFromHost()
+	var hostResolv *network.ResolvFileValues
+	var err error
+
+	for i := 1; i <= 5; i++ {
+		hostResolv, err = network.GetResolvValuesFromHost()
+		if err == nil {
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
 	if err != nil {
-		// wait and retry
 		return fmt.Errorf("Unable to read host resolv file (%v)", err)
 	}
-	// retry for 5 times
+	// retry up to 5 times
 	for i := 0; i < 5; i++ {
 		for _, ns := range hostResolv.NameServers {
 			_, _, err := crcos.RunWithDefaultLocale("ping", "-c4", ns.IPAddress)
--- a/pkg/crc/services/dns/dns_darwin.go
+++ b/pkg/crc/services/dns/dns_darwin.go
@@ -42,12 +42,16 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
 		return *result, err
 	}
 	if needRestart {
-		// Restart the Network but in case of error log it to error info.
-		// If we make it blocking call then in offline use case for mac is
-		// always going to be broken.
 		logging.InfoF("Restarting the host network")
 		if err := restartNetwork(); err != nil {
-			logging.ErrorF("Restarting the host network failed: %v", err)
+			result.Success = false
+			return *result, fmt.Errorf("Restarting the host network failed: %v", err)
+		}
+		// Wait for the Network to come up but in the case of error, log it to error info.
+		// If we make it as fatal call then in offline use case for mac is
+		// always going to be broken.
+		if err := waitForNetwork(); err != nil {
+			logging.Error(err)
 		}
 	} else {
 		logging.InfoF("Network restart not needed")
@@ -98,7 +102,7 @@ func restartNetwork() error {
 		}
 	}
 
-	return waitForNetwork()
+	return nil
 }
 
 // Wait for Network wait till the network is up, since it is required to resolve external dnsquery
--- a/pkg/crc/services/dns/dns_darwin.go
+++ b/pkg/crc/services/dns/dns_darwin.go
@@ -42,11 +42,12 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
 		return *result, err
 	}
 	if needRestart {
-		// Restart the Network on mac
+		// Restart the Network but in case of error log it to error info.
+		// If we make it blocking call then in offline use case for mac is
+		// always going to be broken.
 		logging.InfoF("Restarting the host network")
 		if err := restartNetwork(); err != nil {
-			result.Success = false
-			return *result, err
+			logging.ErrorF("Restarting the host network failed: %v", err)
 		}
 	} else {
 		logging.InfoF("Network restart not needed")
@@ -78,6 +79,7 @@ func createResolverFile(InstanceIP string, path string) (bool, error) {
 	return crcos.WriteFileIfContentChanged(path, resolverFile.Bytes(), 0644)
 }
 
+// restartNetwork is required to update the resolver file on OSx.
 func restartNetwork() error {
 	// https://medium.com/@kumar_pravin/network-restart-on-mac-os-using-shell-script-ab19ba6e6e99
 	netDeviceList, _, err := crcos.RunWithDefaultLocale("networksetup", "-listallnetworkservices")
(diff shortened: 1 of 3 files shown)


pkg/crc/services/dns/dns_darwin.go
```diff
@@ -42,11 +41,12 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
...
 	if needRestart {
-		// Restart the Network on mac
+		// Restart the Network but in case of error log it to error info.
+		// If we make it blocking call then in offline use case for mac is
+		// always going to be broken.
 		logging.InfoF("Restarting the host network")
 		if err := restartNetwork(); err != nil {
-			result.Success = false
-			return *result, err
+			logging.ErrorF("Restarting the host network failed: %v", err)
```
anjannath: :+1:

cfergeau: Alternatively, you could move the `waitForNetwork()` call here and ignore its errors, or return a different error type from `restartNetwork` depending on what failed.
Given the need to ignore the error, I'm under the impression that "ping 8.8.8.8" is not the check we want for checking if network restart is done or not (and I'd rather we don't let google know every time we start a crc VM on mac).

cfergeau:
In the log, I'd have a preference for using "fatal", "will not start", "will fail to start" rather than "blocking"/"non-blocking" which I associate with "we are going to start a go routine to handle network restarts"

pkg/crc/services/dns/dns_darwin.go
praveenkumar: > Alternatively, you could move the waitForNetwork() call here and ignore its errors, or return a different error type from restartNetwork depending on what failed.

@cfergeau you mean something like this?
```golang
if err := restartNetwork(); err != nil {
	logging.ErrorF("Restarting the host network failed: %v", err) => don't ignore this
}
if err := waitForNetwork(); err != nil {
         logging.ErrorF(err) => ignore it here ?
}
```

praveenkumar: > Given the need to ignore the error, I'm under the impression that "ping 8.8.8.8" is not the check we want for checking if network restart is done or not (and I'd rather we don't let google know every time we start a crc VM on mac).

@cfergeau what do you suggest a reliable nameserver we can use instead using google one?

cfergeau: > > Alternatively, you could move the waitForNetwork() call here and ignore its errors, or return a different error type from restartNetwork depending on what failed.
> 
> @cfergeau you mean something like this?
> 
> ```go
> if err := restartNetwork(); err != nil {
> 	logging.ErrorF("Restarting the host network failed: %v", err) => don't ignore this
> }
> if err := waitForNetwork(); err != nil {
>          logging.ErrorF(err) => ignore it here ?
>```

yes :)

cfergeau: > > Given the need to ignore the error, I'm under the impression that "ping 8.8.8.8" is not the check we want for checking if network restart is done or not (and I'd rather we don't let google know every time we start a crc VM on mac).
> 
> @cfergeau what do you suggest a reliable nameserver we can use instead using google one?

As I understand it, the problem we are trying to solve is that the network restart is asynchronous, and we need a way to know when the network is fully restarted and is up. We decided to do that by pinging a well known IP (to avoid potential DNS resolving issues), so we picked 8.8.8.8. for that. 4.2.2.2 might be slightly better (but at this point that's really nitpicking).

"ping to external IP addresses" might even be failing on setups where network connectivity is fine (proxy, machine without internet connection, ...). We could try to build something around `networksetup -getinfo` instead, but there will probably be cornercases where it will report failure when it should not. I dunno if native macos API give a way to handle this. I guess a better long-term solution will be to avoid this network restart so that we don't have to wait for it to complete ;)

praveenkumar: @cfergeau according to https://www.tummy.com/articles/famous-dns-server/ article (published in 2010) I shouldn't use `4.2.2.2` unless I am a customer of Level-3 this is not deliberately setup as a public dns server like google one.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -104,21 +101,24 @@ func restartNetwork() error {
...
 }
 
-// Wait for Network wait till the network is up, since it is required to resolve external dnsquery
-func waitForNetwork(hostResolv *network.ResolvFileValues) error {
+// Wait for Network wait till the network is up in case of a network restart
+// since it is required to resolve external dnsquery
+// We just try to ping google public nameserver '8.8.8.8' and if ping is not
+// succeed then return the error
+func waitForNetwork() error {
+	ns := "8.8.8.8"
```
gbraad: not going to work for China (and possibly Turkey).

Needs to be configurable

pkg/crc/services/dns/dns_darwin.go
cfergeau: You should not use it as your day to day DNS server to resolve DNS queries unless you are a level-3 customer. The way I read this article is that you can use it to check for network connectivity.

cfergeau:
Thinking some more about this issue, I would deal with it this way https://github.com/cfergeau/crc/commits/issue328
Go back to try to ping the nameservers from /etc/resolv.conf *after* the network has been restarted as was done before commit 5419bb458c. As /etc/resolv.conf can be temporarily missing during network restarts, we need to try a few times to read it before giving up.
With that done, if there are still some situations when waitForNetwork() can fail, then we can just ignore its error, ie add this on top of that branch above:
```
diff --git a/pkg/crc/services/dns/dns_darwin.go b/pkg/crc/services/dns/dns_darwin.go
index b79517d..30e56cc 100644
--- a/pkg/crc/services/dns/dns_darwin.go
+++ b/pkg/crc/services/dns/dns_darwin.go
@@ -96,7 +96,9 @@ func restartNetwork() error {
                }
        }
 
-       return waitForNetwork()
+       waitForNetwork()
+
+       return nil
 }
 
 // Wait for Network wait till the network is up, since it is required to resolve external dnsquery
```

pkg/crc/services/dns/dns_darwin.go
cfergeau: >Unless you are a Level-3 customer, absolutely not. Google now has established public DNS servers at 8.8.8.8 which you should use. [...]  It's also a compelling choice if you just need to send out a ping to see if you can reach the Internet, if you don't have IPs you already know like I do. [...] However, some people set up 4.2.2.[1-6] as their standard DNS servers. Don't do this unless you are connected to Level 3.

praveenkumar:
@cfergeau @gbraad After giving it more thought, why we even need to check if network is started or not? We anyway have to support offline use case where it doesn't even matter. Can't we just get rid of that `waitforNetwork()` ?

cfergeau:
dns.go needs to read values from /etc/resolv.conf to copy them to the VM after calling runPostStartForOS so we need the network to be done restarting when we get there, or we might have the same issue with /etc/resolv.conf being missing.

praveenkumar:
> dns.go needs to read values from /etc/resolv.conf to copy them to the VM after calling runPostStartForOS so we need the network to be done restarting when we get there, or we might have the same issue with /etc/resolv.conf being missing.

@cfergeau we don't copy the host `/etc/resolv.conf` to the VM, what we do is read the VM's `/etc/resolv.conf` and add the internal dnsmasq nameserver to it and update that resolv.conf file.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -104,12 +102,25 @@ func restartNetwork() error {
...
 // Wait for Network wait till the network is up, since it is required to resolve external dnsquery
-func waitForNetwork(hostResolv *network.ResolvFileValues) error {
-	// retry for 5 times
+func waitForNetwork() error {
+	var hostResolv *network.ResolvFileValues
+	var err error
+
+	for i := 1; i <= 5; i++ {
+		hostResolv, err = network.GetResolvValuesFromHost()
+		if err == nil {
```
gbraad: :-P

Issue #339, opened 2019-07-23
Update DefaultHostName to latest bundle's host name


anjannath:
https://github.com/code-ready/crc/blob/ac25be8ee06d839d63257ec137adae485647e344/pkg/crc/constants/constants.go#L19

anjannath:
Closed via #340

PR #340, opened 2019-07-23
Issue #339 Update host name of master node

anjannath:


pkg/crc/constants/constants.go
```diff
@@ -16,7 +16,7 @@ const (
 
 	CrcEnvPrefix = "CRC"
 
-	DefaultHostname      = "crc-cvgnz-master-0"
+	DefaultHostname      = "crc-rv4tg-master-0"
```
praveenkumar: 👍

Issue #349, opened 2019-07-24
Add separate hooks for stderr and stdout for logrus


praveenkumar:
Right now, we are putting everything in the `stdout` as part of logging which is not the good thing to do since `stderr` should go to stderr instead stdout. In logrus we can have multiple hooks to put the logs where it belong.

https://github.com/code-ready/crc/blob/master/pkg/crc/logging/logging.go#L108

anjannath:
Fixed in #350

PR #350, opened 2019-07-24
Issue #349 Add separate hooks for stderr/stdout for logrus

praveenkumar:
As of now, we are dumping everything in the stdout which
means the error logs are also part of it instead part of
stderr which makes integration testing tedious and not
we should do as per unix standards.

### DIFF
--- a/pkg/crc/logging/logging.go
+++ b/pkg/crc/logging/logging.go
@@ -1,14 +1,13 @@
 package logging
 
 import (
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh/terminal"
 )
 
@@ -17,41 +16,6 @@ var (
 	LogLevel      string
 	originalHooks = logrus.LevelHooks{}
 )
-
-type fileHook struct {
-	file      io.Writer
-	formatter logrus.Formatter
-	level     logrus.Level
-}
-
-func newFileHook(file io.Writer, level logrus.Level, formatter logrus.Formatter) *fileHook {
-	return &fileHook{
-		file:      file,
-		formatter: formatter,
-		level:     level,
-	}
-}
-
-func (h fileHook) Levels() []logrus.Level {
-	var levels []logrus.Level
-	for _, level := range logrus.AllLevels {
-		if level <= h.level {
-			levels = append(levels, level)
-		}
-	}
-
-	return levels
-}
-
-func (h *fileHook) Fire(entry *logrus.Entry) error {
-	line, err := h.formatter.Format(entry)
-	if err != nil {
-		return err
-	}
-
-	_, err = h.file.Write(line)
-	return err
-}
 
 func OpenLogFile() (*os.File, error) {
 	l, err := os.OpenFile(filepath.Join(constants.LogFilePath), os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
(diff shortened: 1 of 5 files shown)


pkg/crc/logging/logging.go
```diff
@@ -1,14 +1,13 @@
...
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
gbraad: eh, no?

pkg/crc/logging/logging.go
```diff
@@ -1,14 +1,13 @@
...
 import (
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/pkg/errors"
```
gbraad: pkg/errors?

pkg/crc/logging/logging.go
```diff
@@ -1,14 +1,13 @@
...
 
 import (
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
```
gbraad: perhaps this one should go down in a new section?

pkg/crc/logging/stdin_hook.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"io"
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
+
+// This is stdOutHook to send everything except error and Fatal to standard output.
+type stdOutHook struct {
+	stderr    io.Writer
```
anjannath: Typo here `stderr => stdout`  and this file should be called `stdout_hook.go`, also you could probably make this `Hook stuct` generic and reuse it in all the hooks:
```golang
type logrusHook struct {
      writer       io.Write
      formatter    logrus.Formatter
      level        logrus.Level
}

File: test/integration/features/config.feature [hunk 5]
```diff
@@ -122,7 +122,7 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
        And executing "crc config set skip-check-crc-network-active true" succeeds
        # Start CRC
        Then starting CRC with default bundle and default hypervisor fails
-       And stdout contains "Network not found: no network with matching name 'crc'"
+       And stderr contains "network crc not found"
```
jsliacan: I'd go with `no network with matching name` as `network crc not found` is not part of the error message. Also, there's a typo in that error message in `Use 'crc setup' for define the network`. Needs 'to' instead of 'for'. But I guess that doesn't belong into this PR...

pkg/crc/logging/logging.go
praveenkumar: @gbraad didn't get you we want to have different section for imported packages from what std library have so I just move those 2 around, nothing changed.

pkg/crc/logging/logging.go
anjannath: This is only used in one place, can probably be changed to use crc `errors` package
https://github.com/code-ready/crc/blob/6a3e401528bb982ddf1f3a5d26995b3d792f706b/pkg/crc/logging/logging.go#L82

pkg/crc/logging/stdin_hook.go
praveenkumar: @anjannath we can't reuse the struct since all those struct need to satisfy the `Hook` interface so the method should be part of struct.

```
type Hook interface {
	Levels() []Level
	Fire(*Entry) error
}
```

Issue #362, opened 2019-07-30
Abstract retry logic from different places and use it.


praveenkumar:
Right now we are having retry logic different places for different function run, need to abstract it so we can use it for every function.

gbraad:
can you be specific about which we use/have?

praveenkumar:
@gbraad Right now we have it on multiple places https://github.com/code-ready/crc/pull/363 I am still missing some of the places which I am adding slowing but this will give you an idea.

anjannath:
Fixed via #363

PR #363, opened 2019-07-30
Issue #362 Abstract retry logic

praveenkumar:
Multiple places, `for i; i < num; i++` logic is used to
have retry which is non abstracted and also doesn't collect
errors if a function have multiple error. Abstracting this
logic now make it possible to have a callback function which
can be retried for a specific interval.

### DIFF
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -3,6 +3,9 @@ package errors
 import (
 	"fmt"
 	"strings"
+	"time"
+
+	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
 type MultiError struct {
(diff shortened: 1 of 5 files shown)


pkg/crc/errors/multierror.go
```diff
@@ -26,3 +29,34 @@ func (m MultiError) ToError() error {
...
+func (r RetriableError) Error() string {
+	return "Temporary Error: " + r.Err.Error()
+}
+
+// RetryAfter retries a number of attempts, after a delay
+func RetryAfter(attempts int, callback func() error, d time.Duration) (err error) {
+	m := MultiError{}
+	for i := 0; i < attempts; i++ {
+		if i > 0 {
+			log.Debugf("retry loop %d", i)
```
anjannath: Any reason we are using libmachine log package here, instead of crc logging package? The only problem i could see is when this is used outside of the machine package, where we are not setting libmachine logging!

praveenkumar: @anjannath Done.

PR #364, opened 2019-07-30
Add HyperV support

gbraad:


tsedmik:
I have encountered the following issue on Windows:
```
INFO To access the cluster using 'oc', run 'oc login -u kubeadmin -p 9gZtH-EWZLT-mZNFP-bkKwm https://api.crc.testing:6443'
INFO Access the OpenShift web-console here: https://console-openshift-console.apps-crc.testing
INFO Login to the console with user: kubeadmin, password: 9gZtH-EWZLT-mZNFP-bkKwm
ERRO Error occurred: Error approving the node csr Not able to get csr names (exit status 1 : Unable to connect to the server: dial tcp: lookup api.crc.testing: no such host
)
```
For some reason the nameserver is not set correctly
_CRC did not ask for admin privileges to set it properly_ 
```
PS C:\Users\Tomáš Sedmík\tmp> get-dnsclientserveraddress "vEthernet (Default Switch)"

InterfaceAlias               Interface Address ServerAddresses
                             Index     Family
--------------               --------- ------- ---------------
vEthernet (Default Switch)          61 IPv4    {127.0.0.1}
vEthernet (Default Switch)          61 IPv6    {::1}
```
I'm able to manually set the correct server address with `set-dnsclientserveraddress "vEthernet (Default Switch)" -ServerAddress ("172.17.190.28")` (in elevated powershell). This change is not persistent and is changed back to 127.0.0.1 after while :-(

gbraad:
> This change is not persistent and is changed back to 127.0.0.1 after while :-(

This needs some debugging... I would like to know what else you have on this system

gbraad:
@tsedmik @anjannath please add comments of your runs

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -1,11 +1,25 @@
...
 	"github.com/code-ready/crc/pkg/crc/oc"
+
+	"github.com/code-ready/crc/pkg/crc/errors"
+	//"github.com/code-ready/crc/pkg/os/windows/win32"
+	"github.com/code-ready/crc/pkg/os/windows/powershell"
+)
+
+const (
+	// Fall Creators update comes with the "Default Switch"
+	minimumWindowsReleaseId = 1709
```
gbraad: @robin-owen This is important to document, as this is not a step we can automate for people. Sure, 1709 was released in 2017 (Sept), but not guaranteed to be installed.

robin-owen: Noted, thank you. Once this is merged, we can create a docs issue to get that documented.

anjannath:
With the new implementation using `netsh` it successfully adds the dns entries to the interface.

tsedmik:
CRC works well with the new implementation using `netsh` 👍

tsedmik:
> 
> 
> > This change is not persistent and is changed back to 127.0.0.1 after while :-(
> 
> This needs some debugging... I think it is persisted, but gets modified. I would like to know what else you have on this system?

I found out the reason what is responsible for changes server address to `127.0.0.1`. `WSL` is responsible for that.

gbraad:
> `WSL` is responsible for that.

This does not happen with WSL2, so perhaps this can be a recommendation in future (as it uses it's own vswitch named `WSL`).

tsedmik:
I was wrong. `WSL` is not the root cause. I forgot my VPN. I use `Viscosity` for VPN access. Once I turn on the VPN, server address is changed to 127.0.0.1. `WSL` is innocent in this case.

gbraad:
> VPN

OK, as this was expected to begin with. I would wonder what Viscosity tries to accomplish with this.

pkg/crc/preflight/preflight_checks_windows.go
gbraad: @robin-owen please go ahead.

but also we need to document that VPN is currently unsupported.

Issue #385, opened 2019-08-06
Document minimum Windows version for Hyper-V


robin-owen:
We need the "Default Switch" feature for Hyper-V.

Minimum required version is Windows 10 version 1709 (Fall Creators Update).

robin-owen:
Fixed by #386.

PR #386, opened 2019-08-06
Issue #385 Document minimum Windows version

robin-owen:


docs/source/getting-started/content/proc_installing-codeready-containers.adoc
```diff
@@ -14,7 +14,10 @@ These requirements must be met in order to run OpenShift in the {prod} virtual m
 Other applications running on top of OpenShift may require additional resources to run.
 ====
 
-* {prod} requires installation of the following software packages:
+* On {msw}, {prod} requires the Windows 10 Fall Creators Update (version 1709).
```
robin-owen: Should we mention Hyper-V?

anjannath: Its Okay not to mention about Hyper-V. @gbraad WDYT? we don't mention `kvm` module should be loaded. These are also taken care of by the preflight checks.

gbraad: This is OK. We require Windows 10. No need to be SOOOOOOO precise to detail why, as that is part of the rest of the installation. The native target already points to this 'need'.

cfergeau:
And we need the 64 bit edition, 32 bit is not going to be enough for hyperv.

PR #392, opened 2019-08-07
Issue #391 add user to Hyper-V group with net command

### LINKED ISSUE #391: Use net command and win32 api to add user to group

gbraad:
Fixes #391 #398

### DIFF
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -3,6 +3,7 @@ package preflight
 import (
 	//"errors"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 
@@ -10,7 +11,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/oc"
 
 	"github.com/code-ready/crc/pkg/crc/errors"
-	"github.com/code-ready/crc/pkg/os/windows/win32"
 	"github.com/code-ready/crc/pkg/os/windows/powershell"
 )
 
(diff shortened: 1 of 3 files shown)


tsedmik:
It doesn't work with localized usernames - `Tomáš Sedmík` for example

gbraad:
I'll have a look as I have seen this behaviour before, where the name is changed to only use ASCII characters, such as `Tomas Sedmik`

cfergeau:
I've added an additional commit on top of this which should cope better with non-ascii usernames
https://github.com/cfergeau/crc/tree/adduser

gbraad:
Will pull and add to this PR

tsedmik:
It works on my end now 👍 . Tested with `0.89.1-alpha-4.1.6+b05baa3` and `crc.exe setup` successfully added `Hyper-V Administrators` group to my account `Tomáš Sedmík`.

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -115,16 +115,12 @@ func fixUserPartOfHyperVAdmins() (bool, error) {
...
-		return false, errors.New("Unable to get user name")
-	}
-	username := strings.TrimSpace(outUsername)
+	username := os.Getenv("USERNAME")
 
 	netCmdArgs := fmt.Sprintf(`([adsi]"WinNT://./%s,group").Add("WinNT://%s,user")`, groupName, username)
 	_, _, err = powershell.ExecuteAsAdmin(netCmdArgs)
 	if err != nil {
-		return false, errors.New("Error adding user to group")
+		return false, errors.New("Unable to get user name")
```
gbraad: @teuf, why you mention the username here?

anjannath: I guess this is in case `os.Getenv` returned an empty string, this would fail..

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -115,16 +115,12 @@ func fixUserPartOfHyperVAdmins() (bool, error) {
 	}
 	groupName := strings.TrimSpace(strings.Replace(strings.TrimSpace(outGroupName), "BUILTIN\\", "", -1))
 
-	outUsername, _, err := powershell.Execute(`Write-Host $env:USERNAME`)
-	if err != nil {
-		return false, errors.New("Unable to get user name")
-	}
-	username := strings.TrimSpace(outUsername)
+	username := os.Getenv("USERNAME")
```
anjannath: :+1:

gbraad:
I will merge based on feedback. Just not sure why the error is changed... but then again, when this fails, it will never be directly about the execution of the escalated script, but the ACTUAL initial invocation.

pkg/crc/preflight/preflight_checks_windows.go
cfergeau: It's just that I forgot to change the error message, this block was preexisting before the commit.

PR #397, opened 2019-08-09
Issue #396 Add windows target to embed bundle in the binary

### LINKED ISSUE #396: Add windows target to embed bundle and improve the bundle not found error message
Currently we just show a warning message like below, which is not helpful, as there's no information about it in `crc --help`.
```
WARN stat /home/anjan/.crc/crc_libvirt_4.1.6.crcbundle: no such file or directory
```
We shouldn't show an error message if a bundle was passed using `-b` flag.

anjannath:
Don't show an warning message if the bundle was passed to crc
start command. If no bundle was passed to start command, ask
user to run setup command to unpack the bundle to disk

Fixes #396


pkg/crc/validation/validation.go
```diff
@@ -43,7 +43,7 @@ func ValidateMemory(value int) error {
 // ValidateBundle checks if provided bundle path exist
 func ValidateBundle(bundle string) error {
 	if err := ValidatePath(bundle); err != nil {
-		return err
+		return errors.Newf("Run 'crc setup' to unpack the bundle to disk.")
```
gbraad: This actually makes sense...

PR #402, opened 2019-08-09
Issue #400 Add sha256sum to the final release binaries

### LINKED ISSUE #400: Add sha256sum for make cross and also for release makefile target.
Right now, we are not creating the sha256sum for the cross builds and also for the release target in the Makefile, which we mostly require.

praveenkumar:
Fixes #400 #449

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,20 +125,20 @@ fmtcheck: ## Checks for style violation using gofmt
 release: fmtcheck embed_bundle
 	mkdir $(RELEASE_DIR)
 	
-	@mkdir -p $(BUILD_DIR)/crc-$(CRC_VERSION)-darwin-amd64
-	@cp LICENSE $(BUILD_DIR)/darwin-amd64/crc $(BUILD_DIR)/crc-$(CRC_VERSION)-darwin-amd64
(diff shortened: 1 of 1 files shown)


gbraad:
@code-ready/crc-devel WDYT?

cfergeau:
A single file with hashes for all files seems fairly typical, so yes, makes sense.

praveenkumar:
@gbraad @cfergeau If you check odo/oc directory structure it make sense since they are putting every bits in the same root directory but we want to put in different directory as per binary (like macos, linux, windows) so I think to have that for each dir binary may be better than having it on top directory in a single file?

gbraad:
The point would be, we aren't linking to the sha256 file from the landing page. So separation does not make a lot of sense. This is a service for advanced users ... or to describe in the troubleshooting guide. So, I do think a single file still makes more sense.

gbraad:
>  but we want to put in different directory as per binary (like macos, linux, windows)

No... The suggestion was to name files `crc-windows.zip`, `crc-macos.tar.gz`, and `crc-linux.tar.gz` and the folder would be versioned and collection all the deliverables.

anjannath:
You can also probably take care of #449 here!

PR #428, opened 2019-08-19
Issue #393 Add job to circleci for publishing docs to gh-pages

### LINKED ISSUE #393: Automate publishing of docs to https://code-ready.github.io/crc

anjannath:
This will only push the docs update when the changes are
merged to master branch.

Fixes #393

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,3 +45,73 @@ jobs:
     - store_artifacts:
         path: ~/project/out
         destination: crc
+
+  build_docs:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+    - checkout
+    - run:
+        name: Update repo
+        command: sudo apt -y update
+    - run:
+        name: Add podman ppa
+        command: sudo add-apt-repository -y ppa:projectatomic/ppa && sudo apt -y update
+    - run:
+        name: Update repo
+        command: sudo apt -y update
+    - run:
+        name: Install podman
+        command: sudo apt -y install podman
+    - run:
+        name: Install slirp4netns
+        command: sudo apt -y install slirp4netns
+    - run:
+        name: Check podman
+        command: podman info
+    - run:
+        name: Make build_docs
+        command: make build_docs
+    - persist_to_workspace:
+        root: docs/build
+        paths: index.html
+
+  docs_deploy:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+    - attach_workspace:
+        at: docs/build
+    - add_ssh_keys:
+        fingerprints:
+          - "c5:64:d4:79:8f:c7:9b:80:b8:16:71:20:57:6b:8c:94"
+    - run:
+        name: Configure git
+        command: |
+          cd docs/build/
+          git init
+          git remote add origin git@github.com:code-ready/crc.git
+          git config user.email "codereadycontainers@gmail.com"
+          git config user.name "crc-bot"
+    - run:
+        name: Deploy docs to gh-pages branch
+        command: |
+          cd docs/build/
+          git checkout -b gh-pages
+          git add .
+          git commit -m "[skip ci] Docs update"
+          git push origin HEAD -f
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - build
+      - build_docs
+      - docs_deploy:
+          requires:
+            - build
+            - build_docs
+          filters:
+            branches:
+              only: master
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -45,3 +45,64 @@ jobs:
...
+          - "c5:64:d4:79:8f:c7:9b:80:b8:16:71:20:57:6b:8c:94"
+    - run:
+        name: Install and configure dependencies
+        command: |
+          npm install -g --silent gh-pages@2.0.1
+          git config user.email "codereadycontainers@gmail.com"
+          git config user.name "crc-bot"
+    - run:
+        name: Deploy docs to gh-pages branch
+        command: gh-pages -d docs/build/* -s *.html --message "[skip ci] Docs update"
```
anjannath: This commit message is needed to stop circle ci running the job again, but since we don't have a circle ci config in the gh-pages branch, if ran it'll always fail.

.circleci/config.yml
```diff
@@ -45,3 +45,64 @@ jobs:
...
+  docs_deploy:
+    docker:
+      - image: node:8.10.0
+    steps:
+    - checkout
+    - attach_workspace:
+        at: docs/build
+    - add_ssh_keys:
+        fingerprints:
+          - "c5:64:d4:79:8f:c7:9b:80:b8:16:71:20:57:6b:8c:94"
```
anjannath: The corresponding public key is added as a deployment key to crc repo

anjannath:
The Centos CI failure is the known one, but re-starting to have everything green
retest this please

gbraad:
why is an SSH key needed? I thought it should operate with an application token. didn;t we do this for Minishift?

anjannath:
We could also do it using a OAuth token, but in that case the token will be associated with an user account account by using a ssh/deploy key we are not associating with an user accounts but directly with the crc repo, both will need write access so would be similar.

.circleci/config.yml
praveenkumar: @anjannath I am not able to see where are we pushing the resulted html to `gh-pages` branch does that automatic pushed there or we have to specify it someplace?

gbraad:
> Never used — Read/write 

Does this work?

.circleci/config.yml
anjannath: @praveenkumar Its pushed in line 94, the `gh-pages` is a nodejs tool, which is doing it

gbraad: > `npm install -g --silent gh-pages@2.0.1`

I do not see WHY we need this. it is just a `git` action to push to an orphaned branch.

anjannath:
Yup, i tested this with my fork and it pushed the changes..

.circleci/config.yml
anjannath: Remove the dependency, using just `git` now

Makefile
```diff
@@ -95,7 +95,7 @@ test:
 
 .PHONY: build_docs
 build_docs:
-	podman run -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_BUILD_CONTAINER) -b html5 -D $(DOCS_BUILD_DIR) $(DOCS_BUILD_TARGET)
+	podman run -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_BUILD_CONTAINER) -b html5 -D $(DOCS_BUILD_DIR) -o index.html $(DOCS_BUILD_TARGET)
```
praveenkumar: @anjannath you need to run the podman with `sudo` otherwise it is not able to get the network capability, I think it is fixed in later version of podman but package which is part of ubuntu repo might not the updated one. you can see the error from the CI.

```
Storing signatures
ERRO[0014] could not find slirp4netns, the network namespace won't be configured: exec: "slirp4netns": executable file not found in $PATH 
```

Also I am not sure why this job is not failed even podman exited with an error (not sure if the exit code is correct here).

anjannath: This is strange, but it does create the `index.html` file not sure how, when i tested it pushed that file to my branch, instead of running using sudo let me try installing that package.

praveenkumar:
retest this please.

.circleci/config.yml
```diff
@@ -45,3 +45,73 @@ jobs:
...
+          git commit -m "[skip ci] Docs update"
+          git push origin HEAD -f
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - build
+      - build_docs
+      - docs_deploy:
```
praveenkumar: 👍

Issue #439, opened 2019-08-20
Don't install libvirt machine driver in /usr/local


cfergeau:
Since the hyperkit changes, it's possible to install the machine drivers locally in ~/.crc/bin

PR #440, opened 2019-08-20
Issue #439: Install libvirt driver in ~/.crc/bin

cfergeau:
It's currently installed in /usr/local/bin, but since the changes for
the hyperkit driver, we can install the machine drivers in a private
location (~/.crc/bin). This commit changes the libvirt driver to go
there as well, as this means less priviledged operation to do at crc
setup time.

This fixes https://github.com/code-ready/crc/issues/439

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"golang.org/x/sys/unix"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -23,7 +24,6 @@ import (
 )
 
 const (
-	driverBinaryDir             = "/usr/local/bin"
 	libvirtDriverCommand        = "crc-driver-libvirt"
 	libvirtDriverVersion        = "0.12.4"
 	crcDnsmasqConfigFile        = "crc.conf"
@@ -31,9 +31,8 @@ const (
 )
 
 var (
-	libvirtDriverBinaryPath = filepath.Join(driverBinaryDir, libvirtDriverCommand)
-	crcDnsmasqConfigPath    = filepath.Join(string(filepath.Separator), "etc", "NetworkManager", "dnsmasq.d", crcDnsmasqConfigFile)
-	crcDnsmasqConfig        = `server=/crc.testing/192.168.130.1
+	crcDnsmasqConfigPath = filepath.Join(string(filepath.Separator), "etc", "NetworkManager", "dnsmasq.d", crcDnsmasqConfigFile)
+	crcDnsmasqConfig     = `server=/crc.testing/192.168.130.1
 address=/apps-crc.testing/192.168.130.11
 `
 	crcNetworkManagerConfigPath = filepath.Join(string(filepath.Separator), "etc", "NetworkManager", "conf.d", crcNetworkManagerConfigFile)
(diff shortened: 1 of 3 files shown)


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -217,51 +217,35 @@ func fixLibvirtServiceRunning() (bool, error) {
...
 		return false, err
 	}
-	fi, err := os.Stat(path)
-	if err != nil {
-		return false, err
-	}
-	// Check if permissions are correct
-	if fi.Mode()&0011 == 0 {
-		return false, errors.New("crc-driver-libvirt does not have correct permissions")
-	}
```
cfergeau: I can split this change in a separate commit if needed, this changes the way we check if the binary exists/is executable, but it did not have to be part of this commit (this is the same code as in the hyperkit driver).

praveenkumar:
retest this please

gbraad:
This however also needs a possible removal of the previous driver? Or at least test for it somehow?
We do take the driver from the crc/bin directory as priority, right? but we created litter on someone's computer with previous setup runs that stored in /usr/local

cfergeau:
> This however also needs a possible removal of the previous driver? Or at least test for it somehow?
> We do take the driver from the crc/bin directory as priority, right? but we created litter on someone's computer with previous setup runs that stored in /usr/local

Yes, ~/.crc/bin takes priority (I forgot if $PATH will even be considered when we explicitly set the binary path). There will indeed be some leftovers, but we currently have nothing in place for "unsetup" :( I can look at adding a "sudo rm /usr/local/bin/crc-driver-libvirt" somewhere for now.

cfergeau:
Another issue with this change is on upgrades. If you run `crc start` right after an upgrade, it will fail to locate crc-driver-libvirt, and will tell you to run `crc setup`. But `crc setup` can make unwanted changes (in particular with networking), so the preexisting VM may no longer be functional after that. I'm not exactly sure how to avoid this specific problem (I'd prefer to avoid adding changes to look for the driver either in the path we specified, or in $PATH). Maybe for now (pre-beta, no public bundle releases), we can have "non-perfect" upgrades?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -248,6 +248,31 @@ func fixMachineDriverLibvirtInstalled() (bool, error) {
...
+	}
+	logging.Debugf("No older %s installation found", libvirtDriverCommand)
+
+	return true, nil
+}
+
+func fixOldMachineDriverLibvirtInstalled() (bool, error) {
+	oldLibvirtDriverPath := filepath.Join("/usr/local/bin/", libvirtDriverCommand)
+	logging.Debugf("Removing %s", oldLibvirtDriverPath)
+	_, _, err := crcos.RunWithPrivilege("rm", oldLibvirtDriverPath)
```
praveenkumar: @cfergeau shouldn't it `rm -f` otherwise it might prompt user for provide `Y/N` option?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -248,6 +248,31 @@ func fixMachineDriverLibvirtInstalled() (bool, error) {
 	return true, nil
 }
 
+func checkOldMachineDriverLibvirtInstalled() (bool, error) {
```
praveenkumar: @cfergeau Can we add comment on both the functions that this we should remove/revert after 2-3 release since after that no one using older version of crc due to cert issues.

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: I had no such issue during my testing, but it will be better with -f, I'll add it ;)

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: This will not cover someone who briefly tested crc now, and whose next test will be in 6 months. but I agree we can remove the test in a while.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -217,51 +216,61 @@ func fixLibvirtServiceRunning() (bool, error) {
...
+func fixOldMachineDriverLibvirtInstalled() (bool, error) {
+	oldLibvirtDriverPath := filepath.Join("/usr/local/bin/", libvirtDriverCommand)
+	logging.Debugf("Removing %s", oldLibvirtDriverPath)
+	_, _, err := crcos.RunWithPrivilege("rm", "-f", oldLibvirtDriverPath)
 	if err != nil {
-		return false, fmt.Errorf("%s %v: %s", stdOut, err, stdErr)
+		logging.Debugf("Removal of %s failed", oldLibvirtDriverPath)
+		/* Ignoring error, an obsolete file being still present is not a fatal error */
+	} else {
+		logging.Debugf("%s successfully removed", oldLibvirtDriverPath)
```
praveenkumar: 👍

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -217,51 +216,61 @@ func fixLibvirtServiceRunning() (bool, error) {
...
 	logging.Debugf("Checking if %s is installed", libvirtDriverCommand)
+
 	// Check if crc-driver-libvirt is available
-	path, err := exec.LookPath(libvirtDriverCommand)
-	if err != nil {
-		return false, err
-	}
-	fi, err := os.Stat(path)
+	libvirtDriverPath := filepath.Join(constants.CrcBinDir, libvirtDriverCommand)
+	err := unix.Access(libvirtDriverPath, unix.X_OK)
```
praveenkumar: @cfergeau As per `unix` package guide it states `The primary use of this package is inside other packages that provide a more portable interface to the system, such as "os", "time" and "net". Use those packages rather than this one if you can.`. I am OK right now to using it here since this code is not going to run other than linux systems but we have to be careful which all places we should use.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -217,51 +216,61 @@ func fixLibvirtServiceRunning() (bool, error) {
...
+	logging.Debugf("%s is installed in %s", libvirtDriverCommand, constants.CrcBinDir)
 
-	logging.Debugf("Copying %s in %s", libvirtDriverCommand, libvirtDriverBinaryPath)
-	stdOut, stdErr, err := crcos.RunWithPrivilege("mkdir", "-p", driverBinaryDir)
-	if err != nil {
-		return false, fmt.Errorf("%s %v: %s", stdOut, err, stdErr)
+	return true, nil
+}
+
+/* These 2 checks can be removed after a few releases */
```
praveenkumar: 👍

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: I know this is not portable, but I could not find something similar in `os` (the `stat` checks we have/had in the codebase to try and do that had various issues, in particular with 0700 files owned by $anotheruser, or owned by the current user)

praveenkumar: @cfergeau I didn't get the ` in particular with 0700 files owned by $anotheruser, or owned by the current user` this one you mean os.Stat doesn't provide correct info around it?

cfergeau: I was somehow referring to the code I'm removing, it's testing if the executable bit is set for either group or other. This test will unexpectedly fail if the file is 700 and owned by the current user, or and it will succeed when it should not if the file is 070 but the user is not part of the appropriate group. I only wanted to point out it's not simple to achieve the same as unix.Access using os.Stat.

PR #446, opened 2019-08-21
Issue #434 Remove hardcoded block device name when getting disk usage

### LINKED ISSUE #434: crc.exe status does not work on running instance
**Environment**
* CRC version - 0.90.0-alpha-4.1.9+b2559b5
* Windows 10 Pro 1903, Hyper-V

If I start CRC and then try to determine it's state with `crc.exe status`, I get the following error:

```
PS C:\Users\Tomáš Sedmík\Downloads\crc-0.90.0-alpha-windows-amd64> .\crc.exe status --log-level=debug
DEBU Unexpected operator status for authentication: Upgradeable
DEBU Unexpected operator status for console: Upgradeable
DEBU Unexpected operator status for kube-apiserver: Upgradeable
DEBU Unexpected operator status for kube-controller-manager: Upgradeable
DEBU Unexpected operator status for kube-scheduler: Upgradeable
DEBU Unexpected operator status for openshift-apiserver: Upgradeable
DEBU Unexpected operator status for openshift-controller-manager: Upgradeable
DEBU Unexpected operator status for service-ca: Upgradeable
DEBU Unexpected operator status for service-catalog-apiserver: Upgradeable
DEBU Unexpected operator status for service-catalog-controller-manager: Upgradeable
ERRO Error occurred: strconv.ParseInt: parsing "df:": invalid syntax
```

anjannath:
Get the usage details by grepping the mount point `/sysroot` instead
of depending on the device name.

Fixes #434

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -40,10 +40,8 @@ func CheckCertsValidity(driver drivers.Driver) error {
 }
 
 // Return size of disk, used space in bytes and the mountpoint
-func GetDiskUsage(driver drivers.Driver, device string) (int64, int64, error) {
-	cmd := fmt.Sprintf(
-		"df -B1 --output=size,used,target %s | tail -1",
-		device)
+func GetRootPartitionUsage(driver drivers.Driver) (int64, int64, error) {
+	cmd := "df -B1 --output=size,used,target /sysroot | tail -1"
 
 	out, err := drivers.RunSSHCommandFromDriver(driver, cmd)
 
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -424,7 +424,7 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
 			// TODO:get openshift version as well and add to status
 			openshiftStatus = fmt.Sprintf("Running (v4.x)")
 		}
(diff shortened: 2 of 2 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -40,10 +40,8 @@ func CheckCertsValidity(driver drivers.Driver) error {
 }
```
cfergeau: I would add a short description about the "why" to the commit log (depending on the hypervisor being used, the device name may not be /dev/vda3).

pkg/crc/cluster/cluster.go
```diff
@@ -40,10 +40,8 @@ func CheckCertsValidity(driver drivers.Driver) error {
...
 
 // Return size of disk, used space in bytes and the mountpoint
-func GetDiskUsage(driver drivers.Driver, device string) (int64, int64, error) {
-	cmd := fmt.Sprintf(
-		"df -B1 --output=size,used,target %s | tail -1",
-		device)
+func GetRootPartitionUsage(driver drivers.Driver) (int64, int64, error) {
+	cmd := "df -B1 --output=size,used,target | grep sysroot"
```
cfergeau: (alternatively, could be "df -B1 --output=size,used,target sysroot | tail -1", but either way is good)

anjannath: I thought this only takes the device names, and not the mount point, will use this one, looks cleaner no piping!

PR #451, opened 2019-08-22
Use golangci-lint in CI

cfergeau:
This patch series fixes the various issues reported by golangci-lint, and then enable it in the CI.
Some of these issues were actual bugs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,9 @@ jobs:
         name: Make fmtcheck
         command: make fmtcheck
     - run:
+        name: Run golangci-lint
+        command: make lint
+    - run:
         name: Make cross
         command: make cross
     - run:
@@ -71,7 +74,9 @@ jobs:
         command: podman info
     - run:
         name: Make build_docs
-        command: make build_docs
+        # force GOPATH to a sane value as the image we are using has an older go version
+        # with a 'go env GOPATH' output which causes make failures
+        command: make GOPATH="/home/circleci/go" build_docs
     - persist_to_workspace:
         root: docs/build
         paths: index.html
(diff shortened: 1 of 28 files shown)


praveenkumar:
@cfergeau All the commits are related to the `golangci-lint` can we squash those since these are anyway required and it's not like we are changing the function signature (iirc) ?

cfergeau:
> @cfergeau All the commits are related to the `golangci-lint` can we squash those since these are anyway required and it's not like we are changing the function signature (iirc) ?

I don't think they should be squashed, review is much easier when you can tell why a given change is being made (which would be harder with a single commit), and if one of these commit is wrong, bisection/reverting will be easier too.

.circleci/config.yml
```diff
@@ -71,7 +74,9 @@ jobs:
         command: podman info
     - run:
         name: Make build_docs
-        command: make build_docs
+        # force GOPATH to a sane value as the image we are using has an older go version
+        # with a 'go env GOPATH' output which causes make failures
+        command: make GOPATH="/home/circleci/go" build_docs
```
praveenkumar: 👍

.travis.yml
```diff
@@ -6,4 +6,5 @@ go:
 script:
 - make
 - make fmtcheck
+- make lint
```
praveenkumar: 👍

Makefile
```diff
@@ -69,7 +70,7 @@ vendor:
 	GO111MODULE=on go mod vendor
 
 # Get binappend
-binappend:
+$(GOPATH)/bin/binappend-cli:
```
praveenkumar: 👍

PR #456, opened 2019-08-23
[WIP] Issue #442 Print exp for changing encoding to UTF-8 in posh

### LINKED ISSUE #442: crc.exe oc-env does not work well with special characters in the path
**Environment**
* CRC version - 0.90.0-alpha-4.1.9+b2559b5
* Windows 10 Pro 1903, Hyper-V
* Home Dir: C:\Users\Tomáš Sedmík\

If I use `crc.exe oc-env` command to put oc binary on PATH, it does not work in case I have some special characters in the path - see:

```
PS C:\devel\crc> crc.exe oc-env
$Env:PATH = "C:\Users\Tomáš Sedmík\.crc\bin;$Env:PATH"
# & crc oc-env | Invoke-Expression

PS C:\devel\crc> crc.exe oc-env | Invoke-Expression

PS C:\devel\crc> Get-ChildItem Env:PAth

Name                           Value
----                           -----
Path                           C:\Users\Tom├í┼í Sedm├¡k\.crc\bin;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\Program Files\Git\cmd;C:\ProgramData\chocolatey\bin;C:\Program Files\Java\jdk1.8....
```

anjannath:
Posh does not use UTF encoding by default, which leads to
malformed paths in $Env:Path when user tries to add oc to
path using oc-env command. this only happens when there
are accented chars or any special chars are present.

Fixes #442  /cc @tsedmik

### DIFF
--- a/pkg/os/shell/shell.go
+++ b/pkg/os/shell/shell.go
@@ -40,7 +40,7 @@ func GenerateUsageHint(userShell, cmdLine string) string {
 
 	switch userShell {
 	case "powershell":
-		cmd = fmt.Sprintf("& %s | Invoke-Expression", cmdLine)
+		cmd = fmt.Sprintf("[System.Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8; & %s | Invoke-Expression", cmdLine)
 	case "cmd":
 		cmd = fmt.Sprintf("\t@FOR /f \"tokens=*\" %%i IN ('%s') DO @call %%i", cmdLine)
 		comment = "REM"


anjannath:
@gbraad I tried that, wasn't working on a single invocation , so when its ran it does change the encoding, but not for the same invocation, and you need to run that again.. :-/

gbraad:
can we not detect if the expression has non-ASCII characters? and ONLY do this when necessary? it does make the invoke a LOT unfriendlier.

besides, how about the output from Command Prompt? It should also affect @tsedmik

Issue #459, opened 2019-08-25
Add pre-flight check for NetworkManager


gbraad:
After libvirt we check for config files for NetworkManager

https://github.com/code-ready/crc/blob/f185afc39253747c07213ce213fa72b102a5edb6/pkg/crc/preflight/preflight_linux.go#L60

but this assumes that NetworkManager is used. However, it could be people use `network` (rc-scripts) or `systemd-networkd` or try to get this somehow on a headless system. to prevent this now, we should add a test similar to `checkLibvirtEnabled` but with a fix that asks for manual changes (like done with the Windows update) as automagic updates might result in a disconnected state.

anjannath:
Fixed in #461

PR #461, opened 2019-08-26
Issue #459 Add NetworkManager preflight checks for libvirt

praveenkumar:
NetworkManager is required to setup the dnsmasq for application
domain which should be used by the applications created on crc.
Sometime headless os doesn't have NetworkManager by default so
we should warn user before even start.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -455,3 +455,34 @@ func fixCrcNetworkManagerConfig() (bool, error) {
 	logging.Debug("NetworkManager configuration fixed")
 	return true, nil
 }
+func checkNetworkManagerInstalled() (bool, error) {
+	logging.Debug("Checking if 'nmcli' is available")
+	path, err := exec.LookPath("nmcli")
+	if err != nil {
+		return false, errors.New("NetworkManager cli nmcli was not found in path")
+	}
+	logging.Debug("'nmcli' was found in ", path)
+	return true, nil
+}
+func fixNetworkManagerInstalled() (bool, error) {
+	return false, fmt.Errorf("NetworkManager is required and must be installed manually")
+}
+func CheckNetworkManagerIsRunning() (bool, error) {
+	logging.Debug("Checking if NetworkManager.service is running")
+	path, err := exec.LookPath("systemctl")
+	if err != nil {
+		return false, err
+	}
+	stdOut, stdErr, err := crcos.RunWithDefaultLocale(path, "is-active", "NetworkManager")
+	if err != nil {
+		return false, fmt.Errorf("%v : %s", err, stdErr)
+	}
+	if strings.TrimSpace(stdOut) != "active" {
+		return false, errors.New("NetworkManager.service is not running")
+	}
+	logging.Debug("NetworkManager.service is already running")
+	return true, nil
+}
+func fixNetworkManagerIsRunning() (bool, error) {
+	return false, fmt.Errorf("NetworkManager is required. Please make sure it is installed and running manually")
+}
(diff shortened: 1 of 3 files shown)


anjannath:
retest this please

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -455,3 +455,34 @@ func fixCrcNetworkManagerConfig() (bool, error) {
...
+	logging.Debug("Checking if 'nmcli' is available")
+	path, err := exec.LookPath("nmcli")
+	if err != nil {
+		return false, errors.New("NetworkManager cli nmcli was not found in path")
+	}
+	logging.Debug("'nmcli' was found in ", path)
+	return true, nil
+}
+func fixNetworkManagerInstalled() (bool, error) {
+	return false, fmt.Errorf("NetworkManager is required and need to installed manually")
```
cfergeau: "and needs to be installed manually" ("or must be installed manually")

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -455,3 +455,34 @@ func fixCrcNetworkManagerConfig() (bool, error) {
...
+		return false, fmt.Errorf("%v : %s", err, stdErr)
+	}
+	if strings.TrimSpace(stdOut) != "active" {
+		return false, errors.New("NetworkManager.service is not running")
+	}
+	logging.Debug("NetworkManager.service is already running")
+	return true, nil
+}
+func fixNetworkManagerIsRunning() (bool, error) {
+	return false, fmt.Errorf("NetworkManager service should be running and need to run manually")
```
cfergeau: "and must be started manually"? Or maybe something like "NetworkManager is required. Please make sure it is installed and running"

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -455,3 +455,34 @@ func fixCrcNetworkManagerConfig() (bool, error) {
 	logging.Debug("NetworkManager configuration fixed")
 	return true, nil
 }
+func checkNetworkManagerInstalled() (bool, error) {
+	logging.Debug("Checking if 'nmcli' is available")
```
cfergeau: The 'IsRunning' check is enough, if it's not installed, it won't be running, so we don't really need a separate "IsInstalled" check.

praveenkumar: Some time NetworkManager is installed but service is not running, so I think having both should make sense.

cfergeau: Yes, we need to check if it's running. But since it's not going to be running if it's not installed, I would only keep the "running" check, and remove the "installed" check.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -455,3 +455,34 @@ func fixCrcNetworkManagerConfig() (bool, error) {
...
+		return false, fmt.Errorf("%v : %s", err, stdErr)
+	}
+	if strings.TrimSpace(stdOut) != "active" {
+		return false, errors.New("NetworkManager.service is not running")
+	}
+	logging.Debug("NetworkManager.service is already running")
+	return true, nil
+}
+func fixNetworkManagerIsRunning() (bool, error) {
+	return false, fmt.Errorf("NetworkManager is required. Please make sure it is installed and running manually")
```
cfergeau: the "manually" looks a bit out of place here.

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau but if it installed but not running, same as libvirt we have as test or you mean since this is manual so having only one test and tell user to do those manually?

cfergeau: Yes, in the libvirt case, we can fix things when libvirt is not installed/not running. In the networkmanager case, we don't fix anything, we only need to detect something is amiss. So a single check would in my opinion be plenty enough.

Issue #342, opened 2019-07-23
Remove the use of the DefaultHostname (invoke preflight like services)


gbraad:
The constant:

https://github.com/code-ready/crc/blob/ac25be8ee06d839d63257ec137adae485647e344/pkg/crc/constants/constants.go#L19

 is defined during the image creation and the image bundle metadata should have this. It is used in the preflight check

https://github.com/code-ready/crc/blob/f80d90838b28cce58052cc179566d90fc1e319f5/pkg/crc/preflight/preflight_checks_linux.go#L293-L322

But in that case we already know which bundle is going to be used... This is why I suggested to pass the whole metadata/machineconfig to the preflights, and threat them the same as services we start.

gbraad:
This can be handled as part of the preflight refactor (when moving these to be part of services on-start)

gbraad:
We got bitten by this on release ;-)

anjannath:
This check happens before the bundle is extracted, so we don't have access to the bundleInfo, if starting from a cleaned start/new bundle.

cfergeau:
Yes, at the moment they do, which means moving some code around will be needed to fix this ;) Maybe setup will need to take care of extracting the bundle, and then the preflight checks split in "independant checks" "os/distro specific checks" "bundle checks" and so on

anjannath:
I guess we'd want https://github.com/code-ready/snc/issues/33 as a pre-requisite for solving this, wherein setup can just extract the metadata file, and get the needed info.

But, we'd still need to supply the bundle to `setup` command, whereas `setup` is supposed to extract the bundle!

cfergeau:
This was removed by https://github.com/code-ready/crc/commit/035609075fc1e415e228e76dd1bfecb4da680d48

PR #465, opened 2019-08-26
Update hostname hash value from 'tt4fg' to '56mmj'

praveenkumar:
Wait till Centos CI pass before merge.


cfergeau:
Usual comment that this should be read from the bundle file rather than being hardcoded ;) apart from this, looks good.

gbraad:
https://github.com/code-ready/crc/issues/342

Issue #475, opened 2019-08-27
In case of older libvirt version driver version check doesn't provide proper detail.


praveenkumar:
If a user on older linux system and have libvirt < 3.4.0 then installation of driver succeed but the actual check failed without much details to the user. If a user manually try to check the version then it have more info.

```
# crc start -b .crc/crc_libvirt_4.1.9.crcbundle 
INFO Checking if oc binary is cached              
INFO Checking if Virtualization is enabled        
INFO Checking if KVM is enabled                   
INFO Checking if libvirt is installed             
INFO Checking if user is part of libvirt group    
INFO Checking if libvirt is enabled               
INFO Checking if libvirt daemon is running        
INFO Checking if crc-driver-libvirt is installed  
FATA exit status 1 

# crc-driver-libvirt -v
crc-driver-libvirt: /lib64/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by crc-driver-libvirt)
```

cfergeau:
On which system did it happen? RHEL 7.5 and newer have a libvirt more recent than 3.4.0. In addition to printing the error, we could catch this at crc setup time with a libvirt version check (though detecting which minimum version a given crc-driver-libvirt binary needs is a bit tricky/ugly).

gbraad:
This would in general not be an issue for a new user.
Does this need to included in a new beta version?

praveenkumar:
@gbraad no, until a new user is really using an old version of libvirt.

fixed by #476

gbraad:
I saw the merge... so we can postpone this and aim for after the beta, towards GA.

PR #476, opened 2019-08-27
Issue #475 Add stderr to returned err for libvirt driver version check

praveenkumar:
$ crc start -b .crc/crc_libvirt_4.1.9.crcbundle Issue #475
[...]
FATA exit status 1

$ crc-driver-libvirt -v
crc-driver-libvirt: /lib64/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by crc-driver-libvirt)

No explaination given to user if they are using libvirt < 3.4.0, why it happened,
since we are not adding stderr to the error message.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -228,7 +228,7 @@ func checkMachineDriverLibvirtInstalled() (bool, error) {
 	// Check the version of driver if it matches to supported one
 	stdOut, stdErr, err := crcos.RunWithDefaultLocale(libvirtDriverPath, "version")
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("%v : %s", err, stdErr)
 	}
 	if !strings.Contains(stdOut, libvirtDriverVersion) {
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -228,7 +228,7 @@ func checkMachineDriverLibvirtInstalled() (bool, error) {
 	// Check the version of driver if it matches to supported one
```
cfergeau: s/explaination/explanation/ in the commit log

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -228,7 +228,7 @@ func checkMachineDriverLibvirtInstalled() (bool, error) {
...
 	stdOut, stdErr, err := crcos.RunWithDefaultLocale(libvirtDriverPath, "version")
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("%v : %s", err, stdErr)
```
cfergeau: What will the new error look like? 
I guess something like `FATA exit status 1: /lib64/libvirt.so.0: version `LIBVIRT_3.4.0' not found`?
In the future (not in this PR), we can probably improve the error messages we show to the user, with a user readable error message ("libvirt machine driver is non-functional, maybe crc setup will fix it"), and a more detailed/technical error message, which maybe we will only show on log_level debug ("/lib64/libvirt.so.0: version `LIBVIRT_3.4.0' not found`")

PR #485, opened 2019-08-28
Move VM running check before getting bundle details.

praveenkumar:
We don't need to perform bundleName and crcBundleMetadata operation
if the VM is already running and user again use `crc start`.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -176,16 +176,6 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			result.Error = err.Error()
 			return *result, errors.Newf("Error loading bundle metadata: %v", err)
 		}
-		if bundleName != filepath.Base(startConfig.BundlePath) {
-			logging.Fatalf("Bundle '%s' was requested, but loaded VM is using '%s'",
-				filepath.Base(startConfig.BundlePath), bundleName)
-		}
-		if host.Driver.DriverName() != startConfig.VMDriver {
-			err := errors.Newf("VM driver '%s' was requested, but loaded VM is using '%s' instead",
-				startConfig.VMDriver, host.Driver.DriverName())
-			result.Error = err.Error()
-			return *result, err
-		}
 		vmState, err := host.Driver.GetState()
 		if err != nil {
 			result.Error = err.Error()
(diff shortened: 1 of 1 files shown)


cfergeau:
If we move this check, this means that when the VM is already running, we won't be checking the openshift version of the running VM against the openshift version expected by the current crc binary. In my opinion, we want to keep that version check even when the VM is running so that the user can be warned that they are not using the latest bundle version even if `crc start` ran successfully.

praveenkumar:
@cfergeau yes that is my point if a VM is already running then why we need to perform those checks, what if user is using the same `crc` binary so in that case is just not needed but if a user going to use a updated binary then also it just warn that bundle version is old and do nothing which actually make user bit confused for a running cluster. For a stopped VM, it make sense since it going to start a VM with older version of bundle.

cfergeau:
If this is an updated binary and if the user remembers that their old cluster is running, they are not going to run `crc start` because they know it is running already.
So if they run `crc start`, it means there is already some confusion going on ;) Add to that that at the moment, `crc start` is not very explicit whether it started a new instance, or whether it did nothing because the instance was already running. So one might run `crc start` with an updated binary because they want to use the new bundle, and end up running an old version without noticing.

After writing all of this, I think the right thing to do is to change the message when the VM is already running to say something like "A CodeReady Containers instance for OpenShift v%s is already running" (and similar logs when we created a new VM, or started an existing one) which would address what both of us want :)

praveenkumar:
> After writing all of this, I think the right thing to do is to change the message when the VM is already running to say something like "A CodeReady Containers instance for OpenShift v%s is already running" (and similar logs when we created a new VM, or started an existing one) which would address what both of us want :)

@cfergeau Since you already made changes in the message string, I updated the PR now so should be good enough?

pkg/crc/machine/machine.go
```diff
@@ -170,6 +170,22 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error loading host: %v", err)
 		}
 
+		vmState, err := host.Driver.GetState()
+		if err != nil {
+			result.Error = err.Error()
+			return *result, errors.Newf("Error getting the state for host: %v", err)
+		}
+		if vmState == state.Running {
+			openshiftVersion := crcBundleMetadata.GetOpenshiftVersion()
```
cfergeau: I see `bundleName, crcBundleMetadata, err = getBundleMetadataFromDriver(host.Driver)` below this block, is this working as expected? (I did not look very closely nor run it for a test)

praveenkumar: @cfergeau yes `crcBundleMetadata` is used as part of getting openshift version, so `IbundleName, crcBundleMetadata, err = getBundleMetadataFromDriver(host.Driver)` should be above that block.

praveenkumar:
> I would keep them as they are very explicit that the running VM may not be running what the user expects.

@cfergeau If you have very strong opinion about having those warning on a running cluster then I just go ahead and close this PR, since it will not make any other difference.

cfergeau:
> > I would keep them as they are very explicit that the running VM may not be running what the user expects.
> 
> @cfergeau If you have very strong opinion about having those warning on a running cluster then I just go ahead and close this PR, since it will not make any other difference.

Not "very strong opinion", it also depends on how strongly you feel about it :)

praveenkumar:
> Not "very strong opinion", it also depends on how strongly you feel about it :)

@cfergeau looks like we both doesn't have any strong opinion about it so I am closing this PR :)

PR #490, opened 2019-08-28
Issue #488 Add docs_serve target to makefile

### LINKED ISSUE #488: provide way to view docs changes locally
A makefile target `docs_serve`, so that docs changes can be verified locally

anjannath:
Fixes #488

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ RELEASE_DIR ?= release
 DOCS_BUILD_DIR ?= docs/build
 DOCS_BUILD_CONTAINER ?= registry.gitlab.com/gbraad/asciidoctor-centos:latest
 DOCS_PDF_CONTAINER ?= quay.io/crcont/asciidoctor-pdf:latest
+DOCS_SERVE_CONTAINER ?= quay.io/crcont/docs-server:latest
 DOCS_BUILD_TARGET ?= /docs/source/getting-started/master.adoc
 
 GOOS ?= $(shell go env GOOS)
@@ -102,6 +103,10 @@ build_docs:
 .PHONY: build_docs_pdf
 build_docs_pdf:
 	podman run -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_PDF_CONTAINER) -D /$(DOCS_BUILD_DIR) -o doc.pdf $(DOCS_BUILD_TARGET)
+
+.PHONY: docs_serve
+docs_serve: build_docs
+	podman run -v $(CURDIR)/docs:/docs:Z --rm -p 8088:8088/tcp $(DOCS_SERVE_CONTAINER)
 
 .PHONY: clean_docs
 clean_docs:


Makefile
```diff
@@ -103,6 +104,10 @@ build_docs:
 build_docs_pdf:
 	podman run -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_PDF_CONTAINER) -D /$(DOCS_BUILD_DIR) -o doc.pdf $(DOCS_BUILD_TARGET)
 
+.PHONY: docs_serve
+docs_serve: build_docs
+	podman run -v $(CURDIR)/docs:/docs:Z --rm -p 8088:8088/tcp $(DOCS_SERVE_CONTAINER)
+
```
cfergeau: If podman run was successful you can add a `@echo "You can now connect to http://localhost:8088 with your web browser to check the documentation"`

Makefile
```diff
@@ -26,6 +26,7 @@ RELEASE_DIR ?= release
 DOCS_BUILD_DIR ?= docs/build
```
cfergeau: So the goal is to start a webserver that one can connect to in order to check how the docs look like? It would be good to have this explained in the commit log, maybe even put the docker file that you used in the log.

Makefile
anjannath: So it already prints something like below, when it runs:
```
❯ make docs_serve 
podman run -v /home/anjan/github.com/code-ready/crc/docs:/docs:Z --rm registry.gitlab.com/gbraad/asciidoctor-centos:latest -b html5 -D /docs/build -o index.html /docs/source/getting-started/master.adoc
podman run -v /home/anjan/github.com/code-ready/crc/docs:/docs:Z --rm -p 8088:8088/tcp quay.io/crcont/docs-server:latest
Serving HTTP on 0.0.0.0 port 8088 (http://0.0.0.0:8088/) ...
``` 
Do you still think we need to add that message (localhost, 0.0.0.0, 127.0.0.1 all of these would work)

Makefile
anjannath: Okay, i'll add the explanation, but not the Dockerfile though, i added it in the comment, and ideally we should be putting it in the tree, maybe?

anjannath:
This is very slow, takes around a minute for the container to get up and start serving... :(

Makefile
cfergeau: ah, I did not notice there was a message, I agree another one is not needed. If you add a @ before the 'podman' command, I think it will hide the command while still printing the "Serving ..." message.

Makefile
cfergeau: Hmm not sure it's worth having it in the tree

cfergeau:
> This is very slow, takes around a minute for the container to get up and start serving... :(

Ah :( Initially, I wanted to ask why we were starting a container for that, but then I realized I don't have that 'http' python module installed by default on my machine ;) But if starting the container is too slow, maybe checking if the right module is available, and running this directly is a better solution?

anjannath:
> But if starting the container is too slow, maybe checking if the right module is available

I think this module is part of python std library @praveenkumar?

So the container is using python3, for python2 you need to use `python -m SimpleHTTPServer` both are equivalent.

praveenkumar:
@anjannath where is the dockerfile for this serving container? 

> I think this module is part of python std library @praveenkumar?

yes it should be part of python std library but having container help if I want to test the docs on Mac or Windows where default python installation not available.

cfergeau:
> > But if starting the container is too slow, maybe checking if the right module is available
> 
> I think this module is part of python std library @praveenkumar?
> 
> So the container is using python3, for python2 you need to use `python -m SimpleHTTPServer` both are equivalent.

Yup, this was my confusion, I tried `python -m http.server 8088` which is still python2 on my machine. `python3 -m htpp.server 8088` works as expected.


> yes it should be part of python std library but having container help if I want to test the docs on Mac or Windows where default python installation not available.

You can probably directly open the html files in your webbrowser for this kind of testing?

anjannath:
@praveenkumar The dockerfile look as mentioned in https://github.com/code-ready/crc/issues/488#issuecomment-525650590

anjannath:
> You can probably directly open the html files in your webbrowser for this kind of testing?

True, but having a command you can directly run and have the URL to copy paste is convenient, then going to browser, browse to the dir and opening the file :)

cfergeau:
> I'm personally fine with this change, feel free to send an updated version if the concerns which were raised need addressing.

(and feel free to merge this if you think it's good enough for now ;)

PR #491, opened 2019-08-28
Issue #482 Use DefaultBundlePath to check if needs unpacking

anjannath:
Use DefaultBundlePath to check if needs unpacking in crc setup command, on crc start bundle config is set use the configured value, otherwise use DefaultBundlePath.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -9,15 +9,13 @@ import (
 	"github.com/YourFin/binappend"
 	"github.com/kardianos/osext"
 
-	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
-	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/oc"
 )
 
 func checkBundleCached() (bool, error) {
-	if _, err := os.Stat(config.GetString(cmdConfig.Bundle.Name)); os.IsNotExist(err) {
+	if _, err := os.Stat(constants.DefaultBundlePath); os.IsNotExist(err) {
 		return false, err
 	}
 	return true, nil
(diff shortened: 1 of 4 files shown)


cmd/crc/cmd/start.go
```diff
@@ -154,3 +154,10 @@ func getPullSecretFileContent() (string, error) {
 
 	return pullsecret, nil
 }
+
+func getBundlePath() string {
+	if crcConfig.GetString(config.Bundle.Name) == "" {
+		return constants.DefaultBundlePath
+	}
+	return crcConfig.GetString(config.Bundle.Name)
+}
```
cfergeau: Is this going to give us a different behaviour than what we had before? Instead of specifying a default of DefaultBundlePath, you use "" as the default, but you also make sure that each time we use the config value, "" is replaced with DefaultBundlePath, so in the end it's the same as if we had set the default as DefautBundlePath?

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -9,15 +9,13 @@ import (
...
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/oc"
 )
 
 func checkBundleCached() (bool, error) {
-	if _, err := os.Stat(config.GetString(cmdConfig.Bundle.Name)); os.IsNotExist(err) {
+	if _, err := os.Stat(constants.DefaultBundlePath); os.IsNotExist(err) {
 		return false, err
 	}
```
cfergeau: ACK

cmd/crc/cmd/start.go
anjannath: Right, this change is not needed, just the other change is enough!

anjannath:
retest this please

cfergeau:
This is causing https://github.com/code-ready/crc/issues/506

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -9,15 +9,13 @@ import (
...
-	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
-	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/oc"
 )
 
 func checkBundleCached() (bool, error) {
-	if _, err := os.Stat(config.GetString(cmdConfig.Bundle.Name)); os.IsNotExist(err) {
+	if _, err := os.Stat(constants.DefaultBundlePath); os.IsNotExist(err) {
```
praveenkumar: :+1:

Issue #501, opened 2019-08-28
make integration segfault


cfergeau:
Running make integration on my machine segfaults:
```
    $ make integration
    panic: runtime error: index out of range
    
    goroutine 1 [running]:
    github.com/code-ready/crc/test/integration/crcsuite.ParseFlags()
            /home/teuf/go/src/github.com/code-ready/crc/test/integration/crcsuite/prepare.go:64 +0x154
    github.com/code-ready/crc/test/integration.parseFlags()
            /home/teuf/go/src/github.com/code-ready/crc/test/integration/integration_test.go:62 +0x25
    github.com/code-ready/crc/test/integration.TestMain(0xc0000b1f10)
            /home/teuf/go/src/github.com/code-ready/crc/test/integration/integration_test.go:33 +0x34
    main.main()
            _testmain.go:38 +0x10f
    FAIL    github.com/code-ready/crc/test/integration      0.015s
    make: *** [Makefile:120: integration] Error 1
```

This happens because I did not set BUNDLE_LOCATION and PULL_SECRET_PATH.

gbraad:
seems it does not check if the recieved arguments are populated. it should bailout when not provided.

PR #502, opened 2019-08-28
Issue #501: integration: Check we are getting enough arguments

cfergeau:
Running 'make integration' without setting BUNDLE_LOCATION and PULL_SECRET_FILE
results in a segfault because there are not enough arguments:

$ make integration
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/code-ready/crc/test/integration/crcsuite.ParseFlags()
        /home/teuf/go/src/github.com/code-ready/crc/test/integration/crcsuite/prepare.go:64 +0x154
github.com/code-ready/crc/test/integration.parseFlags()
        /home/teuf/go/src/github.com/code-ready/crc/test/integration/integration_test.go:62 +0x25
github.com/code-ready/crc/test/integration.TestMain(0xc0000b1f10)
        /home/teuf/go/src/github.com/code-ready/crc/test/integration/integration_test.go:33 +0x34
main.main()
        _testmain.go:38 +0x10f
FAIL    github.com/code-ready/crc/test/integration      0.015s
make: *** [Makefile:120: integration] Error 1

This commit adds an error message + os.Exit(1) so that it's easier to know what's going on.

This fixes https://github.com/code-ready/crc/issues/501

### DIFF
--- a/test/integration/crcsuite/prepare.go
+++ b/test/integration/crcsuite/prepare.go
@@ -61,6 +61,10 @@ func DownloadBundle(bundleLocation string, bundleDestination string) (string, er
 func ParseFlags() {
 
 	flag.Parse()
+	if flag.NArg() < 2 {
+		fmt.Printf("Invalid number of arguments, the paths to the bundle file and to the pull secret file are required\n")
+		os.Exit(1)
+	}
 	bundleURL = flag.Args()[0]
 	_, bundleName = filepath.Split(bundleURL)
 	pullSecretFile = flag.Args()[1]
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,8 @@ clean: clean_docs
 .PHONY: integration ## Run integration tests
 integration: GODOG_OPTS = --godog.tags=$(GOOS)
 integration:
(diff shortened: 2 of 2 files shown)


test/integration/crcsuite/prepare.go
```diff
@@ -61,6 +61,10 @@ func DownloadBundle(bundleLocation string, bundleDestination string) (string, er
 func ParseFlags() {
 
 	flag.Parse()
+	if flag.NArg() < 2 {
```
praveenkumar: :+1:

praveenkumar:
@cfergeau should we also do that check from Makefile side so if a user want to do `make integration` and the `BUNDLE_PATH` or `PULL_SECRET` not set then tell then and there ( this way we have fixed it both the side. I think we can use same logic which used for bundle embedded side.

cfergeau:
> @cfergeau should we also do that check from Makefile side so if a user want to do `make integration` and the `BUNDLE_PATH` or `PULL_SECRET` not set then tell then and there ( this way we have fixed it both the side. I think we can use same logic which used for bundle embedded side.

Good idea, I'll take a look..

Issue #145, opened 2019-05-21
Inject SSH keys during initial start


gbraad:
At the moment we are (re-)using a single set of SSH keys for the creation of the images, and they are also used for login/configuring during start of the VM.

We should however move to a way in which keys are assigned to an instance on initial start, eg. create the keys, create a disk image (or ISO), mount to the VM, and on initial start replace the SSH keys to be unique to this instance.

gbraad:
would we want to inject the users SSH key as a possibility? If so, we allow access to the VM ... and this is like the behaviour of the installer... just wonder, do we want this?

praveenkumar:
@gbraad what if a user doesn't have SSH key setup and also this will be problematic for windows, I want to go with what we had in minishift, attaching the userdata as a separate disk and consume it as a part of script which run with a unit file.

gbraad:
I would not go down this route as this is tremendously slow on Windows as the image will have to be written three times.

> We have removed that that logic during the refactor 8b1b441 and need to
> be added back since we want to send ssh keys as part of user data.

Inject can be done by using the generic key and then replace them on first connection with SSH with a new key. This key generation can happen within CRC and passed to the driver (but can also be completely contained within CRC). I do not believe the driver axctually has to know about SSH and user to connect with. The driver should ONLY have to deal with hypervisor related interaction.

gbraad:
a datadisk means there needs to be logic inside the VM to handle the extraction of a bitstream... it also needs to have this implementation of 'creation' happening in at least 4 different implementation (libvirt, hyperv, hyperkit and virtualbox), while a replacement from just CRC by connecting to the VM, replacing with a new key, can be shared among all implementations (without having to write support for a non-supported platfom like vbox)

cfergeau:
This one has been fixed by the commit listed in that issue.

PR #514, opened 2019-08-30
Issue #145 Generate ssh keypair inplace

praveenkumar:
Instead using the cached ssh key, it should be generated for every
run when a new instance created.


pkg/crc/machine/machine.go
```diff
@@ -128,6 +130,15 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error getting the state for host: %v", err)
 		}
 
+		if err := ssh.GenerateSSHKey(constants.GetPrivateKeyPath()); err != nil {
```
cfergeau: I'd move this to `updatePublicSSHKey` (and rename it to something different as the 'Public' part of the name would be odd), `Start` is too long as it is, so I'd prefer that we only call helpers from it.

cfergeau:
> Can this be split in
> 
>     * go module update, with the drivers.RunSSHCommandFromDriver API change
> 
>     * new code generating the ssh keypair in place, and updating the VM to make use of it
>       ?

Ah sorry, I did not think this through, updating the module requires changing all the callers, but we can't introduce constants.GetSSHKeyPath() because it's not constant, it's coming from the bundle. My goal in splitting was to reduce the overall diff, and to only get hte 'interesting' changes, but this won't work out, so a single commit as it was initially will be enough...

pkg/crc/machine/machine.go
```diff
@@ -575,3 +581,23 @@ func GetConsoleURL(consoleConfig ConsoleConfig) (ConsoleResult, error) {
...
+	if err != nil {
+		return err
+	}
+	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys", publicKey)
+	_, err = host.RunSSHCommand(cmd, "")
+	if err != nil {
+		return err
+	}
+
+	return nil
```
cfergeau: The end of the function can be `return err`

PR #516, opened 2019-08-30
Issue #497 Add preflight check to see if ran as root/admin

### LINKED ISSUE #497: Add preflight check for root/Administrator
Since crc must not be run as root/Administrator, we can add a check which will detect this and exit with an error when detected.

anjannath:
Fixes #497

### DIFF
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -1,13 +1,13 @@
 package preflight
 
 import (
-	//"errors"
 	"fmt"
 	"os"
 	"strconv"
 	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/os/windows/powershell"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -478,3 +478,15 @@ func CheckNetworkManagerIsRunning() (bool, error) {
 func fixNetworkManagerIsRunning() (bool, error) {
 	return false, fmt.Errorf("NetworkManager is required. Please make sure it is installed and running manually")
 }
+
+func checkIfRunningAsNormalUser() (bool, error) {
+	if os.Geteuid() != 0 {
```
cfergeau: This code is duplicated between linux and darwin. Since os.Geteuid() is also available on Windows (it returns -1), this helper function could go to the generic preflight_checks.go, and it would only be called from the linux and darwin files. On Windows, it would be built but not used.

anjannath: This means we have to name the function something different in case of windows, which actually does this check using the powershell package.

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -126,3 +126,15 @@ func checkIfHyperVVirtualSwitchExists() (bool, error) {
 func fixHyperVVirtualSwitch() (bool, error) {
 	return false, errors.New("Please override the default by adding an external virtual switch and set configuration")
 }
+
+func checkIfRunningAsNormalUserInWindows() (bool, error) {
```
praveenkumar: What about if a user not using `powershell` does this still work for `cmd` side?

@gbraad Can you please have a look to the logic side?

anjannath:
Yes, in the powershell package, the function opens a powershell instance to
execute the command.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -126,3 +126,15 @@ func checkIfHyperVVirtualSwitchExists() (bool, error) {
...
+	if !powershell.IsAdmin() {
+		return true, nil
+	}
+	logging.Debug("Ran as administrator")
+	return false, fmt.Errorf("crc should be ran as a normal user")
+}
+
+func fixRunAsNormalUserInWindows() (bool, error) {
```
praveenkumar: 👍

PR #519, opened 2019-08-30
Fix misc 'crc config' issues

cfergeau:
This PR fixes miscellaneous issues related to "crc config". The most important one is the first commit avoiding writing default values to the config file.
The others are either refactors to gather all configuration-related code in a single place, or cosmetic improvements to "crc config" output.


pkg/crc/config/viper_config.go
```diff
@@ -26,22 +26,28 @@ func GetBool(key string) bool {
...
 		return errors.Newf("Error reading in new config: %s : %v", constants.ConfigFile, err)
 	}
 	return nil
 }
 
+// Unset unsets a given config key
+func Unset(key string) error {
+	delete(ViperConfig, key)
+	/* This apparently preserves default values - unit test would probably be useful.. */
+	return syncViperState(globalViper)
```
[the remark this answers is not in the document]
cfergeau: This is what I want, and the comment is out of place now, I should remove it.
It was about doing globalViper.SetDefault() followed by globalViper.ReadConfig().
I believe the right thing is happening, but I did not check closely ;)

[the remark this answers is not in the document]
anjannath: Yes, its doing the right thing now :)

cmd/crc/cmd/config/config.go
```diff
@@ -61,10 +42,44 @@ to the options that you set when you run the 'crc start' command.
 Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields()
 }
 
+func isPreflightKey(key string) bool {
+	return strings.HasPrefix(key, "skip-") || strings.HasPrefix(key, "warn-")
+}
+
+func less(lhsKey, rhsKey string) bool {
```
praveenkumar: It really took me sometime to understand what this function do, better to put a comment on the function to make sure next time we don't spend much time to figure out what it actually do :)

cmd/crc/cmd/config/get.go
```diff
@@ -26,9 +27,9 @@ to the options that you set when you run the 'crc start' command.`,
...
 func runConfigGet(key string) {
-	if v, ok := config.ViperConfig[key]; ok {
-		output.Out(key, ":", v)
-		return
+	v, err := config.Get(key)
+	if err != nil {
+		errors.ExitWithMessage(1, err.Error())
 	}
-	output.Out("Config not found:", key)
+	output.Out(key, ":", v)
```
praveenkumar: 👍

cmd/crc/cmd/config/set.go
```diff
@@ -25,28 +25,12 @@ to the options that you set when you run the 'crc start' command.`,
 }
 
 func runConfigSet(key string, value interface{}) {
-	_, ok := SettingsList[key]
-	if !ok {
-		errors.ExitWithMessage(1, "Config property does not exist: %s", key)
+	err := config.Set(key, value)
+	if err != nil {
+		errors.ExitWithMessage(1, err.Error())
```
praveenkumar: That you could've done same `if err := config.Set(key, value); err !=nil` but that is not blocker.

pkg/crc/config/viper_config.go
```diff
@@ -81,29 +96,54 @@ func InitViper() error {
...
-// SetDefault sets the default for a config
-func SetDefault(key string, value interface{}) {
+// setDefault sets the default for a config
+func setDefault(key string, value interface{}) {
 	globalViper.SetDefault(key, value)
 }
 
+func SetDefaults() {
+	for _, setting := range allSettings {
+		setDefault(setting.Name, setting.defaultValue)
```
praveenkumar: 👍

cmd/crc/cmd/config/set.go
cfergeau: yes, codebase currently uses a mix of the 2. I'll change it so that it's at least consistent in the set.go file ;)

gbraad:
A new push followed after approval? @cfergeau what did change ?

cfergeau:
> A new push followed after approval? @cfergeau what did change ?

I addressed the 2 comments raised by Praveen, changed `err := ...; if err {` to `if err := ..., err != nil {`, and added a comment to `less`

cmd/crc/cmd/config/config.go
```diff
@@ -61,10 +42,48 @@ to the options that you set when you run the 'crc start' command.
...
 }
 
+func isPreflightKey(key string) bool {
+	return strings.HasPrefix(key, "skip-") || strings.HasPrefix(key, "warn-")
+}
+
+// less is used to sort the config keys. We want to sort first the regular keys, and
+// then the keys related to preflight starting with a skip- or warn- prefix. We want
+// these preflight keys to be grouped by pair: 'skip-bar', 'warn-bar', 'skip-foo', 'warn-foo'
+// would be sorted in that order.
```
praveenkumar: 👍

Issue #530, opened 2019-09-03
Add Docker files to the repo which we use as containers.


praveenkumar:
Right now we are using `dnsmasq`, ` asciidoctor-pdf` and `docs-server` container images from quay but we don't have the Dockerfile in place which let us know how we created those. We should have those files in this repo so if we want to regenerat/update the images we should able to do.

praveenkumar:
fixed by #531

PR #531, opened 2019-09-03
Issue #530 Add Docker files for different container images.

praveenkumar:


images/asciidoctor-pdf/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM registry.redhat.io/ubi8/ruby-25
+MAINTAINER CodeReady Containers <devtools-cdk@redhat.com>
+
+RUN gem install asciidoctor-pdf --pre
+
+VOLUME /docs
+WORKDIR /docs
+
```
anjannath: @praveenkumar i needed to add `USER root`, otherwise the file being created had weird permissions and ownership.

praveenkumar: Ack, I will add that.

gbraad:
not sure if we should use `./images` as the location for this... sounds a lot like graphical assets. any reason why this name is used?

images/dnsmasq/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM centos:7
```
gbraad: no UBI option?

praveenkumar: No, I checked but ubi doesn't have `dnsmasq` default :(

praveenkumar:
> not sure if we should use ./images as the location for this... sounds a lot like graphical assets. any reason why this name is used?

@gbraad most of the project on openshift side use `image` to put dockerfiles like installer => https://github.com/openshift/installer/tree/master/images

images/dnsmasq/Dockerfile
gbraad: so no base UBI that can have DNSmasq installed from the repos?

gbraad:
Check, so `images/` it is ...

images/dnsmasq/Dockerfile
praveenkumar: @gbraad no whatever UBI provide as default repo `dnsmasq` is not present only `bind` is there.

gbraad: @praveenkumar can we inquire someone about this? So for now this is OK to use, but I would think it is better over time to use a RHEL base (especially for downstream).

Issue #477, opened 2019-08-27
Warn user in advance about possible cert expiration


praveenkumar:
We should add a warn message just after the VM creation if the cert of the VM is going to expired soon (assume only 10 days left), Right now we have check which tell user if the cert is already expired but at that time user now have an unusable VM and his workload might be there, putting a fair warning before hand should be good from crc prospective.

gbraad:
Not sure how you see this check to happen. We have to check inside the VM to determine this... or you want to base this of the bundle creation time?

praveenkumar:
@gbraad The check we have inside VM to determine actually error out when the cert is expired we just need to extend this check to warn a user if cert is going to expire within 7-10 days.

gbraad:
I would also go so far to record this certificate time as a precaution inside a data file alongside the VM. This way we can report to the user without having to start the VM, as this process can take quite some time (as warning). It will never be a perfect solution, so the final message comes from the actual VM having started.

praveenkumar:
@gbraad what we can do is to use the metadata json info about `buildtime` we have.

```
"buildInfo": {
    "buildTime": "2019-08-23T18:08:27+00:00",
```

So if a user start for fresh then we use this to show the warning and if a user going to run from an stopped then we should use from the VM to show that warning.

gbraad:
This ties in with: #484 since expiration needs to allow for update.

cfergeau:
Fixed by the commits above.

PR #535, opened 2019-09-05
Issue #477 Add warn if cert going to expire in 7 days.

praveenkumar:
If a user start VM from a clean state then check the bundle build
time and warn user about cert expiry. If user start a stopped VM
then from the VM check the validation of cert and warn the user.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -22,21 +22,44 @@ func WaitForSsh(driver drivers.Driver) error {
 	return errors.RetryAfter(10, checkSshConnectivity, time.Second)
 }
 
-// CheckCertsValidity checks if the cluster certs have expired
-func CheckCertsValidity(driver drivers.Driver) error {
+// CheckCertsValidityUsingBundleBuildTime check if the cluster certs going to expire soon.
+func CheckCertsValidityUsingBundleBuildTime(buildTime time.Time) (bool, int) {
+	certExpiryDate := buildTime.AddDate(0, 1, 0)
+	// Warn user if the cert expiry going to happen starting of the 7 days
+	timeAfter7Days := time.Now().AddDate(0, 0, 7)
+	return timeAfter7Days.After(certExpiryDate), int(time.Until(certExpiryDate).Hours()) / 24
+}
+
+// CheckCertsValidity checks if the cluster certs have expired or going to expire in next 7 days
+func CheckCertsValidity(driver drivers.Driver) (bool, int, error) {
+	certExpiryDate, err := getcertExipryDateFromVM(driver)
+	if err != nil {
+		return false, 0, err
+	}
+	if time.Now().After(certExpiryDate) {
+		return false, 0, fmt.Errorf("Certs have expired, they were valid till: %s", certExpiryDate.Format(time.RFC822))
+	}
+
+	// Warn user if the cert expiry going to happen starting of the 7 days
+	timeAfter7Days := time.Now().AddDate(0, 0, 7)
+	if timeAfter7Days.After(certExpiryDate) {
+		return true, int(time.Until(certExpiryDate).Hours()) / 24, nil
+	}
+	return false, 0, nil
+}
+
+func getcertExipryDateFromVM(driver drivers.Driver) (time.Time, error) {
+	certExpiryDate := time.Time{}
 	certExpiryDateCmd := `date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-client-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds`
 	output, err := drivers.RunSSHCommandFromDriver(driver, certExpiryDateCmd)
 	if err != nil {
-		return err
+		return certExpiryDate, err
 	}
-	certExpiryDate, err := time.Parse(time.RFC3339, strings.TrimSpace(output))
+	certExpiryDate, err = time.Parse(time.RFC3339, strings.TrimSpace(output))
 	if err != nil {
-		return err
+		return certExpiryDate, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -22,21 +22,44 @@ func WaitForSsh(driver drivers.Driver) error {
...
 }
 
-// CheckCertsValidity checks if the cluster certs have expired
-func CheckCertsValidity(driver drivers.Driver) error {
+// CheckCertsValidityUsingBundleBuildTime check if the cluster certs going to expire soon.
+func CheckCertsValidityUsingBundleBuildTime(buildTime time.Time) bool {
+	certExpiryDate := buildTime.AddDate(0, 1, 0)
+	// Warn user if the cert expiry going to happen in 7 days
+	timeAfter7Days := time.Now().AddDate(0, 0, 7)
+	return timeAfter7Days.After(certExpiryDate)
```
anjannath: This will start to return false from the _starting_ of the 7 days.. maybe mention that in the comment.

pkg/crc/cluster/cluster.go
```diff
@@ -22,21 +22,44 @@ func WaitForSsh(driver drivers.Driver) error {
...
+	}
+	if time.Now().After(certExpiryDate) {
+		return false, fmt.Errorf("Certs have expired, they were valid till: %s", certExpiryDate.Format(time.RFC822))
+	}
+
+	// Warn user if the cert expiry going to happen in 7 days
+	timeAfter7Days := time.Now().AddDate(0, 0, 7)
+	if timeAfter7Days.After(certExpiryDate) {
+		return true, nil
+	}
```
anjannath: This will return true from the _starting_ of the 7 days..

pkg/crc/machine/machine.go
```diff
@@ -107,6 +107,16 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 		}
 
+		// Check if certificate is going to expire in next 7 days
+		buildTime, err := crcBundleMetadata.GetBundleBuildTime()
+		if err != nil {
+			result.Error = err.Error()
+			return *result, errors.Newf("Error getting bundle build time: %v", err)
+		}
+		if cluster.CheckCertsValidityUsingBundleBuildTime(buildTime) {
+			logging.Warn("Bundle certificates are going to expire in 7 days, better to use new release")
```
anjannath: Better to say soon, instead of 7 days, as this should start a countdown otherwise.. even on the 6th day it'll say will expire in 7 days.

pkg/crc/machine/machine.go
```diff
@@ -205,10 +215,14 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 	// Check the certs validity inside the vm
 	logging.Info("Verifying validity of the cluster certificates ...")
-	if err := cluster.CheckCertsValidity(host.Driver); err != nil {
+	expiringIn7Days, err := cluster.CheckCertsValidity(host.Driver)
+	if err != nil {
 		result.Error = err.Error()
 		return *result, errors.New(err.Error())
 	}
+	if expiringIn7Days {
+		logging.Warn("Bundle certificates are going to expire in 7 days, better to use new release")
```
anjannath: 7 days should be called _soon_ or something similar, for the same reason above.

pkg/crc/machine/machine.go
praveenkumar: @anjannath Good catch, I added the counter now which will tell within 7 days, how many days left for cert expiry.

pkg/crc/cluster/cluster.go
```diff
@@ -22,21 +22,44 @@ func WaitForSsh(driver drivers.Driver) error {
...
 }
 
-// CheckCertsValidity checks if the cluster certs have expired
-func CheckCertsValidity(driver drivers.Driver) error {
+// CheckCertsValidityUsingBundleBuildTime check if the cluster certs going to expire soon.
+func CheckCertsValidityUsingBundleBuildTime(buildTime time.Time) (bool, int) {
+	certExpiryDate := buildTime.AddDate(0, 1, 0)
+	// Warn user if the cert expiry going to happen starting of the 7 days
+	timeAfter7Days := time.Now().AddDate(0, 0, 20)
+	return timeAfter7Days.After(certExpiryDate), int(time.Until(certExpiryDate).Hours()) / 24
```
anjannath: :+1:

pkg/crc/cluster/cluster.go
```diff
@@ -22,21 +22,44 @@ func WaitForSsh(driver drivers.Driver) error {
 	return errors.RetryAfter(10, checkSshConnectivity, time.Second)
 }
 
-// CheckCertsValidity checks if the cluster certs have expired
-func CheckCertsValidity(driver drivers.Driver) error {
+// CheckCertsValidityUsingBundleBuildTime check if the cluster certs going to expire soon.
+func CheckCertsValidityUsingBundleBuildTime(buildTime time.Time) (bool, int) {
+	certExpiryDate := buildTime.AddDate(0, 1, 0)
+	// Warn user if the cert expiry going to happen starting of the 7 days
+	timeAfter7Days := time.Now().AddDate(0, 0, 20)
```
anjannath: @praveenkumar can you take a look at this again, I think you meant to add 7 days!

praveenkumar: @anjannath Changed, Good catch :+1:

praveenkumar:
retest this please.

praveenkumar:
retest this please.

pkg/crc/cluster/cluster.go
cfergeau: Fwiw, returning a time.Duration rather than a unitless int would probably be slightly more robust.

Issue #539, opened 2019-09-05
Add callback function to inform user if memory/cpu config set.


praveenkumar:
As of now we are not letting users know that if they set `memory` or `cpu` configs as part of config set they need to start a fresh instance to apply those configs.

Something we use to have in minishift side.

```
$  ./minishift config set memory 4096
No Minishift instance exists. New 'memory' setting will be applied on next 'minishift start'

$ ./minishift config set memory 4096
You currently have an existing Minishift instance. Changes to the 'memory' setting are only applied when a new Minishift instance is created.
To let the configuration changes take effect, you must delete the current instance with 'minishift delete' and then start a new one with 'minishift start'.
```

PR #541, opened 2019-09-05
Issue #224 print ok message after config set and unset commands

### LINKED ISSUE #224: Better info message(s) after crc config set
If we set the params via `config set`, the info message printed post operation could be improved

```
 # ./crc config set memory 20480

# ./crc config set cpus 4      
```

here `crc - Local OpenShift 4.x cluster` doesn't indicate anything, either remove it or may be something like `Successfully configured memory to X`.

anjannath:
Fixes #224


anjannath:
Closing, this will fixed with callbacks as part of #539

PR #542, opened 2019-09-05
Issue #539 Add callback function for configs

### LINKED ISSUE #224: Better info message(s) after crc config set

### LINKED ISSUE #539: Add callback function to inform user if memory/cpu config set.

praveenkumar:
Configs which required to provide a message to end user should
be considered using callback functions.

Fixes #539 
Fixes #224


pkg/crc/config/viper_config.go
```diff
@@ -173,35 +176,47 @@ func runValidations(validations []ValidationFnType, value interface{}) (bool, st
...
 	ok, expectedValue := runValidations(allSettings[key].validationFns, value)
 	if !ok {
-		return fmt.Errorf("Config value is invalid: %s, Expected: %s\n", value, expectedValue)
+		return "", fmt.Errorf("Config value is invalid: %s, Expected: %s\n", value, expectedValue)
 	}
 
 	set(key, value)
 
-	return nil
+	callbackMsg := runCallbacks(allSettings[key].callbackFns, key, value)
```
anjannath: :+1:

pkg/crc/config/viper_config.go
```diff
@@ -173,35 +176,47 @@ func runValidations(validations []ValidationFnType, value interface{}) (bool, st
...
+		return "", fmt.Errorf("Config property does not exist: %s", key)
 	}
 
 	if err := unset(key); err != nil {
-		return fmt.Errorf("Error unsetting config property: %s : %v", key, err)
+		return "", fmt.Errorf("Error unsetting config property: %s : %v", key, err)
 	}
 
-	return nil
+	return fmt.Sprintf("Successfully removed config for: %s", key), nil
```
anjannath: :+1:

anjannath:
@praveenkumar Should refer to #539 in the commit message.

praveenkumar:
> Should refer to #539 in the commit message.

Ah yes, will do that.

Issue #399, opened 2019-08-09
Make integration tests working on Windows


tsedmik:
If I run our Integration tests on Windows, they fail. It looks like some test steps are missing (not implemented for Windows) - for example `crc.exe start`

This is the output of integration tests run:
```
make: Entering directory '/cygdrive/c/Jenkins/workspace/crc_pipeline/crc'

powershell.exe : go: downloading github.com/DATA-DOG/godog v0.7.13

At C:\Jenkins\workspace\crc_pipeline@tmp\durable-d39b8f6c\powershellWrapper.ps1:3 char:1

+ & powershell -NoProfile -NonInteractive -ExecutionPolicy Bypass -Comm ...

+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

    + CategoryInfo          : NotSpecified: (go: downloading...G/godog v0.7.13:String) [], RemoteException

    + FullyQualifiedErrorId : NativeCommandError

 

go: downloading github.com/code-ready/clicumber v0.0.0-20190503113956-2563aed4ef12


Log successfully started, logging into: C:\Jenkins\workspace\crc_pipeline\crc\test\integration\out\test-results/integration_2019-8-9_08-49-56.log

Running integration test in: C:\Jenkins\workspace\crc_pipeline\crc\test\integration\out\test-run

Working directory set to: C:\Jenkins\workspace\crc_pipeline\crc\test\integration\out\test-run

Deleted CRC home folder C:\Users\CDKQE\.crc.

The powershell instance has been started and will be used for testing.

Obtaining bundle...

Copying bundle from bundle.crcbundle to C:\Jenkins\workspace\crc_pipeline\crc\test\integration\out\test-run.

Using bundle: 

Feature: Basic test

  Checks whether CRC top-level commands behave correctly.


  Scenario: CRC version                   # features\basic.feature:5

    When executing "crc version" succeeds # shell.go:234 -> github.com/code-ready/clicumber/testsuite.ExecuteCommandSucceedsOrFails

    Then stderr should be empty           # shell.go:300 -> github.com/code-ready/clicumber/testsuite.CommandReturnShouldBeEmpty

    And stdout should contain "version:"  # shell.go:284 -> github.com/code-ready/clicumber/testsuite.CommandReturnShouldContain


  Scenario: CRC help                                # features\basic.feature:11

[...]

PR #543, opened 2019-09-05
Issue #399 - Enable integration tests for Windows hosts

tsedmik:
Signed-off-by: Tomáš Sedmík <tsedmik@redhat.com>

### DIFF
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -51,6 +51,8 @@ func FeatureContext(s *godog.Suite) {
 		StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails)
 	s.Step(`^starting CRC with default bundle and hypervisor "(.*)" (succeeds|fails)$`,
 		StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails)
+	s.Step(`^starting CRC with default bundle and nameserver "(.*)" (succeeds|fails)$`,
+		StartCRCWithDefaultBundleAndNameServerSucceedsOrFails)
 	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds|fails)$`,
 		SetConfigPropertyToValueSucceedsOrFails)
 	s.Step(`^login to the oc cluster (succeeds|fails)$`,
@@ -303,6 +305,16 @@ func StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails(hypervisor string, ex
 		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
 	}
 	cmd = fmt.Sprintf("crc start -d %s -p '%s' %s --log-level debug", hypervisor, pullSecretFile, extraBundleArgs)
+	err := clicumber.ExecuteCommandSucceedsOrFails(cmd, expected)
+
+	return err
+}
+
+func StartCRCWithDefaultBundleAndNameServerSucceedsOrFails(nameserver string, expected string) error {
+
+	var cmd string
+
+	cmd = fmt.Sprintf("crc start -n %s -p '%s' --log-level debug", nameserver, pullSecretFile)
 	err := clicumber.ExecuteCommandSucceedsOrFails(cmd, expected)
 
 	return err
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -72,8 +72,18 @@ Feature: Basic test
         And stdout should contain "Will use root access: set suid"
         And stdout should contain "Setting file permissions"
 
-    @linux
-    Scenario: CRC start on Linux
+    @windows
+    Scenario: CRC setup on Windows
+        When executing "crc setup" succeeds
+        Then stdout should contain "Caching oc binary"
+        Then stdout should contain "Unpacking bundle from the CRC binary"
+        Then stdout should contain "Check Windows 10 release"
+        Then stdout should contain "Hyper-V installed"
+        Then stdout should contain "Is user a member of the Hyper-V Administrators group"
+        Then stdout should contain "Does the Hyper-V virtual switch exist"
+
+    @linux @windows
+    Scenario: CRC start
         When starting CRC with default bundle and default hypervisor succeeds
         Then stdout should contain "The OpenShift cluster is running"
 
--- a/test/integration/features/config.feature
+++ b/test/integration/features/config.feature
@@ -25,7 +25,14 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
             | memory    |    8192 |     4096 |
             | vm-driver | libvirt | hyperkit |
 
-    @darwin @linux
+        @windows
+        Examples: Config settings on Windows
+            | property  | value1                   | value2                                               |
+            | cpus      | 4                        | 3                                                    |
+            | memory    | 8192                     | 4096                                                 |
+            | vm-driver | hyperv                   | hyperkit                                             |
+
+    @linux @darwin @windows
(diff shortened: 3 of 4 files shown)


test/integration/features/basic.feature
```diff
@@ -1,13 +1,13 @@
 Feature: Basic test
     Checks whether CRC top-level commands behave correctly.
 
-    @darwin @linux @windows
```
praveenkumar: why change the order, it doesn't make sense and also now this change part of commit message.

test/integration/features/story_health.feature
```diff
@@ -53,6 +62,20 @@ Feature:
         And executing "oc expose svc/httpd-ex" succeeds
         And with up to "2" retries with wait period of "60s" command "curl -kI http://httpd-ex-testproj.apps-crc.testing" output should contain "HTTP/1.1 200 OK"
 
+    @windows
```
praveenkumar: How come this is different than above scenario, should be clubbed together.

test/integration/features/basic.feature
tsedmik: It was introduce by mistake. I'll change it back

tsedmik:
@praveenkumar PR is ready for another review ;-)

tsedmik:
retest this please

jsliacan:
retest this please

cfergeau:
retest this now

test/integration/features/story_health.feature
praveenkumar: @tsedmik I thought I put the review commit for it but looks like it was not published. What I want is if we are only checking the status of a URI using the curl then we should replace it with a function which can use `http` module https://golang.org/pkg/net/http/ and provide the response back.

jsliacan: @praveenkumar I'll submit this as a separate PR (replace all `curl` calls with a function call that uses Go's `http` module). Thanks for the tip.

jsliacan: Just a follow-up, the change away from `curl` is ready here: #577.

praveenkumar: @jsliacan @tsedmik since #577 already merged now can you update this PR ?

tsedmik:
retest this please

tsedmik:
retest this please

test/integration/features/config.feature
```diff
@@ -58,6 +65,12 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
             | warn-check-crc-network           | true   | false  |
             | warn-check-crc-network-active    | true   | false  |
 
+        @windows
+        Examples: Config warnings on Windows
+            | property                             | value1 | value2 |
+            | warn-check-virtualbox-installed      | true   | false  |
```
cfergeau: This was changed in ` preflight: virtualbox config flags are darwin-only`

test/integration/features/story_health.feature
```diff
@@ -21,7 +21,16 @@ Feature:
...
             | libvirt   |
 
-    @darwin @linux 
+    @windows
+    Scenario: Start CRC on Windows
+        Given executing "crc setup" succeeds
+        When starting CRC with default bundle and default hypervisor succeeds
+        Then stdout should contain "CodeReady Containers instance is running"
+        And executing "crc oc-env | Invoke-Expression" succeeds
+        When with up to "4" retries with wait period of "2m" command "crc status" output should contain "Running (v4.x)"
```
cfergeau: We'll no longer be getting `(4.x)`

cfergeau:
Can you rebase so that CI reruns? I'm surprised some of it was successful. Are we running the integration tests on windows/macos as part of CI yet?

cfergeau:
This still need rebasing.

tsedmik:
Hopefully, ready for merge ;-)

test/integration/features/story_health.feature
```diff
@@ -54,6 +63,7 @@ Feature:
         And executing "oc expose svc/httpd-ex" succeeds
         And with up to "2" retries with wait period of "60s" http response from "http://httpd-ex-testproj.apps-crc.testing" should have status code "200"
 
+    @linux @darwin @windows
```
cfergeau: Do we need to be explicit about all the platforms here? Or would not specifying anything mean that this should always be used?

test/integration/features/story_health.feature
```diff
@@ -22,7 +22,16 @@ Feature:
             | vm-driver |
             | libvirt   |
 
-    @darwin @linux 
+    @windows
+    Scenario: Start CRC on Windows
+        Given executing "crc setup" succeeds
+        When starting CRC with default bundle and nameserver "10.75.5.25" succeeds
```
cfergeau: This nameserver looks like something which is going to be very specific to our CI?

tsedmik: yes, you're right. We discussed it on Slack. Without it we're unable to see outside network.

test/integration/features/story_health.feature
tsedmik: I'll check that. If it is the case, we can remove it as a part of another PR.

jsliacan: @cfergeau we need all tags -- unless I'm missing something fundamental. Sadly, if you specify that you want the tagged items to run, only those will run. So in particular, if you're running a `linux` tag and some part is not tagged as `linux`, it won't run. It clutters the feature files, but using "nonlinux", "nonwindows" and "nondarwin" tags (then negating them at the call when you want linux, windows, or darwin) does not improve situation much and it's confusing. So I left it as it is.

cfergeau: Thanks for checking :)

test/integration/crcsuite/crcsuite.go
```diff
@@ -308,6 +310,20 @@ func StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails(hypervisor string, ex
 	return err
 }
 
+func StartCRCWithDefaultBundleAndNameServerSucceedsOrFails(nameserver string, expected string) error {
+
+	var cmd string
+	var extraBundleArgs string
+
+	if !constants.BundleEmbedded() {
```
praveenkumar: @tsedmik this is not going to work and I think that is what @jsliacan is working on.

jsliacan: @praveenkumar - yes, it won't work with embedded bundle. But this is how the rest of the code on the master is right now. I suggest this can go in as it won't make things worse. And a comprehensive solution is in #724. WDYT?

test/integration/crcsuite/crcsuite.go
```diff
@@ -308,6 +310,20 @@ func StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails(hypervisor string, ex
...
+		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
+	}
```
praveenkumar: @tsedmik @jsliacan I would prefer to remove this logic in this PR then having it. Later once @jsliacan PR is in we should update it with that logic. Sounds good?

test/integration/crcsuite/crcsuite.go
```diff
@@ -308,6 +310,16 @@ func StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails(hypervisor string, ex
 	return err
 }
 
+func StartCRCWithDefaultBundleAndNameServerSucceedsOrFails(nameserver string, expected string) error {
+
+	var cmd string
+
+	cmd = fmt.Sprintf("crc start -n %s -p '%s' --log-level debug", nameserver, pullSecretFile)
```
praveenkumar: 👍

Issue #544, opened 2019-09-05
Make sure integration tests run with embedded bundle as well as a free bundle


jsliacan:
Introduce a way to recognize if the bundle is embedded in the binary or not. This affects how `crc start` is used throughout the tests.

jsliacan:
Closing this as it was addressed by #545. Tests still don't work with embedded bundle, but there is a new issue for it: #673.

PR #545, opened 2019-09-05
Issue #544 Recognize if bundle is embedded or not in integration tests

jsliacan:
I couldn't find any metadata of the `crc` binary that would identify an embedded bundle (except for file size). So, based on the output of `crc setup`, a variable `bundleEmbedded` is set to `true` or `false` and CRC is always started with `-b` flag if `false` and without it if `true`.

### DIFF
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	clicumber "github.com/code-ready/clicumber/testsuite"
+	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/oc"
 )
 
@@ -223,7 +224,13 @@ func LoginToOcClusterSucceedsOrFails(expected string) error {
 
 func StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails(expected string) error {
 
-	cmd := fmt.Sprintf("crc start -b %s -p '%s' --log-level debug", bundleName, pullSecretFile)
+	var cmd string
+	var extraBundleArgs string
+
+	if !constants.BundleEmbedded() {
+		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
+	}
+	cmd = fmt.Sprintf("crc start -p '%s' %s --log-level debug", pullSecretFile, extraBundleArgs)
 	err := clicumber.ExecuteCommandSucceedsOrFails(cmd, expected)
 
 	return err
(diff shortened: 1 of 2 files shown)


anjannath:
@jsliacan There's the `constant.BundleEmbedded()` function for this check, could be useful.
https://github.com/code-ready/crc/blob/d61e39cffd1f741b7afd4d98d602431ed60dcb79/pkg/crc/constants/constants.go#L72

jsliacan:
@anjannath thanks, it turned out to be much simpler that way.

test/integration/crcsuite/crcsuite.go
```diff
@@ -223,15 +224,28 @@ func LoginToOcClusterSucceedsOrFails(expected string) error {
...
 func StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails(expected string) error {
 
-	cmd := fmt.Sprintf("crc start -b %s -p '%s' --log-level debug", bundleName, pullSecretFile)
+	var cmd string
+
+	if constants.BundleEmbedded() {
+		cmd = fmt.Sprintf("crc start -p '%s' --log-level debug", pullSecretFile)
+	} else {
+		cmd = fmt.Sprintf("crc start -b %s -p '%s' --log-level debug", bundleName, pullSecretFile)
+	}
```
cfergeau: Maybe be slightly simpler as:
```
var extraBundleArgs string
if !constants.BundleEmbedded() {
    cmd =  fmt.Sprintf(" -b %s", bundleName
}
cmd =  fmt.Sprintf("crc start -p '%s' --log-level debug %s", pullSecretFile, extraBundleArgs)

File: test/integration/features/basic.feature [hunk 2]
```diff
@@ -60,7 +60,7 @@ Feature: Basic test
     
     @darwin @linux @windows
     Scenario: CRC status check
-        When with up to "10" retries with wait period of "1m" command "crc status" output should not contain "Stopped"
+        When with up to "15" retries with wait period of "1m" command "crc status" output should not contain "Stopped"
```
cfergeau: This change seems unrelated?

jsliacan: Yes, completely unrelated - I just noticed that the PR was not getting through because this step fails. We discussed it on Slack today - some operators take time to come up. They are needed in the next steps, so longer wait seems reasonable. It didn't seem big enough change to warrant a separate PR - was I wrong? Or should I mention it in the commit?

cfergeau: I'd just split it, but mentioning it in the commit log would be good enough so that people know this was not an oversight

test/integration/crcsuite/crcsuite.go
jsliacan: @cfergeau I changed it as you suggested - not sure why it doesn't say "outdated" above the code snippet...

test/integration/features/basic.feature
jsliacan: OK, it's in the log now - next time I'll opt for a separate PR! Thanks for reviewing :)

test/integration/crcsuite/crcsuite.go
```diff
@@ -223,15 +224,27 @@ func LoginToOcClusterSucceedsOrFails(expected string) error {
...
 func StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails(expected string) error {
 
-	cmd := fmt.Sprintf("crc start -b %s -p '%s' --log-level debug", bundleName, pullSecretFile)
+	var cmd string
+	var extraBundleArgs string
+
+	if !constants.BundleEmbedded() {
+		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
+	}
+	cmd = fmt.Sprintf("crc start -p '%s' %s --log-level debug", pullSecretFile, extraBundleArgs)
```
praveenkumar: :+1:

PR #564, opened 2019-09-10
Inform about needed relogin/reboot

### LINKED ISSUE #560: [BUG] Windows Setup: Virtual Switch not Found
### General information

  * OS: Windows 10 1903
  * Hypervisor: Hyper-V

## CRC version
`crc-windows-1.0.0-beta.3-amd64`

### Steps to reproduce

  1. Download CRC Package
  2. Add crc Binary to PATH
  3. run `crc setup`

### Expected
Installation should proceed as default switch exists

### Actual
Installation fails with error: Virtual Switch not Found

### Logs

```
\> crc setup --log-level debug
INFO Caching oc binary
DEBU oc binary already cached
INFO Check Windows 10 release
INFO Hyper-V installed
INFO Is user a member of the Hyper-V Administrators group
DEBU Always add user
INFO Does the Hyper-V virtual switch exist
ERRO Error occurred: Virtual Switch not found
DEBU Virtual Switch not found
ERRO Error occurred: Please override the default by adding an external virtual switch and set configuration
FATA Please override the default by adding an external virtual switch and set configuration
```

gbraad:
Fixes #560

Inform user about needed reboot/relogin when administrative commands fail.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -1,6 +1,7 @@
 package preflight
 
 import (
+	//"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -30,7 +31,7 @@ func checkVersionOfWindowsUpdate() (bool, error) {
 	}
 
 	if yourWindowsReleaseId < minimumWindowsReleaseId {
-		return false, errors.Newf("Please update Windows. Currently %d is the minimum release needed to run. You are running %d", minimumWindowsReleaseId, yourWindowsReleaseId)
+		return false, errors.New(fmt.Sprintf("Please update Windows. Currently %d is the minimum release needed to run. You are running %d", minimumWindowsReleaseId, yourWindowsReleaseId))
 	}
 	return true, nil
 }
@@ -116,7 +117,10 @@ func checkIfHyperVVirtualSwitchExists() (bool, error) {
 		if !strings.Contains(stdErr, "Get-VMSwitch") {
 			// found the default
 			return true, nil
+		} else {
+			return false, errors.New("Incorrect permissions")
 		}
+
 	}
 
 	return false, errors.New("Virtual Switch not found")
@@ -124,7 +128,7 @@ func checkIfHyperVVirtualSwitchExists() (bool, error) {
 
 // Unable to do for now
 func fixHyperVVirtualSwitch() (bool, error) {
-	return false, errors.New("Please override the default by adding an external virtual switch and set configuration")
+	return false, errors.New("Unable to perform Hyper-V administrative commands. Please make sure to re-login or reboot your system.")
 }
 
 func checkIfRunningAsNormalUserInWindows() (bool, error) {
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -80,7 +80,7 @@ func checkIfUserPartOfHyperVAdmins() (bool, error) {
 	@([Security.Principal.WindowsPrincipal][Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole($sid)`
 	stdOut, _, _ := powershell.Execute(checkIfMemberOfHyperVAdmins)
 	if !strings.Contains(stdOut, "True") {
-		return false, errors.New("User is not a member of the Hyper-V administrators group")
+		return false, errors.New("User is not a member of the Hyper-V administrators group.")
```
cfergeau: Not sure we want '.' at the end of messages (this was pointed out to me as part of another review).

gbraad: Fixed. It is a Go convention

gbraad:
retest this please

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -30,7 +30,7 @@ func checkVersionOfWindowsUpdate() (bool, error) {
 	}
 
 	if yourWindowsReleaseId < minimumWindowsReleaseId {
-		return false, errors.Newf("Please update Windows. Currently %d is the minimum release needed to run. You are running %d", minimumWindowsReleaseId, yourWindowsReleaseId)
+		return false, errors.New(fmt.Sprintf("Please update Windows. Currently %d is the minimum release needed to run. You are running %d", minimumWindowsReleaseId, yourWindowsReleaseId))
```
praveenkumar: @gbraad Since preflight checks are directly run from the `start.go` so the errors are wrapped with `errors.Newf` which is the part of `github.com/code-ready/crc/pkg/crc/errors` package but you now using standard `errors` package which not going to print out in stdout when the issue occur.

gbraad: I do not prefer to see:

```
FAIL: Error occured: Please update Windows. Currently 1703 is the minimum release needed to run. You are running ..
FAIL Please update Windows. Currently 1703 is the minimum release needed to run. You are running .."
```

which duplicates messages.

gbraad: The output with `crc/errors` is as follows:
image

Perhaps the last message is NEVER to be printed?

cfergeau: Regarding use of errors.Newf, my understanding was that only Start() would call it, that methods called from Start() should return 'regular' errors instead.

gbraad: With using `errors` it is:
image

gbraad: Regression perhaps?

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -1,6 +1,7 @@
 package preflight
 
 import (
+	//"errors"
```
praveenkumar: @gbraad shouldn't just we remove it instead comment?

praveenkumar:
Merged to master after changes.

Issue #566, opened 2019-09-10
{skip,warn}-check-resolv-conf-file-permissions is badly named


cfergeau:
Since commit c426c9202, we are no longer using /etc/resolv.conf, but we use
/etc/hosts instead. However, the corresponding preflight configuration keys
were not renamed, which makes them misleading.

PR #567, opened 2019-09-10
Issue #566: Rename {skip,warn}-check-resolv-conf-file-permissions

cfergeau:
Since commit c426c9202, we are no longer using /etc/resolv.conf, but we
use /etc/hosts instead. However, the corresponding preflight configuration
keys were not renamed, which makes them misleading. This commit changes
"resolv-conf" to "hosts" in their naming.

This resolves https://github.com/code-ready/crc/issues/566

### DIFF
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -236,11 +236,11 @@ func fixResolverFilePermissions() (bool, error) {
 	return addFileWritePermissionToUser(resolverFile)
 }
 
-func checkResolvConfFilePermissions() (bool, error) {
+func checkHostsFilePermissions() (bool, error) {
 	return isUserHaveFileWritePermission(hostFile)
 }
 
-func fixResolvConfFilePermissions() (bool, error) {
+func fixHostsFilePermissions() (bool, error) {
 	return addFileWritePermissionToUser(hostFile)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -236,11 +236,11 @@ func fixResolverFilePermissions() (bool, error) {
 	return addFileWritePermissionToUser(resolverFile)
 }
 
-func checkResolvConfFilePermissions() (bool, error) {
+func checkHostsFilePermissions() (bool, error) {
```
praveenkumar: 👍

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -97,11 +97,11 @@ func SetupHost(vmDriver string) {
 		config.GetBool(cmdConfig.WarnCheckResolverFilePermissions.Name),
 	)
 
-	preflightCheckAndFix(config.GetBool(cmdConfig.SkipCheckResolvConfFilePermissions.Name),
-		checkResolvConfFilePermissions,
-		fixResolvConfFilePermissions,
+	preflightCheckAndFix(config.GetBool(cmdConfig.SkipCheckHostsFilePermissions.Name),
+		checkHostsFilePermissions,
+		fixHostsFilePermissions,
```
praveenkumar: 👍

Issue #586, opened 2019-09-13
Fix win32 'make lint' issues


cfergeau:
Running "GOOS=windows make lint" reports a few issues at the moment.

gbraad:
shouldn't `make` run `lint` for the current `GOOS`, while `make cross` should also run this for any `GOOS` like it does for building the binaries? It seems lint is affected by `_windows`, `_linux` and `_darwin` in file naming.

So, how aout lint issues for macOS?

PR #587, opened 2019-09-13
Fix make lint issues on Windows

cfergeau:
This is not fixing everything as the last remaining issue was slightly trickier, so we could keep https://github.com/code-ready/crc/issues/586 open a while longer.


praveenkumar:
@cfergeau Can you also look below ones?

```
 $ GOOS=windows make lint
/home/prkumar/work/go/bin/golangci-lint run
pkg/crc/preflight/preflight_checks_common.go:73:6: `checkIfRunningAsNormalUser` is unused (deadcode)
func checkIfRunningAsNormalUser() (bool, error) {
     ^
pkg/crc/preflight/preflight_checks_common.go:81:6: `fixRunAsNormalUser` is unused (deadcode)
func fixRunAsNormalUser() (bool, error) {
     ^
pkg/os/windows/win32/shell32_windows.go:23:32: SA1019: syscall.StringToUTF16Ptr is deprecated: Use UTF16PtrFromString instead.  (staticcheck)
	return uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(input)))
	                              ^
pkg/os/windows/win32/shell32_windows.go:44:9: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (gosimple)
	return errors.New(fmt.Sprintf("win32 error %v", ret))
	       ^
make: *** [Makefile:143: lint] Error 1
```

cfergeau:
> @cfergeau Can you also look below ones?

I've fixed these as well now, even if they were initially intentionally ignored ;)

cfergeau:
retest now please

cfergeau:
retest this please

cfergeau:
retest this please

gbraad:
shouldn't `make` run `lint` for the current `GOOS`, while `make cross` should also run this for any `GOOS` like it does for building the binaries? It seems lint is affected by `_windows`, `_linux` and `_darwin` in file naming.

cfergeau:
retest this please

cfergeau:
> shouldn't `make` run `lint` for the current `GOOS`, while `make cross` should also run this for any `GOOS` like it does for building the binaries? It seems lint is affected by `_windows`, `_linux` and `_darwin` in file naming.

on linux, `make lint` and `GOOS=windows make lint` is giving me different results, so I think it's already doing this? And the new cross-lint target is indeed setting GOOS to the 3 values we are interested in, and running the equivalent of `make lint` with GOOS set. I have the feeling I'm missing your point?

.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         command: make fmtcheck
     - run:
         name: Run golangci-lint
-        command: make lint
+        command: make cross-lint
```
praveenkumar: :+1:

pkg/crc/preflight/preflight_checks_nonwin.go
```diff
@@ -0,0 +1,22 @@
...
+	if os.Geteuid() != 0 {
+		return true, nil
+	}
+	logging.Debug("Ran as root")
+	return false, fmt.Errorf("crc should be ran as a normal user")
+}
+
+func fixRunAsNormalUser() (bool, error) {
+	return false, fmt.Errorf("crc should be ran as a normal user")
+}
```
praveenkumar: :+1:

pkg/os/windows/win32/shell32_windows.go
```diff
@@ -1,45 +1,40 @@
...
 
-func toUintptr(input string) uintptr {
-	return uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(input)))
+func toUint16ptr(input string) *uint16 {
+	uint16ptr, err := syscall.UTF16PtrFromString(input)
+	if err != nil {
+		logging.Warnf("Failed to convert %s to UTF16: %v", input, err)
+	}
+
+	return uint16ptr
```
praveenkumar: :+1:

Issue #588, opened 2019-09-13
[BUG] version have bundle info (which is part of crc version command)


praveenkumar:
Currently, crc version give below output

```
$ ./crc version
version: 1.0.0-beta.2-4.1.11+a0732c0
OpenShift version: 4.1.11 (not embedded in binary)
```

Here `OpenShift version` already specified in a new line so we don't need to put same again to `version` side. Should be look like

```
$ ./crc version
version: 1.0.0-beta.2+a0732c0
OpenShift version: 4.1.11 (not embedded in binary)
```

anjannath:
Fixed via #589

PR #589, opened 2019-09-13
Issue #588 Remove bundle version from version section.

praveenkumar:
Without this `version` output look following, which doesn't make sense
since now Openshift version is part of version command and contain
this detail.

```
$ ./crc version
version: 1.0.0-beta.2-4.1.11+a0732c0
OpenShift version: 4.1.11 (not embedded in binary)
```

With this change, now it looks bit sane.

```
$ ./crc version
version: 1.0.0-beta.2+a0732c0
OpenShift version: 4.1.11 (not embedded in binary)
```

### DIFF
--- a/cmd/crc/cmd/version.go
+++ b/cmd/crc/cmd/version.go
@@ -38,7 +38,7 @@ var versionCmd = &cobra.Command{
 }
 
 func runPrintVersion(arguments []string) {
-	fmt.Printf("version: %s-%s+%s\n", version.GetCRCVersion(), version.GetBundleVersion(), version.GetCommitSha())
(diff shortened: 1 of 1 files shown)


cmd/crc/cmd/version.go
```diff
@@ -38,7 +38,7 @@ var versionCmd = &cobra.Command{
 }
 
 func runPrintVersion(arguments []string) {
-	fmt.Printf("version: %s-%s+%s\n", version.GetCRCVersion(), version.GetBundleVersion(), version.GetCommitSha())
+	fmt.Printf("version: %s+%s\n", version.GetCRCVersion(), version.GetCommitSha())
```
anjannath: :+1:

PR #590, opened 2019-09-13
WIP: Drop host dnsmasq usage

zeenix:
Instead of setting up a DNS on the host, we now ask NetworkManager to
add the internal DNS server to the host. The advantage here is the less
intrusive host setup but the disadvantage is that this solution will
likely not work on systems where there are already 3 name servers listed
in the `/etc/resolv.conf` since libc resolver by default only supports 3
servers.

----
As promised, a WIP branch that tries to get rid of host dnsmasq completely. It currently doesn't work though but issue doesn't seem to be DNS but rather that the VM doesn't seem to be listening on any ports. It is running though, I'm investigating..

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -17,7 +17,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine/libvirt"
-	"github.com/code-ready/crc/pkg/crc/systemd"
 	"github.com/code-ready/crc/pkg/download"
 
 	"github.com/Masterminds/semver"
@@ -28,21 +27,11 @@ import (
 const (
 	libvirtDriverCommand        = "crc-driver-libvirt"
 	libvirtDriverVersion        = "0.12.6"
-	crcDnsmasqConfigFile        = "crc.conf"
-	crcNetworkManagerConfigFile = "crc-nm-dnsmasq.conf"
 	// This is defined in https://github.com/code-ready/machine-driver-libvirt/blob/master/go.mod#L5
 	minSupportedLibvirtVersion = "3.4.0"
 )
 
 var (
-	crcDnsmasqConfigPath = filepath.Join(string(filepath.Separator), "etc", "NetworkManager", "dnsmasq.d", crcDnsmasqConfigFile)
-	crcDnsmasqConfig     = `server=/apps-crc.testing/192.168.130.11
-server=/crc.testing/192.168.130.11
-`
-	crcNetworkManagerConfigPath = filepath.Join(string(filepath.Separator), "etc", "NetworkManager", "conf.d", crcNetworkManagerConfigFile)
-	crcNetworkManagerConfig     = `[main]
-dns=dnsmasq
-`
 	libvirtDriverDownloadURL = fmt.Sprintf("https://github.com/code-ready/machine-driver-libvirt/releases/download/%s/crc-driver-libvirt", libvirtDriverVersion)
 )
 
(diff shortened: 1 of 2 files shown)


zeenix:
Failure log from `crc start`

```
INFO                                              
INFO To access the cluster, first set up your environment by following 'crc oc-env' instructions 
INFO Then you can access it by running 'oc login -u developer -p developer https://api.crc.testing:6443' 
INFO To login as an admin, username is 'kubeadmin' and password is 78UVa-zNj5W-YB62Z-ggxGZ 
INFO                                              
INFO These credentials can also be used to access the OpenShift web console at https://console-openshift-console.apps-crc.testing 
ERRO Error occurred: Error approving the node csr Not able to get csr names (exit status 1 : Unable to connect to the server: dial tcp: lookup api.crc.testing on 192.168.178.1:53: no such host
```

cfergeau:
> ```
> ERRO Error occurred: Error approving the node csr Not able to get csr names (exit status 1 : Unable to connect to the server: dial tcp: lookup api.crc.testing on 192.168.178.1:53: no such host
> ```

Is this really the VM IP? So far it's been 192.168.130.11 (`crc ip`, or `sudo virsh domifaddr crc`)

zeenix:
> Is this really the VM IP? So far it's been 192.168.130.11 (`crc ip`, or `sudo virsh domifaddr crc`)

It's not. As we figured on slack yesterday, the queries are only going to the first server in the `resolv.conf` and that's not the one I add through `nmcli connection modify crc ipv4.dns 192.168.130.11` here.

Although I doubt that's the case, I'll check if there is a way w/o non-root access to add a nameserver on top of `resolve.conf`. Otherwise, dnsmasq is the way to go I'm afraid.

praveenkumar:
Tried but it doesn't solve the issue.

```
$ crc setup
$ crc start
<csr approve is failing>

$ cat /etc/resolv.conf 
# Generated by NetworkManager
search redhat.com
nameserver 10.68.5.26
nameserver 10.64.63.6
nameserver 10.209.130.246
# NOTE: the libc resolver may not support more than 3 nameservers.
# The nameservers listed below may not be recognized.
nameserver 10.208.0.246


// Not able to provide IP during the dig command without adding the nameserver.
$  dig foo.apps-crc.testing
; <<>> DiG 9.11.10-RedHat-9.11.10-1.fc30 <<>> foo.apps-crc.testing
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 4134
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;foo.apps-crc.testing.		IN	A

;; AUTHORITY SECTION:
.			10463	IN	SOA	a.root-servers.net. nstld.verisign-grs.com. 2019092000 1800 900 604800 86400

;; Query time: 59 msec
;; SERVER: 10.68.5.26#53(10.68.5.26)
;; WHEN: Fri Sep 20 15:39:04 IST 2019
;; MSG SIZE  rcvd: 124

```

zeenix:
> Tried but it doesn't solve the issue.

Are you sure you tried this branch? The internal DNS IP should be coming on the top of other servers in your `resolv.conf` but seems like it's not listed at all in there. You need to remove the "dnsmasq" setup manually btw (and remove the dnsmasq config file setup by existing `crc`), on the host.

praveenkumar:
> Are you sure you tried this branch? 
Yes.

> The internal DNS IP should be coming on the top of other servers in your resolv.conf but seems like it's not listed at all in there. 
Same what I thought, but it is not there :(

> You need to remove the "dnsmasq" setup manually btw (and remove the dnsmasq config file setup by existing crc), on the host.
That I did before trying out this PR.

zeenix:
>  That I did before trying out this PR.

Then it's really strange. As said on slack, I'm adding the nameserver in the code through `nmcli`. See if disabling the VPN helps?

praveenkumar:
So give this another try after removing the `crc` network and then running the `setup` command which makes  `192.168.130.11` visible in the `/etc/resolve.conf` as first entry. 

```
$ cat /etc/resolv.conf 
# Generated by NetworkManager
search xxxx
nameserver 192.168.130.11
nameserver xx.xx.xx.xx
nameserver xx.xx.xx.xx
```

But due to this change the VM dns query stopped working and I am not even able to resolve the internal queries :( ( I am not sure why this is happening for internal queries)

```bash
[root@crc-vsqrt-master-0 core]# dig foo.apps-crc.testing 
; <<>> DiG 9.11.4-P2-RedHat-9.11.4-17.P2.el8_0.1 <<>> foo.apps-crc.testing
;; global options: +cmd
;; connection timed out; no servers could be reached

[root@crc-vsqrt-master-0 core]# dig google.com
; <<>> DiG 9.11.4-P2-RedHat-9.11.4-17.P2.el8_0.1 <<>> google.com
;; global options: +cmd
;; connection timed out; no servers could be reached

```

zeenix:
> So give this another try after removing the `crc` network and then running the `setup` command which makes `192.168.130.11` visible in the `/etc/resolve.conf` as first entry.

Thanks. :)

> But due to this change the VM dns query stopped working and I am not even able to resolve the internal queries :( ( I am not sure why this is happening for internal queries)

That's exactly what I see too, so it's nothing wrong with my machine I guess. Try in a bit and it might magically start to work. Also try replacing "192.168.130.1" by "192.168.130.11" in the `resolv.conf` on the node and see if that helps. Somehow it helped in on my machine.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -388,85 +377,25 @@ func fixLibvirtCrcNetworkActive() (bool, error) {
...
-	c := []byte(crcNetworkManagerConfig)
-	_, err := os.Stat(crcNetworkManagerConfigPath)
-	if err != nil {
-		return false, fmt.Errorf("File not found: %s: %s", crcNetworkManagerConfigPath, err.Error())
-	}
-	config, err := ioutil.ReadFile(filepath.Clean(crcNetworkManagerConfigPath))
+	// Set DNS on the network
+	cmd = exec.Command("nmcli", "connection", "modify", "crc",
+		"ipv4.dns", "192.168.130.11",
+		"ipv4.dns-priority", "10")
```
cfergeau: I had to run this command as root, otherwise it fails with "Insufficient privileges"

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -388,85 +377,25 @@ func fixLibvirtCrcNetworkActive() (bool, error) {
...
-
-func fixCrcNetworkManagerConfig() (bool, error) {
-	logging.Debug("Fixing NetworkManager configuration")
-	err := crcos.WriteToFileAsRoot(
-		fmt.Sprintf("write NetworkManager config in %s", crcNetworkManagerConfigPath),
-		crcNetworkManagerConfig,
-		crcNetworkManagerConfigPath,
-	)
+	// We need to do this to get NetworkManager to set the DNS in the resolv.conf
+	cmd = exec.Command("nmcli", "connection", "up", "crc")
```
cfergeau: After running this with sudo, the VM becomes unpingable from the host, so this has more consequences on host networking than adding a 192.168.130.11 entry to /etc/resolv.conf

pkg/crc/preflight/preflight_checks_linux.go
zeenix: Really? That's very strange cause here i don't need to elevate privileges. It kinda beats the point of this exercise if you need root access. :(

zeenix: @gbraad based on this and other hurdles, I don't really see a way forward here. I'd love to be wrong though. Shall I still try to think of ways this could be achieved or maybe I'm just missing some options that you'd have already thought of?

gbraad: we inch forward to a rootless setup. We can not expect all to be solved in day one. especially when even a fallback to hosts would still need root or raised permissions. this is a one time operation... however the driver using user session is a process that runs each single time. right?

zeenix: Setting the dnsmasq is a one time thing already so this won't help at all if it also requires root access, i-e we don't get any closer to non-root setup, which is why I thought perhaps this specific approach is not worth the trouble (I still haven't managed to make it work reliably or know why it won't work).

> however the driver using user session is a process that runs each single time. right?

Yes but i'm not sure about the relevance of that here. Did I miss something?

gbraad:
still WIP ?

zeenix:
> still WIP ?

yes, very much so. :) Hence my question above.

pkg/crc/preflight/preflight_checks_linux.go
gbraad: using goodhosts sounds like a non-alternative for every plaform?

zeenix: how does it get around the permissions for adding entries?

gbraad: grant permissions as part of the setup?
see macOS solution

zeenix: i don't see the advantage then. I understood that the reason we want to drop dnsmasq is move towards rootless access but both the current dnsmasq setup and your proposed solution requires privileged access at setup time *only*. No?

zeenix: your proposed solution is less intrusive for sure though so that reason might be good enough to persue it.

cfergeau: configuring network manager to use dnsmasq (as we do now) feels better to me than giving write permissions to the current user to random files in /etc

zeenix: @cfergeau I agree but I understood that it will be only a setup-time thing. i-e `crc setup` adding the needed hostnames and elevate permissions just of that. Although `crc delete` will need to undo this(?) so it'll also need to elevate permissions.

cfergeau: Ah right, the IP is fixed on linux, so we don't have to make changes every time the VM starts contrary to what happens on macos.
However, new hostnames are dynamically created when new routes are created in openshift, so we still need some kind of dynamic dns, so I'm not sure how something based on /etc/hosts would work. On macos, we only need to change it for 'oc', which only need 2 or 3 static hostnames to be set.

zeenix:
@cfergeau What does `nmcli general permissions` say on your machine? Here it's:

```
...
org.freedesktop.NetworkManager.settings.modify.system             yes     
org.freedesktop.NetworkManager.settings.modify.own                yes     
...
```

Is your user part of `wheel` group?

cfergeau:
```
$ nmcli general permissions
PERMISSION                                                        VALUE   
org.freedesktop.NetworkManager.enable-disable-network             non     
org.freedesktop.NetworkManager.enable-disable-wifi                non     
org.freedesktop.NetworkManager.enable-disable-wwan                non     
org.freedesktop.NetworkManager.enable-disable-wimax               non     
org.freedesktop.NetworkManager.sleep-wake                         non     
org.freedesktop.NetworkManager.network-control                    auth    
org.freedesktop.NetworkManager.wifi.share.protected               non     
org.freedesktop.NetworkManager.wifi.share.open                    non     
org.freedesktop.NetworkManager.settings.modify.system             auth    
org.freedesktop.NetworkManager.settings.modify.own                auth    
org.freedesktop.NetworkManager.settings.modify.hostname           auth    
org.freedesktop.NetworkManager.settings.modify.global-dns         auth    
org.freedesktop.NetworkManager.reload                             auth    
org.freedesktop.NetworkManager.checkpoint-rollback                auth    
org.freedesktop.NetworkManager.enable-disable-statistics          non     
org.freedesktop.NetworkManager.enable-disable-connectivity-check  non     
org.freedesktop.NetworkManager.wifi.scan                          inconnu 
$ groups
teuf wheel libvirt
```

pkg/crc/preflight/preflight_checks_linux.go
gbraad: > configuring network manager to use dnsmasq (as we do now) feels better to me

I do not believe we should pursue getting rid of dnsmasq, but to get to 'rootless' where possible... or at least to a point this is the least needed (preferably only during setup).

note: about macOS, it does NOT remove the entries, and this to me feel bad too.

zeenix:
Based on the discussion here and in the video call, I'm close this PR. I'll keep the issue #591 open, in case someone wants to try another approach.

PR #597, opened 2019-09-16
Use openshift version from bundle in more places

cfergeau:
Now that it's available in the bundle metadata, we can print it in more places.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -177,7 +177,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error loading bundle metadata: %v", err)
 		}
 		if bundleName != filepath.Base(startConfig.BundlePath) {
-			logging.Warnf("Bundle '%s' was requested, but loaded VM is using '%s'",
+			logging.Fatalf("Bundle '%s' was requested, but loaded VM is using '%s'",
 				filepath.Base(startConfig.BundlePath), bundleName)
 		}
 		if host.Driver.DriverName() != startConfig.VMDriver {
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -27,6 +27,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/machine/config"
 
 	"github.com/code-ready/machine/libmachine"
+	"github.com/code-ready/machine/libmachine/drivers"
 	"github.com/code-ready/machine/libmachine/host"
 	"github.com/code-ready/machine/libmachine/log"
 	"github.com/code-ready/machine/libmachine/ssh"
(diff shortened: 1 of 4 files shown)


pkg/crc/machine/machine.go
```diff
@@ -63,17 +64,18 @@ func getCrcBundleInfo(bundlePath string) (*bundle.CrcBundleInfo, error) {
 	return bundle.Extract(bundlePath)
 }
 
-func loadCrcBundleInfo(startConfig *StartConfig, bundleName string) (*bundle.CrcBundleInfo, error) {
-	crcBundleMetadata, err := bundle.GetCachedBundleInfo(bundleName)
-	if err != nil {
-		return nil, err
+func getBundleMetadataFromDriver(driver drivers.Driver) (string, *bundle.CrcBundleInfo, error) {
+	bundleName, err := driver.GetBundleName()
```
praveenkumar: @cfergeau So now we are doing it because all the driver config now have bundle info right?

pkg/crc/machine/machine.go
```diff
@@ -63,17 +64,18 @@ func getCrcBundleInfo(bundlePath string) (*bundle.CrcBundleInfo, error) {
...
+	if bundleName == "" || err != nil {
```
praveenkumar: We should add a comment here about removing `bundleName== ""` condition after 3-4 release since after that it would be not needed.

pkg/crc/machine/machine.go
```diff
@@ -513,8 +513,18 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
...
+			_, crcBundleMetadata, err := getBundleMetadataFromDriver(host.Driver)
+			if err != nil {
+				logging.Debugf("Failed to load bundle metadata: %s", err.Error())
+			}
+			if crcBundleMetadata != nil {
+				openshiftVersion = crcBundleMetadata.GetOpenshiftVersion()
+			}
+			if openshiftVersion == "" {
+				openshiftVersion = "4.x"
+			}
```
praveenkumar: @cfergeau I would do something like following.
```golang
                      openshiftVersion := "4.x"
			_, crcBundleMetadata, err := getBundleMetadataFromDriver(host.Driver)
			if err != nil {
				logging.Debugf("Failed to load bundle metadata: %s", err.Error())
			}
			if crcBundleMetadata.GetOpenshiftVersion() != "" {
				openshiftVersion = crcBundleMetadata.GetOpenshiftVersion()
			}
```

Which condition do you think when there is no error and `crcBundleMetadata` is also nil?

pkg/crc/machine/machine.go
```diff
@@ -180,12 +186,23 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 			result.Status = vmState.String()
 			return *result, nil
 		}
 
 		if vmState != state.Running {
-			logging.Infof("Starting stopped VM ...")
+			openshiftVersion := crcBundleMetadata.GetOpenshiftVersion()
+			if openshiftVersion == "" {
+				logging.Infof("Starting OpenShift VM ...")
```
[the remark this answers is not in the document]
cfergeau: If the VM is already running, we'll print `A CodeReady Containers VM is already running`, if the VM was not running before, we'll print `Starting OpenShift VM" (I think I'll change it to "CodeReady Containers"). This should be enough to tell the difference between "not running before" and "already running before"?

praveenkumar:
@cfergeau following integration test also need to be fixed.
```
21:51:09   Scenario: Check cluster health # features/story_health.feature:25
21:51:09     And stdout contains "Running (v4.x)" # features/story_health.feature:27
21:51:09       Error: output did not match. Expected: 'Running (v4.x)', Actual: 'CRC VM:          Running
21:51:09 OpenShift:       Running (v4.1.14)
21:51:09 Disk Usage:      11.1GB of 16.09GB (Inside the CRC VM)
21:51:09 Cache Usage:     14.99GB
21:51:09 Cache Directory: /home/crc_ci/.crc/cache'
```

pkg/crc/machine/machine.go
praveenkumar: @cfergeau we already updated our machine drivers, when I said crc releases that means the user side when they get those updated machine drivers :) Like with this beta release user should have updated machine driver.

test/integration/features/story_health.feature
```diff
@@ -24,7 +24,7 @@ Feature:
     @darwin @linux 
     Scenario: Check cluster health
         Given executing "crc status" succeeds
-        And stdout contains "Running (v4.x)"
+        And stdout contains "Running (v4."
```
praveenkumar: :+1:

test/integration/features/story_registry.feature
```diff
@@ -11,7 +11,7 @@ Feature: Local image to image-registry to deployment
         When starting CRC with default bundle and hypervisor "<vm-driver>" succeeds
         Then stdout should contain "CodeReady Containers instance is running"
         And executing "eval $(crc oc-env)" succeeds
-        When with up to "4" retries with wait period of "2m" command "crc status" output should contain "Running (v4.x)"
+        When with up to "4" retries with wait period of "2m" command "crc status" output should contain "Running (v4"
```
praveenkumar: @cfergeau you put `.` in every other place and forgot here for `v4`  :)

cfergeau: Ah thanks, I remember thinking about consistency when doing these changes, but I did not double-check they actually were the same ;)

pkg/crc/machine/machine.go
cfergeau: For the latest machine-driver-libvirt (0.12.5), go.mod is https://github.com/code-ready/machine-driver-libvirt/blob/880ceaa894a0bc65d2db413af58fb827ff599c88/go.mod with machine commit `34d5b6049a38`. This commit is one commit _before_ the one adding the error when bundle name is empty (see https://github.com/code-ready/machine/commits/master ). So I don't think the current builds of the machine drivers have this code just yet.

praveenkumar: Ah :( we should've done it before this release. I will create PR for libvirt and hyperkit side to update it with latest machine code.

Issue #601, opened 2019-09-17
Remove some of the expected output of `setup` command from integration tests


jsliacan:
### General information

  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / Hyper-V / VirtualBox / hyperkit

## CRC version
```bash
version: 1.0.0-beta.2+95f7543
OpenShift version: 4.1.14 (not embedded in binary)
```
Remove lines from `test/integration/basic.feature` containing `Will use root access`, as those are not in stdout when setup is executed. Also remove lines mentioning if bundle is embedded or not, as those are not part of the output on each run.

jsliacan:
Solved by #602

PR #602, opened 2019-09-17
Issue #601 Adjust expected output in integration tests to match output of setup cmd

jsliacan:
Did what it says on the tin, plus changed `virtualbox` hyperviser to `hyperkit` when passed as `-d` option to `crc start` in basic tests.

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -13,6 +13,13 @@ Feature: Basic test
         When executing "crc --help" succeeds
         Then stdout should contain "Usage:"
         And stdout should contain "Available Commands:"
+        And stdout should contain "help"
+        And stdout should contain "version"
+        And stdout should contain "setup"
+        And stdout should contain "start"
+        And stdout should contain "stop"
+        And stdout should contain "delete"
+        And stdout should contain "status"
         And stdout should contain "Flags:"
         And stdout should contain 
             """
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -47,22 +47,25 @@ Feature: Basic test
         And stdout should contain "Checking if NetworkManager is installed"
         And stdout should contain "Checking if NetworkManager service is running"
         And stdout should contain "Writing Network Manager config for crc"
-	And stdout should contain "Will use root access: write NetworkManager config in /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf"
+        And stdout should contain "Will use root access: write NetworkManager config in /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf"
         And stdout should contain "Will use root access: execute systemctl daemon-reload command"
         And stdout should contain "Will use root access: execute systemctl stop/start command"
         And stdout should contain "Writing dnsmasq config for crc"
-	And stdout should contain "Will use root access: write dnsmasq configuration in /etc/NetworkManager/dnsmasq.d/crc.conf"
-	And stdout should contain "Will use root access: execute systemctl daemon-reload command"
-	And stdout should contain "Will use root access: execute systemctl stop/start command"
+        And stdout should contain "Will use root access: write dnsmasq configuration in /etc/NetworkManager/dnsmasq.d/crc.conf"
+        And stdout should contain "Will use root access: execute systemctl daemon-reload command"
+        And stdout should contain "Will use root access: execute systemctl stop/start command"
         And stdout should contain "Unpacking bundle from the CRC binary"
-        And stdout should contain "Setup is complete, you can now run 'crc start -b $bundlename' to start a CodeReady Containers instance"
+        And stdout should contain "Setup is complete"
 
     @darwin
     Scenario: CRC setup on Mac
         When executing "crc setup" succeeds
-        Then stdout should contain "Caching oc binary"
+        Then stdout should contain "Checking if running as non-root" 
+        And stdout should contain "Caching oc binary"
         And stdout should contain "Setting up virtualization"
-        And stdout should contain "Setting file permissions for resolver"
+        And stdout should contain "Will use root access: change ownership"
+        And stdout should contain "Will use root access: set suid"
+        And stdout should contain "Setting file permissions"
 
     @linux
     Scenario: CRC start on Linux
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -81,7 +81,7 @@ Feature: Basic test
 
     @darwin
     Scenario: CRC start on Mac
-        When starting CRC with default bundle and hypervisor "virtualbox" succeeds
+        When starting CRC with default bundle and hypervisor "hyperkit" succeeds
         Then stdout should contain "CodeReady Containers instance is running"
     
     @darwin @linux @windows


jsliacan:
retest this please

test/integration/features/basic.feature
```diff
@@ -35,34 +35,25 @@ Feature: Basic test
         And stdout should contain "Setting up KVM"
         And stdout should contain "Installing libvirt service and dependencies" 
         And stdout should contain "Adding user to libvirt group"
-        And stdout should contain "Will use root access: add user to libvirt group"
         And stdout should contain "Enabling libvirt"
         And stdout should contain "Starting libvirt service"
-        And stdout should contain "Will use root access: start libvirtd service"
```
cfergeau: I don't understand why these would not be part of stdout?
```
INFO Starting libvirt 'crc' network               
INFO Writing Network Manager config for crc       
INFO Writing dnsmasq config for crc               
INFO Will use root access: write dnsmasq configuration in /etc/NetworkManager/dnsmasq.d/crc.conf 
INFO Will use root access: execute systemctl daemon-reload command 
INFO Will use root access: execute systemctl stop/start command 
```
This is what I get with a `crc setup` on my test machine.

jsliacan: @cfergeau it looks like the output does not contain the `Will use root access` even on CI (see CentosCI console output). So, I guess, it'll be necessary to remove the `/etc/NetworkManager/dnsmasq.d/crc.conf` file before the feature if we want to test it from fresh.

cfergeau: if you look at https://github.com/code-ready/crc/pull/587 , CI was successful and the integration tests have "Will use root access" https://github.com/cfergeau/crc/blob/win-lint/test/integration/features/basic.feature so I don't understand what's going on :-/
Removing `/etc/NetworkManager/dnsmasq.d/crc.conf` just causes one of the tests to show the "Will use root access" message, not all of them :(

jsliacan: > Removing /etc/NetworkManager/dnsmasq.d/crc.conf just causes one of the tests to show the "Will use root access" message, not all of them :(
This would be enough, right? We want to test that line of execution at least once, so exactly once would be good I guess.

I'm a fair bit confused too! Will look around a bit more for some nicer solution.

cfergeau: > This would be enough, right? We want to test that line of execution at least once, so exactly once would be good I guess.

We also want to test `crc setup` on a blank machine, I was thinking centos CI would do that. If you tell me the "Will use root access:" lines need to be removed, that means we are not testing at all the host machine configuration `crc` is supposed to be doing.
If there is one `crc setup` test checking this, and this is about running `crc setup` a second time on the same host, then it's fine.

jsliacan: @cfergeau sorry, I'm confused. I want to remove the file `/etc/NetworkManager/dnsmasq.d/crc.conf` before the testsuite, not the lines containing `Will use root access:` from the test itself. Then inside the tests, the first time `crc setup` is executed, I'll expect `Will use root access:` lines. The second (and later) time `crc setup` is executed in the tests, I will not expect the `Will use root access:` lines. Which part of configuration that `crc` is supposed to be doing wouldn't be tested this way? Cheers!

cfergeau: The very first time `crc setup` is run, you should get `Will use root access:` lines even without removing `/etc/NetworkManager/dnsmasq.d/crc.conf` before. If you need to remove it, you'll likely only get one such line, rather than multiple ones.
As I was saying in https://github.com/code-ready/crc/pull/602#discussion_r326082144 , I don't think this is needed as we have CI currently passing with the integration test checking for "Will use root access:" and not complaining about it.
So I don't understand what changed and what we are discussing exactly :-/

jsliacan: Cool, I just assumed that "The very first time" means literally only once on each machine. Then simulating "the very first time" by removing the file before the test would make sense.

> If you need to remove it, you'll likely only get one such line, rather than multiple ones.

I got all three when I checked this. Hmmm...

> we have CI currently passing with the integration test checking for "Will use root access:" and not complaining about it.

Agreed, that's the puzzle now.

cfergeau: > Cool, I just assumed that "The very first time" means literally only once on each machine. Then simulating "the very first time" by removing the file before the test would make sense.

Just one more clarification on this. Each preflight check/fix is independent, if you remove `/etc/NetworkManager/dnsmasq.d/crc.conf` then the preflight check for this file existence will fail, and the fix will run. Other unrelated checks/fixes won't be run (for example, the codepath doing "sudo systemctl enable libvirtd" won't be exercised).

> 
> > If you need to remove it, you'll likely only get one such line, rather than multiple ones.
> 
> I got all three when I checked this. Hmmm...
> 
> > we have CI currently passing with the integration test checking for "Will use root access:" and not complaining about it.
> 
> Agreed, that's the puzzle now.

test/integration/features/#basic.feature#
```diff
@@ -0,0 +1,100 @@
...
+    Scenario: CRC version
+        When executing "crc version" succeeds
+        Then stderr should be empty
+        And stdout should contain "version:"
+
+    @darwin @linux @windows
+    Scenario: CRC help
+        When executing "crc --help" succeeds
+        Then stdout should contain "Usage:"
+        And stdout should contain "Available Commands:"
```
gbraad: we might need to check for a subset of commands that should NEVER be hidden/always available.

or perhaps even exclude certain things (since with Minishift we accidently added test commands).

test/integration/features/basic.feature
gbraad: so machines aren't really clean?

test/integration/features/#basic.feature#
jsliacan: I added checks for `help` and `version`, as well as `setup`, `start`, `stop`, `delete`, and `status`. Seemed like the "core" ones.

jsliacan:
For completeness: the `Scenario` about `setup` on Windows is being addressed in #543 .

test/integration/features/basic.feature
jsliacan: @cfergeau @gbraad I haven't encountered a case where `Will use root access:` wasn't part of the output. So the issue seems to have gone away for the time being. This also makes the changes that this PR introduces much smaller - but there are some.

centos_ci.sh
```diff
@@ -124,14 +124,14 @@ function upload_logs() {
...
-    exit 1
-  fi
+	set +e
+	# In Jenkins slave we have pull secret file in the $HOME/payload/crc_pull_secret
+	# this is copied over using https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml#L99
+	make integration BUNDLE_LOCATION=$HOME/Downloads/$BUNDLE PULL_SECRET_FILE=$HOME/payload/crc_pull_secret
+	if [[ $? -ne 0 ]]; then
+		upload_logs $1
+		exit 1
+	fi
```
praveenkumar: @jsliacan  This changes shouldn't be part of this PR. it is just matter of test.

test/integration/features/basic.feature
```diff
@@ -53,6 +69,8 @@ Feature: Basic test
         Then stdout should contain "Checking if running as non-root" 
         And stdout should contain "Caching oc binary"
         And stdout should contain "Setting up virtualization"
+        And stdout should contain "Will use root access: change ownership"
+        And stdout should contain "Will use root access: set suid"
         And stdout should contain "Setting file permissions"
```
cfergeau: Up to this point, none of the changes in this commit seem mac-related?

test/integration/features/basic.feature
```diff
@@ -35,16 +42,25 @@ Feature: Basic test
...
         And stdout should contain "Writing Network Manager config for crc"
+        And stdout should contain "Will use root access: write NetworkManager config in /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf"
+        And stdout should contain "Will use root access: execute systemctl daemon-reload command"
+        And stdout should contain "Will use root access: execute systemctl stop/start command"
         And stdout should contain "Writing dnsmasq config for crc"
+        And stdout should contain "Will use root access: write dnsmasq configuration in /etc/NetworkManager/dnsmasq.d/crc.conf"
+        And stdout should contain "Will use root access: execute systemctl daemon-reload command"
+        And stdout should contain "Will use root access: execute systemctl stop/start command"
+        And stdout should contain "Unpacking bundle from the CRC binary"
         And stdout should contain "Setup is complete"
```
cfergeau: Up to this point, none of the changes in this commit  seem mac-related?

jsliacan: @cfergeau yes, that should be correct. Why is it bad?

cfergeau: They are part of a commit whose log is ` extra: changed virtualbox to hyperkit in crc start for mac with -d...` so it's not really clear why they are in this commit, what they are going to achieve, ...

jsliacan: @cfergeau updated & split into 3 commits topic-wise. Thanks for noticing

PR #619, opened 2019-09-20
Issue #609 Create release info file in json format

### LINKED ISSUE #609: Generate release info in JSON format
After a discussion with @praveenkumar on slack, we decided to have the release info (meta-data) file in JSON format and it needs to be generated via calling the `release` target from makefile.

The generated `release-info.json` file would look like:
```json
{	
	"description": "Bringing a minimal OpenShift 4.1 or newer cluster to your local laptop or desktop computer.",
	"version": {
		"crcVersion": "1.0.0-beta.5",
		"gitSha": "f2aa58c"
	},
	"openshiftVersion": "4.1.14" 
}
```

This would be needed for #484

anjannath:
What this PR does is add a message about available update when the user runs `crc start` command, and asks if he wants to keep using the current version or stop the current version (to download the newer one).
When a new version is available in mirror.openshift.com, and the user starts crc, he'll get a prompt like following:

```
❯ crc start -b ~/Downloads/crc_archives/crc_libvirt_4.1.14.crcbundle -p ~/Documents/pull-secret --log-level debug
INFO A new version (1.0.0-beta.6) has been published 
Would you like to continue using the current version? [y/N]: 
```
If answered No, it prints the URL to download the latest version of `crc`. (going forward we'd automatically download the latest version but that is for later.)

```
❯ crc start -b ~/Downloads/crc_archives/crc_libvirt_4.1.14.crcbundle -p ~/Documents/pull-secret --log-level debug
INFO A new version (1.0.0-beta.6) has been published 
Would you like to continue using the current version? [y/N]: n
Please download the new version from https://cloud.redhat.com/openshift/install/crc/installer-provisioned
```
Note: it'll always interrupt the start once a new version is available

However in https://github.com/code-ready/crc/pull/619#discussion_r328507543 it is mentioned that prompting the user like this, is disruptive, and a WARN msg would be enough.

Fixes #609

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,8 @@ REPOPATH ?= $(ORG)/crc
 PACKAGES := go list ./... | grep -v /out
 SOURCES := $(shell git ls-files  *.go ":^vendor")
 
+RELEASE_INFO := release-info.json
+
 # Check that given variables are set and all have non-empty values,
 # die with an error otherwise.
 #
@@ -154,8 +156,14 @@ $(GOPATH)/bin/gosec:
 gosec: $(GOPATH)/bin/gosec
 	$(GOPATH)/bin/gosec -tests -severity medium  ./...
 
+.PHONY: gen_release_info
+gen_release_info:
+	@cat release-info.json.sample | sed s/@CRC_VERSION@/\"$(CRC_VERSION)\"/ > $(RELEASE_INFO)
+	@sed -i s/@GIT_COMMIT_SHA@/\"$(COMMIT_SHA)\"/ $(RELEASE_INFO)
+	@sed -i s/@OPENSHIFT_VERSION@/\"$(BUNDLE_VERSION)\"/ $(RELEASE_INFO)
+
 .PHONY: release
-release: fmtcheck embed_bundle build_docs_pdf
+release: fmtcheck embed_bundle build_docs_pdf gen_release_info
 	mkdir $(RELEASE_DIR)
 	
 	@mkdir -p $(BUILD_DIR)/crc-macos-$(CRC_VERSION)-amd64
@@ -170,6 +178,8 @@ release: fmtcheck embed_bundle build_docs_pdf
 	@cp LICENSE $(DOCS_BUILD_DIR)/doc.pdf $(BUILD_DIR)/windows-amd64/crc.exe $(BUILD_DIR)/crc-windows-$(CRC_VERSION)-amd64
 	cd $(BUILD_DIR) && zip -r $(CURDIR)/$(RELEASE_DIR)/crc-windows-amd64.zip crc-windows-$(CRC_VERSION)-amd64
 
+	@mv $(RELEASE_INFO) $(RELEASE_DIR)/$(RELEASE_INFO)
+	
 	pushd $(RELEASE_DIR) && sha256sum * > sha256sum.txt && popd
 
 HYPERKIT_BUNDLENAME = $(BUNDLE_DIR)/crc_hyperkit_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -19,6 +19,8 @@ var (
 	NameServer     = cfg.AddSetting("nameserver", nil, []cfg.ValidationFnType{cfg.ValidateIpAddress}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	PullSecretFile = cfg.AddSetting("pull-secret-file", nil, []cfg.ValidationFnType{cfg.ValidatePath}, []cfg.SetFn{cfg.SuccessfullyApplied})
 
+	DisableUpdateCheck = cfg.AddSetting("disable-update-check", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+
 	// Preflight checks
 	SkipCheckVirtualBoxInstalled = cfg.AddSetting("skip-check-virtualbox-installed", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	WarnCheckVirtualBoxInstalled = cfg.AddSetting("warn-check-virtualbox-installed", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -17,6 +17,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/output"
 	"github.com/code-ready/crc/pkg/crc/preflight"
 	"github.com/code-ready/crc/pkg/crc/validation"
+	"github.com/code-ready/crc/pkg/crc/version"
 )
 
 func init() {
@@ -43,6 +44,8 @@ func runStart(arguments []string) {
 	if err := validateStartFlags(); err != nil {
 		errors.Exit(1)
 	}
+
+	checkIfNewVersionAvailable(crcConfig.GetBool(config.DisableUpdateCheck.Name))
 
 	preflight.StartPreflightChecks(crcConfig.GetString(config.VMDriver.Name))
 
@@ -76,6 +79,7 @@ func initStartCmdFlagSet() *pflag.FlagSet {
 	flagSet.IntP(config.CPUs.Name, "c", constants.DefaultCPUs, "Number of CPU cores to allocate to the CRC VM")
 	flagSet.IntP(config.Memory.Name, "m", constants.DefaultMemory, "MiB of Memory to allocate to the CRC VM")
 	flagSet.StringP(config.NameServer.Name, "n", "", "Specify nameserver to use for the instance. (i.e. 8.8.8.8)")
+	flagSet.Bool(config.DisableUpdateCheck.Name, false, "Don't check for update")
 
 	return flagSet
 }
@@ -132,4 +136,20 @@ func getPullSecretFileContent() (string, error) {
 	}
 
 	return pullsecret, nil
+}
+
+func checkIfNewVersionAvailable(noUpdateCheck bool) {
+	if noUpdateCheck {
+		return
+	}
+	isNewVersionAvailable, newVersion, err := version.NewVersionAvailable()
+	if err != nil {
+		logging.Debugf("Error checking if a new version is available: %v", err)
+		return
+	}
+	if isNewVersionAvailable {
+		logging.Warnf("A new version (%s) has been published on %s", newVersion, constants.CrcLandingPageURL)
+		return
+	}
+	logging.Debugf("No new version available, latest version at mirror is %s", newVersion)
 }
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -72,7 +72,7 @@ func initStartCmdFlagSet() *pflag.FlagSet {
 	flagSet := pflag.NewFlagSet("start", pflag.ExitOnError)
 	flagSet.StringP(config.Bundle.Name, "b", constants.DefaultBundlePath, "The system bundle used for deployment of the OpenShift cluster.")
 	flagSet.StringP(config.VMDriver.Name, "d", machine.DefaultDriver.Driver, fmt.Sprintf("The driver to use for the CRC VM. Possible values: %v", machine.SupportedDriverValues()))
-	flagSet.StringP(config.PullSecretFile.Name, "p", "", fmt.Sprintf("File path of Image pull secret for User (Download it from %s)", constants.DefaultPullSecretURL))
+	flagSet.StringP(config.PullSecretFile.Name, "p", "", fmt.Sprintf("File path of Image pull secret for User (Download it from %s)", constants.CrcLandingPageURL))
 	flagSet.IntP(config.CPUs.Name, "c", constants.DefaultCPUs, "Number of CPU cores to allocate to the CRC VM")
 	flagSet.IntP(config.Memory.Name, "m", constants.DefaultMemory, "MiB of Memory to allocate to the CRC VM")
 	flagSet.StringP(config.NameServer.Name, "n", "", "Specify nameserver to use for the instance. (i.e. 8.8.8.8)")
@@ -113,7 +113,7 @@ func getPullSecretFileContent() (string, error) {
 
 	// In case user doesn't provide a file in start command or in config then ask for it.
 	if crcConfig.GetString(config.PullSecretFile.Name) == "" {
-		pullsecret, err = input.PromptUserForSecret("Image pull secret", fmt.Sprintf("Copy it from %s", constants.DefaultPullSecretURL))
+		pullsecret, err = input.PromptUserForSecret("Image pull secret", fmt.Sprintf("Copy it from %s", constants.CrcLandingPageURL))
 		// This is just to provide a new line after user enter the pull secret.
 		fmt.Println()
 		if err != nil {
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -16,6 +16,15 @@ limitations under the License.
 
 package version
 
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/Masterminds/semver"
+	"io/ioutil"
+	"net/http"
+	"time"
+)
+
 // The following variables are private fields and should be set when compiling with ldflags, for example --ldflags="-X github.com/code-ready/crc/pkg/version.crcVersion=vX.Y.Z
 var (
 	// The current version of minishift
@@ -28,6 +37,16 @@ var (
 	bundleVersion = "0.0.0-unset"
 )
 
+const (
+	releaseInfoLink = "https://mirror.openshift.com/pub/openshift-v4/clients/crc/latest/release-info.json"
+)
+
+type CrcReleaseInfo struct {
+	Version struct {
+		LatestVersion string `json:"crcVersion"`
+	}
+}
+
 func GetCRCVersion() string {
 	return crcVersion
 }
@@ -38,4 +57,41 @@ func GetCommitSha() string {
 
 func GetBundleVersion() string {
 	return bundleVersion
+}
+
+func getCRCLatestVersionFromMirror() (*semver.Version, error) {
+	var releaseInfo CrcReleaseInfo
+	client := &http.Client{
+		Timeout: 5 * time.Second,
+	}
+	response, err := client.Get(releaseInfoLink)
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+	releaseMetaData, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal(releaseMetaData, &releaseInfo)
+	if err != nil {
+		return nil, fmt.Errorf("Error unmarshaling JSON metadata: %v", err)
+	}
+	version, err := semver.NewVersion(releaseInfo.Version.LatestVersion)
+	if err != nil {
+		return nil, err
+	}
+	return version, nil
+}
+
+func NewVersionAvailable() (bool, string, error) {
+	latestVersion, err := getCRCLatestVersionFromMirror()
+	if err != nil {
+		return false, "", err
+	}
+	currentVersion, err := semver.NewVersion(GetCRCVersion())
+	if err != nil {
+		return false, "", err
+	}
+	return latestVersion.GreaterThan(currentVersion), latestVersion.String(), nil
 }
--- a/release-info.json.sample
+++ b/release-info.json.sample
@@ -1,0 +1,7 @@
+{
+	"version": {
+		"crcVersion": @CRC_VERSION@,
+		"gitSha": @GIT_COMMIT_SHA@,
+		"openshiftVersion": @OPENSHIFT_VERSION@
+	}
+}
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -23,7 +23,7 @@ const (
 	ConfigFile           = "crc.json"
 	LogFile              = "crc.log"
 	GlobalStateFile      = "globalstate.json"
-	DefaultPullSecretURL = "https://cloud.redhat.com/openshift/install/crc/installer-provisioned" // #nosec G101
+	CrcLandingPageURL    = "https://cloud.redhat.com/openshift/install/crc/installer-provisioned" // #nosec G101
 	PullSecretFile       = "pullsecret.json"
 )
 


Makefile
```diff
@@ -38,6 +38,8 @@ REPOPATH ?= $(ORG)/crc
 PACKAGES := go list ./... | grep -v /out
 SOURCES := $(shell git ls-files  *.go ":^vendor")
 
+JQ ?= ./jq
```
gbraad: So we need an additional dependency ...

Makefile
```diff
@@ -149,6 +161,13 @@ $(GOPATH)/bin/gosec:
 gosec: $(GOPATH)/bin/gosec
 	$(GOPATH)/bin/gosec -tests -severity medium  ./...
 
+.PHONY: gen_release_info
```
gbraad: ... and a similar template ...

Makefile
```diff
@@ -74,6 +76,16 @@ vendor:
...
 
+# Get JQ based on the platform (linux or darwin)
+$(JQ):
+ifeq ($(GOOS),linux)
+	@curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -o jq
+endif
+ifeq ($(GOOS),darwin)
+	@curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-osx-amd64 -o jq
+endif
+	@chmod +x jq
```
praveenkumar: Instead adding the `jq` dependency can we just add the place holder in the sample file and use `envsubst` ?

```bash
$ cat release.json 
{
	"description": "Bringing a minimal OpenShift 4.1 or newer cluster to your local laptop or desktop computer.",
	"version": {
		"crcVersion": "$crcVersion",
		"gitSha": "$gitSha"
	},
	"openshiftVersion": "$openshiftVersion"
}

$ crcVersion=1.0.0-beta.5 gitSha=f2aa58c openshiftVersion=4.1.14 envsubst < release.json 
{
	"description": "Bringing a minimal OpenShift 4.1 or newer cluster to your local laptop or desktop computer.",
	"version": {
		"crcVersion": "1.0.0-beta.5",
		"gitSha": "f2aa58c"
	},
	"openshiftVersion": "4.1.14"
}
```

cmd/crc/cmd/start.go
```diff
@@ -43,6 +44,16 @@ func runStart(arguments []string) {
 	if err := validateStartFlags(); err != nil {
 		errors.Exit(1)
 	}
+	newVersionAvailable, newVersion, err := version.NewVersionAvailable()
+	if err == nil && newVersionAvailable {
+		logging.Infof("A new version (%s) is available for download", newVersion)
+		yes := input.PromptUserForYesOrNo("Would you like to continue using the current version", false)
+		if !yes {
```
praveenkumar: I think we should change this logic, I mean it is confusing that using `!yes` and then asking to download new version. Also there should be a flag which disable this prompt (either in config or as start) ?

Makefile
cfergeau: Or the more usual `sed`?

gbraad: besides, using a 'released' version like this needs trust. I'd rather go for a packaged version... but as I said before, not at all. rather not see another dependency

cmd/crc/cmd/start.go
gbraad: variables named `yes` are hard to decipher, as you need to look at previous lines to give it meaning. Rather see it being named `hasNewVersionAvailable` or similarly.

Makefile
anjannath: removed `jq`, using `sed` now

Makefile
anjannath: Went with using `sed`, as with that we can avoid having to pass the envs.

cmd/crc/cmd/start.go
anjannath: Renamed the variables

anjannath: @praveenkumar Added, config option and flag to start command, `--no-update-check`

go.mod
```diff
@@ -4,6 +4,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/DATA-DOG/godog v0.7.13
 	github.com/YourFin/binappend v0.0.0-20181105185800-0add4bf0b9ad
+	github.com/blang/semver v0.0.0-20190414102917-ba2c2ddd8906
```
praveenkumar: @anjannath with https://github.com/code-ready/crc/commit/d5bf2551426cd05bf52faafe20a8a4aeb026672e one I already started using `github.com/Masterminds/semver` since it is much more updated and not maintained by a single person.

anjannath: @praveenkumar Done, using `github.com/Masterminds/semver` now.

cmd/crc/cmd/start.go
```diff
@@ -133,3 +137,20 @@ func getPullSecretFileContent() (string, error) {
 
 	return pullsecret, nil
 }
+
+func checkIfNewVersionAvailable(noUpdateCheck bool) {
+	if noUpdateCheck {
+		return
+	} else {
```
praveenkumar: @anjannath you don't need `else` here if you already returning it from the `if` block.

pkg/crc/version/version.go
```diff
@@ -39,3 +57,37 @@ func GetCommitSha() string {
...
+	var releaseInfo CrcReleaseInfo
+	response, err := http.Get(releaseInfoLink)
+	if err != nil {
+		return &semver.Version{}, err
+	}
+	releaseMetaData, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return &semver.Version{}, err
+	}
+	response.Body.Close()
```
praveenkumar: shouldn't this be like `defer response.Body.Close()` so that if you hit the error during getting the page you should also close it?

cmd/crc/cmd/start.go
anjannath: Right, i'll push another update, Thanks!

pkg/crc/version/version.go
anjannath: Well, we are already checking the error before the close, and if no errors encountered, then only we close, otherwise we'd try to close on a null.

praveenkumar: Check https://golang.org/pkg/net/http/ once second section, I think as a better practice you should always close the connection.

gbraad: `defer` is always the better option here.

anjannath: Right, that uses `defer` to close, why i went without this is, i didn't want the start command to panic, suppose the `Get` method returns an error, since using `defer` would call the `.Close()` func on `response` all the time, even when its null we might get into a panic.

release-info.json.sample
```diff
@@ -0,0 +1,8 @@
+{
+	"description": "Bringing a minimal OpenShift 4.1 or newer cluster to your local laptop or desktop computer.",
+	"version": {
+		"crcVersion": fillCrcVersion,
+		"gitSha": fillGitCommitSha
+	},
+	"openshiftVersion": fillOpenShiftVersion
```
cfergeau: Mostly personal preference, but I'd add obvious markers for whta's going to be substituted, for example @CRC_VERSION@, @GIT_COMMIT_SHA@, @OPENSHIFT_VERSION@
A bit odd that "openshiftVersion" is not part of the "version" struct.

gbraad: Why a description if this is not intended to be read by a human?

cmd/crc/cmd/start.go
```diff
@@ -133,3 +137,23 @@ func getPullSecretFileContent() (string, error) {
...
+	if err != nil {
+		logging.Debugf("Error checking if a new version is available: %v", err)
+		return
+	}
+	if newVersionAvailable {
+		logging.Infof("A new version (%s) has been published", newVersion)
+		userWantsToContinue := input.PromptUserForYesOrNo("Would you like to continue using the current version", false)
+		if !userWantsToContinue {
+			errors.ExitWithMessage(0, fmt.Sprintf("Please download the new version from %s", constants.DefaultPullSecretURL))
+		}
```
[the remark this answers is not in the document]
anjannath: We'd only prompt when a new version is released, and at that point the user should actually update since the certs are also likely to expire, also a flag to start command `--no-update-check` is added, and similar config option to disable this, we could mention about this in the message. Do you still think we should not show the prompt and just a warning?

release-info.json.sample
anjannath: Can remove the description..

pkg/crc/version/version.go
```diff
@@ -39,3 +57,38 @@ func GetCommitSha() string {
 func GetBundleVersion() string {
 	return bundleVersion
 }
+
+func getCRCLatestVersionFromMirror() (*semver.Version, error) {
+	var releaseInfo CrcReleaseInfo
+	response, err := http.Get(releaseInfoLink)
+	if err != nil {
+		fmt.Println(err)
+		return &semver.Version{}, err
```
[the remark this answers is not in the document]
anjannath: earlier, this was returning `semver.Version` that's why couldn't use nil, since now its returning a pointer, i can use `nil`, Thanks!

anjannath:
> Is the goal to have this included in the RC ?

@gbraad Yes, good to have this in, users would be able to know about the next update sooner, before their certificates expire.

gbraad:
>  Yes, good to have this in,

what was/is not according to schedule, as we have planned/decided to move anything related to update strategy until after the RC/GA

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
cfergeau: Well, once a new version is released, we'll always prompt. And maybe the certs will expire soon, maybe not, maybe we made a minor release to fix a critical bug on a platform you are not using. In any case, a prompt saying "Do you want to abort?" is odd.

anjannath:
> what was/is not according to schedule, as we have planned/decided to move anything related to update strategy until after the RC/GA

I don't remember us actually planning to do this, but since the 4.2 was delayed, we got some more time and its like _good to have_, not necessary for RC!

gbraad:
> good to have

... if strategy and use is clear. since once this is in, we can not just change behaviour.

cmd/crc/cmd/start.go
```diff
@@ -133,3 +137,23 @@ func getPullSecretFileContent() (string, error) {
...
+	newVersionAvailable, newVersion, err := version.NewVersionAvailable()
+	if err != nil {
+		logging.Debugf("Error checking if a new version is available: %v", err)
+		return
+	}
+	if newVersionAvailable {
+		logging.Infof("A new version (%s) has been published", newVersion)
+		userWantsToContinue := input.PromptUserForYesOrNo("Would you like to continue using the current version", false)
+		if !userWantsToContinue {
+			errors.ExitWithMessage(0, fmt.Sprintf("Please download the new version from %s", constants.DefaultPullSecretURL))
```
cfergeau: forgot to mention, ExitWithMessage is a variadic function, so you can remove the Sprintf
```
errors.ExitWithMessage(0, "Please download the new version from %s", constants.DefaultPullSecretURL)
```
Now that I copied/pasted this, I notice `DefaultPullSecretURL` which should be renamed maybe?

anjannath:
> ... if strategy and use is clear. since once this is in, we can not just change behaviour.

strategy wise, these changes are only checking if a new version is released, which we'd always need to know if we want to or need to upgrade. i don't think the actual upgrade strategy is defined by it.

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
anjannath: So, a `WARN` would be enough here? cc @code-ready/crc-devel

gbraad:
> asks if he wants to keep using the current version or stop the current version (to download the newer one).

This means that `start` will have two different behaverioul paths; no new version and just continues. and one that blocks the user on start. This sounds like a bad feature, as this is not something the user will expect to happen. He might start it and walk away to get coffee, and upon return sees a prompt. A user might do this from a CI script, and the build just hangs awaiting input... or the adapter team can not give input, as they use a non-pty terminal.

gbraad:
> Note: it'll always interrupt the start once a new version is available

Please explain `once`: it will always interrupt the flow when a new version is available? or do we keep state about the version we want to ignore?

praveenkumar:
> When a new version is available in mirror.openshift.com, and the user starts crc, he'll get a prompt like following:

I would say just add `warn` message with download URI and not stop for any prompt part for `crc start` we will later introduce `crc update` command then we can easily change this message by adding `crc update` at the end. This way it is less destructive for our `crc start` side also.

anjannath:
>  This sounds like a bad feature, as this is not something the user will expect to happen.

Right, this will be like a sudden behaviour change, although there's a config option and a flag for `crc start` to disable it, but if the user doesn't know about it, that doesn't help much. So the other option is to just show a `WARN` with the update message on start as @cfergeau suggested.

> Please explain once: it will always interrupt the flow when a new version is available? or do we keep state about the version we want to ignore?

It'll always interrupt the flow when a new version is available, until it is updated, once it is updated to the latest one, there's no new version so the prompting stops.

cmd/crc/cmd/start.go
anjannath: @cfergeau will change the name of this constant to `CrcLandingPageURL`.

anjannath:
> Maybe we should only do this once every few days, and stop doing it once we detected a new version is available, but that's additional code that can be done on top of this.

@cfergeau Sure, i can send a follow up patch for these improvements.

gbraad:
how about `INFO Version is still current` on normal flow, and `WARN New version is available` when this is the case. This way there is always the same amount of message and the user can see we check for this. Since when there is no connection, I also expect to see a `WARN Unable to determine if version is current` since the conbnection either failed or the file didn't parse.

anjannath:
@gbraad The messages about `Failed checking for new version` and `Already running the current version` is already part of the debug logs as i think these won't be much useful to the user.

Makefile
```diff
@@ -172,6 +180,8 @@ release: fmtcheck embed_bundle build_docs_pdf
 
 	pushd $(RELEASE_DIR) && sha256sum * > sha256sum.txt && popd
 
+	gen_release_info
```
praveenkumar: @anjannath this will not work since it is a target not the command.
```
$ BUNDLE_DIR=/tmp/bundle make release
rm -rf /home/prkumar/work/github/crc/docs/build
rm -rf out
rm -f /home/prkumar/work/go/bin/crc
rm -rf release
GOARCH=amd64 GOOS=darwin go build -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=1.0.0-dev -X github.com/code-ready/crc/pkg/crc/version.bundleVersion=4.1.14 -X github.com/code-ready/crc/pkg/crc/version.commitSha=7bc3124 -extldflags='-static' -s -w -X github.com/code-ready/crc/pkg/crc/constants.bundleEmbedded=true" -o out/macos-amd64/crc ./cmd/crc
GOOS=linux GOARCH=amd64 go build -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=1.0.0-dev -X github.com/code-ready/crc/pkg/crc/version.bundleVersion=4.1.14 -X github.com/code-ready/crc/pkg/crc/version.commitSha=7bc3124 -extldflags='-static' -s -w -X github.com/code-ready/crc/pkg/crc/constants.bundleEmbedded=true" -o out/linux-amd64/crc ./cmd/crc
GOARCH=amd64 GOOS=windows go build -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=1.0.0-dev -X github.com/code-ready/crc/pkg/crc/version.bundleVersion=4.1.14 -X github.com/code-ready/crc/pkg/crc/version.commitSha=7bc3124 -extldflags='-static' -s -w -X github.com/code-ready/crc/pkg/crc/constants.bundleEmbedded=true" -o out/windows-amd64/crc.exe ./cmd/crc
/home/prkumar/work/go/bin/binappend-cli write out/linux-amd64/crc /tmp/bundle/crc_libvirt_4.1.14.crcbundle
/home/prkumar/work/go/bin/binappend-cli write out/macos-amd64/crc /tmp/bundle/crc_hyperkit_4.1.14.crcbundle
/home/prkumar/work/go/bin/binappend-cli write out/windows-amd64/crc.exe /tmp/bundle/crc_hyperv_4.1.14.crcbundle
podman run -v /home/prkumar/work/github/crc/docs:/docs:Z --rm quay.io/crcont/asciidoctor-pdf:latest -D /docs/build -o doc.pdf /docs/source/getting-started/master.adoc
mkdir release
tar cJSf release/crc-macos-amd64.tar.xz -C out crc-macos-1.0.0-dev-amd64
tar cJSf release/crc-linux-amd64.tar.xz -C out crc-linux-1.0.0-dev-amd64
cd out && zip -r /home/prkumar/work/github/crc/release/crc-windows-amd64.zip crc-windows-1.0.0-dev-amd64
  adding: crc-windows-1.0.0-dev-amd64/ (stored 0%)
  adding: crc-windows-1.0.0-dev-amd64/crc.exe (deflated 61%)
  adding: crc-windows-1.0.0-dev-amd64/LICENSE (deflated 65%)
  adding: crc-windows-1.0.0-dev-amd64/doc.pdf (deflated 68%)
pushd release && sha256sum * > sha256sum.txt && popd
~/work/github/crc/release ~/work/github/crc
~/work/github/crc
gen_release_info
make: gen_release_info: Command not found
make: *** [Makefile:180: release] Error 127
```

This should be like following.
```patch
$ git diff
diff --git a/Makefile b/Makefile
index 9904371..b8cb516 100644
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ REPOPATH ?= $(ORG)/crc
 PACKAGES := go list ./... | grep -v /out
 SOURCES := $(shell git ls-files  *.go ":^vendor")
 
-RELEASE_INFO := $(RELEASE_DIR)/release-info.json
+RELEASE_INFO := release-info.json
 
 # Check that given variables are set and all have non-empty values,
 # die with an error otherwise.
@@ -163,7 +163,7 @@ gen_release_info:
        @sed -i s/@OPENSHIFT_VERSION@/\"$(BUNDLE_VERSION)\"/ $(RELEASE_INFO)
 
 .PHONY: release
-release: fmtcheck embed_bundle build_docs_pdf
+release: fmtcheck embed_bundle build_docs_pdf gen_release_info
        mkdir $(RELEASE_DIR)
        
        @mkdir -p $(BUILD_DIR)/crc-macos-$(CRC_VERSION)-amd64
@@ -180,7 +180,7 @@ release: fmtcheck embed_bundle build_docs_pdf
 
        pushd $(RELEASE_DIR) && sha256sum * > sha256sum.txt && popd
 
-       gen_release_info
+       @mv $(RELEASE_INFO) $(RELEASE_DIR)/$(RELEASE_INFO)
 
 HYPERKIT_BUNDLENAME = $(BUNDLE_DIR)/crc_hyperkit_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
 HYPERV_BUNDLENAME = $(BUNDLE_DIR)/crc_hyperv_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
```

cmd/crc/cmd/config/config.go
```diff
@@ -19,6 +19,8 @@ var (
 	NameServer     = cfg.AddSetting("nameserver", nil, []cfg.ValidationFnType{cfg.ValidateIpAddress}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	PullSecretFile = cfg.AddSetting("pull-secret-file", nil, []cfg.ValidationFnType{cfg.ValidatePath}, []cfg.SetFn{cfg.SuccessfullyApplied})
 
+	NoUpdateCheck = cfg.AddSetting("no-update-check", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
```
gbraad: `disable` probably sounds better
since this allows to also use `enable` as lingo.

there is no good opposite in this context for `no`; eg. `yes-telemetry`

cmd/crc/cmd/start.go
```diff
@@ -133,3 +137,19 @@ func getPullSecretFileContent() (string, error) {
...
+	if noUpdateCheck {
+		return
+	}
+	newVersionAvailable, newVersion, err := version.NewVersionAvailable()
+	if err != nil {
+		logging.Debugf("Error checking if a new version is available: %v", err)
+		return
+	}
+	if newVersionAvailable {
+		logging.Warnf("A new version (%s) has been published. Please download it from %s", newVersion, constants.CrcLandingPageURL)
```
praveenkumar: Should we put a `warn` message or a `info` message. I was under the impression that we agreed to use `info` here. 

Also how about message is like `A new version (1.0.1-dev) has been published on https://cloud.redhat.com/openshift/install/crc/installer-provisioned` so in the future when we have some kind of upgrade mechanism we can also add use `crc update` in the message without making it too big ?

cmd/crc/cmd/start.go
```diff
@@ -44,6 +45,8 @@ func runStart(arguments []string) {
 		errors.Exit(1)
 	}
 
+	checkIfNewVersionAvailable(crcConfig.GetBool(config.NoUpdateCheck.Name))
```
gbraad: ditto

cmd/crc/cmd/start.go
```diff
@@ -72,10 +75,11 @@ func initStartCmdFlagSet() *pflag.FlagSet {
 	flagSet := pflag.NewFlagSet("start", pflag.ExitOnError)
 	flagSet.StringP(config.Bundle.Name, "b", constants.DefaultBundlePath, "The system bundle used for deployment of the OpenShift cluster.")
 	flagSet.StringP(config.VMDriver.Name, "d", machine.DefaultDriver.Driver, fmt.Sprintf("The driver to use for the CRC VM. Possible values: %v", machine.SupportedDriverValues()))
-	flagSet.StringP(config.PullSecretFile.Name, "p", "", fmt.Sprintf("File path of Image pull secret for User (Download it from %s)", constants.DefaultPullSecretURL))
+	flagSet.StringP(config.PullSecretFile.Name, "p", "", fmt.Sprintf("File path of Image pull secret for User (Download it from %s)", constants.CrcLandingPageURL))
```
gbraad: this is not directly related to this PR, right?

cmd/crc/cmd/start.go
```diff
@@ -72,10 +75,11 @@ func initStartCmdFlagSet() *pflag.FlagSet {
...
 	flagSet.IntP(config.CPUs.Name, "c", constants.DefaultCPUs, "Number of CPU cores to allocate to the CRC VM")
 	flagSet.IntP(config.Memory.Name, "m", constants.DefaultMemory, "MiB of Memory to allocate to the CRC VM")
 	flagSet.StringP(config.NameServer.Name, "n", "", "Specify nameserver to use for the instance. (i.e. 8.8.8.8)")
+	flagSet.Bool(config.NoUpdateCheck.Name, false, "Don't check for update")
```
gbraad: ditto... Disable update check

cmd/crc/cmd/start.go
```diff
@@ -113,7 +117,7 @@ func getPullSecretFileContent() (string, error) {
 
 	// In case user doesn't provide a file in start command or in config then ask for it.
 	if crcConfig.GetString(config.PullSecretFile.Name) == "" {
-		pullsecret, err = input.PromptUserForSecret("Image pull secret", fmt.Sprintf("Copy it from %s", constants.DefaultPullSecretURL))
+		pullsecret, err = input.PromptUserForSecret("Image pull secret", fmt.Sprintf("Copy it from %s", constants.CrcLandingPageURL))
```
gbraad: this is related to another PR I believe. this is not about the release info

cmd/crc/cmd/start.go
```diff
@@ -133,3 +137,19 @@ func getPullSecretFileContent() (string, error) {
 
 	return pullsecret, nil
 }
+
+func checkIfNewVersionAvailable(noUpdateCheck bool) {
+	if noUpdateCheck {
+		return
+	}
+	newVersionAvailable, newVersion, err := version.NewVersionAvailable()
```
gbraad: `isNewVersionAvailable`, else it is too close to the variable named `newVersion` which likely contains a string.

pkg/crc/version/version.go
```diff
@@ -39,3 +58,40 @@ func GetCommitSha() string {
...
+	response, err := client.Get(releaseInfoLink)
+	if err != nil {
+		return nil, err
+	}
+	defer response.Body.Close()
+	releaseMetaData, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal(releaseMetaData, &releaseInfo)
```
gbraad: I still do not believe this is more concise than using a simple text file and a regex, as you still need boilerplate code to make this work.

I believe the double nature is lost here...

cmd/crc/cmd/start.go
cfergeau: The URL we want to point people at to find new releases is `constants.DefaultPullSecretURL` (ie same place as where the pull secret is). Since we are using it for more than "URL where the pull secret is" in this patch series, the renaming makes sense here. The renaming could be done in a separate preliminary commit, but so far we did not push much for such fine grained commits.

cmd/crc/cmd/config/config.go
cfergeau: The opposite of `no` is to just drop it, `--update-check` `--no-update-check`. But I almost suggested using disable/enable before, so not objecting to your suggestion ;)

pkg/crc/version/version.go
praveenkumar: @gbraad what you mean by `double nature is lost here` ? To unmarshal a json content that is always required right?

cmd/crc/cmd/start.go
anjannath: This is later used in the if statement and `if newVersionAvailable` sounds more natural than `if isNewVersionAvailable`.

Makefile
```diff
@@ -172,6 +180,8 @@ release: fmtcheck embed_bundle build_docs_pdf
 
 	pushd $(RELEASE_DIR) && sha256sum * > sha256sum.txt && popd
 
+	@mv $(RELEASE_INFO) $(RELEASE_DIR)/$(RELEASE_INFO)
```
praveenkumar: @anjannath can you move this to line number 180 so that we can create the shasum for this file also?

Makefile
```diff
@@ -170,6 +178,8 @@ release: fmtcheck embed_bundle build_docs_pdf
 	@cp LICENSE $(DOCS_BUILD_DIR)/doc.pdf $(BUILD_DIR)/windows-amd64/crc.exe $(BUILD_DIR)/crc-windows-$(CRC_VERSION)-amd64
 	cd $(BUILD_DIR) && zip -r $(CURDIR)/$(RELEASE_DIR)/crc-windows-amd64.zip crc-windows-$(CRC_VERSION)-amd64
 
+	@mv $(RELEASE_INFO) $(RELEASE_DIR)/$(RELEASE_INFO)
```
praveenkumar: :+1:

praveenkumar:
retest this please.

Issue #484, opened 2019-08-28
Introduce update mechanism


gbraad:
- [ ] auto-update ?
   * configuration to disable
- [ ] report update is available
   * warn user of available bundle
   * handle actions to be taken to 'update'
- [ ] no upgrade of instances
- [ ] interpret info from `release.txt`
   * https://mirror.openshift.com/pub/openshift-v4/clients/crc/latest/release.txt

anjannath:
I am going to work on the last task in the list:
* interpret info from release.txt

gbraad:
Any progress and changes to this?

gbraad:
Ref: #1860

PR #620, opened 2019-09-20
Issue #484 Check the latest version of crc on mirror.openshift.com and inform user if a new version is available

anjannath:
Ask the user if wants to continue using the current version or not
when a new version is available on mirror.

Depends on #619


anjannath:
Commit added to the branch in #619

Issue #618, opened 2019-09-20
[BUG]macOS FATA /etc/hosts is not readable/writable by the current user


jr00n:
### General information

  * OS: macOS
  * Hypervisor: hyperkit
  * Did you run `crc setup` before starting it: Yes

## CRC version
```bash
version: 1.0.0-beta.5+f2aa58c
```
  
## CRC status
```bash
CRC VM:          Stopped
OpenShift:       Stopped
Disk Usage:      0B of 0B (Inside the CRC VM)
Cache Usage:     10.48GB
Cache Directory: /Users/jr00n/.crc/cache
```

## CRC config
```bash
output is empty
```

## Host Operating System
```bash
~ » sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G95
```

### Steps to reproduce

  1. rm -rf ~/.crc
  2. crc setup
  3. crc start
  4. 

### Expected
starting cluster

### Actual
FATA /etc/hosts is not readable/writable by the current user
stops

### Logs

[https://gist.github.com/jr00n/9c3bc02e1fe9fdbde042dc7d5b317cfa](https://gist.github.com/jr00n/9c3bc02e1fe9fdbde042dc7d5b317cfa)

praveenkumar:
@jr00n I did try same on my mac machine and didn't able to get the `/etc/hosts` file issue.

```
$ rm -fr ~/.crc
$ ./crc setup
INFO Checking if running as non-root              
INFO Caching oc binary                            
INFO Setting up virtualization with HyperKit      
INFO Will use root access: change ownership of /Users/prkumar/.crc/bin/hyperkit 
Password:
INFO Will use root access: set suid for /Users/prkumar/.crc/bin/hyperkit 
INFO Installing crc-machine-hyperkit              
INFO Will use root access: change ownership of /Users/prkumar/.crc/bin/crc-driver-hyperkit 
INFO Will use root access: set suid for /Users/prkumar/.crc/bin/crc-driver-hyperkit 
INFO Setting file permissions for /etc/resolver/testing 
INFO Setting file permissions for /etc/hosts      
INFO Unpacking bundle from the CRC binary         
Setup is complete, you can now run 'crc start' to start a CodeReady Containers instance

$ ./crc start
INFO Checking if running as non-root              
INFO Checking if oc binary is cached              
INFO Checking if HyperKit is installed            
INFO Checking if crc-driver-hyperkit is installed 
[...]

PR #621, opened 2019-09-20
Issue #618 Make /etc/hosts file in mac user writeable.

praveenkumar:
During the refactor to add gosec in our CI we changed the file
permission from `655` to `500` and this file should be write able
by the user so permission should be `0600`


praveenkumar:
ci error is not related.

Issue #605, opened 2019-09-17
crc start fails saying LIBVIRT_3.4.0 


SaravanaStorageNetwork:
### General information

  * OS: Linux / macOS / Windows  - Linux
  * Hypervisor: KVM / Hyper-V / VirtualBox / hyperkit   KVM
  * Did you run `crc setup` before starting it (Yes/No)?  Yes

## CRC version
```bash
# Put the output of `crc version`
```
  version: 1.0.0-beta.3-4.1.11+e97bf65

## Host Operating System
cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.3 (Maipo)

### Steps to reproduce
 
crc start fails with error saying 
FATA exit status 1 : /root/.crc/bin/crc-driver-libvirt: /lib64/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by /root/.crc/bin/crc-driver-libvirt) 

Is libvirt-3.4 a requirement ? 
I have libvirt installed but the version is not 3.4. 

The doc. needs an update saying libvirt needs to be 3.4 :
https://code-ready.github.io/crc/#required-software-packages_gsg

### Logs
crc start fails as below:

FATA exit status 1 : /root/.crc/bin/crc-driver-libvirt: /lib64/libvirt.so.0: version `LIBVIRT_3.4.0' not found (required by /root/.crc/bin/crc-driver-libvirt) 

You can start crc with `crc start --log-level debug` to collect logs.
Please consider posting this on http://gist.github.com/ and post the link in the issue.

cfergeau:
> ### General information
> 
>     * OS: Linux / macOS / Windows  - Linux
> 
>     * Hypervisor: KVM / Hyper-V / VirtualBox / hyperkit   KVM
> 
>     * Did you run `crc setup` before starting it (Yes/No)?  Yes
> 
> 
> ## CRC version
> 
> ```shell
> # Put the output of `crc version`
> ```
> 
> version: 1.0.0-beta.3-4.1.11+e97bf65
> ## Host Operating System
> 
> cat /etc/redhat-release
> Red Hat Enterprise Linux Server release 7.3 (Maipo)

We indeed expect a more recent RHEL release to be used  (7.5 or newer)

> The doc. needs an update saying libvirt needs to be 3.4 :
> https://code-ready.github.io/crc/#required-software-packages_gsg

Yes, we could have a preflight check about this as well.

gbraad:
We need to add a pre-flight check for this.

[...]

PR #630, opened 2019-09-23
Issue #605: Check libvirt version as part of setup too

cfergeau:
We currently only check libvirt version at "crc start" time, it's more
consistent to also check it at "crc setup" time, which is when we
install libvirt, and the crc-machine-driver binary.

https://github.com/code-ready/crc/issues/605

### DIFF
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -139,6 +139,12 @@ func SetupHost(vmDriver string) {
 		"Starting libvirt service",
 		config.GetBool(cmdConfig.WarnCheckLibvirtRunning.Name),
 	)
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_linux.go
```diff
@@ -139,6 +139,12 @@ func SetupHost(vmDriver string) {
 		"Starting libvirt service",
 		config.GetBool(cmdConfig.WarnCheckLibvirtRunning.Name),
 	)
+	preflightCheckAndFix(config.GetBool(cmdConfig.SkipCheckLibvirtVersionCheck.Name),
+		checkLibvirtVersion,
+		fixLibvirtVersion,
+		"Checking if a supported libvirt version is installed",
+		config.GetBool(cmdConfig.WarnCheckLibvirtVersionCheck.Name),
+	)
```
praveenkumar: :+1:

Issue #631, opened 2019-09-23
[BUG] - ERRO Error occurred: Certs have expired, they were valid till: $NOW


lletelier:
### General information

  * OS: Linux
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes)?

## CRC version
```bash
version: 1.0.0-beta.3-4.1.11+e97bf65

```
  
## CRC status
```bash
version: 1.0.0-beta.3-4.1.11+e97bf65
```

## CRC config
```bash
- bundle                             : /home/roto/.crc/crc_libvirt_4.1.11.crcbundle
- cpus                               : 4
- memory                             : 8192
- nameserver                         : 
- pull-secret-file                   : 
- skip-check-user-in-libvirt-group   : true
- vm-driver                          : libvirt
- warn-check-bundle-cached           : true

```

## Host Operating System
```bash
# PRETTY_NAME="Fedora 30 (Workstation Edition)"
```

### Steps to reproduce

  1. 
```bash 
crc stop
CodeReady Containers instance stopped
```
  2. 
```bash 
crc setup
INFO Checking if NetworkManager is installed      
INFO Checking if NetworkManager service is running 
INFO Caching oc binary                            
INFO Setting up virtualization                    
INFO Setting up KVM                               
INFO Installing libvirt service and dependencies  
INFO Adding user to libvirt group                 
WARN Skipping above check ...                     
INFO Enabling libvirt                             
INFO Starting libvirt service                     
INFO Installing crc-driver-libvirt                
INFO Removing older system-wide crc-driver-libvirt 
INFO Setting up libvirt 'crc' network             
INFO Starting libvirt 'crc' network               
INFO Writing Network Manager config for crc       
INFO Writing dnsmasq config for crc               
INFO Unpacking bundle from the CRC binary    
```
  3. 
```bash 
crc start
INFO Checking if NetworkManager is installed      
INFO Checking if NetworkManager service is running 
INFO Checking if oc binary is cached              
INFO Checking if Virtualization is enabled        
INFO Checking if KVM is enabled                   
INFO Checking if libvirt is installed             
INFO Checking if user is part of libvirt group    
WARN Skipping above check ...                     
INFO Checking if libvirt is enabled               
INFO Checking if libvirt daemon is running        
INFO Checking if crc-driver-libvirt is installed  
INFO Checking if libvirt 'crc' network is available 
INFO Checking if libvirt 'crc' network is active  
[...]

PR #635, opened 2019-09-24
Issue #631 Add a note about cert expire time limit to readme.

praveenkumar:
We have detail about it in the doc troubleshoot section but not
in the readme, better to have it here also.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -4,6 +4,20 @@
 :toc-title:
 :toclevels:
 
+[NOTE]
+====
+Each released `crc` binary expires 30 days after the release due to certificates embedded in the Openshift cluster.
+Using an older `crc` binary will result in an expired certificates error.
+
+To resolve this error:
+
+* link:https://github.com/code-ready/crc/releases[Download the latest CodeReady Containers release] and place the `crc` binary in your `$PATH`.
+* Remove the cluster with certificate errors using the `crc delete` command.
+* Set up the new release with the `crc setup` command.
+* Start the new cluster with the `crc start` command.
+
+====
+
 toc::[]
 
 image:https://travis-ci.org/code-ready/crc.svg?branch=master["Travis CI Build Status", link="https://travis-ci.org/code-ready/crc"]


README.adoc
```diff
@@ -4,6 +4,13 @@
 :toc-title:
 :toclevels:
 
+[NOTE]
+====
+Each released `crc` binary expires 30 days after the release due to certificates embedded in the system bundle.
+Using an older `crc` binary will result in an expired certificates error.
+link:https://github.com/code-ready/crc/releases[Download the latest CodeReady Containers release] to resolve this error.
```
cfergeau: system bundle -> virtual machine image/OpenShift cluster?
We should explain what to do after downloading the latest release (crc delete, which will lose all changes made to the cluster, crc setup, crc start).
The explanation here is much clearer than the one in proc_basic-troubleshooting.adoc so I'd copy it there too.

README.adoc
```diff
@@ -4,6 +4,21 @@
 :toc-title:
 :toclevels:
 
+[NOTE]
+====
+Each released `crc` binary expires 30 days after the release due to certificates embedded in the Openshift cluster.
+Using an older `crc` binary will result in an expired certificates error.
+Following steps are needed to resolve this error.
```
robin-owen:
```adoc
To resolve this error:

* link:https://github.com/code-ready/crc/releases[Download the latest CodeReady Containers release] and place the `crc` binary in your `$PATH`.
* Remove the cluster with certificate errors using the `crc delete` command.
* Set up the new release with the `crc setup` command.
* Start the new cluster with the `crc start` command.
```

I'm not sure that we have to be so specific regarding "and place the `crc` binary in your `$PATH`", but the rest of the suggestion still stands. :)

README.adoc
```diff
@@ -4,6 +4,20 @@
...
+Using an older `crc` binary will result in an expired certificates error.
+
+To resolve this error:
+
+* link:https://github.com/code-ready/crc/releases[Download the latest CodeReady Containers release] and place the `crc` binary in your `$PATH`.
+* Remove the cluster with certificate errors using the `crc delete` command.
+* Set up the new release with the `crc setup` command.
+* Start the new cluster with the `crc start` command.
+
+====
```
cfergeau: Something similar could be added to the documentation too (maybe in https://github.com/code-ready/crc/blob/master/docs/source/getting-started/content/proc_basic-troubleshooting.adoc ?)

robin-owen: Absolutely, but we can do that in another PR. :)

praveenkumar: Yes I will put a PR for that doc update today.

Issue #644, opened 2019-09-26
[Integration test] `crc stop -f` test should be more generic


praveenkumar:
Right now `crc stop -f` can provide `CodeReady Containers instance stopped` if instance is stopped in required time limit or `CodeReady Containers instance forcefully stopped`if the instance doesn't stop during the time limit. 

Our integration test only check `CodeReady Containers instance stopped` and it make it flaky. We should put a regex in place of absolute string.

- https://regex101.com/r/UkT6En/1

jsliacan:
@praveenkumar, the PR that fixes this issue only changes the instances in the `basic.feature` file. There are some more in the other feature files. Should I submit another PR against this issue or create another issue? Sorry that I didn't get to this on time (before merging).

praveenkumar:
@jsliacan Thanks, I will check on other features and put same and then only close this issue, next time I will remember to add one of you folks as part of the review for any integration PR.

PR #645, opened 2019-09-26
Issue #644 Add regex to crc stop -f integration test

praveenkumar:
`crc stop -f` can provide `CodeReady Containers instance stopped` or
 `CodeReady Containers instance forcefully stopped` if instance is
stopped gracefully or not. Better to use regex then absolute string.

- https://regex101.com/r/UkT6En/1

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -98,7 +98,7 @@ Feature: Basic test
     @darwin @linux @windows
     Scenario: CRC forcible stop
         When executing "crc stop -f"
-        Then stdout should contain "CodeReady Containers instance stopped"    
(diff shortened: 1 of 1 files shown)


test/integration/features/basic.feature
```diff
@@ -98,7 +98,7 @@ Feature: Basic test
     @darwin @linux @windows
     Scenario: CRC forcible stop
         When executing "crc stop -f"
-        Then stdout should contain "CodeReady Containers instance stopped"    
+        Then stdout should match "CodeReady Containers instance(.*)stopped"
```
anjannath: :+1:

Issue #643, opened 2019-09-26
Update the troubleshooting docs about cert error


praveenkumar:
This is follow up for https://github.com/code-ready/crc/pull/635#discussion_r328200593 to update the doc section for troubleshooting docs also.

cfergeau:
I believe this is merged now.

PR #646, opened 2019-09-26
Issue #643 Update troubleshooting doc for vm cert expire case.

praveenkumar:


praveenkumar:
retest this please.

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -60,7 +60,7 @@ If your issue has not been resolved by this procedure, perform the following ste
 . If no existing issue addresses the encountered issue, link:https://github.com/code-ready/crc/issues/new[create an issue] and link:https://help.github.com/en/articles/file-attachments-on-issues-and-pull-requests[attach the [filename]`~/.crc/crc.log` file] to the created issue.
 The [filename]`~/.crc/crc.log` file contains detailed debugging and troubleshooting information which can help diagnose the problem that you are experiencing.
 
-= DNS configuration details
+== DNS configuration details
```
cfergeau: "editor"? Which editor?
I put it on purpose at the same level as "Basic Troubleshooting" and "Detailed Troubleshooting". If you want to move in inside "Detailed Troubleshooting", you need to change the "==" to "===" for "linux" and "mac" below I think.
But since there are already 2 sections starting with "=", I don't see why a 3rd one would trigger a warning?

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -1,6 +1,20 @@
...
+
+To resolve this error:
+
+* link:https://github.com/code-ready/crc/releases[Download the latest CodeReady Containers release] and place the `crc` binary in your `$PATH`.
+* Remove the cluster with certificate errors using the `crc delete` command.
+* Set up the new release with the `crc setup` command.
+* Start the new cluster with the `crc start` command.
+
+====
+
```
cfergeau: Not sure whether this should be a "NOTE", or its own bullet point.

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -1,6 +1,20 @@
...
 
+[NOTE]
+====
+Each released `crc` binary expires 30 days after the release due to certificates embedded in the Openshift cluster.
+Using an older `crc` binary will result in an expired certificates error.
+
+To resolve this error:
+
+* link:https://github.com/code-ready/crc/releases[Download the latest CodeReady Containers release] and place the `crc` binary in your `$PATH`.
+* Remove the cluster with certificate errors using the `crc delete` command.
```
cfergeau: I'd emphasize this will cause data loss

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -1,6 +1,20 @@
...
+* Set up the new release with the `crc setup` command.
+* Start the new cluster with the `crc start` command.
+
+====
+
 The majority of issues can be resolved by stopping a running {prod} virtual machine, deleting the virtual machine, and starting a new instance of the virtual machine.
```
cfergeau: These are the same steps as the ones you detailed just above, we could turn what you added in generic instructions that would go there, and then we would refer to these steps  inthe "certificate recovery" description?

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
gbraad: note would make more sense as it stands out.

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
praveenkumar: @cfergeau I am using goland with `adoc` plugin enabled and it was throwing that as warning.

robin-owen: AFAIK, the heading levels should be absolutely fine as they were. We offset the levels when assembling this file into the master assembly, so essentially, all top-level headings here will end up as second-level headings. :)

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
robin-owen: I'd say that a note admonition makes sense to highlight the data loss aspect, but having all of this information in an admonition block is not necessary. In fact, I think it would be better as its own `proc_<name>.adoc` procedure file which we can add to the `assembly_troubleshooting.adoc` file.

@praveenkumar, do you mind if I send you a PR to this branch to show you what I mean? I feel like that would be easier than typing it all out. :)

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
praveenkumar: @robin-owen Ack I will remove this commit then and ignore the plugin warning :)

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
praveenkumar: @robin-owen please go ahead, I will close this PR in favor of yours :)

praveenkumar:
close in favor of #665

PR #626, opened 2019-09-23
Misc cmdline interface and doc improvement around `crc console`

cfergeau:
This PR discourages direct use of the openshift web console URL in favour of usage of `crc console`.

PR #649, opened 2019-09-26
Fix some small naming inconsistencies in variable names

cfergeau:
This is a followup on https://github.com/code-ready/crc/pull/626


cmd/crc/cmd/config/unset.go
```diff
@@ -26,15 +26,15 @@ to the options that you set when you run the 'crc start' command.`,
 }
 
 func runConfigUnset(key string) {
-	unsetMsg, err := config.Unset(key)
+	unsetMessage, err := config.Unset(key)
```
praveenkumar: :+1:

PR #650, opened 2019-09-26
Assorted fixes for preflight checks

cfergeau:
This PR removed unwanted warning when using a binary without an embedded bundle, and add missing preflight config options on Windows.

### DIFF
--- a/cmd/crc/cmd/config/config_darwin.go
+++ b/cmd/crc/cmd/config/config_darwin.go
@@ -6,6 +6,8 @@ import (
 
 var (
 	// Preflight checks
+	SkipCheckVirtualBoxInstalled     = cfg.AddSetting("skip-check-virtualbox-installed", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	WarnCheckVirtualBoxInstalled     = cfg.AddSetting("warn-check-virtualbox-installed", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	SkipCheckResolverFilePermissions = cfg.AddSetting("skip-check-resolver-file-permissions", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	WarnCheckResolverFilePermissions = cfg.AddSetting("warn-check-resolver-file-permissions", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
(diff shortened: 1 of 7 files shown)


cmd/crc/cmd/config/config_darwin.go
```diff
@@ -6,6 +6,8 @@ import (
 
 var (
 	// Preflight checks
+	SkipCheckVirtualBoxInstalled     = cfg.AddSetting("skip-check-virtualbox-installed", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	WarnCheckVirtualBoxInstalled     = cfg.AddSetting("warn-check-virtualbox-installed", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
```
praveenkumar: @gbraad @anjannath is it true that we are completely not supporting VB on windows. iirc if a user can setup his dns server then he can still use VB since cluster should be working internally?

anjannath: I haven't tried VB in windows, but i remember @gbraad mentioning there were some issues with the networking interface. So yes atm it is not supported!

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -68,7 +68,8 @@ func fixHyperVInstalled() (bool, error) {
 	}
 
 	// We do need to error out as a restart might be needed (unfortunately no output redirect possible)
-	return true, errors.New("Please reboot your system")
+	logging.Error("Please reboot your system")
```
anjannath: Did you mean to stop the execution here? if so, this won't stop the execution, for that we need `logging.Fatal`.

cfergeau: I wanted to print a message without stopping execution (since errors.New is printing the message and I did not want to lose that).

cmd/crc/cmd/config/config_darwin.go
cfergeau: Removing these options from the windows build does not mean we are not supporting vbox on Windows. It only acknowledges the fact that we have no preflight checks/fixes for vbox on windows, so these config flags are not doing anything at the moment. If/when we add vbox preflight checks on Windows, then we'll readd these config flags.

gbraad: the issue for vbox is worked around for `dns_windows.go` when vbox is used

pkg/crc/preflight/preflight_checks_windows.go
gbraad: @cfergeau why not stop? The rest of the execution can never succeed properly without this reboot

gbraad: `fixHyperVInstalled` does not need a forced reboot, as the user add can happen. But the user is only member of the group AFTER the reboot, so the `fixUserPartOfHyperVAdministrators` should stop.

PR #651, opened 2019-09-26
Issue #507 Add cleanup feature as part of command set

praveenkumar:


pkg/os/util_linux.go
```diff
@@ -20,3 +20,9 @@ func WriteToFileAsRoot(reason, content, filepath string) error {
 	}
 	return nil
 }
+
+func RemoveAFileAsRoot(reason, filepath string) error {
+	logging.Infof("Will use root access: %s", reason)
+	cmd := exec.Command("sudo", "rm", "-fr", filepath) // #nosec G204
```
cfergeau: This should use RunWithPrivilege (and same for WriteToFileAsRoot actually)

pkg/os/util_linux.go
```diff
@@ -20,3 +20,9 @@ func WriteToFileAsRoot(reason, content, filepath string) error {
 	}
 	return nil
 }
+
+func RemoveAFileAsRoot(reason, filepath string) error {
```
cfergeau: `RemoveFileAsRoot`

pkg/crc/cleanup/cleanup_linux.go
```diff
@@ -0,0 +1,69 @@
...
+		return fmt.Errorf("%+v: %s", err, stdErr)
+	}
+	_, stdErr, err = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-undefine", libvirt.DefaultNetwork)
+	if err != nil {
+		return fmt.Errorf("%+v: %s", err, stdErr)
+	}
+	return nil
+}
+
+func removeNetworkManagerConfigForDNSMasq() error {
```
cfergeau: I think it would be good to keep naming consistent with the preflight checks, so `cleanupCrcNetworkManagerConfig`, `cleanupCrcDnsmasqConfigFile` and so on. Actually, I'd just add a 'cleanup' field to the PreflightCheck structure in https://github.com/cfergeau/crc/blob/preflight-refactor/pkg/crc/preflight/preflight.go , the cleanup command can be generic and only need to iterate over these.

cmd/crc/cmd/cleanup.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/code-ready/crc/pkg/crc/constants"
+)
+
+func init() {
+	rootCmd.AddCommand(cleanupCmd)
+}
+
+var cleanupCmd = &cobra.Command{
+	Use:   "cleanup",
+	Short: fmt.Sprintf("Undo config changes and remove %s dir", constants.MachineBaseDir),
```
gbraad: off-topic, but also on-topic; will we have an `uninstall` command?

cmd/crc/cmd/cleanup.go
```diff
@@ -0,0 +1,33 @@
...
+		runCleanup(args)
+	},
+}
+
+func runCleanup(arguments []string) {
+	err := cleanup.CleanupHost()
+	if err != nil {
+		errors.ExitWithMessage(1, err.Error())
+	}
+	output.Outln("Cleanup is complete")
```
gbraad: Cleanup finished

pkg/crc/cleanup/cleanup_linux.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/machine/libvirt"
+	"github.com/code-ready/crc/pkg/crc/preflight"
+	crcos "github.com/code-ready/crc/pkg/os"
+)
+
+func CleanupHost() error {
+	err := errors.MultiError{}
+	// Remove crc network from libvirt
+	err.Collect(removeCRCLibvirtNetwork())
```
gbraad: this sounds more like an uninstall

praveenkumar:
retest this please.

praveenkumar:
retest this please

pkg/crc/cleanup/cleanup_linux.go
```diff
@@ -0,0 +1,69 @@
...
+	err.Collect(cleanupDnsmasqConfigForCrc())
+
+	// Remove the crc directory
+	err.Collect(cleanupCrcBaseDir())
+
+	return err.ToError()
+}
+
+func cleanupCrcLibvirtNetwork() error {
+	stdOut, stdErr, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-list")
```
cfergeau: net-info/net-uuid would be a better way to test than virsh net-list | grep crc

pkg/crc/cleanup/cleanup_linux.go
```diff
@@ -0,0 +1,69 @@
...
+	err.Collect(cleanupCrcLibvirtNetwork())
+
+	// Removing NetworkManager dnsmasq config for crc
+	err.Collect(cleanupNetworkManagerConfigForDNSMasq())
+
+	// Removing dnsmasq config for crc
+	err.Collect(cleanupDnsmasqConfigForCrc())
+
+	// Remove the crc directory
+	err.Collect(cleanupCrcBaseDir())
```
cfergeau: Is there anything removing the crc libvirt VM?

praveenkumar:
retest this please

pkg/crc/preflight/preflight.go
```diff
@@ -128,6 +135,18 @@ func doFixPreflightChecks(checks []PreflightCheck) {
 	}
 }
 
+func doCleanUpPreflightChecks(checks []PreflightCheck) {
+	for _, check := range checks {
```
cfergeau, line 139: iterating in reverse order would be safer to ensure we undo the changes in the right order.

pkg/crc/preflight/preflight_windows.go
```diff
@@ -68,3 +68,7 @@ func SetupHost() {
 func RegisterSettings() {
 	doRegisterSettings(getPreflightChecks())
 }
+
+func CleanUpHost() {
+	doCleanUpPreflightChecks(getPreflightChecks())
```
cfergeau, line 73: Should this be erroring out until it's implemented on Windows ?

pkg/crc/preflight/preflight.go
```diff
@@ -31,6 +33,7 @@ type PreflightCheck struct {
 	fixDescription   string
 	fix              PreflightFixFunc
 	flags            PreflightCheckFlags
+	cleanup          PreflightCleanUpFunc
```
cfergeau, line 36: no `cleanupDescription` so that we can let the user know what's being done?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -364,6 +364,57 @@ func fixLibvirtCrcNetworkAvailable() error {
...
 }
 
+func removeLibvirtCrcNetwork() error {
+	logging.Debug("Removing libvirt 'crc' network")
+	_, _, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-info", libvirt.DefaultNetwork)
+	if err == nil {
+		cmd := exec.Command("virsh", "--connect", "qemu:///system", "net-destroy", libvirt.DefaultNetwork)
+		err = cmd.Run()
+		buf := new(bytes.Buffer)
+		cmd.Stderr = buf
```
cfergeau, line 374: why not `crcos.RunWithDefaultLocale`?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -364,6 +364,57 @@ func fixLibvirtCrcNetworkAvailable() error {
...
+			return fmt.Errorf("Failed to undefine libvirt 'crc' network")
+		}
+	}
+	logging.Debug("libvirt 'crc' network removed")
+	return nil
+}
+
+func removeCrcVM() error {
+	logging.Debug("Removing 'crc' VM")
+	_, _, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "dumpxml", constants.DefaultName)
```
cfergeau, line 395: `dominfo` rather than `dumpxml`?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -364,6 +364,57 @@ func fixLibvirtCrcNetworkAvailable() error {
...
+	if err == nil {
+		cmd := exec.Command("virsh", "--connect", "qemu:///system", "destroy", constants.DefaultName)
+		err := cmd.Run()
+		buf := new(bytes.Buffer)
+		cmd.Stderr = buf
+		if err != nil {
```
cfergeau, line 401: same comment about `crcos.RunWithDefaultLocale`

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -364,6 +364,57 @@ func fixLibvirtCrcNetworkAvailable() error {
...
+		cmd = exec.Command("virsh", "--connect", "qemu:///system", "undefine", constants.DefaultName)
+		err = cmd.Run()
+		buf = new(bytes.Buffer)
+		cmd.Stderr = buf
+		if err != nil {
+			logging.Debugf("%v : %s", err, buf.String())
+			return fmt.Errorf("Failed to undefine 'crc' VM")
+		}
+	}
+	logging.Debug("'crc' vm is removed")
```
cfergeau, line 414: `vm` or `VM` ?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -474,6 +525,25 @@ func fixCrcDnsmasqConfigFile() error {
...
 }
 
+func removeCrcDnsmasqConfigFile() error {
+	logging.Debug("Removing dnsmasq configuration")
+	err := crcos.RemoveFileAsRoot(
+		fmt.Sprintf("removing dnsmasq configuration in %s", crcDnsmasqConfigPath),
+		crcDnsmasqConfigPath,
+	)
+	if err != nil {
+		return fmt.Errorf("Failed to write dnsmasq config file: %s: %v", crcDnsmasqConfigPath, err)
```
cfergeau, line 535: `s/write/remove/`

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -113,3 +113,7 @@ func SetupHost() {
 func RegisterSettings() {
 	doRegisterSettings(getPreflightChecks())
 }
+
+func CleanUpHost() {
+	doCleanUpPreflightChecks(getPreflightChecks())
```
cfergeau, line 118: Should this be erroring out until it's implemented on macOS ?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -364,6 +364,57 @@ func fixLibvirtCrcNetworkAvailable() error {
...
+		buf = new(bytes.Buffer)
+		cmd.Stderr = buf
+		if err != nil {
+			logging.Debugf("%v : %s", err, buf.String())
+			return fmt.Errorf("Failed to undefine 'crc' VM")
+		}
+	}
+	logging.Debug("'crc' vm is removed")
+	return nil
+}
```
cfergeau, line 416: I think we also want to remove `~/.crc/machines/crc` after destroying the VM. And I wonder if we should also removed the cached bundle when it's embedded in our binary? As crc setup will extract it.

pkg/crc/preflight/preflight_windows.go
praveenkumar: @cfergeau Sure, I can add if that make sense but we need to add it soon after initial stuff merges.

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau yes, we are going to remove `~/.crc/machines/crc` but I am not sure if we should also remove the cache bits because then we are again adding extraction time to user side if he is on the same version.

praveenkumar:
retest this please.

pkg/crc/preflight/preflight_windows.go
cfergeau: I would say it does not hurt to have it until it's implemented, this way we'll get an error if this slips. This should be a one-liner to add such an error, right?

pkg/crc/preflight/preflight_linux.go
```diff
@@ -67,11 +67,13 @@ var libvirtPreflightChecks = [...]PreflightCheck{
 		flags:            SetupOnly,
 	},
 	{
-		configKeySuffix:  "check-crc-network",
-		checkDescription: "Checking if libvirt 'crc' network is available",
-		check:            checkLibvirtCrcNetworkAvailable,
-		fixDescription:   "Setting up libvirt 'crc' network",
-		fix:              fixLibvirtCrcNetworkAvailable,
+		configKeySuffix:    "check-crc-network",
```
cfergeau, line 70: I'm not sure where to mention that the cleanup on linux is only partial for now? (no cleanup of libvirt group for example).

pkg/crc/preflight/preflight.go
```diff
@@ -92,9 +96,19 @@ func (check *PreflightCheck) doFix() error {
 	return check.fix()
 }
 
+func (check *PreflightCheck) doCleanUp() error {
+	if check.cleanupDescription == "" {
+		panic(fmt.Sprintf("Should not happen, empty description for cleanup '%s'", check.configKeySuffix))
+	}
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau for cleanup, we should not allow skip option otherwise the left over still present and user again report back to us.

praveenkumar:
> well, they asked to cleanup crc :) maybe add a --cache option later?

@cfergeau we already have that for `crc delete` option, we can add later same with `crc cleanup` command if that is something user wants.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -364,6 +364,49 @@ func fixLibvirtCrcNetworkAvailable() error {
...
+			return fmt.Errorf("Failed to destroy libvirt 'crc' network")
+		}
+
+		_, stderr, err = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-undefine", libvirt.DefaultNetwork)
+		if err != nil {
+			logging.Debugf("%v : %s", err, stderr)
+			return fmt.Errorf("Failed to undefine libvirt 'crc' network")
+		}
+	}
+	logging.Debug("libvirt 'crc' network removed")
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau ah yes that should've been in the first `if` block.

pkg/crc/preflight/preflight_linux.go
praveenkumar: How about putting that details in the documents?

praveenkumar:
> I would say it does not hurt to have it until it's implemented, this way we'll get an error if this slips. This should be a one-liner to add such an error, right?

@cfergeau I added a warning for windows and mac this way we can tell user if this slips.

pkg/crc/preflight/preflight.go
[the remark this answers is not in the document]
cfergeau: The problem is that we cannot know why the 'skip' config option was set. Maybe it's because they did not want crc to make any changes related to that option in the first place, so trying to cleanup these changes is wrong, and could even mess up the system, as we'd be changing something we did not setup in the first place.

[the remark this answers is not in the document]
praveenkumar: @cfergeau So as of now we are only changing something which CRC do as part of preflight checks even if those are skip and all those are very specific to `CRC` what do you think we are cleaning something which a user can set manually and as part of `crc cleanup` we are removing it ?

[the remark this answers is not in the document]
cfergeau: The comment above does not match what the code is doing, but more on this below

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -472,6 +515,26 @@ func fixCrcDnsmasqConfigFile() error {
 	return nil
 }
 
+func removeCrcDnsmasqConfigFile() error {
+	if _, err := os.Stat(crcDnsmasqConfigPath); !os.IsNotExist(err) {
```
cfergeau, line 519: Getting back to the discussion from above regarding handling of `skip-`, The error is ignored in case the user disabled dnsmasq?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -472,6 +515,26 @@ func fixCrcDnsmasqConfigFile() error {
...
+			crcDnsmasqConfigPath,
+		)
+		if err != nil {
+			return fmt.Errorf("Failed to remove dnsmasq config file: %s: %v", crcDnsmasqConfigPath, err)
+		}
+
+		logging.Debug("Reloading NetworkManager")
+		sd := systemd.NewHostSystemdCommander()
+		if _, err := sd.Reload("NetworkManager"); err != nil {
+			return fmt.Errorf("Failed to restart NetworkManager: %v", err)
```
cfergeau, line 532: This could fail when `skip-check-network-manager-installed` is set.

pkg/crc/preflight/preflight_linux.go
cfergeau: In my opinion, it's useful to have in the commit log too if someone looks at git history to try to understand why something is not part o f `cleanup`.

pkg/crc/preflight/preflight.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau we do check first if file is present but if you very strong about adding the skip functionality here I can add but I don't see any use for this in `cleanup` right now.

pkg/crc/preflight/preflight_linux.go
praveenkumar: @cfergeau sure I can add a line in the commit logs.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -362,6 +362,49 @@ func fixLibvirtCrcNetworkAvailable() error {
 	return nil
 }
 
+func removeLibvirtCrcNetwork() error {
+	logging.Debug("Removing libvirt 'crc' network")
+	_, _, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-info", libvirt.DefaultNetwork)
+	if err == nil {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau  you already answered, if the `crc` network not present their is an error which need to be ignored. I feel better using `err ==nil` instead catching it and then return `nil` :( 
```
if err != nil {
    // not sure why we ignore the error
    return nil
}
```

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau yes we first check if file is exist before doing any other operation to it.

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau we are adding the dnsmasqConfig file to the `/etc/NetworkManager` location so even setup fails if Network manager is not installed and `skip-check-network-manager-installed` is set to true.

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: But why? On a system where `crc setup` successfully ran, this file will be present, so why do we test if it's there, and ignore it if missing?

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: Ah it  fails because the `/etc/NetworkManager` directory does not exist. This does not seem like something intentional though, just luck ;)

pkg/crc/preflight/preflight_checks_linux.go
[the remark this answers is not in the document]
cfergeau: But why do we want to ignore missing `crc` network? In my opinion it's better to use standard go error handling, and add an explanation, rather than using a less usual construct, where people will probably not notice that we intentionnally ignore the error.

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau what is someone use `skip-check-crc-dnsmasq-file` during the setup as second run then this file will be there and `crc setup` will also successfully running, this is just ensuring remove of the file which we might create.

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau not luck, right now our dnsmasq configuration depends on NM so all those config only happen when NM is installed.

pkg/crc/preflight/preflight_checks_linux.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau To me as per go standard if there is an error then return it and if we are just returning `nil` then it just doesn't seems right but if you say we should have a logging debug info which says that there is no crc network then I might go with that something like?
```
if err != nil {
    logging.Debug("No crc network is found for libvirt") 
    return nil
}
```

pkg/crc/preflight/preflight.go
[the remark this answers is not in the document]
cfergeau: Right now, I'm trying to understand how you want this to work, I'm not yet feeling strongly about any options :)
Are you saying that the 'cleanup' methods should ignore as many errors as possible when doing their job, to cope with both the nominal case, and cases where some of the checks are disabled? Which would mean we don't need to check the config options, as the 'cleanup' method will be appropriate in both cases. Or do you have something different in mind?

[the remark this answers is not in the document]
praveenkumar: > Are you saying that the 'cleanup' methods should ignore as many errors as possible when doing their job, to cope with both the nominal case, and cases where some of the checks are disabled?

We will only ignore the error if something is not present, not just ignore errors which need to be processed, Yes I think `cleanup` shouldn't depend on the what changes a user `skip` as part of setup we never know if this is first time setup for the user or something he skipped recently and we are not removing that because of it is skipped in config.

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: I don't see anything in the code making sure `fixCrcNetworkManagerConfig` and `fixCrcDnsmasqConfigFile` are only run if NetworkManager is installed and running (in other word, they don't check if `checkNetworkManagerIsRunning` and `checkNetworkManagerInstalled` return nil). They indeed won't work if NetworkManager is not installed, but only because they will try to write in directories which don't exist. To me this does not need like something intentional, just a corner-case/side effect which happens to do the right thing in our case.

I'd prefer if we did not add more code relying on that assumption in order to work properly when it seems relatively easy to avoid it (ignoring the error seems like it's going to achieve this?)

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: Ok, so this goes back to a discussion I started somewhere else during this review. This is how you want to handle skipped checks. Which is probably fine, but needs to be explained.

pkg/crc/preflight/preflight_checks_linux.go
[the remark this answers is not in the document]
cfergeau: Not even debug, just a comment I guess for future readers of the code.
Per go standard, we need to handle the error after the call. "handle" does not necessarily mean to return it, but make a decision about it. We can decide to ignore it and return nil if we know the rest of the code will be able to deal with it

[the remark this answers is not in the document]
cfergeau: Using `if err == nil` can quickly get ugly if you need to do it several times
```
err := foo()
if err == nil {
    err = bar()
    if err == nil {
        err = foobar()
```

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau where you want it to be explained in commit log or as part of comment?

cfergeau: it probably fits better in the commit log as this is a general description of how things are expected to work, and why the code is written this way.

praveenkumar: Done.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -362,6 +362,54 @@ func fixLibvirtCrcNetworkAvailable() error {
 	return nil
 }
 
+func removeLibvirtCrcNetwork() error {
+	logging.Debug("Removing libvirt 'crc' network")
+	_, _, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-info", libvirt.DefaultNetwork)
+	if err != nil {
+		// Ignore since no crc network exists for libvirt
```
cfergeau, line 369: 'since' -> 'if', but why it's more interesting to say why we are ignoring it

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -362,6 +362,54 @@ func fixLibvirtCrcNetworkAvailable() error {
...
+		return nil
+	}
+	_, stderr, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-destroy", libvirt.DefaultNetwork)
+	if err != nil {
+		logging.Debugf("%v : %s", err, stderr)
```
cfergeau, line 374: We accept setups where the `crc` network does not exist (to be honest, I'm not sure how such a setup would work), but we don't accept setups where a `crc` network exists but was stopped manually? It's hard to know where to draw the line between "this could be a valid user setup" and "we are not going to support this" :(

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -362,6 +362,54 @@ func fixLibvirtCrcNetworkAvailable() error {
...
+	}
+	logging.Debug("libvirt 'crc' network removed")
+	return nil
+}
+
+func removeCrcVM() error {
+	logging.Debug("Removing 'crc' VM")
+	_, _, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "dominfo", constants.DefaultName)
+	if err != nil {
+		// Ignore since no crc vm exists
```
cfergeau, line 391: 'since' -> 'if', or 'ignore' -> 'return early', but probably more descriptive as `User may have run `crc  delete` before 'crc cleanup'`

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -472,6 +520,30 @@ func fixCrcDnsmasqConfigFile() error {
 	return nil
 }
 
+func removeCrcDnsmasqConfigFile() error {
+	if err := checkNetworkManagerInstalled(); err != nil {
+		// Since NetworkManager is not installed then this file will not present.
```
cfergeau, line 525: `When NetworkManager is not present, this file won't exist`

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -472,6 +520,30 @@ func fixCrcDnsmasqConfigFile() error {
...
+		return nil
+	}
+	if _, err := os.Stat(crcDnsmasqConfigPath); !os.IsNotExist(err) {
```
cfergeau, line 528: In my opinion, if we get an error, and it's not a "file does not exist" error, we should return it right away rather than trying to remove the dnsmasq config (I think this is what would happen for now?)
```
_, err := os.Stat(crcDnsmasqConfigPath)
if err != nil {
    if os.IsNotExist(err) {
        return nil
    } else {
        return err
    }
}
logging.Debug("Removing dnsmasq configuration")
````

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 26 above)
Comment (praveenkumar) [reply pkg/crc/preflight/preflight_checks_linux.go:374]: @cfergeau this check is only present if a user manually delete the `~./crc` or `~/.crc/machine` directory and don't want to use crc anymore so he can clean up those using `cleanup` command.

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 29 above)
Comment (praveenkumar) [reply pkg/crc/preflight/preflight_checks_linux.go:528]: @cfergeau As per go docs https://golang.org/pkg/os/#File.Stat is there is an error it is only the PathError https://golang.org/pkg/os/#PathError this logic is used a lot of place without any issue https://gist.github.com/mattes/d13e273314c3b3ade33f

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 25 above)
Comment (praveenkumar) [reply pkg/crc/preflight/preflight_checks_linux.go:369]: > why it's more interesting to say why we are ignoring it

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 29 above)
Comment (cfergeau) [reply pkg/crc/preflight/preflight_checks_linux.go:530]: Yes but `PathError` is not always a "file not  found" error as indicated in https://gist.github.com/mattes/d13e273314c3b3ade33f it could for example match `os.IsPermission()`. So we are ignoring some errors which are not "file not found" errors.

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 26 above)
Comment (cfergeau) [reply pkg/crc/preflight/preflight_checks_linux.go:375]: Removing `~/.crc` is not going to change anything with respect to libvirt network definitions.

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 25 above)
Comment (cfergeau) [reply pkg/crc/preflight/preflight_checks_linux.go]: Why can it happen? `no crc network, user may have removed it themselves, just ignore it`

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 26 above)
Comment (praveenkumar) [reply pkg/crc/preflight/preflight_checks_linux.go:375]: @cfergeau sorry, I didn't get what you are suggesting in term of destroying the crc network, do you mean we shouldn't destroy it and just undefine it? My understanding if you directly `undefine` the network without `destroy` then it will be still `active` but transient. 

```
$ sudo virsh net-list
 Name              State    Autostart   Persistent
----------------------------------------------------
 crc               active   yes         yes

$ sudo virsh net-undefine crc
Network crc has been undefined

$ sudo virsh net-list
 Name              State    Autostart   Persistent
----------------------------------------------------
 crc               active   no          no
```

File: pkg/crc/preflight/preflight_checks_linux.go (hunk 29 above)
Comment (praveenkumar) [reply pkg/crc/preflight/preflight_checks_linux.go:530]: @cfergeau  The whole check is only to know if file is present of not and then try to process it. `!os.IsNotExist(err)` make sure all the other errors which are part of `os` package shouldn't need to handle one by one just to know if file is present or not.

```golang
package main

import (
	"os"
	"fmt"
)

func main() {
	 _, err := os.Stat("/etc/libvirt/libvirt.conf")
	if !os.IsNotExist(err) {
		fmt.Println("File Exist")
	}
	fmt.Println(err)
}
```

Following program tells that File exist even it doesn't have required permission and that all we want since we are deleting the file with sudo anyway, the way we created as part of the setup?

```
$ go run test.go 
File Exist
stat /etc/libvirt/libvirt.conf: permission denied
```

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: You said 
> @cfergeau this check is only present if a user manually delete the ~./crc or ~/.crc/machine directory and don't want to use crc anymore so he can clean up those using cleanup command.

I was answering to that, removing `~/.crc` does not change anything with respect to libvirt `crc` network being there or not. I'm not asking for any change, but I don't understand why we need to check for `crc` network existence.

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: if `os.Stat` fails with an error which is not "file not found", maybe we'll manage to delete the file, maybe we won't, we cannot know. So I'm only asking if we want to keep going and try to delete anyway and report an error only if deletion fails? Or do we want to return early reporting the error? 
All I'm saying is that it's not clear at all if the code is trying on purpose to delete the file when an error happened, or if it's an oversight, so I'm asking about it.

Issue #658, opened 2019-09-27
Ask user before delete the VM and warn them about data lose


praveenkumar:
We should ask user promptly about deleting the crc vm when they execute `crc delete` and also warn them about data lose in the VM. there should be a `force` flag to enable scripting purpose.

praveenkumar:
fixed by #659

PR #660, opened 2019-09-27
Issue #658 Ask user confirmation before delete the VM

praveenkumar:
Also add force flag to make it easy for scripting

### DIFF
--- a/cmd/crc/cmd/delete.go
+++ b/cmd/crc/cmd/delete.go
@@ -37,12 +37,14 @@ func runDelete(arguments []string) {
 	if clearCache {
 		deleteCache()
 	}
-
-	_, err := machine.Delete(deleteConfig)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/crc/cmd/delete.go
```diff
@@ -37,12 +37,14 @@ func runDelete(arguments []string) {
 	if clearCache {
 		deleteCache()
 	}
-
-	_, err := machine.Delete(deleteConfig)
-	if err != nil {
-		errors.Exit(1)
+	yes := input.PromptUserForYesOrNo("Do you want to delete the crc VM, it will cause data lose in the VM", globalForce)
```
anjannath: I'd go with `Do you want to delete the crc VM (you'll lose all data in the VM)` in brackets, as `PromptUserForYesOrNo()` would add a question mark at the end of the sentence.

Issue #669, opened 2019-09-30
Add check if crc VM exist before running the complete command.


praveenkumar:
If the VM is not exist then asking for user permission to delete it or even showing that crc will stop doesn't make sense. We should first check about instance and then run next flow of command.

```
$ ./crc stop
Stopping CodeReady Containers instance... this may take a few minutes
ERRO Error occurred: Machine "crc" does not exist. Use "crc start" to add a new one. 

$ ./crc delete
Do you want to delete the crc VM? [y/N]: y
ERRO Error occurred: Machine "crc" does not exist. Use "crc start" to add a new one. 
```

PR #670, opened 2019-09-30
Issue #669 Check if crc VM exist before executing cmd functions.

praveenkumar:
Commands which require user input or expose some message should check
if crc vm exist, before display the message (or take user input).

```
$ ./crc stop
Stopping CodeReady Containers instance... this may take a few minutes
ERRO Error occurred: Machine "crc" does not exist. Use "crc start" to add a new one.

$ ./crc delete
Do you want to delete the crc VM? [y/N]: y
ERRO Error occurred: Machine "crc" does not exist. Use "crc start" to add a new one.
```

### DIFF
--- a/cmd/crc/cmd/console.go
+++ b/cmd/crc/cmd/console.go
@@ -48,6 +48,9 @@ func runConsole(arguments []string) {
 	consoleConfig := machine.ConsoleConfig{
 		Name: constants.DefaultName,
 	}
+
+	exitIfMachineMissing(consoleConfig.Name)
+
 	result, err := machine.GetConsoleURL(consoleConfig)
 	if err != nil {
 		errors.Exit(1)
--- a/cmd/crc/cmd/delete.go
+++ b/cmd/crc/cmd/delete.go
@@ -34,6 +34,9 @@ func runDelete(arguments []string) {
 	deleteConfig := machine.DeleteConfig{
 		Name: constants.DefaultName,
 	}
+
+	exitIfMachineMissing(deleteConfig.Name)
+
 	if clearCache {
 		deleteCache()
 	}
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -489,20 +489,20 @@ func Ip(ipConfig IpConfig) (IpResult, error) {
 
 func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
 	result := &ClusterStatusResult{Name: statusConfig.Name, Success: true}
-	api := libmachine.NewClient(constants.MachineBaseDir, constants.MachineCertsDir)
-	_, err := api.Exists(statusConfig.Name)
+	libMachineAPIClient := libmachine.NewClient(constants.MachineBaseDir, constants.MachineCertsDir)
+	_, err := libMachineAPIClient.Exists(statusConfig.Name)
 	if err != nil {
 		result.Success = false
 		result.Error = err.Error()
 		return *result, errors.New(err.Error())
 	}
-	defer api.Close()
+	defer libMachineAPIClient.Close()
 
 	openshiftStatus := "Stopped"
 	var diskUse int64
 	var diskSize int64
 
-	host, err := api.Load(statusConfig.Name)
+	host, err := libMachineAPIClient.Load(statusConfig.Name)
 	if err != nil {
 		result.Success = false
 		result.Error = err.Error()
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -400,6 +400,7 @@ func Stop(stopConfig StopConfig) (StopResult, error) {
 		result.Error = err.Error()
 		return *result, errors.New(err.Error())
 	}
+	defer libMachineAPIClient.Close()
 
 	result.State, _ = host.Driver.GetState()
 
@@ -424,6 +425,7 @@ func PowerOff(PowerOff PowerOffConfig) (PowerOffResult, error) {
 		result.Error = err.Error()
 		return *result, errors.New(err.Error())
 	}
+	defer libMachineAPIClient.Close()
 
 	if err := host.Kill(); err != nil {
 		result.Success = false
@@ -446,6 +448,7 @@ func Delete(deleteConfig DeleteConfig) (DeleteResult, error) {
 		result.Error = err.Error()
 		return *result, errors.New(err.Error())
 	}
+	defer libMachineAPIClient.Close()
 
 	m := errors.MultiError{}
 	m.Collect(host.Driver.Remove())
@@ -475,6 +478,7 @@ func Ip(ipConfig IpConfig) (IpResult, error) {
 		result.Error = err.Error()
 		return *result, errors.New(err.Error())
 	}
+	defer libMachineAPIClient.Close()
 	if result.IP, err = host.Driver.GetIP(); err != nil {
 		result.Success = false
 		result.Error = err.Error()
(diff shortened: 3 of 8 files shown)


cmd/crc/cmd/console.go
```diff
@@ -48,6 +48,12 @@ func runConsole(arguments []string) {
 	consoleConfig := machine.ConsoleConfig{
 		Name: constants.DefaultName,
 	}
+
+	// Check if the crc instance is exist
+	if _, _, err := machine.GetHost(consoleConfig.Name); err != nil {
+		errors.ExitWithMessage(1, err.Error())
+	}
+
```
cfergeau: I don't think all these tests should be done in cmd/crc/cmd/, they belong in the corresponding pkg/crc/machine/ methods so that the system tray gets the same behaviour. Only "crc delete" is a bit trickier if you want to avoid the user interaction when the VM is not running.

pkg/crc/machine/machine.go
```diff
@@ -392,14 +398,13 @@ func Stop(stopConfig StopConfig) (StopResult, error) {
...
-	libMachineAPIClient := libmachine.NewClient(constants.MachineBaseDir, constants.MachineCertsDir)
-	host, err := libMachineAPIClient.Load(stopConfig.Name)
-
+	libMachineAPIClient, host, err := GetHost(stopConfig.Name)
 	if err != nil {
 		result.Success = false
 		result.Error = err.Error()
 		return *result, errors.New(err.Error())
 	}
+	defer libMachineAPIClient.Close()
```
cfergeau: did you test each command with the addition of this libMachineAPIClient.Close()? I think I tried to add this some time ago, and that in some commands it was causing a crash.

cmd/crc/cmd/console.go
cfergeau: It's also missing a `defer firstRetVal.Close()` similar to what you fixed in the previous commit ;)

cmd/crc/cmd/console.go
```diff
@@ -48,6 +48,12 @@ func runConsole(arguments []string) {
 	consoleConfig := machine.ConsoleConfig{
 		Name: constants.DefaultName,
 	}
+
+	// Check if the crc instance is exist
```
cfergeau: "check if the crc instance exists"

cmd/crc/cmd/console.go
praveenkumar: @cfergeau right so system tray also should first try to see the exist of the VM and the only run other operation `GetHost` is something done in the `pkg/crc/machine` side not part of the `cmd` side. 

> It's also missing a defer firstRetVal.Close() similar to what you fixed in the previous commit ;)

Ah right, will do that.

pkg/crc/machine/machine.go
praveenkumar: I didn't check for all the commands but checked for delete/stop and that's working, I will check for others and see if that cause any issue.

praveenkumar: @cfergeau checked all the commands, didn't see any issue about causing a crash.

cmd/crc/cmd/console.go
praveenkumar: Done

cmd/crc/cmd/console.go
praveenkumar: > It's also missing a defer firstRetVal.Close() similar to what you fixed in the previous commit ;)

Done, good catch 👍

cfergeau: > @cfergeau right so system tray also should first try to see the exist of the VM and the only run other operation 

My understanding is that this is precisely the kind of code duplication we want to avoid "in order to implement 'Stop', the systray implementation should do X, then Y, then Z". It's probably fine to have some sort of generic "preflight" check which has to be called before calling "Stop()"

praveenkumar: > probably fine to have some sort of generic "preflight" check which has to be called before calling "Stop()"

What do you mean by having a kind of `preflight` , doesn't `GetHost` is actually doing it? Also on the tray side if you want to just consume the `Stop` , `Start` you can directly do since those functionality doesn't change at all.

cmd/crc/cmd/delete.go
```diff
@@ -34,6 +34,14 @@ func runDelete(arguments []string) {
...
 		Name: constants.DefaultName,
 	}
+
+	// Check if the crc instance exists
+	libMachineAPIClient, _, err := machine.GetHost(deleteConfig.Name)
+	if err != nil {
+		errors.ExitWithMessage(1, err.Error())
+	}
+	defer libMachineAPIClient.Close()
+
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau Done, PTAL.

cmd/crc/cmd/delete.go
```diff
@@ -34,6 +34,16 @@ func runDelete(arguments []string) {
...
+	// Check if the crc instance exists
+	exist, err := machine.MachineExists(deleteConfig.Name)
+	if err != nil {
+		errors.ExitWithMessage(1, err.Error())
+	}
+	if !exist {
+		errors.ExitWithMessage(1, machineNotExistsErrorMessage)
+	}
+
```
cfergeau: This could go to a helper `checkMachineExists(machineName string)`/`assertMachineExists`/`exitIfMachineMissing` rather than duplicating the same code multiple times

cmd/crc/cmd/delete.go
```diff
@@ -34,6 +34,16 @@ func runDelete(arguments []string) {
 	deleteConfig := machine.DeleteConfig{
 		Name: constants.DefaultName,
 	}
+
+	// Check if the crc instance exists
+	exist, err := machine.MachineExists(deleteConfig.Name)
```
cfergeau: 'exists' I think

cmd/crc/cmd/constants.go
```diff
@@ -1,7 +1,8 @@
 package cmd
 
 const (
-	commandName      = "crc"
-	descriptionShort = "Local OpenShift 4.x cluster"
-	descriptionLong  = "Deploy local instance of an OpenShift 4.x cluster for development purpose"
+	commandName                  = "crc"
+	descriptionShort             = "Local OpenShift 4.x cluster"
+	descriptionLong              = "Deploy local instance of an OpenShift 4.x cluster for development purpose"
+	machineNotExistsErrorMessage = "Machine \"crc\" does not exist. Use \"crc start\" to add a new one."
```
cfergeau: "nonExistingMachineErrorMessage", or "missingMachineErrorMessage" is better I think. But if the machine existence check is moved to a helper, we don't have to define a constant for the error message.

cmd/crc/cmd/delete.go
praveenkumar: @cfergeau make sense, updated :+1:

cmd/crc/cmd/delete.go
```diff
@@ -34,6 +34,10 @@ func runDelete(arguments []string) {
 	deleteConfig := machine.DeleteConfig{
 		Name: constants.DefaultName,
 	}
+
+	// Check if the crc instance exists
+	exitIfMachineMissing(deleteConfig.Name)
```
cfergeau: The comment is now redundant as you have a very explicit function name

praveenkumar:
@cfergeau should the exit code in this case is `0` or `1` I am incline to have it `0` since the VM is not even exist so the command just display that and exit with message ?

cfergeau:
> @cfergeau should the exit code in this case is `0` or `1` I am incline to have it `0` since the VM is not even exist so the command just display that and exit with message ?

I would return 1 as it's an error to call it on a non existing VM. That's similar to calling `virsh undefine nonexistentvm` which returns 1

cmd/crc/cmd/delete.go
praveenkumar: @cfergeau Done

PR #671, opened 2019-09-30
Update bundle location and bundle version

anjannath:


centos_ci.sh
```diff
@@ -108,7 +108,7 @@ function perform_artifacts_upload() {
 
 function get_bundle() {
   mkdir $HOME/Downloads
-  curl -L 'https://github.com/code-ready/crc-ci-jobs/releases/download/4.0.1/crc_bundle.zip' -o $HOME/Downloads/bundle.zip
+  curl -L 'https://storage.googleapis.com/crc-bundle-github-ci/crc_bundle.zip' -o $HOME/Downloads/bundle.zip
```
cfergeau: Could we use "BUNDLE_VERSION" so that it's relatively easy to know which bundle is being downloaded?

Makefile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-BUNDLE_VERSION = 4.1.14
+BUNDLE_VERSION = 4.2.0-0.nightly-2019-09-26-192831
```
cfergeau: Commit log is missing a lot of context explaining why we are switching to stable 4.1 to a nightly 4.2 now.

centos_ci.sh
anjannath: we are using the simple name `crc_bundle.zip` to keep this link unchanged. To find out which bundle was downloaded/used you need to anyway go check the CI logs, which already shows the bundle name after unzipping.

Makefile
anjannath: I guess this is just a regular update, since the new bits also have this bundle, @praveenkumar would be able to give more information here.

praveenkumar: @cfergeau what more details do we require, we are shifting to 4.1 to 4.2 side and right now only nightly bits are available?

centos_ci.sh
cfergeau: The link would also be mostly unchanged with curl -L "..../releases/download/$BUNDLE_VERSION/..." :)

Makefile
cfergeau: Yes, something like this, and also that we'll aim for 4.2 in the final release, and that we only do it now because before that there were no publicly available 4.2 bits.

centos_ci.sh
cfergeau: One more comment on that, having a non-versioned URL will get in the way if we maintain 2 branches which expect 2 different bundles.

Issue #674, opened 2019-10-01
Change crc log file permission before uploading to centos CI artifact server


praveenkumar:
Right now the `crc.log` file have `0600` as file permission but when we upload same file to the centos CI artifact server then this file get `permission denied` during browsing http://artifacts.ci.centos.org/minishift/crc/pr/670/crc_2019_09_30_10_54_11.log

We need to change the permission of that file `0644` before upload.

anjannath:
Fixed in https://github.com/code-ready/crc/commit/7f501486b434b5afec12a5ce2e048e45539ae4f9

PR #675, opened 2019-10-01
Issue #674 Change log file permission before uploading to CI artifact server

praveenkumar:
Right now the crc.log file have 0600 as file permission but when we upload
same file to the centos CI artifact server then this file get permission
denied during browsing

http://artifacts.ci.centos.org/minishift/crc/pr/670/crc_2019_09_30_10_54_11.log

We need to change the permission of that file 0644 before upload.

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -118,6 +118,8 @@ function upload_logs() {
   # http://stackoverflow.com/a/22908437/1120530; Using --relative as --rsync-path not working
   mkdir -p pr/$ghprbPullId/
   cp -R test/integration/out/test-results/* pr/$ghprbPullId/
+  # Change the file permission to 0644 so after rsync it can be readable by other user.
+  chmod 0644 $HOME/.crc/crc.log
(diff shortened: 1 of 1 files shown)


centos_ci.sh
```diff
@@ -118,6 +118,8 @@ function upload_logs() {
   # http://stackoverflow.com/a/22908437/1120530; Using --relative as --rsync-path not working
   mkdir -p pr/$ghprbPullId/
   cp -R test/integration/out/test-results/* pr/$ghprbPullId/
+  # Change the file permission to 0644 so after rsync it can be readable by other user.
+  chmod 0644 $HOME/.crc/crc.log
```
anjannath: Wonder how it was working before! :-/

praveenkumar: @anjannath previously our file permission was different and after adding gosec we made those modifications.

anjannath: Ahh.. Okay, that makes sense!

Issue #673, opened 2019-09-30
Integration tests don't run with embedded bundle binary


jsliacan:
Running integration tests (independent of the binary) against a compiled binary results in `constants.BundleEmbedded` returning false as the tests weren't compiled with the binary. 

Suggested solution is to pass a flag to `make integration` specifying if the bundle is embedded or not.

jsliacan:
Resolved via #724

PR #676, opened 2019-10-01
Issue #673 Integration tests run with embedded bundle binary

jsliacan:
Made changes to integration tests such that they can now run against a binary with embedded bundle.

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -130,7 +130,7 @@ function run_tests() {
   set +e
   # In Jenkins slave we have pull secret file in the $HOME/payload/crc_pull_secret
   # this is copied over using https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml#L99
-  make integration BUNDLE_LOCATION=$HOME/Downloads/$BUNDLE PULL_SECRET_FILE=$HOME/payload/crc_pull_secret
+  make integration PULL_SECRET_FILE=$HOME/payload/crc_pull_secret BUNDLE_LOCATION=$HOME/Downloads/$BUNDLE
   if [[ $? -ne 0 ]]; then
     upload_logs $1
     exit 1
--- a/developing.adoc
+++ b/developing.adoc
@@ -63,10 +63,9 @@ Clicumber allows running commands in a persistent shell instance (`bash`, `tcsh`
 To start integration tests, run:
 
 ```bash
-$ make integration BUNDLE_LOCATION=<bundle location> PULL_SECRET_FILE=<pull secret path>
+$ make integration PULL_SECRET_FILE=<pull secret path> BUNDLE_LOCATION=<bundle location>
 ```
-where `<bundle location>` is either the bundle's URL or its path in the filesystem,
-and `<pull secret path>` is the path to a file containing your OpenShift pull secret.
+where `<bundle location>` is the bundle's URL or its path in the filesystem if the bundle is not embedded. If bundle is embedded in the binary, `BUNDLE_LOCATION` is not passed to `make integration`. The `<pull secret path>` is the path to a file containing your OpenShift pull secret.
 The paths must be absolute paths.
 
 ===== How to run only a subset of all integration tests
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -30,14 +30,15 @@ import (
 	"time"
 
 	clicumber "github.com/code-ready/clicumber/testsuite"
-	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/oc"
 )
 
 var (
 	CRCHome        string
+	bundleEmbedded bool
 	bundleURL      string
 	bundleName     string
+	bundleVersion  string
 	pullSecretFile string
 )
 
@@ -91,21 +92,25 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeFeature(func(this *gherkin.Feature) {
 
-		if _, err := os.Stat(bundleName); os.IsNotExist(err) {
-			// Obtain the bundle to current dir
-			fmt.Println("Obtaining bundle...")
-			bundle, err := DownloadBundle(bundleURL, ".")
-			if err != nil {
-				fmt.Printf("Failed to obtain CRC bundle, %v\n", err)
+		if !bundleEmbedded {
+			if _, err := os.Stat(bundleName); os.IsNotExist(err) {
+				// Obtain the bundle to current dir
+				fmt.Println("Obtaining bundle...")
+				fmt.Printf("bundleURL = %s\n", bundleURL)
+				bundle, err := DownloadBundle(bundleURL, ".")
+				if err != nil {
+					fmt.Printf("Failed to obtain CRC bundle, %v\n", err)
+					os.Exit(1)
+				}
+				fmt.Println("Using bundle:", bundle)
+			} else if err != nil {
+				fmt.Printf("Unknown error obtaining the bundle %v.\n", bundleName)
 				os.Exit(1)
+			} else {
+				fmt.Println("Using existing bundle:", bundleName)
 			}
-			fmt.Println("Using bundle:", bundle)
-		} else if err != nil {
-			fmt.Printf("Unknown error obtaining the bundle %v.\n", bundleName)
-			os.Exit(1)
-		} else {
-			fmt.Println("Using existing bundle:", bundleName)
 		}
+		fmt.Println("Expecting bundle to be embedded in the CRC binary.")
 
 	})
 }
@@ -260,7 +265,7 @@ func StartCRCWithDefaultBundleAndDefaultHypervisorSucceedsOrFails(expected strin
 	var cmd string
 	var extraBundleArgs string
 
-	if !constants.BundleEmbedded() {
+	if !bundleEmbedded {
 		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
 	}
 	cmd = fmt.Sprintf("crc start -p '%s' %s --log-level debug", pullSecretFile, extraBundleArgs)
@@ -274,7 +279,7 @@ func StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails(hypervisor string, ex
 	var cmd string
 	var extraBundleArgs string
 
-	if !constants.BundleEmbedded() {
+	if !bundleEmbedded {
 		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
 	}
 	cmd = fmt.Sprintf("crc start -d %s -p '%s' %s --log-level debug", hypervisor, pullSecretFile, extraBundleArgs)
--- a/test/integration/crcsuite/prepare.go
+++ b/test/integration/crcsuite/prepare.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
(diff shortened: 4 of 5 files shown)


jsliacan:
retest this please

jsliacan:
retest this please

cfergeau:
CI failed on this one too, I'm not sure what happened :-/

jsliacan:
@cfergeau I checked a few failed PRs today and all failed because of: `Error checking if a new version is available: Error unmarshaling JSON metadata: invalid character '<' looking for beginning of value` discussed on Slack today. I assume it'll be addressed soon.

cfergeau:
> @cfergeau I checked a few failed PRs today and all failed because of: `Error checking if a new version is available: Error unmarshaling JSON metadata: invalid character '<' looking for beginning of value` discussed on Slack today. I assume it'll be addressed soon.

This message appears in debug, but I don't think it's fatal. Or was it?

jsliacan:
You're probably right, this time it failed in multiple places too :(.

developing.adoc
```diff
@@ -65,8 +65,7 @@ To start integration tests, run:
 ```bash
 $ make integration BUNDLE_LOCATION=<bundle location> PULL_SECRET_FILE=<pull secret path>
 ```
-where `<bundle location>` is either the bundle's URL or its path in the filesystem,
-and `<pull secret path>` is the path to a file containing your OpenShift pull secret.
+where `<bundle location>` is either the bundle's URL, its path in the filesystem, or `embedded` if the bundle is embedded in the binary. The `<pull secret path>` is the path to a file containing your OpenShift pull secret.
```
cfergeau: Have you considered not setting BUNDLE_LOCATION at all when the bundle is embedded?

test/integration/crcsuite/crcsuite.go
```diff
@@ -286,11 +318,41 @@ func StartCRCWithDefaultBundleAndHypervisorSucceedsOrFails(hypervisor string, ex
...
 	err := clicumber.ExecuteCommandSucceedsOrFails(cmd, expected)
 
 	return err
 }
+
+func SetBundleName() error {
+
+	files, err := ioutil.ReadDir(CRCHome)
+	if err != nil {
+		fmt.Println("Cannot list files in %s: %s.\n", CRCHome)
```
cfergeau: It's missing `, err`

test/integration/crcsuite/prepare.go
```diff
@@ -61,12 +61,19 @@ func DownloadBundle(bundleLocation string, bundleDestination string) (string, er
...
+	if flag.NArg() < 3 {
 		fmt.Printf("Invalid number of arguments, the paths to the bundle file and to the pull secret file are required\n")
 		os.Exit(1)
 	}
-	bundleURL = flag.Args()[0]
-	_, bundleName = filepath.Split(bundleURL)
+
+	if flag.Args()[0] == "embedded" {
+		bundleEmbedded = true
+		bundleVersion = flag.Args()[2]
```
cfergeau: From 'bundleVersion', don't we have a good idea of the expected bundleName? This would avoid the magic in `SetBundleName`

developing.adoc
jsliacan: @cfergeau yeah, but I figured it's better explicit. If integration can be run without the argument, then it's not immediately obvious that sometimes BUNDLE_LOCATION is needed. Reminds me of `crc` binary being explicitly stated vs implicitly taken from top of the `PATH`.

test/integration/crcsuite/prepare.go
jsliacan: @cfergeau We do, up to hypervisor - which is part of the name but not part of bundleVersion. If you happen to see a way that avoids `SetBundleName`, I'm all ears because I don't like it much either, but didn't manage to circumvent it. Thanks! I'll also have a think once more.

cfergeau: If the bundle is not embedded, the full name was passed on the command line
If the bundle is embedded, we know the version, and I'd make the assumption that the hypervisor is the native one for the platform (ie not virtualbox). Similar to GetDefaultBundle().

jsliacan: @cfergeau I changed it to what I think you meant. There are a few dodgy bits:

- The args to `make integration` are order sensitive. And I had to swap order since `BUNDLE_LOCATION` is optional now.
- I don't check for the situation when bundle is embedded and user still passes `BUNDLE_LOCATION`. As it is, if bundle is embedded, user must not specify bundle (that's how you tell tests that bundle is embedded). Not sure if that's good -- could be too implicit?
- bundle version is taken from Makefile, and not from the actual bundle. So the Makefile which comes with tests could contain a different version than the bundle, e.g. if user runs older CRC.

Some of these can be changed, e.g. the first one: I can probably pass bundle location to tests via tags. But this will be a separate PR as it affects other args too.

cfergeau:
```
10:32:23 Obtaining bundle...
10:32:23 panic: runtime error: slice bounds out of range [:4] with length 0
10:32:23 
10:32:23 goroutine 1 [running]:
10:32:23 github.com/code-ready/crc/test/integration/crcsuite.DownloadBundle(0x0, 0x0, 0x9531a5, 0x1, 0x0, 0x0, 0x0, 0x0)
10:32:23 	/home/crc_ci/payload/test/integration/crcsuite/prepare.go:20 +0x66a
10:32:23 github.com/code-ready/crc/test/integration/crcsuite.FeatureContext.func3(0xc0000ca370)
10:32:23 	/home/crc_ci/payload/test/integration/crcsuite/crcsuite.go:99 +0x2dd
10:32:23 github.com/DATA-DOG/godog.(*Suite).runFeature(0xc000266000, 0xc0001f2fc0)
10:32:23 	/home/crc_ci/gopath/pkg/mod/github.com/!d!a!t!a-!d!o!g/godog@v0.7.13/suite.go:524 +0x6d3
10:32:23 github.com/DATA-DOG/godog.(*Suite).run(0xc000266000)
10:32:23 	/home/crc_ci/gopath/pkg/mod/github.com/!d!a!t!a-!d!o!g/godog@v0.7.13/suite.go:220 +0xaf
10:32:23 github.com/DATA-DOG/godog.(*runner).run(0xc00025eb40, 0x14)
10:32:23 	/home/crc_ci/gopath/pkg/mod/github.com/!d!a!t!a-!d!o!g/godog@v0.7.13/run.go:76 +0xae
10:32:23 github.com/DATA-DOG/godog.RunWithOptions(0x95373b, 0x3, 0x978470, 0x0, 0x0, 0x0, 0x7ffeea0e499a, 0x5, 0x954b82, 0x6, ...)
10:32:23 	/home/crc_ci/gopath/pkg/mod/github.com/!d!a!t!a-!d!o!g/godog@v0.7.13/run.go:173 +0x8d8
10:32:23 github.com/code-ready/crc/test/integration.TestMain(0xc0001b1ed8)
10:32:23 	/home/crc_ci/payload/test/integration/integration_test.go:35 +0x187
10:32:23 main.main()
10:32:23 	_testmain.go:40 +0x17c
10:32:23 FAIL	github.com/code-ready/crc/test/integration	0.044s
10:32:23 FAIL
```

Some issues with `if bundleLocation[:4] != "http" {` in `DownloadBundle` in `test/integration/crcsuite/prepare.go`

centos_ci.sh
```diff
@@ -129,7 +129,7 @@ function run_tests() {
   set +e
   # In Jenkins slave we have pull secret file in the $HOME/payload/crc_pull_secret
   # this is copied over using https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml#L99
-  make integration BUNDLE_LOCATION=$HOME/Downloads/$BUNDLE PULL_SECRET_FILE=$HOME/payload/crc_pull_secret
+  make integration PULL_SECRET_FILE=$HOME/payload/crc_pull_secret BUNDLE_LOCATION=$HOME/Downloads/$BUNDLE
```
cfergeau: actually order here is probably not important as what matters is that the 3 variables are defined. ACK to putting them in the same order as in Makefile, less confusing that way.

test/integration/crcsuite/crcsuite.go
```diff
@@ -91,21 +92,25 @@ func FeatureContext(s *godog.Suite) {
...
 
-		if _, err := os.Stat(bundleName); os.IsNotExist(err) {
-			// Obtain the bundle to current dir
-			fmt.Println("Obtaining bundle...")
-			bundle, err := DownloadBundle(bundleURL, ".")
-			if err != nil {
-				fmt.Printf("Failed to obtain CRC bundle, %v\n", err)
+		if !bundleEmbedded {
+			if _, err := os.Stat(bundleName); os.IsNotExist(err) {
+				// Obtain the bundle to current dir
```
cfergeau: "Copy" instead of "Obtain" ?

test/integration/crcsuite/crcsuite.go
```diff
@@ -91,21 +92,25 @@ func FeatureContext(s *godog.Suite) {
...
+				fmt.Println("Obtaining bundle...")
+				fmt.Printf("bundleURL = %s\n", bundleURL)
+				bundle, err := DownloadBundle(bundleURL, ".")
+				if err != nil {
+					fmt.Printf("Failed to obtain CRC bundle, %v\n", err)
+					os.Exit(1)
+				}
+				fmt.Println("Using bundle:", bundle)
+			} else if err != nil {
+				fmt.Printf("Unknown error obtaining the bundle %v.\n", bundleName)
```
cfergeau: Unknown or unexpected?

test/integration/crcsuite/prepare.go
```diff
@@ -62,12 +63,43 @@ func ParseFlags() {
...
+		var hypervisor string
+		switch os := runtime.GOOS; os {
+		case "darwin":
+			hypervisor = "hyperkit"
+		case "linux":
+			hypervisor = "libvirt"
+		case "windows":
+			hypervisor = "hyperv"
+		default:
+			fmt.Printf("Unsupported OS: %s", os)
```
cfergeau: I would exit the binary at this point.

test/integration/crcsuite/prepare.go
```diff
@@ -62,12 +63,43 @@ func ParseFlags() {
...
+		}
+
+		bundleName = fmt.Sprintf("crc_%s_%s.crcbundle", hypervisor, bundleVersion)
+
+	}
+
+	// separate bundle
+	// 3: on laptop
+	// 4: on Centos CI
+	if flag.NArg() > 2 {
```
cfergeau: Not sure I understand the 3: on laptop and 4: on Centos CI, we only make use of arg 1 and 2?

jsliacan: I think we make use of arg 0 (pull secret) and arg 1 (bundle location). There's also arg 2 (bundle version) which is ignored except when bundle is embedded (and bundle location is not passed), in which case it's actually arg 1. For some reason, on Centos CI, there's a 4th argument which is always there (I forgot what it was). Luckily though, we don't care about the case when bundle is embedded on CentosCI, because we don't test that case there. So it can be clumped with `flag.NArg() >= 3` on a laptop (i.e. not on CentosCI). 

This was the best I could do if "embedded" was supposed to be specified by absence of `BUNDLE_LOCATION`. I agree it's really confusing and really flimsy hunk of code! Hence I decided to rework it: see my [binloc branch](https://github.com/jsliacan/crc/tree/binloc). The solution I had in mind for specifying the binary location doesn't work yet, so it's not there.

PR #680, opened 2019-10-02
[WIP] Update dnsmasq configuration

anjannath:


praveenkumar:
@anjannath can we close this, I don't think it is solving any problem also when you use `resolv-file` then every request will go through defined nameserver on this file (in this case google one) so might not work if some organization block those nameservers. 

```
-r, --resolv-file=<file>
Read the IP addresses of the upstream nameservers from <file>, instead of /etc/resolv.conf. For the format of this file see resolv.conf(5). The only lines relevant to dnsmasq are nameserver ones. Dnsmasq can be told to poll more than one resolv.conf file, the first file name specified overrides the default, subsequent ones add to the list. This is only allowed when polling; the file with the currently latest modification time is the one used.
```

anjannath:
@praveenkumar Right, in that case we should remove the file from the container. Closing

Issue #681, opened 2019-10-02
Legacy references to virtualbox make integration tests fail on mac


jsliacan:
### General information

  * OS: macOS
  * Hypervisor: hyperkit

Not all references to `virtualbox` in tests rely on `virtualbox` being the used hyperviser, but some do: 
```
Scenario Outline: CRC config checks (skips) # features/config.feature:64
    Then stdout should contain "Skipping above check ..." # features/config.feature:68
      Error: output did not match. Expected: 'Skipping above check ...', Actual: 'level=info msg="Checking if running as non-root"
level=info msg="Caching oc binary"
level=info msg="Setting up virtualization with HyperKit"
level=info msg="Installing crc-machine-hyperkit"
level=info msg="Setting file permissions for /etc/resolver/testing"
level=info msg="Setting file permissions for /etc/hosts"
level=info msg="Unpacking bundle from the CRC binary"
Setup is complete, you can now run 'crc start' to start a CodeReady Containers instance'
```
It's probably a good idea to substitute virtualbox with hyperkit everywhere in tests.

PR #682, opened 2019-10-02
Issue #681 Replace references to virtualbox in tests by hyperkit

jsliacan:


praveenkumar:
retest this please.

test/integration/features/config.feature
```diff
@@ -74,7 +74,7 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
         @darwin
         Examples:
             | property                             | value1 | value2 |
-            | skip-check-virtualbox-installed      | true   | false  |
+            | skip-check-hyperkit-installed        | true   | false  |
```
praveenkumar: @jsliacan this looks like some extra whitespace at the end of `skip-check-hyperkit-installed`, `warn-check-hyperkit-installed` and `hyperkit`

jsliacan: I reformatted all tables in `config.feature`, but some of those you pointed out didn't change. It doesn't look like there are extra white spaces though if you look here (I use an auto-fmt for feature files in emacs): https://github.com/jsliacan/crc/blob/killvboxrefs/test/integration/features/config.feature

cfergeau: The formatting is done with whitespaces, the ones which are highlighted in this diff is the 'ed' before which is changed to '  '. Change looks good to me, especially as CI passes ;)

PR #689, opened 2019-10-04
Rework cert expiry code

cfergeau:
This removes duplicated code, and tries to make the expiration tests easier to read.
This also fixes a bug with the warning when the check based on the bundle build time fails. Before this commit it could say that the bundle is going to expire in -5 days. Now this errors out in such situations.
The commit also adds a link to the documentation explaining how to solve certificate issues.
The custom errors/type matching might be a bit too sophisticated, I can rework this if needed.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -22,30 +22,25 @@ func WaitForSsh(sshRunner *ssh.SSHRunner) error {
 	return errors.RetryAfter(60, checkSshConnectivity, time.Second)
 }
 
-// CheckCertsValidityUsingBundleBuildTime check if the cluster certs going to expire soon.
-func CheckCertsValidityUsingBundleBuildTime(buildTime time.Time) (bool, int) {
-	certExpiryDate := buildTime.AddDate(0, 1, 0)
-	// Warn user if the cert expiry going to happen starting of the 7 days
-	timeAfter7Days := time.Now().AddDate(0, 0, 7)
-	return timeAfter7Days.After(certExpiryDate), int(time.Until(certExpiryDate).Hours()) / 24
-}
+type CertExpiryState int
+
+const (
+	Unknown CertExpiryState = iota
+	CertNotExpired
+	CertExpired
+)
 
 // CheckCertsValidity checks if the cluster certs have expired or going to expire in next 7 days
-func CheckCertsValidity(sshRunner *ssh.SSHRunner) (bool, int, error) {
+func CheckCertsValidity(sshRunner *ssh.SSHRunner) (CertExpiryState, error) {
 	certExpiryDate, err := getcertExpiryDateFromVM(sshRunner)
 	if err != nil {
-		return false, 0, err
+		return Unknown, err
 	}
 	if time.Now().After(certExpiryDate) {
-		return false, 0, fmt.Errorf("Certs have expired, they were valid till: %s", certExpiryDate.Format(time.RFC822))
+		return CertExpired, fmt.Errorf("Certs have expired, they were valid till: %s", certExpiryDate.Format(time.RFC822))
 	}
 
-	// Warn user if the cert expiry going to happen starting of the 7 days
-	timeAfter7Days := time.Now().AddDate(0, 0, 7)
-	if timeAfter7Days.After(certExpiryDate) {
-		return true, int(time.Until(certExpiryDate).Hours()) / 24, nil
-	}
-	return false, 0, nil
+	return CertNotExpired, nil
 }
 
 func getcertExpiryDateFromVM(sshRunner *ssh.SSHRunner) (time.Time, error) {
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -131,16 +131,6 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error getting bundle metadata: %v", err)
 		}
 
-		// Check if certificate is going to expire in next 7 days
-		buildTime, err := crcBundleMetadata.GetBundleBuildTime()
-		if err != nil {
-			result.Error = err.Error()
-			return *result, errors.Newf("Error getting bundle build time: %v", err)
-		}
-		if goingToExpire, duration := cluster.CheckCertsValidityUsingBundleBuildTime(buildTime); goingToExpire {
-			logging.Warnf("Bundle certificates are going to expire in %d days, better to use new release", duration)
-		}
-
 		openshiftVersion := crcBundleMetadata.GetOpenshiftVersion()
 		if openshiftVersion == "" {
 			logging.Infof("Creating VM...")
@@ -257,13 +247,13 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	// Check the certs validity inside the vm
 	needsCertsRenewal := false
 	logging.Info("Verifying validity of the cluster certificates ...")
-	expiringIn7Days, duration, err := cluster.CheckCertsValidity(sshRunner)
+	certExpiryState, err := cluster.CheckCertsValidity(sshRunner)
 	if err != nil {
-		needsCertsRenewal = true
-	} else if exists {
-		// Only show when VM is started from stopped state.
-		if expiringIn7Days {
-			logging.Warnf("Bundle certificates are going to expire in %d days, better to use new release", duration)
+		if certExpiryState == cluster.CertExpired {
+			needsCertsRenewal = true
+		} else {
+			result.Error = err.Error()
+			return *result, errors.New(err.Error())
 		}
 	}
 	// Add nameserver to VM if provided by User


pkg/crc/cluster/cluster.go
```diff
@@ -22,33 +21,7 @@ func WaitForSsh(sshRunner *ssh.SSHRunner) error {
...
-	// Warn user if the cert expiry going to happen starting of the 7 days
-	timeAfter7Days := time.Now().AddDate(0, 0, 7)
-	if timeAfter7Days.After(certExpiryDate) {
-		return true, int(time.Until(certExpiryDate).Hours()) / 24, nil
-	}
-	return false, 0, nil
-}
-
-func getcertExipryDateFromVM(sshRunner *ssh.SSHRunner) (time.Time, error) {
+func GetCertExpiryDateFromVM(sshRunner *ssh.SSHRunner) (time.Time, error) {
```
gbraad: not used to see us passing around the `SSHRunner` instead of `Host` in the context of VM.

pkg/crc/machine/certs-expiry.go
```diff
@@ -0,0 +1,66 @@
...
+	if daysLeft < 0 {
+		return certsExpiredError{expiryTime}
+	} else if daysLeft < 7 {
+		return certsExpiringError{daysLeft}
+	}
+
+	return nil
+}
+
+func checkVMCertsExpiry(sshRunner *crcssh.SSHRunner) error {
```
gbraad: It would be nicer to keep this as `Host` in the context of a VM. I know this wasn't in the original code ...

pkg/crc/machine/machine.go
```diff
@@ -128,14 +128,16 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error getting bundle metadata: %v", err)
 		}
 
-		// Check if certificate is going to expire in next 7 days
-		buildTime, err := crcBundleMetadata.GetBundleBuildTime()
+		// Check if certificate is expired/going to expire in the next 7 days
+		err := checkBundleCertsExpiry(crcBundleMetadata)
```
gbraad: `s/Expire/Expiration` ?

pkg/crc/machine/certs-expiry.go
gbraad: Don't really like the use of `error` here. I believe this should be `bool, error`. In which `err` is only used when an actual error occurs.

pkg/crc/machine/machine.go
```diff
@@ -253,17 +255,19 @@ func Start(startConfig StartConfig) (StartResult, error) {
 
 	// Check the certs validity inside the vm
 	logging.Info("Verifying validity of the cluster certificates ...")
-	expiringIn7Days, duration, err := cluster.CheckCertsValidity(sshRunner)
+	err = checkVMCertsExpiry(sshRunner)
```
gbraad:
```
valid, err = checkVMCertsExpiry(sshRunner)

if !valid && err != nil {
  
```

as this would make the code easier to understand than having to interfere the use of an error to communicate this state

pkg/crc/cluster/cluster.go
cfergeau: Now that we are changing the ssh key of the VM without explicit Host interaction, this SSHRunner is needed so that we know which ssh key we should be using. At some point, we probably should introduce a crc.VM type which would have host.Host and SSHRunner fields, but this does not exist at the moment.

pkg/crc/machine/machine.go
cfergeau: Probably https://writingexplained.org/expiry-vs-expiration-difference

pkg/crc/machine/certs-expiry.go
cfergeau: And I did not really like having a somehow 'magic' boolean in the previous implementation ;)
Would an enum work for you?
```
const (
    CertExpired = iota,
    CertExpiringIn7Days,
)
```

pkg/crc/machine/machine.go
cfergeau: Does 'valid' mean 'certificate is expiring in more than 7 days'? Or something else?

pkg/crc/cluster/cluster.go
gbraad: OK. we need to work on that abstraction sooner than later

pkg/crc/machine/certs-expiry.go
gbraad: something more explicit is always better. I think an `enum` would work better in this case... as it avoid a magic value; both processing logic and determining logic can/should use the same value in that case.

pkg/crc/machine/machine.go
gbraad: enum option is better, as this tries to capture a tri-state; valid, still valid and expired.

pkg/crc/machine/machine.go
cfergeau: git master is already using Expiry rather than Expiration. I'll add a commit to the end of the series changing Expiry to Expiration, we can the discuss whether to keep it or not :)

pkg/crc/machine/certs-expiry.go
cfergeau: Could not fully switch to enums if I want to hold the expiry date somewhere, so hybrid approach with error/enum.

pkg/crc/machine/machine.go
cfergeau: I added a commit on top of the PR doing the change, but the codebase is currently (before this PR) using expiry, so I'm not sure it's worth changing.

go.mod
```diff
@@ -25,3 +25,5 @@ require (
 	golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7
 	gopkg.in/AlecAivazis/survey.v1 v1.8.5
 )
+
+go 1.13
```
praveenkumar: @cfergeau why this mention the go version and also this is quite higher version and on RHEL side we still have 1.12.x

cfergeau: This is automatically added by go 1.13, and was not intentionally added to the PR, I'll remove it.
https://blog.ksub.org/bytes/2019/09/15/go.mods-go-directive/ has some explanations regarding what this means

praveenkumar: @cfergeau thanks for sharing, so that means it is just look for the feature set available for a specific version and not stop user to build it on older version as long as it have same feature set. I think it would be better to have 1.12 for now since we are going to use this as part of internal CI and then move to 1.13 once it is updated to RHEL side.

cfergeau: Yeah, agreed, I was removing it at the same time I was writing these comments :) Note that the commit changing "expiry" to "expiration" is mostly a RFC as I think both words are ok in the context of certificates.

pkg/crc/machine/certs-expiration.go
```diff
@@ -0,0 +1,67 @@
...
+	durationTillExpiration := time.Until(expirationTime)
+	return int(durationTillExpiration.Hours() / 24)
+}
+
+func checkCertsExpiration(expirationTime time.Time) (CertExpirationState, error) {
+	daysLeft := daysTillExpiration(expirationTime)
+	if daysLeft < 0 {
+		return CertExpired, certsExpiredError{expirationTime}
+	} else if daysLeft < 7 {
+		return CertExpiringIn7Days, nil
```
praveenkumar: @cfergeau This will always give warning that cert is going to expire in 7 days even only 1-2 days left which is wrong and user thinks he still have 7 days to delete/recreate the environment. It should be function instead part of const which take the number of day left as arg.

cfergeau: Ah right, maybe it's enough to change the warning from "will expire in 7 days" to "will expire in less than 7 days"?

gbraad: if no `-days` option is possible, say `less than 7 days left`

pkg/crc/cluster/cluster.go
```diff
@@ -22,30 +22,25 @@ func WaitForSsh(sshRunner *ssh.SSHRunner) error {
...
-	timeAfter7Days := time.Now().AddDate(0, 0, 7)
-	return timeAfter7Days.After(certExpiryDate), int(time.Until(certExpiryDate).Hours()) / 24
-}
+type CertExpiryState int
+
+const (
+	Unknown CertExpiryState = iota
+	CertNotExpired
+	CertExpired
+)
```
praveenkumar: 👍

pkg/crc/machine/machine.go
```diff
@@ -257,13 +247,13 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
-		needsCertsRenewal = true
-	} else if exists {
-		// Only show when VM is started from stopped state.
-		if expiringIn7Days {
-			logging.Warnf("Bundle certificates are going to expire in %d days, better to use new release", duration)
+		if certExpiryState == cluster.CertExpired {
+			needsCertsRenewal = true
+		} else {
+			result.Error = err.Error()
+			return *result, errors.New(err.Error())
```

PR #693, opened 2019-10-07
Issue #692 Unmarshall json only when http response status is 200

### LINKED ISSUE #692: Handle HTTP errors while fetching version information for update check
While GET ing version info we should take care of the different HTTP responses.

anjannath:
Fixes #692

### DIFF
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -69,6 +69,9 @@ func getCRCLatestVersionFromMirror() (*semver.Version, error) {
 		return nil, err
 	}
 	defer response.Body.Close()
+	if response.StatusCode != 200 {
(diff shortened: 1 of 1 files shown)


gbraad:
haven't loopked yet, but not just 200.
also, only when able to parse properly.

debug log errors, but treat errors as no value.

cfergeau:
One note for future improvements, we have a download.Download() helper which downloads a file from a http(s) location using cavaliercoder/grab. In getCRCLatestVersionFromMirror(), this helper is not really what we want because a io.Reader is more appropriate than a file for what is being done. It would be more consistent though to have a helper returning an io.Reader in the download package, and to use it in getCRCLatestVersionFromMirror().

PR #694, opened 2019-10-07
[hold] Update bundle version to 4.2.0-rc.0

anjannath:


cfergeau:
Pushing this as otherwise CI is broken since the bundle it uses was switched to rc.0

Issue #698, opened 2019-10-09
[BUG] Literal \n in crc.log


syedriko:
### General information

  * OS: Linux, Fedora 30
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? Yes

## CRC version
```bash
$ ./crc version
version: 1.0.0-rc.0+34371d3
OpenShift version: 4.2.0-0.nightly-2019-09-26-192831 (embedded in binary)
$ 
```
  
## CRC status
```bash
$ ./crc status
ERRO Error occurred: Machine "crc" does not exist. Use "crc start" to add a new one. 
$
```

## CRC config
```bash
$ ./crc config view
$
```

## Host Operating System
```bash
$ cat /etc/os-release
NAME=Fedora
VERSION="30 (Workstation Edition)"
ID=fedora
VERSION_ID=30
VERSION_CODENAME=""
PLATFORM_ID="platform:f30"
PRETTY_NAME="Fedora 30 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:30"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f30/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=30
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=30
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
$
```

### Steps to reproduce

  1. crc setup
  2. grep '\\n' ~/.crc/crc.log

### Expected
No output from "grep '\\n' ~/.crc/crc.log"

### Actual
$ grep '\\n' ~/.crc/crc.log
time="2019-10-08T16:31:17-04:00" level=debug msg="Download saved to /tmp/crc316183009/oc.tar.gz \n"
[...]

PR #699, opened 2019-10-09
Issue #698: Literal \n in crc.log

syedriko:
This is a fix for somewhat of an antipattern in forming log entries. For example, this
```
logging.Debugf("Download saved to %v \n", resp.Filename)
```
results in 
```
time="2019-10-08T16:31:17-04:00" level=debug msg="Download saved to /tmp/crc316183009/oc.tar.gz \n"
```

The following is an analysis of why that happens. The parameters of logging.Debugf() don't produce a complete line of output but just the "msg" field of a log entry. The logger takes care of terminating lines by adding a linebreak at the end of a log entry:
https://github.com/sirupsen/logrus/blob/de736cf91b921d56253b4010270681d33fdf7cb5/text_formatter.go#L218

But why do we end up with a literal \n in the output? Look here:
https://github.com/sirupsen/logrus/blob/de736cf91b921d56253b4010270681d33fdf7cb5/text_formatter.go#L318

f.needsQuoting() returns true due to the trailing linebreak in value and our log message runs through fmt.Sprintf("%q", stringVal), which turns the linebreak back into "\n"

Just so it's as clear as possible, this is the msg log field going into the fmt.Sprintf("%q", stringVal):

```
(dlv) print stringVal
"Download saved to /tmp/crc285030223/oc.tar.gz \n"
(dlv) print []byte(stringVal)
[]uint8 len: 47, cap: 47, [68,111,119,110,108,111,97,100,32,115,97,118,101,100,32,116,111,32,47,116,109,112,47,99,114,99,50,56,53,48,51,48,50,50,51,47,111,99,46,116,97,114,46,103,122,32,10]
(dlv)
```

Note that dlv prints out ASCII 10 as \n.

And here is what is returned from fmt.Sprintf("%q", stringVal):

```
(dlv) print s
"\"Download saved to /tmp/crc285030223/oc.tar.gz \\n\""
(dlv) print []byte(s)
[]uint8 len: 50, cap: 50, [34,68,111,119,110,108,111,97,100,32,115,97,118,101,100,32,116,111,32,47,116,109,112,47,99,114,99,50,56,53,48,51,48,50,50,51,47,111,99,46,116,97,114,46,103,122,32,92,110,34]
(dlv)
```

and we end up with a log entry that looks like this:
```
time="2019-10-08T20:57:58-04:00" level=debug msg="Download saved to /tmp/crc285030223/oc.tar.gz \n"
```

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -58,7 +58,7 @@ func fixBundleCached() (bool, error) {
 func checkOcBinaryCached() (bool, error) {
 	oc := oc.OcCached{}
 	if !oc.IsCached() {
-		return false, errors.New("oc binary is not cached.")
+		return false, errors.New("oc binary is not cached")
 	}
 	logging.Debug("oc binary already cached")
 	return true, nil
@@ -67,7 +67,7 @@ func checkOcBinaryCached() (bool, error) {
 func fixOcBinaryCached() (bool, error) {
 	oc := oc.OcCached{}
 	if err := oc.EnsureIsCached(); err != nil {
-		return false, fmt.Errorf("Not able to download oc %v", err)
+		return false, fmt.Errorf("Unable to download oc %v", err)
 	}
 	logging.Debug("oc binary cached")
 	return true, nil
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -155,7 +155,7 @@ func fixLibvirtVersion() (bool, error) {
 }
 
 func checkLibvirtVersion() (bool, error) {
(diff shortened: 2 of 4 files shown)


praveenkumar:
ok to test

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -67,7 +67,7 @@ func checkOcBinaryCached() (bool, error) {
 func fixOcBinaryCached() (bool, error) {
 	oc := oc.OcCached{}
 	if err := oc.EnsureIsCached(); err != nil {
-		return false, fmt.Errorf("Not able to download oc %v", err)
+		return false, fmt.Errorf("Unable to download oc %v", err)
 	}
 	logging.Debug("oc binary cached")
 	return true, nil
```
cfergeau: The changes in this file are unrelated to '\n'

gbraad:
The changes that are unrelated are acceptable

PR #717, opened 2019-10-14
Update bundle version to 4.2.0-rc.5

anjannath:


gbraad:
It would be ideal if only a single change would be needed in future.

Issue #647, opened 2019-09-26
Make explicit which crc binary is used for integration tests


jsliacan:
Currently, the topmost `crc` binary on the `PATH` is used for integration tests (without explicitly defining it anywhere). Make this explicit and/or let the user specify which binary to use with the tests.

jsliacan:
Resolved via #724

PR #724, opened 2019-10-15
Issue #673, #647 Integration tests use specific binary and run with embedded bundle

jsliacan:
1. Addresses Issue #673 so the tests work with embedded bundle
2. Updates how we pass arguments to `make integration`.
3. User can specify the location of the binary they want to use with tests: Issue #647.

More details:
1. Unless the `--bundle-location` flag is used, the bundle is assumed to be embedded.
2. Arguments pertinent to integration tests are now passed via Go flags. The following 4 are defined:
   - `--bundle-location`: default is embedded
   - `--bundle-version`
   - `--pull-secret-file`
   - `--crc-binary`
3. The crc binary is assumed to be called `crc`. The way the binary is specified for the tests is by prepending `PATH` with the user provided binary location. The reason for not wrapping all `crc` commands is greater transparency/readability: this way the commands will be in plain view in the feature files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,11 +122,19 @@ clean: clean_docs
        
 
 .PHONY: integration ## Run integration tests
-integration: GODOG_OPTS = --godog.tags=$(GOOS)
 integration:
-	@$(call check_defined, BUNDLE_LOCATION, "'make integration' requires BUNDLE_LOCATION to contain the full path to a bundle file")
-	@$(call check_defined, PULL_SECRET_FILE, "'make integration' requires PULL_SECRET_FILE to point to a file with the pull secret to use")
-	@go test --timeout=90m $(REPOPATH)/test/integration -v --tags=integration $(GODOG_OPTS) $(BUNDLE_LOCATION) $(PULL_SECRET_FILE)
+GODOG_OPTS = --godog.tags=$(GOOS)
+ifndef PULL_SECRET_FILE
+	PULL_SECRET_FILE = --pull-secret-file=$(HOME)/Downloads/crc-pull-secret
+endif
+ifndef BUNDLE_LOCATION
+	BUNDLE_LOCATION = --bundle-location=$(HOME)/Downloads/crc_libvirt_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
+endif
+ifndef CRC_BINARY
+	CRC_BINARY = --crc-binary=$(GOPATH)/bin
+endif
+integration:
+	@go test --timeout=90m $(REPOPATH)/test/integration -v $(PULL_SECRET_FILE) $(BUNDLE_LOCATION) $(CRC_BINARY) --bundle-version=$(BUNDLE_VERSION) --tags=integration $(GODOG_OPTS)
 
 .PHONY: fmt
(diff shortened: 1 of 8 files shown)


cfergeau:
retest this please

Makefile
```diff
@@ -122,11 +122,19 @@ clean: clean_docs
...
 
 .PHONY: integration ## Run integration tests
-integration: GODOG_OPTS = --godog.tags=$(GOOS)
 integration:
-	@$(call check_defined, BUNDLE_LOCATION, "'make integration' requires BUNDLE_LOCATION to contain the full path to a bundle file")
-	@$(call check_defined, PULL_SECRET_FILE, "'make integration' requires PULL_SECRET_FILE to point to a file with the pull secret to use")
-	@go test --timeout=90m $(REPOPATH)/test/integration -v --tags=integration $(GODOG_OPTS) $(BUNDLE_LOCATION) $(PULL_SECRET_FILE)
+GODOG_OPTS = --godog.tags=$(GOOS)
+ifndef PULL_SECRET_FILE
+	PULL_SECRET_FILE = --pull-secret-file=$(HOME)/Downloads/crc-pull-secret
```
praveenkumar: @jsliacan why we want to hard code this pull secret file instead using the ENV variable?

test/integration/crcsuite/crcsuite.go
```diff
@@ -73,11 +78,71 @@ func FeatureContext(s *godog.Suite) {
...
+			usr, _ := user.Current()
+			CRCBinary = filepath.Join(usr.HomeDir, "go", "bin")
+		}
+
+		// put CRC binary location on top of PATH
+		path := os.Getenv("PATH")
+		var separator string
+		switch platform := runtime.GOOS; platform {
+		case "darwin":
+			separator = ":"
```
praveenkumar: There is already `os.PathListSeparator` so you shouldn't need to manually define it.

Makefile
jsliacan: @praveenkumar I'm not sure I understand, it's only hard coded if ENV variable is not defined. So in practice, it stays as before, except now it also has a default value.

test/integration/crcsuite/crcsuite.go
jsliacan: Thanks! I only found `filepath.SplitList` method. This is useful.

test/integration/features/story_registry.feature
```diff
@@ -41,3 +41,13 @@ Feature: Local image to image-registry to deployment
         When executing "oc logs -f dc/hello" succeeds
         Then stdout should contain "Hello, it works!"
 
+    Scenario: Clean up
+        Given executing "sudo podman images" succeeds
+        When stdout contains "localhost/hello"
+        Then executing "sudo podman image rm localhost/hello:test" succeeds
+        And executing "oc delete project testproj-img" succeeds
+        When executing "crc stop -f" succeeds
+        Then stdout should match "The OpenShift cluster(.*)stopped"
```
cfergeau: This needs to be changed to
```
 Then stdout should match "(.*)[Ss]topped the OpenShift cluster"
```

PR #744, opened 2019-10-18
Issue #741 add message about degraded state

### LINKED ISSUE #741: Cluster reports to be in error state / operators are running in degraded state
Due to recent changes to OpenShift 4.2, the "Cluster Health" will be reported on the Dashboard / console. And looks as follows:

image

and reads: "Red Hat OpenShift Container Platform is in an error state". This message is incorrectly reported for CRC usage. We purposely disable the `monitoring` and `machine-config` operators to preserve resources, however OpenShift itself will therefore conclude the cluster is in a degraded or unhealthy state.

Although we have added documentation around this, a user, newcomer or familiar to CRC, might believe the server is unusable. It is therefore adviced to add a message to the end of the `crc start` output.

gbraad:
Fixes #741 

... not really happy with the message, but something needs to be there

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -66,6 +66,8 @@ func runStart(arguments []string) {
 	}
 	if commandResult.Status == "Running" {
 		output.Outln("The OpenShift cluster is running")
+
+		logging.Warn("The cluster might report a degraded or error state. This is expected since several operators have been disable to lower the resource usage. For more information, please consult the documentation")
 	} else {
 		logging.Warnf("Unexpected status of the OpenShift cluster: %s", commandResult.Status)
 	}


cmd/crc/cmd/start.go
```diff
@@ -66,6 +66,8 @@ func runStart(arguments []string) {
 	}
 	if commandResult.Status == "Running" {
 		output.Outln("The OpenShift cluster is running")
+
+		output.Outln("Note: the cluster might report an error state. This is due to disabling the following operator: 'monitoring' and is expected.")
```
anjannath: It might be better to use a `warn` o/p to make this msg stand out then the rest?

praveenkumar: It is not only monitoring but we also disable the machine config.

gbraad: It does not complain about `machine-config` when it comes to the degraded state, according to feedback.

cmd/crc/cmd/start.go
```diff
@@ -66,6 +66,8 @@ func runStart(arguments []string) {
 	}
 	if commandResult.Status == "Running" {
 		output.Outln("The OpenShift cluster is running")
+
+		logging.Warn("The cluster might report a degraded or error state. This is expected since several operators have been disable to lower the resource usage. For more information, please consult the documentation")
```
praveenkumar: :+1:

PR #745, opened 2019-10-18
Issue #743 Make user checks skippable

### LINKED ISSUE #743: [BUG] User pre-flight checks are not skippable
All pre-flight checks needs to be skippable ...

gbraad:
Fixes #743 allow user checks to be skippable for privilged users that insist on wanting to use Admin/root users

### DIFF
--- a/cmd/crc/cmd/config/config_windows.go
+++ b/cmd/crc/cmd/config/config_windows.go
@@ -6,6 +6,8 @@ import (
 
 var (
 	// Preflight checks
+	SkipCheckAdministratorUser   = cfg.AddSetting("skip-check-administrator-user", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	WarnCheckAdministratorUser   = cfg.AddSetting("warn-check-administrator-user", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	SkipCheckWindowsVersionCheck = cfg.AddSetting("skip-check-windows-version", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	WarnCheckWindowsVersionCheck = cfg.AddSetting("warn-check-windows-version", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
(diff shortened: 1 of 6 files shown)


cfergeau:
If we make this skippable on all platforms, I think I'd just remove the check altogether, as this implies running as non-root does not really matter.
I'm fine with having the check skippable on Windows to avoid a well-defined issue.

gbraad:
We should never prevent a user from being able to start the tool, plus the UI conecpt of "User is in control" applies. When a check is skippable, it does not mean it is irrelevant, we just prevent a deadlock situation in which the user is prevented from progressing.

Although I agree that this change is not as severe on macOS and/or Linux, I do think we need to remain consistent among the plaforms.

cmd/crc/cmd/config/config_windows.go
```diff
@@ -6,6 +6,8 @@ import (
 
 var (
 	// Preflight checks
+	SkipCheckAdministratorUser   = cfg.AddSetting("skip-check-administrator-user", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	WarnCheckAdministratorUser   = cfg.AddSetting("warn-check-administrator-user", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
```
praveenkumar: :+1:

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -8,10 +8,10 @@ import (
 
 // StartPreflightChecks performs the preflight checks before starting the cluster
 func StartPreflightChecks(vmDriver string) {
-	preflightCheckSucceedsOrFails(false,
+	preflightCheckSucceedsOrFails(config.GetBool(cmdConfig.SkipCheckRootUser.Name),
 		checkIfRunningAsNormalUser,
 		"Checking if running as non-root",
-		false,
+		config.GetBool(cmdConfig.WarnCheckRootUser.Name),
```
praveenkumar: :+1:

pkg/crc/preflight/preflight_windows.go
```diff
@@ -7,10 +7,10 @@ import (
 
 // StartPreflightChecks performs the preflight checks before starting the cluster
 func StartPreflightChecks(vmDriver string) {
-	preflightCheckSucceedsOrFails(false,
+	preflightCheckSucceedsOrFails(config.GetBool(cmdConfig.SkipCheckAdministratorUser.Name),
 		checkIfRunningAsNormalUserInWindows,
 		"Checking if running as normal user",
-		false,
+		config.GetBool(cmdConfig.WarnCheckAdministratorUser.Name),
```
praveenkumar: :+1:

PR #759, opened 2019-10-23
Issue #758 Force encoding for GetDefaultSwitchName

### LINKED ISSUE #758: Default Switch using special characters might fail start (VM creation in driver)
In Swedish the Default Switch is named "Standardväxel" which might cause issues upon creation of the VM.

image

```
INFO Creating CodeReady Containers VM for OpenShift 4.2.0-0.nightly-2019-09-26-192831...

ERRO Error occurred: Error creating host: Error creating the VM. Error with pre-create check: "virtual switch \"Standardv�xel\" not found"
```

gbraad:
Fixes #758

### DIFF
--- a/pkg/os/windows/network/vswitch_windows.go
+++ b/pkg/os/windows/network/vswitch_windows.go
@@ -10,7 +10,7 @@ import (
 const hypervDefaultVirtualSwitchId = "c08cb7b8-9b3c-408e-8e30-5e16a3aeb444"
 
 func GetDefaultSwitchName() (bool, string) {
-	getDefaultSwitchNameCmd := fmt.Sprintf("Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", hypervDefaultVirtualSwitchId)
+	getDefaultSwitchNameCmd := fmt.Sprintf("[Console]::OutputEncoding = [Text.Encoding]::UTF8; Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", hypervDefaultVirtualSwitchId)
 	stdOut, stdErr, _ := powershell.Execute(getDefaultSwitchNameCmd)
 
 	// If stdErr contains the command then execution failed


pkg/os/windows/network/vswitch_windows.go
```diff
@@ -10,7 +10,7 @@ import (
 const hypervDefaultVirtualSwitchId = "c08cb7b8-9b3c-408e-8e30-5e16a3aeb444"
 
 func GetDefaultSwitchName() (bool, string) {
-	getDefaultSwitchNameCmd := fmt.Sprintf("Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", hypervDefaultVirtualSwitchId)
+	getDefaultSwitchNameCmd := fmt.Sprintf("[Console]::OutputEncoding = [Text.Encoding]::UTF8; Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", hypervDefaultVirtualSwitchId)
 	stdOut, stdErr, _ := powershell.Execute(getDefaultSwitchNameCmd)
```
cfergeau: Is this output specification something we can/should do in powershell.Execute rather than for each command?

praveenkumar:
retest this please

gbraad:
keeps failing on CI

praveenkumar:
retest this please

pkg/os/windows/network/vswitch_windows.go
gbraad: we should ideally have a session in use, so we can execute them in succession and have a setup for the session that would handle this.

anjannath: This can be done as an improvement task, for now this fixes the default switch issue.

praveenkumar: @cfergeau if you agree then we can merge it so it can be out with this release.

cfergeau: Dunno, something like this (untested..) seems better to me, I don't think we ever want non-utf8 output from powershell
```
iff --git a/pkg/os/windows/powershell/powershell_windows.go b/pkg/os/windows/powershell/powershell_windows.go
index d6af4463..71054ccc 100644
--- a/pkg/os/windows/powershell/powershell_windows.go
+++ b/pkg/os/windows/powershell/powershell_windows.go
@@ -32,6 +32,7 @@ var (
                "$currentPrincipal = New-Object Security.Principal.WindowsPrincipal([Security.Principal.WindowsIdentity]::GetCurrent())",
                "$currentPrincipal.IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)",
        }
+       utf8Encoding = "[Console]::OutputEncoding = [Text.Encoding]::UTF8"
 )
 
 func LocatePowerShell() string {
@@ -54,6 +55,7 @@ func IsAdmin() bool {
 
 func Execute(args ...string) (stdOut string, stdErr string, err error) {
        args = append([]string{"-NoProfile", "-NonInteractive", "-ExecutionPolicy", "RemoteSigned", "-Command"}, args...)
+       args[0] = strings.Join(utf8Encoding, args[0], ";")
        cmd := exec.Command(LocatePowerShell(), args...) // #nosec G204
        cmd.SysProcAttr = &syscall.SysProcAttr{HideWindow: true}
 ```

gbraad:
@praveenkumar merge? should be in this release ...

PR #792, opened 2019-11-06
Fix 2 doc related bug

cfergeau:
The initial paragraph in the 'oc debug nodes/...' commit will most likely need to be rewritten :)

### DIFF
--- a/docs/source/getting-started/content/proc_basic-troubleshooting.adoc
+++ b/docs/source/getting-started/content/proc_basic-troubleshooting.adoc
@@ -50,6 +50,24 @@ If your issue has not been resolved by this procedure, perform the following ste
 . If no existing issue addresses the encountered issue, link:https://github.com/code-ready/crc/issues/new[create an issue] and link:https://help.github.com/en/articles/file-attachments-on-issues-and-pull-requests[attach the [filename]`~/.crc/crc.log` file] to the created issue.
 The [filename]`~/.crc/crc.log` file contains detailed debugging and troubleshooting information which can help diagnose the problem that you are experiencing.
 
+= Getting shell access to the OpenShift cluster
+
+Direct access to the OpenShift cluster is not needed for regular cluster use, and is strongly discouraged.
+If such access is needed for troubleshooting or debugging purposes, follow these steps:
+
+  . Enable `oc` access to the cluster, logging in as `kubeadmin`. For detailed steps, see <<accessing-the-openshift-cluster_{context}>>.
+
+  . Run [command]`oc get nodes`. The output will be similar to this:
++
+[subs="+quotes,attributes",options="nowrap"]
+----
+$ oc get nodes
+NAME                 STATUS   ROLES           AGE    VERSION
+crc-shdl4-master-0   Ready    master,worker   7d7h   v1.14.6+7e13ab9a7
+----
+
+  . Run [command]`oc debug nodes/crc-shdl4-master-0` where `crc-shdl4-master-0` is the name of the node which was printed at the previous step.
+
 = DNS configuration details
 
 This section describes how {prod} configures networking.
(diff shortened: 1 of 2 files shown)


docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -50,6 +50,29 @@ If your issue has not been resolved by this procedure, perform the following ste
 . If no existing issue addresses the encountered issue, link:https://github.com/code-ready/crc/issues/new[create an issue] and link:https://help.github.com/en/articles/file-attachments-on-issues-and-pull-requests[attach the [filename]`~/.crc/crc.log` file] to the created issue.
 The [filename]`~/.crc/crc.log` file contains detailed debugging and troubleshooting information which can help diagnose the problem that you are experiencing.
 
+= Getting shell access to the OpenShift cluster
+
+One of the core premises of OpenShift 4 (which is based on Red Hat CoreOS) is that one should not need to make direct changes to the cluster nodes.
+Instead, OpenShift operators will be doing all the needed changes on the nodes when you deploy a service or an application.
+Direct access to the OpenShift cluster is thus not needed in regular use, and strongly discouraged.
```
robin-owen: I believe this is all that we need to say in our documentation. If we need to include information about OpenShift 4/Red Hat CoreOS and their philosophy, we should do so earlier in the docs (perhaps in the Understanding CodeReady Containers section).

In short, I don't think the previous two sentences are necessary, and if they are, they belong elsewhere. :)

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -50,6 +50,29 @@ If your issue has not been resolved by this procedure, perform the following ste
...
+
+  . Run [command]`oc get nodes`. The output will be similar to this:
++
+[subs="+quotes,attributes",options="nowrap"]
+----
+$ oc get nodes
+NAME                 STATUS   ROLES           AGE    VERSION
+crc-shdl4-master-0   Ready    master,worker   7d7h   v1.14.6+7e13ab9a7
+----
++
```
robin-owen: The extra `+` is not needed here.

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -50,6 +50,29 @@ If your issue has not been resolved by this procedure, perform the following ste
...
+
+  . Run [command]`oc debug nodes/crc-shdl4-master-0` where `crc-shdl4-master-0` is the name of the node which was printed at the previous step.
+
+  . You now have a shell running on the {rh-prod} node.
```
robin-owen: This isn't actually a step, and as such can be removed. :)

docs/source/getting-started/content/proc_basic-troubleshooting.adoc
```diff
@@ -50,6 +50,29 @@ If your issue has not been resolved by this procedure, perform the following ste
...
++
+[subs="+quotes,attributes",options="nowrap"]
+----
+$ oc get nodes
+NAME                 STATUS   ROLES           AGE    VERSION
+crc-shdl4-master-0   Ready    master,worker   7d7h   v1.14.6+7e13ab9a7
+----
++
+
+  . Run [command]`oc debug nodes/crc-shdl4-master-0` where `crc-shdl4-master-0` is the name of the node which was printed at the previous step.
```
robin-owen: I wonder if we'd be better off supplying a generic name for the node here rather than an explicit one which is inferred to be replaceable. Does that make sense?

cfergeau: Nope, I'm not sure what you mean :(
Use `<node>` for example instead of `crc-shdl4-master-0` and explain that the value of <node> comes from the previous command? Or do you mean something else?

cfergeau: I've pushed a new iteration with all the comments addressed except for this one, I'll push another series once what you expect is clearer to me :)

robin-owen: Honestly, we're probably fine with this as it's written. I just wonder whether it's a good idea or not to use the node name explicitly, but this may be clearer as-is than using a generic name would be. :)

Issue #802, opened 2019-11-11
[Integration] Update config tests


jsliacan:
Skips and Warns have changed since the tests were written. Some other configurable properties were added in the meantime too. Tests need to be updated.

jsliacan:
Resolved by #803

PR #803, opened 2019-11-11
Issue #802 Integration tests: update config feature

jsliacan:


test/integration/features/config.feature
```diff
@@ -9,67 +9,99 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
...
+        And unsetting config property "disable-update-check" succeeds
+
 # WARNINGS
 
     Scenario Outline: CRC config checks (warnings)
         When setting config property "<property>" to value "<value1>" succeeds
         Then "JSON" config file "crc.json" in CRC home folder contains key "<property>" with value matching "<value1>"
         When setting config property "<property>" to value "<value2>" succeeds
         Then "JSON" config file "crc.json" in CRC home folder contains key "<property>" with value matching "<value2>"
+        And unsetting config property "<property>" succeeds
```
cfergeau: `And "JSON" config file "crc.json" in CRC home folder does not contain key "<property>"`
Can this statement work at the moment? Or would it need some changes in crcsuite.go?

test/integration/features/config.feature
```diff
@@ -9,67 +9,99 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
...
     Scenario: CRC config checks (bundle version)
         When setting config property "bundle" to value "current bundle" succeeds
         And "JSON" config file "crc.json" in CRC home folder contains key "bundle" with value matching "current bundle"
         And setting config property "bundle" to value "/path/to/nonexistent/bundle/crc_hypervisor_version.tar.xz" fails
 
+    @linux @darwin @windows
+    Scenario: CRC config checks (update check)
+        Given setting config property "disable-update-check" to value "true" succeeds
+        When executing "crc setup" succeeds
+        Then stdout should not contain "new version"
```
cfergeau: Shouldn't we never get 'new version' in the output? Integration will most of the time test either a new release, or a git version, in both cases there will not be any newer version available even without setting the configuration key?
I think right now there's a bug where versions built from git output this warning, but if we fix it, then this check will always succeed, so I don't know how good it is to add it?

jsliacan: > I think right now there's a bug where versions built from git output this warning
Ah, that's a bug. Good :). I can remove it then, true.

test/integration/features/config.feature
jsliacan: Thanks. I assumed I hadn't implemented the `does not contain` option in https://github.com/code-ready/clicumber/blob/02a92c8f7eceabab30fac8c66ef7d93c6a2c3f5c/testsuite/testsuite.go#L139, but it's there. Thanks. I now updated the feature file to check the crc.json file after `crc config unset`.

test/integration/features/config.feature
cfergeau: I consider this to be a bug, I don't know what others in the team think about it :)

gbraad: it is a bug... git versions should never be checked against the released versions.
this was reported before, but forgot to add an issue... sorry

cfergeau: https://github.com/code-ready/crc/issues/806

Issue #807, opened 2019-11-14
[Integration] Wrong type for os.PathListSeparator when constructing new PATH


jsliacan:
The Go's `os.PathListSeparator` is a `rune` type, whereas [test/integration/crcsuite/crcsuite.go#L100](https://github.com/code-ready/crc/blob/f7e8581f31a1e5e32accfd8dd3a0d1414560477c/test/integration/crcsuite/crcsuite.go#L100) assumes it is a string:
```
newPath := fmt.Sprintf("%s%s%s", CRCBinary, os.PathListSeparator, path)
```
This causes issues on RHEL7/8 and MacOS.

jsliacan:
Resolved by #808

PR #808, opened 2019-11-14
Issue #807 Integration tests: Print PathListSeparator as a char

jsliacan:
Resolves Issue #807.

### DIFF
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -97,7 +97,7 @@ func FeatureContext(s *godog.Suite) {
 
 		// put CRC binary location on top of PATH
 		path := os.Getenv("PATH")
-		newPath := fmt.Sprintf("%s%s%s", CRCBinary, os.PathListSeparator, path)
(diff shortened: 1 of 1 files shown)


test/integration/crcsuite/crcsuite.go
```diff
@@ -97,7 +97,7 @@ func FeatureContext(s *godog.Suite) {
 
 		// put CRC binary location on top of PATH
 		path := os.Getenv("PATH")
-		newPath := fmt.Sprintf("%s%s%s", CRCBinary, os.PathListSeparator, path)
+		newPath := fmt.Sprintf("%s%s%s", CRCBinary, string(os.PathListSeparator), path)
```
cfergeau: have you tried `%c` and no cast?

jsliacan: `%c` works too, yes. Is it better to do it that way?

cfergeau: It seems less convoluted this way, we use the correct printf modifier for the type we want to print, rather than converting the value to string, and then printing it.

PR #819, opened 2019-11-19
Issue #722 Allow to use 'crc' named Virtual Switch as alternative

### LINKED ISSUE #722: Handle possible alternative network/virtual switch for use with Hyper-V
Not all editions of Windows 10 have the Default Switch, such as Enterprise, but also in some case people want to use a network that allows remote access. In that case, the Default Switch is either not available or not usable.

The idea would be to add a `crc` named Virtual Switch can be used as a primary option when available. it has not been deciced if this should be behind a config option or not, but I'd rather want, just like in the case of libvirt, to have a statically named option that can be used, and hav ethe "Default Switch" act as the default/fallback when not found.

gbraad:
Fixes #722 by allowing an alternative named Virtual Switch to be used. Startup will report back the name that is used;

```
PS C:\Users\gbraad\Downloads> .\crc.exe setup
INFO Caching oc binary
INFO Check Windows 10 release
INFO Hyper-V installed
INFO Is user a member of the Hyper-V Administrators group
INFO Does the Hyper-V virtual switch exist
Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster
```


gbraad:
Just for the message:
```
INFO Does the Hyper-V virtual switch exist
```

that logs the name of the VSwitch this change is already important, as it convery a lot of meaning to show what is being used.

anjannath:
If using the `crc` switch then we also have to set the Dns server for it to the VM ip address during post start.

anjannath:
https://github.com/code-ready/crc/blob/master/pkg/crc/services/dns/dns_windows.go#L26 should call `winnet.SelectSwitchByNameOrDefault(switchName)` to get the appropriate switch.

gbraad:
I had fixed this, but seems it refuses to push this (all up to date?)
Will repush

gbraad:
@anjannath @praveenkumar updated ...

pkg/crc/machine/hyperv/driver_windows.go
```diff
@@ -18,7 +18,8 @@ func CreateHost(machineConfig config.MachineConfig) *hyperv.Driver {
 	hypervDriver.BundleName = machineConfig.BundleName
 	hypervDriver.Memory = machineConfig.Memory
 
-	_, switchName := winnet.GetDefaultSwitchName()
+	// Determine the Virtual Switch to be used
+	_, switchName := winnet.SelectSwitchByNameOrDefault(AlternativeNetwork)
```
anjannath: :+1:

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -126,25 +127,13 @@ func fixUserPartOfHyperVAdmins() (bool, error) {
...
-			logging.Debug(err.Error())
-			return false, errors.New("Failed checking if Hyper-V Default Switch exists")
-		}
-
-		if !strings.Contains(stdErr, "Get-VMSwitch") {
-			// found the default
-			return true, nil
-		} else {
-			return false, errors.New("Incorrect permissions")
-		}
```
anjannath: :+1:

pkg/os/windows/network/vswitch_windows.go
```diff
@@ -9,6 +9,31 @@ import (
...
+	if strings.Contains(stdErr, "Get-VMSwitch") {
+		return false
+	}
+
+	if strings.Contains(stdOut, name) {
+		return true
+	}
+	return false
+}
+
```
anjannath: :+1:

anjannath:
Centos CI failure is not related to the PR

gbraad:
retest this please

gbraad:
Thanks ... will contact @robin-owen to ad this to the release notes ...

Issue #786, opened 2019-11-05
[Integration] Add an app deployment from marketplace.


praveenkumar:
We need to create integration test for marketplace, there are few operators which are present and can be used for a simple integration test for CI.

jsliacan:
Solved by #823

PR #823, opened 2019-11-20
Issue #786 Integration: Install operator from marketplace user story

jsliacan:


praveenkumar:
retest this please.

jsliacan:
retest this please

test/integration/crcsuite/prepare.go
```diff
@@ -56,6 +58,60 @@ func DownloadBundle(bundleLocation string, bundleDestination string) (string, er
...
+func CopyFilesToTestDir() {
+
+	pwd, err := os.Getwd()
+	if err != nil {
+		fmt.Printf("Error retrieving current dir: %s", err)
+		os.Exit(1)
+	}
+	pDir := path.Dir(pwd)
+	pDir = path.Dir(pDir)
+	pDir = path.Dir(pDir)
```
praveenkumar: @jsliacan What you want to simulate here ` cd ../..` like this?

jsliacan: Exactly. I don't know the absolute path of the directory, and I don't want to use `/` as this should work on all platforms. Didn't find a way to go up the tree by _n_ steps at once... any suggestions welcome :)

praveenkumar: @jsliacan How about https://play.golang.org/p/Tkz7vPJ5rNq one?

test/integration/crcsuite/prepare.go
```diff
@@ -56,6 +58,60 @@ func DownloadBundle(bundleLocation string, bundleDestination string) (string, er
...
+
+	for _, file := range files {
+
+		sFileName := filepath.Join(dataDir, file.Name())
+		destLoc, _ := os.Getwd()
+		fmt.Printf("Copying %s to %s\n", sFileName, destLoc)
+
+		sFile, err := os.Open(sFileName)
+		if err != nil {
+			fmt.Printf("Error occured opening file: %s", err)
```
cfergeau: 'occurred' (multiple times)

test/integration/crcsuite/prepare.go
```diff
@@ -56,6 +58,60 @@ func DownloadBundle(bundleLocation string, bundleDestination string) (string, er
...
+	files, err := ioutil.ReadDir(dataDir)
+	if err != nil {
+		fmt.Printf("Error occured loading data files: %s", err)
+		os.Exit(1)
+	}
+
+	for _, file := range files {
+
+		sFileName := filepath.Join(dataDir, file.Name())
+		destLoc, _ := os.Getwd()
```
cfergeau: This is not going to change while looping over the files, is it? This could be moved out of the loop.

test/integration/crcsuite/prepare.go
```diff
@@ -56,6 +58,60 @@ func DownloadBundle(bundleLocation string, bundleDestination string) (string, er
...
+		fmt.Printf("Copying %s to %s\n", sFileName, destLoc)
+
+		sFile, err := os.Open(sFileName)
+		if err != nil {
+			fmt.Printf("Error occured opening file: %s", err)
+			os.Exit(1)
+		}
+		defer sFile.Close()
+
+		_, dFileName := filepath.Split(sFileName)
```
cfergeau: isn't this `dFileName := file.Name()`?

test/integration/crcsuite/prepare.go
cfergeau: Maybe we could pass the path to testdata/ to the tests somehow rather than hardcoding it's `$(getcwd)/../../..`?

test/integration/crcsuite/prepare.go
jsliacan: Good point, `destLoc` is independent of the `file`. Thanks

test/integration/crcsuite/prepare.go
cfergeau: > @jsliacan How about https://play.golang.org/p/Tkz7vPJ5rNq one?

`l := strings.Split(s, string(filepath.Separator))` can use `filepath.SplitList` instead.

praveenkumar: @cfergeau I think `SplitList` is for path list separator not for a single path. https://golang.org/pkg/path/filepath/#SplitList

jsliacan: I think `filepath.SplitList` splits a list of paths not a path (e.g. based on `:` for Unix): https://golang.org/pkg/path/filepath/#SplitList.  As for passing the testdata dir to `make integration` -- tempting, but I worry that we're already passing way too many things. Plus this info is deduceable, just not in a nice way it seems. Thanks though, I'll change it to something and we'll see.

cfergeau: Oops, sorry, read too quickly the go doc :)

jsliacan: I changed it, but it's not much nicer than it was! Hmmm...

jsliacan:
Right now the CentosCI is being killed by the timout of 90m in `minishift-ci-jobs` [here](https://github.com/minishift/minishift-ci-jobs/blob/4610c14778c601441289adba1f10abe98d1b74ef/minishift-ci-index.yaml#L532). Should we increase it to ~120? We might have to decide on a subset of tests for PR checks... 2h is quite a bit of time. WDYT?

praveenkumar:
> Right now the CentosCI is being killed by the timeout of 90m in minishift-ci-jobs here. Should we increase it to ~120? 

@jsliacan Sure, go ahead with that.

jsliacan:
retest this please

Issue #828, opened 2019-11-21
Improve CSR handling


cfergeau:
At the end of 'crc start' we need to approve CSRs. However, we do this after displaying the message which says that the cluster is ready, which is odd as this still takes a bit of time.
We are also approving all CSRs, including ones which were already approved, which makes things slower than they could

PR #829, opened 2019-11-21
Improve CSR handling

cfergeau:
This PR fixes https://github.com/code-ready/crc/issues/828

### DIFF
--- a/pkg/crc/oc/clusteroperator.go
+++ b/pkg/crc/oc/clusteroperator.go
@@ -9,7 +9,7 @@ import (
 
 var ignoreClusterOperators = []string{"monitoring", "machine-config", "marketplace"}
 
-type ClusterOperator struct {
+type K8sResource struct {
 	Items []struct {
 		Metadata struct {
 			Name string `json:"name"`
@@ -32,7 +32,7 @@ func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
 		return false, fmt.Errorf("%s - %v", stderr, err)
 	}
 
-	var co ClusterOperator
+	var co K8sResource
 
 	err = json.Unmarshal([]byte(data), &co)
 	if err != nil {
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -32,7 +32,7 @@ func CheckCertsValidityUsingBundleBuildTime(buildTime time.Time) (bool, int) {
 
 // CheckCertsValidity checks if the cluster certs have expired or going to expire in next 7 days
 func CheckCertsValidity(sshRunner *ssh.SSHRunner) (bool, int, error) {
-	certExpiryDate, err := getcertExipryDateFromVM(sshRunner)
+	certExpiryDate, err := getcertExpiryDateFromVM(sshRunner)
 	if err != nil {
 		return false, 0, err
 	}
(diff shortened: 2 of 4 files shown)


praveenkumar:
retest this please

pkg/crc/oc/clusteroperator.go
```diff
@@ -9,7 +9,7 @@ import (
 
 var ignoreClusterOperators = []string{"monitoring", "machine-config", "marketplace"}
 
-type ClusterOperator struct {
+type K8sResource struct {
```
praveenkumar: `K8sResource` is kind of a very generalize way and can have anything (like pod, deployment, rs ..etc.) , If I really want to change name then I would always include which type of k8s resource is this one.

cfergeau: This `ClusterOperator` structure is generic enough to match what is need for `Csr`, hence the more generic name. I don't know how to call it otherwise? I can change this to
```
type Csr ClusterOperator
type ClusterOperator struct {
 ...
}
```
I suspect we might be reinventing struct already defined in https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go for example, might be worth using the kubernetes API directly at some point rather than going through `oc`.

praveenkumar: > I suspect we might be reinventing struct already defined in https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go for example, might be worth using the kubernetes API directly at some point rather than going through oc

I think, in past we did try that and found out it is k8s dep is too heavy to include in the vendor directory than just parsing some of the `oc` command output.

praveenkumar: > This ClusterOperator structure is generic enough to match what is need for Csr, hence the more generic name. I don't know how to call it otherwise? I can change this to

What's wrong with following
```
var csrs K8sResource
```
to 
```
var csrs ClusterOperator
```

cfergeau: > I think, in past we did try that and found out it is k8s dep is too heavy to include in the vendor directory than just parsing some of the `oc` command output.

This means instead of having k8s in vendor/, we have k8s hidden in the `oc` binary, I don't think this really is different, just hidden :)

cfergeau: > What's wrong with following
> 
> ```
> var csrs K8sResource
> ```
> 
> to
> 
> ```
> var csrs ClusterOperator
> ```

Why are we using something named `ClusterOperator` in order to parse `oc get csr -o yaml` output? This would be confusing/misleading.

praveenkumar: > This means instead of having k8s in vendor/, we have k8s hidden in the oc binary, I don't think this really is different, just hidden :)

This is different `oc` binary we anyway required to all the kind of operation and also after the cluster is running, what you are suggesting here is completely drop  oc commands an just use the k8s apis (even with login or any kind of stuff) this is what odo do atm. I am not sure how much effort will it take for us to replace all the oc commands with the API endpoints.

praveenkumar: > Why are we using something named ClusterOperator in order to parse oc get csr -o yaml output? This would be confusing/misleading.

yes, I read again so  k8sresource struct same here for clusteroperator and csr so it is just reused 👍

PR #830, opened 2019-11-21
Fix centosci

praveenkumar:


centos_ci.sh
```diff
@@ -79,6 +79,11 @@ function prepare_ci_user() {
 }
 
 function setup_golang() {
+  # Install the required version of golang
+  curl -L -O https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz
+  tar -C /usr/local -xzf go${GO_VERSION}.linux-amd64.tar.gz
+  export PATH=$PATH:/usr/local/go/bin
+
```
cfergeau: The commit log does not explain why we cannot use the epel version. Might be worthwhile to file an epel bug as well?

praveenkumar: @cfergeau I am going to use same go version as we are using for other CI which is 1.12.x and which is part of internal build system.

centos_ci.sh
```diff
@@ -3,6 +3,7 @@
 # bundle location
 BUNDLE_VERSION=4.2.7
 BUNDLE=crc_libvirt_$BUNDLE_VERSION.crcbundle
+GO_VERSION=1.12.8
```
cfergeau: If we pick an upstream release rather than something curated by a distro, this really should be the latest one, so 1.12.13 in order to benefit from any important bug fix/CVE fix which was done since 1.12.8.

praveenkumar:
@cfergeau if you have anything other suggestion then let me know or just merge it.

PR #831, opened 2019-11-21
Implement cert renewal

cfergeau:
This implements automatic renewal of certificates when they are expired. I'm only marking this as a draft for now as I'm still making some minor adjustment to it (rework the oc interface commit, there might still be some debug leftovers, I need to test this some more, ...), but given the timezone differences, this will give other people from the team a chance to take an early look. On my side, I don't expect big changes to the last 2 commits which are the actual implementation of the feature.

### DIFF
--- a/pkg/crc/machine/cert-renewal.go
+++ b/pkg/crc/machine/cert-renewal.go
@@ -1,0 +1,251 @@
+package machine
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/oc"
+	"github.com/code-ready/crc/pkg/crc/ssh"
+	"github.com/code-ready/crc/pkg/crc/systemd"
+	"github.com/code-ready/crc/pkg/crc/version"
+)
+
+var (
+	releaseImage = fmt.Sprintf("quay.io/openshift-release-dev/ocp-release:%s", version.GetBundleVersion())
+)
+
+func getRecoveryKubeConfig(output string) (string, error) {
+	pattern, err := regexp.Compile(`(?m)export KUBECONFIG=([\w-/.]*)`)
+	if err != nil {
+		return "", err
+	}
+	recoveryKubeconfig := []byte{}
+	for _, submatches := range pattern.FindAllStringSubmatchIndex(output, -1) {
+		recoveryKubeconfig = pattern.ExpandString(recoveryKubeconfig, "$1", output, submatches)
+	}
+	if len(recoveryKubeconfig) == 0 {
+		return "", fmt.Errorf("Could not parse recovery kubeconfig path")
+	}
+
+	return string(recoveryKubeconfig), nil
+}
+
+type recoveryPod struct {
+	sshRunner    *ssh.SSHRunner
+	releaseImage string
+	kaoImage     string
+	kubeconfig   string
+}
+
+func (recoveryPod recoveryPod) GetKubeconfigPath() string {
+	return recoveryPod.kubeconfig
+}
+
+func (recoveryPod recoveryPod) Run(args ...string) (string, string, error) {
+	cmd := fmt.Sprintf("sudo oc %s", strings.Join(args, " "))
+	logging.Debugf("kube-apiserver-operator: Running %s", cmd)
+	stdout, err := recoveryPod.sshRunner.Run(cmd)
+	return stdout, "", err
+}
+
+func (recoveryPod *recoveryPod) runPodCommand(cmd string) (string, error) {
+	podmanCmd := fmt.Sprintf("sudo podman run -it --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator '%s'", recoveryPod.kaoImage)
+	podmanCmd = podmanCmd + " " + cmd
+	return recoveryPod.sshRunner.Run(podmanCmd)
+}
+
+func startRecoveryPod(sshRunner *ssh.SSHRunner) (*recoveryPod, error) {
+	recoveryPod := recoveryPod{}
+	recoveryPod.sshRunner = sshRunner
+	recoveryPod.releaseImage = releaseImage
+
+	logging.Debugf("Using release image %s", recoveryPod.releaseImage)
+	kaoImage, err := sshRunner.Run(fmt.Sprintf("sudo oc adm release info --registry-config='/var/lib/kubelet/config.json' '%s' --image-for=cluster-kube-apiserver-operator", releaseImage))
+	if err != nil {
+		return nil, err
+	}
+	recoveryPod.kaoImage = strings.TrimSpace(kaoImage)
+	logging.Debugf("kube-apiserver-operator image: %s", kaoImage)
+
+	_, err = sshRunner.Run(fmt.Sprintf("sudo podman pull --authfile='/var/lib/kubelet/config.json' %s", kaoImage))
+	if err != nil {
+		return nil, err
+	}
+	output, err := recoveryPod.runPodCommand("recovery-apiserver create")
+	if err != nil {
+		return nil, err
+	}
+
+	logging.Debug("Recovery pod is running")
+
+	recoveryPod.kubeconfig, err = getRecoveryKubeConfig(output)
+	if err != nil {
+		return nil, err
+	}
+	logging.Debugf("Recovery KUBECONFIG: [%s]", recoveryPod.kubeconfig)
+
+	return &recoveryPod, nil
+}
+
+func waitForApiServer(machineName string) error {
+	localOc := oc.UseOCWithConfig(machineName)
+	waitForApiServer := func() error {
+		_, stderr, err := localOc.RunOcCommand("get", "node")
+		if err != nil {
+			logging.Debugf("oc invocation failed: %s", stderr)
+			return &errors.RetriableError{Err: err}
+		}
+		return nil
+	}
+
+	return errors.RetryAfter(300, waitForApiServer, time.Second)
+}
+
+func waitForRecoveryPod(oc oc.OcConfig) error {
+	waitForRecoveryPod := func() error {
+		_, _, err := oc.RunOcCommand("get", "namespace", "kube-system")
+		if err != nil {
+			return &errors.RetriableError{Err: err}
+		}
+		return nil
+	}
+	return errors.RetryAfter(60, waitForRecoveryPod, time.Second)
+}
+
+func waitForPendingCsrs(oc oc.OcConfig) error {
+	waitForPendingCsr := func() error {
+		output, _, err := oc.RunOcCommand("get", "csr")
+		if err != nil {
+			return &errors.RetriableError{Err: err}
+		}
+		matched, err := regexp.MatchString("Pending", output)
+		if err != nil {
+			return &errors.RetriableError{Err: err}
+		}
+		if !matched {
+			return &errors.RetriableError{Err: fmt.Errorf("No Pending CSR")}
+		}
+		return nil
+	}
+
+	return errors.RetryAfter(60, waitForPendingCsr, time.Second)
+}
+
+func RegenerateCertificates(sshRunner *ssh.SSHRunner, machineName string) error {
+	/* FIXME: Need to error out if openshift version is older than 4.1.17 */
+
+	sd := systemd.NewInstanceSystemdCommander(sshRunner)
+	startedKubelet, err := sd.Start("kubelet")
+	if err != nil {
+		logging.Debugf("Error starting kubelet service: %v", err)
+		return err
+	}
+	if startedKubelet {
+		defer sd.Stop("kubelet") //nolint:errcheck
+	}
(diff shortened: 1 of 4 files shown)


pkg/crc/machine/cert-renewal.go
```diff
@@ -0,0 +1,215 @@
...
+
+func (recoveryPod recoveryPod) Run(args ...string) (string, string, error) {
+	cmd := fmt.Sprintf("sudo oc %s", strings.Join(args, " "))
+	logging.Debugf("Running %s", cmd)
+	logging.Warnf("Running %s", cmd)
+	stdout, err := recoveryPod.sshRunner.Run(cmd)
+	return stdout, "", err
+}
+
+func (recoveryPod *recoveryPod) runPodCommand(cmd string) (string, error) {
```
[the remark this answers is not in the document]
cfergeau: Naming can probably be improved, but I don't think `runPodmanCommand` is a good name. The `recoveryPod` struct is an abstration over the `cluster-kube-apiserver-operator`. In order to run it/interact with it, you go through `podman` but imo that's an implementation detail. What matters is that you can pass commands to this process (`create`, `destroy`, `regenerate-certificates`), and also that it provides you with a `kubeconfig` file to use to interact with the recovery api server.

pkg/crc/machine/cert-renewal.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	forceRedeploymentPatch := fmt.Sprintf(`'{"spec": {"forceRedeploymentReason": "recovery-%s"}}'`, time.Now().Format(time.RFC3339))
+	for _, resourceType := range []string{"kubeapiserver", "kubecontrollermanager", "kubescheduler"} {
+		_, _, err := oc.RunOcCommand(recoveryPod,
+			fmt.Sprintf("patch %s cluster", resourceType),
+			fmt.Sprintf("--patch=%s", forceRedeploymentPatch),
+			"--type=merge")
+		if err != nil {
+			logging.Warnf("Error forcing redeployment of %s: %v", resourceType, err)
```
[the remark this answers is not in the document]
cfergeau: Thanks, I'll add it.

pkg/crc/machine/cert-renewal.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		logging.Debugf("Error waiting for second pending CSR: %v", err)
+		return err
+	}
+	err = oc.ApproveNodeCSR()
+	if err != nil {
+		logging.Debugf("Error approving second pending CSR: %v", err)
+		return err
+	}
+
```
cfergeau: Once this is done, we need to wait some more or the next steps won't work. It's not clear what we should be waiting on though nor how long :( this is why it's not in this iteration yet.

praveenkumar:
@cfergeau FYI, below is what I did for `cert-renewal.go` file and tested and it does as expected.

```golang
package machine

import (
	"encoding/base64"
	"fmt"
	"github.com/code-ready/crc/pkg/crc/ssh"
)

var (
	certRotationScriptTemplate = `#!/bin/bash
# This script follows https://docs.openshift.com/container-platform/4.1/backup_and_restore/disaster_recovery/scenario-3-expired-certs.html
# https://github.com/openshift/openshift-docs/blob/master/modules/dr-recover-expired-control-plane-certs.adoc

set -x

export KAO_IMAGE="openshift/cert-recovery"
RECOVERY_OUTPUT=$(podman run -it --rm --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" recovery-apiserver create)
KUBECONFIG=$(echo $RECOVERY_OUTPUT|grep "export KUBECONFIG"|tail -1 | sed "s/^.*export KUBECONFIG=//" | tr -d '\r\n')

export KUBECONFIG

until oc get namespace kube-system ; do echo 'Waiting for recovery apiserver to come up.'; sleep 1; done

podman run -it --rm --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" regenerate-certificates

oc patch kubeapiserver cluster -p='{"spec": {"forceRedeploymentReason": "recovery-'"$( date --rfc-3339=ns )"'"}}' --type=merge
oc patch kubecontrollermanager cluster -p='{"spec": {"forceRedeploymentReason": "recovery-'"$( date --rfc-3339=ns )"'"}}' --type=merge
oc patch kubescheduler cluster -p='{"spec": {"forceRedeploymentReason": "recovery-'"$( date --rfc-3339=ns )"'"}}' --type=merge

/usr/local/bin/recover-kubeconfig.sh >kubeconfig
mv kubeconfig /etc/kubernetes/kubeconfig

oc get configmap kube-apiserver-to-kubelet-client-ca -n openshift-kube-apiserver-operator --template='{{ index .data "ca-bundle.crt" }}' > /etc/kubernetes/kubelet-ca.crt

systemctl stop kubelet
rm -rf /var/lib/kubelet/pki /var/lib/kubelet/kubeconfig

crictl stopp $(sudo crictl pods -q)
crictl rmp $(sudo crictl pods -q)

systemctl start kubelet

until oc get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 1; done
for CSR in $(oc get csr | grep Pending | awk '{print $1}'); do
	oc adm certificate approve $CSR;
done

until oc get csr | grep Pending; do echo 'Waiting for second CSR request.'; sleep 1; done
for CSR in $(oc get csr | grep Pending | awk '{print $1}'); do
	oc adm certificate approve $CSR;
done

#echo "kubeconfig: $KUBECONFIG"
podman run -it --rm --network=host -v /etc/kubernetes/:/etc/kubernetes/:Z --entrypoint=/usr/bin/cluster-kube-apiserver-operator "${KAO_IMAGE}" recovery-apiserver destroy
`
)



func RegenerateCertificates(sshRunner *ssh.SSHRunner) error {
	/* FIXME: Need to error out if openshift version is older than 4.1.17 */
	startedKubelet, err := sd.Start("kubelet")
	if err != nil {
		logging.Debugf("Error starting kubelet service: %v", err)
		return err
	}
	if startedKubelet {
		defer sd.Stop("kubelet")
	}

	// Change the template to base64 and scp to VM
	base64CertRecoveryScript := base64.StdEncoding.EncodeToString([]byte(certRotationScriptTemplate))
	_, err = sshRunner.Run(fmt.Sprintf("cat <<EOF | sudo tee /tmp/certRecoveryScript.base64\n%s\nEOF", base64CertRecoveryScript))
	if err != nil {
		return err
	}
	// change the recoverycert script
	_, err = sshRunner.Run("base64 -d </tmp/certRecoveryScript.base64 > /tmp/certRecoveryScript.sh && chmod +x /tmp/certRecoveryScript.sh")
	if err != nil {
		return err
	}
	// execute the recovery script
	output, err := sshRunner.Run("sudo /tmp/certRecoveryScript.sh")
	if err != nil {
		return err
	}
	logging.Debug(output)

	return nil
}
```

cfergeau:
> @cfergeau FYI, below is what I did for `cert-renewal.go` file and tested and it does as expected.

Yeah, I considered base64 encoding/decoding, but transferring that script and running it was not my final goal, just initial experiment. I prefer to do as much as possible in native go code, so I switched to doing that instead of spending too much time on transferring that script.

pkg/crc/machine/machine.go
```diff
@@ -255,17 +255,19 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 	logging.Info("Verifying validity of the cluster certificates ...")
 	expiringIn7Days, duration, err := cluster.CheckCertsValidity(sshRunner)
 	if err != nil {
-		result.Error = err.Error()
-		return *result, errors.New(err.Error())
-	}
-	// Only show when VM is started from stopped state.
-	if exists {
-		if expiringIn7Days {
+		//result.Error = err.Error()
```
praveenkumar: @cfergeau I think these comments shouldn't be there right?

cfergeau: yeah left overs :)

pkg/crc/oc/oc.go
```diff
@@ -10,23 +10,44 @@ import (
...
+
 // UseOcWithConfig return the oc binary along with valid kubeconfig
 func UseOCWithConfig(machineName string) OcConfig {
-	oc := OcConfig{
+	localRunner := OcLocalRunner{
 		OcBinaryPath:   filepath.Join(constants.CrcBinDir, constants.OcBinaryName),
 		KubeconfigPath: filepath.Join(constants.MachineInstanceDir, machineName, "kubeconfig"),
 	}
-	return oc
+	return NewOcConfig(localRunner)
```
praveenkumar: 👍

pkg/crc/machine/cert-renewal.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/oc"
+	"github.com/code-ready/crc/pkg/crc/ssh"
+	"github.com/code-ready/crc/pkg/crc/systemd"
+	"github.com/code-ready/crc/pkg/crc/version"
+)
+
+var (
+	releaseImage = fmt.Sprintf("quay.io/openshift-release-dev/ocp-release:%s", version.GetBundleVersion())
```
praveenkumar: @cfergeau Are will still using this, since with new bundle we already have `openshift/cert-recovery` tag?

cfergeau: Yes, I haven't tested using `openshift/cert-recovery` yet, so I'd favour sticking with this for now, and switching to `openshift/cert-recovery` in a follow-up PR after the release.

pkg/crc/machine/cert-renewal.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return nil, err
+	}
+	recoveryPod.kaoImage = strings.TrimSpace(kaoImage)
+	logging.Debugf("kube-apiserver-operator image: %s", kaoImage)
+
+	_, err = sshRunner.Run(fmt.Sprintf("sudo podman pull --authfile='/var/lib/kubelet/config.json' %s", kaoImage))
+	if err != nil {
+		return nil, err
+	}
```
praveenkumar: @cfergeau we can replace this with.
```
recoveryPod.kaoImage = "openshift/cert-recovery:latest" 
```

pkg/crc/machine/machine.go
```diff
@@ -255,17 +255,17 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
-	// Only show when VM is started from stopped state.
-	if exists {
-		if expiringIn7Days {
+		needsCertsRenewal = true
+	} else if expiringIn7Days {
+		// Only show when VM is started from stopped state.
+		if exists {
 			logging.Warnf("Bundle certificates are going to expire in %d days, better to use new release", duration)
 		}
+		needsCertsRenewal = true
```
praveenkumar: @cfergeau why we need to cert renewal if there is already 7 days remaining?

pkg/crc/machine/cert-renewal.go
praveenkumar: We need to put a follow up issue for same, since that is not tested.

pkg/crc/machine/machine.go
cfergeau: Initial idea was to be proactive, and to renew the certs before they even expire. But since the renewal process is quite long, only doing it when absolutely required is a good idea. I'll change it.

praveenkumar: @cfergeau sure, please change it , it is really add around 3mins to complete process.

pkg/crc/machine/cert-renewal.go
```diff
@@ -0,0 +1,251 @@
...
+	logging.Debugf("Recovery KUBECONFIG: [%s]", recoveryPod.kubeconfig)
+
+	return &recoveryPod, nil
+}
+
+func waitForApiServer(machineName string) error {
+	localOc := oc.UseOCWithConfig(machineName)
+	waitForApiServer := func() error {
+		_, stderr, err := localOc.RunOcCommand("get", "node")
```
praveenkumar: 👍

pkg/crc/machine/machine.go
```diff
@@ -255,14 +255,13 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+	needsCertsRenewal := false
 	logging.Info("Verifying validity of the cluster certificates ...")
 	expiringIn7Days, duration, err := cluster.CheckCertsValidity(sshRunner)
 	if err != nil {
-		result.Error = err.Error()
-		return *result, errors.New(err.Error())
-	}
-	// Only show when VM is started from stopped state.
-	if exists {
+		needsCertsRenewal = true
```

pkg/crc/machine/machine.go
```diff
@@ -344,8 +342,26 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 		}
 
+		err = pullsecret.AddPullSecretToInstanceDisk(sshRunner, pullSecret)
+		if err != nil {
+			result.Error = err.Error()
+			return *result, errors.Newf("Failed to update user pull secret or cluster ID: %v", err)
+		}
+	}
+
+	if needsCertsRenewal {
```

PR #839, opened 2019-11-25
Issue #838 Add virtualbox deprecation notice

### LINKED ISSUE #838: Remove Virtualbox support (deprecated)
Our tests have concluded that native virtualization is easier to deal with in setting up, we even automate this as much as possible, On an aside, they also consume less resources. Mostly due to the fact they are tailored towards the host platform.

However, VirtualBox on Windows and macOS as driver were provided for certain situation where policies prevent the user from self-admin, etc. Windows users for example are often confronted by the restriction that a virtualization suite has been chosen by the other tools they use or by the administrative team. This is less likely on Linux. Due to the ease of use and streamlining the out of the box experience, we have been focussed to pursue native solutions foremost.

While CRC provided a driver for VirtualBox, this was an unsupported and mostly untested driver. We have received positive results from users during the alpha and beta testing period. We have included the driver for those who are willing to tinker when a native hypervisor is unavailable, such as due to company restrictions and/or policies. But be aware, from our own tests we conclude that VirtualBox overall consumes more memory and is slower in use than the native options we provide.

Although, we do not prevent people to add the needed drivers and support, we really do not have the resources and time on our schedule to handle more platforms and hypervisor combinations. We have learned from dealing with Minikube and Minishift and decided to not support these drivers from a basic installation.  Take note, nothing prevents outside contributors to add support. But note, 'supported' means more than just adding the setup and driver code, but also being able to run regular tests, integration ... And this is what we talk about here.

Therefore, we have decided to remove the VirtualBox support from our codebase. Starting with v1.2.0 we showed a deprecation notice and for v1.3.0 we have planned to remove the driver implementation. This means we will also not publish VirtualBox image bundles.

 — CRC Team

Note: in case of another hypervisor, VirtualBox will default to use full system virtualization (using the emulation backend) as it does not have access to hypercalls. be aware, that when WSL2 is enabled in future (or from Insider) you are unable to use VirtualBox in that case.

gbraad:
Fixes #838 Adds a deprecation notice when the vbox driver is used.

### DIFF
--- a/pkg/crc/machine/driver_darwin.go
+++ b/pkg/crc/machine/driver_darwin.go
@@ -3,6 +3,7 @@ package machine
 import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	crcos "github.com/code-ready/crc/pkg/os"
 
 	"github.com/code-ready/crc/pkg/crc/machine/config"
@@ -39,6 +40,7 @@ func getDriverOptions(machineConfig config.MachineConfig) interface{} {
 	switch machineConfig.VMDriver {
 
 	case "virtualbox":
+		logging.Warn("Virtualbox support is deprecated and will be removed in the next release.")
 		driver = virtualbox.CreateHost(machineConfig)
 	case "hyperkit":
 		driver = hyperkit.CreateHost(machineConfig)
--- a/pkg/crc/machine/driver_windows.go
+++ b/pkg/crc/machine/driver_windows.go
@@ -2,6 +2,7 @@ package machine
 
 import (
 	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	crcos "github.com/code-ready/crc/pkg/os"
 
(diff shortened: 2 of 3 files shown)


pkg/crc/machine/driver_windows.go
```diff
@@ -37,6 +38,7 @@ func getDriverOptions(machineConfig config.MachineConfig) interface{} {
 	switch machineConfig.VMDriver {
 
 	case "virtualbox":
+		output.Outln("Virtualbox support is deprecated and will be removed in the next release.")
```
praveenkumar: Instead of `output.Outln` I would use `logging.Warn` so that in the output it atleast show a warning to the user about this deprecation.

gbraad: I was considering this... If you think this is better, I'll change

gbraad:
Updated to use `logging.Warn()`

pkg/crc/machine/driver_darwin.go
```diff
@@ -39,6 +40,7 @@ func getDriverOptions(machineConfig config.MachineConfig) interface{} {
 	switch machineConfig.VMDriver {
 
 	case "virtualbox":
+		logging.Warn("Virtualbox support is deprecated and will be removed in the next release.")
```
praveenkumar: 👍

Issue #840, opened 2019-11-26
Use tagged image for recovery process instead fetching from internet


praveenkumar:
Right now https://github.com/code-ready/crc/blob/master/pkg/crc/machine/cert-renewal.go#L66-L77 we are trying to fetch the image content from internet but since https://github.com/code-ready/snc/pull/132 is in and we already tag that image so we don't need to fetch it from internet anymore.

cfergeau:
`kaoImage := "openshift/cert-recovery"` should do the trick.

praveenkumar:
Fixed by #842

PR #842, opened 2019-11-26
Issue #840 Use tagged image instead fetching from registry

praveenkumar:
We already tag cluster kube-apiserver-operator image as part of our
bundle generation script. Using the tag is going to avoid fetching
the image details from quay registry instead directly use of this tag.

### DIFF
--- a/pkg/crc/machine/cert-renewal.go
+++ b/pkg/crc/machine/cert-renewal.go
@@ -11,11 +11,10 @@ import (
 	"github.com/code-ready/crc/pkg/crc/oc"
 	"github.com/code-ready/crc/pkg/crc/ssh"
 	"github.com/code-ready/crc/pkg/crc/systemd"
-	"github.com/code-ready/crc/pkg/crc/version"
 )
 
 var (
-	releaseImage = fmt.Sprintf("quay.io/openshift-release-dev/ocp-release:%s", version.GetBundleVersion())
+	kaoImage = "openshift/cert-recovery"
 )
 
 func getRecoveryKubeConfig(output string) (string, error) {
@@ -35,10 +34,9 @@ func getRecoveryKubeConfig(output string) (string, error) {
 }
 
 type recoveryPod struct {
-	sshRunner    *ssh.SSHRunner
-	releaseImage string
-	kaoImage     string
-	kubeconfig   string
+	sshRunner  *ssh.SSHRunner
+	kaoImage   string
+	kubeconfig string
 }
 
 func (recoveryPod recoveryPod) GetKubeconfigPath() string {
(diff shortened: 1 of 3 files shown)


pkg/crc/machine/cert-renewal.go
```diff
@@ -61,20 +59,9 @@ func (recoveryPod *recoveryPod) runPodCommand(cmd string) (string, error) {
...
-	logging.Debugf("Using release image %s", recoveryPod.releaseImage)
-	kaoImage, err := sshRunner.Run(fmt.Sprintf("sudo oc adm release info --registry-config='/var/lib/kubelet/config.json' '%s' --image-for=cluster-kube-apiserver-operator", releaseImage))
-	if err != nil {
-		return nil, err
-	}
-	recoveryPod.kaoImage = strings.TrimSpace(kaoImage)
+	recoveryPod.kaoImage = kaoImage
 	logging.Debugf("kube-apiserver-operator image: %s", kaoImage)
 
-	_, err = sshRunner.Run(fmt.Sprintf("sudo podman pull --authfile='/var/lib/kubelet/config.json' %s", kaoImage))
```
cfergeau: With these references to `config.json` gone, some of the machine.go changes introduced by `3f84454e machine: Trigger certificate renewal when they are expired` can probably be removed:
```
diff --git a/pkg/crc/machine/machine.go b/pkg/crc/machine/machine.go
index b550443c..582211aa 100644
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -341,12 +341,6 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			result.Error = err.Error()
 			return *result, errors.Newf("Error copying kubeconfig file  %v", err)
 		}
-
-		err = pullsecret.AddPullSecretToInstanceDisk(sshRunner, pullSecret)
-		if err != nil {
-			result.Error = err.Error()
-			return *result, errors.Newf("Failed to update user pull secret or cluster ID: %v", err)
-		}
 	}
 
 	if needsCertsRenewal {
diff --git a/pkg/crc/pullsecret/pullsecret.go b/pkg/crc/pullsecret/pullsecret.go
index 81ee535e..1c40a81c 100644
--- a/pkg/crc/pullsecret/pullsecret.go
+++ b/pkg/crc/pullsecret/pullsecret.go
@@ -62,7 +62,7 @@ func AddPullSecretAndClusterID(sshRunner *ssh.SSHRunner, pullSec string, kubecon
 	}
 
 	// Add user pull secret to the instance
-	if err := AddPullSecretToInstanceDisk(sshRunner, pullSec); err != nil {
+	if err := addPullSecretToInstanceDisk(sshRunner, pullSec); err != nil {
 		return err
 	}
 	return nil
@@ -78,7 +78,7 @@ func addpullSecretSpecToInstance(sshRunner *ssh.SSHRunner, pullSec string) error
 	return nil
 }
 
-func AddPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error {
+func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error {
 	output, err := sshRunner.Run(fmt.Sprintf("cat <<EOF | sudo tee /var/lib/kubelet/config.json\n%s\nEOF", pullSec))
 	if err != nil {
 		return err
```

cfergeau:
Also, in shortlog "instead" -> "instead of"

pkg/crc/machine/cert-renewal.go
praveenkumar: @cfergeau yes we can revert those, will put another commit to revert it back.

Issue #832, opened 2019-11-22
OpenShift 4.0 ?


liangxia:
Red Hat CodeReady Containers reference OpenShift 4.0 in the docs [1], but 4.0 is not a valid version for OpenShift.

[1]
https://code-ready.github.io/crc/#understanding-codeready-containers_gsg
https://github.com/code-ready/crc#introduction

PR #844, opened 2019-11-26
Issue #832 update Openshift version in the doc

praveenkumar:
Openshift 4.x started with 4.1 as starting version 4.0 was dev preview
and never released to public so in the doc 4.1 and later make more sense.

### DIFF
--- a/docs/source/getting-started/content/con_understanding-codeready-containers.adoc
+++ b/docs/source/getting-started/content/con_understanding-codeready-containers.adoc
@@ -1,7 +1,7 @@
 [id="understanding-codeready-containers_{context}"]
 = Understanding {prod}
 
-{rh-prod} brings a minimal OpenShift 4.0 or newer cluster to your local computer.
+{rh-prod} brings a minimal OpenShift 4 cluster to your local computer.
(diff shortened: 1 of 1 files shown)


docs/source/getting-started/content/con_understanding-codeready-containers.adoc
```diff
@@ -1,7 +1,7 @@
 [id="understanding-codeready-containers_{context}"]
 = Understanding {prod}
 
-{rh-prod} brings a minimal OpenShift 4.0 or newer cluster to your local computer.
+{rh-prod} brings a minimal OpenShift 4.1 or newer cluster to your local computer.
```
cfergeau: Or just "brings a minimal OpenShift 4 cluster to your local computer." ?

robin-owen: I agree with Christophe here (and I believe I made the same suggestion in the past at some point).

praveenkumar: @robin-owen @cfergeau Done 👍

Issue #846, opened 2019-11-26
[Integration tests] Make it possible to run feature files separately


jsliacan:
When running `config.feature` independently, bundle hasn't been unpacked from the `crc` binary yet and the test already tries to access it.

Issue #845, opened 2019-11-26
[Integration tests] Update location to which embedded bundle is unpacked


jsliacan:
Right now, tests assume bundle is at the old location `$HOME/.crc/<bundle>` instead of the correct one: `$HOME/.crc/cache/<bundle>`.

PR #847, opened 2019-11-26
Issue #845, #846 [Integration tests] Bundle path and independent feature files

jsliacan:
This patch addresses #845 and #846.

### DIFF
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -274,7 +274,7 @@ func FileExistsInCRCHome(fileName string) error {
 func ConfigFileInCRCHomeContainsKeyMatchingValue(format string, configFile string, condition string, keyPath string, expectedValue string) error {
 
 	if expectedValue == "current bundle" {
-		expectedValue = bundleName
+		expectedValue = fmt.Sprintf(".*%s", bundleName)
 	}
 	configPath := filepath.Join(CRCHome, configFile)
 
@@ -406,7 +406,7 @@ func SetConfigPropertyToValueSucceedsOrFails(property string, value string, expe
 	if value == "current bundle" {
 
 		if bundleEmbedded {
-			value = filepath.Join(CRCHome, bundleName)
+			value = filepath.Join(CRCHome, "cache", bundleName)
 		} else {
 			value = bundleName
 		}
--- a/test/integration/features/config.feature
+++ b/test/integration/features/config.feature
@@ -42,6 +42,7 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
 
     @linux @darwin @windows
     Scenario: CRC config checks (bundle version)
+        Given executing "crc setup" succeeds
         When setting config property "bundle" to value "current bundle" succeeds
         And "JSON" config file "crc.json" in CRC home folder contains key "bundle" with value matching "current bundle"
         And setting config property "bundle" to value "/path/to/nonexistent/bundle/crc_hypervisor_version.tar.xz" fails


Makefile
```diff
@@ -128,7 +128,8 @@ ifndef PULL_SECRET_FILE
 	PULL_SECRET_FILE = --pull-secret-file=$(HOME)/Downloads/crc-pull-secret
 endif
 ifndef BUNDLE_LOCATION
-	BUNDLE_LOCATION = --bundle-location=$(HOME)/Downloads/crc_libvirt_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
+	BUNDLE_LOCATION = --bundle-location=embedded
+##$(HOME)/Downloads/crc_libvirt_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
```
cfergeau: This comment can be removed I think? I see this is reverted in the next commit, so not sure what is intended here?

test/integration/crcsuite/crcsuite.go
```diff
@@ -406,7 +411,7 @@ func SetConfigPropertyToValueSucceedsOrFails(property string, value string, expe
 	if value == "current bundle" {
 
 		if bundleEmbedded {
-			value = filepath.Join(CRCHome, bundleName)
+			value = filepath.Join(CRCHome, "cache", bundleName)
 		} else {
 			value = bundleName
 		}
```
cfergeau: Have you considered putting these 4 lines in a `getBundleName` function since it's now duplicated in 2 places?

test/integration/crcsuite/crcsuite.go
```diff
@@ -274,7 +274,12 @@ func FileExistsInCRCHome(fileName string) error {
 func ConfigFileInCRCHomeContainsKeyMatchingValue(format string, configFile string, condition string, keyPath string, expectedValue string) error {
 
 	if expectedValue == "current bundle" {
-		expectedValue = bundleName
+
+		if bundleEmbedded {
+			expectedValue = filepath.Join(CRCHome, "cache", bundleName)
+		} else {
+			expectedValue = bundleName
+		}
```
cfergeau: Most of this is reverted in the next commit?

test/integration/features/config.feature
```diff
@@ -42,6 +42,7 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
 
     @linux @darwin @windows
     Scenario: CRC config checks (bundle version)
+        Given executing "crc setup" succeeds
```
cfergeau: Is this the main reason for this commit? I'm confused by the changes in the previous commit which get almost reverted in this one.

Makefile
jsliacan: @cfergeau that's my mistake. I had to change that for testing (changes only apply to embedded bundles). But I only checked if everything was back to non-embedded after the last commit. Hence the back and forth. Sorry, I'll address this.

test/integration/crcsuite/crcsuite.go
jsliacan: Yes, I tested it on windows in the meantime. I use regex to match the content of a config file against the expected value. If the expected value is a path, the backslashes escape characters inside regex :(. I'll clean up the commits though, thanks.

test/integration/features/config.feature
jsliacan: Yes, that's the only reason. Without running `crc setup`, bundle won't be unpacked to it's location, and so config setting with that bundle cannot be set. This is not an issue if `basic.feature` is run before this feature (as `crc setup` is executed in basic tests).

test/integration/crcsuite/crcsuite.go
jsliacan: Now those 4 lines are actually different, so extracting them to a function didn't seem good anymore. Thanks

PR #852, opened 2019-11-27
Doc updates

cfergeau:
Mention cert renewal process in the docs, and mention odo too.

### DIFF
--- a/docs/source/getting-started/content/proc_accessing-the-openshift-cluster.adoc
+++ b/docs/source/getting-started/content/proc_accessing-the-openshift-cluster.adoc
@@ -89,5 +89,6 @@ The `monitoring` cluster operator is expected to report `Unknown` availability.
 ====
 +
   . The link:https://docs.openshift.com/container-platform/latest/applications/projects/working-with-projects.html[OpenShift documentation] covers the creation of projects and applications.
+  link:https://docs.openshift.com/container-platform/4.2/cli_reference/openshift_developer_cli/understanding-odo.html[OpenShift Do] (`odo`) can also be used to create OpenShift projects and applications from the command line.
 +
 See <<troubleshooting-codeready-containers_{context}>> if you cannot access the {prod} OpenShift cluster.
--- a/docs/source/getting-started/content/proc_troubleshooting-expired-certificates.adoc
+++ b/docs/source/getting-started/content/proc_troubleshooting-expired-certificates.adoc
@@ -1,12 +1,16 @@
 [id="troubleshooting-expired-certificates_{context}"]
 = Troubleshooting expired certificates
 
-Each released `{bin}` binary includes an embedded system bundle that expires 30 days after the release due to certificates embedded in the OpenShift cluster.
+Prior to {rh-prod} 1.2.0, the system bundle included in each released `{bin}` binary expired 30 days after the release due to certificates embedded in the OpenShift cluster.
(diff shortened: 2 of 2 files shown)


docs/source/getting-started/content/proc_troubleshooting-expired-certificates.adoc
```diff
@@ -1,12 +1,16 @@
 [id="troubleshooting-expired-certificates_{context}"]
 = Troubleshooting expired certificates
 
-Each released `{bin}` binary includes an embedded system bundle that expires 30 days after the release due to certificates embedded in the OpenShift cluster.
+Prior to {rh-prod} 1.2.0, each released `{bin}` binary included a system bundle that expires 30 days after the release due to certificates embedded in the OpenShift cluster.
```
robin-owen: Don't we still include the embedded system bundle?

docs/source/getting-started/content/proc_troubleshooting-expired-certificates.adoc
```diff
@@ -1,12 +1,16 @@
...
 Using an older `{bin}` binary or system bundle will result in an expired certificates error.
 
+Starting from {prod} 1.2.0, the embedded certificates can be automatically renewed by {bin}.
+The certificate renewal process is triggered when needed when [command]`{bin} start` is run.
+When certificates need to be renewed, this can adds up to 5 minutes to the cluster startup time.
```
robin-owen: We try to avoid phrasal verbs (for example, "startup"). Reword to "this can add up to five minutes to the start time of the cluster"?

docs/source/getting-started/content/proc_accessing-the-openshift-cluster.adoc
```diff
@@ -89,5 +89,6 @@ The `monitoring` cluster operator is expected to report `Unknown` availability.
 ====
 +
   . The link:https://docs.openshift.com/container-platform/latest/applications/projects/working-with-projects.html[OpenShift documentation] covers the creation of projects and applications.
+  link:https://docs.openshift.com/container-platform/4.2/cli_reference/openshift_developer_cli/understanding-odo.html[OpenShift Do] (`odo`) can also be used to create OpenShift projects and applications from the command line.
```
robin-owen: :+1:

docs/source/getting-started/content/proc_troubleshooting-expired-certificates.adoc
cfergeau: 'embedded' is repeated twice in the same sentence, which is not so nice. I don't think the 'embedded bundle' VS 'bundle' is really important here.

robin-owen: Sure, that makes sense. But the inclusion of "prior to {rh-prod} 1.2.0" implies that since 1.2.0, there is no included system bundle. Does that make sense?

cfergeau: Oooh ok, I thought you were ascking about the removal of 'embedded'
Rephrase it to "Prior to {rh-prod} 1.2.0, the system bundle included in each released `{bin}` binary expired 30 days after the release due to certificates embedded in the OpenShift cluster." ?

robin-owen: Sounds good to me. :+1:

Issue #859, opened 2019-12-03
[Integration] Allow user to run integration tests without removing `.crc` folder


jsliacan:
It is not always desirable that the `.crc` folder is removed by the integration tests. The suggested way to tell the tests not to remove the folder is to place a `.keep` file directly inside it.

jsliacan:
Solved by #860

PR #860, opened 2019-12-03
Issue #859 Integration: run tests and keep .crc folder

jsliacan:
Allow user to run integration tests and keep the `.crc` folder. This is done by placing a `.keep` file inside the `~/.crc` folder

### DIFF
--- a/test/integration/crcsuite/util.go
+++ b/test/integration/crcsuite/util.go
@@ -5,6 +5,7 @@ package crcsuite
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	clicumber "github.com/code-ready/clicumber/testsuite"
 )
@@ -22,14 +23,21 @@ func DeleteCRC() error {
 // Remove CRC home folder ~/.crc
 func RemoveCRCHome() error {
 
-	err := os.RemoveAll(CRCHome)
+	keepFile := filepath.Join(CRCHome, ".keep")
 
-	if err != nil {
-		fmt.Printf("Problem deleting CRC home folder %s.\n", CRCHome)
-		return err
+	_, err := os.Stat(keepFile)
+	if err != nil { // cannot get keepFile's status
+		err = os.RemoveAll(CRCHome)
+
+		if err != nil {
+			fmt.Printf("Problem deleting CRC home folder %s.\n", CRCHome)
+			return err
+		}
+
+		fmt.Printf("Deleted CRC home folder %s.\n", CRCHome)
+		return nil
+
 	}
-
-	fmt.Printf("Deleted CRC home folder %s.\n", CRCHome)
-	return nil
-
+	// keepFile exists
+	return fmt.Errorf("Folder %s not removed as per request: %s present.\n", CRCHome, keepFile)
 }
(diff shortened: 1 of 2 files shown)


test/integration/crcsuite/util.go
```diff
@@ -22,14 +23,23 @@ func DeleteCRC() error {
...
+		if err != nil {
+			fmt.Printf("Problem deleting CRC home folder %s.\n", CRCHome)
+			return err
+		}
+
+		fmt.Printf("Deleted CRC home folder %s.\n", CRCHome)
+		return nil
+
+	} else {
+		fmt.Printf("Folder '.crc' not removed as per request: '.keep' file found inside.\n")
```
cfergeau: `"Folder '%s' not removed ...", CRCHome` ?

test/integration/crcsuite/util.go
```diff
@@ -22,14 +23,23 @@ func DeleteCRC() error {
 // Remove CRC home folder ~/.crc
 func RemoveCRCHome() error {
 
-	err := os.RemoveAll(CRCHome)
+	keepFile := filepath.Join(CRCHome, ".keep")
 
-	if err != nil {
-		fmt.Printf("Problem deleting CRC home folder %s.\n", CRCHome)
+	_, err := os.Stat(keepFile)
+	if os.IsNotExist(err) {
```
cfergeau: dunno if it's better to be explicit with err?
`if err != nil && os.IsNotExist(err) {`

jsliacan: Could it be that you meant `if err != nil || os.IsNotExist(err) {`? To cover the situation that there is an error, but it's not the `IsNotExist` one. Correct?

cfergeau: Ah, I was thinking of doing the opposite, if `~/.crc/.keep` is not there for sure, then delete `~/.crc`, otherwise keep it. From that perspective, what I suggested looks fine to me, "if there was an error, and this error indicates `~/.crc/.keep` is not present, then delete `~/.crc`"
Note that my suggestion is not different from the current `if os.IsNotExist(err) {`, as `if os.IsNotExist(nil) {` is going to be false.

If you prefer the other way, always delete unless you are sure .keep is present, then I don't think the current code does that.

test/integration/crcsuite/util.go
```diff
@@ -22,14 +23,22 @@ func DeleteCRC() error {
...
+		if err != nil {
+			fmt.Printf("Problem deleting CRC home folder %s.\n", CRCHome)
+			return err
+		}
 
+		fmt.Printf("Deleted CRC home folder %s.\n", CRCHome)
+		return nil
+
+	} else { // keepFile exists
+		fmt.Printf("Folder %s not removed as per request: %s present.\n", CRCHome, keepFile)
```
cfergeau: I think I would error out and tell the user to delete/move away ~/.crc themselves as I expect this is going in the way of the tests?

jsliacan: It shouldn't break tests. Just the test result will mean something different (as the clean install won't be tested). 

But I think I now understand what you wanted -- essentially just be given a chance to back up your folder before running the tests. I changed that part accordingly. I didn't provide a message for the user, as if they have a `.keep` file in the folder, they know why they put it there. WDYT?

Issue #337, opened 2019-07-22
Running CRC behind proxy


piggyvenus:
Export `HTTP_PROXY` and `HTTPS_PROXY`.
Getting error:
```
sudo crc setup
INFO Caching oc binary                            
ERRO Error occurred: Download failed: Head https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/macosx/oc.tar.gz: dial tcp: lookup mirror.openshift.com on x.x.x.x:53: no such host 
FATA Not able to download oc Download failed: Head https://mirror.openshift.com
/pub/openshift-v4/clients/oc/latest/macosx/oc.tar.gz: dial tcp: lookup mirror.openshift.com on x.x.x.x:53: no such host 
```

gbraad:
Currently not possible. We will add proxy support when OpenShift has added support on cluster level.

cfergeau:
For what it's worth, the cavaliercoder/grab module that we use for our downloads seems to be able to use proxies, so I would have expected oc/machine driver downloads to work. Needs more testing and debugging ;)

praveenbommali:
We're not able to access in  corporate environment and  getting the below issue  
ERRO Download failed: Head https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/windows/oc.zip: dial tcp 54.173.18.88:443: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
FATA Unable to download oc Download failed: Head https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/windows/oc.zip: dial tcp 54.173.18.88:443: connectex: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.

praveenbommali:
How to  set the proxy while  running 
`>> crc setup  `

praveenkumar:
@praveenbommali we still don't have proxy support for crc right now, it is on our high priority list and will be start working soon on that.

cfergeau:
https://github.com/code-ready/crc/issues/509 would workaround these issues with downloading oc/machine-drivers  when a proxy is involved

cfergeau:
Another issue we will have when adding proxy support is `hosts -R 3 quay.io` which we currently use, but will need to change to something different.

jepries:
Sorry to rehash this, it does look like in your 1.2.0 bundle, some of this has been fixed.  I was able to run 

    crc setup

[...]

PR #863, opened 2019-12-03
Issue #337 Add basic layout for proxy support

praveenkumar:
At this momemt user can specify the proxy details as part of crc
config file using

- `crc config set http-proxy` to set http-proxy
- `crc config set https-proxy` to set https-proxy
- `crc config set no-proxy` to set no-proxy

This is used for start and setup command. Right now setup command able
to fetch the content over proxy if it is set in the config file.


praveenkumar:
> but it will also honor the http_proxy/https_proxy environment variables.
I'm not even sure we need to have configuration options for proxy (except on Windows)

Right, I will add that in the log (honor proxy env variable), but we do need at least config option for proxy just to be platform independent setting.

cfergeau:
> but we do need at least config option for proxy just to be platform independent setting.

or we could add support for system-wide proxy settings on Windows and set it automatically too?

praveenkumar:
> or we could add support for system-wide proxy settings on Windows and set it automatically too?

@cfergeau Didn't get you, how crc will add support to have system-wide proxy setting on windows or you mean a windows user should set it himself and crc should just consume it?

cfergeau:
> > or we could add support for system-wide proxy settings on Windows and set it automatically too?
> 
> @cfergeau Didn't get you, how crc will add support to have system-wide proxy setting on windows or you mean a windows user should set it himself and crc should just consume it?

Apparently proxy settings can be set in the control center, and then queried through powershell or the registry, wondering if we should integrate with that and try not to provide `crc config` support for this?

praveenkumar:
> if we should integrate with that and try not to provide crc config support for this?

@cfergeau I really want to have it at this moment, most of time users just have proxy setup in the browser and not systemwide so we have to put document how to do it in windows, having it as part of `config` makes easy for just single entry in doc.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	if !govalidator.IsURL(proxyUrl) {
+		return fmt.Errorf("Proxy URL '%s' is not valid.", proxyUrl)
+	}
+	return nil
+}
+
+// parseProxySpecialChar parse the URI and convert special char to hex
+func parseProxySpecialChar(proxyUrl string, defaultScheme string) string {
```
cfergeau: I don't understand why we need to do this :(
`parse` is misleading, `escape` is more descriptive.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,201 @@
...
+		!strings.HasPrefix(proxyUrl, "https://") {
+		proxyUrl = fmt.Sprintf("%s://%s", defaultScheme, proxyUrl)
+	}
+
+	u, _ := url.Parse(proxyUrl)
+	scheme := fmt.Sprintf("%s://", u.Scheme)
+	if strings.HasPrefix(proxyUrl, scheme) {
+		s := strings.Replace(proxyUrl, scheme, "", 1)
+		s = url.PathEscape(s)
+		return scheme + s
```
cfergeau: if my proxy URL is `https://cfergeau%40redhat.com:password@example.com/proxy/path`, `url.PathEscape` is going to corrupt the URL.

cmd/crc/cmd/start.go
```diff
@@ -58,6 +59,7 @@ func runStart(arguments []string) {
 		NameServer:    crcConfig.GetString(config.NameServer.Name),
 		GetPullSecret: getPullSecretFileContent,
 		Debug:         isDebugLog(),
+		Proxy:         handleProxyConfig,
```
cfergeau: I think you can pass directly the `ProxyConfig` here rather than a callback? Nothing is using this `Proxy` field though?

cmd/crc/cmd/setup.go
```diff
@@ -32,6 +32,16 @@ func runSetup(arguments []string) {
...
 	}
+
+	// Proxy Setup
+	proxy, err := handleProxyConfig()
+	if err != nil {
+		errors.ExitWithMessage(1, err.Error())
+	}
+	if proxy != nil {
+		proxy.ApplyToEnvironment()
+	}
```
cfergeau: Maybe this should be handled in the `download` package directly? Or are there more codepaths which need proxy?

cmd/crc/cmd/start.go
```diff
@@ -155,3 +157,21 @@ func checkIfNewVersionAvailable(noUpdateCheck bool) {
...
 	logging.Debugf("No new version available. The latest version is %s", newVersion)
 }
+
+func handleProxyConfig() (*network.ProxyConfig, error) {
+	httpProxy := crcConfig.GetString(config.HttpProxy.Name)
+	httpsProxy := crcConfig.GetString(config.HttpsProxy.Name)
+	noProxy := crcConfig.GetString(config.NoProxy.Name)
+
+	proxyConfig, err := network.NewProxyConfig(httpProxy, httpsProxy, noProxy)
+
```
cfergeau: have you considered having `ProxyConfig` do the `crcConfig.GetString` calls itself rather than doing it here? We want *all* users of ProxyConfig to get values from crc config if they are available, so better to do it from the `ProxyConfig` code?

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,201 @@
...
+	noProxy    []string
+}
+
+// NewProxyConfig creates a proxy configuration with the specified parameters. If a empty string is passed
+// the corresponding environment variable is checked.
+func NewProxyConfig(httpProxy string, httpsProxy string, noProxy string) (*ProxyConfig, error) {
+	defaultScheme := "http"
+
+	// We only parse the Proxy Special char if it is coming from the config file
+	// or user defined flag. In case of Env variable we just ignore parsing.
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau we had issue with this on minishift side where user try to put the special char as part of config without escaping it and on the code side we are not able to validate this, if you have something in mind how to validate it before user add it as part of config I would happy to error out and point to some doc.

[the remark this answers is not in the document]
cfergeau: ah, this is https://github.com/minishift/minishift/issues/2283
I'd argue that this cannot be solved with software and the user has to provide a proper URL. 
If I set my proxy to "http://foo:bar:baz@proxy.example.com", am I user foo:bar with password baz, or user foo with password bar:baz? If I now set it to "http://foo:bar%3Abaz@proxy.example.com", is my password bar:baz with an escaped ':', or is it really bar%3Abaz, in which case crc would need to escape the %?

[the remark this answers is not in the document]
praveenkumar: @cfergeau make sense, in that case I am going to remove this escape function completely and in the documentation just add a note that user need to handle special character in the proxy themselves.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,201 @@
...
+		httpsProxy = parseProxySpecialChar(httpsProxy, defaultScheme)
+	}
+
+	if httpProxy == "" {
+		httpProxy = os.Getenv("http_proxy")
+		if httpProxy == "" {
+			httpProxy = os.Getenv("HTTP_PROXY")
+		}
+	}
+	err := ValidateProxyURL(httpProxy, defaultScheme)
```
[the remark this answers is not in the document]
praveenkumar: govalidator.IsURL is provide a better way to parse a url to check it validity, but since we already removing `parseProxySpecialChar()` one so that should be not an issue.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,201 @@
...
+		noProxy:    np,
+	}
+
+	return &config, nil
+}
+
+// ProxyConfig returns a the proxy configuration as a slice, one element for each of the potential settings
+// HTTP_PROXY, HTTPS_PROXY and NO_PROXY. If proxies are not enabled an empty slice is returned.
+func (p *ProxyConfig) ProxyConfig() []string {
+	config := []string{}
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I added it because I thought we will use it when cluster wide proxy part need to implemented but we can add it later if required (removing it now).

cmd/crc/cmd/setup.go
praveenkumar: @cfergeau this will ensure that before running any code path, the proxy is apply to env, this doesn't hurt us and in future if we add any code path which require proxy shouldn't have to worry about adding it in that function.

cmd/crc/cmd/start.go
praveenkumar: @cfergeau right now yes, nothing is using it but it should be used as soon as we start the cluster proxy part so I just put the place holder before.

This callback will allow me to return the error in case of fail to validate the proxy URI.

cmd/crc/cmd/start.go
praveenkumar: @cfergeau I didn't get you? you mean move this to package level and have something following? if this is what have in mind then no since `config` is coming from `cmd` side not from the `pkg` side so I don't want to have anything which come from `cmd` to `pkg` otherwise moving forward we might hit the cyclic dependencies. 

```golang
func GetProxyConfig() (*ProxyConfig, error) {
	defaultScheme := "http"
        
        httpProxy := crcConfig.GetString(config.HttpProxy.Name)
	if httpProxy == "" {
		httpProxy = getProxyFromEnv("http_proxy")
	}
	err := ValidateProxyURL(httpProxy, defaultScheme)
	if err != nil {
		return nil, err
	}
```

cmd/crc/cmd/setup.go
gbraad: well... is might not hurt us, but it does have a side-effect, it applies changes to the users shell environment.

cmd/crc/cmd/start.go
gbraad: also not following the argument, but as pointed out we avoid moving code into pkg that handles with basic config setup.

cmd/crc/cmd/start.go
```diff
@@ -155,3 +157,21 @@ func checkIfNewVersionAvailable(noUpdateCheck bool) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if proxyConfig.IsEnabled() {
+		logging.Info("Using proxy for the setup")
+		logging.Debugf("HTTP-PROXY: %s, HTTPS-PROXY: %s, NO-PROXY: %s", proxyConfig.HttpProxy(), proxyConfig.HttpsProxy(), proxyConfig.NoProxy())
```
gbraad: how about authentication strings inside the proxy variables?

pkg/crc/network/proxy.go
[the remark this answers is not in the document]
cfergeau: If doing this escaping manually turns out to be problematic, an alternative could be to provide a way to set the proxy username and proxy password separately.

cmd/crc/cmd/start.go
praveenkumar: @gbraad this is debug info so only depend on user if want to share.

gbraad: Is the user aware that this might happen. We always should prevent logging username and/or passwords, redact them when possible.

pkg/crc/network/proxy.go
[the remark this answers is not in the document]
gbraad: this is not what happens when a enviornment variable is used....

also, the username, password would suggest NTLM would be possible, while added to the standard string would make it more obvious we only do basic authentication.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,143 @@
...
+		httpProxy:  httpProxy,
+		httpsProxy: httpsProxy,
+		noProxy:    np,
+	}
+
+	return &config, nil
+}
+
+func getProxyFromEnv(proxyScheme string) string {
+	p := os.Getenv(proxyScheme)
```
gbraad: @cfergeau these can contain proxy authentication... so when a separate option is used, how to reliably recontruct or handle this?

cmd/crc/cmd/start.go
cfergeau: `config` is coming from `pkg`: `import crcConfig "github.com/code-ready/crc/pkg/crc/config"`
Requiring every user of the ProxyConfig package to know that at creation time, you need to pass it configuration values you fetched yourself is awkward/error-prone.

cfergeau: Alternative which might work: add `DefaultHttpProxy`, `DefaultHttpsProxy`, `DefaultNoProxy` variables to the proxy package, use these in `NewProxyConfig` instead of its 3 args, and set `Default*` from the config in the start command.

praveenkumar: @cfergeau I am still missing your point of view but I am talking about `config.HttpProxy.Name` part where we get the value of it. This `config` is part of `cmd` and not `pkg` or you are suggesting something like?
```
httpProxy := crcConfig.GetString("http-proxy") => using directly the 'http-proxy' string instead of getting it from the configs?
```

cmd/crc/cmd/start.go
cfergeau: > @cfergeau right now yes, nothing is using it but it should be used as soon as we start the cluster proxy part so I just put the place holder before.

I'd prefer that we add this when it starts getting used. Maybe during review of the cluster part, we'll realize things need to be done differently.

> This callback will allow me to return the error in case of fail to validate the proxy URI.

I'd prefer if we could make this simpler.. ;)

cmd/crc/cmd/start.go
praveenkumar: @gbraad I do agree on that part so what you suggest, can we just not print it at all then?

cmd/crc/cmd/start.go
praveenkumar: > Alternative which might work: add DefaultHttpProxy, DefaultHttpsProxy, DefaultNoProxy variables to the proxy package, use these in NewProxyConfig instead of its 3 args, and set Default* from the config in the start command.

@cfergeau you mean like below?  how does that make difference, I still not sure what is missing? 

```golang
func handleProxyConfig() (*network.ProxyConfig, error) {
	proxyConfig.DefaultHttpProxy := crcConfig.GetString(config.HttpProxy.Name)
	[...]
	proxyConfig, err := network.NewProxyConfig()
```

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	config = append(config, fmt.Sprintf("NO_PROXY=%s", p.NoProxy()))
+	config = append(config, fmt.Sprintf("no_proxy=%s", p.NoProxy()))
+
+	return config
+}
+
+// Override proxy for local/intermediate proxy
+func (p *ProxyConfig) OverrideHttpProxy(proxy string) {
+	p.httpProxy = proxy
```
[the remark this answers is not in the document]
gbraad: I think this came from Minishift, as they were intended as the override of values given by the user, as they were used in te flow of cluster.go to override the values by using an intermediate proxy instead. I think the meaning here is not the same, and perhaps if unused, better to leave out.

cmd/crc/cmd/start.go
cfergeau: url.Parse() would allow to print only the bits you want to show.

gbraad: I just wanted to suggest the same... like `Host` is possibly the only important part.

cmd/crc/cmd/start.go
cfergeau: >  but I am talking about config.HttpProxy.Name part where we get the value of it.
Ah, right, I missed this one :) I would define it in dns/proxy.go, this way any user of the proxy API has easy access to the config keys to use. But apparently we only want `cmd/` to make use of `pkg/crc/config`?

cmd/crc/cmd/start.go
gbraad: ... and perhaps this should even be functions on `ProxyConfig`...

cmd/crc/cmd/start.go
cfergeau: > > Alternative which might work: add DefaultHttpProxy, DefaultHttpsProxy, DefaultNoProxy variables to the proxy package, use these in NewProxyConfig instead of its 3 args, and set Default* from the config in the start command.
> 
> @cfergeau you mean like below? how does that make difference, I still not sure what is missing?
> 
> ```go
> func handleProxyConfig() (*network.ProxyConfig, error) {
> 	proxyConfig.DefaultHttpProxy := crcConfig.GetString(config.HttpProxy.Name)
> 	[...]
> 	proxyConfig, err := network.NewProxyConfig()
> ```

In `init()` or some function called early, we'd initialize proxy from config, either by directly setting the `Default*Proxy` variables, or through a dedicated helper (`proxyConfig.InitProxyDefaults()`).
Once you've done that, the `handleProxyConfig()` delegate is no longer needed, any code can call `proxyConfig, err := network.NewProxyConfig()` directly and check for errors. The `Default*Proxy` variables are just there if we want to avoid reading the config from `pkg/`

pkg/crc/network/proxy.go
cfergeau: Context seems to be missing, I assume this was an answer to my comment saying we can have separate options if we want the user to be able to pass unescaped passwords/usernames?
We have to define priority rules between env and config anyways, so we'll just say "anything defined in the config overrides anything set in the environment" (or the opposite). This is not much different from setting both http_proxy and the corresponding configuration option.

cmd/crc/cmd/start.go
cfergeau: you want the scheme, the port number. If user and password are set, I think I would print `https://username:xxxx@proxy.example.com` (real username, but replace the password with xxx. This makes it obvious something was set.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/asaskevich/govalidator"
+)
+
+var defaultNoProxies = []string{".apps-crc.testing", "127.0.0.1", "api.crc.testing", "localhost"}
```
cfergeau: `apps-crc.testing` and `crc.testing` should be coming from the bundle, not be hardcoded here.

pkg/crc/network/proxy.go
praveenkumar: @cfergeau As of now anything which is part of config should overwrite what's in the ENV.

> With that said, I dont' think we need dedicated username/password options just now, this is just something to keep in mind if user feedback tells us it's needed.

Yes, agreed 👍

cmd/crc/cmd/start.go
praveenkumar: > But apparently we only want cmd/ to make use of pkg/crc/config?

@cfergeau No, that is not the case but there is also a `config` which is part of `cmd/config` which we don't want to use in `pkg` level.

>Once you've done that, the handleProxyConfig() delegate is no longer needed, any code can call proxyConfig, err := network.NewProxyConfig() directly and check for errors.

Doesn't this will make everything implicit? Right now we know only start and setup need the proxy and no other cmd really required it. Now from the start side if any other function require then we can pass this as part of function parameter (which will make it explicit) instead directly invoking the `network.NewProxyConfig` from that function. I think it just matter of taste I would like to have it passed explicitly.

cfergeau: > Doesn't this will make everything implicit?

It makes sense to me that a proxy is used whenever one is needed rather than having to be explicitly provided before proxy can be used. For example, cavaliercoder/grab uses http_proxy by default.

>  I think it just matter of taste I would like to have it passed explicitly.

My main objection is that this is currently done through a function delegate, as long as it's replaced with something simpler, I'm fine with both explicit and implicit.

cfergeau: Since it's unused at the moment, I'd just remove it from that commit so that we move forward with it, and rediscuss that when there is an actual usecase for it.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	return p.httpProxy
+}
+
+// HttpsProxy returns the configured value for the HTTPS proxy. The empty string is returned in case HTTPS proxy is not set.
+func (p *ProxyConfig) HttpsProxy() string {
+	return p.httpsProxy
+}
+
+// HttpProxy with hidden credentials
+func (p *ProxyConfig) HttpProxyWithHiddenCredentials() string {
```
cfergeau: I'd name it `HttpProxyForDisplay()`, it's in my opinion clearer when we want to use it.
I think I would just add a generic `UriStringForDisplay()` and call `UriStringForDisplay(HttpsProxy())` whenever needed. I know Gerard suggested otherwise, so let's keep things this way for now

cmd/crc/cmd/start.go
praveenkumar: > Since it's unused at the moment, I'd just remove it from that commit so that we move forward with it, and rediscuss that when there is an actual usecase for it.

@cfergeau it is used for setup command atm.

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// AddNoProxy appends the specified host to the list of no proxied hosts.
+func (p *ProxyConfig) AddNoProxy(host string) {
+	p.noProxy = append(p.noProxy, host)
+}
+
+// Sets the current config as environment variables in the current process.
+func (p *ProxyConfig) ApplyToEnvironment() {
+	if !p.IsEnabled() {
```
cfergeau: Just a note for potential future improvements, https://golang.org/pkg/os/exec/#example_Command_environment could be used as a more fine-grained alternative to `ApplyToEnvironment()`

cmd/crc/cmd/setup.go
```diff
@@ -32,6 +32,7 @@ func runSetup(arguments []string) {
 	if err := validateSetupFlags(); err != nil {
 		errors.Exit(1)
 	}
+
 	preflight.SetupHost()
```
cfergeau: This can be removed :)

pkg/crc/network/proxy.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// ProxyConfig keeps the proxy configuration for the current environment
+type ProxyConfig struct {
+	HttpProxy  string
+	HttpsProxy string
+	NoProxy    string
+}
+
+// NewProxyConfig creates a proxy configuration with the specified parameters. If a empty string is passed
```
cfergeau: nit: "an empty string"

praveenkumar:
> This is no longer adding api.crc.testing and *.crc-apps.testing to NoProxy, I assume this will be part of another commits?

@cfergeau Right this should added when cluster wide proxy implemented.

Issue #865, opened 2019-12-04
A typo in the output of `crc config --help`


jsliacan:
## CRC version
```bash
crc version: 1.3.0-dev+e65ff7f
OpenShift version: 4.2.8 (not embedded in binary)
```

### Steps to reproduce

  1. `crc config --help`

### Expected
configurable properties should be entered *after* subcommand (or *with* it). As in
`crc config set pull-secret-file ~/Downloads/my-pull-secret`, the subcommand being `set`. 

### Actual
```
Modifies crc configuration properties.
Configurable properties (enter as SUBCOMMAND):
```

PR #866, opened 2019-12-04
Issue #865: update output of a help string for crc config

jsliacan:
Suggested alteration to the `crc config --help` output. Other alternatives would be:
- (use in conjunction with a SUBCOMMAND)
- (enter with SUBCOMMAND)
- ...

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -38,7 +38,7 @@ var (
 
 func init() {
 	ConfigCmd.Long = `Modifies crc configuration properties.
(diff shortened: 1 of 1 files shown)


cmd/crc/cmd/config/config.go
```diff
@@ -38,7 +38,7 @@ var (
 
 func init() {
 	ConfigCmd.Long = `Modifies crc configuration properties.
-Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields()
+Configurable properties (enter after SUBCOMMAND): ` + "\n\n" + configurableFields()
```
praveenkumar: 👍

PR #868, opened 2019-12-04
UX - copy paste and login as a kubeadmin

rinormaloku:
After using the tool a couple of times it was always a little cumbersome to log in as an admin.

This way the user has only to copy-paste.

### DIFF
--- a/cmd/crc/cmd/console.go
+++ b/cmd/crc/cmd/console.go
@@ -60,8 +60,8 @@ func runConsole(arguments []string) {
 		output.Outln(result.ClusterConfig.WebConsoleURL)
 	}
 	if consolePrintCredentials {
-		output.Outln("To login as a regular user, username is 'developer' and password is 'developer'.")
-		output.Outf("To login as an admin, username is 'kubeadmin' and password is '%s'.\n", result.ClusterConfig.KubeAdminPass)
+		output.Outf("To login as a regular user, run 'oc login -u developer -p developer %s'.\n", result.ClusterConfig.ClusterAPI)
+		output.Outf("To login as an admin, run 'oc login -u kubeadmin -p %s %s'\n", result.ClusterConfig.KubeAdminPass, result.ClusterConfig.ClusterAPI)
 	}
 	if consolePrintURL || consolePrintCredentials {
 		return
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -378,7 +378,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		logging.Infof("")
 		logging.Infof("To access the cluster, first set up your environment by following 'crc oc-env' instructions")
 		logging.Infof("Then you can access it by running 'oc login -u developer -p developer %s'", result.ClusterConfig.ClusterAPI)
-		logging.Infof("To login as an admin, username is 'kubeadmin' and password is %s", result.ClusterConfig.KubeAdminPass)
+		logging.Infof("To login as an admin, run 'oc login -u kubeadmin -p %s %s'", result.ClusterConfig.KubeAdminPass, result.ClusterConfig.ClusterAPI)
 		logging.Infof("")
 		logging.Infof("You can now run 'crc console' and use these credentials to access the OpenShift web console")
 	}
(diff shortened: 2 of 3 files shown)


gbraad:
Please create an issue to refer to and explain the need and what is expected. This helps with traceability and testing pruposes.

gbraad:
@code-ready/crc-devel WDYT? Do we need to enforce the use of issues?

gbraad:
OK to test

cfergeau:
Can you also update the strings in https://github.com/code-ready/crc/blob/master/cmd/crc/cmd/console.go#L62  ?

rinormaloku:
@cfergeau 
Do you want the outputs on the machine and console to be identical?

Previously they differed and only this one [code-line](https://github.com/code-ready/crc/blob/master/pkg/crc/machine/machine.go#L380) specified an actual kubectl command.

I am asking as I don't know where the second command is displayed (meaning https://github.com/code-ready/crc/blob/master/pkg/crc/machine/machine.go#L380)

Will fix the test, missed that one. Thanks!

cfergeau:
> @cfergeau
> Do you want the outputs on the machine and console to be identical?
> 
> Previously they differed and only this one [code-line](https://github.com/code-ready/crc/blob/master/pkg/crc/machine/machine.go#L380) specified an actual kubectl command.
> 

Yes they were different, but I think they should display the same message.

> I am asking as I don't know where the second command is displayed (meaning https://github.com/code-ready/crc/blob/master/pkg/crc/machine/machine.go#L380), and as such want your confirmation.
> 

The machine.go one (which is your initial change) appears at the end of `crc start`. The console.go messages are shown with `crc console --credentials` when the VM is running.

> Will fix the test, missed that one. Thanks!

The test change is only needed if you change console.go.  For your machine.go change, changing the test would break CI :) (the test checks `crc console --credentials`, not the output of `crc start`).

rinormaloku:
@cfergeau made the changes in `console.go` and in the tests. ✌️

cfergeau:
This fails CI, you need to use `out.Outf` rather than `out.Outln`. The messages passed to `out.Outf` need to be terminated with `\n`.

cfergeau:
> This fails CI, you need to use `out.Outf` rather than `out.Outln`. The messages passed to `out.Outf` need to be terminated with `\n`.
With that fixed, ACK from me.

gbraad:
Please add issue number #878 to commit messages.

cfergeau:
I've addressed these misc issues in https://github.com/cfergeau/crc/tree/oc-login
But cannot easily suggest these changes in this PR :(

gbraad:
@cfergeau not sure what to do as you have added these commits to a branch on your repo?

cfergeau:
> @cfergeau not sure what to do as you have added these commits to a branch on your repo?

I also have no idea how this is supposed to work in this PR workflow... A PR seems strongly tied to its initial branch, so the person owning the branch would have to take these changes if we want to add them to the PR? The other alternative I can think of is to open a new PR, but that means losing the discussion we had here.

gbraad:
PRs are owned and when merged will close this and the possible associated issue.

When someone rebases on top of a PR, this link is gone. The person will still be associated as the owner, but the PR will end up as being unmerged.

rinormaloku:
@gbraad added the Issue in the commit and squashed ✌️
@cfergeau fixed the Outf 🤦‍♂️☺️

cmd/crc/cmd/console.go
```diff
@@ -60,8 +60,8 @@ func runConsole(arguments []string) {
 		output.Outln(result.ClusterConfig.WebConsoleURL)
 	}
 	if consolePrintCredentials {
-		output.Outln("To login as a regular user, username is 'developer' and password is 'developer'.")
-		output.Outf("To login as an admin, username is 'kubeadmin' and password is '%s'.\n", result.ClusterConfig.KubeAdminPass)
+		output.Outf("To login as a regular user, run 'oc login -u developer -p developer %s'.", result.ClusterConfig.ClusterAPI)
+		output.Outf("To login as an admin, run 'oc login -u kubeadmin -p %s %s'", result.ClusterConfig.KubeAdminPass, result.ClusterConfig.ClusterAPI)
```
cfergeau: You need both strings to end with `\n`

rinormaloku: done ;)

rinormaloku:
> Please add issue number to commit messages for changelog and traceability.

@gbraad commit messages are updated to have the issue number

Issue #854, opened 2019-11-28
Add host->cluster DNS checks


cfergeau:
At the moment we check VM->cluster DNS, VM->external DNS, but I could not find a startup check that the host can resolve `api.crc.testing` and `non-existing-app.apps-crc.testing` after the dnsmasq container was started.
We should add such a check, and verify that it matches the VM IP. Not sure whether this should abort `crc start` or just be a warning though.

----

  * [ ] check `api.crc.testing`
  * [ ] check `foo.crc-apps.testing`

gbraad:
Unfortunately this check has been reverted.

cfergeau:
This was reverted because of https://github.com/code-ready/crc/issues/893
Since our builds have CGO_ENABLED=0, the binary does not make use of `/etc/resolver/` for DNS resolution, but relies on hardcoded entries in `/etc/hosts`, so checking for foo.crc.testing fails.

cfergeau:
As indicated in #918, for now we could have a check for only `api.crc.testing`.
Regarding the check for `foo.crc-apps.testing`, we can do it only on linux/windows to avoid the issue we have on macos

gbraad:
See first message

PR #869, opened 2019-12-06
Refactor to set pullsecret and clusterID

praveenkumar:
This refactor allow us to perform more cluster resource changes like
add proxy, insecure registry etc. at cluster level.

Note: Now we are going to depend on the dns even before approving the csr so if local dns is not working then CRC not even able to start.


pkg/crc/machine/machine.go
```diff
@@ -675,3 +671,33 @@ func updateSSHKeyPair(sshRunner *crcssh.SSHRunner) error {
...
+		if _, err := sd.Stop("kubelet"); err != nil {
+			m.Collect(err)
+		}
+		if err := cluster.StopAndRemovePodsInVM(sshRunner); err != nil {
+			m.Collect(err)
+		}
+		rerr = m.ToError()
+	}()
+
+	// Update the user pull secret before kubelet start.
```
cfergeau: Looks like outdated comment? The kubelet was started just before.

cfergeau:
One thing I forgot to mention, if we start to rely more on the host->VM DNS being functional, then it's a good opportunity to fix https://github.com/code-ready/crc/issues/854 to ensure it's there before we try to use it.

praveenkumar:
> One thing I forgot to mention, if we start to rely more on the host->VM DNS being functional, then it's a good opportunity to fix #854 to ensure it's there before we try to use it.

@cfergeau yes that's right.

pkg/crc/machine/machine.go
praveenkumar: @cfergeau By default the pull secret set to null string in our bundle so logically we are updating the pull secret but since on UX side if we say updating the pull secret then user might think that there is already existing pull secret so we just say `Adding user's pull secret` :)

cfergeau: My problem with this comment is `before kubelet start` since the code is:
```
if _, err := sd.Start("kubelet"); err != nil {
		return fmt.Errorf("Error starting kubelet: %s", err)
}
//...
// Update the user pull secret before kubelet start.
```

pkg/crc/cluster/cluster.go
```diff
@@ -77,3 +92,62 @@ func GetRootPartitionUsage(sshRunner *ssh.SSHRunner) (int64, int64, error) {
...
+func AddPullSecret(sshRunner *ssh.SSHRunner, oc oc.OcConfig, pullSec string) error {
+	if err := addPullSecretToInstanceDisk(sshRunner, pullSec); err != nil {
+		return err
+	}
+
+	base64OfPullSec := base64.StdEncoding.EncodeToString([]byte(pullSec))
+	cmdArgs := []string{"patch", "secret", "pull-secret", "-p",
+		fmt.Sprintf(`{"data":{".dockerconfigjson":"%s"}}`, base64OfPullSec),
+		"-n", "openshift-config", "--type", "merge"}
+	logging.Debugf("Executing oc %s\n", strings.Join(cmdArgs, " "))
```
cfergeau: This belongs to oc.RunOcCommand rather than here if you want such a log.
`logging.Debugf("Patching cluster config for new pull secret")` for a debug log which goes in this method.
This is also leaking the pull secret to debug logs.

pkg/crc/cluster/cluster.go
```diff
@@ -77,3 +92,62 @@ func GetRootPartitionUsage(sshRunner *ssh.SSHRunner) (int64, int64, error) {
...
+	addPullSecret := func() error {
+		_, stderr, err := oc.RunOcCommand(cmdArgs...)
+		if err != nil {
+			logging.Debug(stderr)
+			return &errors.RetriableError{Err: err}
+		}
+		return nil
+	}
+	return errors.RetryAfter(80, addPullSecret, time.Second)
```
cfergeau: I'd prefer if we did not have this somehow arbitrary RetryAfter in these methods. A prerequisite for calling this methods is that the api server is up, if needed we can add a method waiting for the api server to be up, and which would contain such a RetryAfter call.

pkg/crc/cluster/cluster.go
```diff
@@ -77,3 +92,62 @@ func GetRootPartitionUsage(sshRunner *ssh.SSHRunner) (int64, int64, error) {
...
 	return diskSize, diskUsage, nil
 }
+
+func AddPullSecret(sshRunner *ssh.SSHRunner, oc oc.OcConfig, pullSec string) error {
+	if err := addPullSecretToInstanceDisk(sshRunner, pullSec); err != nil {
+		return err
+	}
+
+	base64OfPullSec := base64.StdEncoding.EncodeToString([]byte(pullSec))
+	cmdArgs := []string{"patch", "secret", "pull-secret", "-p",
```
cfergeau: This used to be `oc replace`, not sure why you switched to `oc patch`?

pkg/crc/network/utils.go
```diff
@@ -80,3 +82,17 @@ func DetermineHostIP(instanceIP string) (string, error) {
 
 	return "", errors.New("unknown error occurred")
 }
+
+func CheckCRCLocalDNSReachableFromHost(crcbundleMetadata *bundle.CrcBundleInfo) error {
+	ips, err := net.LookupIP(fmt.Sprintf("api.%s.%s", crcbundleMetadata.ClusterInfo.ClusterName, crcbundleMetadata.ClusterInfo.BaseDomain))
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau since we first check already on the VM which make sure that dnsmasq is running on the VM so on the host it should be good without retry.

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau `oc replace` need whole file to be available but since resource already available we just need to use patch to update it.

pkg/crc/machine/machine.go
praveenkumar: @cfergeau Ah I got it now, I will remove the `before kublet start` part from comment.

pkg/crc/network/utils.go
```diff
@@ -92,3 +94,17 @@ func UriStringForDisplay(uri string) (string, error) {
 	}
 	return uri, nil
 }
+
+func CheckCRCLocalDNSReachableFromHost(crcbundleMetadata *bundle.CrcBundleInfo) error {
+	ip, err := net.LookupIP(fmt.Sprintf("api.%s.%s", crcbundleMetadata.ClusterInfo.ClusterName, crcbundleMetadata.ClusterInfo.BaseDomain))
+	if err != nil {
+		return err
+	}
+	logging.Debugf("api.crc.testing resolved to %s", ip)
```
cfergeau: `logging.Debugf("%s resolved to %s", domainName, ip)`

pkg/crc/network/utils.go
```diff
@@ -92,3 +94,17 @@ func UriStringForDisplay(uri string) (string, error) {
...
+	ip, err = net.LookupIP(fmt.Sprintf("foo.%s", crcbundleMetadata.ClusterInfo.AppsDomain))
+	if err != nil {
+		return err
+	}
+	logging.Debugf("foo.apps-crc.testing resolved to %s", ip)
```
cfergeau: logging.Debugf("%s resolved to %s", appsDomainName, ip)

pkg/crc/network/utils.go
```diff
@@ -92,3 +94,17 @@ func UriStringForDisplay(uri string) (string, error) {
 	}
 	return uri, nil
 }
+
+func CheckCRCLocalDNSReachableFromHost(crcbundleMetadata *bundle.CrcBundleInfo) error {
```
cfergeau: Wondering if we need the dependency on the `bundle`package, or if this should be
` func CheckCRCLocalDNSReachableFromHost(domainName, appsDomainName string) error` ?

pkg/crc/machine/machine.go
```diff
@@ -682,3 +678,32 @@ func updateSSHKeyPair(sshRunner *crcssh.SSHRunner) error {
 
 	return err
 }
+
+func configureCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, pullSecret string) (rerr error) {
+	m := errors.MultiError{}
```
cfergeau: This can be moved in the defer lambda, as it's only used there.

pkg/crc/machine/machine.go
```diff
@@ -682,3 +678,32 @@ func updateSSHKeyPair(sshRunner *crcssh.SSHRunner) error {
...
+
+	defer func() {
+		// Stop the kubelet service.
+		if _, err := sd.Stop("kubelet"); err != nil {
+			m.Collect(err)
+		}
+		if err := cluster.StopAndRemovePodsInVM(sshRunner); err != nil {
+			m.Collect(err)
+		}
+		rerr = m.ToError()
```
cfergeau: wondering if we really need a MultiError here though .If sd.Stop() fails, it's very llikely cluster.StopAndRemovePodsInVM will fail too, so we could just report the sd.Stop() error in that case? And if sd.Stop succeeds, then we only have the StopAndRemovePodsInVM error to potentially report.

pkg/crc/cluster/cluster.go
```diff
@@ -77,3 +81,79 @@ func GetRootPartitionUsage(sshRunner *ssh.SSHRunner) (int64, int64, error) {
...
+func waitForOpenshiftAPIServer(oc oc.OcConfig) error {
+	waitForApiServer := func() error {
+		_, stderr, err := oc.RunOcCommand("get", "pods")
+		if err != nil {
+			logging.Debug(stderr)
+			return &errors.RetriableError{Err: err}
+		}
+		return nil
+	}
+	return errors.RetryAfter(80, waitForApiServer, time.Second)
```
cfergeau: Isn't 80s very long?

praveenkumar: @cfergeau no that what we had before also so I am not changing it and then break something :)

PR #871, opened 2019-12-09
Update proc_starting-the-virtual-machine.adoc

### LINKED ISSUE #870: Bundle 'crc_hyperkit_4.2.8.crcbundle' was requested, but the existing VM is using 'crc_hyperkit_4.2.2.crcbundle'
Environment info   
  System Version:	macOS 10.15.1 (19B88)
  Kernel Version:	Darwin 19.0.0

Running as non-root

Downloaded and extracted CRC 1.2 (crc-macos-amd64.tar.xz) from https://cloud.redhat.com/openshift/install/crc/installer-provisioned.

After adding the extracted location to my path, I ran 'crc setup'
mbp-3:crc-macos-1.2.0-amd64 user1$ crc setup
INFO Caching oc binary                            
INFO Setting up virtualization with HyperKit      
INFO Installing crc-machine-hyperkit              
INFO Setting file permissions for /etc/resolver/testing 
INFO Setting file permissions for /etc/hosts      
Setup is complete, you can now run 'crc start' to start the OpenShift cluster

Then ran 'crc start'

mbp-3:crc-macos-1.2.0-amd64 user1$ crc start
INFO Checking if HyperKit is installed            
INFO Checking if crc-driver-hyperkit is installed 
INFO Checking file permissions for /etc/hosts     
FATA Bundle 'crc_hyperkit_4.2.8.crcbundle' was requested, but the existing VM is using 'crc_hyperkit_4.2.2.crcbundle'

erpenbach:
Fixes #870 

added Note to setup instructions to account for problem such as  "Bundle 'crc_hyperkit_4.2.8.crcbundle' was requested, but the existing VM is using 'crc_hyperkit_4.2.2.crcbundle'"

### DIFF
--- a/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
+++ b/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
@@ -38,5 +38,5 @@ $ {bin} start --vm-driver virtualbox --bundle _path_to_system_bundle_
 [NOTE]
 ====
 * The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
-* If errors are encountered on the start command due to a mismatch in the hyperkit bundle, use the '{bin} delete' command to remove older versions.  These older versions may have user data and care should be taken before issuing the command to insure past work is not lost. 
+* If errors are encountered on the start command due to a mismatch in dependencies (such as the hyperkit bundle), use the '{bin} delete' command to remove older versions.  These older versions may have user data and care should be taken before issuing the command to insure past work is not lost. 
 ====
--- a/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
+++ b/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
@@ -37,5 +37,6 @@ $ {bin} start --vm-driver virtualbox --bundle _path_to_system_bundle_
 
 [NOTE]
 ====
-The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
+* The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
+* If errors are encountered on the start command due to a mismatch in the hyperkit bundle, use the '{bin} delete' command to remove older versions.  These older versions may have user data and care should be taken before issuing the command to insure past work is not lost. 
 ====
--- a/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
+++ b/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
@@ -38,5 +38,5 @@ $ {bin} start --vm-driver virtualbox --bundle _path_to_system_bundle_
 [NOTE]
 ====
 * The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
-* If errors are encountered on the start command due to a mismatch in dependencies (such as the hyperkit bundle), use the '{bin} delete' command to remove older versions.  These older versions may have user data and care should be taken before issuing the command to insure past work is not lost. 
+* If errors are encountered on the start, check the Troubleshooting section for potential solutions. 
 ====
--- a/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
+++ b/docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
@@ -38,5 +38,5 @@ $ {bin} start --vm-driver virtualbox --bundle _path_to_system_bundle_
 [NOTE]
 ====
 * The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
-* If errors are encountered on the start, check the Troubleshooting section for potential solutions. 
+* If errors are encountered on the start, check the <<troubleshooting-codeready-containers_{context}>> section for potential solutions. 
 ====
--- a/docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
+++ b/docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
@@ -1,7 +1,10 @@
 [id="troubleshooting-bundle-version-mismatch_{context}"]
 = Troubleshooting bundle version mismatch
 
-As {prod} releases are set up and started, instances are created with bundle information and instance data.  When a new {prod} release is setup, the bundle information and instance data is not updated due to the customization in the previous instance data.  This will lead to errors on [command]`{bin} start` command. The error will be similar to this:
+As {prod} releases are set up and started, instances are created with bundle information and instance data.  
+When a new {prod} release is setup, the bundle information and instance data is not updated due to the customization in the previous instance data.  
+This will lead to errors on [command]`{bin} start` command. 
+The error will be similar to this:
 
 ----
 $ {bin} start
@@ -26,6 +29,7 @@ $ {bin} delete
 
 [NOTE]
 ====
-The [command]`{bin} delete` command will result in the loss of data stored in the {prod} virtual machine. Steps should be taken to save any instance information before issuing the command. 
+The [command]`{bin} delete` command will result in the loss of data stored in the {prod} virtual machine. 
+Steps should be taken to save any instance information before issuing the command. 
 ====
 
--- a/docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
+++ b/docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
@@ -1,10 +1,7 @@
 [id="troubleshooting-bundle-version-mismatch_{context}"]
 = Troubleshooting bundle version mismatch
 
-As {prod} releases are setup and started, instances are created with bundle information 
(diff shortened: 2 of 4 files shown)


docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
```diff
@@ -37,5 +37,6 @@ $ {bin} start --vm-driver virtualbox --bundle _path_to_system_bundle_
 
 [NOTE]
 ====
-The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
+* The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
+* If errors are encountered on the start command due to a mismatch in the hyperkit bundle, use the '{bin} delete' command to remove older versions.  These older versions may have user data and care should be taken before issuing the command to insure past work is not lost.
```
gbraad: This is not specifc to Hyperkit, so this is incorrect.

erpenbach: I modified the message to generalize for other dependencies and referenced the hyperkit bundle as an example.  I could make it more specific if there is a set list of dependencies, however I expect this could change over time.  Also I would expect that the support on start or setup handles this better in the future to the point where this not is no longer needed.  Rather than wait until that point, I wanted to get something else out for others to find to avoid spending time trying to resolve this.

docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
```diff
@@ -38,5 +38,5 @@ $ {bin} start --vm-driver virtualbox --bundle _path_to_system_bundle_
 [NOTE]
 ====
 * The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
-* If errors are encountered on the start command due to a mismatch in the hyperkit bundle, use the '{bin} delete' command to remove older versions.  These older versions may have user data and care should be taken before issuing the command to insure past work is not lost. 
+* If errors are encountered on the start command due to a mismatch in dependencies (such as the hyperkit bundle), use the '{bin} delete' command to remove older versions.  These older versions may have user data and care should be taken before issuing the command to insure past work is not lost.
```
cfergeau: I would put this in the troubleshooting section, where there is more space and a sample message can be added.
"dependencies" is not the right term. "A mismatch in bundle versions" maybe. I would not mention at all the hyperkit bundle.
You have an extra space before "These older". I believe it's "to ensure past work...", not "to insure"

erpenbach:
I've created a section in the Troubleshooting chapter and also added a Note in the crc setup section.  Without a way to handle the data from previous versions, documenting this has spread across a number of places. 
It might be better to document this in the "Known Issues" section instead, especially if this is a short term issue.  This way it can still be found and doesn't intrude across the documentation.  Please let me know what you prefer.

docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
erpenbach: I've created a section in the Troubleshooting chapter and also added a Note in the crc setup section. Without a way to handle the data from previous versions, documenting this has spread across a number of places.
It might be better to document this in the "Known Issues" section instead, especially if this is a short term issue. This way it can still be found and doesn't intrude across the documentation. Please let me know what you prefer.

docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
```diff
@@ -37,5 +37,6 @@ $ {bin} start --vm-driver virtualbox --bundle _path_to_system_bundle_
 
 [NOTE]
 ====
-The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
+* The cluster takes a minimum of four minutes to start the necessary containers and operators before serving a request.
+* If errors are encountered on the start, check the Troubleshooting section for potential solutions.
```
robin-owen: Could you link (`xref`) the troubleshooting section here so that the reader doesn't have to go looking for it? :)

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
```diff
@@ -0,0 +1,34 @@
+[id="troubleshooting-bundle-version-mismatch_{context}"]
+= Troubleshooting bundle version mismatch
+
+As {prod} releases are setup and started, instances are created with bundle information
```
robin-owen: "set up" instead of "setup". This applies generally.

Also, could you please keep each sentence on its own line? We don't particularly care about the line length, but want to keep each sentence isolated so that diffs are cleaner. :)

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
```diff
@@ -0,0 +1,34 @@
+[id="troubleshooting-bundle-version-mismatch_{context}"]
+= Troubleshooting bundle version mismatch
+
+As {prod} releases are setup and started, instances are created with bundle information 
+and instance data.  When a new {prod} release is setup, the bundle information and instance data
+is not updated due to the customization in the previous instance data.  This will lead to errors on 
+[command]`{bin} start` command 
+
+----
```
robin-owen: What's the point of this source block? Can we clarify that by adding a bit of text before the block appears?

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
```diff
@@ -0,0 +1,34 @@
...
+To resolve the error, use the [command]`{bin} delete` command:
+
+[subs="+quotes,attributes"]
+----
+$ {bin} delete
+----
+
+[NOTE]
+====
+The [command]`{bin} delete` command will result in the loss of data stored in the {prod} virtual machine.
```
robin-owen: This isn't a note about this PR specifically, but it does feel like maybe we could use this information in a snippet that can be reused through the docs. It appears a few times.

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
cfergeau: Isn't this explained by `This will lead to errors on [command]`{bin} start` command`?

cfergeau:
https://github.com/cfergeau/crc/commits/pr871 should have the changes requested by @robin-owen

docs/source/getting-started/content/proc_starting-the-virtual-machine.adoc
erpenbach: I've added a link to the troubleshooting section

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
erpenbach: I've added a clarification.

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
erpenbach: I've made this more specific to the situation

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
```diff
@@ -0,0 +1,31 @@
+[id="troubleshooting-bundle-version-mismatch_{context}"]
+= Troubleshooting bundle version mismatch
+
+As {prod} releases are set up and started, instances are created with bundle information and instance data.  When a new {prod} release is setup, the bundle information and instance data is not updated due to the customization in the previous instance data.  This will lead to errors on [command]`{bin} start` command. The error will be similar to this:
```
robin-owen: Could you please put each sentence on its own line?

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
```diff
@@ -0,0 +1,31 @@
...
+To resolve the error, issue the [command]`{bin} delete` command before attempting to start the instances:
+
+[subs="+quotes,attributes"]
+----
+$ {bin} delete
+----
+
+[NOTE]
+====
+The [command]`{bin} delete` command will result in the loss of data stored in the {prod} virtual machine. Steps should be taken to save any instance information before issuing the command.
```
robin-owen: Likewise, each sentence on its own line, please. :)

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
erpenbach: Moved sentences to their own line.

docs/source/getting-started/content/proc_troubleshooting-bundle-version-mismatch.adoc
erpenbach: Moved sentences to their own line.

Issue #873, opened 2019-12-09
[BUG] Hide User pull secret even using debug option


praveenkumar:
User pull secret shouldn't be shown even debug log is enabled because most of the time not aware of what all the details debug log have and might be put on the issue tracker.

PR #875, opened 2019-12-09
Issue #873 Move ssh command invocation from machine to sshRunner

praveenkumar:
`RunSSHCommandFromDriver` function is part of driver util in the
machine code and can be extracted to sshRunner so we can make required
changes here instead having it part of machine codebase.

### DIFF
--- a/pkg/crc/ssh/ssh.go
+++ b/pkg/crc/ssh/ssh.go
@@ -1,7 +1,9 @@
 package ssh
 
 import (
+	"fmt"
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/machine/libmachine/drivers"
 )
 
@@ -20,9 +22,29 @@ func CreateRunnerWithPrivateKey(driver drivers.Driver, privateKey string) *SSHRu
 
 // Create a host using the driver's config
 func (runner *SSHRunner) Run(command string) (string, error) {
-	return drivers.RunSSHCommandFromDriver(runner.driver, runner.privateSSHKey, command)
+	return runner.runSSHCommandFromDriver(command)
 }
 
 func (runner *SSHRunner) SetPrivateKeyPath(path string) {
 	runner.privateSSHKey = path
+}
+
+func (runner *SSHRunner) runSSHCommandFromDriver(command string) (string, error) {
+	client, err := drivers.GetSSHClientFromDriver(runner.driver, runner.privateSSHKey)
+	if err != nil {
+		return "", err
+	}
+
+	logging.Debugf("About to run SSH command:\n%s", command)
+
+	output, err := client.Output(command)
+	logging.Debugf("SSH cmd err, output: %v: %s", err, output)
+	if err != nil {
+		return "", fmt.Errorf(`ssh command error:
+command : %s
+err     : %v
+output  : %s`, command, err, output)
+	}
+
+	return output, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/crc/ssh/ssh.go
```diff
@@ -20,9 +22,29 @@ func CreateRunnerWithPrivateKey(driver drivers.Driver, privateKey string) *SSHRu
...
 func (runner *SSHRunner) Run(command string) (string, error) {
-	return drivers.RunSSHCommandFromDriver(runner.driver, runner.privateSSHKey, command)
+	return runSSHCommandFromDriver(runner.driver, runner.privateSSHKey, command)
 }
 
 func (runner *SSHRunner) SetPrivateKeyPath(path string) {
 	runner.privateSSHKey = path
 }
+
+func runSSHCommandFromDriver(d drivers.Driver, sshPrivateKeyPath, command string) (string, error) {
```
cfergeau: `func (runner *SSHRunner) runSSHCommandFromDriver(command string) (string, error) {}`

pkg/crc/ssh/ssh.go
```diff
@@ -20,9 +22,29 @@ func CreateRunnerWithPrivateKey(driver drivers.Driver, privateKey string) *SSHRu
...
+	client, err := drivers.GetSSHClientFromDriver(d, sshPrivateKeyPath)
+	if err != nil {
+		return "", err
+	}
+
+	logging.Debugf("About to run SSH command:\n%s", command)
+
+	output, err := client.Output(command)
+	logging.Debugf("SSH cmd err, output: %v: %s", err, output)
```
cfergeau: imo that's a good opportunity to cleanup a bit the debug output which is far too verbose.

pkg/crc/ssh/ssh.go
praveenkumar: @cfergeau why do you want this function to part of `SSHRunner` struct? I thought this is something we are not exposing as part of `CreateRunner` or `CreateRunnerWithPrivateKey` invocation?

pkg/crc/ssh/ssh.go
praveenkumar: @cfergeau anything particular in mind since it show what command is going to run and then tell about error/output of that command, I think that much info we do need right?

pkg/crc/ssh/ssh.go
cfergeau: The main benefit is that we don't need to pass sshPrivateKey and drivers.Driver separately, they are only known.

pkg/crc/ssh/ssh.go
cfergeau: Ah the logs I have in mind might come from libmachine generic rpc code, I'd have to look again at a debug log to remember what I find too verbose :)

pkg/crc/ssh/ssh.go
```diff
@@ -22,23 +22,27 @@ func CreateRunnerWithPrivateKey(driver drivers.Driver, privateKey string) *SSHRu
...
+func (runner *SSHRunner) runSSHCommandFromDriver(command string, runPrivate bool) (string, error) {
 	client, err := drivers.GetSSHClientFromDriver(runner.driver, runner.privateSSHKey)
 	if err != nil {
 		return "", err
 	}
 
-	logging.Debugf("About to run SSH command:\n%s", command)
+	if !runPrivate {
+		logging.Debugf("About to run SSH command:\n%s", command)
+	}
```
cfergeau: We still want some logs indicating something is being done ("About to run SSH command with output hidden")

praveenkumar: @cfergeau Done.

Issue #877, opened 2019-12-10
[Integration tests] Absolute path on Windows does not begin with a (back)slash


jsliacan:
### General information

  * OS: Windows

### Steps to reproduce

  1. Run `make integration`

### Expected
Integration tests run without a hiccup.

### Actual
```
Error occurred loading data files: open \C:\Jenkins\workspace\crc_pipeline\crc\test\testdata: The filename, directory name, or volume label syntax is incorrect.FAIL	github.com/code-ready/crc/test/integration	0.920s
```

This occurs because [here](https://github.com/code-ready/crc/blob/fc30212d0a0705996e1bdd9b81e4c129fd4bf108/test/integration/crcsuite/prepare.go#L70) we prepend a path with filepath separator to make it absolute. This doesn't work on Windows.

jsliacan:
Addressed in #879

PR #879, opened 2019-12-10
Issue #877 Integration: Absolute path starts without separator on Win…

jsliacan:
…dows

### DIFF
--- a/test/integration/crcsuite/prepare.go
+++ b/test/integration/crcsuite/prepare.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/code-ready/crc/pkg/download"
@@ -67,9 +68,14 @@ func CopyFilesToTestDir() {
 	}
 
 	l := strings.Split(cwd, string(filepath.Separator))
-	dataDirPieces := append([]string{string(filepath.Separator)}, l[:len(l)-3]...)
-	dataDirPieces = append(dataDirPieces, "testdata")
+	dataDirPieces := append(l[:len(l)-3], "testdata")
+	var volume string
+	if runtime.GOOS == "windows" {
+		volume = filepath.VolumeName(cwd)
+		dataDirPieces = dataDirPieces[1:] // drop volume from list of dirs
+	}
 	dataDir := filepath.Join(dataDirPieces...)
+	dataDir = fmt.Sprintf("%s%c%s", volume, filepath.Separator, dataDir) // prepend volume back
 
 	files, err := ioutil.ReadDir(dataDir)
 	if err != nil {


test/integration/crcsuite/prepare.go
```diff
@@ -67,7 +68,10 @@ func CopyFilesToTestDir() {
 	}
 
 	l := strings.Split(cwd, string(filepath.Separator))
-	dataDirPieces := append([]string{string(filepath.Separator)}, l[:len(l)-3]...)
+	dataDirPieces := l[:len(l)-3]
+	if runtime.GOOS != "windows" {
+		dataDirPieces = append([]string{string(filepath.Separator)}, dataDirPieces...)
+	}
```
cfergeau: Probably slightly more readable as
```
	dataDirPieces := l[:len(l)-3]
        dataDirPieces = append(dataDirPieces, "testdata")
	dataDir := filepath.Join(dataDirPieces...)
	if runtime.GOOS != "windows" {
           dataDir:= filepath.Join("/", dataDir)
        }
```

jsliacan: Thanks! Right now I have a different problem though: it turns out that `filepath.Join` doesn't like `:\` and joins `C:` and `Users` into `C:Users`. I need to get around this somehow, that's why it's WIP for now. It doesn't work at all as it is.

cfergeau: Might be easier to just get back to the initial code (filepath.Base 3 times)

jsliacan: WDYT about this version? (not sure why it doesn't show as 'outdated') I tested it on Windows and it seemed to work, but I'll do more testing tomorrow.

test/integration/crcsuite/prepare.go
```diff
@@ -67,9 +68,14 @@ func CopyFilesToTestDir() {
...
-	dataDirPieces := append([]string{string(filepath.Separator)}, l[:len(l)-3]...)
-	dataDirPieces = append(dataDirPieces, "testdata")
+	dataDirPieces := append(l[:len(l)-3], "testdata")
+	var volume string
+	if runtime.GOOS == "windows" {
+		volume = filepath.VolumeName(cwd)
+		dataDirPieces = dataDirPieces[1:] // drop volume from list of dirs
+	}
 	dataDir := filepath.Join(dataDirPieces...)
+	dataDir = fmt.Sprintf("%s%c%s", volume, filepath.Separator, dataDir) // prepend volume back
```
cfergeau: `dataDir := filepath.Join(volume, filepath.Separator, dataDir)`? or is this not working as expected when volume is empty?

jsliacan: It should work fine with any volume (didn't test), but it's a hack: we pretend that `\` is a folder name, then `filepath.Join` creates a path `C:\\\datadir` and simplifies it to `C:\datadir`. We'd also have to cast `filepath.Separator` to `string`. Going to test it properly now...

cfergeau: Yet another alternative on top of the current code is to set `volume` to `filepath.Separator` in the linux case, and to use `dataDir := filepath.Join(volume, dataDirPieces...)`

cfergeau:
retest this please

test/integration/crcsuite/prepare.go
jsliacan: @teuf , since `filepath.Join` refuses to take two arguments: `string` and `[]string`, it seems that we have a choice between these two code blocks
```
if runtime.GOOS != "windows" {
	volume = string(filepath.Separator)
}
dataDirPieces = append([]string{volume}, dataDirPieces...)
dataDir := filepath.Join(dataDirPieces...)
```
and the current
```
	volume = filepath.VolumeName(cwd)
	dataDirPieces = dataDirPieces[1:]
}
It would be cleaner if we didn't need L4 in the first block, but I couldn't do it without first prepending `volume` to `dataDirPieces`. Since it's not cleaner the other way, I'd just leave it as it is and merge it (it's tested on all platforms). Unless I'm missing some shortcut! WDYT? 
dataDir := filepath.Join(dataDirPieces...)
dataDir = fmt.Sprintf("%s%c%s", volume, filepath.Separator, dataDir)
```

cfergeau: > since `filepath.Join` refuses to take two arguments: `string` and `[]string`,

Did you try with `filepath.Separator(volume, dataDirPieces...)` (3 trailing dots)? This should 'convert' the `[]string` to something appropriate  for `filepath.Join()`

jsliacan: You meant `filepath.Join(volume, dataDirPieces...)`? Here's what I had:
```
l := strings.Split(cwd, string(filepath.Separator))
dataDirPieces := append(l[:len(l)-3], "testdata")

var volume string
if runtime.GOOS != "windows" {
	volume = string(filepath.Separator)
}
dataDir := filepath.Join(volume, dataDirPieces...)
```
and here's what I got when I ran `make integration` with it
```
test/integration/crcsuite/prepare.go:78:26: too many arguments in call to filepath.Join
	have (string, []string...)
	want (...string)
```

jsliacan: It looks like `filepath.Join` is not happy with a "loose bunch of arguments + 1", but is happy with a "loose bunch of arguments".

cfergeau: Indeed, I did not expect that ;) 
Reproducible with
```
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
       a := []string{"a", "b"}
       b := "c"
	fmt.Println(filepath.Join(b, a...))
}
```

cfergeau:
> Assuming this works on windows, I'm fine with landing this as is, or with small additional changes :)

Still fine with landing this as long as you are happy with it, and it succeeds on windows.

jsliacan:
> Still fine with landing this as long as you are happy with it, and it succeeds on windows.

@cfergeau, I just ran some tests on Windows with these changes and it seems to be doing what we want. Should be OK. Thanks

cfergeau:
> > Still fine with landing this as long as you are happy with it, and it succeeds on windows.
> 
> @cfergeau, I just ran some tests on Windows with these changes and it seems to be doing what we want. Should be OK. Thanks

Let's merge this then ^^

PR #881, opened 2019-12-10
Update README for automatic certificate renewal in 1.2.0

CoaxVex:


cfergeau:
Now that it's hopefully no longer be something that is user-visible, I'd just remove that section from the README. This is already in the documentation.

CoaxVex:
Sure, here is an update with the note removed.

PR #884, opened 2019-12-12
Issue #337 Add proxy for cluster and instance

praveenkumar:


pkg/crc/cluster/cluster.go
```diff
@@ -138,6 +139,41 @@ func StopAndRemovePodsInVM(sshRunner *ssh.SSHRunner) error {
...
+	}
+	return nil
+}
+
+func AddProxyToKubeletAndCriO(sshRunner *ssh.SSHRunner, proxy *network.ProxyConfig) error {
+	proxyTemplate := `[Service]
+Environment=HTTP_PROXY=%s
+Environment=HTTPS_PROXY=%s
+Environment=NO_PROXY=%s`
+	defaultNoProxyForCluster := ".cluster.local,.svc,10.128.0.0/14,172.30.0.0/16"
```
cfergeau: Is this documented somewhere? Maybe it's better to put this directly in the template?

pkg/crc/cluster/cluster.go
```diff
@@ -138,6 +139,41 @@ func StopAndRemovePodsInVM(sshRunner *ssh.SSHRunner) error {
...
+	noProxy := fmt.Sprintf("%s,%s", defaultNoProxyForCluster, proxy.NoProxy)
+
+	p := fmt.Sprintf(proxyTemplate, proxy.HttpProxy, proxy.HttpsProxy, noProxy)
+	_, err := sshRunner.Run(fmt.Sprintf("cat <<EOF | sudo tee /etc/systemd/system/crio.service.d/10-default-env.conf\n%s\nEOF", p))
```
cfergeau: probably worth a comment saying what this is doing (or a small helper with a good name since it's done twice).
Note that this does not have to go to this specific `10-default-env.conf` file, we could also create our own `20-crc-proxy-settings.conf`

pkg/crc/cluster/cluster.go
```diff
@@ -138,6 +139,41 @@ func StopAndRemovePodsInVM(sshRunner *ssh.SSHRunner) error {
...
+		return err
+	}
+	_, stderr, err := oc.RunOcCommand(cmdArgs...)
+	if err != nil {
+		return fmt.Errorf("Failed to add proxy details %v: %s", err, stderr)
+	}
+	return nil
+}
+
+func AddProxyToKubeletAndCriO(sshRunner *ssh.SSHRunner, proxy *network.ProxyConfig) error {
```
cfergeau: might be worth documenting that this will change the systemd unit files to add the environment variables needed for proxy support, and that after calling this, both "kubelet" and "crictl" systemd units need to be restarted.
It's also not clear whether `systemd daemon-reload` needs to be called or not after this https://serverfault.com/questions/413397/how-to-set-environment-variable-in-systemd-service#413408

pkg/crc/machine/machine.go
```diff
@@ -351,7 +351,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 
 	ocConfig := oc.UseOCWithConfig(startConfig.Name)
 	if !exists {
-		if err := configureCluster(ocConfig, sshRunner, pullSecret); err != nil {
+		if err := configureCluster(ocConfig, sshRunner, crcBundleMetadata.ClusterInfo.BaseDomain, pullSecret, instanceIP); err != nil {
```
cfergeau: If the user changes their proxy settings after creating the VM, they will have to go through `crc delete` first. Need to keep that in mind to document it ;)

pkg/crc/cluster/cluster.go
```diff
@@ -138,6 +139,41 @@ func StopAndRemovePodsInVM(sshRunner *ssh.SSHRunner) error {
...
+func AddProxyConfigToCluster(oc oc.OcConfig, proxy *network.ProxyConfig) error {
+	cmdArgs := []string{"patch", "proxy", "cluster", "-p",
+		fmt.Sprintf(`{"spec":{"httpProxy":"%s", "httpsProxy":"%s", "noProxy":"%s"}}`, proxy.HttpProxy, proxy.HttpsProxy, proxy.NoProxy),
+		"-n", "openshift-config", "--type", "merge"}
+
+	if err := waitForOpenshiftAPIServer(oc); err != nil {
+		return err
+	}
+	_, stderr, err := oc.RunOcCommand(cmdArgs...)
+	if err != nil {
```
cfergeau: Nit: in the same function, it would be better to be consistent with error handling :)
```
if err := ...; err != nil {
}
```
VS
```
err := ...
if err != nil {
}
```

pkg/crc/cluster/cluster.go
```diff
@@ -138,6 +139,41 @@ func StopAndRemovePodsInVM(sshRunner *ssh.SSHRunner) error {
...
+Environment=NO_PROXY=%s`
+	defaultNoProxyForCluster := ".cluster.local,.svc,10.128.0.0/14,172.30.0.0/16"
+	noProxy := fmt.Sprintf("%s,%s", defaultNoProxyForCluster, proxy.NoProxy)
+
+	p := fmt.Sprintf(proxyTemplate, proxy.HttpProxy, proxy.HttpsProxy, noProxy)
+	_, err := sshRunner.Run(fmt.Sprintf("cat <<EOF | sudo tee /etc/systemd/system/crio.service.d/10-default-env.conf\n%s\nEOF", p))
+	if err != nil {
+		return err
+	}
+	_, err = sshRunner.RunPrivate(fmt.Sprintf("cat <<EOF | sudo tee /etc/systemd/system/kubelet.service.d/10-default-env.conf\n%s\nEOF", p))
```
cfergeau: why `RunPrivate` ? I don't think there's anything secret in that command?

pkg/crc/machine/machine.go
```diff
@@ -680,23 +680,48 @@ func updateSSHKeyPair(sshRunner *crcssh.SSHRunner) error {
...
 	sd := systemd.NewInstanceSystemdCommander(sshRunner)
 	if _, err := sd.Start("kubelet"); err != nil {
 		return fmt.Errorf("Error starting kubelet: %s", err)
 	}
 
+	proxy, err := network.NewProxyConfig()
+	if err != nil {
+		return err
+	}
+
```
cfergeau: I'd somehow prefer to keep things separated, one helper function to configure pullsecret/cluster id, another one to configure proxy.

pkg/crc/network/proxy.go
```diff
@@ -47,6 +47,8 @@ func NewProxyConfig() (*ProxyConfig, error) {
 
 	if DefaultProxy.NoProxy != "" {
 		np = fmt.Sprintf("%s,%s", np, DefaultProxy.NoProxy)
+		// Remove whitespace from string
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau if a user specify the noproxy variable from the config variable then it might be case they have spaces in between.

pkg/crc/network/proxy.go
```diff
@@ -47,6 +47,8 @@ func NewProxyConfig() (*ProxyConfig, error) {
...
+		np = strings.Replace(np, " ", "", -1)
```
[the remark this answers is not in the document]
praveenkumar: That doesn't all remove intermediate spaces. https://play.golang.org/p/W9FEsu67lCX

pkg/crc/network/proxy.go
```diff
@@ -79,8 +81,9 @@ func (p *ProxyConfig) HttpsProxyForDisplay() string {
 }
 
 // AddNoProxy appends the specified host to the list of no proxied hosts.
-func (p *ProxyConfig) AddNoProxy(host string) {
-	p.NoProxy = fmt.Sprintf("%s,%s", host, p.NoProxy)
+func (p *ProxyConfig) AddNoProxy(hosts string) {
```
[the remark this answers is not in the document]
praveenkumar: >  you want to be able to pass ".example.com,.example2.com"

yes, that was the intention, I should change the signature for this to `host ...string` so it will become a variadic method.

pkg/crc/machine/machine.go
```diff
@@ -680,23 +680,48 @@ func updateSSHKeyPair(sshRunner *crcssh.SSHRunner) error {
...
+		}
 		if err := cluster.StopAndRemovePodsInVM(sshRunner); err != nil {
 			rerr = err
 		}
 		rerr = nil
 	}()
 
+	if proxy.IsEnabled() {
+		logging.Info("Adding proxy configuration to the cluster ...")
+		proxy.AddNoProxy(fmt.Sprintf(".%s,%s", baseDomainName, instanceIP))
```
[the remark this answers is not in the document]
praveenkumar: > Is there anything directly connecting to the instanceIP? 
No, but this proxy also going to part of the kubelet and crio so I don't want to have any surprise if internally anything happen with IP.

>  why don't we add apps-crc.testing too?
We are adding `.testing` so this should take care of all other subdomain right?

pkg/crc/cluster/cluster.go
praveenkumar: > Is this documented somewhere? 

This you can get once the proxy is applied and check the status of it, this does take some time to come-up since some of the operator restart due to the change but below is what it looks like.

`10.128.0.0/14` is the pod network and `172.30.0.0/16` is service network.

```
$ oc get proxy -oyaml
[...]
  status:
    httpProxy: http://<my_proxy>
    httpsProxy: http://<my_proxy>
    noProxy: .cluster.local,.svc,10.128.0.0/14,127.0.0.1,172.30.0.0/16,api-int.crc.testing,api.crc.testing,etcd-0.crc.testing,localhost
[...]
```

> Maybe it's better to put this directly in the template?
@cfergeau sure we can put that in the template.

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau the proxy variable which have password are going to be exposed if we are not use `RunPrivate` on this.

pkg/crc/machine/machine.go
praveenkumar: @cfergeau sure, let me do that change

pkg/crc/cluster/cluster.go
praveenkumar: > Note that this does not have to go to this specific 10-default-env.conf file, we could also create our own 20-crc-proxy-settings.conf

Yes, since this file is used when you enable proxy in production environment, I don't want to make any change in the naming here, if someone look into the VM.

pkg/crc/network/proxy.go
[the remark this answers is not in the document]
cfergeau: "in between"? why isn't this caught in the validation of the config string if this is going to be an issue?

pkg/crc/network/proxy.go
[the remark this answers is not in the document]
cfergeau: I know, but it was not obvious where the spaces were expected to be :)

pkg/crc/network/proxy.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau there is no validation for noproxy, we might add it and ask tell user to not to provide any space in between but I feel this one is better handling our side for it.

pkg/crc/cluster/cluster.go
cfergeau: I don't understand that link that you quote `noProxy: .cluster.local,.svc,10.128.0.0/14,127.0.0.1,172.30.0.0/16,api-int.crc.testing,api.crc.testing,etcd-0.crc.testing,localhost`
Is it you during your testing which specified `api-int.crc.testing`, `etcd-0.crc.testing`? or was it added automatically by some operator?

By "documented", I meant "is there some openshift documentation saying we should do that?" ?

pkg/crc/machine/machine.go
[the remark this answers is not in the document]
cfergeau: > We are adding .testing so this should take care of all other subdomain right?

Ah true, I assumed `baseDomainName` was `crc.testing`, my bad :)

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau there is no doc from openshift side about it, this is taken care by the proxy operator and added automatically but since this `status` doesn't visible just after making the change in the proxy and need bit time so I used these value as constant because those are the one needed for `kubelet` and `crio` both

pkg/crc/network/proxy.go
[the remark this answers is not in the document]
cfergeau: Or we can avoid the problem altogether with
```diff
diff --git a/pkg/crc/config/validations.go b/pkg/crc/config/validations.go
index 9563b8be..ed15ce1a 100644
--- a/pkg/crc/config/validations.go
+++ b/pkg/crc/config/validations.go
@@ -3,6 +3,7 @@ package config
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/network"
@@ -77,10 +78,18 @@ func ValidatePath(value interface{}) (bool, string) {
 	return true, ""
 }
 
-// ValidateURI check if given URI is valid
+// ValidateURI checks if given URI is valid
 func ValidateURI(value interface{}) (bool, string) {
 	if err := network.ValidateProxyURL(value.(string)); err != nil {
 		return false, err.Error()
 	}
 	return true, ""
 }
+
+// ValidateNoProxy checks if the NoProxy string has the correct format
+func ValidateNoProxy(value interface{}) (bool, string) {
+	if strings.Contains(value.(string), " ") {
+		return false, "NoProxy string can't contain spaces"
+	}
+	return true, ""
+}
```

pkg/crc/network/proxy.go
```diff
@@ -81,8 +81,8 @@ func (p *ProxyConfig) HttpsProxyForDisplay() string {
 }
 
 // AddNoProxy appends the specified host to the list of no proxied hosts.
-func (p *ProxyConfig) AddNoProxy(hosts string) {
-	hosts = strings.Replace(hosts, " ", "", -1)
+func (p *ProxyConfig) AddNoProxy(host ...string) {
+	hosts := strings.Join(host, ",")
 	p.NoProxy = fmt.Sprintf("%s,%s", hosts, p.NoProxy)
 }
```
cfergeau: That's a bit odd as this method was changed in the previous commit, and is somehow changed back here.
Might also work as
```
host = append(host, p.NoProxy)
p.NoProxy = strings.Join(host, ",")
```

pkg/crc/cluster/cluster.go
```diff
@@ -138,6 +139,41 @@ func StopAndRemovePodsInVM(sshRunner *ssh.SSHRunner) error {
...
+		return err
+	}
+	if _, stderr, err := oc.RunOcCommand(cmdArgs...); err != nil {
+		return fmt.Errorf("Failed to add proxy details %v: %s", err, stderr)
+	}
+	return nil
+}
+
+// AddProxyToKubeletAndCriO add the systemd drop-in proxy configuration file to the instance,
+// both the services (kubelet and crio) need to be restarted after this change.
```
cfergeau: "adds the systemd"
"both services"

pkg/crc/cluster/cluster.go
```diff
@@ -138,6 +139,41 @@ func StopAndRemovePodsInVM(sshRunner *ssh.SSHRunner) error {
...
+func AddProxyToKubeletAndCriO(sshRunner *ssh.SSHRunner, proxy *network.ProxyConfig) error {
+	proxyTemplate := `[Service]
+Environment=HTTP_PROXY=%s
+Environment=HTTPS_PROXY=%s
+Environment=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,%s`
+
```
cfergeau: Quoting your earlier message on that topic
> @cfergeau there is no doc from openshift side about it, this is taken care by the proxy operator and added automatically but since this status doesn't visible just after making the change in the proxy and need bit time so I used these value as constant because those are the one needed for kubelet and crio both

If these values are added automatically, I'd prefer to let the proxy operator deal with them instead of hardcoding them here, and then need to change them if these values are ever changed to something different on the OpenShift side.

praveenkumar: >  I'd prefer to let the proxy operator deal with them instead of hardcoding them here, and then need to change them if these values are ever changed to something different on the OpenShift side.

@cfergeau problem is proxy operator is not able to make changes to in the kubelet/crio side, this is job of machine config operator on the node and since for crc this is not possible so we do need to put it here.

pkg/crc/network/proxy.go
praveenkumar: @cfergeau this will only work if we take `host` as slice instead of string.

pkg/crc/cluster/cluster.go
```diff
@@ -182,13 +182,14 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
...
-func waitForOpenshiftAPIServer(oc oc.OcConfig) error {
+func waitForOpenshiftAPIServer(oc oc.OcConfig, resource string) error {
 	waitForApiServer := func() error {
-		_, stderr, err := oc.RunOcCommand("get", "pods")
+		stdout, stderr, err := oc.RunOcCommand("get", resource)
 		if err != nil {
 			logging.Debug(stderr)
 			return &errors.RetriableError{Err: err}
 		}
+		logging.Debug(stdout)
```
cfergeau: dunno if that added debug is intentional?

pkg/crc/cluster/cluster.go
cfergeau: This kind of context belongs in the commit log, very useful info to have to understand why this was implemented this way.

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau yes this is intentional, would help us if some resource not able to get after a given retry.

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau I will add it to commit log and also as part of comment on top of this function.

Issue #656, opened 2019-09-27
Remove hypervisor string from the validation message.


praveenkumar:
As of now if a user provide the older version of the bundle he/she gets following error, since now our bundle follow all the hypervisor naming we should directly tell them what to use.

```
$ ./crc start -b ~/Downloads/crc_libvirt_4.1.11.crcbundle 
ERRO Error occurred: crc_libvirt_4.1.11.crcbundle bundle is not supported by this binary, you must use crc_<hypervisor>_4.1.14.crcbundle 
```

cfergeau:
On macos and windows, both native and vbox hypervisors can be used. I agree we could make this message better as most of the time we'll have a bundle embedded in the binary and can recommend using this somehow.

praveenkumar:
@cfergeau right, not high priority but in the future we can look it and see how it can be done.

cfergeau:
Now that vbox is gone, much easier to do ;)

PR #885, opened 2019-12-12
Issue 656: validation: Replace "<hypervisor>" with actual hypervisor name

cfergeau:
When the bundle version is not correct, we currently print a message
referring to "crc_<hypervisor>_4.2.8.crcbundle". However, now that
virtualbox support has been dropped, we can directly mention the native
hypervisor in the message.

This fixes https://github.com/code-ready/crc/issues/656

### DIFF
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -53,7 +53,7 @@ func ValidateBundle(bundle string) error {
 	releaseBundleVersion := version.GetBundleVersion()
 	userProvidedBundleVersion := filepath.Base(bundle)
 	if !strings.Contains(userProvidedBundleVersion, fmt.Sprintf("%s.crcbundle", releaseBundleVersion)) {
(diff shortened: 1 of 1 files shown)


pkg/crc/validation/validation.go
```diff
@@ -53,7 +53,7 @@ func ValidateBundle(bundle string) error {
 	releaseBundleVersion := version.GetBundleVersion()
 	userProvidedBundleVersion := filepath.Base(bundle)
 	if !strings.Contains(userProvidedBundleVersion, fmt.Sprintf("%s.crcbundle", releaseBundleVersion)) {
-		return errors.Newf("%s bundle is not supported by this binary, please use crc_<hypervisor>_%s.crcbundle", userProvidedBundleVersion, releaseBundleVersion)
+		return errors.Newf("%s bundle is not supported by this binary, please use %s", userProvidedBundleVersion, constants.GetDefaultBundle())
```
praveenkumar: :+1:

PR #886, opened 2019-12-12
Embed the binaries we need in 'crc' to support offline installs

### LINKED ISSUE #509: Support 'offline' installs
At the moment, at `crc setup` time, we download `oc`, and the crc machine driver that we need. This will fail if run on a machine with limited network connectivity, so it would be better if all the needed binaries were included in the tarball or embedded.

cfergeau:
At the moment, crc needs external network access in order to download
and install oc and its machine drivers. This is not going to be possible
in some environment where external access is disabled.
This patch series adds a new helper used at build time which will download
the misc required binaries and embed them in the 'crc' binary. Then these
binaries will be used instead of downloading them.
This fixes #509

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,12 @@ DOCS_BUILD_TARGET ?= /docs/source/getting-started/master.adoc
 
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
+HOST_BUILD_DIR=$(BUILD_DIR)/$(GOOS)-$(GOARCH)
 GOPATH ?= $(shell go env GOPATH)
 ORG := github.com/code-ready
 REPOPATH ?= $(ORG)/crc
 
-PACKAGES := go list ./...
+PACKAGES := go list --tags build ./...
 SOURCES := $(shell git ls-files  *.go ":^vendor")
 
 RELEASE_INFO := release-info.json
@@ -72,10 +73,6 @@ default: install
 vendor:
 	GO111MODULE=on go mod vendor
 
-# Get binappend
-$(GOPATH)/bin/binappend-cli:
-	GO111MODULE=off go get -u github.com/yourfin/binappend-cli
-
 # Start of the actual build targets
 
 .PHONY: install
@@ -91,12 +88,15 @@ $(BUILD_DIR)/linux-amd64/crc: $(SOURCES)
 $(BUILD_DIR)/windows-amd64/crc.exe: $(SOURCES)
 	GOARCH=amd64 GOOS=windows go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/windows-amd64/crc.exe ./cmd/crc
 
+$(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
+	go build --tags="build" -ldflags="$(LDFLAGS)" -o $(HOST_BUILD_DIR)/crc-embedder ./cmd/crc-embedder
+
 .PHONY: cross ## Cross compiles all binaries
 cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/windows-amd64/crc.exe
 
 .PHONY: test
 test:
-	go test -v -ldflags="$(VERSION_VARIABLES)" $(shell $(PACKAGES))
+	go test --tags build -v -ldflags="$(VERSION_VARIABLES)" $(shell $(PACKAGES))
 
 .PHONY: build_docs
 build_docs:
@@ -148,21 +148,22 @@ $(GOPATH)/bin/golangci-lint:
 	GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.17.1
 
 # Run golangci-lint against code
+GOLANGCI_FLAGS=--build-tags build
 .PHONY: lint cross-lint
 lint: $(GOPATH)/bin/golangci-lint
-	$(GOPATH)/bin/golangci-lint run
+	$(GOPATH)/bin/golangci-lint $(GOLANGCI_FLAGS) run
 
 cross-lint: $(GOPATH)/bin/golangci-lint
-	GOOS=darwin $(GOPATH)/bin/golangci-lint run
-	GOOS=linux $(GOPATH)/bin/golangci-lint run
-	GOOS=windows $(GOPATH)/bin/golangci-lint run
+	GOOS=darwin $(GOPATH)/bin/golangci-lint $(GOLANGCI_FLAGS) run
+	GOOS=linux $(GOPATH)/bin/golangci-lint $(GOLANGCI_FLAGS) run
+	GOOS=windows $(GOPATH)/bin/golangci-lint $(GOLANGCI_FLAGS) run
 
 $(GOPATH)/bin/gosec:
 	GO111MODULE=on go get github.com/securego/gosec/cmd/gosec@2.0.0
 
 # Run gosec against code
 gosec: $(GOPATH)/bin/gosec
-	$(GOPATH)/bin/gosec -tests -severity medium  ./...
+	$(GOPATH)/bin/gosec -tags build -tests -severity medium  ./...
 
 .PHONY: gen_release_info
 gen_release_info:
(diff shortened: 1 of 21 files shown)


pkg/crc/constants/constants.go
```diff
@@ -25,8 +26,27 @@ const (
 	GlobalStateFile      = "globalstate.json"
 	CrcLandingPageURL    = "https://cloud.redhat.com/openshift/install/crc/installer-provisioned" // #nosec G101
 	PullSecretFile       = "pullsecret.json"
+
+	DefaultOcURLDarwin  = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/macosx/oc.tar.gz"
+	DefaultOcURLLinux   = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz"
+	DefaultOcURLWindows = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/windows/oc.zip"
```
praveenkumar: Why we need those `DefaultOcURL<OS>` variables when we are getting it from `GetOcUrl` side?

cfergeau: I forgot to remove these :) Thanks!

praveenkumar:
retest this please.

pkg/crc/machine/hyperkit/constants.go
```diff
@@ -1,3 +1,5 @@
+//+build darwin build
```
cfergeau: It seems using a `crc` build tag when building the `crc` binary (ie making this line `//+build darwin !crc`) would mean less changes to `Makefile`, I can post this alternate version.

Makefile
```diff
@@ -199,7 +199,7 @@ check_bundledir:
 	@$(call check_defined, BUNDLE_DIR, "Embedding bundle requires BUNDLE_DIR set to a directory containing CRC bundles for all hypervisors")
 
 embed_bundle: LDFLAGS += $(BUNDLE_EMBEDDED)
-embed_bundle: clean cross $(GOPATH)/bin/binappend-cli check_bundledir $(HYPERKIT_BUNDLENAME) $(HYPERV_BUNDLENAME) $(LIBVIRT_BUNDLENAME)
-	$(GOPATH)/bin/binappend-cli write $(BUILD_DIR)/linux-amd64/crc $(LIBVIRT_BUNDLENAME)
-	$(GOPATH)/bin/binappend-cli write $(BUILD_DIR)/macos-amd64/crc $(HYPERKIT_BUNDLENAME)
-	$(GOPATH)/bin/binappend-cli write $(BUILD_DIR)/windows-amd64/crc.exe $(HYPERV_BUNDLENAME)
+embed_bundle: clean cross build-aux/crc-embedder check_bundledir $(HYPERKIT_BUNDLENAME) $(HYPERV_BUNDLENAME) $(LIBVIRT_BUNDLENAME)
+	build-aux/crc-embedder embed --goos=darwin --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/macos-amd64/crc
```
gbraad: wondering why this is `build-aux`. is this intended as a place for more build related tooling?

cfergeau: If we need more build tools in the future, yes they could go there. But this is just a name I picked from my autotools day, I don't mind changing it to something different.

gbraad: just wondering... not a problem. it almost suggest you have more tools on mind,... if so, what are those?

cfergeau: nope, definitely no additional tool in mind :) Not even sure it would be convenient to build 2 binaries from the same dir with go.

gbraad: actually, the patter for that is `cmd/build-aux`, just like `cmd/crc`. the second `cmd` in `cmd/crc/cmd` is actually for cobra/viper.

praveenkumar:
@cfergeau this now need a rebase.

praveenkumar:
retest this please,

build-aux/cmd/list.go
```diff
@@ -0,0 +1,38 @@
...
+		runList(args)
+	},
+}
+
+func runList(args []string) {
+	if len(args) != 1 {
+		logging.Fatalf("list takes exactly one argument")
+	}
+	binaryPath := args[0]
+	output.Outln("List")
```
praveenkumar: This is not required otherwise it will show as below.
```
List
Data files embedded in out/macos-amd64/crc:
	 crc-driver-hyperkit
	 crc_hyperkit_4.2.12.crcbundle
	 hyperkit
	 oc.tar.gz
```

Makefile
```diff
@@ -91,12 +87,15 @@ $(BUILD_DIR)/linux-amd64/crc: $(SOURCES)
 $(BUILD_DIR)/windows-amd64/crc.exe: $(SOURCES)
 	GOARCH=amd64 GOOS=windows go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/windows-amd64/crc.exe ./cmd/crc
 
+build-aux/crc-embedder: $(SOURCES) build-aux/crc-embedder.go
```
praveenkumar: why we require `build-aux/crc-embedder.go` for this target? I tested it on mac without adding it and it does work without any issue.

Makefile
```diff
@@ -199,7 +199,7 @@ check_bundledir:
...
 
 embed_bundle: LDFLAGS += $(BUNDLE_EMBEDDED)
-embed_bundle: clean cross $(GOPATH)/bin/binappend-cli check_bundledir $(HYPERKIT_BUNDLENAME) $(HYPERV_BUNDLENAME) $(LIBVIRT_BUNDLENAME)
-	$(GOPATH)/bin/binappend-cli write $(BUILD_DIR)/linux-amd64/crc $(LIBVIRT_BUNDLENAME)
-	$(GOPATH)/bin/binappend-cli write $(BUILD_DIR)/macos-amd64/crc $(HYPERKIT_BUNDLENAME)
-	$(GOPATH)/bin/binappend-cli write $(BUILD_DIR)/windows-amd64/crc.exe $(HYPERV_BUNDLENAME)
+embed_bundle: clean cross build-aux/crc-embedder check_bundledir $(HYPERKIT_BUNDLENAME) $(HYPERV_BUNDLENAME) $(LIBVIRT_BUNDLENAME)
+	build-aux/crc-embedder embed --goos=darwin --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/macos-amd64/crc
+	build-aux/crc-embedder embed --goos=linux --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/linux-amd64/crc
+	build-aux/crc-embedder embed --goos=windows --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/windows-amd64/crc.exe
```
praveenkumar: We should use `--log-level debug` here since this is the step which is going to run on the CI so better to have debug logs if possible.

Makefile
```diff
@@ -91,12 +87,15 @@ $(BUILD_DIR)/linux-amd64/crc: $(SOURCES)
 $(BUILD_DIR)/windows-amd64/crc.exe: $(SOURCES)
 	GOARCH=amd64 GOOS=windows go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/windows-amd64/crc.exe ./cmd/crc
 
+build-aux/crc-embedder: $(SOURCES) build-aux/crc-embedder.go
+	go build --tags="build" -ldflags="$(LDFLAGS)" -o build-aux/crc-embedder ./build-aux
```
praveenkumar: Shouldn't we use `BUILD_DIR` for all the binary generation part? Following patch have all the diff which I think should be considered.

```patch
-build-aux/crc-embedder: $(SOURCES) build-aux/crc-embedder.go
-       go build --tags="build" -ldflags="$(LDFLAGS)" -o build-aux/crc-embedder ./build-aux
+build-aux/crc-embedder: $(SOURCES)
+       go build --tags="build" -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/crc-embedder ./build-aux
 
 .PHONY: cross ## Cross compiles all binaries
 cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/windows-amd64/crc.exe
@@ -200,6 +200,6 @@ check_bundledir:
 
 embed_bundle: LDFLAGS += $(BUNDLE_EMBEDDED)
 embed_bundle: clean cross build-aux/crc-embedder check_bundledir $(HYPERKIT_BUNDLENAME) $(HYPERV_BUNDLENAME) $(LIBVIRT_BUNDLENAME)
-       build-aux/crc-embedder embed --goos=darwin --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/macos-amd64/crc
-       build-aux/crc-embedder embed --goos=linux --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/linux-amd64/crc
-       build-aux/crc-embedder embed --goos=windows --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/windows-amd64/crc.exe
+       $(BUILD_DIR)/crc-embedder embed --goos=darwin --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/macos-amd64/crc --log-level debug
+       $(BUILD_DIR)/crc-embedder embed --goos=linux --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/linux-amd64/crc --log-level debug
+       $(BUILD_DIR)/crc-embedder embed --goos=windows --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/windows-amd64/crc.exe --log-level debug
```

Makefile
cfergeau: Ah right it should be part of $(SOURCES). I'll try to drop it.

Makefile
cfergeau: dunno what are the conventions for go + Makefile :) if you prefer things this way, no objections from me, I'll change.

Makefile
```diff
@@ -31,11 +31,12 @@ DOCS_BUILD_TARGET ?= /docs/source/getting-started/master.adoc
 
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
+HOST_BUILD_DIR=$(BUILD_DIR)/$(GOOS)-$(GOARCH)
```
praveenkumar: @cfergeau Any specific reason not directly using the `$(BUILD_DIR)` one instead creating a new directory?

cfergeau: The `out` directory contains `out/macos-amd64/crc`, `out/linux-amd64/crc` and `out/win-amd64/crc.exe`, it felt odd to add a `out/crc-embedder` binary rather than putting it in the correct `out/$os-$arch/` directory.

praveenkumar: Ack 👍

cfergeau: actually, I suspect on macos, `$(GOOS)-$(GOARCH)` is going to be `darwin-amd64`, not `macos-amd64`, so an additional commit may be needed.

cfergeau: Confirmed, but this does not cause a build failure, it just created `out/darwin-amd64/crc-embedder` in addition to `out/macos-amd64/crc`. I'll keep it that way for now.

PR #902, opened 2019-12-19
Add proxy variables to oc-env

cfergeau:
When a proxy is configured in crc, we should add http_proxy/https_proxy/no_proxy to the variables to set in `oc-env`
This pull request achieves that, though it starts by misc improvements to network.Proxy which are not strictly required for that change.

### DIFF
--- a/cmd/crc/cmd/oc_env.go
+++ b/cmd/crc/cmd/oc_env.go
@@ -16,8 +16,6 @@ var ocEnvCmd = &cobra.Command{
 	Use:   "oc-env",
 	Short: "Add the 'oc' binary to PATH",
 	Long:  `Add the OpenShift client binary 'oc' to PATH`,
-	// This is required to make sure root command Persistent PreRun not run.
-	PersistentPreRun: func(cmd *cobra.Command, args []string) {},
 	Run: func(cmd *cobra.Command, args []string) {
 		userShell, err := shell.GetShell(forceShell)
 		if err != nil {
--- a/cmd/crc/cmd/oc_env.go
+++ b/cmd/crc/cmd/oc_env.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/output"
 	"github.com/code-ready/crc/pkg/os/shell"
 	"github.com/spf13/cobra"
@@ -22,7 +23,16 @@ var ocEnvCmd = &cobra.Command{
 			errors.ExitWithMessage(1, "Error running the oc-env command: %s", err.Error())
 		}
 
+		proxyConfig, err := machine.GetProxyConfig(constants.DefaultName)
+		if err != nil {
+			errors.Exit(1)
+		}
 		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
+		if proxyConfig.IsEnabled() {
+			output.Outln(shell.GetEnvString(userShell, "HTTP_PROXY", proxyConfig.HttpProxy))
+			output.Outln(shell.GetEnvString(userShell, "HTTPS_PROXY", proxyConfig.HttpsProxy))
+			output.Outln(shell.GetEnvString(userShell, "NO_PROXY", proxyConfig.GetNoProxyString()))
+		}
 		output.Outln(shell.GenerateUsageHint(userShell, "crc oc-env"))
 	},
 }
--- a/cmd/crc/cmd/oc_env.go
+++ b/cmd/crc/cmd/oc_env.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/output"
 	"github.com/code-ready/crc/pkg/os/shell"
 	"github.com/spf13/cobra"
 )
@@ -19,6 +20,7 @@ var (
 )
 
 type OcShellConfig struct {
+	UserShell string
 	shell.ShellConfig
 	OcDirPath string
 	UsageHint string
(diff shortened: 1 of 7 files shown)


cmd/crc/cmd/oc_env.go
```diff
@@ -22,7 +23,16 @@ var ocEnvCmd = &cobra.Command{
 			errors.ExitWithMessage(1, "Error running the oc-env command: %s", err.Error())
 		}
 
+		clusterConfig, err := machine.GetClusterConfig(constants.DefaultName)
```
praveenkumar: @cfergeau I am trying to understand why we need to create this function only to get proxy info? shouldn't we abstracted the proxy to use it for any method/function? Since now the root.cmd `prerun` happen so we know if the proxy is set of not and can be get using `network.NewProxyConfig()` . Are you trying to only show those proxy detail if the VM is running ?

pkg/crc/machine/machine.go
```diff
@@ -703,9 +703,21 @@ func configureCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, baseDom
...
+		proxy.AddNoProxy(fmt.Sprintf(".%s", baseDomainName))
+	}
+
+	return proxy, nil
+}
+
 func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd *systemd.InstanceSystemdCommander,
 	baseDomainName, instanceIP string) (err error) {
-	proxy, err := network.NewProxyConfig()
+	proxy, err := getProxyConfig(baseDomainName)
```
praveenkumar: @cfergeau I don't think this helper is required, what we are doing with this helper is just adding the `base domain` to noproxy what is wrong it to add along with instanceIP?

cfergeau: On the host, we don't need `instanceIP` in `NO_PROXY` as nothing should try to access the cluster directly through its IP, but always through `api.crc.testing`. And the logic of calling `network.NewProxyConfig()` and then adding the base domain name to `NO_PROXY` will be needed twice by the end of the patch series, so I moved this to a helper.

cmd/crc/cmd/oc_env.go
cfergeau: In order to be able to add `BaseDomain` to `NO_PROXY`, we need to read it from a `crc-bundle-info.json` file. At the moment, this means going through `~/.crc/machines/crc/config.json` to get the `BundleName`, and then looking up the correct bundle from `~/.crc/cache`. This is why this is not using `network.NewProxyConfig` directly.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

cmd/crc/cmd/oc_env.go
praveenkumar: @cfergeau Got it, but I still think that we should get the basedomain instead of the whole cluster config.

```golang
	proxy, err := network.NewProxyConfig()
	if err != nil {
		return nil, err
	}
	if proxy.IsEnabled() {
               baseDomain := machine.GetbaseDomain()
               proxy.AddNoProxy(fmt.Sprintf(".%s", baseDomainName))

	       output.Outln(shell.GetEnvString(userShell, "HTTP_PROXY", clusterConfig.ProxyConfig.HttpProxy))
	       output.Outln(shell.GetEnvString(userShell, "HTTPS_PROXY", clusterConfig.ProxyConfig.HttpsProxy))
	       output.Outln(shell.GetEnvString(userShell, "NO_PROXY", clusterConfig.ProxyConfig.GetNoProxyString()))
	}
```

cfergeau: Ah, `GetClusterConfig` dates from when I hoped to share this code with `GetConsoleURL`, this can indeed be reworked a bit

gbraad: will this rework be done now or postponed?

pkg/crc/machine/machine.go
gbraad: I think this is needed, since when using the hostname, which gets translated to the IP address, you will perform a get request. which can be forced to be resolved over the local machines proxy settings.

You are adding it anyway: https://github.com/code-ready/crc/pull/902/files#diff-06d0e75b8f65eb58c0f83515f2a7217dR771 Right?

cmd/crc/cmd/oc_env.go
cfergeau: I changed machine.GetClusterConfig() to machine.GetProxyConfig(). I don't want to duplicate in cmd/ the logic to create the proxy config (in particular the `proxy.AddNoProxy(fmt.Sprintf(".%s", baseDomainName))` part)

pkg/crc/machine/machine.go
cfergeau: It's added for the initial machine creation/proxy configuration. It's not added as part of `crc oc-env`

cmd/crc/cmd/oc_env.go
```diff
@@ -1,65 +1,39 @@
...
+		if err != nil {
+			errors.ExitWithMessage(1, "Error running the oc-env command: %s", err.Error())
+		}
+
+		proxyConfig, err := machine.GetProxyConfig(constants.DefaultName)
 		if err != nil {
 			errors.Exit(1)
 		}
-		_ = executeOcTemplateStdout(shellCfg)
+		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
```
praveenkumar: @cfergeau once this is in then user is not able to execute `oc-env` if the VM is not present, which is not the case right now. This look like the change from user side, cc @gbraad ? should we just ignore the error if the machine not present?

```
$ ./crc oc-env
export PATH="/home/prkumar/.crc/bin:$PATH"
# eval $(crc oc-env)

--- After this PR ---
$ ./crc oc-env
ERRO Machine "crc" does not exist. Use "crc start" to add a new one.
```

gbraad: In this case we should use an `echo` to force the printout.

pkg/crc/machine/machine.go
```diff
@@ -614,6 +620,33 @@ func addNameServerToInstance(sshRunner *crcssh.SSHRunner, ns string) error {
 	return nil
 }
 
+// Return proxy config if VM is present
+func GetProxyConfig(machineName string) (*network.ProxyConfig, error) {
```

Issue #872, opened 2019-12-09
Refactor preflight checks


cfergeau:
This change simplifies the use of the prflight checks

PR #907, opened 2019-12-20
Issue #872: Remove no longer used config constants

cfergeau:
Since e552b17571, the skip/warn config variables are automatically generated
so the corresponding config_*.go files declaring these variables are no
longer needed.


praveenkumar:
LGTM

PR #922, opened 2020-01-07
Issue #854 Add host->cluster DNS checks

cfergeau:
This check will make sure that host can able to resolve the dns
for api and app domain of crc using dnsmasq which is running
inside the VM.

This patch is heavily based on a patch from Praveen Kumar
<kumarpraveen.nitdgp@gmail.com>, with the addition of the code excluding macos
from the apps domain check.

This fixes https://github.com/code-ready/crc/issues/854

### DIFF
--- a/pkg/crc/network/utils.go
+++ b/pkg/crc/network/utils.go
@@ -2,8 +2,10 @@ package network
 
 import (
 	"fmt"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	"net"
 	"net/url"
+	"runtime"
 
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/ssh"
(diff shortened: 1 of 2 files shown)


pkg/crc/network/utils.go
```diff
@@ -92,3 +94,24 @@ func UriStringForDisplay(uri string) (string, error) {
 	}
 	return uri, nil
 }
+
+func CheckCRCLocalDNSReachableFromHost(clusterName, domainName, appsDomainName string) error {
+	ip, err := net.LookupIP(fmt.Sprintf("api.%s.%s", clusterName, domainName))
```
praveenkumar: @cfergeau Did we test this on OSx since api entry is only added to hosts file so not sure if `net.LookupIP` works for it? we need to test the osx build which is cross compiled on linux, build which is done on osx will not going error out for same.

cfergeau: Yes, I tested it on macos yesterday, and did not have issues. However, this was a binary built on the macos machine, not a cross-build, now that you raised this question, this deserves additional testing (but I don't have access to the mac today).

gbraad: it should, but please verify a cross-build. do macOS based builds use CGO_ENABLED?

praveenkumar: > do macOS based builds use CGO_ENABLED?

No, that way we are not able to make cross builds :(

I tested the cross bit from centos artifact and it is working.

pkg/crc/network/utils.go
```diff
@@ -92,3 +94,24 @@ func UriStringForDisplay(uri string) (string, error) {
...
+	if err != nil {
+		return err
+	}
+	logging.Debugf("api.%s.%s resolved to %s", clusterName, domainName, ip)
+
+	if runtime.GOOS != "darwin" {
```
praveenkumar: 👍

pkg/crc/network/utils.go
praveenkumar: Should we also verify it on windows before merging it?

pkg/crc/network/utils.go
gbraad: I know this is needed, but less charmed about it...

cfergeau: We could go add some `utils_darwin.go` , `utils_linux.go`... to deal with the os-specific bits, but this seemed overkill.

pkg/crc/network/utils.go
cfergeau: Would not hurt if someone has a Windows install handy for testing.

pkg/crc/network/utils.go
gbraad: I was considering the same... and agree about the overkill. just concerned this gets forgotten.

pkg/crc/network/utils.go
gbraad: will perfrom a test, but currently have limited access to resources.

praveenkumar: I did tested this on windows :+1:

PR #931, opened 2020-01-13
Embed macOS tray in crc binary and install it on crc setup

### LINKED ISSUE #896: Add/Install the tray as part of crc setup on macOS
We have to integrate the tray with crc. For this on macOS we can create and add _launchd_ scripts to auto start the daemon and tray app as part of `crc setup`.

https://medium.com/swlh/how-to-use-launchd-to-run-services-in-macos-b972ed1e352

anjannath:
Fixes #896 
Depends on #861

### DIFF
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -30,7 +30,8 @@ const (
 	CrcLandingPageURL    = "https://cloud.redhat.com/openshift/install/crc/installer-provisioned" // #nosec G101
 	PullSecretFile       = "pullsecret.json"
 
-	DefaultOcUrlBase = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest"
+	DefaultOcUrlBase   = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest"
+	CrcTrayDownloadURL = "https://github.com/code-ready/tray-macos/releases/download/v%s/crc-tray-macos.tar.gz"
 )
 
 var ocUrlForOs = map[string]string{
@@ -115,4 +116,8 @@ func GetPublicKeyPath() string {
 
 func GetPrivateKeyPath() string {
 	return filepath.Join(MachineInstanceDir, DefaultName, "id_rsa")
+}
+
+func GetCrcTrayDownloadURL() string {
+	return fmt.Sprintf(CrcTrayDownloadURL, version.GetCRCTrayVersion())
 }
--- a/pkg/crc/preflight/preflight_check_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_check_tray_darwin.go
@@ -1,0 +1,235 @@
+package preflight
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	goos "os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	goTemplate "text/template"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	dl "github.com/code-ready/crc/pkg/download"
+	"github.com/code-ready/crc/pkg/embed"
+	"github.com/code-ready/crc/pkg/extract"
+	"github.com/code-ready/crc/pkg/os"
+	"github.com/pkg/errors"
+)
+
+const (
+	trayPlistTemplate = `<?xml version='1.0' encoding='UTF-8'?>
+	<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+	<plist version='1.0'>
+		<dict>
+			<key>Label</key>
+			<string>crc.tray</string>
+			<key>ProgramArguments</key>
+			<array>
+				<string>{{ .BinaryPath }}</string>
+			</array>
+			<key>StandardOutPath</key>
+			<string>{{ .StdOutFilePath }}</string>
+			<key>Disabled</key>
+			<false/>
+			<key>RunAtLoad</key>
+			<true/>
+		</dict>
+	</plist>`
+
+	daemonPlistTemplate = `<?xml version='1.0' encoding='UTF-8'?>
+	<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+	<plist version='1.0'>
+		<dict>
+			<key>Label</key>
+			<string>crc.daemon</string>
+			<key>ProgramArguments</key>
+			<array>
+				<string>{{ .BinaryPath }}</string>
+				<string>daemon</string>
+				<string>--log-level</string>
+				<string>debug</string>
+			</array>
+			<key>StandardOutPath</key>
+			<string>{{ .StdOutFilePath }}</string>
+			<key>KeepAlive</key>
+			<true/>
+			<key>Disabled</key>
+			<false/>
+		</dict>
+	</plist>`
+
+	daemonAgentLabel = "crc.daemon"
+	trayAgentLabel   = "crc.tray"
+)
+
+var (
+	launchAgentDir       = filepath.Join(constants.GetHomeDir(), "Library", "LaunchAgents")
+	daemonPlistFilePath  = filepath.Join(launchAgentDir, "crc.daemon.plist")
+	trayPlistFilePath    = filepath.Join(launchAgentDir, "crc.tray.plist")
+	stdOutFilePathDaemon = filepath.Join(constants.CrcBaseDir, ".crcd-agent.log")
+	stdOutFilePathTray   = filepath.Join(constants.CrcBaseDir, ".crct-agent.log")
+)
+
+type AgentConfig struct {
+	BinaryPath     string
+	StdOutFilePath string
+}
+
+func checkTrayExistsAndRunning() error {
+	logging.Debug("Checking if daemon plist file exists")
+	if !os.FileExists(daemonPlistFilePath) {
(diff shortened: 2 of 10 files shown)


pkg/crc/constants/constants.go
```diff
@@ -29,7 +29,8 @@ const (
 	CrcLandingPageURL    = "https://cloud.redhat.com/openshift/install/crc/installer-provisioned" // #nosec G101
 	PullSecretFile       = "pullsecret.json"
 
-	DefaultOcUrlBase = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest"
+	DefaultOcUrlBase   = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest"
+	CrcTrayDownloadURL = "https://github.com/anjannath/crc-macos-tray/releases/download/%s/crc_tray_1.0.0-alpha.tar.gz"
```
anjannath: This URL will have to changed, once we move the tray repo to code-ready namespace

praveenkumar: @anjannath Since the repo is already migrated can you change this URI now also should this be part of this contants file or be part of macos specific constant file?

anjannath: So it has to be in the constants file, since the `crc-embedder` should be able to find even if we build the embedded binary in a non macOS environment. Its based on teuf's patches for embedding oc and other assets.

pkg/crc/version/version.go
```diff
@@ -35,6 +35,9 @@ var (
 
 	// Bundle version which used for the release.
 	bundleVersion = "0.0.0-unset"
+
+	// Tray version to be embedded in binary
+	crcTrayVersion = "0.0.0-unset"
```
cfergeau: I believe we could directly put 1.0.0-alpha here rather than passing it through linker flags, as we don't need CRC_TRAY_VERSION in the Makefile, only in the go code.

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -0,0 +1,248 @@
...
+		return err
+	}
+	err = ioutil.WriteFile(plistPath, plistContent.Bytes(), 0644)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func loadPlist(plistFilePath string) error {
```
cfergeau: `launchPlist` may be a better function name?

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -0,0 +1,248 @@
...
+		return false
+	}
+	if strings.TrimSpace(string(out)) == "-" {
+		return false
+	}
+	return true
+}
+
+func plistFileExists(plistFilePath string) bool {
+	return os.FileExists(plistFilePath)
```
cfergeau: why the indirection/helper? We could directly use `os.FileExists(plistFilePath)`  throughout this file?

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -0,0 +1,248 @@
...
+	return nil
+}
+
+// check if a service (daemon,tray) is running
+func agentRunning(label string) bool {
+	// This command return a PID if the process
+	// is running, otherwise returns "-"
+	launchctlListCommand := `launchctl list | grep %s | awk '{print $1}'`
+	cmd := fmt.Sprintf(launchctlListCommand, label)
+	out, err := exec.Command("bash", "-c", cmd).Output() // #nosec G204
```
cfergeau: It seems `launchtl list %s` without the `grep` could work?

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -39,13 +39,23 @@ var dnsPreflightChecks = [...]PreflightCheck{
...
 
+var traySetupCheck = PreflightCheck{
+	configKeySuffix:  "check-tray-setup",
+	checkDescription: "Checking if the tray is installed and running",
+	check:            checkTrayExistsAndRunning,
+	fixDescription:   "Installing and setting up tray app",
+	fix:              fixTrayExistsAndRunning,
+	flags:            SetupOnly,
+}
+
```
cfergeau: My feeling is that this preflight check is doing too much at the same time, we usually have smaller tests, one checking it's installed, another checking it's running, ...
This probably can be adjusted later on.

cfergeau:
I guess longer term users won't have to go through `crc setup` to install/start the tray? I'd expect to download somethnig from the web, install it from Finder, and start it through the GUI after install.

pkg/crc/version/version.go
anjannath: Its kept in the makefile, for ease of updating the version of the tray, when a new version of CRC is released we bump the version in the Makefile and along with it we can bump the tray version too.

pkg/crc/preflight/preflight_check_tray_darwin.go
anjannath: the output from `launchctl list crc.daemon` is more verbose, we don't need all that information.

anjannath:
> I guess longer term users won't have to go through crc setup to install/start the tray? I'd expect to download somethnig from the web, install it from Finder, and start it through the GUI after install.

we went with the `crc setup` way so that user doesn't have to do any manual work. and it had to be embedded in the binary itself. what you are suggesting is to create some kind of installer, which is still being discussed. So i don't know at the moment.

pkg/crc/preflight/preflight_check_tray_darwin.go
anjannath: Right, let me fix this!

pkg/crc/version/version.go
cfergeau: For the machine drivers, the version is not in the Makefile, only in the go code, I'd be consistent here, this means less magic constants needed when building crc.

pkg/crc/preflight/preflight_check_tray_darwin.go
cfergeau: If this was `func (launch LaunchCtl) loadPlist(...)`, then we could name it `loadPlist`. In this case, it's just `loadPlist`, so we have no indication that `launchctl` is involved. With just `loadPlist`, I think the function reads the plist file in memory, maybe parsing it. I don't think it's going to pass it as an argument to `launchctl load`.

pkg/crc/preflight/preflight_darwin.go
anjannath: I was thinking about the same, but then this was done to make the setup process be skip-able by configure just one config option.

pkg/crc/version/version.go
anjannath: OK, removed the version constant from Makefile.

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -0,0 +1,244 @@
...
+	// load tray
+	err = launchctlLoadPlist(trayPlistFilePath)
+	if err != nil {
+		return err
+	}
+	if !agentRunning(trayAgentLabel) {
+		if err = startAgent(trayAgentLabel); err != nil {
+			return err
+		}
+	}
```
cfergeau: Fwiw, this whole function is twice the same code duplicated

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -0,0 +1,244 @@
...
+	return nil
+}
+
+func launchctlLoadPlist(plistFilePath string) error {
+	_, err := exec.Command("launchctl", "load", plistFilePath).Output() // #nosec G204
+	if err != nil {
+		return err
+	}
+	return nil
```
cfergeau: This can be just `return err` rather than
```
	if err != nil {
		return err
	}
	return nil
```

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func startAgent(label string) error {
+	_, err := exec.Command("launchctl", "start", label).Output() // #nosec G204
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
cfergeau: Same comment

praveenkumar:
retest this please.

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -0,0 +1,235 @@
...
+	t, err := goTemplate.New("plist").Parse(template)
+	if err != nil {
+		return err
+	}
+	err = t.Execute(&plistContent, config)
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(plistPath, plistContent.Bytes(), 0644)
+	return err
```
cfergeau: Could be `return ioutil.WriteFile(...)`, but fine with me.

pkg/crc/preflight/preflight_darwin.go
cfergeau: with the "experimental feature" thing gerard added, this could now be split if desired.

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -40,7 +40,7 @@ var dnsPreflightChecks = [...]PreflightCheck{
 }
 
 var traySetupCheck = PreflightCheck{
-	configKeySuffix:  "check-tray-setup",
+	configKeySuffix:  "",
```
cfergeau: why remove this?

anjannath: It doesn't have any use now, since we are controlling the check's happening through the flag? so if i do `crc setup --enable-optional-features` and that config is set to `true` the setup will be skipped. and in the case of just `crc setup` that flag is not at all being used.

Issue #929, opened 2020-01-10
[BUG] make release fails if ~/.crc doesn't exist


praveenkumar:
```
$ make release
[...]
out/linux-amd64/crc-embedder embed --log-level debug --goos=darwin --bundle-dir=bundle out/macos-amd64/crc
level=fatal msg="Failed to open logfile: open /home/prkumar/.crc/crc.log: no such file or directory"
make: *** [Makefile:204: embed_bundle] Error 1
```

cfergeau:
It fails to create `~/.crc` if it does not exist yet. If `~/.crc/crc.log` is removed, `make release` still works.

PR #932, opened 2020-01-13
Issue #929: logging: Ensure ~/.crc exists before creating log file

cfergeau:
crc-embedder currently fails if ~/.crc does not exist. This is because
it tries to create a log file before ~/.crc is created. Adding a call to
EnsureBaseDirExists() before trying to create the log file fixes this
issue.

This fixes https://github.com/code-ready/crc/issues/929

### DIFF
--- a/pkg/crc/logging/logging.go
+++ b/pkg/crc/logging/logging.go
@@ -18,6 +18,10 @@ var (
 )
 
 func OpenLogFile() (*os.File, error) {
+	err := constants.EnsureBaseDirExists()
+	if err != nil {
+		return nil, err
+	}
 	l, err := os.OpenFile(filepath.Join(constants.LogFilePath), os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0600)
 	if err != nil {
 		return nil, err


praveenkumar:
retest this please.

pkg/crc/logging/logging.go
```diff
@@ -18,6 +18,10 @@ var (
 )
 
 func OpenLogFile() (*os.File, error) {
+	err := constants.EnsureBaseDirExists()
+	if err != nil {
+		return nil, err
+	}
```
praveenkumar: @cfergeau This should be part of `crc-embedder/root.go` script instead here, same way it is part of `crc/root.go` because it is not only creating the log directory but initial structure so should be part of init method of root.

cfergeau: With Anjan's refactor of this code, we can move the responsibility of creating the directories to the caller.
With the current code, where it's `pkg/crc/logging/logging.go` which decides where to put its log file, to me it makes sense that it also makes sure the directory exists.

praveenkumar: @cfergeau you mean https://github.com/code-ready/crc/pull/861/commits/1f9070d6146b63f2a443c03c6433146022bd8012 this refactor? This is just adding a path argument, or it is a different one?

cfergeau: Yes, this one. Currently, where the log is stored is an implementation detail of the 'logging' package, so in my opinion it makes sense that the `logging` package takes care of creating the path.
After Anjan's refactor, where the log is stored is specified by users of the package iirc. The location of the log file is no longer internal to the `logging` package, so it makes sense that users of `logging` are required to make sure the path exists.

PR #933, opened 2020-01-13
Fix permissions of embedded binaries

cfergeau:


pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -130,7 +124,7 @@ func extractOrDownloadBinary(url string) error {
 		return err
 	}
 	logging.Debugf("Installing %s", binaryName)
-	binaryPath, err := extractBinary(binaryName)
+	binaryPath, err := extractBinary(binaryName, 0755)
```
praveenkumar: :+1:

PR #940, opened 2020-01-20
Update dnsmasq

praveenkumar:


images/dnsmasq/Dockerfile
```diff
@@ -1,13 +1,12 @@
-FROM centos:7
+FROM centos:8
```
gbraad: We didn't ?

hmmm... but we still need to be ask about this package, so we can use a UBI image instead. /cc: @fatherlinux

images/dnsmasq/Dockerfile
```diff
@@ -2,12 +2,11 @@ FROM centos:8
...
-    systemctl enable dnsmasq
+    yum clean all
 
 COPY dnsmasq.conf /etc/
 COPY resolv.dnsmasq.conf /etc/
 
 EXPOSE 53/udp
 
-CMD [ "/sbin/init" ]
+ENTRYPOINT ["/usr/sbin/dnsmasq", "-d"]
```
cfergeau: From `man dnsmasq`:
```
       -d, --no-daemon
              Debug  mode:  don't fork to the background, don't write a pid file, don't change user id, generate a complete cache dump on receipt on SIGUSR1, log to stderr as well as syslog, don't fork new
              processes to handle TCP queries. Note that this option is for use in debugging only, to stop dnsmasq daemonising in production, use -k.
```
I'd use the long option name fwiw, so `--keep-in-foreground`

images/dnsmasq/Dockerfile
cfergeau: iirc most container images are using rhel7 ubi, it would be more consistent to keep using centos:7 even if there will be no sharing. But both `centos:7` and `centos:8` are fine with me.

images/dnsmasq/Dockerfile
praveenkumar: @cfergeau I thought `-d` make more sense in our case since we are able to collect more logs if it crashed ?

praveenkumar: But `don't fork new processes to handle TCP queries` might be good point not to use it since there is a lot of request which is going through local dns server, I will update it

cfergeau: They say not no use `-d` in production, I'm not going to try to second guess dnsmasq's authors :)

praveenkumar:
retest this please.

PR #942, opened 2020-01-21
Issue #337 Document running CRC behind a proxy

robin-owen:


docs/source/getting-started/content/proc_starting-codeready-containers-behind-proxy.adoc
```diff
@@ -0,0 +1,19 @@
...
+$ {bin} config set https-proxy http://example.proxy.com:__<port>__
+$ {bin} config set no-proxy __<comma-separated-no-proxy-entries>__
+----
+
+The [command]`{bin}` binary will be able to use the defined proxy once set via environment variables or {prod} configuration.
+
+[NOTE]
+====
+* Proxy-related values set in the configuration for {prod} have priority over values set via environment variables.
+* Ensure that you export `.testing` as part of the `no_proxy` or `NO_PROXY` environment variable before logging in to the cluster using the [command]`oc` command on the host.
```
robin-owen: Should this actually be a prerequisite?

@praveenkumar, @teuf, what do you think?

robin-owen:
According to @gbraad, we should mention that `SOCKS` proxies are not supported, as these are not supported by OpenShift itself.

docs/source/getting-started/content/proc_starting-codeready-containers-behind-proxy.adoc
cfergeau: If you use `eval $(crc oc-env)` before using `oc`, you don't need any special step, so I'm not sure we need to mention this here.

praveenkumar: So if a user already have `oc` binary installed he will tend to just copy paste what is the output of `crc start` instead using `eval $(crc oc-env)` and in this case we really need to let the user know otherwise he will get something following.
```
$ ./crc status --log-level debug
DEBU HTTP-PROXY: http://proxy-user1:xxx@xxx:3128, HTTPS-PROXY: http://proxy-user1:xxx@ec2-xxx:3128, NO-PROXY: [127.0.0.1 localhost] 
ERRO Unable to connect to the server: Service Unavailable
 - exit status 1 
```

cfergeau: The output of `crc start` starts with `To access the cluster, first set up your environment by following 'crc oc-env' instructions`

cfergeau: We can mention somewhere `oc` needs in `NO_PROXY`, but we should not mislead users in thinking it's mandatory to do it manually, nor to think `crc oc-env` is optional.

praveenkumar: > We can mention somewhere oc needs in NO_PROXY, but we should not mislead users in thinking it's mandatory to do it manually, nor to think crc oc-env is optional.

@cfergeau right, @robin-owen can you put some wording around it?

robin-owen: @praveenkumar @cfergeau Sure, we can add that information. For now, I added that information as a prerequisite without explicitly stating what it was for -- I can add that in. On that topic, though, I'd also like to explicitly mention that the "recommended way" to use `oc` is to run `eval $(crc oc-env)` and use the version that we have bundled with `crc`.

robin-owen: Updated the PR. Please add any comments to the newer diff. :)

docs/source/getting-started/content/proc_starting-codeready-containers-behind-proxy.adoc
```diff
@@ -0,0 +1,31 @@
+[id="starting-codeready-containers-behind-proxy_{context}"]
+= Starting {prod} behind a proxy
+
+.Prerequisites
+
+* To use an existing [command]`oc` binary on your host machine, export the `.testing` domain as part of the `no_proxy` environment variable.
```
praveenkumar: :+1:

Issue #947, opened 2020-01-22
[BUG] network operator fails because of invalid proxy config


praveenkumar:
NoProxy is `[]string` type for `ProxyConfig` struct and we are directly using it with `%s` which is causing issue for different operators. 

```
$  ./crc status --log-level debug
DEBU HTTP-PROXY: http://proxy-user1:xxx@ec2-xxxx:3128, HTTPS-PROXY: http://proxy-user1:xxx@ec2-xxxx.com:3128, NO-PROXY: [127.0.0.1 localhost .testing] 
DEBU Unexpected operator status for insights: Disabled 
DEBU network operator is degraded, Reason: InvalidProxyConfig 
DEBU About to run SSH command:
df -B1 --output=size,used,target /sysroot | tail -1 
DEBU SSH cmd err, output: <nil>: 32719437824 8936333312 /sysroot 
CRC VM:          Running
OpenShift:       Stopped
Disk Usage:      8.936GB of 32.72GB (Inside the CRC VM)
Cache Usage:     30.93GB
Cache Directory: /Users/prkumar/.crc/cache
```

PR #948, opened 2020-01-22
Issue #947 use GetNoProxyString() to set no proxy for proxy resource

praveenkumar:
NoProxy is `[]string` type for ProxyConfig struct and we are directly using it with `%s` for proxy resource which need a `string` type.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -100,8 +100,8 @@ func setProxyDefaults() {
 	}
 
 	if proxyConfig.IsEnabled() {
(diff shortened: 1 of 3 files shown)


praveenkumar:
> And since proxy.NoProxy could be renamed to proxy.noProxy to avoid it being used outside of the proxy package.

@cfergeau but we are using it atm in root.go, I will double check if we can make it internal to package to avoid misuse.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

Issue #898, opened 2019-12-17
Access to  OperatorHub through proxy is not working


yeplaa:
### General information

  * OS: Linux centos 7.7
  * Hypervisor: KVM 
  * Did you run `crc setup` before starting it (Yes/No)? YEs

## CRC version
```bash
# Put the output of `crc version`
crc version: 1.2.0+c2e3c0f
OpenShift version: 4.2.8 (embedded in binary)

  
## CRC status
```bash
# Put the output of `crc status`
CRC VM:          Running
OpenShift:       Running (v4.2.8)
Disk Usage:      16.73GB of 32.2GB (Inside the CRC VM)
Cache Usage:     11.01GB
Cache Directory: /home/centos/.crc/cache


## CRC config
```bash
# Put the output of `crc config view`
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
# put the output of `sw_vers` in case of Mac
# Put the output of `systeminfo` in case of Windows
cat /etc/os-release
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"



### Steps to reproduce

I have used your https://github.com/code-ready/crc/wiki/Use-proxy-with-CRC doc mandatory to deploy crc front of a proxy. My proxy configuration is deployed as indicated into the doc.
The crc deployment is good.

### Expected
I want deploy operator by OperatorHub.

### Actual

I have currently a problem with the OperatorHub with the "No OperatorHub Items Found" message.
In the Operatorsource configuration the endpoint access is KO : 

oc get opsrc   -n openshift-marketplace
NAME                  TYPE          ENDPOINT              REGISTRY              DISPLAYNAME           PUBLISHER   STATUS        MESSAGE                                                                                                                                 AGE
[...]

PR #954, opened 2020-01-23
Issue #898 Add proxy configuration to openshift-marketplace

praveenkumar:
Add proxy configuration to openshift-marketplace

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -176,6 +176,21 @@ Environment=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,%s`
 	return nil
 }
 
+// Only marketplace operator doesn't get updated with proxy when it is set using proxy resource
+// All other operators which have `config.openshift.io/inject-proxy` annotation get updated except marketplace.
+func AddProxyConfigToMarketplaceOperator(oc oc.OcConfig, proxy *network.ProxyConfig) error {
+	cmdArgs := []string{"set", "env", "deployment", "marketplace-operator", "-n", "openshift-marketplace",
+		fmt.Sprintf(`HTTP_PROXY="%s", HTTPS_PROXY="%s", NO_PROXY="%s"}}`, proxy.HttpProxy, proxy.HttpsProxy, proxy.GetNoProxyString()),
+	}
+	if err := oc.WaitForOpenshiftResource("deployment"); err != nil {
+		return err
+	}
+	if _, stderr, err := oc.RunOcCommand(cmdArgs...); err != nil {
+		return fmt.Errorf("Failed to add proxy details to marketplace-operator %v: %s", err, stderr)
+	}
+	return nil
+}
+
 func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error {
 	_, err := sshRunner.RunPrivate(fmt.Sprintf("cat <<EOF | sudo tee /var/lib/kubelet/config.json\n%s\nEOF", pullSec))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cfergeau:
> As of now all other operators which have config.openshift.io/inject-proxy
annotation get updated except marketplace when proxy resource is defined.

How do you check this? I tried to look at `oc get deployments --all-namespaces -o yaml`, find deployments with `inject-proxy`:
```
$  ~/.crc/bin/oc get deployment -n openshift-image-registry  -o yaml |grep inject-proxy
      config.openshift.io/inject-proxy: cluster-image-registry-operator
```

Then I looked at `oc get pods -n openshift-image-registry`, and then `oc debug pods/cluster-image-registry-operator-f9697f69d-mzcgb  -n openshift-image-registry` and `printenv` does not show a HTTP_PROXY variable defined even if it's there in `kubelet` environment.

Should I be testing this differently?

praveenkumar:
> Then I looked at oc get pods -n openshift-image-registry, and then oc debug pods/cluster-image-registry-operator-f9697f69d-mzcgb -n openshift-image-registry and printenv does not show a HTTP_PROXY variable defined even if it's there in kubelet environment.

@cfergeau you can just check using `oc get pod <podname> -n openshift-image-registry -oyaml` to check the proxy env in the container spec, I think if you do `oc rsh <pod_name> -n openshift-image-registry` and then `env` or `printenv` should also list it.

cfergeau:
> > Then I looked at oc get pods -n openshift-image-registry, and then oc debug pods/cluster-image-registry-operator-f9697f69d-mzcgb -n openshift-image-registry and printenv does not show a HTTP_PROXY variable defined even if it's there in kubelet environment.
> 
> @cfergeau you can just check using `oc get pod <podname> -n openshift-image-registry -oyaml` to check the proxy env in the container spec, I think if you do `oc rsh <pod_name> -n openshift-image-registry` and then `env` or `printenv` should also list it.

```sh
$ for name in $(oc get pods -n openshift-image-registry -oname); do oc get $name -n openshift-image-registry -oyaml |grep -i proxy; done
$
```

```sh
$ sudo cat /proc/$(pgrep -f kubelet)/environ
...HTTP_PROXY=http://squid.redhat.comHTTPS_PROXY=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,127.0.0.1,localhost,.testing,192.168.130.11...
```

Proxy support is enabled in kubelet as far as I can tell, but is not propagated to the `openshift-image-registry` operator.

cfergeau:
> Proxy support is enabled in kubelet as far as I can tell, but is not propagated to the `openshift-image-registry` operator.

This was a local issue, I was using an outdated crc binary which was setting an invalid no_proxy configuration. After recreating the cluster, it's better.

I looked again at the deployments with `inject-proxy` and at the corresponding pods, and now most of them have the proxy settings set.
One notable exception was `openshift-ingress`, but maybe that's expected.

I also tried `oc scale --replicas=1 -n openshift-insights deployment insights-operator` and then looked at `oc get pods -n openshift-insights -oyaml` and HTTP_PROXY is not set there.

So this PR feels like a short-term workaround for an issue we don't fully understand.
I don't necessarily object to adding this as a stop gap, but this will probably come back sooner or later :)

praveenkumar:
retest this please.

praveenkumar:
retest this please.

cfergeau:
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/podtemplatespec.go#L105 might be why `openshift-image-registry` does the right thing wrt proxy.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

praveenkumar:
retest this please

PR #959, opened 2020-01-24
Fix misc no_proxy issues

cfergeau:


pkg/crc/machine/machine.go
```diff
@@ -290,6 +290,13 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, errors.Newf("Error determining host IP: %v", err)
 	}
 
+	proxyConfig, err := getProxyConfig(crcBundleMetadata.ClusterInfo.BaseDomain)
+	if err != nil {
+		result.Error = err.Error()
+		return *result, errors.Newf("Error getting proxy configuration: %v", err)
+	}
+	proxyConfig.ApplyToEnvironment()
```
praveenkumar: @cfergeau should this will be like 
```golang
if proxyConfig.IsEnabled() {
    proxyConfig.ApplyToEnvironment()
}
```

pkg/crc/machine/machine.go
```diff
@@ -765,29 +765,32 @@ func getProxyConfig(baseDomainName string) (*network.ProxyConfig, error) {
...
+		// Restart the crio service
+		if proxy.IsEnabled() {
+			// Restart reload the daemon and then restart the service
+			// So no need to explicit reload the daemon.
+			if _, ferr := sd.Restart("crio"); ferr != nil {
+				err = ferr
+			}
 		}
+	}()
+	proxy.ApplyToEnvironment()
```
praveenkumar: @cfergeau why we again have to apply to env?

pkg/crc/machine/machine.go
```diff
@@ -537,6 +537,18 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
...
+		if err != nil {
+			result.Error = err.Error()
+			return *result, errors.Newf("Error loading bundle metadata: %v", err)
+		}
+		proxyConfig, err := getProxyConfig(crcBundleMetadata.ClusterInfo.BaseDomain)
+		if err != nil {
+			result.Error = err.Error()
+			return *result, errors.Newf("Error getting proxy configuration: %v", err)
+		}
+		proxyConfig.ApplyToEnvironment()
```
praveenkumar: same here shouldn't we first check if proxy is enable before apply to env?

cfergeau: I can add it if you want, `proxyConfig.ApplyToEnvironment()` starts with `if !proxy.IsEnabled() return`

pkg/crc/machine/machine.go
cfergeau: This `proxy.ApplyToEnvironment()` call was already there before this PR, it could indeed be removed by the first commit, I'll do that.

pkg/crc/machine/machine.go
praveenkumar: @cfergeau Ah right :( no need

pkg/crc/machine/machine.go
```diff
@@ -757,34 +776,32 @@ func getProxyConfig(baseDomainName string) (*network.ProxyConfig, error) {
 }
 
 func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd *systemd.InstanceSystemdCommander,
-	baseDomainName, instanceIP string) (err error) {
-	proxy, err := getProxyConfig(baseDomainName)
-	if err != nil {
-		return err
+	proxy *network.ProxyConfig, instanceIP string) (err error) {
+	if !proxy.IsEnabled() {
+		return nil
```

Issue #962, opened 2020-01-27
Add option to allow the use of preview/experimental features


gbraad:
\

anjannath:
Fixed via f2a4bc9526d72c736d73cbe1cafd883aa6fab55f

PR #964, opened 2020-01-27
Experimental embed tray

gbraad:
Combines #931 and #962 to prevent the tray from installing by default

This should expose a flag for `setup` named `enable-experimental-features` to select additional tech-preview features to be enabled. This should be used as:

```
$ crc setup --enable-experimental-features
```

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -19,7 +19,8 @@ var (
 	NameServer     = cfg.AddSetting("nameserver", nil, []cfg.ValidationFnType{cfg.ValidateIpAddress}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	PullSecretFile = cfg.AddSetting("pull-secret-file", nil, []cfg.ValidationFnType{cfg.ValidatePath}, []cfg.SetFn{cfg.SuccessfullyApplied})
 
-	DisableUpdateCheck = cfg.AddSetting("disable-update-check", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	DisableUpdateCheck   = cfg.AddSetting("disable-update-check", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	ExperimentalFeatures = cfg.AddSetting("enable-experimental-features", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
 
 	// Proxy Configuration
 	HttpProxy  = cfg.AddSetting("http-proxy", nil, []cfg.ValidationFnType{cfg.ValidateURI}, []cfg.SetFn{cfg.SuccessfullyApplied})
--- a/pkg/crc/preflight/preflight_darwin.go
+++ b/pkg/crc/preflight/preflight_darwin.go
@@ -39,6 +39,15 @@ var dnsPreflightChecks = [...]PreflightCheck{
 	},
 }
 
+var traySetupCheck = PreflightCheck{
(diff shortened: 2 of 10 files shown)


gbraad:
Please check if this does the job on macOS. It should only enable the systray features when experimental flag is given.

gbraad:
Can @praveenkumar or @cfergeau test this quickly? It will only do the `setup` command... (and therefore on `start` it will not have any knowledge of the experimental stuff.

cmd/crc/cmd/config/config.go
```diff
@@ -19,7 +19,8 @@ var (
 	NameServer     = cfg.AddSetting("nameserver", nil, []cfg.ValidationFnType{cfg.ValidateIpAddress}, []cfg.SetFn{cfg.SuccessfullyApplied})
 	PullSecretFile = cfg.AddSetting("pull-secret-file", nil, []cfg.ValidationFnType{cfg.ValidatePath}, []cfg.SetFn{cfg.SuccessfullyApplied})
 
-	DisableUpdateCheck = cfg.AddSetting("disable-update-check", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	DisableUpdateCheck   = cfg.AddSetting("disable-update-check", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	ExperimentalFeatures = cfg.AddSetting("enable-experimental-features", nil, []cfg.ValidationFnType{cfg.ValidateBool}, []cfg.SetFn{cfg.SuccessfullyApplied})
```
cfergeau: I could not check this on the mac, but I tried on linux and marked some parts of the preflights as experimental. You need the diff below for `--enable-experimental-features` (on the commandline) to work. `crc config enable-experimentatl-features` is still working fine after this patch.

```diff
diff --git a/cmd/crc/cmd/setup.go b/cmd/crc/cmd/setup.go
index d36725de..8fd61345 100644
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -18,6 +18,7 @@ func init() {
 	setupCmd.Flags().StringVarP(&vmDriver, config.VMDriver.Name, "d",
 		machine.DefaultDriver.Driver, fmt.Sprintf("The driver to use for the OpenShift cluster. Possible values: %v", machine.SupportedDriverValues()))
 	setupCmd.Flags().Bool(config.ExperimentalFeatures.Name, false, "Allow the use of experimental features")
+	crcConfig.BindFlagSet(setupCmd.Flags())
 
 	rootCmd.AddCommand(setupCmd)
 }
```

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -55,7 +55,11 @@ func getPreflightChecks() []PreflightCheck {
 	checks = append(checks, nonWinPreflightChecks[:]...)
 	checks = append(checks, hyperkitPreflightChecks[:]...)
 	checks = append(checks, dnsPreflightChecks[:]...)
-	checks = append(checks, traySetupCheck)
+
+	// Experimental feature
+	if EnableExperimentalFeatures {
```
cfergeau: Do we want a warning to let users know the feature exists?

cfergeau:
> One minor bug to fix, one question, I think there were a few small things to adjust in Anjan's PR, apart from this, looks good

And lack of commit logs :(

cmd/crc/cmd/config/config.go
gbraad: Not sure why this is needed... as the `vmdriver` worked like this (which should be removed actually; follow-up)

cfergeau: I suspect the `vmdriver` parameter was buggy too, we probably did not use values set in the config.

praveenkumar:
closing it, since already merged by other PR.

PR #967, opened 2020-01-27
Issue #962 Remove configKeySuffix field from tray preflight check

anjannath:
This was missed in PR #931


pkg/crc/preflight/preflight_darwin.go
```diff
@@ -40,7 +40,6 @@ var dnsPreflightChecks = [...]PreflightCheck{
 }
 
 var traySetupCheck = PreflightCheck{
-	configKeySuffix:  "",
```
praveenkumar: :+1:

PR #973, opened 2020-01-28
Issue #966 Remove driver option from setup

### LINKED ISSUE #966: Remove `vm-driver` flag from setup command
We no longer need this flag as only the default drivers are supported.

praveenkumar:
CRC only support the native hypervisor for each platform and doesn't
provide any different driver to connect a non-native hypervisor. This
PR will remove unused `driver` option from setup.

Fix #966

### DIFF
--- a/pkg/crc/config/validations.go
+++ b/pkg/crc/config/validations.go
@@ -22,14 +22,6 @@ func ValidateBool(value interface{}) (bool, string) {
 	return false, "must be true or false"
 }
 
-// ValidateDriver checks if driver is valid in the config
-func ValidateDriver(value interface{}) (bool, string) {
-	if err := validation.ValidateDriver(value.(string)); err != nil {
-		return false, err.Error()
-	}
-	return true, ""
-}
-
 // ValidateCPUs checks if provided cpus count is valid in the config
 func ValidateCPUs(value interface{}) (bool, string) {
 	v, err := strconv.Atoi(value.(string))
--- a/test/integration/features/story_health.feature
+++ b/test/integration/features/story_health.feature
@@ -4,23 +4,14 @@ Feature:
     project and deploy an app. Check on the app and delete the
     project. Stop and delete CRC.
 
-    Scenario Outline: Start CRC
+    @linux @darwin
+    Scenario: Start CRC
         Given executing "crc setup" succeeds
-        When starting CRC with default bundle and hypervisor "<vm-driver>" succeeds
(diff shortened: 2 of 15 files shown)


gbraad:
Does this also still exist on `start` ? I do see the integration test using `-d libvirt`, which makes no sense anymore.

gbraad:
https://github.com/code-ready/crc/blob/d1b374c1c0e35921bb3bcf66d3a97e6f854e4b39/test/integration/crcsuite/crcsuite.go#L372

test/integration/features/story_health.feature
```diff
@@ -4,24 +4,15 @@ Feature:
     project and deploy an app. Check on the app and delete the
     project. Stop and delete CRC.
 
-    Scenario Outline: Start CRC
+    @linux @darwin
+    Scenario: Start CRC
```
cfergeau: Not sure about these changes from `Scenario Outline` to `Scenario`? This happens several times in this patch.

praveenkumar: @cfergeau if you use `Scenario Outline` you need to put `examples` fields with table like it was there before the change but since we dont' need the table so it need to be changed also.

jsliacan: @cfergeau as @praveenkumar says, maybe we can delete the table, change `Scenario Outline` to `Scenario` and replace `starting CRC with default bundle and hypervisor <vm-driver> succeeds` with 
```
starting CRC with default bundle and default hypervisor succeeds
```
This will **not** use the `--vm-driver` option. I can clean up the "hypervisor" wording from the text later.

Another option is that I submit a PR tomorrow changing these things, then you rebase onto that (if you want).

cfergeau: Did you look at the PR? I believe this already is doing the changes you are thinking of?

jsliacan: My bad, just looked at these comments instead of the whole PR.

pkg/crc/config/validations.go
```diff
@@ -22,14 +22,6 @@ func ValidateBool(value interface{}) (bool, string) {
 	return false, "must be true or false"
 }
 
-// ValidateDriver checks if driver is valid in the config
-func ValidateDriver(value interface{}) (bool, string) {
-	if err := validation.ValidateDriver(value.(string)); err != nil {
```
cfergeau: after this change, `validation.ValidateDriver` is unused, so this method can also be removed.

Issue #636, opened 2019-09-24
oc-env should indicate what to do next


cfergeau:
On linux/macos, it could also have:
```
echo "You can now login to the OpenShift cluster with `oc login xxx`
```

cfergeau:
It's a bit odd though to show it all the time, as this needs to only be done once.

gbraad:
@cfergeau is this still needed...

also see #708

cfergeau:
>The idea is not for oc-env to provide a user-guide to using the oc command. this is something that should come from documentation and even a higher-level documentation.

I'm fine with that resolution.

PR #975, opened 2020-01-28
Issue #636: Mention oc login at the end of oc-env

cfergeau:
This fixes https://github.com/code-ready/crc/issues/636

### DIFF
--- a/cmd/crc/cmd/oc_env.go
+++ b/cmd/crc/cmd/oc_env.go
@@ -23,16 +23,17 @@ var ocEnvCmd = &cobra.Command{
 			errors.ExitWithMessage(1, "Error running the oc-env command: %s", err.Error())
 		}
 
-		proxyConfig, err := machine.GetProxyConfig(constants.DefaultName)
+		clusterConfig, err := machine.GetClusterConfig(constants.DefaultName)
 		if err != nil {
 			errors.Exit(1)
 		}
 		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
-		if proxyConfig.IsEnabled() {
-			output.Outln(shell.GetEnvString(userShell, "HTTP_PROXY", proxyConfig.HttpProxy))
-			output.Outln(shell.GetEnvString(userShell, "HTTPS_PROXY", proxyConfig.HttpsProxy))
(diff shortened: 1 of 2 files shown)


cmd/crc/cmd/oc_env.go
```diff
@@ -23,16 +23,17 @@ var ocEnvCmd = &cobra.Command{
...
-		if proxyConfig.IsEnabled() {
-			output.Outln(shell.GetEnvString(userShell, "HTTP_PROXY", proxyConfig.HttpProxy))
-			output.Outln(shell.GetEnvString(userShell, "HTTPS_PROXY", proxyConfig.HttpsProxy))
-			output.Outln(shell.GetEnvString(userShell, "NO_PROXY", proxyConfig.GetNoProxyString()))
+		if clusterConfig.ProxyConfig.IsEnabled() {
+			output.Outln(shell.GetEnvString(userShell, "HTTP_PROXY", clusterConfig.ProxyConfig.HttpProxy))
+			output.Outln(shell.GetEnvString(userShell, "HTTPS_PROXY", clusterConfig.ProxyConfig.HttpsProxy))
+			output.Outln(shell.GetEnvString(userShell, "NO_PROXY", clusterConfig.ProxyConfig.GetNoProxyString()))
 		}
+		output.Outf("echo \"You can now login to the OpenShift cluster with 'oc login -p developer -u developer %s'\"\n", clusterConfig.ClusterAPI)
```
praveenkumar: @cfergeau only for developer user or also for admin?

cfergeau: Not sure we should advertise the admin user too strongly?

praveenkumar: @cfergeau most of the user who want to experiment with operators like making any config changes or installing a new one require the the admin even with operator hub so I think it should be good to have both user mentioned here like we do for `crc console credentials`.

cfergeau: But if we push too much the admin credentials, won't crc users always use that, and just ignore the `developer` user?

cfergeau:
Ignore the commits I pushed, I did these by mistake ;)

cmd/crc/cmd/oc_env.go
praveenkumar: @cfergeau it's just user choice we should provide them both the options like we are providing for other user facing commands.

Issue #978, opened 2020-01-29
Unable to use OpenShift 4.3 for CRC


gbraad:
Currently we are unable to produce usable images for CRC that run OpenShift 4.3. Details for this can be found in: https://bugzilla.redhat.com/show_bug.cgi?id=1795163

If you need to run 4.3, please consider using the installer to create a regular cluster (avoiding to set up a SNC).

tmds:
@gbraad I'd like to run OpenShift 4.3 on my dev machine and found this issue.
Any recommendation on how I can run 4.3 (other can `crc`)?
Do you think there will be a version of `crc` soon that supports 4.3?

gbraad:
We have a workaround that is planned for the coming release. Code freeze is this Friday.

praveenkumar:
I am closing this issue since now we have a workaround which works for 4.3.x and also made a release.

gbraad:
We still have issues with 4.3.x and 4.4 that are tracked on our projects boards https://github.com/code-ready/crc/projects/24

PR #979, opened 2020-01-29
Update bundle from 4.3.0 to 4.2.16

praveenkumar:
Since we are facing issues with 4.3.0 bundles[0], we need to revert
back to 4.2.x to make CI stable again.

- https://bugzilla.redhat.com/show_bug.cgi?id=1795163

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


gbraad:
Rather call this downgrade and use the issue #978 or so.

Also, what about the release tags?

praveenkumar:
retest this please.

praveenkumar:
retest this please.

PR #987, opened 2020-02-03
[WIP] Issue #978 Enable required deployments

praveenkumar:


praveenkumar:
retest this please.

praveenkumar:
retest this please.

pkg/crc/cluster/cluster.go
```diff
@@ -198,3 +219,47 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
...
+	}
+
+	if _, stderr, err := oc.RunOcCommand(cmdArgs...); err != nil {
+		return fmt.Errorf("Failed to start Openshift apiserver  %v: %s", err, stderr)
+	}
+	return nil
+}
+
+func WaitforRequestHeaderClientCaFile(oc oc.OcConfig) error {
+	cmdArgs := []string{"get", "cm/extension-apiserver-authentication", `-ojsonpath={.data.requestheader-client-ca-file}`,
```
cfergeau: It might be possible to use `oc observe configmap -n kube-system` to watch for the changes, but seems a bit tricky to do properly.

pkg/crc/machine/machine.go
```diff
@@ -382,7 +382,21 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	}
 	result.KubeletStarted = kubeletStarted
 
-	time.Sleep(time.Minute * 3)
+	if err := cluster.StartKubeApiserverOperator(ocConfig); err != nil {
+		result.Error = err.Error()
+		return *result, errors.New(err.Error())
+	}
+
+	if err := cluster.WaitforRequestHeaderClientCaFile(ocConfig); err != nil {
```
cfergeau: With the addition of that method, we don't strictly need the whole "enable required deployments" code, we could limit ourselves to restarting `openshift-apiserver` when `requestheader-client-ca-file` changes.

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau yes looks like if we use `oc observe` we need to have a script associate with it which will make it bit complicated then just query ourself and see if that resource is updated.

pkg/crc/machine/machine.go
praveenkumar: Yes, we can do that also but I do find this bit more appealing since it provide us more control over crc side then snc side (which is a long process to test simple changes) . cc @gbraad wdyt?

praveenkumar:
closing this since we might not going to use it.

PR #992, opened 2020-02-07
Issue #978 Wait for updation of client-ca request header

praveenkumar:


gbraad:
What is the reason for this change and how can this be tested?

pkg/crc/cluster/cluster.go
```diff
@@ -198,3 +198,39 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
...
+	if err != nil {
+		return fmt.Errorf("Failed to get request header client ca file %v: %s", err, stderr)
+	}
+	for stdout == "" {
+		stdout, stderr, err = oc.RunOcCommand(cmdArgs...)
+		if err != nil {
+			return fmt.Errorf("Failed to get request header client ca file %v: %s", err, stderr)
+		}
+		logging.Debug(stdout)
+		time.Sleep(2 * time.Second)
```
cfergeau: This should have a timeout in case the header never appears.

pkg/crc/machine/machine.go
```diff
@@ -378,6 +378,17 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, errors.Newf("Error starting kubelet: %s", err)
 	}
 	if kubeletStarted {
+		logging.Info("Waiting for updation of client-ca request header ...")
```
cfergeau: "update", not "updation". Since it's a workaround for an openshift bug, I would put this at logging.Debug(), and this deserves an explanation why we do that.

pkg/crc/cluster/cluster.go
```diff
@@ -198,3 +198,39 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
 	}
 	return nil
 }
+
+func WaitforRequestHeaderClientCaFile(oc oc.OcConfig) error {
+	cmdArgs := []string{"get", "cm/extension-apiserver-authentication", `-ojsonpath={.data.requestheader-client-ca-file}`,
```
cfergeau: `configmap` instead of `cm` maybe?

praveenkumar:
retest this please.

praveenkumar:
retest this please.

pkg/crc/cluster/cluster.go
```diff
@@ -198,3 +198,44 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
...
+	for i := 0; i < 90; i++ {
+		stdout, stderr, err := oc.RunOcCommand(cmdArgs...)
+		if err != nil {
+			return fmt.Errorf("Failed to get request header client ca file %v: %s", err, stderr)
+		}
+		logging.Debug(stdout)
+		if stdout != "" {
+			break
+		}
+		time.Sleep(2 * time.Second)
```
cfergeau: Could this use `errors.RetryAfter`?

pkg/crc/machine/machine.go
```diff
@@ -378,6 +378,23 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, errors.Newf("Error starting kubelet: %s", err)
 	}
 	if kubeletStarted {
+		// With 4.3, when cluster come up and request header client-ca regenerated then
```
cfergeau: "when the cluster comes up", this only happens on first start, I guess this should be explicit.
"request header client-ca regenerated " I'm not sure I understand this part of the sentence :-/

pkg/crc/machine/machine.go
```diff
@@ -378,6 +378,23 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+		// openshift-apiserver pod is not able to consume it and as a result it shows
+		// `certificate has expired or is not yet valid` in log message. To resolve
```
cfergeau: "openshift-apiserver pod is not able to consume it and as a result it shows `certificate has expired or is not yet valid` in log message. " This part is not accurate, right after boot, this "certificate has expired or is not yet valid" message appears in the logs, and `client-ca` is not set in the config map. Then the certificates and CAs are regenerated, `client-ca` shows up in the config map, and then the message changes to "unknown authority"

pkg/crc/cluster/cluster.go
praveenkumar: `errors.RetryAfter` is used in case of error retry but here we are retry because of the stdout shouldn't be empty string, error can be nil in this case since the configmap resource is available and command will successful by returning an empty string.

praveenkumar:
retest this please.

praveenkumar:
retest this please.

pkg/crc/cluster/cluster.go
cfergeau: but nothing prevents us from returning our own RetriableError when stdout is empty?
untested, but something like this:
```
diff --git a/pkg/crc/cluster/cluster.go b/pkg/crc/cluster/cluster.go
index 73a755a6..cabb9238 100644
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -198,3 +198,44 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
 	}
 	return nil
 }
+
+func WaitforRequestHeaderClientCaFile(oc oc.OcConfig) error {
+	if err := oc.WaitForOpenshiftResource("configmaps"); err != nil {
+		return err
+	}
+
+	lookupRequestHeaderClientCa := func() error {
+		cmdArgs := []string{"get", "configmaps/extension-apiserver-authentication", `-ojsonpath={.data.requestheader-client-ca-file}`,
+			"-n", "kube-system"}
+
+		stdout, stderr, err := oc.RunOcCommand(cmdArgs...)
+		if err != nil {
+			return fmt.Errorf("Failed to get request header client ca file %v: %s", err, stderr)
+		}
+		if stdout == "" {
+			return &errors.RetriableError{Err: fmt.Errorf("missing .data.requestheader-client-ca-file")}
+		}
+		logging.Debugf("Found .data.requestheader-client-ca-file: %s", stdout)
+		return nil
+	}
+
+	return errors.RetryAfter(90, lookupRequestHeaderClientCa, 2*time.Second)
+}
```

praveenkumar: @cfergeau yes if we create our own error message then we can use it, as you described. I don't want to use it this non error message but if you insist I can just add same :)

cfergeau: Not going to insist, was just a suggestion ;) If you prefer it without RetryAfter, fine with me.

pkg/crc/cluster/cluster.go
```diff
@@ -198,3 +198,44 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
 	}
 	return nil
 }
+
+func WaitforRequestHeaderClientCaFile(oc oc.OcConfig) error {
+	if err := oc.WaitForOpenshiftResource("configmaps"); err != nil {
+		return err
+	}
+
+	// Timeout after 180 seconds
```
cfergeau: I don't think this comment is required.

pkg/crc/cluster/cluster.go
```diff
@@ -198,3 +198,44 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
...
+			return &errors.RetriableError{Err: fmt.Errorf("missing .data.requestheader-client-ca-file")}
+		}
+		logging.Debugf("Found .data.requestheader-client-ca-file: %s", stdout)
+		return nil
+	}
+	return errors.RetryAfter(90, lookupRequestHeaderClientCa, 2*time.Second)
+}
+
+func DeleteOpenshiftApiServerPod(oc oc.OcConfig) error {
+	cmdArgs := []string{"delete", "pod", "--all", "-n", "openshift-apiserver"}
```
cfergeau: this could be moved inside `deleteOpenshiftApiserverPod` definition

pkg/crc/cluster/cluster.go
```diff
@@ -198,3 +198,44 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
...
+		if stdout == "" {
+			return &errors.RetriableError{Err: fmt.Errorf("missing .data.requestheader-client-ca-file")}
+		}
+		logging.Debugf("Found .data.requestheader-client-ca-file: %s", stdout)
+		return nil
+	}
+	return errors.RetryAfter(90, lookupRequestHeaderClientCa, 2*time.Second)
+}
+
+func DeleteOpenshiftApiServerPod(oc oc.OcConfig) error {
```
cfergeau: Maybe use plural, `Pods` since the method is using `--all`?

praveenkumar:
> 4 and 5 are a bit confusing, imo it's 5. which happens first, /etc/kubernetes/static-pod-resources/..../aggregator-ca is regenerated, this new CA gets set as request-header's client-ca, and openshift-apiserver reads this value.

@cfergeau new CA gets set as request-header's client-ca in the config map  is kind of implicit that /etc/kubernetes/static-pod-resources/../aggregrator-ca is generated, if you think we should mention that also explicit I will update it but step 5 does happen after step-4, until the CA is not part of config map apiserver-pod not able to read it and show error logs `missing content for CA bundle "client-ca::kube-system::extension-apiserver-authentication::requestheader-client-ca-file"` mentioned in the step-3.

cfergeau:
> > 4 and 5 are a bit confusing, imo it's 5. which happens first, /etc/kubernetes/static-pod-resources/..../aggregator-ca is regenerated, this new CA gets set as request-header's client-ca, and openshift-apiserver reads this value.
> 
> @cfergeau new CA gets set as request-header's client-ca in the config map is kind of implicit that /etc/kubernetes/static-pod-resources/../aggregrator-ca is generated, 

the fact that request-header-client-ca corresponds to the content of the aggregator-ca.crt file can be useful info I believe? Or is it too much information?

> if you think we should mention that also explicit I will update it but step 5 does happen after step-4, until the CA is not part of config map apiserver-pod not able to read it and show error logs `missing content for CA bundle "client-ca::kube-system::extension-apiserver-authentication::requestheader-client-ca-file"` mentioned in the step-3.

It is the "the CA which is regenerated." which felt out of order, the CA is regenerated first, then 4 and 5 happen.

And I totally forgot to mention that it's missing a few sentences at the end :) That when the openshift-apiserver is in that state, the cluster is non functional, that a restart of the openshift-apiserver pod is enough to clear that error and get a working cluster. And that this problem started happening with OpenShift 4.3, and that this is a work-around while the root cause is being identified.

Issue #999, opened 2020-02-11
CRC daemon ignoring config changes if already running


jsliacan:
### General information

  * OS: macOS
  * Hypervisor: hyperkit

## CRC version
```bash
crc version: 1.5.0+e1c8fb8
OpenShift version: 4.2.14 (embedded in binary)
```
## Context
```
$ crc setup
$ crc daemon
$ crc config set pull-secret-file <location of pull secret file>
$ nc -U ~/.crc/crc.sock
{"command":"start"}
```

### Expected
CRC starts as usual

### Actual
```
{"Name":"crc","Status":"","Error":"open : no such file or directory","ClusterConfig":{"KubeConfig":"","KubeAdminPass":"","ClusterAPI":"","WebConsoleURL":"","ProxyConfig":null},"KubeletStarted":false}
```
with `~/.crc/crcd.log` containing the following
```
time="2020-02-11T10:46:52+01:00" level=error msg="Error reading from socket"                                                                                                                                                                  
time="2020-02-11T10:47:09+01:00" level=debug msg="Received Request:{\"command\":\"start\"}\n"                                                                                                                                                 
time="2020-02-11T10:47:09+01:00" level=error msg="Failed to get pull secret: open : no such file or directory"                                                                                                                                
```
When the user restarts the daemon or first sets the location of the pull secret and then runs the daemon, all is OK.

PR #1000, opened 2020-02-11
Issue #999 Call WatchConfig in viper initialization

anjannath:
Since we initialize viper in root.go and the daemon runs that
only once, the new changes to the config is not reflected on the
daemon side, this should make daemon aware of changes made to the
config file, after it has been launched.

This needs to be tested specially on macOS and windows, since it depends on file system notifications to work as expected.

### DIFF
--- a/pkg/crc/config/viper_config.go
+++ b/pkg/crc/config/viper_config.go
@@ -101,6 +101,7 @@ func InitViper() error {
 	if err != nil {
 		return fmt.Errorf("Error reading configuration file '%s': %v", constants.ConfigFile, err)
 	}
+	v.WatchConfig()
 	globalViper = v
 	return v.Unmarshal(&changedConfigs)
 }


anjannath:
> I did not check how we are using config values in our codebase, but could this cause weird issues if one starts a long running operation (crc start comes to mind), and changes some of the config values while it's running?

We read the config values once for start, and then use it, this change makes sure that when you do read a config value, its the latest one.

cfergeau:
> We read the config values once for start, and then use it,

`git grep` seems to confirm this:
```
$ git grep "crc/config"
cmd/crc/cmd/config/config.go:   cfg "github.com/code-ready/crc/pkg/crc/config"
cmd/crc/cmd/config/get.go:      "github.com/code-ready/crc/pkg/crc/config"
cmd/crc/cmd/config/set.go:      "github.com/code-ready/crc/pkg/crc/config"
cmd/crc/cmd/config/unset.go:    "github.com/code-ready/crc/pkg/crc/config"
cmd/crc/cmd/config/view.go:     "github.com/code-ready/crc/pkg/crc/config"
cmd/crc/cmd/root.go:    "github.com/code-ready/crc/pkg/crc/config"
cmd/crc/cmd/setup.go:   crcConfig "github.com/code-ready/crc/pkg/crc/config"
cmd/crc/cmd/start.go:   crcConfig "github.com/code-ready/crc/pkg/crc/config"
pkg/crc/api/handlers.go:        crcConfig "github.com/code-ready/crc/pkg/crc/config"
pkg/crc/preflight/preflight.go: cfg "github.com/code-ready/crc/pkg/crc/config"
```
Almost no users outside of `cmd/`

pkg/crc/config/viper_config.go
```diff
@@ -101,6 +101,7 @@ func InitViper() error {
 	if err != nil {
 		return fmt.Errorf("Error reading configuration file '%s': %v", constants.ConfigFile, err)
 	}
+	v.WatchConfig()
```
praveenkumar: :+1:

PR #1001, opened 2020-02-11
Add podman-env command

### LINKED ISSUE #961: Add `podman-env` command to allow access to the VM container runtime
As described in the #874 it is possible to re-use the VM to run containers from podman. This task will add the needed commands to crc, such as:

  * [x] add `podman-env`
     * [x] this sets up the environment inside the VM
     * [x] sets env vars to allow access to the runtime
  * [x] embed the `podman-remote` binary

zeenix:
Fixes #961 

Add subcommand to setup environment variables to use `podman-remote` with the CRC VM.

This handles:

- [x]  the environment setup
- [x] Binary bundling/installation for all 3 OSs.

### DIFF
--- a/cmd/crc/cmd/podman_env.go
+++ b/cmd/crc/cmd/podman_env.go
@@ -1,0 +1,46 @@
+package cmd
+
+import (
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/machine"
+	"github.com/code-ready/crc/pkg/crc/output"
+	"github.com/code-ready/crc/pkg/os/shell"
+	"github.com/spf13/cobra"
+)
+
+var podmanEnvCmd = &cobra.Command{
+	Use:   "podman-env",
+	Short: "Setup podman environment",
+	Long:  `Setup environment for 'podman' binary to access podman on CRC VM`,
+	Run: func(cmd *cobra.Command, args []string) {
+		userShell, err := shell.GetShell(forceShell)
+		if err != nil {
+			errors.ExitWithMessage(1, "Error running the podman-env command: %s", err.Error())
+		}
+
+		ipConfig := machine.IpConfig{
+			Name:  constants.DefaultName,
+			Debug: isDebugLog(),
+		}
+
+		exitIfMachineMissing(ipConfig.Name)
+
+		result, err := machine.Ip(ipConfig)
+		if err != nil {
+			errors.Exit(1)
+		}
+
+		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_USER", constants.DefaultSSHUser))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_HOST", result.IP))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_IDENTITY_FILE", constants.GetPrivateKeyPath()))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_IGNORE_HOSTS", "1"))
+		output.Outln(shell.GenerateUsageHint(userShell, "crc podman-env"))
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(podmanEnvCmd)
+	podmanEnvCmd.Flags().StringVar(&forceShell, "shell", "", "Set the environment for the specified shell: [fish, cmd, powershell, tcsh, bash, zsh]. Default is auto-detect.")
+}
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -30,8 +30,9 @@ const (
 	CrcLandingPageURL    = "https://cloud.redhat.com/openshift/install/crc/installer-provisioned" // #nosec G101
 	PullSecretFile       = "pullsecret.json"
 
-	DefaultOcUrlBase   = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest"
-	CrcTrayDownloadURL = "https://github.com/code-ready/tray-macos/releases/download/v%s/crc-tray-macos.tar.gz"
+	DefaultOcUrlBase     = "https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest"
+	DefaultPodmanUrlBase = "https://storage.googleapis.com/libpod-master-releases"
+	CrcTrayDownloadURL   = "https://github.com/code-ready/tray-macos/releases/download/v%s/crc-tray-macos.tar.gz"
 )
 
 var ocUrlForOs = map[string]string{
@@ -46,6 +47,16 @@ func GetOcUrlForOs(os string) string {
 
 func GetOcUrl() string {
 	return GetOcUrlForOs(runtime.GOOS)
+}
+
+var PodmanUrlForOs = map[string]string{
+	"darwin":  fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-darwin-amd64.zip"),
+	"linux":   fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-linux---amd64.zip"),
+	"windows": fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-windows-amd64.zip"),
+}
+
+func GetPodmanUrl() string {
+	return PodmanUrlForOs[runtime.GOOS]
 }
 
 var defaultBundleForOs = map[string]string{
--- a/pkg/crc/oc/oc_cache.go
+++ b/pkg/crc/oc/oc_cache.go
@@ -14,12 +14,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-const (
-	OC_CACHE_DIR = "oc"
-	TARGZ        = "tar.gz"
-	ZIP          = "zip"
-)
-
 // Oc is a struct with methods designed for dealing with the oc binary
 type OcCached struct{}
 
--- a/pkg/crc/podman/podman_cache.go
+++ b/pkg/crc/podman/podman_cache.go
@@ -1,0 +1,118 @@
+package podman
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/download"
+	"github.com/code-ready/crc/pkg/embed"
+	"github.com/code-ready/crc/pkg/extract"
+	crcos "github.com/code-ready/crc/pkg/os"
+	"github.com/pkg/errors"
+)
+
+// Podman is a struct with methods designed for dealing with the podman binary
+type PodmanCached struct{}
+
+func (podman *PodmanCached) EnsureIsCached() error {
+	if !podman.IsCached() {
+		err := podman.cachePodman()
+		if err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
+
+func (podman *PodmanCached) IsCached() bool {
+	if _, err := os.Stat(filepath.Join(constants.CrcBinDir, constants.PodmanBinaryName)); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func (podman *PodmanCached) getPodman(destDir string) (string, error) {
+	logging.Debug("Trying to extract podman from crc binary")
+	archiveName := filepath.Base(constants.GetPodmanUrl())
+	destPath := filepath.Join(destDir, archiveName)
+	err := embed.Extract(archiveName, destPath)
+	if err != nil {
+		logging.Debug("Downloading podman")
+		return download.Download(constants.GetPodmanUrl(), destDir, 0600)
+	}
+
+	return destPath, err
+}
+
+// cachePodman downloads and caches the podman binary into the CRC directory
+func (podman *PodmanCached) cachePodman() error {
+	if podman.IsCached() {
+		return nil
+	}
+
+	// Create tmp dir to download the podman tarball
+	tmpDir, err := ioutil.TempDir("", "crc")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpDir)
+	assetTmpFile, err := podman.getPodman(tmpDir)
+	if err != nil {
+		return err
+	}
+
+	// Extract the tarball and put it the cache directory.
+	err = extract.Uncompress(assetTmpFile, tmpDir)
+	if err != nil {
+		return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
+	}
+
+	binaryName := constants.PodmanBinaryName
+	binaryPath, err := findBinary(tmpDir, binaryName)
+	if err != nil {
+		return err
+	}
+
+	// Copy the requested asset into its final destination
+	outputPath := constants.CrcBinDir
+	err = os.MkdirAll(outputPath, 0750)
+	if err != nil && !os.IsExist(err) {
+		return errors.Wrap(err, "Cannot create the target directory.")
+	}
+
+	finalBinaryPath := filepath.Join(outputPath, binaryName)
+	err = crcos.CopyFileContents(binaryPath, finalBinaryPath, 0500)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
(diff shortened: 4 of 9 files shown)


cmd/crc/cmd/podman_env.go
```diff
@@ -0,0 +1,46 @@
...
+		result, err := machine.Ip(ipConfig)
+		if err != nil {
+			errors.Exit(1)
+		}
+
+		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_USER", constants.DefaultSSHUser))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_HOST", result.IP))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_IDENTITY_FILE", constants.GetPublicKeyPath()))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_IGNORE_HOSTS", "1"))
```
cfergeau: I'm not sure what this one is doing? I could not find much doc around it

zeenix: i don't know either actually, I just added this cause the original config @praveenkumar tried had this, so i just kept this around. Maybe Praveen knows?

zeenix:
> Overall looks good, though if our target is windows/macos, I think we also need the embedding part before getting this in?

Given that people can easily install podman binary on Mac, I think this is still useful even w/o embedding. For Windows too, even if installation there isn't as easy/simple.

Actually, i think for now this is all we should do until/unless podman folks can provide us a canonical location to download the latest binaries for Mac. @gbraad ?

cmd/crc/cmd/podman_env.go
praveenkumar: @cfergeau this is part of https://github.com/containers/libpod/blob/7af4074c6f96d0fabd2ab5dd6c5773053f80aa08/cmd/podman/main_remote.go one and not in the documents, may be better to have it as part of commit logs also. I am not sure if podman team have any plan to include those env in the documents.

gbraad: would it be possible to use a template instead?

gbraad:
While `easy` to install from brew on macOS, I can only merge this when the full integration is added with the `podman-remote` embed.  Please add needed changes to simplify testing.

zeenix:
> While `easy` to install from brew on macOS, I can only merge this when the full integration is added with the `podman-remote` embed.

The main task is to expose podman and since it's useful on its own, I suggested these changes to go in before binary embedding.

> Please add needed changes to simplify testing.

Of course. I had misunderstood our chat with Dan on the issue but now that i re-read it, it's clear that we can just get the Mac binary from [this location](https://storage.googleapis.com/libpod-master-releases/podman-remote-latest-master-darwin-amd64.zip) and embed it.

praveenkumar:
retest this please.

zeenix:
I just pushed my WIP patch for binary embedding. It currently doesn't work cause extractor code only extracts top-level items from the archive. I'll modify it to recurse into dirs until it finds the file..

gbraad:
Please add referring issue to the commit descriptions. had a hard time to track this when referring Dusty to this.

zeenix:
> Please add referring issue to the commit descriptions. had a hard time to track this when referring Dusty to this.

Had forgotten and it's fixed now. Maybe GH actions can be used here to notify PR submitter that they may have forgtten to reference the issue they're solving?

pkg/crc/podman/podman_cache.go
```diff
@@ -0,0 +1,95 @@
...
+	err := embed.Extract(archiveName, destPath)
+	if err != nil {
+		logging.Debug("Downloading podman")
+		return download.Download(constants.GetPodmanUrl(), destDir, 0600)
+	}
+
+	return destPath, err
+}
+
+// cachePodman downloads and caches the podman binary into the minishift directory
```
cfergeau: minishift?

zeenix: copy&paste error from oc_cache.go :)

zeenix:
I fixed and tested the binary download and extract part (finally and in the end it was super easy) and tested that bit. Binary embedding hasn't been tested yet cause I need to learn how to do that. Also need to test on Mac and Windows and test if the podman binary that we download actually can do the remote stuff (e2e test).

cmd/crc/cmd/podman_env.go
cfergeau: I removed the template stuff from `oc-env` as in my opinion it made the code harder to read for no real gain.

pkg/crc/oc/oc_cache.go
```diff
@@ -14,12 +14,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-const (
-	OC_CACHE_DIR = "oc"
-	TARGZ        = "tar.gz"
-	ZIP          = "zip"
-)
-
```
cfergeau: s/redundant/unused?

pkg/crc/podman/podman_cache.go
```diff
@@ -0,0 +1,118 @@
...
+		if filepath.Base(path) == binaryName {
+			binaryPath = path
+
+			return filepath.SkipDir
+		}
+
+		return nil
+	})
+	if binaryPath == "" && err == nil {
+		err = fmt.Errorf("Failed to find `{}` binary in `{}`.", binaryName, dirPath)
```
cfergeau: This fails `make cross-lint`, `{}` should be `%s`

pkg/crc/podman/podman_cache.go
```diff
@@ -0,0 +1,118 @@
...
+	err = crcos.CopyFileContents(binaryPath, finalBinaryPath, 0500)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func findBinary(dirPath string, binaryName string) (string, error) {
+	var binaryPath string = ""
```
cfergeau: I don't think setting it to "" is required.

pkg/crc/podman/podman_cache.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+	defer os.RemoveAll(tmpDir)
+	assetTmpFile, err := podman.getPodman(tmpDir)
+	if err != nil {
+		return err
+	}
+
+	// Extract the tarball and put it the cache directory.
+	err = extract.Uncompress(assetTmpFile, tmpDir)
```
cfergeau: Extending UncompressWithFilter to extract only the binary you are interested in was not possible?

cfergeau:
ah, and on linux, I'd really prefer using the distro package if it's recent enough.

praveenkumar:
@zeenix have you tested this on mac with our 4.3.x bundle? I am not able to get anything after setting the env variable, every time it  exit with `125` code :(

pkg/crc/podman/podman_cache.go
cfergeau: Given a tarball, UncompressWithFilter will only extract the files which names match the filter (ie the filter returns `'true`)

praveenkumar: Ah right, following logic take care of it :+1:  
```
	// the target location where the dir/file should be created
		path := filepath.Join(targetDir, header.Name)
		if fileFilter != nil && !fileFilter(path) {
			logging.Debugf("untar: Skipping %s", path)
			continue
		}
```

pkg/crc/podman/podman_cache.go
zeenix: lol, sorry drank too much Rust-aid :-D

pkg/crc/podman/podman_cache.go
zeenix: This is why it took so long. I was first trying to modify `UncompressWithFilter` to work for me but then I realized it'll be too much trouble than it's worth.

zeenix:
> @zeenix have you tested this on mac with our 4.3.x bundle?

As I wrote in https://github.com/code-ready/crc/pull/1001#issuecomment-592167612, no I've not yet. You're talking of podman binary exiting with errors? If so, I don't think this patch is responsible. We're just unpacking the binary that is provided [here](https://storage.googleapis.com/libpod-master-releases/podman-remote-latest-master-darwin-amd64.zip).

> ah, and on linux, I'd really prefer using the distro package if it's recent enough.

Why be inconsistent if we don't need to be? Actually here, IMO there is more reason to be consistent: the binary name is the same for all platforms, which isn't the case for the distro-shipped podman. So users won't get confused and irritated if they try CRC+podman on multiple platforms.

gbraad:
> ah, and on linux, I'd really prefer using the distro package if it's recent enough.

we can not gaurantee compatibility. I'd prefer not to run inconsistent versions. We are delivering a full OOTB experience. if the user wants to use the distro package, it is a choice he makes. In that case he just overrides the path settings. We have discussed this before on the actual issue and agreed to go ahead with with an embedded version on all distros... like we do with `oc`.

gbraad:
still WIP ?

cmd/crc/cmd/podman_env.go
```diff
@@ -0,0 +1,46 @@
...
+		if err != nil {
+			errors.ExitWithMessage(1, "Error running the podman-env command: %s", err.Error())
+		}
+
+		ipConfig := machine.IpConfig{
+			Name:  constants.DefaultName,
+			Debug: isDebugLog(),
+		}
+
+		exitIfMachineMissing(ipConfig.Name)
```
gbraad: Why are we using this name ?

and why not perform this ceck before the IpConfig as `exitIfMachineMissing(constants.DefaultName)` ?

For now we purposely perform all these tests with the DefaultName... IF we would ever move away from this, it would be a simple accross the board search.

cmd/crc/cmd/podman_env.go
gbraad: when considering all platform,s, it would be easy to put all templates specifc to a platform in a platform typed source code file. now it is all contained inside the actual code and harder to maintain too... specially as the environment/shell is harder to determine to be correct in the current way.

gbraad: ... even if the template is not consider, the name `GetEnvString` does not seem correct when you realize what it does.

cmd/crc/cmd/podman_env.go
zeenix: this code is based on https://github.com/code-ready/crc/blob/master/cmd/crc/cmd/oc_env.go#L26 and that does exactly this as well.

pkg/crc/podman/podman_cache.go
zeenix: unless there are strong feelings here, I'm resolving this thread.

cmd/crc/cmd/podman_env.go
gbraad: Should be refactored as a new issue in that case as bailout conditions should be considered first

gbraad:
> still WIP ?

Title got changed; WIP state removed

pkg/crc/constants/constants.go
```diff
@@ -48,6 +49,16 @@ func GetOcUrl() string {
 	return GetOcUrlForOs(runtime.GOOS)
 }
 
+var PodmanUrlForOs = map[string]string{
+	"darwin":  fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-darwin-amd64.zip"),
+	"linux":   fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-linux---amd64.zip"),
+	"windows": fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-windows-amd64.msi"),
```
gbraad: Why are we using the msi ?

gbraad: I also created #1061 as the below example is not a good practice. This can be done without the `map` and `runtime.GOOS` selection by moving these file identifiers to the respective `_darwin.go`, `_windows.go`, etc files and just use as `PodmanUrl` instead.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -19,6 +20,12 @@ var genericPreflightChecks = [...]PreflightCheck{
 		fixDescription:   "Caching oc binary",
 		fix:              fixOcBinaryCached,
 	},
+	{
+		checkDescription: "Checking if podman binary is cached",
+		check:            checkPodmanBinaryCached,
+		fixDescription:   "Caching podman binary",
```
gbraad: We are not caching the `podman` binary, but the `podman-remote` binary.

gbraad:
`crc setup` fails on `Caching podman` as a `.msi` is used instead of an archive. The extraction will fail with an exception.

an .MSI is an installer executable/script that is used to provision a machine with an application by an adminstrator.

```diff
-       "windows": fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-windows-amd64.msi"),
+       "windows": fmt.Sprintf("%s/%s", DefaultPodmanUrlBase, "podman-remote-latest-master-windows-amd64.zip"),
```

is needed to make this work.

gbraad:
Support on Windows does not work as expected for now:

```
Service:
Error: Unable to obtain server version information: unexpected EOF
```

cmd/crc/cmd/podman_env.go
```diff
@@ -0,0 +1,46 @@
...
+
+		result, err := machine.Ip(ipConfig)
+		if err != nil {
+			errors.Exit(1)
+		}
+
+		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_USER", constants.DefaultSSHUser))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_HOST", result.IP))
+		output.Outln(shell.GetEnvString(userShell, "PODMAN_IDENTITY_FILE", constants.GetPublicKeyPath()))
```
gbraad: Are yoy sure this should be the public key?

gbraad:
When using the correct key
```
PS C:\Users\gbraad\.crc\bin> $env:PODMAN_IDENTITY_FILE="C:\Users\gbraad\.crc\machines\crc\id_rsa"
```
more happens, but still fails:
```
PS C:\Users\gbraad\.crc\bin> .\podman.exe --log-level debug ps
time="2020-03-04T17:54:43+08:00" level=debug msg="unable to load configuration file at C:\\Users\\gbraad\\AppData\\podman\\podman-remote.conf"
time="2020-03-04T17:54:43+08:00" level=debug msg="creating a varlink bridge based on user input"
time="2020-03-04T09:54:42Z" level=debug msg="Using conmon: \"/usr/bin/conmon\""
time="2020-03-04T09:54:42Z" level=debug msg="Initializing boltdb state at /var/home/core/.local/share/containers/storage/libpod/bolt_state.db"
time="2020-03-04T09:54:42Z" level=debug msg="Using graph driver overlay"
time="2020-03-04T09:54:42Z" level=debug msg="Using graph root /var/home/core/.local/share/containers/storage"
time="2020-03-04T09:54:42Z" level=debug msg="Using run root /run/user/1000"
time="2020-03-04T09:54:42Z" level=debug msg="Using static dir /var/home/core/.local/share/containers/storage/libpod"
time="2020-03-04T09:54:42Z" level=debug msg="Using tmp dir /run/user/1000/libpod/tmp"
time="2020-03-04T09:54:42Z" level=debug msg="Using volume path /var/home/core/.local/share/containers/storage/volumes"
time="2020-03-04T09:54:42Z" level=debug msg="Set libpod namespace to \"\""
time="2020-03-04T09:54:42Z" level=debug msg="Not configuring container store"
time="2020-03-04T09:54:42Z" level=debug msg="Initializing event backend journald"
time="2020-03-04T09:54:42Z" level=debug msg="using runtime \"/usr/bin/runc\""
time="2020-03-04T17:54:44+08:00" level=error msg="unexpected EOF"
```

cmd/crc/cmd/podman_env.go
praveenkumar: It should be the private key instead public one.

cmd/crc/cmd/podman_env.go
zeenix: ok, will revert back then.

pkg/crc/constants/constants.go
zeenix: > Why are we using the msi ?

Copy&paste error. Will fix.

cmd/crc/cmd/podman_env.go
anjannath: Instead of this we could also create the config file for podman-remote. https://github.com/containers/libpod/blob/master/docs/source/markdown/podman-remote.conf.5.md

gbraad: if you check the issue itself we decided against this, as that would mess up exisitng and future connections. now they are scoped to the terminal session.

Issue #952, opened 2020-01-22
After initial `crc start -b xxx.crcbundle`, `crc start` without `-b` should work


cfergeau:
At the moment, it looks for `~/.crc/cache/xxx.crcbundle` and fail since it's missing,
which is not going to be a problem as
- the bundle is uncompressed in `~/.crc/cache/xxx`
- `~/.crc/machines/crc` already exists, so we don't need the data from `~/.crc/cache`

praveenkumar:
@cfergeau yes, I think we did discuss this, have to find out where was that discussion :)

zhangli19817:
$ crc start
ERRO file '/Users/luffy/.crc/cache/crc_hyperkit_4.2.13.crcbundle' does not exist 
ERRO Please provide the path to a valid bundle using the -b option 

Question:

where's the target crc_hyperkit_4.2.13.crcbundle?
on this mac platform, with flag -b, what's this parameter refer to?

cfergeau:
If you use a released version from `mirror.openshift.com`, `crc setup` will unpack the needed bundle to `~/.crc/cache`

PR #1003, opened 2020-02-12
Don't require bundle archive to exist if bundle is unpacked already

cfergeau:
All the work is done in the 2nd commit, "Issue #952: Don't require bundle archive if unpacked bundle exists"

This fixes https://github.com/code-ready/crc/issues/952


praveenkumar:
retest this please.

pkg/crc/validation/validation.go
```diff
@@ -30,19 +31,24 @@ func ValidateMemory(value int) error {
...
 
 // ValidateBundle checks if provided bundle path exist
-func ValidateBundle(bundle string) error {
-	if err := ValidatePath(bundle); err != nil {
-		if constants.BundleEmbedded() {
-			return errors.Newf("Run 'crc setup' to unpack the bundle to disk")
-		} else {
-			return errors.Newf("Please provide the path to a valid bundle using the -b option")
+func ValidateBundle(bundlePath string) error {
+	// check if this bundle version is cached already, in which case we don't
```
cfergeau: `validation.ValidateBundle()` is also used by `crc config set bundle`, and I don't think the newly added shortcut makes sense there, so this will need more work (or more testing)

praveenkumar:
retest this please

pkg/crc/validation/validation.go
cfergeau: I can confirm that this impacts `crc config set bundle` in a bad way:
```
$ ./crc.git config set bundle /fake/path/crc_libvirt_4.3.1.crcbundle
Successfully configured bundle to /fake/path/crc_libvirt_4.3.1.crcbundle

$ ls -al /fake/path/
ls: cannot access '/fake/path/': No such file or directory
```

cfergeau:
Didn't take time to revise this yet, https://github.com/code-ready/crc/pull/1003#discussion_r380157960 is still blocking its merge.

test/integration/features/config.feature
```diff
@@ -37,6 +37,18 @@ Checks whether CRC `config set` command works as expected in conjunction with `c
             | nameserver       | 120.0.0.1 |   999.999.999.999 |
             | pull-secret-file |    /Users | /nonexistent-file |
 
+   @linux
+   Scenario: Check bundle version validation (command line)
+       Given executing "touch crc_libvirt_99.9.9.crcbundle" succeeds
+       When executing "crc start --bundle crc_libvirt_99.9.9.crcbundle" fails
```
jsliacan, line 43: What happens here if binary has bundle embedded in it? Fails with the same message? Can you actually provide an external bundle to such a binary at all (even if correct version)?

pkg/crc/validation/validation.go
gbraad: There is no validation check for the bundle?

gbraad:
> Didn't take time to revise this yet, #1003 (comment) is still blocking its merge.

Did this happen now?

cfergeau:
Closing this as it's outdated and Guillaume's PR will eventually achieve the same purpose.

Issue #1006, opened 2020-02-13
[BUG] Use UBI as base image for dnsmasq


praveenkumar:
Now since UBI image repository added the dnsmasq it would make sense to use UBI instead of centos for dnsmasq image.

BZ :- https://bugzilla.redhat.com/show_bug.cgi?id=1766581

praveenkumar:
This is merged with #1007

PR #1007, opened 2020-02-13
Issue #1006 Use ubi instead of centos as base image

praveenkumar:
dnsmasq rpm now available on ubi-8-appstream repo, so better to use
ubi then centos base image.

- BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1766581


gbraad:
After some quick tests we can use this instead.

PR #1008, opened 2020-02-13
Improvements to the daemon

### LINKED ISSUE #949: [Improvement] Usage of go routine for different request to daemon
We need to  reconsider running the different request using go routine asynchronously, `start` ,`stop` and `delete` should be blocking calls which is not right now and we are using the tray UI to disable those action.

- https://github.com/code-ready/crc/pull/861/files#r365775688

### LINKED ISSUE #951: [Improvement] Send json response in case of failure to encode string
Right now, we are sending `Failed` as string instead the json response when encoding is failed.

- https://github.com/code-ready/crc/pull/861/files#r365786147

anjannath:
Fixes #949 
Fixes #951

### DIFF
--- a/pkg/crc/api/api.go
+++ b/pkg/crc/api/api.go
@@ -50,12 +50,13 @@ func (api CrcApiServer) handleClusterOperations() {
 
 func (api CrcApiServer) handleRequest(req commandRequest, conn net.Conn) {
 	defer conn.Close()
+	var result string
 	if handler, ok := api.handlers[req.Command]; ok {
-		result := handler(req.Args)
-		writeStringToSocket(conn, result)
+		result = handler(req.Args)
 	} else {
-		writeStringToSocket(conn, fmt.Sprintf("Unknown command supplied: %s", req.Command))
+		result = encodeErrorToJson(fmt.Sprintf("Unknown command supplied: %s", req.Command))
 	}
+	writeStringToSocket(conn, result)
 }
 
 func (api CrcApiServer) handleConnections(conn net.Conn) {
@@ -84,7 +85,8 @@ func (api CrcApiServer) handleConnections(conn net.Conn) {
 		if !addRequestToChannel(r, api.clusterOpsRequestsChan) {
 			defer conn.Close()
 			logging.Error("Channel capacity reached, unable to add new request")
-			writeStringToSocket(conn, "Channel capacity reached, unable to add new request")
+			errMsg := encodeErrorToJson("Sockets channel capacity reached, unable to add new request")
+			writeStringToSocket(conn, errMsg)
 			return
 		}
 	} else {
--- a/pkg/crc/api/api.go
+++ b/pkg/crc/api/api.go
@@ -9,27 +9,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
-type ArgsType map[string]string
-type handlerFunc func(ArgsType) string
-
-type CrcApiServer struct {
-	listener               net.Listener
-	clusterOpsRequestsChan chan clusterOpsRequest
-	handlers               map[string]handlerFunc // relates commands to handler func
-}
-
-// commandRequest struct is used to decode the json request from tray
-type commandRequest struct {
-	Command string            `json:"command"`
-	Args    map[string]string `json:"args,omitempty"`
-}
-
-// clusterOpsRequest struct is used to store the command request and associated socket
-type clusterOpsRequest struct {
-	command commandRequest
-	socket  net.Conn
-}
-
 func CreateApiServer(socketPath string) (CrcApiServer, error) {
 	listener, err := net.Listen("unix", socketPath)
 	if err != nil {
--- a/pkg/crc/api/api.go
+++ b/pkg/crc/api/api.go
@@ -13,14 +13,21 @@ type ArgsType map[string]string
 type handlerFunc func(ArgsType) string
 
 type CrcApiServer struct {
-	listener net.Listener
-	handlers map[string]handlerFunc // relates commands to handler func
+	listener               net.Listener
+	clusterOpsRequestsChan chan clusterOpsRequest
+	handlers               map[string]handlerFunc // relates commands to handler func
 }
 
 // commandRequest struct is used to decode the json request from tray
 type commandRequest struct {
 	Command string            `json:"command"`
 	Args    map[string]string `json:"args,omitempty"`
+}
+
+// clusterOpsRequest struct is used to store the command request and associated socket
+type clusterOpsRequest struct {
+	command commandRequest
+	socket  net.Conn
 }
 
 func CreateApiServer(socketPath string) (CrcApiServer, error) {
@@ -30,7 +37,8 @@ func CreateApiServer(socketPath string) (CrcApiServer, error) {
 		return CrcApiServer{}, err
 	}
 	apiServer := CrcApiServer{
-		listener: listener,
+		listener:               listener,
+		clusterOpsRequestsChan: make(chan clusterOpsRequest, 10),
 		handlers: map[string]handlerFunc{
 			"start":         startHandler,
 			"stop":          stopHandler,
@@ -44,18 +52,34 @@ func CreateApiServer(socketPath string) (CrcApiServer, error) {
 }
 
 func (api CrcApiServer) Serve() {
+	go api.handleClusterOperations() // go routine that handles start, stop and delete calls
 	for {
 		conn, err := api.listener.Accept()
 		if err != nil {
 			logging.Error("Error establishing communication: ", err.Error())
 			continue
 		}
-		go api.handleConnection(conn)
+		api.handleConnections(conn) // handle version, status, webconsole, etc. requests
 	}
 }
 
-func (api CrcApiServer) handleConnection(conn net.Conn) {
+func (api CrcApiServer) handleClusterOperations() {
+	for req := range api.clusterOpsRequestsChan {
+		api.handleRequest(req.command, req.socket)
+	}
+}
+
+func (api CrcApiServer) handleRequest(req commandRequest, conn net.Conn) {
 	defer conn.Close()
+	if handler, ok := api.handlers[req.Command]; ok {
+		result := handler(req.Args)
+		writeStringToSocket(conn, result)
+	} else {
+		writeStringToSocket(conn, fmt.Sprintf("Unknown command supplied: %s", req.Command))
+	}
+}
+
+func (api CrcApiServer) handleConnections(conn net.Conn) {
 	inBuffer := make([]byte, 1024)
 	var req commandRequest
 	numBytes, err := conn.Read(inBuffer)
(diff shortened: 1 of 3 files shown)


pkg/crc/api/api.go
```diff
@@ -69,12 +69,34 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
 		logging.Error("Error decoding request: ", err.Error())
 		return
 	}
-
-	if handler, ok := api.handlers[req.Command]; ok {
-		result := handler(req.Args)
-		writeStringToSocket(conn, result)
+	if req.Command == "start" || req.Command == "stop" || req.Command == "delete" {
+		// queue new request to channel
```
cfergeau: The commit log says
" Issue #949 Make cluster operation calls serve in order

This patch makes start, stop and delete calls serve in order
and one at a time"

Why are these special, and why do we treat all other requests without the channel? Which problem are we trying to solve? This needs to be in the commit log

pkg/crc/api/api.go
```diff
@@ -69,12 +69,34 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
...
+		}(requests)
 	} else {
-		writeStringToSocket(conn, fmt.Sprintf("Unknown command supplied: %s", req.Command))
+		go func() {
+			if handler, ok := api.handlers[req.Command]; ok {
+				result := handler(req.Args)
+				writeStringToSocket(conn, result)
+			} else {
+				writeStringToSocket(conn, fmt.Sprintf("Unknown command supplied: %s", req.Command))
+			}
```
cfergeau: Don't duplicate that code twice.

pkg/crc/api/handlers.go
```diff
@@ -65,6 +72,8 @@ func getPullSecretFileContent() (string, error) {
 }
 
 func deleteHandler(_ ArgsType) string {
+	mutex.Lock()
+	defer mutex.Unlock()
 	delConfig := machine.DeleteConfig{Name: constants.DefaultName}
 	r, _ := machine.Delete(delConfig)
```
cfergeau: Btw, it since you now have a way to report errors, you could handle the error returned by these commands.

anjannath: So the error inside `r` is handled on the tray side, for the daemon the errors are already logged in the machine package.

pkg/crc/api/api.go
```diff
@@ -68,8 +68,6 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
 			if handler, ok := api.handlers[req.Command]; ok {
 				result := handler(req.Args)
 				writeStringToSocket(conn, result)
-			} else {
-				writeStringToSocket(conn, fmt.Sprintf("Unknown command supplied: %s", req.Command))
```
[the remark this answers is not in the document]
anjannath: Since we are checking if the command is one of `start, stop, delete` we have a guarantee that Unknown command supplied is not going to happen...

anjannath:
@cfergeau I think i addressed all your review comments, PTALA, Thanks!

pkg/crc/api/api.go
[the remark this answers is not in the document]
cfergeau: ah this could be moved to `Issue #949 Make cluster operation calls serve in order ` I believe?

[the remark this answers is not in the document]
cfergeau: Actually rather than being removed, I'd just add a small `handleRequest` helper for
```
			if handler, ok := api.handlers[req.Command]; ok {
				result := handler(req.Args)
				writeStringToSocket(conn, result)
			} else {
				writeStringToSocket(conn, fmt.Sprintf("Unknown command supplied: %s", req.Command))
			}
```

pkg/crc/api/api.go
```diff
@@ -69,12 +69,34 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
...
-
-	if handler, ok := api.handlers[req.Command]; ok {
-		result := handler(req.Args)
-		writeStringToSocket(conn, result)
+	if req.Command == "start" || req.Command == "stop" || req.Command == "delete" {
+		// queue new request to channel
+		if !addRequestToChannel(req, requests) {
+			defer conn.Close()
+			logging.Error("Channel capacity reaced, unable to add new request")
+			writeStringToSocket(conn, "Channel capacity reaced, unable to add new request")
```
cfergeau: "reached" twice

pkg/crc/api/handlers.go
cfergeau: yeah, more of a design issue with the `machine` package than anything to do with your code :( We really should not hide errors inside results returned by `machine` methods :(

pkg/crc/api/api.go
```diff
@@ -69,12 +69,34 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
...
+			return
+		}
+		go func(requests <-chan commandRequest) {
+			req := <-requests
```
cfergeau: Since we spawn one goroutine each time we add an item to the channel, I don't think this will help serialize start/stop/delete requests, they will still potentially be run concurrently. I'm only talking about this goroutine/channel use, which maybe could be improved. I know you've also added mutexes

pkg/crc/api/handlers.go
```diff
@@ -13,13 +14,17 @@ import (
...
 func statusHandler(_ ArgsType) string {
 	statusConfig := machine.ClusterStatusConfig{Name: constants.DefaultName}
 	clusterStatus, _ := machine.Status(statusConfig)
 	return encodeStructToJson(clusterStatus)
 }
 
 func stopHandler(_ ArgsType) string {
+	mutex.Lock()
+	defer mutex.Unlock()
 	stopConfig := machine.StopConfig{
```
cfergeau: at this point, I think I'd prefer that we achieve the serialization at the api.go level rather than adding mutexes to the `machine` package.

anjannath:
> With that said, it seems the worst that can happen is bad error messages (which may be handled in the tray already to show something better?)

Yes, if we get an error from the daemon that is handled in tray to show the status as unknown.

> We should also document what we consider is acceptable to call concurrently, and what is not to make it easier to decide between "this is a bug we need to fix" and "this is not supported"
Which info do you need to get in your regular "status" calls?

For now we show the status in a window, so all the info that is part of `crc status` command is needed, later when we poll regularly only the VM status would be enough to change the color of the tray icon to indicate that a VM is running/stopped.

pkg/crc/api/api.go
anjannath: Yes, so the channel is used as a queue to add the incoming requests, this makes sure the requests are served in order, and the mutex lets only one of start/stop/delete handler to run at a time.

pkg/crc/api/api.go
anjannath: Ahh, this i fixed in the last commit a9056c9bea085dcf490b7915a505b646a7703fb7

pkg/crc/api/api.go
cfergeau: I don't think this is working as intended.
Let's say you receive 2 requests, A and B.
First request gets added to the queue, a go routine is spawned
Second request gets added to the queue, a second go routine is spawned

One of the go routine will read from the channel and get request A
The other go routine will read from the channel and get request B
However, I don't think we have anything ensuring which go routine will call its handler first, they will be racing with each other.

gbraad: @anjannath Can you verify this? Could just a mutex be enough?

anjannath: @gbraad I did look into this and also had a chat with teuf, what he explains in the comment does happen and the requests gets served out of order, he suggested to use a single go routine for the three cluster operations (start,stop,delete) in which case we don't need the mutex. Just pushed an update!

anjannath:
@cfergeau Updated the PR, PTALA

pkg/crc/api/api.go
```diff
@@ -69,12 +90,22 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
...
 		return
 	}
-
-	if handler, ok := api.handlers[req.Command]; ok {
-		result := handler(req.Args)
-		writeStringToSocket(conn, result)
+	if req.Command == "start" || req.Command == "stop" || req.Command == "delete" {
+		// queue new request to channel
+		if !addRequestToChannel(req, api.clusterOpsRequestsChan) {
+			defer conn.Close()
```
cfergeau: better to use a new struct holding both the request and the connection, and to add it to a single channel. Having 2 channels which must have exactly the same number of items could break easily if we don't pay attention.

pkg/crc/api/api.go
```diff
@@ -44,18 +48,35 @@ func CreateApiServer(socketPath string) (CrcApiServer, error) {
...
 		}
-		go api.handleConnection(conn)
+		api.handleConnections(conn) // handle version, status, webconsole, etc. requests
 	}
 }
 
-func (api CrcApiServer) handleConnection(conn net.Conn) {
-	defer conn.Close()
+func (api CrcApiServer) handleClusterOperations() {
+	for req := range api.clusterOpsRequestsChan {
```
cfergeau: https://golangr.com/range-channel/ says that buffered channels can only be iterated over after being closed?
You could also go with a `for {}` loop

anjannath: Not sure, this is working as is, maybe this changed now? if the channel is closed the range will throw an error and the go routine dies, anyways we don't want the channel to be closed.

pkg/crc/api/api.go
anjannath: Yes, i was also considering this.. thanks!

pkg/crc/api/api.go
```diff
@@ -69,12 +90,22 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
...
+		}
+		if !addSocketsToChannel(conn, api.socketsChan) {
+			defer conn.Close()
+			logging.Error("Channel capacity reaced, unable to add new request")
+			writeStringToSocket(conn, "Channel capacity reaced, unable to add new request")
+			return
+		}
 	} else {
-		writeStringToSocket(conn, fmt.Sprintf("Unknown command supplied: %s", req.Command))
+		go api.handleRequest(req, conn)
```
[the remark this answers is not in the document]
anjannath: Yes this has to be run async, since we need to handle the other requests as well.

> The bigger question is whether they can be run concurrently with start/stop/.. operations.

Yes, there shouldn't be any issues running the status, version, and webconsoleurl command concurrently, in case they fail we handle the error. and it shouldn't kill the daemon process.

Also to implement polling to get the status update periodically and since a GUI is involved we need to offer these to be run even when a time consuming background process is on going in the background.

pkg/crc/api/api.go
anjannath: done..

pkg/crc/api/api.go
```diff
@@ -44,18 +31,35 @@ func CreateApiServer(socketPath string) (CrcApiServer, error) {
...
 func (api CrcApiServer) Serve() {
+	go api.handleClusterOperations() // go routine that handles start, stop and delete calls
 	for {
 		conn, err := api.listener.Accept()
 		if err != nil {
 			logging.Error("Error establishing communication: ", err.Error())
 			continue
 		}
-		go api.handleConnection(conn)
+		api.handleConnections(conn) // handle version, status, webconsole, etc. requests
```
gbraad: what is the motivation behind removing the goroutine?

anjannath: Now the go routine is called from `api.handleConnections` to make the cluster operations `start/stop/delete` be served in a different go routine.

pkg/crc/api/api.go
cfergeau: Or maybe golangr.com is wrong. From other sites/my testing, closing the channel is needed to stop the loop, but before the channel is closed, this will loop forever.

anjannath: Yes, and will block on the channel read when its empty..

pkg/crc/api/api.go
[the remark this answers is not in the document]
cfergeau: I'm still *very* uncomfortable running concurrently code which was not written for that, but I can't spot potential problems either so..

pkg/crc/api/api.go
```diff
@@ -69,12 +93,20 @@ func (api CrcApiServer) handleConnection(conn net.Conn) {
 		logging.Error("Error decoding request: ", err.Error())
 		return
 	}
-
-	if handler, ok := api.handlers[req.Command]; ok {
-		result := handler(req.Args)
-		writeStringToSocket(conn, result)
+	if req.Command == "start" || req.Command == "stop" || req.Command == "delete" {
+		// queue new request to channel
```
cfergeau: I believe a bit more context why they are treated differently from other requests would be useful here.
(they are slow operations, and change the VM state so they have to run sequentially. We don't want other operations querying the status of the VM to be blocked by these, so they are treated by a dedicated go routine)

anjannath: You mean in the commit log or here as a comment?

cfergeau: As a comment, as for someone reading the code, it's not going to be obvious at all why we special case commands.

anjannath: Done!

PR #1011, opened 2020-02-14
Issue #1009 Create ~/Library/LaunchAgents dir during setup

### LINKED ISSUE #1009: crc setup fails to install the tray in some versions of macOS
In some versions of macOS the `~/Library/LaunchAgents` dir is not present by default, we can check for the existence of this dir during setup and create it if it doesn't exists instead of failing.

See https://github.com/code-ready/tray-macos/issues/10#issuecomment-579763400

anjannath:
This PR fixes the following things:
1. In some macOS version `~/Library/LaunchAgents` dir is not present by default, we are creating it now as part of crc setup.
2. When a cached tray binary is present in ~/.crc/bin`, crc setup doesn't download/install the tray binary even if `checkTrayExistsAndRunning()` fails. we are overwriting the old tray now whenever the check function fails.
3. Check if the version of the tray is the expected one in `checkTrayExistsAndRunning()`

Fixes #1009


pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -103,11 +103,9 @@ func fixTrayExistsAndRunning() error {
 		return err
 	}
 	// get the tray app
-	if !trayAppCached() {
-		err := downloadOrExtractTrayApp()
-		if err != nil {
-			return err
-		}
+	err := downloadOrExtractTrayApp()
```
cfergeau: This is added to `fixTrayExistsAndRunning()`, which only at `crc setup` time if `checkTrayExistsAndRunning` returned an error, so when is this expected to run exactly? Or does this assume the system tray has been stopped manually?
It would be better if we had a way of checking the version of the systray, and if the version is older than what the binary can install, then we upgrade. This is how it is done for the machine drivers.

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -99,6 +99,9 @@ func checkTrayExistsAndRunning() error {
 }
 
 func fixTrayExistsAndRunning() error {
+	if err := ensureLaunchAgentsDirExists(); err != nil {
+		return err
+	}
 	// get the tray app
```
cfergeau: This could go in the `!trayAppCached` block.

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -2,6 +2,7 @@ package preflight
 
 import (
 	"bytes"
+
```
cfergeau: I don't think this blank link is needed ;)

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -232,3 +244,38 @@ func ensureLaunchAgentsDirExists() error {
...
+}
+
+func checkTrayVersion() bool {
+	v, err := getTrayVersion(constants.TrayBinaryPath)
+	if err != nil {
+		logging.Error(err.Error())
+	}
+	currentVersion, err := semver.NewVersion(v)
+	if err != nil {
+		logging.Error(err.Error())
```
cfergeau: why no `return false` in this case?

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -232,3 +244,38 @@ func ensureLaunchAgentsDirExists() error {
...
+	}
+	expectedVersion, err := semver.NewVersion(version.GetCRCTrayVersion())
+	if err != nil {
+		logging.Error(err.Error())
```
cfergeau: same here

pkg/crc/preflight/preflight_check_tray_darwin.go
cfergeau: Now that the preflight check will fail if the tray is outdated, how does upgrade work exactly when the tray is running already? we'll install the new binary but won't restart the tray so that it's used, will we?

anjannath: @cfergeau Added a restart at the end of the fix function, thanks!

pkg/crc/preflight/preflight_check_tray_darwin.go
anjannath: done!

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -151,6 +162,16 @@ func fixTrayExistsAndRunning() error {
...
+
+	// Restart daemon and tray agents
+	err = restartAgent(trayAgentLabel)
+	if err != nil {
+		return err
+	}
+	err = restartAgent(daemonAgentLabel)
+	if err != nil {
+		return err
+	}
```
cfergeau: I don't think this belongs in this commit, but rather in the one always installing the binary when the preflight fails.

anjannath:
The Travis CI failure is because vendor dir is not updated with the new package.

Issue #1016, opened 2020-02-17
[BUG] kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)


drpaneas:
### General information

  * OS: Fedora 31 Workstation Edition
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? Yes

## CRC version
```bash
crc version: 1.5.0+e1c8fb8
OpenShift version: 4.2.14 (embedded in binary)
```
  
## CRC status
```bash
CRC VM:          Stopped
OpenShift:       Stopped
Disk Usage:      0B of 0B (Inside the CRC VM)
Cache Usage:     11.98GB
Cache Directory: /home/drpaneas/.crc/cache
```

## CRC config
```bash
# Nothing
```

## Host Operating System
```bash
NAME=Fedora
VERSION="31 (Workstation Edition)"
ID=fedora
VERSION_ID=31
VERSION_CODENAME=""
PLATFORM_ID="platform:f31"
PRETTY_NAME="Fedora 31 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:31"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f31/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=31
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=31
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

### Steps to reproduce

  1. crc setup
  2. crc start

### Expected
To work without `ERRO`

### Actual
`ERRO Failed to connect to the CRC VM with SSH`

### Issue

I couldn't connect to the machine, simply because the machine (VM) was having a kernel panic. So it make sense to not being able to ssh into it. However, I would expect maybe a better `log` message, e.g. trying to `ping` first, rather than `ssh` first. Anyways, here's the kernel panic from inside the VM:

---
[...]

PR #1026, opened 2020-02-19
Host might be unreachable

drpaneas:
In case SSH fails this might also be caused due to a booting
problem of the VM (see #1016 - kernel panic - not syncing: VFS)

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -240,7 +240,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	logging.Debug("Waiting until ssh is available")
 	if err := cluster.WaitForSsh(sshRunner); err != nil {
 		result.Error = err.Error()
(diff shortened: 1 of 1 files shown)


gbraad:
Add to whitelist

gbraad:
@robin-owen @praveenkumar WDYT?

pkg/crc/machine/machine.go
```diff
@@ -240,7 +240,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	logging.Debug("Waiting until ssh is available")
 	if err := cluster.WaitForSsh(sshRunner); err != nil {
 		result.Error = err.Error()
-		return *result, errors.New("Failed to connect to the CRC VM with SSH")
+		return *result, errors.New("Failed to connect to the CRC VM with SSH -- host might be unreachable")
```
cfergeau: Fwiw, "host" can be confusing here, the host is the machine running the VM, while it's the VM which is unreachable.

Issue #1022, opened 2020-02-18
Split tray setup checks into individual checks


anjannath:
Right now different checks related to the tray are invoked as one unit, this issue is to decompose those into more fine grained individual checks like all other preflight checks/setups.

https://github.com/code-ready/crc/blob/8674e345a7f843aa01803fc261a2d7b6539bec37/pkg/crc/preflight/preflight_check_tray_darwin.go#L81-L99

gbraad:
... and maybe a separate file / subset also ?

anjannath:
Fived in #1029

PR #1029, opened 2020-02-20
Issue #1022 Split tray setup logic into smaller checks

anjannath:
Its mostly the existing code, just split into individual
functions, and accordingly Preflight structs are created
for each check function

The checks performed:
1. Check if tray binary is present
2. Check if tray/daemon plist is present (if not fix would create plist and load using `launchctl load`)
3. Check if tray version is matching (if not the fix would download/extract the tray and restart the tray)
4. Check if both tray/daemon running.

### DIFF
--- a/pkg/crc/constants/constants_darwin.go
+++ b/pkg/crc/constants/constants_darwin.go
@@ -8,5 +8,6 @@ const (
 )
 
 var (
-	TrayBinaryPath = filepath.Join(CrcBinDir, TrayBinaryName)
+	TrayAppBundlePath = filepath.Join(CrcBinDir, TrayBinaryName)
+	TrayBinaryPath    = filepath.Join(TrayAppBundlePath, "Contents", "MacOS", "CodeReady Containers")
 )
--- a/pkg/crc/preflight/preflight_check_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_check_tray_darwin.go
@@ -85,39 +85,14 @@ type TrayVersion struct {
 	ShortVersion string `plist:"CFBundleShortVersionString"`
 }
 
-func checkTrayExistsAndRunning() error {
-	logging.Debug("Checking if daemon plist file exists")
+func checkIfDaemonPlistFileExists() error {
 	if !os.FileExists(daemonPlistFilePath) {
-		return errors.New("Daemon plist file does not exist")
-	}
-	logging.Debug("Checking if crc agent running")
-	if !agentRunning(daemonAgentLabel) {
-		return errors.New("crc daemon is not running")
-	}
-	logging.Debug("Checking if tray plist file exists")
-	if !os.FileExists(trayPlistFilePath) {
-		return errors.New("Tray plist file does not exist")
-	}
-	logging.Debug("Checking if tray agent running")
-	if !agentRunning(trayAgentLabel) {
-		return errors.New("Tray is not running")
-	}
-	logging.Debug("Check if correct version of tray exists")
-	if !checkTrayVersion() {
-		return errors.New("cached tray version is older then expected")
+		return fmt.Errorf("Daemon plist file does not exist")
 	}
 	return nil
 }
 
-func fixTrayExistsAndRunning() error {
-	if err := ensureLaunchAgentsDirExists(); err != nil {
-		return err
-	}
-	// get the tray app
-	err := downloadOrExtractTrayApp()
-	if err != nil {
-		return err
-	}
+func fixDaemonPlistFileExists() error {
 	currentExecutablePath, err := goos.Executable()
 	if err != nil {
 		return err
@@ -126,52 +101,87 @@ func fixTrayExistsAndRunning() error {
 		BinaryPath:     currentExecutablePath,
 		StdOutFilePath: stdOutFilePathDaemon,
 	}
+	return fixPlistFileExists(daemonPlistTemplate, daemonConfig, daemonPlistFilePath)
+}
 
-	trayConfig := AgentConfig{
-		BinaryPath:     filepath.Join(constants.CrcBinDir, constants.TrayBinaryName, "Contents", "MacOS", "CodeReady Containers"),
-		StdOutFilePath: stdOutFilePathTray,
-	}
-	logging.Debug("Creating daemon plist")
-	err = createPlist(daemonPlistTemplate, daemonConfig, daemonPlistFilePath)
-	if err != nil {
-		return err
-	}
-	logging.Debug("Creating tray plist")
-	err = createPlist(trayPlistTemplate, trayConfig, trayPlistFilePath)
-	if err != nil {
-		return err
-	}
-
-	// load crc daemon
-	err = launchctlLoadPlist(daemonPlistFilePath)
-	if err != nil {
-		return err
-	}
-	if !agentRunning(daemonAgentLabel) {
-		if err = startAgent(daemonAgentLabel); err != nil {
-			return err
-		}
-	}
-	// load tray
-	err = launchctlLoadPlist(trayPlistFilePath)
-	if err != nil {
-		return err
-	}
-	if !agentRunning(trayAgentLabel) {
-		if err = startAgent(trayAgentLabel); err != nil {
-			return err
-		}
-	}
-	// restart tray and daemon agents
-	err = restartAgent(daemonAgentLabel)
-	if err != nil {
-		return err
-	}
-	err = restartAgent(trayAgentLabel)
-	if err != nil {
-		return err
+func checkIfTrayPlistFileExists() error {
+	if !os.FileExists(trayPlistFilePath) {
+		return fmt.Errorf("Tray plist file does not exist")
 	}
 	return nil
+}
+
+func fixTrayPlistFileExists() error {
+	trayConfig := AgentConfig{
+		BinaryPath:     constants.TrayBinaryPath,
+		StdOutFilePath: stdOutFilePathTray,
+	}
+	return fixPlistFileExists(trayPlistTemplate, trayConfig, trayPlistFilePath)
+}
+
+func checkIfDaemonAgentRunning() error {
+	if !agentRunning(daemonAgentLabel) {
+		return fmt.Errorf("crc daemon is not running")
+	}
+	return nil
+}
+
+func fixDaemonAgentRunning() error {
+	return startAgent(daemonAgentLabel)
+}
+
+func checkIfTrayAgentRunning() error {
+	if !agentRunning(trayAgentLabel) {
+		return fmt.Errorf("Tray is not running")
+	}
+	return nil
+}
+
+func fixTrayAgentRunning() error {
+	return startAgent(trayAgentLabel)
+}
+
+func checkTrayVersion() error {
+	v, err := getTrayVersion(constants.TrayAppBundlePath)
+	if err != nil {
+		logging.Error(err.Error())
+		return err
+	}
+	currentVersion, err := semver.NewVersion(v)
+	if err != nil {
+		logging.Error(err.Error())
+		return err
+	}
+	expectedVersion, err := semver.NewVersion(version.GetCRCTrayVersion())
+	if err != nil {
+		logging.Error(err.Error())
+		return err
+	}
+
+	if expectedVersion.GreaterThan(currentVersion) {
+		return fmt.Errorf("Cached version is older then latest version: %s < %s", currentVersion.String(), expectedVersion.String())
+	}
+	return nil
+}
+
+func fixTrayVersion() error {
+	// get the tray app
+	err := downloadOrExtractTrayApp()
+	if err != nil {
+		return err
+	}
+	return restartAgent(trayAgentLabel)
+}
+
+func checkTrayBinaryPresent() error {
+	if !os.FileExists(constants.TrayBinaryPath) {
+		return fmt.Errorf("Tray binary does not exist")
+	}
+	return nil
+}
+
+func fixTrayBinaryPresent() error {
+	return downloadOrExtractTrayApp()
 }
 
 func createPlist(template string, config AgentConfig, plistPath string) error {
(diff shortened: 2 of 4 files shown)


pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -127,48 +105,111 @@ func fixTrayExistsAndRunning() error {
...
-	err = launchctlLoadPlist(trayPlistFilePath)
-	if err != nil {
+	return nil
+}
+
+func fixDaemonAgentRunning() error {
+	if err := startAgent(daemonAgentLabel); err != nil {
 		return err
 	}
+	return nil
```
cfergeau: `return startAgent(...)`

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -127,48 +105,111 @@ func fixTrayExistsAndRunning() error {
...
+		return fmt.Errorf("Tray is not running")
+	}
+	return nil
+}
+
+func fixTrayAgentRunning() error {
+	if err := startAgent(trayAgentLabel); err != nil {
+		return err
+	}
+	return nil
```
cfergeau: `return startAgent(...)`

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -127,48 +105,111 @@ func fixTrayExistsAndRunning() error {
...
 	err = createPlist(daemonPlistTemplate, daemonConfig, daemonPlistFilePath)
 	if err != nil {
 		return err
 	}
+	// load plist and restart agent
+	if err := launchctlLoadPlist(daemonPlistFilePath); err != nil {
+		return err
+	}
+
+	if err := restartAgent(daemonAgentLabel); err != nil {
```
cfergeau: I'm not sure `launchtlLoadPlist` and `restartAgent` should be done here. We created the plist because it did not exist, our job is done. Another preflight check will load/start the agent if it's needed. If the plist was missing initially, there is nothing to *re*start anyway, is there?

pkg/launchd/launchd.go
```diff
@@ -0,0 +1,95 @@
...
+
+// LoadPlist loads a launchd agents' plist file
+func LoadPlist(plistFilePath string) error {
+	_, err := exec.Command("launchctl", "load", plistFilePath).Output() // #nosec G204
+	return err
+}
+
+// StartAgent starts a launchd agent
+func StartAgent(label string) error {
+	_, err := exec.Command("launchctl", "start", label).Output() // #nosec G204
```
cfergeau: One additional comment, launchctl man page says:

> Start the specified job by label. The expected use of this sub-command is for debugging and testing so that one can manually kick-start an on-demand server.

so maybe it's not really useful to call this?

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -127,48 +105,111 @@ func fixTrayExistsAndRunning() error {
...
+
+func checkIfTrayPlistFileExists() error {
+	if !os.FileExists(trayPlistFilePath) {
+		return fmt.Errorf("Tray plist file does not exist")
+	}
+	return nil
+}
+
+func fixTrayPlistFileExists() error {
+	if err := ensureLaunchAgentsDirExists(); err != nil {
```
cfergeau: This is very similar to `fixDaemonPlistFileExists`, for the longer methods in this file, it would be good to try to factor the common code in helper functions.

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -39,12 +39,42 @@ var dnsPreflightChecks = [...]PreflightCheck{
...
-	fix:              fixTrayExistsAndRunning,
-	flags:            SetupOnly,
+var traySetupChecks = [...]PreflightCheck{
+	{
+		checkDescription: "Checking installed tray version",
+		check:            checkTrayVersion,
+		fixDescription:   "Installing and setting up tray app",
+		fix:              fixTrayVersion,
+		flags:            SetupOnly,
+	},
```
cfergeau: `fixTrayVersion` expects the tray to be installed and running since it has a call to `RestartAgent`. On a system where the tray was never installed, I don't think this is going to work as expected. We may need a test checking if the tray binary has been unpacked and is available, then the check that the tray is installed and running, and then the version check.
Not sure how to best approach this, as I think some other corner cases are going to be odd :-/

pkg/launchd/launchd.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(plistPath, plistContent.Bytes(), 0644)
+	return err
+}
+
+// LoadPlist loads a launchd agents' plist file
+func LoadPlist(plistFilePath string) error {
+	_, err := exec.Command("launchctl", "load", plistFilePath).Output() // #nosec G204
```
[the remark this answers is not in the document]
anjannath: It would certainly be better, but not very convenient yes, keybase actually has a nice launchd wrapper library which i later found out, but we don't need much of the functionality there..

pkg/launchd/launchd.go
```diff
@@ -0,0 +1,95 @@
...
+	_, err := exec.Command("launchctl", "start", label).Output() // #nosec G204
+	return err
+}
+
+// RestartAgent restarts a launchd agent
+func RestartAgent(label string) error {
+	_, err := exec.Command("launchctl", "stop", label).Output() // #nosec G204
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
gbraad: but is this used? Unneeded/unused code is not helpful.
We should only introduce code that is actually used by us.

pkg/crc/preflight/preflight_darwin.go
gbraad: a defensive strategy is not bad

we can expect:

   * there is no tray
   * there is a tray
      * it is not running
      * it is running
         * you are having the right permissions to stop/start

pkg/crc/preflight/preflight_check_tray_darwin.go
anjannath: There's really not much common code here, and already using the `createPlist` helper.. we can have an helper to generate the struct for the plist creation..

pkg/crc/preflight/preflight_darwin.go
anjannath: > fixTrayVersion expects the tray to be installed and running since it has a call to RestartAgent. On a system where the tray was never installed, I don't think this is going to work as expected.

No, it just expects the tray to be installed, it doesn't have to be running. Installed for the tray means two things -- the tray binary is present and the launchd plist file exists.

pkg/crc/preflight/preflight_check_tray_darwin.go
anjannath: Yes, we only need to restart in case of an upgrade, so this is only needed in `fixTrayVersion`

pkg/crc/preflight/preflight_darwin.go
anjannath: I believe the following steps should handle both a new install and upgrade of the tray:
1. Check if tray binary is present
2. Check if tray/daemon plist is present (if not fix would create plist and load using `launchctl load`)
3. Check if tray version is matching (if not the fix would download/extract the tray and restart the tray)
4. Check if both tray/daemon running.

cfergeau: > No, it just expects the tray to be installed, it doesn't have to be running.

But `checkTrayVersion/fixTrayVersion` is the very first check in `traySetupChecks`, at this point you don't know if the tray is installed or not.

cfergeau: > I believe the following steps should handle both a new install and upgrade of the tray:

Seems fine.

pkg/crc/preflight/preflight_check_tray_darwin.go
cfergeau:
```
diff --git a/pkg/crc/preflight/preflight_check_tray_darwin.go b/pkg/crc/preflight/preflight_check_tray_darwin.go
index c91e232f..4e727cc0 100644
--- a/pkg/crc/preflight/preflight_check_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_check_tray_darwin.go
@@ -45,9 +45,6 @@ func checkIfDaemonPlistFileExists() error {
 }
 
 func fixDaemonPlistFileExists() error {
-	if err := ensureLaunchAgentsDirExists(); err != nil {
-		return err
-	}
 	currentExecutablePath, err := goos.Executable()
 	if err != nil {
 		return err
@@ -59,23 +56,7 @@ func fixDaemonPlistFileExists() error {
 		Args:           []string{"daemon", "--log-level", "debug"},
 	}
 
-	logging.Debug("Creating daemon plist")
-	err = launchd.CreatePlist(daemonConfig, daemonPlistFilePath)
-	if err != nil {
-		logging.Debug("failed while creating plist:", err.Error())
-		return err
-	}
-	// load plist and restart agent
-	if err := launchd.LoadPlist(daemonPlistFilePath); err != nil {
-		logging.Debug("Failed while loading plist:", err.Error())
-		return err
-	}
-
-	if err := launchd.RestartAgent(daemonAgentLabel); err != nil {
-		logging.Debug("Failed while restarting daemon:", err.Error())
-		return err
-	}
-	return nil
+	return fixPlistFileExists(daemonConfig, daemonPlistFilePath)
 }
 
 func checkIfTrayPlistFileExists() error {
@@ -85,29 +66,35 @@ func checkIfTrayPlistFileExists() error {
 	return nil
 }
 
+func fixPlistFileExists(agentConfig launchd.AgentConfig, plistFilePath string) error {
+	if err := ensureLaunchAgentsDirExists(); err != nil {
+		return err
+	}
+	logging.Debugf("Creating plist for %s", agentConfig.Label)
+	err := launchd.CreatePlist(agentConfig, plistFilePath)
+	if err != nil {
+		return err
+	}
+	// load plist and restart agent
+	if err := launchd.LoadPlist(plistFilePath); err != nil {
+		logging.Debug("failed while creating plist:", err.Error())
+		return err
+	}
+
+	if err := launchd.RestartAgent(agentConfig.Label); err != nil {
+		logging.Debug("Failed while loading plist:", err.Error())
+		return err
+	}
+	return nil
+}
 func fixTrayPlistFileExists() error {
-	if err := ensureLaunchAgentsDirExists(); err != nil {
-		return err
-	}
 	trayConfig := launchd.AgentConfig{
 		Label:          trayAgentLabel,
 		BinaryPath:     filepath.Join(constants.CrcBinDir, constants.TrayBinaryName, "Contents", "MacOS", "CodeReady Containers"),
 		StdOutFilePath: stdOutFilePathTray,
 	}
-	logging.Debug("Creating tray plist")
-	err := launchd.CreatePlist(trayConfig, trayPlistFilePath)
-	if err != nil {
-		return err
-	}
-	// load plist and restart agent
-	if err := launchd.LoadPlist(trayPlistFilePath); err != nil {
-		return err
-	}
 
-	if err := launchd.RestartAgent(trayAgentLabel); err != nil {
-		return err
-	}
-	return nil
+	return fixPlistFileExists(trayConfig, trayPlistFilePath)
 }
 
 func checkIfDaemonAgentRunning() error {
```

pkg/launchd/launchd.go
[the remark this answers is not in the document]
cfergeau: Yes

pkg/crc/preflight/preflight_darwin.go
anjannath: > But checkTrayVersion/fixTrayVersion is the very first check in traySetupChecks, at this point you don't know if the tray is installed or not.

Ahh, right my bad..

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -100,22 +52,27 @@ func fixDaemonPlistFileExists() error {
...
+	daemonConfig := launchd.AgentConfig{
+		Label:          daemonAgentLabel,
 		BinaryPath:     currentExecutablePath,
 		StdOutFilePath: stdOutFilePathDaemon,
+		Args:           []string{"daemon", "--log-level", "debug"},
 	}
 
 	logging.Debug("Creating daemon plist")
-	err = createPlist(daemonPlistTemplate, daemonConfig, daemonPlistFilePath)
+	err = launchd.CreatePlist(daemonConfig, daemonPlistFilePath)
```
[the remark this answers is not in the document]
anjannath: They can be formed from the labels, but keeping them as constants just reads better i think..

anjannath:
@cfergeau I added a new commit for the `fixPlistFile` changes you suggested, couldn't put it in the first commit :(

pkg/launchd/launchd.go
anjannath: So in case of when we update the tray binary, we forcefully want to start the tray so that the new version is ran..

cfergeau:
> @cfergeau I added a new commit for the `fixPlistFile` changes you suggested, couldn't put it in the first commit :(

If this was because of merge conflicts, https://github.com/cfergeau/crc/commits/launchd should adress this.

pkg/crc/preflight/preflight_check_tray_darwin.go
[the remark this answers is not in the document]
cfergeau: `preflight_check_tray_darwin.go` does not need to know that plist files go to `$HOME/Library/LaunchAgents`. That's an implementation detail of the launch package. Once the path is hidden from `preflight_check_tray_darwin.go`, we can remove plist names from the public interface since they are easy to autogenerate.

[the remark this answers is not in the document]
cfergeau: see https://github.com/cfergeau/crc/commit/db96e3c72e4a279a7a184925f44a576a78210dbc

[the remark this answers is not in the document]
cfergeau: > see [cfergeau@db96e3c](https://github.com/cfergeau/crc/commit/db96e3c72e4a279a7a184925f44a576a78210dbc)

Which probably needs a bit more work to use the `daemonAgentLabel` `trayAgentLabel` constants rather than hardcoding these :)

pkg/crc/constants/constants_darwin.go
```diff
@@ -8,5 +8,6 @@ const (
 )
 
 var (
-	TrayBinaryPath = filepath.Join(CrcBinDir, TrayBinaryName)
+	TrayAppBundlePath = filepath.Join(CrcBinDir, TrayBinaryName)
+	TrayBinaryPath    = filepath.Join(TrayAppBundlePath, "Contents", "MacOS", "CodeReady Containers")
```
cfergeau: `TrayBinaryPath` is not being used at the moment, seems like an oversight.

pkg/launchd/launchd.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// StartAgent starts a launchd agent
+func StartAgent(label string) error {
+	return exec.Command("launchctl", "start", label).Run() // #nosec G204
+}
+
+// StopAgent stops a launchd agent
+func StopAgent(label string) error {
+	return exec.Command("launchctl", "stop", label).Run()
```
cfergeau: It's missing `// #nosec G204`

pkg/crc/preflight/preflight_check_tray_darwin.go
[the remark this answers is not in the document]
anjannath: > >    see cfergeau@db96e3c

> Which probably needs a bit more work to use the daemonAgentLabel trayAgentLabel constants rather than hardcoding these :)

Looks like it needs only small change, i can cherry pick your commit and do that work on top, Thanks!

pkg/crc/constants/constants_darwin.go
anjannath: This is needed in the trayPlistConfig, but am again forming this path there, should use the constant.

pkg/crc/preflight/preflight_darwin.go
gbraad: @anjannath resolved or will be resolved in follow-up?

anjannath: @gbraad Its resolved, i made the changes

gbraad: I see you have `checkTrayBinaryPresent`

anjannath:
@cfergeau PTALA

cfergeau:
`Refactor fixTrayPlistFile and daemonPlistFile functions` can be squashed in `Issue #1022 Split tray setup logic into smaller checks` if you think that makes more sense as a single commit.

Issue #1015, opened 2020-02-17
/var/lib/kubelet/config.json created with wrong permissions


cfergeau:
Noticed this when checking machine-config-operator/machine-config-daemon logs:

```
E0217 15:09:48.620882    4622 writer.go:127] Marking Degraded due to: unexpected on-disk state validating against rendered-master-8c392e87b6032fdfb6f8a3ffb4aa6fee                                                
```

cfergeau:
I copied the wrong log line, the correct one was complaining about `/var/lib/kubelet/config.json` permissions being `0664` while `0600` is expected

cfergeau:
```
E0217 20:22:13.809612    4901 daemon.go:1341] mode mismatch for file: "/var/lib/kubelet/config.json"; expected: -rw-------; received: -rw-r--r--
```

PR #1037, opened 2020-02-25
Issue #1015: Set /var/lib/kubelet/config.json permissions to 0600

cfergeau:
oc logs  -n openshift-machine-config-operator
machine-config-daemon-5t5sj -c machine-config-daemon

E0217 20:22:13.809612    4901 daemon.go:1341] mode mismatch for file:
"/var/lib/kubelet/config.json"; expected: -rw-------; received:
-rw-r--r--

This fixes https://github.com/code-ready/crc/issues/1015

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -196,10 +196,11 @@ func AddProxyConfigToMarketplaceOperator(oc oc.OcConfig, proxy *network.ProxyCon
 }
 
 func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error {
-	err := sshRunner.SetTextContentAsRoot("/var/lib/kubelet/config.json", pullSec, 0644)
+	err := sshRunner.SetTextContentAsRoot("/var/lib/kubelet/config.json", pullSec, 0600)
 	if err != nil {
 		return err
 	}
+
 	return nil
 }
 
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -167,11 +167,11 @@ Environment=HTTPS_PROXY=%s
 Environment=NO_PROXY=%s,%s`
 	p := fmt.Sprintf(proxyTemplate, proxy.HttpProxy, proxy.HttpsProxy, internalNoProxy, proxy.GetNoProxyString())
 	// This will create a systemd drop-in configuration for proxy (both for kubelet and crio services) on the VM.
-	_, err := sshRunner.RunPrivate(fmt.Sprintf("cat <<EOF | sudo tee /etc/systemd/system/crio.service.d/10-default-env.conf\n%s\nEOF", p))
+	err := sshRunner.SetTextContentAsRoot("/etc/systemd/system/crio.service.d/10-default-env.conf", p, 0644)
 	if err != nil {
 		return err
 	}
-	_, err = sshRunner.RunPrivate(fmt.Sprintf("cat <<EOF | sudo tee /etc/systemd/system/kubelet.service.d/10-default-env.conf\n%s\nEOF", p))
+	err = sshRunner.SetTextContentAsRoot("/etc/systemd/system/kubelet.service.d/10-default-env.conf", p, 0644)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -200,6 +200,8 @@ func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error
 	if err != nil {
 		return err
 	}
+	_, _ = sshRunner.RunPrivate("sudo chmod 600 /var/lib/kubelet/config.json")
```
praveenkumar: @cfergeau I would prefer to have a single command for ssh, something like following so we also don't ignore the error.
```patch
diff --git a/pkg/crc/cluster/cluster.go b/pkg/crc/cluster/cluster.go
index e17235d..e0419cd 100644
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -196,7 +196,7 @@ func AddProxyConfigToMarketplaceOperator(oc oc.OcConfig, proxy *network.ProxyCon
 }
 
 func addPullSecretToInstanceDisk(sshRunner *ssh.SSHRunner, pullSec string) error {
-       _, err := sshRunner.RunPrivate(fmt.Sprintf("cat <<EOF | sudo tee /var/lib/kubelet/config.json\n%s\nEOF", pullSec))
+       _, err := sshRunner.RunPrivate(fmt.Sprintf("sudo install -m 600 /dev/null /var/lib/kubelet/config.json && cat <<EOF | sudo tee /var/lib/kubelet/config.json\n%s\nEOF", pullSec))
        if err != nil {
                return err
        }
```

cfergeau: Ignoring the error was intentional as it's no big deal if we fail to change the permissions, it's only causing a warnings in some operator's logs. I'll try your suggestion though as I'm fine either way.

pkg/crc/ssh/ssh.go
```diff
@@ -25,6 +27,13 @@ func (runner *SSHRunner) Run(command string) (string, error) {
 	return runner.runSSHCommandFromDriver(command, false)
 }
 
+func (runner *SSHRunner) SetTextContentAsRoot(destFilename string, content string, mode os.FileMode) error {
+	command := fmt.Sprintf("sudo install -m 0%o /dev/null %s && cat <<EOF | sudo tee %s\n%s\nEOF", mode, destFilename, destFilename, content)
+	_, err := runner.RunPrivate(command)
```
praveenkumar: Any specific reason we need to run those in private, since these command then will not be part of debug output.

pkg/crc/ssh/ssh.go
```diff
@@ -25,6 +27,13 @@ func (runner *SSHRunner) Run(command string) (string, error) {
 	return runner.runSSHCommandFromDriver(command, false)
 }
 
+func (runner *SSHRunner) SetTextContentAsRoot(destFilename string, content string, mode os.FileMode) error {
```
praveenkumar: :+1:

pkg/crc/ssh/ssh.go
```diff
@@ -28,6 +28,7 @@ func (runner *SSHRunner) Run(command string) (string, error) {
 }
 
 func (runner *SSHRunner) SetTextContentAsRoot(destFilename string, content string, mode os.FileMode) error {
+	logging.Debugf("Creating %s with permissions 0%o in the CRC VM", destFilename, mode)
```
praveenkumar: This should be part of earlier commit I think.

pkg/crc/ssh/ssh.go
cfergeau: If you look at the 4 locations using `SetTextContentAsRoot`, 3 of them were using `RunPrivate` even if for 2 of these, it may have been a copy and paste mistake. With the debug log that we are copying a file to the VM, it's probably good enough.
If we want to dump the ssh command when possible, we can decide whether to use RunPrivate or not according to `mode` (`0?00` -> private, otherwise, not private).

pkg/crc/ssh/ssh.go
cfergeau: yes!

PR #1038, opened 2020-02-25
Add proxy setting to ingress and image registry operators

cfergeau:
After some investigations with tcpconnect, I noticed these operators/pods have an inject-proxy annotation and are not using the proxy setting in crc

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -178,17 +178,12 @@ Environment=NO_PROXY=%s,%s`
 	return nil
 }
 
-// crc clusters are not running the cluster-version-operator, which is responsible for adding proxy env variables
-// to deployments which have the 'inject-proxy' annotation. This helper will be used to add the proxy env variables
-// to the deployments which should have it
-func AddProxyConfigToDeployment(oc oc.OcConfig, deployment string, namespace string, proxy *network.ProxyConfig) error {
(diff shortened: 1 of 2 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -178,21 +178,55 @@ Environment=NO_PROXY=%s,%s`
...
+	}
+
+	err = addProxyConfigToDeployment(oc, "openshift-controller-manager-operator", "openshift-controller-manager-operator", proxy)
+	if err != nil {
+		logging.Debugf("Failed to add proxy settings to image registry operator: %v", err)
+		multiErr.Collect(err)
+	}
+
+	return multiErr.ToError()
 }
```
anjannath: Was just wondering if you had considered using a map of strings to store the deployment name and namespace then looping over it to add the proxy config instead of duplicating?

praveenkumar: we can do that but I think if @cfergeau test around the bundle which have CVO not disabled make all those obsolete then we might not even need this one.

cfergeau:
This one is probably going to be obsoleted by https://github.com/code-ready/snc/pull/157

cfergeau:
Closing for now, the branch will still be around in case we realize it's needed later :) 
But since snc no longer disables the CVO, this should no longer be needed.

Issue #1013, opened 2020-02-14
CheckCRCLocalDNSReachableFromHost should check the IP it resolves


cfergeau:
Currently CheckCRCLocalDNSReachableFromHost tries to resolve api.crc.testing and foo.crc-apps.testing, but it's not verifying they resolved to the VM IP
It's possible to have non-working DNS configurations where unknown names would resolve to an IP, so it would be better to check we resolved these hostnames to a known IP.

PR #1042, opened 2020-02-26
Issue #1013: Check resolved IP in CheckCRCLocalDNSReachableFromHost

cfergeau:
CheckCRCLocalDNSReachableFromHost checks that the api.crc.testing and
foo.apps-crc.testing hostnames can be resolved, but it does not check
this returns the IP of the VM. This commit adds an additional check
that the resolved IP matches the IP of the VM as this could help crc
detect some host misconfigurations early.

This resolves https://github.com/code-ready/crc/issues/1013

I've only tested this on linux.

### DIFF
--- a/pkg/crc/network/utils.go
+++ b/pkg/crc/network/utils.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/ssh"
+
+	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 )
 
 func executeCommandOrExit(sshRunner *ssh.SSHRunner, command string, errorMessage string) string {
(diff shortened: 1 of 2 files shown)


pkg/crc/network/utils.go
```diff
@@ -95,23 +95,42 @@ func UriStringForDisplay(uri string) (string, error) {
...
+	apiHostname := fmt.Sprintf("api.%s.%s", clusterName, domainName)
+	ip, err := net.LookupIP(apiHostname)
 	if err != nil {
 		return err
 	}
-	logging.Debugf("api.%s.%s resolved to %s", clusterName, domainName, ip)
+	logging.Debugf("%s resolved to %s", apiHostname, ip)
+	if !matchIP(ip, expectedIP) {
+		logging.Debugf("%s is expected to resolve to %s", apiHostname, expectedIP)
+		return fmt.Errorf("Invalid IP for %s", apiHostname)
```
praveenkumar: @cfergeau can we also add what we are expecting and what we got when failure happen otherwise  user have to again run with debug mode to see what going on.

pkg/crc/network/utils.go
```diff
@@ -95,23 +97,42 @@ func UriStringForDisplay(uri string) (string, error) {
...
+func CheckCRCLocalDNSReachableFromHost(bundle *bundle.CrcBundleInfo, expectedIP string) error {
+	apiHostname := bundle.GetAPIHostname()
+	ip, err := net.LookupIP(apiHostname)
 	if err != nil {
 		return err
 	}
-	logging.Debugf("api.%s.%s resolved to %s", clusterName, domainName, ip)
+	logging.Debugf("%s resolved to %s", apiHostname, ip)
+	if !matchIP(ip, expectedIP) {
+		logging.Warnf("%s resolved to %s but %s was expected", apiHostname, ip, expectedIP)
```
praveenkumar: :+1:

pkg/crc/network/utils.go
```diff
@@ -95,23 +97,42 @@ func UriStringForDisplay(uri string) (string, error) {
...
-		ip, err = net.LookupIP(fmt.Sprintf("foo.%s", appsDomainName))
+		appsHostname := bundle.GetAppHostname("foo")
+		ip, err = net.LookupIP(appsHostname)
 		if err != nil {
 			return err
 		}
-		logging.Debugf("foo.%s resolved to %s", appsDomainName, ip)
+		logging.Debugf("%s resolved to %s", appsHostname, ip)
+		if !matchIP(ip, expectedIP) {
+			logging.Warnf("%s resolved to %s but %s was expected", appsHostname, ip, expectedIP)
```
praveenkumar: :+1:

praveenkumar:
I tested it on the mac and it was working as expected.

pkg/crc/network/utils.go
```diff
@@ -121,7 +123,7 @@ func CheckCRCLocalDNSReachableFromHost(clusterName, domainName, appsDomainName,
 		 * in this case, /etc/resolver/ will not be used, so we won't
 		 * have wildcard DNS for our domains
 		 */
-		appsHostname := fmt.Sprintf("foo.%s", appsDomainName)
+		appsHostname := bundle.GetAppHostname("foo")
```
praveenkumar: :+1:

praveenkumar:
retest this please.

pkg/crc/network/utils.go
cfergeau: If you think it's better this way, I'll squash the `fixup` commit

praveenkumar: @cfergeau yes I think it much better, I will have to test this tomorrow on mac.

Issue #149, opened 2019-05-21
Document how to configure the VM


robin-owen:
Related: PR #143

robin-owen:
Work on this is ongoing.

robin-owen:
Fixed by #1045.

PR #1045, opened 2020-02-27
Issue #149 Document configuration

robin-owen:
This PR primarily documents configuring the virtual machine, as that procedure requires more steps than general configuration (which is simply `crc config set <property>`).

The documentation for configuration in general is intentionally non-specific. We guide the reader as to how to view the current configuration and modify it, but do not list the configurable properties. These can be discovered via `crc config --help`, and not listing them explicitly removes a potential maintenance concern.

Configuration of preflight checks for the `crc start` command are also briefly mentioned.

### DIFF
--- a/docs/source/getting_started/content/con_about-codeready-containers-configuration.adoc
+++ b/docs/source/getting_started/content/con_about-codeready-containers-configuration.adoc
@@ -1,0 +1,12 @@
+[id="about-codeready-containers-configuration_{context}"]
+= About {prod} configuration
+
+You can use the [command]`{bin} config` command to configure both the [command]`{bin}` binary and the {prod} virtual machine.
+The [command]`{bin} config` command requires a subcommand to act on the configuration.
+The available subcommands are `get`, `set,` `unset`, and `view`.
+The `get`, `set`, and `unset` subcommands operate on named configurable properties.
+Run the [command]`{bin} config --help` command to list the available properties.
+
+You can also use the [command]`{bin} config` command to configure the behavior of the startup checks for the [command]`{bin} start` and [command]`{bin} setup` commands.
+By default, startup checks report an error and stop execution when their conditions are not met.
+To change this behavior, set the value of a property starting with `skip-check` or `warn-check` to `true` to skip the check or report a warning rather than an error, respectively.
--- a/docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
+++ b/docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
@@ -1,0 +1,52 @@
+[id="configuring-the-virtual-machine_{context}"]
+= Configuring the virtual machine
+
+Use the `cpus` and `memory` properties to configure the default number of vCPUs and amount of memory available to the {prod} virtual machine, respectively.
+
+[IMPORTANT]
+====
+You cannot modify the configuration of an existing {prod} virtual machine.
+To enable configuration changes, you must delete the existing virtual machine and create a new one.
+
+To create the new virtual machine, first delete the existing {prod} virtual machine, then start a new virtual machine with the configuration changes:
+
+[subs="+quotes,attributes"]
+----
+$ {bin} delete
+$ {bin} start
+----
+====
+
+[WARNING]
+====
+The [command]`{bin} delete` command results in the loss of data stored in the {prod} virtual machine.
+Save any desired information stored in the virtual machine before running this command.
+====
+
+.Procedure
+
+* To increase the number of vCPUs available to the virtual machine:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config set cpus _<number>_
+----
++
+The default value for the `cpus` property is `4`.
+The number of vCPUs to allocate must be greater than or equal to the default.
+
+* To increase the memory available to the virtual machine:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config set memory _<number-in-mib>_
+----
++
+[NOTE]
+====
+Values for available memory must be supplied in mebibytes (MiB).
+One gibibyte (GiB) of memory is equal to 1024 MiB.
+====
++
+The default value for the `memory` property is `8192`.
+The amount of memory to allocate must be greater than or equal to the default.
--- a/docs/source/getting_started/content/snip_crc-delete.adoc
+++ b/docs/source/getting_started/content/snip_crc-delete.adoc
@@ -5,6 +5,6 @@ $ {bin} delete
 +
 [WARNING]
 ====
-The [command]`{bin} delete` command will result in the loss of data stored in the {prod} virtual machine.
-Save any desired instance information before running this command.
+The [command]`{bin} delete` command results in the loss of data stored in the {prod} virtual machine.
+Save any desired information stored in the virtual machine before running this command.
 ====
--- a/docs/source/getting_started/content/assembly_configuring-codeready-containers.adoc
+++ b/docs/source/getting_started/content/assembly_configuring-codeready-containers.adoc
@@ -1,0 +1,8 @@
+[id="configuring-codeready-containers_{context}"]
+= Configuring {prod}
+
+include::con_about-codeready-containers-configuration.adoc[leveloffset=+1]
+
+include::proc_viewing-codeready-containers-configuration.adoc[leveloffset=+1]
+
+include::proc_configuring-the-virtual-machine.adoc[leveloffset=+1]
--- a/docs/source/getting_started/content/proc_viewing-codeready-containers-configuration.adoc
+++ b/docs/source/getting_started/content/proc_viewing-codeready-containers-configuration.adoc
@@ -1,0 +1,32 @@
+[id="viewing-codeready-containers-configuration_{context}"]
+= Viewing {prod} configuration
+
+The {prod} binary provides commands to view configurable properties and the current {prod} configuration.
+
+.Procedure
+
+* To view the available configurable properties:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config --help
+----
+
+* To view the values for a configurable property:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config get _<property>_
+----
+
+* To view the complete current configuration:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config view
+----
++
+[NOTE]
+====
+The [command]`{bin} config view` command does not return any information if the configuration consists of default values.
+====
--- a/docs/source/getting_started/master.adoc
+++ b/docs/source/getting_started/master.adoc
@@ -10,6 +10,8 @@ include::content/assembly_installation.adoc[leveloffset=+1]
 
 include::content/assembly_using-codeready-containers.adoc[leveloffset=+1]
 
+include::content/assembly_configuring-codeready-containers.adoc[leveloffset=+1]
+
 include::content/assembly_networking.adoc[leveloffset=+1]
 
 include::content/assembly_administrative-tasks.adoc[leveloffset=+1]


docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,10 @@
+[id="understanding-codeready-containers-configuration_{context}"]
+= Understanding {prod} configuration
+
+Both the [command]`{bin}` binary and the {prod} virtual machine can be configured using the [command]`{bin} config` command.
+The [command]`{bin} config` command requires a subcommand to act on the configuration.
+The available subcommands are `get`, `set,` `unset`, and `view`.
+The `get`, `set`, and `unset` subcommands operate using named configurable properties.
+
+Preflight checks for the [command]`{bin} start` command can be configured to be skipped or to report a warning rather than an error.
```
cfergeau: This also applies to `crc setup`

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,10 @@
+[id="understanding-codeready-containers-configuration_{context}"]
+= Understanding {prod} configuration
+
+Both the [command]`{bin}` binary and the {prod} virtual machine can be configured using the [command]`{bin} config` command.
+The [command]`{bin} config` command requires a subcommand to act on the configuration.
+The available subcommands are `get`, `set,` `unset`, and `view`.
```
cfergeau: `get` and `unset` are not documented in the rest of this PR, dunno if that's intentional?

robin-owen: Good point. I'll add documentation for those.

robin-owen: I've added documentation for `crc config get <property>`, but I believe `unset` more or less explains itself.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,10 @@
+[id="understanding-codeready-containers-configuration_{context}"]
+= Understanding {prod} configuration
```
stoobie: As a general rule, I stay away from "Understanding" as a title. There was a discussion in the Style council about the best way to title topics like this, but you can never guarantee understanding. "About {prod} configuration" is more accurate, or "{prod} configuration explained" (Which I don't like the sound of, but is more accurate than 'understanding')

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+Both the [command]`{bin}` binary and the {prod} virtual machine can be configured using the [command]`{bin} config` command.
```
stoobie: Passive voice "can be configured". Suggestion:

You can use the [command]`{bin} config` command to configure both the [command]`{bin}` binary and the {prod} virtual machine.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
stoobie: Why do you use the [command] prefix for `{bin} config`, but not for these subcommands?

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,10 @@
...
+The [command]`{bin} config` command requires a subcommand to act on the configuration.
+The available subcommands are `get`, `set,` `unset`, and `view`.
+The `get`, `set`, and `unset` subcommands operate using named configurable properties.
```
stoobie: Is the word "operate" here really necessary? I understand the sentence without it, at least I think I do.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+Preflight checks for the [command]`{bin} start` and [command]`{bin} setup` commands can be configured to be skipped or to report a warning rather than an error.
```
stoobie: I don't know what "preflight" means in this context. Is "preflight" a concept that you've already explained somewhere previously? If not, it sounds like jargon or slang to me, especially for translators.

"can be configured" is passive voice. I suggest rephrasing as I suggested in my previous similar comment.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,10 @@
...
+The configurable properties for preflight checks begin with `skip-check` or `warn-check`, respectively.
```
stoobie: I don't understand the technical concept here, so discard this command if it's out of place...

I'm commenting on "begin with". If you have a bunch of configurable properties, there is no real beginning or end. A list of properties _does_ have a beginning and end, or if you have to define them in a certain order.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
```diff
@@ -0,0 +1,50 @@
+[id="configuring-the-virtual-machine_{context}"]
+= Configuring the virtual machine
+
+The default number of vCPUs and amount of memory available to the {prod} virtual machine can be configured using the `cpus` and `memory` configurable properties, respectively.
```
stoobie: "can be configured" passive voice. 

In this case, I do understand the word "respectively". You're saying that the "cpus" property defines the number of vCPUs and the "memory" property defines the amount of memory available to the vm.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+[IMPORTANT]
+====
+Changes to the configuration of the virtual machine require the creation of a new {prod} virtual machine.
```
stoobie: It sounds like I can't actually change the configuration of my current vm. Is that the case?
Are you saying that if I change the configuration of the vm, I have to create a new vm for those changes to take effect? If so, what happens to my existing vm? Do I need to destroy it?

gbraad: At the moment we can not modify a VM in-place due to a limitation in the libmachine library we use.

Yes, a new VM is needed.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
```diff
@@ -0,0 +1,50 @@
...
+To create the new virtual machine, first delete the existing {prod} instance, then start a new instance with the configuration changes:
```
stoobie: Is an instance the same as a vm? If so, is there a reason to use a different word here (instance vs vm)?

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
```diff
@@ -0,0 +1,50 @@
...
+[subs="+quotes,attributes"]
+----
+$ {bin} delete
+$ {bin} start
+----
+====
+
+[WARNING]
+====
+The [command]`{bin} delete` command will result in the loss of data stored in the {prod} virtual machine.
```
stoobie: "will result" => "results"

Unless there's a clear reason not to, use present tense. A clear reason means you can't avoid it. In this case, the present tense is clearer and less jargony.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
gbraad: This discussion has been had before... CRC instance is essential the VM. We already had document-wide change for this before.

Orginally we avoided the term VM as this was related to how some companies have a policy against the use of a hypervisor. We have since changed our thinking.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
```diff
@@ -0,0 +1,50 @@
...
+Save any desired instance information before running this command.
+====
+
+.Procedure
+
+* To increase the number of vCPUs available to the virtual machine:
```
stoobie: I just had a conversation with Junko about this lead-in sentence. I prefer the way you wrote it here, but she told me that broken sentences like this require manual translator intervention, so it's better to have a complete sentence in lead-ins. The IBM Style Guide doesn't explicitly say that you have to have a complete sentence here, but its examples all include something like "complete the following step(s)" or in this case I'd write "enter the following command:"

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
stoobie: I'm just saying that as a reader, I don't intuitively understand that "instance" = "vm", so that's confusing.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
stoobie: OK, that's not clear to me when I read this.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
gbraad: @robin-owen @stoobie WDYT?

stoobie: @gbraad Better.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
stoobie: The following would be clearer to me:

You cannot modify the configuration of an existing virtual machine. To enable configuration changes, you need delete the current vm and create a new vm.

gbraad:
@stoobie you add comments, but you do not request for changes?

stoobie:
> @stoobie you add comments, but you do not request for changes?

@gbraad Sorry, I'm a noob with Github. I intend that you take my comments and suggestions seriously. Consider them as change requests.

How should I add change requests?

gbraad:
@stoobie That is why I wanted to confirm this, as it already got approved by an engineer.

image

Choose "Request changes" and confirm with "Submit review". Be sure to add a comment in the field, else it will refuse to submit.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
robin-owen: We've used "run the following command" in Minishift docs before, so I'm fine with that. This would require some changes across the entire documentation set, though, so it might be worth creating a new docs issue to encompass the changes across the entire set.

For what it's worth, I keep hearing contradictory things regarding wording such as this. The contradictions largely seem to come from minimalism telling us to remove the fluff, while localization benefits from complete sentences.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
robin-owen: @gbraad I'm inclined to go with "virtual machine", as that's unambiguous.

Like some other things in this review, that would require changes across the rest of the docs set. I don't have a problem with that, but would want to create a separate issue to address those changes in the near future.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
robin-owen: Good call on "preflight" being jargon. This is a term that was used previously in Minishift for checks which are run prior to starting the VM which provisions the OpenShift cluster.

Perhaps "startup checks" or "pre-start checks" would be more understandable. What do you think?

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
robin-owen: Succinctly, because `crc config` is a command -- `get`, `set`, `unset`, and `view` are subcommands that cannot be directly run.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
robin-owen: Fair enough. In this case, I think "About {prod} configuration" is a suitable title.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
robin-owen: "Operate on" makes the most sense.

In this context, a named configurable property is the accurate term for... well, a configurable property that has a name. These can be listed with `crc config --help`.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
robin-owen: "Begin with" is referring to the names of configurable properties. A relatively large number of them exist for the pre-start (or "preflight") checks that are run on `crc setup` and `crc start`. All of the names of properties related to these checks start with `skip-check` or `warn-check`.

For example, here are some of the properties:
 * skip-check-bundle-cached
 * warn-check-bundle-cached
 * skip-check-crc-dnsmasq-file
 * warn-check-crc-dnsmasq-file
 * skip-check-crc-network
 * warn-check-crc-network
 * skip-check-crc-network-active
 * warn-check-crc-network-active

In this case, we can probably clarify that simply by stating that "The names of configurable properties for pre-start checks begin with `skip-check` or `warn-check`."

The "respectively" regards whether the check is skipped entirely or reports a warning rather than an error.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
gbraad: the term pre-flight checks is mostly the internal naming and how we talk about them. Startup checks are more correct in term to the user.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
gbraad: We hope to resolve this in the future... but for now a caution note might be needed as this is destructive.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
gbraad: The problem came from the use when we talk about both the VM and cluster. the consensus is to use "virtual machine" when talking about the means, but we are starting an "OpenShift cluster" (and for this we use a virtual machine". Perhaps this is why "instance" avoids having to choose between these two? :-/

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
robin-owen: I won't address this particular point in this PR simply because the requested change is larger in scope than this PR alone. For consistency, I'm okay with this PR going in as-is with the lead-in sentences. I'll create another issue to address lead-in sentences across the documentation set.

robin-owen:
@stoobie I've updated the PR to address your change requests. Do you mind re-reviewing the changes made?

stoobie:
> @stoobie That is why I wanted to confirm this, as it already got approved by an engineer.
> 
> image
> 
> Choose "Request changes" and confirm with "Submit review". Be sure to add a comment in the field, else it will refuse to submit.

Thank you for that explanation! I was wondering what the difference was between just adding comments and using the Review Changes button.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
stoobie: Yeah, that's fine. We have the same issues in our docs. With anything like this that requires a lot of changes for the sake of consistency, it's usually more practical to start doing the "right" thing and then extend to legacy content in a separate phase of work.

stoobie: As far as your remark about minimalism...one thing that Ingrid mentioned is that minimalism does not necessarily mean fewer words. It means being as concise *and clear* as possible. In this case, I agree that the extra words are fluffy, and because of that, it was important for me to hear directly from Junko what the exact issues are. So minimalistic writing does need to take into account other issues, like translation. And sometimes clarity requires more words, but that doesn't meant that more words are fluff.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
stoobie: Ah...so I get it.

So how about this:

> You can configure the [command]`{bin} start` and [command]`{bin} setup` commands such that the preflight check skips or reports a warning rather than an error.

> To configure the commands, add one of the following prefixes to each property:

> `skip-check`::Skips the preflight check. For example: `skip-check-bundle-cached`
> `warn-check`::Reports a warning rather than an error. For example: `warn-check-bundle-cached`

> For example:
F

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
stoobie: > In this context, a named configurable property is the accurate term for... well, a configurable property that has a name. These can be listed with `crc config --help`.

Ok. The word "named" sounds redundant (i.e. fluff) to me. Is there a configurable property without a name? I'm assuming there isn't.

docs/source/getting_started/content/con_understanding-codeready-containers-configuration.adoc
stoobie: Is a startup check a check that takes place when something is started? I don't see the full context here, so I'm guessing you're talking about a virtual machine starting up (i.e. booting). Is that correct?

Anyway, "startup check" is more intuitive and less jargony to my ear.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
stoobie: @gbraad I'm not sure what you're referring to when you say "this is destructive". A caution  is fine. I was just commenting on what I understood from the text.

stoobie:
> For @stoobie

@gbraad I'm sorry but I don't understand if something is being requested of me in this comment.

stoobie:
> @stoobie I've updated the PR to address your change requests. Do you mind re-reviewing the changes made?

@robin-owen I added my comments. Again, I apologize for my inexperience with Github, but if there's something I need to do at this point so that you can merge this, please tell me clearly, like @gbraad did when telling me how to review changes.

gbraad:
I marked the PR as "Request changes" to make sure this does not get merged; and as mentioned before a comment is needed; hence the "for @stoobie" as a follow-up to indicate why i do this. seems this got lost ... my excuse.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
```diff
@@ -0,0 +1,52 @@
+[id="configuring-the-virtual-machine_{context}"]
+= Configuring the virtual machine
+
+Use the `cpus` and `memory` properties to configure the default number of vCPUs and amount of memory available to the {prod} virtual machine, respectively.
```
stoobie: This is fine but getting back to your comment about minimalism, I would have written this:

> The `cpus` property configures the default number of vCPUs.
> The  `memory` property configures the amount of memory available to the {prod} virtual machine.

docs/source/getting_started/content/snip_crc-delete.adoc
```diff
@@ -5,6 +5,6 @@ $ {bin} delete
 +
 [WARNING]
 ====
-The [command]`{bin} delete` command will result in the loss of data stored in the {prod} virtual machine.
-Save any desired instance information before running this command.
+The [command]`{bin} delete` command results in the loss of data stored in the {prod} virtual machine.
+Save any desired information stored in the virtual machine before running this command.
```
stoobie: much better.

docs/source/getting_started/content/con_about-codeready-containers-configuration.adoc
```diff
@@ -0,0 +1,12 @@
...
+
+You can use the [command]`{bin} config` command to configure both the [command]`{bin}` binary and the {prod} virtual machine.
+The [command]`{bin} config` command requires a subcommand to act on the configuration.
+The available subcommands are `get`, `set,` `unset`, and `view`.
+The `get`, `set`, and `unset` subcommands operate on named configurable properties.
+Run the [command]`{bin} config --help` command to list the available properties.
+
+You can also use the [command]`{bin} config` command to configure the behavior of the startup checks for the [command]`{bin} start` and [command]`{bin} setup` commands.
+By default, startup checks report an error and stop execution when their conditions are not met.
+To change this behavior, set the value of a property starting with `skip-check` or `warn-check` to `true` to skip the check or report a warning rather than an error, respectively.
```
stoobie: Much better.

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
robin-owen: I'm aware of the goal of minimalism, and that it doesn't mean fewer words. :)

In this particular context (lead-in sentences), though, I've explicitly heard advice going in opposing directions. ;)

docs/source/getting_started/content/proc_configuring-the-virtual-machine.adoc
robin-owen: Fair enough, but those sentences seem more like part of the procedure itself rather than an introduction to the steps being performed in my opinion. :)

PR #1046, opened 2020-02-27
 Use 'RunOcCommandPrivate' method for pull secret

praveenkumar:


pkg/os/exec.go
```diff
@@ -30,6 +29,15 @@ func run(command string, args []string, env map[string]string) (string, string,
...
 }
 
+func run(command string, args []string, env map[string]string) (string, string, error) {
+	logging.Debugf("Running '%s %s'", command, strings.Join(args, " "))
+	return runCmd(command, args, env)
+}
+
+func runPrivate(command string, args []string, env map[string]string) (string, string, error) {
+	return runCmd(command, args, env)
+}
```
cfergeau: Maybe you can add a `private` arg to `run` rather than having multiple one-line helpers? This would be closer to what is done in `ssh.go` (merely a suggestion).
We need a log similar to `logging.Debugf("About to run SSH command with hidden output")` though in the `runPrivate` case.

pkg/os/exec.go
```diff
@@ -20,7 +20,6 @@ func run(command string, args []string, env map[string]string) (string, string,
 	stdErr := new(bytes.Buffer)
 	cmd.Stdout = stdOut
 	cmd.Stderr = stdErr
-	logging.Debugf("Running '%s %s'", command, strings.Join(args, " "))
 	err := cmd.Run()
 	if err != nil {
 		logging.Debugf("Command failed: %v", err)
```
cfergeau: Wondering if `oc` will print the failed command/args and leak the pull-secret?

pkg/crc/machine/cert-renewal.go
```diff
@@ -50,6 +50,13 @@ func (recoveryPod recoveryPod) Run(args ...string) (string, string, error) {
 	return stdout, "", err
 }
 
+func (recoveryPod recoveryPod) RunPrivate(args ...string) (string, string, error) {
+	cmd := fmt.Sprintf("sudo oc %s", strings.Join(args, " "))
+	logging.Debugf("kube-apiserver-operator: Running %s", cmd)
```
cfergeau: We don't want this log

pkg/os/exec.go
praveenkumar: @cfergeau this is what I thought first but then we need to change the `run` function signature everywhere to add that private arg value to false where we don't want it. I did thought about it and I think abstraction make it bit better then changing the signature everywhere.

pkg/os/exec.go
praveenkumar: My assumtion is oc command mostly fail either the apiserver is not up or the resource is not valid and both the cases I think we are safe.

pkg/crc/machine/cert-renewal.go
praveenkumar: Ah right, we will remove this one.

pkg/os/exec.go
praveenkumar: > We need a log similar to logging.Debugf("About to run SSH command with hidden output") though in the runPrivate case.

Good point, will update this accordingly.

Issue #747, opened 2019-10-19
Corrupt bundles are not overwritten by `crc setup`


ajaysimha:
### General information

  * OS: Linux (Fedora 30)
  * Hypervisor: KVM 
  * Did you run `crc setup` before starting it (Yes/No)? Yes

## CRC version
```bash
# [asimha@asimha-fedora crc-linux-1.0.0-rc.0-amd64]$ crc version
version: 1.0.0-rc.0+34371d3
OpenShift version: 4.2.0-0.nightly-2019-09-26-192831 (embedded in binary)

```
  
## CRC status
```bash
# Put `crc status` output here
```

## CRC config
```bash
# Put `crc config view` output here
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
# put the output of `sw_vers` in case of Mac
# Put the output of `systeminfo` in case of Windows
```
[asimha@asimha-fedora Downloads]$ cat /etc/redhat-release 
Fedora release 30 (Thirty)

### Steps to reproduce

  1. Download CRC and pull-secret.txt
  2. run crc setup
  3. run crc start
  4. 

### Expected
complete without errors

### Actual
[asimha@asimha-fedora crc-linux-1.0.0-rc.0-amd64]$ ./crc start -p ~/Downloads/pull-secret.txt 
INFO Checking if running as non-root              
INFO Checking if oc binary is cached              
INFO Checking if Virtualization is enabled        
INFO Checking if KVM is enabled                   
INFO Checking if libvirt is installed             
INFO Checking if user is part of libvirt group    
INFO Checking if libvirt is enabled               
INFO Checking if libvirt daemon is running        
INFO Checking if a supported libvirt version is installed 
INFO Checking if crc-driver-libvirt is installed  
INFO Checking if libvirt 'crc' network is available 
INFO Checking if libvirt 'crc' network is active  
INFO Checking if NetworkManager is installed      
INFO Checking if NetworkManager service is running 
INFO Checking if /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf exists 
INFO Checking if /etc/NetworkManager/dnsmasq.d/crc.conf exists 
INFO Extracting bundle: crc_libvirt_4.2.0-0.nightly-2019-09-26-192831.crcbundle ... 
ERRO Error occurred: Error getting bundle metadata: xz: data is truncated or corrupt 

### Logs

You can start crc with `crc start --log-level debug` to collect logs.
[...]

PR #1049, opened 2020-02-28
Issue #747 Enable sha256sum check for disk image

praveenkumar:
As of now there is no way to verify if the extracted disk image
is in sane condition or corupted. Using this patch now we can
match the sha256sum which is part of our bundle metadata to actual
disk image sha256sum and let user know if there is mismatch.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -1,9 +1,13 @@
 package machine
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"runtime"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/machine.go
```diff
@@ -73,6 +77,34 @@ func getCrcBundleInfo(bundlePath string) (*bundle.CrcBundleInfo, error) {
...
+	if err != nil {
+		return err
+	}
+	if expectedSha != gotSha {
+		return fmt.Errorf("Expected sha %s Got %s", expectedSha, gotSha)
+	}
+	return nil
+}
+
+func getSha256Sum(filename string) (string, error) {
```
praveenkumar: I want this function to public but didn't find at which package it should be part of or a complete new pacakge.

gbraad: this is generic functionality to obtain a digest of a file, right? so either a namespace `files` or `digest`.

or perhaps even `validation`... but for that you need to provide the the compare to verification string and return a boolean.

PR #1053, opened 2020-03-02
Enable size check for disk image

praveenkumar:
As of now there is no way to verify if the extracted disk image
is in sane condition or corupted. Using this patch now we can
match the size which is part of our bundle metadata to actual
disk image size and let user know if there is mismatch.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -136,6 +136,12 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error getting bundle metadata: %v", err)
 		}
 
+		logging.Infof("Checking size of the disk image %s ...", crcBundleMetadata.GetDiskImagePath())
+		if err := crcBundleMetadata.CheckDiskImageSize(); err != nil {
+			result.Error = err.Error()
+			return *result, errors.Newf("Invalid bundle disk image '%s', %v", crcBundleMetadata.GetDiskImagePath(), err)
+		}
+
 		openshiftVersion := crcBundleMetadata.GetOpenshiftVersion()
 		if openshiftVersion == "" {
 			logging.Infof("Creating VM...")
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/crc/machine/machine.go
```diff
@@ -73,6 +75,25 @@ func getCrcBundleInfo(bundlePath string) (*bundle.CrcBundleInfo, error) {
 	return bundle.Extract(bundlePath)
 }
 
+func checkDiskImageSize(bundleinfo *bundle.CrcBundleInfo) error {
+	diskImagePath := bundleinfo.GetDiskImagePath()
+	expectedSizeStr := bundleinfo.GetDiskSize()
+	expectedSize, err := strconv.ParseInt(expectedSizeStr, 10, 64)
+	if err != nil {
+		return err
+	}
```
cfergeau: The int conversion should be done in `bundleinfo.GetDiskSize()`

pkg/crc/machine/machine.go
```diff
@@ -136,6 +157,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error getting bundle metadata: %v", err)
 		}
 
+		if err := checkDiskImageSize(crcBundleMetadata); err != nil {
+			result.Error = err.Error()
+			return *result, errors.Newf("Error to match disk size %v", err)
```
cfergeau: "Error checking disk image size"?
But it's not a great description of what the problem is, as it won't help in solving it. Maybe "Invalid bundle disk image '%s'", bundleinfo.GetDiskImagePath()?

pkg/crc/machine/machine.go
```diff
@@ -73,6 +74,24 @@ func getCrcBundleInfo(bundlePath string) (*bundle.CrcBundleInfo, error) {
 	return bundle.Extract(bundlePath)
 }
 
+func checkDiskImageSize(bundleinfo *bundle.CrcBundleInfo) error {
+	diskImagePath := bundleinfo.GetDiskImagePath()
+	expectedSize, err := bundleinfo.GetDiskSize()
+	if err != nil {
+		return err
+	}
+	logging.Infof("Checking size of the disk image %s ...", diskImagePath)
```
gbraad: can we try not to put all functionality in machine.go ?

it would be better if we only perform a boolean action in machine.go, but put the code in a separate file or even inside a more logical place.

pkg/crc/machine/machine.go
```diff
@@ -136,6 +155,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error getting bundle metadata: %v", err)
 		}
 
+		if err := checkDiskImageSize(crcBundleMetadata); err != nil {
```
gbraad: only this should be here...

... as this is part of the metadata check function, it makes sense to move the check code into the metadata.go file.

praveenkumar: @gbraad Ack

Issue #1041, opened 2020-02-26
'crc status' should show OpenShift's status as "Starting" instead of "Stopped"


dharmit:
When starting CRC, the OpenShift web console is available pretty soon but doing `crc status` shows that OpenShift is "Stopped". After a few minutes, it turns to "Running". The "Stopped" status is apparently because few Operators are still starting up while rest of the bits have started.

It would be helpful if the status of OpenShift shows as "Starting" or something other than "Stopped" so that it doesn't get misinterpreted by the end-user.

```sh
❯ crc status
CRC VM:          Running
OpenShift:       Stopped    <------   why is this "Stopped"?
Disk Usage:      10.11GB of 32.2GB (Inside the CRC VM)
Cache Usage:     22.99GB
Cache Directory: /home/dshah/.crc/cache

❯ crc version
crc version: 1.5.0+e1c8fb8
OpenShift version: 4.2.14 (embedded in binary)
```

cfergeau:
This should be fixed by https://github.com/code-ready/crc/commit/960a6a2909ac9b187e6d8ac533cc27f5dd0a0a02

karthik101:
What does it mean its fixed? do we need to download new crc version? Is there no alternate fix ?

cfergeau:
Yes you'd need a newer crc version to benefit from that fix.

PR #1058, opened 2020-03-03
Issue #1041 Add intermediate `Starting` status for the cluster

praveenkumar:
When starting CRC, some of the operators don't available even
after 3 min wait. A user who execute `crc status` sees openshift status
as stopped which is confusing since cluster is not completely stopped
but waiting for some of operators to be available. After this change, user
can see `Starting` as status.

```
$crc status
CRC VM:          Running
OpenShift:       Starting
Disk Usage:      10.11GB of 32.2GB (Inside the CRC VM)
Cache Usage:     22.99GB
```

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -582,6 +582,10 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
 				openshiftVersion = crcBundleMetadata.GetOpenshiftVersion()
 			}
 			openshiftStatus = fmt.Sprintf("Running (v%s)", openshiftVersion)
+		} else if operatorsStatus.Degraded {
+			openshiftStatus = fmt.Sprintf("Degraded")
+		} else if operatorsStatus.Progressing {
+			openshiftStatus = fmt.Sprintf("Starting")
 		}
 		sshRunner := crcssh.CreateRunner(host.Driver)
 		diskSize, diskUse, err = cluster.GetRootPartitionUsage(sshRunner)
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -567,13 +567,13 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
 
 		// check if all the clusteroperators are running
 		ocConfig := oc.UseOCWithConfig(statusConfig.Name)
-		operatorsRunning, err := oc.GetClusterOperatorStatus(ocConfig)
+		operatorsStatus, err := oc.GetClusterOperatorStatus(ocConfig)
 		if err != nil {
 			result.Success = false
 			result.Error = err.Error()
 			return *result, errors.New(err.Error())
 		}
-		if operatorsRunning {
+		if operatorsStatus.Available {
 			openshiftVersion := "4.x"
 			_, crcBundleMetadata, err := getBundleMetadataFromDriver(host.Driver)
 			if err != nil {
--- a/pkg/crc/oc/clusteroperator.go
+++ b/pkg/crc/oc/clusteroperator.go
@@ -25,19 +25,28 @@ type K8sResource struct {
 	} `json:"items"`
 }
 
-func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
-	allAvailable := true
+// https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource
+type ClusterStatus struct {
+	Available   bool
+	Degraded    bool
+	Progressing bool
+	Disabled    bool
+}
+
+func GetClusterOperatorStatus(oc OcConfig) (*ClusterStatus, error) {
+	cs := &ClusterStatus{}
 	data, stderr, err := oc.RunOcCommand("get", "co", "-ojson")
 	if err != nil {
-		return false, fmt.Errorf("%s - %v", stderr, err)
+		return cs, fmt.Errorf("%s - %v", stderr, err)
 	}
 
 	var co K8sResource
 
 	err = json.Unmarshal([]byte(data), &co)
 	if err != nil {
-		return false, err
+		return cs, err
 	}
+	cs.Available = true
 	for _, c := range co.Items {
 		if contains(c.Metadata.Name, ignoreClusterOperators) {
 			continue
@@ -47,31 +56,31 @@ func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
 			case "Available":
 				if con.Status != "True" {
 					logging.Debug(c.Metadata.Name, " operator not available, Reason: ", con.Reason)
-					allAvailable = false
+					cs.Available = false
 				}
 			case "Degraded":
-				if con.Status != "False" {
+				if con.Status == "True" {
 					logging.Debug(c.Metadata.Name, " operator is degraded, Reason: ", con.Reason)
-					allAvailable = false
+					cs.Degraded = true
 				}
 			case "Progressing":
-				if con.Status != "False" {
+				if con.Status == "True" {
 					logging.Debug(c.Metadata.Name, " operator is still progressing, Reason: ", con.Reason)
-					allAvailable = false
+					cs.Progressing = true
 				}
 			case "Upgradeable":
 				continue
 			case "Disabled":
-				if con.Status != "False" {
+				if con.Status == "True" {
 					logging.Debug(c.Metadata.Name, " operator is disabled, Reason: ", con.Reason)
-					allAvailable = false
+					cs.Disabled = true
 				}
 			default:
 				logging.Debugf("Unexpected operator status for %s: %s", c.Metadata.Name, con.Type)
 			}
 		}
 	}
-	return allAvailable, nil
+	return cs, nil
 }
 
 func contains(value string, list []string) bool {
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -186,7 +186,7 @@ func CheckClusterOperatorsWithRetry(retryCount int, retryWait string) error {
 		if err != nil {
 			return err
 		}
-		if s == true {
+		if s.Available == true {
 			return nil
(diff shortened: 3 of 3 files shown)


pkg/crc/machine/machine.go
```diff
@@ -582,6 +582,8 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
 				openshiftVersion = crcBundleMetadata.GetOpenshiftVersion()
 			}
 			openshiftStatus = fmt.Sprintf("Running (v%s)", openshiftVersion)
+		} else {
+			openshiftStatus = "Starting"
```
cfergeau: `func GetClusterOperatorStatus(oc OcConfig) (bool, error)` should return a more fine-grained status than just a boolean, if at least one operator is `Degraded`, I'm not sure `Starting` is correct.

praveenkumar: @cfergeau isn't it always at least one operator when the return value from `GetClusterOperatorStatus` is not false or you want which all are in degraded state?

gbraad: if we want to be finegrained we probably need to determine this ourselves.

cfergeau: `false` can mean "Starting" if one or more operators are "progressing", but it could also mean "Degraded" if one or more operators are "degraded". So hardcoding that `false` means "starting" in `machine.go` does not seem very good, I prefer that we do it in `GetClusterOperatorStatus` where all the logic already is.

praveenkumar: @cfergeau you want to have `Degraded` as `stopped` condition? As of now there can be following conditions and I want to understand your prospective on those?
- Available => True  (everything working)
- Degraded => An operator reports Degraded when its current state does not match its desired state over a period of time resulting in a lower quality of service.  (What you prefer here? ). Does this mean openshift is Stopped or in Starting state (since we are not using any intermediate state). 
 Ref: https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md
- Progressing => Starting

Do we want an intermediate state which says `Degraded` as whole cluster, if that is the case then we need to put some refactor, otherwise for me a `false` return from `GetClusterOperatorStatus` means it is starting.

cfergeau: I don't want to hardcode in `machine.go` that if `GetClusterOperatorStatus` returns `false`, then it means that the cluster is still being started. This is not always going to be true, as as you say, there's also 'degraded' to encode, and also potentially 'stopped'.

`GetClusterOperatorStatus` should return an enum instead of a boolean.
I'm fine if `GetClusterOperatorStatus` is changed for now to only report `starting` in addition to `started` while we figure out what to do with `degraded`.

praveenkumar: @cfergeau https://github.com/code-ready/crc/pull/1058/commits/eb6373f7d5562f325f1f6d29df49cd9616b56ca3 is what I come up with, now we don't skip any of the cluster state and make use of it.

gbraad:
have you asked the OP to give this a try?

dharmit:
I tried the binary and bundle for this PR. What I'm seeing is that `crc status` shows `Running` while `crc start` is still not finished.

```sh
$ ./crc start -b crc_libvirt_4.3.1.crcbundle
...
...
INFO Check DNS query from host ...                
INFO Copying kubeconfig file to instance dir ...  
INFO Adding user's pull secret ...                
INFO Updating cluster ID ...
```
In another window:
```sh
$ ./crc status
CRC VM:          Running
OpenShift:       Running (v4.3.1)
Disk Usage:      10.2GB of 32.72GB (Inside the CRC VM)
Cache Usage:     43.67GB
Cache Directory: /home/dshah/.crc/cache
```

praveenkumar:
> I tried the binary and bundle for this PR. What I'm seeing is that crc status shows Running while crc start is still not finished.

@dharmit this is because it is the previous state of cluster, how long you waited before did `crc status` ?

pkg/crc/oc/clusteroperator.go
```diff
@@ -25,19 +25,27 @@ type K8sResource struct {
 	} `json:"items"`
 }
 
-func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
-	allAvailable := true
+type ClusterStatus struct {
+	Available  bool
+	Degraded   bool
+	Processing bool
```
cfergeau: `Processing` or `Progressing`?

pkg/crc/oc/clusteroperator.go
```diff
@@ -25,19 +25,27 @@ type K8sResource struct {
 	} `json:"items"`
 }
 
-func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
-	allAvailable := true
+type ClusterStatus struct {
```
cfergeau: Do you expect to have multiple fields `true` at the same time? Or could it be an enumeration?
```
type ClusterStatus int
const (  
        Available = ClusterStatus iota
        Degraded
        Progressing
)
```

pkg/crc/oc/clusteroperator.go
praveenkumar: yes it is `Progressing`, Good catch :+1:

dharmit:
> @dharmit this is because it is the previous state of cluster, how long you waited before did `crc status` ?

I did a `crc delete` using the existing `crc` binary located somewhere in the `PATH` and then started new VM with `./crc` and bundle in cwd. I'm not exactly sure about how long I waited but I did it when the latest message displayed by `./crc start` on the console was what I mentioned in my previous comment.

pkg/crc/oc/clusteroperator.go
praveenkumar: @cfergeau we can have multiple fields `true` at the same time, there is a chance when an operator is available but also degrading like it need more replica but only few are available.

pkg/crc/oc/clusteroperator.go
```diff
@@ -47,31 +55,31 @@ func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
 			case "Available":
 				if con.Status != "True" {
 					logging.Debug(c.Metadata.Name, " operator not available, Reason: ", con.Reason)
-					allAvailable = false
+					cs.Available = false
```
cfergeau: I think I would keep the name `AllAvailable`. The other boolean are `true` when at least one operator has the state, this one is different as it's `true` only if all operators are available.

pkg/crc/machine/machine.go
```diff
@@ -582,6 +582,8 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
 				openshiftVersion = crcBundleMetadata.GetOpenshiftVersion()
 			}
 			openshiftStatus = fmt.Sprintf("Running (v%s)", openshiftVersion)
```
cfergeau: Before the check for `Progressing`, I would add a
```
} else if operatorsStatus.Degraded {
			openshiftStatus = fmt.Sprintf("Unknown")
} else if operatorsStatus.Processing {
			openshiftStatus = fmt.Sprintf("Starting")
}
```

pkg/crc/oc/clusteroperator.go
```diff
@@ -47,31 +55,31 @@ func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
...
-					allAvailable = false
+					cs.Disabled = true
 				}
 			default:
 				logging.Debugf("Unexpected operator status for %s: %s", c.Metadata.Name, con.Type)
 			}
 		}
 	}
-	return allAvailable, nil
+	return cs, nil
```
cfergeau: If one operator can report `(Progressing, Available)` and all other operators report `(Available)`, with that change we'll be returning a `ClusterStatus` with both `Progressing` and `Available` set to `true`. With the code before this PR, we were returning `false`. I don't know if this can happen for real though.

pkg/crc/machine/machine.go
praveenkumar: @cfergeau sure, I thought we are just going to ignore this state for sometime but I then why not  put `Degraded` instead `unknown` ?

pkg/crc/oc/clusteroperator.go
praveenkumar: it might happen for real if you just delete a single pod for a namespace which have mutiple replica set but that mostly manual or if CVO really start something wired :P

pkg/crc/oc/clusteroperator.go
```diff
@@ -47,31 +55,31 @@ func GetClusterOperatorStatus(oc OcConfig) (bool, error) {
...
+					cs.Processing = true
 				}
 			case "Upgradeable":
 				continue
 			case "Disabled":
-				if con.Status != "False" {
+				if con.Status == "True" {
 					logging.Debug(c.Metadata.Name, " operator is disabled, Reason: ", con.Reason)
-					allAvailable = false
+					cs.Disabled = true
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau for sake of each state I think better to have it, in case it required.

pkg/crc/oc/clusteroperator.go
cfergeau: I'd tend to be pessimistic ("if something is still progressing, the cluster is not fully running") rather than optimistic ("even if something is still progressing, let's declare the cluster running"). But that really is a cornercase

pkg/crc/oc/clusteroperator.go
cfergeau: I believe we discussed this, but I forgot what the conclusion was.

cfergeau:
Still 2 minor suggestions, but probably good either way

pkg/crc/oc/clusteroperator.go
praveenkumar: @cfergeau `ClusterStatus.Available  make more sense to ClusterStatus.AllAvailable since ClusterSTatus already mean the whole cluster state instead of a single operator state.` was the end of the discussion and we both agreed to it.

pkg/crc/oc/clusteroperator.go
praveenkumar: If something is processing and any bit is available then the cluster is running.

Issue #1066, opened 2020-03-05
Remove logging.Fatal from machine start


anjannath:
When a crc vm with a different bundle versions exists, we error out calling logging.Fatal which calls os.Exit() and makes the whole process exit, which is bad for the daemon. 

https://github.com/code-ready/crc/blob/3ba6f260e5276c814077706fde0789bc238ecfbf/pkg/crc/machine/machine.go#L187-L190

PR #1067, opened 2020-03-05
Issue #1066 Remove call to logging.Fatal from machine.Start

anjannath:
In case of the daemon, we don't want an error in machine.Start()
to exit the daemon process, logging.Fatal calls os.Exit which
kills the process

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -185,8 +185,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error loading bundle metadata: %v", err)
 		}
 		if bundleName != filepath.Base(startConfig.BundlePath) {
-			logging.Fatalf("Bundle '%s' was requested, but the existing VM is using '%s'",
+			logging.Debugf("Bundle '%s' was requested, but the existing VM is using '%s'",
 				filepath.Base(startConfig.BundlePath), bundleName)
+			result.Error = fmt.Sprintf("Bundle '%s' was requested, but the existing VM is using '%s'",
+				filepath.Base(startConfig.BundlePath), bundleName)
+			return *result, errors.New(result.Error)
 		}
 		vmState, err := host.Driver.GetState()
 		if err != nil {


pkg/crc/machine/machine.go
```diff
@@ -185,8 +185,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error loading bundle metadata: %v", err)
 		}
 		if bundleName != filepath.Base(startConfig.BundlePath) {
-			logging.Fatalf("Bundle '%s' was requested, but the existing VM is using '%s'",
+			logging.Debugf("Bundle '%s' was requested, but the existing VM is using '%s'",
 				filepath.Base(startConfig.BundlePath), bundleName)
+			result.Error = fmt.Sprintf("Bundle '%s' was requested, but the existing VM is using '%s'",
+				filepath.Base(startConfig.BundlePath), bundleName)
+			return *result, errors.Newf("Found a CodeReady Containers VM using bundle %s", filepath.Base(startConfig.BundlePath))
```
cfergeau: Is it intentional that the returned error is not "Bundle %s was requested but ..." ?

anjannath: Yes, i thought normal error would show a brief info and the debug log would give the details with the bundle names.

anjannath: If you think this needs to be changed to "Bundle %s was requested but ..." that is fine with me, it does make the message more clear.

cfergeau: I'm worried about what the user-visible message will be. At the moment it's "Bundle '%s' was requested, but the existing VM is using '%s'", which is somehow explicit as to what the problem is. If with this change we show "Found a CodeReady Containers VM using bundle %s", imo this is less explicit.

anjannath: Okay, let me change it back to the original message..

pkg/crc/machine/machine.go
```diff
@@ -185,7 +185,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error loading bundle metadata: %v", err)
 		}
 		if bundleName != filepath.Base(startConfig.BundlePath) {
-			logging.Fatalf("Bundle '%s' was requested, but the existing VM is using '%s'",
+			logging.Debugf("Bundle '%s' was requested, but the existing VM is using '%s'",
+				filepath.Base(startConfig.BundlePath), bundleName)
+			result.Error = fmt.Sprintf("Bundle '%s' was requested, but the existing VM is using '%s'",
+				filepath.Base(startConfig.BundlePath), bundleName)
+			return *result, errors.Newf("Bundle '%s' was requested, but the existing VM is using '%s'",
```
cfergeau: could be errors.Newf(result.Error) to avoid doing the fmt.Sprintf twice.

pkg/crc/machine/machine.go
```diff
@@ -185,8 +185,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 			return *result, errors.Newf("Error loading bundle metadata: %v", err)
 		}
 		if bundleName != filepath.Base(startConfig.BundlePath) {
-			logging.Fatalf("Bundle '%s' was requested, but the existing VM is using '%s'",
+			logging.Debugf("Bundle '%s' was requested, but the existing VM is using '%s'",
 				filepath.Base(startConfig.BundlePath), bundleName)
+			result.Error = fmt.Sprintf("Bundle '%s' was requested, but the existing VM is using '%s'",
+				filepath.Base(startConfig.BundlePath), bundleName)
+			return *result, errors.New(result.Error)
```

PR #1068, opened 2020-03-05
Remove openshift-marketplace operator proxy settings

cfergeau:
Since we are now using a bundle with the cluster-version-operator (CVO)
enabled, crc no longer needs to propagate the proxy settings itself as
CVO does it for the deployments with the `inject-proxy` annotation.

This reverts commits 8674e3, 75414b and f8441b

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -153,8 +153,6 @@ func AddProxyConfigToCluster(oc oc.OcConfig, proxy *network.ProxyConfig) error {
 	return nil
 }
 
-const internalNoProxy = ".cluster.local,.svc,10.128.0.0/14,172.30.0.0/16"
-
 // AddProxyToKubeletAndCriO adds the systemd drop-in proxy configuration file to the instance,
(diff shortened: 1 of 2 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -164,8 +162,8 @@ func AddProxyToKubeletAndCriO(sshRunner *ssh.SSHRunner, proxy *network.ProxyConf
 	proxyTemplate := `[Service]
 Environment=HTTP_PROXY=%s
 Environment=HTTPS_PROXY=%s
-Environment=NO_PROXY=%s,%s`
-	p := fmt.Sprintf(proxyTemplate, proxy.HttpProxy, proxy.HttpsProxy, internalNoProxy, proxy.GetNoProxyString())
+Environment=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,%s`
+	p := fmt.Sprintf(proxyTemplate, proxy.HttpProxy, proxy.HttpsProxy, proxy.GetNoProxyString())
```
praveenkumar: :+1:

PR #1072, opened 2020-03-06
Wait for proxy to propagated to cluster

praveenkumar:
Since we are not disabling the CVO so in case of proxy, CVO
take more time to propagate it to different operators, at this
moment we are checking operator status in every 2 sec (60 times)
till all are available.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -223,3 +223,19 @@ func DeleteOpenshiftApiServerPods(oc oc.OcConfig) error {
 
 	return errors.RetryAfter(60, deleteOpenshiftApiserverPods, time.Second)
 }
+
+func CheckProxySettingsForOperator(oc oc.OcConfig, proxy *network.ProxyConfig, deployment, namespace string) (bool, error) {
+	if !proxy.IsEnabled() {
+		logging.Debugf("No proxy in use")
+		return true, nil
+	}
+	cmdArgs := []string{"set", "env", "deployment", deployment, "--list", "-n", namespace}
+	out, _, err := oc.RunOcCommand(cmdArgs...)
+	if err != nil {
+		return false, err
+	}
+	if strings.Contains(out, proxy.HttpsProxy) || strings.Contains(out, proxy.HttpProxy) {
+		return true, nil
+	}
+	return false, nil
+}
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -415,6 +415,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, errors.Newf("Error approving the node csr %v", err)
 	}
 
+	if proxyConfig.IsEnabled() {
+		logging.Info("Waiting for the proxy configuration to be applied ...")
+		waitForProxyPropagation(ocConfig, proxyConfig)
+	}
+
 	// If no error, return usage message
 	if result.Error == "" {
 		logging.Infof("")
@@ -829,4 +834,23 @@ func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd
 	}
 
 	return nil
+}
+
+func waitForProxyPropagation(ocConfig oc.OcConfig, proxyConfig *network.ProxyConfig) {
+	checkProxySettingsForOperator := func() error {
+		proxySet, err := cluster.CheckProxySettingsForOperator(ocConfig, proxyConfig, "redhat-operators", "openshift-marketplace")
+		if err != nil {
+			logging.Debugf("Error getting proxy setting for openshift-marketplace operator %v", err)
+			return &errors.RetriableError{Err: err}
+		}
+		if !proxySet {
+			logging.Debug("Proxy changes for cluster in progress")
+			return &errors.RetriableError{Err: fmt.Errorf("")}
+		}
+		return nil
+	}
+
+	if err := errors.RetryAfter(60, checkProxySettingsForOperator, 2*time.Second); err != nil {
+		logging.Debug("Failed to propagate proxy settings to cluster")
+	}
 }


pkg/crc/machine/machine.go
```diff
@@ -415,6 +415,11 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return *result, errors.Newf("Error approving the node csr %v", err)
 	}
 
+	if proxyConfig.IsEnabled() {
+		logging.Info("Adding proxy details to cluster ...")
```
cfergeau: "Waiting for the proxy configuration to be applied" or something like this? "Adding" sounds like we are doing an action on th ecluster when we are only waiting for things to happen :)

pkg/crc/machine/machine.go
```diff
@@ -830,3 +835,22 @@ func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd
...
+		}
+		if !status {
+			logging.Debug("Proxy changes for cluster in progress")
+			return &errors.RetriableError{Err: fmt.Errorf("all operators are still not in available state")}
+		}
+		return nil
+	}
+
+	if err := errors.RetryAfter(60, clusterstatus, 2*time.Second); err != nil {
+		logging.Debug("Cluster is still in progress state after proxy is applied.")
```
cfergeau: "Proxy settings have not been applied yet"?

pkg/crc/machine/machine.go
```diff
@@ -830,3 +835,22 @@ func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd
...
+func checkProxyPropagatedToCluster(ocConfig oc.OcConfig, proxyConfig *network.ProxyConfig) {
+	clusterstatus := func() error {
+		status, err := cluster.CheckProxyIsPartOfAuthOperator(ocConfig, proxyConfig)
+		if err != nil {
+			logging.Debugf("Error getting the clusterstatus %v", err)
+			return &errors.RetriableError{Err: err}
+		}
+		if !status {
+			logging.Debug("Proxy changes for cluster in progress")
+			return &errors.RetriableError{Err: fmt.Errorf("all operators are still not in available state")}
```
cfergeau: Is this the right error message or does this need to be adjusted?

pkg/crc/machine/machine.go
```diff
@@ -830,3 +835,22 @@ func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd
 
 	return nil
 }
+
+func checkProxyPropagatedToCluster(ocConfig oc.OcConfig, proxyConfig *network.ProxyConfig) {
```
cfergeau: waitForProxyPropagation?

pkg/crc/machine/machine.go
```diff
@@ -830,3 +835,22 @@ func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd
...
+	clusterstatus := func() error {
+		status, err := cluster.CheckProxyIsPartOfAuthOperator(ocConfig, proxyConfig)
+		if err != nil {
+			logging.Debugf("Error getting the clusterstatus %v", err)
```
cfergeau: "clusterstatus" does not sound correct?

pkg/crc/cluster/cluster.go
```diff
@@ -223,3 +223,15 @@ func DeleteOpenshiftApiServerPods(oc oc.OcConfig) error {
 
 	return errors.RetryAfter(60, deleteOpenshiftApiserverPods, time.Second)
 }
+
+func CheckProxyIsPartOfAuthOperator(oc oc.OcConfig, proxy *network.ProxyConfig) (bool, error) {
```
cfergeau: `AuthOperator`? this is checking `openshift-marketplace`? I would pass the operator name as an argument to that method, and call it `CheckProxySettingsForOperator(...)`, `OperatorHasProxy`, ... or something like this?

pkg/crc/cluster/cluster.go
```diff
@@ -223,3 +223,15 @@ func DeleteOpenshiftApiServerPods(oc oc.OcConfig) error {
...
+	cmdArgs := []string{"set", "env", "deployment", "redhat-operators", "--list", "-n", "openshift-marketplace"}
+	out, _, err := oc.RunOcCommandPrivate(cmdArgs...)
+	if err != nil {
+		return false, err
+	}
+	if strings.Contains(out, proxy.HttpsProxy) {
```
cfergeau: What if only HttpProxy is set?

pkg/crc/machine/machine.go
```diff
@@ -830,3 +835,22 @@ func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd
 
 	return nil
 }
+
+func checkProxyPropagatedToCluster(ocConfig oc.OcConfig, proxyConfig *network.ProxyConfig) {
+	clusterstatus := func() error {
+		status, err := cluster.CheckProxyIsPartOfAuthOperator(ocConfig, proxyConfig)
```
cfergeau: `status` is not really accurate, `proxySet` maybe?

pkg/crc/cluster/cluster.go
```diff
@@ -223,3 +223,15 @@ func DeleteOpenshiftApiServerPods(oc oc.OcConfig) error {
 
 	return errors.RetryAfter(60, deleteOpenshiftApiserverPods, time.Second)
 }
+
+func CheckProxyIsPartOfAuthOperator(oc oc.OcConfig, proxy *network.ProxyConfig) (bool, error) {
+	cmdArgs := []string{"set", "env", "deployment", "redhat-operators", "--list", "-n", "openshift-marketplace"}
```
cfergeau: Maybe it's worth starting the method with `if !proxy.IsEnabled() { return true, nil }`, potentially with a `logging.Debugf` to know what's going on?

pkg/crc/machine/machine.go
praveenkumar: @cfergeau since here we are already check for `IsEnabled()` does it make sense to check https://github.com/code-ready/crc/pull/1072/files#r388997725 here?

pkg/crc/cluster/cluster.go
praveenkumar: Ah yes I have to put an or operator here, had in mind but just missed when putting the code :(

pkg/crc/cluster/cluster.go
praveenkumar: Ah I was testing Auth before but yes I will pass the namespace and deployment of operator as argument.

pkg/crc/machine/machine.go
cfergeau: We are adding a new exported `CheckProxyIsPartOfAuthOperator` method in the `cluster` package. It's currently only called from `machine.go`, but since it's part of a different package, maybe one day there will be more users calling it. Either we consider it's an error to call `CheckProxyIsPartOfAuthOperator` if `proxy.IsEnabled()` is false, in which case it would still be nice for `CheckProxyIsPartOfAuthOperator` to have a warning about it, or we consider it's ok to call `CheckProxyIsPartOfAuthOperator` if `proxy.IsEnabled()` is false, and then it should return true. In any case, I don't think we should rely on the rest of the code to have properly checked that proxy is enabled before calling that method.

pkg/crc/cluster/cluster.go
```diff
@@ -223,3 +223,15 @@ func DeleteOpenshiftApiServerPods(oc oc.OcConfig) error {
 
 	return errors.RetryAfter(60, deleteOpenshiftApiserverPods, time.Second)
 }
+
+func CheckProxySettingsForOperator(oc oc.OcConfig, proxy *network.ProxyConfig, deployment, namespace string) (bool, error) {
+	cmdArgs := []string{"set", "env", "deployment", deployment, "--list", "-n", namespace}
+	out, _, err := oc.RunOcCommandPrivate(cmdArgs...)
```
cfergeau: why `Private`?

pkg/crc/machine/machine.go
```diff
@@ -830,3 +835,22 @@ func configProxyForCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, sd
...
+		}
+		if !proxySet {
+			logging.Debug("Proxy changes for cluster in progress")
+			return &errors.RetriableError{Err: fmt.Errorf("")}
+		}
+		return nil
+	}
+
+	if err := errors.RetryAfter(60, checkProxySettingsForOperator, 2*time.Second); err != nil {
+		logging.Debug("Proxy settings have not been applied yet to cluster")
```
cfergeau: Sounds a bit like "the settings are not applied yet, but they are going to be applied later", which we don't know as we just failed to apply them. I would reword this to be less optismistic "Failed to propagate proxy settings to cluster" or something like this.

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau as part of output it list the proxy details.

praveenkumar:
retest this please

Issue #1060, opened 2020-03-04
[BUG] Set `admin` context before using the kubeconfig to interact with cluster


praveenkumar:
Recently we found out if a user by mistake do following then our kubeconfig which is part of the bundle gets different context and also current context set to a different cluster.

```
$ export KUBECONFIG=~/.crc/cache/<bundle>/kubeconfig
$ run kind <= way to deploy a single cluster
-- Check the contexts of the kubeconfig file --
$ oc config get-contexts  --kubeconfig ~/.crc/cache/<bundle>/kubeconfig
CURRENT   NAME                                      CLUSTER                AUTHINFO                          NAMESPACE
          admin                                     crc                    admin                             
          default/api-crc-testing:6443/kube:admin   api-crc-testing:6443   kube:admin/api-crc-testing:6443   default
*         kind-kind                                 kind-kind              kind-kind        
```

With this if a user now run the crc cluster then he might get something following
```
Unable to connect to the server: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"kubernetes\")"
```

As part of CRC we should always set the current context as `admin` even there is no other app consumed this kubeconfig to make sure it will not break.

cfergeau:
Fixed by https://github.com/code-ready/crc/commit/f91d978eb2a225a8cd70baddd75f3f7ea5a8b2ca

priyankakhetawat:
while running
crc start  facing the same issue
DEBU retry loop: attempt 11 out of 80
DEBU Running 'C:\Users\pkhetawat\.crc\bin\oc\oc.exe get csr --kubeconfig C:\Users\pkhetawat\.crc\machines\crc\kubeconfig --context admin --cluster crc'
DEBU Command failed: exit status 1
DEBU stdout:
DEBU stderr: Unable to connect to the server: dial tcp: lookup api.crc.testing: no such host
DEBU Unable to connect to the server: dial tcp: lookup api.crc.testing: no such host
DEBU error: Temporary Error: exit status 1 - sleeping 1s

windows version10 18363

PR #1101, opened 2020-03-13
Issue #1060 Add admin context to oc runner

praveenkumar:
If a user by mistake do following then the kubeconfig file which is part
of the bundle gets another entry for context and also current context
sets to a different cluster.

```
$ export KUBECONFIG=~/.crc/cache/<bundle>/kubeconfig
$ run kind <= way to deploy a single cluster
-- Check the contexts of the kubeconfig file --
$ oc config get-contexts  --kubeconfig ~/.crc/cache/<bundle>/kubeconfig
CURRENT   NAME                                      CLUSTER                AUTHINFO                          NAMESPACE
          admin                                     crc                    admin
          default/api-crc-testing:6443/kube:admin   api-crc-testing:6443   kube:admin/api-crc-testing:6443   default
*         kind-kind                                 kind-kind              kind-kind
```

With this kubeconfig file, if a user tries to run the crc cluster then
he might get something following.
```
Unable to connect to the server: x509: certificate signed by unknown authority (possibly because
of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"kubernetes\")"
```

This patch make sure each oc command run use `admin` context which is default for crc cluster.

### DIFF
--- a/pkg/crc/oc/oc.go
+++ b/pkg/crc/oc/oc.go
@@ -19,7 +19,9 @@ type OcRunner interface {
 }
 
 type OcConfig struct {
-	runner OcRunner
+	runner  OcRunner
+	Context string
+	Cluster string
 }
 
 type OcLocalRunner struct {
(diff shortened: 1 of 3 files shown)


pkg/crc/oc/oc.go
```diff
@@ -25,6 +26,7 @@ type OcConfig struct {
 type OcLocalRunner struct {
 	OcBinaryPath   string
 	KubeconfigPath string
+	Context        string
```
cfergeau: Have you considered putting this in `OcConfig` rather than `OcLocalRunner`? (but then `KubeconfigPath` would belong there too). Juts random thoughts at this point to avoid duplicating too much between `oc.go` and `cert-renewal.go`

pkg/crc/oc/oc.go
```diff
@@ -39,22 +41,27 @@ func (oc OcLocalRunner) GetKubeconfigPath() string {
...
 		OcBinaryPath:   filepath.Join(constants.CrcBinDir, constants.OcBinaryName),
 		KubeconfigPath: filepath.Join(constants.MachineInstanceDir, machineName, "kubeconfig"),
+		Context:        constants.DefaultContext,
 	}
 	return NewOcConfig(localRunner)
 }
 
 func (oc OcConfig) RunOcCommand(args ...string) (string, string, error) {
-	args = append(args, "--kubeconfig", oc.runner.GetKubeconfigPath())
+	args = append(args, "--kubeconfig", oc.runner.GetKubeconfigPath(), "--context", oc.runner.GetContext())
```
cfergeau: I believe we also want to add `--cluster crc` as this:
```
$ oc config get-contexts  --kubeconfig ~/.crc/cache/<bundle>/kubeconfig
CURRENT   NAME                                      CLUSTER                AUTHINFO                          NAMESPACE
          admin                                     crc                    admin
          default/api-crc-testing:6443/kube:admin   api-crc-testing:6443   kube:admin/api-crc-testing:6443   default
*         kind-kind                                 kind-kind              kind-kind
```
gives me the feeling that we could have an `admin` context in the `kind-kind` namespace/cluster too?

pkg/crc/oc/oc.go
```diff
@@ -45,21 +47,25 @@ func UseOCWithConfig(machineName string) OcConfig {
...
 
 func (oc OcConfig) RunOcCommandPrivate(args ...string) (string, string, error) {
-	args = append(args, "--kubeconfig", oc.runner.GetKubeconfigPath())
+	args = append(args, "--kubeconfig", oc.runner.GetKubeconfigPath(), "--context", oc.Context, "--cluster", oc.Cluster)
 	return oc.runner.RunPrivate(args...)
 }
 
-func NewOcConfig(runner OcRunner) OcConfig {
-	return OcConfig{runner: runner}
+func NewOcConfig(runner OcRunner, context string, clusterName string) OcConfig {
```
cfergeau: Is `context` something only valid in the context of `clusterName`? Ie should we have `clusterName` first in the argument list, and then `context`?

praveenkumar: @cfergeau yes both are independent options in the `oc options`

```
10:52 $ oc options 2>&1 | grep cluster
      --cluster='': The name of the kubeconfig cluster to use
10:52 $ oc options 2>&1 | grep context
      --context='': The name of the kubeconfig context to use
```

Issue #1132, opened 2020-03-31
[spike] Remove dns dependency on the host


praveenkumar:
Right now we are depending on different method on different platform to provide a dns resolver on the host to resolve the routes for crc `apps-crc.testing` and `api.crc.testing` . Current way of handling those depend on the platform and creating some issues which can be avoided if we completely remove it and depend on only `/etc/hosts` file using something like https://github.com/goodhosts/cli one.

cfergeau:
>  some issues which can be avoided
Can this be specific about the exact issues this spike wants to solve?
Using https://github.com/goodhosts/cli would introduce a usability regression, and a difference in behaviour compared with a regular openshift install.

gbraad:
Can you describe what is tried to be solved... and how the changed scenario woudl look like?
Especially, as we need to find a common approach if possible and when needed what the platform specific changes would be.

Especially on Windows, we need to avoid the use of UAC when possible. This is requesting to execute something as admin, which happens by a modal dialog blocking the user.

gbraad:
Also see #926; the current solution for Windows is not optimal and causes issues outside the use of CRC

gbraad:
#926 got resolved witn `cleanup`.

anjannath:
Part of this was done in #1136 for macOS and Linux the windows part remains.

cfergeau:
Only the basics were done, host DNS still needs to be configured for the user-created routes to work.

guillaumerose:
The three platforms are now using the admin helper (which includes goodhost lib).

Issue #926, opened 2020-01-09
[BUG] Windows : CRC negatively impacts DNS resolution when not running.


BarDweller:
### General information

  * OS:  Windows
  * Hypervisor:Hyper-V
  * Did you run `crc setup` before starting it YES

## CRC version
```bash
crc version: 1.3.0+918756b
OpenShift version: 4.2.10 (embedded in binary)
```
  
## CRC status
```bash
CRC VM:          Stopped
OpenShift:       Stopped
Disk Usage:      0B of 0B (Inside the CRC VM)
Cache Usage:     24.92GB
Cache Directory: C:\Users\Ozzy\.crc\cache
```

## CRC config
```bash
- cpus                                  : 6
- memory                                : 32768
```

## Host Operating System
```bash
OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.18362 N/A Build 18362
OS Manufacturer:           Microsoft Corporation
```

### Steps to reproduce

  1. Install crc.. do the whole crc setup, crc start, use crc for a bit, then do crc stop.
  2. Try using a web browser, or building with maven, notice all dns resolutions are now taking multiple seconds to resolve instead of being instant.
  3. do crc start (in my case .. crc start -n 8.8.8.8)
  4. Try using a web browser, or building with maven, notice all dns resolutions are now working as expected again.. 

### Expected

When CRC isn't running, it shouldn't be screwing up the platforms dns resolution. 

### Actual

CRC's adding of a DNS Server to the default virtual switch means every DNS lookup tries to go to the CRC ip (192.168.75.246 in my case) .. which is only available when CRC is running. This lookup adds noticeable seconds to DNS resolutions, and seems to cause issues with Maven/Docker builds that expect DNS lookups to occur in a timely manner. 

Removing the DNS entry via the network connections panel with start/run/ncpa.cpl and right clicking vEthernet (Default Switch), double click Internet Protocol Version TCP/IP v4, and removing the ip address from the 'Use the following DNS server addresses' box, restored normal DNS behavior to my system.

gbraad:
We are aware of this. Still looking into a solution that might not need the local DNS resolution configured on the interface.

BarDweller:
Any chance of an update in the docs somewhere to help mitigate the impact of this until it's resolved? It took me quite a while to figure out what CRC had done, so I could undo it. 

The impact is quite nasty, as it affects the developers machine even if they remove CRC.

BarDweller:
[...]

PR #1136, opened 2020-04-01
Issue #1132 Use /etc/hosts instead of dnsmasq

praveenkumar:
**Fixes:** Issue #1132 

## Solution/Idea

Idea here to remove the dependency on host side dns changes and rely on the `/etc/hosts` file.

## Proposed changes

There shouldn't be any changes the way we start the crc but now by default you can't open the web browser until this we add to the `/etc/hosts`, ( We need to add console and registry url along with api and outh which is why it is WIP).

## Testing

Also this is right now only tested on linux, this will work on the mac since that we are already using but we need more work for windows.


cfergeau:
This is missing several "big picture" details to fully grasp what this is doing :) Some of the commits (maybe later ones?) will need to detail *why* we are doing this change. Why are we using a separate binary rather than maintaining `/etc/hosts` ourselves? What will the UI like be when the user deploys a new application and wants to access it?

jeffsaremi:
Hi @praveenkumar @gbraad 
This is on Windows and I'm not sure if this change will take care of it or not?
Currently the installation process removes the list of default DNS servers on the "crc" external V-switch. We find this very intrusive. Is there a way through this change or other options to avoid this behavior?

thanks

gbraad:
We are looking into this. It is best to discuss on the main issue associated, the spike about the removal/change. We are gathering details about what this should look like

gbraad:
Also see #926

cfergeau:
> This is missing several "big picture" details to fully grasp what this is doing :) Some of the commits (maybe later ones?) will need to detail _why_ we are doing this change. Why are we using a separate binary rather than maintaining `/etc/hosts` ourselves? What will the UI like be when the user deploys a new application and wants to access it?

There is still no good overview of what the overall goal of this series is, how things are expected to work after it, why we are making this change, ...

pkg/crc/services/dns/dns_linux.go
```diff
@@ -1,12 +1,36 @@
...
+	for _, hostname := range hostnames {
+		if _, _, err := crcos.RunWithDefaultLocale(goodhostPath, "rm", hostname); err != nil {
+			return err
+		}
+		if _, _, err := crcos.RunWithDefaultLocale(goodhostPath, "add", instanceIP, hostname); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
cfergeau, line 39: Can the changes in `dns_linux.go` be put in a separate commit?

pkg/crc/services/dns/dns_linux.go
```diff
@@ -1,12 +1,36 @@
...
+		serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift")); err != nil {
+		result.Success = false
+		return *result, err
+	}
 	result.Success = true
 	return *result, nil
 }
+
+// updateHostsConfFile updates the host's /etc/hosts file with Instance IP.
+func updateHostsConfFile(instanceIP string, hostnames ...string) error {
```
cfergeau, line 27: Maybe just `updateHostsFile`?

pkg/crc/services/dns/dns_linux.go
```diff
@@ -12,7 +12,10 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
 	// We might need to set the firewall here to forward
 	// Update /etc/hosts file for host
 	if err := updateHostsConfFile(serviceConfig.IP, serviceConfig.BundleMetadata.GetAPIHostname(),
-		serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift")); err != nil {
+		serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift"),
+		serviceConfig.BundleMetadata.GetAppHostname("console-openshift-console"),
+		serviceConfig.BundleMetadata.GetAppHostname("default-route-openshift-image-registry"),
+		serviceConfig.BundleMetadata.GetAppHostname("foo")); err != nil {
```
cfergeau, line 18: Why is this linux specific? We'll want exactly the same thing on all platforms.
I don't think we should add `foo` here, but rather we should use the test for `foo` to indicate whether dnsmasq-based DNS is working or not. This will be useful when debugging networking issues.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -31,7 +31,9 @@ type resolverFileValues struct {
 func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *services.ServicePostStartResult) (services.ServicePostStartResult, error) {
 	// Update /etc/hosts file for host
 	if err := updateHostsConfFile(serviceConfig.IP, serviceConfig.BundleMetadata.GetAPIHostname(),
-		serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift")); err != nil {
+		serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift"),
+		serviceConfig.BundleMetadata.GetAppHostname("console-openshift-console"),
+		serviceConfig.BundleMetadata.GetAppHostname("default-route-openshift-image-registry")); err != nil {
```
cfergeau, line 36: There seems to be a mix of
1) change the existing code to make use of goodhosts-cli to update /etc/hosts
2) add console-openshift-console/default-route-openshift-image-registry to /etc/hosts

I would split this commit along these lines, and merge 1) with the next commit in this series ("Remove `/etc/hosts` file modification on MacOS"). Or replace this with a commit using the same code on linux and macos.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -104,3 +111,24 @@ func fixPodmanBinaryCached() error {
...
+}
+
+func fixGoodHostBinaryCached() error {
+	goodhostPath := filepath.Join(constants.CrcBinDir, constants.GoodhostBinaryName)
+	goodhost := goodhost.GoodhostCached{}
+	if err := goodhost.EnsureIsCached(); err != nil {
+		return fmt.Errorf("Unable to download goodhost remote binary %v", err)
+	}
+	logging.Debug("goodhost remote binary cached")
+	return setSuid(goodhostPath)
```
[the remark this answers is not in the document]
gbraad: shouldn't the binary be embedded?

pkg/crc/goodhost/goodhost_cache.go
```diff
@@ -0,0 +1,101 @@
...
+		return false
+	}
+	return true
+}
+
+func matchGoodhostBinaryName(filename string) bool {
+	return filepath.Base(filename) == constants.GoodhostBinaryName
+}
+
+func (goodhost *GoodhostCached) getgoodhost(destDir string) (string, error) {
```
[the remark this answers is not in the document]
gbraad: It should be Goodhosts to reflect the name of the library

pkg/crc/cache/cache.go
```diff
@@ -48,7 +48,7 @@ func CacheBinary(cache Cache, matchBinaryName func(string) bool) error {
 
 	for _, extractedFilePath := range extractedFiles {
 		finalBinaryPath := filepath.Join(outputPath, filepath.Base(extractedFilePath))
-		err = crcos.CopyFileContents(extractedFilePath, finalBinaryPath, 0500)
+		err = crcos.CopyFileContents(extractedFilePath, finalBinaryPath, 0555)
```
cfergeau, line 51: Maybe this should be done by `setSUID` instead?

pkg/crc/cache/cache.go
```diff
@@ -0,0 +1,58 @@
...
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/extract"
+	crcos "github.com/code-ready/crc/pkg/os"
+	"github.com/pkg/errors"
+)
+
+type Cache interface {
+	GetBinary(destDir string) (string, error)
```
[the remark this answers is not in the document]
cfergeau: True, `IsCached` is still needed, but can't this be all generic code? I'm not sure an interface is needed, just a type with the appropriate methods should do the trick?

pkg/extract/extract.go
```diff
@@ -139,10 +139,6 @@ func unzip(archive, target string, fileFilter func(string) bool) ([]string, erro
 			return nil, fmt.Errorf("%s: illegal file path", path)
 		}
 
-		if fileFilter != nil && !fileFilter(path) {
-			logging.Debugf("untar: Skipping %s", path)
-			continue
-		}
```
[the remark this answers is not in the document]
cfergeau: The current code is
```
// the target location where the dir/file should be created
		path := filepath.Join(targetDir, header.Name)
		if fileFilter != nil && !fileFilter(path) {
			logging.Debugf("untar: Skipping %s", path)
			continue
		}

		// check the file type
		switch header.Typeflag {
		// if it's a file create it
		case tar.TypeReg, tar.TypeGNUSparse:
			// tar.Next() will externally only iterate files, so we might have to create intermediate directories here
			if err = os.MkdirAll(filepath.Dir(path), 0750); err != nil {
				return err
			}
			file, err := os.OpenFile(path, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, header.FileInfo().Mode())
			if err != nil {
				return err
			}
			defer file.Close()

			// copy over contents
			if _, err := io.Copy(file, tarReader); err != nil {
				return err
			}
		}
```
There is a call to `mkdirAll(path)` before trying to create the file from the archive, is this not working?

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
cfergeau: This could be squashed in the appropriate commits for example https://github.com/cfergeau/crc/commit/1adb910b1214995e97eff5a532aa8cfea7046ead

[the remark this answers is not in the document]
cfergeau: On second though, `OcCache`, `PodmanCache`, ... are not needed so this can become https://github.com/cfergeau/crc/commit/3b1d32143a1ba749ab8130a1a7503eb737444819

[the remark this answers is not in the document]
praveenkumar: @cfergeau we also need to include the path because right now we are dumping everything in `~/.crc/bin` so this will again need a refactor when https://github.com/code-ready/crc/pull/1217 worked on.

pkg/extract/extract.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau no this not going to work, because if you check the current code the `path` here is got from `path := filepath.Join(targetDir, header.Name)` and `header.Name` only contain the file name not the complete path so in our current scenario we check first the `fileFilter(path)` and skip that directory creation if the binary is inside a sub-directory. 

Assume binary name is `oc` which we want to extract and it is available in the tarball like `A/B/oc` so as per current flow when tar.Next() hit the `B` then fileFilter check if the `B` is `oc` if not then continue without creating the directory `B` and then when tar.Next() counter to `oc` it identify the binary is `oc` and now it hit the ` os.MkdirAll(filepath.Dir(path), 0750)` where the `path` will only have `<target_dir>/oc` instead `<target_dir>/B/oc` and it will fail.

praveenkumar:
retest this please.

pkg/crc/cache/cache.go
```diff
@@ -23,6 +24,14 @@ func New(binarName string, archiveURL string, destDir string) *Cache {
...
 }
 
+func NewOcCache(destDir string) *Cache {
+	return New(constants.OcBinaryName, constants.GetOcUrl(), destDir)
+}
+
+func NewPodmanCache(destDir string) *Cache {
+	return New(constants.PodmanBinaryName, constants.GetPodmanUrl(), destDir)
+}
+
```
cfergeau, line 34: I think Gerard objected to having this in the generic `Cache` package, this could be moved to one of the preflight files for now.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -84,7 +84,7 @@ func setSuid(path string) error {
 	}
 
 	/* Can't do this before the chown as the chown will reset the suid bit */
-	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s", path)
+	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s,g+x", path)
```
cfergeau: Hmm this assumes that the file is u+x, this assumes the user running crc is part of the `wheel` group. Not sure we want to address this right now, but `setSuid` could probably still fail in some corner cases.

pkg/crc/cache/cache.go
```diff
@@ -32,6 +32,10 @@ func NewPodmanCache(destDir string) *Cache {
 	return New(constants.PodmanBinaryName, constants.GetPodmanUrl(), destDir)
 }
 
+func NewGoodhostCache(destDir string) *Cache {
+	return New(constants.GoodhostBinaryName, constants.GetGoodHostUrl(), destDir)
```
cfergeau, line 36: `Goodhost` or `GoodHost`? And this really should be `Goodhosts` or `GoodHosts`  (applies throughout the PR)

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -26,6 +26,12 @@ var genericPreflightChecks = [...]PreflightCheck{
 		fixDescription:   "Caching podman remote binary",
 		fix:              fixPodmanBinaryCached,
 	},
+	{
+		checkDescription: "Checking if goodhosts binary is cached",
+		check:            checkGoodhostBinaryCached,
+		fixDescription:   "Caching goodhosts binary",
+		fix:              fixGoodHostBinaryCached,
+	},
```
cfergeau, line 34: Do we want  to make it mandatory by default?

pkg/crc/preflight/preflight_checks_common.go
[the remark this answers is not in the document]
cfergeau: I think at the moment it's not being embedded.

pkg/crc/services/dns/dns_darwin.go
cfergeau: This also duplicates the linux code.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -144,35 +146,14 @@ func waitForNetwork() error {
...
-	}
-	return nil
-}
-
-func replaceHostname(hosts *goodhosts.Hosts, ipAddress string, hostname string) error {
-	logging.Debugf("Updating %s to %s in /etc/hosts file", hostname, ipAddress)
-	if hosts.HasHostname(hostname) {
-		if err := hosts.RemoveByHostname(hostname); err != nil {
+		if _, _, err := crcos.RunWithDefaultLocale(goodhostPath, "add", instanceIP, hostname); err != nil {
 			return err
```
cfergeau, line 155: Would probably be nicer (more readable) to have a helper to add/remove an entry to /etc/hosts.

pkg/crc/cache/cache.go
praveenkumar: @cfergeau I am going to use `Goodhost` throughout the PR.

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau I am incline to make it mandatory and since this is on our crc-repo side so we have full control over it.

pkg/crc/preflight/preflight_checks_common.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau it is part of embed.go in the same PR so it is going to be embedded.

pkg/crc/cache/cache.go
gbraad: I mind to have the values IN the cache package as that makes it non-generic.
if they exist and get set outside the cache they would make this fiel really generic and reusable without having to modify this file... and just from constansts or the preflight that uses it

praveenkumar:
> I'm still missing a clear description of what is the exact goal with this change. Is it just a workaround for when we fail to configure DNS so that the cluster at least starts?

@cfergeau Yes, that is the initial plan for this PR.

>  Do we plan for this to replace DNS based configuration?

This PR doesn't replace the existing DNS based configuration but just act as an addon so if the DNS configuration fail then at least cluster start and user can able to login using cli or webconsole.

>What are the plans for adding/removing additional entries for user apps?

At this moment adding/removing user app or any other route is not scope of this PR, in future we can have a `crc route add/remove` command for this (but as I said this is not scope for this PR and shouldn't block for this usecase.)

cfergeau:
> > Do we plan for this to replace DNS based configuration?
> 
> This PR doesn't replace the existing DNS based configuration but just act as an addon so if the DNS configuration fail then at least cluster start and user can able to login using cli or webconsole.
> 

My feeling is that we are basically moving the failure from `crc start` to a later point when the user is trying to create an application. I guess we should at least add a warning that we are running in a degraded state, and that application routes are not going to automatically work?

pkg/crc/cache/cache.go
cfergeau: `git grep  goodhost[^s]` still returns some results

pkg/crc/cache/cache.go
cfergeau: This does not seem to be addressed in the latest iteration? Is there nothing to address?

pkg/crc/preflight/preflight_checks_common.go
cfergeau: I realized it can't be skipped at all, so we'll always have the 2 DNS configuration methods running in parallel. Enabling it by default or not is one thing, but I really think it should be possible to disable it entirely, at least for debugging/testing purposes. Then you will have people who don't want (can't) run a 5.5MB suid binary.

cfergeau:
> > > Do we plan for this to replace DNS based configuration?
> > 
> > 
> > This PR doesn't replace the existing DNS based configuration but just act as an addon so if the DNS configuration fail then at least cluster start and user can able to login using cli or webconsole.
> 
> My feeling is that we are basically moving the failure from `crc start` to a later point when the user is trying to create an application. I guess we should at least add a warning that we are running in a degraded state, and that application routes are not going to automatically work?

Overall the patch series looks good wrt achieving its goal. However I'm really dubious about not being able to disable this at all, especially as it will make bugs in the 'main' implementation harder to notice, harder to debug, ...

Issue #1141, opened 2020-04-06
[BUG] crc setup not able to update the plist file for a different release


praveenkumar:
### General information

  * OS: macOS
  * Hypervisor: hyperkit
  * Did you run `crc setup` before starting it (Yes/No)? Yes

When a user use a different version of crc and runs `crc setup  --enable-experimental-features` then expected scenario is crc daemon should get updated and have info about the version which the user try to use instead it will just use the older version.

Reproduce:
- Download 1.7.0 version of CRC on fresh system (no tray and daemon)
- Run `crc setup  --enable-experimental-features`
- Check the about section in the tray ( It will show correct version of CRC and bundle)
- Delete the cluster
- Download 1.8.0 and repeat those steps and again check `About` section in the tray and it is still going to point from `1.7.0` release.

In our check we just see if plist file exists and doesn't check the path string.
```
	<array>
				<string>/Users/prkumar/Downloads/crc-1.7.0/crc</string>
				<string>daemon</string>
				<string>--log-level</string>
				<string>debug</string>
	</array>
```

gbraad:
Likely the executable was in use and therefore you see the old version. The setup should detect a running instance, stop it and then replace before starting it again. (this was mentioned as a scenario in on of the comments on a PR).

anjannath:
We do that for the tray, as in we check if the version running is the same as the one requested, if not then we restart. In case of the daemon am thinking to always create the plist file with the current executable's path and restart it whenever setup is performed.

gbraad:
> always create the plist file with the current executable's path 

this would be missing... but best would be to 'stop' first, as you do not want the executable to remain in memory or have a filelock due to it being active.

praveenkumar:
fixed by #1142

Issue #1140, opened 2020-04-06
[BUG] daemon and tray silently fail to load as part of setup if `launchctl remove` used


praveenkumar:
### General information

  * OS:  macOS 
  * Hypervisor: hyperkit
  * Did you run `crc setup` before starting it (Yes/No)? yes

If a user use `launchctl remove crc.daemon` or `launchctl remove crc.tray` then setup command is not able to start those launch agents without any error or warning.

gbraad:
> start those launch agents without any error or warning.

not sure how this should be read. Can you show the error in some way... or you suggest there is no error, and therefore silently fails?

praveenkumar:
> not sure how this should be read. Can you show the error in some way... or you suggest there is no error, and therefore silently fails?

If you try following, you expect to have tray icon and daemon running after the setup command which is not happening and also there is no error from the setup command which checks for the daemon and tray.

```
$  launchctl remove crc.daemon
$  launchctl remove crc.tray

$ ./crc setup --enable-experimental-features --log-level debug
INFO Checking if oc binary is cached              
DEBU oc binary already cached                     
INFO Checking if podman remote binary is cached   
DEBU podman remote binary already cached          
INFO Checking if goodhost binary is cached        
DEBU goodhost remote binary already cached        
INFO Checking if CRC bundle is cached in '$HOME/.crc' 
INFO Checking if running as non-root              
INFO Checking if HyperKit is installed            
DEBU Checking if hyperkit is installed            
INFO Checking if crc-driver-hyperkit is installed 
DEBU Checking if crc-driver-hyperkit is installed 
DEBU Running '/Users/prkumar/.crc/bin/crc-driver-hyperkit version' 
DEBU crc-driver-hyperkit is already installed in /Users/prkumar/.crc/bin/crc-driver-hyperkit 
INFO Checking file permissions for /etc/resolver/testing 
INFO Checking if tray binary is installed         
INFO Checking if launchd configuration for daemon exists 
INFO Checking if launchd configuration for tray exists 
INFO Checking installed tray version              
INFO Checking if CodeReady Containers daemon is running 
INFO Check if CodeReady Containers tray is running 
Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster
```

gbraad:
> > silently fails

> there is no error 

👍

anjannath:
So this is due to two things..
[...]

PR #1142, opened 2020-04-07
Make tray setup work when agents are removed and setup is ran from a different crc binary

anjannath:
**Fixes:** Issue #1140 #1141 

## Solution/Idea

1. When daemon/tray agent is removed using the `launchctl remove` command, `crc setup` was unable to restart it, since the plist file was removed and running `launchctl start` requires the agent to be loaded, therefore we load the agent plist before trying to start it as part of preflight check.
2. When the plist file already exists `crc setup` wouldn't recreate the daemon plist file with the current crc binary's path, so we make the `DaemonPlistFileExists` check to always fail and trigger the recreation of the plist file whenever setup is ran.

## Proposed changes
1. Load plist before starting it
2. Improve `launchd.AgentRunning` function to handle when an agent is removed from launchd
3. Force creation of daemon plist every time `crc setup` is ran

## Testing

1. run `launchctl remove crc.tray; launchctl remove crc.daemon`
2. Now when `crc setup --enable-experimental-fetures` is ran you should have a working tray
3. Download an older version of crc (1.6.0) and enable the tray using `crc setup --enable-experimental-features`
4. Check the crc version from the tray by clicking on _About_
5. Now run `crc setup --enable-experimetal-features` from the crc binary built from this PR.
6. Again check the version like in step 4, and the commit sha from the crc version should match the commit id from HEAD of this patch set. or just check inside `~/Library/LaunchAgents/crc.daemon.plist` that the first string in the `ProgramArguments` key is the path to the crc binary used in step 5.

### DIFF
--- a/pkg/crc/preflight/preflight_check_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_check_tray_darwin.go
@@ -79,6 +79,9 @@ func checkIfDaemonAgentRunning() error {
 }
 
 func fixDaemonAgentRunning() error {
+	if err := launchd.LoadPlist(daemonAgentLabel); err != nil {
+		return err
+	}
 	return launchd.StartAgent(daemonAgentLabel)
 }
 
@@ -90,6 +93,9 @@ func checkIfTrayAgentRunning() error {
 }
 
 func fixTrayAgentRunning() error {
+	if err := launchd.LoadPlist(trayAgentLabel); err != nil {
+		return err
+	}
 	return launchd.StartAgent(trayAgentLabel)
 }
 
--- a/pkg/crc/preflight/preflight_check_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_check_tray_darwin.go
@@ -86,6 +86,7 @@ func checkIfDaemonAgentRunning() error {
 }
 
 func fixDaemonAgentRunning() error {
+	logging.Debug("Starting daemon agent")
 	if err := launchd.LoadPlist(daemonAgentLabel); err != nil {
 		return err
 	}
@@ -100,6 +101,7 @@ func checkIfTrayAgentRunning() error {
 }
 
 func fixTrayAgentRunning() error {
+	logging.Debug("Starting tray agent")
 	if err := launchd.LoadPlist(trayAgentLabel); err != nil {
 		return err
 	}
@@ -130,6 +132,7 @@ func checkTrayVersion() error {
 }
 
 func fixTrayVersion() error {
+	logging.Debug("Downloading/extracting tray binary")
 	// get the tray app
 	err := downloadOrExtractTrayApp()
 	if err != nil {
@@ -146,6 +149,7 @@ func checkTrayBinaryPresent() error {
 }
 
 func fixTrayBinaryPresent() error {
+	logging.Debug("Downloading/extracting tray binary")
 	return downloadOrExtractTrayApp()
 }
 
--- a/pkg/crc/preflight/preflight_check_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_check_tray_darwin.go
@@ -35,13 +35,20 @@ type TrayVersion struct {
 }
 
 func checkIfDaemonPlistFileExists() error {
-	if !launchd.PlistExists(daemonAgentLabel) {
-		return fmt.Errorf("Daemon plist file does not exist")
-	}
-	return nil
+	// crc setup can be ran from any location in the
+	// users computer, and we need to update the plist
+	// file with the path of the crc binary which was
+	// used to run setup, to force it this check needs
+	// to always fail so the fix routine is triggered
+
+	return fmt.Errorf("Ignoring this check and triggering creation of daemon plist")
 }
 
 func fixDaemonPlistFileExists() error {
+	// Try to remove the daemon agent from launchd
+	// and recreate its plist
+	_ = launchd.Remove(daemonAgentLabel)
+
 	currentExecutablePath, err := goos.Executable()
 	if err != nil {
 		return err
--- a/pkg/os/launchd/launchd_darwin.go
+++ b/pkg/os/launchd/launchd_darwin.go
@@ -128,3 +128,8 @@ func AgentRunning(label string) bool {
 	}
 	return false
 }
+
+// Remove removes the agent from launchd
+func Remove(label string) error {
+	return exec.Command("launchctl", "remove", label).Run() // #nosec G204
+}
--- a/pkg/os/launchd/launchd_darwin.go
+++ b/pkg/os/launchd/launchd_darwin.go
@@ -7,6 +7,7 @@ import (
 	goos "os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"text/template"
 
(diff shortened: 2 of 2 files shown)


pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -35,13 +35,22 @@ type TrayVersion struct {
...
-		return fmt.Errorf("Daemon plist file does not exist")
-	}
-	return nil
+	// crc setup can be ran from any location in the
+	// users computer, and we need to update the plist
+	// file with the path of the crc binary which was
+	// used to run setup, to force it this check needs
+	// to always fail so the fix routine is triggered
+
+	return fmt.Errorf("Daemon plist file does not exist")
```
praveenkumar, line 44: If this check always going to return the `error` then what is the use of this check we should just remove it.

anjannath: Ohh, that'd be nice, this is just there to trigger the fix, can we make the fix run without having the check..?

praveenkumar: @anjannath why not, may be you can change the name without `fix` prefix and run it for every single time?

anjannath: @praveenkumar In that way we have to completely move this out of the preflight checks, now we need this to fill the `preflightCheck` struct, if we leave out the check member in that struct, i think we'll hit a segfault.

pkg/os/launchd/launchd_darwin.go
```diff
@@ -103,7 +104,11 @@ func StopAgent(label string) error {
 
 // RestartAgent restarts a launchd agent
 func RestartAgent(label string) error {
-	err := StopAgent(label)
+	err := LoadPlist(label)
```
cfergeau, line 107: Not clear why we need this change, if the agent was already running, then wasn't the plist also loaded already?

pkg/os/launchd/launchd_darwin.go
```diff
@@ -120,8 +125,8 @@ func AgentRunning(label string) bool {
 	if err != nil {
 		return false
 	}
-	if strings.TrimSpace(string(out)) == "-" {
-		return false
+	if match, err := regexp.MatchString(`^\d*$`, strings.TrimSpace(string(out))); err == nil && match {
```
cfergeau, line 128: I'd add a small comment such as "match pid". 
`^\d*$` will match an empty string, I don't think we want that if we're looking for a pid.

pkg/os/launchd/launchd_darwin.go
```diff
@@ -7,6 +7,7 @@ import (
 	goos "os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
```
cfergeau, line 11: In the commit log of " Issue #1140 Improve AgentRunning function " can you change "o/p" to "output"? took me a while to expand it :)

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -42,6 +42,7 @@ func checkIfDaemonPlistFileExists() error {
 }
 
 func fixDaemonPlistFileExists() error {
```
cfergeau, line 47: Commit log is "Issue #1140 Load plist file before trying to start a launchd agent", feel free to detail why this is needed, why this was working before without this, ...

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -63,6 +64,7 @@ func checkIfTrayPlistFileExists() error {
 }
 
 func fixTrayPlistFileExists() error {
+	logging.Debug("Creating tray plist file")
```
cfergeau, line 75: the addition of debugging statements could go in its own commit.

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -79,6 +81,10 @@ func checkIfDaemonAgentRunning() error {
 }
 
 func fixDaemonAgentRunning() error {
+	logging.Debug("Starting daemon agent")
+	if err := launchd.LoadPlist(daemonAgentLabel); err != nil {
+		return err
+	}
 	return launchd.StartAgent(daemonAgentLabel)
```
cfergeau, line 96: almost tempted to suggest using a `fixAgentRunnig` helper

pkg/crc/preflight/preflight_check_tray_darwin.go
cfergeau: I don't know if it's expensive to recreate this plist file, or if it has side effects we want to avoid, but an alternative could be to change `checkIfDaemonPlistFileExists` to 
- generate a string containing to the content the plist file should have for the current binary
- compare that string with the content of the plist file we have on disk
- only return an error if this does not match
If recreating the plist file is cheap enough, then maybe this is overkill..

pkg/os/launchd/launchd_darwin.go
anjannath: Ahh, thanks so much, should be `^\d+$`

pkg/crc/preflight/preflight_check_tray_darwin.go
anjannath: i didn't notice an increase in time consumed, so its very cheap to create it, happens near instantly

cfergeau: it's not just file creation, but file creation + LoadPlist + RestartAgent which I was asking if there are noticeable side effects.

pkg/os/launchd/launchd_darwin.go
anjannath: This was a left over from debug/testing, fixed now!

pkg/os/launchd/launchd_darwin.go
anjannath: Done!

pkg/crc/preflight/preflight_check_tray_darwin.go
anjannath: oh, I did not notice any side effects, seems to work as expected

pkg/os/launchd/launchd_darwin.go
```diff
@@ -113,15 +114,17 @@ func RestartAgent(label string) error {
 // AgentRunning checks if a launchd service is running
 func AgentRunning(label string) bool {
 	// This command return a PID if the process
-	// is running, otherwise returns "-"
+	// is running, otherwise returns "-" and empty
+	// ouput if the agent is not loaded in launchd
```
cfergeau, line 118: "ouput" -> "output"
and I think I would say "or empty output" instead of "and .." since we'll either get "-" if loaded but not running, or "" if not loaded.

pkg/crc/preflight/preflight_check_tray_darwin.go
praveenkumar: @cfergeau @anjannath I would still think if we are going to keep it then better to update the error message because  `Daemon plist file does not exist` seems confusing to me, what about just put `Ignoring this check` as error message and explanation already part of comments?

pkg/crc/preflight/preflight_check_tray_darwin.go
```diff
@@ -35,13 +35,20 @@ type TrayVersion struct {
...
-		return fmt.Errorf("Daemon plist file does not exist")
-	}
-	return nil
+	// crc setup can be ran from any location in the
+	// users computer, and we need to update the plist
+	// file with the path of the crc binary which was
+	// used to run setup, to force it this check needs
+	// to always fail so the fix routine is triggered
+
+	return fmt.Errorf("Ignoring this check and triggering creation of daemon plist")
```
praveenkumar, line 44: 👍

Issue #1147, opened 2020-04-08
Move docs modules to `docs/source/topics`


robin-owen:
Currently, the modules for documentation exist in the `docs/source/getting_started/content` directory, which implies that modules are only usable within the Getting Started Guide. Per modular documentation guidelines, these should be placed in a more globally-accessible location in case we introduce more books/guides in the future which also require the use of individual modules. For this reason, we should move all of the modules to a new `docs/source/topics` directory.

robin-owen:
Fixed by #1148.

PR #1148, opened 2020-04-08
Issue #1147 Move docs modules to accessible directory

robin-owen:
Per modular documentation guidelines, modules should exist in an
accessible location for potential use in multiple guides.

**Fixes:** Issue #1147


docs/source/topics/proc_starting-monitoring-alerting-telemetry.adoc
```diff
@@ -20,22 +20,22 @@ For more information, see <<configuring-the-virtual-machine_{context}>>.
 
 ** On Linux or {mac}:
 +
-[subs="+quotes,attributes"]
+[subs="+quotes"]
```
cfergeau, line 23: These ",attributes" removals seem unrelated, are they intentional?

robin-owen: Yes, they're intentional, but only because I happened to notice them while building the docs to test how they build with the modules moved around.

The inclusion of "attributes" here causes warnings while building the docs about undefined attributes (the `{end}` portions of the source blocks are interpreted as attributes by AsciiDoctor).

I can move these changes into their own dedicated commit if you'd prefer.

cfergeau: I don't feel that strongly about the separate commit (even if that's better ;), I wanted to make sure this was not an oversight.

PR #1153, opened 2020-04-13
Issue #507 Implement cleanup feature for macos

praveenkumar:
**Fixes:** Issue #507 

## Solution/Idea

Implement the cleanup for macos so user can run `crc cleanup` and it removes the file which created by setup command including the tray.

## Proposed changes

1. Add `cleanup` command.

## Testing

Run `crc cleanup` command and it should remove the files and change the file permission back to original user which created by `crc setup` command.

### DIFF
--- a/pkg/crc/preflight/preflight_darwin.go
+++ b/pkg/crc/preflight/preflight_darwin.go
@@ -2,8 +2,6 @@ package preflight
 
 import (
 	"fmt"
-
-	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
 // SetupHost performs the prerequisite checks and setups the host to run the cluster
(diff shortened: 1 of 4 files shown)


anjannath:
Not related to this PR but to cleanup feature, i think the cleanups should also do the checks, if we perform the checks in `doPreflightCleanUp` and skip over the ones that fails, we'll only do cleanup for the needed ones, now even when tray is not installed, it tries to unload and remove it.

praveenkumar:
> Not related to this PR but to cleanup feature, i think the cleanups should also do the checks, if we perform the checks in doPreflightCleanUp and skip over the ones that fails, we'll only do cleanup for the needed ones, now even when tray is not installed, it tries to unload and remove it.

@anjannath that is by design, we (me and @cfergeau ) had same conversation when implementing it for linux, we don't care if something is present or not but our cleanup stuff should run for each steps even a skip one.

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -71,18 +77,22 @@ var traySetupChecks = [...]PreflightCheck{
...
-		check:            checkIfDaemonAgentRunning,
-		fixDescription:   "Starting CodeReady Containers daemon",
-		fix:              fixDaemonAgentRunning,
-		flags:            SetupOnly,
+		checkDescription:   "Checking if CodeReady Containers daemon is running",
+		check:              checkIfDaemonAgentRunning,
+		fixDescription:     "Starting CodeReady Containers daemon",
+		fix:                fixDaemonAgentRunning,
+		flags:              SetupOnly,
+		cleanupDescription: "Unload CodeReady Containers daemon",
```
cfergeau, line 85: Unloading?

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -50,18 +52,22 @@ var traySetupChecks = [...]PreflightCheck{
...
-		check:            checkIfDaemonPlistFileExists,
-		fixDescription:   "Creating launchd configuration for daemon",
-		fix:              fixDaemonPlistFileExists,
-		flags:            SetupOnly,
+		checkDescription:   "Checking if launchd configuration for daemon exists",
+		check:              checkIfDaemonPlistFileExists,
+		fixDescription:     "Creating launchd configuration for daemon",
+		fix:                fixDaemonPlistFileExists,
+		flags:              SetupOnly,
+		cleanupDescription: "Removing launchd configuration for daemon",
```
cfergeau, line 60: This is consistent with the other messages, but maybe they should all be changed to  `CodeReady Containers daemon`

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -117,6 +127,12 @@ func RegisterSettings() {
 }
 
 func CleanUpHost() {
-	logging.Warn("Cleanup is not supported for MacOS")
-	doCleanUpPreflightChecks(getPreflightChecks())
+	// A user can use setup with experiment flag
```
cfergeau, line 130: the flag is named 'experimental'.
I guess we could name it in full `with the --enable-experimental-features` flag

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -117,6 +127,12 @@ func RegisterSettings() {
...
+	// and not use cleanup with same flag, to avoid
```
cfergeau, line 131: `cleanup` does not have an 'experimental' flag, does it?

PR #1154, opened 2020-04-13
Misc 4.4 changes

praveenkumar:
4.4 bundles specific change to run it as sane as possible.

### DIFF
--- a/pkg/crc/services/dns/template.go
+++ b/pkg/crc/services/dns/template.go
@@ -20,7 +20,7 @@ address=/{{ .AppsDomain }}/{{ .IP }}
 address=/etcd-0.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
 address=/api.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
 address=/api-int.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
-address=/{{ .Hostname }}.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
+address=/{{ .Hostname }}.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .InternalIP }}
 `
 )
 
@@ -31,6 +31,7 @@ type dnsmasqConfFileValues struct {
 	Hostname    string
 	IP          string
 	AppsDomain  string
+	InternalIP  string
 }
 
 func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error {
@@ -43,9 +44,10 @@ func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error
 		AppsDomain:  serviceConfig.BundleMetadata.ClusterInfo.AppsDomain,
 		ClusterName: serviceConfig.BundleMetadata.ClusterInfo.ClusterName,
 		IP:          serviceConfig.IP,
+		InternalIP:  serviceConfig.BundleMetadata.Nodes[0].InternalIP,
 	}
 
-	dnsConfig, err := createDnsConfigFile(dnsmasqConfFileValues)
+	dnsConfig, err := createDnsConfigFile(dnsmasqConfFileValues, dnsmasqConfTemplate)
 	if err != nil {
 		return err
 	}
@@ -53,10 +55,10 @@ func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error
 	return serviceConfig.SSHRunner.CopyData([]byte(dnsConfig), "/var/srv/dnsmasq.conf")
 }
 
-func createDnsConfigFile(values dnsmasqConfFileValues) (string, error) {
+func createDnsConfigFile(values dnsmasqConfFileValues, tmpl string) (string, error) {
 	var dnsConfigFile bytes.Buffer
 
-	t, err := template.New("dnsConfigFile").Parse(dnsmasqConfTemplate)
+	t, err := template.New("dnsConfigFile").Parse(tmpl)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/crc/services/dns/template.go
+++ b/pkg/crc/services/dns/template.go
@@ -2,8 +2,6 @@ package dns
 
 import (
 	"bytes"
-	"encoding/base64"
-	"fmt"
 	"text/template"
 
 	"github.com/code-ready/crc/pkg/crc/services"
@@ -52,14 +50,7 @@ func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error
 		return err
 	}
 
-	encodeddnsConfig := base64.StdEncoding.EncodeToString([]byte(dnsConfig))
-	_, err = serviceConfig.SSHRunner.Run(
-		fmt.Sprintf("echo '%s' | openssl enc -base64 -d | sudo tee /var/srv/dnsmasq.conf > /dev/null",
-			encodeddnsConfig))
-	if err != nil {
-		return err
-	}
-	return nil
+	return serviceConfig.SSHRunner.CopyData([]byte(dnsConfig), "/var/srv/dnsmasq.conf")
 }
 
 func createDnsConfigFile(values dnsmasqConfFileValues) (string, error) {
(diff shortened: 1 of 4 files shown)


praveenkumar:
retest this please.

jsliacan:
Makefile still has `4.3.8` bundle version in it.

pkg/crc/services/dns/dns.go
```diff
@@ -40,6 +41,26 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) (services.Servi
...
+	}
+	// Todo: Move those to snc script when start building for 4.4 bundles
+	// Add node-ip field in kublet service and set it to the static IP `192.168.126.11`
+	if _, err := serviceConfig.SSHRunner.Run(`sudo sed -i.back '/kubelet /a\      \--node-ip=192.168.126.11 \\' /etc/systemd/system/kubelet.service`); err != nil {
+		result.Success = false
+		result.Error = err.Error()
+		return *result, err
+	}
+	// Reload the kubelet service since the service changed happen
+	if _, err = serviceConfig.SSHRunner.Run("sudo systemctl daemon-reload"); err != nil {
```
cfergeau, line 58: This is calling "daemon-reload", not I'm not sure this triggers a restart of `kubelet.service` if it changed?

pkg/crc/services/dns/dns.go
```diff
@@ -10,11 +10,12 @@ import (
...
 
 const (
-	dnsServicePort              = 53
-	dnsConfigFilePathInInstance = "/var/srv/dnsmasq.conf"
-	dnsContainerIP              = "10.88.0.8"
-	dnsContainerImage           = "quay.io/crcont/dnsmasq:latest"
-	publicDNSQueryURI           = "quay.io"
+	dnsServicePort                  = 53
+	dnsConfigFilePathInInstance     = "/var/srv/dnsmasq.conf"
+	dnsHostConfigFilePathInInstance = "/var/srv/hosts.openshift"
```
cfergeau, line 15: This change belongs to the other commit I think ?

pkg/crc/services/dns/dns.go
```diff
@@ -40,6 +41,26 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) (services.Servi
 	// Remove the dnsmasq container if it exists during the VM stop cycle
 	_, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq")
 
+	// Add a virtual dummy network interface and set the IP as `192.168.126.11` which is the IP used to create the bundle
```
cfergeau, line 44: Shouldn't we add this IP in the bundle metadata file rather than hardcoding it?

pkg/crc/services/dns/dns.go
```diff
@@ -40,6 +41,26 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) (services.Servi
...
+	if _, err := serviceConfig.SSHRunner.Run("sudo ip link add eth10 type dummy && sudo ip addr add 192.168.126.11/24 dev eth10"); err != nil {
+		result.Success = false
+		result.Error = err.Error()
+		return *result, err
+	}
+	// Todo: Move those to snc script when start building for 4.4 bundles
+	// Add node-ip field in kublet service and set it to the static IP `192.168.126.11`
```
cfergeau, line 51: So this PR will be changed before we release for 4.4 ?

pkg/crc/services/dns/template.go
```diff
@@ -22,7 +22,10 @@ address=/{{ .AppsDomain }}/{{ .IP }}
 address=/etcd-0.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
 address=/api.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
 address=/api-int.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
-address=/{{ .Hostname }}.{{ .ClusterName}}.{{ .BaseDomain }}/{{ .IP }}
+addn-hosts=/etc/hosts.openshift
+`
+	dnsmasqHostsTemplate = `192.168.126.11 {{ .Hostname }}.{{ .ClusterName}}.{{ .BaseDomain }}
+{{ .IP }} {{ .Hostname }}.{{ .ClusterName}}.{{ .BaseDomain }}
```
cfergeau, line 28: This has to be done in a hosts file, this can't be done through regular DNS?

pkg/crc/services/dns/template.go
```diff
@@ -59,13 +62,27 @@ func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error
...
+			encodeddnsConfig))
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
-func createDnsConfigFile(values dnsmasqConfFileValues) (string, error) {
+func createDnsConfigFile(values dnsmasqConfFileValues, temp string) (string, error) {
```
cfergeau, line 82: `template` instead of `temp`, I read it as `temporary`

pkg/crc/services/dns/dns.go
praveenkumar: @cfergeau you mean the comment right?

cfergeau: Ah, I was a bit confused by the comment, but also because I did not look at the code (kubelet is not running at this point)
`Reload the kubelet systemd unit file since it was modified`?

cfergeau: Isn't this going to be adding a `--node-ip=192.168.126.11` line each time we run `crc start` though?

pkg/crc/services/dns/dns.go
gbraad: why change? this would not effect 4.3, ...

cfergeau: I'm referring to the comment which says some of the changes in this PR really belongs to snc. So I'm asking whether this is going to be changed before this PR is merged or not.

gbraad: right. but if this is part of SNC i would fail to see how this sruvives a 'reboot' without creating a config file or custom systemd unit.

pkg/crc/services/dns/template.go
praveenkumar: @cfergeau regular dns you mean the dnsmasq config? that is not possible because for same domain if you need different ip you need to use the host file.

pkg/crc/services/dns/template.go
praveenkumar: @cfergeau `template` is name of imported module so I would go for `tmpl` :)

pkg/crc/services/dns/template.go
```diff
@@ -59,13 +64,27 @@ func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error
...
+	dnsHostConfig, err := createDnsConfigFile(dnsmasqConfFileValues, dnsmasqHostsTemplate)
+	if err != nil {
+		return err
+	}
+
+	encodeddnsConfig = base64.StdEncoding.EncodeToString([]byte(dnsHostConfig))
+	_, err = serviceConfig.SSHRunner.Run(
+		fmt.Sprintf("echo '%s' | openssl enc -base64 -d | sudo tee /var/srv/hosts.openshift > /dev/null",
+			encodeddnsConfig))
+	if err != nil {
```
cfergeau, line 77: Instead of duplicating this a 3rd time, can we add a helper?
https://github.com/cfergeau/crc/commit/51a0ad953873d2f8375a52a4addfdba203ab17f5

pkg/crc/services/dns/template.go
```diff
@@ -43,20 +44,21 @@ func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error
...
+	dnsConfig, err := createDnsConfigFile(dnsmasqConfFileValues, dnsmasqConfTemplate)
 	if err != nil {
 		return err
 	}
 
 	return serviceConfig.SSHRunner.CopyData([]byte(dnsConfig), "/var/srv/dnsmasq.conf")
 }
 
-func createDnsConfigFile(values dnsmasqConfFileValues) (string, error) {
+func createDnsConfigFile(values dnsmasqConfFileValues, tmpl string) (string, error) {
```
cfergeau, line 58: This is no longer mandatory tohave this as an argument.

praveenkumar: @cfergeau yes but in case we again need to add the hosts file for dnsmasq.

cfergeau: Then the argument can be readded again at this time ;)

PR #1155, opened 2020-04-14
Implement cleanup command for windows

praveenkumar:
**Fixes:** Issue #507

### DIFF
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -11,6 +11,7 @@ import (
 	winnet "github.com/code-ready/crc/pkg/os/windows/network"
 	"github.com/code-ready/crc/pkg/os/windows/powershell"
 
+	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/machine/hyperv"
 )
 
@@ -165,4 +166,40 @@ func checkIfRunningAsNormalUser() error {
 	}
 	logging.Debug("Ran as administrator")
 	return fmt.Errorf("crc should be ran as a normal user")
+}
+
+func removeDnsServerAddress() error {
+	resetDnsCommand := `Set-DnsClientServerAddress -InterfaceAlias ("vEthernet (crc)") -ResetServerAddresses`
+	if exist, defaultSwitch := winnet.GetDefaultSwitchName(); exist {
+		resetDnsCommand = fmt.Sprintf(`Set-DnsClientServerAddress -InterfaceAlias ("vEthernet (%s)","vEthernet (crc)") -ResetServerAddresses`, defaultSwitch)
+	}
+	if _, _, err := powershell.ExecuteAsAdmin("Remove dns entry for default switch", resetDnsCommand); err != nil {
+		return err
+	}
+	return nil
+}
+
+func removeCrcVM() (err error) {
+	defer func() {
+		if ferr := os.RemoveAll(constants.MachineInstanceDir); ferr != nil {
+			logging.Debugf("Error removing %s dir: %v", constants.MachineInstanceDir, err)
+			err = fmt.Errorf("Error removing %s dir", constants.MachineInstanceDir)
+		}
+	}()
+	if _, _, err := powershell.Execute("Get-VM -Name crc"); err != nil {
+		// This means that there is no crc VM exist
+		return nil
+	}
+	stopVMCommand := fmt.Sprintf(`Stop-VM -Name "%s" -Force`, constants.DefaultName)
+	if _, _, err := powershell.Execute(stopVMCommand); err != nil {
+		// ignore the error as this is useless (prefer not to use nolint here)
+		return err
+	}
+	removeVMCommand := fmt.Sprintf(`Remove-VM -Name "%s" -Force`, constants.DefaultName)
+	if _, _, err := powershell.Execute(removeVMCommand); err != nil {
+		// ignore the error as this is useless (prefer not to use nolint here)
+		return err
+	}
+	logging.Debug("'crc' VM is removed")
+	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -166,3 +167,41 @@ func checkIfRunningAsNormalUser() error {
 	logging.Debug("Ran as administrator")
 	return fmt.Errorf("crc should be ran as a normal user")
 }
+
+func removeDnsServerAddress() error {
+	//Set-DNSClientServerAddress "vEthernet (Default Switch)" -ResetServerAddresses
+	resetDnsForDefaultSwitchCommand := `Set-DnsClientServerAddress "vEthernet (Default Switch)" -ResetServerAddresses`
```
gbraad, line 173: Maybe combine this in a single command to execute, so we don't show two UAC popups.

praveenkumar: @gbraad I also thought same but may be https://docs.microsoft.com/en-us/powershell/module/dnsclient/set-dnsclientserveraddress?view=win10-ps#syntax `InterfaceAlias` might work, will need to check.

gbraad: you likely just need to add both commands with a separator. the Alias is an alternative name for the network interface, which "crc" and "default switch" actually are.

praveenkumar: @gbraad so in this case alias is much better than adding the commands with separator right?

gbraad: I am not sure what you mean:

```
PS> Set-DnsClientServerAddress "crc" -ResetServerAddresses`
```

is the same as:

```
PS> Set-DnsClientServerAddress -InterfaceAlias "vEthernet (Default Switch)" -ResetServerAddresses
```

I mean to combine them to execute as:

```
PS Set-DnsClientServerAddress "crc" -ResetServerAddresses; Set-DnsClientServerAddress "default switch" -ResetServerAddresses

Comment (praveenkumar) [reply pkg/crc/preflight/preflight_checks_windows.go:173]: @gbraad How about following?

```
PS> Set-DnsClientServerAddress -InterfaceAlias ("default switch","crc") -ResetServerAddresses
```

Comment (gbraad) [reply pkg/crc/preflight/preflight_checks_windows.go:173]: that should be:


```
PS> Set-DnsClientServerAddress -InterfaceAlias ("vEthernet (Default Switch)","vEthernet (crc)") -ResetServerAddresses
```

as Alias needs the actual `vEthernet` prefix. Also note, that "Default Switch" might not be named as such on localized versions of Windows. This is already part of the code and best to use this replacement value.

Comment (gbraad) [reply pkg/crc/preflight/preflight_checks_windows.go:173]: https://github.com/code-ready/crc/blob/20d32e725098acf94d9032f0799e43224b9b82f5/pkg/os/windows/network/vswitch_windows.go#L37-L47

Comment (gbraad) [reply pkg/crc/preflight/preflight_checks_windows.go:173]: which would return nothing if the default does not exist. and could prevent the second device...

but it would still need to be collected in a single statement to prevent the UAC popup IF we by accident have configured it once (before they started to use the one named crc)

**Comment by anjannath** [comment]
@gbraad am getting UAC prompts when it tries and removes the dns entries.

File: pkg/crc/preflight/preflight_checks_windows.go (hunk 1 above)
Comment (praveenkumar) [reply pkg/crc/preflight/preflight_checks_windows.go:173]: @gbraad How about following, this way we delete the both and single time UAC popup?

```
func removeDnsServerAddress() error {
	resetDnsForDefaultSwitchCommand := `Set-DnsClientServerAddress -InterfaceAlias ("vEthernet (crc)") -ResetServerAddresses`
	if exist, defaultSwitch := winnet.GetDefaultSwitchName(); exist {
		resetDnsForDefaultSwitchCommand = fmt.Sprintf(`Set-DnsClientServerAddress -InterfaceAlias ("vEthernet (%s)","vEthernet (crc)") -ResetServerAddresses`, defaultSwitch)
	}
	if _, _, err := powershell.ExecuteAsAdmin("Remove dns entry for default switch", resetDnsForDefaultSwitchCommand); err != nil {
		return err
	}
	return nil
}
```

File: pkg/crc/preflight/preflight_checks_windows.go [hunk 2]
```diff
@@ -166,3 +167,39 @@ func checkIfRunningAsNormalUser() error {
 	logging.Debug("Ran as administrator")
 	return fmt.Errorf("crc should be ran as a normal user")
 }
+
+func removeDnsServerAddress() error {
+	resetDnsForDefaultSwitchCommand := `Set-DnsClientServerAddress -InterfaceAlias ("vEthernet (crc)") -ResetServerAddresses`
```
gbraad, line 172: default switch?

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -166,3 +167,39 @@ func checkIfRunningAsNormalUser() error {
...
+	if exist, defaultSwitch := winnet.GetDefaultSwitchName(); exist {
+		resetDnsForDefaultSwitchCommand = fmt.Sprintf(`Set-DnsClientServerAddress -InterfaceAlias ("vEthernet (%s)","vEthernet (crc)") -ResetServerAddresses`, defaultSwitch)
```
gbraad, line 174: why not insert the string for the default, by appending...

or collect the options and return the `-InterfaceAlias ("vEthernet (crc)", ...` accordingly.

gbraad: `resetDnsForDefaultSwitchCommand` is not the right name when only `crc` is used.

Rather call this `resetDnsForVirtualSwitchCommand`

Issue #1156, opened 2020-04-14
Apply IBM style guide to documentation


robin-owen:

robin-owen:
Fixed by #1157.

PR #1157, opened 2020-04-14
Issue #1156 Apply IBM style guide to documentation

robin-owen:
**Fixes:** Issue #1156

Note that the IBM style guide advocates the use of "notebook" rather than "laptop" when describing hardware. I've fixed the one time we use "laptop" in the documentation, but we do not directly control the page on cloud.redhat.com which is named "Install on Laptop" and includes wording such as "Red Hat CodeReady Containers brings a minimal OpenShift 4.2 or newer cluster to your local laptop or desktop computer."

### DIFF
--- a/docs/source/topics/assembly_accessing-the-openshift-cluster.adoc
+++ b/docs/source/topics/assembly_accessing-the-openshift-cluster.adoc
@@ -1,7 +1,7 @@
 [id="accessing-the-openshift-cluster_{context}"]
 = Accessing the OpenShift cluster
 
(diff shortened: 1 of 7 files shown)


robin-owen:
Also, I do not have a strong opinion regarding "laptop" or "notebook", but the more I look into it, the more I believe we should use "laptop", as that's the more inclusive wording. As far as I can tell, the term "notebook" is often used for lightweight laptops.

cfergeau:
Ok, this replaces 'via' with 'through', and one instance of 'on' with 'about', fine with me.

Issue #1152, opened 2020-04-10
[BUG] NetworkManager configs should be readable by non-root user


siteshwar:
### General information

  * OS: Linux 
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? Yes

## CRC version
```bash
crc version: 1.8.0+0a318dc
OpenShift version: 4.3.8 (embedded in binary)

```
  
## CRC status
```bash
CRC VM:          Stopped
OpenShift:       Stopped
Disk Usage:      0B of 0B (Inside the CRC VM)
Cache Usage:     11.8GB
Cache Directory: /home/situ/.crc/cache
```

## CRC config
```bash
```

## Host Operating System
```bash
NAME=Fedora
VERSION="31 (KDE Plasma)"
ID=fedora
VERSION_ID=31
VERSION_CODENAME=""
PLATFORM_ID="platform:f31"
PRETTY_NAME="Fedora 31 (KDE Plasma)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:31"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f31/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=31
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=31
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="KDE Plasma"
VARIANT_ID=kde
```

### Steps to reproduce

  1. Download tar archive for linux from [here](https://cloud.redhat.com/openshift/install/crc/installer-provisioned)
  2. Run `crc setup` as a non-root user
  3. Run `crc start` as non-root user

### Expected
`crc start` should run succesfully.

### Actual
During setup I get these errors:

```
FATA Error opening file: /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf: open /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf: permission denied 
[...]

PR #1158, opened 2020-04-14
Issue #1152 Add file mode as argument for file creation

praveenkumar:
As of now if a user uses umask as `0077` which is serve security
level then any file created by default not readable by any different
user. With this setup if that user run `crc setup` it will pass without
any issue but then `crc start` fails with following error.

```
FATA Error opening file: /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf: open /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf: permission denied
```

This is happen since we create those file using `sudo` and don't provide
any file-mode which makes those file only readable by root user and thus this failure.

**Fixes:** Issue #1152

### DIFF
--- a/pkg/os/util_linux.go
+++ b/pkg/os/util_linux.go
@@ -3,13 +3,15 @@ package os
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
-func WriteToFileAsRoot(reason, content, filepath string) error {
+func WriteToFileAsRoot(reason, content, filepath string, mode os.FileMode) error {
 	logging.Infof("Will use root access: %s", reason)
 	cmd := exec.Command("sudo", "tee", filepath) // #nosec G204
 	cmd.Stdin = strings.NewReader(content)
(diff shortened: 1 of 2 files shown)


siteshwar:
LGTM

pkg/os/util_linux.go
```diff
@@ -18,6 +20,10 @@ func WriteToFileAsRoot(reason, content, filepath string) error {
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("Failed writing to file as root: %s: %s: %v", filepath, buf.String(), err)
 	}
+	if _, _, err := RunWithPrivilege(fmt.Sprintf("Changing permission for %s to %d ", filepath, mode),
+		"chmod", strconv.FormatUint(uint64(mode), 8), filepath); err != nil {
```
cfergeau, line 24: Could also be `fmt.Sprintf("%o", uint64(mode))`

cfergeau: os.FileMode is a `uint32`, so maybe slightly more expected to cast to `uint32`?

praveenkumar: @cfergeau yes but `strconv.FormatUint` doesn't allow to use uint32 so have to cast. Also as per https://stackoverflow.com/questions/24886015/how-to-convert-uint32-to-string using `strconv.FormatUint` is much better than `fmt.Sprintf`

cfergeau: > The better option is strconv.FormatUint because it is faster, has less memory allocations (benchmark examples here or here).

This is not really relevant in this code as we will not be doing the conversion really often, and this is used in a function which called `system()`.

PR #1176, opened 2020-04-21
preflight: Allow to skip podman check

cfergeau:
## Solution/Idea                                                                                                                                                                                                                              
                                                                                                                                                                                                                                              
Added `skip-check-podman-cached` and `warn-check-podman-cached` config options so that user can avoid podman being installed by crc
                                                                                                                                                                                                                                                                                                                                                                                                                          
## Testing                                                                                                                                                                                                                                                                                                                                                                                                                                               

0. `rm ~/.crc/bin/podman*`                                                                                                                                                                                                                                             
1. `crc config set       skip-check-podman-cached true`                                                                                                                                                                     
2. `crc setup` succeeds

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -22,6 +22,7 @@ var genericPreflightChecks = [...]PreflightCheck{
 		fix:              fixOcBinaryCached,
 	},
 	{
+		configKeySuffix:  "check-podman-cached",
 		checkDescription: "Checking if podman remote binary is cached",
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -22,6 +22,7 @@ var genericPreflightChecks = [...]PreflightCheck{
 		fix:              fixOcBinaryCached,
 	},
 	{
+		configKeySuffix:  "check-podman-cached",
```
praveenkumar, line 25: :+1:

PR #1177, opened 2020-04-21
Add cleanup command to tests

jsliacan:
**Fixes:** Issue #1279 

**Relates to:** Issue # PR #

## Solution/Idea

Simply add the step to `basic.feature` file that uses `crc cleanup` to remove the relevant settings. Check the stdout against expected stdout. Then use `crc cleanup` through other feature files after the crux of the story is finished.

## Proposed changes

1. Modify `*.feature` files except for `config.feature`, where CRC is not started.

## Testing

1. Run `make integration` and all tests should pass, including the `crc cleanup` related steps.

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -137,3 +137,39 @@ Feature: Basic test
     Scenario: CRC delete
         When executing "crc delete -f" succeeds
         Then stdout should contain "Deleted the OpenShift cluster"
+
+    @darwin
+    Scenario Outline: CRC clean-up
+        When executing "crc cleanup" succeeds
+        Then stdout should contain "Unload CodeReady Containers tray"
+        And stdout should contain "Unload CodeReady Containers daemon"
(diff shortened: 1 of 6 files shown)


anjannath:
This would need updates, as the preflight/cleanups have changed

jsliacan:
@anjannath - maybe I'll discard this PR (we're using cleanup command already, just not checking its output). I'll have to rework tests to follow the cucumber's progressively more enforced and restrictive design principles: https://github.com/cucumber/godog/issues/335. Basically every "Feature" needs to be crammed into a single Scenario. And each Scenario only has Steps as building blocks. So there won't be any space to check these individual outputs anymore (it's kind of unit-testish anyway).

jsliacan:
@praveenkumar , the centos ci check did not pass, but one thing that didn't fail (when you open the console output) is the cleanup scenario :).

praveenkumar (approved):

PR #1181, opened 2020-04-22
Update bundle version from 4.3.10 to 4.4.0-rc.11

praveenkumar:


pkg/crc/machine/machine.go
```diff
@@ -372,12 +372,12 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		}
 	}
 
-	// Start kubelet inside the VM
+	// Check if kubelet service is running inside the VM
 	sd := systemd.NewInstanceSystemdCommander(sshRunner)
-	kubeletStarted, err := sd.Start("kubelet")
+	kubeletStarted, err := sd.IsActive("kubelet")
```
cfergeau, line 377: This only leaves `configureCluster` to start the kubelet service inside the VM. This function is only called when creating a VM, not when starting a pre-existing VM. This means `crc start; crc stop; crc start` now fails
```
INFO Check DNS query from host ...
ERRO kubelet service is not running: ssh command error:
command : sudo systemctl -f is-active kubelet
err     : exit status 3
output  : inactive
```

praveenkumar: @cfergeau yes, I updated the PR after that, now it have another commit which should do it.

pkg/crc/machine/machine.go
```diff
@@ -769,12 +773,7 @@ func updateSSHKeyPair(sshRunner *crcssh.SSHRunner) error {
 	return err
 }
 
-func configureCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, proxyConfig *network.ProxyConfig, pullSecret, instanceIP string) (rerr error) {
-	sd := systemd.NewInstanceSystemdCommander(sshRunner)
-	if _, err := sd.Start("kubelet"); err != nil {
-		return fmt.Errorf("Error starting kubelet: %s", err)
-	}
-
+func configureCluster(ocConfig oc.OcConfig, sshRunner *crcssh.SSHRunner, proxyConfig *network.ProxyConfig, pullSecret, instanceIP string, sd *systemd.InstanceSystemdCommander) (rerr error) {
```
cfergeau, line 776: `configProxyForCluster` can recreate a `systemd.Instance` from the `sshRunner` when needed, I don't think it's really useful to pass around `sd` in function args.

pkg/crc/machine/machine.go
cfergeau: It would be better to reorder the commits so that we don't break something in one of the early commits, and then fix it in a later one.

praveenkumar:
> It would be better to reorder the commits so that we don't break something in one of the early commits, and then fix it in a later one.

@cfergeau yes, I am doing the reordering for this commit.

Issue #1143, opened 2020-04-07
Add "quick start" guide using `odo`


robin-owen:
The documentation would benefit from a "quick start" guide to direct the reader on how to get started using CodeReady Containers with `odo` to create and deploy a simple example application.

Ideally, we would like to use a simple Java application for the deployment via `odo`. At the moment, this may not be possible, and we will have to start writing documentation using a simple Node.js application instead -- this can be reevaluated later.

robin-owen:
Fixed by #1187.

PR #1187, opened 2020-04-24
Issue #1143 Add docs for using `odo` with CRC

robin-owen:
**Fixes:** Issue #1143

## Solution/Idea

We want to encourage the use of `odo` with the OpenShift cluster running in the CodeReady Containers VM for development purposes. This PR introduces a short guide for using `odo` to deploy a sample Java application with links to the `odo` documentation for more information.

## Proposed changes

1. Add the "Deploying a sample application with `odo`" procedure.
2. Add additional attributes for `odo` documentation.
3. Remove the mention of `odo` in "Accessing the OpenShift cluster with `oc`". We now include a procedure which goes over this process and includes links to `odo` documentation.

### DIFF
--- a/docs/source/topics/proc_deploying-sample-application-with-odo.adoc
+++ b/docs/source/topics/proc_deploying-sample-application-with-odo.adoc
@@ -1,0 +1,80 @@
+[id="deploying-sample-application-with-odo_{context}"]
+= Deploying a sample application with `odo`
+
+You can use OpenShift Do ([command]`odo`) to create OpenShift projects and applications from the command line.
+This procedure deploys a sample application to the OpenShift cluster running in the {prod} virtual machine.
+
+.Prerequisites
+
+* You have installed [command]`odo`.
+For more information, see link:{odo-docs-url-installing}[Installing `odo`] in the [command]`odo` documentation.
+* The {prod} virtual machine is running.
+For more information, see <<starting-the-virtual-machine_{context}>>.
+
+.Procedure
+
+To deploy a sample application through [command]`odo`, follow these steps:
+
+. Log in to the running {prod} OpenShift cluster as the `developer` user:
++
+[subs="+quotes,attributes"]
+----
+$ odo login -u developer -p developer
+----
+
+. Create a project for your application:
++
+[subs="+quotes,attributes"]
+----
+$ odo project create sample-app
+----
+
+. Create a directory for your components:
++
+[subs="+quotes,attributes"]
+----
+$ mkdir sample-app
+$ cd sample-app
+----
+
+. Create a component from a sample application on GitHub:
++
+[subs="+quotes,attributes"]
+----
+$ odo create nodejs --git https://github.com/openshift/nodejs-ex
+----
++
+[NOTE]
+====
+Creating a component from a remote Git repository will rebuild the application each time you run the [command]`odo push` command.
+To create a component from a local Git repository, see link:{odo-docs-url-single-component}[Creating a single-component application with `odo`] in the [command]`odo` documentation.
+====
+
+. Create a URL and add an entry to the local configuration file:
++
+[subs="+quotes,attributes"]
+----
+$ odo url create --port 8080
+----
+
+. Push the changes:
++
+[subs="+quotes,attributes"]
+----
+$ odo push
+----
++
+Your component is now deployed to the cluster with an accessible URL.
+
+. List the URLs and check the desired URL for the component:
++
+[subs="+quotes,attributes"]
+----
+$ odo url list
+----
+
+. View the deployed application using the generated URL.
+
+.Additional resources
+
+* For more information about using [command]`odo`, see the link:{odo-docs-url}[`odo` documentation].
--- a/docs/source/meta/product-attributes.adoc
+++ b/docs/source/meta/product-attributes.adoc
@@ -26,4 +26,6 @@
 
 // URLs
 :crc-download-url: https://cloud.redhat.com/openshift/install/crc/installer-provisioned
-:odo-docs-url: https://docs.openshift.com/container-platform/latest/cli_reference/openshift_developer_cli/creating-a-single-component-application-with-odo.html
+:odo-docs-url: https://docs.openshift.com/container-platform/latest/cli_reference/openshift_developer_cli/understanding-odo.html
+:odo-docs-url-installing: https://docs.openshift.com/container-platform/latest/cli_reference/openshift_developer_cli/installing-odo.html
+:odo-docs-url-single-component: https://docs.openshift.com/container-platform/latest/cli_reference/openshift_developer_cli/creating-a-single-component-application-with-odo.html
--- a/docs/source/topics/assembly_using-codeready-containers.adoc
+++ b/docs/source/topics/assembly_using-codeready-containers.adoc
@@ -7,6 +7,8 @@ include::proc_starting-the-virtual-machine.adoc[leveloffset=+1]
 
 include::assembly_accessing-the-openshift-cluster.adoc[leveloffset=+1]
 
+include::proc_deploying-sample-application-with-odo.adoc[leveloffset=+1]
+
 include::proc_stopping-the-virtual-machine.adoc[leveloffset=+1]
 
 include::proc_deleting-the-virtual-machine.adoc[leveloffset=+1]
--- a/docs/source/topics/proc_accessing-the-openshift-cluster-with-oc.adoc
+++ b/docs/source/topics/proc_accessing-the-openshift-cluster-with-oc.adoc
@@ -50,4 +50,3 @@ See <<troubleshooting-codeready-containers_{context}>> if you cannot access the 
 .Additional resources
 
 * The link:https://docs.openshift.com/container-platform/latest/applications/projects/working-with-projects.html[OpenShift documentation] covers the creation of projects and applications.
-* You can also use link:{odo-docs-url}[OpenShift Do] (`odo`) to create OpenShift projects and applications from the command line.


robin-owen:
@boczkowska, could you provide a peer review for this content?

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
```diff
@@ -0,0 +1,60 @@
...
+* [command]`odo` is installed.
+For more information, see link:{odo-docs-url-installing}[Installing `odo`] in the [command]`odo` documentation.
+* The {prod} virtual machine is running.
+For more information, see <<starting-the-virtual-machine_{context}>>.
+* You must log in through [command]`oc` as the `developer` user.
+For more information, see <<accessing-the-openshift-cluster-with-oc_{context}>>.
+
+.Procedure
+
+To deploy a sample Java application through [command]`odo`, follow these steps:
```
cfergeau, line 18: This is both redundant with https://docs.openshift.com/container-platform/4.3/cli_reference/openshift_developer_cli/creating-a-single-component-application-with-odo.html , and at the same time slightly different (java instead of nodejs, odo create --git rather than direct git clone call, ...) I don't know if we should duplicate all these steps in crc documentation ?

robin-owen: This is also greatly simplified and doesn't go into detail.

The main idea here is to quickly demonstrate how to create and deploy a sample application using both together. Without any documentation "gluing" these concepts, the idea remains somewhat abstract.

If we only want to document how to connect `odo` and `crc`, then there's really almost nothing to document -- start CRC, then `odo login -u developer -p developer` -- that's it. I thought we should go into a little bit more detail here to have the user end up with a visible deployment.

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
```diff
@@ -0,0 +1,60 @@
...
+----
+$ mkdir java-sample
+$ cd java-sample
+----
+
+. Create a Java component from a sample application on GitHub:
++
+[subs="+quotes,attributes"]
+----
+$ odo create java --git https://github.com/spring-projects/spring-petclinic.git
```
praveenkumar, line 39: I am bit concern about having a sample app as java, it might take more resources than a simple python/nodejs example, I would first try this before merge this PR. When a new user just start the crc he really have very limited resource to play with the cluster and I don't want first app to be crash because of mem restriction we have :(

robin-owen: I have tried this. It works with the default CRC settings on 4.3.10.

We also discussed using a Java sample application during one of the CRC docs sync meetings and agreed that Java is more appropriate for the audience.

cfergeau: Quarkus? :) We could add an automated tests for these steps which run once in a while and lets us know if something broke, as it would indeed look bad if the example from our doc failed.

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
cfergeau: Ok, you're our doc owner, so that's your call, fine with me :)

robin-owen: I do agree that it's essentially redundant, but that's not always a bad thing. :)

One of the ideas here is to use minimal steps to get something visibly working. Fortunately, all of the steps described are stable and there shouldn't be any maintenance cost re: keeping it up to date (unless we keep having memory constraint issues, as mentioned by @praveenkumar).

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
robin-owen: Alternatively, if we can find the lightest possible Java sample application, that would also be nice. :)

Honestly, though, I don't have strong opinions about whether we should use Java or Python/Node.js personally.

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
boczkowska: @cfergeau odo also has a guide with Java https://docs.openshift.com/container-platform/4.3/cli_reference/openshift_developer_cli/creating-a-multicomponent-application-with-odo.html and a list of sample apps in different languages https://docs.openshift.com/container-platform/4.3/cli_reference/openshift_developer_cli/using-sample-applications.html

@robin-owen regarding `odo create --git` vs `git clone`, as it was mentioned in odo channel, `git clone` would probably be quicker on push for users. It adds more steps to the procedure so your call. :)

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
```diff
@@ -0,0 +1,60 @@
...
+----
+
+. Push the changes:
++
+[subs="+quotes,attributes"]
+----
+$ odo push
+----
++
+Your component is now deployed to the cluster with an accessible URL.
```
boczkowska, line 56: Suggestion: add some verification step. It could be `odo url list` - it displays the URL(s) of the component or `odo list` - it displays all of the components of the application OR, instead of an extra command, an output of `odo push`, it should be something like

```Validation
 ✓  Checking component [7ms]

 Configuration changes
 ✓  Initializing component
 ✓  Creating component [134ms]

 Applying URL changes
 ✓  URL myurl: <URL HERE>

 Pushing to component <component-name> of type local
 ✓  Checking files for pushing [657850ns]
 ✓  Waiting for component to start [6s]
 ✓  Syncing files to the component [408ms]
 ✓  Building component [7s]
 ✓  Changes successfully pushed to component
```
I can quickly test it and give you the exact output with this app if needed.

robin-owen: Fair enough -- originally, I included the `odo url list` step in the procedure, but noticed that the output of `odo push` prominently displayed the URL.

Personally, I'd like to avoid using output from commands in the documentation -- the commands themselves are unlikely to change, but the same can't be said for the output.

I'll go ahead and add the `odo url list` step back into the procedure just to be sure. :)

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
robin-owen: @boczkowska Right, that's the issue I was trying to figure out myself. ;)

Considering this is more of a demonstration than a detailed guide, I considered less commands to type (read: steps to follow) was appropriate. More information is always available from the `odo` docs directly. :)

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
robin-owen: Updated. Thank you!

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
gbraad: > I am bit concern about having a sample app as java, it might take more resources

This was discussed in several docs-sync. Java seems more appropriate with our other offerings. Odo's documentation can serve the other purposes.

robin-owen: It's also worth noting that the only components listed as supported by `odo catalog list components` are `java` and `nodejs`. This suggests, in my opinion, that we should use one of these sample applications as the example for this procedure. If the Java application is too heavy for the average CodeReady Containers OpenShift cluster, that only leaves one option.

@code-ready/crc-devel, please let me know what you think -- would you prefer that we use the Java sample, or a Node.js sample? If possible, I'd like to have this documentation merged before the 1.10 release. Thank you!

cfergeau: That specific  https://github.com/spring-projects/spring-petclinic.git application is imo too heavy, in that it needs to download a ton of stuff before building. Hopefully there are more lightweight java apps which can be used instead.

cfergeau: It seems to be the "official" java sample for odo :( https://docs.openshift.com/online/starter/cli_reference/openshift_developer_cli/using-sample-applications.html#java

dharmit: @robin-owen you're right about nodejs and java support in odo. They're the only two supported component types right now (and with specific versions)

@cfergeau we're actively working on Quarkus support but the work is tied to DevFile work that's also happening in parallel. I'm optimistic of its support landingin  soon! It is being tracked at https://github.com/openshift/odo/issues/2800.

At the moment, I'm not aware of any examples other than the ones mentioned in the docs. I'll ping you folks as soon as I have something lightweight and java. :disappointed:

gbraad: what is the ETA on Quarkus support? If you can't say here, ping me on slack. A rough estimate would be good enough

robin-owen:
I've updated this PR to use a Node.js sample application and include an admonition about the use of a remote Git repository (with a link to the `odo` documentation for using a local repo).

The wording has also been altered to not refer to the type of sample application so that we can change this more easily once a more lightweight Java sample is available.

@code-ready/crc-devel, please take a look. :)

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
dharmit: There's no straightforward answer to this but I think it'll be around July for it to be generally available. In experimental mode that works with devfile, it'll land in a week or two. But "experimental mode" is prone to breaking changes.

docs/source/topics/proc_deploying-sample-application-with-odo.adoc
```diff
@@ -0,0 +1,80 @@
...
+----
+$ mkdir sample-app
+$ cd sample-app
+----
+
+. Create a component from a sample application on GitHub:
++
+[subs="+quotes,attributes"]
+----
+$ odo create nodejs --git https://github.com/openshift/nodejs-ex
```
praveenkumar, line 44: :+1:

PR #1196, opened 2020-04-28
Disable the podman remote

praveenkumar:
## Solution/Idea

Disable the podman remote support.

## Proposed changes

- No podman remote binary download for any platform
- No podman remote binary embedded for any platform
- `podman-env` will error out for each platform.

## Testing

```
$ ./crc setup --log-level debug
DEBU oc binary already cached                     
DEBU Currently podman remote is not supported     
[...]

$ ./crc podman-env
Currently not supported.
```

### DIFF
--- a/cmd/crc/cmd/podman_env.go
+++ b/cmd/crc/cmd/podman_env.go
@@ -5,7 +5,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/output"
-	crcos "github.com/code-ready/crc/pkg/os"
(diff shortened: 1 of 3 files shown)


cmd/crc/cmd/podman_env.go
```diff
@@ -17,9 +16,7 @@ var podmanEnvCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 
 		// See issue #961; Currently does not work on Windows in combination with the CRC vm.
-		if crcos.CurrentOS() == crcos.WINDOWS {
-			errors.ExitWithMessage(1, "Currently not supported on this platform.")
-		}
+		errors.ExitWithMessage(1, "Currently not supported.")
```
gbraad, line 19: we state this is experimental...

PR #1217, opened 2020-05-05
Update the oc binary cache path

praveenkumar:
Right now we are storing all the binaries which are part of CRC
to `~/.crc/bin` and it causing the issue with env commands where
we set the path to `~/.crc/bin:$PATH` and all the binaries available
in the `~/.crc/bin` also available to user path and create confusion
for some of system specific binaries.

### DIFF
--- a/cmd/crc/cmd/oc_env.go
+++ b/cmd/crc/cmd/oc_env.go
@@ -27,7 +27,7 @@ var ocEnvCmd = &cobra.Command{
 		if err != nil {
 			errors.Exit(1)
 		}
-		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
+		output.Outln(shell.GetPathEnvString(userShell, constants.CrcOcBinDir))
 		if proxyConfig.IsEnabled() {
 			output.Outln(shell.GetEnvString(userShell, "HTTP_PROXY", proxyConfig.HttpProxy))
 			output.Outln(shell.GetEnvString(userShell, "HTTPS_PROXY", proxyConfig.HttpsProxy))
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -67,7 +67,11 @@ func fixBundleCached() error {
 
 // Check if oc binary is cached or not
 func checkOcBinaryCached() error {
-	oc := cache.NewOcCache(constants.CrcBinDir)
+	// Remove oc binary from older location and ignore the error
+	// We should remove this code after 3-4 releases. (after 2020-07-10)
+	os.Remove(filepath.Join(constants.CrcBinDir, "oc"))
+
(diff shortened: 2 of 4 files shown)


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -62,6 +62,10 @@ func fixBundleCached() error {
 
 // Check if oc binary is cached or not
 func checkOcBinaryCached() error {
+	// Remove oc binary from older location and ignore the error
+	// We should remove this code after 3-4 releases.
```
cfergeau, line 66: maybe add a date or a release version so that we have a starting point for the "3-4 releases"?

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -62,6 +62,10 @@ func fixBundleCached() error {
...
+	os.Remove(filepath.Join(constants.CrcBinDir, "oc"))
```
cfergeau, line 67: I think gosec will complain if it's not `_ = os.Remove(...)`

cmd/crc/cmd/oc_env.go
```diff
@@ -27,7 +27,7 @@ var ocEnvCmd = &cobra.Command{
 		if err != nil {
 			errors.Exit(1)
 		}
-		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
+		output.Outln(shell.GetPathEnvString(userShell, constants.CrcOcBinDir))
```
cfergeau, line 30: This is going to conflict with https://github.com/code-ready/crc/pull/1136/commits
I don't know in which order you want to merge these? The commits up to `Use Cache interface to cache oc and podman binary` in that PR could be finished and merged without the hosts changes if you prefer to have the refactor first.

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau If we are consist about releases then I can add the date but like this time we are not releasing so bit hard to put an date, I can add a tentative one if that helps?

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau it does put the severity is not hight so CI ignore it like any other errors like we don't handle the error for `fmt.Printf`.

cmd/crc/cmd/oc_env.go
praveenkumar: @cfergeau right, I rebased the #1136 and hope that will merge soon so I can update this one.

pkg/crc/preflight/preflight_checks_common.go
cfergeau: An approximate date corresponding to when the PR was written should be good enough
`// We should remove this code after 3-4 releases (after 2020-05-11)`

PR #1220, opened 2020-05-06
Issue #1211 Explicit add base domain to resolver in MacOS

praveenkumar:
It might happen that a user is using multiple resolver which might
have same domain name in that case resolver man page advise to use
the `domain` field explicit to avoid any kind of dns resolution issues.

## Testing

After crc start happen check the content of `/etc/resolver/testing` file it should have domain details

```
$ cat /etc/resolver/testing 
port 53
domain testing
nameserver 192.168.64.92
search_order 1
```

### DIFF
--- a/pkg/crc/services/dns/dns_darwin.go
+++ b/pkg/crc/services/dns/dns_darwin.go
@@ -18,12 +18,14 @@ import (
 
 const (
 	resolverFileTemplate = `port {{ .Port }}
+domain {{ .Domain }}
 nameserver {{ .IP }}
 search_order {{ .SearchOrder }}`
 )
 
 type resolverFileValues struct {
 	Port        int
+	Domain      string
 	IP          string
 	SearchOrder int
 }
@@ -37,7 +39,8 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
 	}
 
 	// Write resolver config to host
-	needRestart, err := createResolverFile(serviceConfig.IP, filepath.Join("/", "etc", "resolver", serviceConfig.BundleMetadata.ClusterInfo.BaseDomain))
+	needRestart, err := createResolverFile(serviceConfig.IP, serviceConfig.BundleMetadata.ClusterInfo.BaseDomain,
+		serviceConfig.BundleMetadata.ClusterInfo.BaseDomain)
 	if err != nil {
 		result.Success = false
 		return *result, err
@@ -64,11 +67,12 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
 	return *result, nil
 }
 
-func createResolverFile(InstanceIP string, path string) (bool, error) {
+func createResolverFile(InstanceIP string, domain string, filename string) (bool, error) {
 	var resolverFile bytes.Buffer
 
 	values := resolverFileValues{
 		Port:        dnsServicePort,
+		Domain:      domain,
 		IP:          InstanceIP,
 		SearchOrder: 1,
 	}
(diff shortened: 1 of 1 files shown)


pkg/crc/services/dns/dns_darwin.go
```diff
@@ -37,7 +39,8 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *se
 	}
 
 	// Write resolver config to host
-	needRestart, err := createResolverFile(serviceConfig.IP, filepath.Join("/", "etc", "resolver", serviceConfig.BundleMetadata.ClusterInfo.BaseDomain))
+	needRestart, err := createResolverFile(serviceConfig.IP, serviceConfig.BundleMetadata.ClusterInfo.BaseDomain,
+		filepath.Join("/", "etc", "resolver", serviceConfig.BundleMetadata.ClusterInfo.BaseDomain))
```
cfergeau, line 43: I would move that `filepath.Join()` to `createResolverFile`, as we are not going to be creating resolver files outside of `/etc/resolver/` anyway.

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -18,12 +18,14 @@ import (
 
 const (
 	resolverFileTemplate = `port {{ .Port }}
+domain {{ .Domain }}
```
cfergeau, line 21: Did you test this with a filename != domain name ? Only documentation about the format of that file which I found is https://hints.macworld.com/article.php?story=2004062902195410 which is very old, but says that `domain` must be the first entry in the file.

pkg/crc/services/dns/dns_darwin.go
praveenkumar: `createResolverFile` This function is independent of file so I don't want to move that to this functio, to me provide the info and file path to this function and it will create the content for a resolve file, it happened to be a static file in our case if we really want to do then might be change the function name to `createBaseDomainResolverFile` or something.

pkg/crc/services/dns/dns_darwin.go
praveenkumar: @cfergeau yes tested it and it does work, also in the article I just notice it mentioned `you can use any name if you specify the option domain foobar.com as the first one in that file` not first entry in the file ( not sure if I missed something )

```
$ ls /etc/resolver/
foo

17:52 $ cat /etc/resolver/foo
port 53
domain testing
nameserver 192.168.64.92
search_order

17:52 $ ping -c 4 bha.apps-crc.testing
PING bha.apps-crc.testing (192.168.64.92): 56 data bytes
64 bytes from 192.168.64.92: icmp_seq=0 ttl=64 time=0.222 ms
64 bytes from 192.168.64.92: icmp_seq=1 ttl=64 time=0.259 ms
64 bytes from 192.168.64.92: icmp_seq=2 ttl=64 time=0.461 ms
64 bytes from 192.168.64.92: icmp_seq=3 ttl=64 time=0.298 ms

--- bha.apps-crc.testing ping statistics ---
4 packets transmitted, 4 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 0.222/0.310/0.461/0.091 ms

```

pkg/crc/services/dns/dns_darwin.go
cfergeau: oh, forgot to suggest to rename the file from `$domain` to `crc-resolver` or something. 

> This function is independent of file 
And my point is that it does not make sense to expect the caller to know that `createResolverFile` must be used with a file based in `/etc/resolver/` :) we can keep a `filename` arg if you think it's useful, but I'd still move the `filepath.Join("/", "etc", "resolver", filename)` call in `createResolverFile`.

Issue #795, opened 2019-11-08
Systemtray (Windows)


gbraad:
Tasks
-----

  * [x] service #796
  * [x] system tray entry
    * [x] allow to start cluster
    * [x] allow to stop cluster
    * [x] represent cluster state
    ~~opt. start terminal with oc environment set~~
    * [x] allow to set configurations (#1246)

anjannath:
Few issues with the setup:
1. Daemon service is not writing the logs to `~/.crc/crcd.log` haven't able to figure out why.
2. in total now we get 6 UAC prompts which gets a bit annoying. (for the daemon service installation and starting it)
3. To install the daemon service we need the user's password, so the user gets a prompt to enter their password during setup.

With the above issues present, the PR filed for this,  can install the service and the tray.

gbraad:
@anjannath can you add issue numbers for those about config, etc?

anjannath:
@gbraad Linked to #1246 pr for the config api for daemon, the other related issues are part of the tray repo, which i think are listed in a card in the project board so will update it there.

anjannath:
Issue for the config window: https://github.com/code-ready/tray-windows/issues/5

PR #1221, opened 2020-05-06
Tray setup for windows

anjannath:
**Fixes:** #795 

## Solution/Idea

Make the crc daemon run as a service in windows and automate tray installation process from `crc setup --enable-experimental-features`

## Proposed changes

Add preflight and clean up checks

## Testing

1. `crc setup --enable-experimental-features` succeeds
2. should have a working system tray
3. The tray and the daemon service should auto start after windows boots

For debugging we can check if the _CodeReady Containers_ service was created in the service manager (in run box `services.msc`

### DIFF
--- a/cmd/crc/cmd/daemon.go
+++ b/cmd/crc/cmd/daemon.go
@@ -1,11 +1,6 @@
 package cmd
 
 import (
-	"os"
-
-	"github.com/code-ready/crc/pkg/crc/api"
-	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/spf13/cobra"
 )
 
@@ -21,19 +16,4 @@ var daemonCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		runDaemon()
 	},
-}
-
-func runDaemon() {
-	//setup separate logfile for daemon
-	logging.CloseLogging()
-	logging.InitLogrus(logging.LogLevel)
-	logging.SetupFileHook(constants.DaemonLogFilePath)
-
-	// Remove if an old socket is present
-	os.Remove(constants.DaemonSocketPath)
-	crcApiServer, err := api.CreateApiServer(constants.DaemonSocketPath)
-	if err != nil {
-		logging.Fatal("Failed to launch daemon", err)
-	}
-	crcApiServer.Serve()
 }
--- a/cmd/crc/cmd/daemon_windows.go
+++ b/cmd/crc/cmd/daemon_windows.go
@@ -1,0 +1,20 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/code-ready/crc/pkg/crc/api"
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+)
+
+func runDaemon() {
+	//setup separate logfile for daemon
+	logging.CloseLogging()
+	logging.InitLogrus(logging.LogLevel)
+	logging.SetupFileHook(constants.DaemonLogFilePath)
+
+	// Remove if an old socket is present
+	os.Remove(constants.DaemonSocketPath)
+	api.RunCrcDaemonService("CodeReady Containers", false)
+}
--- a/pkg/crc/api/api_windows.go
+++ b/pkg/crc/api/api_windows.go
@@ -1,0 +1,76 @@
+package api
+
+import (
+	"fmt"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/debug"
+	"golang.org/x/sys/windows/svc/eventlog"
+)
+
+var elog debug.Log
+
+type crcDaemonService struct {
+	socketPath string
+}
+
+func (m *crcDaemonService) Execute(args []string, r <-chan svc.ChangeRequest, changes chan<- svc.Status) (ssec bool, errno uint32) {
+	const cmdsAccepted = svc.AcceptStop | svc.AcceptShutdown
+	crcApiServer, err := CreateApiServer(m.socketPath)
+	if err != nil {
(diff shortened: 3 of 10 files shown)


pkg/crc/preflight/preflight_windows.go
```diff
@@ -55,11 +55,61 @@ var hypervPreflightChecks = [...]PreflightCheck{
...
+		flags:            SetupOnly,
+	},
+	{
+		checkDescription: "Checking if tray is running",
+		check:            checkTrayRunning,
+		fixDescription:   "Starting CodeReady Containers tray",
+		fix:              fixTrayRunning,
+		flags:            SetupOnly,
+	},
+}
```
praveenkumar, line 101: @anjannath don't we also need the cleanup steps so `crc cleanup` can also deal with removing the tray part on windows?

anjannath: Yes, that is needed, i was thinking to create another PR for it, but i can push those commits here also, haven't finished the cleanup part yet.

praveenkumar: Sure, better to have it part of this one but a separate commit would be OK.

cmd/crc/cmd/daemon.go
```diff
@@ -1,3 +1,5 @@
+// +build !windows
```
gbraad, line 1: would it be possible to be explicit about what we compile it for?

cmd/crc/cmd/daemon_windows.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(daemonCmd)
+}
+
+var daemonCmd = &cobra.Command{
```
gbraad, line 16: this code is duplicated between Windows and unix. When using a nonwindows.go file we can use the daemon.go for the shared code.

cmd/crc/cmd/daemon.go
anjannath: i could change the file name to `daemon_nonwin.go`

cmd/crc/cmd/daemon_windows.go
anjannath: Done, thanks!

pkg/crc/preflight/preflight_windows.go
anjannath: @praveenkumar Added the cleanup part in https://github.com/code-ready/crc/pull/1221/commits/b2fa823e398c20c6ea60b3b1645de282e7f229c5

pkg/crc/api/api_windows.go
```diff
@@ -0,0 +1,74 @@
...
+loop:
+	for { // nolint
+		select {
+		case c := <-r:
+			switch c.Cmd {
+			case svc.Interrogate:
+				changes <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
+				elog.Info(1, "Shutting down CRC daemon service") // nolint
+				break loop
```
gbraad, line 41: does this need to be named?

would a function named `MainLoop()` with the `for {};` not work with a `return` instead of a break?

anjannath: @gbraad Updated with the changes you suggested, ptal if that's what you had in mind!

pkg/crc/api/api_windows.go
```diff
@@ -0,0 +1,79 @@
...
+
+	MainLoop(r, changes)
+
+	// Above MainLoop will return when the daemon service is stopped
+	changes <- svc.Status{State: svc.Stopped}
+	return
+}
+
+func MainLoop(r <-chan svc.ChangeRequest, changes chan<- svc.Status) {
+	for { // nolint
```
gbraad, line 40: can this be `while(running)` and toggle the boolean instead of using a return?

gbraad:
```diff
func MainLoop(r <-chan svc.ChangeRequest, changes chan<- svc.Status) {
+       var running = true
-	for { // nolint
+       while(running) {
		select {
		case c := <-r:
			switch c.Cmd {
			case svc.Interrogate:
				changes <- c.CurrentStatus
			case svc.Stop, svc.Shutdown:
				elog.Info(1, "Shutting down CRC daemon service") // nolint
-				return
+                               running = false
			default:
				elog.Error(1, fmt.Sprintf("unexpected control request #%d", c)) // nolint
			}
		}
	}
+       return
}
```

gbraad: can `elog` be named `eventlog` ?

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	return nil
+}
+
+func fixTrayBinaryVersion() error {
+	return fixTrayBinaryExists()
+}
+
+func checkTrayBinaryAddedToStartupFolder() error {
+	if os.FileExists(filepath.Join(startUpFolder, constants.TrayShortCutName)) {
```
gbraad, line 155: Shortcut is a singleword. No camelcase needed

pkg/crc/api/api_windows.go
anjannath: No, there's already a import with that name..

https://github.com/code-ready/crc/pull/1221/files#diff-8e01d564e3fa24ba40438055729bc063R9

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Changed, thanks!

pkg/crc/api/api_windows.go
anjannath: > can this be `while(running)` and toggle the boolean instead of using a return?

@gbraad Updated! Had to remove the return from `MainLoop` as linter was complaining, anyways once the loop break it has to return, there;s no code after it..

gbraad: even in that case we can use aliases for those

Issue #723, opened 2019-10-15
[Epic] User can click to perform all usual command line actions (system trays)


anjannath:
As planned, the Goal is to have a systemtray for crc. It'll make repetitive actions like start/stop possible with a few clicks. This would be useful for developers who tend to stay on the IDE . They'd perform all interactions with the cluster using IDE provided plugins. In that scenario having a tray to manage the crc cluster would fit in without much friction.

We'd need a server component (crc daemon) to perform the actions on behalf of  the  GUI component.

- [x] system process; daemon/network service #796 
- [x] allow changing configuration through the daemon #1246 

systray
- [x] macOS #794 
- [x] Windows #795
   - [x] https://github.com/code-ready/tray-windows/issues/5

gbraad:
@anjannath ping

anjannath:
@gbraad macOS system tray can be moved out of experimental, remaining thing for the windows system tray to be out of experimental is

 - https://github.com/code-ready/tray-windows/issues/5

gbraad:
Will happen at the same time..

gbraad:
This issue can be closed as the epic has been resolved. Everything now are follow-up issues

PR #1246, opened 2020-05-21
Config api for the daemon

anjannath:
~I did these changes on top of #1221 so those needs to go in before this, when reviewing just review this commit https://github.com/code-ready/crc/pull/1246/commits/149b27f7ab684c210a66bef5720d60fe9d23fb56~
 Part of #723

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -77,7 +77,11 @@ func less(lhsKey, rhsKey string) bool {
 
 func configurableFields() string {
 	var fields []string
-	keys := cfg.AllConfigKeys()
+	var keys = make([]string, len(cfg.AllConfigs()))
+
+	for key := range cfg.AllConfigs() {
+		keys = append(keys, key)
+	}
 	sort.Slice(keys, func(i, j int) bool {
 		return less(keys[i], keys[j])
 	})
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -2,11 +2,13 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 
 	"github.com/code-ready/crc/cmd/crc/cmd/config"
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/validation"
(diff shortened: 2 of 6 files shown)


gbraad:
> - setconfig example
>   {"command":"setconfig", "args":{"bundle":"path/to/bundle", "cpu":"4"}}
> - getconfig example
>   {"command":"getconfig"} witouth the args object it will return all
>   {"command":"getconfig", "args":{"properties":["bundle", "memory"]}}
> - unsetconfig example
>   {"command":"unsetconfig", "args":{"properties":["bundle"]}}

This makes sense

pkg/crc/api/types.go
```diff
@@ -25,6 +26,20 @@ type clusterOpsRequest struct {
...
-	Args    map[string]string `json:"args,omitempty"`
+	Command string `json:"command"`
+	Args    json.RawMessage
+}
+
+// setOrUnsetConfigResult struct is used to return the result of
+// setconfig/unsetconfig command
+type setOrUnsetConfigResult struct {
+	Error   string
+	Success bool
```
cfergeau, line 37: Imo it's a mistake that some crc APIs return both an error and a success boolean. If `Success` is true, then `Error` must be `""`, if `Success` is false, then `Error` must be  non-empty, so they are mutually exclusive, so I would either send a message with only an `Error` field, or a message with the command result. No need for `Success`, unless maybe if the response would be empty otherwise.

I wonder if set/unset should return the list of config values it set/unset?

anjannath: Yes they are mutually exclusive, only the error would be enough to know if succeeded or failed but for `setconfig` and `unsetconfig` there is no command result, so `Success` is used.

> I wonder if set/unset should return the list of config values it set/unset?

I thought about this, in case any of the set/unset on the provided keys fails, we are getting back an error (err msg contains the name of the keys but harder to extract)

anjannath:
>Why the difference between the 2? Why not {"command":"unsetconfig", "args":{"bundle", "cpu"}}?

For `setconfig` we need a key value pair, but for `unsetconfig` and `getconfig` we just need the keys, so sending them inside the _properties_ array.

pkg/crc/config/viper_config.go
```diff
@@ -153,6 +153,19 @@ func AllConfigKeys() []string {
...
+	for _, key := range AllConfigKeys() {
+		v, err := Get(key)
+		if err != nil {
+			allConfigs[key] = nil
+		}
+		allConfigs[key] = v
+	}
+	return allConfigs
+}
+
```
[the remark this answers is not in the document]
anjannath: Yeah, we can get the list of keys from this function. if you think i should remove the `AllConfigKeys` then i can do that, i think right now only `cmd/config` uses it.

pkg/crc/api/handlers.go
```diff
@@ -64,18 +66,135 @@ func getPullSecretFileContent() (string, error) {
...
+		if err != nil {
+			multiError.Collect(err)
+		}
+		err = crcConfig.WriteConfig()
+		if err != nil {
+			multiError.Collect(err)
+		}
+	}
+	if len(multiError.Errors) != 0 {
+		setConfigResult.Error = fmt.Sprintf("%v", multiError.ToError())
```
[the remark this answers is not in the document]
anjannath: This is inline to what you mention in https://github.com/code-ready/crc/pull/1246#discussion_r429869470 so, `setconfig` will return

```json
{
    "properties": ["bundle", "pull-secret-file"],
    "Error": "xxxxxxxxx"
}
```

or when no errors..

```json
{
    "properties": ["bundle", "pull-secret-file", "memory"],
    "Error": ""
}
```

anjannath:
> > > Why the difference between the 2? Why not {"command":"unsetconfig", "args":{"bundle", "cpu"}}?
> > 
> > 
> > For `setconfig` we need a key value pair, but for `unsetconfig` and `getconfig` we just need the keys, so sending them inside the _properties_ array.
> 
> I don't understand what the `properties` array is bringing us? Why not have a `properties` array in the case of `setconfig` too then?

For `setconfig` we have an object, key/value pair so we can directly provide that to args, but for `unset/getconfig` we just the need the properties'/keys' names, so sending them in a array makes sense, and i named it as _properties_. we cannot pass it like you mentioned.. however we can do this..

```json
{"command":"unsetconfig", "args":{"bundle":"", "cpu":""}}
```

but using an array is better..
```json
{"command":"unsetconfig", "args":{"properties":["bundle", "cpu"]}}}
```

We can also have the _properties_ in `setconfig`, but its just one more layer..

```json
{"command":"setconfig", "args": {"properties": {"bundle":"path/to/bundle", "cpu":4}}}
```

cfergeau:
> > > > Why the difference between the 2? Why not {"command":"unsetconfig", "args":{"bundle", "cpu"}}?
> > > 
> > > 
> > > For `setconfig` we need a key value pair, but for `unsetconfig` and `getconfig` we just need the keys, so sending them inside the _properties_ array.
> > 
> > 
> > I don't understand what the `properties` array is bringing us? Why not have a `properties` array in the case of `setconfig` too then?
> 
> For `setconfig` we have an object, key/value pair so we can directly provide that to args, but for `unset/getconfig` we just the need the properties'/keys' names, so sending them in a array makes sense, and i named it as _properties_. we cannot pass it like you mentioned.. however we can do this..
> 
> ```json
> {"command":"unsetconfig", "args":{"bundle":"", "cpu":""}}
> ```
> 
> but using an array is better..
> 
> ```json
> {"command":"unsetconfig", "args":{"properties":["bundle", "cpu"]}}}
> ```
> 
> We can also have the _properties_ in `setconfig`, but its just one more layer..
> 
> ```json
> {"command":"setconfig", "args": {"properties": {"bundle":"path/to/bundle", "cpu":4}}}
> ```

In one case using an array (which is an extra layer) is better, but in the other case, using an additional layer seems to be a problem? :)

I think in both cases this should be wrapped in `properties`. If we ever want to add additional parameters to the `setconfig` command, the way it's currently done means we would have to use a parameter name which does not conflict with the name of a config key. So better to just wrap this in `properties`, and  be consistent and keep our options open for future extensions of the protocol.

pkg/crc/config/viper_config.go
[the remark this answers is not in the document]
anjannath: Done!

anjannath:
@cfergeau i think all the review commits are addressed, please have a look again, thanks!

pkg/crc/api/handlers.go
```diff
@@ -64,18 +66,139 @@ func getPullSecretFileContent() (string, error) {
...
 }
 
-func webconsoleURLHandler(_ ArgsType) string {
+func webconsoleURLHandler(_ json.RawMessage) string {
 	consoleConfig := machine.ConsoleConfig{Name: constants.DefaultName}
 	r, _ := machine.GetConsoleURL(consoleConfig)
 	return encodeStructToJSON(r)
 }
 
+func setConfigHandler(args json.RawMessage) string {
```
guillaumerose, line 81: A test in api_test.go would be really good!

pkg/crc/api/handlers.go
```diff
@@ -2,24 +2,26 @@ package api
...
 
-func statusHandler(client machine.Client, _ ArgsType) string {
+func statusHandler(client machine.Client, _ json.RawMessage) string {
 	statusConfig := machine.ClusterStatusConfig{Name: constants.DefaultName}
 	clusterStatus, _ := client.Status(statusConfig)
 	return encodeStructToJSON(clusterStatus)
 }
 
-func stopHandler(client machine.Client, _ ArgsType) string {
+func stopHandler(client machine.Client, _ json.RawMessage) string {
```
praveenkumar, line 24: @anjannath should this be in a different commit message where you changed the `ArgsType` to `json.RawMessage` ?

anjannath: Yeah, makes sense to separate this,will do thanks!

cmd/crc/cmd/config/config.go
```diff
@@ -77,7 +77,11 @@ func less(lhsKey, rhsKey string) bool {
 
 func configurableFields() string {
 	var fields []string
-	keys := cfg.AllConfigKeys()
+	var keys []string
```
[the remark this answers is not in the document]
cfergeau: And this causes https://github.com/code-ready/crc/issues/1487 ;)

Issue #1260, opened 2020-05-28
Permission denied error for goodhosts on crc start


anjannath:
### General information

  * OS: macOS 
  * Hypervisor:  hyperkit
  * Did you run `crc setup` before starting it (Yes/No)? yes

## CRC version
`master`
  
### Logs
```
DEBU About to run SSH command:
sudo podman run  --ip 10.88.0.8 --name dnsmasq -v /var/srv/dnsmasq.conf:/etc/dnsmasq.conf -p 53:53/udp --privileged -d quay.io/crcont/dnsmasq:latest
DEBU SSH cmd err, output: <nil>: 32841d539580fc457b0ab6834dfec600dbd158bcd6222c0581a9a6e3df4ef68f
DEBU Running '/Users/anjan/.crc/bin/goodhosts rm api.crc.testing'
DEBU Command failed: fork/exec /Users/anjan/.crc/bin/goodhosts: permission denied
DEBU stdout:
DEBU stderr:
ERRO Error running post start: fork/exec /Users/anjan/.crc/bin/goodhosts: permission denied
Making call to close driver server
(crc) Calling .Close
Successfully made call to close driver server
Making call to close connection to plugin binary
Making call to close driver server
(crc) Calling .Close
```

lhoss:
Could following permission issue (on Fedora 31), be related ?
https://github.com/code-ready/crc/issues/908#issuecomment-635910183
> ERRO Error creating host: Error creating the VM: Error creating machine: Error in driver during machine creation: virError(Code=38, Domain=18, Message='Cannot access storage file '/home/lhoss/.crc/machines/crc/crc' (as uid:107, gid:107): Permission denied')

anjannath:
Fixed in #1263

PR #1263, opened 2020-05-29
Issue 1260 Add execute permission for users which are not wheel group

praveenkumar:
**Fixes:** Issue #1260 

## Solution/Idea

`chmod g+x` only set execute permission for the users which are part of currently set group for the file. Looks like by default the normal user in mac or linux is not part of wheel group and file execution permission is not set for the user which leads to permission denied error.

## Testing

### DIFF
--- a/pkg/crc/network/utils.go
+++ b/pkg/crc/network/utils.go
@@ -10,6 +10,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 	"github.com/code-ready/crc/pkg/crc/ssh"
+	crcos "github.com/code-ready/crc/pkg/os"
 )
 
 func executeCommandOrExit(sshRunner *ssh.SSHRunner, command string, errorMessage string) string {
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_nonwin.go
```diff
@@ -58,7 +58,7 @@ func setSuid(path string) error {
 	}
 
 	/* Can't do this before the chown as the chown will reset the suid bit */
-	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s,g+x", path)
+	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s,g+x,o+x", path)
```
anjannath, line 61: 👍

pkg/crc/network/utils.go
```diff
@@ -125,6 +126,11 @@ func CheckCRCLocalDNSReachableFromHost(bundle *bundle.CrcBundleInfo, expectedIP
 		appsHostname := bundle.GetAppHostname("foo")
 		ip, err = net.LookupIP(appsHostname)
 		if err != nil {
+			// Right now goodhosts fallback is not implemented in windows so
+			// this checks should still return the error.
+			if crcos.CurrentOS() == crcos.WINDOWS {
+				return err
+			}
```
anjannath, line 133: @praveenkumar Just for clarification, earlier if this failed for windows it didn't error out, but now we will, (even though it doesn't have the `goodhosts` workaround) this will prevent me from a successful `crc start` right? earlier i could start crc and manually add entries to resolve the app domains..

praveenkumar: No, we made this change with #1136 and in windows since we don't have goodhosts implementation yet to add the entry to hosts file we will fall back to our earlier default way where we error out as soon as the dns resolution not works.

anjannath: Okay, thanks for the clarification!

PR #1267, opened 2020-06-01
Update tray version from 1.0.0-alpha.3 to 1.0.0-alpha.4

anjannath:


pkg/crc/version/version.go
```diff
@@ -41,7 +41,7 @@ var (
 const (
 	releaseInfoLink = "https://mirror.openshift.com/pub/openshift-v4/clients/crc/latest/release-info.json"
 	// Tray version to be embedded in binary
-	crcTrayVersion = "1.0.0-alpha.3"
+	crcTrayVersion = "1.0.0-alpha.4"
```
praveenkumar, line 44: :+1:

Issue #1278, opened 2020-06-02
Latest CRC setup fails


aceeric:
I'm running Ubuntu 18.04. I had been using CRC 1.10 then ran into the cert renewal issue yesterday. Reverted to 1.8 yesterday but had other issues. Noticed today that a new version (1.11) is available and downloaded and extracted:
```
$ ls -l
total 17003000
-rwxrwxr-x 1 eace eace 2195573725 May 31 23:55 crc
-rw-rw-r-- 1 eace eace 2179670828 Jun  2 12:38 crc-linux-amd64.tar.xz
```
Before downloading, I removed the 1.8 VM with `crc delete`. Then:
```
$ ./crc version
crc version: 1.11.0+883ca49
OpenShift version: 4.4.5 (embedded in binary)
```
Then:
```
$ ./crc setup --log-level=debug
INFO Checking if oc binary is cached              
DEBU oc binary already cached                     
INFO Checking if podman remote binary is cached   
DEBU Currently podman remote is not supported     
INFO Checking if goodhosts binary is cached       
DEBU goodhost binary already cached               
DEBU /home/eace/.crc/bin/goodhosts does not have the SUID bit set (-r-x------) 
INFO Caching goodhosts binary                     
DEBU goodhost binary cached                       
DEBU Making /home/eace/.crc/bin/goodhosts suid    
INFO Will use root access: change ownership of /home/eace/.crc/bin/goodhosts 
DEBU Running '/usr/bin/sudo chown root:wheel /home/eace/.crc/bin/goodhosts' 
DEBU Command failed: exit status 1                
DEBU stdout:                                      
DEBU stderr: chown: invalid group: ‘root:wheel’   
FATA Unable to set ownership of /home/eace/.crc/bin/goodhosts to root:wheel:  exit status 1: chown: invalid group: ‘root:wheel’ 
```
So:
```
$ mv /home/eace/.crc/bin/goodhosts /home/eace/.crc/bin/goodhosts.save
eace@ubuntu-desktop:~/codeready$ ./crc setup --log-level=debug
INFO Checking if oc binary is cached              
DEBU oc binary already cached                     
INFO Checking if podman remote binary is cached   
DEBU Currently podman remote is not supported     
INFO Checking if goodhosts binary is cached       
DEBU goodhost binary is not cached                
INFO Caching goodhosts binary                     
DEBU Trying to extract oc from crc binary         
[...]

PR #1280, opened 2020-06-03
Issue #1278: Don't use 'wheel' group in setSuid 

cfergeau:
This group does not exist on ubuntu, and the user does not need to be
part of any specific group since commit 70800d0 'Add execute
permission for users which are not wheel group'.

Best way to test would be to run `crc setup` on Ubuntu and make sure `~/.crc/bin/goodhosts` can be run, an alternative is to `rm -rf ~/.crc/bin && crc setup` and then check with `ls -al ~/.crc/bin/goodhosts` that the binary is not part of the wheel group.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_nonwin.go
+++ b/pkg/crc/preflight/preflight_checks_nonwin.go
@@ -58,7 +58,7 @@ func setSuid(path string) error {
 	}
 
 	/* Can't do this before the chown as the chown will reset the suid bit */
-	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s,g+x,o+x", path)
+	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s,g+x", path)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_checks_nonwin.go
```diff
@@ -51,14 +51,14 @@ func extractBinary(binaryName string, mode os.FileMode) (string, error) {
...
+	stdOut, stdErr, err := crcos.RunWithPrivilege(fmt.Sprintf("change ownership of %s", path), "chown", "root", path)
 	if err != nil {
-		return fmt.Errorf("Unable to set ownership of %s to root:wheel: %s %v: %s",
+		return fmt.Errorf("Unable to set ownership of %s to root: %s %v: %s",
 			path, stdOut, err, stdErr)
 	}
 
 	/* Can't do this before the chown as the chown will reset the suid bit */
-	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s,g+x,o+x", path)
+	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("set suid for %s", path), "chmod", "u+s,g+x", path)
```
praveenkumar, line 61: 👍

PR #1282, opened 2020-06-03
Refactor cache code

praveenkumar:
This is mostly refactor of the current code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 BUNDLE_VERSION = 4.5.0-rc.5
+# OC_VERSION and BUNDLE_VERSION are going to same for release artifacts but
+# different for nightly and CI bits where bundle version would be any random
+# string or dd-mm-yyyy format.
+OC_VERSION ?= ${BUNDLE_VERSION}
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.12.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
@@ -57,6 +61,7 @@ __check_defined = \
 # Linker flags
 VERSION_VARIABLES := -X $(REPOPATH)/pkg/crc/version.crcVersion=$(CRC_VERSION) \
     -X $(REPOPATH)/pkg/crc/version.bundleVersion=$(BUNDLE_VERSION) \
+    -X $(REPOPATH)/pkg/crc/version.ocVersion=$(OC_VERSION) \
 	-X $(REPOPATH)/pkg/crc/version.commitSha=$(COMMIT_SHA)
 
 BUNDLE_EMBEDDED := -X $(REPOPATH)/pkg/crc/constants.bundleEmbedded=true
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -85,6 +85,9 @@ func (c *Cache) CacheBinary() error {
 
 	for _, extractedFilePath := range extractedFiles {
 		finalBinaryPath := filepath.Join(c.destDir, filepath.Base(extractedFilePath))
+		// If the file exists then remove it (ignore error) first before copy because with `0500` permission
+		// it is not possible to overwrite the file.
+		os.Remove(finalBinaryPath)
 		err = crcos.CopyFileContents(extractedFilePath, finalBinaryPath, 0500)
 		if err != nil {
 			return err
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -2,6 +2,11 @@ package cache
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/download"
@@ -9,9 +14,6 @@ import (
 	"github.com/code-ready/crc/pkg/extract"
 	crcos "github.com/code-ready/crc/pkg/os"
 	"github.com/pkg/errors"
-	"io/ioutil"
-	"os"
-	"path/filepath"
 )
 
 type Cache struct {
@@ -81,11 +83,20 @@ func (c *Cache) CacheBinary() error {
 		return err
 	}
 
-	// Extract the tarball and put it the cache directory.
-	extractedFiles, err := extract.UncompressWithFilter(assetTmpFile, tmpDir, false,
-		func(filename string) bool { return filepath.Base(filename) == c.binaryName })
-	if err != nil {
-		return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
+	var extractedFiles []string
+	// Check the file is tarball or not
+	if IsTarball(assetTmpFile) {
+		// Extract the tarball and put it the cache directory.
+		extractedFiles, err = extract.UncompressWithFilter(assetTmpFile, tmpDir, false,
+			func(filename string) bool { return filepath.Base(filename) == c.binaryName })
+		if err != nil {
+			return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
+		}
+	} else {
+		extractedFiles = append(extractedFiles, assetTmpFile)
+		if filepath.Base(assetTmpFile) != c.binaryName {
+			logging.Warnf("Binary name is %s but extracted file name is %s", c.binaryName, filepath.Base(assetTmpFile))
+		}
 	}
 
 	// Copy the requested asset into its final destination
@@ -104,7 +115,6 @@ func (c *Cache) CacheBinary() error {
 			return err
 		}
 	}
-
 	return nil
 }
 
@@ -134,4 +144,14 @@ func (c *Cache) CheckVersion() error {
 		return &VersionMismatchError{CurrentVersion: currentVersion, ExpectedVersion: c.version}
 	}
 	return nil
+}
+
+func IsTarball(filename string) bool {
+	tarballExtensions := []string{".tar", ".tar.gz", ".tar.xz", ".zip", ".tar.bz2", ".crcbundle"}
+	for _, extension := range tarballExtensions {
+		if strings.HasSuffix(strings.ToLower(filename), extension) {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -24,16 +24,16 @@ func New(binarName string, archiveURL string, destDir string) *Cache {
 	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
 }
 
-func NewOcCache(destDir string) *Cache {
-	return New(constants.OcBinaryName, constants.GetOcUrl(), destDir)
+func NewOcCache() *Cache {
+	return New(constants.OcBinaryName, constants.GetOcUrl(), constants.CrcOcBinDir)
 }
 
-func NewPodmanCache(destDir string) *Cache {
-	return New(constants.PodmanBinaryName, constants.GetPodmanUrl(), destDir)
+func NewPodmanCache() *Cache {
+	return New(constants.PodmanBinaryName, constants.GetPodmanUrl(), constants.CrcBinDir)
 }
 
-func NewGoodhostsCache(destDir string) *Cache {
-	return New(constants.GoodhostsBinaryName, constants.GetGoodhostsUrl(), destDir)
+func NewGoodhostsCache() *Cache {
+	return New(constants.GoodhostsBinaryName, constants.GetGoodhostsUrl(), constants.CrcBinDir)
 }
 
 func (c *Cache) IsCached() bool {
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -1,10 +1,7 @@
 package cache
 
 import (
-	"io/ioutil"
-	"os"
-	"path/filepath"
-
+	"fmt"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/download"
@@ -12,28 +9,42 @@ import (
 	"github.com/code-ready/crc/pkg/extract"
 	crcos "github.com/code-ready/crc/pkg/os"
 	"github.com/pkg/errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 )
 
 type Cache struct {
 	binaryName string
 	archiveURL string
 	destDir    string
+	version    string
+	getVersion func() (string, error)
 }
 
-func New(binarName string, archiveURL string, destDir string) *Cache {
-	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
+type VersionMismatchError struct {
+	ExpectedVersion string
+	CurrentVersion  string
 }
 
-func NewOcCache() *Cache {
-	return New(constants.OcBinaryName, constants.GetOcUrl(), constants.CrcOcBinDir)
+func (e *VersionMismatchError) Error() string {
+	return fmt.Sprintf("expected: %s but got: %s", e.ExpectedVersion, e.CurrentVersion)
 }
 
-func NewPodmanCache() *Cache {
-	return New(constants.PodmanBinaryName, constants.GetPodmanUrl(), constants.CrcBinDir)
+func New(binaryName string, archiveURL string, destDir string, version string, getVersion func() (string, error)) *Cache {
+	return &Cache{binaryName: binaryName, archiveURL: archiveURL, destDir: destDir, version: version, getVersion: getVersion}
 }
 
-func NewGoodhostsCache() *Cache {
-	return New(constants.GoodhostsBinaryName, constants.GetGoodhostsUrl(), constants.CrcBinDir)
+func NewOcCache(version string, getVersion func() (string, error)) *Cache {
+	return New(constants.OcBinaryName, constants.GetOcUrl(), constants.CrcOcBinDir, version, getVersion)
+}
+
+func NewPodmanCache(version string, getVersion func() (string, error)) *Cache {
+	return New(constants.PodmanBinaryName, constants.GetPodmanUrl(), constants.CrcBinDir, version, getVersion)
+}
+
+func NewGoodhostsCache(version string, getVersion func() (string, error)) *Cache {
+	return New(constants.GoodhostsBinaryName, constants.GetGoodhostsUrl(), constants.CrcBinDir, version, getVersion)
 }
 
 func (c *Cache) IsCached() bool {
@@ -44,7 +55,7 @@ func (c *Cache) IsCached() bool {
 }
 
 func (c *Cache) EnsureIsCached() error {
-	if !c.IsCached() {
+	if !c.IsCached() || c.CheckVersion() != nil {
 		err := c.CacheBinary()
 		if err != nil {
 			return err
@@ -55,7 +66,7 @@ func (c *Cache) EnsureIsCached() error {
 
 // CacheBinary downloads and caches the requested binary into the CRC directory
 func (c *Cache) CacheBinary() error {
-	if c.IsCached() {
+	if c.IsCached() && c.CheckVersion() == nil {
 		return nil
 	}
 
@@ -108,4 +119,19 @@ func (c *Cache) getBinary(destDir string) (string, error) {
 	}
 
 	return destPath, err
+}
+
+func (c *Cache) CheckVersion() error {
+	// Check if version string is non-empty
+	if c.version == "" {
+		return nil
+	}
+	currentVersion, err := c.getVersion()
+	if err != nil {
+		return err
+	}
+	if currentVersion != c.version {
+		return &VersionMismatchError{CurrentVersion: currentVersion, ExpectedVersion: c.version}
+	}
+	return nil
 }
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -5,11 +5,14 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/cache"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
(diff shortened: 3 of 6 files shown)


pkg/crc/cache/cache.go
```diff
@@ -130,3 +135,13 @@ func (c *Cache) CheckVersion() error {
 	}
 	return nil
 }
+
+func IsTarball(filename string) bool {
+	tarballExtensions := []string{".tar", ".tar.gz", ".tar.xz", ".zip", ".crcbundle"}
+	for _, extension := range tarballExtensions {
+		if strings.HasSuffix(path.Base(filename), extension) {
```
cfergeau, line 142: I would do a case-insensitive comparison. Maybe add tar.bz2 to the list of archives? (I forgot if we handle this or not for now).

cfergeau: Wondering if rather than detecting tarball vs uncompressed, we should not be making sure that all the binaries we get are compressed?

pkg/crc/cache/cache.go
```diff
@@ -18,22 +20,24 @@ type Cache struct {
...
 
-func New(binarName string, archiveURL string, destDir string) *Cache {
-	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
+func New(binarName string, archiveURL string, destDir string, version string, versionArg []string) *Cache {
+	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir, version: version, versionArg: versionArg}
 }
 
-func NewOcCache(destDir string) *Cache {
-	return New(constants.OcBinaryName, constants.GetOcUrl(), destDir)
+func NewOcCache(destDir string, version string, versionArg []string) *Cache {
```
cfergeau, line 32: I should have thought of this before, but I don't think having `destDir` as a `NewOcCache` parameter is really useful, we want the `oc` binary to go to a welll known location, we won't have some parts of the code putting it in /some/dir, and other parts of the code putting it in /other/dir. We could add a `GetBinaryPath()` or such to this package.

pkg/crc/cache/cache.go
```diff
@@ -106,3 +110,20 @@ func (c *Cache) getBinary(destDir string) (string, error) {
...
 }
+
+func (c *Cache) CheckVersion() error {
+	// Check if version string is non-empty
+	if c.version == "" {
+		return nil
+	}
+	// Check the version of a binary matches to supported one
+	binaryPath := filepath.Join(c.destDir, c.binaryName)
+	stdOut, stdErr, err := crcos.RunWithDefaultLocale(binaryPath, c.versionArg...)
```
cfergeau, line 129: My feeling is that it would be more flexible to pass a `GetVersion` method rather than using `c.VersionArg`. This would give us the opportunity to do more sophisticated version extraction if needed, instead of just `strings.Contains`

cfergeau:
Forgot to mention, this is a good opportunity to add tests for the Cache package.

pkg/crc/cache/cache.go
praveenkumar: > I would do a case-insensitive comparison.
Sure we can add it.

>  Maybe add tar.bz2 to the list of archives? (I forgot if we handle this or not for now).

No, atm we don't handle it as part of extract.

pkg/crc/cache/cache.go
praveenkumar: @cfergeau this makes it really explicit, we we add `GetBinaryPath()` then how you want to handle podman, oc  or any other future binary which will have it's own path? or you want to provide the path as per binaryName?

pkg/crc/cache/cache.go
praveenkumar: @cfergeau so `GetVersion` just provide the curent version of binary or will it do something more? `strings.Contains` is part of `CheckVersion` where we compare what is installed vs what we expect.

cfergeau: It would return the current version of the binary (which could come from running the binary), then `CheckVersion` could use a semantic version check instead, or use `==` directly, rather than a very wide `strings.Contains`.

pkg/crc/cache/cache.go
cfergeau: `func (c *Cache) GetBinaryPath() (string, error)` this would be a method, so it would be `ocCache.GetBinaryPath()`, `podmanCache.GetBinaryPath()`, ... It would not do a lot more than `return filepath.join(c.DestDir, c.BinaryName)` (approximate syntax/implementation :)

cfergeau:
retest this please.

pkg/crc/cache/cache.go
praveenkumar: @cfergeau the only way semantic version check works if each binary we use follows it, currently even the tray bits we control for windows not able to follow semantic versioning.

pkg/crc/cache/cache.go
praveenkumar: @cfergeau `return filepath.join(c.DestDir, c.BinaryName)` even here we do need `destDir` and how do we get it?

cfergeau:
```
diff --git a/pkg/crc/cache/cache.go b/pkg/crc/cache/cache.go
index 72a5babc..6b54f826 100644
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -24,8 +24,8 @@ func New(binarName string, archiveURL string, destDir string) *Cache {
        return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
 }
 
-func NewOcCache(destDir string) *Cache {
-       return New(constants.OcBinaryName, constants.GetOcUrl(), destDir)
+func NewOcCache() *Cache {
+       return New(constants.OcBinaryName, constants.GetOcUrl(), constants.CrcOcBinDir)
 }
 
 func NewPodmanCache(destDir string) *Cache {
@@ -36,6 +36,10 @@ func NewGoodhostsCache(destDir string) *Cache {
        return New(constants.GoodhostsBinaryName, constants.GetGoodhostsUrl(), destDir)
 }
 
+func (c *Cache) GetBinaryPath() string {
+       return filepath.Join(c.destDir, c.binaryName)
+}
+
 func (c *Cache) IsCached() bool {
        if _, err := os.Stat(filepath.Join(c.destDir, c.binaryName)); os.IsNotExist(err) {
                return false
diff --git a/pkg/crc/preflight/preflight_checks_common.go b/pkg/crc/preflight/preflight_checks_common.go
index 0fa00942..5f22d5d1 100644
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -71,7 +71,7 @@ func checkOcBinaryCached() error {
        // We should remove this code after 3-4 releases. (after 2020-07-10)
        os.Remove(filepath.Join(constants.CrcBinDir, "oc"))
 
-       oc := cache.NewOcCache(constants.CrcOcBinDir)
+       oc := cache.NewOcCache()
        if !oc.IsCached() {
                return errors.New("oc binary is not cached")
        }
@@ -80,7 +80,7 @@ func checkOcBinaryCached() error {
 }
 
 func fixOcBinaryCached() error {
-       oc := cache.NewOcCache(constants.CrcOcBinDir)
+       oc := cache.NewOcCache()
        if err := oc.EnsureIsCached(); err != nil {
                return fmt.Errorf("Unable to download oc %v", err)
        }
```
really 2 separate suggestions, we should imo be removing `destDir` from `NewOcCache` and similar methods. The `GetBinaryPath()` suggestion is more a "would such an API be useful?" question.

pkg/crc/cache/cache.go
cfergeau: I think the tray versioning is valid semantic versioning, except that the semantic spec explicitly says these are alpha/test releases and don't really define how to compare them ;) Comparing with `==` is still better than `strings.Contains` in my opinion.

pkg/crc/cache/cache.go
```diff
@@ -69,12 +74,17 @@ func (c *Cache) CacheBinary() error {
...
-	// Extract the tarball and put it the cache directory.
-	extractedFiles, err := extract.UncompressWithFilter(assetTmpFile, tmpDir,
-		func(filename string) bool { return filepath.Base(filename) == c.binaryName })
-	if err != nil {
-		return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
+	var extractedFiles []string
+	// Check the file is tarball or not
+	if IsTarball(assetTmpFile) {
+		// Extract the tarball and put it the cache directory.
+		extractedFiles, err = extract.UncompressWithFilter(assetTmpFile, tmpDir,
```
guillaumerose, line 81: Let say I don't have oc and goodhost on my machine, will it unzip two times the archive ?

pkg/crc/cache/cache.go
```diff
@@ -85,12 +95,14 @@ func (c *Cache) CacheBinary() error {
 
 	for _, extractedFilePath := range extractedFiles {
```
guillaumerose, line 96: At this point, I think extractedFiles will always have a length of 1 or 0. Maybe we can avoid the loop ?

pkg/crc/cache/cache.go
```diff
@@ -123,3 +130,13 @@ func (c *Cache) CheckVersion() error {
 	}
 	return nil
 }
+
+func IsTarball(filename string) bool {
+	tarballExtensions := []string{".tar", ".tar.gz", ".tar.xz", ".zip", ".tar.bz2", ".crcbundle"}
+	for _, extension := range tarballExtensions {
+		if strings.HasSuffix(strings.ToLower(path.Base(filename)), extension) {
```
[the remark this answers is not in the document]
guillaumerose: You can look at `filepath.Ext` also.

pkg/crc/cache/cache.go
praveenkumar: @guillaumerose no it shouldn't extract it twice, this logic check if the file is a tarball type or not if yes the only it will run the extract otherwise just download.

pkg/crc/cache/cache.go
praveenkumar: @guillaumerose we put this so in future if we want to have more files like shasum or License ext. to be extracted.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -244,37 +243,24 @@ func fixLibvirtServiceRunning() error {
...
+		return err
 	}
-	logging.Debugf("%s is already installed in %s", libvirt.MachineDriverCommand, libvirtDriverPath)
+	logging.Debugf("%s is already installed", libvirt.MachineDriverCommand)
 	return nil
 }
 
 func fixMachineDriverLibvirtInstalled() error {
 	logging.Debugf("Installing %s", libvirt.MachineDriverCommand)
-	_, err := extractBinary(libvirt.MachineDriverCommand, 0755)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau yes that I know, I kept it just this way to have initial structure reviewed since on the Mac side also after refactor this will not needed.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -244,37 +243,24 @@ func fixLibvirtServiceRunning() error {
 func checkMachineDriverLibvirtInstalled() error {
 	logging.Debugf("Checking if %s is installed", libvirt.MachineDriverCommand)
 
-	// Check if crc-driver-libvirt is available
-	libvirtDriverPath := filepath.Join(constants.CrcBinDir, libvirt.MachineDriverCommand)
-	err := unix.Access(libvirtDriverPath, unix.X_OK)
-	if err != nil {
-		return fmt.Errorf("%s is not executable", libvirtDriverPath)
-	}
-
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau atm no, we should add that also.

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
praveenkumar: yes I think we can use the filepath.Ext

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -124,3 +130,12 @@ func fixGoodhostsBinaryCached() error {
 	logging.Debug("goodhost binary cached")
 	return setSuid(goodhostPath)
 }
+
+func getCurrentOcversion() (string, error) {
+	ocPath := filepath.Join(constants.CrcOcBinDir, constants.OcBinaryName)
+	stdOut, _, err := crcos.RunWithDefaultLocale(ocPath, "version", "--client")
```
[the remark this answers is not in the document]
praveenkumar: > wondering if the parsing could be simpler/more robust with -oyaml or -ojson? 

yes we can use the json or yaml output and put into the struct to get just the version info but even with current scenario it is not that bad.

> On the other hand, this makes it very similar to getCurrentLibvirtDriverVersion() introduced later which could have its use.

Both have different argument to get the version, and that was initially the point to pass function for version info.

pkg/crc/cache/cache.go
```diff
@@ -106,3 +108,15 @@ func (c *Cache) getBinary(destDir string) (string, error) {
...
 }
+
+func (c *Cache) CheckVersion() error {
+	// Check if version string is non-empty
+	if c.version == "" {
+		return nil
+	}
+	currentVersion, err := c.getVerison()
+	if currentVersion != c.version {
+		return fmt.Errorf("%s does not have right version \n Required: %s \n Got: %s use 'crc setup' command.\n%v\n", c.binaryName, c.version, currentVersion, err)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau as of now whole Cache implementation is used as part of preflight and setup so I think this is good way to return consistent error in case of mismatch or you just want to have error till it says `required: foo but Got: bar` ?

pkg/crc/cache/cache.go
```diff
@@ -1,39 +1,41 @@
...
 	binaryName string
 	archiveURL string
 	destDir    string
+	version    string
+	getVerison func() (string, error)
 }
 
-func New(binarName string, archiveURL string, destDir string) *Cache {
-	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
+func New(binarName string, archiveURL string, destDir string, version string, getVerison func() (string, error)) *Cache {
```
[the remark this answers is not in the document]
praveenkumar: Ack

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
cfergeau: For `.tar.gz`/`.tar.xz`/`.tar.bz2`, `filepath.Ext` is not going to do the right thing.

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
cfergeau: I'd consider adding an error type
```
type VersionMismatchError {
    ExpectedVersion string
    CurrentVersion string
}
func (e *VersionMismatchError) Error() string { return fmt.Sprintf("expected: %s but got: %s", e.ExpectedVersion, e.CurrentVersion) }
```
This way, the higher level code can easily generate `fmt.Errorf("%s does not have right version \n Required: %s \n Got: %s use 'crc setup' command.\n%v\n", c.binaryName, c.version, currentVersion, err)` if it wants.

pkg/crc/preflight/preflight_checks_linux.go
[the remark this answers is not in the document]
cfergeau: If CI was functional, the PR would not pass CI :)

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
praveenkumar: we can just check `.gz` `.xz` and `.bz2` extensions instead using fully `.tar.gz` right?

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau in that case in all the high level functions (caller of `CheckVersion` ) need to add same kind of string or create another variable which take this error message and append the other string.

pkg/crc/preflight/preflight_checks_common.go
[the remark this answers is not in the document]
cfergeau: > Both have different argument to get the version, and that was initially the point to pass function for version info.

Yup, only thinking out loud that maybe it will be useful that they are similar so that we can use something like `func() { return genericGetVersion("version", "--client") }` as the `getVersion` method.

pkg/crc/cache/cache.go
guillaumerose: oh thanks! I thought it would extract the crcbundle but it's in fact the archive containing the bits in the upstream mirror.

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
cfergeau: `foo.gz` is a single file compressed with gzip, `foo.tar.gz` is a tar archive (multiple files) compressed with gzip, passing `foo.gz` to `uncompress` would fail with `"Unknown file format when trying to uncompress %s"`

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
cfergeau: Not sure I understand your objection? If the higher level code is fine with a generic message, `e.Error()` will return such a message (`expected: %s but got: %s"`). If they want a more specific message more appropriate in the context of the higher level code, they can build one (ie they ignore `e.Error()` and use the `ExpectedVersion`/`CurrentVersion` fields).

pkg/crc/cache/cache.go
```diff
@@ -1,39 +1,41 @@
...
+	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir, version: version, getVerison: getVerison}
 }
 
-func NewOcCache(destDir string) *Cache {
-	return New(constants.OcBinaryName, constants.GetOcUrl(), destDir)
+func NewOcCache(destDir string, version string, getVersion func() (string, error)) *Cache {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau version is something get from the constant but getVersion is not something constant and need to be passed. Even for version assume in case of libvirt driver to get that constant we need to import the libvrit driver constant to cache which doesn't look good. so I will keep it as it is for now.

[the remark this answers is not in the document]
cfergeau: >  we need to import the libvrit driver constant to cache which doesn't look good.
 
The miscellaneous specialized OcCache, PodmanCache, ... already don't look that great in the generic cache package, so I would not mind getting the constant from where we need it. These specialized caches could be split off in their own file to 'hide' these dependencies from the main cache file :)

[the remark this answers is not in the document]
cfergeau: And I don't think it's that great to export a method with arguments which are always going to be the same...
```
pkg/crc/preflight/preflight_checks_common.go:   oc := cache.NewOcCache(version.GetOcVersion(), getCurrentOcversion)
pkg/crc/preflight/preflight_checks_common.go:   oc := cache.NewOcCache(version.GetOcVersion(), getCurrentOcversion)
```
This just makes the method more complicated to use, the occache package is harder to understand as you need to look in multiple places for a clear image, ...

pkg/crc/cache/cache.go
```diff
@@ -80,12 +82,17 @@ func (c *Cache) CacheBinary() error {
...
+	// Check the file is tarball or not
+	if IsTarball(assetTmpFile) {
+		// Extract the tarball and put it the cache directory.
+		extractedFiles, err = extract.UncompressWithFilter(assetTmpFile, tmpDir,
+			func(filename string) bool { return filepath.Base(filename) == c.binaryName })
+		if err != nil {
+			return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
+		}
+	} else {
+		extractedFiles = append(extractedFiles, assetTmpFile)
```
cfergeau, line 95: I think this should at least warn if `assetTmpFile` does not match `c.binaryName`?

Makefile
```diff
@@ -13,6 +13,10 @@
 # limitations under the License.
 
 BUNDLE_VERSION = 4.4.8
+# OC_VERSION and BUNDLE_VERSION are going to same for release artifacts but
+# different for nightly and CI bits where bundle version would be any random
+# string or dd-mm-yyyy format.
+OC_VERSION = 4.4.8
```
cfergeau, line 19: This is better done as `OC_VERSION ?= $(BUNDLE_VERSION)` to avoid duplicating the bundle version a 3rd time. You can then run `make OC_VERSION=4.4.10` when you want to override it. However, it might be better to detect that we are in a CI/nightly context as opposed to a release context, and then use either `latest-4.x` or maybe https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/4.6.0-0.nightly-2020-06-30-020342/ to also test newest `oc` code and catch regressions early.

pkg/crc/cache/cache.go
praveenkumar: @cfergeau Added the warn message.

Makefile
praveenkumar: > This is better done as OC_VERSION ?= $(BUNDLE_VERSION) to avoid duplicating the bundle version a 3rd time. You can then run make OC_VERSION=4.4.10 when you want to override it. 

Done

>  However, it might be better to detect that we are in a CI/nightly context as opposed to a release context, and then use either latest-4.x or maybe https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/4.6.0-0.nightly-2020-06-30-020342/ to also test newest oc code and catch regressions early.

yes we have to think that if this is possible because the base url is change here so to support that kind of thing we also need to make the base url configurable from the makefile.

cfergeau: >  we also need to make the base url configurable from the makefile.

In my opinion, it's more useful to have a release/nightly toggle, or even autodetection in the Makefile.

praveenkumar: > In my opinion, it's more useful to have a release/nightly toggle, or even autodetection in the Makefile.

@cfergeau yes but somewhere we need to mention it, even we do release/nightly toggle then download url also need respect that choice, anyway we can take this as separate PR.

cfergeau: Yes, I'm still of the opinion that even this OC_VERSION Makefile variable should have been part of this separate PR which would focus on adding CI/nightly support rather than slipping it in this not directly related PR.

praveenkumar: @cfergeau filled https://github.com/code-ready/crc/issues/1362 one to track it.

pkg/crc/cache/cache.go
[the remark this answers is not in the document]
cfergeau: I don't think we ever reached a consensus on this one...

Issue #1275, opened 2020-06-02
Use functions from cache pkg to download tray


anjannath:
### General information

  * OS: Windows
  

`crc setup --enable-experimental-features` is not able to start the system tray in windows.

Setup looks for the tray in `~\.crc\bin\tray-windows` but the tray executable is downloaded/extracted directly inside `~\.crc\bin\` without the `tray-windows` directory.

anjannath:
The bug is fixed now, but need to refactor the tray download code to use cache pkg

PR #1285, opened 2020-06-04
[WIP] Fix tray download location in crc home folder

anjannath:
**Fixes:** Issue #1275 

## Testing

`crc setup --enable-experimental-features` should start the tray and daemon services successfully

### DIFF
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -20,8 +20,8 @@ type Cache struct {
 	destDir    string
 }
 
-func New(binarName string, archiveURL string, destDir string) *Cache {
-	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
+func New(binaryName string, archiveURL string, destDir string) *Cache {
+	return &Cache{binaryName: binaryName, archiveURL: archiveURL, destDir: destDir}
 }
 
 func NewOcCache(destDir string) *Cache {
--- a/pkg/crc/cache/cache_tray_windows.go
+++ b/pkg/crc/cache/cache_tray_windows.go
@@ -1,0 +1,59 @@
+package cache
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/extract"
+	crcos "github.com/code-ready/crc/pkg/os"
+	"github.com/pkg/errors"
+)
+
+type TrayCache struct {
+	Cache
+}
+
+func (tc *TrayCache) CacheBinary() error {
+	if tc.IsCached() {
+		return nil
+	}
+
+	// Create tmp dir to download the requested tarball
+	tmpDir, err := ioutil.TempDir("", "crc")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpDir)
+	assetTmpFile, err := tc.getBinary(tmpDir)
+	if err != nil {
+		return err
+	}
+
+	// Extract the tarball and put it the cache directory.
+	extractedFiles, err := extract.Uncompress(assetTmpFile, tmpDir)
+	if err != nil {
+		return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
+	}
+
+	// Copy the requested asset into its final destination
+	err = os.MkdirAll(tc.destDir, 0750)
+	if err != nil && !os.IsExist(err) {
+		return errors.Wrap(err, "Cannot create the target directory.")
+	}
+
+	for _, extractedFilePath := range extractedFiles {
+		finalBinaryPath := filepath.Join(tc.destDir, filepath.Base(extractedFilePath))
+		err = crcos.CopyFileContents(extractedFilePath, finalBinaryPath, 0500)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func NewWindowsTrayCache(destDir string) *TrayCache {
+	return &TrayCache{Cache{binaryName: constants.TrayBinaryName, archiveURL: constants.GetCRCWindowsTrayDownloadURL(), destDir: destDir}}
+}
--- a/pkg/crc/preflight/preflight_checks_tray_windows.go
+++ b/pkg/crc/preflight/preflight_checks_tray_windows.go
@@ -2,19 +2,16 @@ package preflight
 
 import (
 	"fmt"
-	"io/ioutil"
 	goos "os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 
+	"github.com/code-ready/crc/pkg/crc/cache"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/input"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/version"
-	dl "github.com/code-ready/crc/pkg/download"
-	"github.com/code-ready/crc/pkg/embed"
-	"github.com/code-ready/crc/pkg/extract"
 	"github.com/code-ready/crc/pkg/os"
 	"github.com/code-ready/crc/pkg/os/windows/powershell"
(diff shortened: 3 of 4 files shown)


pkg/crc/cache/cache_tray_windows.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/code-ready/crc/pkg/extract"
+	crcos "github.com/code-ready/crc/pkg/os"
+	"github.com/pkg/errors"
+)
+
+type TrayCache struct {
+	Cache
+}
+
+func (tc *TrayCache) CacheBinary() error {
```
anjannath, line 18: The original `CacheBinary()` function only extracts a single binary that matches the filter, for the tray we want to extract the complete directory

cfergeau: I don't think you need to duplicate the whole function when there's only a slight difference:
`extractedFiles, err := extract.Uncompress(assetTmpFile, tmpDir)` VS `extractedFiles, err := extract.UncompressWithFilter(assetTmpFile, tmpDir, ...)` You could have an internal helper with more args (a file filter) which would be used by both `CacheBinary` and `TrayCache.CacheBinary`. I don't think you need `TrayCache` at all, it could probably be called `TrayCache.CacheBinary` could probably be `Cache.CacheDir` or something like that?

pkg/crc/cache/cache.go
```diff
@@ -20,8 +20,8 @@ type Cache struct {
 	destDir    string
 }
 
-func New(binarName string, archiveURL string, destDir string) *Cache {
-	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
+func New(binaryName string, archiveURL string, destDir string) *Cache {
+	return &Cache{binaryName: binaryName, archiveURL: archiveURL, destDir: destDir}
 }
```
cfergeau, line 25: nit: This typo fix is totally unrelated to these changes, only mentioning it since it took me a while to notice `binarName` vs `binaryName`

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -148,6 +113,12 @@ func checkTrayBinaryVersion() error {
 }
 
 func fixTrayBinaryVersion() error {
+	// Kill older running tray
+	if err := checkTrayRunning(); err == nil {
+		if err := stopTray(); err != nil {
+			logging.Debugf("Failed to kill running tray: %v", err)
+		}
+	}
```
cfergeau, line 121: This code used to be in `fixTrayBinaryExists`, not clear why it moved.

pkg/crc/cache/cache_tray_windows.go
anjannath: I avoided to add a `cache.CacheDir` because i wasn't sure if calling it on the other objects like `podmanCache` would result in a side effect..

pkg/crc/cache/cache.go
anjannath: will put this is in a separate commit

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Yes, but it should have been here all along, we can only expect a running tray when there is one available, and only when it is available and the versions don't match we need to update the version in which case we'll have to stop the running tray, in normal case when `fixTrayBinaryExists` is called, it'll most probably not have a running tray, as that binary doesn't exist yet

pkg/crc/cache/cache_tray_windows.go
anjannath: @cfergeau Is this what you were suggesting..
```diff
diff --git a/pkg/crc/cache/cache.go b/pkg/crc/cache/cache.go
index 72a5bab..28cf083 100644
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -20,8 +20,8 @@ type Cache struct {
 	destDir    string
 }
 
-func New(binarName string, archiveURL string, destDir string) *Cache {
-	return &Cache{binaryName: binarName, archiveURL: archiveURL, destDir: destDir}
+func New(binaryName string, archiveURL string, destDir string) *Cache {
+	return &Cache{binaryName: binaryName, archiveURL: archiveURL, destDir: destDir}
 }
 
 func NewOcCache(destDir string) *Cache {
@@ -58,7 +58,20 @@ func (c *Cache) CacheBinary() error {
 	if c.IsCached() {
 		return nil
 	}
+	return c.cache(func(filename string) bool {
+		return filepath.Base(filename) == c.binaryName
+	})
+}
 
+// CacheDir downloads and caches the requested asset's dir into crc home
+func (c *Cache) CacheDir() error {
+	if c.IsCached() {
+		return nil
+	}
+	return c.cache(nil)
+}
+
+func (c *Cache) cache(fileFilter func(string) bool) error {
 	// Create tmp dir to download the requested tarball
 	tmpDir, err := ioutil.TempDir("", "crc")
 	if err != nil {
@@ -70,9 +83,13 @@ func (c *Cache) CacheBinary() error {
 		return err
 	}
 
-	// Extract the tarball and put it the cache directory.
-	extractedFiles, err := extract.UncompressWithFilter(assetTmpFile, tmpDir,
-		func(filename string) bool { return filepath.Base(filename) == c.binaryName })
+	// Extract the tarball and put it in the CrcBin directory.
+	extractedFiles, err := extract.UncompressWithFilter(
+		assetTmpFile,
+		tmpDir,
+		fileFilter,
+	)
+
 	if err != nil {
 		return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
 	}
diff --git a/pkg/crc/cache/cache_tray_windows.go b/pkg/crc/cache/cache_tray_windows.go
new file mode 100644
index 0000000..17010db
--- /dev/null
+++ b/pkg/crc/cache/cache_tray_windows.go
@@ -0,0 +1,7 @@
+package cache
+
+import "github.com/code-ready/crc/pkg/crc/constants"
+
+func NewWindowsTrayCache(destDir string) *Cache {
+	return New(constants.TrayBinaryName, constants.GetCRCWindowsTrayDownloadURL(), destDir)
+}

```

pkg/crc/preflight/preflight_checks_tray_windows.go
cfergeau: can you move this to its own commit? The change to use `Cache` should be purely code moving around, with no behaviour change, the explanation you just gave deserves to be in a git commit log :)

pkg/crc/cache/cache_tray_windows.go
cfergeau: > @cfergeau Is this what you were suggesting..

If `UncompressWithFilter` knows how to deal with a `nil` filter, yes, looks like it! Maybe move the `c.IsCached()` to the helper too?

> The original `CacheBinary()` function only extracts a single binary that matches the filter, for the tray we want to extract the complete directory

Did not occur to me initially, but this can also be added to the commit log, so that future you (or other readers) can remember why this was done this way. What is in that directory though in addition to the tray binary?

> I avoided to add a `cache.CacheDir` because i wasn't sure if calling it on the other objects like `podmanCache` would result in a side effect..

Hmm if someone tries to use this on an inappropriate instance, I would say it's their problem/their bug. We want to avoid non-generic code in this package (OcCache/PodmanCache/... should not have been added there, this was requested during the review but slipped through), so adding a non-generic TrayCache class is going the wrong way. So let's try to add the helper you need to the base class, we can refine the API later if it's not good enough.

anjannath: > What is in that directory though in addition to the tray binary?

In case of windows there's a dll of a json parser library, that am using and an app config file, in case of macOS the whole app is actually a directory, but macOS would recognize it as an app bundle, it also contains some app manifests, static resources like the icons.

anjannath: @cfergeau one thing i missed in that diff, we'd also need something like `cache.EnsureDirIsCached`

anjannath:
Closing as this might need more work after #1282

PR #1290, opened 2020-06-05
Misc logs (and more) improvements

cfergeau:
Some fixes/log improvements I've accumulated over time.  Most of them are related to log messages. I've tried to adjust the error when certificates expire, but I'm not fully satisfied with how it's worded, mostly because I was not sure if I should use very long/multiline error strings.
I'm fine with splitting this in smaller pull requests if we don't want to take everything at once.


pkg/crc/errors/multierror.go
```diff
@@ -58,5 +58,12 @@ func RetryAfter(attempts int, callback func() error, d time.Duration) (err error
 		logging.Debugf("error: %v - sleeping %s", err, d)
 		time.Sleep(d)
 	}
-	return m.ToError()
+	err = m.ToError()
+
+	/* Should only be reached in case of a timeout, in which case err will be set */
+	if err == nil {
+		panic("")
+	}
```
praveenkumar, line 66: why we need to check for `err == nil` and set the panic ?

cfergeau: I think because it was not obvious to me that err would never be nil while I was reading this code, so I tried it/made it obvious. I can indeed remove this if you prefer as this leans on debug/test code.

praveenkumar: @cfergeau if err is nil the we always send the return value as nil quite early https://github.com/code-ready/crc/pull/1290/commits/8a29f67ff01aef58d0142c501f5c8282d82f50c0#diff-9c9d6b3a60f814f0535684659ec8cc14R49-R51  so  here if the err is collected it would never be the nil, better to remove it.

praveenkumar:
@cfergeau this need a rebase :(

pkg/crc/errors/multierror.go
cfergeau: Yes, `err` cannot be `nil` due to how the code is written, hence the `panic`, it's just some documentation/future proofing for this code. I'll rebase and drop it.

praveenkumar:
@cfergeau looks like another rebase because of https://github.com/code-ready/crc/pull/1317 one  :( , Also go ahead and merge it once you done with rebase

/lgtm

Issue #1242, opened 2020-05-15
[BUG] Certificates do not renew successfully using crc v1.10


btannous:
### General information

  * OS: Linux (fedora 31)
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? Yes (always)
  * Running CRC on: Desktop, 8core, 32g memory, nothing else running (accessing remotely)

## CRC version
```bash
❯ ./crc version
crc version: 1.10.0+9025021
OpenShift version: 4.4.3 (embedded in binary)
```

## Host Operating System
```bash
❯ cat /etc/os-release
NAME=Fedora
VERSION="31 (Server Edition)"
ID=fedora
VERSION_ID=31
VERSION_CODENAME=""
PLATFORM_ID="platform:f31"
PRETTY_NAME="Fedora 31 (Server Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:31"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f31/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=31
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=31
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Server Edition"
VARIANT_ID=server
```

### Steps to reproduce

- blocked UDP port 123 out of my local gateway
- sudo date -s "9 JULY 2020 11:11:11"
- downloaded/extracted crc-1.10
- crc delete <just in case i had an old one>
- crc setup
- crc start -c 8 -m 25000 ...

### Expected
Certificates get renewed and cluster comes up.

### Actual
Certificates do not get renewed.

### Logs
https://gist.github.com/dea1a60dddaf7becd05200d0a2eed3e5

https://gist.github.com/btannous/1ebb796dbb747a2f18294403aea6c15d

gbraad:
We are currently investigating this ...

cfergeau:
[...]

PR #1292, opened 2020-06-05
Update certificate recovery codebase

praveenkumar:
With 4.4 [1] automatic cert rotation works so we don't need to
rely on the manual process. We still need to approve the csr
manually which is not automated yet.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1819623

**Fixes:** Issue #1242 

## Solution/Idea

Idea is to make sure the cert recovery process works without any issue.

### DIFF
--- a/pkg/crc/machine/cert-renewal.go
+++ b/pkg/crc/machine/cert-renewal.go
@@ -28,7 +28,7 @@ func waitForPendingCsrs(oc oc.OcConfig) error {
 		return nil
 	}
 
-	return errors.RetryAfter(60, waitForPendingCsr, time.Second)
+	return errors.RetryAfter(120, waitForPendingCsr, time.Second*5)
 }
 
 func RegenerateCertificates(sshRunner *ssh.SSHRunner, machineName string) error {
--- a/pkg/crc/machine/cert-renewal.go
+++ b/pkg/crc/machine/cert-renewal.go
@@ -3,7 +3,6 @@ package machine
 import (
 	"fmt"
 	"regexp"
-	"strings"
 	"time"
 
 	"github.com/code-ready/crc/pkg/crc/errors"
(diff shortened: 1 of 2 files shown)


btannous:
@praveenkumar if this is the fix for what you mentioned in slack, tested here -- seems good! https://gist.github.com/3790f396add70b53b796f0c5913ddbe2

pkg/crc/machine/cert-renewal.go
```diff
@@ -125,12 +28,10 @@ func waitForPendingCsrs(oc oc.OcConfig) error {
 		return nil
 	}
 
-	return errors.RetryAfter(60, waitForPendingCsr, time.Second)
+	return errors.RetryAfter(60, waitForPendingCsr, time.Second*5)
```
cfergeau, line 31: I'd change this in a separate commit.

pkg/crc/machine/cert-renewal.go
```diff
@@ -140,105 +41,32 @@ func RegenerateCertificates(sshRunner *ssh.SSHRunner, machineName string) error
...
-	if err != nil {
-		logging.Debugf("Error restarting kubelet service: %v", err)
-		return err
-	}
-
+	oc := oc.UseOCWithConfig(machineName)
 	/* 2 CSRs to approve, one right after kubelet restart, the other one a few dozen seconds after
-	approving the first one */
+	approving the first one
+	- First one is requested by system:serviceaccount and address to node-bootstrapper
```
cfergeau, line 47: I don't understand what you mean by "address to xxx"

pkg/crc/machine/cert-renewal.go
```diff
@@ -140,105 +41,32 @@ func RegenerateCertificates(sshRunner *ssh.SSHRunner, machineName string) error
...
 	err = oc.ApproveNodeCSR()
 	if err != nil {
-		logging.Debugf("Error approving second pending CSR: %v", err)
+		logging.Debugf("Error approving second pending (system:node) CSR: %v", err)
 		return err
 	}
 
-	/* The cluster needs to settle for a few minutes before connections to
-	 * api.crc.testing:6443 gets a valid TLS certificate
-	 */
```
cfergeau, line 237: With the automated cert recovery, the apiserver is immediatly available?

pkg/crc/machine/cert-renewal.go
praveenkumar: @cfergeau this is more like for me to remember which cert address to which component so when I file an issue I don't need to everytime go back to working cluster and check those details.

pkg/crc/machine/cert-renewal.go
praveenkumar: @cfergeau Here we don't create the recovery apiserver which need to be deleted so our current apiserver handle it and we don't need to now check for default  api connections.

pkg/crc/machine/cert-renewal.go
praveenkumar: @cfergeau sure, will do that.

pkg/crc/machine/cert-renewal.go
```diff
@@ -13,102 +12,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/systemd"
 )
 
-var (
-	kaoImage = "openshift/cert-recovery"
```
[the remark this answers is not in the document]
praveenkumar: yes we need to remove that step from snc now, also in the snc side this is just a tag doesn't add any addition image but good to remove those.

pkg/crc/machine/cert-renewal.go
cfergeau: I don't mind having "node-bootstrapper" and the other one being mentioned, it's the "address to" formulation which I don't understand, so the comment is not as useful as it could be.

pkg/crc/machine/cert-renewal.go
cfergeau: This was not only checking if we can reach the api server, this was also checking if the certificates of the apiserver are valid. Actually I'm not sure how this all works, the certificate for the api.crc.testing:6443 connection are renewed automatically, so that we can approve the CSR for a different set of certificates?

praveenkumar: @cfergeau before the recovery cert server used to run on `7443` port so we need to delete that recovery pod after certs are created and csr are approved and then we used this to make sure our default apiserver which runs on `6443` port able to comeup, now we don't need that different recovery pod so this step also not needed.

pkg/crc/machine/cert-renewal.go
cfergeau: How much time is needed before getting the first csr?  Less than the 4/5 minutes mentioned in the log? And once we got one CSR, isn't it very long to wait for 5 more minutes?

praveenkumar: @cfergeau usually second csr we get around 30sec-1min but this is not like waiting again for 5 mins but break as soon as the csr requests comes right, 5 min is a timeout in our condition.

cfergeau: How long does the 1st one take to come up?
And if there is a failure, 5 minutes is a very wasteful wait if we know after 2 minutes that it's never going to succeed.

praveenkumar: > How long does the 1st one take to come up?

@cfergeau first one does take around 4-5 mins, didn't record exact time but around that on my testing instance.

cfergeau: If first CSR request takes 4/5 minutes to appear, and since this time is dependant on the machine running crc, isn't a 5 minutes timeout possibly going to be a bit too short for the 1st request? Might be better to add a `timeout` arg to `waitForPendingCsrs, and wait for 8 to 10 minutes (or is it too much?) for the first one, and 2 minutes for the second one.

praveenkumar: @cfergeau yesterday when @anjannath trying this PR the second CSR also took around 4min to appear so I am not sure if we put different timeout for different csr requests if you want to increase the first one to 8-10 mins then I would also wait around same time for second csr.

PR #1297, opened 2020-06-08
Issue #1286 Add log on as user permission for current user during setup

### LINKED ISSUE #1286: Permissions needed for running the daemon service under normal user
To run the daemon service under the user that is logged in, we need _LogOnAsService_ permission for the user (steps: https://www.opcsupport.com/s/article/How-do-I-configure-a-user-account-to-have-logon-as-a-service-permissions) We do this because we want to use the same `crc` homedir both for our CLI tool and the daemon.

Adding this permission is a security risk as pointed out in slack, another concern is about resource consumption, in the sense that the daemon process can keep running even when the user is logged out. (microsoft link: https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/log-on-as-a-service#vulnerability)

The alternative is to use the `LocalSystem` account to run the daemon service, but since `crc` uses the `%USERPROFILE%/.crc` as the crc homedir, using the `LocalSystem` account creates the `.crc` dir inside `C:\Windows\system32` which we don't want.

Now to fix this we either need to modify our `GetHomeDir()` to decide the homedir based on if its the daemon process (provided we decide to go with using the `LocalSystem` account)
or we keep using the current user, but find a way to automate the addition of _LogOnAsService_ privilege after considering the security and resource consumption concerns.

anjannath:
Fixes #1286 

~in order to use the LocalSystem account to run the daemon service
we also need to take care of where the .crc dir is created, as by
default when using the LocalSystem account the users home dir is at
C:\Windows\System32\config\systemprofile but we want the .crc dir to
be created in the logged in user's home directory.~

~We capture the user's home dir when crc setup is run and pass it as an
argument in the service invocation command, e.g `crc.exe daemon c:\Users\anjan`~

Dropped the commit using the `LocalSystem` account, instead we use the current user account to the run the daemon service, and also allow _log on as a service_ for the user, so that a crc user doesn't have to do it manually.

### How to test:
1. run `crc setup --enable-experimental-features --log-level debug`
2. Should have a working tray up and running
3. If errors are encountered then Open and check if `secpol.msc` > Local User Rights > Log on as service contains the current user's name.


pkg/crc/constants/constants_nonwin.go
```diff
@@ -0,0 +1,10 @@
+// +build !windows
+
+package constants
+
+import "os"
+
+// GetHomeDir returns the home directory for the current user
+func GetHomeDir() string {
```
cfergeau, line 8: How is this different from https://golang.org/pkg/os/#UserHomeDir ? Could we use this instead?

anjannath: Yup, for the non-win case we can use this..

anjannath: Done!

pkg/crc/constants/constants_windows.go
```diff
@@ -11,4 +16,32 @@ const (
...
+		homeDir := filepath.Join(homeDrive, homePath)
+		if _, err := os.Stat(homeDir); err == nil {
+			return homeDir
+		}
+	}
+	if userProfile := os.Getenv("USERPROFILE"); len(userProfile) > 0 {
+		if _, err := os.Stat(userProfile); err == nil {
+			return userProfile
+		}
+	}
```
praveenkumar, line 45: @anjannath doesn't this also replaced with `os.UserHomeDir()` as per doc it does return USERPROFILE in case of windows.

anjannath: @praveenkumar Ahh, sorry i missed it thought it was getting the homedir with some other logic, changed to use `os.UserHomeDir()`

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -96,12 +96,12 @@ func fixTrayBinaryExists() error {
 		}
 	}
 	archivePath := filepath.Join(tmpArchivePath, filepath.Base(constants.GetCRCWindowsTrayDownloadURL()))
-	outputPath := constants.CrcBinDir
-	//err = goos.MkdirAll(filepath.Join(outputPath, "tray-windows"), 0750)
+	outputPath := constants.TrayBinaryDir
+	err = goos.MkdirAll(outputPath, 0750)
```
cfergeau, line 100: Note for later: `extract.unzip()` is creating the destination directory, but `extract.untar()` does not do it. It would be a worthwhile improvement to do it in both cases, and not have to do it here.

pkg/crc/constants/constants_windows.go
```diff
@@ -12,3 +17,28 @@ const (
...
+	// argument in the service executable invocation
+	isIntSess, err := svc.IsAnInteractiveSession()
+	if err == nil {
+		if !isIntSess && len(os.Args) > 2 {
+			return os.Args[2]
+		}
+	}
+
+	if homeDrive, homePath := os.Getenv("HOMEDRIVE"), os.Getenv("HOMEPATH"); len(homeDrive) > 0 && len(homePath) > 0 {
+		homeDir := filepath.Join(homeDrive, homePath)
```
cfergeau, line 36: Is this required now that we use `os.UserHomeDir()`?

pkg/crc/constants/constants_windows.go
```diff
@@ -12,3 +17,28 @@ const (
...
+func GetHomeDir() string {
+	// if not running in interactive mode, it is the daemon service
+	// in that case we have the user's home directory as the third
+	// argument in the service executable invocation
+	isIntSess, err := svc.IsAnInteractiveSession()
+	if err == nil {
+		if !isIntSess && len(os.Args) > 2 {
+			return os.Args[2]
+		}
+	}
```
cfergeau, line 33: I don't think this should go in such a generic function. If someone reuse `GetHomeDir` for non-tray code in other parts of CRC, they probably won't be expecting to sometimes get the 2nd cmd line arg as a home dir value. Maybe set an env var with argv[2] early in `crc daemon` run, and use that here?

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -41,26 +40,9 @@ func fixDaemonServiceInstalled() error {
...
-	accountName := string(stdOut)
-	accountName = strings.TrimSpace(accountName)
-
-	// get the password from user
-	password, err := input.PromptUserForSecret("Enter account login password for service installation", "This is the login password of your current account, needed to install the daemon service")
-	if err != nil {
-		return fmt.Errorf("Unable to get login password: %v", err)
-	}
-	err = service.Create(constants.DaemonServiceName, binPathWithArgs, accountName, password)
+	err = service.Create(constants.DaemonServiceName, binPathWithArgs)
```
cfergeau, line 45: Is the daemon service going to run with elevated privileges compared to the user running `crc setup`?

anjannath: Yes, the daemon service will run with the privileges of the `LocalSystem`account which is there just for service and has extensive privileges on the system..

pkg/crc/constants/constants_windows.go
anjannath: So, argv[2] will only be returned if the service is not running as an interactive process, and provided `crc` is a cli tool, this is not likely to occur outside of the tray scenario.

I actually tried using a variable to store `argv[2]` then using it in the constants pkg, but that creates import cycle..

pkg/crc/constants/constants_windows.go
anjannath: Yes, this is to take care of situations when the `%USERPROFILE%` is not set in windows

pkg/crc/constants/constants_windows.go
gbraad: I would suggest to write logs when this happens... this can be, as said, very confusing... especially when debugging.

also, would it be better to use the username and reconstruct the path when actually needed? the user might have moved his path... or does this get properly cleaned with `crc cleanup` ?

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Actually looking at the code for `extract.unzip()` it is supposed to create the destination directory.. i am not passing the correct target then..  https://github.com/code-ready/crc/blob/1b0dda67aad99118d0688f151483b5cf51f22625/pkg/extract/extract.go#L130-L132

pkg/crc/constants/constants_windows.go
anjannath: Yes, `crc cleanup` will remove the service..

anjannath: > I would suggest to write logs when this happens... this can be, as said, very confusing... especially when debugging.

we are depending on `svc.IsAnInteractiveSession` to behave properly, in normal situations the `os.Args[2]` won't be considered, but only when running as a service in windows. even if we put logs here it doesn't help us to debug the normal working of crc.

gbraad: > we are depending on svc.IsAnInteractiveSession to behave proper

Right, but we have no way to understand what happened. So I suggest that logging can help to understand if an interactive session and therefore the parameter was used. The parameter would have to be properly described, including possible localized characters to behave as expected.

if an error occurs, I would like to understand if this was due to the early return as an interactivesession or arg value was used or the regular flow.

cfergeau: > I actually tried using a variable to store argv[2] then using it in the constants pkg, but that creates import cycle..
In `cmd/daemon.go`, call something like `constants.SetHomeDir(argv[2])` ? This should avoid the import cycle? (even if it's ugly to have a setter for constants :( )

gbraad: he wanted to avoid storing the value...

at the moment the value `argv[2]` does not convey any meaning. the problem is where to place such logic.

pkg/crc/preflight/preflight_checks_tray_windows.go
cfergeau: It's a bit worrying to run something with elevated privileges,  and to be able to configure where it's going to write on the system (through the "home dir" argument), even if I don't really have a specific situation where this could be abused.

gbraad: What could an alternative to this be?

anjannath: Yes, to change the services config "home dir" argument it needs admin privileges.

cfergeau: I really don't know what alternatives there are, I haven't looked closely at that code and how the daemon on Windows works. However your question made me realize that the commit log is not explaining _why_ we want to run the daemon under a LocalSystem account (ie c&p most of #1286, and maybe mention that currently when trying to start the service, some users get the error described in https://www.opcsupport.com/s/article/How-do-I-configure-a-user-account-to-have-logon-as-a-service-permissions )

anjannath:
@gbraad @cfergeau Dropped the use of `LocalSystem` account, now the daemon service would run under the user account.

Now we are automatically adding the log on as service permission for the user during setup.

go.mod
```diff
@@ -8,10 +8,8 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
 	github.com/cavaliercoder/grab v2.0.0+incompatible
 	github.com/code-ready/clicumber v0.0.0-20190801094041-02a92c8f7ece
-	github.com/code-ready/goodhosts v0.0.0-20190712111040-f090f3f77c26 // indirect
```
gbraad, line 11: ???

gbraad: never dare to question these tools, but often it is hard to understand what is happening ...

anjannath: So, i ran `make vendor` because i had to add this new package `golang.org/x/text` and that created this changes,

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -204,3 +201,34 @@ func stopTray() error {
...
+	enabled, err := secpol.UserAllowedToLogonAsService(username)
+	if enabled {
+		return nil
+	}
+	return fmt.Errorf("Failed to check if user is allowed to log on as service: %v", err)
+}
+
+func fixUserHasServiceLogonEnabled() error {
+	// get the username
+	username := getCurrentUsernmae()
```
gbraad, line 216: typo...

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -204,3 +201,34 @@ func stopTray() error {
...
+
+	return secpol.AllowUserToLogonAsService(username)
+}
+
+func getCurrentUsernmae() string {
```
gbraad, line 221: ^^^

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -204,3 +201,34 @@ func stopTray() error {
...
+	user, err := user.Current()
+	if err != nil {
+		logging.Debugf("Unable to get current user's name: %v", err)
+		return ""
+	}
+	return strings.TrimSpace(strings.Split(user.Username, "\\")[1])
+}
+
+func disableUserServiceLogon() error {
```
gbraad, line 230: this might not have been set by us... so, just like other fixes we might want to record what we have modified to make CRC work (and undo this on a cleanup... only the failed issues+fixes that were applied)

pkg/os/windows/secpol/secpol_windows.go
```diff
@@ -0,0 +1,238 @@
...
+		return ""
+	}
+	return path
+}
+
+// UserAllowedToLogonAsService returns true if the user's sid is present in 'seservicelogonright' config key
+func UserAllowedToLogonAsService(username string) (bool, error) {
+	/* To find out the user is allowed to log on as service by security policy
+	 * we export the User rights from the system security db at %windrive/defltbase.sdb using secedit.exe
+	 * to a temporary file and check if the given user's SID has the user right 'seservicelogonright'
```
gbraad, line 54: > to a temporary file

where is this file? and is it removed?

pkg/os/windows/secpol/secpol_windows.go
```diff
@@ -0,0 +1,238 @@
...
+	 */
+
+	securityInfoFilePath, err := exportConfigFromSecurityDbToFile()
+	if err != nil {
+		logging.Debugf("Unable to get export config: %v", err)
+		return ""
+	}
+
+	defer func() {
+		_ = goos.RemoveAll(filepath.Dir(securityInfoFilePath))
```
gbraad, line 145: Removed as instructed in `exportConfigFromSecurityDbToFile`

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Yeah, no way to know if we made those changes, yes future improvement!

pkg/os/util_windows.go
```diff
@@ -0,0 +1,27 @@
...
+
+func ReadFileUTF16(filename string) ([]byte, error) {
+	// Read the file into a []byte:
+	raw, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	// Make an tranformer that converts MS-Win default to UTF8:
+	win16le := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)
```
gbraad, line 19: Ignore byte order mark? Why ???

pkg/os/windows/secpol/secpol_windows.go
```diff
@@ -0,0 +1,238 @@
...
+	}()
+
+	// open the security config file and get the line containing the config for configName
+	content, err := os.ReadFileUTF16(securityInfoFilePath)
```
gbraad, line 149: I am not sure why we need to handle the encoding here...

pkg/os/util_windows.go
gbraad: This function is not documented so unsure why this is needed

anjannath: This is needed since the file created by secedit.exe /ecport is not utf-8 and when read directly without the conversion, the text are malformed.

pkg/os/windows/secpol/secpol_windows.go
anjannath: We need to, or else the text is malformed when read directly, https://github.com/code-ready/crc/pull/1297#discussion_r439995256

gbraad: Can you be more precies in the naming... as it is unclear now if this is `ReadFileAsUTF16LE` or `ReadFileUTF16BE`

anjannath: Yes, this should be `ReadFileUTF16LE`

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: fixed

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: fixed

pkg/os/windows/secpol/secpol_windows.go
anjannath: Done!

pkg/os/util_windows.go
cfergeau: It's documented at https://godoc.org/golang.org/x/text/encoding/unicode#UTF16
`UseBOM` would probably be slightly more correct, unless we are sure Windows UTF16 files *never* start with a BOM.

gbraad:
So many requests for Admin permission:

```
INFO Checking if user is allowed to log on as a service
INFO Will run as admin: Running secedit export command
INFO Enabling service log on for user
INFO Will run as admin: Running secedit export command
INFO Will run as admin: Running secedit configure command
INFO Checking if daemon service is installed
INFO Installing daemon service
INFO Will run as admin: Stopping service: CodeReady Containers
INFO Will run as admin: Uninstalling service: CodeReady Containers
? Enter account login password for service installation [? for help] ********
INFO Will run as admin: Install daemon service
INFO Checking if daemon service is running
INFO Starting daemon service
INFO Will run as admin: Starting service: CodeReady Containers
FATA Could not start service: CodeReady Containers
```

and since I use a MicrosoftAccount, I am not so sure what my password should be? I use automated login from a connected device to unlock. :-/

gbraad:
I performed a cleanup and when re-running setup it shows:


```
INFO Will run as admin: Stopping service: CodeReady Containers
INFO Will run as admin: Uninstalling service: CodeReady Containers
```

gbraad:
image
No description

pkg/os/util_windows.go
```diff
@@ -0,0 +1,30 @@
...
+func ReadFileUTF16LE(filename string) ([]byte, error) {
+	// Read the file into a []byte:
+	raw, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	// Make an tranformer that converts MS-Win default to UTF8:
+	win16le := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM)
+	// Make a transformer that is like win16be, but abides by BOM:
```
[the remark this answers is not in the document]
gbraad: I am just as confused due to the not clearly named function. the Byte order mark is ignored to force reading inthe non-default manner; little-endian or big-endian (IIRC, LE is the default on Wintel). The BOM will indicate with 0xFFFE to be LE... else with 0xFEFF is BE.

anjannath:
> Regarding Add check, fix and cleanup for log on as service By reading the commit log (and even in issue #1286) , it's really not clear what problem this is solving, ie if we try to use the daemon without this change, what goes wrong?

The daemon is ran as a windows service, because we want it to keep running in the background and have an interface to stop, restart the daemon. atm we cannot start the daemon without these changes because the user might not have the _log on as a service_ right, so these changes makes sure that the user has those rights.

> i read it's to start the daemon, but os.Command('crc', 'daemon').Start() would also run the daemon, so the problem which is being solved needs to be stated better

Okay, so we cannot run it as this during setup, as when the cmd/powershell is closed i think the daemon would also die, and  we don't want to depend on the tray to start the daemon process

anjannath:
@gbraad There is no option to pass description in`sc.exe create` command, we can run `sc.exe descritption` again to add a description, but that also needs elevated privilege, so one more UAC :(

cfergeau:
> > Regarding Add check, fix and cleanup for log on as service By reading the commit log (and even in issue #1286) , it's really not clear what problem this is solving, ie if we try to use the daemon without this change, what goes wrong?
> 
> The daemon is ran as a windows service, because we want it to keep running in the background and have an interface to stop, restart the daemon. atm we cannot start the daemon without these changes because the user might not have the _log on as a service_ right, so these changes makes sure that the user has those rights.

In the commit log, it might be worth adding that without this 'log on as a service' rights, `fixDaemonServiceRunning` (or `fixDaemonServiceInstalled`?) will fail? Or will these methods succeed, and then it will fail later? A more detailed description of the issue would probably be useful if we need to revisit that code in the future.

> we don't want to depend on the tray to start the daemon process

This is the part I would question/discuss. Given the complexity of the alternative (this PR which needs hundreds of lines of code to start the daemon differently), why not have the tray start it on demand?

anjannath:
> > we don't want to depend on the tray to start the daemon process
> 
> This is the part I would question/discuss. Given the complexity of the alternative (this PR which needs hundreds of lines of code to start the daemon differently), why not have the tray start it on demand?

Okay, so if we are not using a service and the tray starts the daemon process, there are also difficulties i think,  we need to know the crc binary location, but we don't require for it to be added to path, or be in a specific location, idk `crc setup` could somehow handle it...

go.mod
guillaumerose: Yes there is unused vendoring. I did the same here: https://github.com/code-ready/crc/pull/1312

guillaumerose: it's now merged, you can rebase to avoid this change.

anjannath: Sure, will do!

anjannath:
We have taken the same approach in case of macOS, we use `launchd` to run the daemon as an _agent_ which is run under the logged on user, and starts automatically when the user logs in. In case of windows keeping the approach consistent we are using the windows service manager, and hence the daemon service for windows, i've already mentioned that using a service in windows and an agent in macOS we get the o/s provided interface to start/stop/uninstall the daemon.

In case of the windows service it is running under the local user and i've listed out the differences of using the _LocalSystem_ vs a local user account in #1286 
We have to reduce the number of UACs during `crc setup` which i think can be solved in a follow up PR if that is acceptable.

If we now go ahead with an approach of the tray starting the daemon, then we have to do the same for macOS to have a consistent system, which means we need to re-engineer the tray part. @cfergeau what would you suggest we do here?

jsliacan:
ACK. Tested this latest version on Windows 10 machines (with and without cygwin) and setup now works nicely. The operation of tray also works for basic operations (start, stop, delete, status, and combination of these). LGTM!

PR #1312, opened 2020-06-15
Run `go mod tidy`. Remove unnecessary dependencies in go.mod and go.sum.

guillaumerose:
Minor cleanup. This was prompted by Jetbrains Goland when I opened the go.mod file.

I ran `go mod tidy` then `go mod vendor`. Vendor directory was already in a clean state, nothing changed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,9 @@ jobs:
         name: Make
         command: make
     - run:
+        name: Verify vendoring
+        command: make vendorcheck
+    - run:
         name: Make fmtcheck
         command: make fmtcheck
     - run:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ go:
 
 script:
 - make
+- make vendorcheck
 - make fmtcheck
 - make gosec
 - make cross
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ default: install
 # Create and update the vendor directory
 .PHONY: vendor
 vendor:
+	GO111MODULE=on go mod tidy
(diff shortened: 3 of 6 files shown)


praveenkumar:
@guillaumerose Should we add a target in the Makefile and check in the CI also or make it part of `make vendor` because in future this will again happen.

vdemeester:
> @guillaumerose Should we add a target in the Makefile and check in the CI also or make it part of `make vendor` because in future this will again happen.

@praveenkumar we do that on `tektoncd/cli` (and other tekton project) to validate that the vendor is always up-to-date :wink:

guillaumerose:
I added `go mod tidy` in the Makefile. 

For the CI, the change is a little bit bigger. `go mod verify` doesn't still check the vendor/ directory. We need to have something like https://github.com/tektoncd/cli/blob/master/hack/verify-codegen.sh - should I copy this script ?

praveenkumar:
> For the CI, the change is a little bit bigger. go mod verify doesn't still check the vendor/ directory. We need to have something like https://github.com/tektoncd/cli/blob/master/hack/verify-codegen.sh - should I copy this script ?

I agree we might need a script to fail the CI when vendor directory or the go.mod/sum doesn't match what it should be but that should be a different commit/PR.

guillaumerose:
Looking at other way to do it, can we delete the vendor/ directory ? Is there a reason to keep it ? Number of dependencies is low and public.

guillaumerose:
I added the script to verify the vendoring. PTAL.

guillaumerose:
thanks for the merge!

PR #1316, opened 2020-06-18
Bump language version of crc go module

guillaumerose:


praveenkumar:
@guillaumerose The go version we have in the CI should be matching with what RHEL offical repo have (which is still 1.13.4) so that when we do the release (happen on RHEL machine) we shouldn't have any surprise there.

praveenkumar:
@guillaumerose we should take https://github.com/code-ready/crc/pull/1316/commits/4de6ee99d4c2f558a10f5bf80946cce2a44e40bd one in but not the go version one.

guillaumerose:
Thanks @praveenkumar ! I remove the commit about the upgrade.

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/code-ready/crc
 
+go 1.13
```
praveenkumar, line 3: :+1:

PR #1317, opened 2020-06-18
Enable gocritic linter

guillaumerose:
This is a proposition, not sure we should merge it.

It catches errors like capitalized variable name and switch with default case only.
On the other hand, it forces us to rewrite some if-else statements with switchs.

### DIFF
--- a/pkg/extract/extract.go
+++ b/pkg/extract/extract.go
@@ -36,21 +36,22 @@ func uncompress(tarball, targetDir string, fileFilter func(string) bool) ([]stri
 	}
 	defer file.Close()
 
-	if strings.HasSuffix(tarball, ".tar.xz") || strings.HasSuffix(tarball, ".crcbundle") {
+	switch {
+	case strings.HasSuffix(tarball, ".tar.xz"), strings.HasSuffix(tarball, ".crcbundle"):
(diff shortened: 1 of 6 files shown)


pkg/extract/extract.go
```diff
@@ -36,21 +36,22 @@ func uncompress(tarball, targetDir string, fileFilter func(string) bool) ([]stri
 	}
 	defer file.Close()
 
-	if strings.HasSuffix(tarball, ".tar.xz") || strings.HasSuffix(tarball, ".crcbundle") {
+	switch {
+	case strings.HasSuffix(tarball, ".tar.xz") || strings.HasSuffix(tarball, ".crcbundle"):
```
cfergeau, line 40: I think I'd write this as
```
case strings.HasSuffix(tarball, ".tar.xz"):
fallthrough
case strings.HasSuffix(tarball, ".crcbundle"):
```
or maybe
```
case strings.HasSuffix(tarball, ".tar.xz"), strings.HasSuffix(tarball, ".crcbundle"):
```

guillaumerose: Nice! I didn't know about the syntax with a comma. Thanks! I changed it.

PR #1318, opened 2020-06-18
Remove fmtcheck from release target

praveenkumar:
b374b3f remove the fmtcheck target from makefile but not
removed from release target.


guillaumerose:
sorry I missed this :/

PR #1319, opened 2020-06-18
Add source code to openshift-ci image

praveenkumar:
Since our integration test depend on the source code we need to
put that in the image creation so during the test step we can
able to scp to the VM and run the integration test.

### DIFF
--- a/images/openshift-ci/Dockerfile
+++ b/images/openshift-ci/Dockerfile
@@ -7,6 +7,7 @@ COPY . .
 RUN make cross
 
 FROM centos:7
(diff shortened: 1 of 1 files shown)


praveenkumar:
> What do they need exactly? 

@cfergeau In Openshift-CI the container which run test need to have source code so we can scp to the created VM. 

> Dunno what other projects are doing? 

Openshift related projects like installer depend on the openshift-tests binary which perform the e2e tests and we don't have any testing binary.

> Another option could be to copy built binaries and the test descriptions to the image? 

yes but if we copy the test descriptions then what is issue of copying the source, at least then we can just run make integration as we run on our dev box.

PR #1321, opened 2020-06-19
Add unzip pacakge to openshift-ci image

praveenkumar:
This is required to unzip the downloaded bundle.

### DIFF
--- a/images/openshift-ci/Dockerfile
+++ b/images/openshift-ci/Dockerfile
@@ -20,6 +20,7 @@ RUN yum update -y && \
     libvirt-client \
     libvirt-libs \
     nss_wrapper \
+    unzip \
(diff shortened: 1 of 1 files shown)


praveenkumar:
> The bundles are compressed with xz, so not sure why it needs unzip?

We take that xz bundle create a zip with password and upload it to GCP from where it is used for centos CI till now.

cfergeau:
> > The bundles are compressed with xz, so not sure why it needs unzip?
> 
> We take that xz bundle create a zip with password and upload it to GCP from where it is used for centos CI till now.

Can you add this background information to the commit log? ("OpenShift CI needs to access the bundle from GCP which is wrapped in a password-protected zip" or such)

PR #1324, opened 2020-06-19
RFC Move all drivers inside crc binary

guillaumerose:
Not finished yet.

Pros:
* No more drivers. 2 binaries removed.
* No more RPC between crc and drivers.
* Allow us to delete more machine code and slowly diverge.

Cons:
* crc is not statically compiled on linux anymore because of libvirt-go


guillaumerose:
Hyperkit code still to be run as root so we cannot do all of that sadly.

PR #1336, opened 2020-06-24
Add OWNERS file

praveenkumar:
This file is used by openshift ci bot to assign a reviewer and also
provide permission to specific member to approve it.


cfergeau:
Honestly, I would have expected an announcement/a discussion about all the process changes forced on us when we start using OpenShift CI **before** these changes are done.

gbraad:
> I would have expected an announcement/a discussion about all the process changes 

I do not think any of us was aware of this 'need'. However the need to gate with our build on OpenShift CI outways the choice of using the CI.

We can discuss about how to adhere to the process; ignore, follow, configure, etc.

PR #1343, opened 2020-06-29
Add progress bar for bundle extraction

guillaumerose:
This PR adds multiple things:
* testdata for the xz format
* a fix for a `defer file.Close()` in a loop
* a progress bar for bundle extraction (in a terminal only)

Output:
```
INFO Extracting bundle: crc_hyperkit_4.4.5.crcbundle ...
crc_hyperkit_4.4.5/kubeadmin-password: 23 B / 23 B [-----------------------------------------------------------------------------------------] 100.00% ? p/s 0s
crc_hyperkit_4.4.5/kubeconfig: 11.55 KiB / 11.55 KiB [---------------------------------------------------------------------------------------] 100.00% ? p/s 0s
crc_hyperkit_4.4.5/id_rsa_crc: 2.53 KiB / 2.53 KiB [-----------------------------------------------------------------------------------------] 100.00% ? p/s 0s
crc_hyperkit_4.4.5/crc.qcow2: 490.22 MiB / 9.15 GiB [--->_______________________________________________________________________] 5.23% 34.99 MiB p/s ETA 4m13s
```

Tell me what you think!


gbraad:
> Output:

Too detailed... can this be brought back to an overall extraction (displaying the extraction of the archive instead of contents)?

Also, how does this work on a non PTY enabled terminal? like from the adapter in Eclipse?

guillaumerose:
Yes there is too much details. The issue is that we don't have the total length of the uncompressed data. I can change the template and remove information.

What if I only display the bar for object larger than 1 GB ?

gbraad:
> What if I only display the bar for object larger than 1 GB ?

That would make more sense, although the others file are relevant as part of debug perhaps?

cfergeau:
> Also, how does this work on a non PTY enabled terminal? like from the adapter in Eclipse?

Aren't they planning to use the daemon API as soon as it's available/haven't they been asking for json output for our commands?

guillaumerose:
Other try, less chatty:

``` 
INFO Extracting bundle: crc_libvirt_4.4.8.crcbundle ... 
crc.qcow2: 445.91 MiB / 9.40 GiB [---->______________________________________________________________________________________] 4.63%
``` 

Without a pty:
```
level=info msg="Checking if NetworkManager service is running"
level=info msg="Checking if /etc/NetworkManager/conf.d/crc-nm-dnsmasq.conf exists"
level=info msg="Checking if /etc/NetworkManager/dnsmasq.d/crc.conf exists"
level=info msg="Extracting bundle: crc_libvirt_4.4.8.crcbundle ..."
```

pkg/extract/extract.go
```diff
@@ -117,9 +119,18 @@ func untarFile(tarReader *tar.Reader, header *tar.Header, path string) error {
 	}
 	defer file.Close()
 
+	var reader io.Reader
+	reader = tarReader
+	if isatty.IsTerminal(os.Stdout.Fd()) && header.FileInfo().Size() >= 1_000_000_000 {
```
cfergeau, line 124: I'd start even at 100MB or so.

pkg/extract/extract.go
```diff
@@ -117,9 +119,18 @@ func untarFile(tarReader *tar.Reader, header *tar.Header, path string) error {
 	}
 	defer file.Close()
 
+	var reader io.Reader
+	reader = tarReader
```
cfergeau, line 123: This is only implemented for tar files? (rationale in commit log)

pkg/extract/extract.go
```diff
@@ -117,9 +119,18 @@ func untarFile(tarReader *tar.Reader, header *tar.Header, path string) error {
...
+	if isatty.IsTerminal(os.Stdout.Fd()) && header.FileInfo().Size() >= 1_000_000_000 {
+		bar := pb.Simple.Start64(header.FileInfo().Size())
+		bar.Set("prefix", fmt.Sprintf("%s: ", filepath.Base(header.Name)))
```
cfergeau, line 126: my main grief is that the bar is far too wide on fullscreen term, but that's a minor complaint, much better with one :)

cfergeau:
> `Extract function that untar a single file.`defer file.Close()` is no……w called after each file.`
> 
> is too long for a short log, you can split this in a real short log, and a longer description.
> 
> ```
> untar: call defer file.Close() after each file
> 
> deferred calls are only called when the current function ends, which means untar() is currently holding all files open until it has finished processing the whole archive. By introducing a `untarSingleFile()` helper, we can ensure each file is closed as soon as it's fully uncompressed.
> ```

I also had commented on this shortlog but it seems it was lost (?):
` Add progress bar for bundle extraction (only in a terminal) `
The `only in a terminal` could be split in the actual log, I don't think it's required to keep it in the short log.

pkg/extract/extract.go
```diff
@@ -10,19 +10,21 @@ import (
...
 }
 
-func Uncompress(tarball, targetDir string) ([]string, error) {
-	return uncompress(tarball, targetDir, nil)
+func Uncompress(tarball, targetDir string, showProgress bool) ([]string, error) {
+	return uncompress(tarball, targetDir, nil, showProgress)
 }
 
-func uncompress(tarball, targetDir string, fileFilter func(string) bool) ([]string, error) {
+func uncompress(tarball, targetDir string, fileFilter func(string) bool, showProgress bool) ([]string, error) {
```
praveenkumar, line 27: where is the `showProgress` bool is used in this function, I think you need to pass it to the untarFile function and hide the progessbar as per this bool.

anjannath:
The progress bar looks nice 👍 

I thought we also wanted to show a progress bar after? but this would be a bit difficult i think because, we don't know when its going to end, maybe a indefinite spinner kind of thing.
```
```

pkg/extract/extract.go
guillaumerose: fixed, thanks!

pkg/extract/extract.go
guillaumerose: I set the max width parameter to 80. It will not fill the terminal anymore.

pkg/extract/extract.go
guillaumerose: I added support for zip file also.

pkg/extract/extract.go
guillaumerose: done. In crcbundle, only the qcow2 file is larger than 100MB, so still only one progress bar.

guillaumerose:
I tried to add a max width to the progress bar but it messes up with carriage return when 2 progress bars are shown. I tried to add newline, etc. but it's not pretty. 
I'm back with a fullscreen progress bar.

pkg/extract/extract.go
```diff
@@ -181,7 +194,16 @@ func unzipFile(path string, file *zip.File) error {
...
 	defer targetFile.Close()
 
+	var reader io.Reader
+	reader = fileReader
+	if showProgress && file.FileInfo().Size() >= minSizeForProgressBar {
+		bar := pb.Simple.Start64(file.FileInfo().Size())
+		bar.Set("prefix", fmt.Sprintf("%s: ", filepath.Base(file.Name)))
+		defer bar.Finish()
+		reader = bar.NewProxyReader(fileReader)
+	}
```
cfergeau, line 204: I believe this could go in a small helper shared with the tar code, especially as the `defer bar.Finish()` is not required, reader.Close() will also do it https://github.com/cheggaaa/pb/blob/ca7ab10606663e9efdbc3c1ad157459b5a2b0edf/reader.go#L21

guillaumerose: Good catch! I refactored a bit more and now I have https://github.com/code-ready/crc/pull/1343/files#diff-106940100fecd5f362aaeae069041830R198
The two functions are very similar but calling one from the other makes me introduce more complexity than required. WDYT ?

pkg/extract/extract_test.go
```diff
@@ -99,9 +99,9 @@ func testUncompress(archiveName string, fileFilter func(string) bool, files File
 
 	var fileList []string
 	if fileFilter != nil {
-		fileList, err = UncompressWithFilter(archiveName, destDir, fileFilter)
+		fileList, err = UncompressWithFilter(archiveName, destDir, false, fileFilter)
```
praveenkumar, line 102: :+1

pkg/extract/extract.go
```diff
@@ -175,13 +183,30 @@ func unzipFile(path string, file *zip.File) error {
...
 	defer targetFile.Close()
 
 	// #nosec G110
-	_, err = io.Copy(targetFile, fileReader)
+	_, err = io.Copy(targetFile, reader)
 	return err
 }
+
+func progressBarReader(reader io.Reader, info os.FileInfo, showProgress bool) (io.Reader, func()) {
+	if showProgress && info.Size() >= minSizeForProgressBar {
```
praveenkumar, line 201: :+1:

guillaumerose:
Any idea why prow doesn't merge the PR ? https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#the-code-review-process 
It should do it automatically.

gbraad:
it did... right?

PR #1350, opened 2020-07-01
preflight: macos: reset /etc/hosts permissions to 0644

cfergeau:
Up to crc 1.10, we were mistakenly setting /etc/hosts permissions to
0600. Since it was owned by the user running crc, this was not causing
noticeable issue. Now we no longer rely on the /etc/hosts file being
owned by the user, but it seems it's possible for some tool to reset
/etc/hosts ownership back to root:wheel without changing its
permissions.
When this happens, crc fails to start because of DNS problems. This
probably has an impact on other tools as well.
This commit readds a preflight check to reset permissions/ownership of
/etc/hosts to avoid this problematic 0600 / root:wheel situation.

## Solution/Idea

To prevent having a /etc/hosts file with 0600 permissions forever, this commit (re) adds a preflight check to reset the permissions and ownership back to some acceptable default values (0644 and root:wheel)

## Testing

1. run setup from an older (1.10 I believe, or before that) crc release on macos
2. check /etc/hosts permissions are 0600
3. run setup with these changes
4. check /etc/hosts permissions/ownership

### DIFF
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -20,6 +20,7 @@ import (
 const (
 	resolverDir  = "/etc/resolver"
 	resolverFile = "/etc/resolver/testing"
+	hostsFile    = "/etc/hosts"
 )
 
 func basenameFromUrl(url string) (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -130,6 +131,35 @@ func fixMachineDriverHyperKitInstalled() error {
...
+
+func fixEtcHostsFilePermissions() error {
+	stdOut, stdErr, err := crcos.RunWithPrivilege(fmt.Sprintf("change ownership of %s", hostsFile), "chown", "root:wheel", hostsFile)
+	if err != nil {
+		return fmt.Errorf("Unable to change ownership of %s: %s %v: %s", hostsFile, stdOut, err, stdErr)
+	}
+
+	stdOut, stdErr, err = crcos.RunWithPrivilege(fmt.Sprintf("change permissions of %s", hostsFile), "chmod", "644", hostsFile)
+	if err != nil {
+		return fmt.Errorf("Unable to change permissions of %s to 0644: %s %v: %s", hostsFile, stdOut, err, stdErr)
```
praveenkumar, line 157: @cfergeau does the default permission on the `/etc/hosts` is `0644` for mac (like for a user who never used the CRC) ?

praveenkumar: I just checked on fresh machine and looks like the permission is 644 so this looks good.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -20,6 +20,7 @@ import (
 const (
 	resolverDir  = "/etc/resolver"
 	resolverFile = "/etc/resolver/testing"
+	hostsFile    = "/etc/hosts"
```
praveenkumar, line 23: :+1:

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -130,6 +131,35 @@ func fixMachineDriverHyperKitInstalled() error {
...
+	logging.Debugf("Checking if /etc/hosts ownership/permissions need to be adjusted after crc upgrade")
+	fileinfo, err := os.Stat(hostsFile)
+	if err != nil {
+		return err
+	}
+	// Older crc releases were setting /etc/hosts permissions to 0600 and ownership to the current user
+	// This will cause issues if ownership is reset to root:wheel with permissions
+	// issue if other tools
+	if fileinfo.Mode().Perm() == 0600 {
+		return fmt.Errorf("%s permissions are not 0644", hostsFile)
```
praveenkumar, line 144: :+1:

PR #1351, opened 2020-07-01
Add deployment config and deployment both for rollout status

praveenkumar:
with 16d015a, we removed deploymentconfig because with latest version
of oc didn't have it but looks like it is again come up so better to
check for both resources.

### DIFF
--- a/test/integration/features/story_health.feature
+++ b/test/integration/features/story_health.feature
@@ -49,7 +49,7 @@ Feature:
             """
             service "httpd-ex" created
             """
-        When executing "oc rollout status deployment httpd-ex" succeeds
+        When executing "oc rollout status dc/httpd-ex || oc rollout status deployment httpd-ex" succeeds
         Then stdout should contain "successfully rolled out"
         And executing "oc expose svc/httpd-ex" succeeds
         And with up to "2" retries with wait period of "60s" http response from "http://httpd-ex-testproj.apps-crc.testing" has status code "200"


test/integration/features/story_health.feature
```diff
@@ -49,7 +49,7 @@ Feature:
             """
             service "httpd-ex" created
             """
-        When executing "oc rollout status deployment httpd-ex" succeeds
+        When executing "oc rollout status dc/httpd-ex || oc rollout status deployment httpd-ex" succeeds
```
anjannath, line 52: I've only seen it create _deployment_ but just to be safe..

jsliacan:
We could check for both, but only Deployments should be used soon I think, here: https://redhat:redhat@docs.openshift.com/container-platform/4.5/release_notes/ocp-4-5-release-notes.html#ocp-4-5-oc-new-app-deployment-resources.  I'm aiming to remove all uses of DeploymentConfig throughout the tests. But there are other things that became obsolete, like "CatalogSourceConfig"s etc.

praveenkumar:
> We could check for both, but only Deployments should be used soon I think, here: https://redhat:redhat@docs.openshift.com/container-platform/4.5/release_notes/ocp-4-5-release-notes.html#ocp-4-5-oc-new-app-deployment-resources. I'm aiming to remove all uses of DeploymentConfig throughout the tests. But there are other things that became obsolete, like "CatalogSourceConfig"s etc.

I agree as per doc it should be the case but right now even with latest nightly I saw the deployment config instead of deployment so better to have this until it settle (mostly when 4.6 nightly bits start).

cfergeau:
> I agree as per doc it should be the case but right now even with latest nightly I saw the deployment config instead of deployment so better to have this until it settle (mostly when 4.6 nightly bits start).

Including latest nightly of `oc`? The link Jakub gave seems to say it's an `oc` change rather than a cluster change.

cfergeau:
This was tracked in https://bugzilla.redhat.com/show_bug.cgi?id=1836833 and seems to only be done in oc 4.5.x

Issue #1338, opened 2020-06-25
[BUG] crc setup broken on Windows with user name containg space when using experimental-features


jeffmaury:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop / Baremetal-Server / VM: Laptop

## CRC version
```bash
# Put `crc version` output here
CodeReady Containers version: 1.12.0+6710aff
OpenShift version: 4.4.8 (embedded in binary)
```
  
## CRC status
```bash
# Put `crc status --log-level debug` output here
DEBU CodeReady Containers version: 1.12.0+6710aff
DEBU OpenShift version: 4.4.8 (embedded in binary)
Machine 'crc' does not exist. Use 'crc start' to create it.
```

## CRC config
```bash
# Put `crc config view` output here
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
# put the output of `sw_vers` in case of Mac
# Put the output of `systeminfo` in case of Windows
Nom de l’hôte:                              DESKTOP-U2EVGBM                                                                                           
Nom du système d’exploitation:              Microsoft Windows 10 Professionnel                                                                        
Version du système:                         10.0.18363 N/A version 18363                                                                              
Fabricant du système d’exploitation:        Microsoft Corporation                                                                                     
Configuration du système d’exploitation:    Station de travail autonome                                                                               
Type de version du système d’exploitation:  Multiprocessor Free                                                                                       
Propriétaire enregistré:                    Jeff MAURY                                                                                                
Organisation enregistrée:                                                                                                                             
Identificateur de produit:                  00330-80819-37372-AA630                                                                                   
Date d’installation originale:              18/11/2019, 15:37:22                                                                                      
Heure de démarrage du système:              11/06/2020, 09:01:18                                                                                      
Fabricant du système:                       LENOVO                                                                                                    
Modèle du système:                          20LAS3NJ0H                                                                                                
Type du système:                            x64-based PC                                                                                              
Processeur(s):                              1 processeur(s) installé(s).                                                                              
                                            [01] : Intel64 Family 6 Model 142 Stepping 10 GenuineIntel ~1910 MHz                                      
Version du BIOS:                            LENOVO N27ET36W (1.22 ), 04/07/2019                                                                       
Répertoire Windows:                         C:\WINDOWS                                                                                                
[...]

PR #1353, opened 2020-07-02
Issue #1338 User user.Uid from os/user to get sid in windows

anjannath:
We were using Get-LocalUser powershell cmdlet before but it resulted in
a bug as the _username_ was not enclosed in quotes "", usernames having
spaces in them would cause the function to fail

**Fixes:** #1338

### DIFF
--- a/pkg/os/windows/secpol/secpol_windows.go
+++ b/pkg/os/windows/secpol/secpol_windows.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	goos "os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


anjannath:
Was a typo in the commit msg, s/User/Use

pkg/os/windows/secpol/secpol_windows.go
```diff
@@ -25,13 +26,12 @@ Revision=1
...
-	stdOut, stdErr, err := powershell.Execute(fmt.Sprintf(cmd, username))
-	if stdErr != "" || err != nil {
-		return "", fmt.Errorf("%s: %v", stdErr, err)
+func getSidOfCurrentUser() (string, error) {
+	u, err := user.Current()
+	if err != nil {
+		return "", err
 	}
-	return strings.TrimSpace(stdOut), nil
+	return u.Uid, nil
```
praveenkumar, line 34: :+1:

PR #1355, opened 2020-07-02
Refactor logging code

anjannath:
Remove the file hook and set the logfile as the output destination
for logrus, this enables the daemon service in windows to put logs
in the log file, both from the machine side and the logs from logrus

In case of the daemon service in windows there is no stdout and stderr devices
and the logs from logrus doesn't appear in the logfile this commit sets the
logfile as the primary output for logrus, which seems to solve the issue.

### Testing
1. Remove the existing daemon (`crc cleanup`)
2. install daemon from this PR (`crc setup --enable-experimental-features`)
3. check ~/.crc/crcd.log for logs while interacting from the tray, it should logs that are from logrus


anjannath:
@cfergeau Updated the commit log!

gbraad:
> Remove the file hook and set logfile as the output
> destination for logrus

Can you explain why this is needed?

Also, how can this be tested?

anjannath:
@gbraad The motivation for the refactor was to solve why the daemon service was not sending logs to the logfile, whereas logs from the machine package was written. My guess was that since we discard all the logs and then use hooks afterwards, in case of the daemon service there was no stdout and stderr devices, and those gets discarded before it fires the filehook, this refactor sets the logfile as the primary output for logrus in the beginning, which seems to solve the issue. 

This doesn't change the existing behaviour of crc, there for no testing instructions, but to test one could delete the existing daemon service and install the daemon again from this PR and logs that are using logrus should appear on the log file (~/.crc/crcd.log)

gbraad:
Can you describe this in the git commit descriptions and the issue template (update message 1)?
Even when NO specific testing exists, in that case it should be confirmed no behavarioul changes happen.

cmd/crc/cmd/daemon.go
```diff
@@ -14,6 +16,10 @@ var daemonCmd = &cobra.Command{
 	Long:   "Run the crc daemon",
 	Hidden: true,
 	Run: func(cmd *cobra.Command, args []string) {
+		// setup separate logging for daemon
+		logging.CloseLogging()
+		logging.InitLogrus(logging.LogLevel, constants.DaemonLogFilePath)
+
```
cfergeau, line 22: For what it's worth, this code moving logging set up from `daemon_nonwin.go`/`daemon_windows.go` could have been done as a first commit to minimize the changes in the commit reworking the logging system.

PR #1359, opened 2020-07-03
Update bundle version from 4.4.8 to 4.5.0-rc.5

praveenkumar:


praveenkumar:
> I assume nothing is confused by the -rc.5?

@cfergeau you mean not using the `rc` bits here or anything else?

praveenkumar:
/hold 

Will also try to fix the operator test case with same PR.

cfergeau:
> > I assume nothing is confused by the -rc.5?
> 
> @cfergeau you mean not using the `rc` bits here or anything else?

Not suggesting changing anything, just mentioning that I can't think of crc code which would be impacted/confused by the `-rc.5`

PR #1363, opened 2020-07-07
Prepare to update ~/.kube/config with crc contexts

guillaumerose:
This is preliminary work for updating kubeconfig

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -602,7 +602,7 @@ func Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
 
 		// check if all the clusteroperators are running
 		ocConfig := oc.UseOCWithConfig(statusConfig.Name)
-		operatorsStatus, err := cluster.GetClusterOperatorStatus(ocConfig)
+		operatorsStatus, err := cluster.GetClusterOperatorsStatus(ocConfig)
 		if err != nil {
 			openshiftStatus = "Not Reachable"
 			logging.Debug(err.Error())
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 
@@ -35,24 +36,38 @@ type ClusterStatus struct {
 	Disabled    bool
 }
 
-func GetClusterOperatorStatus(ocConfig oc.OcConfig) (*ClusterStatus, error) {
-	cs := &ClusterStatus{}
+func GetClusterOperatorStatus(ocConfig oc.OcConfig, operator string) (*ClusterStatus, error) {
+	return getStatus(ocConfig, []string{operator})
+}
+
+func GetClusterOperatorsStatus(ocConfig oc.OcConfig) (*ClusterStatus, error) {
+	return getStatus(ocConfig, []string{})
+}
+
+func getStatus(ocConfig oc.OcConfig, selector []string) (*ClusterStatus, error) {
+	cs := &ClusterStatus{
+		Available: true,
+	}
+
 	data, stderr, err := ocConfig.RunOcCommand("get", "co", "-ojson")
 	if err != nil {
 		return cs, fmt.Errorf("%s", stderr)
 	}
 
 	var co K8sResource
-
-	err = json.Unmarshal([]byte(data), &co)
-	if err != nil {
+	if err := json.Unmarshal([]byte(data), &co); err != nil {
 		return cs, err
 	}
-	cs.Available = true
+
+	found := false
 	for _, c := range co.Items {
 		if contains(c.Metadata.Name, ignoreClusterOperators) {
 			continue
 		}
+		if len(selector) > 0 && !contains(c.Metadata.Name, selector) {
+			continue
+		}
+		found = true
 		for _, con := range c.Status.Conditions {
 			switch con.Type {
 			case "Available":
@@ -81,6 +96,9 @@ func GetClusterOperatorStatus(ocConfig oc.OcConfig) (*ClusterStatus, error) {
 				logging.Debugf("Unexpected operator status for %s: %s", c.Metadata.Name, con.Type)
 			}
 		}
+	}
+	if !found {
+		return nil, errors.New("no cluster operator found")
 	}
 	return cs, nil
 }
--- a/pkg/crc/cluster/clusteroperator_test.go
+++ b/pkg/crc/cluster/clusteroperator_test.go
@@ -5,30 +5,57 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/oc"
+
+	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetClusterOperatorStatus(t *testing.T) {
-	status, err := GetClusterOperatorStatus(oc.OcConfig{
-		Runner: &mockRunner{file: filepath.Join("testdata", "co.json")},
-	})
-	assert.NoError(t, err)
-	assert.Equal(t, &ClusterStatus{
+var (
+	available = &ClusterStatus{
 		Available: true,
-	}, status)
-}
-
-func TestGetClusterOperatorStatusProgressing(t *testing.T) {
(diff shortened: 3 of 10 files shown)


pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,109 @@
...
+
+func currentContext(rootConfig string) string {
+	if _, err := os.Stat(rootConfig); os.IsNotExist(err) {
+		return ""
+	}
+	// #nosec G204
+	currentCtxCmd := exec.Command(ocBinary(), "config", "current-context")
+	currentCtxCmd.Stderr = os.Stderr
+	currentCtxCmd.Env = []string{fmt.Sprintf("KUBECONFIG=%s", rootConfig)}
+	context, err := currentCtxCmd.Output()
```
praveenkumar, line 76: For this you can use https://github.com/code-ready/crc/blob/master/pkg/os/exec.go#L54 one also use `--kubeconfig` parameter for oc command instead of adding to env.

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,109 @@
...
+	loginCmd.Stderr = os.Stderr
+	err := loginCmd.Run()
+	if err != nil {
+		return "", err
+	}
+	// #nosec G204
+	currentContextCmd := exec.Command(ocBinary(), "config", "current-context")
+	currentContextCmd.Stderr = os.Stderr
+	currentContextCmd.Env = []string{fmt.Sprintf("KUBECONFIG=%s", kubeconfig)}
+	currentContext, err := currentContextCmd.Output()
```
praveenkumar, line 100: same as above using crc os package.

praveenkumar:
@guillaumerose The biggest challenge for this to work every single time is how our auth operator able to serve request. Since we just wait for a 3 mins and assume everything is up and running for the cluster but in real that is not true for some of the scenario (like proxy case) or even a bit slow system, that's the reason we have `crc status` command to check the operator status and the login is taken care by auth so if that is not available in 3 mins then this is going to error out and our end user will think something broken with CRC, which might not be the case. 

We thought of https://github.com/code-ready/crc/pull/242 but some part of it still not implemented.

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,109 @@
...
+	dev, err := writeContext(dir, config, "developer", "developer", "crc-dev")
+	if err != nil {
+		return err
+	}
+	admin, err := writeContext(dir, config, "kubeadmin", config.KubeAdminPass, "crc-admin")
+	if err != nil {
+		return err
+	}
+
+	rootConfig := filepath.Join(constants.GetHomeDir(), ".kube", "config")
```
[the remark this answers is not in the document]
guillaumerose: To simplify, we can do:
* If KUBECONFIG is set, print a warning and do nothing. `You need to login, etc.`
* otherwise use the default file.

I believe that if KUBECONFIG is set, it will be really hard to not mess up with the configuration.

pkg/crc/machine/machine.go
```diff
@@ -439,6 +445,7 @@ func Start(startConfig StartConfig) (StartResult, error) {
 
 	// If no error, return usage message
 	if result.Error == "" {
+
```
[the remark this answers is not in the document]
guillaumerose: Good catch!

pkg/crc/machine/kubeconfig.go
guillaumerose: It's weird but config merge doesn't work with `--kubeconfig`. I had to keep the env variable.

``` 
$ kubectl --kubeconfig c1:c2 config view
apiVersion: v1
clusters: []
contexts: []
current-context: ""
kind: Config
preferences: {}
users: []

$ KUBECONFIG=c1:c2 kubectl config view  
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: https://api.crc.testing:6443
...

```

pkg/crc/machine/kubeconfig.go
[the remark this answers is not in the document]
praveenkumar: @guillaumerose in minishift we used to do that without any issue. But there we used to parse the kubeconfig file instead using the oc commands (that doesn't need to be done here) . 

- Check if KUBECONFIG set the use this ase rootConfig (if file doesn't exist then create, I think same we are going to do with `~/.kube/config` file )
- otherwise default file `~/.kube/config` 

Because I am sure if we make this change the next request would be honoring the KUBECONFIG parameter.

pkg/crc/machine/kubeconfig.go
praveenkumar: @guillaumerose Ah yes, https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files explains it.

guillaumerose:
I added code to wait the authentication operator. Ready for a new round of review!

/retest
/unhold

pkg/crc/machine/machine.go
```diff
@@ -426,6 +425,24 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+	if err := errors.RetryAfter(60, func() error {
+		status, err := cluster.GetClusterOperatorStatus(ocConfig, "authentication")
+		if err != nil {
+			return &errors.RetriableError{Err: err}
+		}
+		if isReady(status) {
+			return nil
+		}
+		return &errors.RetriableError{Err: errors.New("cluster operator authentication not ready")}
+	}, 2*time.Second); err != nil {
```
praveenkumar, line 439: @gbraad This extra 2 mins in the best scenario doesn't going to increase the startup time but in worst scenario it will increase the startup time by 2 mins.

gbraad: OK

The time of 10 minutes is the optimal path, but thanks for informing.

pkg/crc/machine/kubeconfig.go
guillaumerose: I refactored this, it should be better.

pkg/crc/machine/kubeconfig.go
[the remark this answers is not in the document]
guillaumerose: I change the logic. Let me know if it's better.

praveenkumar:
Overall this PR looks good but we still have issue with merging the kubeconfig file looks like the token are not updated in case of merge, this is limitation of kubeconfig merge side. :( 

```yaml
$ oc config view --kubeconfig /tmp/crc588552271/crc-admin 
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: https://api.crc.testing:6443
  name: api-crc-testing:6443
contexts:
- context:
    cluster: api-crc-testing:6443
    namespace: default
    user: kube:admin
  name: crc-admin
current-context: crc-admin
kind: Config
preferences: {}
users:
- name: kube:admin
  user:
    token: LhtL4sIvSID3oJuimLzWIXuedx0DOtbRS1-B2sM0_II
$ oc config view --kubeconfig /tmp/crc588552271/crc-dev 
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: https://api.crc.testing:6443
  name: api-crc-testing:6443
contexts:
- context:
    cluster: api-crc-testing:6443
    user: developer
  name: crc-dev
current-context: crc-dev
kind: Config
preferences: {}
users:
- name: developer
  user:
    token: b_2RzZUeiQ1yFKsBSziC6K6UT9Hz_tl9oyS-L3jPb3I
$ export KUBECONFIG=/tmp/crc588552271/crc-dev:/tmp/crc588552271/crc-admin:~/.kube/config
$ oc config view
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: https://api.crc.testing:6443
  name: api-crc-testing:6443
contexts:
- context:
    cluster: api-crc-testing:6443
    namespace: default
    user: kube:admin
  name: crc-admin
- context:
    cluster: api-crc-testing:6443
    user: developer
  name: crc-dev
current-context: crc-dev
kind: Config
preferences: {}
users:
- name: developer
  user:
    token: nd8VTMdYe60t5vrB_tVClg8xEBsIWfhT2OCWsF3hAWQ
- name: kube:admin
  user:
    token: VfIpu5_TZTvCkVTEOYG4TwtQNGB0pFQiC3WmnE2_SGs
```

The another problem is if someone set the `KUBECONFIG` env with merge type `foo:bar:~/.kube/config`  then we are honoring this variable but it will fails with no file found when writing back to kubeconfig file.

```
ERRO open foo:bar:~/.kube/config: no such file or directory 
```

I am happy to take other commits (mostly refactor) as it is and the last one can be a separate PR.

PR #1372, opened 2020-07-08
Remove go-isatty use

cfergeau:
The logging package is already using terminal.IsTerminal for the same purpose.

## Solution/Idea

This removes  use of a package which is redundant with another one that we already use.

## Testing

When unpacking a bundle, a progress bar should be shown on a tty, and not shown when not on a tty (same behaviour as before)


praveenkumar:
@cfergeau looks like vendor check error in the CI please try to update the vendor and again push?

 ``
/Users/distiller/project is out of date. Please run make vendor
make: *** [vendorcheck] Error 1
```

cfergeau:
Running `make vendor` locally does not make any changes, so I don't know how to fix CI.

guillaumerose:
Running `make vendor` with go1.13, unwanted file are removed. Strange!
I amended your last commit.

Issue #1327, opened 2020-06-22
Reduce the number of UAC prompts during setup and cleanup in windows


anjannath:
After #1297 now there are many UAC prompts that the user now has to see, which is not a very good user experience. We need to reduce the number of these prompt.

PR #1373, opened 2020-07-09
Issue #1327 Reduce the UAC Prompts during setup and cleanup on windows

anjannath:
**Fixes:** Issue #1327 

## Solution/Idea
Instead of prompting user for elevated access all the time during setup and cleanup for installing the tray, perform all the required steps/commands from a powershell script and prompt user for elevated access only once.

## Testing

1. `crc setup --enable-experimental-features` successfully installs the system tray and ask for UAC only once.
2. `crc cleanup` successfully removes the system tray and ask for UAC prompt once (it asks 2 times, but the  later one is for removing dns entry from default switch)

Note: the ps scripts with comments are here: https://gist.github.com/anjannath/6ede4ac5aa6d50d458581a5f14c7e9b2


gbraad:
Will test this tomorrow morning

gbraad:
@anjannath can you provide more pointers for QE to understand what they can test AFTER the `crc setup` has run? As before, the steps were clearly defined as part of preflight, but now they are combined it would be hard for @jsliacan to understand what he can specificially look into to determine if the setup succeeded, or not.

Also, this might have to be tested on an international system and if possible a domain-joined machine.

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -4,95 +4,118 @@ import (
...
 	password, err := input.PromptUserForSecret("Enter account login password for service installation", "This is the login password of your current account, needed to install the daemon service")
 	if err != nil {
 		return fmt.Errorf("Unable to get login password: %v", err)
 	}
-	err = service.Create(constants.DaemonServiceName, binPathWithArgs, accountName, password)
+
+	// escape '$' with '`' if present in password
+	// so that powershell doesn't try to exapnd it as variable
+	if strings.Contains(password, "$") {
+		password = strings.Replace(password, "$", "`$", -1)
```
cfergeau, line 64: Is this the only character which needs escaping? What if the password contains a `? Or if the password is `"; somepowershellcommand`?

cfergeau:
I have fuzzy memories that we needed to split some of the tray related preflight checks to handle some corner cases, will these cases still be handled properly?

anjannath:
> I have fuzzy memories that we needed to split some of the tray related preflight checks to handle some corner cases, will these cases still be handled properly?

as far as i remember, we wanted to split it because, we wanted to show the user exactly what step was being done during the setup. But now with these changes, its just one script that does everything, so no individual checks only `Installing CodeReady Containers tray...` and `Uninstalling CodeReady Containers tray...` to find out what step failed in the whole process we have to manually debug. (also the script is removed, after executing)

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: We already include the string inside quotes, which escapes most of the things but not the `$`, but i need to test with the backquote and quotes..

gbraad:
> because, we wanted to show the user exactly what step was being done during the setup.

Yep. The finegrained nature is because we also do not want to have a single "Installing CRC" with some progress indicator but shrouded in mystery. Especially as some of the issues with user accounts remains. So this tend to be clearer WHY somethings fails without the need to use a debug option.

But for the tray I do not mind, as this is a feature to enabled. but it is also a side effect due to UAC. Acceptable.

gbraad:
@jsliacan can you verify this works as-is?

jsliacan:
Checked this on Win10 in Bangalore. Only two things:

1. This password does not work: `test@"_\/` (with an extra ` at the end):
```
PS C:\Users\CDKQE> crc setup --enable-experimental-features                                                             
INFO Checking if goodhosts binary is cached                                                                             
NFO Checking if user is a member of the Hyper-V Administrators group                                                   
INFO Checking if tray binary is present                                                                                 
INFO Checking if tray is installed                                                                                      
INFO Installing CodeReady Containers tray                                                                               
? Enter account login password for service installation [? for help] **********                                         
INFO Will run as admin: Installing system tray for CodeReady Containers                                                 
FATA Installation script didn't execute successfully: CreateFile C:\Users\CDKQE\AppData\Local\Temp\crc032378815\success: The system cannot find the file specified. 
```

2. Sometimes we print `System Tray` like [here](https://github.com/code-ready/crc/pull/1373/files#diff-eb52fc4ccbf11c95e895fcf521912a36R27) and sometimes we use `system tray` like [here](https://github.com/code-ready/crc/pull/1373/files#diff-eb52fc4ccbf11c95e895fcf521912a36R76), [here](https://github.com/code-ready/crc/pull/1373/files#diff-eb52fc4ccbf11c95e895fcf521912a36R112), and [here](https://github.com/code-ready/crc/pull/1373/files#diff-34a389830398302de6ccaba5fa1733cbR61). There could be other instances which I overlooked.

Otherwise -- looks good!

pkg/crc/preflight/preflight_tray_powershell_windows.go
```diff
@@ -0,0 +1,170 @@
+package preflight
+
+import (
+	"fmt"
+	"strings"
+)
+
+var (
+	trayInstallationScript = []string{
```
praveenkumar, line 9: @anjannath why don't you use something like so you don't need to use a different function just to join them with a new line?
```golang
taryInstallationScriptTemplate = `$ErrorActionPreference = "Stop"
$password = "%s"
$tempDir = "%s"
...
`
```

Issue #1238, opened 2020-05-13
Document non-customizable cluster settings


robin-owen:
The OpenShift cluster running in the CodeReady Containers VM contains some configurable options which should not (or cannot) be modified by the end-user.

The following settings are non-customizable:

1. Domain name (`*.crc.testing`).
2. Internal cluster settings such as the service network/address range for internal cluster communication.
    1. The cluster uses the 172 address range. This can cause issues when, for example, a proxy is run in the same address space.

We can document this information in the existing section "Differences from a production OpenShift installation".

robin-owen:
Reopen, we're addressing this with #1375.

robin-owen:
Fixed by #1375.

PR #1375, opened 2020-07-09
Issue #1238 Document non-customizable cluster settings

robin-owen:
**Fixes:** Issue #1238

## Solution/Idea

Document settings which should not (or cannot) be modified by the end-user. This information is included in the "Differences from a production OpenShift installation" section.

### DIFF
--- a/docs/source/topics/con_differences-from-production-openshift-install.adoc
+++ b/docs/source/topics/con_differences-from-production-openshift-install.adoc
@@ -10,3 +10,11 @@
 ** For the same reason, there is no upgrade path to newer OpenShift versions.
 * The OpenShift instance runs in a virtual machine.
 This may cause other differences, particularly with external networking.
+
+{prod} also includes the following non-customizable cluster settings.
+These settings should not be modified:
+
+* Use of the ***.crc.testing** domain.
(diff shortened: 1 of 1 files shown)


docs/source/topics/con_differences-from-production-openshift-install.adoc
```diff
@@ -10,3 +10,11 @@
 ** For the same reason, there is no upgrade path to newer OpenShift versions.
 * The OpenShift instance runs in a virtual machine.
 This may cause other differences, particularly with external networking.
+
+{prod} also includes the following non-customizable cluster settings.
+These settings should not be modified:
+
+* Use of the ***.crc.testing** domain.
```
praveenkumar, line 17: @robin-owen this might be changeable, we need to look bit more closely apart from this all other points are valid.

robin-owen: @praveenkumar Should we remove this as a non-customizable setting in that case?

praveenkumar: @robin-owen I think let's have it here atm and then when we have a solution we can actually try to remove it.

Issue #1378, opened 2020-07-10
Failed to propagate proxy settings to cluster


jeet1990:
### General information

  * OS: Linux 
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it -YES

## CRC version
```bash
# Put the output of `crc version`
```
[saan@localhost ~]$ crc version
CodeReady Containers version: 1.12.0+6710aff
OpenShift version: 4.4.8 (embedded in binary)
[saan@localhost ~]$

  
## CRC status
```bash
# Put the output of `crc status`
```

[saan@localhost ~]$ crc status
CRC VM:          Running
OpenShift:       Running (v4.4.8)
Disk Usage:      12.15GB of 32.72GB (Inside the CRC VM)
Cache Usage:     12.27GB
Cache Directory: /home/saan/.crc/cache
[saan@localhost ~]$

## CRC config
```bash
# Put the output of `crc config view`
```

[saan@localhost ~]$ crc config view
- cpus                                  : 6
- http-proxy                            : 1x.1xx.xxx.xxx:8080
- https-proxy                           : 1x.1xx.xxx.xx:8080
- memory                                : 12000
- no-proxy                              : localhost,127.0.0.1,192.168.*.*,*.apps-crc.testing

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
# put the output of `sw_vers` in case of Mac
# Put the output of `systeminfo` in case of Windows
```
[saan@localhost ~]$ cat /etc/os-release
NAME="Red Hat Enterprise Linux"
VERSION="8.2 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.2"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.2 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8.2:GA"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.2"

### Actual
getting the error.
[...]

PR #1379, opened 2020-07-10
Check that proxy URL starts with http:// or https://

guillaumerose:
**Relates to:** Issue #1378

The validation didn't check that the proxy URL starts with http or https.


pkg/crc/network/proxy.go
```diff
@@ -133,6 +133,9 @@ func ValidateProxyURL(proxyUrl string) error {
 		return nil
 	}
 
+	if !strings.HasPrefix(proxyUrl, "https://") && !strings.HasPrefix(proxyUrl, "http://") {
+		return fmt.Errorf("Proxy URL '%s' is not valid: url should start with http:// or https://", proxyUrl)
```
praveenkumar, line 137: As per doc https://docs.openshift.com/container-platform/4.4/networking/enable-cluster-wide-proxy.html#enable-cluster-wide-proxy, under section `3. Configure the necessary fields for the proxy:` it is documented that `A proxy URL to use for creating HTTPS connections outside the cluster. If this is not specified, then httpProxy is used for both HTTP and HTTPS connections. The URL scheme must be http; https is currently not supported.` so we might need to only support `http` part

guillaumerose: Nice ! I changed the condition and added an explicit error message for https.

PR #1383, opened 2020-07-13
Add mock bundle logic in makefile

praveenkumar:
In the CI we can run the release target using mock
bundles which will help us to detect any issue due to changes
in release target.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ script:
 - make cross-lint
 - make cross
 - make test
-- alias podman=docker
+- make BUNDLE_DIR=/tmp MOCK_BUNDLE=true release
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,5 @@ script:
 - make cross-lint
 - make cross
 - make test
+- alias podman=docker
+- make BUNDLE_DIR=/tmp MOCK_BUNDLE=true release
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,12 @@ language: go
 go:
 - 1.13.4
 
+before_script:
+  - echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04 /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+  - curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04/Release.key | sudo apt-key add -
+  - sudo apt-get update -qq
+  - sudo apt-get -qq -y install podman slirp4netns
+
 script:
 - make
 - make vendorcheck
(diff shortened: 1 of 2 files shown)


guillaumerose:
Excellent idea!

gbraad:
You mean, to force the build and tests to run against an empty (but existing) cluster ?

I fail to see how this would work, as the bundle logic would fail with this, right?

.travis.yml
```diff
@@ -1,11 +1,20 @@
...
 language: go
 
 go:
 - 1.13.4
 
+before_script:
+  - echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04 /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+  - curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04/Release.key | sudo apt-key add -
+  - sudo apt-get update -qq
+  - sudo apt-get -qq -y install podman slirp4netns
```
anjannath, line 12: Why do we need to install podman, does `make release` need it?

praveenkumar: @anjannath yes make release need the podman

Makefile
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+SHELL := /bin/bash
+
```
cfergeau, line 16: If we want to avoid forcing a bash dependency, I believe an alternative is to go with a subshell:
```
diff --git a/Makefile b/Makefile
index db4ec7ce..4d2ea398 100644
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 
-SHELL := /bin/bash
-
 BUNDLE_VERSION = 4.5.1
 # OC_VERSION and BUNDLE_VERSION are going to same for release artifacts but
 # different for nightly and CI bits where bundle version would be any random
@@ -190,8 +188,8 @@ release: cross-lint embed_bundle build_docs_pdf gen_release_info
 	cd $(BUILD_DIR) && zip -r $(CURDIR)/$(RELEASE_DIR)/crc-windows-amd64.zip crc-windows-$(CRC_VERSION)-amd64
 
 	@mv $(RELEASE_INFO) $(RELEASE_DIR)/$(RELEASE_INFO)
-	
-	pushd $(RELEASE_DIR) && sha256sum * > sha256sum.txt && popd
+
+	( cd $(RELEASE_DIR) && sha256sum * > sha256sum.txt )
 
 HYPERKIT_BUNDLENAME = $(BUNDLE_DIR)/crc_hyperkit_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
 HYPERV_BUNDLENAME = $(BUNDLE_DIR)/crc_hyperv_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
```

Issue #1384, opened 2020-07-13
daemon: listen on named pipes in windows instaead of unix socket


anjannath:
One use case of the daemon is to provide an api to be used from vscode extension which uses javascript, nodejs.
On windows trying to connect to the daemon socket `~/.crc/crcd.sock` from javascipt returns error.

according to nodejs docs, the alternative is to use named pipes on windows (https://nodejs.org/api/net.html#net_identifying_paths_for_ipc_connections)

Linked pull requests:
  * [ ] #1385 
  * [ ] https://github.com/code-ready/tray-windows/issues/6

PR #1385, opened 2020-07-14
Issue #1384 Listen using named pipe for daemon on windows

anjannath:
One use case of the daemon is to provide an api to be used
from vscode extension which uses javascript on nodejs.

On windows however it is not possible to connect to unix sockets
from nodejs, according to their docs, the alternative is to  use
named pipes.

https://nodejs.org/api/net.html#net_identifying_paths_for_ipc_connections

Issue on the tray repo: https://github.com/code-ready/tray-windows/issues/6


pkg/crc/api/api.go
```diff
@@ -9,14 +9,21 @@ import (
...
+type listenerType int
+
+const (
+	localListenerType listenerType = iota
+	remoteListenerType
+)
+
 func CreateApiServer(socketPath string) (CrcApiServer, error) {
-	listener, err := net.Listen("unix", socketPath)
+	l, err := getListener(socketPath, localListenerType)
```
cfergeau, line 20: you can keep the `listener` name

pkg/crc/api/api.go
```diff
@@ -30,6 +37,18 @@ func CreateApiServer(socketPath string) (CrcApiServer, error) {
...
 
+func getListener(socketPath string, listenT listenerType) (net.Listener, error) {
+	switch listenT {
+	case localListenerType:
+		l, err := createIPCListener(socketPath)
+		return l, err
+	case remoteListenerType:
+		return nil, fmt.Errorf("Unimplemented listener requested.")
+	default:
+		return nil, fmt.Errorf("Unimplemented listener requested.")
```
cfergeau, line 48: We probably can get rid of the `listenerType` argument for now, and readd it when there is a patch series which needs it?

pkg/crc/api/listener_windows.go
```diff
@@ -0,0 +1,28 @@
...
+	 * can be found at:
+	 *	- https://docs.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-string-format
+	 *	- https://itconnect.uw.edu/wares/msinf/other-help/understanding-sddl-syntax/
+	 */
+	pipeConfig := winio.PipeConfig{
+		SecurityDescriptor: "D:P(A;;GA;;;BU)",
+	}
+	listener, err := winio.ListenPipe("\\\\.\\pipe\\crc", &pipeConfig)
+	if err != nil {
+		logging.Error("Failed to create socket: ", err.Error())
```
cfergeau, line 23: `Failed to create named pipe xxxx: err.Error()`

pkg/crc/api/listener_windows.go
```diff
@@ -0,0 +1,28 @@
...
+	/* the named pipe is  configured that built in users are able
+	 * to read,write from the pipe, info about security descriptor
+	 * can be found at:
+	 *	- https://docs.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-string-format
+	 *	- https://itconnect.uw.edu/wares/msinf/other-help/understanding-sddl-syntax/
+	 */
+	pipeConfig := winio.PipeConfig{
+		SecurityDescriptor: "D:P(A;;GA;;;BU)",
+	}
+	listener, err := winio.ListenPipe("\\\\.\\pipe\\crc", &pipeConfig)
```
cfergeau, line 21: why aren't we using `nil` as the config, which seems to be a supported value ( https://github.com/microsoft/go-winio/blob/master/pipe_test.go )

pkg/crc/api/listener_nonwin.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"net"
+
+	"github.com/code-ready/crc/pkg/crc/logging"
+)
+
+func createIPCListener(socketPath string) (net.Listener, error) {
+	listener, err := net.Listen("unix", socketPath)
+	if err != nil {
+		logging.Error("Failed to create socket: ", err.Error())
```
cfergeau, line 14: This could be "Failed to create unix socket at $socketPath: err.Error()"

pkg/crc/api/listener_windows.go
anjannath: We need this to successfully connect to the pipe, when its `nil`, it creates the pipe with restricted permissions and we can't connect to it, setting the `SecurityDescriptor` makes it accessible to other users on the machine.

pkg/crc/api/api.go
anjannath: Yeah, definitely as i was talking to @gbraad about the remote scenario! will remove this when removing wip.

pkg/crc/api/listener_windows.go
cfergeau: Hmm I assume this is required because of how the way we start the daemon? Does this mean anybody with an account on the Windows machine will be able to talk with the daemon, and, for example, get the pull secret of the user running crc?

pkg/crc/api/api.go
```diff
@@ -10,11 +10,11 @@ import (
 )
 
 func CreateApiServer(socketPath string) (CrcApiServer, error) {
-	listener, err := net.Listen("unix", socketPath)
+	listener, err := getListener(socketPath)
```
cfergeau, line 13: This can call `createIPCListener()` directly now since `getListener()` is just `return createIPCListener()`

pkg/crc/api/listener_windows.go
anjannath: > Hmm I assume this is required because of the way we start the daemon?

Not exactly, if we pass a nil `PipeConfig` the pipe uses the default settings.. from ms docs (https://docs.microsoft.com/en-us/windows/win32/ipc/named-pipe-security-and-access-rights)

`You can specify a security descriptor for a named pipe when you call the CreateNamedPipe function. The security descriptor controls access to both client and server ends of the named pipe. If you specify NULL, the named pipe gets a default security descriptor. The ACLs in the default security descriptor for a named pipe grant full control to the LocalSystem account, administrators, and the creator owner. They also grant read access to members of the Everyone group and the anonymous account.`


> Does this mean anybody with an account on the Windows machine will be able to talk with the daemon, and, for example, get the pull secret of the user running crc?

Yes, this gives all the users in the machine access to the pipe, i'll change this to give permission only to the current user. Thanks!

Although the microsoft docs mention that it gives '_administrators_' access to the pipe by default, i wasn't able to actually read/write to it despite being an admin :-/

pkg/crc/api/listener_nonwin.go
```diff
@@ -0,0 +1,18 @@
...
+package api
+
+import (
+	"net"
+
+	"github.com/code-ready/crc/pkg/crc/logging"
+)
+
+func createIPCListener(socketPath string) (net.Listener, error) {
+	listener, err := net.Listen("unix", socketPath)
```
guillaumerose, line 12: We can also check the socketPath length. For Darwin, the maximum is 104 bytes. For Linux, it's 108. (ref https://github.com/moby/vpnkit/pull/506 or https://github.com/moby/moby/pull/13408). 

But we can do that in an other PR!

pkg/crc/api/listener_windows.go
cfergeau: > Although the microsoft docs mention that it gives 'administrators' access to the pipe by default, i wasn't able to actually read/write to it despite being an admin :-/

Yes, still not clear to me why the default security descriptor would not work? What is special about the reader process or the writer process? They are not running as the current user?

anjannath: > What is special about the reader process or the writer process? They are not running as the current user

They aren't special, well the daemon is a service.. and both are running as the current user.

the daemon creates the pipe so it has all the privileges (creator owner), now if we are having the default security descriptor, the client should be able to read/write to it, because the default security descriptor allows members of the _local Admin group_ to do so. but while testing, this was not the case, so i explicitly added the current user's sid.

One way i'll try to find out why its not working is, right now the daemon which is a service creates the pipe maybe that's the reason, so i'll test by creating a pipe not from the service but with a test program and see if it behaves the same.

cfergeau: Yes, this is what I meant when I asked if there was anything special about the reader or the writer. I would not be surprised if the way we launch the daemon as a service was causing this. I still wonder if it's possible to spawn the daemon directly from the tray process since I think it would make everything much simpler... :)

anjannath: @cfergeau I don't think this would simplify we'll have to figure out which crc binary to use for the daemon from the tray, also the whole setup process is driven by the `crc setup` command.

> One way i'll try to find out why its not working is, right now the daemon which is a service creates the pipe maybe that's the reason, so i'll test by creating a pipe not from the service but with a test program and see if it behaves the same.

I did try out this, and if not running as a service, it works without the permission modifications.

I think since we are only adding the current users SID it makes it more secure, that only the current user can access the pipe.

cfergeau: > @cfergeau I don't think this would simplify we'll have to figure out which crc binary to use for the daemon from the tray, also the whole setup process is driven by the crc setup command.

I don't think this would add up as more code than "running as a service" + "UAC prompt removal" + "pipe permissions", I'd expect it would be much smaller than that.

guillaumerose:
This is still a good think to do!

Issue #1387, opened 2020-07-14
Implement clock-skew flag to skew time in the VM to test cert rotation in the CI


praveenkumar:
It would be good to create the VM with skew time so that we can test the cert rotation as part of CI.  We can make this `time-skew` flag hidden also so that it is not visible to end users.

```
$ crc start --clock-skew <number of days>
```

cfergeau:
If you want to change the time of the crc VM, it can also be achieved by changing its libvirt domain definition, without stopping chronyd or running any guest-side  commands.

gbraad:
while right for the firstset of tests, this has to be reproducible on other platforms. I would suggest an external command to interact with the VM, but not sure if this needs to be part of the `crc` binary

praveenkumar:
This is fixed by #1390

PR #1390, opened 2020-07-15
Issue #1387 Add stop-chronyd hidden bool flag to start

praveenkumar:
To test the cert rotation we need to stop the chronyd service
in the VM so it will not sync with ntp server. On the host side
it is platform specific, like on linux host you do need to stop
the chronyd service and set the time in future (around 2 month)
`sudo date -s "2 month"` and then start crc with `--stop-chronyd`
flag to test out cert rotation.

**Fixes:** Issue #1387 

## Solution/Idea

Test out cert rotation in the CI

## Proposed changes
Add a hidden option `stop-chronyd` as part of start flag

## Testing
On the linux host 

```
$ sudo systemctl stop chronyd.service
$ sudo date -s "2 month"
$ ./crc start -b ~/Downloads/crc_libvirt_4.5.1.crcbundle --log-level debug --stop-chronyd
```

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -4,8 +4,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"time"
 
 	"github.com/code-ready/crc/pkg/crc/cluster"
@@ -242,6 +244,14 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		return startError(startConfig.Name, "Failed to connect to the CRC VM with SSH -- host might be unreachable", err)
 	}
 	logging.Info("CodeReady Containers VM is running")
+
+	// Start network time synchronization if `CRC_DEBUG_ENABLE_STOP_NTP` not set
+	if b, _ := strconv.ParseBool(os.Getenv("CRC_DEBUG_ENABLE_STOP_NTP")); !b {
+		logging.Info("Stopping network time synchronization in CodeReady Containers VM")
+		if _, err := sshRunner.Run("sudo timedatectl set-ntp on"); err != nil {
+			return startError(startConfig.Name, "Failed to stop network time synchronization", err)
+		}
+	}
 
 	// Check the certs validity inside the vm
 	needsCertsRenewal := false
--- a/test/integration/features/cert_rotation.feature
+++ b/test/integration/features/cert_rotation.feature
@@ -1,0 +1,25 @@
+@cert_rotation @linux
+Feature: Check the cert is rotation happen after it expire
+
+    The user try to use crc after one month. he/she expect the
+    cert rotation happen successfully and able to deploy the app and check its
+    accessibility.
+
+    Scenario: Set clock to 3 month ahead on the host
+      Given executing "sudo timedatectl set-ntp off" succeeds
+      Then executing "sudo date -s '3 month'" succeeds
+
+    Scenario: Start CRC
+      Given executing "crc setup" succeeds
+      When starting CRC with default bundle along with stopped network time synchronization succeeds
+      Then stdout should contain "Started the OpenShift cluster"
+      And executing "eval $(crc oc-env)" succeeds
+      When with up to "4" retries with wait period of "2m" command "crc status --log-level debug" output matches ".*Running \(v\d+\.\d+\.\d+.*\).*"
+      Then login to the oc cluster succeeds
+
+    Scenario: Set clock back to original time
+      Given executing "sudo timedatectl set-ntp on" succeeds
+
+    Scenario: CRC delete
+      When executing "crc delete -f" succeeds
+      Then stdout should contain "Deleted the OpenShift cluster"
--- a/developing.adoc
+++ b/developing.adoc
@@ -81,6 +81,24 @@ make integration GODOG_OPTS="--godog.tags='@basic && @windows'" BUNDLE_LOCATION=
 
 Please notice `@basic && @windows`, where `@basic` tag stands for `basic.feature` file and `@windows` tag for integration tests designed for Windows.
 
+[[how-to-test-cert-rotation]]
+=== How to test cert rotation
+
+On linux platform first stop the network time sync using:
+```
+$ sudo timedatectl set-ntp off
+```
+
+Set the time 2 month ahead:
+```
+$ sudo date -s '2 month'
+```
+
+Start the crc with `CRC_DEBUG_ENABLE_STOP_NTP=true` set:
+```
+$ CRC_DEBUG_ENABLE_STOP_NTP=true crc start
+```
+
 
 [[integration-test-logs]]
 === Logs
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -41,6 +41,8 @@ func FeatureContext(s *godog.Suite) {
 		RemoveCRCHome)
 	s.Step(`^starting CRC with default bundle (succeeds|fails)$`,
 		StartCRCWithDefaultBundleSucceedsOrFails)
+	s.Step(`^starting CRC with default bundle along with stopped network time synchronization (succeeds|fails)$`,
+		StartCRCWithDefaultBundleWithStopNetworkTimeSynchronizationSucceedsOrFails)
 	s.Step(`^starting CRC with default bundle and nameserver "(.*)" (succeeds|fails)$`,
 		StartCRCWithDefaultBundleAndNameServerSucceedsOrFails)
 	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds|fails)$`,
@@ -341,6 +343,20 @@ func StartCRCWithDefaultBundleSucceedsOrFails(expected string) error {
 		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
 	}
 	cmd = fmt.Sprintf("crc start -p '%s' %s --log-level debug", pullSecretFile, extraBundleArgs)
+	err := clicumber.ExecuteCommandSucceedsOrFails(cmd, expected)
+
+	return err
+}
+
+func StartCRCWithDefaultBundleWithStopNetworkTimeSynchronizationSucceedsOrFails(expected string) error {
+
+	var cmd string
+	var extraBundleArgs string
+
+	if bundleEmbedded == false {
+		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
+	}
+	cmd = fmt.Sprintf("CRC_DEBUG_ENABLE_STOP_NTP=true crc start -p '%s' %s --log-level debug", pullSecretFile, extraBundleArgs)
 	err := clicumber.ExecuteCommandSucceedsOrFails(cmd, expected)
 
 	return err
--- a/test/integration/features/story_registry.feature
+++ b/test/integration/features/story_registry.feature
@@ -6,20 +6,18 @@ Feature: Local image to image-registry to deployment
     project/namespace. They deploy and expose the app and check its
     accessibility.
 
-    Scenario: Start CRC
+    Scenario: Start CRC and login to cluster
         Given executing "crc setup" succeeds
         When starting CRC with default bundle succeeds
         Then stdout should contain "Started the OpenShift cluster"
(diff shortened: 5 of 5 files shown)


cfergeau:
> we need to stop the chronyd service in the VM so it will not sync with ntp server

Nope, this is one way of achieving this, not the only one.

praveenkumar:
> Nope, this is one way of achieving this, not the only one.

@cfergeau I didn't mention that this is only way or you are saying this commit is interpreted like this ?

guillaumerose:
If I am correct, chronyd can still be working for couple of seconds between the VM boot and the ssh command. Not sure if it can cause unwanted failures.

It could be handy to have `crc ssh` command (hidden and only for testing?)

praveenkumar:
> If I am correct, chronyd can still be working for couple of seconds between the VM boot and the ssh command. Not sure if it can cause unwanted failures.

@guillaumerose yes but during my testing I didn't see any time mismatch and even on the CI, there is possibility but in that case host time would still be ahead 2 months so start might fail and so does the test.

> It could be handy to have crc ssh command (hidden and only for testing?)

@guillaumerose I didn't get how ssh command will help in cert rotation testing or you mean in general to have ssh command for testing different bit then I do agree but in the past we made a decision not providing an option to ssh command to end user.

cfergeau:
> > Nope, this is one way of achieving this, not the only one.
> 
> @cfergeau I didn't mention that this is only way or you are saying this commit is interpreted like this ?

> « To test the cert rotation we need to stop the chronyd service in the VM so it will not sync with ntp server. »

I tend to read "we need to" as "the only way to test cert rotation is to stop the chronyd service in the VM"

cmd/crc/cmd/start.go
```diff
@@ -38,6 +38,7 @@ var startCmd = &cobra.Command{
 
 var (
 	startCmdFlagSet = initStartCmdFlagSet()
+	stopChronyd     = "stop-chronyd"
```
cfergeau, line 41: For what it's worth, I would have gone with an environment variable (CRC_DEBUG_ENABLE_TIME_SKEW with a value in seconds) rather than a command line flag.

cmd/crc/cmd/start.go
```diff
@@ -80,6 +82,11 @@ func initStartCmdFlagSet() *pflag.FlagSet {
 	flagSet.IntP(config.Memory.Name, "m", constants.DefaultMemory, "MiB of memory to allocate to the OpenShift cluster")
 	flagSet.StringP(config.NameServer.Name, "n", "", "IPv4 address of nameserver to use for the OpenShift cluster")
 	flagSet.Bool(config.DisableUpdateCheck.Name, false, "Don't check for update")
+	// clock-skew flag is used to test certificate rotation
+	flagSet.Bool(stopChronyd, false, "Stop the chronyd service")
+	if err := flagSet.MarkHidden(stopChronyd); err != nil {
+		logging.Errorf("Not able to mark %s as hide: %v", stopChronyd, err)
```
cfergeau, line 88: `hide` -> `hidden`

pkg/crc/machine/machine.go
```diff
@@ -253,6 +253,15 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	}
 	logging.Info("CodeReady Containers VM is running")
 
+	// Stop Chronyd service
+	if startConfig.StopChronyd {
+		logging.Info("Stopping chronyd service in CodeReady Containers VM")
+		if _, err := sshRunner.Run("sudo systemctl stop chronyd"); err != nil {
```
cfergeau, line 259: This feels more low-level than `sudo timedatectl set-ntp off`, any reason to directly interact with `chronyd`?

test/integration/features/cert_rotation.feature
```diff
@@ -0,0 +1,25 @@
+@cert_rotation @linux
+Feature: Check the cert is rotation happen after it expire
+
+    The user try to use crc after one month. he/she expect the
+    cert rotation happen successfully and able to deploy the app and check its
+    accessibility.
+
+    Scenario: Set clock to 3 month ahead on the host
+      Given executing "sudo systemctl stop chronyd" succeeds
+      Then executing "sudo date -s '3 month'" succeeds
```
cfergeau, line 10: Maybe this should also be done by `crc --stop-chronyd`?

praveenkumar: @cfergeau that would be platform depend so I would prefer to have user decide on this.

pkg/crc/machine/machine.go
praveenkumar: @cfergeau I didn't check `timedatectl set-ntp off ` but looking at the man page looks like it also start/stop underline service in timedatectl side it is `systemd-timesyncd.service` or `chronyd` so I am not sure if it have any improvement over how we are interacting.

```
set-ntp [BOOL]
           Takes a boolean argument. Controls whether network time synchronization is active and enabled (if available). If the argument is true, this enables and starts the first existing network synchronization service. If the
           argument is false, then this disables and stops the known network synchronization services. The way that the list of services is built is described below.

$ timedatectl set-ntp true
           ==== AUTHENTICATING FOR org.freedesktop.timedate1.set-ntp ===
           Authentication is required to control whether network time synchronization shall be enabled.
           Authenticating as: user
           Password: ********
           ==== AUTHENTICATION COMPLETE ===

           $ systemctl status systemd-timesyncd.service
           ● systemd-timesyncd.service - Network Time Synchronization
              Loaded: loaded (/usr/lib/systemd/system/systemd-timesyncd.service; enabled)
              Active: active (running) since Mo 2015-03-30 14:20:38 CEST; 5s ago
                Docs: man:systemd-timesyncd.service(8)
            Main PID: 595 (systemd-timesyn)
              Status: "Using Time Server 216.239.38.15:123 (time4.google.com)."
              CGroup: /system.slice/systemd-timesyncd.service
                      └─595 /usr/lib/systemd/systemd-timesyncd

```

cmd/crc/cmd/start.go
praveenkumar: > CRC_DEBUG_ENABLE_TIME_SKEW with a value in seconds) rather than a command line flag.

@cfergeau  This is just a bool flag and yes we can go with env variable if you prefer.

praveenkumar:
> I tend to read "we need to" as "the only way to test cert rotation is to stop the chronyd service in the VM". Since there were other alternatives discussed in the past, I'd expect a mention somewhere why this was chosen over the other solutions (or just "we need to be able to arbitrarily set the VM time. One way of doing that is to stop the chronyd service in the VM to prevent ntp syncs")

@cfergeau The other solution ( using libvirt domain modification) is only for linux, I don't think we actually have any other solutions which is platform independent apart from stopping the ntp sync from the VM and also from the host.

cfergeau:
> > I tend to read "we need to" as "the only way to test cert rotation is to stop the chronyd service in the VM". Since there were other alternatives discussed in the past, I'd expect a mention somewhere why this was chosen over the other solutions (or just "we need to be able to arbitrarily set the VM time. One way of doing that is to stop the chronyd service in the VM to prevent ntp syncs")
> 
> @cfergeau The other solution ( using libvirt domain modification) is only for linux, I don't think we actually have any other solutions which is platform independent apart from stopping the ntp sync from the VM and also from the host.

I'd like to have the 'platform independance' mentioned in the commit log since you mentioned this rationale twice in answer to my comments :)

test/integration/features/cert_rotation.feature
cfergeau: Feels odd to add code for half of the solution to crc, and declare the user is on their own for the other half.

pkg/crc/machine/machine.go
cfergeau: > @cfergeau I didn't check `timedatectl set-ntp off ` but looking at the man page looks like it also start/stop underline service in timedatectl side it is `systemd-timesyncd.service` or `chronyd` so I am not sure if it have any improvement over how we are interacting.

This shields us from the NTP service changing in the future, if there's a switch in the future from `chronyd` to `systemd-timesyncd`, `timedatectl` command line will be unchanged, while `systemctl stop chronyd` would stop working and we'd need to change our code.

cmd/crc/cmd/start.go
cfergeau: It should explicitly set the time in the VM, or you are going to have a short window when `chronyd` is running, and you are going to race with this and assume the time is not going to be reset through NTP before you get a chance to disable NTP.

But I'm also wondering if it would not be better to just run externally a loop trying to ssh into the VM and shutdown NTP, this way no need for very special code in crc, and all time hacks would be located in the same place (integration tests)

praveenkumar: @cfergeau it's not with integration test, if test failed on the CI I want to do it locally then this gives me a better way to recreate env then running integraiton test locally. About the race condition it the time sync with ntp on the VM then also host time differ and we will have failed test.

cfergeau: > it's not with integration test, if test failed on the CI I want to do it locally then this gives me a better way to recreate env then running integraiton test locally.

You need to know and run some manual steps locally already if you want to reproduce the CI environment locally, so this does not feel like a very strong argument :-/

> About the race condition it the time sync with ntp on the VM then also host time differ and we will have failed test.

But we don't want CI to fail because it triggered a race condition in the integration test implementation, do we?

praveenkumar:
> Should this be done in a small separate go helper which would ssh in a loop to run the "stop ntp" command, and which would also have the host side code to set the time/date, or which would warn you to do it manually on platforms where this cannot be done? This way we don't need to change crc code, we still have something which can easily be run for debugging, and this is even easier to use as this also is doing the host side changes.

@cfergeau This hack is just to add have integration test for cert rotation, I don't want to spend time on creating a separate go helper right now (we can put a follow up issue and then look when time permit) . This code base is not run by the user and it is only for developer so developer/QE knows how to make host specific time change on platform specific. I don't mind if this env variable should be part of development doc.

cfergeau:
> @cfergeau This hack is just to add have integration test for cert rotation, I don't want to spend time on creating a separate go helper right now (we can put a follow up issue and then look when time permit) . This code base is not run by the user and it is only for developer so developer/QE knows how to make host specific time change on platform specific. I don't mind if this env variable should be part of development doc.

If there is no time to consider alternatives now, I would suggest adding this as a patch to the CI build rather than having this in the upstream codebase.

cfergeau:
I don't have a fully functional cluster to test how this behaves after a stop/start cycle, but see https://github.com/cfergeau/crc/commit/2e33b5cc7c5e5ad8ffa0e06e71ceff27911c523c which runs a command in the VM through SSH as soon as possible

guillaumerose:
Even the code of Christophe is not race condition free. Should we change the SNC VM and disable some services by default? crc will launch them. 
It's maybe too much for testing purpose but test flakiness is a real burden.

praveenkumar:
> Even the code of Christophe is not race condition free. Should we change the SNC VM and disable some services by default? crc will launch them.

@guillaumerose yes this is another option but if we are doing that, it means we again need to execute a unwanted command every single time on user side which take around 5-10 sec to execute.

praveenkumar:
> I don't have a fully functional cluster to test how this behaves after a stop/start cycle, but see cfergeau@2e33b5c which runs a command in the VM through SSH as soon as possible

@cfergeau How come this is any better than what we have right now? with this change we now need to maintain another command just to check this feature to me it feels like bit extra.

gbraad:
I would prefer to be able to run a command, as suggested (right?) but hidden behind a flag or just a hidden command.

I do not see the need for a specific command to handle this. unless it can be modularized and uised for multiple purposes, hence a ssh command run? but this is not for generic use.

cfergeau:
> > I don't have a fully functional cluster to test how this behaves after a stop/start cycle, but see [cfergeau/crc@2e33b5c](https://github.com/cfergeau/crc/commit/2e33b5c) which runs a command in the VM through SSH as soon as possible
> 
> @cfergeau How come this is any better than what we have right now? with this change we now need to maintain another command just to check this feature to me it feels like bit extra.

But the code is totally separate from `crc`, this means we did not add very specialized code to the main `crc` binary for a test. Where do we draw the line now? Is it acceptable to add one env var per integration test, plus a 2 line change in the main `crc` binary to change one small thing needed for each integration test? The separate binary had the advantage of cleanly keeping the small needed hack on the testing side, rather than in the code which we are releasing.

PR #1393, opened 2020-07-15
Remove the crc machine directory before removing the VM

praveenkumar:
Right now removeCRCVM() function return as soon as it sees there
is no crc domain which might give false positive for cleanup scenario
consider following.

- crc start
- sudo virsh --connect qemu:///system destroy crc
- sudo virsh --connect qemu:///system undefine crc
- crc delete
Do you want to delete the OpenShift cluster? [y/N]: y
(crc) Failed to fetch machine
ERRO Failed to fetch machine 'crc'
- crc cleanup --log-level debug
[...]
DEBU Running 'virsh --connect qemu:///system domstate crc'
DEBU stdout:
DEBU stderr: error: failed to get domain 'crc'
INFO Removing /etc/NetworkManager/dnsmasq.d/crc.conf file
- ls ~/.crc/machines/crc/
config.json  crc  id_rsa  id_rsa.pub  kubeconfig

Above you can see we just check about crc domain status and return
and step not reached to delete the machine dir.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -366,6 +366,12 @@ func removeLibvirtCrcNetwork() error {
 
 func removeCrcVM() error {
 	logging.Debug("Removing 'crc' VM")
+
+	if err := os.RemoveAll(constants.MachineInstanceDir); err != nil {
+		logging.Debugf("Error removing %s dir: %v", constants.MachineInstanceDir, err)
+		return fmt.Errorf("Error removing %s dir", constants.MachineInstanceDir)
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -366,6 +366,12 @@ func removeLibvirtCrcNetwork() error {
 
 func removeCrcVM() error {
 	logging.Debug("Removing 'crc' VM")
+
+	if err := os.RemoveAll(constants.MachineInstanceDir); err != nil {
```
cfergeau, line 370: One key thing to know is that `os.RemoveAll("/non/existing/path")` will return `nil` if any of the path components is non-existing. Without this, this would have fixed one corner case, and broken another one.
I'm mentioning this for posterity as I had to check how `os.RemoveAll` worked :)

Issue #1216, opened 2020-05-05
create/update kubeconfig on cluster start 


gorkem:
Can we create or update a `kubeconfig`  as part of the cluster start? We can have one cluster 2 contexts (kubeadmin & developer). If KUBECONFIG is set on the environment we can append the new files to the KUBECONFIG env. variable print its location to the crc output if the user would like to make that more permanent.  If KUBECONFIG is not set, we can set it to this new file and  $HOME/.kube/config.

praveenkumar:
@gorkem we do have a kubeconfig file `~/.crc/machines/crc/kubeconfig` which have `system:admin` context, I think for developer and kubeadmin are the users and if we set the context on the current kubeconfig or using a new one it will based on the token which expire after ~24 hour iirc. Any specific reason we do need this?

gorkem:
I do not think anybody keeps CRC running for longer than 24 hours :), and crc start should update the token. I was not aware that we had a kubeconfig and can't recall a reference to it on  docs. I checked the file and the context is named `admin`. It should reflect that it is crc. Perhaps `crc-admin` and `crc-developer` is more descriptive. 

There are multiple use cases that this can help with. For instance when I start CRC through tray and move to my IDE my cluster does not appear on the list until I a `oc login`. It allows me to use a tool like `kubectx` easily, especially if we can change the naming as descriptive as I explained above. It does shorten the path from `crc start` to `odo push` a couple of steps.

gorkem:
This is still valid

praveenkumar:
This is fixed by #1396 , closing it.

PR #1396, opened 2020-07-17
Issue #1216 Update ~/.kube/config with crc contexts

guillaumerose:
**Fixes:** Issue #1216

Using the same strategy as minishift (https://github.com/minishift/minishift/pull/2881/files), update kubeconfig with crc-developer and crc-admin context.

It vendors k8s code and a little part of the openshift/oc `tokencmd.RequestToken`. It implies a large list of replace in go.mod but most of them are only used for dependency resolution and are not included in the final vendor directory. This list is taken from https://github.com/openshift/oc/blob/release-4.5/go.mod#L78 (I tried to trim it without success).
It also allows us to remove this kind of struct https://github.com/code-ready/crc/blob/master/pkg/crc/cluster/clusteroperator.go#L15


cfergeau:
One thing to have in mind is that this doubles build times (from a few seconds to 2*a few seconds) and binary size (from 10MB to 20MB).

pkg/crc/machine/machine.go
```diff
@@ -41,25 +41,23 @@ func init() {
 	}
 }
 
-func fillClusterConfig(bundleInfo *bundle.CrcBundleInfo, clusterConfig *ClusterConfig) error {
+func getClusterConfig(bundleInfo *bundle.CrcBundleInfo) (*ClusterConfig, error) {
```
cfergeau, line 44: I don't think this is related to adding crc contexts to ~/.kube/config?

pkg/crc/machine/machine.go
```diff
@@ -726,39 +745,35 @@ func GetProxyConfig(machineName string) (*network.ProxyConfig, error) {
...
 	// We might need to improve and use crc status logic, only
 	// return if the Openshift is running as part of status.
 	libMachineAPIClient := libmachine.NewClient(constants.MachineBaseDir, constants.MachineCertsDir)
 	host, err := libMachineAPIClient.Load(consoleConfig.Name)
 
 	if err != nil {
-		result.Success = false
-		result.Error = err.Error()
-		return *result, errors.New(err.Error())
+		return ConsoleResult{Error: err.Error()}, errors.New(err.Error())
```
cfergeau, line 755: Same comment here. This change looks more 'go-ish' to me, but I don't think it's required in this already huge change (because of the vendoring), and it's also inconsistent with the way all other methods return their result.

go.mod
```diff
@@ -12,21 +11,64 @@ require (
...
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.18.2
+	k8s.io/kubectl => github.com/openshift/kubernetes-kubectl v0.0.0-20200507115706-2f87de22f81a
+	k8s.io/kubelet => k8s.io/kubelet v0.18.2
+	k8s.io/kubernetes => github.com/openshift/kubernetes v1.17.0-alpha.0.0.20200427141011-f0879866c662
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.18.2
+	k8s.io/metrics => k8s.io/metrics v0.18.2
+	k8s.io/node-api => k8s.io/node-api v0.18.2
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.18.2
+	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.18.2
+	k8s.io/sample-controller => k8s.io/sample-controller v0.18.2
```
praveenkumar, line 73: @guillaumerose Are we sure these replace we really require?

pkg/crc/machine/machine.go
guillaumerose: yes removed!

go.mod
guillaumerose: I removed 3 more dependencies. `k8s.io/kubernetes` triggers a lot of imports :/

go.mod
```diff
@@ -12,22 +11,62 @@ require (
...
+	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.18.2
+	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.18.2
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.18.2
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.18.2
+	k8s.io/kubectl => github.com/openshift/kubernetes-kubectl v0.0.0-20200507115706-2f87de22f81a
+	k8s.io/kubelet => k8s.io/kubelet v0.18.2
+	k8s.io/kubernetes => github.com/openshift/kubernetes v1.17.0-alpha.0.0.20200427141011-f0879866c662
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.18.2
+	k8s.io/metrics => k8s.io/metrics v0.18.2
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.18.2
```
praveenkumar, line 71: @guillaumerose I am bit worry about this change, how we are going to maintain it in future, is it something taken from https://github.com/openshift/oc/blob/master/go.mod here?

guillaumerose: yes it's taken from oc. I just removed what I could.

go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/code-ready/crc
 go 1.13
 
 require (
-	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/Masterminds/semver v1.5.0
```
cfergeau, line 6: Fwiw, given their size, I'd suggest moving the go.mod/go.sum/vendor updates to a separate commit. They do not really make sense without the newly added code, but this would make the commits easier to look at/ignore.

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+	// Make sure .kube/config exist if not then this will create
+	_, _ = os.OpenFile(kubeconfig, os.O_RDONLY|os.O_CREATE, 0600)
+
+	cfg, err := clientcmd.LoadFromFile(kubeconfig)
+	if err != nil {
+		return err
+	}
+
+	cfg.Clusters[clusterName] = &api.Cluster{
+		Server:                constants.DefaultAPIURL,
```
cfergeau, line 35: shouldn't `clusterName` be derived from `constants.DefaultAPIURL`? And we should avoid hardcoding this as we can use `clusterConfig.ClusterAPI` when calling this function.

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+		InsecureSkipTLSVerify: true,
```
cfergeau, line 36: `~/.crc/machines/crc/kubeconfig` contains the CA authority which the cluster uses, I think we should make use of it rather than disabling TLS verification.

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+	if cfg.CurrentContext == "" {
+		cfg.CurrentContext = developerContext
+	}
+
+	return clientcmd.WriteToFile(*cfg, kubeconfig)
+}
+
+func addContext(cfg *api.Config, context, username, password string) error {
+	token, err := tokencmd.RequestToken(&restclient.Config{
+		Host: constants.DefaultAPIURL,
```
cfergeau, line 55: same comment about `DefaultAPIURL`

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+		TLSClientConfig: restclient.TLSClientConfig{
+			Insecure: true,
```
cfergeau, line 57: and same comment about disabling TLS verifications.

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	cfg.Contexts[context] = &api.Context{
+		Cluster:   clusterName,
+		AuthInfo:  context,
+		Namespace: "default",
+	}
+	return nil
+}
+
+// getGlobalKubeConfigPath returns the path to the first entry in KUBECONFIG environment variable
```
cfergeau, line 74: `the KUBECONFIG...`

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+// or if KUBECONFIG not set then $HOME/.kube/config
```
cfergeau, line 75: `is not set`

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+func getGlobalKubeConfigPath() string {
+	pathList := filepath.SplitList(os.Getenv("KUBECONFIG"))
+	if len(pathList) > 0 {
+		// Tools should write to last entry in the KUBECONFIG file instead first one.
```
cfergeau, line 79: `the last entry` ... `instead of the first one`

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,84 @@
...
+		// oc cluster up also follow same.
```
cfergeau, line 80: `follows` (or `does the same`?)

pkg/crc/machine/machine.go
```diff
@@ -417,6 +417,23 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
+		}
+		if isReady(status) {
+			return nil
+		}
+		return &errors.RetriableError{Err: goerrors.New("cluster operator authentication not ready")}
+	}, 2*time.Second); err != nil {
+		logging.Warn("Cannot update kubeconfig. Cluster operator authentication still not ready after 2min.")
+	} else if err := WriteKubeconfig(clusterConfig.KubeAdminPass); err != nil {
+		return startError(startConfig.Name, "Cannot update kubeconfig", err)
+	}
```
cfergeau, line 435: Can you put this in a helper?

go.mod
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/kubeconfig.go
guillaumerose: done :+1:

pkg/crc/machine/machine.go
guillaumerose: done :+1:

go.mod
praveenkumar: @guillaumerose Can we mention then in commit logs or may be a comment on this `go.mod` file so it would be easy for us t to track changes in future.

praveenkumar:
@guillaumerose Overall, this looks good to me, able to test it also. I think we should have it as part of single commit instead of vendoring as part of separate commit. I know that @cfergeau suggested it as part of review but to me it doesn't make sense to have vendor part as different commit then using actual module/package.

guillaumerose:
I merged the 2 commits. We should be good now!

The default context is admin. I chose it because otherwise, `oc get all` and `kubectl get all` return lot of errors and admin has already access to default project. Is it ok?

pkg/crc/machine/kubeconfig.go
```diff
@@ -0,0 +1,150 @@
...
+
+func addContext(cfg *api.Config, clusterConfig *ClusterConfig, ca []byte, context, username, password string) error {
+	host, err := hostname(clusterConfig.ClusterAPI)
+	if err != nil {
+		return err
+	}
+	token, err := tokencmd.RequestToken(&restclient.Config{
+		Host: clusterConfig.ClusterAPI,
+		TLSClientConfig: restclient.TLSClientConfig{
+			CAData: ca,
```
praveenkumar, line 123: @guillaumerose This makes the call from the host to the VM, that means host dns should be configured and working to resolve the `clusterConfig.ClusterAPI` hostname, I was looking into a way to get token using curl which we can execute directly inside the VM and get token.

```
$ curl -u developer:developer -kv -H "X-CSRF-Token: 1" 'https://oauth-openshift.apps-crc.testing/oauth/authorize?client_id=openshift-challenging-client&response_type=token' 2>&1 | grep -oP "access_token=\K[^&]*"
```

guillaumerose: It's now using a custom dialer that always connect to the VM IP. It skips the DNS resolution.

praveenkumar:
@guillaumerose looks like in case of proxy it is failing in the CI.

```
Command stderr: level=error msg="Cannot update kubeconfig: proxyconnect tcp: dial tcp 192.168.130.11:3128: connect: connection refused"
Cannot update kubeconfig: proxyconnect tcp: dial tcp 192.168.130.11:3128: connect: connection refused
```

guillaumerose:
Yes I will look at it. It also explains why some users might have issue with `crc` if `oc` talks to the proxy!

cfergeau:
> Yes I will look at it. It also explains why some users might have issue with `crc` if `oc` talks to the proxy!

This should be handled by `oc-env`
```
$ ./out/linux-amd64/crc oc-env
export PATH="/home/teuf/.crc/bin/oc:$PATH"
export HTTP_PROXY="http://squid.redhat.com:3128"
export HTTPS_PROXY="http://squid.redhat.com:3128"
export NO_PROXY="127.0.0.1,localhost,.testing"
# eval $(crc oc-env)
```

guillaumerose:
Why do we want the `oc` command to use a proxy ? Maybe there are some remote interactions (not only the local server?).

praveenkumar:
> Why do we want the oc command to use a proxy ? Maybe there are some remote interactions (not only the local server?).

@guillaumerose it is mostly because `noproxy` otherwise `.testing` also route from proxy which we want to avoid.

PR #1398, opened 2020-07-17
Limit the use of os.Exit(). Push up errors

guillaumerose:
Reduce the number of os.Exit() and use return err instead (esp. in non cmd/ code). This is required for the daemon: we don't want it to crash in the middle of a call. In the cmd package, it gives us a better understanding of the flow.

### DIFF
--- a/cmd/crc/cmd/console.go
+++ b/cmd/crc/cmd/console.go
@@ -2,7 +2,7 @@ package cmd
 
 import (
 	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/exit"
 	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/output"
 	"github.com/pkg/browser"
@@ -40,7 +40,7 @@ func runConsole(arguments []string) {
 
 	result, err := machine.GetConsoleURL(consoleConfig)
 	if err != nil {
-		errors.Exit(1)
+		exit.WithoutMessage(1)
 	}
 
 	if consolePrintURL {
@@ -55,11 +55,11 @@ func runConsole(arguments []string) {
 	}
 
 	if !machine.IsRunning(result.State) {
-		errors.ExitWithMessage(1, "The OpenShift cluster is not running, cannot open the OpenShift Web Console.")
+		exit.WithMessage(1, "The OpenShift cluster is not running, cannot open the OpenShift Web Console.")
 	}
 	output.Outln("Opening the OpenShift Web Console in the default browser...")
 	err = browser.OpenURL(result.ClusterConfig.WebConsoleURL)
 	if err != nil {
-		errors.ExitWithMessage(1, "Failed to open the OpenShift Web Console, you can access it by opening %s in your web browser.", result.ClusterConfig.WebConsoleURL)
+		exit.WithMessage(1, "Failed to open the OpenShift Web Console, you can access it by opening %s in your web browser.", result.ClusterConfig.WebConsoleURL)
 	}
 }
--- a/cmd/crc/cmd/console.go
+++ b/cmd/crc/cmd/console.go
@@ -1,6 +1,9 @@
 package cmd
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/exit"
 	"github.com/code-ready/crc/pkg/crc/machine"
@@ -27,20 +30,24 @@ var consoleCmd = &cobra.Command{
 	Short:   "Open the OpenShift Web Console in the default browser",
 	Long:    `Open the OpenShift Web Console in the default browser or print its URL or credentials`,
 	Run: func(cmd *cobra.Command, args []string) {
-		runConsole(args)
+		if err := runConsole(args); err != nil {
+			exit.WithMessage(1, err.Error())
+		}
 	},
 }
 
-func runConsole(arguments []string) {
+func runConsole(arguments []string) error {
 	consoleConfig := machine.ConsoleConfig{
 		Name: constants.DefaultName,
 	}
 
-	exitIfMachineMissing(consoleConfig.Name)
+	if err := checkIfMachineMissing(consoleConfig.Name); err != nil {
+		return err
+	}
 
 	result, err := machine.GetConsoleURL(consoleConfig)
 	if err != nil {
-		exit.WithoutMessage(1)
+		return err
 	}
 
 	if consolePrintURL {
@@ -51,15 +58,16 @@ func runConsole(arguments []string) {
 		output.Outf("To login as an admin, run 'oc login -u kubeadmin -p %s %s'\n", result.ClusterConfig.KubeAdminPass, result.ClusterConfig.ClusterAPI)
 	}
 	if consolePrintURL || consolePrintCredentials {
-		return
+		return nil
 	}
 
 	if !machine.IsRunning(result.State) {
-		exit.WithMessage(1, "The OpenShift cluster is not running, cannot open the OpenShift Web Console.")
+		return errors.New("The OpenShift cluster is not running, cannot open the OpenShift Web Console")
 	}
 	output.Outln("Opening the OpenShift Web Console in the default browser...")
 	err = browser.OpenURL(result.ClusterConfig.WebConsoleURL)
 	if err != nil {
-		exit.WithMessage(1, "Failed to open the OpenShift Web Console, you can access it by opening %s in your web browser.", result.ClusterConfig.WebConsoleURL)
+		return fmt.Errorf("Failed to open the OpenShift Web Console, you can access it by opening %s in your web browser", result.ClusterConfig.WebConsoleURL)
 	}
+	return nil
 }
--- a/cmd/crc/cmd/delete.go
+++ b/cmd/crc/cmd/delete.go
@@ -4,13 +4,12 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/spf13/cobra"
-
 	"github.com/code-ready/crc/pkg/crc/constants"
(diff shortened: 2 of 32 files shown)


cmd/crc/cmd/delete.go
```diff
@@ -24,31 +24,32 @@ var deleteCmd = &cobra.Command{
...
-	deleteConfig := machine.DeleteConfig{
-		Name: constants.DefaultName,
-	}
-
+func runDelete(arguments []string) error {
 	if clearCache {
 		deleteCache()
 	}
 
-	exitIfMachineMissing(deleteConfig.Name)
```
gbraad, line 42: we left the option for a name because some people want the multi instance option. at the moment we do not prevent people using a derived command that used the Name tto deal with this. 

in our own commands we do not use this.

cmd/crc/cmd/ip.go
```diff
@@ -18,21 +17,25 @@ var ipCmd = &cobra.Command{
...
-		runIP(args)
+		if err := runIP(args); err != nil {
+			errors.ExitWithMessage(1, err.Error())
+		}
 	},
 }
 
-func runIP(arguments []string) {
+func runIP(arguments []string) error {
 	ipConfig := machine.IpConfig{
```
gbraad, line 27: same here...

cmd/crc/cmd/status.go
```diff
@@ -41,14 +43,14 @@ type Status struct {
 	CacheDir        string
 }
 
-func runStatus() {
-	statusConfig := machine.ClusterStatusConfig{Name: constants.DefaultName}
```
gbraad, line 45: same here...

cmd/crc/cmd/podman_env.go
```diff
@@ -14,36 +10,17 @@ var podmanEnvCmd = &cobra.Command{
...
-		exitIfMachineMissing(ipConfig.Name)
-
-		result, err := machine.Ip(ipConfig)
-		if err != nil {
-			errors.Exit(1)
+		if err := runPodmanEnv(args); err != nil {
+			errors.ExitWithMessage(1, err.Error())
 		}
-
-		output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
```
gbraad, line 38: This is relevant and will be restored soon.

gbraad:
the use of the name in the structs has bee nkept since this means that people, who do use a multi instance setup (with different images/bundles) can still use the command as they only need to patch the name in the DefaultName location. this has been requested abd mentioned several times. Removing this is not suggested... refactored is OK. We allowed the input of JSON as would be on the pkg-level as input from the daemon.

The changes to podman are NOT suggested to be removed as this will be dealt with soon for Podman 2.0.

cmd/crc/cmd/delete.go
guillaumerose: OK, I will put it back.

cmd/crc/cmd/podman_env.go
guillaumerose: I will argue this is why we use git. When it will be ready, we can found it really quickly in the history and restore it.
Also, the linter doesn't like dead code and it refuses to build. Here, the error was hidden by in os.Exit before.

cmd/crc/cmd/podman_env.go
```diff
@@ -25,32 +19,6 @@ var podmanEnvCmd = &cobra.Command{
...
-		return err
-	}
-
-	output.Outln(shell.GetPathEnvString(userShell, constants.CrcBinDir))
-	output.Outln(shell.GetEnvString(userShell, "PODMAN_USER", constants.DefaultSSHUser))
-	output.Outln(shell.GetEnvString(userShell, "PODMAN_HOST", result.IP))
-	output.Outln(shell.GetEnvString(userShell, "PODMAN_IDENTITY_FILE", constants.GetPrivateKeyPath()))
-	output.Outln(shell.GetEnvString(userShell, "PODMAN_IGNORE_HOSTS", "1"))
-	output.Outln(shell.GenerateUsageHint(userShell, "crc podman-env"))
-	return nil
```
praveenkumar: @guillaumerose This is not going to dead in near future, we had this functionality and then because of issue with podman we just put `not supported`.

guillaumerose:
I added back podman-env code and removed the part about grouping machine commands under a struct. 
The diff is now smaller.

pkg/crc/network/utils.go
```diff
@@ -13,13 +13,13 @@ import (
 	crcos "github.com/code-ready/crc/pkg/os"
 )
 
-func executeCommandOrExit(sshRunner *ssh.SSHRunner, command string, errorMessage string) string {
+func executeCommandOrExit(sshRunner *ssh.SSHRunner, command string, errorMessage string) (string, error) {
```
cfergeau, line 16: This helper can probably be removed, and sshRunner.Run be called directly? At least it should be renamed.

cmd/crc/cmd/console.go
```diff
@@ -64,15 +70,16 @@ func runConsole(arguments []string) {
 		output.Outf("To login as an admin, run 'oc login -u kubeadmin -p %s %s'\n", result.ClusterConfig.KubeAdminPass, result.ClusterConfig.ClusterAPI)
 	}
 	if consolePrintURL || consolePrintCredentials {
-		return
+		return nil
 	}
 
 	if !machine.IsRunning(result.State) {
-		errors.ExitWithMessage(1, "The OpenShift cluster is not running, cannot open the OpenShift Web Console.")
+		return errors.New("The OpenShift cluster is not running, cannot open the OpenShift Web Console.")
```
cfergeau, line 77: Beware of `errors.New()` as it unconditionally prints the error message. For example, running `crc console` when crc is not running now gives
```
ERRO The OpenShift cluster is not running, cannot open the OpenShift Web Console. 
The OpenShift cluster is not running, cannot open the OpenShift Web Console.
```
when it used to print:
```
$ ./out/linux-amd64/crc console
The OpenShift cluster is not running, cannot open the OpenShift Web Console.
```

cmd/crc/cmd/delete.go
```diff
@@ -39,16 +41,19 @@ func runDelete(arguments []string) {
...
+	if err := exitIfMachineMissing(deleteConfig.Name); err != nil {
```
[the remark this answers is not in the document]
guillaumerose: yes, done

pkg/crc/errors/atexit.go
```diff
@@ -31,8 +31,8 @@ type exitHandlerFunc func(int) bool
 // exitHandlers keeps track of the list of registered exit handlers. Handlers are applied in the order defined in this list.
 var exitHandlers = []exitHandlerFunc{}
 
-// Exit runs all registered exit handlers and then exits the program with the specified exit code using os.Exit.
-func Exit(code int) {
+// exit runs all registered exit handlers and then exits the program with the specified exit code using os.exit.
+func exit(code int) {
```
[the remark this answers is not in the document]
guillaumerose: Previously we could do `Exit()` or `ExitWithMessage()`. Now, only `ExitWithMessage()`. I changed the commit message to be more explicit.

pkg/crc/network/utils.go
guillaumerose: yes done

cmd/crc/cmd/console.go
guillaumerose: outch! This was unexpected. I will fix it.

guillaumerose: I fixed it by:
* Moving ExitWithMessage to a new package so that we import less the error package and don't use it by mistake
* Use the custom errors package only where it's needed (eg in pkg/machine).

cmd/crc/cmd/podman_env.go
guillaumerose: Added back

praveenkumar:
https://github.com/code-ready/crc/pull/1398/commits/e1f6372aca11d11de79a025c004c0fd4a9622ab8 should be first one and then other 2 commits.

guillaumerose:
@praveenkumar I put the package move commit first. It's now good for review.

guillaumerose:
Rebased! PTAL :)

cmd/crc/cmd/console.go
```diff
@@ -51,15 +58,16 @@ func runConsole(arguments []string) {
 		output.Outf("To login as an admin, run 'oc login -u kubeadmin -p %s %s'\n", result.ClusterConfig.KubeAdminPass, result.ClusterConfig.ClusterAPI)
 	}
 	if consolePrintURL || consolePrintCredentials {
-		return
+		return nil
 	}
 
 	if !machine.IsRunning(result.State) {
-		errors.ExitWithMessage(1, "The OpenShift cluster is not running, cannot open the OpenShift Web Console.")
+		return errors.New("The OpenShift cluster is not running, cannot open the OpenShift Web Console")
```
praveenkumar, line 65: @guillaumerose since you removed this `dot` one of the integration test is failing, can you update the integration test for same?

guillaumerose: Thanks, I changed the test.

cmd/crc/cmd/root.go
```diff
@@ -81,14 +81,15 @@ func setConfigDefaults() {
...
-func exitIfMachineMissing(name string) {
+func checkIfMachineMissing(name string) error {
 	exists, err := machine.Exists(name)
 	if err != nil {
-		errors.ExitWithMessage(1, err.Error())
+		return err
 	}
 	if !exists {
-		errors.ExitWithMessage(1, "Machine '%s' does not exist. Use 'crc start' to create it.", name)
+		return fmt.Errorf("Machine '%s' does not exist. Use 'crc start' to create it", constants.DefaultName)
```
praveenkumar, line 90: @guillaumerose same here, in the CI it is failing with :( 

```
output did not match. Expected: 'Machine 'crc' does not exist. Use 'crc start' to create it.', Actual: 'Machine 'crc' does not exist. Use 'crc start' to create it'
```

guillaumerose: Fixed thanks

PR #1400, opened 2020-07-17
Fix cache.Executable() check

cfergeau:
This check is incorrect as it will return true for a file with
permissions 744 which is not owned by the current user. This switches
back to using unix.Access, which reverts parts of b8aba44ea and
b59dcf96f
unix.Access is not available on Windows, which is why the generic cache
implementation cannot be switched to it. An alternative to calling
unix.Access directly from the preflight files would be to use
unix.Access in cache_nonwin.go, and have a different implementation in
cache_windows.go, but I was not sure how to implement this on Windows.

**Fixes:** Issue #N

## Solution/Idea

## Proposed changes

1. Add `start` command.
3. ...

## Testing

7. ...

### DIFF
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -155,12 +155,3 @@ func IsTarball(filename string) bool {
 	}
 	return false
 }
-
-func (c *Cache) Executable() bool {
-	f := filepath.Join(c.destDir, c.binaryName)
-	fInfo, err := os.Stat(f)
-	if err != nil {
-		return false
-	}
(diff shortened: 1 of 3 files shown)


cfergeau:
Meant to cancel this, and it only dropped my edition of the template, but not the PR itself :-/
Quick note how to test this: check that removing executable permissions of `machine-driver-libvirt`/`machine-driver-hyperkit` is fixed by `crc setup`. The previous code was not doing what it was supposed to do, but this was hidden by some subsequent code in the preflight checks.

Issue #1227, opened 2020-05-08
Add JSON output to commands to support IDE integration


mohitsuman:
### Enhancement Request
~~To improve CRC integration with IDEs, we can leverage `crc daemon` command to be consumed directly from IDE.~~

- ~~Currently the daemon is started along with the system tray, there needs to be a way to start the daemon without loading the tray.~~

----

See #13. We will add support to allow the basic commands `status`, `start`, `stop` and `version` to output JSON for use by the IDE integration.

praveenkumar:
@mohitsuman `crc daemon` already exist, what did you try and doesn't work or what more you want as part of daemon, it will help us to prioritize this issue. Do update it once you have more details.

gbraad:
@anjannath provided some code to be able to talk to the daemon:

```
const net = require('net');
const homedir = require('os').homedir();
const path = require('path');

const SOCKET_PATH = path.join(homedir, ".crc", "crc.sock");

function sendCommand(command, args) {
    if (command === "") {
        console.log("Command is required!!")
        return
    }

    sock = net.createConnection(SOCKET_PATH, () => {
        sock.on('error', (e) => {
            console.error(e)
        })

        cmd = {
            "command": command,
            "args": args
        }

        console.log(JSON.stringify(cmd))

        sock.write(JSON.stringify(cmd));

        // receive json response
        sock.on('data', (d) => {
            console.log(d.toString())
        })

        sock.on('error', (e) => {
            console.log(e)
        })

    })
}

sendCommand("version")
```

However, we need to plan a follow-up meeting for this. Originally we did not intent to run a daemon process on Linux, but this might be needed?

dgolovin:
This specific feature we can implement as a vscode status line extension. It would look exactly the same as implementation for system tray on windows. 
Thank you for example @gbraad. I like that is is possible to detect if demon is running. That would let to initialize status line element after vscode is started.

If linux and macos implementation could just provide the same functionality except system tray. That would let vscode to provide this feature on all supported platforms: win, mac and linux.

gbraad:
Windows woukd likely have to use named pipes
See #1384 for more info and related PR

gbraad:
The daemon at the moment only exists for Windows and macOS.
You would prefer to run this without systray?

gbraad:
Modified the issue to relfect the discussion we had. We will add JSON putput to the commands as origjnally proposed by #13.

cfergeau:
[...]

PR #1402, opened 2020-07-20
Add json output format for crc status

guillaumerose:
Partial fix for #1227

```
$ crc status -o json               
{
  "virtualMachineStatus": "Running",
  "openshiftStatus": "Running",
  "openshiftVersion": "4.5.1",
  "diskUsage": 12852875264,
  "diskSize": 32720793600,
  "cacheUsage": 9769732330,
  "cacheDir": "/home/guillaumerose/.crc/cache"
}

$ crc status        
CRC VM:          Running
OpenShift:       Running (v4.5.1)
Disk Usage:      12.86GB of 32.72GB (Inside the CRC VM)
Cache Usage:     9.77GB
Cache Directory: /home/guillaumerose/.crc/cache

$ crc status -o wide
CRC VM:          Running
OpenShift:       Running (v4.5.1)
Disk Usage:      12.86GB of 32.72GB (Inside the CRC VM)
Cache Usage:     9.77GB
Cache Directory: /home/guillaumerose/.crc/cache

```

### DIFF
--- a/cmd/crc/cmd/client.go
+++ b/cmd/crc/cmd/client.go
@@ -1,0 +1,18 @@
+package cmd
+
+import "github.com/code-ready/crc/pkg/crc/machine"
+
+type client interface {
+	Status(statusConfig machine.ClusterStatusConfig) (machine.ClusterStatusResult, error)
+	Exists(name string) (bool, error)
+}
+
+type libmachineClient struct{}
+
+func (*libmachineClient) Status(statusConfig machine.ClusterStatusConfig) (machine.ClusterStatusResult, error) {
+	return machine.Status(statusConfig)
+}
+
+func (*libmachineClient) Exists(name string) (bool, error) {
+	return machine.Exists(name)
+}
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -22,7 +22,7 @@ var statusCmd = &cobra.Command{
 	Short: "Display status of the OpenShift cluster",
 	Long:  "Show details about the OpenShift cluster",
 	Run: func(cmd *cobra.Command, args []string) {
-		if err := runStatus(args); err != nil {
+		if err := runStatus(); err != nil {
 			exit.WithMessage(1, err.Error())
 		}
 	},
@@ -43,7 +43,7 @@ type Status struct {
 	CacheDir        string
 }
 
-func runStatus(args []string) error {
+func runStatus() error {
 	statusConfig := machine.ClusterStatusConfig{Name: constants.DefaultName}
 
 	if err := checkIfMachineMissing(statusConfig.Name); err != nil {
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -6,7 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"text/template"
+	"text/tabwriter"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/exit"
@@ -36,13 +36,6 @@ var statusCmd = &cobra.Command{
 		}
 	},
 }
-
-var statusFormat = `CRC VM:          {{.CrcStatus}}
-OpenShift:       {{.OpenShiftStatus}}
-Disk Usage:      {{.DiskUsage}}
-Cache Usage:     {{.CacheUsage}}
-Cache Directory: {{.CacheDir}}
-`
 
 type Status struct {
 	CrcStatus       string `json:"crcStatus"`
@@ -90,17 +83,31 @@ func runStatus(writer io.Writer, client client, cacheDir, outputFormat string) e
 		encoder.SetIndent("", "  ")
 		return encoder.Encode(status)
 	}
-	return printStatus(writer, status, statusFormat)
+	return printStatus(writer, status)
 }
 
-func printStatus(writer io.Writer, status interface{}, statusFormat string) error {
-	tmpl, err := template.New("status").Parse(statusFormat)
-	if err != nil {
-		return fmt.Errorf("Error creating status template: %s", err.Error())
+func printStatus(writer io.Writer, status Status) error {
+	w := tabwriter.NewWriter(writer, 0, 0, 1, ' ', 0)
+	lines := []struct {
+		left, right string
+	}{
+		{"CRC VM", status.CrcStatus},
+		{"OpenShift", status.OpenShiftStatus},
+		{"Disk Usage", status.DiskUsage},
+		{"Cache Usage", status.CacheUsage},
+		{"Cache Directory", status.CacheDir},
 	}
-	err = tmpl.Execute(writer, status)
-	if err != nil {
-		return fmt.Errorf("Error executing status template: %s", err.Error())
+	for _, line := range lines {
+		if err := printLine(w, line.left, line.right); err != nil {
+			return err
+		}
+	}
+	return w.Flush()
+}
+
+func printLine(w *tabwriter.Writer, left string, right string) error {
+	if _, err := fmt.Fprintf(w, "%s:\t%s\n", left, right); err != nil {
+		return err
 	}
 	return nil
 }
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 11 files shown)


cmd/crc/cmd/status.go
```diff
@@ -34,23 +45,24 @@ Cache Directory: {{.CacheDir}}
 `
 
 type Status struct {
-	CrcStatus       string
-	OpenShiftStatus string
-	DiskUsage       string
-	CacheUsage      string
-	CacheDir        string
+	CrcStatus       string `json:"crcStatus"`
```
gbraad, line 48: We do not care much about the actual markup, but rather about providing the needed information. This would be more important is we would ever rename something.

cmd/crc/cmd/status.go
```diff
@@ -59,23 +71,30 @@ func runStatus() {
...
+		return fmt.Errorf("Error finding size of cache: %s", err.Error())
 	}
 	cacheUsage := units.HumanSize(float64(size))
 	diskUse := units.HumanSize(float64(clusterStatus.DiskUse))
 	diskSize := units.HumanSize(float64(clusterStatus.DiskSize))
 	diskUsage := fmt.Sprintf("%s of %s (Inside the CRC VM)", diskUse, diskSize)
 	status := Status{clusterStatus.CrcStatus, clusterStatus.OpenshiftStatus, diskUsage, cacheUsage, cacheDir}
-	printStatus(status, statusFormat)
+
+	if outputFormat == jsonFormat {
```
gbraad, line 82: this is more generic, so the writer is better suited to be created in a generic way as ALL commands will use this.

guillaumerose: I prefer to start small and don't over-engineer it in this PR. But for sure, when we will have 2 commands with `-o json` we will need to do something.

cmd/crc/cmd/status.go
guillaumerose: Markup is better if it's consistent with other products. OpenShift and Kubernetes use lower camel case. CRC should do the same.

cmd/crc/cmd/status.go
```diff
@@ -90,17 +83,31 @@ func runStatus(writer io.Writer, client client, cacheDir, outputFormat string) e
...
+func printStatus(writer io.Writer, status Status) error {
+	w := tabwriter.NewWriter(writer, 0, 0, 1, ' ', 0)
+	lines := []struct {
+		left, right string
+	}{
+		{"CRC VM", status.CrcStatus},
+		{"OpenShift", status.OpenShiftStatus},
+		{"Disk Usage", status.DiskUsage},
+		{"Cache Usage", status.CacheUsage},
+		{"Cache Directory", status.CacheDir},
```
praveenkumar, line 102: I would also like some more details about this change, like what is the benefit of using tabwritter over templates.

cmd/crc/cmd/status.go
```diff
@@ -38,10 +38,11 @@ var statusCmd = &cobra.Command{
 }
 
 type Status struct {
-	CrcStatus       string `json:"crcStatus"`
+	CrcStatus       string `json:"virtualMachineStatus"`
```
praveenkumar: This change don't belong to this commit.

cmd/crc/cmd/status.go
```diff
@@ -68,12 +69,13 @@ func runStatus(writer io.Writer, client client, cacheDir, outputFormat string) e
...
-		DiskSize:        clusterStatus.DiskSize,
-		CacheUsage:      size,
-		CacheDir:        cacheDir,
+		CrcStatus:        clusterStatus.CrcStatus,
+		OpenShiftStatus:  clusterStatus.OpenshiftStatus,
+		OpenShiftVersion: clusterStatus.OpenshiftVersion,
+		DiskUsage:        clusterStatus.DiskUse,
+		DiskSize:         clusterStatus.DiskSize,
+		CacheUsage:       size,
+		CacheDir:         cacheDir,
```
praveenkumar, line 78: This is again public facing UX change so I would like to have discussion about it before merge this one.

cmd/crc/cmd/status.go
guillaumerose: In the following commit, I change the struct to remove strings and have raw numbers for usage. It was easier to build the original template by hand than with template.

cmd/crc/cmd/status.go
guillaumerose: The status message didn't change. With json output, for sure, now we cannot change it like this.

But for now, I don't think someone is relying on this struct ? Does someone vendor crc ?

cmd/crc/cmd/status.go
guillaumerose: yes you're right. I was not sure of this change. WDYT of virtualMachineStatus ?

cmd/crc/cmd/status.go
praveenkumar: > Does someone vendor crc ?

@guillaumerose not that I am aware of.

cmd/crc/cmd/status.go
praveenkumar: @guillaumerose Can you put that in the commit log the benefit it is going to provide moving forward or right now.

guillaumerose: I added the why in the commit log

cmd/crc/cmd/status.go
anjannath: The trays depend on the struct, so will need changes there

guillaumerose: Yes, looking at the macOS tray we use CrcStatus, OpenshiftStatus, DiskUse and DiskSize :
```
➜  tray-macos git:(master) rg "status\."
CodeReady Containers/DetailedStatusViewController.swift
44:                    if status.Success {
45:                        self.vmStatus.stringValue = status.CrcStatus
46:                        self.ocpStatus.stringValue = status.OpenshiftStatus
47:                        self.diskUsage.stringValue = "\(Units(bytes: status.DiskUse).getReadableUnit()) of \(Units(bytes: status.DiskSize).getReadableUnit()) (Inside the VM)"
```

This PR (apart from breaking the json due the case) removes the version from OpenshiftStatus. 

What should we do ? The tray is still experimental so we can edit in both and release at the same time. WDYT of the case of the json ? I took the same convention as kubernetes/openshift.

guillaumerose: In fact this struct is not used by the tray. This struct is only use in cmd package. Handler use ClusterStatusResult. So we are good.

cmd/crc/cmd/client.go
```diff
@@ -0,0 +1,18 @@
...
+type client interface {
+	Status(statusConfig machine.ClusterStatusConfig) (machine.ClusterStatusResult, error)
+	Exists(name string) (bool, error)
+}
+
+type libmachineClient struct{}
+
+func (*libmachineClient) Status(statusConfig machine.ClusterStatusConfig) (machine.ClusterStatusResult, error) {
+	return machine.Status(statusConfig)
+}
```
praveenkumar, line 14: @guillaumerose this new `client` for the cmd is only for unit test purpose so that it can be mocked and also extended for other commands ( like start,stop ..etc.) ?

guillaumerose: Yes !

cmd/crc/cmd/status.go
anjannath: Yeah, sorry.. i thought i commented on the `ClusterStatusResult` structure changes, we can change in both, that's no issue. and i am fine with the case of the json, although i find Upper camel case more good to look at..

cmd/crc/cmd/status.go
anjannath: i'd prefer it to be called `crcStatus` only, as this also relates to the struct member name, if that makes sense..

cmd/crc/cmd/status.go
```diff
@@ -14,7 +15,14 @@ import (
...
 )
 
+const jsonFormat = "json"
+
+var (
+	outputFormat string
+)
+
 func init() {
+	statusCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "Output format. One of: json|wide")
```
anjannath, line 25: The `wide` format can be removed, as its not been implemented.

anjannath:
Looks good to me 👍

cmd/crc/cmd/status.go
guillaumerose: Yes good point. What could be the name of the default format ? For kubectl, it's wide if I'm correct.

cmd/crc/cmd/status.go
guillaumerose: I put back `crcStatus`

cmd/crc/cmd/status.go
anjannath: `kubectl` wide gives more information, then without a `-o` flag which is the default. i don't think we need to provide the default also as option for the output flag.

guillaumerose: I removed wide and just let json. Is it better? `Output format. One of: json` ?

PR #1403, opened 2020-07-20
Remove license headers left over

guillaumerose:


cmd/crc/cmd/console.go
```diff
@@ -1,16 +1,3 @@
-/*
-Copyright 2016 The Kubernetes Authors All rights reserved.
```
gbraad, line 2: Not sure why this mentions kubernetes?!? C&P issue?

guillaumerose:
I added the missing headers and a CI step to verify we don't miss one.

praveenkumar:
@gbraad @guillaumerose can't we remove the license headers from all the files since we already have a license file or these headers means something different?

PR #1404, opened 2020-07-20
Add unit test for api

guillaumerose:
This adds a very basic unit test for the API. I plan to improve it by using a machine that can be mocked so that we can check json errors.


pkg/crc/api/api.go
```diff
@@ -15,6 +15,10 @@ func CreateApiServer(socketPath string) (CrcApiServer, error) {
 		logging.Error("Failed to create socket: ", err.Error())
 		return CrcApiServer{}, err
 	}
+	return createApiServerWithListener(listener)
+}
+
+func createApiServerWithListener(listener net.Listener) (CrcApiServer, error) {
```
praveenkumar, line 21: :+1:

anjannath:
Awesome, thanks!!

PR #1405, opened 2020-07-20
Rename log-level param to loglevel

guillaumerose:
Automatically convert numerical level to string level so that `--loglevel=0` is equivalent to info and `--loglevel=1` and more to debug.

This PR makes `crc` command behave like `oc`.

There is a breaking change in the tray setup. The command line changed and I'm not sure if it will be migrated automatically. It's still an experimental feature so I guess it's ok.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -21,7 +21,7 @@ assignees: ''
   
 ## CRC status
 ```bash
-# Put `crc status --log-level debug` output here
+# Put `crc status --loglevel debug` output here
 ```
 
 ## CRC config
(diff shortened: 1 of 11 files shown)


praveenkumar:
@guillaumerose this is not breaking change for tray but also it is public facing command argument change, in the past we already get a backfire when did something like this, I will hold this until we discuss this change and also send a mail about this change.

/hold

gbraad:
> I will hold this until we discuss this change and also send a mail about this change.

renames to the CLI can not happen... aliases should be added in that case, as this impacts documentation, issue template, etc if this is actually changing from `--log-level` to `--loglevel`.

guillaumerose:
I will add an alias then. How long should we keep both ?

gbraad:
> I will add an alias then. How long should we keep both ?

I do not feel very strong about the removal. It depends on the docs, issue template that gets updated and a deprecation note?

but it would be more likely that this might become the standard command, but the older `log-level` remains a permanent alias.

guillaumerose:
I propose:
* Drop this PR
* Wait for the merge of machine inside this repo
* Remove logrus and use the same log system as oc, odo, etc. (glog if I'm correct). It will force us to use numeric loglevel.

praveenkumar:
> I propose:

@guillaumerose works for me, so closing this PR as of now.

/close

Issue #1401, opened 2020-07-19
[BUG] ERRO Error creating host: Error creating the VM: Error creating machine: Error in driver during machine creation: exit status 1


jpdade:
### General information

  * OS: Windows 10 Pro
  * Hypervisor: Hyper-V 
  * Did you run `crc setup` before starting it (Yes)?
  * Running CRC on: Laptop 

## CRC version
```bash
# Put `crc version` output here
```

```
  CodeReady Containers version: 1.13.0+8070bae
OpenShift version: 4.5.1 (embedded in binary)
```

## CRC status
```bash
# Put `crc status --log-level debug` output here
```

```
Machine 'crc' does not exist. Use 'crc start' to create it.
```

## CRC config
```bash
# Put `crc config view` output here
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
# put the output of `sw_vers` in case of Mac
# Put the output of `systeminfo` in case of Windows
```

```
C:\Users\HP\Downloads\crc-windows-amd64\crc-windows-1.13.0-amd64>systeminfo

Host Name:                 LAPTOP-CODEREAD
OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.18363 N/A Build 18363
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Standalone Workstation
OS Build Type:             Multiprocessor Free
Registered Owner:          Windows User
Registered Organization:
Product ID:                00330-80000-00000-AA357
Original Install Date:     7/9/2020, 7:25:29 PM
System Boot Time:          7/19/2020, 12:08:27 PM
System Manufacturer:       HP
System Model:              HP Laptop 15-db0xxx
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: AMD64 Family 21 Model 112 Stepping 0 AuthenticAMD ~1300 Mhz
BIOS Version:              Insyde F.28, 5/8/2020
Windows Directory:         C:\Windows
System Directory:          C:\Windows\system32
Boot Device:               \Device\HarddiskVolume1
System Locale:             en-us;English (United States)
Input Locale:              en-us;English (United States)
Time Zone:                 (UTC-05:00) Eastern Time (US & Canada)
Total Physical Memory:     7,646 MB
Available Physical Memory: 3,963 MB
Virtual Memory: Max Size:  8,862 MB
Virtual Memory: Available: 4,299 MB
Virtual Memory: In Use:    4,563 MB
[...]

PR #1407, opened 2020-07-21
Ensure the host has at least 9GB of RAM

guillaumerose:
Helpful for https://github.com/code-ready/crc/issues/1401

The implementation of `memory.TotalMemory()` seems good. It uses the same mechanism for Windows as Kubelet.

https://github.com/pbnjay/memory/blob/master/memory_windows.go#L26
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/winstats/perfcounter_nodestats.go#L232

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -10,9 +10,11 @@ import (
 	"github.com/code-ready/crc/pkg/crc/cache"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/validation"
 	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/code-ready/crc/pkg/embed"
 	crcos "github.com/code-ready/crc/pkg/os"
+	"github.com/docker/go-units"
 )
 
 var genericPreflightChecks = [...]PreflightCheck{
@@ -42,6 +44,15 @@ var genericPreflightChecks = [...]PreflightCheck{
 		fixDescription:   "Unpacking bundle from the CRC binary",
 		fix:              fixBundleCached,
 		flags:            SetupOnly,
+	},
+	{
+		configKeySuffix:  "check-ram",
+		checkDescription: "Checking minimum RAM requirements",
+		check: func() error {
+			return validation.ValidateEnoughMemory(constants.DefaultMemory)
+		},
+		fixDescription: fmt.Sprintf("crc requires at least %s to run", units.HumanSize(float64(constants.DefaultMemory*1024*1024))),
+		flags:          NoFix,
 	},
 }
 
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -10,7 +10,10 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/errors"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/version"
+	"github.com/docker/go-units"
+	"github.com/pbnjay/memory"
 )
 
 // ValidateCPUs checks if provided cpus count is valid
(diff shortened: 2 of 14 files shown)


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -43,6 +45,21 @@ var genericPreflightChecks = [...]PreflightCheck{
...
+		checkDescription: "Checking minimum RAM requirements",
+		check:            checkRAM,
+		fixDescription:   "crc requires at least 9GB to run",
+		flags:            NoFix,
+	},
+}
+
+func checkRAM() error {
+	totalMemory := memory.TotalMemory()
+	if totalMemory < 9_000_000_000 {
```
jerolimov, line 59: Can this check compare totalMemory against `crcConfig.GetInt(config.Memory.Name)` because it is possible to increase the memory setting?

https://github.com/code-ready/crc/blob/a3bd657a7c04292f59454e7dbfb62f54ac367b6d/cmd/crc/cmd/start.go#L55

https://github.com/code-ready/crc/blob/master/docs/source/topics/proc_configuring-the-virtual-machine.adoc

guillaumerose: Do you mean something like `totalMemory < max(9GB, crcConfig.GetInt(config.Memory.Name))` ? I think we really want to prevent user from running crc on too little machine (docs says `The amount of memory to assign must be greater than or equal to the default.`).

jerolimov: Yes. It was just an idea to ensure if someone sets 16 GB we check totalMemory against this configuration.

You are right, setting a configuration with less than 9 GB (default) is already not possible:

```
➜  ~ crc config set memory 1024
ERRO requires memory in MiB >= 9216               
Value '1024' for configuration property 'memory' is invalid, reason: requires memory in MiB >= 9216
```

praveenkumar, line 59: @guillaumerose we already have a constant which consist of default memory so shouldn't we use something like

```patch
 func checkRAM() error {
        totalMemory := memory.TotalMemory()
-       if totalMemory < 9_000_000_000 {
+       logging.Debugf("Total memory of system is %d Bytes", totalMemory)
+       if totalMemory < constants.DefaultMemory*1024*1024 {
                return fmt.Errorf("9GB of memory are required to run crc (%s found)", units.HumanSize(float64(totalMemory)))
        }
        return nil
```

Also we should move it so some other package and make it public so even with `crc config set memory <>` we can use this as part of validation https://github.com/code-ready/crc/blob/master/pkg/crc/validation/validation.go#L25

gbraad: If you have TOO little memory, Hyper-V will be prevented to start. And this NOT related to total memory, but free memory.

gbraad: While I agree this is already better, it is actually the amount of free memory that prevents Hyper-V to start a VM.

guillaumerose: @praveenkumar nice! I will change this

guillaumerose:
It now checks total memory at `setup`, `start` and `config set memory` time.

pkg/crc/validation/validation.go
```diff
@@ -26,6 +29,19 @@ func ValidateMemory(value int) error {
...
+// ValidateEnoughMemory checks if enough memory is installed on the host
+func ValidateEnoughMemory(value int) error {
+	totalMemory := memory.TotalMemory()
+	logging.Debugf("Total memory of system is %d bytes", totalMemory)
+	valueBytes := value * 1024 * 1024
+	if totalMemory < uint64(valueBytes) {
+		return fmt.Errorf("only %s of memory found (%s required)",
+			units.HumanSize(float64(totalMemory)),
+			units.HumanSize(float64(valueBytes)))
+	}
```
praveenkumar, line 44: :+1:

marcelkottmann:
Someone may update the documentation on https://access.redhat.com/documentation/en-us/red_hat_codeready_containers/1.0/html/getting_started_guide/getting-started-with-codeready-containers_gsg

> 1.3.1. Hardware requirements
> CodeReady Containers requires the following system resources:
> 
> 4 virtual CPUs (vCPUs)
> 8 GB of memory
> 35 GB of storage space

I get the following error: 

> crc requires at least 9.664GB to run

cfergeau:
This is the documentation for crc 1.0, the documentation for crc 1.17 is https://access.redhat.com/documentation/en-us/red_hat_codeready_containers/1.17/html/getting_started_guide/installation_gsg#minimum-system-requirements_gsg

Issue #1386, opened 2020-07-14
Use oc runner from the VM to execute all the oc commands 


praveenkumar:
Right now we are executing the oc commands from the host for different purpose (like cluster ID, pull secret, approving csr ...etc.) this will always make initial crc start depend on the dns configuration of the host which sometime doesn't work as expected. We should try to convert all those oc commands which are depend on the host dns to the oc ssh runner and execute it.

praveenkumar:
fixed by #1408

PR #1408, opened 2020-07-21
Issue #1386 Run all oc commands from the VM

praveenkumar:
Currently running oc commands from the host need dns to be working,
and something it doesn't work as expected, running all oc commands
from the VM to make sure that atleast initial startup doesn't depend
on the host side.

### DIFF
--- a/pkg/crc/oc/oc.go
+++ b/pkg/crc/oc/oc.go
@@ -1,9 +1,12 @@
 package oc
 
 import (
+	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/ssh"
 	crcos "github.com/code-ready/crc/pkg/os"
 )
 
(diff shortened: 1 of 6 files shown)


praveenkumar:
Need to update the tests.
/hold

pkg/crc/oc/oc.go
```diff
@@ -82,3 +85,39 @@ func (oc OcConfig) RunOcCommandPrivate(args ...string) (string, string, error) {
...
 	return oc.Runner.RunPrivate(args...)
 }
+
+type OcVMConfig struct {
+	Runner  *ssh.SSHRunner
+	Context string
+	Cluster string
+}
+
+func UseOCWithVMConfig(sshRunner *ssh.SSHRunner) OcVMConfig {
```
guillaumerose, line 95: You can avoid to define this new type by creating a new Runner (and maybe remove use of stderr?).

``` 
type OcSSHRunner struct {
	Runner *ssh.SSHRunner
}

func (r OcSSHRunner) Run(args ...string) (string, string, error) {
	command := fmt.Sprintf("r --kubeconfig /opt/kubeconfig %s", strings.Join(args, " "))
	stdout, err := r.Runner.Run(command)
	return stdout, "", err
}

func (r OcSSHRunner) RunPrivate(args ...string) (string, string, error) {
	command := fmt.Sprintf("r --kubeconfig /opt/kubeconfig %s", strings.Join(args, " "))
	stdout, err := r.Runner.RunPrivate(command)
	return stdout, "", err
}

func UseOCWithSSH(ssh *ssh.SSHRunner) OcConfig {
	return OcConfig{
		Runner: OcSSHRunner{
			Runner: ssh,
		},
		Context: constants.DefaultContext,
		Cluster: constants.DefaultName,
	}
}
```

guillaumerose:
What if `oc` prints something both on stdout and stderr and we need to parse it as json ? Maybe it never happens.

PR #1410, opened 2020-07-21
Enable more linters esp. golint

guillaumerose:
Activate common linters. 
golint triggered lot of small changes. I find the code to be a little bit better now. Function names reads better.

### DIFF
--- a/pkg/crc/machine/hyperkit/driver_darwin.go
+++ b/pkg/crc/machine/hyperkit/driver_darwin.go
@@ -8,7 +8,7 @@ import (
 	"github.com/code-ready/machine/libmachine/drivers"
 )
 
-type hyperkitDriver struct {
+type Driver struct {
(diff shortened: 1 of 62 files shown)


pkg/crc/machine/hyperkit/driver_darwin.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/code-ready/machine/libmachine/drivers"
 )
 
-type hyperkitDriver struct {
+type Driver struct {
```
praveenkumar, line 11: @guillaumerose Does golint also complain about type names like `libvirtDriver` or `hyperkitDriver` ?

guillaumerose: Yes because the package name is the prefix of the struct name.

PR #1415, opened 2020-07-23
Update proc_starting-monitoring-alerting-telemetry.adoc

sradco:
Using 12GB was insufficient and resulted in insufficient memory errors.

## Solution/Idea
Increase recommended memory size

## Proposed changes

Update recommended memory size to 20 GB

## Testing

1.Install crc with monitoring, alerting and telemetry enabled with 12GB
Compared to 20GB.
2. Check for insufficient memory errors in the UI.

### DIFF
--- a/docs/source/topics/proc_starting-monitoring-alerting-telemetry.adoc
+++ b/docs/source/topics/proc_starting-monitoring-alerting-telemetry.adoc
@@ -11,7 +11,8 @@ Telemetry functionality is responsible for listing your cluster in the link:http
 For more information, see link:{crc-gsg-url}#accessing-the-openshift-cluster-with-oc_gsg[Accessing the OpenShift cluster with `oc`].
 * You must log in through [command]`oc` as the `kubeadmin` user.
 * You must assign additional memory to the {prod} virtual machine.
-At least 20 GiB of memory (a value of `20480`) is recommended.
+At least 14 GiB of memory (a value of `14336`) is recommended, for core functionality. 
+For additional workload additional memory will be required.
 For more information, see link:{crc-gsg-url}#configuring-the-virtual-machine_gsg[Configuring the virtual machine].
 
 .Procedure


docs/source/topics/proc_starting-monitoring-alerting-telemetry.adoc
```diff
@@ -11,7 +11,7 @@ Telemetry functionality is responsible for listing your cluster in the link:http
 For more information, see link:{crc-gsg-url}#accessing-the-openshift-cluster-with-oc_gsg[Accessing the OpenShift cluster with `oc`].
 * You must log in through [command]`oc` as the `kubeadmin` user.
 * You must assign additional memory to the {prod} virtual machine.
-At least 12 GiB of memory (a value of `12288`) is recommended.
```
praveenkumar, line 14: @sradco Are you sure putting this much memory needed just for enable monitoring and telemetry because we did test it with ~14GB without any issue.

docs/source/topics/proc_starting-monitoring-alerting-telemetry.adoc
```diff
@@ -11,7 +11,7 @@ Telemetry functionality is responsible for listing your cluster in the link:http
...
+At least 20 GiB of memory (a value of `20480`) is recommended.
```
robin-owen, line 14: @sradco 20 GiB of memory seems excessive to me. I tested this scenario recently on version 1.13 while allocating 12 GiB to the virtual machine. When no other Operators are required, this works -- and we note elsewhere that enabling more Operators will require more resources.

robin-owen:
@sradco Do you mind addressing the comments that we've left on this PR? As far as I can tell, the current documentation is accurate, but if you're having a different experience, we'd love to hear more about it. :)

sradco:
> @sradco Do you mind addressing the comments that we've left on this PR? As far as I can tell, the current documentation is accurate, but if you're having a different experience, we'd love to hear more about it. :)

@robin-owen Sorry for the late reply. I used a strong machine for the testing. I enabled the monitoring and telemetry and got insufficient memory errors in the UI. I did not try it with 15GB, it might be enough.

robin-owen:
@sradco No worries, thanks for the reply!

This is interesting to me, though -- this is a classic example of "I can't reproduce this on my machine." I also know that the developers are trying to keep the minimum system requirements below 16 GiB if possible, as this is all of the memory available to a standard Macbook Pro.

Based on @praveenkumar's answer, it seems we're testing with 14 GiB. If that's the amount that we're testing this functionality with, perhaps that should be our recommendation in the docs?

praveenkumar:
@robin-owen https://github.com/code-ready/snc/blob/master/snc.sh#L356 is what we set when create the bundle so this should be the minimum in our doc (which is 14GB), Also we need to remember that this memory is required when there is no other deployment only core stuff is up and running so if user need to put workload then again need to increase the memory according to the requirement.

docs/source/topics/proc_starting-monitoring-alerting-telemetry.adoc
sradco: Updated to 14GB.

sradco:
@praveenkumar  Updated to 14GB

PR #1417, opened 2020-07-24
Use helper function to ensure errors are consistent

guillaumerose:
In machine pkg, instead of doing
```
result.Error = err.Error()
return *result, errors.Newf("Failed to get pull secret: %v", err)
``` 
do
```
return startError(startConfig.Name, "Cannot initialize libmachine", err)
``` 

It highlights some errors without a description. Some result were not valid (Error not set, etc.).

Also extract the libmachine creation so that callers can set logging verbosity if they want.

### DIFF
--- a/pkg/crc/machine/errors.go
+++ b/pkg/crc/machine/errors.go
@@ -9,3 +9,56 @@ func startError(name string, description string, err error) (StartResult, error)
 		Error: fullErr.Error(),
 	}, fullErr
 }
+
+func stopError(name string, description string, err error) (StopResult, error) {
+	fullErr := errors.Newf("%s: %v", description, err)
+	return StopResult{
+		Name:    name,
+		Success: false,
+		Error:   fullErr.Error(),
+	}, fullErr
+}
+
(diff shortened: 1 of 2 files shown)


gbraad:
Much preferred as it allows override and logging from a central place

praveenkumar:
@guillaumerose Can we have a different file which contain all the error function like `startError` `statusError` etc. in the same package, current approach is just making this `machine.go` file bit longer.

guillaumerose:
I moved the error functions in a new file. Thanks @praveenkumar for the suggestion!

pkg/crc/machine/errors.go
```diff
@@ -0,0 +1,64 @@
+package machine
+
+import "github.com/code-ready/crc/pkg/crc/errors"
+
+func startError(name string, description string, err error) (StartResult, error) {
```
cfergeau, line 5: Yup, very nice to have the error creation centralized somewhere as this was indeed not always fully consistent. One minor nit as always, I'm wondering if `reportStartError` or such would have been a better name.

PR #1421, opened 2020-07-28
Update godog dependency

praveenkumar:


test/integration/crcsuite/crcsuite.go
```diff
@@ -5,6 +5,7 @@ package crcsuite
 import (
 	"crypto/tls"
 	"fmt"
+	"github.com/cucumber/messages-go/v10"
```
jsliacan, line 8: Maybe we should move this to what's now line 20 (the other paragraph of imports).

jsliacan:
One more thing though... We'll probably have to make analogous changes to Clicumber

PR #1422, opened 2020-07-28
Remove custom build tag

guillaumerose:
It seems that the build tag `build` can be removed now. I guess a previous PR simplified the code.

### DIFF
--- a/pkg/crc/machine/hyperkit/constants.go
+++ b/pkg/crc/machine/hyperkit/constants.go
@@ -1,5 +1,3 @@
-//+build darwin build
-
 package hyperkit
 
 import "fmt"
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ GOPATH ?= $(shell go env GOPATH)
 ORG := github.com/code-ready
 REPOPATH ?= $(ORG)/crc
 
-PACKAGES := go list --tags build ./...
+PACKAGES := go list ./...
 SOURCES := $(shell git ls-files  *.go ":^vendor")
 
 RELEASE_INFO := release-info.json
@@ -89,14 +89,14 @@ $(BUILD_DIR)/windows-amd64/crc.exe: $(SOURCES)
 	GOARCH=amd64 GOOS=windows go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/windows-amd64/crc.exe ./cmd/crc
 
 $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
-	go build --tags="build" -ldflags="$(LDFLAGS)" -o $(HOST_BUILD_DIR)/crc-embedder ./cmd/crc-embedder
+	go build -ldflags="$(LDFLAGS)" -o $(HOST_BUILD_DIR)/crc-embedder ./cmd/crc-embedder
 
 .PHONY: cross ## Cross compiles all binaries
 cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/windows-amd64/crc.exe
 
 .PHONY: test
 test:
-	go test --tags build -v -ldflags="$(VERSION_VARIABLES)" $(shell $(PACKAGES))
+	go test -v -ldflags="$(VERSION_VARIABLES)" $(shell $(PACKAGES))
 
 .PHONY: build_docs
 build_docs:
--- a/pkg/crc/machine/libvirt/constants.go
+++ b/pkg/crc/machine/libvirt/constants.go
@@ -1,5 +1,3 @@
-//+build linux build
-
 package libvirt
 
 import "fmt"


pkg/crc/machine/hyperkit/constants.go
```diff
@@ -1,5 +1,3 @@
-//+build darwin build
```
praveenkumar, line 1: @guillaumerose I think we still need `+build darwin` otherwise this will be included for all the platform right?

guillaumerose: Yes but linux and windows binaries will not use them and the compiler should not include them. In this case, if someone wants to use it, it will have to specify the package `hyperkit.FOO`, so it's very unlikely it will be used by mistake.

With this branch: 
``` 
$ strings crc | grep "code-ready/machine-driver-hyperkit"
nothing!
$ strings crc-embedder | grep "code-ready/machine-driver-hyperkit"
found it!
```

I like to avoid as much as possible tags and _goos file so that when I develop or compile for my platform, I can see all errors/compilation issues. They are not hidden until I do a `make cross`.
For instance, `util_windows.go` which contains `ReadFileUTF16LE` with nothing platform specific inside could be merged with `util.go`.

Also, if we do that, when you will see a _windows.go file, you will know there is something really specific inside.

WDYT?

praveenkumar: I agree but think about if I am putting something for libvirt/hyperkit and this build tag is not there then by mistake I can use `constants.HyperkitCommand` even for linux side and then make would be successful right?
 
```patch
$ git diff
diff --git a/pkg/crc/preflight/preflight_checks_linux.go b/pkg/crc/preflight/preflight_checks_linux.go
index bd85d0b..4d5c20f 100644
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -3,6 +3,7 @@ package preflight
 import (
        "bytes"
        "fmt"
+       "github.com/code-ready/crc/pkg/crc/machine/hyperkit"
        "io/ioutil"
        "os"
        "os/exec"
@@ -219,7 +220,7 @@ func fixLibvirtServiceRunning() error {
 
 func checkMachineDriverLibvirtInstalled() error {
        logging.Debugf("Checking if %s is installed", libvirt.MachineDriverCommand)
-
+       logging.Debugf("I can check hyperkit command now", hyperkit.HyperkitCommand)
        machineDriverLibvirt := cache.NewMachineDriverLibvirtCache(libvirt.MachineDriverVersion, getCurrentLibvirtDriverVersion)
        if !machineDriverLibvirt.IsCached() {
                return fmt.Errorf("%s binary is not cached", libvirt.MachineDriverCommand)
```

```
 $ strings crc | grep "code-ready/machine-driver-hyperkit"
tls: either ServerName or InsecureSkipVerify must be specified in the tls.Configx509: invalid signature: parent certificate cannot sign this kind of certificateNetworkManager is required. Please make sure it is installed and running manuallyconfig version is from the future -- you should upgrade your Docker Machine clienthttps://mirror.openshift.com/pub/openshift-v4/clients/crc/latest/release-info.jsoninvalid pull secret, '%q' JSON-object requires either 'auth' or 'credsStore' fieldrefusing to use HTTP_PROXY value in CGI environment; see golang.org/s/cgihttpproxyx509: a root or intermediate certificate is not authorized to sign for this name:  (possibly because of %q while trying to verify candidate authority certificate %q)Cluster TLS certificates have expired, renewing them... [will take up to 5 minutes]Set up local virtualization and networking infrastructure for the OpenShift clusterjson: invalid use of ,string struct tag, trying to unmarshal unquoted value into %vx509: issuer has name constraints but leaf contains unknown or unconstrained name: {{string . "prefix"}}{{counters . }} {{bar . }} {{percent . }}{{string . "suffix"}}^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=|[A-Za-z0-9+\/]{4})$https://github.com/code-ready/machine-driver-hyperkit/releases/download/v%s/hyperkitx509: signature algorithm specifies an %s public key, but have public key of type %TOpen the OpenShift Web Console in the default browser or print its URL or credentialsreflect.Value.Interface: cannot return value obtained from unexported field or methodx509: failed to parse private key (use ParseECPrivateKey instead for this key format) has no exported methods of suitable type (hint: pass a pointer to value of that type)Please provide a configuration property and its value as in 'crc config set KEY VALUE'^([a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62}){1}(\.[a-zA-Z0-9_]{1}[a-zA-Z0-9_-]{0,62})*[\._]?$x509: a root or intermediate certificate is not authorized for an extended key usage: x509: failed to parse public key (use ParsePKCS1PublicKey instead for this key format)Modifies crc configuration properties.
To access the cluster, first set up your environment by following 'crc oc-env' instructionsYou can now run 'crc console' and use these credentials to access the OpenShift web consolehttps://github.com/code-ready/machine-driver-libvirt/releases/download/%s/crc-driver-libvirttls: handshake hash for a client certificate requested after discarding the handshake bufferFailed to open the OpenShift Web Console, you can access it by opening %s in your web browsertls: unsupported certificate: private key is *ed25519.PrivateKey, expected ed25519.PrivateKeyhttps://github.com/code-ready/machine-driver-hyperkit/releases/download/v%s/crc-driver-hyperkitsudo podman run  --ip %s --name dnsmasq -v %s:/etc/dnsmasq.conf -p 53:%d/udp --privileged -d %s3617de4a96262c6f5d9e98bf9292dc29f8f41dbd289a147ce9da3113b5f0b8c00a60b1ce1d7e819d7a431d7c90ea0e5faa87ca22be8b05378eb1c71ef320ad746e1d3b628ba79b9859f741e082542a385502f25dbf55296c3a545e3872760ab7b3312fa7e23ee7e4988e056be3f82d19181d9c6efe8141120314088f5013875ac656398d8a2ed19d2a85c8edd3ec2aef{{string . "prefix"}}{{counters . }} {{bar . }} {{percent . }} {{speed . }}{{string . "suffix"}}Failed to renew TLS certificates: please check if a newer CodeReady Containers release is availablecipher: the nonce can't have zero length, or the security of the key will be immediately compromisedasn1: time did not serialize back to the original value and may be invalid: given %q, but serialized as %qSet the environment for the specified shell: [fish, cmd, powershell, tcsh, bash, zsh]. Default is auto-detect.NS=%s; cat /etc/resolv.conf |grep -i "^nameserver $NS" || echo "nameserver $NS" | sudo tee -a /etc/resolv.confEnable the remote features by doing a blank import of the viper/remote package: '_ github.com/spf13/viper/remote'The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.

```

guillaumerose: Yes but it simplifies a lot the code. Maybe others will have opinions ? @gbraad @anjannath and @cfergeau

gbraad: as long as current code for hyperkit does not end up on Linux, or Hyper-V on macOS (to a degree due to this being part of the generic library), I am OK. but we should not intermix

cmd/crc-embedder/cmd/embed.go
```diff
@@ -7,14 +7,12 @@ import (
...
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
-	"github.com/code-ready/crc/pkg/download"
-
 	"github.com/code-ready/crc/pkg/crc/machine/hyperkit"
 	"github.com/code-ready/crc/pkg/crc/machine/libvirt"
-
-	"github.com/YourFin/binappend"
+	"github.com/code-ready/crc/pkg/download"
 	"github.com/spf13/cobra"
```
cfergeau, line 16: I don't think this needs to be reordered?

cmd/crc-embedder/cmd/extract.go
```diff
@@ -4,7 +4,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/output"
 	"github.com/code-ready/crc/pkg/embed"
-
 	"github.com/spf13/cobra"
 )
```
cfergeau, line 9: Not needed here either

cmd/crc-embedder/cmd/list.go
```diff
@@ -1,10 +1,9 @@
 package cmd
 
 import (
+	"github.com/YourFin/binappend"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/output"
-
-	"github.com/YourFin/binappend"
 	"github.com/spf13/cobra"
 )
```
cfergeau, line 8: Why reorder?

cmd/crc-embedder/cmd/root.go
```diff
@@ -4,7 +4,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/output"
-
 	"github.com/spf13/cobra"
 )
```
cfergeau, line 9: Not the right commit for that imo.

pkg/crc/machine/hyperkit/constants.go
cfergeau: The binary sizes on all 3 platforms are identical before/after this change, so golang dead code elimination seems to be doing its job here. I don't see a lot of code simplification in this PR though? just tag removal?

cmd/crc-embedder/cmd/embed.go
guillaumerose: I removed this change

pkg/crc/machine/hyperkit/constants.go
guillaumerose: Yes just tag removal. Are we ok with this?

cfergeau: if there are no other code simplifications this allows, I would agree with Praveen and Gerard, removing the build tags prevents the compiler from telling us about problematic code, so it feels better to keep it.

guillaumerose:
I will come back with better arguments ! :D

PR #1427, opened 2020-07-30
Windows preflight fixes

anjannath:
- Add a preflight check for windows edition
- Clearly mention to reboot and re-run `crc setup` in windows when setup fails due to not being able to run hyperv commands.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -721,7 +721,7 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	if err != nil {
 		return err
 	}
-	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys", publicKey)
+	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys; chmod 644 /home/core/.ssh/authorized_keys", publicKey)
 	_, err = sshRunner.Run(cmd)
 	if err != nil {
 		return err
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -246,6 +246,14 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	}
 	logging.Info("CodeReady Containers VM is running")
 
+	// Post VM start immediately update SSH key and copy kubeconfig to instance
+	// dir and VM
+	if !exists {
+		if err := updateSSHKeyAndCopyKubeconfigToVM(sshRunner, startConfig, crcBundleMetadata); err != nil {
+			return startError(startConfig.Name, "Error updating public key", err)
+		}
+	}
+
 	// Start network time synchronization if `CRC_DEBUG_ENABLE_STOP_NTP` is not set
 	if stopNtp, _ := strconv.ParseBool(os.Getenv("CRC_DEBUG_ENABLE_STOP_NTP")); !stopNtp {
 		logging.Info("Starting network time synchronization in CodeReady Containers VM")
@@ -328,31 +336,6 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	logging.Info("Check DNS query from host ...")
 	if err := network.CheckCRCLocalDNSReachableFromHost(crcBundleMetadata, instanceIP); err != nil {
 		return startError(startConfig.Name, "Failed to query DNS from host", err)
-	}
-
-	// Additional steps to perform after newly created VM is up
-	if !exists {
-		logging.Info("Generating new SSH key")
-		if err := updateSSHKeyPair(sshRunner); err != nil {
-
-			return startError(startConfig.Name, "Error updating public key", err)
-		}
-		// Copy Kubeconfig file from bundle extract path to machine directory.
-		// In our case it would be ~/.crc/machines/crc/
-		logging.Info("Copying kubeconfig file to instance dir ...")
-		kubeConfigFilePath := filepath.Join(constants.MachineInstanceDir, startConfig.Name, "kubeconfig")
-		err := crcos.CopyFileContents(crcBundleMetadata.GetKubeConfigPath(),
-			kubeConfigFilePath,
-			0644)
-		if err != nil {
-			return startError(startConfig.Name, "Error copying kubeconfig file", err)
-		}
-
-		logging.Info("Copying kubeconfig file to CRC VM ...")
-		err = sshRunner.CopyFile(kubeConfigFilePath, "/opt/kubeconfig", 0644)
-		if err != nil {
-			return startError(startConfig.Name, "Error copying kubeconfig file in VM", err)
-		}
 	}
 
 	ocConfig := oc.UseOCWithSSH(sshRunner)
@@ -746,6 +729,32 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	sshRunner.SetPrivateKeyPath(constants.GetPrivateKeyPath())
 
 	return err
+}
+
+func updateSSHKeyAndCopyKubeconfigToVM(sshRunner *crcssh.Runner, startConfig StartConfig, crcBundleMetadata *bundle.CrcBundleInfo) error {
+	logging.Info("Generating new SSH Key pair ...")
+	if err := updateSSHKeyPair(sshRunner); err != nil {
+		return fmt.Errorf("Error updating SSH Keys: %v", err)
+	}
+
+	// Copy Kubeconfig file from bundle extract path to machine directory.
+	// In our case it would be ~/.crc/machines/crc/
+	logging.Info("Copying kubeconfig file to instance dir ...")
+	kubeConfigFilePath := filepath.Join(constants.MachineInstanceDir, startConfig.Name, "kubeconfig")
+	err := crcos.CopyFileContents(crcBundleMetadata.GetKubeConfigPath(),
+		kubeConfigFilePath,
+		0644)
+	if err != nil {
+		return fmt.Errorf("Error copying kubeconfig file to instance dir: %v", err)
+	}
+
+	logging.Info("Copying kubeconfig file to CRC VM ...")
+	err = sshRunner.CopyFile(kubeConfigFilePath, "/opt/kubeconfig", 0644)
+	if err != nil {
+		return fmt.Errorf("Error copying kubeconfig file to VM: %v", err)
+	}
+
+	return nil
 }
 
 func configureCluster(ocConfig oc.Config, sshRunner *crcssh.Runner, proxyConfig *network.ProxyConfig, pullSecret, instanceIP string) (rerr error) {
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -41,6 +41,30 @@ func checkVersionOfWindowsUpdate() error {
 	return nil
 }
 
+func checkWindowsEdition() error {
+	windowsEditionCmd := `(Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").EditionID`
+
+	stdOut, _, err := powershell.Execute(windowsEditionCmd)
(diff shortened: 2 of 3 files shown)


gbraad:
can we also make sure the DNS check happens AFTER the key rotation? in this case people do not see the wrong error they try to restart the stopped VM, as the keys in that case are else not found..

anjannath:
@gbraad Added 6083a7ca4076b000ad32123a4172f57737e590d6 for the ssh keys issue.

pkg/crc/machine/machine.go
```diff
@@ -757,7 +757,7 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	if err != nil {
 		return err
 	}
-	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys", publicKey)
+	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys; chmod 644 /home/core/.ssh/authorized_keys", publicKey)
```
cfergeau, line 760: > but in case of fedora coreos the file's initial permissions are too open

This would be good to report a bug against fedora coreos so that they can fix it. Or maybe it's just that on fedora coreos, ~/.ssh/authorized_keys initially does not exist, and when a new file is created, it's getting 664 permissions by default?

pkg/crc/machine/machine.go
```diff
@@ -321,11 +329,6 @@ func Start(startConfig StartConfig) (StartResult, error) {
 
 	// Additional steps to perform after newly created VM is up
 	if !exists {
-		logging.Info("Generating new SSH key")
-		if err := updateSSHKeyPair(sshRunner); err != nil {
-
-			return startError(startConfig.Name, "Error updating public key", err)
-		}
 		// Copy Kubeconfig file from bundle extract path to machine directory.
 		// In our case it would be ~/.crc/machines/crc/
```
cfergeau, line 333: This commit avoids issues with the ssh key not being available, but won't we be getting issues with a missing kubeconfig where we were getting issues with the missing ssh key before?

pkg/crc/preflight/preflight_windows.go
```diff
@@ -15,6 +15,13 @@ var hypervPreflightChecks = [...]Check{
 		fixDescription:   "Please manually update your Windows 10 installation",
 		flags:            NoFix,
 	},
+	{
+		configKeySuffix:  "check-windows-edition",
+		checkDescription: "Checking Windows edition",
+		check:            checkWindowsEdition,
+		fixDescription:   "Your Windows edition is not supported. Consider using Professional or Enterprise edition of Windows",
```
cfergeau, line 22: `editions` instead of `edition`

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -41,6 +41,22 @@ func checkVersionOfWindowsUpdate() error {
...
+	if err != nil {
+		logging.Debug(err.Error())
+		return fmt.Errorf("Failed to get Windows edition")
+	}
+
+	windowsEdition := strings.TrimSpace(stdOut)
+	if windowsEdition == "Professional" || windowsEdition == "Enterprise" {
+		return nil
+	}
+	return fmt.Errorf("Supported windows editions are Professional and Enterprise. Windows %s edition is not supported", windowsEdition)
```
cfergeau, line 57: Upper case `W` is needeed in `Windows`

cfergeau: This could be a good use case for a `switch/case`?

pkg/crc/preflight/preflight_windows.go
```diff
@@ -40,7 +40,7 @@ var hypervPreflightChecks = [...]Check{
 		configKeySuffix:  "check-hyperv-switch",
 		checkDescription: "Checking if the Hyper-V virtual switch exist",
 		check:            checkIfHyperVVirtualSwitchExists,
-		fixDescription:   "Unable to perform Hyper-V administrative commands. Please make sure to re-login or reboot your system",
+		fixDescription:   "Unable to perform Hyper-V administrative commands. Please reboot your system and run 'crc setup' again.",
```
cfergeau, line 50: The commit log says a re-login is sometimes necessary, but this removes it from the message? Do we need to run crc setup again after rebooting to configure the virtual switch?

anjannath: >The commit log says a re-login is sometimes necessary, but this removes it from the message?

Ahh, need to modify the commit msg then, it should also say to reboot, i've had less luck wiht re-login, but reboot always seems to work.

>Do we need to run crc setup again after rebooting to configure the virtual switch?

user doesn't have to run `crc setup` again, but i added this so that user knows if all the setup steps were successfully done.

How about asking to run `crc start`, `Unable to perform Hyper-V administrative commands. Please reboot your system and run 'crc start'.` as the same preflight cheks will be done for start also.

pkg/crc/machine/machine.go
anjannath: Need to check again, i did check for this but forgot.. :( if its preexisting with wrong permissions will report a bug

pkg/crc/machine/machine.go
anjannath: Yes, we'd be getting issues with that, makes sense to copy the kubeconfig also earlier..

pkg/crc/machine/machine.go
```diff
@@ -253,6 +253,14 @@ func Start(startConfig StartConfig) (StartResult, error) {
...
 	}
 
+	// Generate and copy new ssh key to the newly created vm
+	if !exists {
+		logging.Info("Generating new SSH key")
+		if err := updateSSHKeyPair(sshRunner); err != nil {
+			return startError(startConfig.Name, "Error updating public key", err)
+		}
+	}
+
```
praveenkumar, line 280: Then it should be after L#255, as soon as ssh is available.

anjannath: Aha (its as soon as possible :-), we are only doing these two things after the start. i think it makes sense to change the ssh key first and then send the kubeconfig.

This would obviously not prevent the start failures, but at least we'll know why a second start fails when the initial start failed, now that the potential missing `id_rsa` and `kubeconfig` errors should be avoided, we might see the real error why it fails.

pkg/crc/machine/machine.go
```diff
@@ -253,6 +253,31 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		}
 	}
 
+	// Additional steps to perform after newly created VM is up
```
guillaumerose, line 256: Can you extract this in a new method ?

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -41,6 +41,28 @@ func checkVersionOfWindowsUpdate() error {
 	return nil
 }
 
+func checkWindowsEdition() error {
+	windowsEditionCmd := `(Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").EditionID`
+
+	stdOut, _, err := powershell.Execute(windowsEditionCmd)
+	if err != nil {
+		logging.Debug(err.Error())
+		return fmt.Errorf("Failed to get Windows edition")
```
guillaumerose, line 50: `return errors.Wrap(err, "Failed to get Windows edition")`  is better because it doesn't hide the original error.

cfergeau: For debugging, the original error is useful, but at the  same time, it goes against showing user-friendly error messages in our UI. So both options are desirable :D We are currently very bad at that anyway, sonot saying one way is better than the other in this case.

pkg/crc/preflight/preflight_windows.go
cfergeau: I did not  follow you :) If after a reboot, `setup` is never required, the message can just be `Please reboot your system to complete the setup process`. If `crc setup` is needed in some cases, then the current message is good in these cases :)

pkg/crc/machine/machine.go
anjannath: > Or maybe it's just that on fedora coreos, ~/.ssh/authorized_keys initially does not exist, and when a new file is created, it's getting 664 permissions by default?

This is the case, and i think since the suggested way is to inject keys using ignition which uses the `authorized_keys.d` directory not sure if this should be reported.

pkg/crc/preflight/preflight_windows.go
anjannath: Changed the message to `Please reboot your system to complete the setup process`.

I had the weird idea of making the users run `crc setup` again just to verify if the setup succeeded earlier ;-)

pkg/crc/machine/machine.go
cfergeau: I don't think it needs to be reported, creating a new file will honour your umask settings, and that's about it. If the default permissions don't match what you need, then you get to change them yourself, which is what this commit is doing

pkg/crc/machine/machine.go
```diff
@@ -246,6 +246,14 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	}
 	logging.Info("CodeReady Containers VM is running")
 
+	// Post VM start immediately update SSH key and copy kubeconfig to instance
+	// dir and VM
+	if !exists {
+		if err := updateSSHKeyAndCopyKubeconfigToVM(sshRunner, startConfig, crcBundleMetadata); err != nil {
```
cfergeau, line 252: It's not just `ToVM`, it's both to VM and to instance dir. Fwiw, I don't really think updating the ssh key and copying the kubeconfig file to the right places need to be in the same helper.

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -41,6 +41,28 @@ func checkVersionOfWindowsUpdate() error {
...
+	}
+
+	windowsEdition := strings.TrimSpace(stdOut)
+	logging.Debugf("Running on Windows %s edition", windowsEdition)
+
+	switch windowsEdition {
+	case "Professional":
```
cfergeau, line 57: I spend some time testing this, more than expected ^^ The good news is that on my French "Windows 10 Professionnel", "Professional" was not localized. The relatively bad news is that I tested "ProfessionalN" and "ProfessionalEducationN" Windows releases which both seem to have hyperv support (I managed to get `crc setup` to succeed after disabling the edition check, and running it a second time after reboot).

And I realized that on new systems (without hyperv), we need to mention somewhere that rebooting and re-running `crc setup` is required. After enabling hyperv, a reboot is required, and since hyperv was not enabled, all the subsequent setup steps were not properly done, so crc setup needs to be rerun 
```


        INFO Checking if goodhosts binary is cached
        INFO Checking Windows edition
        WARN Skipping above check ...
        INFO Installing Hyper-V
        INFO Will run as admin: enable Hyper-V
        ERRO Please reboot your system
        INFO Checking if user is a member of the Hyper-V Administrators group
        INFO Adding user to the Hyper-V Administrators group
        INFO Will run as admin: add user to hyperv admins group
        INFO Enabling Hyper-V service
        INFO Will run as admin: enable Hyper-V service
        FATA Unable to perform Hyper-V administrative commands. Please reboot your system to complete the setup process
```

anjannath: > And I realized that on new systems (without hyperv), we need to mention somewhere that rebooting and re-running crc setup is required. After enabling hyperv, a reboot is required, and since hyperv was not enabled, all the subsequent setup steps were not properly done, so crc setup needs to be rerun

maybe this was the case i also encountered and needed a reboot, but forgot exactly, good that this is clear now.. will revert the message to `Please reboot your system and run 'crc setup' to complete the setup process`

cfergeau: > Interesting, i didn't knew about these editions, so need to look at the whole list of editions that support hyperv and add all of those name to the check..

My feeling is that `strings.HasPrefix(edition, "professional")` would cover these.

PR #1428, opened 2020-07-31
Determine compress algorithm by checking magic bytes

guillaumerose:
This allows us to change the format of crcbundle to what we want.


cfergeau:
Did you consider using a dedicated go module for this content sniffing? https://stackoverflow.com/questions/29838185/how-to-detect-additional-mime-type-in-golang (not saying this should be done, merely asking :)

guillaumerose:
Very good point! I was too happy to write it myself. It's better now :)

pkg/extract/extract_test.go
```diff
@@ -9,21 +9,59 @@ import (
...
+	}
+	filteredFiles fileMap = map[string]string{
+		"/a/b/c.txt": "ccc",
+	}
+	archives = []string{
+		"test.tar",
+		"test.tar.gz",
+		"test.zip",
+		"test.tar.xz",
+	}
```
praveenkumar, line 32: During this unit test can we also put a `crcbundle` tarball file in the testdata folder just to make sure it actually able to detect it as a tarball?

guillaumerose: Yes this is exactly the second test https://github.com/code-ready/crc/pull/1428/files#diff-92bf4f1e87f4458b933a72c9b54411d5R42

crcbundle can be any type now and it checks for it.

pkg/extract/extract_test.go
```diff
@@ -9,21 +9,59 @@ import (
...
+func TestUnCompressBundle(t *testing.T) {
+	dir, err := ioutil.TempDir("", "bundles")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	bundle := filepath.Join(dir, "test.crcbundle")
+	for _, archive := range archives {
+		require.NoError(t, crcos.CopyFileContents(filepath.Join("testdata", archive), bundle, 0600))
+		assert.NoError(t, testUncompress(bundle, nil, files))
+		assert.NoError(t, testUncompress(bundle, fileFilter, filteredFiles))
```
praveenkumar, line 51: :+1:

Issue #1430, opened 2020-07-31
CRC VM does not start after 30 attempts


rrapolu12:
As part of troubleshooting, i have run `crc delete `and `crc cleanup` and then i ran `crc setup` and `crc start.`

i see the VM starts after 4 min, but it somehow throws and an unknown error.

i am attaching the crc.log 
[crc.log](https://github.com/code-ready/crc/files/5007904/crc.log)

praveenkumar:
@rrapolu12 Can you please file the bug using the template, you can close this issue and try to file a new one using the template so we will have more info around the environment also.

praveenkumar:
closing this issue since the reporter didn't provide required info.

PR #1431, opened 2020-07-31
Propogate errors outside of network package

guillaumerose:
It would help issue https://github.com/code-ready/crc/issues/1430

### DIFF
--- a/pkg/crc/network/utils.go
+++ b/pkg/crc/network/utils.go
@@ -1,7 +1,6 @@
 package network
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"net/url"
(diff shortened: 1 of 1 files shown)


pkg/crc/network/utils.go
```diff
@@ -12,18 +11,14 @@ import (
...
-// HostIPs returns the IP addresses assigned to the host
-func HostIPs() []string {
+// hostIPs returns the IP addresses assigned to the host
+func hostIPs() ([]string, error) {
 	ips := []string{}
 
-	ifaces, _ := net.Interfaces()
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return nil, err
```
anjannath, line 20: 👍

praveenkumar:
This patch should be tested on the windows, I think we ignore the errors because it takes some iteration to get the interface IP on windows. @jsliacan can you test this on windows  when have some time?

guillaumerose:
Yes the loop is here https://github.com/code-ready/crc/blob/master/pkg/crc/machine/machine.go#L284
It's always a retriable error so it should be fine.

PR #1434, opened 2020-08-04
Use only machine.GetConsoleURL and remove machine.GetProxyConfig

guillaumerose:
and a small refactoring around errors.


cfergeau:
I knew there was a reason for the slight differences, but could not find it during the review:
Initial state:
```
$ ./out/linux-amd64/crc status
Machine 'crc' does not exist. Use 'crc start' to create it
```

With commit 437ea92fdab 
```
$ ./out/linux-amd64/crc oc-env
ERRO Error getting the state for host: Failed to fetch machine 'crc' 
Error getting the state for host: Failed to fetch machine 'crc'
```

With commit 437ea92fdab^
```
$ ./out/linux-amd64/crc oc-env
export PATH="/home/teuf/.crc/bin/oc:$PATH"
export HTTP_PROXY="http://squid.redhat.com:3128"
export HTTPS_PROXY="http://squid.redhat.com:3128"
export NO_PROXY="127.0.0.1,localhost,.testing"
# eval $(crc oc-env)
```

If the VM already exists but is stopped, both work commits give the expected output. Not necessarily a huge issue, just thought I'd point it out.

guillaumerose:
`ERRO Error getting the state for host: Failed to fetch machine 'crc' ` is triggered when you started the machine and then deleted the crc machine inside qemu but not in ~/.crc. Am I right ? Otherwise you would have a machine loading error.

cfergeau:
I can double check my setup, but I only did `crc delete` on this machine, I did not manually mess with `~/.crc/machines` or `virsh undefine`

cfergeau:
Hmm yes, my system must have been in a confused state when I tested this yesterday, behaviour is consistent now that I retest this, and it's indeed saying the VM does not exist.

PR #1435, opened 2020-08-04
Issue #1227 Use a type to hold all machine functions

guillaumerose:
This helps to test the API and the command line.


pkg/crc/machine/machine.go
```diff
@@ -107,7 +107,7 @@ func createLibMachineClient(debug bool) (*libmachine.Client, func(), error) {
 	}, nil
 }
 
-func Start(startConfig StartConfig) (StartResult, error) {
+func (client *LibMachineClient) Start(startConfig StartConfig) (StartResult, error) {
 	var crcBundleMetadata *bundle.CrcBundleInfo
 
 	libMachineAPIClient, cleanup, err := createLibMachineClient(startConfig.Debug)
```
cfergeau, line 113: Seeing this, I don't think the `client *LibMachineClient` naming is quite right, as even if we have a `LibMachineClient` already, we need to call a method named `createLibMachineClient` to get a different kind of LibMachineClient?
`client *MachineClient` is probably good enough as a name, as it's a client used in the `machine` package?

pkg/crc/api/client_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/code-ready/crc/pkg/crc/machine"
+	"github.com/code-ready/crc/pkg/crc/network"
+)
+
+type mockClient struct {
+}
+
+func (c mockClient) Delete(deleteConfig machine.DeleteConfig) (machine.DeleteResult, error) {
+	return machine.DeleteResult{}, errors.New("not implemented")
+}
```
cfergeau, line 16: This almost duplicates what is in `cmd/crc/cmd/client_test.go`, does it have to be this way?

pkg/crc/api/client_test.go
```diff
@@ -8,37 +8,44 @@ import (
...
 	return machine.PowerOffResult{}, errors.New("not implemented")
 }
 
-func (c mockClient) Start(startConfig machine.StartConfig) (machine.StartResult, error) {
+func (c *mockClient) Start(startConfig machine.StartConfig) (machine.StartResult, error) {
 	return machine.StartResult{}, errors.New("not implemented")
 }
 
-func (c mockClient) Stop(stopConfig machine.StopConfig) (machine.StopResult, error) {
+func (c *mockClient) Stop(stopConfig machine.StopConfig) (machine.StopResult, error) {
```
cfergeau, line 38: This could already be a `c *mockClient` when introduced in the previous commit I believe?

pkg/crc/machine/machine.go
guillaumerose: When I changed to MachineClient the linter complains:

```
pkg/crc/machine/client.go:14:6: type name will be used as machine.MachineClient by other packages, and that stutters; consider calling this Client (golint)
type MachineClient struct{}
``` 

Other idea ? It's hard to find  a good name here :(

pkg/crc/machine/client.go
```diff
@@ -0,0 +1,20 @@
...
+	return &RealClient{}
+}
+
+func (*RealClient) Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
+	return Status(statusConfig)
+}
+
+func (*RealClient) Exists(name string) (bool, error) {
+	return Exists(name)
+}
```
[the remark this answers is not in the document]
guillaumerose: done :+1:

pkg/crc/api/client_test.go
guillaumerose: done :+1: 

Introduced a fakemachine package with this Client.

pkg/crc/api/client_test.go
guillaumerose: done :+1:

pkg/crc/machine/machine.go
guillaumerose: I chose `client` for the moment (and the interface is `Client`). Not sure if it's better but it's a start.

cmd/crc/cmd/console.go
```diff
@@ -41,11 +41,12 @@ func runConsole(arguments []string) error {
 		Name: constants.DefaultName,
 	}
 
-	if err := checkIfMachineMissing(machine.NewClient(), consoleConfig.Name); err != nil {
+	client := machine.NewClient()
+	if err := checkIfMachineMissing(client, consoleConfig.Name); err != nil {
```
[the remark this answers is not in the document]
guillaumerose: Added a line about mock and tests.

pkg/crc/machine/machine.go
cfergeau: Yup, only remembered the linter would complain while away from the computer :) I guess if we try to call it `Machine` the linter will also fail? `Instance` might do, but not much better than `Client`. Or `Manager` as this really is about managing our crc machine(s)?

PR #1436, opened 2020-08-05
Add integration test for proxy

praveenkumar:


guillaumerose:
Maybe add an assertion to check that the proxy is actually used. I must admit it's not easy!

An other way of doing this is to run the proxy as a go routines of the test binary (with https://github.com/elazarl/goproxy for example). It removes the extra container, makes it portable to Windows and you can check that it is used. (as you want, your PR is totally fine. We can improve later)

praveenkumar:
> Maybe add an assertion to check that the proxy is actually used. I must admit it's not easy!

@guillaumerose assertion in the integration test? 

> An other way of doing this is to run the proxy as a go routines of the test binary (with https://github.com/elazarl/goproxy for example). It removes the extra container, makes it portable to Windows and you can check that it is used. (as you want, your PR is totally fine. We can improve later)

Yes this is the module I also looked and can be used as as daemon process, something we used to do in minishift, Also I think we do need the docker for Mac/windows installed by default and use docker instead of podman so all those tests can be executed on the win/mac side also.

guillaumerose:
>assertion in the integration test?

Yes, in this case, it could be an assertion on squid logs. Checking the traffic.

cfergeau:
Running goproxy code in a goroutine inside the test process may be simpler/more flexible than going through an external container image.

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,31 @@
+@proxy @linux
+Feature: Test the proxy
+
+  The user try to use crc behind proxy. he/she expect the
+  crc start happen successfully and able to deploy the app and check its
+  accessibility.
+
+  Scenario: Setup the proxy container using podman
+    Given executing "sudo podman run --name squid -d -p 3128:3128 quay.io/crcont/squid" succeeds
```
cfergeau, line 9: maybe there are pre-made squid images which we could use?

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,31 @@
...
+    Then executing "crc config set https-proxy http://192.168.130.1:3128" succeeds
+    When starting CRC with default bundle succeeds
+    Then stdout should contain "Started the OpenShift cluster"
+    And executing "eval $(crc oc-env)" succeeds
+    When with up to "4" retries with wait period of "2m" command "crc status --log-level debug" output matches ".*Running \(v\d+\.\d+\.\d+.*\).*"
+    Then login to the oc cluster succeeds
+
+  Scenario: Remove the proxy container and host proxy env (which set because of oc-env)
+    Given executing "sudo podman stop squid" succeeds
+    And executing "sudo podman rm squid" succeeds
```
cfergeau, line 23: Is this intentional that you did not use `podman run --rm`, which would avoid the need for this `podman rm`?

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,31 @@
...
+    And executing "unset HTTP_PROXY HTTPS_PROXY NO_PROXY" succeeds
+
+
+  Scenario: CRC delete and remove proxy setting from config
```
cfergeau, line 27: 'proxy settings' maybe?

test/integration/features/proxy.feature
praveenkumar: @cfergeau you mean instead maintaining `quay.io/crcont/squid` ourself?  I really want to control the image build ourself for any crc specific usage. I don't trust existing images, not sure what base image they chose and how often it updated. Another good thing maintaining ourself, if we need any change we just do it and rebuild then waiting for existing image maintainer.

test/integration/features/proxy.feature
praveenkumar: @cfergeau `--rm` option doesn't behave well with `-d` the way we are running. Also old version of podman doesn't support to use `--rm` with `-d` option (what docker used to do).

```
$ podman run --name squid --rm -d -p 3128:3128 quay.io/crcont/squid 
42560e1d8a7a157ced7fa393c1c475715dc532af70c292c1315009d6c77503e5

$ podman stop squid
Error: container 42560e1d8a7a157ced7fa393c1c475715dc532af70c292c1315009d6c77503e5 does not exist in database: no such container
```

test/integration/features/proxy.feature
cfergeau: > I don't trust existing images, not sure what base image they chose and how often it updated. 

And we know we'll almost never update it, not sure it's better ;)

> Another good thing maintaining ourself, if we need any change we just do it and rebuild then waiting for existing image maintainer.

We always have the option to switch to our own image when we have such needs. And in the meantime, that's one less thing we have to take care of.

PR #1437, opened 2020-08-05
Remove integration build tag

guillaumerose:
Simplify the build code and fix lint errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,6 @@ GOPATH ?= $(shell go env GOPATH)
 ORG := github.com/code-ready
 REPOPATH ?= $(ORG)/crc
 
-PACKAGES := go list --tags build ./...
 SOURCES := $(shell git ls-files  *.go ":^vendor")
 
 RELEASE_INFO := release-info.json
@@ -96,7 +95,7 @@ cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/wi
 
 .PHONY: test
 test:
-	go test --tags build -v -ldflags="$(VERSION_VARIABLES)" $(shell $(PACKAGES))
+	go test --tags build -v -ldflags="$(VERSION_VARIABLES)" ./pkg/... ./cmd/...
 
 .PHONY: build_docs
 build_docs:
@@ -120,7 +119,6 @@ clean: clean_docs
 	rm -f $(GOPATH)/bin/crc
 	rm -rf $(RELEASE_DIR)
        
-
 .PHONY: integration ## Run integration tests
 integration:
 GODOG_OPTS = --godog.tags=$(GOOS)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -121,6 +120,9 @@ clean: clean_docs
 	rm -rf $(RELEASE_DIR)
        
 
+$(HOST_BUILD_DIR)/integration.test: $(SOURCES)
```
praveenkumar, line 123: What is the use of this target, may be put that details to commit log also?

Makefile
```diff
@@ -96,7 +95,7 @@ cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/wi
 
 .PHONY: test
 test:
-	go test --tags build -v -ldflags="$(VERSION_VARIABLES)" $(shell $(PACKAGES))
+	go test --tags build -v -ldflags="$(VERSION_VARIABLES)" ./pkg/... ./cmd/...
```
praveenkumar, line 98: :+1:

Makefile
guillaumerose: I forgot to add it but we should check we can build integration test on all platforms. Perhaps in a new target like `make cross-integration`.

praveenkumar: @guillaumerose but then crc binary for mac/windows not going to run on linux so `make cross-integration` will not work on a single platform.

guillaumerose: I added the CI target, let me know what you think of https://github.com/code-ready/crc/pull/1437/commits/fa753f28b43688cf69ac09b78a107726bddff89e !

Makefile
```diff
@@ -119,7 +118,12 @@ clean: clean_docs
 	rm -rf $(BUILD_DIR)
 	rm -f $(GOPATH)/bin/crc
 	rm -rf $(RELEASE_DIR)
-       
+
+.PHONY: build_integration
+build_integration: $(SOURCES)
+	GOOS=linux   go test ./test/integration/ -c -o $(BUILD_DIR)/linux-amd64/integration.test
+	GOOS=windows go test ./test/integration/ -c -o $(BUILD_DIR)/windows-amd64/integration.test.exe
+	GOOS=darwin  go test ./test/integration/ -c -o $(BUILD_DIR)/darwin-amd64/integration.test
```
praveenkumar, line 126: @guillaumerose This is really new for me, I didn't know we can even do that :(. can this binary used directly starting the integration tests with the arguments?

praveenkumar: @guillaumerose Also for mac target we use `macos-amd64` instead `darwin-amd64`

guillaumerose: Fixed macos/darwin mistake.

Yes we can use the binary! The only issue I have is that it uses `os.Getwd` and it really wants to be launched in `test/integration`.

The Makefile should look like

``` 
integration: 
	(cd ./test/integration && $(BUILD_DIR)/linux-amd64/integration.test --test.timeout=120m -test.v $(PULL_SECRET_FILE) $(BUNDLE_LOCATION) $(CRC_BINARY) --bundle-version=$(BUNDLE_VERSION) $(GODOG_OPTS))
```

I will open a new PR with this change, if it's ok for you :)

praveenkumar: Ah so it is only create the binary which doesn't include feature files and that's the reason needed to launched from `test/integration` ?

guillaumerose: yes!

PR #1439, opened 2020-08-05
exitHandlers are never used. Removing them.

guillaumerose:
Small cleanup!


praveenkumar:
This again came from minishift side where we used those for some of mocking side.

PR #1441, opened 2020-08-05
sshRunner API cleanups and fix small error in NTP logs

cfergeau:
The first 3 commits are merely minor code cleanups/improvements, the last commit fixes an erroneous log message, which says the NTP service is being stopped, when it's being started.

## Proposed changes

1. Add ssh.CopyFile
2. Merge ssh.CopyData and ssh.SetTextContentAsRoot
3. Improve variable name
4. Fix log message

## Testing

Check that during cluster startup (creation?), there's a `Starting network time synchronization in CodeReady Containers VM` log message, not a `Stopping network time synchronization in CodeReady Containers VM` log.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -346,9 +346,9 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		if err != nil {
 			return startError(startConfig.Name, "Error copying kubeconfig file", err)
 		}
-		// Copy kubeconfig file inside the VM
-		kubeconfigContent, _ := ioutil.ReadFile(kubeConfigFilePath)
-		_, err = sshRunner.RunPrivate(fmt.Sprintf("cat <<EOF | sudo tee /opt/kubeconfig\n%s\nEOF", string(kubeconfigContent)))
+
+		logging.Info("Copying kubeconfig file to CRC VM ...")
+		err = sshRunner.CopyFile(kubeConfigFilePath, "/opt/kubeconfig", 0644)
 		if err != nil {
 			return startError(startConfig.Name, "Error copying kubeconfig file in VM", err)
 		}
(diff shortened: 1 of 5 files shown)


pkg/crc/machine/machine.go
```diff
@@ -346,9 +346,9 @@ func Start(startConfig StartConfig) (StartResult, error) {
 		if err != nil {
 			return startError(startConfig.Name, "Error copying kubeconfig file", err)
 		}
-		// Copy kubeconfig file inside the VM
-		kubeconfigContent, _ := ioutil.ReadFile(kubeConfigFilePath)
-		_, err = sshRunner.RunPrivate(fmt.Sprintf("cat <<EOF | sudo tee /opt/kubeconfig\n%s\nEOF", string(kubeconfigContent)))
+
+		logging.Info("Copying kubeconfig file to CRC VM ...")
```
praveenkumar, line 350: Copy kubeconfig file is going to be part of snc so we don't need to copy it from here, but we can revert it as soon as it lands to snc side.

pkg/crc/machine/machine.go
```diff
@@ -245,8 +245,8 @@ func Start(startConfig StartConfig) (StartResult, error) {
 	}
 	logging.Info("CodeReady Containers VM is running")
 
-	// Start network time synchronization if `CRC_DEBUG_ENABLE_STOP_NTP` not set
-	if b, _ := strconv.ParseBool(os.Getenv("CRC_DEBUG_ENABLE_STOP_NTP")); !b {
+	// Start network time synchronization if `CRC_DEBUG_ENABLE_STOP_NTP` is not set
+	if stopNtp, _ := strconv.ParseBool(os.Getenv("CRC_DEBUG_ENABLE_STOP_NTP")); !stopNtp {
```
praveenkumar, line 249: :+1:

PR #1443, opened 2020-08-06
Add feature to automate proxy ca cert

praveenkumar:
## Solution/Idea

This PR is try to automate the steps which are part of https://github.com/code-ready/crc/wiki/Custom-CA-cert-for-proxy page

## Proposed changes

- Add the `proxy-ca-file` option in the config

## Testing

- Create the ssl cert using openssl
```
$ openssl req -new -newkey rsa:2048 -sha256 -days 365 -nodes -x509 -extensions v3_ca -keyout myCA.pem  -out myCA.pem
```
- Use the specific squid config to test it out
```
$ cat squid.conf
#
# Recommended minimum configuration:
#

# Example rule allowing access from your local networks.
# Adapt to list your (internal) IP networks from where browsing
# should be allowed
acl localnet src 0.0.0.1-0.255.255.255	# RFC 1122 "this" network (LAN)
acl localnet src 10.142.0.0/8		# RFC 1918 local private network (LAN)
acl localnet src 172.16.0.0/12		# RFC 1918 local private network (LAN)
acl localnet src 192.168.0.0/16		# RFC 1918 local private network (LAN)
acl localnet src 103.127.32.1		# RFC 1918 local private network (LAN)
acl localnet src 35.237.1.197		# RFC 1918 local private network (LAN)

acl SSL_ports port 443
acl Safe_ports port 80		# http
acl Safe_ports port 21		# ftp
acl Safe_ports port 443		# https
acl Safe_ports port 70		# gopher
acl Safe_ports port 210		# wais
acl Safe_ports port 1025-65535	# unregistered ports
acl Safe_ports port 280		# http-mgmt
acl Safe_ports port 488		# gss-http
acl Safe_ports port 591		# filemaker
acl Safe_ports port 777		# multiling http
acl CONNECT method CONNECT

#
# Recommended minimum Access Permission configuration:
#
# Deny requests to certain unsafe ports
http_access deny !Safe_ports

# Deny CONNECT to other than secure SSL ports
http_access deny CONNECT !SSL_ports

# Only allow cachemgr access from localhost
http_access allow localhost manager
http_access deny manager

# We strongly recommend the following be uncommented to protect innocent
# web applications running on the proxy server who think the only
# one who can access services on "localhost" is a local user
http_access deny to_localhost

#
# INSERT YOUR OWN RULE(S) HERE TO ALLOW ACCESS FROM YOUR CLIENTS
#

# Example rule allowing access from your local networks.
# Adapt localnet in the ACL section to list your (internal) IP networks
# from where browsing should be allowed
http_access allow localnet
http_access allow localhost

# And finally deny all other access to this proxy
http_access deny all

# Squid normally listens to port 3128
http_port 3128 ssl-bump \
  cert=/etc/squid/ssl_cert/myCA.pem \
  generate-host-certificates=on dynamic_cert_mem_cache_size=16MB
# Uncomment and adjust the following to add a disk cache directory.
#cache_dir ufs /var/spool/squid 100 16 256

# Leave coredumps in the first cache dir
coredump_dir /var/spool/squid

#
# Add any of your own refresh_pattern entries above these.
#
refresh_pattern ^ftp:		1440	20%	10080
refresh_pattern ^gopher:	1440	0%	1440
refresh_pattern -i (/cgi-bin/|\?) 0	0%	0
refresh_pattern .		0	20%	4320
acl to_metadata dst 169.254.169.254
http_access deny to_metadata

sslcrtd_program  /usr/lib64/squid/security_file_certgen -s /var/lib/ssl_db -M 4MB
acl step1 at_step SslBump1

ssl_bump peek step1
ssl_bump bump all
ssl_bump splice all
```
- Run the squid container image with mount of those file
```
$ podman run --name squid -v <host_path>/squid.conf:/etc/squid/squid.conf:Z -v <host_path>/myCA.pem:/etc/squid/ssl_cert/myCA.pem:Z -p 3128:3128 -d  quay.io/crcont/squid
```
- Run the CRC with proxy
```
$ ./crc config view
- http-proxy                            : http://192.168.130.1:3128
- https-proxy                           : http://192.168.130.1:3128
- proxy-ca-file                         : <PATH>/myCA.pem
```

- Check the proxy logs from podman
```
$ podman exec -it squid /bin/bash
<container> # tail -f /var/log/squid/access.log
```

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -23,9 +23,10 @@ var (
 
 	// Proxy Configuration
 
-	HTTPProxy  = cfg.AddSetting("http-proxy", nil, []cfg.ValidationFnType{cfg.ValidateURI}, []cfg.SetFn{cfg.SuccessfullyApplied})
-	HTTPSProxy = cfg.AddSetting("https-proxy", nil, []cfg.ValidationFnType{cfg.ValidateURI}, []cfg.SetFn{cfg.SuccessfullyApplied})
-	NoProxy    = cfg.AddSetting("no-proxy", nil, []cfg.ValidationFnType{cfg.ValidateNoProxy}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	HTTPProxy   = cfg.AddSetting("http-proxy", nil, []cfg.ValidationFnType{cfg.ValidateURI}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	HTTPSProxy  = cfg.AddSetting("https-proxy", nil, []cfg.ValidationFnType{cfg.ValidateURI}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	NoProxy     = cfg.AddSetting("no-proxy", nil, []cfg.ValidationFnType{cfg.ValidateNoProxy}, []cfg.SetFn{cfg.SuccessfullyApplied})
+	ProxyCAFile = cfg.AddSetting("proxy-ca-file", nil, []cfg.ValidationFnType{cfg.ValidatePath}, []cfg.SetFn{cfg.SuccessfullyApplied})
 )
 
 var (
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
+	"strings"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	"github.com/code-ready/crc/pkg/crc/config"
@@ -97,15 +99,37 @@ func setProxyDefaults() {
 	httpProxy := config.GetString(cmdConfig.HTTPProxy.Name)
 	httpsProxy := config.GetString(cmdConfig.HTTPSProxy.Name)
 	noProxy := config.GetString(cmdConfig.NoProxy.Name)
+	proxyCAFile := config.GetString(cmdConfig.ProxyCAFile.Name)
 
-	proxyConfig, err := network.NewProxyDefaults(httpProxy, httpsProxy, noProxy)
+	proxyCAData, err := getProxyCAData(proxyCAFile)
+	if err != nil {
+		exit.WithMessage(1, fmt.Sprintf("not able to read proxyCAFile %s: %v", proxyCAFile, err.Error()))
+	}
+
+	proxyConfig, err := network.NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAData)
 	if err != nil {
 		exit.WithMessage(1, err.Error())
 	}
 
 	if proxyConfig.IsEnabled() {
-		logging.Debugf("HTTP-PROXY: %s, HTTPS-PROXY: %s, NO-PROXY: %s", proxyConfig.HTTPProxyForDisplay(),
-			proxyConfig.HTTPSProxyForDisplay(), proxyConfig.GetNoProxyString())
+		logging.Debugf("HTTP-PROXY: %s, HTTPS-PROXY: %s, NO-PROXY: %s, proxyCAFile: %s", proxyConfig.HTTPProxyForDisplay(),
+			proxyConfig.HTTPSProxyForDisplay(), proxyConfig.GetNoProxyString(), proxyCAFile)
 		proxyConfig.ApplyToEnvironment()
 	}
+}
+
+func getProxyCAData(proxyCAFile string) (string, error) {
+	if proxyCAFile == "" {
+		return "", nil
+	}
+	proxyCACert, err := ioutil.ReadFile(proxyCAFile)
+	if err != nil {
+		return "", err
+	}
+	// Before passing string back to caller function, remove the empty lines in the end
+	return trimTrailingEOL(string(proxyCACert)), nil
+}
+
+func trimTrailingEOL(s string) string {
+	return strings.TrimRight(s, "\n")
 }
--- a/docs/source/topics/proc_starting-codeready-containers-behind-proxy.adoc
+++ b/docs/source/topics/proc_starting-codeready-containers-behind-proxy.adoc
@@ -22,6 +22,16 @@ $ {bin} config set https-proxy http://example.proxy.com:__<port>__
 $ {bin} config set no-proxy __<comma-separated-no-proxy-entries>__
 ----
 +
+
+. If proxy uses custom CA then use the `{bin} config set` command as follows:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config set proxy-ca-file __<Path to custom CA file>__
+----
++
+
+
 The [command]`{bin}` binary will be able to use the defined proxy once set through environment variables or {prod} configuration.
 
 [NOTE]
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
@@ -119,16 +120,79 @@ func UpdateClusterID(ocConfig oc.Config) error {
 	return nil
 }
 
-func AddProxyConfigToCluster(ocConfig oc.Config, proxy *network.ProxyConfig) error {
-	cmdArgs := []string{"patch", "proxy", "cluster", "-p",
-		fmt.Sprintf(`'{"spec":{"httpProxy":"%s", "httpsProxy":"%s", "noProxy":"%s"}}'`, proxy.HTTPProxy, proxy.HTTPSProxy, proxy.GetNoProxyString()),
-		"-n", "openshift-config", "--type", "merge"}
+func AddProxyConfigToCluster(sshRunner *ssh.Runner, ocConfig oc.Config, proxy *network.ProxyConfig) error {
+	type trustedCA struct {
+		Name string `json:"name"`
+	}
+
+	type proxySpecConfig struct {
+		HTTPProxy  string    `json:"httpProxy"`
+		HTTPSProxy string    `json:"httpsProxy"`
+		NoProxy    string    `json:"noProxy"`
+		TrustedCA  trustedCA `json:"trustedCA"`
+	}
+
+	type patchSpec struct {
+		Spec proxySpecConfig `json:"spec"`
+	}
+
+	patch := &patchSpec{
+		Spec: proxySpecConfig{
+			HTTPProxy:  proxy.HTTPProxy,
+			HTTPSProxy: proxy.HTTPSProxy,
+			NoProxy:    proxy.GetNoProxyString(),
+		},
+	}
 
 	if err := WaitForOpenshiftResource(ocConfig, "proxy"); err != nil {
 		return err
 	}
+
+	if proxy.ProxyCACert != "" {
+		trustedCAName := "user-ca-bundle"
+		logging.Debug("Adding proxy CA cert to cluster")
+		if err := addProxyCACertToCluster(sshRunner, ocConfig, proxy, trustedCAName); err != nil {
+			return err
+		}
+		patch.Spec.TrustedCA = trustedCA{Name: trustedCAName}
+	}
+
+	patchEncode, err := json.Marshal(patch)
+	if err != nil {
+		return fmt.Errorf("Failed to encode to json: %v", err)
+	}
+	logging.Debugf("Patch string %s", string(patchEncode))
+
+	cmdArgs := []string{"patch", "proxy", "cluster", "-p", fmt.Sprintf("'%s'", string(patchEncode)), "-n", "openshift-config", "--type", "merge"}
 	if _, stderr, err := ocConfig.RunOcCommand(cmdArgs...); err != nil {
 		return fmt.Errorf("Failed to add proxy details %v: %s", err, stderr)
+	}
+	return nil
+}
+
+func addProxyCACertToCluster(sshRunner *ssh.Runner, ocConfig oc.Config, proxy *network.ProxyConfig, trustedCAName string) error {
+	proxyConfigMapFileName := fmt.Sprintf("/tmp/%s.json", trustedCAName)
+	proxyCABundleTemplate := `{
+  "apiVersion": "v1",
+  "data": {
+    "ca-bundle.crt": "%s"
+  },
+  "kind": "ConfigMap",
+  "metadata": {
+    "name": "%s",
+    "namespace": "openshift-config"
+  }
+}
+`
+	// Replace the carriage return with `\n` char
+	p := fmt.Sprintf(proxyCABundleTemplate, strings.ReplaceAll(proxy.ProxyCACert, "\n", `\n`), trustedCAName)
+	err := sshRunner.CopyData([]byte(p), proxyConfigMapFileName, 0644)
+	if err != nil {
+		return err
+	}
+	cmdArgs := []string{"apply", "-f", proxyConfigMapFileName}
+	if _, stderr, err := ocConfig.RunOcCommand(cmdArgs...); err != nil {
+		return fmt.Errorf("Failed to add proxy cert details %v: %s", err, stderr)
 	}
 	return nil
 }
@@ -151,6 +215,21 @@ Environment=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,%s`
 	}
 	err = sshRunner.CopyData([]byte(p), "/etc/systemd/system/kubelet.service.d/10-default-env.conf", 0644)
 	if err != nil {
+		return err
+	}
+
+	if proxy.ProxyCACert != "" {
+		logging.Debug("Adding proxy CA cert to instance")
+		return addProxyCACertToInstance(sshRunner, proxy)
+	}
+	return nil
+}
+
+func addProxyCACertToInstance(sshRunner *ssh.Runner, proxy *network.ProxyConfig) error {
+	if err := sshRunner.CopyData([]byte(proxy.ProxyCACert), "/etc/pki/ca-trust/source/anchors/openshift-config-user-ca-bundle.crt", 0600); err != nil {
+		return err
+	}
+	if _, err := sshRunner.Run("sudo update-ca-trust"); err != nil {
 		return err
 	}
 	return nil
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -795,7 +795,7 @@ func configProxyForCluster(ocConfig oc.Config, sshRunner *crcssh.Runner, sd *sys
 
 	logging.Info("Adding proxy configuration to the cluster ...")
 	proxy.AddNoProxy(instanceIP)
-	if err := cluster.AddProxyConfigToCluster(ocConfig, proxy); err != nil {
+	if err := cluster.AddProxyConfigToCluster(sshRunner, ocConfig, proxy); err != nil {
 		return err
 	}
 
--- a/pkg/crc/network/proxy.go
+++ b/pkg/crc/network/proxy.go
@@ -15,15 +15,17 @@ var (
 
 // ProxyConfig keeps the proxy configuration for the current environment
 type ProxyConfig struct {
-	HTTPProxy  string
-	HTTPSProxy string
-	noProxy    []string
+	HTTPProxy   string
+	HTTPSProxy  string
+	noProxy     []string
+	ProxyCACert string
 }
 
-func NewProxyDefaults(httpProxy, httpsProxy, noProxy string) (*ProxyConfig, error) {
+func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCACert string) (*ProxyConfig, error) {
 	DefaultProxy = ProxyConfig{
-		HTTPProxy:  httpProxy,
-		HTTPSProxy: httpsProxy,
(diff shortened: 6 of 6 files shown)


cmd/crc/cmd/root.go
```diff
@@ -96,15 +98,26 @@ func setProxyDefaults() {
...
+
+	// If proxyCAFile is not empty string then read the content and pass it to NewProxyDefaults
+	var proxyCACert []byte
+	if proxyCAFile != "" {
+		var err error
+		proxyCACert, err = ioutil.ReadFile(proxyCAFile)
+		if err != nil {
+			exit.WithMessage(1, fmt.Sprintf("not able to read proxyCAFile %s: %v", proxyCAFile, err.Error()))
+		}
+	}
```
cfergeau, line 111: I'd go with a `getProxyCAData` helper rather than having this low-level code here, + the trim/conversion to string in the `NewProxyDefault` call.

cmd/crc/cmd/root.go
```diff
@@ -96,15 +98,26 @@ func setProxyDefaults() {
 	httpProxy := config.GetString(cmdConfig.HTTPProxy.Name)
 	httpsProxy := config.GetString(cmdConfig.HTTPSProxy.Name)
 	noProxy := config.GetString(cmdConfig.NoProxy.Name)
+	proxyCAFile := config.GetString(cmdConfig.ProxyCAFile.Name)
+
+	// If proxyCAFile is not empty string then read the content and pass it to NewProxyDefaults
```
cfergeau, line 103: `is not an empty string` or `is not empty`

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +119,56 @@ func UpdateClusterID(ocConfig oc.Config) error {
 	return nil
 }
 
-func AddProxyConfigToCluster(ocConfig oc.Config, proxy *network.ProxyConfig) error {
+func AddProxyConfigToCluster(sshRunner *ssh.Runner, ocConfig oc.Config, proxy *network.ProxyConfig) error {
+	trustedCAName := "user-ca-bundle"
```
cfergeau, line 123: This could go in the `if proxy.ProxyCACert != "" {` block

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +119,56 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
 	}
+	if proxy.ProxyCACert != "" {
+		logging.Debug("Adding proxy CA cert to cluster")
+		if err := addProxyCACertToCluster(sshRunner, ocConfig, proxy, trustedCAName); err != nil {
+			return err
+		}
+		cmdArgs = []string{"patch", "proxy", "cluster", "-p",
+			fmt.Sprintf(`'{"spec":{"httpProxy":"%s", "httpsProxy":"%s", "noProxy":"%s", "trustedCA":{"name":"%s"}}}'`,
+				proxy.HTTPProxy, proxy.HTTPSProxy, proxy.GetNoProxyString(), trustedCAName),
+			"-n", "openshift-config", "--type", "merge"}
```
cfergeau, line 139: Let's not duplicate this. Maybe this is doable with a `struct` serialized to json rather than a `fmt.Sprintf` in order to handle the serialization of the optional `trustedCA` more easily?

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +119,56 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
+  },
+  "kind": "ConfigMap",
+  "metadata": {
+    "name": "%s",
+    "namespace": "openshift-config"
+  }
+}
+`
+	p := fmt.Sprintf(proxyCABundleTemplate, strings.ReplaceAll(proxy.ProxyCACert, "\n", `\n`), trustedCAName)
+	err := sshRunner.SetTextContentAsRoot(fmt.Sprintf("/tmp/%s.json", trustedCAName), p, 0644)
```
cfergeau, line 161: This method is gone, it's `CopyData` now

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +119,56 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
+	if err != nil {
+		return err
+	}
+	cmdArgs := []string{"apply", "-f", fmt.Sprintf("/tmp/%s.json", trustedCAName)}
```
cfergeau, line 165: `tmpFilename := fmt.Sprintf("/tmp/%s.json", trustedCAName)` and then use that. Maybe go with `filepath.Join()` and `filepath.ToSlash()`?

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +119,56 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
+	if _, stderr, err := ocConfig.RunOcCommand(cmdArgs...); err != nil {
+		return fmt.Errorf("Failed to add proxy cert details %v: %s", err, stderr)
+	}
```
cfergeau, line 168: This would almost deserve a helper `ocConfig.Apply(jsonFile string)` if this is commonly needed. An intermediate file was needed because it's not convenient to pass the json data on the command line?

pkg/crc/cluster/cluster.go
```diff
@@ -153,6 +189,21 @@ Environment=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,%s`
...
+
+	if proxy.ProxyCACert != "" {
+		logging.Debug("Adding proxy CA cert to instance")
+		return addProxyCACertToInstance(sshRunner, proxy)
+	}
+	return nil
+}
+
+func addProxyCACertToInstance(sshRunner *ssh.Runner, proxy *network.ProxyConfig) error {
+	if err := sshRunner.SetTextContentAsRoot("/etc/pki/ca-trust/source/anchors/openshift-config-user-ca-bundle.crt", proxy.ProxyCACert, 0600); err != nil {
```
cfergeau, line 201: We do this manually because this should be done by the machine-config-operator, which we are not running?

pkg/crc/machine/machine.go
```diff
@@ -801,7 +801,7 @@ func configProxyForCluster(ocConfig oc.Config, sshRunner *crcssh.Runner, sd *sys
 
 	logging.Info("Adding proxy configuration to the cluster ...")
 	proxy.AddNoProxy(instanceIP)
-	if err := cluster.AddProxyConfigToCluster(ocConfig, proxy); err != nil {
+	if err := cluster.AddProxyConfigToCluster(sshRunner, ocConfig, proxy); err != nil {
```
cfergeau, line 804: `sshRunner` and `ocConfig` are slightly redundant, wondering if we should create the `ocConfig` 'on demand' in the cluster package, or have a way to get the ssh runner from the `ocConfig`? Thinking out loud without spending much time on the existing code, maybe none of these suggestions make sense :)

cmd/crc/cmd/root.go
```diff
@@ -96,15 +98,26 @@ func setProxyDefaults() {
...
+	proxyConfig, err := network.NewProxyDefaults(httpProxy, httpsProxy, noProxy, strings.TrimRight(string(proxyCACert), "\r\n"))
 	if err != nil {
 		exit.WithMessage(1, err.Error())
 	}
 
 	if proxyConfig.IsEnabled() {
-		logging.Debugf("HTTP-PROXY: %s, HTTPS-PROXY: %s, NO-PROXY: %s", proxyConfig.HTTPProxyForDisplay(),
-			proxyConfig.HTTPSProxyForDisplay(), proxyConfig.GetNoProxyString())
+		logging.Debugf("HTTP-PROXY: %s, HTTPS-PROXY: %s, NO-PROXY: %s, proxyCACert: %s", proxyConfig.HTTPProxyForDisplay(),
+			proxyConfig.HTTPSProxyForDisplay(), proxyConfig.GetNoProxyString(), proxyConfig.ProxyCACert)
```
cfergeau, line 120: Isn't this the ProxyCACert too verbose for debug output? Maybe it's good enough to print only the filename *if we managed to read it successfully*?

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +119,56 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
+    "ca-bundle.crt": "%s"
+  },
+  "kind": "ConfigMap",
+  "metadata": {
+    "name": "%s",
+    "namespace": "openshift-config"
+  }
+}
+`
+	p := fmt.Sprintf(proxyCABundleTemplate, strings.ReplaceAll(proxy.ProxyCACert, "\n", `\n`), trustedCAName)
```
cfergeau, line 160: slightly confused here? is this replacing actual carriage returns with the caracters `\n` ? or is this doing something else?

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau That's right

cmd/crc/cmd/root.go
praveenkumar: @cfergeau yes, it was my own debug purpose, have to add the filename only.

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau yes, I am adding same as comment.

praveenkumar:
> Regarding the steps you give for testing these changes, are they going to be added to the integration tests?

@cfergeau not right now, since we already now have basic proxy test in place. if you folks agree then we can extend those to cover cert case also.

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +120,82 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
+		patch.Spec.TrustedCA = trustedCA{Name: trustedCAName}
+	}
+
+	patchEncode, err := json.Marshal(patch)
+	if err != nil {
+		return fmt.Errorf("Failed to encode to json: %v", err)
+	}
+	logging.Debugf("Patch string %s", string(patchEncode))
+
+	cmdArgs := []string{"patch", "proxy", "cluster", "-p", fmt.Sprintf("'%s'", string(patchEncode)), "-n", "openshift-config", "--type", "merge"}
```
cfergeau, line 166: Is the `fmt.Sprintf` required?

cmd/crc/cmd/config/config.go
```diff
@@ -23,9 +23,10 @@ var (
```
cfergeau, line 23: `5 files changed, 120 insertions(+), 19 deletions(-)`
Better to have more than one line/60 chars to describe such changes :) If something is listed in "List main as well as consequential changes you introduced or had to introduce." in the PR description, it needs to appear in the commit log as well.

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +120,82 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
-func AddProxyConfigToCluster(ocConfig oc.Config, proxy *network.ProxyConfig) error {
-	cmdArgs := []string{"patch", "proxy", "cluster", "-p",
-		fmt.Sprintf(`'{"spec":{"httpProxy":"%s", "httpsProxy":"%s", "noProxy":"%s"}}'`, proxy.HTTPProxy, proxy.HTTPSProxy, proxy.GetNoProxyString()),
-		"-n", "openshift-config", "--type", "merge"}
+func AddProxyConfigToCluster(sshRunner *ssh.Runner, ocConfig oc.Config, proxy *network.ProxyConfig) error {
+	type trustedCA struct {
+		Name string `json:"name"`
+	}
+
+	type proxyPatch struct {
```
cfergeau, line 128: I would name this `proxyConfig`...

pkg/crc/cluster/cluster.go
```diff
@@ -119,20 +120,82 @@ func UpdateClusterID(ocConfig oc.Config) error {
...
+		HTTPProxy  string    `json:"httpProxy"`
+		HTTPSProxy string    `json:"httpsProxy"`
+		NoProxy    string    `json:"noProxy"`
+		TrustedCA  trustedCA `json:"trustedCA"`
+	}
+
+	type spec struct {
```
cfergeau, line 135: ... and this `patchSpec` if you want to have the name `patch` in one of the types.

cmd/crc/cmd/root.go
```diff
@@ -96,15 +98,32 @@ func setProxyDefaults() {
...
+
+func getProxyCAData(proxyCAFile string) (string, error) {
+	if proxyCAFile == "" {
+		return "", nil
+	}
+	proxyCACert, err := ioutil.ReadFile(proxyCAFile)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimRight(string(proxyCACert), "\r\n"), nil
```
cfergeau, line 128: Not clear what the `strings.TrimRight` call is for? Is this Windows specific?

pkg/crc/cluster/cluster.go
cfergeau: The name of the tmp file is still generated twice. I would hardcode `user-ca-bundle` in `addProxyCACertToCluster`, no need for it to be an argument, as OpenShift expects it to be named this way if I'm not mistaken, so if you want to add a CA cert to the cluster, you have to use the correct `user-ca-bundle` name.

cfergeau:
Can't  find again Guillaume's comment, but would definitely be nice to have an automated test for the proxy CA :)

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau yes because we are executing command as ssh runner so it should be passed as `'<json_string>'` otherwise you will get error about json patch not valid.

pkg/crc/cluster/cluster.go
praveenkumar: there is already `ProxyConfig` struct in network so I don't want to have any misunderstanding it here, also struct is to create patch for proxy config so I OK with current naming.

pkg/crc/cluster/cluster.go
praveenkumar: yes `patchSpec` make more sense :+1:

cmd/crc/cmd/root.go
praveenkumar: In windows if user open the cert file in notepad or any windows application it does add the `\r\n` trailing so this will just to add extra layer on our side to get the cert data in correct way.

cmd/crc/cmd/config/config.go
praveenkumar: @cfergeau we are just adding a feature, I am not sure what I can add as part of description :confused:

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau https://docs.openshift.com/container-platform/4.5/networking/enable-cluster-wide-proxy.html#nw-proxy-configure-object_config-cluster-wide-proxy it says `user-ca-bundle` but name can be anything because it's the configmap name which we need to match here, about twice generating the `/tmp` file I can move it to top block of function.

pkg/crc/machine/machine.go
praveenkumar: Not right now because the interface we use have only 2 methods so we can support ocConfig from the host or from the VM, we can change it if we are moving away with host oc config but that would be a different refactor.

praveenkumar:
> , no need for it to be an argument, as OpenShift expects it to be named this way if I'm not mistaken, so if you want to add a CA cert to the cluster, you have to use the correct user-ca-bundle name.

@cfergeau not the name is just configmap name and it can be different as per doc it just uses `user-ca-bundle` https://docs.openshift.com/container-platform/4.5/networking/enable-cluster-wide-proxy.html#nw-proxy-configure-object_config-cluster-wide-proxy

pkg/crc/cluster/cluster.go
cfergeau: oh I did not noticed the added quotes ^^

cmd/crc/cmd/config/config.go
cfergeau: what it does, how to use it, ... (which reminds me this PR is missing a docs change)

pkg/crc/cluster/cluster.go
cfergeau: `specProxyConfig` maybe? I don't think this format would be only valid for `patch`? For example if we parse what `oc get` returns, we'd use the same struct?

cmd/crc/cmd/root.go
cfergeau: I'd move that to a helper at least, and probably add a comment that it's mostly needed on Windows. Why is this removing them completely rather than changing them to `\n`? If the removal is desirable, why not remove `\n` on linux as well?

cmd/crc/cmd/config/config.go
praveenkumar: @cfergeau yes, we need to add the doc for it.

pkg/crc/cluster/cluster.go
praveenkumar: if `oc get proxy` then you can use same struct to extract the `spec` data or may be other resource which have same kind of data as part of spec, but I can go with `proxySpecConfig`

cmd/crc/cmd/root.go
praveenkumar: @cfergeau It does remove the endline for linux and windows both, I will put a comment, shouldn't need a helper function just to perform string operation using standard strings package. https://play.golang.org/p/0gKkIhLLxij

cfergeau: > It does remove the endline for linux and windows both

Ah, right, "\r\n" is a list of characters to remove, not a substring to remove.

> shouldn't need a helper function just to perform string operation using standard strings package.

This makes the code much easier to read.
Compare this line of thought:

`strings.TrimRight(string(proxyCACert), "\r\n")` What is `TrimRight` doing again? Ah yes, removing characters from a string. Ok these characters are "\r\n" which are end of line characters, so it's removing the end of lines from the string (and this assumes the reader knows that \r is used for end of line on Windows, and \n on all platforms)

with:

`removeLineEndings(string(proxyCACert))` ok, they remove the end of line markers from the string

docs/source/topics/proc_starting-codeready-containers-behind-proxy.adoc
```diff
@@ -22,6 +22,16 @@ $ {bin} config set https-proxy http://example.proxy.com:__<port>__
...
+
+. If proxy uses custom CA then use the `{bin} config set` command as follows:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config set proxy-ca-file __<Path to custom CA file>__
+----
++
+
+
```
cfergeau, line 34: @robin-owen small doc change for you to take a look at

pkg/crc/cluster/cluster.go
cfergeau: But all line endings were removed by `getProxyCAData`, so is there anything left to replace?

cmd/crc/cmd/root.go
cfergeau: Wondering if removing these line endings is the right thing to do. https://serverfault.com/questions/466683/can-an-ssl-certificate-be-on-a-single-line-in-a-file-no-line-breaks says not to remove line endings from a PEM file (which is what openshift wants)

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau not all, it only remove the end of file line not intermediate one.

cmd/crc/cmd/root.go
praveenkumar: @cfergeau this is not removing the intermediate end line `\n` but only remove if the file end have more new lines something like.
```
-----BEGIN CERTIFICATE-----
Zc8uiGg99pmd008E2cmGErqNzPclK9INbPjwxl9zIgTcXpF/UtBVVS1Z7Xo83qfs
GMPdMzYt3F+hQpQYLhlGF+5Jf2x6Zb1zjT/aM3GMdGi+6OPV+YHl/wgUe45N6kRu
DZnVL167+W3l2nD+mShO/7xqwHiZuz0GHyBHRp5i43blH0vg67sJ8guN81papupY
2v78FcvR8j0ELfgyehtWpjF3vzKSq/rY2sMXTYzSPqMeksseHNhSMYU6Ww9h9pyM
aVnw9vahqf7nKHHcC2VRTUgkQfn9yDmmBOo0nQ8Xgfpd65/PaxVfBnuKfEkXBfpM
-----END CERTIFICATE-----
<multiple new lines here>
```

cfergeau: Ah right, sorry again, I really should read carefully what `TrimRight` is doing :( I would still use a helper for this (trimTrailingEOL) to make the code self documenting..
One more question: it seems we assume we may get `\r\n` at the end of the file, but that all other EOL must be `\n`? I'd personally expect that we don't deal with `\r` at all, or that we deal with it everywhere (replace `\r\n` with `\n`, and then remove trailing `\n`

praveenkumar: @cfergeau the assumption we are having here is at the file end only user might add new lines by mistake and not in between so default endline now doesn't change to `\r\n` only the newly added on if any for windows.

cfergeau: I'd go for "all or nothing", either we remove all `\r` + trailing new lines, or we don't remove any `\r`. Handling them in some specific cases but not others is unexpected

cfergeau (approved):

PR #1450, opened 2020-08-10
Add ssh test

guillaumerose:
It tests native and external connection.


cfergeau:
> It tests native and external connection.

I would have expected to have this in the commit log.

guillaumerose:
I updated the commit message. Waiting for CI and I merge it, thanks for the review!

PR #1451, opened 2020-08-10
Merge libvirt driver in crc repository

guillaumerose:
Preliminary step for a bigger simplification of libmachine code.

It adds a new binary in this repo. This binary is compiled when doing `make cross`, `make install` or `make embed_bundle`. I had to introduce a variation of the embedder to just embed local files.

I am not really convinced about the file:// scheme. If you have an other idea, I take it :)


pkg/drivers/libvirt/libvirt_nonlinux.go
```diff
@@ -0,0 +1,12 @@
...
+
+package libvirt
+
+import (
+	"github.com/code-ready/machine/drivers/errdriver"
+	"github.com/code-ready/machine/libmachine/drivers"
+)
+
+func NewDriver(hostName, storePath string) drivers.Driver {
+	return &errdriver.Driver{}
```
praveenkumar, line 11: @guillaumerose is it there just because if someone use `libvirt.NewDriver` in case of mac/windows it will return the error?

guillaumerose: yes it was a bad idea. It was only for the linter to be able to compile. I removed it.

cfergeau:
> It adds a new binary in this repo. This binary is compiled when doing `make cross`, `make install` or `make embed_bundle`. I had to introduce a variation of the embedder to just embed local files.

Not clear this binary is really required, it may be enough for the libvirt driver to make sync calls to the driver implementation rather than going through a separate binary.

guillaumerose:
It would imply that crc is statically compiled and therefore it won't run if libvirt is not installed on the system.

cfergeau:
> It would imply that crc is statically compiled and therefore it won't run if libvirt is not installed on the system.

Ah right, we could try using https://golang.org/src/plugin/plugin.go to be able to report better errors if libvirt is not installed, but that's more work :)

cfergeau:
> I am not really convinced about the file:// scheme. If you have an other idea, I take it :)

`crc` needs to know that the libvirt machine driver was locally provided so that it errors out.
`crc-embedder` needs to know where the libvirt machine driver is so that it can be embedded in the bundle.

I would separate these 2 needs, and use some magic value for `MachineDriverDownloadURL` (could be `local:///` for example), and pass the correct path to the libvirt driver to `crc-embedder`
Also note that building crc-libvirt-driver as part of crc, and making binaries available online are not necessarily exclusive, is it?

cfergeau:
Is the first commit a copy of some files from https://github.com/code-ready/machine-driver-libvirt/ to the crc repository? Or are more changes needed? Having what you did explained in the commit log makes things much easier to look at :)

guillaumerose:
Giving up for the moment.

PR #1452, opened 2020-08-10
Use shared types between libvirt/hyperkit machine drivers and crc

cfergeau:
Use shared types between libvirt/hyperkit machine drivers and crc. This should not have any user-visible changes over crc use, so nothing specific to test.


praveenkumar:
@cfergeau https://github.com/code-ready/crc/pull/1452/commits/5fc9083e8dcc5ac8ec4808bd17bbd60ca049eb04 and https://github.com/code-ready/crc/pull/1452/commits/fbc8b36395a833bd71969984056266e39f00cb75 commits should be squash to single commit because it make sense to update the vendor directory also when you update the go mod file otherwise one commit able to compile when uses vendor and other doesn't.

cfergeau:
> @cfergeau [5fc9083](https://github.com/code-ready/crc/commit/5fc9083e8dcc5ac8ec4808bd17bbd60ca049eb04) and [fbc8b36](https://github.com/code-ready/crc/commit/fbc8b36395a833bd71969984056266e39f00cb75) commits should be squash to single commit because it make sense to update the vendor directory also when you update the go mod file otherwise one commit able to compile when uses vendor and other doesn't.

I know, but it also makes the git diff very messy. However, since 'Use platform driver types from machine module' has the go.mod/go.sum changes, I believe this will build just fine though, it is just `make vendorcheck` and similar targets which will fail before the "update vendor dir" commit, which is an acceptable trade-off in my opinion.

praveenkumar:
> Even rm -rf vendor/ && make cross works fine on my machine. So maybe it got cached somewhere, or maybe the vendor dir is really only there as a failsafe.

@cfergeau by default from Go-1.11 vendor director is ignored because of go module, if you want to check the vendor stuff then you need to use `GO111MODULE=off make cross` https://golang.org/cmd/go/#hdr-Module_support (make sure you put the code under `$GOPATH/src/github.com/code-ready/crc`

praveenkumar:
> I know, but it also makes the git diff very messy. However, since 'Use platform driver types from machine module' has the go.mod/go.sum changes,

@cfergeau yes and that's my worry as I communicated on the slack, one commit only have the go mod update and second have the vendor update which is because of first commit so even it is messy but it would be easy when you revert or want to know which version of machine code we updated to have this vendor directory update.

cfergeau:
> > I know, but it also makes the git diff very messy. However, since 'Use platform driver types from machine module' has the go.mod/go.sum changes,
> 
> @cfergeau yes and that's my worry as I communicated on the slack, one commit only have the go mod update and second have the vendor update which is because of first commit so even it is messy but it would be easy when you revert or want to know which version of machine code we updated to have this vendor directory update.

If you revert without syncing the vendor directory, CI will fail and tell you about it. Even if you forget to update it, I don't think it will cause issues since by default it will fetch the modules? Matching vendor/ update with module version change will indeed be a bit more complicated, but it's not going to be terribly difficult, nor is something which needs to be done often?

Issue #1449, opened 2020-08-10
crc setup fails with setting up KVM on vanilla CentOS 7


cg2p:
### General information

  * OS: Linux 
  * Hypervisor: virtual private server (antsle) / LXC
  * Did you run `crc setup` before starting it (Yes/No)? No - attempting the setup process

## CRC version
```bash
CodeReady Containers version: 1.14.0+36ad776
OpenShift version: 4.5.4 (embedded in binary)
```
  
## CRC status
```bash
Machine 'crc' does not exist. Use 'crc start' to create it
```

## CRC config
```bash
```

## Host Operating System
```bash
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

### Steps to reproduce
1. as username who has sudo privileges
```
# vi /etc/sudoers
# add
<username>  ALL=(ALL) NOPASSWD:ALL
```
2. crc setup 
 

### Expected
Setup works

### Actual
crc setup fails
```
INFO Checking if oc binary is cached              
INFO Checking if podman remote binary is cached   
INFO Checking if goodhosts binary is cached       
INFO Checking if CRC bundle is cached in '$HOME/.crc' 
INFO Checking minimum RAM requirements            
INFO Checking if running as non-root              
INFO Checking if Virtualization is enabled        
INFO Checking if KVM is enabled                   
INFO Setting up KVM                               
FATA Failed to load kvm module  
```

cfergeau:
Can you post the output of `crc setup --log-level debug` so that the actual failure can be seen? Can you make sure that virtualization is not disabled in your bios? `dmesg` will have a line about it if this is the reason for the failure.

cg2p:
```
DEBU CodeReady Containers version: 1.14.0+36ad776 
DEBU OpenShift version: 4.5.4 (embedded in binary) 
INFO Checking if oc binary is cached              
[...]

PR #1454, opened 2020-08-11
Fix loading of kvm/kvm_intel/kvm_amd modules

praveenkumar:
**Fixes:** Issue #1449 

## Solution/Idea
Enable the kvm_intel and kvm_amd kernel module according to processor.

## Testing

With current master following will fail to start the crc
```
$ sudo modprobe -r kvm_intel
$ $ lsmod | grep kvm
$ crc setup
DEBU Trying to load kvm module                    
DEBU exit status 1 : modprobe: ERROR: could not insert 'kvm': Operation not permitted 
FATA Failed to load kvm module
```

If you manually execute the `sudo modprobe kvm` then the start will fail
```
$ sudo modprobe kvm
$ $ lsmod | grep kvm
kvm                   815104  0
irqbypass              16384  1 kvm
$ ./crc setup
$ crc start
...
DEBU Checking if /dev/kvm exists                  
DEBU kvm kernel module is not loaded              
FATA kvm kernel module is not loaded  
```

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -80,7 +80,7 @@ func checkKvmEnabled() error {
 
 func fixKvmEnabled() error {
 	logging.Debug("Trying to load kvm module")
-	stdOut, stdErr, err := crcos.RunWithDefaultLocale("modprobe", "kvm")
+	stdOut, stdErr, err := crcos.RunWithPrivilege("Load kvm kernel module","modprobe", "kvm")
 	if err != nil {
 		return fmt.Errorf("Failed to load kvm module: %s %v: %s", stdOut, err, stdErr)
 	}
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -42,19 +42,12 @@ dns=dnsmasq
 func checkVirtualizationEnabled() error {
 	logging.Debug("Checking if the vmx/svm flags are present in /proc/cpuinfo")
 	// Check if the cpu flags vmx or svm is present
-	out, err := ioutil.ReadFile("/proc/cpuinfo")
+	flags, err := getCPUFlags()
 	if err != nil {
-		logging.Debugf("Failed to read /proc/cpuinfo: %v", err)
-		return fmt.Errorf("Failed to read /proc/cpuinfo")
-	}
-	re := regexp.MustCompile(`flags.*:.*`)
-
-	flags := re.FindString(string(out))
-	if flags == "" {
-		return fmt.Errorf("Could not find cpu flags from /proc/cpuinfo")
+		return err
 	}
 
-	re = regexp.MustCompile(`(vmx|svm)`)
+	re := regexp.MustCompile(`(vmx|svm)`)
 
 	cputype := re.FindString(flags)
 	if cputype == "" {
@@ -80,10 +73,26 @@ func checkKvmEnabled() error {
 
 func fixKvmEnabled() error {
 	logging.Debug("Trying to load kvm module")
-	stdOut, stdErr, err := crcos.RunWithPrivilege("Load kvm kernel module","modprobe", "kvm")
+	flags, err := getCPUFlags()
 	if err != nil {
-		return fmt.Errorf("Failed to load kvm module: %s %v: %s", stdOut, err, stdErr)
+		return err
 	}
+
+	switch {
+	case strings.Contains(flags, "vmx"):
+		stdOut, stdErr, err := crcos.RunWithPrivilege("Load kvm_intel kernel module", "modprobe", "kvm_intel")
+		if err != nil {
+			return fmt.Errorf("Failed to load kvm intel module: %s %v: %s", stdOut, err, stdErr)
+		}
+	case strings.Contains(flags, "svm"):
+		stdOut, stdErr, err := crcos.RunWithPrivilege("Load kvm_amd kernel module", "modprobe", "kvm_amd")
+		if err != nil {
+			return fmt.Errorf("Failed to load kvm amd module: %s %v: %s", stdOut, err, stdErr)
+		}
+	default:
+		logging.Debug("Unable to detect processor details")
+	}
+
 	logging.Debug("kvm module loaded")
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -624,3 +626,19 @@ func getCurrentLibvirtDriverVersion() (string, error) {
...
+func getCPUFlags() (string, error) {
+	// Check if the cpu flags vmx or svm is present
+	out, err := ioutil.ReadFile("/proc/cpuinfo")
+	if err != nil {
+		logging.Debugf("Failed to read /proc/cpuinfo: %v", err)
+		return "", fmt.Errorf("Failed to read /proc/cpuinfo")
+	}
+	re := regexp.MustCompile(`flags.*:.*`)
+
+	flags := re.FindString(string(out))
```
guillaumerose, line 639: Maybe use this library https://godoc.org/github.com/shirou/gopsutil/cpu#InfoStat ?

praveenkumar: @guillaumerose this is what I also suggested to @cfergeau but since our usecase is not that much to include this complete library as dependency so I went without it.

guillaumerose: if we merge https://github.com/code-ready/crc/pull/1396, we will have this library included :angel:

praveenkumar: @guillaumerose Ah, may be we will merge it soon :)

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -80,13 +80,9 @@ func checkKvmEnabled() error {
...
 	logging.Debug("Trying to load kvm module")
-	cmd := exec.Command("modprobe", "kvm")
-	buf := new(bytes.Buffer)
-	cmd.Stderr = buf
-	err := cmd.Run()
+	stdOut, stdErr, err := crcos.RunWithDefaultLocale("modprobe", "kvm")
 	if err != nil {
-		logging.Debugf("%v : %s", err, buf.String())
-		return fmt.Errorf("Failed to load kvm module")
+		return fmt.Errorf("Failed to load kvm module: %s %v: %s", stdOut, err, stdErr)
```
cfergeau, line 78: Fwiw, what would be nice would be to define an `RunError` type
```
type RunError struct {
    stdOut string
    stdErr string
    err Error
}
```
and make use of `return fmt.Errorf("Failed to load kvm module: %w", runError)
(see https://blog.golang.org/go1.13-errors )

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -308,13 +304,11 @@ func fixLibvirtCrcNetworkAvailable() error {
 	// For time being we are going to override the crc network according what we have in our binary template.
 	// We also don't care about the error or output from those commands atm.
 	// #nosec G204
-	cmd := exec.Command("virsh", "--connect", "qemu:///system", "net-destroy", libvirt.DefaultNetwork)
-	_ = cmd.Run()
+	_, _ , _ = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-destroy", libvirt.DefaultNetwork)
 	// #nosec G204
-	cmd = exec.Command("virsh", "--connect", "qemu:///system", "net-undefine", libvirt.DefaultNetwork)
-	_ = cmd.Run()
+	_,_,_ = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-undefine", libvirt.DefaultNetwork)
```
cfergeau: Can you make the spacing consistent with the previous call?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -80,10 +73,30 @@ func checkKvmEnabled() error {
 
 func fixKvmEnabled() error {
 	logging.Debug("Trying to load kvm module")
-	stdOut, stdErr, err := crcos.RunWithPrivilege("Load kvm kernel module","modprobe", "kvm")
+	stdOut, stdErr, err := crcos.RunWithPrivilege("Load kvm kernel module", "modprobe", "kvm")
```
cfergeau, line 76: Can you move this to the initial commit, since this is when the missing space was introduced.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -304,9 +317,9 @@ func fixLibvirtCrcNetworkAvailable() error {
 	// For time being we are going to override the crc network according what we have in our binary template.
 	// We also don't care about the error or output from those commands atm.
 	// #nosec G204
-	_, _ , _ = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-destroy", libvirt.DefaultNetwork)
+	_, _, _ = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-destroy", libvirt.DefaultNetwork)
 	// #nosec G204
-	_,_,_ = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-undefine", libvirt.DefaultNetwork)
+	_, _, _ = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "net-undefine", libvirt.DefaultNetwork)
```
cfergeau, line 322: Ah, this belongs to the previous commit where I asked to make the spacing consistent :+1:

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau I think then this should go as part of runCmd where we can wrap `err` and `stderr` as part of struct.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -42,19 +42,12 @@ dns=dnsmasq
...
-	re := regexp.MustCompile(`flags.*:.*`)
-
-	flags := re.FindString(string(out))
-	if flags == "" {
-		return fmt.Errorf("Could not find cpu flags from /proc/cpuinfo")
+		return err
 	}
 
-	re = regexp.MustCompile(`(vmx|svm)`)
+	re := regexp.MustCompile(`(vmx|svm)`)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I then need to use 2 different `strings.Contains` since it doesn't allow the regex :(

PR #1455, opened 2020-08-11
Check bundle presence by looking at bindata

guillaumerose:
Small simplification.

I think I will also open a PR to introduce a function that validates embedded data (all files are present, etc).


guillaumerose:
@praveenkumar is this PR ok for you ? just to be sure.

Issue #1175, opened 2020-04-20
CRC start needs admin permissions mid script


jlynchMicron:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? Yes

## CRC version
```bash
# Put the output of `crc version`
crc version: 1.8.0+0a318dc
OpenShift version: 4.3.8 (embedded in binary)
```
  
## CRC status
```bash
# Put the output of `crc status`
ERRO error: CreateFile C:\Users\jlynch\.crc\machines\crc\kubeconfig: The system cannot find the file specified.
 - exit status 1
```

## CRC config
```bash
# Put the output of `crc config view`
- http-proxy                            : ### corp proxy ###
- https-proxy                           : ### corp proxy ###
```

## Host Operating System
```bash
Host Name:                 JLYNCH10
OS Name:                   Microsoft Windows 10 Enterprise
OS Version:                10.0.17763 N/A Build 17763
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Member Workstation
OS Build Type:             Multiprocessor Free
Registered Owner:          Micron
Registered Organization:   Micron Technology, Inc.
Product ID:                00329-00000-00003-AA175
Original Install Date:     11/12/2019, 12:22:10 PM
System Boot Time:          4/20/2020, 2:26:03 PM
System Manufacturer:       HP
System Model:              HP EliteDesk 800 G3 SFF
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 94 Stepping 3 GenuineIntel ~3192 Mhz
BIOS Version:              HP P01 Ver. 02.31, 10/15/2019
Windows Directory:         C:\WINDOWS
System Directory:          C:\WINDOWS\system32
Boot Device:               \Device\HarddiskVolume2
System Locale:             en-us;English (United States)
Input Locale:              en-us;English (United States)
Time Zone:                 (UTC-07:00) Mountain Time (US & Canada)
Total Physical Memory:     32,647 MB
Available Physical Memory: 17,104 MB
Virtual Memory: Max Size:  37,511 MB
Virtual Memory: Available: 20,028 MB
Virtual Memory: In Use:    17,483 MB
Page File Location(s):     C:\pagefile.sys
Domain:                    na.micron.com
Logon Server:              \\WDCNABO3
Hotfix(s):                 11 Hotfix(s) Installed.
[...]

PR #1457, opened 2020-08-11
Issue #1175 Clarify required privilege elevation in docs

robin-owen:
**Fixes:** Issue #1175

### DIFF
--- a/docs/source/topics/proc_starting-the-virtual-machine.adoc
+++ b/docs/source/topics/proc_starting-the-virtual-machine.adoc
@@ -8,6 +8,7 @@ The [command]`{bin} start` command starts the {prod} virtual machine and OpenShi
 * To avoid networking-related issues, ensure that you are not connected to a VPN and that your network connection is reliable.
 * You set up the host machine through the [command]`{bin} setup` command.
 For more information, see link:{crc-gsg-url}#setting-up-codeready-containers_gsg[Setting up {prod}].
+* On {msw}, ensure that your user account can elevate to Administrator privileges.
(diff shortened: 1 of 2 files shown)


robin-owen:
Note that I'm unsure about the required permissions for `crc start` on Linux/macOS -- is the ability to elevate privileges using `sudo` required? If so, I can update this PR with that information as well.

anjannath:
> Note that I'm unsure about the required permissions for crc start on Linux/macOS -- is the ability to elevate privileges using sudo required?

Not needed for linux, but for macOS during `crc start` a network restart might be needed and to perform that sudo is used.

praveenkumar:
> Not needed for linux, but for macOS during crc start a network restart might be needed and to perform that sudo is used.

No on the mac also we don't need any sudo operation as part of `crc start` the network restart happen using logged in user.

docs/source/topics/proc_starting-the-virtual-machine.adoc
```diff
@@ -8,6 +8,7 @@ The [command]`{bin} start` command starts the {prod} virtual machine and OpenShi
 * To avoid networking-related issues, ensure that you are not connected to a VPN and that your network connection is reliable.
 * You set up the host machine through the [command]`{bin} setup` command.
 For more information, see link:{crc-gsg-url}#setting-up-codeready-containers_gsg[Setting up {prod}].
+* On {msw}, ensure that your user account can elevate to Administrator privileges.
```
praveenkumar, line 11: :+1:

PR #1458, opened 2020-08-12
Add Windows CI (compile and unit tests)

guillaumerose:
Use AppVeyor

### DIFF
--- a/cmd/crc/cmd/status_test.go
+++ b/cmd/crc/cmd/status_test.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/code-ready/crc/pkg/crc/machine/fakemachine"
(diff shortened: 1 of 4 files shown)


guillaumerose:
Very basic checks right now. Unit tests are green now.

praveenkumar:
@guillaumerose I can see https://ci.appveyor.com/project/guillaumerose/crc/builds/34612389 which means project is under your name, I think it should be part of `code-ready` so anyone in the team restart the job if needed.

guillaumerose:
Links and perm are ok now. It should be ready to merge.

cmd/crc/cmd/status_test.go
```diff
@@ -52,7 +53,7 @@ func TestJsonStatus(t *testing.T) {
   "cacheDir": "%s"
 }
 `
-	assert.Equal(t, fmt.Sprintf(expected, cacheDir), out.String())
+	assert.Equal(t, fmt.Sprintf(expected, strings.Replace(cacheDir, `\`, `\\`, -1)), out.String())
```
praveenkumar, line 56: @guillaumerose I am not sure what you are accomplice here, do you want to replace `\` by `\\` in windows so that `cacheDir` should take only single `\` when `fmt.Sprintf` executed? https://play.golang.org/p/-XJ9XdtqjwS

guillaumerose: In order to create valid JSON string, I need to escape backslashes.

guillaumerose:
Rebased !

PR #1465, opened 2020-08-18
Remove kubeconfig file copy to instance

praveenkumar:
We have added it as part of cf3064e but since now this file
is copied during snc/createdisk step, we don't need to copy
it over here again.

- https://github.com/code-ready/snc/pull/222


anjannath:
@praveenkumar Needs rebase!

anjannath (approved):

Issue #705, opened 2019-10-10
How to access CRC or apps deployed on CRC remotely


morningspace:
### General information

  * OS: Linux
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes)?

## CRC version
```bash
# Put the output of `crc version`
version: 1.0.0-beta.5+f2aa58c
OpenShift version: 4.1.14 (embedded in binary)
```
  
## CRC status
```bash
# Put the output of `crc status`
CRC VM:          Running
OpenShift:       Running (v4.x)
Disk Usage:      13.68GB of 16.09GB (Inside the CRC VM)
Cache Usage:     14.99GB
Cache Directory: /home/morningspace/.crc/cache
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
... ...
```

### Steps to reproduce

After I deploy CRC on a linux server running in LAN. How can I access CRC web console, or apps deployed onto it (if any), from another machine in the same network?

It looks the current CRC is  designed to be used on local machine. So, it has ways such as Network Manager for Linux or `/etc/hosts` + `/etc/resolver/testing` for Mac to route requests outside to OpenShift running inside a VM. But it seems all happen on the same machine.

On the other hand, the original `oc cluster up` has the ability to support cross-machine access, because it runs on the machine directly (w/o VM) and it depends on the simple public DNS service `nip.io`(by default).

This seems to be the missing part in CRC? Or, if there's anything I missed.

I'd think this is a common usage scenario for dev/test/demo purpose at team level.

### Expected
I can access CRC or apps deployed on CRC from another machine

### Actual
I can only access it on the machine which is deployed CRC.

gbraad:
> I can only access it on the machine which is deployed CRC.

This was the same for Minishift unless you used the 'generic driver' to deploy this without a VM. This is by design as your are running this on a virtual network. The intended use-case is to use this for development and any unauthorized reote access would not be a wanted situation. 

However, there are ways around this, but they are for now outside the scope of CRC. You could for instance allow the use of `ssh` with a proxy. Or, depending on the platform, reconfigure the network to use a remotely accessible network segment.

morningspace:
[...]

PR #1473, opened 2020-08-20
Issue #705 Document remote set up, client connection

robin-owen:
This is based on the work of Trevor McKay published as a Gist: https://gist.github.com/tmckayus/8e843f90c44ac841d0673434c7de0c6a

Thank you for the assistance, @tmckayus!

**Fixes:** Issue #705


docs/source/topics/proc_connecting-to-remote-instance.adoc
```diff
@@ -0,0 +1,63 @@
+[id="connecting-to-remote-instance_{context}"]
+= Connecting to a remote {prod} instance
+
+Follow this procedure to connect a client machine to a remote server running a {prod} OpenShift cluster.
```
praveenkumar, line 4: Should we somewhere mention that client and remote machine is considered Fedora/CentOS/RHEL based because all our doc commands are based on those platform.

PR #1484, opened 2020-08-26
Simplify oc package

cfergeau:
## Solution/Idea

This is just a refactoring of the 'oc' package in order to make it a bit easier to understand, and to have less code duplication. The gist of the changes is the introduction of an os.CommandRunner interface (instead of oc.Runner) which can be reused by the systemd code in the future (https://github.com/cfergeau/crc/commits/systemd)

## Testing

There should be no behaviour change at all, so no specific testing needed, except to make sure things are still working as before.

### DIFF
--- a/pkg/os/exec.go
+++ b/pkg/os/exec.go
@@ -60,3 +60,26 @@ func RunWithDefaultLocale(command string, args ...string) (string, string, error
 func RunWithDefaultLocalePrivate(command string, args ...string) (string, string, error) {
 	return runPrivate(command, args, defaultLocaleEnv)
 }
+
+type CommandRunner interface {
+	Run(command string, args ...string) (string, string, error)
+	RunPrivate(command string, args ...string) (string, string, error)
+	RunPrivileged(reason string, cmdAndArgs ...string) (string, string, error)
+}
+type localRunner struct{}
+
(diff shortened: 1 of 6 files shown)


pkg/os/exec.go
```diff
@@ -51,26 +51,35 @@ func RunWithPrivilege(reason string, cmdAndArgs ...string) (string, string, erro
...
+	RunPrivileged(reason string, cmdAndArgs ...string) (string, string, error)
 }
+type localRunner struct{}
 
-func RunWithDefaultLocalePrivateAndEnv(command string, args []string, extraEnv map[string]string) (string, string, error) {
-	env := map[string]string{"LC_ALL": "C", "LANG": "C"}
-	for k, v := range extraEnv {
-		env[k] = v
-	}
-	return runPrivate(command, args, env)
```
praveenkumar, line 75: These both removed function is used to get `oc config view` because `oc` doesn't respect `--kubeconfig` option when you try to merge the 2 different kubeconfig, this use case is not required since it we are using oc package now. I think in the future this might be again need refactor when some command require and env variable set before executing. I am not sure we should have it right now or again do refactor around when this requirement comes.

cfergeau: Yup, it can always be readded if/when needed

PR #1488, opened 2020-08-27
Add 10m timeout to golangci run to stop travis CI failure

praveenkumar:
By default the timeout set to `1m` which is way to less for
travis CI to run the linter.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,7 @@
 run:
   build-tags:
     - build
+  timeout: 10m
 
 linters:
   enable:


anjannath (approved):
I feel like 10m is too much, but ideally it'll be done before that, only in some real issue we'll wait the full 10m before failing.

cfergeau:
> I feel like 10m is too much, but ideally it'll be done before that, only in some real issue we'll wait the full 10m before failing.

Yes, 10m feels very long, why is the travis CI taking that long to run golangci?

Makefile
```diff
@@ -148,12 +148,12 @@ $(GOPATH)/bin/golangci-lint:
 # Run golangci-lint against code
 .PHONY: lint cross-lint
 lint: $(GOPATH)/bin/golangci-lint
-	$(GOPATH)/bin/golangci-lint run
+	$(GOPATH)/bin/golangci-lint run --timeout 10m
```
cfergeau, line 151: This should be done in `.golangci.yml`

praveenkumar:
> Yes, 10m feels very long, why is the travis CI taking that long to run golangci?

@cfergeau @anjannath This is I just put for safer side, even it takes 2 mins.

Issue #1487, opened 2020-08-27
[BUG] Text formatting issue in crc config help output


anjannath:
### General information

  * OS: All OS

## CRC version
```bash
1.15.0
```

### Steps to reproduce

  1.  `crc config -h`

### Actual
```bash
❯ crc config -h
Modifies crc configuration properties.
Configurable properties (enter as SUBCOMMAND): 

 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * 
 * bundle
 * cpus
 * disable-update-check
 * enable-experimental-features
 * http-proxy
 * https-proxy
 * memory
 * nameserver
 * no-proxy
 * proxy-ca-file
 * pull-secret-file
 * start-okd-cluster
 * skip-check-bundle-cached
 * warn-check-bundle-cached
 * skip-check-crc-dnsmasq-file
 * warn-check-crc-dnsmasq-file
 * skip-check-crc-network
 * warn-check-crc-network
 * skip-check-crc-network-active
 * warn-check-crc-network-active
 * skip-check-kvm-enabled
 * warn-check-kvm-enabled
 * skip-check-libvirt-driver
 * warn-check-libvirt-driver
 * skip-check-libvirt-installed
 * warn-check-libvirt-installed
 * skip-check-libvirt-running
 * warn-check-libvirt-running
 * skip-check-libvirt-version
 * warn-check-libvirt-version
 * skip-check-network-manager-config
 * warn-check-network-manager-config
 * skip-check-network-manager-installed
 * warn-check-network-manager-installed
 * skip-check-network-manager-running
 * warn-check-network-manager-running
 * skip-check-podman-cached
 * warn-check-podman-cached
 * skip-check-ram
 * warn-check-ram
 * skip-check-root-user
 * warn-check-root-user
 * skip-check-user-in-libvirt-group
 * warn-check-user-in-libvirt-group
 * skip-check-virt-enabled
 * warn-check-virt-enabled

Usage:
  crc config SUBCOMMAND [flags]
  crc config [command]

Available Commands:
  get         Get a crc configuration property
  set         Set a crc configuration property
[...]

PR #1489, opened 2020-08-27
config: Don't preallocate array in configurableFields

cfergeau:
If we preallocate the 'keys' array in configurableFields() to the number
of config keys, and then append the config keys to that preallocated
array, we'll end up with an array containing len(cfg.AllConfigs()) empty
keys followed by the len(cfg.AllConfigs()) keys, which is not what we
want.

This commit removes the preallocation of the 'keys' array as this is
unlikely to make a noticeable difference when the 'config' command is
being used.

This fixes https://github.com/code-ready/crc/issues/1487

**Fixes:** Issue #1487

## Testing

1. crc config -h no longer have N empty entries `* ` followed by N `* config-name` entries

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -78,7 +78,7 @@ func less(lhsKey, rhsKey string) bool {
 
 func configurableFields() string {
 	var fields []string
-	var keys = make([]string, len(cfg.AllConfigs()))
+	var keys []string
 
 	for key := range cfg.AllConfigs() {
 		keys = append(keys, key)


cmd/crc/cmd/config/config.go
```diff
@@ -78,7 +78,7 @@ func less(lhsKey, rhsKey string) bool {
 
 func configurableFields() string {
 	var fields []string
-	var keys = make([]string, len(cfg.AllConfigs()))
+	var keys []string
```
anjannath, line 81: :+1:

anjannath (approved):

Issue #1492, opened 2020-08-30
Invalid prerequisites check: Windows 10 Student could run crc


bahadurbaniya:
### General information

  * OS: Linux / macOS / Windows
  * Hypervisor: Hyper-V 
  * Did you run `crc setup` before starting it ? **YES**
 
## CRC version
```bash
# Put the output of `crc version`
C:\Users\Baha>crc version
CodeReady Containers version: 1.15.0+e317bed
OpenShift version: 4.5.7 (embedded in binary)
```
  
## CRC status
```bash
# Put the output of `crc status`
 C:\Users\Baha>crc status
Machine 'crc' does not exist. Use 'crc start' to create it
```

## CRC config
```bash
# Put the output of `crc config view`

C:\Users\Baha>crc config view
no output

```
## Host Operating System
```bash
# Put the output of `systeminfo` in case of Windows

Host Name:                 BB-XPS
OS Name:                   Microsoft Windows 10 Education
OS Version:                10.0.18363 N/A Build 18363
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Standalone Workstation
OS Build Type:             Multiprocessor Free
Registered Owner:          Baha
Registered Organization:   Microsoft
Product ID:                00328-00252-07944-AA592
Original Install Date:     23/02/2020, 13:41:50
System Boot Time:          29/08/2020, 6:38:05
System Manufacturer:       Dell Inc.
System Model:              XPS 13 9350
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 78 Stepping 3 GenuineIntel ~501 Mhz
BIOS Version:              Dell Inc. 1.12.2, 15/12/2019
Windows Directory:         C:\WINDOWS
System Directory:          C:\WINDOWS\system32
Boot Device:               \Device\HarddiskVolume1
System Locale:             en-gb;English (United Kingdom)
Input Locale:              en-us;English (United States)
Time Zone:                 (UTC+10:00) Brisbane
Total Physical Memory:     16,252 MB
Available Physical Memory: 2,142 MB
Virtual Memory: Max Size:  26,092 MB
Virtual Memory: Available: 5,074 MB
Virtual Memory: In Use:    21,018 MB
Page File Location(s):     C:\pagefile.sys
Domain:                    WORKGROUP
Logon Server:              \\BB-XPS
Hotfix(s):                 14 Hotfix(s) Installed.
                           [01]: KB4562900
[...]

PR #1494, opened 2020-08-31
Issue #1492 Only windows 10 Home edition doesn't support Hyperv

anjannath:
As per the chart in wikipedia all the other editions of windows
has support for running vms using hyperv, so only filtering out
Home edition of windows in the preflight check
https://en.wikipedia.org/wiki/Windows_10_editions#Comparison_chart

Fixes: #1492 

## Solution/Idea

As mentioned in the issue, Win10 Education edition supports Hyperv so crc can run on it, therefore modified the pre-flight check to filter out only Home edition.
## Testing

Should be able to run `crc start` successfully on win10 Education edition, (all others excluding Home edition)

### DIFF
--- a/docs/source/topics/ref_minimum-system-requirements.adoc
+++ b/docs/source/topics/ref_minimum-system-requirements.adoc
@@ -26,8 +26,8 @@ To assign more resources to the {prod} virtual machine, see link:{crc-gsg-url}#c
 
 === {msw}
 
-* On {msw}, {prod} requires the Windows 10 Pro Fall Creators Update (version 1709) or newer.
-{prod} does not work on earlier versions or editions of {msw}.
+* On {msw}, {prod} requires the Windows 10 Fall Creators Update (version 1709) or newer.
+{prod} does not work on earlier versions of {msw}.
 {msw} 10 Home Edition is not supported.
 
 === {mac}
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -53,16 +53,11 @@ func checkWindowsEdition() error {
 	windowsEdition := strings.TrimSpace(stdOut)
 	logging.Debugf("Running on Windows %s edition", windowsEdition)
 
-	if strings.HasPrefix(windowsEdition, "Professional") {
-		return nil
+	if strings.HasPrefix(windowsEdition, "Home") {
+		return fmt.Errorf("Windows Home edition is not supported")
 	}
 
-	switch windowsEdition {
-	case "Enterprise":
-		return nil
-	default:
-		return fmt.Errorf("Supported Windows editions are Professional and Enterprise. Windows %s edition is not supported", windowsEdition)
-	}
+	return nil
 }
 
 func checkHyperVInstalled() error {


pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -53,16 +53,11 @@ func checkWindowsEdition() error {
 	windowsEdition := strings.TrimSpace(stdOut)
 	logging.Debugf("Running on Windows %s edition", windowsEdition)
 
-	if strings.HasPrefix(windowsEdition, "Professional") {
-		return nil
+	if windowsEdition == "Home" {
```
cfergeau, line 56: This table you link to has a row saying:
> Has N or KN variants? Yes

so I think you need to check for a `Home` prefix, and not just check for `==`.

Does the documentation also needs to be updated?

anjannath: > so I think you need to check for a Home prefix, and not just check for ==.

 will update it, i missed that row, Thanks!

> Does the documentation also need to be updated?

We could add to it, we only mention Home edition is not supported.
> On Microsoft Windows, CodeReady Containers requires the Windows 10 Pro Fall Creators Update (version 1709) or newer. CodeReady Containers does not work on earlier versions or editions of Microsoft Windows. Microsoft Windows 10 Home Edition is not supported.

docs/source/topics/ref_minimum-system-requirements.adoc
```diff
@@ -26,9 +26,9 @@ To assign more resources to the {prod} virtual machine, see link:{crc-gsg-url}#c
 
 === {msw}
 
-* On {msw}, {prod} requires the Windows 10 Pro Fall Creators Update (version 1709) or newer.
-{prod} does not work on earlier versions or editions of {msw}.
-{msw} 10 Home Edition is not supported.
+* On {msw}, Except {msw} 10 Home Edition {prod} works on all other Windows 10 Editions.
+For Pro Edition {prod} requires the Windows 10 Pro Fall Creators Update (version 1709) or newer.
+{prod} does not work on earlier versions of Windows 10 Pro Edition.
```
cfergeau, line 31: Maybe this? 
```
* On {msw}, {prod} requires the Windows 10 Fall Creators Update (version 1709) or newer.
{prod} does not work on earlier versions of {msw}.
{msw} 10 Home Edition is not supported.
```

@robin-owen ^^

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -53,16 +53,11 @@ func checkWindowsEdition() error {
...
+		return fmt.Errorf("Windows Home edition is not supported")
 	}
 
-	switch windowsEdition {
-	case "Enterprise":
-		return nil
-	default:
-		return fmt.Errorf("Supported Windows editions are Professional and Enterprise. Windows %s edition is not supported", windowsEdition)
-	}
+	return nil
```
cfergeau, line 60: :+1:

cfergeau (commented):
ACK from me for the code part, but Kevin's approval is needed for the documentation changes.

docs/source/topics/ref_minimum-system-requirements.adoc
anjannath: @cfergeau Your version reads more clearer, i can change it to that, but lets see what @robin-owen thinks.

robin-owen: @anjannath The change @cfergeau recommends looks good to me, that's what I would change it to. :)

robin-owen (approved):

PR #1498, opened 2020-09-01
Refactor systemd code

cfergeau:
The systemd code currently has one file for systemd interactions on the host, and a second very similar file for systemd interaction on the guest. This PR refactors this code to make use of os.CommandRunner and have as much common code as possible between the 2 implementations.

## Proposed changes

1. The systemd package uses os.CommandRunner
2. linux preflight checks use the systemd package

## Testing

There should be no changes in behaviour. `crc setup` on linux and `crc start` on all platforms run the code this PR modifies, but if there is any noticeable difference, that's a bug.


pkg/crc/machine/machine.go
```diff
@@ -365,35 +366,34 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 	}
 
 	// Check if kubelet service is running inside the VM
-	kubeletStarted, err := sd.IsActive("kubelet")
-	if err != nil {
+	kubeletStatus, err := sd.Status("kubelet")
+	if err != nil || kubeletStatus != states.Running {
 		return startError(startConfig.Name, "kubelet service is not running", err)
 	}
-	if kubeletStarted {
```
guillaumerose, line 372: Good ! Thanks for this removal.

pkg/crc/systemd/systemd_linux_test.go
```diff
@@ -0,0 +1,98 @@
...
+	runner := mockSystemdRunner{
+		test: t,
+	}
+
+	return &Commander{
+		commandRunner: &runner,
+	}, &runner
+}
+
+func TestSystemd(t *testing.T) {
```
guillaumerose, line 22: Interesting!

If I was using Java, I would have written something like

```java
when(runner.Run("sysctl", "enable", "foo")).thenReturn(...)
when(runner.Run("sysctl", "enable", "bar")).thenFail(...)

assertThat(systemctl.Enable("foo")).isTrue()
assertThat(systemctl.Enable("bar")).isFalse()
```

With Go, I would have probably set in the stone what the runner should answer.

```go
mockSystemdRunner{
  commands: map[string]bool{
    "systemctl enable foo": true,
    "systemctl enable bar": false,
  }
}
```

Don't change anything! Both tests are good for me :)

guillaumerose (approved):

PR #1499, opened 2020-09-01
Misc docs generation related changes

anjannath:
## Solution/Idea

Currently we  have three different images for docs related tasks like  generating html, pdf doing a preview of the changes in a browser, instead of having three different images with the potential of having more when we want to add something (Like link checking), it is better to combine all of them into one single image that we can later keep adding to for all docs related requirements.
    
## Proposed changes

Add dockerfile for docs-builder image.The new image combines all requirement for docs generation tasks in a single image and adds link checking for the docs

## Testing

All the `make` targets related to documentation generation: `build_docs, build_docs_pdf, docs_serve` should work exactly like before.
A new target for link checking `make check_links` is introduced which should be verified.


Makefile
```diff
@@ -15,10 +15,8 @@ SOURCE_DIRS = cmd pkg test
 RELEASE_DIR ?= release
 
 # Docs build related variables
-DOCS_BUILD_DIR ?= docs/build
-DOCS_BUILD_CONTAINER ?= registry.gitlab.com/gbraad/asciidoctor-centos:latest
-DOCS_PDF_CONTAINER ?= quay.io/crcont/asciidoctor-pdf:latest
-DOCS_SERVE_CONTAINER ?= quay.io/crcont/docs-server:latest
+DOCS_BUILD_DIR ?= /docs/build
+DOCS_BUILD_CONTAINER ?= quay.io/anjan/docs-builder
```
anjannath, line 19: Need to change this to `quay.io/crcont/docs-builder` before merging, haven't pushed the image to crcont org yet!

images/docs-builder/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+		;;
+	build_docs_pdf)
+		echo "Generating pdf from docs source"
+		shift 
+		asciidoctor-pdf $@
+		;;
+	check_links)
+		echo "Checking if all links are alive in docs source"
+		find . -name \*.adoc | xargs -n1 asciidoc-link-check -q
+		;;
```
cfergeau, line 22: Minor potential refinement, this 'check_links' code could be added at the same time as you add support for it in the Makefile/CI (just a comment, no changes required). Seeing that ' Add make target to check links in the docs ' modifies this part of the script, better to have all the changes in a single place

images/docs-builder/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+	docs_serve)
+		cd build
+		echo "Make sure port 8088 of the is exposed"
```
cfergeau, line 25: `of the`? Is this a log to tell what's going on? or something the user has to fix?

images/docs-builder/entrypoint.sh
cfergeau: Fwiw, all other targes have 'docs' in the name, either I'd expect a 'docs' mention here, or I'd remove it from the other targets as this is a 'doc-builder' image already.

Makefile
cfergeau: :+1:

Makefile
```diff
@@ -97,15 +95,15 @@ test:
...
 
 .PHONY: build_docs_pdf
 build_docs_pdf:
-	podman run -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_PDF_CONTAINER) -D /$(DOCS_BUILD_DIR) -o doc.pdf $(DOCS_BUILD_TARGET)
+	podman run -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_BUILD_CONTAINER) build_docs_pdf -D $(DOCS_BUILD_DIR) -o doc.pdf $(DOCS_BUILD_TARGET)
 
 .PHONY: docs_serve
 docs_serve: build_docs
-	podman run -v $(CURDIR)/docs:/docs:Z --rm -p 8088:8088/tcp $(DOCS_SERVE_CONTAINER)
+	podman run -it -v $(CURDIR)/docs:/docs:Z --rm -p 8088:8088/tcp $(DOCS_BUILD_CONTAINER) docs_serve
```
cfergeau, line 106: why `-it` now?

Makefile
```diff
@@ -105,6 +105,10 @@ build_docs_pdf:
 docs_serve: build_docs
 	podman run -it -v $(CURDIR)/docs:/docs:Z --rm -p 8088:8088/tcp $(DOCS_BUILD_CONTAINER) docs_serve
 
+.PHONY: check_links
+check_links:
+	podman run -it -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_BUILD_CONTAINER) check_links
```
cfergeau, line 110: is the `-it` needed here?  Do the docs need to be built before we can check the links? or is it acting on the source asciidoc data?

images/docs-builder/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM registry.redhat.io/ubi8
```
cfergeau, line 1: We used to use `FROM registry.redhat.io/ubi8/ruby-25`, is this not good enough for your needs?

images/docs-builder/links_whitelist.json
```diff
@@ -0,0 +1,9 @@
+{
+    "ignorePatterns": [
+        { "pattern": "^https://api.crc.testing" },
+        { "pattern": "^http://api.crc.testing" },
+        { "pattern": "https://console-openshift-console.apps-crc.testing" },
+        { "pattern": "http://example.proxy.com" }
```
cfergeau, line 6: Totally unrelated to this PR, but `proxy.example.com` would be better than `example.proxy.com` as `example.com` is explicitly reserved for use in examples https://tools.ietf.org/html/rfc2606

images/docs-builder/entrypoint.sh
anjannath: Just to tell whats going on..

Makefile
anjannath: > is the -it needed here?

we get a nicer colored output and to stop the process inside the container by `^C`

>  Do the docs need to be built before we can check the links? or is it acting on the source asciidoc data?

No we don't need to build, its acting on The Source.

Makefile
anjannath: It should've been there from before, we need the stdin open to send a `^C` to kill the server, `-t` is probably not needed here, unlike in https://github.com/code-ready/crc/pull/1499#discussion_r481845232 where we can get a more nicer colored output because of this.

images/docs-builder/Dockerfile
anjannath: we also need Nodejs, the link checking tool is a nodejs script tool, so need to install with npm, it'd probably be possible to install nodejs on the Ruby image, but i went with a generic base image.

images/docs-builder/links_whitelist.json
anjannath: Simple enough change, added a commit to take care of this.

images/docs-builder/entrypoint.sh
anjannath: Changed to `docs_check_links`

Makefile
cfergeau: hmm ok, separate commit with just that change would have made sense, as here it's not clear if it's intentional or not, nor why this is done.

images/docs-builder/entrypoint.sh
```diff
@@ -0,0 +1,32 @@
...
+		shift 
+		asciidoctor-pdf $@
+		;;
+	check_links)
+		echo "Checking if all links are alive in docs source"
+		find . -name \*.adoc | xargs -n1 asciidoc-link-check -q
+		;;
+	docs_serve)
+		cd build
+		echo "Make sure port 8088 of the container is exposed"
```
cfergeau, line 25: The `Make sure` confuses me :-/ ` docs can be browsed at http://localhost:8080`?

anjannath: That is already printed by http server :)

```
podman run -it -v /home/anjan/github.com/code-ready/crc/docs:/docs:Z --rm -p 8088:8088/tcp quay.io/anjan/docs-builder docs_serve 
Make sure port 8088 of the container is exposed
Serving HTTP on 0.0.0.0 port 8088 (http://0.0.0.0:8088/) ...
^C
Keyboard interrupt received, exiting.
``` 

so when doing `make serve_docs` this is not really help but if the image is separately used and you just run `podman run -it docs-builder docs_serve` then this will remind you to expose the port. we can remove it..

cfergeau: Hmm yeah I think I would just remove it, if someone wants to run this by hand, it's their problem to use the right commandline :)

images/docs-builder/entrypoint.sh
```diff
@@ -18,7 +18,7 @@ case $1 in
 		;;
 	check_links)
 		echo "Checking if all links are alive in docs source"
-		find . -name \*.adoc | xargs -n1 asciidoc-link-check -q
+		find . -name \*.adoc | xargs -n1 asciidoc-link-check -c /links_whitelist.json -q
```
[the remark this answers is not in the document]
anjannath: Done, brought the check links part here from the first commit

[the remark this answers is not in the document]
cfergeau: :+1:

cfergeau (approved):
Looks good to me

anjannath:
Pushed the docs-builder image to `quay.io/crcont/docs-builder` and using that one in Makefile now.

images/docs-builder/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+MAINTAINER CodeReady Containers <devtools-cdk@redhat.com>
+
+RUN yum update -y
+RUN yum install -y ruby rubygems python38
```
praveenkumar, line 6: Better to do, it will only create single layer and also clean the yum cache so the end result of image should be lesser in size.

```
RUN yum update -y && \
    yum install --setopt=tsflags=nodocs -y \
    ruby rubygems python38 && \
    yum clean all && rm -rf /var/cache/yum/*
```

images/docs-builder/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN gem install asciidoctor --pre
```
praveenkumar, line 7: Again do it in singler layer if possible

```
RUN gem install asciidoctor asciidoctor-pdf --pre
```

Makefile
```diff
@@ -103,7 +103,7 @@ build_docs_pdf:
 
 .PHONY: docs_serve
 docs_serve: build_docs
-	podman run -v $(CURDIR)/docs:/docs:Z --rm -p 8088:8088/tcp $(DOCS_BUILD_CONTAINER) docs_serve 
+	podman run -it -v $(CURDIR)/docs:/docs:Z --rm -p 8088:8088/tcp $(DOCS_BUILD_CONTAINER) docs_serve
```
praveenkumar, line 106: @anjannath `it` is for interactive mode, why do need the docs_serve to run as interactive session?

anjannath: So that we can terminate the http server process with `ctrl+c`

images/docs-builder/Dockerfile
cfergeau: Can we squash the layers somehow rather than having a hard to read line?

anjannath: Yeah, non trivial, have to look at how to squash the layers, but can be handled later imo, since this is an image we don't ship to the users, size shouldn't be that much of an issue.

Issue #977, opened 2020-01-28
OKD-based, registration-free version of CRC


omkensey:
Creating this issue per request of vrutkovs at https://kubernetes.slack.com/archives/C6BRQSH2S/p1580198984018400?thread_ts=1580180011.017200&cid=C6BRQSH2S

### Expected
An OKD-based version of CRC is available that does not require portal registration or authenticating to a container registry to start up.

### Actual
Using CRC requires setting up a Red Hat portal account to download the `crc` release and the necessary pull secret to pull the OpenShift containers from the authenticated Red Hat container registry.

vrutkovs:
OKD releases (not yet released officially) - https://origin-release.svc.ci.openshift.org/

* Based on OCP 4.4 nightlies
* Fedora CoreOS base + machine-os-content
* Forked MCO (`fcos` branch)
* Forked installer (`fcos` branch)

gbraad:
We have been looking into this and had many discussions to allow for an OKD-based release. However, s you can see, this is at the moment still based on forks instead of upstreamed code. This is why this can, at this point in time, still not be considered for release. We are however working on this.

Thanks @vrutkovs for the heads up ;-)!

gbraad:
@vrutkovs any updates?

anjannath:
We made some progress to have OKD based, registration free _bundles_ (one still need to manually change things in crc to use those bundles). the scripts in our snc repo needed minimal changes during my testing to successfully generate OKD bundles. PR in snc repo https://github.com/code-ready/snc/pull/216

overview of the changes i did to the scripts:
- No need to remove the pull-secret config and file, as we inject a dummy pull-secret for okd installation
- Switch the mirror to download okd bits
- Change the ostree hash greping pattern, it starts with fedora-coreos in okd (we need these info to use the image with hyperkit)
- The root partition is not luks-encryption ready in focs, so modify guestfish commands accordingly (createdisk.sh)

guillaumerose:
Related to https://github.com/openshift/community/issues/54

guillaumerose:
OKD release of CRC is available here: https://www.okd.io/crc.html

The UX around the pull secret need to be enhanced but it's fully functionnal.

gbraad:
Currently @cgruver handles this. Closing the issue as this needs a very differently timeline now.

PR #1505, opened 2020-09-03
Add OKD compatibility to crc for building CRC for OKD bundles

cgruver:
**Fixes:** Issue #977 

**Relates to:** https://github.com/code-ready/snc/issues/212

## Solution/Idea

I modified Makefile, constants.go, and version.go to allow for building a `crc` image which uses OKD instead of OCP.  The building of `crc` should not be affected by these changes, as it is the default.   `crc` for OKD can be built by setting environment variables.

## Proposed changes

1. Removed `DefaultOcURLBase` from `constants.go` and added it as a build config variable to `version.go`
1. Modified the creation of `ocURLForOs` to account for okd vs. ocp URLs
1. Modified `Makefile` to allow overriding `BUNDLE_VERSION`
1. Modified `Makefile` to add `MIRROR` as an override for `version.defaultOcURLBase` and added `https://mirror.openshift.com/pub/openshift-v4/clients/ocp` as a default value.

## Testing

I have tested the OKD builds of snc and crc.  I do not have the facility to test OCP builds.

To build for OKD:

Build `snc`:  See Pull Request - https://github.com/code-ready/snc/pull/230

```bash
export BUNDLE_VERSION=${OPENSHIFT_VERSION}                                  # Using OPENSHIFT_VERSION that was set for snc.
export MIRROR=https://github.com/openshift/okd/releases/download     
export BUNDLE_DIR=/tmp/snc                                                                      # where snc was built in /tmp/snc

make embed_bundle
```

Use `{"auths":{"fake":{"auth": "Zm9vOmJhcgo="}}}` as your pull secret when prompted during `crc start`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,15 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION = 4.5.0-0.okd-2020-08-12-020541
+BUNDLE_VERSION ?= 4.5.7
 # OC_VERSION and BUNDLE_VERSION are going to same for release artifacts but
 # different for nightly and CI bits where bundle version would be any random
 # string or dd-mm-yyyy format.
 OC_VERSION ?= ${BUNDLE_VERSION}
+OPENSHIFT_VERSION ?= ${BUNDLE_VERSION}
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.15.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
+MIRROR ?= https://mirror.openshift.com/pub/openshift-v4/clients/ocp
 
 # Go and compilation related variables
 BUILD_DIR ?= out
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -159,7 +161,7 @@ cross-lint: $(GOPATH)/bin/golangci-lint
 gen_release_info:
 	@cat release-info.json.sample | sed s/@CRC_VERSION@/\"$(CRC_VERSION)\"/ > $(RELEASE_INFO)
 	@sed -i s/@GIT_COMMIT_SHA@/\"$(COMMIT_SHA)\"/ $(RELEASE_INFO)
-	@sed -i s/@OPENSHIFT_VERSION@/\"$(BUNDLE_VERSION)\"/ $(RELEASE_INFO)
+	@sed -i s/@BUNDLE_VERSION@/\"$(BUNDLE_VERSION)\"/ $(RELEASE_INFO)
```
anjannath, line 164: This is not needed i think, we only have `@OPENSHIFT_VERSION@` in that file

anjannath:
@cgruver The first three commits and the last two commits can be squashed, so there will be two commits, or maybe it makes sense to have it all in one commit, now i see changes from previous commits are reverted in later commits.

cfergeau (commented):
Nit: This PR is not "adding OKD compatibility", it's more about adding knobs which can be changed externally in order to do OKD builds. If you don't know what these knobs should be set to, no OKD build for you. Maybe this can be done through `make okd`, or maybe this will need a `BUNDLE_TYPE` Makefile variable which will be used in the targets which need it. Or maybe this should be dynamic, with a single crc binary being able to support both bundles?

You'll need to cleanup the git history too, probably by squashing everything together, and adding a bit of background on what you did, https://www.berrange.com/posts/2012/06/27/thoughts-on-improving-openstack-git-commit-practicehistory/

cgruver:
I will clean up the commits and re-submit this.  @cfergeau Thanks for the tips!

cgruver:
Submitted clean request: https://github.com/code-ready/crc/pull/1511

PR #1513, opened 2020-09-08
[WIP] Remove hard coded node IP for kubelet

praveenkumar:
This is an experiment PR to test e2e if we make those changes and those changes should be part of snc side instead here.


praveenkumar:
Looks like even with those changes the cluster is not able to behave correctly :( 

/close

PR #1520, opened 2020-09-14
Add json support for setup/stop/delete/cleanup commands

guillaumerose:
**Relates to:** Issue #1227

* Sometimes we prompt the user a yes/no question. With json output, he has to add ` --force` (a little bit like mv/cp).

### DIFF
--- a/cmd/crc/cmd/delete_test.go
+++ b/cmd/crc/cmd/delete_test.go
@@ -1,0 +1,38 @@
+package cmd
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/code-ready/crc/pkg/crc/machine/fakemachine"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPlainDelete(t *testing.T) {
+	cacheDir, err := ioutil.TempDir("", "cache")
+	require.NoError(t, err)
+	defer os.RemoveAll(cacheDir)
+
+	out := new(bytes.Buffer)
+	assert.NoError(t, runDelete(out, fakemachine.NewClient(), true, cacheDir, true, true, ""))
+	assert.Equal(t, "Deleted the OpenShift cluster\n", out.String())
+
+	_, err = os.Stat(cacheDir)
+	assert.True(t, os.IsNotExist(err))
+}
+
+func TestJSONDelete(t *testing.T) {
+	cacheDir, err := ioutil.TempDir("", "cache")
+	require.NoError(t, err)
+	defer os.RemoveAll(cacheDir)
+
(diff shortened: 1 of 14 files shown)


pkg/crc/input/input.go
```diff
@@ -1,17 +1,23 @@
...
 	survey "gopkg.in/AlecAivazis/survey.v1"
 )
 
 func PromptUserForYesOrNo(message string, force bool) bool {
 	if force {
 		return true
 	}
+	if !terminal.IsTerminal(int(os.Stdin.Fd())) {
+		return false
+	}
```
cfergeau, line 20: Do we have the equivalent of `yum --assumeyes`/`yum --assumeno`?

cmd/crc/cmd/delete_test.go
```diff
@@ -0,0 +1,41 @@
...
+	require.NoError(t, err)
+	defer os.RemoveAll(cacheDir)
+
+	out := new(bytes.Buffer)
+	assert.NoError(t, runDelete(out, fakemachine.NewClient(), true, cacheDir, false, true, jsonFormat))
+	assert.Equal(t, `{
+  "success": true
+}
+`, out.String())
+
```
cfergeau, line 38: Did you try `assert.JSONEq(t, ``{"success": true}``, out.String())` to avoid the slightly hard to parse/read multiline string?

cfergeau (commented):
A few comments, still thinking that shortlog-only commits should be the exception rather than the norm :)

cmd/crc/cmd/setup.go
```diff
@@ -31,13 +36,30 @@ func runSetup(arguments []string) error {
...
-		return err
+	err := preflight.SetupHost()
+	return render(&setupResult{
+		Success: err == nil,
+		Error:   errorMessage(err),
+	}, os.Stdout, outputFormat)
+}
+
+type setupResult struct {
+	Success bool   `json:"success"`
```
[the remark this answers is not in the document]
guillaumerose: sucess/errors was already present in the daemon API. I just kept it. If there is no error, the JSON output would be `{}` or `{errors:""}` ? It feels empty :)

pkg/crc/input/input.go
guillaumerose: No we don't have this.

cmd/crc/cmd/delete_test.go
guillaumerose: I didn't know this assertions, thanks! I will use it now.

cfergeau (approved):
Looks good to me
>  Sometimes we prompt the user a yes/no question. With json output, he has to add --force (a little bit like mv/cp).

This could be described in the commit log for the stop command and the one changing getPullSecret/yes-no rather than just in the PR description

PR #1523, opened 2020-09-15
Improve version mismatch wording

guillaumerose:
Before:
```
$ crc start -p ~/pull-secret.txt -b ~/bundle_dir/crc_libvirt_4.5.9.crcbundle -ojson
{
  "success": false,
  "error": "expected: 4.5.9 but got: 4.5.1"
}
``` 

After:
```
$ crc start -p ~/pull-secret.txt -b ~/bundle_dir/crc_libvirt_4.5.9.crcbundle -ojson
{
  "success": false,
  "error": "oc version mismatch: 4.5.9 expected but 4.5.1 found in the cache"
}

```

### DIFF
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -25,12 +25,13 @@ type Cache struct {
 }
 
 type VersionMismatchError struct {
+	BinaryName      string
 	ExpectedVersion string
 	CurrentVersion  string
 }
 
 func (e *VersionMismatchError) Error() string {
-	return fmt.Sprintf("expected: %s but got: %s", e.ExpectedVersion, e.CurrentVersion)
+	return fmt.Sprintf("%s version mismatch: %s expected but %s found in the cache", e.BinaryName, e.ExpectedVersion, e.CurrentVersion)
 }
 
 func New(binaryName string, archiveURL string, destDir string, version string, getVersion func() (string, error)) *Cache {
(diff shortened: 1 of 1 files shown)


pkg/crc/cache/cache.go
```diff
@@ -25,12 +25,13 @@ type Cache struct {
...
 
 type VersionMismatchError struct {
+	BinaryName      string
 	ExpectedVersion string
 	CurrentVersion  string
 }
 
 func (e *VersionMismatchError) Error() string {
-	return fmt.Sprintf("expected: %s but got: %s", e.ExpectedVersion, e.CurrentVersion)
+	return fmt.Sprintf("%s version mismatch: %s expected but %s found in the cache. Run 'crc setup' to update the cache.", e.BinaryName, e.ExpectedVersion, e.CurrentVersion)
```
cfergeau, line 34: Mentioning `crc setup` here (in the cache package) feels a bit out of place, and it's also not really relevant when using json output I would say. Might be better in `crc start` as 
```
var e VersionMismatchError;
if errors.As(err, &e) {
    logging.Errorf("%s. Run 'crc setup' to update the cache.", e.Error())
}
```

Or am I overthinking this?

guillaumerose: I could to that but this check is called in the preflight package, so it won't be that closer to the start command. 

I can remove the "Run crc setup" part. This only happens when I update crc, I guess I can write a better check on top of all commands.

praveenkumar: This version mismatch is only part of preflight checks and right now all the preflight checks should be fixed by running the `crc setup` so I think adding it in `crc start` or even every commands is bit over doing.

guillaumerose: I removed the howto fix it part.

cfergeau (approved):

PR #1526, opened 2020-09-16
Add userland tcpip stack for the VM

guillaumerose:
Linux:

* `crc config set network-mode vsock`
* then `crc daemon` running in an other console
* and finally ` crc start`  as usual, with bundle >= 4.6.1.

Darwin:

* `crc config set network-mode vsock`
* `crc config set enable-experimental-features`
* and finally ` crc start`  as usual, with bundle >= 4.6.1.

Windows:
* `crc config set network-mode vsock`
* then `crc daemon` running in an other console
* and finally ` crc start`  as usual, with bundle >= 4.6.1.


cmd/crc/cmd/daemon.go
```diff
@@ -20,6 +38,106 @@ var daemonCmd = &cobra.Command{
...
+			GatewayIP:         "192.168.127.1",
+			GatewayMacAddress: "\x5A\x94\xEF\xE4\x0C\xDD",
+			DNSRecords: map[string]net.IP{
+				"gateway.crc.testing.":              net.ParseIP("192.168.127.1"),
+				"apps-crc.testing.":                 net.ParseIP("192.168.127.2"),
+				"etcd-0.crc.testing.":               net.ParseIP("192.168.127.2"),
+				"api.crc.testing.":                  net.ParseIP("192.168.127.2"),
+				"api-int.crc.testing.":              net.ParseIP("192.168.127.2"),
+				"oauth-openshift.apps-crc.testing.": net.ParseIP("192.168.127.2"),
+				"crc-zqfk6-master-0.crc.testing.":   net.ParseIP("192.168.126.11"),
```
praveenkumar, line 65: instead of hard code those domains we should take it from bundle metadata side.

guillaumerose:
Still todo: change the flag that enables it and find a better group for vsock device.

gbraad:
please rebase

Makefile
```diff
@@ -74,6 +74,9 @@ vendorcheck:
 .PHONY: install
 install: $(SOURCES)
 	go install -ldflags="$(LDFLAGS)" ./cmd/crc
+ifeq ($(GOOS),linux)
+	sudo setcap CAP_NET_BIND_SERVICE=+eip $(shell go env GOPATH)/bin/crc
```
gbraad, line 78: is this a necessity on ther users machine too to perform?

cmd/crc/cmd/config/config.go
```diff
@@ -32,6 +34,7 @@ func RegisterSettings(cfg *config.Config) {
 	cfg.AddSetting(PullSecretFile, "", config.ValidatePath, config.SuccessfullyApplied)
 	cfg.AddSetting(DisableUpdateCheck, false, config.ValidateBool, config.SuccessfullyApplied)
 	cfg.AddSetting(ExperimentalFeatures, false, config.ValidateBool, config.SuccessfullyApplied)
+	cfg.AddSetting(NetworkMode, string(network.DefaultMode), network.ValidateMode, network.SuccessfullyAppliedMode)
```
gbraad, line 37: this eventually needs documentation. what are the possible values? what is the default? a short oneliner what it does?

cmd/crc/cmd/console.go
```diff
@@ -33,7 +33,7 @@ var consoleCmd = &cobra.Command{
 	Short:   "Open the OpenShift Web Console in the default browser",
 	Long:    `Open the OpenShift Web Console in the default browser or print its URL or credentials`,
 	Run: func(cmd *cobra.Command, args []string) {
-		if renderErr := runConsole(os.Stdout, machine.NewClient(), consolePrintURL, consolePrintCredentials, outputFormat); renderErr != nil {
+		if renderErr := runConsole(os.Stdout, newMachine(), consolePrintURL, consolePrintCredentials, outputFormat); renderErr != nil {
```
gbraad, line 36: where is this defined? It seems multiple files are referring to this.., so this is a static/shared function in the cmd scope for sure. how do we handle shared code?

pkg/crc/machine/machine.go
```diff
@@ -424,6 +432,20 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 	}, err
 }
 
+func getSSHPort(vsockNetwork bool) int {
```
gbraad, line 435: not a huge fan of passing in a boolean to control the behaviour, but also not really of a static variable...

pkg/crc/machine/libvirt/driver_linux.go
```diff
@@ -11,7 +12,12 @@ func CreateHost(machineConfig config.MachineConfig) *libvirt.Driver {
 
 	config.InitVMDriverFromMachineConfig(machineConfig, libvirtDriver.VMDriver)
 
-	libvirtDriver.Network = DefaultNetwork
+	if machineConfig.NetworkMode == network.VSockMode {
+		libvirtDriver.Network = ""
```
gbraad, line 16: When empty, do we remove the interface from the VM? If so, perhaps 'better' to annotate that here. 

Either we can choose to have that happen from the driver code by magic (side-effect behaviour) or explicitly do this.

cmd/crc/cmd/daemon.go
gbraad: `crc-zqfk6-master-0` is not consustent, metadata.
however these other values not defined in the SNC generation step.

I do think a more centralized approach to the IP settings is necessary

pkg/crc/cluster/cluster.go
```diff
@@ -29,7 +29,7 @@ func WaitForSSH(sshRunner *ssh.Runner) error {
 		return nil
 	}
 
-	return errors.RetryAfter(60*time.Second, checkSSHConnectivity, time.Second)
+	return errors.RetryAfter(300*time.Second, checkSSHConnectivity, time.Second)
```
gbraad, line 32: from 1 minute to 5?! what is the justification here?

go.mod
```diff
@@ -76,3 +76,7 @@ replace (
 )
 
 replace vbom.ml/util => github.com/fvbommel/util v0.0.2
+
+replace github.com/miekg/dns => github.com/guillaumerose/dns-1 v1.1.5-0.20200731203003-fe9b991fbece
```
gbraad, line 80: what is the specific change in this library?
preferably not use a fork unless there is nothing else possible.

https://github.com/guillaumerose/dns-1/commit/fe9b991fbece42eb3ef3a118dc043aecf1f48d1c

gbraad: By going from https://golang.org/pkg/net/#UDPConn to https://golang.org/pkg/net/#PacketConn you asre using a more generic interface. Why was this necessary?

gbraad: is this overriding the implementation that gvisor uses?

gbraad (changes requested):
Do not like to see a `replace` for an external module and therefore a fork as there are changes, but are they necessary?

cmd/crc/cmd/daemon.go
guillaumerose: In fact, I tested with multiples bundles and this record is not used at all.  So we don't depend on the master node name apparently.

Makefile
guillaumerose: Good question, from what I understand, the flag is preserved if we tar/untar the binary. It maybe deserves a preflight check?

guillaumerose: This is necessary to be able to bind ports < 1024

pkg/crc/machine/libvirt/driver_linux.go
guillaumerose: You can see it in a different way: when network is empty, there is no interface attached. For the libvirt case, we don't remove it, we just don't add the interface.

I will add a note here.

pkg/crc/cluster/cluster.go
guillaumerose: Drivers are currently responsible of part of the wait: Start returns once the IP is available. This is not possible to do that with vsock. The startup polling is only here and therefore need to be extended a little bit.

go.mod
guillaumerose: *gonet.UDPConn is not *net.UDPConn. I can't use a gonet.UDPConn as a *net.UDPConn, it doesn't compile. 
I opened https://github.com/miekg/dns/pull/1170

Makefile
gbraad: Peopl;e do many different things to our tarball. We can not rely on preserving this I guess. A check might always be neater than a fail ;-)

cmd/crc/cmd/daemon.go
gbraad: the name might internally be recorded somehow... so, why not test this without? this way we can confirm if this is a necessity or not. Things might have changed over time.

pkg/crc/cluster/cluster.go
gbraad: 'a little bit.' ?

would another optimization be possibile in future? in this case instead of `GetIP` this would need to become a `Dial` and see if the socket responds or a `ping` to see if the other endpoint responds correctly.

pkg/crc/machine/libvirt/driver_linux.go
gbraad: oh sorry, 'Linux'. Still thinking of the first-serve Windows ;-).

But anyway, is empty not the default?

pkg/crc/network/types.go
```diff
@@ -12,3 +19,33 @@ type ResolvFileValues struct {
 	SearchDomains []SearchDomain
 	NameServers   []NameServer
 }
+
+type Mode string
+
+const (
+	DefaultMode Mode = "default"
```
gbraad, line 26: I do not see start options using these values, or the config help.
We need to somehow print possible options with help/description?

gbraad:
> Run crc start --enable-experimental-features, then crc daemon running in an other console.

Aren't we missing the `NetworkMode` for startup of the VM?

pkg/crc/cluster/cluster.go
guillaumerose: 300 is by security. libvirt driver wait time is 3 * 60sec, hyperkit 2 * 60sec, hyperv driver is infinite (https://github.com/code-ready/machine/blob/master/drivers/hyperv/hyperv.go#L314 - lol) !

pkg/crc/machine/libvirt/driver_linux.go
guillaumerose: yes it was just to make it obvious for the reader

go.mod
gbraad: Ah, check. I thought these were related. Confusing names ;-)

gbraad:
```
$ make cross
```

does not do `setuid` ?

Makefile
gbraad: `make cross` and `make release` do not perform the `make install` target, so this will not be part of the release binary and tarball.

go.mod
gbraad: Added comments to that PR. While positive, we need to see what can be done when this is not merged soon-ish

gbraad: Approved: https://github.com/miekg/dns/pull/1174

cmd/crc/cmd/daemon.go
guillaumerose: I retest it multiples times and I can see DNS requests for the `crc-.*-master-0.crc.testing.` domain but even if it doesn't work, the cluster seems to work fine. 🤔

pkg/crc/machine/client.go
```diff
@@ -18,11 +19,12 @@ type Client interface {
...
+	name        string
+	debug       bool
+	networkMode network.Mode
 }
 
-func NewClient(name string, debug bool) Client {
+func NewClient(name string, debug bool, networkMode network.Mode) Client {
 	return &client{
 		name:  name,
 		debug: debug,
```
cfergeau, line 30: This is missing a `networkMode: networkMode`,`

cfergeau:
I add that change
```
--- a/pkg/crc/cluster/csr.go
+++ b/pkg/crc/cluster/csr.go
@@ -21,7 +21,7 @@ func WaitForOpenshiftResource(ocConfig oc.Config, resource string) error {
                logging.Debug(stdout)
                return nil
        }
-       return crcerrors.RetryAfter(80*time.Second, waitForAPIServer, time.Second)
+       return crcerrors.RetryAfter(160*time.Second, waitForAPIServer, time.Second)
 }
```
otherwise I'm getting timeouts during startup  (tested with 4.6 so it might be unrelated to the vsock changes)

pkg/crc/preflight/preflight_linux.go
```diff
@@ -106,18 +125,72 @@ var libvirtPreflightChecks = [...]Check{
...
+	}
+	return nil
+}
+
+func fixVsock() error {
+	_, _, err := crcos.RunWithPrivilege("modprobe vhost_vsock", "modprobe", "vhost_vsock")
+	if err != nil {
+		return err
+	}
+	_, _, err = crcos.RunWithPrivilege("chown /dev/vsock", "chown", "root:libvirt", "/dev/vsock")
```
cfergeau, line 165: Fwiw, this is a bit confusing, even if we set the group to `libvirt`, this has nothing to do with libvirt, all that we want is that the user running `crc daemon` can read/write `/dev/vsock`. Since another preflight check ensures this user is part of the `libvirt` group, this ends up doing the right thing. But the group name could be anything, the libvirt daemon/library/... does not care about the `/dev/vsock` permissions.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -106,18 +125,72 @@ var libvirtPreflightChecks = [...]Check{
...
+		return errors.New("cannot cast info")
+	}
+	if info.Mode()&0060 == 0 {
+		return errors.New("/dev/vsock doesn't have the right permissions")
+	}
+	return nil
+}
+
+func fixVsock() error {
+	_, _, err := crcos.RunWithPrivilege("modprobe vhost_vsock", "modprobe", "vhost_vsock")
```
cfergeau, line 161: This won't persist on reboot, which means `crc setup` will need to be run again after rebooting.

cfergeau:
I was not successful trying this with a 4.6.0-rc3 bundle, the authentication operator never comes up because `curl -k https://oauth-openshift.apps-crc.testing/healthz` takes 15 seconds when done inside a container. It is the DNS resolution which is taking that long. This was tested on a RHEL8 machine.

pkg/crc/api/api.go
```diff
@@ -6,26 +6,32 @@ import (
...
 
-func createAPIServerWithListener(listener net.Listener, newConfig newConfigFunc, handler RequestHandler) (CrcAPIServer, error) {
+func createAPIServerWithListener(listener net.Listener, config newConfigFunc, machine newMachineFunc) (CrcAPIServer, error) {
 	apiServer := CrcAPIServer{
 		listener:               listener,
-		newConfig:              newConfig,
 		clusterOpsRequestsChan: make(chan clusterOpsRequest, 10),
-		handler:                handler,
+		handlerFactory: func() (RequestHandler, error) {
+			cfg, err := config()
```
anjannath, line 26: Won't this re-create the issue we faced earlier in https://github.com/code-ready/crc/issues/1557? we solved that by re-creating the config for each handler

guillaumerose: Handler and his fields are recreated for each request, so it should be fine.

pkg/crc/preflight/preflight_linux.go
guillaumerose: very good point! Should I add a file in /etc/modprobe.d ?

pkg/crc/preflight/preflight_linux.go
guillaumerose: Yes totally true. I reused the group libvirt but I can pick something else. Is there an other good group ? Should I create a new one like vsock ?

pkg/crc/machine/client.go
guillaumerose: thanks, fixed.

cfergeau:
> I was not successful trying this with a 4.6.0-rc3 bundle, the authentication operator never comes up because `curl -k https://oauth-openshift.apps-crc.testing/healthz` takes 15 seconds when done inside a container. It is the DNS resolution which is taking that long. This was tested on a RHEL8 machine.

I redid the same tests with a fresh rhel8.2 install, and the 4.6.1 bundle, and I'm having the same issue, DNS resolution taking 15s when using curl, resulting in various timeouts, and the authentication operator never coming up.

cfergeau:
Wanted to try this on fedora 33:
```
INFO Creating CodeReady Containers VM for OpenShift 4.6.1...
Error creating machine: Error creating the VM: Error creating machine: Error in driver during machine creation: virError(Code=1, Domain=10, Message='internal error: qemu unexpectedly closed the monitor : 2020-10-30T12:50:45.145226Z qemu-system-x86_64: -device vhost-vsock-pci,id=vsock0,guest-cid=3,vhostfd=30,bus=pci.0,addr=0x7: device is modern-only, use disable-legacy=on')
```
probably need to adjust the libvirt XML somehow.

cfergeau:
```
    <vsock model='virtio-non-transitional'>
      <cid auto='yes'/>
    </vsock>
```
(`virtio-non-transitional` instead of `virtio`) works, see https://libvirt.org/formatdomain.html#elementsVirtioTransitional for more details.
I suspect this is a strong suggestion to migrate to a q35 machine type instead of `<type arch='x86_64' machine='pc-i440fx-5.1'>`

pkg/crc/api/api.go
anjannath: Ahh, okay in here: https://github.com/code-ready/crc/pull/1526/files#diff-1b652f0fd14d465febc80f3ed2ed4b0f9e5a5cce6ed35c06869acd0bda0d95c4R61
Thanks!

guillaumerose:
Only missing part now is the DNS resolution of the master node (crc-zqfk6-master-0.crc.testing.) but it seems to work fine without it.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -106,18 +127,91 @@ var libvirtPreflightChecks = [...]Check{
...
+	}
+	getcap, err := exec.Command("getcap", binary).Output()
+	if err != nil {
+		return err
+	}
+	if !strings.Contains(string(getcap), "cap_net_bind_service+eip") {
+		return fmt.Errorf("capabilities are not correct for %s", binary)
+	}
+	info, err := os.Stat("/dev/vsock")
+	if err != nil {
```
cfergeau, line 153: Fwiw, I would put the `getcap` check and the `/dev/vsock` check in 2 separate preflights. One might want to avoid one of the tests while keeping the other.

cfergeau: The checks should also be in `preflight_checks_linux.go` even if the separation is really artificial. When done as part of `crc start`, this test happens too late as the daemon is already supposed to be running at that point, so even `crc setup && crc start` won't fix the issue, even if the preflight will nolonger fail.

cfergeau (commented):
```
$ git show --stat ":/Add userland tcpip"  ":^vendor"
commit 3d98c7a68dafd1d1d1af88fac38a4c7447362727
Author: Guillaume Rose <gurose@redhat.com>
Date:   Wed Sep 30 15:12:36 2020 +0200

    Add userland tcpip stack for the VM
[...]
 30 files changed, 698 insertions(+), 190 deletions(-)
```
is quite bad, a 700 lines change can most likely be described in more than 36 characters.

I did not review everything in detail (especially not the preflight changes), but a couple of notes from initial testing:
- having to start the daemon separately is not user-friendly (but this should not be very hard to change)
- it does not support at all new routes, which means following https://docs.openshift.com/container-platform/4.6/cli_reference/developer_cli_odo/creating-a-multicomponent-application-with-odo.html won't work at all. 2 things are missing: a new entry in `/etc/hosts`, and a redirection for port 80. The former is easy to workaround by adding the entry manually, but the lack of port 80 redirection is harder to deal with. It probably should be redirected by default together with port 443. This may be good enough for initial wider testing for people having issues with the regular DNS handling, but this still needs improving. https://github.com/code-ready/crc/pull/1537 would help with that, but if it's complicated to integrate right now, we could also have a manual command to notify crc that we need a new route (ie a new hostname, and possibly a new port to redirect).

Some limitations of this solution is if the cluster user wants to use a port which is already in use on the host machine (which could be happening with 443/80, and is already the case with port 22 being mapped to 2222), and also https://github.com/code-ready/crc/issues/1627 describes a situation where `goodhosts` cannot be used as is.

cmd/crc/cmd/daemon.go
cfergeau: > It remains the hardcoded domain .crc.testing. I guess we can keep it as it is for the moment. It won't change in a near future.

This hardcoded domain seems like it could be avoided with a `crc route add` command, it could be called early by `machine.Start()`  once it knows which domains the cluster needs.

cfergeau:
> > I was not successful trying this with a 4.6.0-rc3 bundle, the authentication operator never comes up because `curl -k https://oauth-openshift.apps-crc.testing/healthz` takes 15 seconds when done inside a container. It is the DNS resolution which is taking that long. This was tested on a RHEL8 machine.
> 
> I redid the same tests with a fresh rhel8.2 install, and the 4.6.1 bundle, and I'm having the same issue, DNS resolution taking 15s when using curl, resulting in various timeouts, and the authentication operator never coming up.

And after updating to the latest iteration of the code, these timeouts are gone, not sure what was going on before.

cmd/crc/cmd/daemon.go
```diff
@@ -21,10 +36,83 @@ var daemonCmd = &cobra.Command{
...
+				"apps-crc.testing.":                 net.ParseIP("192.168.127.2"),
+				"etcd-0.crc.testing.":               net.ParseIP("192.168.127.2"),
+				"api.crc.testing.":                  net.ParseIP("192.168.127.2"),
+				"api-int.crc.testing.":              net.ParseIP("192.168.127.2"),
+				"oauth-openshift.apps-crc.testing.": net.ParseIP("192.168.127.2"),
+			},
+			Forwards: map[string]string{
+				":2222": "192.168.127.2:22",
+				":6443": "192.168.127.2:6443",
+				":443":  "192.168.127.2:443",
```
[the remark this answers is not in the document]
guillaumerose: daemon will print error on stderr if it can't bind. This can be improved!

cmd/crc/cmd/daemon.go
```diff
@@ -21,10 +36,83 @@ var daemonCmd = &cobra.Command{
 		logging.CloseLogging()
 		logging.InitLogrus(logging.LogLevel, constants.DaemonLogFilePath)
 
-		runDaemon()
+		go runDaemon()
```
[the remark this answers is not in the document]
guillaumerose: yes, error handling is bad here also. If the daemon goroutine crash, the daemon will stay up.

cmd/crc/cmd/daemon.go
```diff
@@ -21,10 +36,83 @@ var daemonCmd = &cobra.Command{
...
+		go func() {
+			if err := http.Serve(ln, vn.Mux()); err != nil {
+				errCh <- err
+			}
+		}()
+	}
+	go func() {
+		for {
+			fmt.Printf("%v sent to the VM, %v received from the VM\n", units.HumanSize(float64(vn.BytesSent())), units.HumanSize(float64(vn.BytesReceived())))
+			time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
guillaumerose: yes added!

cmd/crc/cmd/daemon.go
```diff
@@ -21,10 +36,83 @@ var daemonCmd = &cobra.Command{
...
+		if err := run(&types.Configuration{
+			Debug:             isDebugLog(),
+			CaptureFile:       captureFile(),
+			MTU:               4000,
+			Subnet:            "192.168.127.0/24",
+			GatewayIP:         "192.168.127.1",
+			GatewayMacAddress: "\x5A\x94\xEF\xE4\x0C\xDD",
+			DNSRecords: map[string]net.IP{
+				"gateway.crc.testing.":              net.ParseIP("192.168.127.1"),
+				"apps-crc.testing.":                 net.ParseIP("192.168.127.2"),
```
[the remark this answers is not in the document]
guillaumerose: yes, dns record matching works with suffix here https://github.com/guillaumerose/gvisor-tap-vsock/blob/master/pkg/services/dns/dns.go#L52

I simplified the other records.

cmd/crc/cmd/daemon.go
```diff
@@ -21,10 +36,83 @@ var daemonCmd = &cobra.Command{
...
+				"etcd-0.crc.testing.":               net.ParseIP("192.168.127.2"),
```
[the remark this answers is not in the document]
guillaumerose: removed!

cmd/crc/cmd/daemon.go
```diff
@@ -21,10 +36,83 @@ var daemonCmd = &cobra.Command{
...
+			endpoints = append(endpoints, fmt.Sprintf("unix://%s", apiSock))
+			if runtime.GOOS == "linux" {
+				endpoints = append(endpoints, transport.DefaultURL)
+			}
+		}
+
+		if err := run(&types.Configuration{
+			Debug:             isDebugLog(),
+			CaptureFile:       captureFile(),
+			MTU:               4000,
```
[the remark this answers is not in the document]
guillaumerose: performance are a little bit better with a large mtu. I added a comment.

pkg/crc/services/dns/dns.go
```diff
@@ -21,6 +22,30 @@ func init() {
...
+		// override resolv.conf file
+		resolvFileValues := network.ResolvFileValues{
+			SearchDomains: []network.SearchDomain{
+				{
+					Domain: fmt.Sprintf("%s.%s", serviceConfig.Name, serviceConfig.BundleMetadata.ClusterInfo.BaseDomain),
+				},
+			},
+			NameServers: []network.NameServer{
+				{
+					IPAddress: "192.168.127.1",
```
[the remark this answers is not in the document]
guillaumerose: yes this is the gateway of the virtual network. I extracted a constant.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -106,18 +127,91 @@ var libvirtPreflightChecks = [...]Check{
...
+	fixDescription:   "Checking if vsock is correctly configured",
+	fix:              fixVsock,
+}
+
+func checkVsock() error {
+	binary, err := os.Executable()
+	if err != nil {
+		return err
+	}
+	getcap, err := exec.Command("getcap", binary).Output()
```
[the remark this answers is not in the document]
guillaumerose: done

pkg/crc/preflight/preflight_linux.go
```diff
@@ -106,18 +127,91 @@ var libvirtPreflightChecks = [...]Check{
...
+		return errors.New("cannot cast info")
+	}
+	if info.Mode()&0060 == 0 {
+		return errors.New("/dev/vsock doesn't have the right permissions")
+	}
+	return nil
+}
+
+func fixVsock() error {
+	binary, err := os.Executable()
```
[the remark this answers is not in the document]
guillaumerose: done

pkg/crc/preflight/preflight_linux.go
```diff
@@ -106,18 +127,91 @@ var libvirtPreflightChecks = [...]Check{
...
+var vsockPreflightChecks = Check{
+	configKeySuffix:  "check-vsock",
+	checkDescription: "Checking if vsock is correctly configured",
+	check:            checkVsock,
+	fixDescription:   "Checking if vsock is correctly configured",
+	fix:              fixVsock,
+}
+
+func checkVsock() error {
+	binary, err := os.Executable()
```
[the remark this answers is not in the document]
guillaumerose: done

pkg/crc/machine/machine.go
```diff
@@ -426,6 +446,20 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 	}, nil
 }
 
+func getSSHPort(vsockNetwork bool) int {
+	if vsockNetwork {
+		return 2222
```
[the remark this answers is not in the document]
guillaumerose: done!

pkg/crc/machine/machine.go
```diff
@@ -249,6 +250,18 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 		}
 	}
 
+	if runtime.GOOS == "darwin" {
+		dst := filepath.Join(constants.MachineInstanceDir, client.name, "00000002.00000400")
```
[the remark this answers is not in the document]
guillaumerose: added a comment and extracted the function.

```
// makeDaemonVisibleToHyperkit crc daemon is launched in background and doesn't know where hyperkit is running.
// In order to vsock to work with hyperkit, we need to put the unix socket in the hyperkit working directory with a
// special name. The name is the hex representation of the cid and the vsock port.
// This function adds the unix socket in the hyperkit directory.
```

pkg/crc/machine/hyperkit/driver_darwin.go
```diff
@@ -19,5 +20,7 @@ func CreateHost(machineConfig config.MachineConfig) *hyperkit.Driver {
 	hyperkitDriver.InitrdPath = machineConfig.Initramfs
 	hyperkitDriver.HyperKitPath = filepath.Join(constants.CrcBinDir, "hyperkit")
 
+	hyperkitDriver.VMNet = machineConfig.NetworkMode == network.DefaultMode
```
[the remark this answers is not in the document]
guillaumerose: my editor complains when I add the parenthesis :)

pkg/crc/services/dns/dns_windows.go
```diff
@@ -18,6 +21,13 @@ const (
 )
 
 func runPostStartForOS(serviceConfig services.ServicePostStartConfig) error {
+	if serviceConfig.NetworkMode == network.VSockMode {
+		return goodhosts.UpdateHostsFile(serviceConfig.IP, serviceConfig.BundleMetadata.GetAPIHostname(),
+			serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift"),
+			serviceConfig.BundleMetadata.GetAppHostname("console-openshift-console"),
+			serviceConfig.BundleMetadata.GetAppHostname("default-route-openshift-image-registry"))
```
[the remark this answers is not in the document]
guillaumerose: extracted in a function

pkg/crc/goodhosts/goodhosts_windows.go
```diff
@@ -0,0 +1,12 @@
+package goodhosts
+
+import (
+	"strings"
+
+	"github.com/code-ready/crc/pkg/os/windows/powershell"
+)
+
+func execute(args ...string) error {
+	_, _, err := powershell.ExecuteAsAdmin("modifying host file", strings.Join(append([]string{goodhostPath}, args...), " "))
```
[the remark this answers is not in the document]
guillaumerose: done

pkg/crc/preflight/preflight_windows.go
```diff
@@ -83,15 +90,55 @@ var traySetupChecks = [...]Check{
...
+		checkDescription: "Checking if vsock is correctly configured",
+		check:            checkVsock,
+		fixDescription:   "Checking if vsock is correctly configured",
+		fix:              fixVsock,
+	},
+}
+
+const (
+	registryDirectory = `HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Virtualization\GuestCommunicationServices`
+	registryKey       = "00000400-FACB-11E6-BD58-64006A7986D3"
```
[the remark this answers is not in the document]
guillaumerose: added a comment.

This is what activates the communication between the vm and the host

pkg/crc/services/dns/dns.go
```diff
@@ -21,6 +22,30 @@ func init() {
 }
 
 func RunPostStart(serviceConfig services.ServicePostStartConfig) error {
+	if serviceConfig.NetworkMode == network.VSockMode {
+		if runtime.GOOS == "linux" || runtime.GOOS == "darwin" {
```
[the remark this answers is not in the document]
guillaumerose: done

pkg/crc/api/api.go
```diff
@@ -6,26 +6,32 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/code-ready/crc/pkg/crc/machine"
-
 	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
-func CreateAPIServer(socketPath string, newConfig newConfigFunc, client machine.Client) (CrcAPIServer, error) {
+func CreateAPIServer(socketPath string, config newConfigFunc, machine newMachineFunc) (CrcAPIServer, error) {
```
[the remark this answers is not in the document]
guillaumerose: done!

guillaumerose:
I think I went on all the comments. A lot is still need to be done for sure..

guillaumerose:
/retest
retest this please

gbraad (approved):

cfergeau:
The log of ` Add userland tcpip stack for the VM` is a missed opportunity for giving a lot of details about what this is about, how this works, ... It was definitely not eligible to a one-line commit log.

PR #1528, opened 2020-09-17
Host IP is not necessary anymore in startup sequence

guillaumerose:
Less code!


pkg/crc/machine/machine.go
```diff
@@ -284,20 +284,6 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 		return startError(startConfig.Name, "Error getting the IP", err)
 	}
 
-	var hostIP string
```
praveenkumar, line 287: It was added when we tried to use the coredns and never removed this dead code :+1:

praveenkumar (approved):

praveenkumar:
There is issue with linter which need to fixed before merge.

```
: finding mvdan.cc/lint v0.0.0-20170908181259-adc824a0674b
GOOS=darwin /Users/distiller/go/bin/golangci-lint run
pkg/crc/machine/machine.go:119:10: ineffectual assignment to `err` (ineffassign)
	exists, err := client.Exists(startConfig.Name)
	        ^
```

/lgtm cancel

pkg/crc/machine/machine.go
cfergeau: can this be added to the commit log?

guillaumerose:
Interesting lint error. I didn't change anything in this place. :thinking:

pkg/crc/machine/machine.go
```diff
@@ -284,20 +284,6 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
...
-	determineHostIP := func() error {
-		hostIP, err = network.DetermineHostIP(instanceIP)
```
guillaumerose, line 289: oh! this err was written in place of this err declaration:

```
exists, err := client.Exists(startConfig.Name)
```

lint error appear now because we never touch this variable.

pkg/crc/machine/machine.go
guillaumerose: done :+1:

praveenkumar (approved):

PR #1529, opened 2020-09-18
Better display MultiError when errors are the same

guillaumerose:
Before:

```
Failed to renew TLS certificates: please check if a newer CodeReady Containers release is available: Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
Temporary Error: No Pending CSR
...
my console is full :)
``` 

After

```
Failed to renew TLS certificates: please check if a newer CodeReady Containers release is available: Temporary error: No Pending CSR (x37)
Temporary error: ssh command error:
command : oc get csr --context admin --cluster crc --kubeconfig /opt/kubeconfig
err     : exit status 1
output  : The connection to the server api.crc.testing:6443 was refused - did you specify the right host or port?
 (x1)
Temporary error: No Pending CSR (x82)
```

Slightly better as it fits on one screen.

### DIFF
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -36,7 +36,7 @@ type RetriableError struct {
 }
 
 func (r RetriableError) Error() string {
-	return "Temporary Error: " + r.Err.Error()
+	return "Temporary error: " + r.Err.Error()
 }
 
 // RetryAfter retries a number of attempts, after a delay
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -12,22 +12,43 @@ type MultiError struct {
 	Errors []error
 }
 
+func (m MultiError) Error() string {
+	if len(m.Errors) == 0 {
+		return ""
+	}
+	if len(m.Errors) == 1 {
+		return m.Errors[0].Error()
+	}
+
+	var aggregatedErrors []string
+
+	count := 1
+	current := m.Errors[0].Error()
+	for i := 1; i < len(m.Errors); i++ {
+		if m.Errors[i].Error() == current {
+			count++
+			continue
+		}
+		aggregatedErrors = append(aggregatedErrors, errorWithCount(current, count))
+		count = 1
+		current = m.Errors[i].Error()
+	}
+	aggregatedErrors = append(aggregatedErrors, errorWithCount(current, count))
+
+	return strings.Join(aggregatedErrors, "\n")
+}
+
 func (m *MultiError) Collect(err error) {
 	if err != nil {
 		m.Errors = append(m.Errors, err)
 	}
 }
 
-func (m MultiError) ToError() error {
-	if len(m.Errors) == 0 {
-		return nil
+func errorWithCount(current string, count int) string {
+	if count == 1 {
+		return current
 	}
-
-	errStrings := []string{}
-	for _, err := range m.Errors {
-		errStrings = append(errStrings, err.Error())
-	}
-	return fmt.Errorf(strings.Join(errStrings, "\n"))
+	return fmt.Sprintf("%s (x%d)", current, count)
 }
 
 // RetriableError is an error that can be tried again
@@ -40,26 +61,24 @@ func (r RetriableError) Error() string {
 }
 
 // RetryAfter retries a number of attempts, after a delay
-func RetryAfter(attempts int, callback func() error, d time.Duration) (err error) {
+func RetryAfter(attempts int, callback func() error, d time.Duration) error {
 	m := MultiError{}
 	for i := 0; i < attempts; i++ {
 		if i > 0 {
 			logging.Debugf("retry loop: attempt %d out of %d", i, attempts)
 		}
-		err = callback()
+		err := callback()
 		if err == nil {
 			return nil
 		}
 		m.Collect(err)
 		if _, ok := err.(*RetriableError); !ok {
 			logging.Debugf("non-retriable error: %v", err)
-			return m.ToError()
+			return m
 		}
 		logging.Debugf("error: %v - sleeping %s", err, d)
 		time.Sleep(d)
 	}
-	err = m.ToError()
-
 	logging.Debugf("RetryAfter timeout after %d tries", attempts)
-	return err
+	return m
 }
--- a/pkg/crc/errors/multierror_test.go
+++ b/pkg/crc/errors/multierror_test.go
@@ -33,7 +33,7 @@ func TestRetryAfterMaxAttempts(t *testing.T) {
 		calls++
 		return &RetriableError{Err: errors.New("failed")}
 	}, 0)
-	assert.EqualError(t, ret, "Temporary Error: failed\nTemporary Error: failed\nTemporary Error: failed")
+	assert.EqualError(t, ret, "Temporary error: failed\nTemporary error: failed\nTemporary error: failed")
 	assert.Equal(t, 3, calls)
 }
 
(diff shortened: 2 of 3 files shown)


pkg/crc/errors/multierror.go
```diff
@@ -22,12 +22,30 @@ func (m MultiError) ToError() error {
...
+	}
+
+	var aggregatedErrors []string
 
-	errStrings := []string{}
-	for _, err := range m.Errors {
-		errStrings = append(errStrings, err.Error())
+	count := 1
+	current := m.Errors[0].Error()
+	for i := 1; i < len(m.Errors); i++ {
```
cfergeau, line 33: Sorry, not going to be a real review, but more a list of design questions
Wondering if  `for i, err := range(m.Errors) {` could be used somehow? (I know you to skip the first item).
Would an `AggregatedError` type be better than `errorWithCount` returning just a string? This could  even implement `Unwrap()` ( https://blog.golang.org/go1.13-errors )

Why isn't `Multierror` implementing `error` directly rather than creating a fake error with a flattened string?

pkg/crc/errors/multierror.go
```diff
@@ -22,12 +22,30 @@ func (m MultiError) ToError() error {
...
+		if m.Errors[i].Error() == current {
+			count++
+			continue
+		}
+		aggregatedErrors = append(aggregatedErrors, errorWithCount(current, count))
+		count = 1
+		current = m.Errors[i].Error()
 	}
-	return fmt.Errorf(strings.Join(errStrings, "\n"))
+	aggregatedErrors = append(aggregatedErrors, errorWithCount(current, count))
```
cfergeau, line 42: I think I'd use a `strings.Builder` rather than a `[]string` + `append`/`strings.Join()`

pkg/crc/errors/multierror.go
```diff
@@ -22,12 +22,30 @@ func (m MultiError) ToError() error {
...
+
+	var aggregatedErrors []string
 
-	errStrings := []string{}
-	for _, err := range m.Errors {
-		errStrings = append(errStrings, err.Error())
+	count := 1
+	current := m.Errors[0].Error()
+	for i := 1; i < len(m.Errors); i++ {
+		if m.Errors[i].Error() == current {
```
cfergeau, line 34: Maybe also check `reflect.TypeOf` in addition to the string value?

cfergeau (commented):
Patch as it is looks good (ie no obvious bugs/flaws), however I had plenty of design questions, I feel we could use more structured error types rather than just strings.

pkg/crc/errors/multierror_test.go
```diff
@@ -49,3 +49,24 @@ func TestRetryAfterSuccessAfterFailures(t *testing.T) {
...
+	assert.Equal(t, "Temporary Error: No Pending CSR (x4)", MultiError{
+		Errors: []error{
+			errors.New("Temporary Error: No Pending CSR"),
+			errors.New("Temporary Error: No Pending CSR"),
+			errors.New("Temporary Error: No Pending CSR"),
+			errors.New("Temporary Error: No Pending CSR"),
+		},
+	}.ToError().Error())
+
+	assert.Equal(t, "No Pending CSR (x2)\nConnection refused (x2)\nNo Pending CSR (x1)", MultiError{
```
cfergeau, line 63: Maybe the 'x1' could be omitted?

pkg/crc/errors/multierror.go
praveenkumar: in the background of `Join` implementation, it does uses the `Builder` itself.
```golang
// Join concatenates the elements of its first argument to create a single string. The separator
// string sep is placed between elements in the resulting string.
func Join(elems []string, sep string) string {
	switch len(elems) {
	case 0:
		return ""
	case 1:
		return elems[0]
	}
	n := len(sep) * (len(elems) - 1)
	for i := 0; i < len(elems); i++ {
		n += len(elems[i])
	}

	var b Builder
	b.Grow(n)
	b.WriteString(elems[0])
	for _, s := range elems[1:] {
		b.WriteString(sep)
		b.WriteString(s)
	}
	return b.String()
}
```

pkg/crc/errors/multierror_test.go
guillaumerose: done :+1:

pkg/crc/errors/multierror.go
guillaumerose: Error() always return a string. I don't see what you mean.

pkg/crc/errors/multierror.go
guillaumerose: I removed the ToError() function and now MultiError is a proper error. For the Unwrap() function, I did nothing. Maybe in follow up?

cfergeau: Yeah, this really was more random ramblings where you can pick and choose what makes sense than a list of things which must be done right now :)

pkg/crc/errors/multierror.go
cfergeau: I was wondering if something like this on top of your changes could be useful so that we check both error type and string:

```
diff --git a/pkg/crc/errors/multierror.go b/pkg/crc/errors/multierror.go
index a7d11afd..a28f5866 100644
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -2,6 +2,7 @@ package errors
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
@@ -12,6 +13,25 @@ type MultiError struct {
 	Errors []error
 }
 
+type errorWithCount struct {
+	errorCount int
+	err        error
+}
+
+func (err *errorWithCount) Error() string {
+	if err.errorCount == 1 {
+		return err.err.Error()
+	}
+	return fmt.Sprintf("%s (x%d)", err.err.Error(), err.errorCount)
+}
+
+func equalErr(err1, err2 error) bool {
+	if reflect.TypeOf(err1) != reflect.TypeOf(err2) {
+		return false
+	}
+	return err1.Error() == err2.Error()
+}
+
 func (m MultiError) Error() string {
 	if len(m.Errors) == 0 {
 		return ""
@@ -20,22 +40,27 @@ func (m MultiError) Error() string {
 		return m.Errors[0].Error()
 	}
 
-	var aggregatedErrors []string
+	var aggregatedErrors []error
 
 	count := 1
-	current := m.Errors[0].Error()
+	current := m.Errors[0]
 	for i := 1; i < len(m.Errors); i++ {
-		if m.Errors[i].Error() == current {
+		if equalErr(m.Errors[i], current) {
 			count++
 			continue
 		}
-		aggregatedErrors = append(aggregatedErrors, errorWithCount(current, count))
+		aggregatedErrors = append(aggregatedErrors, &errorWithCount{errorCount: count, err: current})
 		count = 1
-		current = m.Errors[i].Error()
+		current = m.Errors[i]
 	}
-	aggregatedErrors = append(aggregatedErrors, errorWithCount(current, count))
+	aggregatedErrors = append(aggregatedErrors, &errorWithCount{errorCount: count, err: current})
 
-	return strings.Join(aggregatedErrors, "\n")
+	var strBuilder strings.Builder
+	for _, err := range aggregatedErrors {
+		strBuilder.WriteString(err.Error())
+		strBuilder.WriteString("\n")
+	}
+	return strBuilder.String()
 }
 
 func (m *MultiError) Collect(err error) {
@@ -44,13 +69,6 @@ func (m *MultiError) Collect(err error) {
 	}
 }
 
-func errorWithCount(current string, count int) string {
-	if count == 1 {
-		return current
-	}
-	return fmt.Sprintf("%s (x%d)", current, count)
-}
-
 // RetriableError is an error that can be tried again
 type RetriableError struct {
 	Err error
```
(this does not pass `make test` because of an extra `\n` at the end of `Error()` and I'd actually move most of this to `Collect()` rather than `Error()` - maybe I'll do all of that as a followup)

cfergeau (approved):
Still looks good to me

PR #1530, opened 2020-09-18
Update 'hyperkit' binary to newer version

cfergeau:
## Solution/Idea

This PR does some refactoring of the cache/versioning code, and then adds version checks for the hyperkit binary before updating it to a newer version.

## Proposed changes

1. machine drivers/oc/goodhosts caching code is refactored, but there should not be any user-visible changes (apart from slightly different debug output)
2. hyperkit binary is upgraded from `v0.20180403-36-ge3b37f` to `v0.20200224-44-gb54460`
4. ...

## Testing

1. Check that caching (including older version upgrades) of machine drivers/oc/goodhosts did not regress
2. Check that hyperkit is upgraded to a newer version after `crc setup` (won't work as expected at the moment as the newer binary is not uploaded yet to gtihub.com)

### DIFF
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -148,12 +148,11 @@ func (c *Cache) CacheBinary() error {
 }
 
 func (c *Cache) getBinary(destDir string) (string, error) {
-	logging.Debug("Trying to extract oc from crc binary")
+	logging.Debugf("Trying to extract %s from crc binary", c.binaryName)
 	archiveName := filepath.Base(c.archiveURL)
 	destPath := filepath.Join(destDir, archiveName)
 	err := embed.Extract(archiveName, destPath)
 	if err != nil {
-		logging.Debug("Downloading oc")
 		return download.Download(c.archiveURL, destDir, 0600)
 	}
 
@@ -170,12 +169,15 @@ func (c *Cache) CheckVersion() error {
 		return err
 	}
 	if currentVersion != c.version {
-		return &VersionMismatchError{
+		err := &VersionMismatchError{
 			BinaryName:      c.binaryName,
 			CurrentVersion:  currentVersion,
 			ExpectedVersion: c.version,
 		}
+		logging.Debugf("%s", err.Error())
+		return err
 	}
+	logging.Debugf("Found %s version %s", c.binaryName, c.version)
 	return nil
 }
 
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/code-ready/crc/pkg/download"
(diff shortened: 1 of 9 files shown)


praveenkumar:
/hold

we need to first release the machine-driver-hyperkit with new hyperkit version before merging this one.

cfergeau:
> /hold
> 
> we need to first release the machine-driver-hyperkit with new hyperkit version before merging this one.

Are you referring to https://github.com/code-ready/machine-driver-hyperkit/blob/5f2c6ee9b191903b0eb8f1473a40ec7d4a51fc25/pkg/hyperkit/driver.go#L39 ? Or to something else?
If you are referring to the former, I don't think it's strongly tied to the hyperkit binary version we are using, as current crc releases use hyperkit `v0.20180403-36-ge3b37f` while the machine driver's go.mod uses `github.com/moby/hyperkit v0.0.0-20171204115932-858492e3d919`
I've tried the newer binary without machine driver changes and did not notice issues.

cfergeau:
If your comment was referring to https://github.com/code-ready/machine-driver-hyperkit/blob/5f2c6ee9b191903b0eb8f1473a40ec7d4a51fc25/pkg/hyperkit/driver.go#L39 , I'd prefer to delay updating this until we look at adding disk resizing to `machine-driver-hyperkit` as it will be required at this point, and I'm not sure the update is trivial (there might be a hard dependency on `qcow-tool` being available)

pkg/crc/cache/cache.go
```diff
@@ -137,17 +165,20 @@ func (c *Cache) CheckVersion() error {
 	if c.version == "" {
 		return nil
 	}
-	currentVersion, err := c.getVersion()
+	currentVersion, err := c.getVersion(c)
```
guillaumerose, line 168: `c.getVersion(c)` this line seems particularly weird. 

I would expect getVersion to take a binary filepath as parameter and return a version. WDYT? `getVersionGeneric` only needs the binary. Making it static would also help.

praveenkumar:
I was referring to https://github.com/code-ready/crc/pull/1530/commits/4dc811828a98d9541af357aed47fc6b48e79a379 one since now the version check is in place for hyperkit binary and we are downloading this binary from the https://github.com/code-ready/machine-driver-hyperkit/releases/tag/v0.12.7 which contains an old version so if we merge this PR to master and someone test it on mac then he always going to get version mismatch for hyperkit binary

pkg/crc/cache/cache.go
cfergeau: ah yes, should be better, thanks!

cfergeau:
> I was referring to [4dc8118](https://github.com/code-ready/crc/commit/4dc811828a98d9541af357aed47fc6b48e79a379) one since now the version check is in place for hyperkit binary and we are downloading this binary from the https://github.com/code-ready/machine-driver-hyperkit/releases/tag/v0.12.7 which contains an old version so if we merge this PR to master and someone test it on mac then he always going to get version mismatch for hyperkit binary

I dropped the last commit which was the problematic one, this should allow to move forward with the rest of the changes, and to add the missing commit at release time.

guillaumerose (approved):

PR #1532, opened 2020-09-21
Rework cert renewal process

guillaumerose:
cert renewal is not accurate (at least from what I saw). We only have one cert to approve. The second one is automatically approved by the cluster machine approver. This PR removes the race.

### DIFF
--- a/pkg/crc/cluster/cert-renewal.go
+++ b/pkg/crc/cluster/cert-renewal.go
@@ -5,52 +5,64 @@ import (
 	"strings"
 	"time"
 
-	"github.com/code-ready/crc/pkg/crc/errors"
+	crcerrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/oc"
+	"github.com/code-ready/crc/pkg/crc/ssh"
 )
 
-func waitForPendingCsrs(ocConfig oc.Config, signerName string) error {
-	waitForPendingCsr := func() error {
+func waitForPendingCSRs(ocConfig oc.Config, signerName string) error {
+	return crcerrors.RetryAfter(8*time.Minute, func() error {
 		output, _, err := ocConfig.RunOcCommand("get", "csr")
 		if err != nil {
-			return &errors.RetriableError{Err: err}
+			return &crcerrors.RetriableError{Err: err}
 		}
 		if strings.Contains(output, "Pending") && strings.Contains(output, signerName) {
 			return nil
 		}
-		return &errors.RetriableError{Err: fmt.Errorf("No Pending CSR with signerName %s", signerName)}
-	}
-
-	return errors.RetryAfter(8*time.Minute, waitForPendingCsr, time.Second*5)
+		return &crcerrors.RetriableError{Err: fmt.Errorf("No Pending CSR with signerName %s", signerName)}
+	}, time.Second*5)
 }
 
-func WaitAndApprovePendingCSRs(ocConfig oc.Config, client, server bool) error {
-	/* 2 CSRs to approve, one right after kubelet restart, the other one a few dozen seconds after
-	approving the first one
-	- First one is requested by system:serviceaccount:openshift-machine-config-operator:node-bootstrapper
-	- Second one is requested by system:node:<node_name> */
+func ApproveCSRAndWaitForCertsRenewal(sshRunner *ssh.Runner, ocConfig oc.Config, client, server bool) error {
+	// First, kubelet starts and tries to connect to API server. If its certificate is expired, it asks for a new one
+	// Admin needs to approve it. The Kubernetes controller manager will then issue the cert, kubelet will fetch it and use it.
+	// Kubelet stores the cert in /var/lib/kubelet/pki/kubelet-client-current.pem
 	if client {
 		logging.Info("Kubelet client certificate has expired, renewing it... [will take up to 8 minutes]")
-		if err := waitForPendingCsrs(ocConfig, "kubernetes.io/kube-apiserver-client-kubelet"); err != nil {
+		if err := waitForPendingCSRs(ocConfig, kubeletClientSignerName); err != nil {
 			logging.Debugf("Error waiting for pending kube-apiserver-client-kubelet CSR: %v", err)
 			return err
 		}
-		if err := ApproveNodeCSR(ocConfig, "kubernetes.io/kube-apiserver-client-kubelet"); err != nil {
+		if err := approveNodeCSR(ocConfig, kubeletClientSignerName); err != nil {
+			logging.Debugf("Error approving pending kube-apiserver-client-kubelet CSR: %v", err)
+			return err
+		}
+		if err := crcerrors.RetryAfter(5*time.Minute, waitForCertRenewal(sshRunner, kubeletClientCert), time.Second*5); err != nil {
 			logging.Debugf("Error approving pending kube-apiserver-client-kubelet CSR: %v", err)
 			return err
 		}
 	}
+	// API server needs to connect to kubelet for some features like logs, port forwards. This communication is backed by a cert
+	// store in /var/lib/kubelet/pki/kubelet-server-current.pem
+	// After kubelet connected to the API server, if the serving cert is expireed, kubelet asks for a new CSR.
+	// This CSR is automatically approved by the cluster-machine-approver. The k8s controller manager issues the cert and kubelet fetches it.
 	if server {
 		logging.Info("Kubelet serving certificate has expired, waiting for automatic renewal... [will take up to 8 minutes]")
-		if err := waitForPendingCsrs(ocConfig, "kubernetes.io/kubelet-serving"); err != nil {
-			logging.Debugf("Error waiting for pending kubelet-serving CSR: %v", err)
-			return err
-		}
-		if err := ApproveNodeCSR(ocConfig, "kubernetes.io/kubelet-serving"); err != nil {
-			logging.Debugf("Error approving pending kubelet-serving CSR: %v", err)
-			return err
-		}
+		return crcerrors.RetryAfter(5*time.Minute, waitForCertRenewal(sshRunner, kubeletServerCert), time.Second*5)
 	}
 	return nil
+}
+
+func waitForCertRenewal(sshRunner *ssh.Runner, cert string) func() error {
+	return func() error {
+		expired, err := checkCertValidity(sshRunner, cert)
+		if err != nil {
+			return err
+		}
+		if !expired {
+			return nil
+		}
+		return &crcerrors.RetriableError{Err: fmt.Errorf("certificate %s still expired", cert)}
+	}
 }
--- a/pkg/crc/cluster/cert-renewal.go
+++ b/pkg/crc/cluster/cert-renewal.go
@@ -8,8 +8,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/oc"
-	"github.com/code-ready/crc/pkg/crc/ssh"
-	"github.com/code-ready/crc/pkg/crc/systemd"
 )
 
 func waitForPendingCsrs(ocConfig oc.Config) error {
@@ -31,43 +29,26 @@ func waitForPendingCsrs(ocConfig oc.Config) error {
 	return errors.RetryAfter(8*time.Minute, waitForPendingCsr, time.Second*5)
 }
 
-func RegenerateCertificates(sshRunner *ssh.Runner, ocConfig oc.Config) error {
-	sd := systemd.NewInstanceSystemdCommander(sshRunner)
-	err := sd.Start("kubelet")
-	if err != nil {
-		logging.Debugf("Error starting kubelet service: %v", err)
-		return err
-	}
-	// FIXME: regression compared to previous API
-	startedKubelet := false
-	if startedKubelet {
-		defer sd.Stop("kubelet") //nolint:errcheck
-	}
+func RegenerateCertificates(ocConfig oc.Config) error {
 	/* 2 CSRs to approve, one right after kubelet restart, the other one a few dozen seconds after
 	approving the first one
 	- First one is requested by system:serviceaccount:openshift-machine-config-operator:node-bootstrapper
 	- Second one is requested by system:node:<node_name> */
-	err = waitForPendingCsrs(ocConfig)
-	if err != nil {
+	if err := waitForPendingCsrs(ocConfig); err != nil {
 		logging.Debugf("Error waiting for first pending (node-bootstrapper) CSR: %v", err)
 		return err
 	}
-	err = ApproveNodeCSR(ocConfig)
-	if err != nil {
+	if err := ApproveNodeCSR(ocConfig); err != nil {
 		logging.Debugf("Error approving first pending (node-bootstrapper) CSR: %v", err)
 		return err
 	}
-
-	err = waitForPendingCsrs(ocConfig)
-	if err != nil {
+	if err := waitForPendingCsrs(ocConfig); err != nil {
 		logging.Debugf("Error waiting for second pending (system:node) CSR: %v", err)
 		return err
 	}
-	err = ApproveNodeCSR(ocConfig)
-	if err != nil {
+	if err := ApproveNodeCSR(ocConfig); err != nil {
 		logging.Debugf("Error approving second pending (system:node) CSR: %v", err)
 		return err
 	}
-
 	return nil
 }
--- a/pkg/crc/cluster/cert-renewal.go
+++ b/pkg/crc/cluster/cert-renewal.go
@@ -2,7 +2,7 @@ package cluster
 
 import (
 	"fmt"
-	"regexp"
+	"strings"
 	"time"
 
 	"github.com/code-ready/crc/pkg/crc/errors"
@@ -10,45 +10,47 @@ import (
 	"github.com/code-ready/crc/pkg/crc/oc"
 )
 
-func waitForPendingCsrs(ocConfig oc.Config) error {
+func waitForPendingCsrs(ocConfig oc.Config, signerName string) error {
 	waitForPendingCsr := func() error {
 		output, _, err := ocConfig.RunOcCommand("get", "csr")
 		if err != nil {
 			return &errors.RetriableError{Err: err}
 		}
-		matched, err := regexp.MatchString("Pending", output)
-		if err != nil {
-			return &errors.RetriableError{Err: err}
+		if strings.Contains(output, "Pending") && strings.Contains(output, signerName) {
+			return nil
 		}
-		if !matched {
-			return &errors.RetriableError{Err: fmt.Errorf("No Pending CSR")}
-		}
-		return nil
+		return &errors.RetriableError{Err: fmt.Errorf("No Pending CSR with signerName %s", signerName)}
 	}
 
 	return errors.RetryAfter(8*time.Minute, waitForPendingCsr, time.Second*5)
 }
 
-func RegenerateCertificates(ocConfig oc.Config) error {
+func WaitAndApprovePendingCSRs(ocConfig oc.Config, client, server bool) error {
 	/* 2 CSRs to approve, one right after kubelet restart, the other one a few dozen seconds after
 	approving the first one
 	- First one is requested by system:serviceaccount:openshift-machine-config-operator:node-bootstrapper
 	- Second one is requested by system:node:<node_name> */
-	if err := waitForPendingCsrs(ocConfig); err != nil {
-		logging.Debugf("Error waiting for first pending (node-bootstrapper) CSR: %v", err)
-		return err
+	if client {
+		logging.Info("Kubelet client certificate has expired, renewing it... [will take up to 8 minutes]")
+		if err := waitForPendingCsrs(ocConfig, "kubernetes.io/kube-apiserver-client-kubelet"); err != nil {
+			logging.Debugf("Error waiting for pending kube-apiserver-client-kubelet CSR: %v", err)
+			return err
+		}
+		if err := ApproveNodeCSR(ocConfig, "kubernetes.io/kube-apiserver-client-kubelet"); err != nil {
+			logging.Debugf("Error approving pending kube-apiserver-client-kubelet CSR: %v", err)
+			return err
+		}
 	}
-	if err := ApproveNodeCSR(ocConfig); err != nil {
-		logging.Debugf("Error approving first pending (node-bootstrapper) CSR: %v", err)
-		return err
-	}
-	if err := waitForPendingCsrs(ocConfig); err != nil {
-		logging.Debugf("Error waiting for second pending (system:node) CSR: %v", err)
-		return err
-	}
-	if err := ApproveNodeCSR(ocConfig); err != nil {
-		logging.Debugf("Error approving second pending (system:node) CSR: %v", err)
-		return err
+	if server {
+		logging.Info("Kubelet serving certificate has expired, waiting for automatic renewal... [will take up to 8 minutes]")
+		if err := waitForPendingCsrs(ocConfig, "kubernetes.io/kubelet-serving"); err != nil {
+			logging.Debugf("Error waiting for pending kubelet-serving CSR: %v", err)
+			return err
+		}
+		if err := ApproveNodeCSR(ocConfig, "kubernetes.io/kubelet-serving"); err != nil {
+			logging.Debugf("Error approving pending kubelet-serving CSR: %v", err)
+			return err
+		}
 	}
 	return nil
 }
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -32,9 +32,30 @@ func WaitForSSH(sshRunner *ssh.Runner) error {
 	return errors.RetryAfter(60*time.Second, checkSSHConnectivity, time.Second)
 }
 
-// CheckCertsValidity checks if the cluster certs have expired or going to expire in next 7 days
-func CheckCertsValidity(sshRunner *ssh.Runner) (bool, error) {
-	certExpiryDate, err := getcertExpiryDateFromVM(sshRunner)
+func CheckCertsValidity(sshRunner *ssh.Runner) (bool, bool, error) {
+	client, err := CheckKubeletClientCertValidity(sshRunner)
+	if err != nil {
+		return false, false, err
+	}
+	server, err := CheckKubeletServerCertValidity(sshRunner)
+	if err != nil {
+		return false, false, err
+	}
+	return client, server, nil
+}
+
+// CheckKubeletClientCertValidity checks if the kubelet server cert have expired
+func CheckKubeletServerCertValidity(sshRunner *ssh.Runner) (bool, error) {
+	return checkCertValidity(sshRunner, "/var/lib/kubelet/pki/kubelet-server-current.pem")
+}
+
+// CheckKubeletClientCertValidity checks if the kubelet client cert have expired
+func CheckKubeletClientCertValidity(sshRunner *ssh.Runner) (bool, error) {
+	return checkCertValidity(sshRunner, "/var/lib/kubelet/pki/kubelet-client-current.pem")
+}
+
+func checkCertValidity(sshRunner *ssh.Runner, cert string) (bool, error) {
+	certExpiryDate, err := getCertExpiryDateFromVM(sshRunner, cert)
 	if err != nil {
 		return false, err
 	}
@@ -45,9 +66,9 @@ func CheckCertsValidity(sshRunner *ssh.Runner) (bool, error) {
 	return false, nil
 }
 
-func getcertExpiryDateFromVM(sshRunner *ssh.Runner) (time.Time, error) {
+func getCertExpiryDateFromVM(sshRunner *ssh.Runner, cert string) (time.Time, error) {
 	certExpiryDate := time.Time{}
-	certExpiryDateCmd := `date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-client-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds`
+	certExpiryDateCmd := fmt.Sprintf(`date --date="$(sudo openssl x509 -in %s -noout -enddate | cut -d= -f 2)" --iso-8601=seconds`, cert)
 	output, err := sshRunner.Run(certExpiryDateCmd)
 	if err != nil {
 		return certExpiryDate, err
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -32,52 +32,39 @@ func WaitForSSH(sshRunner *ssh.Runner) error {
 	return errors.RetryAfter(60*time.Second, checkSSHConnectivity, time.Second)
 }
 
+const (
+	kubeletServerCert = "/var/lib/kubelet/pki/kubelet-server-current.pem"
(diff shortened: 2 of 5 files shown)


pkg/crc/machine/machine.go
```diff
@@ -399,11 +400,6 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 		log.Warnf("Cannot update kubeconfig: %v", err)
 	}
 
-	// Approve the node certificate.
-	if err := cluster.ApproveNodeCSR(ocConfig); err != nil {
-		return startError(startConfig.Name, "Error approving the node csr", err)
-	}
-
```
praveenkumar, line 398: This is now not required because we are using `192.168.126.11` as kubelet IP but before 4.4 timeframe when we had different IP for kubelet then it was required to approve pending node certificate.

Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION = 4.5.9
+BUNDLE_VERSION ?= 4.5.9
```
praveenkumar, line 3: @guillaumerose This change shouldn't be part of this commit message.

pkg/crc/cluster/cluster.go
```diff
@@ -73,24 +91,31 @@ func AddPullSecretInTheCluster(ocConfig oc.Config, pullSec *PullSecret) error {
...
-	}
 	_, stderr, err := ocConfig.RunOcCommandPrivate(cmdArgs...)
 	if err != nil {
 		return fmt.Errorf("Failed to add Pull secret %v: %s", err, stderr)
 	}
 	return nil
 }
 
-func UpdateClusterID(ocConfig oc.Config) error {
+func EnsureClusterIDIsNotEmpty(ocConfig oc.Config) error {
```
praveenkumar, line 101: Why we are adding thee check, this is not something user facing and we automatic add it as part of code, I am not sure what benefit it provide, if it is not set that mean our code which set this is buggy and need to be fixed.

pkg/crc/machine/machine.go
```diff
@@ -240,6 +236,10 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 		return startError(startConfig.Name, "Error getting the state", err)
 	}
 
+	privateKeyPath := constants.GetPrivateKeyPath()
+	if _, err := os.Stat(constants.GetPrivateKeyPath()); err != nil && os.IsNotExist(err) {
```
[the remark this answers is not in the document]
guillaumerose: done :+1:

pkg/crc/machine/machine.go
```diff
@@ -681,36 +681,47 @@ func (*client) GetConsoleURL(consoleConfig ConsoleConfig) (ConsoleResult, error)
...
 
 func updateSSHKeyAndCopyKubeconfig(sshRunner *crcssh.Runner, startConfig StartConfig, crcBundleMetadata *bundle.CrcBundleInfo) error {
-	logging.Info("Generating new SSH Key pair ...")
 	if err := updateSSHKeyPair(sshRunner); err != nil {
 		return fmt.Errorf("Error updating SSH Keys: %v", err)
 	}
 
+	kubeConfigFilePath := filepath.Join(constants.MachineInstanceDir, startConfig.Name, "kubeconfig")
+	if _, err := os.Stat(kubeConfigFilePath); err == nil {
+		return nil
```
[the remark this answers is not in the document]
guillaumerose: In case the file doesn't exist, an error will happen and the copy code just after will be triggered.

Makefile
guillaumerose: done :+1:

pkg/crc/cluster/cluster.go
guillaumerose: I would like to remove most of the branches in the `machine.Start` code. If the creation code could be the same as re-start code, it would be easier to follow. 

It also ensures after each start that the clusterID, the pull secret, etc. are correctly in place.

praveenkumar:
Except https://github.com/code-ready/crc/pull/1532#discussion_r491879049, lgtm

guillaumerose:
I refactored a bit the Start() function to reduce its size. Let me know what you think!

praveenkumar:
@guillaumerose conflict need to be resolved, I do like separating the `start` functionality in it's own file.

pkg/crc/cluster/cert-renewal.go
```diff
@@ -31,43 +29,26 @@ func waitForPendingCsrs(ocConfig oc.Config) error {
...
-
-	err = waitForPendingCsrs(ocConfig)
-	if err != nil {
+	if err := waitForPendingCsrs(ocConfig); err != nil {
 		logging.Debugf("Error waiting for second pending (system:node) CSR: %v", err)
 		return err
 	}
-	err = ApproveNodeCSR(ocConfig)
-	if err != nil {
+	if err := ApproveNodeCSR(ocConfig); err != nil {
```
cfergeau: Half of the changes in this commit replace `err = foo(); if err != nil {}` with `if err := foo(); err != nil {}` which I'm not necessarily opposed to, but would probably be better done in a separate commit rather than cluttering this one.

cfergeau (commented):
` Be selective when approving CSR. Either it's a client cert or a serve… `
Short log is too long, even github UI split it :) It's imo too terse to explain what this is about.

pkg/crc/cluster/cluster.go
cfergeau: this rationale belongs in the commit log

cfergeau (commented):
` Simplify CheckCertsValidity: use a boolean` please explain what your rationale is rather than leaving this as an exercise for the readers.
```cert-renewal: Simplify CheckCertsValidity

All that is needed from CheckCertsValidity() is to indicate whether certificates have expired and need to be renewed, or if no further action is needed. This can be done through a boolean rather than using the CertExpiryState enum as done currently.
```

pkg/crc/machine/start.go
```diff
@@ -341,6 +325,25 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 	}, err
 }
 
+func checkDNS(servicePostStartConfig services.ServicePostStartConfig) error {
+	// Check DNS lookup before starting the kubelet
+	if queryOutput, err := dns.CheckCRCLocalDNSReachable(servicePostStartConfig); err != nil {
+		return errors.Wrapf(err, "Failed internal DNS query: %s", queryOutput)
+	}
+	logging.Info("Check internal and public DNS query ...")
+
```
cfergeau, line 321: Good opportunity to fix the spacing?
```
<blank line>
logging.Info("Check internal and public DNS query ...")
if queryOutput, err := dns.CheckCRCPublicDNSReachable(...)
```

pkg/crc/machine/start.go
```diff
@@ -50,50 +51,45 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
...
-		// Retrieve metadata info
-		diskPath := crcBundleMetadata.GetDiskImagePath()
-		machineConfig.DiskPathURL = fmt.Sprintf("file://%s", filepath.ToSlash(diskPath))
-		machineConfig.SSHKeyPath = crcBundleMetadata.GetSSHKeyPath()
-		machineConfig.KernelCmdLine = crcBundleMetadata.Nodes[0].KernelCmdLine
-		machineConfig.Initramfs = crcBundleMetadata.GetInitramfsPath()
-		machineConfig.Kernel = crcBundleMetadata.GetKernelPath()
-
-		_, err := createHost(libMachineAPIClient, machineConfig)
+		_, err := createHost(libMachineClient, config.MachineConfig{
```
cfergeau: slight preference for this as it's easier to parse:
```
machineConfig := config.MachineConfig {
...
}
_, err := createHost(libMachineClient, machineConfig)
```
(nit: it's neither a renaming nor related to `IsRunning` -> this could actually be done in `Extract functions that return large structs`)

pkg/crc/machine/start.go
```diff
@@ -304,6 +287,31 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 	}, err
 }
 
+func machineConfig(startConfig StartConfig, bundle *bundle.CrcBundleInfo) config.MachineConfig {
```
cfergeau, line 290: `machineConfigNew()`?

pkg/crc/machine/start.go
```diff
@@ -304,6 +287,31 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
...
+		Memory:        startConfig.Memory,
+		DiskPathURL:   fmt.Sprintf("file://%s", filepath.ToSlash(bundle.GetDiskImagePath())),
+		SSHKeyPath:    bundle.GetSSHKeyPath(),
+		KernelCmdLine: bundle.Nodes[0].KernelCmdLine,
+		Initramfs:     bundle.GetInitramfsPath(),
+		Kernel:        bundle.GetKernelPath(),
+	}
+}
+
+func postStartConfig(startConfig StartConfig, sshRunner *crcssh.Runner, instanceIP string, bundle *bundle.CrcBundleInfo) services.ServicePostStartConfig {
```
cfergeau, line 304: `postStartConfigNew` (or `newPostStartConfig`)

cfergeau (commented):
> ` Always generate SSH keys if they don't exist`

Please explain in the commit log the rationale for doing this, why this is better than the current (working) code.


> `Use a struct to memoize the the pull secret` -> memorize (store ?) Commit log is missing some details regarding what your intent is.


> `cluster.RegenerateCertificates doesn't need to start kubelet anymore.`

Before this change, `cluster.RegenerateCertificates` was stopping the kubelet service, and it was restarted after it returned. Maybe a `systemctl restart kubelet` will still be needed, not sure how well it copes with cert renewals without restarts these days.

> `Wait for kubelet certs to be renewed`

This moves lots of code around, I'd expect the commit log to give a summary of what changed.

>` Ensure clusterID and pull secret are correctly set`

The commit log should explain why you want to do this. I see a (very good) reason for doing this, but that's just my guess, so better to spell it explicitly in the commit log so that everyone is on the same line.


> ` Move machine.Start in its own file `

As in the previous commits, stating explicitly why you do that (because it's huge I guess) never hurts :)


> ` Extract method that check DNS on the host and in the VM `

Why? (actually here I thought the commit was doing something entirely different from its log...) `start: Add helper for host/VM DNS checks`?


> ` Rename variables, remove machine.IsRunning `

Can you at least keep the "machine.IsRunning" removal separate? You could give the list of renames you are doing in the commit log so that they are easier to discuss if people disagree. As far as I can see, its `crcBundleMetadata` -> `bundle` and `libMachineAPIClient` -> `libMachineClient`, `vmState` -> `vmStatus`


> ` GetOpenshiftVersion() always return a value `

```
Remove handling of GetOpenshiftVersion() returning ""

Bundles shipped in current crc versions always have the OpenShift version set in their metadata, so we can remove the fallback code which was there in case it's not set.
```

> `Extract functions that return large structs`

Once again, not really clear what the commit is about :(

```
Add helpers to create MachineConfig and PostStartConfig

These struct have quite a few fields to initialise, this commit moves their creation to dedicated helpers.
```

pkg/crc/cluster/cluster.go
guillaumerose: added

pkg/crc/machine/machine.go
```diff
@@ -250,12 +250,12 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
...
-	if !exists {
-		if err := updateSSHKeyAndCopyKubeconfig(sshRunner, startConfig, crcBundleMetadata); err != nil {
-			return startError(startConfig.Name, "Error updating public key", err)
-		}
+	if err := updateSSHKeyAndCopyKubeconfig(sshRunner, startConfig, crcBundleMetadata); err != nil {
+		return startError(startConfig.Name, "Error updating public key", err)
 	}
 
+	sshRunner = crcssh.CreateRunnerWithPrivateKey(host.Driver, constants.GetPrivateKeyPath())
+
```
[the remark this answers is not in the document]
guillaumerose: My issue with this is that the private key of the runner is mutated in a sub-function of machine.Start. It's not obvious when reading start code that the ssh runner will mutate.

Now, starting here, we are sure the private key is `constants.GetPrivateKeyPath()`.

pkg/crc/machine/machine.go
```diff
@@ -250,12 +250,12 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 
 	// Post VM start immediately update SSH key and copy kubeconfig to instance
 	// dir and VM
-	if !exists {
-		if err := updateSSHKeyAndCopyKubeconfig(sshRunner, startConfig, crcBundleMetadata); err != nil {
-			return startError(startConfig.Name, "Error updating public key", err)
-		}
+	if err := updateSSHKeyAndCopyKubeconfig(sshRunner, startConfig, crcBundleMetadata); err != nil {
```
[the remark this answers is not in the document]
guillaumerose: The commit message is more precise now.

pkg/crc/cluster/cluster.go
```diff
@@ -76,11 +76,7 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 	return diskSize, diskUsage, nil
 }
 
-func AddPullSecret(sshRunner *ssh.Runner, ocConfig oc.Config, pullSec string) error {
-	if err := addPullSecretToInstanceDisk(sshRunner, pullSec); err != nil {
-		return err
-	}
-
+func AddPullSecretInTheCluster(ocConfig oc.Config, pullSec string) error {
```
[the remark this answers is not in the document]
guillaumerose: I ended up the serie with `EnsurePullSecretPresentInTheCluster` and `EnsurePullSecretPresentOnInstanceDisk`. Is it ok?

pkg/crc/cluster/cluster.go
```diff
@@ -223,13 +227,28 @@ func addProxyCACertToInstance(sshRunner *ssh.Runner, proxy *network.ProxyConfig)
...
+	}
+	val, err := p.Getter()
+	if err == nil {
+		p.value = val
+	}
+	return val, err
+}
+
+func AddPullSecretToInstanceDisk(sshRunner *ssh.Runner, pullSec *PullSecret) error {
+	content, err := pullSec.Value()
```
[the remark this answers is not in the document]
guillaumerose: I renamed `pullSec` to `pullSecret`, but kept content since it used 2 lines after.

pkg/crc/cluster/cluster.go
```diff
@@ -243,12 +243,17 @@ func (p *PullSecret) Value() (string, error) {
 	return val, err
 }
 
-func AddPullSecretToInstanceDisk(sshRunner *ssh.Runner, pullSec *PullSecret) error {
+func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSec *PullSecret) error {
+	filename := "/var/lib/kubelet/config.json"
```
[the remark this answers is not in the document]
guillaumerose: Constant extracted!

pkg/crc/machine/start.go
guillaumerose: done

pkg/crc/machine/start.go
guillaumerose: IsRunning is gone, removed by an other commit.

pkg/crc/machine/start.go
guillaumerose: renamed !

pkg/crc/machine/start.go
guillaumerose: renamed !

guillaumerose:
I think I went on all messages. It should be good for an other round!

cfergeau (commented):
The `SetPrivateKey() is also removed so that sshRunner is not mutated in a sub function of machine.Start().` log is obsolete since you addressed this differently in master

pkg/crc/machine/machine.go
```diff
@@ -676,16 +675,26 @@ func (*client) GetConsoleURL(consoleConfig ConsoleConfig) (ConsoleResult, error)
...
+		}
 	}
 
 	// Read generated public key
 	publicKey, err := ioutil.ReadFile(constants.GetPublicKeyPath())
 	if err != nil {
 		return err
 	}
+
+	authorizedKeys, err := sshRunner.Run("cat /home/core/.ssh/authorized_keys")
```
cfergeau: `test -e /home/core/.ssh/authorized_keys` seems to be doing the job

cfergeau:
There is a `memoize` typo in one of the commit logs

guillaumerose:
All done! For memoize/memorize, I was referring https://en.wikipedia.org/wiki/Memoization

pkg/crc/cluster/cluster.go
cfergeau: >  I am not sure what benefit it provide, if it is not set that mean our code which set this is buggy and need to be fixed.

It would help to recover in case there was a transient cluster failure when the cluster was created, or if the user pressed ctrl+c during the initial cluster creation, ...

pkg/crc/machine/start.go
```diff
@@ -108,30 +103,30 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 					filepath.Base(startConfig.BundlePath),
 					bundleName))
 		}
-		vmState, err := host.Driver.GetState()
+		vmStatus, err := host.Driver.GetState()
```
cfergeau, line 90: fwiw, after applying this PR, I still get
```
$ git grep vmState
pkg/crc/machine/machine.go:     vmState, err := host.Driver.GetState()
pkg/crc/machine/machine.go:             State:         vmState,
pkg/crc/machine/start.go:       vmState, err := host.Driver.GetState()
pkg/crc/machine/start.go:               Status:         vmState.String(),
```

pkg/crc/machine/start.go
```diff
@@ -297,13 +280,38 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
...
+}
+
 func checkDNS(servicePostStartConfig services.ServicePostStartConfig) error {
 	// Check DNS lookup before starting the kubelet
 	if queryOutput, err := dns.CheckCRCLocalDNSReachable(servicePostStartConfig); err != nil {
 		return errors.Wrapf(err, "Failed internal DNS query: %s", queryOutput)
 	}
-	logging.Info("Check internal and public DNS query ...")
 
+	logging.Info("Check internal and public DNS query ...")
```
cfergeau, line 314: Not the right commit, this belongs to `Add helper for host/VM DNS checks`

cfergeau (commented):
This PR seems to be a bit of multiple independant PRs, I'd expect `Always make sure the pull secret is in the cluster` and `Always make sure the cluster ID is not empty` to come before the commits related to certificate regeneration, so that we have first the commits making VM creation/VM start more similar, and then the certificate regeneration commits

cfergeau (commented):
One note regarding `Move machine.Start in its own file`, this impacts `tig  blame`, a bit `git log`, ...

`Verify both client and server kubelet certificates` introduces a significant behaviour change in my tests, this seems to be unexpected from my reading of the commit log?

cfergeau:
At this point, I'd be inclined to suggest switching to 2 PRs, one with the cert renewal commits:
```
215cd38a Wait for kubelet certs to be renewed
af2ed7eb Move code related to certs in the same go file
90ea3320 Make K8sResource a private struct
3cf6c3d9 Verify both client and server kubelet certificates
25ca1e47 Move cluster.RegenerateCertificates right after kubelet startup in machine.Start()
```
and the other with the rest of the commits (but the introduction of `start.go` makes this slightly painful :( The other commits look good to me. Since they are not really related to cert renewal, it would be good to land them while taking a closer look at the cert renewal  changes.

pkg/crc/cluster/cluster.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/pborman/uuid"
 )
 
+// #nosec G101
+const pullSecretInTheVM = "/var/lib/kubelet/config.json"
```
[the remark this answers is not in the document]
guillaumerose: done in https://github.com/code-ready/crc/pull/1548

pkg/crc/cluster/cluster.go
```diff
@@ -243,12 +246,16 @@ func (p *PullSecret) Value() (string, error) {
 	return val, err
 }
 
-func AddPullSecretToInstanceDisk(sshRunner *ssh.Runner, pullSecret *PullSecret) error {
+func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret *PullSecret) error {
+	if _, err := sshRunner.Run(fmt.Sprintf("ls %s", pullSecretInTheVM)); err == nil {
```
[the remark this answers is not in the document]
guillaumerose: done in https://github.com/code-ready/crc/pull/1548

pkg/crc/machine/machine.go
```diff
@@ -337,11 +341,7 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 			return startError(startConfig.Name, "Error Setting cluster config", err)
 		}
 
-		logging.Info("Adding user's pull secret ...")
-		if err := cluster.AddPullSecretToInstanceDisk(sshRunner, startConfig.PullSecret); err != nil {
-			return startError(startConfig.Name, "Failed to update user pull secret", err)
-		}
-
```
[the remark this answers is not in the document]
guillaumerose: done in https://github.com/code-ready/crc/pull/1548

pkg/crc/cluster/cert-renewal.go
```diff
@@ -31,43 +29,26 @@ func waitForPendingCsrs(ocConfig oc.Config) error {
...
-	err := sd.Start("kubelet")
-	if err != nil {
-		logging.Debugf("Error starting kubelet service: %v", err)
-		return err
-	}
-	// FIXME: regression compared to previous API
-	startedKubelet := false
-	if startedKubelet {
-		defer sd.Stop("kubelet") //nolint:errcheck
-	}
```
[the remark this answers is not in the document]
guillaumerose: Apparently yes, kubelet uses the new certificate without restarting.

pkg/crc/machine/start.go
guillaumerose: Dropped this commit

pkg/crc/machine/start.go
guillaumerose: I dropped this commit, it should be ok for this one.

guillaumerose:
The log is far more lighter now with the PR split.

guillaumerose:
About the second cert approval, I triggered the race condition yesterday.

Both crc and machine approver want to approve the kubelet serving certificate. If crc wins, it works. If machine approver wins, crc command hangs.

guillaumerose:
PR rebased. PTAL.

cfergeau:
Still unclear whether this PR is expected to trigger significant behaviour change on `crc start`, after re-reading the commit logs, my feeling is that this is unexpected?

guillaumerose:
There is no significant changes, just bug fixes. It changes:
* Previously, we checked client cert expiry date and waited both the client and serving cert to be rotated. The serving cert might not need to be rotated, but crc still wait for a CSR to approve and hangs. This PR removes the dead lock.
* If the serving cert is expired, crc try to approve it. But, an other process, the cluster machine approver, also tries to do it. If crc wins, no locks. If the approver wins, crc is in a dead lock. This PR removes the approval of the serving cert.

cfergeau:
> There is no significant changes, just bug fixes. It changes:
> 
>     * Previously, we checked client cert expiry date and waited both the client and serving cert to be rotated. The serving cert might not need to be rotated, but crc still wait for a CSR to approve and hangs. This PR removes the dead lock.
> 
>     * If the serving cert is expired, crc try to approve it. But, an other process, the cluster machine approver, also tries to do it. If crc wins, no locks. If the approver wins, crc is in a dead lock. This PR removes the approval of the serving cert.

Could we be now trying too hard/doing too much? With the 4.5.9 bundle from 1.16 and git master:
```
$ crc delete
$ crc start
[...]
INFO Verifying validity of the cluster certificates ...
date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-client-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds
DEBU SSH command results: err: <nil>, output: 2020-10-13T10:06:06+00:00
DEBU Creating /var/srv/dnsmasq.conf with permissions 0644 in the CRC VM
DEBU About to run SSH command with hidden output                              
DEBU SSH command succeeded                      
sudo podman rm -f dnsmasq              
[...]
```
With this PR and the same bundle:
```
$ crc delete
$ crc start
[...]
INFO Verifying validity of the kubelet certificates ...                                                                                                                                               
date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-client-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds
DEBU SSH command results: err: <nil>, output: 2020-10-13T10:06:06+00:00
date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-server-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds
DEBU SSH command results: err: <nil>, output: 2020-09-14T09:06:42+00:00
DEBU Certs have expired, they were valid till: 14 Sep 20 09:06 +0000
[...]
INFO Cluster TLS certificates have expired, renewing them... [will take up to 5 minutes]
```
With git master, if I run `date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-server-current.pem -noout -enddate | cut -d= -f 2)` in the VM right after crc start, the server cert is valid for 1 more week without any explicit action from crc.

guillaumerose:
Yes the serving cert is an extra check. With current master, it is silently updated (during the 3min sleep time maybe). Until the cert is updated, user can't exec in a container nor fetching container logs. 
I hope to remove the final sleep(3 min) with this kind of checks.

pkg/crc/machine/machine.go
cfergeau: I'd add an explanation about this in the commit log. Especially as this seems to be silently removed in ` Wait for kubelet certs to be renewed`

guillaumerose: Added an extra commit for this!

pkg/crc/cluster/cert-renewal.go
```diff
@@ -2,53 +2,53 @@ package cluster
...
-	if err := waitForPendingCsrs(ocConfig); err != nil {
-		logging.Debugf("Error waiting for second pending (system:node) CSR: %v", err)
-		return err
+	if client {
+		if err := waitForPendingCsrs(ocConfig, "kubernetes.io/kube-apiserver-client-kubelet"); err != nil {
+			logging.Debugf("Error waiting for first pending (node-bootstrapper) CSR: %v", err)
+			return err
+		}
+		if err := ApproveNodeCSR(ocConfig, "kubernetes.io/kube-apiserver-client-kubelet"); err != nil {
+			logging.Debugf("Error approving first pending (node-bootstrapper) CSR: %v", err)
```
cfergeau, line 38: It's no longer first/second, but client/server CSRs. I guess it would be possible to only get the second CSR for example in some scenarios.

pkg/crc/cluster/cert-renewal.go
```diff
@@ -1,54 +1,72 @@
...
+			if err != nil {
+				return err
+			}
+			if !expired {
+				return nil
+			}
+			return &crcerrors.RetriableError{Err: errors.New("kubelet client certificate still expired")}
+		}, time.Second*5); err != nil {
+			logging.Debugf("Error approving first pending (node-bootstrapper) CSR: %v", err)
 			return err
```
cfergeau, line 52: Have you considered adding a `waitForCertRenewal()` helper?

pkg/crc/cluster/cert-renewal.go
```diff
@@ -1,54 +1,72 @@
...
 		}
 	}
+	// API server needs to connect to kubelet for some features like logs, port forwards. This communication is backed by a cert
+	// store in /var/lib/kubelet/pki/kubelet-server-current.pem
+	// After kubelet connected to the API server, if the serving cert is expireed, kubelet asks for a new CSR.
+	// This CSR is automatically approved by the cluster-machine-approver. The k8s controller manager issues the cert and kubelet fetches it.
```
cfergeau, line 58: This seems to be explaining why waitForPendingCSRs/ApproveNodeCSR was removed in the `if server` case. This should be mentioned/explained in the commit log as well.

cfergeau (commented):
>  Move RegenerateCertificates right after kubelet startup in machine.Start 

> cluster.RegenerateCertificates doesn't need to stop/start kubelet anymore.

Maybe add explicitly that it's not needed anymore with newer openshift releases, since kubelet uses the new certificate without restarting?


> Verify both client and server kubelet certificates

> When approving CSR, now we select exactly the one we expect by specifying the signerName.

It would be good for future reference to expand this commit log to describe more clearly the behaviour before, the new behaviour, ..
Before we were only checking the client certificate `/var/lib/kubelet/pki/kubelet-client-current.pem`, but expecting 2 CSRs, one which is generated when the client certificate is expired, and another one which is generated when the server certificate is expired. After this PR, crc also checks for expiration of the server certificate ``/var/lib/kubelet/pki/kubelet-server-current.pem`, and `WaitAndApprovePendingCSRs` has been improved so that depending on which certificate(s) has expired, it waits/approves the correct CSR.


It might also be worth adding 
> Yes the serving cert is an extra check. With current master, it is silently updated (during the 3min sleep time maybe). Until the cert is updated, user can't exec in a container nor fetching container logs.
I hope to remove the final sleep(3 min) with this kind of checks.

in one of the commits logs. The more we record regarding why a change is being made, our intent, ..., the easier it will be if we look at this code again a few months/years from now after having forgotten most of these discussions.

pkg/crc/cluster/cert-renewal.go
```diff
@@ -1,73 +1,72 @@
...
+	crcerrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/oc"
 	"github.com/code-ready/crc/pkg/crc/ssh"
-	"github.com/code-ready/crc/pkg/crc/systemd"
 )
 
-func waitForPendingCsrs(ocConfig oc.Config) error {
-	waitForPendingCsr := func() error {
+func waitForPendingCsrs(ocConfig oc.Config, signerName string) error {
```
cfergeau, line 15: Since you are changing the method, you can fix the spelling `waitForPendingCSRs` to be consistent with the other method with `CSRs` in its name.

pkg/crc/machine/machine.go
```diff
@@ -334,9 +335,9 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 
 	ocConfig := oc.UseOCWithSSH(sshRunner)
 
-	if certsExpired {
+	if clientExpired || serverExpired {
 		logging.Info("Cluster TLS certificates have expired, renewing them... [will take up to 5 minutes]")
```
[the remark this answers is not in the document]
guillaumerose: yes done!

pkg/crc/cluster/cert-renewal.go
guillaumerose: yes changed!

pkg/crc/cluster/cert-renewal.go
guillaumerose: yes changelog is better now

pkg/crc/cluster/cert-renewal.go
guillaumerose: yes done!

pkg/crc/cluster/cert-renewal.go
guillaumerose: done

cfergeau (approved):

PR #1533, opened 2020-09-21
Only send logs to stderr

guillaumerose:
`crc start -ojson | jq .` is broken because INFO logs are sent to stdout. This PR hopefully restores the default behaviour of logs.


cfergeau:
Dunno, to me it makes sense that informative logs go to stdout, and warning/error logs would go to stderr?

cfergeau:
Wondering if the real problem is that we are mixing 'regular command output' with 'info logs'? regular command output must be either 'human readable' or 'json' depending on `-o` value. 'info logs' should probably not be shown at all unless requested, or at least it should be easy to enable/disable them?
I would say whether they go to stdout or stderr is orthogonal to that, even if it seems that in oc, they either go to stderr or to a file.

praveenkumar:
The default output file descriptor for logrus is stderr https://github.com/sirupsen/logrus/blob/42baed85eb233bfc674562af0ef6e99c45f5f384/logger.go#L84-L93 , I do agree on https://github.com/code-ready/crc/pull/1533#issuecomment-695989251 one either we disable the logger stdout hook when user use `-o json` I don't think we are using any `fmt.Println`, everything is coming from logger side.

guillaumerose:
logging package overrides default behaviour of logrus in order to send logs both to a file and to the console. 
It can't use `logrus.SetOut(io.MultiWriter(file, os.Stderr))` because it would remove colors in the shell. It needs to use a hook to keep the colors.

In machine.Start, we should perhaps don't use the logger if we want to print valuable information for the user. I don't think the output format should change the logger behaviour.

anjannath:
> Wondering if the real problem is that we are mixing 'regular command output' with 'info logs'? regular command output must be either 'human readable' or 'json' depending on `-o` value. 'info logs' should probably not be shown at all unless requested, or at least it should be easy to enable/disable them?
> I would say whether they go to stdout or stderr is orthogonal to that, even if it seems that in oc, they either go to stderr or to a file.

This can be done if we set the default log level to 0, which is not there but we can set it to error or panic, we are mostly using debug and info logs, and debuf/info logs can be enabled by `--log-level`

cfergeau:
> In machine.Start, we should perhaps don't use the logger if we want to print valuable information for the user. I don't think the output format should change the logger behaviour.

Yup, I would say this is the root issue, which clashes with "I don't think the output format should change the logger behaviour.". Logging to stderr just provides an easy workaround for this issue (and it seems logging to stderr is standard behaviour for at least golang applications if logrus does this by default)

test/integration/features/basic.feature
```diff
@@ -37,54 +37,54 @@ Feature: Basic test
...
-        And stdout should contain "Installing crc-machine-hyperkit"
-        And stdout should contain "Will use root access: change ownership"
-        And stdout should contain "Will use root access: set suid"
-        And stdout should contain "Checking file permissions"
+        Then stderr should contain "Caching oc binary"
+        And stderr should contain "Checking if running as non-root"
+        And stderr should contain "Checking if HyperKit is installed"
+        And stderr should contain "Checking if crc-driver-hyperkit is installed"
+        And stderr should contain "Installing crc-machine-hyperkit"
+        And stderr should contain "Will use root access: change ownership"
```
praveenkumar, line 75: :+1:

praveenkumar (approved):

PR #1535, opened 2020-09-21
Only check bundle version if the bundle is embedded

guillaumerose:
This check is useful for released version of crc. For development, we need to be able to launch any bundle.

### DIFF
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -56,6 +56,10 @@ func ValidateBundle(bundle string) error {
 	releaseBundleVersion := version.GetBundleVersion()
 	userProvidedBundleVersion := filepath.Base(bundle)
 	if !strings.Contains(userProvidedBundleVersion, fmt.Sprintf("%s.crcbundle", releaseBundleVersion)) {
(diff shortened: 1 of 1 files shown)


cfergeau (approved):
Yep, sounds like a reasonable compromise between forcing a given bundle version and making testing easier.
@praveenkumar @gbraad would be good to have your ack as well

gbraad:
Wish we could rename `constants`

gbraad (approved):

pkg/crc/validation/validation.go
```diff
@@ -56,6 +56,10 @@ func ValidateBundle(bundle string) error {
 	releaseBundleVersion := version.GetBundleVersion()
 	userProvidedBundleVersion := filepath.Base(bundle)
 	if !strings.Contains(userProvidedBundleVersion, fmt.Sprintf("%s.crcbundle", releaseBundleVersion)) {
+		if !constants.BundleEmbedded() {
+			logging.Warnf("Using unsupported bundle %s", userProvidedBundleVersion)
```
praveenkumar, line 60: :+1:

praveenkumar (approved):

PR #1537, opened 2020-09-21
Hack: replace the DNS on the host by a loop running in the daemon that keep in sync /etc/hosts and openshift routes

guillaumerose:
/wip

I dont know if it's a good solution or not, but starting a discussion.

Pros:
* Remove the dependency on NetworkManager and dnsmasq
* It could work on Windows if we have a solution for goodhosts (a privileged daemon with the simplest/smaller API possible?).

Cons:
* Requires a running daemon (not privileged on Linux/macOS)


cmd/crc/cmd/daemon.go
```diff
@@ -20,6 +32,78 @@ var daemonCmd = &cobra.Command{
...
+	factory := informers.NewSharedInformerFactory(clientset, 0)
+	informer := factory.Route().V1().Routes().Informer()
+	defer close(stopCh)
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			route := obj.(*v1.Route)
+			fmt.Printf("added: %s %s\n", route.GetName(), route.Spec.Host)
+			if err := goodhosts.AddToHostsFile(ip, route.Spec.Host); err != nil {
+				logging.Error(err)
+			}
```
gbraad, line 86: can be extracted

cmd/crc/cmd/daemon.go
```diff
@@ -20,6 +32,78 @@ var daemonCmd = &cobra.Command{
...
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			old := oldObj.(*v1.Route)
+			route := newObj.(*v1.Route)
+			fmt.Printf("updated: %s (%s -> %s)\n", route.GetName(), old.Spec.Host, route.Spec.Host)
+			if err := goodhosts.RemoveFromHostsFile(ip, old.Spec.Host); err != nil {
+				logging.Error(err)
+			}
+			if err := goodhosts.AddToHostsFile(ip, route.Spec.Host); err != nil {
+				logging.Error(err)
+			}
```
gbraad, line 97: can be re-used

cmd/crc/cmd/daemon.go
```diff
@@ -20,6 +32,78 @@ var daemonCmd = &cobra.Command{
...
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			old := oldObj.(*v1.Route)
+			route := newObj.(*v1.Route)
+			fmt.Printf("updated: %s (%s -> %s)\n", route.GetName(), old.Spec.Host, route.Spec.Host)
+			if err := goodhosts.RemoveFromHostsFile(ip, old.Spec.Host); err != nil {
+				logging.Error(err)
+			}
```
gbraad, line 94: can be extracted

cmd/crc/cmd/daemon.go
```diff
@@ -20,6 +32,78 @@ var daemonCmd = &cobra.Command{
...
+			if err := goodhosts.AddToHostsFile(ip, route.Spec.Host); err != nil {
+				logging.Error(err)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			route := obj.(*v1.Route)
+			fmt.Printf("deleted: %s %s\n", route.GetName(), route.Spec.Host)
+			if err := goodhosts.RemoveFromHostsFile(ip, route.Spec.Host); err != nil {
+				logging.Error(err)
+			}
```
gbraad, line 104: can be re-used

cmd/crc/cmd/daemon.go
```diff
@@ -20,6 +32,78 @@ var daemonCmd = &cobra.Command{
...
+			fmt.Printf("added: %s %s\n", route.GetName(), route.Spec.Host)
+			if err := goodhosts.AddToHostsFile(ip, route.Spec.Host); err != nil {
+				logging.Error(err)
+			}
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			old := oldObj.(*v1.Route)
+			route := newObj.(*v1.Route)
+			fmt.Printf("updated: %s (%s -> %s)\n", route.GetName(), old.Spec.Host, route.Spec.Host)
+			if err := goodhosts.RemoveFromHostsFile(ip, old.Spec.Host); err != nil {
```
gbraad, line 92: refactor => updateRoute() -> ...

gbraad:
this is mostly a large PR due to the dependencies. perhaps better to break this up in an additional commit for the go.mod stuff?

praveenkumar:
>  better to break this up in an additional commit for the go.mod stuff?

go mod stuff should be part of the feature itself, even it that increase the PR because we are going to use the imported module for the feature and getting it in a separate commit doesn't make sense.

gbraad:
> the feature and getting it in a separate commit doesn't make sense.

for review purposes... plus, it can be the first commit to prepare the environment.
the compiler should not have to complain about this.

cfergeau:
I'm still in favour of updating `vendor/` in a separate commit :) I'm not even sure how well `vendor/` works seeing the recent breakage https://github.com/code-ready/crc/commit/7240de6554e6b8fe0d372ce190b94b26d9de1724
The `go.mod`/`go.sub` changes are better done in the main commit as it needs them.

pkg/crc/services/dns/dns_linux.go
```diff
@@ -8,10 +8,7 @@ import (
 func runPostStartForOS(serviceConfig services.ServicePostStartConfig, result *services.ServicePostStartResult) (services.ServicePostStartResult, error) {
 	// We might need to set the firewall here to forward
 	// Update /etc/hosts file for host
-	if err := goodhosts.UpdateHostsFile(serviceConfig.IP, serviceConfig.BundleMetadata.GetAPIHostname(),
-		serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift"),
-		serviceConfig.BundleMetadata.GetAppHostname("console-openshift-console"),
-		serviceConfig.BundleMetadata.GetAppHostname("default-route-openshift-image-registry")); err != nil {
+	if err := goodhosts.UpdateHostsFile(serviceConfig.IP, serviceConfig.BundleMetadata.GetAPIHostname()); err != nil {
```
cfergeau, line 11: how does this work on cluster startup? do we get a sync for all the routes the cluster knows? or would we need to add it ourself? I guess on cluster stop it would also be nice to remove all the known routes from /etc/hosts?

cfergeau (commented):
Since this removes the need for all the NetworkManager configuration that we need to do on the host (and to a lesser extent, the /etc/resolver change on the mac), this is imo a very interesting experiment, as I've always found these configuration changes very invasive.

guillaumerose:
related to https://github.com/code-ready/crc/issues/1132

guillaumerose:
Keeping this PR as an example but it would be not this code for sure. It would require a UAC prompt for Windows for each change.

cfergeau:
>  It would require a UAC prompt for Windows for each change.

I don't really see this as a blocker, the route monitoring code would still be the same, only `routesController` would need to stop using `goodhosts` directly, and I expect the replacement would have a similar interface?

guillaumerose:
The discovery of the kubeconfig is also a problem. It doesn't exist after setup, it can change if delete/start occurs. It should maybe land in a container in the VM with a proper service account.

PR #1538, opened 2020-09-22
This PR updates golangci-lint to its latest release, and fixes the few additional issues the newer version is reporting.

cfergeau:
There should be no user-visible changes.


guillaumerose:
I think there is an issue with the Makefile. golangci-lint (+ deps) should not be in our vendor, we should just use the binary. LGTM otherwise, thanks!

praveenkumar:
> I think there is an issue with the Makefile. golangci-lint (+ deps) should not be in our vendor, we should just use the binary. LGTM otherwise, thanks!

I agree I think it is happening because by default `go get` will update the go.mod file also so we need something like https://stackoverflow.com/questions/56842385/using-go-get-to-download-binaries-without-adding-them-to-go-mod in our makefile .

guillaumerose (approved):

PR #1539, opened 2020-09-22
WIP: Add command to gather diagnostics from the user

guillaumerose:
This PR adds `crc diagnose` command. It collects information both from the host and the VM and put everything in a zip file. It would allow us to debug user/customer issue but also CI issue (I added `crc diagnose` in the AfterSuite but I guess it's not enough).

Zip example: [diagnose.zip](https://github.com/code-ready/crc/files/5260341/diagnose.zip)

Let me know what you think!

### DIFF
--- a/cmd/crc/cmd/diagnose.go
+++ b/cmd/crc/cmd/diagnose.go
@@ -1,0 +1,208 @@
+package cmd
+
+import (
+	"archive/zip"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/exit"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/machine"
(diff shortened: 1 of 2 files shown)


cmd/crc/cmd/diagnose.go
```diff
@@ -0,0 +1,208 @@
...
+			Dir:    constants.CrcBaseDir,
+			Target: "tree.txt",
+		},
+		file(constants.ConfigPath),
+		file(constants.LogFilePath),
+		file(constants.DaemonLogFilePath),
+		file(filepath.Join(constants.MachineBaseDir, "machines", constants.DefaultName, "config.json")),
+		file("/etc/hosts"),
+		vmFile("/etc/resolv.conf"),
+		vmFile("/etc/hosts"),
```
cfergeau, line 60: `file("/etc/os-release")`, `file("/proc/cpuinfo")`, `journalctl -u NetworkManager`, NetworkManager configuration files, ... ?

cfergeau:
Seems definitely useful as an upstream debugging tool, though for downstearm I'm not sure how (or if it should?) this would integrate with eg sosreport on RHEL? But we could definitely try this and see how useful this is with debugging crc issues.

guillaumerose:
Will reopen this PR if this becomes a priority.

PR #1540, opened 2020-09-22
Get golangci-lint binary without updating go mod

praveenkumar:
Currently when we update the golangci-lint version it also
update the go.mod file which is unwanted.

- https://stackoverflow.com/questions/56842385/using-go-get-to-download-binaries-without-adding-them-to-go-mod

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ fmt:
 	@gofmt -l -w $(SOURCE_DIRS)
 
 $(GOPATH)/bin/golangci-lint:
-	go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.27.0
+	pushd /tmp && GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.27.0 && popd
 
 # Run golangci-lint against code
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -145,7 +145,7 @@ fmt:
 	@gofmt -l -w $(SOURCE_DIRS)
 
 $(GOPATH)/bin/golangci-lint:
-	go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.27.0
+	pushd /tmp && GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.27.0 && popd
```
cfergeau, line 148: quite ugly :(( Is GO111MODULE required?

cfergeau (approved):

Makefile
praveenkumar: yes without it, if you have $GOPATH set as default, it will complain about it and exit

```
go: cannot use path@version syntax in GOPATH mode
make: *** [Makefile:148: /home/prkumar/work/go/bin/golangci-lint] Error 1
```

PR #1542, opened 2020-09-24
Create only one immutable sshRunner in machine.Start

guillaumerose:
This sshRunner now uses 2 private keys: one from the bundle, one that we created for the rotation.
If the ssh key rotation was aborted by the user (ctrl+c at the wrong moment, or so), machine.Start will be able to recover the next time.

It also removes the dependency on drivers.Driver in ssh package.

### DIFF
--- a/pkg/crc/ssh/ssh.go
+++ b/pkg/crc/ssh/ssh.go
@@ -14,16 +14,16 @@ import (
 )
 
 type Runner struct {
-	ip            string
-	port          int
-	privateSSHKey string
+	ip             string
+	port           int
+	privateSSHKeys []string
 }
 
-func CreateRunnerWithPrivateKey(ip string, port int, privateKey string) *Runner {
+func CreateRunner(ip string, port int, privateKeys ...string) *Runner {
 	return &Runner{
-		ip:            ip,
-		port:          port,
-		privateSSHKey: privateKey,
+		ip:             ip,
+		port:           port,
+		privateSSHKeys: privateKeys,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/crc/ssh/ssh.go
```diff
@@ -14,16 +14,16 @@ import (
...
-	ip            string
-	port          int
-	privateSSHKey string
+	ip             string
+	port           int
+	privateSSHKeys []string
 }
 
-func CreateRunnerWithPrivateKey(ip string, port int, privateKey string) *Runner {
+func CreateRunnerWithPrivateKey(ip string, port int, privateKeys ...string) *Runner {
```
cfergeau, line 22: The method name should also be `CreateRunnerWithPrivateKeys`, or just `CreateRunner` since there is no longer a method with that name.

cfergeau (approved):

pkg/crc/ssh/ssh.go
guillaumerose: done 👍

pkg/crc/ssh/ssh.go
```diff
@@ -58,8 +54,15 @@ func (runner *Runner) CopyFile(srcFilename string, destFilename string, mode os.
 }
 
 func (runner *Runner) runSSHCommandFromDriver(command string, runPrivate bool) (string, error) {
+	var availableKeys []string
+	for _, privateKey := range runner.privateSSHKeys {
+		if _, err := os.Stat(privateKey); err == nil {
+			availableKeys = append(availableKeys, privateKey)
```
praveenkumar, line 60: :+1:

praveenkumar (approved):

PR #1543, opened 2020-09-25
Add unit tests for preflight checks

guillaumerose:
Count test will save us from mistake when doing some refactorings.

### DIFF
--- a/pkg/crc/preflight/preflight_test.go
+++ b/pkg/crc/preflight/preflight_test.go
@@ -1,0 +1,73 @@
+package preflight
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/code-ready/crc/pkg/crc/config"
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_test.go
```diff
@@ -2,12 +2,23 @@ package preflight
...
 )
 
+func TestCount(t *testing.T) {
+	cfg := config.New(config.NewEmptyInMemoryStorage())
+	RegisterSettings(cfg)
+	assert.Len(t, cfg.AllConfigs(), map[string]int{
+		"darwin":  16,
+		"linux":   34,
+		"windows": 20,
+	}[runtime.GOOS])
```
praveenkumar, line 19: :+1:

praveenkumar (approved):

PR #1545, opened 2020-09-28
Proxy refactorings: logs and instanceIP in NO_PROXY

guillaumerose:
* Add in proxy struct where the proxy CA is coming from
* Log when overriding proxy env variables.
* Add instanceIP in NO_PROXY at the top and not only in a small part of `machine.Start()`

After this one, we can open a follow-up that removes calls to `ApplyToEnvironment` in machine pkg. Since `oc` is called with ssh, we don't need it (apart from crc version check but it's done in cmd pkg).

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -310,7 +310,7 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 
 	// Check DNS lookup from host to VM
 	logging.Info("Check DNS query from host ...")
-	if err := network.CheckCRCLocalDNSReachableFromHost(crcBundleMetadata, instanceIP); err != nil {
+	if err := dns.CheckCRCLocalDNSReachableFromHost(crcBundleMetadata, instanceIP); err != nil {
 		return nil, errors.Wrap(err, "Failed to query DNS from host")
 	}
 
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -715,14 +715,10 @@ func updateSSHKeyAndCopyKubeconfig(sshRunner *crcssh.Runner, name string, crcBun
 }
 
 func getProxyConfig(baseDomainName string) (*network.ProxyConfig, error) {
-	proxy, err := network.NewProxyConfig()
-	if err != nil {
-		return nil, err
-	}
+	proxy := network.GetProxyConfig()
 	if proxy.IsEnabled() {
 		proxy.AddNoProxy(fmt.Sprintf(".%s", baseDomainName))
 	}
-
 	return proxy, nil
 }
 
--- a/pkg/crc/network/proxy.go
+++ b/pkg/crc/network/proxy.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
@@ -18,13 +19,20 @@ type ProxyConfig struct {
 	HTTPProxy   string
 	HTTPSProxy  string
 	noProxy     []string
+	ProxyCAFile string
 	ProxyCACert string
 }
 
-func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCACert string) (*ProxyConfig, error) {
+func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAFile string) (*ProxyConfig, error) {
+	proxyCACert, err := getProxyCAData(proxyCAFile)
+	if err != nil {
+		return nil, err
+	}
+
 	DefaultProxy = ProxyConfig{
 		HTTPProxy:   httpProxy,
 		HTTPSProxy:  httpsProxy,
+		ProxyCAFile: proxyCAFile,
 		ProxyCACert: proxyCACert,
 	}
 
@@ -48,6 +56,7 @@ func NewProxyConfig() (*ProxyConfig, error) {
 	config := ProxyConfig{
 		HTTPProxy:   DefaultProxy.HTTPProxy,
 		HTTPSProxy:  DefaultProxy.HTTPSProxy,
+		ProxyCAFile: DefaultProxy.ProxyCAFile,
 		ProxyCACert: DefaultProxy.ProxyCACert,
 	}
 
@@ -67,6 +76,22 @@ func NewProxyConfig() (*ProxyConfig, error) {
 	}
 
 	return &config, nil
+}
+
+func getProxyCAData(proxyCAFile string) (string, error) {
+	if proxyCAFile == "" {
+		return "", nil
+	}
+	proxyCACert, err := ioutil.ReadFile(proxyCAFile)
+	if err != nil {
+		return "", err
+	}
+	// Before passing string back to caller function, remove the empty lines in the end
+	return trimTrailingEOL(string(proxyCACert)), nil
+}
+
+func trimTrailingEOL(s string) string {
+	return strings.TrimRight(s, "\n")
 }
 
 func getProxyFromEnv(proxyScheme string) string {
--- a/pkg/crc/network/proxy.go
+++ b/pkg/crc/network/proxy.go
@@ -97,6 +97,11 @@ func getProxyFromEnv(proxyScheme string) string {
 	return p
 }
 
+func setProxyEnv(key, value string) {
+	setEnv(strings.ToLower(key), value)
+	setEnv(strings.ToUpper(key), value)
+}
+
 func setEnv(key, value string) {
 	before := os.Getenv(key)
 	if before != value {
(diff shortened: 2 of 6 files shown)


pkg/crc/network/proxy.go
```diff
@@ -69,12 +69,12 @@ func NewProxyConfig() (*ProxyConfig, error) {
...
-	p := os.Getenv(proxyScheme)
-	if p == "" {
-		p = os.Getenv(strings.ToUpper(proxyScheme))
+func getEnv(key string) string {
+	p := os.Getenv(key)
+	if p != "" {
+		return p
 	}
-	return p
+	return os.Getenv(strings.ToUpper(key))
```
cfergeau, line 96: Looking for both `env_var` and `ENV_VAR` in the environment is actually something I've only seen for proxy variables, not something I'd do in general.

pkg/crc/network/proxy.go
```diff
@@ -18,7 +18,7 @@ var (
 type ProxyConfig struct {
 	HTTPProxy   string
 	HTTPSProxy  string
-	noProxy     []string
+	NoProxy     []string
```
praveenkumar, line 22: Making this exportable means the consumer of `ProxyConfig` can make use of `NoProxy`? In the commit logs it says `user might want to verify` (which user in this context) ?

pkg/crc/network/proxy_test.go
```diff
@@ -15,7 +15,7 @@ func TestValidateProxyURL(t *testing.T) {
 }
 
 func TestHidePassword(t *testing.T) {
-	url, err := URIStringForDisplay("https://user:secret@proxy.org:123")
+	url, err := hidePassword("https://user:secret@proxy.org:123")
```
praveenkumar, line 18: should this change be part of this commit, it should only have stringer change.

pkg/crc/network/proxy.go
```diff
@@ -118,16 +129,16 @@ func (p *ProxyConfig) ApplyToEnvironment() {
 	}
 
 	if p.HTTPProxy != "" {
-		os.Setenv("HTTP_PROXY", p.HTTPProxy)
-		os.Setenv("http_proxy", p.HTTPProxy)
+		setEnv("HTTP_PROXY", p.HTTPProxy)
+		setEnv("http_proxy", p.HTTPProxy)
```
cfergeau, line 133: Maybe go for `setProxyEnv("http_proxy", p.HTTPProxy)` which will set both upper and lowercase variables?

cfergeau (commented):
need to take a closer look as iirc there were a few subtleties to the proxy code, but seems to make sense overall.

pkg/crc/network/proxy.go
```diff
@@ -44,24 +44,24 @@ func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAFile string) (*Prox
...
 		noProxy = getEnv("no_proxy")
 	}
 	DefaultProxy.setNoProxyString(noProxy)
 
-	return NewProxyConfig()
+	return nil
 }
 
-// NewProxyConfig creates a proxy configuration with the specified parameters. If an empty string is passed
+// GetProxyConfig creates a proxy configuration with the specified parameters. If an empty string is passed
```
praveenkumar, line 63: This comment should be change now, it is now not creating the proxy config but updating the proxyConfig struct which is set by `SetproxyDefaults` and also the env check changes are not part of it now.

cfergeau:
> Add instanceIP in NO_PROXY at the top and not only in a small part of machine.Start()

Actually, I'm not sure `instanceIP` should be in `NO_PROXY` in general. In `Start()`, we need/used to need to access the VM before DNS is up, so could not use just `api.crc.testing` for HTTP interactions with the VM. The rest of the code and people using crc should probably use `api.crc.testing` or another DNS name, not the instance IP. If adding it results in simpler code, why not.

pkg/crc/network/proxy.go
guillaumerose: I dropped the rename :+1:

pkg/crc/network/proxy.go
guillaumerose: done :+1:

pkg/crc/network/proxy.go
guillaumerose: This struct is exposed in the daemon API when asking for status or start. I thought it would be good to expose this also.

pkg/crc/network/proxy.go
guillaumerose: done :+1:

pkg/crc/network/proxy_test.go
guillaumerose: I added more details in the commit message. Adding Stringer remove the need to have `URIStringForDisplay` public. I renamed it to be more in line with what it does.

guillaumerose:
> If adding it results in simpler code, why not.

Sadly no, it's not simpler but I found this less prone to errors. I don't like the mutation of the ProxyConfig in the middle of the start sequence.

I guess this NO_PROXY change was useful when `oc` ran on the host. But now, I guess no. So far, only external interactions are:
* fetch the latest crc version
* run oc login to initiate the kubeconfig (proxy already disabled).
Anything else? Otherwise, I will remove the `proxyConfig.ApplyToEnvironment()` in machine pkg.

pkg/crc/network/proxy.go
praveenkumar: @guillaumerose This info should be part of of logs to get the context :+1:

cfergeau:
> > If adding it results in simpler code, why not.
> 
> Sadly no, it's not simpler but I found this less prone to errors. I don't like the mutation of the ProxyConfig in the middle of the start sequence.

On that matter, see https://github.com/code-ready/crc/commit/91bf39357c , apparently we never want to have the instance IP in the `NO_PROXY` environment variable, it's only needed in the configuration which is applied inside the VM. In order to avoid the proxy configuration change, maybe `configProxyForCluster()` should be:
```
clusterProxyConfig := proxy.Copy()
clusterProxyConfig.AddNoProxy(instanceIP)
```
> 
> I guess this NO_PROXY change was useful when `oc` ran on the host. But now, I guess no. So far, only external interactions are:
> 
>     * fetch the latest crc version
> 
>     * run oc login to initiate the kubeconfig (proxy already disabled).
>       Anything else? Otherwise, I will remove the `proxyConfig.ApplyToEnvironment()` in machine pkg.

There are also the fetching of the misc binaries we need, which should not happen on most systems since they are bundled, and some of the DNS checks which are not using proxy, but which should.

cfergeau (commented):
> `Add ProxyConfig.NoProxy to JSON output`

I agree with Praveen that it's not obvious/clear what the goal is - edit: it is clear from the shortlog, but the long log is confusing :)

> `ProxyConfig implements Stringer`

This sounds as if it's already the case before the commit, I usually go with `Implement Stringer interface in ProxyConfig` or `Add String() implementation to ProxyConfig`.
I think long log needs `s/is now be/can now be/`

> `Reduce the use of the global variable for more clarity`

Really not clear to me what you mean by that :)
> `Reduce the use of the global variable for more clarity`

I was worried this would reintroduce https://github.com/code-ready/crc/commit/3ae05dac5 , but this seems ok.

Still looks good to me, some comments, but nothing major. I haven't run this for testing though ;)

pkg/crc/machine/machine.go
```diff
@@ -178,8 +178,17 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
...
 		}
 		if vmState == state.Running {
-			logging.Infof("A CodeReady Containers VM for OpenShift %s is already running", crcBundleMetadata.GetOpenshiftVersion())
-			clusterConfig, err := getClusterConfig(crcBundleMetadata)
+			openshiftVersion := crcBundleMetadata.GetOpenshiftVersion()
+			if openshiftVersion == "" {
+				logging.Info("A CodeReady Containers VM is already running")
+			} else {
+				logging.Infof("A CodeReady Containers VM for OpenShift %s is already running", openshiftVersion)
+			}
```
cfergeau, line 186: I think I forgot to point out this small rebase bug

pkg/crc/network/proxy.go
```diff
@@ -96,6 +97,16 @@ func getProxyFromEnv(proxyScheme string) string {
 	return p
 }
 
+func setEnv(key, value string) {
+	before := os.Getenv(key)
+	if before != value {
+		logging.Warnf("Overriding environment variable %s: previously %s, now %s", key, before, value)
```
cfergeau, line 108: `logging.Debugf` should be enough? This looks a bit weird when `before` is empty, could be wrapped between `[%s]` maybe, or special cased.

guillaumerose:
let's give up this, too old.

PR #1547, opened 2020-09-28
 Remove handling of GetOpenshiftVersion() returning ""

guillaumerose:
split of #1532 in order to make it more readable

Bundles shipped in current crc versions always have the OpenShift version set in their metadata, so we can remove the fallback code which was there in case it's not set.

### DIFF
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -138,12 +138,7 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 			return startError(startConfig.Name, fmt.Sprintf("Invalid bundle disk image '%s'", crcBundleMetadata.GetDiskImagePath()), err)
 		}
 
-		openshiftVersion := crcBundleMetadata.GetOpenshiftVersion()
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/machine.go
```diff
@@ -138,12 +138,7 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 			return startError(startConfig.Name, fmt.Sprintf("Invalid bundle disk image '%s'", crcBundleMetadata.GetDiskImagePath()), err)
 		}
 
-		openshiftVersion := crcBundleMetadata.GetOpenshiftVersion()
-		if openshiftVersion == "" {
-			logging.Info("Creating VM...")
-		} else {
-			logging.Infof("Creating CodeReady Containers VM for OpenShift %s...", openshiftVersion)
-		}
+		logging.Infof("Creating CodeReady Containers VM for OpenShift %s...", crcBundleMetadata.GetOpenshiftVersion())
```
praveenkumar, line 141: 👍

praveenkumar (approved):

PR #1548, opened 2020-09-28
Make creation and start sequence look the same

guillaumerose:
split of https://github.com/code-ready/crc/pull/1532

* Always check if ssh key is present and authorized_keys are up-to-date
* Always verify cluster ID
* Always verify pull secret (on disk and in the cluster)

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -16,6 +16,9 @@ import (
 	"github.com/pborman/uuid"
 )
 
+// #nosec G101
+const vmPullSecretPath = "/var/lib/kubelet/config.json"
+
 func WaitForSSH(sshRunner *ssh.Runner) error {
 	checkSSHConnectivity := func() error {
 		_, err := sshRunner.Run("exit 0")
@@ -251,12 +254,16 @@ func (p *PullSecret) Value() (string, error) {
 	return val, err
 }
 
-func AddPullSecretToInstanceDisk(sshRunner *ssh.Runner, pullSecret *PullSecret) error {
+func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret *PullSecret) error {
+	if _, err := sshRunner.Run(fmt.Sprintf("test -e %s", vmPullSecretPath)); err == nil {
+		return nil
+	}
+	logging.Info("Adding user's pull secret to instance disk...")
 	content, err := pullSecret.Value()
 	if err != nil {
 		return err
 	}
-	return sshRunner.CopyData([]byte(content), "/var/lib/kubelet/config.json", 0600)
+	return sshRunner.CopyData([]byte(content), vmPullSecretPath, 0600)
 }
 
 func WaitforRequestHeaderClientCaFile(ocConfig oc.Config) error {
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -84,8 +84,12 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 	return diskSize, diskUsage, nil
 }
 
-func AddPullSecretInTheCluster(ocConfig oc.Config, pullSec string) error {
-	base64OfPullSec := base64.StdEncoding.EncodeToString([]byte(pullSec))
+func AddPullSecretInTheCluster(ocConfig oc.Config, pullSec *PullSecret) error {
+	content, err := pullSec.Value()
+	if err != nil {
+		return err
+	}
+	base64OfPullSec := base64.StdEncoding.EncodeToString([]byte(content))
 	cmdArgs := []string{"patch", "secret", "pull-secret", "-p",
 		fmt.Sprintf(`'{"data":{".dockerconfigjson":"%s"}}'`, base64OfPullSec),
 		"-n", "openshift-config", "--type", "merge"}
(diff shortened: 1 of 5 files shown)


pkg/crc/machine/types.go
```diff
@@ -24,7 +25,7 @@ type StartConfig struct {
 	Debug bool
 
 	// User Pull secret
-	GetPullSecret GetPullSecretFunc
+	PullSecret *cluster.PullSecret
```
guillaumerose, line 28: > I don't see much difference in behaviour with or without this type, before we had a separate callback and string, now we have a struct holding the callback and the string, but ultimately we call the callback at about the same time, and we the struct will be carrying a string all along. In a way, the new type is somehow worse than the explicit start.GetPullSecret(), it's not obvious that _, err = startConfig.PullSecret.Value() is here to ensure we ask for the pull secret.
I don't mind this change, but I feel you are overselling it ;)

Yes you're right but now, with the EnsurePullSecret* functions, if by any chance, the pull secret is required it will need to ask for it. We don't know in advance if the pull secret will be needed or not in this case.

pkg/crc/machine/machine.go
```diff
@@ -111,24 +111,23 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 	}
 
 	// Pre-VM start
-	var pullSecret string
 	exists, err := client.Exists(startConfig.Name)
 	if err != nil {
 		return startError(startConfig.Name, "Cannot determine if VM exists", err)
 	}
 	if !exists {
+		_, err = startConfig.PullSecret.Value()
```
cfergeau, line 119: This deserves a comment such as `// Ask early for pull secret if it hasn't been requested yet`, as it's not obvious that a call to `Value()` could display a prompt

pkg/crc/cluster/cluster.go
```diff
@@ -87,7 +88,24 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
...
+		return err
+	}
+	decoded, err := base64.StdEncoding.DecodeString(stdout)
+	if err != nil {
+		return err
+	}
+	if err := validation.ImagePullSecret(string(decoded)); err == nil {
+		return nil
+	}
+
```
cfergeau, line 107: Should this check that `string(decoded)` is the same as `pullSec.Value()`?

guillaumerose: If we do that, each time we do a `crc start` it will ask for the pull secret.

pkg/crc/machine/machine.go
guillaumerose: done :+1:

pkg/crc/machine/machine.go
```diff
@@ -344,7 +344,11 @@ func (client *client) Start(startConfig StartConfig) (StartResult, error) {
 		}
 
 		logging.Info("Adding user's pull secret ...")
-		if err := cluster.AddPullSecret(sshRunner, ocConfig, pullSecret); err != nil {
+		if err := cluster.AddPullSecretToInstanceDisk(sshRunner, pullSecret); err != nil {
+			return startError(startConfig.Name, "Failed to update user pull secret", err)
```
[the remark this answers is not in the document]
guillaumerose: done :+1:

pkg/crc/machine/machine.go
```diff
@@ -681,16 +680,26 @@ func (*client) GetConsoleURL(consoleConfig ConsoleConfig) (ConsoleResult, error)
 }
 
 func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
-	// Generate ssh key pair
-	if err := ssh.GenerateSSHKey(constants.GetPrivateKeyPath()); err != nil {
-		return fmt.Errorf("Error generating ssh key pair: %v", err)
+	if _, err := os.Stat(constants.GetPrivateKeyPath()); err != nil && os.IsNotExist(err) {
```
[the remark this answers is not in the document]
guillaumerose: done :+1:

guillaumerose:
Thinking twice about the pull secret, I'm wondering if we should move it to `crc setup` and store it somewhere in ~/.crc.
There is already a config property for it and we need it for the daemon (in order to handle start command). WDYT?

cfergeau:
> Thinking twice about the pull secret, I'm wondering if we should move it to `crc setup` and store it somewhere in ~/.crc.
> There is already a config property for it and we need it for the daemon (in order to handle start command). WDYT?

I think we used to do that, but stopped for security reasons, see https://github.com/code-ready/crc/commit/da50d52534037a

cfergeau (approved):
Forgot to ack this I think.
One loose end which can be improved later is that `crc start` is not 100% error-proof yet:
1) run `crc start` -> this fails because it's interrupted/...
2) run `crc start` again, `crc` says that the cluster is already running (but it's actually non-functional because it was interrupted at 1)
3) running `crc stop -f; crc start` will fix the cluster (at least for missing ssh key/cluster id/pull secret)

Ideally, at 2) it would detect the cluster is not in working order, and suggest to stop/start to see if this helps. We could reuse code from this PR and use the same `check`/`fix` design as in the preflight checks. One drawback this has is that this would slow down `crc start` when the cluster is already running.

guillaumerose:
Rebased, will merge after the release cut.

cfergeau (approved):

guillaumerose:
Rebased. Hope the CI to be green.

Issue #1549, opened 2020-09-28
Unable to perform cert check when using native SSH implementation after key pivot


jsliacan:
Error checking `/var/lib/kubelet/pki/kubelet-client-current.pem` during the `crc start` process on Win10 with `4.6.0-fc.7` bundle. 

### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? Yes

## CRC version
```bash
CodeReady Containers version: 1.16.0+9cfefb2
OpenShift version: 4.6.0-fc.7 (not embedded in binary)
```
  
## CRC status
```bash
CRC VM:          Running
OpenShift:       Running (v4.6.0-fc.7)
Disk Usage:      12.99GB of 32.72GB (Inside the CRC VM)
Cache Usage:     15.34GB
Cache Directory: C:\Users\crcqe\.crc\cache
```

## CRC config
```bash
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
# put the output of `sw_vers` in case of Mac
# Put the output of `systeminfo` in case of Windows
```

### Steps to reproduce

  1. `crc setup`
  2. `crc start`

### Expected
CRC starts without errors.

### Actual
```
> crc setup 
INFO Checking if oc binary is cached
INFO Checking if podman remote binary is cached
INFO Checking if goodhosts binary is cached
INFO Checking if CRC bundle is cached in '$HOME/.crc'
INFO Checking minimum RAM requirements
INFO Checking if running as normal user
INFO Checking Windows 10 release
INFO Checking Windows edition
INFO Checking if Hyper-V is installed and operational
INFO Checking if user is a member of the Hyper-V Administrators group
INFO Checking if Hyper-V service is enabled
INFO Checking if the Hyper-V virtual switch exist
INFO Found Virtual Switch to use: Default Switch
Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster
```
and 
```
> crc start -b C:\Users\crcqe\Downloads\crc_hyperv_4.6.0-fc.7.crcbundle -p C:\Users\crcqe\Downloads\pull-secret --log-level=debug
DEBU CodeReady Containers version: 1.16.0+9cfefb2
DEBU OpenShift version: 4.6.0-fc.7 (not embedded in binary)
DEBU Total memory of system is 34186895360 bytes
[...]

PR #1550, opened 2020-09-29
Bump driver version and update machine revision in go.mod file

praveenkumar:
**Fixes:** Issue #1549


cfergeau (commented):
Might be nice to keep a link to https://app.circleci.com/pipelines/github/moby/hyperkit/jobs/769/artifacts so that we have a trace where the hyperkit binary came from

praveenkumar:
> Might be nice to keep a link to https://app.circleci.com/pipelines/github/moby/hyperkit/jobs/769/artifacts so that we have a trace where the hyperkit binary came from

@cfergeau I will put that in the release notes so we know from it is coming from.

PR #1553, opened 2020-09-29
Fix display of non-string value in SuccessfullyApplied callback

guillaumerose:
Use of cast.ToString instead of %v because spf13/cast package is widely used in this package.

This was introduced when enforcing the type of config properties.

### DIFF
--- a/pkg/crc/config/callbacks.go
+++ b/pkg/crc/config/callbacks.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"fmt"
+
+	"github.com/spf13/cast"
 )
 
 func RequiresRestartMsg(key string, _ interface{}) string {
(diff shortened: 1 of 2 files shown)


pkg/crc/config/callbacks.go
```diff
@@ -11,5 +13,5 @@ func RequiresRestartMsg(key string, _ interface{}) string {
 }
 
 func SuccessfullyApplied(key string, value interface{}) string {
-	return fmt.Sprintf("Successfully configured %s to %s", key, value)
+	return fmt.Sprintf("Successfully configured %s to %s", key, cast.ToString(value))
```
praveenkumar, line 16: @guillaumerose In the commit logs you mentioned remove `%v` but here we are already using `%s` and the value already casted to string?

guillaumerose: yes sorry I tried to say badly that I could either use cast.ToString or just use %v. I chose cast.ToString because we use it a lot in this package.

guillaumerose: added !

PR #1554, opened 2020-09-29
Add json ouput to crc console command

praveenkumar:


cmd/crc/cmd/console.go
```diff
@@ -31,45 +33,94 @@ var consoleCmd = &cobra.Command{
...
+		ClusterConfig:           toConsoleClusterConfig(&result),
+		Error:                   errorMessage(err),
+		ConsolePrintURL:         consolePrintURL,
+		ConsolePrintCredentials: consolePrintCredentials,
+	}, writer, outputFormat)
+}
+
+type consoleResult struct {
+	Success                 bool           `json:"success"`
+	State                   state.State    `json:"state"`
```
guillaumerose, line 67: This is redundant with `crc status`. Keep it private ?

cmd/crc/cmd/console.go
```diff
@@ -31,45 +33,94 @@ var consoleCmd = &cobra.Command{
...
+	Error                   string         `json:"error,omitempty"`
+	ClusterConfig           *clusterConfig `json:"clusterConfig,omitempty"`
+	ConsolePrintURL         bool           `json:"consolePrintUrl"`
```
guillaumerose, line 70: This is an implementation details. I don't think we need it in the json ouput. Just make it private ?

praveenkumar: sure we can make those private.

cmd/crc/cmd/console.go
praveenkumar: Ack

guillaumerose (approved):

cfergeau:
We probably should not add the `clusterConfig` block when the cluster does not exist:
```
$ ./out/linux-amd64/crc delete
$ ./out/linux-amd64/crc console -ojson
{
  "success": false,
  "error": "Machine 'crc' does not exist. Use 'crc start' to create it",
  "clusterConfig": {
    "webConsoleUrl": "",
    "url": "",
    "adminCredentials": {
      "username": "kubeadmin",
      "password": ""
    },
    "developerCredentials": {
      "username": "developer",
      "password": "developer"
    }
  }
}
```

PR #1558, opened 2020-09-30
Recreate config object before calling each handler to always get the latest configs

### LINKED ISSUE #1557: [BUG] Configs are not in sync when mixing the daemon api and cli to set/unset properties
### General information

  * OS: macOS

### Expected

The configs in the cli and configs in the daemon are identical

### Actual

Configs from the cli and the daemon are different. Below i am reproducing the issue, by setting the config property `cpus`

```bash
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ crc config view
- cpus                                  : 10
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ nc -U ~/.crc/crc.sock
{"command":"setconfig", "args":{"properties":{"cpus":8}}}
{"Error":"","Properties":["cpus"]}%                                                                                                                                                                                                         ╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ crc config view
- cpus                                  : 8
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ crc config set cpus 7
Changes to configuration property 'cpus' are only applied when a new CRC instance is created.
If you already have a CRC instance, then for this configuration change to take effect, delete the CRC instance with 'crc delete' and start a new one with 'crc start'.
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ crc config view
- cpus                                  : 7
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ nc -U ~/.crc/crc.sock
{"command":"getconfig", "args":{"properties":["cpus"]}}
{"Error":"","Configs":{"cpus":8}}%                                                                                                                                                                                                          ╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ crc config unset cpus
Successfully unset configuration property 'cpus'
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ crc config view
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹master›
╰─$ nc -U ~/.crc/crc.sock
{"command":"getconfig", "args":{"properties":["cpus"]}}
{"Error":"","Configs":{"cpus":8}}%
```

### LINKED ISSUE #1576: [BUG] while doing start from api preflight checks are not performed

anjannath:
Fixes #1557 
Fixes #1576 

The `viper.WatchConfig` is not able to update the config properties on the daemon side, when the  same config property is set via cli and also from daemon. 
To solve this for config 'set' we watch writes to the config file and use `set` method from `config` to explicitly set the config properties read from the config file

## Testing

Below what am trying to do is, set/get config by mixing cli and the daemon api, whatever was set last should be returned from both cli and daemon.

```bash
## Checking the current config
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹config_watch›
╰─$ nc -U ~/.crc/crc.sock
{"command":"getconfig", "args":{"properties":["cpus"]}}
{"Error":"","Configs":{"cpus":5}}%

## Set cpus using cli                                                                                                                                                                                                          ╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹config_watch›
╰─$ crc config set cpus 8
Changes to configuration property 'cpus' are only applied when a new CRC instance is created.
If you already have a CRC instance, then for this configuration change to take effect, delete the CRC instance with 'crc delete' and start a new one with 'crc start'.

## Set cpus using daemon api
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹config_watch›
╰─$ nc -U ~/.crc/crc.sock
{"command":"setconfig", "args":{"properties":{"cpus":9}}}
{"Error":"","Properties":["cpus"]}%                                                                                                                                                                                                         

## Get cpus using cli
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹config_watch›
╰─$ crc config view
- cpus                                  : 9

## Again set using cli
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹config_watch›
╰─$ crc config set cpus 10                                                                                                                                                                                                            130 ↵
Changes to configuration property 'cpus' are only applied when a new CRC instance is created.
If you already have a CRC instance, then for this configuration change to take effect, delete the CRC instance with 'crc delete' and start a new one with 'crc start'.

## Get using daemon api
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹config_watch›
╰─$ nc -U ~/.crc/crc.sock
{"command":"getconfig", "args":{"properties":["cpus"]}}
{"Error":"","Configs":{"cpus":10}}%
```

### DIFF
--- a/cmd/crc/cmd/daemon_nonwin.go
+++ b/cmd/crc/cmd/daemon_nonwin.go
@@ -6,15 +6,14 @@ import (
 	"os"
 
 	"github.com/code-ready/crc/pkg/crc/api"
-	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
-func runDaemon(config crcConfig.Storage) {
+func runDaemon() {
 	// Remove if an old socket is present
 	os.Remove(constants.DaemonSocketPath)
-	crcAPIServer, err := api.CreateAPIServer(constants.DaemonSocketPath, config)
+	crcAPIServer, err := api.CreateAPIServer(constants.DaemonSocketPath, newConfig)
 	if err != nil {
 		logging.Fatal("Failed to launch daemon", err)
 	}
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -44,14 +44,10 @@ func init() {
 		logging.Fatal(err.Error())
 	}
 	var err error
-	viper, err = crcConfig.NewViperStorage(constants.ConfigPath, constants.CrcEnvPrefix)
+	config, viper, err = newViperConfig()
 	if err != nil {
 		logging.Fatal(err.Error())
 	}
-	config = crcConfig.New(viper)
-	cmdConfig.RegisterSettings(config)
-
-	preflight.RegisterSettings(config)
 
 	// subcommands
 	rootCmd.AddCommand(cmdConfig.GetConfigCmd(config))
@@ -132,4 +128,15 @@ func getProxyCAData(proxyCAFile string) (string, error) {
 
 func trimTrailingEOL(s string) string {
 	return strings.TrimRight(s, "\n")
+}
+
+func newViperConfig() (*crcConfig.Config, *crcConfig.ViperStorage, error) {
+	viper, err := crcConfig.NewViperStorage(constants.ConfigPath, constants.CrcEnvPrefix)
+	if err != nil {
+		return nil, nil, err
+	}
+	cfg := crcConfig.New(viper)
+	cmdConfig.RegisterSettings(cfg)
+	preflight.RegisterSettings(cfg)
+	return cfg, viper, nil
 }
--- a/pkg/crc/api/api.go
+++ b/pkg/crc/api/api.go
@@ -6,25 +6,24 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine"
 )
 
-func CreateAPIServer(socketPath string, config config.Storage) (CrcAPIServer, error) {
+func CreateAPIServer(socketPath string, newConfig newConfigFunc) (CrcAPIServer, error) {
 	listener, err := net.Listen("unix", socketPath)
 	if err != nil {
 		logging.Error("Failed to create socket: ", err.Error())
 		return CrcAPIServer{}, err
 	}
-	return createAPIServerWithListener(listener, machine.NewClient(), config)
+	return createAPIServerWithListener(listener, machine.NewClient(), newConfig)
 }
 
-func createAPIServerWithListener(listener net.Listener, client machine.Client, config config.Storage) (CrcAPIServer, error) {
+func createAPIServerWithListener(listener net.Listener, client machine.Client, newConfig newConfigFunc) (CrcAPIServer, error) {
 	apiServer := CrcAPIServer{
 		client:                 client,
-		config:                 config,
 		listener:               listener,
+		newConfig:              newConfig,
 		clusterOpsRequestsChan: make(chan clusterOpsRequest, 10),
 		handlers: map[string]handlerFunc{
 			"start":         startHandler,
@@ -63,7 +62,12 @@ func (api CrcAPIServer) handleRequest(req commandRequest, conn net.Conn) {
 	defer conn.Close()
 	var result string
 	if handler, ok := api.handlers[req.Command]; ok {
-		result = handler(api.client, api.config, req.Args)
+		config, err := api.newConfig()
+		if err != nil {
+			result = encodeErrorToJSON(fmt.Sprintf("Failed to initialize new config store: %v", err))
+		} else {
+			result = handler(api.client, config, req.Args)
+		}
 	} else {
 		result = encodeErrorToJSON(fmt.Sprintf("Unknown command supplied: %s", req.Command))
 	}
--- a/pkg/crc/api/api.go
+++ b/pkg/crc/api/api.go
@@ -7,7 +7,6 @@ import (
 	"net"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
-	"github.com/code-ready/crc/pkg/crc/machine"
 )
 
 func CreateAPIServer(socketPath string, newConfig newConfigFunc) (CrcAPIServer, error) {
@@ -16,26 +15,15 @@ func CreateAPIServer(socketPath string, newConfig newConfigFunc) (CrcAPIServer, 
 		logging.Error("Failed to create socket: ", err.Error())
 		return CrcAPIServer{}, err
 	}
-	return createAPIServerWithListener(listener, machine.NewClient(), newConfig)
+	return createAPIServerWithListener(listener, newConfig, newHandler())
 }
 
-func createAPIServerWithListener(listener net.Listener, client machine.Client, newConfig newConfigFunc) (CrcAPIServer, error) {
+func createAPIServerWithListener(listener net.Listener, newConfig newConfigFunc, handler RequestHandler) (CrcAPIServer, error) {
 	apiServer := CrcAPIServer{
-		client:                 client,
 		listener:               listener,
 		newConfig:              newConfig,
 		clusterOpsRequestsChan: make(chan clusterOpsRequest, 10),
-		handlers: map[string]handlerFunc{
-			"start":         startHandler,
-			"stop":          stopHandler,
-			"status":        statusHandler,
-			"delete":        deleteHandler,
-			"version":       versionHandler,
-			"setconfig":     setConfigHandler,
-			"unsetconfig":   unsetConfigHandler,
-			"getconfig":     getConfigHandler,
-			"webconsoleurl": webconsoleURLHandler,
-		},
+		handler:                handler,
 	}
 	return apiServer, nil
 }
@@ -61,14 +49,35 @@ func (api CrcAPIServer) handleClusterOperations() {
 func (api CrcAPIServer) handleRequest(req commandRequest, conn net.Conn) {
 	defer conn.Close()
 	var result string
-	if handler, ok := api.handlers[req.Command]; ok {
-		config, err := api.newConfig()
-		if err != nil {
-			result = encodeErrorToJSON(fmt.Sprintf("Failed to initialize new config store: %v", err))
-		} else {
-			result = handler(api.client, config, req.Args)
-		}
-	} else {
+
+	config, err := api.newConfig()
+	if err != nil {
+		logging.Error(err.Error())
+		result = encodeErrorToJSON(fmt.Sprintf("Failed to initialize new config store: %v", err))
+		writeStringToSocket(conn, result)
+		return
+	}
+
+	switch req.Command {
+	case "start":
+		result = api.handler.Start(config, req.Args)
+	case "stop":
+		result = api.handler.Stop()
+	case "status":
+		result = api.handler.Status()
+	case "delete":
+		result = api.handler.Delete()
+	case "version":
+		result = api.handler.GetVersion()
+	case "setconfig":
+		result = api.handler.SetConfig(config, req.Args)
+	case "unsetconfig":
+		result = api.handler.UnsetConfig(config, req.Args)
+	case "getconfig":
+		result = api.handler.GetConfig(config, req.Args)
+	case "webconsoleurl":
+		result = api.handler.GetWebconsoleInfo()
+	default:
 		result = encodeErrorToJSON(fmt.Sprintf("Unknown command supplied: %s", req.Command))
 	}
 	writeStringToSocket(conn, result)
@@ -91,21 +100,28 @@ func (api CrcAPIServer) handleConnections(conn net.Conn) {
 	// start, stop and delete are slow operations, and change the VM state so they have to run sequentially.
 	// We don't want other operations querying the status of the VM to be blocked by these,
 	// so they are treated by a dedicated go routine
-	if req.Command == "start" || req.Command == "stop" || req.Command == "delete" {
+
+	switch req.Command {
+	case "start", "stop", "delete":
 		// queue new request to channel
 		r := clusterOpsRequest{
 			command: req,
 			socket:  conn,
 		}
 		if !addRequestToChannel(r, api.clusterOpsRequestsChan) {
-			defer conn.Close()
 			logging.Error("Channel capacity reached, unable to add new request")
 			errMsg := encodeErrorToJSON("Sockets channel capacity reached, unable to add new request")
 			writeStringToSocket(conn, errMsg)
-			return
+			conn.Close()
 		}
-	} else {
+
+	case "status", "version", "setconfig", "getconfig", "unsetconfig", "webconsoleurl":
 		go api.handleRequest(req, conn)
+
+	default:
+		err := encodeErrorToJSON(fmt.Sprintf("Unknown command supplied: %s", req.Command))
+		writeStringToSocket(conn, err)
+		conn.Close()
 	}
 }
 
--- a/pkg/crc/api/api_test.go
+++ b/pkg/crc/api/api_test.go
@@ -4,6 +4,7 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
@@ -12,6 +13,7 @@ import (
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	"github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/machine/fakemachine"
 	"github.com/code-ready/crc/pkg/crc/preflight"
 	"github.com/code-ready/crc/pkg/crc/version"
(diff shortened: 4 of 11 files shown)


pkg/crc/config/viper_config.go
```diff
@@ -36,7 +39,33 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
 	}
 	// only watch config file if not running in a terminal/interactively
 	if !terminal.IsTerminal(int(os.Stdout.Fd())) {
-		v.WatchConfig()
+		go func() {
+			var tempConfig map[string]interface{} = make(map[string]interface{})
```
praveenkumar, line 44: why not just `var tempConfig = make(map[string]interface{})`

pkg/crc/config/viper_config.go
```diff
@@ -37,6 +37,7 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error reading configuration file '%s': %v", configFile, err)
 	}
+	v.WatchConfig()
```
praveenkumar, line 40: Why don't you make this change as part of https://github.com/code-ready/crc/pull/1558/commits/3a15992263c6aad9c922031dea22505b7935c371 commit itself? I mean it doesn't make sense you first remove the Watchconfig and then add it back?

pkg/crc/config/viper_config.go
```diff
@@ -33,7 +34,11 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error reading configuration file '%s': %v", configFile, err)
 	}
-	v.WatchConfig()
+	// only watch config file if not running in a terminal/interactively
+	if !terminal.IsTerminal(int(os.Stdout.Fd())) {
```
cfergeau, line 42: This is hardcoding some crc-specific knowledge in a package which should be generic. I'd prefer that you add a "WatchConfig()" method to the viper storage that the daemon calls explicitly.

pkg/crc/config/viper_config.go
anjannath: Will do, i removed it then the test failed so added it again!

cfergeau (commented):
> the viper.WatchConfig is not able to update the config properties
on the daemon side, when the  same config property is set via cli
and also from daemon.

This is imo missing a "why this is not working"? This commit is mostly duplicating the code from viper so it would be nice to understand why their code is not working, and ours is?

pkg/crc/config/viper_config.go
```diff
@@ -34,6 +38,42 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
 		return nil, fmt.Errorf("error reading configuration file '%s': %v", configFile, err)
 	}
 	v.WatchConfig()
+	// only watch config file if not running in a terminal/interactively
+	if !terminal.IsTerminal(int(os.Stdout.Fd())) {
+		go func() {
+			var tempConfig map[string]interface{} = make(map[string]interface{})
+			w, err := fsnotify.NewWatcher()
+			if err != nil {
```
cfergeau, line 46: Have you considered using `func (v *Viper) OnConfigChange(run func(in fsnotify.Event))` instead of reimplementing the whole file watching?

anjannath: Yes, i did try using `OnConfigChange` at first, but this only gets called if we have the `viper.WatchConfig`, and `viper.WatchConfig` automatically calls `ReadInConfig`, which is supposed to update the config but it doesn't.

So i avoided using these two methods and directly reimplement it in here.

anjannath:
> This is imo missing a "why this is not working"? This commit is mostly duplicating the code from viper so it would be nice to understand why their code is not working, and ours is?

Yes it is mostly duplicating the watch part, but in viper when it detects changes on the file, it calls `ReadInConfig`, in this commit the config is set using `viper.Set`. I suspect `viper.ReadInConfig` behaves differently when called multiple time.

pkg/crc/config/viper_config.go
anjannath: Changed

pkg/crc/config/viper_config.go
cfergeau: Even if `viper.WatchConfig` is not working as expected, wouldn't it be less code to use `WatchConfig` + `OnConfigChange` rather than having a watch ourselves?

anjannath: Yes, it'll be less code, will switch to using `WatchConfig` + `OnConfigChange`

cfergeau:
> > This is imo missing a "why this is not working"? This commit is mostly duplicating the code from viper so it would be nice to understand why their code is not working, and ours is?
> 
> Yes it is mostly duplicating the watch part, but in viper when it detects changes on the file, it calls `ReadInConfig`, in this commit the config is set using `viper.Set`. I suspect `viper.ReadInConfig` behaves differently when called multiple time.
> 
> But, i don't have an answer for the why, as this only fixes the `Set` part, if something is `Unset` we have the same not in sync issue again.

I did not look at `Unset` which might be its own can of worms, but I investigated why `Set` is not working. First, the exact bug is that after doing `{"command":"setconfig", "args":{"properties":{"cpus":8}}}` in the daemon, then `crc config set cpus 6` will no longer be reflected in the daemon process.
What is happening:
- `v.Set("cpus", 8)` will be roughly doing `v.override["cpus"] = 8`
- the config watch will call `ReadInConfig()` which reloads the config file, and sets the value of the `v.config` map to the config file content
- however, `v.Get("cpus")` will first look for `"cpus"` in `v.override`, and only if it's not present there, it will search for it in `v.config`. So once `v.Set()` has been called in the process for a config key, `WatchConfig` will never again refresh its value.

pkg/crc/config/viper_config.go
```diff
@@ -34,6 +38,42 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
...
+				logging.Error(err)
+				return
+			}
+			for {
+				select {
+				case e := <-w.Events:
+					if e.Op == fsnotify.Write {
+						cnf, err := ioutil.ReadFile(configFile)
+						if err != nil {
+							logging.Debug(err)
```
guillaumerose, line 61: Add `continue`. In case of error, cnf might be nil and crash couple of lines after.

pkg/crc/config/viper_config.go
```diff
@@ -34,6 +38,42 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
...
+						}
+						if err := json.Unmarshal(cnf, &tempConfig); err != nil {
+							logging.Error(err)
```
guillaumerose, line 64: Same as above, `continue`

pkg/crc/config/viper_config.go
```diff
@@ -8,8 +8,12 @@ import (
 	"os"
 	"strings"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"golang.org/x/crypto/ssh/terminal"
+
```
guillaumerose, line 15: Extra space :)

pkg/crc/config/viper_config.go
```diff
@@ -34,6 +38,42 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
...
+					if e.Op == fsnotify.Write {
+						cnf, err := ioutil.ReadFile(configFile)
+						if err != nil {
+							logging.Debug(err)
+						}
+						if err := json.Unmarshal(cnf, &tempConfig); err != nil {
+							logging.Error(err)
+						}
+						for key, value := range tempConfig {
+							v.Set(key, value)
```
guillaumerose, line 67: Maybe log the error if Set fails.

guillaumerose:
Looks that we don't have the choice. Can you try to run `go test` with `-race` option ? Currently it crashes due to WatchConfig.

cfergeau:
> Looks that we don't have the choice. Can you try to run `go test` with `-race` option ? Currently it crashes due to WatchConfig.
> 
> [spf13/viper#174](https://github.com/spf13/viper/issues/174)

Ah that's right, avoiding usage of `viper.WatchConfig` and `OnConfigChange` (and reimplement the file watching) would allow us to avoid that race.

guillaumerose:
Thinking about it twice, I am wondering if for the daemon we could create a new instance of config.Storage each time a handler is called. It would probably solve this issue.

cfergeau:
> Thinking about it twice, I am wondering if for the daemon we could create a new instance of config.Storage each time a handler is called. It would probably solve this issue.

Sounds closer to the behaviour we want/expect. crc code is not going to be able to deal well witha config parameter being change during the execution of a command/handler.

anjannath:
> Thinking about it twice, I am wondering if for the daemon we could create a new instance of config.Storage each time a handler is called. It would probably solve this issue.

Yes, that way the behavior should be same as running from the cli, trying this way!

anjannath:
> Thinking about it twice, I am wondering if for the daemon we could create a new instance of config.Storage each time a handler is called. It would probably solve this issue.

This indeed works as we expect it to, will update the PR, thanks a lot!

pkg/crc/api/api_test.go
```diff
@@ -167,6 +153,14 @@ func TestGetconfigApi(t *testing.T) {
 	}, getconfigRes)
 }
 
+func setupInMemoryConfigAndHandle(fn handlerFunc, client machine.Client, args json.RawMessage) string {
+	storage := config.NewEmptyInMemoryStorage()
+	cfg := config.New(storage)
+	cmdConfig.RegisterSettings(cfg)
+
+	return fn(client, cfg, args)
+}
```
cfergeau, line 162: Not clear why you need to pass `handlerFunc()`  to this? Seems like this would be simpler as
```
func inMemoryConfigFactory() config.Config {
storage := config.NewEmptyInMemoryStorage()
	cfg := config.New(storage)
	cmdConfig.RegisterSettings(cfg)

        return cfg
}
```
and `handleRequest` becomes
```
	if handler, ok := api.handlers[req.Command]; ok {
		config := api.configFactory() // or api.newConfig() or ... - this is the field which is currently called prehook
		result = handler(api.client, config, req.Args)
	} else {
...
```

anjannath: For the `NewEmptyInMemoryStorage` there is no error returned, so this is simple, but in case of the `NewViperStorage` it returns an error, and we'll have to check that in `handleRequest` before calling the handler, so then it'll be.. 

```golang
        if handler, ok := api.handlers[req.Command]; ok {
		config, err := api.newConfig()
                if err != nil {
                       result = encodeErrorToJSON(err.Error())
                } else {
		       result = handler(api.client, config, req.Args)
                }
	} else {
```

cfergeau: Seems ok this way. This also allows you to add a test for the error case if needed.

anjannath: Oh, you mean as in the previous comment https://github.com/code-ready/crc/pull/1558#discussion_r500123532 i thought you'd not like the nested ifs :)

cfergeau: Indeed not a big fan of the nesting, but it seems better than the alternatives in this case.

anjannath: Made the changes!

pkg/crc/api/handlers.go
```diff
@@ -16,6 +16,17 @@ import (
...
+func NewConfig() (crcConfig.Storage, error) {
+	viper, err := crcConfig.NewViperStorage(constants.ConfigPath, constants.CrcEnvPrefix)
+	if err != nil {
+		return nil, err
+	}
+	cfg := crcConfig.New(viper)
+	config.RegisterSettings(cfg)
+
+	return cfg, nil
+}
```
cfergeau, line 30: Does this have to live in `pkg/crc/api/handlers.go`? `cmd/crc/cmd/root.go` has very similar code, maybe it's a good opportunity to share it?

cfergeau (commented):
Would not hurt to have a few more details in the commit logs, for example why `Run preflight checks on start via api`? (because it's been forgotten, and we want the same behaviour as in `cmd/crc/cmd/start.go`)

anjannath:
> Would not hurt to have a few more details in the commit logs, for example why `Run preflight checks on start via api`? (because it's been forgotten, and we want the same behaviour as in `cmd/crc/cmd/start.go`)

Sorry, will update the commit message :)

pkg/crc/api/handlers.go
anjannath: will have to change the return type to `config.Config` instead of `config.Storage`

cmd/crc/cmd/root.go
```diff
@@ -133,3 +129,14 @@ func getProxyCAData(proxyCAFile string) (string, error) {
 func trimTrailingEOL(s string) string {
 	return strings.TrimRight(s, "\n")
 }
+
+func newConfig() (*crcConfig.Config, error) {
```
anjannath, line 133: Now am thinking if i should have put this in `cmd/config`!!

anjannath:
@cfergeau @guillaumerose Rebased and updated the PR, please have a look again!

pkg/crc/api/types.go
```diff
@@ -11,13 +11,15 @@ import (
 
 type handlerFunc func(machine.Client, config.Storage, json.RawMessage) string
 
+type newViperConfigFunc func() (*config.Config, *config.ViperStorage, error)
```
cfergeau, line 10: surprised to see concrete types here rather than interfaces? Would be nicer as:
```diff
-type newViperConfigFunc func() (*config.Config, *config.ViperStorage, error)
+type newConfigFunc func() (config.Storage, error)
```

pkg/crc/api/api.go
```diff
@@ -61,15 +49,36 @@ func (api CrcAPIServer) handleClusterOperations() {
...
+	case "setconfig":
+		result = api.handler.SetConfig(config, req.Args)
+	case "unsetconfig":
+		result = api.handler.UnsetConfig(config, req.Args)
+	case "getconfig":
+		result = api.handler.GetConfig(config, req.Args)
+	case "webconsoleurl":
+		result = api.handler.GetWebconsoleInfo()
+	default:
+		result = encodeErrorToJSON("Unexpected error occurred")
```
cfergeau, line 81: You can keep `result = encodeErrorToJSON(fmt.Sprintf("Unknown command supplied: %s", req.Command))`

pkg/crc/api/api.go
```diff
@@ -91,21 +100,28 @@ func (api CrcAPIServer) handleConnections(conn net.Conn) {
...
-	} else {
+
+	case "status", "version", "setconfig", "getconfig", "unsetconfig", "webconsoleurl":
 		go api.handleRequest(req, conn)
+
+	default:
+		err := encodeErrorToJSON(fmt.Sprintf("Unknown command supplied: %s", req.Command))
+		writeStringToSocket(conn, err)
+		conn.Close()
 	}
```
cfergeau, line 125: I don't think this change in `handleConnections` is related to the handler interface change? Maybe indirectly because you moved the "unknown command" reporting here rathre than in `handleRequest`?

pkg/crc/api/handlers.go
```diff
@@ -17,22 +17,32 @@ import (
...
 		Name:  constants.DefaultName,
 		Debug: true,
 	}
-	commandResult, _ := client.Stop(stopConfig)
+	commandResult, _ := h.MachineClient.Stop(stopConfig)
 	return encodeStructToJSON(commandResult)
 }
 
-func startHandler(client machine.Client, crcConfig crcConfig.Storage, args json.RawMessage) string {
+func (h *Handler) Start(crcConfig crcConfig.Storage, args json.RawMessage) string {
```
cfergeau, line 46: Would it make sense to have the `crcConfig` be part of your handler?

cmd/crc/cmd/daemon_nonwin.go
```diff
@@ -13,7 +13,7 @@ import (
 func runDaemon() {
 	// Remove if an old socket is present
 	os.Remove(constants.DaemonSocketPath)
-	crcAPIServer, err := api.CreateAPIServer(constants.DaemonSocketPath, newViperConfig)
+	crcAPIServer, err := api.CreateAPIServer(constants.DaemonSocketPath, newViperConfig, api.NewHandler())
```
cfergeau, line 16: For `CreateAPIServer` you will always be passing `api.NewHandler()`, the tests use `createAPIWithListener` directly, maybe you don't need to add the `handler` argumen to `CreateAPIServer` and only add it to `CreateAPIWithListener`?

pkg/crc/api/api_test.go
```diff
@@ -30,6 +32,21 @@ func newFakeHandler(client *fakemachine.Client) *fakeHandler {
...
+		startErr := &machine.StartResult{
+			Name:  "crc",
+			Error: fmt.Sprintf("Incorrect arguments given: %s", err.Error()),
+		}
+		return encodeStructToJSON(startErr)
+	}
+	startConfig := getStartConfig(crcConfig, parsedArgs)
+	status, _ := f.MachineClient.Start(startConfig)
+	return encodeStructToJSON(status)
+}
```
cfergeau, line 48: What would be great in here would be to have a way to call the normal `api.Start()` handler, with a mock'ed `MachineClient`, but also with mock'ed preflight checks. These fake preflight checks would be just one or 2 checks which would set booleans which we'd use to make sure they were run. I don't know if we can dynamically replace `getPreflightChecks` at runtime ^^ (just food for thoughts, I don't think there is much which can be done right now)

pkg/crc/api/handlers.go
anjannath: not really, we can have the `newConfigFunc` as part of the handler, since we need to re-create the config every time, and only some of the handler functions need it, so thought it'd be better to just pass it those ones who need it.

pkg/crc/api/api.go
anjannath: Yeah indirectly, there are two places where commands are filtered, and i went with different message, actually both should report "Unknown command"

cmd/crc/cmd/daemon_nonwin.go
anjannath: Done!

pkg/crc/api/handlers.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/version"
 )
 
-func statusHandler(client machine.Client, crcConfig crcConfig.Storage, _ json.RawMessage) string {
+func statusHandler(client machine.Client, _ crcConfig.Storage, _ json.RawMessage) string {
```
cfergeau: Would fit better in ` Ignore unused crcConfig in some of the handlers `

cfergeau (approved):
Looks good to me!

guillaumerose:
A good move!

/lgtm

Issue #1556, opened 2020-09-29
Add TLS CAs to json output


cfergeau:
Any json output which returns an https URL should also
contain the TLS CAs which are going to be used when connecting 
to this https endpoint, as it's likely to be using a custom CA.

gbraad:
What would you include? the CN, the fingerprint or the actual content?

cfergeau:
It must be the actual content so that one can do `curl --cacert xxx https://api.crc.testing:6443` after extracting the CA from the JSON data.

guillaumerose:
PR merged

PR #1560, opened 2020-10-01
Issue #1556 Add cluster CA info to json output

praveenkumar:
As part of cluster config we should have cluster CA info also so the
respective client can consume it instead using insecure parameter.

```
$ crc start -b ~/Downloads/crc_libvirt_4.5.9.crcbundle -ojson | jq -r .clusterConfig.cacert > cluster.cacert

$ curl https://api.crc.testing:6443/version
curl: (60) SSL certificate problem: self signed certificate in certificate chain
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.

$ curl --cacert cluster.cacert https://api.crc.testing:6443/version
{
  "major": "1",
  "minor": "18+",
  "gitVersion": "v1.18.3+6c42de8",
  "gitCommit": "6c42de8",
  "gitTreeState": "clean",
  "buildDate": "2020-09-04T14:16:02Z",
  "goVersion": "go1.13.4",
  "compiler": "gc",
  "platform": "linux/amd64"
}
```

**Fixes:** #1556


pkg/crc/machine/kubeconfig.go
```diff
@@ -53,6 +53,14 @@ func isReady(status *cluster.Status) bool {
 	return status.Available && !status.Progressing && !status.Degraded && !status.Disabled
 }
 
+func GetClusterCACert(kubeconfigFile string) (string, error) {
```
guillaumerose, line 56: This function looks like `certificateAuthority` few lines below. Can we merge both ?

pkg/crc/machine/kubeconfig_test.go
```diff
@@ -0,0 +1,39 @@
...
...= "-----BEGIN CERTIFICATE-----\nMIIDODCCAiCgAwIBAgIIRVfCKNUa1wIwDQYJKoZIhvcNAQELBQAwJjEkMCIGA1UE\nAwwbaW5ncmVzcy1vcGVyYXRvckAxNTk5OTkxMDc4MB4XDTIwMDkxMzA5NTgwOFoX\nDTIyMDkxMzA5NTgwOVowHTEbMBkGA1UEAwwSKi5hcHBzLWNyYy50ZXN0aW5nMIIB\nIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAtClYnM1FYiH+wcYiGIUnLvNi\nzLWWjsqJrQ/4Snnu61HDaU5w3An9yDZojyaJdCmWUg6CKXDiCoJB+lxMFdyaolXU\ndohJ9vr2wt6iuNfshmtxUwiBhI9ZBsVhztWdu3cgnUcYW8KMyUmajiEyXD8Npvba\nZ4ifUsjAYE1LByZzPIkmBmKPnv0fFZu1ejgg7HuQlYmfN/pJHudWMvwZJ8b3Zhqn\nsA0cDsaCDU1SJk6cQvNJFgF38+IWNhJpiQlGfFwKkal64/RYcR9yv8/pMV+7jj1z\nBPI9yxbFgLuxLxlyS8Kxnz9ln/4V0ZB12qU3c6aM4Ew3uhYv8ZDSOnpTUSjrRwID\nAQABo3MwcTAOBgNVHQ8BAf8EBAMCBaAwEwYDVR0lBAwwCgYIKwYBBQUHAwEwDAYD\nVR0TAQH/BAIwADAdBgNVHQ4EFgQUTtNCBUyuPpf4M4YhlYBqJhZGQYYwHQYDVR0R\nBBYwFIISKi5hcHBzLWNyYy50ZXN0aW5nMA0GCSqGSIb3DQEBCwUAA4IBAQBkGZiv\nBqqEDKUTiifFfTxQXJzO5OBBTUDqSntrknAw0sidPgn9A8a2gGdCr7mKEH16FQ7N\n1SpkjXWZghE/1pZTaS7JVcT6+Yuplfy5Reoim/Nlu8ulDDfBSSp9S9BO+Q/lFJdM\nomwyIu0sSXI79ColhLirDinEmyQMDHdmTJ+kJfctpSH27L4j5osJxtNSAEABmUgX\n1HQ7FYzrNmys5OcK9SlDn0vCyOTHyqBbxRnF6L9Ec2bU6KD0DvEBurBpAYGMc9ss\nBD0FXdHEPg7HP0Nep79jhe10IXGrghtep3D5jUjbj1I4DGSsQ8y4df2vo6IFd96A\nf0uUS/73sncN64gl\n-----END CERTIFICATE-----\n"
```
cfergeau, line 37: Wondering if this should also be base64-encoded in the json we return with `-ojson`?

praveenkumar: @cfergeau if we make it base64 encoded than to have it in the file we again need to decode it and then have it which is one more step.

pkg/crc/machine/kubeconfig.go
praveenkumar: @guillaumerose right, done :+1:

pkg/crc/machine/kubeconfig_test.go
guillaumerose: in ~/.kube/config CA is base64, I guess we should follow them.

praveenkumar: @guillaumerose yes but if we parse that using client tool then the return []byte are decoded one, if you folks want the encoded one then we need to again encode it.

cfergeau: I know base64 is inconvenient, but the certificates contains `\n` characters, I have no idea how well this is going to be supported if we put this unencoded in yaml data?

gbraad:
Are we returning a cert? And if so, is this the public or private part?

pkg/crc/machine/kubeconfig_test.go
gbraad: Everything between the lines:  `-----BEGIN CERTIFICATE -----` and. `-----END CERTIFICATE -----` are already BASE64 encoded. 

I would instead here focus on the line ending issue... though, does YAML not support multi-line entries? Or what about just stripping the line endings?

praveenkumar: It is not encoded right now, if all agree I just encode it and then consumers need to decode it and use accordingly.

praveenkumar:
> Are we returning a cert? And if so, is this the public or private part?

@gbraad we are returning the CA (certificate authority) not any certs.

gbraad:
How do you identify the CAcert, as a private or public certificate? It is a chain of trust, so this itself is a certificate

cfergeau:
> How do you identify the CAcert, as a private or public certificate? It is a chain of trust, so this itself is a certificate

It's the public one as this allows you to check your connection to api.crc.testing, I'd expect the private one (which you'd use to sign new certificates) to be much harder to get access to, and in this case we don't need it.

pkg/crc/machine/machine.go
```diff
@@ -53,7 +54,12 @@ func getClusterConfig(bundleInfo *bundle.CrcBundleInfo) (*ClusterConfig, error)
 	if err != nil {
 		return nil, err
 	}
+	clusterCACert, err := certificateAuthority(bundleInfo.GetKubeConfigPath())
+	if err != nil {
+		return nil, err
+	}
 	return &ClusterConfig{
+		ClusterCACert: base64.StdEncoding.EncodeToString(clusterCACert),
```
praveenkumar, line 62: @cfergeau this will make sure that `-ojson` will have base64 encoded certs.

guillaumerose (approved):

PR #1561, opened 2020-10-01
Allow daemon start handler to accept pull secret path as an argument

anjannath:
**Relates to:** https://github.com/code-ready/tray-macos/issues/24

## Testing
```bash
$ nc -U ~/.crc/crc.sock
{"command":"start", "args":{"pullSecretFile":"/User/anjan/Downloads/pull-secret-ocp"}}
{"Name":"crc","Status":"","Error":"Failed to get pull secret: open /User/anjan/Downloads/pull-secret-ocp: no such file or directory","ClusterConfig":{"KubeConfig":"","KubeAdminPass":"","ClusterAPI":"","WebConsoleURL":"","ProxyConfig":null},"KubeletStarted":false}
```

### DIFF
--- a/pkg/crc/api/api_test.go
+++ b/pkg/crc/api/api_test.go
@@ -36,6 +36,7 @@ func TestApi(t *testing.T) {
 	tt := []struct {
 		command       string
 		clientFailing bool
+		args          json.RawMessage
 		expected      map[string]interface{}
 	}{
 		{
@@ -74,6 +75,41 @@ func TestApi(t *testing.T) {
 				"Success":          false,
 			},
 		},
+		{
+			command:       "start",
+			clientFailing: true,
+			args:          json.RawMessage(`{"pull-secret":"/Users/fake/pull-secret"}`),
+			expected: map[string]interface{}{
+				"Name":           "crc",
+				"Status":         "",
+				"Error":          "Incorrect arguments given: json: unknown field \"pull-secret\"",
+				"KubeletStarted": false,
+				"ClusterConfig": map[string]interface{}{
+					"KubeConfig":    "",
+					"KubeAdminPass": "",
+					"ClusterAPI":    "",
+					"WebConsoleURL": "",
+					"ProxyConfig":   nil,
+				},
+			},
+		},
+		{
+			command: "start",
+			args:    json.RawMessage(`{"pullSecretFile":"/Users/fake/pull-secret"}`),
+			expected: map[string]interface{}{
+				"Name":           "crc",
+				"Status":         "",
+				"Error":          "",
+				"KubeletStarted": true,
+				"ClusterConfig": map[string]interface{}{
+					"KubeConfig":    "/tmp/kubeconfig",
+					"KubeAdminPass": "foobar",
+					"ClusterAPI":    "https://foo.testing:6443",
+					"WebConsoleURL": "https://console.foo.testing:6443",
+					"ProxyConfig":   nil,
+				},
+			},
+		},
 	}
 	for _, test := range tt {
 		client.Failing = test.clientFailing
@@ -82,6 +118,7 @@ func TestApi(t *testing.T) {
 
 		jsonReq, err := json.Marshal(commandRequest{
 			Command: test.command,
+			Args:    test.args,
 		})
 		assert.NoError(t, err)
 		_, err = client.Write(jsonReq)
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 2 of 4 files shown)


pkg/crc/api/handlers.go
```diff
@@ -30,15 +30,28 @@ func stopHandler(client machine.Client, crcConfig crcConfig.Storage, _ json.RawM
...
-		GetPullSecret: getPullSecretFileContent(crcConfig),
-		Debug:         true,
+		Name:       constants.DefaultName,
+		BundlePath: crcConfig.Get(config.Bundle).AsString(),
+		Memory:     crcConfig.Get(config.Memory).AsInt(),
+		CPUs:       crcConfig.Get(config.CPUs).AsInt(),
+		NameServer: crcConfig.Get(config.NameServer).AsString(),
+		Debug:      true,
+	}
+	if args != nil {
```
guillaumerose, line 42: An unit test for this would be good to have :)

pkg/crc/api/types.go
```diff
@@ -47,3 +47,8 @@ type getConfigResult struct {
 	Error   string
 	Configs map[string]interface{}
 }
+
+// startArgs is used to get the pull secret file path as argument for start handler
+type startArgs struct {
+	PullSecretFile string `json:"pullSecretFile"`
```
praveenkumar, line 53: So we are not accepting cpu/mem etc as part of startArgs as of now and either take from config or use defaults?

anjannath: Yeah, right now i only implemented to pass in a pull secret (because if its not set you can't start the cluster) the memory, cpu configs i think should be part of the separate config window, using the config api

pkg/crc/api/handlers.go
anjannath: Done!

pkg/crc/api/handlers.go
```diff
@@ -30,15 +30,28 @@ func stopHandler(client machine.Client, crcConfig crcConfig.Storage, _ json.RawM
...
+		a := startArgs{}
+		_ = json.Unmarshal(args, &a)
+		if a == (startArgs{}) {
+			startErr := &machine.StartResult{
+				Name:  constants.DefaultName,
+				Error: "Incorrect argument name. Should be: pullSecretFile",
```
cfergeau, line 48: This is conflating 2 distinct things imo
1. `pullSecretFile` is not set, while it's mandatory
2. an invalid argument was passed

When you add more arguments, this check for `invalid argument name` will no longer work,, but when you change it, we might forget to force the presence of `pullSecretFile`. If you want mark  `pullSecretFile` as mandatory, use `if a.PullSecretFile == ""`, if you want to check for invalid args, the code might need to be more involved and make use of https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields.
Apart from this, looks good to me.

pkg/crc/api/api_test.go
```diff
@@ -74,6 +75,40 @@ func TestApi(t *testing.T) {
...
+					"KubeConfig":    "",
+					"KubeAdminPass": "",
+					"ClusterAPI":    "",
+					"WebConsoleURL": "",
+					"ProxyConfig":   nil,
+				},
+			},
+		},
+		{
+			command: "start",
```
guillaumerose, line 96: If you implement this https://github.com/code-ready/crc/blob/master/pkg/crc/machine/fakemachine/client.go#L86 the test will be better.

pkg/crc/api/handlers.go
```diff
@@ -30,15 +30,28 @@ func stopHandler(client machine.Client, crcConfig crcConfig.Storage, _ json.RawM
...
+		BundlePath: crcConfig.Get(config.Bundle).AsString(),
+		Memory:     crcConfig.Get(config.Memory).AsInt(),
+		CPUs:       crcConfig.Get(config.CPUs).AsInt(),
+		NameServer: crcConfig.Get(config.NameServer).AsString(),
+		Debug:      true,
+	}
+	if args != nil {
+		a := startArgs{}
+		_ = json.Unmarshal(args, &a)
+		if a == (startArgs{}) {
```
guillaumerose, line 45: It's better to check `json.Unmarshal` error. You can write:

```
var parsedArguments startArgs
if err := json.Unmarshal(args, &a); err != nil {
  // do something
}
if parsedArguments.PullSecretFile == "" {
   // do something
}
```

pkg/crc/api/handlers.go
```diff
@@ -30,15 +30,28 @@ func stopHandler(client machine.Client, crcConfig crcConfig.Storage, _ json.RawM
...
+			startErr := &machine.StartResult{
+				Name:  constants.DefaultName,
+				Error: "Incorrect argument name. Should be: pullSecretFile",
+			}
+			return encodeStructToJSON(startErr)
+		}
+		startConfig.GetPullSecret = getPullSecretFileContent(a.PullSecretFile)
+	} else {
```
guillaumerose, line 53: You can extract a function that takes config + args and return a startConfig.

pkg/crc/api/handlers.go
anjannath: The error from unmarshal is only generated if its provided invalid json or not a pointer, which is not that useful. I'd still need to check if args is empty to bypass this unmarshalling part.

pkg/crc/api/handlers.go
anjannath: Thanks, modified the code, PTALA!

pkg/crc/api/api_test.go
anjannath: Done!

pkg/crc/api/handlers.go
anjannath: Done!

guillaumerose (approved):

pkg/crc/api/types.go
gbraad: I am not following why this needs to be an exception. The config could have a text input or file requestor to locate the pull secret? Or does the daemon else not understand this change? Not seeing why this needs to be a start argument.

gbraad:
Why does this need to be a start argument for the daemon process?

pkg/crc/api/types.go
anjannath: This is only for the startHandler, we want to avoid the manual step of having to configure the `pull-secret-file` via the cli before being able to start the cluster through the api.

pkg/crc/api/handlers.go
```diff
@@ -31,22 +32,45 @@ func stopHandler(client machine.Client, crcConfig crcConfig.Storage, _ json.RawM
...
+	}
+	if args.PullSecretFile == "" {
+		startConfig.PullSecret = &cluster.PullSecret{
+			Getter: getPullSecretFileContent(cfg.Get(config.PullSecretFile).AsString()),
+		}
+		return startConfig
+	}
+	startConfig.PullSecret = &cluster.PullSecret{
+		Getter: getPullSecretFileContent(args.PullSecretFile),
+	}
```
praveenkumar, line 70: In this block only the `getPullSecretFileContent` function parameter is changed, we could use something like following
```golang
pullSecretFile := args.PullSecretFile
if pullSecretFile == "" {
   pullSecretFile = cfg.Get(Config.PullSecretFile).AsString()
}
startConfig.PullSecret = &cluster.PullSecret{
		Getter: getPullSecretFileContent(pullSecretFile),
	}
return startConfig
```

anjannath: Done, Thanks!

praveenkumar (approved):

PR #1565, opened 2020-10-01
Use time limit instead of number of attempts for retry loops

guillaumerose:
Writing the maximum duration the program will wait is more precise than
using a number of attempts.

Example:
errors.RetryAfter(120, waitForPendingCsr, time.Second*5) is equivalent
of waiting 120 * (execution time of waitForPendingCsr + 5 seconds).
This is way more than 120 seconds, the duration expected to last when
this line was added.
Now, it will try as much as possible within 120 seconds and wait 5
seconds between each try.

### DIFF
--- a/pkg/crc/cluster/cert-renewal.go
+++ b/pkg/crc/cluster/cert-renewal.go
@@ -28,7 +28,7 @@ func waitForPendingCsrs(ocConfig oc.Config) error {
 		return nil
 	}
 
-	return errors.RetryAfter(120, waitForPendingCsr, time.Second*5)
+	return errors.RetryAfter(8*time.Minute, waitForPendingCsr, time.Second*5)
 }
 
 func RegenerateCertificates(sshRunner *ssh.Runner, ocConfig oc.Config) error {
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -25,7 +25,7 @@ func WaitForSSH(sshRunner *ssh.Runner) error {
 		return nil
 	}
 
-	return errors.RetryAfter(60, checkSSHConnectivity, time.Second)
+	return errors.RetryAfter(60*time.Second, checkSSHConnectivity, time.Second)
 }
 
 type CertExpiryState int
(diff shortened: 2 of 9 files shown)


pkg/crc/errors/multierror.go
```diff
@@ -60,17 +60,18 @@ func (r RetriableError) Error() string {
...
+// RetryAfter retries for a certain duration, after a delay
+func RetryAfter(limit time.Duration, callback func() error, d time.Duration) error {
 	m := MultiError{}
-	for i := 0; i < attempts; i++ {
-		if i > 0 {
-			logging.Debugf("retry loop: attempt %d out of %d", i, attempts)
-		}
+	timeLimit := time.Now().Add(limit)
+	attempt := 0
+	for time.Now().Before(timeLimit) {
```
cfergeau, line 68: This is a very theoritical concern, but would it be worth having `for time.Now().Before(timeLimit) && attempt < 2` so that we are guaranteed at least 2 tries, first one takes ages and times out, second one succeeds as soon as we try it because durin the "takes ages" part to of the first try, everything got setup as needed?

guillaumerose: Yes, it seems reasonable. I'm adding it.

cfergeau (approved):

pkg/crc/cluster/cert-renewal.go
```diff
@@ -28,7 +28,7 @@ func waitForPendingCsrs(ocConfig oc.Config) error {
 		return nil
 	}
 
-	return errors.RetryAfter(120, waitForPendingCsr, time.Second*5)
+	return errors.RetryAfter(120*time.Second, waitForPendingCsr, time.Second*5)
```
praveenkumar, line 31: @guillaumerose before this change this check used to run around 120*(waitforpendingProcess + 5) seconds but now it will be just 120 seconds and on slow system I am sure that is not going to enough and might fail so we should atleast consider to have it 5-8 mins if not 10-12 mins.

pkg/crc/cluster/cluster.go
```diff
@@ -263,7 +263,7 @@ func WaitforRequestHeaderClientCaFile(ocConfig oc.Config) error {
 		logging.Debugf("Found .data.requestheader-client-ca-file: %s", stdout)
 		return nil
 	}
-	return errors.RetryAfter(90, lookupRequestHeaderClientCa, 2*time.Second)
+	return errors.RetryAfter(90*time.Second, lookupRequestHeaderClientCa, 2*time.Second)
```
praveenkumar, line 266: Same with this.

pkg/crc/cluster/cert-renewal.go
cfergeau: If there was a problem and the CSR never comes, we cannot really expect the user to wait for 10-12 minutes before reporting the error.

guillaumerose: It's good to have this debate! I changed to 8min.

pkg/crc/cluster/cluster.go
guillaumerose: I changed to 8min.

PR #1575, opened 2020-10-06
preflight: Improve libvirt/qemu detection

cfergeau:
At the moment, we decide if libvirt/qemu need to be installed by
checking for the availability of the 'virsh' binary. On current fedoras,
it's possible to install only the `libvirt-client` package, and no
other libvirt/qemu packages. This will install `virsh`, but
libvirtd/qemu-system-x86_64 which crc need will be missing.

This commit extends the checkLibvirtInstalled preflight check to parse
`virsh capabilities` to look for an appropriate hypervisor (ie qemu
binary)

## Testing

Without this PR:
1. Run 'yum remove 'qemu*' 'libvirt*'"
2. Run 'yum install libvirt-client'
3. Run 'crc setup' -> it succeeds
4. Run 'crc start' -> libvirt failure to start the crc VM

After this PR:
1. Run 'yum remove 'qemu*' 'libvirt*'"
2. Run 'yum install libvirt-client'
3. Run 'crc setup' -> it installs the needed libvirt/qemu packages
4. Run 'crc start' -> success


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -98,13 +99,52 @@ func fixKvmEnabled() error {
 	return nil
 }
 
+type virCapabilitiesGuest struct {
```
guillaumerose, line 102: Perhaps use this library https://github.com/libvirt/libvirt-go-xml/blob/master/capabilities.go#L293

guillaumerose (approved):

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -98,13 +99,46 @@ func fixKvmEnabled() error {
...
+
+	logging.Debug("Checking 'virsh capabilities' for libvirtd/qemu availability")
+	caps, err := getLibvirtCapabilities()
+	if err != nil {
+		return err
+	}
+
+	foundHvm := false
+	for _, guest := range caps.Guests {
+		if guest.OSType == "hvm" && guest.Arch.Name == "x86_64" {
```
praveenkumar, line 132: This check will fail on a different Arch like ppc64 if someone build the crc on that platform, instead of hard coding `x86_64` shouldn't we use `runtime.GOARCH` ?

cfergeau: Yup, I was a bit worried about this. `runtime.GOARCH` returns `amd64` for `x86_64`, so we need to do some mapping from goarch to libvirt arch. When adding a new arch, it will be failing until we add the new mapping. I think what I can do is read `host.cpu.arch` from `virsh capabilities` and match it here.

praveenkumar: you mean `guest.Arch.Name === host.CPU.Arch` which means what ever the host arch should be guest one, that seems bit clean.

cfergeau: Done in the new revision

PR #1579, opened 2020-10-07
Fix unit test for startHandler of daemon api

anjannath:
A new field `ClusterCACert` was added to `ClusterConfig` struct
which caused the unit test to fail, added it to the expected cluster config

### DIFF
--- a/pkg/crc/api/api_test.go
+++ b/pkg/crc/api/api_test.go
@@ -90,6 +90,7 @@ func TestApi(t *testing.T) {
 					"ClusterAPI":    "",
 					"WebConsoleURL": "",
 					"ProxyConfig":   nil,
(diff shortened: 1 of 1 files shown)


pkg/crc/api/api_test.go
```diff
@@ -102,6 +103,7 @@ func TestApi(t *testing.T) {
 				"Error":          "",
 				"KubeletStarted": true,
 				"ClusterConfig": map[string]interface{}{
+					"ClusterCACert": "MIIDODCCAiCgAwIBAgIIRVfCKNUa1wIwDQYJ",
```
praveenkumar, line 106: :+1

praveenkumar (approved):

PR #1586, opened 2020-10-09
Restart openshift-apiserver only if aggregator-client-ca is expired

guillaumerose:
This cert is eventually renewed at startup. If it is renewed, we need to
wait for it and restart openshift-apiserver. Since this cert is located
both in a configmap and on the disk, we prefer to read it from the disk:
it is faster to access and more reliable.

If the cert is not expired, openshift-apiserver will correctly start.
`crc start` will take less time when the bundle is fresh (< 30 days).

follow up of https://github.com/code-ready/crc/pull/1532

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -37,6 +37,8 @@ const (
 	kubeletClientCert = "/var/lib/kubelet/pki/kubelet-client-current.pem"
 
 	kubeletClientSignerName = "kubernetes.io/kube-apiserver-client-kubelet"
+
+	aggregatorClientCert = "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt"
 )
 
 func CheckCertsValidity(sshRunner *ssh.Runner) (bool, bool, error) {
@@ -65,6 +67,10 @@ func checkCertValidity(sshRunner *ssh.Runner, cert string) (bool, error) {
 		return true, nil
 	}
 	return false, nil
+}
+
+func CheckAggregatorClientCAValidity(sshRunner *ssh.Runner) (bool, error) {
+	return checkCertValidity(sshRunner, aggregatorClientCert)
 }
 
 // Return size of disk, used space in bytes and the mountpoint
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -41,16 +41,20 @@ const (
 	aggregatorClientCert = "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt"
 )
 
-func CheckCertsValidity(sshRunner *ssh.Runner) (bool, bool, error) {
+func CheckCertsValidity(sshRunner *ssh.Runner) (bool, bool, bool, error) {
 	client, err := checkCertValidity(sshRunner, kubeletClientCert)
 	if err != nil {
-		return false, false, err
+		return false, false, false, err
 	}
 	server, err := checkCertValidity(sshRunner, kubeletServerCert)
 	if err != nil {
-		return false, false, err
+		return false, false, false, err
 	}
-	return client, server, nil
+	aggregatorClient, err := checkCertValidity(sshRunner, aggregatorClientCert)
+	if err != nil {
+		return false, false, false, err
+	}
+	return client, server, aggregatorClient, nil
 }
 
 func checkCertValidity(sshRunner *ssh.Runner, cert string) (bool, error) {
@@ -67,10 +71,6 @@ func checkCertValidity(sshRunner *ssh.Runner, cert string) (bool, error) {
 		return true, nil
 	}
 	return false, nil
-}
-
-func CheckAggregatorClientCAValidity(sshRunner *ssh.Runner) (bool, error) {
-	return checkCertValidity(sshRunner, aggregatorClientCert)
 }
 
 // Return size of disk, used space in bytes and the mountpoint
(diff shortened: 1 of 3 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -67,6 +67,28 @@ func getcertExpiryDateFromVM(sshRunner *ssh.Runner) (time.Time, error) {
...
+	certExpiryDate, err := getAggregatorClientCAExpiryDate(sshRunner)
+	if err != nil {
+		return false, err
+	}
+	return time.Now().After(certExpiryDate), nil
+}
+
+func getAggregatorClientCAExpiryDate(sshRunner *ssh.Runner) (time.Time, error) {
+	certExpiryDate := time.Time{}
+	certExpiryDateCmd := `date --date="$(sudo openssl x509 -in /etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt -noout -enddate | cut -d= -f 2)" --iso-8601=seconds`
```
cfergeau, line 80: Feel like we should have a helper which takes a `sshRunner` and a `path` as arguments and returns a cert expiry date (or a boolean indicating if the cert is expired).

pkg/crc/cluster/cluster.go
```diff
@@ -291,23 +313,15 @@ func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret *Pu
 	return sshRunner.CopyData([]byte(content), vmPullSecretPath, 0600)
 }
 
-func WaitforRequestHeaderClientCaFile(ocConfig oc.Config) error {
-	if err := WaitForOpenshiftResource(ocConfig, "configmaps"); err != nil {
-		return err
-	}
-
+func WaitforRequestHeaderClientCaFile(sshRunner *ssh.Runner) error {
```
cfergeau, line 316: `WaitFor...` while you change this.

pkg/crc/cluster/cluster.go
```diff
@@ -291,23 +313,15 @@ func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret *Pu
...
-		stdout, stderr, err := ocConfig.RunOcCommand(cmdArgs...)
+		expired, err := CheckAggregatorClientCAValidity(sshRunner)
 		if err != nil {
-			return fmt.Errorf("Failed to get request header client ca file %v: %s", err, stderr)
+			return fmt.Errorf("Failed to expiry date: %v", err)
 		}
-		if stdout == "" {
-			return &errors.RetriableError{Err: fmt.Errorf("missing .data.requestheader-client-ca-file")}
+		if expired {
+			return &errors.RetriableError{Err: fmt.Errorf("certificate still expired")}
```
cfergeau, line 323: Were you able to test this part of the PR? This PR is again doing 2 things at once, it checks certificate validity before doing the `waitForRequestHeaderClientCaFile`/`deletePods` dance, which is good for startup speed. But it also changes the implementation of `waitForRequestHeaderClientCaFile`, and this is not immediately obvious from the commit log. Imo we need to be careful with this part of the change as this workaround has been troublesome in the past. Maybe split it in a second commit?

guillaumerose: I tested it with an old bundle (4.5.4 from August) and the last one (4.5.14). I will split it 2 commits!

pkg/crc/cluster/cluster.go
guillaumerose: Yes I'm waiting for https://github.com/code-ready/crc/pull/1532 to get this in a better shape. I opened it collect feedback on "is it a good thing or not to do?"

guillaumerose: done

pkg/crc/cluster/cluster.go
guillaumerose: done

pkg/crc/cluster/cluster.go
```diff
@@ -291,23 +313,15 @@ func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret *Pu
...
+func WaitforRequestHeaderClientCaFile(sshRunner *ssh.Runner) error {
 	lookupRequestHeaderClientCa := func() error {
-		cmdArgs := []string{"get", "configmaps/extension-apiserver-authentication", `-ojsonpath={.data.requestheader-client-ca-file}`,
-			"-n", "kube-system"}
-
-		stdout, stderr, err := ocConfig.RunOcCommand(cmdArgs...)
+		expired, err := CheckAggregatorClientCAValidity(sshRunner)
 		if err != nil {
-			return fmt.Errorf("Failed to get request header client ca file %v: %s", err, stderr)
+			return fmt.Errorf("Failed to expiry date: %v", err)
```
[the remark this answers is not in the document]
guillaumerose: done

guillaumerose:
Any thoughts on this @praveenkumar ? Should we get that in ?

pkg/crc/cluster/cluster.go
```diff
@@ -41,16 +41,20 @@ const (
 	aggregatorClientCert = "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt"
 )
 
-func CheckCertsValidity(sshRunner *ssh.Runner) (bool, bool, error) {
+func CheckCertsValidity(sshRunner *ssh.Runner) (bool, bool, bool, error) {
```
cfergeau, line 44: Time for some API doc comments, or to put these 3 bools in a struct ;)

guillaumerose: I changed it to a map, it is better! :)

Issue #1438, opened 2020-08-05
Duplicate error message on errors


cfergeau:
70ed4aec introduced a regression in some cases, for example when the VM is stopped and one tries `crc stop`

Before:
```
$ ./out/linux-amd64/crc stop
Stopping the OpenShift cluster, this may take a few minutes...
ERRO Cannot stop machine: Machine "crc" is already stopped.
$
```

After:
```
$ ./out/linux-amd64/crc stop
Stopping the OpenShift cluster, this may take a few minutes...
ERRO Cannot stop machine: Machine "crc" is already stopped.
Cannot stop machine: Machine "crc" is already stopped.
$
```

PR #1587, opened 2020-10-09
Simplify machine Client API, decouple it from the daemon

guillaumerose:
Be having the same type for the daemon API and the machine package, we were forced to return errors in 2 formats: one in the returned object (Success: false, Error: err) and in the second returned object (error).

This PR cleans this and pushes this "complexity" in the daemon package. Benefits are:
* no more Success and Error field where we can check the second returned object (of error type)
* Being able to adjust what we log when a failure occurs: for cmd, let the cmd do something with the error. for the daemon api, log it and send it as json.
* Far less types and lines in machine package.
* Use type as far as possible (for instance, state can be represented as an enum in one more layer and not as a string)
* It will be easier to implement Client interface (esp. when composing Client in a Client, error handling will be easier).

The inconvenient is that daemon API and output json for cmd are still not unified.

Fix https://github.com/code-ready/crc/issues/1438

---

I wanted to this a long time ago. Now the machine package looks like traditional Go. I wanted to write a new implementation of Client this morning that would use a sub-Client, it worked well until I had to deal with errors: lot of lines with little value.

Example:
```
type ProtectedClient struct {
	Underlying Client
	sem        *semaphore.Weighted
}

// Before
func (p *ProtectedClient) Delete(deleteConfig DeleteConfig) (DeleteResult, error) {
	if p.sem.TryAcquire(1) {
		defer p.sem.Release(1)
		return p.Underlying.Delete(deleteConfig)
	}
	return DeleteResult{
		Name:    deleteConfig.Name,
		Success: false,
		Error:   "cannot delete while an operation is still in progress",
	}, errors.New("cannot delete while an operation is still in progress")
}

// After
func (p *ProtectedClient) Delete(name string) error {
	if p.sem.TryAcquire(1) {
		defer p.sem.Release(1)
		return p.Underlying.Delete(name)
	}
	return errors.New("cannot delete while an operation is still in progress")
}
```


cmd/crc/cmd/status.go
```diff
@@ -60,7 +60,7 @@ func runStatus(writer io.Writer, client machine.Client, cacheDir, outputFormat s
 		return fmt.Errorf("Error finding size of cache: %s", err.Error())
 	}
 	status := &status{
-		CrcStatus:        clusterStatus.CrcStatus,
+		CrcStatus:        clusterStatus.CrcStatus.String(),
```
cfergeau, line 63: rather than "an enum" in the commit log, could be "Use the `state.State` enum ...`

pkg/crc/machine/errors.go
```diff
@@ -1,27 +1,17 @@
...
-	"fmt"
-
-	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/pkg/errors"
 )
 
 func startError(description string, err error) (*StartResult, error) {
-	fullErr := logErrorf("%s: %v", description, err)
-	return nil, fullErr
+	return nil, errors.Wrap(err, description)
```
cfergeau, line 8: At this point, these helpers can be removed.

cfergeau (commented):
Overall looks good, some comments.
I'd add the `DeleteResult is only used in the daemon API. Having only an err is fine
for other usage.` wording (or something similar) to most of these `Simplify machine.Client xxx` commits.

pkg/crc/api/handlers.go
```diff
@@ -17,13 +17,13 @@ import (
 	"github.com/code-ready/crc/pkg/crc/version"
 )
 
-func statusHandler(client machine.Client, crcConfig crcConfig.Storage, _ json.RawMessage) string {
+func statusHandler(client AdaptedClient, crcConfig crcConfig.Storage, _ json.RawMessage) string {
```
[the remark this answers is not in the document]
guillaumerose: yes, I just merged conflicts. Got luck, it was not so hard!

pkg/crc/machine/machine.go
```diff
@@ -134,7 +134,7 @@ func (client *client) updateVMConfig(startConfig StartConfig, api libmachine.API
 func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 	var crcBundleMetadata *bundle.CrcBundleInfo
 
-	libMachineAPIClient, cleanup, err := createLibMachineClient(startConfig.Debug)
+	libMachineAPIClient, cleanup, err := createLibMachineClient(client.debug)
 	defer cleanup()
```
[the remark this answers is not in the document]
guillaumerose: very good point! Added.

cmd/crc/cmd/console.go
```diff
@@ -40,18 +40,13 @@ var consoleCmd = &cobra.Command{
 }
 
 func showConsole(client machine.Client) (*machine.ConsoleResult, error) {
-	consoleConfig := machine.ConsoleConfig{
-		Name: constants.DefaultName,
-	}
-
-	if err := checkIfMachineMissing(client, consoleConfig.Name); err != nil {
+	if err := checkIfMachineMissing(client, constants.DefaultName); err != nil {
```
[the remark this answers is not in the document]
guillaumerose: yes done

cmd/crc/cmd/status.go
guillaumerose: :+1:

pkg/crc/machine/errors.go
guillaumerose: I was tempted but it forces us to add an extra explanation for startup error. It's easy to see in Start() if an error is not using `startError`. 

I removed the helper for console and status. Errors were not bringing much info.

pkg/crc/api/adapter.go
```diff
@@ -1,16 +1,18 @@
...
 	GetConsoleURL(consoleConfig machine.ConsoleConfig) (machine.ConsoleResult, error)
 	Start(startConfig machine.StartConfig) (machine.StartResult, error)
 	Status(statusConfig machine.ClusterStatusConfig) (machine.ClusterStatusResult, error)
-	Stop(stopConfig machine.StopConfig) (machine.StopResult, error)
+	Stop(stopConfig machine.StopConfig) Result
 }
 
-type DeleteResult struct {
+type Result struct {
 	Name    string
```
cfergeau, line 19: This could have been named `Result` from the start, rather than being first `DeleteResult` and then `Result`

pkg/crc/machine/machine.go
```diff
@@ -513,16 +513,16 @@ func (*client) IP(ipConfig IPConfig) (string, error) {
 	return ip, nil
 }
 
-func (*client) Status(statusConfig ClusterStatusConfig) (ClusterStatusResult, error) {
+func (*client) Status(statusConfig ClusterStatusConfig) (*ClusterStatusResult, error) {
 	libMachineAPIClient, cleanup, err := createLibMachineClient(false)
 	defer cleanup()
 	if err != nil {
-		return statusError(statusConfig.Name, "Cannot initialize libmachine", err)
+		return statusError("Cannot initialize libmachine", err)
```
cfergeau: This would feel more 'natural' as `return nil, statusError(...)`

pkg/crc/machine/machine.go
```diff
@@ -517,12 +517,12 @@ func (client *client) Status() (*ClusterStatusResult, error) {
 	libMachineAPIClient, cleanup, err := createLibMachineClient(false)
 	defer cleanup()
 	if err != nil {
-		return statusError("Cannot initialize libmachine", err)
+		return nil, err
```
cfergeau, line 520: This does not feel very consistent, `Simplify machine.Client Stop` directly removed `stopError` and inlined it instead:
`return state.None, errors.Wrap(err, "Cannot initialize logging")`

With this commit, why do we remove the `error.Wrap()` from console/status/...? I'd just keep it as this makes it easier to know where in the code the error comes from. And why do we treat `startError()` differently from `stopError()`?

cfergeau (commented):
Some smallish comments left

pkg/crc/api/adapter.go
guillaumerose: done

pkg/crc/machine/machine.go
guillaumerose: I removed all statusError/etc functions.

pkg/crc/machine/machine.go
guillaumerose: I reworked this. It should be correct now. I removed the function error function at the same type I touch the related machine function.

cfergeau (approved):

guillaumerose:
Passed on one CI. Merging it.

PR #1588, opened 2020-10-09
Use a config that discards all preflights for API tests

guillaumerose:
It allows us to test the real implementation of the start handler.
If a check doesn't have a key, it cannot be skipped. Adding missing
ones.

Follow up for #1558

### DIFF
--- a/pkg/crc/api/api_test.go
+++ b/pkg/crc/api/api_test.go
@@ -4,48 +4,21 @@ package api
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


pkg/crc/api/api_test.go
```diff
@@ -57,7 +30,9 @@ func TestApi(t *testing.T) {
 	require.NoError(t, err)
 
 	client := fakemachine.NewClient()
-	api, err := createAPIServerWithListener(listener, setupNewInMemoryConfig, newFakeHandler(client))
+	api, err := createAPIServerWithListener(listener, setupNewInMemoryConfig, &Handler{
+		MachineClient: client,
+	})
```
praveenkumar, line 35: :+1:

praveenkumar (approved):

PR #1597, opened 2020-10-14
cluster: Use k8s/openshift API types

cfergeau:
We currently use a generic k8sResource type to parse kubernetes
resources, which has become a bit weird as it has fields specific to
CSRs, and fields specific to ClusterOperators.

Since we already have a dependency on the k8s code, we can make use of
native k8s/openshift types rather than hand-writing our own struct.
This 'only' grows the vendor directory by ~600kB

## Testing

This is only an internal code change, there should be 0 change in crc behaviour.


pkg/crc/cluster/csr.go
```diff
@@ -41,12 +42,16 @@ func approveNodeCSR(ocConfig oc.Config, signerName string) error {
 		if len(csr.Status.Conditions) != 0 {
 			continue
 		}
-		if signerName != csr.Spec.SignerName {
-			logging.Debugf("Unexpected unapproved csr %s (signerName: %s)", csr.Metadata.Name, csr.Spec.SignerName)
+		var signerName string
+		if csr.Spec.SignerName != nil {
+			signerName = *csr.Spec.SignerName
+		}
+		if signerName != *csr.Spec.SignerName {
```
cfergeau, line 49: ouch, this is buggy, I overrode the 'signerName' variable by mistake

pkg/crc/cluster/csr.go
```diff
@@ -41,12 +42,16 @@ func approveNodeCSR(ocConfig oc.Config, signerName string) error {
 		if len(csr.Status.Conditions) != 0 {
 			continue
 		}
-		if signerName != csr.Spec.SignerName {
-			logging.Debugf("Unexpected unapproved csr %s (signerName: %s)", csr.Metadata.Name, csr.Spec.SignerName)
+		var signerName string
+		if csr.Spec.SignerName != nil {
+			signerName = *csr.Spec.SignerName
```
praveenkumar, line 47: @cfergeau you mean here?

cfergeau: yep, in that block. I thought I had pushed a fix for this already

guillaumerose (approved):

PR #1607, opened 2020-10-20
Disk resize

cfergeau:
**Fixes:** Issue #127

## Solution/Idea

This uses the machine driver's "UpdateConfigRaw" API introduced in the previous release in order to implement disk resizing support. This needs machine driver support as well.

## Proposed changes

1. Added --disk-size argument to the `start` command.
2. Added 'disk-size' config key

disk size is always in GiB (1024*1024*1024 bytes) and can't be less than 31 GiB as this is the size used by our bundles, and disks can't be shrunk.

## Testing

1. `start` without any argument/special configuration succeeds
2. `crc start --disk-size xxx` succeeds and resizes the disk
3. `crc start` succeeds after `crc config set disk-size xxx`
4. in both cases, disk is resized (can be checked with `ssh ... df -h`, or on linux through `virsh vol-info crc.qcow2 crc`)

For now this is not supported on macos
7. ...

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -13,6 +13,7 @@ const (
 	Bundle               = "bundle"
 	CPUs                 = "cpus"
 	Memory               = "memory"
+	DiskSize             = "disk-size"
 	NameServer           = "nameserver"
 	PullSecretFile       = "pull-secret-file"
 	DisableUpdateCheck   = "disable-update-check"
@@ -28,6 +29,7 @@ func RegisterSettings(cfg *config.Config) {
 	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundle, config.SuccessfullyApplied)
 	cfg.AddSetting(CPUs, constants.DefaultCPUs, config.ValidateCPUs, config.RequiresRestartMsg)
 	cfg.AddSetting(Memory, constants.DefaultMemory, config.ValidateMemory, config.RequiresRestartMsg)
+	cfg.AddSetting(DiskSize, constants.DefaultDiskSize, config.ValidateDiskSize, config.RequiresRestartMsg)
 	cfg.AddSetting(NameServer, "", config.ValidateIPAddress, config.SuccessfullyApplied)
 	cfg.AddSetting(PullSecretFile, "", config.ValidatePath, config.SuccessfullyApplied)
 	cfg.AddSetting(DisableUpdateCheck, false, config.ValidateBool, config.SuccessfullyApplied)
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -31,6 +31,7 @@ func init() {
 	flagSet.StringP(cmdConfig.PullSecretFile, "p", "", fmt.Sprintf("File path of image pull secret (download from %s)", constants.CrcLandingPageURL))
 	flagSet.IntP(cmdConfig.CPUs, "c", constants.DefaultCPUs, "Number of CPU cores to allocate to the OpenShift cluster")
 	flagSet.IntP(cmdConfig.Memory, "m", constants.DefaultMemory, "MiB of memory to allocate to the OpenShift cluster")
+	flagSet.UintP(cmdConfig.DiskSize, "d", constants.DefaultDiskSize, "Total size in GiB of the disk used by the OpenShift cluster")
 	flagSet.StringP(cmdConfig.NameServer, "n", "", "IPv4 address of nameserver to use for the OpenShift cluster")
 	flagSet.Bool(cmdConfig.DisableUpdateCheck, false, "Don't check for update")
 
@@ -65,6 +66,7 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 	startConfig := machine.StartConfig{
 		BundlePath: config.Get(cmdConfig.Bundle).AsString(),
 		Memory:     config.Get(cmdConfig.Memory).AsInt(),
+		DiskSize:   config.Get(cmdConfig.DiskSize).AsInt(),
 		CPUs:       config.Get(cmdConfig.CPUs).AsInt(),
 		NameServer: config.Get(cmdConfig.NameServer).AsString(),
 		PullSecret: &cluster.PullSecret{
@@ -159,6 +161,9 @@ func validateStartFlags() error {
 		return err
 	}
 	if err := validation.ValidateCPUs(config.Get(cmdConfig.CPUs).AsInt()); err != nil {
+		return err
+	}
+	if err := validation.ValidateDiskSize(config.Get(cmdConfig.DiskSize).AsInt()); err != nil {
 		return err
 	}
 	if err := validation.ValidateBundle(config.Get(cmdConfig.Bundle).AsString()); err != nil {
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/cavaliercoder/grab v2.0.0+incompatible
 	github.com/cheggaaa/pb/v3 v3.0.4
 	github.com/code-ready/clicumber v0.0.0-20200728062640-1203dda97f67
-	github.com/code-ready/machine v0.0.0-20201008083613-1420a5d7f1e2
+	github.com/code-ready/machine v0.0.0-20201022100236-4405fef4d0cc
 	github.com/cucumber/godog v0.9.0
 	github.com/cucumber/messages-go/v10 v10.0.3
 	github.com/docker/docker v1.13.1 // indirect
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -128,6 +128,19 @@ func (client *client) updateVMConfig(startConfig StartConfig, api libmachine.API
 		return err
 	}
 
+	/* Disk size */
+	if startConfig.DiskSize != constants.DefaultDiskSize {
+		if err := setDiskSize(host, startConfig.DiskSize); err != nil {
+			logging.Debugf("Failed to update CRC disk configuration: %v", err)
+			if err != drivers.ErrNotImplemented {
+				return err
+			}
+		}
+		if err := api.Save(host); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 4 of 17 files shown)


pkg/crc/machine/machine.go
```diff
@@ -262,6 +275,11 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 		return nil, errors.Wrap(err, "Error updating public key")
 	}
 
+	// Trigger disk resize, this will be a no-op if no disk size change is needed
+	if _, err = sshRunner.Run("sudo xfs_growfs / >/dev/null"); err != nil {
```
guillaumerose, line 279: Does it work with OKD?

pkg/crc/machine/machine.go
```diff
@@ -128,6 +128,19 @@ func (client *client) updateVMConfig(startConfig StartConfig, api libmachine.API
 		return err
 	}
 
+	/* Disk size */
+	if startConfig.DiskSize != 0 {
+		if err := setDiskSize(host, startConfig.DiskSize); err != nil {
+			logging.Debugf("Failed to update CRC disk configuration: %v", err)
+			if err != drivers.ErrNotImplemented {
+				return err
```
guillaumerose, line 136: I guess this is where macOS user will currently fall ?

cmd/crc/cmd/start.go
```diff
@@ -31,6 +31,7 @@ func init() {
 	flagSet.StringP(cmdConfig.PullSecretFile, "p", "", fmt.Sprintf("File path of image pull secret (download from %s)", constants.CrcLandingPageURL))
 	flagSet.IntP(cmdConfig.CPUs, "c", constants.DefaultCPUs, "Number of CPU cores to allocate to the OpenShift cluster")
 	flagSet.IntP(cmdConfig.Memory, "m", constants.DefaultMemory, "MiB of memory to allocate to the OpenShift cluster")
+	flagSet.UintP(cmdConfig.DiskSize, "d", 0, "Total size in GiB of the disk used by the CodeReady Container instance")
```
gbraad, line 34: CodeReady Container instance or the OpenShift c;uster? 

Just like memory, disk is NOT assigned to the cluster, but we say so in the help.

pkg/crc/validation/validation.go
```diff
@@ -31,6 +31,17 @@ func ValidateMemory(value int) error {
 	return ValidateEnoughMemory(value)
 }
 
+func ValidateDiskSize(value int) error {
+	/* If no disk size was set on the command line or in the preferences, we'll validate '0' */
+	if value == 0 {
+		return nil
+	}
+	if value < constants.DefaultDiskSize {
```
gbraad, line 39: :+1

go.mod
```diff
@@ -75,4 +75,7 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.18.2
 )
 
-replace vbom.ml/util => github.com/fvbommel/util v0.0.2
+replace (
+	github.com/code-ready/machine => ../machine
```
praveenkumar, line 79: I think this need to be removed once you have changes in machine side.

pkg/crc/machine/machine.go
praveenkumar: @guillaumerose yes it should work with OKD also.

pkg/crc/machine/machine.go
praveenkumar: Same my concern, `startConfig.DiskSize` will have the default DiskSize, shouldn't we compare `startConfig.DiskSize != constants.DefaultDiskSize` ?

gbraad (approved):

gbraad:
@cfergeau can you describe clearly for @kowen what is supported and what not? This needs to be part of the release note,s. But let's also consider to actually document this. Can this be a follow PR?

cfergeau:
> @cfergeau can you describe clearly for @kowen what is supported and what not? This needs to be part of the release note,s. But let's also consider to actually document this. Can this be a follow PR?

The `machine` PR https://github.com/code-ready/machine/pull/48 adds support for disk resize on Windows (though I haven't retested it recently). The `machine-driver-libvirt` PR adds support on linux https://github.com/code-ready/machine-driver-libvirt/pull/68
hyperkit won't be supported for now, the hyperkit machine driver needs to be updated to use the latest version of https://github.com/moby/hyperkit/tree/master/go to get resizing support https://github.com/moby/hyperkit/commit/7e0fa508edd58db99f011f14a59c09ba1fd4f4a2

This PR description https://github.com/code-ready/crc/pull/1607#issue-506926874 has a user-level explanation regarding how to use it. Disk size is specified in GiB, though this can easily be changed to GB if more user friendly.

pkg/crc/machine/machine.go
cfergeau: This is checking for `UpdateConfigRaw` failures, at the moment the hyperkit driver only ignores disk sizes, so it won't return an error, it needs an update in order to return an error. This is one part where `UpdateConfigRaw` could be improved, return an error if it was called but did not  modify any fields.

cmd/crc/cmd/start.go
cfergeau: ah, it's indeed inconsistent, I'll fix.

pkg/crc/machine/machine.go
cfergeau: > Same my concern, startConfig.DiskSize will have the default DiskSize, shouldn't we compare startConfig.DiskSize != constants.DefaultDiskSize ?

I don't understand what you are asking :( However, I did not want the 'disk-size' config to default to `constants.DefaultDiskSize` as this would cause unexpected behaviour when doing
`crc start --disk-size 50; crc stop; crc start`.
It would try to resize the disk back to `constants.DefaultDiskSize`, but the libvirt driver errors out when trying to do that. So it's better to keep the old size, and `0` is used as a means to indicate "no value was set"

cfergeau: > This is one part where UpdateConfigRaw could be improved, return an error if it was called but did not modify any fields.

The `machine-driver-hyperkit` implementation may be something like this: (untested)
```
From c83e1a0c18f5df18940e686dd51737bd93fe67d1 Mon Sep 17 00:00:00 2001
From: Christophe Fergeau <cfergeau@redhat.com>
Date: Wed, 21 Oct 2020 20:36:26 +0200
Subject: [PATCH] Report an error when trying to change unsupported field

Only memory size and number of CPUS should be changed with
`UpdateConfigRaw`. If neither of these changed, the call
to UpdateConfigRaw is unsupported.
---
 pkg/hyperkit/driver.go | 7 +++++++
 1 file changed, 7 insertions(+)

diff --git a/pkg/hyperkit/driver.go b/pkg/hyperkit/driver.go
index 084b54e..c8bbd90 100644
--- a/pkg/hyperkit/driver.go
+++ b/pkg/hyperkit/driver.go
@@ -416,6 +416,13 @@ func (d *Driver) UpdateConfigRaw(rawConfig []byte) error {
 		return err
 	}
 
+	if newDriver.Memory == d.Memory && newDriver.CPU == d.CPU {
+		/* For now only changing memory and CPU is supported/tested.
+		 * If none of these changed, we might be trying to change another
+		 * value, which is may or may not work, return ErrNotImplemented for now
+		 */
+		return drivers.ErrNotImplemented
+	}
 	*d = newDriver
 
 	return nil
-- 
2.28.0
```

guillaumerose: I tried on macOS and the flag is just discarded. I guess it's fine for a first release.

cfergeau: https://github.com/code-ready/machine-driver-hyperkit/pull/21 should improve this, together with the commit I added to this PR to show a warning when `ErrNotImplemented` is returned.

guillaumerose: code-ready/machine-driver-hyperkit#21 is not for disk size. Should we add it ?

cmd/crc/cmd/config/config.go
```diff
@@ -28,6 +29,7 @@ func RegisterSettings(cfg *config.Config) {
 	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundle, config.SuccessfullyApplied)
 	cfg.AddSetting(CPUs, constants.DefaultCPUs, config.ValidateCPUs, config.RequiresRestartMsg)
 	cfg.AddSetting(Memory, constants.DefaultMemory, config.ValidateMemory, config.RequiresRestartMsg)
+	cfg.AddSetting(DiskSize, 0, config.ValidateDiskSize, config.RequiresRestartMsg)
```
guillaumerose, line 32: Why 0 and not DefaultDiskSize ?

pkg/crc/validation/validation.go
```diff
@@ -31,6 +31,17 @@ func ValidateMemory(value int) error {
 	return ValidateEnoughMemory(value)
 }
 
+func ValidateDiskSize(value int) error {
+	/* If no disk size was set on the command line or in the preferences, we'll validate '0' */
+	if value == 0 {
```
guillaumerose, line 36: I guess this check is related to 0 being the default value.

cmd/crc/cmd/config/config.go
cfergeau: Initial concern was to be able to differentiate between "no value was specified on the command line" and "user specified a value which happens to be the default one". Without this, initial iterations of the code would end up calling `if err := setDiskSize(host, cfg.DefaultDiskSize)`, which would trigger an error if the disk had been resized before.
However, now the `setDiskSize` call is wrapped in a `if diskSize != 0` test. If we change this test to `if diskSize != cfg.DefaultDiskSize` then everything should work. I'll test this now.

praveenkumar: @cfergeau doesn't our validation run initially and error out if the provided disk size is not greater than the default one and we only run the disk resize part when the disk size is provided greater than default?

cfergeau: The "only run the disk resize part when the disk size is provided greater than default" is the part which is not done yet, we only have `if startConfig.diskSize != 0`, I think this could be replaced with `if startConfig.diskSize != constants.DefaultDiskSize`, but this needs testing

cfergeau: I've force-pushed an updated series, only difference is a `fixup` commit which replaces the `0` default value with a `constants.DefaultDiskSize` default value, we can go with this one I think, I'll squash it.

guillaumerose (approved):

Issue #1613, opened 2020-10-22
[BUG] Replace generic term 'binary' in path


deboer-tim:
### Steps to reproduce

At https://cloud.redhat.com/openshift/install/crc/installer-provisioned the text reads:
`Download and extract the CodeReady Containers archive for your operating system and place the binary in your $PATH .`

The term binary is generic and I found it confusing the first time since the archive is also a binary. We should switch to a better term, I'd suggest 'executable':
`Download and extract the CodeReady Containers archive for your operating system and place the executable in your $PATH.`

cfergeau:
The documentation changes have been merged, however these probably deserves to be changed too:
```
$ git grep -i "\".*binary.*\""
cmd/crc-embedder/cmd/embed.go:	Short: "Embed data files in crc binary",
cmd/crc-embedder/cmd/extract.go:	Short: "Extract data file embedded in the crc binary",
cmd/crc-embedder/cmd/list.go:	Short: "List data files embedded in the crc binary",
cmd/crc-embedder/cmd/root.go:	Short: "Build helper for crc for binary embedding",
cmd/crc/cmd/oc_env.go:	Short: "Add the 'oc' binary to PATH",
cmd/crc/cmd/version.go:		fmt.Sprintf("OpenShift version: %s (%sembedded in binary)\n", v.OpenshiftVersion, embedded),
pkg/crc/cache/cache.go:			logging.Warnf("Binary name is %s but extracted file name is %s", c.binaryName, filepath.Base(assetTmpFile))
pkg/crc/cache/cache.go:	logging.Debugf("Trying to extract %s from crc binary", c.binaryName)
pkg/crc/machine/client/client_darwin.go:	"github.com/code-ready/machine/libmachine/drivers/plugin/localbinary"
pkg/crc/machine/client/client_linux.go:	"github.com/code-ready/machine/libmachine/drivers/plugin/localbinary"
pkg/crc/machine/client/client_windows.go:	"github.com/code-ready/machine/libmachine/drivers/plugin/localbinary"
pkg/crc/preflight/preflight_checks_common.go:		checkDescription: "Checking if oc binary is cached",
pkg/crc/preflight/preflight_checks_common.go:		fixDescription:   "Caching oc binary",
pkg/crc/preflight/preflight_checks_common.go:		checkDescription: "Checking if podman remote binary is cached",
pkg/crc/preflight/preflight_checks_common.go:		fixDescription:   "Caching podman remote binary",
[...]

PR #1614, opened 2020-10-22
Issue #1613 Docs: Use "executable" instead of "binary"

robin-owen:
**Fixes:** Issue #1613

## Solution/Idea

"Executable" is more specific and conveys more useful information.


cfergeau (approved):
\lgtm
There are quite a few places in the codebase where this could be changed too.

PR #1616, opened 2020-10-26
Replace use of 'binary' terminology with 'executable'

cfergeau:
Fixes: Issue #1613

## Solution/Idea

This replaces the 'binary' terminology when talking about an executable file with 'executable'. As pointed out in #1613, 'binary' can be ambiguous.

## Proposed changes

1. There should only be cosmetic changes to user visible strings ('binary' -> 'executable'), with no change in behaviour.


anjannath:
retest this please
/retest

guillaumerose (approved):

PR #1622, opened 2020-10-29
Use value instead castValue for error message

praveenkumar:
invalidprop variable is used to create the error message and it respect
the strings instead of casted value. if casted value is used then following
msg shown to user which is not a good ux.

```
$ crc config set cpus 2
Value '%!!(MISSING)s(int=2)' for configuration property 'cpus' is invalid, reason: requires CPUs >= 4
```


cfergeau (approved):
I think 'respects the string' -> 'expects a string'
Apart from that, ack.

PR #1625, opened 2020-10-30
Minor fixes for disk size

guillaumerose:


cfergeau (approved):
` hyperv: Don't try to convert DiskCapacity from GiB to bytes ` is missing a `go mod vendor`.
Thanks for fixing my mistakes!

PR #1626, opened 2020-10-30
Use latest hyperkit driver 

guillaumerose:
I merged the vsock changes and it requires for the default and current use case to set VMNet = true.

### DIFF
--- a/pkg/crc/machine/hyperkit/constants.go
+++ b/pkg/crc/machine/hyperkit/constants.go
@@ -6,7 +6,7 @@ import "fmt"
 
 const (
 	MachineDriverCommand = "crc-driver-hyperkit"
-	MachineDriverVersion = "0.12.9"
+	MachineDriverVersion = "0.12.10-beta1"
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/hyperkit/constants.go
```diff
@@ -6,7 +6,7 @@ import "fmt"
 
 const (
 	MachineDriverCommand = "crc-driver-hyperkit"
-	MachineDriverVersion = "0.12.9"
+	MachineDriverVersion = "0.12.10-beta1"
```
cfergeau, line 9: I see you put a built binary in https://github.com/code-ready/machine-driver-hyperkit/releases/tag/v0.12.10-beta1
Fwiw, setting the version number in the `machine-driver-hyperkit` repo to 0.12.10, and changing it here might have been enough, people could build it from git, and then copy it to the right lo cation.

cfergeau (approved):

PR #1629, opened 2020-11-02
Use oc executable embedded in the bundle

guillaumerose:
oc executable is now in the bundle. This removes the hard link between
an openshift/snc release and a crc release. A crc release can use any
bundle compatible: for instance an okd bundle.

---

oc is not available after crc setup now. The bundle needs to be extracted first (with crc start).
still need to double check on Windows

### DIFF
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
-	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/code-ready/crc/pkg/download"
 	"github.com/code-ready/crc/pkg/embed"
 	"github.com/code-ready/crc/pkg/extract"
@@ -39,10 +38,6 @@ func New(executableName string, archiveURL string, destDir string, version strin
 	return &Cache{executableName: executableName, archiveURL: archiveURL, destDir: destDir, version: version, getVersion: getVersion}
 }
 
-func getCurrentOcVersion(executablePath string) (string, error) {
-	return getVersionGeneric(executablePath, "version", "--client")
-}
-
 func (c *Cache) GetExecutablePath() string {
 	return filepath.Join(c.destDir, c.executableName)
 }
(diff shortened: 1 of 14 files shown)


pkg/crc/cache/cache.go
```diff
@@ -51,13 +46,13 @@ func (c *Cache) GetBinaryName() string {
...
 
-/* getVersionGeneric runs the cached binary with 'args', and assumes the version string
+/* GetVersionGeneric runs the cached binary with 'args', and assumes the version string
  * was output on stdout's first line with this format:
  * something: <version>
  *
  * It returns <version> as a string
  */
-func getVersionGeneric(binaryPath string, args ...string) (string, error) {
+func GetVersionGeneric(binaryPath string, args ...string) (string, error) {
```
cfergeau, line 55: This does not seem needed.

pkg/crc/constants/constants.go
```diff
@@ -128,9 +114,21 @@ func GetHomeDir() string {
...
+			return err
+		}
+	}
+	if _, err := os.Stat(CrcOcBinDir); err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+		if err := os.MkdirAll(CrcOcBinDir, 0750); err != nil {
+			return err
+		}
```
cfergeau, line 131: I'd do that in `installSymlinkOrCopy`

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -15,13 +15,6 @@ import (
 )
 
 var genericPreflightChecks = [...]Check{
-	{
-		configKeySuffix:  "check-oc-cached",
-		checkDescription: "Checking if oc binary is cached",
-		check:            checkOcBinaryCached,
-		fixDescription:   "Caching oc binary",
-		fix:              fixOcBinaryCached,
-	},
```
cfergeau, line 24: Once you do https://github.com/code-ready/crc/pull/1630, we could have a preflight running only at `crc start` time ensuring an `oc` binary is available.

cfergeau (approved):

pkg/crc/cache/cache.go
guillaumerose: yes the linter was complaining the function was not used for Windows. nolint annotations didnt work. It was the easiest way to make it pass.

pkg/crc/preflight/preflight_checks_common.go
guillaumerose: yes!

pkg/crc/cache/cache.go
cfergeau:
```
-func getVersionGeneric(executablePath string, args ...string) (string, error) {
+func getVersionGeneric(executablePath string, args ...string) (string, error) { //nolint:deadcode,unused
```
works for me with `golangci-lint` 1.31 and `make cross-lint`

guillaumerose: thanks it works! I forgot the comma

pkg/crc/constants/constants.go
cfergeau: I'd still do that in `installSymlinkOrCopy` :) Why isn't this also creating `~/.crc/cache`?
This apparently could be simplified as:
```
func EnsureBaseDirectoriesExist() error {
    return os.MkdirAll(CrcOcBinDir, 0750)
}
```

https://golang.org/pkg/os/#MkdirAll
> If path is already a directory, MkdirAll does nothing and returns nil.

cfergeau (approved):

guillaumerose:
/retest
retest this please

PR #1630, opened 2020-11-02
crc setup prepares everything before crc start even bundle extraction

guillaumerose:
This moves the bundle extraction from start to setup action for most
users.

---

I always found weird that setup is half done in `crc setup` and half done in `crc start`.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -9,6 +9,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/cache"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 	"github.com/code-ready/crc/pkg/crc/validation"
 	"github.com/code-ready/crc/pkg/embed"
 	"github.com/docker/go-units"
@@ -30,11 +31,11 @@ var genericPreflightChecks = [...]Check{
 		fix:              fixGoodhostsExecutableCached,
 	},
 	{
-		configKeySuffix:  "check-bundle-cached",
-		checkDescription: "Checking if CRC bundle is cached in '$HOME/.crc'",
-		check:            checkBundleCached,
-		fixDescription:   "Unpacking bundle from the CRC executable",
-		fix:              fixBundleCached,
+		configKeySuffix:  "check-bundle-extracted",
+		checkDescription: "Checking if CRC bundle is extracted in '$HOME/.crc'",
+		check:            checkBundleExtracted,
+		fixDescription:   "Extracting bundle from the CRC executable",
+		fix:              fixBundleExtracted,
 		flags:            SetupOnly,
 	},
 	{
@@ -48,7 +49,7 @@ var genericPreflightChecks = [...]Check{
 	},
 }
 
-func checkBundleCached() error {
+func checkBundleExtracted() error {
 	if !constants.BundleEmbedded() {
 		return nil
 	}
@@ -58,7 +59,7 @@ func checkBundleCached() error {
 	return nil
 }
 
-func fixBundleCached() error {
+func fixBundleExtracted() error {
 	// Should be removed after 1.19 release
 	// This check will ensure correct mode for `~/.crc/cache` directory
 	// in case it exists.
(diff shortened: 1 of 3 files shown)


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -38,7 +39,7 @@ var genericPreflightChecks = [...]Check{
 	},
 	{
 		configKeySuffix:  "check-bundle-cached",
-		checkDescription: "Checking if CRC bundle is cached in '$HOME/.crc'",
+		checkDescription: "Checking if CRC bundle is installed in '$HOME/.crc'",
 		check:            checkBundleCached,
 		fixDescription:   "Unpacking bundle from the CRC binary",
 		fix:              fixBundleCached,
```
cfergeau, line 45: Here we have `installed`, `Cached` and `unpacking` in just a few lines, can you use this commit as an opportunity to make this consistent?

cfergeau (commented):
Can you expand a bit in the commit log about "for most users"? My reading is that this is about binaries with embedded bundles, and that if one builds from git/wants to use a non-embedded bundle, we'll still use code running at `crc start` time?
Otherwise looks good.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -74,12 +75,15 @@ func fixBundleCached() error {
 	}
 	if constants.BundleEmbedded() {
 		bundleDir := filepath.Dir(constants.DefaultBundlePath)
-		err := os.MkdirAll(bundleDir, 0775)
-		if err != nil && !os.IsExist(err) {
+		if err := os.MkdirAll(bundleDir, 0775); err != nil && !os.IsExist(err) {
```
cfergeau, line 78: I think `os.IsExist()` will never be returned if `bundleDir` already exists and is a directory. So this check is unneeded.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -65,7 +66,7 @@ func checkBundleCached() error {
 	return nil
```
cfergeau, line 66: This is only checking for `~/.crc/cache/crc_libvirt_4.6.1.crcbundle`, this needs to start with a check for `~/.crc/cache/crc_libvirt_4.6.1/`, or something more sophisticated which checks the directory has a valid bundle.

pkg/crc/preflight/preflight_checks_common.go
guillaumerose: I put installed

pkg/crc/preflight/preflight_checks_common.go
guillaumerose: done

pkg/crc/preflight/preflight_checks_common.go
guillaumerose: yes, more generally the name of the bundle is both coming from the crcbundle file on disk and from the directory in the bundle. This can cause mismatch sometimes for developers when the name is not the same. 
I guess we should revisit that.

At the end of the extraction we could also put a small empty file ".done" to indicate that the extraction is complete.

guillaumerose:
/retest
retest this please

pkg/crc/preflight/preflight_checks_common.go
cfergeau: Yeah my point was that the preflight check is not doing its job fully, if I `rm -rf ~/.crc/cache/crc_libvirt_4.6.1.crcbundle` and the run `crc setup`, the preflight check won't catch this, and `fixBundleExtracted` will not be called. There is a `bundle.Extract()` call in `getCrcBundleInfo()` which will get triggered and fix that, but it's not really meant for that situation.

test/integration/features/basic.feature
```diff
@@ -77,6 +78,7 @@ Feature: Basic test
     @windows
     Scenario: CRC setup on Windows
         When executing "crc setup" succeeds
+        Then stderr should contain "Caching oc executable"
         Then stderr should contain "Unpacking bundle from the CRC executable" if bundle is embedded
```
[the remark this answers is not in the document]
guillaumerose: Yes, fixed, thx.

pkg/crc/preflight/preflight_checks_common.go
cfergeau: Imo this still misses a check which calls `GetCachedBundleInfo` and then makes sure the various files it lists exist.

cfergeau (approved):
Looks good to me even if there is still an open question.

pkg/crc/preflight/preflight_checks_common.go
guillaumerose: I will open a follow-up with this. It is a more massive refactoring.

guillaumerose:
Follow-up will be created this/next week.

guillaumerose:
retest this please

guillaumerose:
/retest
retest this please

PR #1632, opened 2020-11-03
Use golang native ssh implementation on all platforms

guillaumerose:
Bring machine/ssh code in and prune everything not necessary anymore.

Successfully tested on the three platforms.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/docker/docker v1.13.1 // indirect
 	github.com/docker/go-units v0.4.0
 	github.com/h2non/filetype v1.1.0
+	github.com/hectane/go-acl v0.0.0-20190604041725-da78bae5fc95
 	github.com/imdario/mergo v0.3.9 // indirect
 	github.com/libvirt/libvirt-go-xml v6.8.0+incompatible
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/docker/docker v1.13.1 // indirect
 	github.com/docker/go-units v0.4.0
 	github.com/h2non/filetype v1.1.0
+	github.com/hectane/go-acl v0.0.0-20190604041725-da78bae5fc95
```
gbraad, line 18: is this to change the permissions of generated certs?

pkg/crc/ssh/keys_windows.go
```diff
@@ -0,0 +1,52 @@
...
+			return err
+		}
+
+	}
+
+	return nil
+}
+
+func windowsChmod(filePath string, fileMode os.FileMode) error {
+	if err := acl.Chmod(filePath, fileMode); err != nil {
```
gbraad, line 47: ah, so yes it is.

gbraad (approved):

guillaumerose:
rebase with the diagnostics tarball in case I found something useful before merge.

PR #1634, opened 2020-11-03
delete: Fix message when user refuses cluster deletion

cfergeau:
Currently, after `crc delete`, the user is asked
"Do you want to delete the OpenShift cluster [y/N]?"
If they answer 'no', or just press enter, `crc` will still display
"Deleted the OpenShift cluster" even if the cluster was not actually
deleted.
This commit ensures this is only displayed when `client.Delete()` was
called.

## Testing

1. make sure you have a cluster created
2. run `crc delete`
3. press 'N' or 'enter'
4. "Deleted the OpenShift cluster is not displayed"

### DIFF
--- a/cmd/crc/cmd/delete.go
+++ b/cmd/crc/cmd/delete.go
@@ -33,9 +33,9 @@ var deleteCmd = &cobra.Command{
 	},
 }
 
-func deleteMachine(client machine.Client, clearCache bool, cacheDir string, interactive, force bool) error {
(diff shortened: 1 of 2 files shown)


guillaumerose:
I guess you can add a unit test for that. We should have everything for.

cfergeau:
> I guess you can add a unit test for that. We should have everything for.

Not sure the test I added is what you had in mind, I could not really think of a better way.

guillaumerose (approved):

Issue #1635, opened 2020-11-03
Degraded ingress controller causes CI failure


guillaumerose:
In this CI failure (https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/code-ready_crc/1534/pull-ci-code-ready-crc-master-e2e-crc/1323617626810748928), we can read that the cluster correctly started, crc status is fine but as soon we try to login it fails.

Logs say for crc status:
```
time="2021-02-03T14:05:06Z" level=debug msg="ingress operator is degraded, Reason: IngressControllersDegraded"
``` 

But it has reported a Running status since the test continued directly after the first call to crc status.

The diagnostic tarball gives us this information about the ingress operator:
``` 
  Conditions:
    Last Transition Time:  2020-10-26T04:48:13Z
    Message:               desired and current number of IngressControllers are equal
    Reason:                AsExpected
    Status:                True
    Type:                  Available
    Last Transition Time:  2020-10-26T04:48:13Z
    Message:               desired and current number of IngressControllers are equal
    Reason:                AsExpected
    Status:                False
    Type:                  Progressing
    Last Transition Time:  2021-02-03T13:59:18Z
    Message:               Some ingresscontrollers are degraded: default
    Reason:                IngressControllersDegraded
    Status:                True
    Type:                  Degraded
``` 

It is both Degraded and Available at the same time.

I guess there is a mismatch in our code here: https://github.com/code-ready/crc/blob/b4a6188088da46a717be13fca4ed6653f08026f6/pkg/crc/machine/machine.go#L571

The switch sees Available and set openshiftStatus to Running. It doesn't fall in the Degraded case just after.

[diagnose-1612361107.zip](https://github.com/code-ready/crc/files/5482078/diagnose-1612361107.zip)

PR #1636, opened 2020-11-03
Fix priority between Degraded and Available statuses

guillaumerose:
Fixes https://github.com/code-ready/crc/issues/1635

### DIFF
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -83,7 +83,7 @@ func getStatus(ocConfig oc.Config, selector []string) (*Status, error) {
 		}
 	}
 	if !found {
-		return nil, errors.New("no cluster operator found")
+		return cs, errors.New("no cluster operator found")
 	}
 	return cs, nil
 }
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -35,12 +35,12 @@ func getStatus(ocConfig oc.Config, selector []string) (*Status, error) {
 
 	data, _, err := ocConfig.RunOcCommandPrivate("get", "co", "-ojson")
 	if err != nil {
-		return cs, err
+		return nil, err
 	}
 
 	var co openshiftapi.ClusterOperatorList
 	if err := json.Unmarshal([]byte(data), &co); err != nil {
-		return cs, err
+		return nil, err
 	}
 
 	found := false
@@ -82,7 +82,7 @@ func getStatus(ocConfig oc.Config, selector []string) (*Status, error) {
 		}
 	}
 	if !found {
-		return cs, errors.New("no cluster operator found")
+		return nil, errors.New("no cluster operator found")
 	}
 	return cs, nil
 }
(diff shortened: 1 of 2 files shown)


guillaumerose:
retest this please

pkg/crc/cluster/clusteroperator.go
```diff
@@ -83,7 +83,7 @@ func getStatus(ocConfig oc.Config, selector []string) (*Status, error) {
 		}
 	}
 	if !found {
-		return nil, errors.New("no cluster operator found")
+		return cs, errors.New("no cluster operator found")
```
cfergeau, line 85: This is similar to what other error path are doing in this method, so ok. However, if a caller is not checking for err before using the returned `Status`, then it's a bug in the caller. Being aware of this misbehaviour through a crash rather than having the program keep running while doing weird things is in a way a good thing :)

`return &Status{}` be better in all the error branches, as `cs` is initialized with `Available: true` which may be unwanted in error cases.

pkg/crc/machine/machine.go
```diff
@@ -554,36 +554,37 @@ func (client *client) Status() (*ClusterStatusResult, error) {
...
 		OpenshiftVersion: crcBundleMetadata.GetOpenshiftVersion(),
 		DiskUse:          diskUse,
 		DiskSize:         diskSize,
 	}, nil
 }
 
+func getOpenShiftStatus(sshRunner *crcssh.Runner) string {
+	status, err := cluster.GetClusterOperatorsStatus(oc.UseOCWithSSH(sshRunner))
+	if err != nil {
+		logging.Debugf("cannot get openshift status: %v", err)
```
cfergeau, line 573: `OpenShift`

pkg/crc/machine/machine.go
```diff
@@ -559,7 +559,7 @@ func (client *client) Status() (*ClusterStatusResult, error) {
 		return nil, errors.Wrap(err, "Cannot get root partition usage")
 	}
 	return &ClusterStatusResult{
-		CrcStatus:        vmStatus,
+		CrcStatus:        state.Running,
```
cfergeau, line 562: I would mention in the commit log that this change is made possible by one of the previous commits.

cfergeau (approved):

cfergeau:
> Fixes #1635

Fwiw, it's also useful to have in the commit log of ` Some operators can be both degraded and available at the same time ` so that it's easier to get back to the corresponding issue/corresponding PR if needed in the future. I just did this during the review because I had vague memories of https://github.com/code-ready/crc/pull/1058#discussion_r394510100

pkg/crc/machine/machine.go
guillaumerose: done

pkg/crc/machine/machine.go
guillaumerose: done

pkg/crc/cluster/clusteroperator.go
guillaumerose: with the refactorings I did, at the end, we can get rid of this. I added a commit on top to return nil status in case of an error.

The issue was here: https://github.com/code-ready/crc/blob/master/pkg/crc/machine/machine.go#L558. Now we properly handle this case.

guillaumerose:
centos green. Let's merge it.

Issue #1638, opened 2020-11-03
Insufficient memory to schedule httpd-example in the centos CI


guillaumerose:
This failure https://ci.centos.org/job/codeready-crc-pr/2428 is due to a lack of memory.

Scheduler logs:
``` 
157:I1103 14:17:19.868380       1 factory.go:445] "Unable to schedule pod; no fit; waiting" pod="testproj/httpd-example-1-6hdqv" err="0/1 nodes are available: 1 Insufficient memory."
160:I1103 14:18:49.863218       1 factory.go:445] "Unable to schedule pod; no fit; waiting" pod="testproj/httpd-example-1-6hdqv" err="0/1 nodes are available: 1 Insufficient memory."
163:I1103 14:19:10.368390       1 factory.go:445] "Unable to schedule pod; no fit; waiting" pod="testproj/httpd-example-1-6hdqv" err="0/1 nodes are available: 1 Insufficient memory."
``` 

Although `describe nodes` reports a good state:
```
Conditions:
  Type             Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
  ----             ------  -----------------                 ------------------                ------                       -------
  MemoryPressure   False   Tue, 03 Nov 2020 14:14:42 +0000   Tue, 03 Nov 2020 14:02:07 +0000   KubeletHasSufficientMemory   kubelet has sufficient memory available
  DiskPressure     False   Tue, 03 Nov 2020 14:14:42 +0000   Tue, 03 Nov 2020 14:02:07 +0000   KubeletHasNoDiskPressure     kubelet has no disk pressure
  PIDPressure      False   Tue, 03 Nov 2020 14:14:42 +0000   Tue, 03 Nov 2020 14:02:07 +0000   KubeletHasSufficientPID      kubelet has sufficient PID available
  Ready            True    Tue, 03 Nov 2020 14:14:42 +0000   Tue, 03 Nov 2020 14:02:07 +0000   KubeletReady                 kubelet is posting ready status
``` 

```
$ df -h
              total        used        free      shared  buff/cache   available
Mem:           8557        6205         143          23        2208        2816
Swap:             0           0           0

```
[diagnose-1604413151.zip](https://github.com/code-ready/crc/files/5482338/diagnose-1604413151.zip)

guillaumerose:
This also happen on the openshift CI.

guillaumerose:
same on a local machine.

guillaumerose:
describe nodes:
```
[...]

PR #1641, opened 2020-11-04
Give more memory to the VM in health e2e test

guillaumerose:
Related to https://github.com/code-ready/crc/issues/1638

Temporary fix until a better solution is found.


jsliacan (approved):
LGTM, tried it, 12G is enough. Thanks

PR #1645, opened 2020-11-05
Add explicit order to all preflight checks

guillaumerose:
This ensures the order is always what we want. For instance, root check
is always the first check performed now.

Before:
```
INFO Checking if oc executable is cached          
INFO Checking if goodhosts executable is cached   
INFO Checking for obsolete crc-driver-libvirt     
INFO Checking if libvirt 'crc' network is available 
```

After:
``` 
INFO Checking if libvirt 'crc' network is available 
INFO Checking for obsolete crc-driver-libvirt     
INFO Checking if goodhosts executable is cached   
INFO Checking if oc executable is cached          

``` 

With this + https://github.com/code-ready/crc/pull/1630, I would like to make the bundle extraction to be the last item of the list (because it is the longest).

### DIFF
--- a/pkg/crc/preflight/preflight.go
+++ b/pkg/crc/preflight/preflight.go
@@ -2,6 +2,7 @@ package preflight
 
 import (
 	"fmt"
+	gosort "sort"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	"github.com/code-ready/crc/pkg/crc/config"
@@ -24,6 +25,7 @@ type FixFunc func() error
 type CleanUpFunc func() error
 
 type Check struct {
+	order              int
 	configKeySuffix    string
 	checkDescription   string
 	check              CheckFunc
@@ -101,6 +103,18 @@ func (check *Check) doCleanUp() error {
 	logging.Infof("%s", check.cleanupDescription)
 
 	return check.cleanup()
+}
+
+func sort(checks []Check) {
+	gosort.Slice(checks, func(i, j int) bool {
+		if checks[i].order < checks[j].order {
+			return true
+		}
+		if checks[i].order > checks[j].order {
+			return false
+		}
+		return checks[i].configKeySuffix < checks[j].configKeySuffix
+	})
 }
 
 func doPreflightChecks(config config.Storage, checks []Check) error {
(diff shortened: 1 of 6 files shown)


cfergeau (commented):
Could you achieve a similar result with just an optional `priority` field which could be set to `RunFirst` or `RunLast`? My feeling is that in 90% of the tests, the mandatory `order` value is not really needed, and a very arbitrary number.

pkg/crc/preflight/preflight.go
```diff
@@ -103,6 +105,18 @@ func (check *Check) doCleanUp() error {
...
+	gosort.Slice(checks, func(i, j int) bool {
+		if checks[i].order < checks[j].order {
+			return true
+		}
+		if checks[i].order > checks[j].order {
+			return false
+		}
+		return checks[i].configKeySuffix < checks[j].configKeySuffix
+	})
+}
```
cfergeau, line 118: Looking at the comparison function, this is ever more complicated than I thought, the genericPreflightChecks won't be executed in the order they are listed in the source file even if they have the same order, if one forgets to set `order`, their check will be the very first one, ... Overall I fear it's going to be very hard to predict the order the tests are going to be run in without looking at the code, when you add a new test, you have to check the order of most of the other ones if you want it to be run at the appropriate time, ... I'd really favour avoiding this field, and just tagging the few tests that we want to be first or last, but for all the others, just use the order they are in the `getPreflightChecks` call (before sorting the array).

guillaumerose: yes, not easy. I found it also not really nice to read.

I would like to do the same as init or OpenShift to organize their manifests (0000_50_foo.yaml, etc.) Is there a way to do that more nicely?

cfergeau: I never found that way of ordering very nice when you are not very familiar with the project. When you want to add a new file, which priority should you pick? How do you know if it should be run early/late, if some priorities are reserved or not, if you can reuse the same priority as another file or not, ... 

I still think that just a RunFirst/RunLast tag for the few tests we want to (de)prioritize would be plenty enough, and easy to understand. I think it would work ok for us as we won't have hundreds of preflights.

guillaumerose:
Will re-order by hand.

PR #1653, opened 2020-11-09
Bump libmachine. No more RPC for hyperv driver.

guillaumerose:


jsliacan (approved):
Not going to comment on the individual changes. Overall starts OK (after stop too). No issue with starting CRC.

PR #1656, opened 2020-11-10
Allow to resize disk when machine is created

cfergeau:
For some reason, during initial implementation of disk resizing, setting
the disk size when the VM is initially created was missing. This commit
adds this, but this also needs a change in the libvirt machine driver.

## Testing

1. `start --disk-size 40` succeeds first time after `setup` succeeded
2. `df -h` inside the VM, or `virsh -c qemu:///system vol-info crc.qcow2 crc` confirm that the image was resized to 40GiB

### DIFF
--- a/pkg/crc/machine/config/config.go
+++ b/pkg/crc/machine/config/config.go
@@ -10,6 +10,7 @@ type MachineConfig struct {
 	Name            string
 	Memory          int
 	CPUs            int
+	DiskSize        int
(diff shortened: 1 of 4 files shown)


guillaumerose (approved):
Would be great to have an e2e test for that

cfergeau:
> Would be great to have an e2e test for that

yep, I think Jakub was working on one.

Issue #1346, opened 2020-07-01
crc cleanup on macOS doesn't remove existing crc vm


anjannath:
On the other platforms (windows and linux) `crc cleanup` removes an existing crc vm but not in the case of macOS

praveenkumar:
It is because we don't have any management cli for hyperkit like we have for libvirt and hyperV, in this case we might need to grab the running hyperkit process and make sure it is crc related then kill it.

anjannath:
Ahh, just checked the `removeCrcVM()` for windows, and it manually deletes the vm without using the machine api, can we not use `machine.Delete()` from the cleanup function?

gbraad:
@anjannath this is referenced by e2f2062. Anything still to be done?

cfergeau:
> Ahh, just checked the `removeCrcVM()` for windows, and it manually deletes the vm without using the machine api, can we not use `machine.Delete()` from the cleanup function?

More than a year later, I'm wondering the same thing, why this ends up duplicating code preexisting in the machine drivers ;)

PR #1658, opened 2020-11-10
Remove hyperkit machines directory during crc cleanup

anjannath:
Fixes: #1346 

The second commit does the removing of hyperkit directory.

### DIFF
--- a/pkg/crc/preflight/preflight.go
+++ b/pkg/crc/preflight/preflight.go
@@ -5,6 +5,7 @@ import (
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	"github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/network"
 )
@@ -143,6 +144,7 @@ func doFixPreflightChecks(config config.Storage, checks []Check) error {
 }
 
 func doCleanUpPreflightChecks(checks []Check) error {
+	var mErr errors.MultiError
 	// Do the cleanup in reverse order to avoid any dependency during cleanup
 	for i := len(checks) - 1; i >= 0; i-- {
 		check := checks[i]
@@ -151,10 +153,16 @@ func doCleanUpPreflightChecks(checks []Check) error {
 		}
 		err := check.doCleanUp()
 		if err != nil {
-			return err
+			// If an error occurs in a cleanup function
+			// we log/collect it and  move to the  next
+			logging.Debug(err)
+			mErr.Collect(err)
 		}
 	}
-	return nil
+	if len(mErr.Errors) == 0 {
+		return nil
+	}
+	return mErr
 }
 
 func doRegisterSettings(cfg config.Schema, checks []Check) {
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -3,9 +3,12 @@ package preflight
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"os/user"
+	"path/filepath"
 
 	"github.com/code-ready/crc/pkg/crc/cache"
+	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/network"
 	crcos "github.com/code-ready/crc/pkg/os"
@@ -180,5 +183,18 @@ func addFileWritePermissionToUser(filename string) error {
 	}
 	logging.Debugf("%s is readable/writable by current user", filename)
 
+	return nil
+}
+
+func stopCRCHyperkitProcess() error {
+	path, err := exec.LookPath("pkill")
+	if err != nil {
+		logging.Debugf("Could not find 'pkill'. %v", err)
+		return fmt.Errorf("Could not find 'pkill'. %w", err)
+	}
+	if _, _, err := crcos.RunWithDefaultLocale(path, "-f", filepath.Join(constants.CrcBinDir, "hyperkit")); err != nil {
+		logging.Debugf("Failed to kill 'hyperkit' process. %v", err)
+		return fmt.Errorf("Failed to kill 'hyperkit' process. %w", err)
+	}
 	return nil
 }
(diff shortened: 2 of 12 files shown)


pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -182,3 +185,27 @@ func addFileWritePermissionToUser(filename string) error {
...
 	return nil
 }
+
+func removeCRCHyperkitMachine() error {
+	err := os.RemoveAll(filepath.Join(constants.MachineInstanceDir, "crc"))
+	if err != nil {
+		logging.Debugf("Failed to remove hyperkit machine directory: %v", err)
+		return err
+	}
+	return nil
```
cfergeau, line 195: windows/linux already have code doing that, have you considered moving this to `preflight_checks_common.go` since we want to do it for all platforms? I haven't looked if it's doable.

pkg/crc/preflight/preflight.go
```diff
@@ -151,7 +151,9 @@ func doCleanUpPreflightChecks(checks []Check) error {
 		}
 		err := check.doCleanUp()
 		if err != nil {
-			return err
+			// If an error occurs in a cleanup function
+			// we log it and  move on to the next  func
+			logging.Debug(err)
```
guillaumerose, line 156: You can aggregate errors with errors.MultiError and return it instead of nil.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -182,3 +185,27 @@ func addFileWritePermissionToUser(filename string) error {
...
+}
+
+func stopCRCHyperkitProcess() error {
+	path, err := exec.LookPath("pkill")
+	if err != nil {
+		logging.Debugf("Could not find 'pkill'. %v", err)
+		return fmt.Errorf("Could not find 'pkill'. %v", err)
```
guillaumerose, line 202:
``` 
return errors.Wrap(err, "Could not find 'pkill'")
```

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -182,3 +185,27 @@ func addFileWritePermissionToUser(filename string) error {
...
+	}
+	cmd := exec.Command(path, "-f", filepath.Join(constants.CrcBinDir, "hyperkit"))
+	err = cmd.Run()
+	if err != nil {
```
guillaumerose, line 206: If you like it: 
`if err := exec.Command(path, "-f", filepath.Join(constants.CrcBinDir, "hyperkit")).Run(); err != nil`

Maybe use crcos.RunWithDefaultLocale ?

pkg/crc/preflight/preflight_checks_darwin.go
anjannath: Good idea, can create a `removeCRCMachineDir()` in `preflight_checks_common.go`

pkg/crc/preflight/preflight_checks_darwin.go
anjannath: this should be ok without `crcos.RunWithDefaultLocale`, as we are not using the o/p from this command..

cfergeau: In my  opinion, we should never use `exec.Command`/`cmd.Run`, and always go through `RunWithDefaultLocale`. Even if it does not matter here, this could be copied and pasted to do something different where it would matter, which is why I prefer to use `RunWithDefaultLocale` most of the time (and also consistency, `exec.Command.Run()` and `RunWithDefaultLocale` looks very different)

pkg/crc/preflight/preflight.go
```diff
@@ -151,10 +153,13 @@ func doCleanUpPreflightChecks(checks []Check) error {
...
 		if err != nil {
-			return err
+			// If an error occurs in a cleanup function
+			// we log/collect it and  move to the  next
+			logging.Debug(err)
+			mErr.Collect(err)
 		}
 	}
-	return nil
+	return mErr
```
cfergeau, line 162: I think we want to return nil if there was no error, and `mErr` if there was at least one error?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -385,13 +385,6 @@ func removeLibvirtCrcNetwork() error {
 }
 
 func removeCrcVM() error {
-	logging.Debug("Removing 'crc' VM")
```
cfergeau, line 388: I'd keep this message.

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: Or `fmt.Errorf("Could not find 'pkill'. %w", err)` ?

cfergeau (approved):
Couple of minor comments, but overall looks good.

pkg/crc/preflight/preflight.go
cfergeau: Won't an empty `mErr` be seen sa an error anyway? On retrospect, I'm not sure it was a good idea when I pushed for `multierror` to implement the `error` interface, a `ToError()` method was probably better :)
Apart from this, /lgtm

anjannath: Yes, empty `mErr` is seen as an error. Should we then add a `ToError()` method in `multierror`?

cfergeau: I would go for a quick work-around in this PR, and a follow-up PR if you want to clean that up with better API

anjannath: For now checking at the end if `mErr` is empty and then return nil or error based on that..
```diff
-       return nil
+       if len(mErr.Errors) <= 0 {
+               return nil
+       }
+       return goerrors.New(mErr.Error())
```

pkg/crc/preflight/preflight.go
```diff
@@ -151,10 +154,16 @@ func doCleanUpPreflightChecks(checks []Check) error {
...
+			// we log/collect it and  move to the  next
+			logging.Debug(err)
+			mErr.Collect(err)
 		}
 	}
-	return nil
+	if len(mErr.Errors) == 0 {
+		return nil
+	}
+	return goerrors.New(mErr.Error())
```
cfergeau, line 166: I don't think you need to create a new error, just
```
    if len(mErr.Errors) == 0 {
		return nil
	}
	return mErr
```

cfergeau (commented):
I'm a bit confused, I thought you had addressed https://github.com/code-ready/crc/pull/1658/commits/5b0cba100fcbdc0b2ee8a6432ae6438094bcc4a7#r525327212 and https://github.com/code-ready/crc/pull/1658/commits/8e1932202f83b226611dc6ff3dd4b90a72acf99e#r525330505 previously?

cfergeau (approved):

PR #1659, opened 2020-11-10
Fixes for daemon service in windows

### LINKED ISSUE #1637: Daemon on Windows is running as admin
### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? Yes

## CRC version
```bash
CodeReady Containers version: 1.18.0+bb304aa
OpenShift version: 4.6.1 (embedded in binary)
```
 
### Steps to reproduce

  1. `crc cleanup`
  2. `crc config set pull-secret-file <location>`
  3. `crc setup --enable-experimental-feature`
  4. Go to the tray and start CRC from there.

### Expected
After a while, CRC has status Running.

### Actual
Get a pop-up notification `Cluster did not start...`. 

### Logs (from crcd.log)

```
time="2020-11-03T18:42:14+05:30" level=debug msg="Received Request:{\"command\":\"start\"}"
time="2020-11-03T18:42:14+05:30" level=info msg="Checking if oc binary is cached"
time="2020-11-03T18:42:14+05:30" level=debug msg="Running 'C:\\Users\\Anjan\\.crc\\bin\\oc\\oc.exe version --client'"
time="2020-11-03T18:42:14+05:30" level=debug msg="Found oc.exe version 4.6.1"
time="2020-11-03T18:42:14+05:30" level=debug msg="oc binary already cached"
time="2020-11-03T18:42:14+05:30" level=info msg="Checking if podman remote binary is cached"
time="2020-11-03T18:42:14+05:30" level=debug msg="Currently podman remote is not supported"
time="2020-11-03T18:42:14+05:30" level=info msg="Checking if goodhosts binary is cached"
time="2020-11-03T18:42:14+05:30" level=debug msg="goodhost binary already cached"
time="2020-11-03T18:42:14+05:30" level=info msg="Checking minimum RAM requirements"
time="2020-11-03T18:42:14+05:30" level=debug msg="Total memory of system is 17066782720 bytes"
time="2020-11-03T18:42:14+05:30" level=info msg="Checking if running as normal user"
time="2020-11-03T18:42:14+05:30" level=debug msg="Ran as administrator"
time="2020-11-03T18:42:14+05:30" level=debug msg="crc should be ran as a normal user"
```

anjannath:
This partially fixes #1637, in case the user is using a standard user account then, this will install the daemon service with the standard user as the logon user and everything should work fine (starting the cluster through the daemon is possible and doesn't fail the "is-root" check)

If the user is using an Administrator user account then the "is-root" check still fails and daemon can't complete the start.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_tray_windows.go
+++ b/pkg/crc/preflight/preflight_checks_tray_windows.go
@@ -91,7 +91,10 @@ func fixTrayInstalled() error {
 	if _, err = goos.Stat(filepath.Join(tempDir, "success")); goos.IsNotExist(err) {
 		return fmt.Errorf("Installation script didn't execute successfully: %v", err)
 	}
-	return nil
+
+	// start the tray process
+	_, _, err = powershell.Execute("Start-Process", constants.TrayExecutablePath)
+	return err
 }
 
 func escapeWindowsPassword(password string) string {
--- a/pkg/crc/preflight/preflight_checks_tray_windows.go
+++ b/pkg/crc/preflight/preflight_checks_tray_windows.go
@@ -61,8 +61,10 @@ func fixTrayInstalled() error {
 
 	// sanitize password
 	password = escapeWindowsPassword(password)
+	username := goos.Getenv("USERNAME")
 
 	psScriptContent := genTrayInstallScript(
+		username,
 		password,
 		tempDir,
 		binPathWithArgs,
@@ -121,7 +123,9 @@ func removeTray() error {
 		_ = goos.RemoveAll(tempDir)
 	}()
 
+	username := goos.Getenv("USERNAME")
 	psScriptContent := genTrayRemovalScript(
+		username,
 		trayProcessName,
 		constants.TrayShortcutName,
 		constants.DaemonServiceName,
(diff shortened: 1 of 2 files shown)


anjannath:
This still have another issue that came up while discussing with @adrianriobo  

The start from daemon won't succeed, as for setting up the DNS on the "Default Switch" the daemon needs Admin rights, which cannot be asked through a UAC prompt as the service cannot interact with the desktop, to make the service interactive the Log on user needs to be "LocalSystem" which again is an Administrator account.

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -61,11 +60,13 @@ func fixTrayInstalled() error {
 
 	// sanitize password
 	password = escapeWindowsPassword(password)
+	username := goos.Getenv("USERNAME")
```
guillaumerose, line 63: Maybe not relevant here I don't know but you can probably use `user.Current()`

anjannath: Yes, but `user.Username` returns full display name with the machine name like `DESKTOP-HFRVFP1\Anjan` i only require the account part like from `USERNAME` variable.. 
```posh
PS C:\Users\Anjan\github.com\code-ready\crc> $env:USERNAME
Anjan
```

anjannath:
We need another solution, with these patches only `crc setup` succeeds but starting crc from the tray still fails in the `Running as admin user..` peflight check.

But it works if the user is using a standard account, in that case the daemon service doesn't run as admin.

PR #1661, opened 2020-11-11
[WIP] Restructuring tests: an example of splitting basic tests into integration and e2e

jsliacan:
## Idea

This PR introduces initial changes so we have a concrete example when we talk about restructuring the CRC tests.

By adhering to Cucumber's design principles (thus not needing BeforeFeature hooks), we can upgrade to godog v0.11.0 and get support for jUnit file export. This means that we need to shift our more detailed (integration) tests away from Cucumber. Ginkgo was suggested as a possible candidate.

So this is a preview of splitting the tests into _integration_ and _e2e_ on the example of `basic` tests. The integration part is in Ginkgo and the e2e part is in Cucumber.

I reused a lot of Guillaume's Ginkgo set-up code from Tea Time discussion.

**Still missing**
- Test on all 3 platforms (only Linux has been tested)
- Test with standalone bundle (only release bits were tested)
- Various combinations of small commands in Ginkgo tests. See [this map](https://drive.google.com/file/d/1Zdk5zm-PLkiDJTFaGeJxVfs_zaCp64jW/view?usp=sharing). The `config`, `ip`, `console`, `oc-env`, `podman-env` subtrees are not tested yet. Many flag combinations are also missing as of now.
- Need to designate a _focus_ subset of basic tests to run as an "important" subset when not everything needs to be tested.
- Did not update dependency to godog v0.11.0, hence no jUnit output for e2e tests yet.

**Warning**
- Renamed `PULL_SECRET_FILE` env var to `PULL_SECRET_LOCATION` to be analogous to `BUNDLE_LOCATION` var. This `Makefile` change will affect pipelines.

## Proposed changes

1. Move the integration component of the `basic` tests to Ginkgo framework in `test/ginkgo`
2. Add disk/mem/cpu resize integration tests in Ginkgo.
3. Condense the Cucumber e2e tests component to adhere to Cucumber design principles. E.g. independent Scenarios.
4. Produce jUnit files with test results to upload to ReportPortal (or elsewhere). 

## Testing

1. Set `BUNDLE_LOCATION`(if using standalone bundle) and `PULL_SECRET_LOCATION` variables.
2. Go to `test/ginkgo` and trigger the integration tests by typing `ginkgo`. Check jUnit output in `test/ginkgo/results`.
3. To run e2e basic test, modify makefile to only run `<os> && basic` tags and type `make integration` from repo dir.


test/ginkgo/resize_test.go
```diff
@@ -0,0 +1,113 @@
+package test_test
+
+import (
+	//"bytes"
```
guillaumerose, line 4: Use GoLand, it works really well!

Makefile
```diff
@@ -125,7 +125,7 @@ build_integration: $(SOURCES)
 integration:
 GODOG_OPTS = --godog.tags=$(GOOS)
 ifndef PULL_SECRET_FILE
-	PULL_SECRET_FILE = --pull-secret-file=$(HOME)/Downloads/crc-pull-secret
+	PULL_SECRET_LOCATION = --pull-secret-file=$(HOME)/Downloads/crc-pull-secret
```
guillaumerose, line 128: Is it possible to start without this change? CI is broken with this. Or maybe handle both parameters.

go.mod
```diff
@@ -41,7 +45,7 @@ require (
 	gopkg.in/AlecAivazis/survey.v1 v1.8.5
 	howett.net/plist v0.0.0-20181124034731-591f970eefbb
 	k8s.io/api v0.19.0
-	k8s.io/client-go v0.19.0
+	k8s.io/client-go v8.0.0+incompatible
```
guillaumerose, line 48: Weird change

test/ginkgo/resize_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+		It("start CRC with custom memory, cpu, and disk", func() {
+			Expect(RunCRCExpectSuccess("start", "-b", bundleLocation, "-p", pullSecretLocation, "--memory", "12000", "--cpus", "6", "--disk-size", "40")).To(ContainSubstring("Started the OpenShift cluster"))
+		})
+
+		switch os := runtime.GOOS; os {
+		case "linux":
+
+			It("check memory size", func() {
+				out, err := RunOnHost("ssh", "-i", credPath, "core@192.168.130.11", "cat", "/proc/meminfo")
```
guillaumerose, line 38: You can use the same code as collect.go:

```
	client := machine.NewClient(constants.DefaultName, true, network.DefaultMode)
	ip, err := client.IP()
	if err != nil {
		return err
	}
	ssh, err := ssh.NewClient(constants.DefaultSSHUser, ip, 22, &ssh.Auth{
		Keys: []string{constants.GetPrivateKeyPath()},
	})
	if err != nil {
		return err
	}
	ssh.Output("cat /proc/meminfo")
```

It will work for all platforms o/

test/ginkgo/resize_test.go
```diff
@@ -0,0 +1,113 @@
...
+				Expect(out).Should(MatchRegexp(`MemTotal:[\s]*11\d{6}`))
+			})
+
+			It("check number of cpus", func() {
+				out, err := RunOnHost("ssh", "-i", credPath, "core@192.168.130.11", "cat", "/proc/cpuinfo")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(out).Should(MatchRegexp(`processor[\s]*\:[\s]*5`))
+			})
+			It("check size of VM disk", func() {
+				out, err := RunOnHostWithPrivilege("virsh", "vol-info", "crc.qcow2", "crc")
```
guillaumerose, line 49: Check the size inside the VM so that it will  be the same check for all platforms.

test/ginkgo/testsuite_test.go
```diff
@@ -0,0 +1,101 @@
...
+var credPath string
+var userHome string
+var versionInfo VersionAnswer
+
+var bundleLocation string
+var pullSecretLocation string
+
+func TestTest(t *testing.T) {
+
+	rand.Seed(time.Now().UTC().UnixNano()) // what's this for?
```
guillaumerose, line 51: This is needed if we use rand package and ask for random number. We don't use it in this PR, you can remove it!

test/ginkgo/testsuite_test.go
```diff
@@ -0,0 +1,101 @@
...
+	RunSpecsWithDefaultAndCustomReporters(t, "Test Suite", []Reporter{junitReporter})
+	//RunSpecs(t, "Test Suite")
+}
+
+var _ = BeforeSuite(func() {
+
+	// set userHome
+	usr, err := user.Current()
+	if err != nil {
+		os.Exit(1)
```
guillaumerose, line 66: `Expect(err).NotTo(HaveOccurred())` instead of os.Exit.

guillaumerose:
Good start! Let's make the CI green so that we can all iterate on this

go.mod
jsliacan: Yeah, not manual. No idea.

Makefile
jsliacan: I left integration target as it was and added a `gintegration` target so we can control the two testsuites separately. 

I find `PULL_SECRET_FILE` vs `BUNDLE_LOCATION` inconsistent. But `--pull-secret-file` is an option for `crc start`... So, I left it as `FILE` everywhere for now.

test/ginkgo/resize_test.go
jsliacan: I rewrote it using this code. Thanks! However, it's the crc code that we're now using to test CRC. It raises questions :). I remember I once tried to use the sshRunner function but was told that it's better not to use CRC code (which, strictly speaking, makes sense). We can discuss on Slack.

test/ginkgo/testsuite_test.go
jsliacan: I did that, but I remember reading that `Expect` should not be used outside of `It`. Maybe I missed a caveat (like `BeforeSuite` being similar construct as `It`?).

guillaumerose:
I rebased the PR and make it work on CentOS CI but it's very red: https://ci.centos.org/job/codeready-crc-pr/2685/console

jsliacan:
Will resubmit in smaller chunks.

PR #1666, opened 2020-11-12
Move shell package from libmachine to crc

guillaumerose:
This is the only location where this package is used.


cfergeau (commented):
Why are the tests not working on appveyor? (I can guess, but should be in the log)

guillaumerose:
I added the error from AppVeyor. I guess it's an error due to cygwin or it is launched (cmd instead of powershell?).

pkg/os/shell/shell_windows_test.go
```diff
@@ -33,6 +35,8 @@ func TestGetNameAndItsPpidOfParent(t *testing.T) {
 }
 
 func TestGetNameAndItsPpidOfGrandParent(t *testing.T) {
+	t.Skip("not working on AppVeyor")
```
praveenkumar, line 38: Why would we have those tests if we are going to skip those, I think this will skipped on any windows system. I think it is failing because we are asserting powershell, even in appveyour it execute on cmd?

guillaumerose: I tried to use ps on AppVeyor but because of we use `make test`, there is a sh.exe also in the party... :/

pkg/os/shell/shell_windows.go
```diff
@@ -1,5 +1,92 @@
...
+	"strings"
+	"syscall"
+	"unsafe"
+)
+
 var (
 	supportedShell = []string{"cmd", "powershell"}
 )
+
+// re-implementation of private function in https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L945
```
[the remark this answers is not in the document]
guillaumerose: done

pkg/os/shell/shell_windows.go
```diff
@@ -1,5 +1,90 @@
...
+	"strings"
+	"syscall"
+	"unsafe"
+)
+
 var (
 	supportedShell = []string{"cmd", "powershell"}
 )
+
+// re-implementation of private function in https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go
```
cfergeau, line 15: https://github.com/mitchellh/go-ps can probably be used if we don't want to c&p this code directly (but not expecting any changes related to this just now)

pkg/os/shell/shell_windows_test.go
```diff
@@ -8,6 +8,8 @@ import (
 )
 
 func TestDetect(t *testing.T) {
+	t.Skip("AppVeyor runs test with make/sh")
+
```
cfergeau, line 12: Is there any way to detect that we are running in appveyor? Someone could be developping on Windows and running `make test` there, and they'd get messages about AppVeyor?

pkg/os/shell/shell_windows_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDetect(t *testing.T) {
+	defer func(shell string) { os.Setenv("SHELL", shell) }(os.Getenv("SHELL"))
+	os.Setenv("SHELL", "")
+
+	shell, err := Detect()
+
+	assert.Equal(t, "powershell", shell)
```
praveenkumar, line 18: @guillaumerose I am thinking to use `Contains` here, this way we don't have to worry if it is powershell or cmd as long as it is part of supported shell?
```
assert.Contains(t, supportedShell, shell)
```

guillaumerose:
Funny, someone pinged me to merge this https://github.com/docker/machine/pull/4788#issuecomment-734809216

I guess we should include it :)

pkg/os/shell/shell_windows_test.go
guillaumerose: Good suggestion! I did this and remove the 2nd test. It was bringing more information.

PR #1670, opened 2020-11-12
Additional improvements around disk resizing, mostly for macos

cfergeau:
This PR adds an explicit error when the user tries disk resizing on macos.
It also removes annoying warnings when using `crc start` on this platform.

## Testing

2 main changes:
- `crc start --disk-size 32` fails on macos
- `crc start && crc stop && crc start` no longer shows warnings on macos

Checking that disk resizing/memory/vcpu changes did not regress could be useful as well.

### DIFF
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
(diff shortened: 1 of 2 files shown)


gbraad (approved):

pkg/crc/validation/validation.go
```diff
@@ -35,6 +36,11 @@ func ValidateDiskSize(value int) error {
 	if value < constants.DefaultDiskSize {
 		return fmt.Errorf("requires disk size in GiB >= %d", constants.DefaultDiskSize)
 	}
+	// https://github.com/code-ready/machine-driver-hyperkit/issues/18
+	if runtime.GOOS == "darwin" && value > constants.DefaultDiskSize {
+		return fmt.Errorf("Disk resizing is not supported yet on macOS")
```
gbraad, line 41: `Not yet` supported or `Not supported`

Yet might give an unrealistic idea about a timeline.

PR #1672, opened 2020-11-13
Docs: Update `odo` command with `--s2i` flag

robin-owen:
As reported here: https://github.com/openshift/odo/issues/4069


borod108:
@robin-owen tnx for this one, helped me especially since "--git --s2i" does not work while "--s2i --git" does so it is frustrating, really hope this gets merged soon :)

PR #1676, opened 2020-11-16
Connect ssh only once, and reuse the same connection

guillaumerose:
It removes plenty of ssh handshakes and should improve the startup time.
    
When crc change the private key, the connection with the previous key is
kept opened and still works.

### DIFF
--- a/pkg/crc/ssh/ssh_test.go
+++ b/pkg/crc/ssh/ssh_test.go
@@ -48,7 +48,8 @@ func TestRunner(t *testing.T) {
 	})
 
 	addr := listener.Addr().String()
-	runner := CreateRunner(ipFor(addr), portFor(addr), clientKeyFile)
+	runner, err := CreateRunner(ipFor(addr), portFor(addr), clientKeyFile)
+	assert.NoError(t, err)
 
 	bin, err := runner.Run("echo hello")
 	assert.NoError(t, err)
--- a/pkg/crc/ssh/ssh_test.go
+++ b/pkg/crc/ssh/ssh_test.go
@@ -36,7 +36,7 @@ func TestRunner(t *testing.T) {
 	require.NoError(t, err)
 	defer listener.Close()
 
-	createSSHServer(t, listener, clientKey, func(input string) (byte, string) {
+	totalConn := createSSHServer(t, listener, clientKey, func(input string) (byte, string) {
 		escaped := fmt.Sprintf("%q", input)
 		if escaped == `"echo hello"` {
 			return 0, "hello"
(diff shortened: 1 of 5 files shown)


cfergeau:
Can the ssh connection time out if it's idle for too long? Does the code deal with this?

guillaumerose:
> Can the ssh connection time out if it's idle for too long? 

There is no code in the client for that. The timeout could happen on the server side, or from the NAT. Both are unlikely to happen.

> Does the code deal with this?

No, first failure will be fatal.

cfergeau (changes requested):
Please no one-line logs on non-trivial changes...

pkg/crc/ssh/ssh_test.go
```diff
@@ -61,9 +62,12 @@ func TestRunner(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, "hello", stdout)
 	assert.Empty(t, stderr)
+
+	assert.Equal(t, 1, *totalConn)
 }
 
-func createSSHServer(t *testing.T, listener net.Listener, clientKey *rsa.PrivateKey, fun func(string) (byte, string)) {
+func createSSHServer(t *testing.T, listener net.Listener, clientKey *rsa.PrivateKey, fun func(string) (byte, string)) *int {
```
cfergeau, line 69: Could return just `int` I think?

cfergeau (commented):
Looks reasonable for our very specific usecase. Does it make it harder to switch  back to using an external ssh binary if needed though? The go-native ssh client hasn't been widely tested yet (it's new in 1.19.0), so I'd prefer to keep our options open for a little while longer.

pkg/crc/ssh/ssh_test.go
guillaumerose: this counter is not set directly but modified during the test by the ssh server. I need this pointer :/

guillaumerose:
> Does it make it harder to switch back to using an external ssh binary if needed though? 

No, the interface is still the same.

> The go-native ssh client hasn't been widely tested yet (it's new in 1.19.0), so I'd prefer to keep our options open for a little while longer.

Understand, we can wait 1-2 weeks (for 1.20.0?). 1.19.0 will be out for enough time I assume.

praveenkumar:
If we are targeting this change to be part of 1.20.0 release and it have all the required changes, I would like to push it right away so we have time till the release if something goes wrong.

cfergeau:
> If we are targeting this change to be part of 1.20.0 release and it have all the required changes, I would like to push it right away so we have time till the release if something goes wrong.

If there were critical bugs with the native-go ssh implementation, I guess we would have heard of it by now? Let's try to land this now and see how this goes.

cfergeau (approved):

PR #1703, opened 2020-11-19
[WIP] Monitoring

jsliacan:
Here to be discussed & changed. Possibly replaced by better approach & discarded. 

### Idea

Create a feature file which starts CRC and regularly calls `oc node describe`. Ouput of this and a couple of other commands is then pushed to https://github.com/jsliacan/crc-data. This job is run more-or-less perpetually on an Openstack instance. 

# Note
Can run this PR without worries, it doesn't have my github token so no spoiler uploads.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 SHELL := /bin/bash
 
 BUNDLE_VERSION = 4.6.6
+BUNDLE_VERSION = 4.6.3
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.19.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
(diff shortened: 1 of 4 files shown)


guillaumerose:
I would prefer a new small go program to do that (and you can even remove clicumber with that, far more simpler!). WDYT?

adrianriobo:
> I would prefer a new small go program to do that (and you can even remove clicumber with that, far more simpler!). WDYT?

Yes I would prefer have it independently, maybe it can be run from a scheduled container?, My initial thought, at least to get information (at least for ocp pods) with some format would be use something like crictl stats -o .., but an specific development also makes total sense

jsliacan:
Sure - sounds good. When I was writing that thing, I reused as much as I could (including jenkins jobs for running tests). We can change it to anything we want.

jsliacan:
Will do this differently - separate module.

PR #1714, opened 2020-11-23
Check if the user is root must be the first check and for all commands

guillaumerose:
It needs to be done before anything (esp. creating the base directory).
`crc daemon`, `crc config`, etc. must be run as user.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
@@ -40,6 +41,9 @@ var (
 )
 
 func init() {
+	if err := checkIfRunningAsNormalUser(); err != nil {
+		logging.Fatal(err.Error())
+	}
 	if err := constants.EnsureBaseDirectoriesExist(); err != nil {
 		logging.Fatal(err.Error())
 	}
@@ -147,4 +151,12 @@ func newMachine() machine.Client {
 
 func newMachineWithConfig(config crcConfig.Storage) machine.Client {
 	return machine.NewClient(constants.DefaultName, isDebugLog(), network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString()))
+}
+
+func checkIfRunningAsNormalUser() error {
+	if os.Geteuid() != 0 {
+		return nil
+	}
+	logging.Debug("Ran as root")
+	return fmt.Errorf("crc should be ran as a normal user")
 }
--- a/pkg/crc/preflight/preflight_checks_nonwin.go
+++ b/pkg/crc/preflight/preflight_checks_nonwin.go
@@ -11,24 +11,6 @@ import (
 	crcos "github.com/code-ready/crc/pkg/os"
 )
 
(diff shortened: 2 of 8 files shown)


cmd/crc/cmd/root.go
```diff
@@ -148,3 +152,11 @@ func newMachine() machine.Client {
...
 	return machine.NewClient(constants.DefaultName, isDebugLog(), network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString()))
 }
+
+func checkIfRunningAsNormalUser() error {
+	if os.Geteuid() != 0 {
+		return nil
+	}
+	logging.Debug("Ran as root")
+	return fmt.Errorf("crc should be ran as a normal user")
+}
```
praveenkumar, line 162: Note: Till now we have skip option for the preflight checks even for root user, with this change, it will be not available anymore. I am not sure we really required to have skip functionality for linux and mac.

praveenkumar:
CI need to be fixed since as part of setup we check `And stderr should contain "Checking if running as non-root"` which need to adjusted or removed.

guillaumerose:
CI fixed, thanks!

guillaumerose:
I just realize I need to double check on Windows.

guillaumerose:
Works fine on Windows, os.Geteuid() returns -1.

cmd/crc/cmd/root.go
cfergeau: This is one of the `skip-` options which might be used by some people, so yeah, not sure it's a good thing to remove it.

guillaumerose: so we should remove this check totally ?

praveenkumar: @guillaumerose we added this check to force user to use crc as normal user mode instead admin/root, let's discuss this if we really want to remove it.

praveenkumar: https://github.com/code-ready/crc/pull/745 some discussion in the past when we made it skip-able I think it was because of windows.

cfergeau: If the main problem is creating `~/.crc` with the wrong permissions, workaround could be to check if running as root and if  $HOME is `/root` or `/var/root` (on macOS). Another one could be to have an additional preflight check making sure ~/.crc is readable/writable the current user and fixing it.

anjannath:
There's another is-admin check on windows, you may want to remove that as well? https://github.com/code-ready/crc/blob/cc9db2fcceaf8c94a64f865e8a1202d27aaf9513/pkg/crc/preflight/preflight_windows.go#L16

cmd/crc/cmd/root.go
anjannath: This is removing the check from preflight so the `skip` config will be gone, maybe we can make it skippable by using some environment variables `CRC_SKIP_ROOT_CHECK=true crc` so that users have a way to skip it if they want..

guillaumerose:
Any strong opinion on this ? Otherwise I am going to close it. It is not something essential. Just putting the root check up front and moving the mkdirAll of the base dir could be enough.

cfergeau:
> Any strong opinion on this ? Otherwise I am going to close it. It is not something essential. Just putting the root check up front and moving the mkdirAll of the base dir could be enough.

As this is one of the few config options I disable, I'd rather keep it ;)

PR #1715, opened 2020-11-23
Update bundle from 4.6.3 to 4.6.6

praveenkumar:
**Fixes:** Issue #N

## Solution/Idea

## Proposed changes

1. Add `start` command.
3. ...

## Testing

7. ...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 SHELL := /bin/bash
 
-
-BUNDLE_VERSION = 4.6.3
+BUNDLE_VERSION = 4.6.6
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.19.0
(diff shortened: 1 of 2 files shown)


guillaumerose (approved):

cfergeau (commented):
Why 4.6.5 (which http://mirror.openshift.com/pub/openshift-v4/clients/ocp/candidate/ points at) rather than 4.6.4, which is http://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/ ?
I don't mind either way, but good to know if it's because you want CI to test the latest, or because we *must* have at least this version for the next release, ...
If we upgrade to 4.6.5 now, and nothing changes on mirror.openshift.com before next crc release, I'd be in favour of downgrading to 4.6.4 for the release.

praveenkumar:
test this please

PR #1716, opened 2020-11-23
[WIP] extract embed bundle directly from crc, remove intermediate bundle on disk

guillaumerose:
It saves ~2GB of disk space for the end user.

Requires https://github.com/YourFin/binappend/pull/1 - this is only to not break the compatibility with an eventual zip bundle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION = 4.6.3
+BUNDLE_VERSION = 4.6.1
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.19.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
--- a/go.mod
+++ b/go.mod
@@ -84,3 +84,5 @@ replace (
 replace vbom.ml/util => github.com/fvbommel/util v0.0.2
 
 replace golang.org/x/sys => golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6
+
+replace github.com/YourFin/binappend => github.com/guillaumerose/binappend v0.0.0-20201123104445-14e342367bb5
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/embed"
 	"github.com/code-ready/crc/pkg/extract"
 	crcos "github.com/code-ready/crc/pkg/os"
 )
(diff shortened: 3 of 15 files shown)


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -47,7 +47,6 @@ var bundleExtractionCheck = Check{
 	check:            checkBundleExtracted,
 	fixDescription:   "Extracting bundle from the CRC executable",
 	fix:              fixBundleExtracted,
-	flags:            SetupOnly,
```
cfergeau: When running during `start`, only `checkBundleExtracted` will be run, and `start` will fail if `checkBundleExtracted` fails, `fixBundleExtracted` won't be run during `start`. Not sure this is the behaviour we want.

Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION = 4.6.3
+BUNDLE_VERSION = 4.6.1
```
cfergeau, line 3: Seems unwanted

go.mod
```diff
@@ -84,3 +84,5 @@ replace (
 replace vbom.ml/util => github.com/fvbommel/util v0.0.2
 
 replace golang.org/x/sys => golang.org/x/sys v0.0.0-20200826173525-f9321e4c35a6
+
+replace github.com/YourFin/binappend => github.com/guillaumerose/binappend v0.0.0-20201123104445-14e342367bb5
```
cfergeau, line 88: This will need to go as well before merging.

cfergeau (commented):
Last 2 commits are missing commit logs.

go.mod
gbraad: Any update on this?

pkg/crc/preflight/preflight_linux.go
```diff
@@ -190,7 +190,7 @@ func getPreflightChecksForDistro(distro linux.OsType, networkMode network.Mode)
 		}
 	}
 
-	return checks
+	return append(checks, bundleExtractionCheck)
```
gbraad, line 193: so we don't break consistency with the platform files

```
	checks = append(checks, bundleExtractionCheck)

	return checks
```

go.mod
guillaumerose: No news after 2 weeks. I believe it will never be merged :/

cfergeau: Still no progress on this one :(

pkg/crc/machine/bundle/metadata.go
```diff
@@ -128,7 +130,10 @@ func ExtractFromExecutable(sourcepath string) (*CrcBundleInfo, error) {
 		return nil, err
 	}
 
-	return GetCachedBundleInfo(filepath.Base(sourcepath))
+	if err := writeExtractionDone(bundleName); err != nil {
+		return nil, err
+	}
+	return GetCachedBundleInfo(bundleName)
 }
```
cfergeau, line 137: Fwiw, the code duplication between `ExtractFromExecutable` and `Extract` seems to be avoidable if you add an `extractFromReader` helper, and open the file in `Extract` rather than opening it in `extract.Uncompress`

guillaumerose:
With https://github.com/code-ready/crc/pull/1931, we will be able to delete the intermediate .crcbundle on the disk.

Issue #1705, opened 2020-11-19
systemd code does not detect the state of 'socket' units properly


cfergeau:
There is a bug report about issues starting crc when libvirtd socket activation is in use. libvirtd.socket is enabled, libvirtd.service is not enabled, and is shutting down when idle. And `crc start` fails to detect that `libvirtd.socket` is running. This forces the user to run `crc setup` every time (which will restart `libvirtd`). 
It turns out that `systemctl status libvirtd.socket` returns:
```
● libvirtd.socket - Libvirt local socket
Loaded: loaded (/usr/lib/systemd/system/libvirtd.socket; enabled; vendor preset: disabled)
Active: active (listening) since Thu 2020-11-19 18:41:01 +04; 7min ago
Listen: /run/libvirt/libvirt-sock (Stream)
Tasks: 0 (limit: 100800)
Memory: 0B
CGroup: /system.slice/libvirtd.socket
```
The important part is `Active: active (listening)`, our code only parses `active (running)`, so `checkLibvirtServiceRunning` never detects that `libvirtd.socket` is started, and so the preflight fails.

PR #1721, opened 2020-11-24
Improve/fix systemd state parsing

cfergeau:
The current code does not handle correctly stopped/not found units, and it also does not support systemd socket units.
This PR will fix this. It starts with preparatory work so that local command runners and ssh command runners behave the same
(separate stdout and stderr), since at the moment local command runners use separate output streams, while ssh command runners
combine stdout and stderr.

**Fixes:** Issue #1705

## Solution/Idea

This fixes the 'libvirtd is running' preflight check on systems using socket activation for libvirtd (libvirtd.socket is `active (listening`) and libvirtd.service is stopped after a few minutes of inactivity).

## Testing

This should be testable on rhel 8.3 by enabling/starting libvirtd.socket, stopping libvirtd.service, and running `crc setup`, the 'check-libvirt-running' test should succeed without any attempt at fixing it up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,5 @@ before_script:
   - sudo apt-get -qq -y install podman
 
 script:
-- make
-- make vendorcheck
-- make cross-lint
-- make cross
-- make test
-- make build_integration
+- make check
 - make BUNDLE_DIR=/tmp MOCK_BUNDLE=true release
--- a/pkg/crc/ssh/client.go
+++ b/pkg/crc/ssh/client.go
@@ -1,6 +1,7 @@
 package ssh
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -12,7 +13,7 @@ import (
 )
 
 type Client interface {
-	Output(command string) (string, error)
+	Run(command string) ([]byte, []byte, error)
 	Close()
 }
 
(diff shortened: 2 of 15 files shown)


test/integration/crcsuite/collect.go
```diff
@@ -188,7 +188,7 @@ func (collector *VMCommandCollector) Collect(w Writer) error {
 	if err != nil {
 		return err
 	}
-	out, err := ssh.Output(collector.Command)
+	out, _, err := ssh.Run(collector.Command)
 	if err != nil {
```
cfergeau, line 192: These changes in `collect.go` belong in the previous commit, I've already fixed it locally.

praveenkumar:
/lgtm

https://github.com/code-ready/crc/pull/1721#discussion_r530353550 @cfergeau Are you going to force push that?

pkg/crc/ssh/client.go
```diff
@@ -85,19 +86,24 @@ func (client *NativeClient) session() (*ssh.Client, *ssh.Session, error) {
 	return conn, session, err
 }
 
-func (client *NativeClient) Output(command string) (string, error) {
+func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
 	conn, session, err := client.session()
 	if err != nil {
-		return "", err
+		return []byte{}, []byte{}, err
```
guillaumerose, line 92: `return nil, nil, err` is also fine.

pkg/crc/systemd/states/state.go
```diff
@@ -9,6 +9,7 @@ type State int
 const (
 	Unknown State = iota
 	Running
+	Listening
```
guillaumerose, line 12: We should use a `string` instead of `iota`. It would simplify this code a lot.

`type State string`

guillaumerose (approved):

cfergeau:
> /lgtm
> 
> [#1721 (comment)](https://github.com/code-ready/crc/pull/1721#discussion_r530353550) @cfergeau Are you going to force push that?

Yes I will, I was expecting to need some changes, and to push everything at once :)

pkg/crc/ssh/client.go
cfergeau: Fixed.

test/integration/crcsuite/collect.go
cfergeau: Fixed in the latest push.

pkg/crc/systemd/states/state.go
cfergeau: Not a big fan of using strings for enums, but maybe it's coming from my C background.

.travis.yml
```diff
@@ -12,10 +12,5 @@ before_script:
   - sudo apt-get -qq -y install podman
 
 script:
-- make
-- make vendorcheck
-- make cross-lint
-- make cross
-- make test
-- make build_integration
+- make check
```
praveenkumar, line 15: shouldn't this added to other CI as well like circleCI, appveyor and centos one?

praveenkumar (approved):

Issue #1328, opened 2020-06-22
[BUG] DNS records are not cleaned up in /etc/hosts after `crc cleanup`


guillaumerose:
### General information

  * OS: Linux & macOS
  * Hypervisor: KVM / hyperkit
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop

### Steps to reproduce

  1. crc setup
  2. crc cleanup
  3. check /etc/hosts, custom dns records are still present.

guillaumerose:
This is in  master for a while now.

PR #1722, opened 2020-11-25
Use admin-helper instead of goodhosts

guillaumerose:
**Fixes:** Issue #1328

It uses https://github.com/code-ready/admin-helper instead of goodhosts CLI.
I introduced a new command called clean that remove all entries by a domain suffix. It is enough to clean up what was added by crc.

### DIFF
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -38,6 +38,9 @@ const (
 	VsockSSHPort = 2222
 
 	OkdPullSecret = `{"auths":{"fake":{"auth": "Zm9vOmJhcgo="}}}` // #nosec G101
+
+	ClusterDomain = ".crc.testing"
+	AppsDomain    = ".apps-crc.testing"
 )
 
 var podmanURLForOs = map[string]string{
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -29,7 +29,7 @@ const (
 	DaemonLogFile             = "crcd.log"
 	CrcLandingPageURL         = "https://cloud.redhat.com/openshift/install/crc/installer-provisioned" // #nosec G101
 	DefaultPodmanURLBase      = "https://storage.googleapis.com/libpod-master-releases"
-	DefaultGoodhostsCliBase   = "https://github.com/code-ready/goodhosts-cli/releases/download/v1.0.0"
+	DefaultAdminHelperCliBase = "https://github.com/code-ready/admin-helper/releases/download/0.0.1"
 	CRCMacTrayDownloadURL     = "https://github.com/code-ready/tray-macos/releases/download/v%s/crc-tray-macos.tar.gz"
 	CRCWindowsTrayDownloadURL = "https://github.com/code-ready/tray-windows/releases/download/v%s/crc-tray-windows.zip"
 	DefaultContext            = "admin"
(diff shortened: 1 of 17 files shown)


pkg/crc/constants/constants.go
```diff
@@ -38,6 +38,9 @@ const (
 
 	VSockGateway = "192.168.127.1"
 	VsockSSHPort = 2222
+
+	BaseDomain = ".crc.testing"
+	AppsDomain = ".apps-crc.testing"
```
praveenkumar, line 43: We usually take this from the bundle info instead having it in constants.

pkg/crc/machine/machine.go
```diff
@@ -67,6 +67,12 @@ func getCrcBundleInfo(bundlePath string) (*bundle.CrcBundleInfo, error) {
 	bundleName := filepath.Base(bundlePath)
 	bundleInfo, err := bundle.GetCachedBundleInfo(bundleName)
 	if err == nil {
+		if fmt.Sprintf(".%s", bundleInfo.ClusterInfo.AppsDomain) != constants.AppsDomain {
+			return nil, fmt.Errorf("unexpected bundle, it must have %s apps domain", constants.AppsDomain)
+		}
+		if bundleInfo.GetAPIHostname() != fmt.Sprintf("api%s", constants.BaseDomain) {
+			return nil, fmt.Errorf("unexpected bundle, it must have %s base domain", constants.BaseDomain)
+		}
```
praveenkumar, line 75: Ah, this is just validating the whatever we are putting on the bundle should be same what we put in the constants. Now if we ever want to change the snc side then here also we do need to change :( wouldn't it better to have this info at one place where the bundle is created?

guillaumerose: An alternative would be to browse all installed bundles to get all base domains and apps domains, but this is code doesn't exist yet! WDYT?

cfergeau: I was thinking we'd add some markers in `/etc/hosts` in order to do the cleanup?
```
### These entries were automatically by CodeReady Containers (crc)
....
### End of entries automatically added by crc
```
We'll need to hardcode `*.crc.testing` during the transition.

guillaumerose: If I remove these 2 checks but keep for now hardcoded the domain for the clean, would you accept it for the next release ? 
For sure, we need to add markers but I would prefer not to do a big bang.

Meanwhile, I try to improve the code in bundle package. It could help us for that.

gbraad:
Renamed to WIP as scope of Admin-helper is of yet not clearly defined

guillaumerose:
The scope of this PR is really limited. It is just an upgrade of goodhosts CLI + a new sub command. It doesn't engage us too much. It's not WIP and the user benefit is great: no more extra entries in the hosts file after crc cleanup.

pkg/crc/constants/constants.go
guillaumerose: I will open a follow up to put everything in common. These constants are also used in crc setup for the dns part.

guillaumerose:
/unhold

Rebased! PTAL.

pkg/crc/constants/constants.go
cfergeau: The end-goal is to kill them, not to make use of them everywhere :)

guillaumerose:
retest this please

guillaumerose:
retest this please

Issue #1724, opened 2020-11-27
[Integration] Switch to quay or fedora registry instead docker


praveenkumar:
Docker recently limited the pulls for anonymous user and something the integration tests fails because of that.
 
```
21:52:06     command 'sudo podman build -t hello:test -f Dockerfile', expected to succeed, exited with exit code: 125
21:52:06 Command stdout: STEP 1: FROM centos:latest
21:52:06 
21:52:06 Command stderr: time="2020-11-26T16:21:36Z" level=error msg="HEADER map[Cache-Control:[no-cache] Content-Type:[application/json]]"
21:52:06 time="2020-11-26T16:21:38Z" level=error msg="HEADER map[Cache-Control:[no-cache] Content-Type:[application/json]]"
21:52:06 time="2020-11-26T16:21:42Z" level=error msg="HEADER map[Cache-Control:[no-cache] Content-Type:[application/json]]"
21:52:06 time="2020-11-26T16:21:50Z" level=error msg="HEADER map[Cache-Control:[no-cache] Content-Type:[application/json]]"
21:52:06 Error: error creating build container: The following failures happened while trying to pull image specified by "centos:latest" based on search registries in /etc/containers/registries.conf:
21:52:06 * "localhost/centos:latest": Error initializing source docker://localhost/centos:latest: error pinging docker registry localhost: Get https://localhost/v2/: dial tcp [::1]:443: connect: connection refused
21:52:06 * "registry.access.redhat.com/centos:latest": Error initializing source docker://registry.access.redhat.com/centos:latest: Error reading manifest latest in registry.access.redhat.com/centos: name unknown: Repo not found
21:52:06 * "registry.redhat.io/centos:latest": Error initializing source docker://registry.redhat.io/centos:latest: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication
21:52:06 * "docker.io/library/centos:latest": Error initializing source docker://centos:latest: Error reading manifest latest in docker.io/library/centos: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

gbraad:
As you can see multiple registries are specified, but the can't satisfy the requested build.

[...]

PR #1726, opened 2020-11-27
Issue #1724: Do not use Dockerhub centos images

jsliacan:
**Fixes:** Issue #1724 

## Solution

Modify the Dockerfile to pull specifically from quay.io. The image we're pulling is a mirror of the upstream CentOS repository (probably along the lines of: https://www.openshift.com/blog/mitigate-impact-of-docker-hub-pull-request-limits). 

## Testing

1. The restriction of integration tests to `story_registry` feature should pass OK.

### DIFF
--- a/test/testdata/Dockerfile
+++ b/test/testdata/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:latest
+FROM registry.centos.org/centos/centos:latest
 
 RUN groupadd -r crcek -g 1234 && useradd -u 1234 -r -g crcek -m -d /crcek -s /sbin/nologin -c "CRC user" crcek && chmod 755 /crcek
 
--- a/images/openshift-ci/Dockerfile
+++ b/images/openshift-ci/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /go/src/github.com/code-ready/crc
 COPY . .
 RUN make cross
 
-FROM centos:7
+FROM registry.centos.org/centos/centos:7
 COPY --from=builder /go/src/github.com/code-ready/crc /opt/crc
 COPY --from=builder /go/src/github.com/code-ready/crc/out/linux-amd64/crc /bin/crc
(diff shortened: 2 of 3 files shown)


test/testdata/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:latest
+FROM quay.io/app-sre/centos:latest
```
cfergeau, line 1: Wondering if it would make sense to use https://registry.centos.org/centos/centos ? I don't know which one will be maintained better?

jsliacan: I don't really mind. I can swap it if it's preferred, since it does look more official :)

gbraad: We rely on quay.io already from the cluster, so this seems to not introduce another unknown into the tests.

guillaumerose:
there is also 2 other uses of centos images (see /images directory)

test/testdata/Dockerfile
cfergeau: My point was not about `quay.io`, but rather about how well `app-sre/centos:latest` is maintained. If it's updated once in a blue moon, it's not a great image to depend on.

jsliacan: Updated this and other two uses of centos images with the centos' own registry. Checked if it works in case of the tests, but no the other two cases.

guillaumerose:
/lgtm
retest this please

anjannath:
Need to update commit message to _Pull from centos registry not Dockerhub_

anjannath (approved):

gbraad (approved):

test/testdata/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:latest
+FROM registry.centos.org/centos/centos:latest
```
anjannath, line 1: To save space, we can use a smaller image like alpine linux..

gbraad: it is about a consistent scenario. images like these can be 'cached' and tagged beforehand so space should not be an issue

PR #1730, opened 2020-11-30
Add unit tests to bundle package

guillaumerose:
Before doing a larger refactoring, add unit tests.


pkg/crc/machine/bundle/metadata.go
```diff
@@ -19,39 +19,56 @@ import (
...
+type DiskImage struct {
+	Name     string `json:"name"`
+	Format   string `json:"format"`
+	Size     string `json:"size"`
+	Checksum string `json:"sha256sum"`
+}
+
+type DriverInfo struct {
+	Name string `json:"name"`
 }
```
cfergeau, line 66: This hunk would make more sense in a separate commit after the commit adding the unit test, as this is exactly the kind of change where we need the unit test to make sure we did not break anything.

pkg/crc/machine/bundle/repository.go
```diff
@@ -0,0 +1,78 @@
+package bundle
+
```
cfergeau, line 2: Is this code movement only? are there expected  behaviour changes? Would be useful to know in the commit log :)

pkg/crc/machine/bundle/repository.go
```diff
@@ -39,26 +39,47 @@ func (bundle *CrcBundleInfo) readBundleInfo() error {
...
-	if err := bundleInfo.installSymlinkOrCopy(); err != nil {
+	return &bundleInfo, nil
+}
+
+func (r *Repository) Use(bundleName string) (*CrcBundleInfo, error) {
+	bundleInfo, err := r.Get(bundleName)
+	if err != nil {
+		return nil, err
+	}
+	if err := bundleInfo.installSymlinkOrCopy(r.OcBinDir); err != nil {
```
cfergeau, line 65: This is pre-existing, but this method name is missing a hint that it is about the `oc` binary.

pkg/crc/machine/bundle/repository.go
```diff
@@ -89,11 +81,23 @@ func (bundle *CrcBundleInfo) installSymlinkOrCopy(ocBinDir string) error {
...
+func (r *Repository) Extract(path string) error {
+	_, err := extract.Uncompress(path, r.CacheDir, true)
+	return err
+}
+
+var defaultRepo = &Repository{
+	CacheDir: constants.MachineCacheDir,
+	OcBinDir: constants.CrcOcBinDir,
+}
+
```
cfergeau, line 93: The introduction of `defaultRepo` could have been squashed in the previous commit rather than here.

pkg/crc/machine/bundle/repository.go
```diff
@@ -89,11 +81,23 @@ func (bundle *CrcBundleInfo) installSymlinkOrCopy(ocBinDir string) error {
 	return os.Symlink(ocInBundle, ocInBinDir)
 }
 
-func Extract(sourcepath string) (*CrcBundleInfo, error) {
-	_, err := extract.Uncompress(sourcepath, constants.MachineCacheDir, true)
-	if err != nil {
+func (r *Repository) Extract(path string) error {
```
cfergeau, line 84: Wondering if this would be better as `func (repo *Repository)`? I usually have a hard time with one or two letter variable names.

cfergeau (commented):
A high level summary of what the end goal is with the new `Repository` type could be useful?

pkg/crc/machine/bundle/metadata.go
guillaumerose: I tried but the test change completely because the struct also change.
We don't break anything here because in the unit test, I take the raw json, unmarshall, marshall and compare to the raw json given as input.

pkg/crc/machine/bundle/repository.go
guillaumerose: done

pkg/crc/machine/bundle/repository.go
guillaumerose: Only move, added comment.

pkg/crc/machine/bundle/repository.go
guillaumerose: It was only once in the previous commit, the commit was already large, no need for it ;)

pkg/crc/machine/bundle/metadata.go
cfergeau: It depends where you got the initial raw json from.  If it's the sample one from snc, then it's good, if it's one which was generated by a previous run of the test case, it's less good :)

pkg/crc/machine/bundle/repository.go
guillaumerose: done also :)

pkg/crc/machine/bundle/metadata.go
guillaumerose: Took it from my local running crc installation. So, from the snc release CI.

cfergeau (approved):

PR #1647, opened 2020-11-06
Add basic Ubuntu support

guillaumerose:
Only what is needed if used with vsock

```
guillaumerose@gurose-crc-ubuntu:~$ ./crc setup
INFO Checking if podman remote executable is cached 
INFO Checking if goodhosts executable is cached   
INFO Caching goodhosts executable                 
INFO Will use root access: change ownership of /home/guillaumerose/.crc/bin/goodhosts 
INFO Will use root access: set suid for /home/guillaumerose/.crc/bin/goodhosts 
INFO Checking if CRC bundle is cached in '$HOME/.crc' 
INFO Checking minimum RAM requirements            
INFO Checking if running as non-root              
INFO Checking if Virtualization is enabled        
INFO Checking if KVM is enabled                   
INFO Checking if libvirt is installed             
INFO Installing libvirt service and dependencies  
INFO Will use root access: install virtualization related packages 
INFO Checking if user is part of libvirt group    
INFO Adding user to libvirt group                 
INFO Will use root access: add user to libvirt group 
User guillaumerose added to libvirt group. Please log out and login to get access to libvirt daemon

... logout / login ...

guillaumerose@gurose-crc-ubuntu:~$ ./crc setup
INFO Checking if podman remote executable is cached 
INFO Checking if goodhosts executable is cached   
INFO Checking if CRC bundle is cached in '$HOME/.crc' 
INFO Checking minimum RAM requirements            
INFO Checking if running as non-root              
INFO Checking if Virtualization is enabled        
INFO Checking if KVM is enabled                   
INFO Checking if libvirt is installed             
INFO Checking if user is part of libvirt group    
INFO Checking if libvirt daemon is running        
INFO Checking if a supported libvirt version is installed 
INFO Checking if crc-driver-libvirt is installed  
INFO Installing crc-driver-libvirt                
INFO Checking for obsolete crc-driver-libvirt     
INFO Checking if libvirt 'crc' network is available 
INFO Setting up libvirt 'crc' network             
INFO Checking if libvirt 'crc' network is active  
INFO Starting libvirt 'crc' network               
INFO Checking if AppArmor is configured           
INFO Updating AppArmor configuration              
INFO Will use root access: updating AppArmor configuration 
INFO Will use root access: Changing permission for /etc/apparmor.d/libvirt/TEMPLATE.qemu to 0644  
Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster
```

Issue #1624, opened 2020-10-29
Document a possible workaround for Fedora 33


gbraad:
Ref: #1572

gbraad:
@cfergeau did you look into the issue? Do we still need to document this?

cfergeau:
Praveen has been testing this change:
```
# cat /etc/systemd/resolved.conf.d/crc.conf 
[Resolve]
DNS=192.168.130.11
Domains=~. testing
# resolvectl dns
Global: 192.168.130.11
Link 2 (eth0): 10.11.142.1 10.11.5.19
# resolvectl domain
Global: ~. testing
```
with the hope that it would be generic enough to work both on f33 and ubuntu. This could also be handled by following https://fedoramagazine.org/systemd-resolved-introduction-to-split-dns/ , but Praveen was worried this would not work on Ubuntu which uses systemd-networkd. https://manpages.ubuntu.com/manpages/cosmic/man1/resolvectl.1.html documents that when systemd-networkd is in use, per-interface nameservers must be set in the network interface file, while when it's not in use, it's done in a different place.

I did not test any of this myself.

gbraad:
Do we report this currently as a known issue? An automated solution for this doesnt exist yet

robin-owen:
@gbraad This isn't recorded as a known issue at this time. We can add it to the known issues asynchronously, however.

cfergeau:
> Praveen has been testing this change:
> 
> ```
> # cat /etc/systemd/resolved.conf.d/crc.conf 
> [Resolve]
> DNS=192.168.130.11
> Domains=~. testing
> # resolvectl dns
> Global: 192.168.130.11
> Link 2 (eth0): 10.11.142.1 10.11.5.19
> # resolvectl domain
> Global: ~. testing
> ```
> 
> with the hope that it would be generic enough to work both on f33 and ubuntu. This could also be handled by following https://fedoramagazine.org/systemd-resolved-introduction-to-split-dns/ , but Praveen was worried this would not work on Ubuntu which uses systemd-networkd. https://manpages.ubuntu.com/manpages/cosmic/man1/resolvectl.1.html documents that when systemd-networkd is in use, per-interface nameservers must be set in the network interface file, while when it's not in use, it's done in a different place.
> 
> I did not test any of this myself.

[...]

Issue #1572, opened 2020-10-05
Implement preflights for  f33


cfergeau:
Fedora 33 is using systemd-resolved by default https://fedoraproject.org/wiki/Changes/systemd-resolved 
We need to test how this impacts crc DNS setup

childofthewired:
Internal Wildcard DNS setup via DNSMasq is ends up broken.

cfergeau:
https://fedoramagazine.org/systemd-resolved-introduction-to-split-dns/ might be useful.

dgolovin:
@praveenkumar I have the same issue, just installed Fedora 33 and crc seems to run fine OOTB, but nothing can be deployed to the cluster with odo. Will have to swithc back to deb until this is fixed.

cfergeau:
Can you be a bit more specific about 'nothing can be deployed to the cluster with odo'? I  would expect that application deployment goes fine with odo, but then the URL to which you deployed the application would not be reachable. 
So far I've only tested f33 up to "the cluster starts up and `oc get co` is good", I'll have to run more tests with odo.

The goodhosts fallback code is apparently enough to get a cluster working reasonably well (`oc get co` is fine, but `WARN Wildcard DNS resolution for apps-crc.testing does not appear to be working` is shown during startup)

childofthewired:
@cfergeau after deploying CRC we are able to log into the console with "crc console" and we can deploy an application in the UI, but the wildcard DNS is not setup properly for the route.

When logging in from the console, we can 'oc get project' and get a list of our projects including the ones that we created, and all the objects in it.

cfergeau:
ok, if it's only wildcard DNS which is not working, this is good news as this was the part most likely to break. This can be worked around by adding manually entries in /etc/hosts if needed, though Praveen is working on a better fix.

jlyle:
> @cfergeau after deploying CRC we are able to log into the console with "crc console" and we can deploy an application in the UI, but the wildcard DNS is not setup properly for the route.
> 
> When logging in from the console, we can 'oc get project' and get a list of our projects including the ones that we created, and all the objects in it.

@cfergeauI I am working with @childofthewired  and added the application name and ip of the cluster to my /etc/hosts and it resolved properly, thank you

cfergeau:
An alternative which worked fine for me is:
```
sudo resolvectl domain crc ~testing
sudo resolvectl dns crc 192.168.130.11
sudo resolvectl default-route crc false
```

[...]

PR #1732, opened 2020-11-30
Add fedora 33 support

cfergeau:
**Fixes:** Issue #1572

**Relates to:** Issue #1624

## Solution/Idea

This adds support for NetworkManager + systemd-resolved as used by fedora 33 (and ubuntu desktop).
This makes use of the per-distro preflight checks which were introduced during last sprint so that older distro releases can keep using NM + dnsmasq while newer distros will use NM + systemd-resolved.
Some limitations of this series:
- ubuntu support is not enabled, it can be added after #1647 lands
- this does not check if systemd-resolved is listening on port 53, this can be configured through /etc/systemd/resolved.conf
- there is no support for systemd-networkd + systemd-resolved (no NetworkManager) setups
- the selection of systemd-resolved/dnsmasq is static depending on the detected distro

## Proposed changes

1. Modifies `crc setup` to run different preflight checks depending on the distro which is in use (fedora 33 VS others)

## Testing

1. `crc setup` DNS configuration should not regress on previously working distros (fedora <= 32, RHEL8)
2. `crc setup` DNS configuration should work on fedora 33, including wildcard DNS (`anything.apps-crc.testing`)


cfergeau:
(since they conflict quite a bit, https://github.com/cfergeau/crc/tree/f33-ubuntu has a version of this PR rebased on top of https://github.com/code-ready/crc/pull/1647 )

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+const (
+	crcDnsmasqConfigFile        = "crc.conf"
+	crcNetworkManagerConfigFile = "crc-nm-dnsmasq.conf"
+	crcNMDispatcherFile         = "99-crc.sh"
+)
+
+var (
+	crcDnsmasqConfigPath = filepath.Join(string(filepath.Separator), "etc", "NetworkManager", "dnsmasq.d", crcDnsmasqConfigFile)
```
gbraad, line 71: extract `NMConfigRoot` as `filepath.Join(string(filepath.Separator), "etc", "NetworkManager")` as this is beinng re-used serveral times

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -0,0 +1,292 @@
...
+		checkDescription:   fmt.Sprintf("Checking if %s exists", crcNMDispatcherPath),
+		check:              checkCrcNMDispatcherFile,
+		fixDescription:     "Writing NetworkManager dispatcher file for crc",
+		fix:                fixCrcNMDispatcherFile,
+		cleanupDescription: fmt.Sprintf("Removing %s file", crcNMDispatcherPath),
+		cleanup:            removeCrcNMDispatcherFile,
+	},
+}
+
+func checkNMConfigFile(path string, expectedContent string) error {
```
gbraad, line 119: Generic file-exists-and-matches-content check, might be better of in a different pkg.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -166,22 +166,18 @@ func getPreflightChecksForDistro(distro *linux.OsRelease, networkMode network.Mo
...
-			checks = append(checks, dnsmasqPreflightChecks[:]...)
-		}
 	default:
-		logging.Warnf("distribution-specific preflight checks are not implemented for %s", distroID(distro))
-		if networkMode == network.DefaultMode {
-			checks = append(checks, nmPreflightChecks[:]...)
-			checks = append(checks, dnsmasqPreflightChecks[:]...)
-		}
+		logging.Warnf("distribution-specific preflight checks are not implemented for '%s'", distroID(distro))
+		fallthrough
```
praveenkumar, line 178: I don't get this one, why we want to go to other part of switch if we don't support or implemented the distribution specific checks?

praveenkumar:
@cfergeau we also need to adjust the integration test to incorporate the changes in the output messages.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -191,6 +195,17 @@ func commonChecks() []Check {
 	return checks
 }
 
+func usesSystemdResolved(osRelease *linux.OsRelease) bool {
+	switch distroID(osRelease) {
+	case linux.Ubuntu:
```
guillaumerose, line 206: never reached because usesSystemdResolved is only called if `case linux.RHEL, linux.CentOS, linux.Fedora:`

cfergeau: yes I know, it's one change less to enable it for ubuntu though. I can remove this for now if you prefer.

pkg/crc/preflight/preflight_linux.go
cfergeau: This keeps the pre-existing behaviour, which is how crc has more or less always behaved, we try to configure NM+dnsmasq regardless of the distro. Once we are a little bit smarter about distribution detection (use `ID_LIKE` rather than `ID`), we can indeed justu ignore it.

pkg/crc/preflight/preflight_linux.go
cfergeau: I'll remove it for now in the next iteration of this PR.

pkg/crc/preflight/preflight_checks_network_linux.go
cfergeau: Fixed.

pkg/crc/preflight/preflight_linux.go
cfergeau: I tried to improve the commit log.

pkg/crc/preflight/preflight_checks_network_linux.go
cfergeau: Good point, fits well with https://github.com/code-ready/crc/blob/cc9db2fcceaf8c94a64f865e8a1202d27aaf9513/pkg/os/util.go#L64

cfergeau:
> @cfergeau we also need to adjust the integration test to incorporate the changes in the output messages.

I did not have time to look at this, I'll do it tomorrow, hopefully the CI run will help.

guillaumerose:
I think you can remove all the lines checking logs in integration test (`stderr should contain ...`). They bring very little value:
* we check already the number of preflight tests, so we will hardly miss one. 
* `And stderr should contain` makes you think log lines are in a particular order but it is not.
* logs are not something we deliver or something that need to be stable for the user.

guillaumerose (approved):

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return err
+	}
+	logging.Debug("NetworkManager configuration is good")
+	return nil
+}
+
+func fixCrcNetworkManagerConfig() error {
+	logging.Debug("Fixing NetworkManager configuration")
+	err := fixNMConfigFile(crcNetworkManagerConfigPath, crcNetworkManagerConfig, 0644)
```
gbraad, line 196: Not a huge fan of abbr. for tech or names.
Esp. if the file contains both `NM` + `NetworkManager`

Note: purposely abbreviate text in this comment to indicate the confusing nature. It is inconsistent

gbraad (changes requested):
Be consistent about naming things the same.

besides this, OK

pkg/crc/preflight/preflight_checks_network_linux.go
cfergeau: Fixed

cfergeau:
> I think you can remove all the lines checking logs in integration test (`stderr should contain ...`). They bring very little value:
> 
>     * we check already the number of preflight tests, so we will hardly miss one.
> 
>     * `And stderr should contain` makes you think log lines are in a particular order but it is not.
> 
>     * logs are not something we deliver or something that need to be stable for the user.

We could introduce a regression in one of the 'check' method, which would return the wrong result. The integration 
test could detect some of these regressions which would be harder to catch through unit tests.

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -15,6 +15,13 @@ import (
 )
 
 var nmPreflightChecks = [...]Check{
+	{
+		configKeySuffix:  "check-systemd-networkd-not-running",
```
praveenkumar, line 19: Shouldn't the `configKeySuffix` should be `check-systemd-networkd-running` as per description and then user can skip this check with config set?

cfergeau: yeah should work this way too, and it's a less awkward name, thanks!

PR #1734, opened 2020-12-01
Check bundle integrity more in depth

guillaumerose:
* Verify all files are present
* Verify the size of the image disk
* Verify version of the bundle

The version check is particularly interesting when we look at upgrade. Currently we exit if the bundle is different but we could ask the user to whether continue or delete the VM if the version is still supported.


pkg/crc/machine/bundle/repository.go
```diff
@@ -45,13 +41,13 @@ type Repository struct {
 }
 
 func (repo *Repository) Get(bundleName string) (*CrcBundleInfo, error) {
-	var bundleInfo CrcBundleInfo
-	bundleInfo.cachedPath = getCachedBundlePath(repo.CacheDir, bundleName)
-	if !bundleInfo.isCached() {
-		return nil, fmt.Errorf("could not find cached bundle info in %s", bundleInfo.cachedPath)
+	path := getCachedBundlePath(repo.CacheDir, bundleName)
+	if _, err := os.Stat(path); err != nil {
+		return nil, errors.Wrapf(err, "could not find cached bundle info in %s", path)
```
cfergeau, line 33: I'd prefer we stop using `errors.Wrapf` in new code as I don't think this integrates well with https://blog.golang.org/go1.13-errors (or am I wrong?)

cfergeau (approved):
Looks fine, some minor comments!

pkg/crc/machine/bundle/repository.go
guillaumerose: good to know! I didnt know. 

Apparently pkg/errors does the job: https://github.com/pkg/errors/blob/master/errors.go#L162

cfergeau: Ah indeed, got misled by the end of their README https://github.com/pkg/errors/blob/master/README.md#roadmap

guillaumerose:
Windows CI was complaining. Fixing it hopefully!

pkg/crc/machine/bundle/metadata.go
```diff
@@ -116,19 +118,44 @@ func (bundle *CrcBundleInfo) GetOpenshiftVersion() string {
...
+		bundle.GetInitramfsPath(),
+	}
+
+	for _, file := range files {
+		if file == "" {
+			continue
+		}
+		if _, err := os.Stat(file); err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("%s not found in bundle", filepath.Base(file))
```
cfergeau, line 138: This seems like a good candidate for a wrapped error?

cfergeau (approved):

cfergeau (approved):
Commit log of `Current crc supports >=1.0 <2.0 versions of snc ` might  be a good opportunity to document how we are planning to handle versioning.

pkg/crc/machine/bundle/repository_test.go
```diff
@@ -52,9 +53,27 @@ func TestExtract(t *testing.T) {
...
+	bundle, err := repo.Get(testBundle(t))
 	assert.NoError(t, err)
 	assert.Equal(t, "4.6.1", bundle.ClusterInfo.OpenShiftVersion)
+
+	assert.NoError(t, bundle.Verify())
+	_ = os.Remove(bundle.GetKubeConfigPath())
+	assert.EqualError(t, bundle.Verify(), "kubeconfig not found in bundle")
+}
+
+func testBundle(t *testing.T) string {
```
anjannath, line 67: 👍🏽

anjannath (approved):

pkg/crc/machine/bundle/metadata.go
guillaumerose: I wanted to put a more friendly error otherwise the user would see "stat /longpath/file: no such file or directory".

cfergeau: Friendly error is very nice, but information was lost on the way ("file was missing"), which could be useful to users of this API. We probably should start adding a `UserFriendlyError { friendlyMessage string, userUnfriendlyError error }`, and only accept this error type in the upper layers. Just food for thoughts, let's keep things as is  for now.

praveenkumar (approved):

Issue #1707, opened 2020-11-20
[Integration Test] Add feature for monitoring operator


praveenkumar:
We don't have any tests in place for monitoring operator but it is mentioned in the document around how to enable it. 

Rough test scenario should be following
- Start the crc with 16GB ram, 6 vcpu
- Patch the cluster version operator (remove monitoring from the override list) 
   - https://code-ready.github.io/crc/#starting-monitoring-alerting-telemetry_gsg
- Wait till the all the pods are in running state in `openshift-monitoring` namespace
- execute `oc adm top node` or `oc adm top pods -A` which show the resource consumption for the node.

PR #1736, opened 2020-12-01
[WIP] Issue #1707: Integration: Add Scenario enabling monitoring operator

jsliacan:
_Needs testing on Windows_. MacOS/Linux work OK.

**Fixes:** Issue #1707 

## Solution

Instead of adding a Feature file, I added 1 Scenario into `story_health`. This utilizes a running cluster at the end of a test which only creates an app (i.e. after deleting the project, the cluster should be usable for enabling `cluster-monitoring-operator`).

## Testing

1. Run `make integration` with only `story_health` feature. It should pass OK.

### DIFF
--- a/test/integration/crcsuite/crcsuite.go
+++ b/test/integration/crcsuite/crcsuite.go
@@ -9,6 +9,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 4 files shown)


jsliacan:
Not needed thanks to #1774.

PR #1740, opened 2020-12-02
Unit for memory sizes are is MB, not MiB

cfergeau:
The libvirt machine driver expects memory sizes to be in MB (megabytes,
1000*1000 bytes), while crc documents memory sizes to be in MiB.
The difference is not too bad, if X MB is enough, X MiB will most likely
be enough, and vice versa.
Let's fix this though to avoid potential confusion.

https://github.com/code-ready/machine-driver-libvirt/blob/29905467a281ab60c40cc00a38957919fbdd1f00/pkg/libvirt/libvirt.go#L118-L122
https://github.com/code-ready/machine-driver-libvirt/blob/29905467a281ab60c40cc00a38957919fbdd1f00/pkg/libvirt/domain.go#L13-L16


cfergeau:
Actually, hyperkit uses MiB for memory size, and I believe hyperv as well, though the documentation is not explicit about it. So it looks like it's better to fix this in the libvirt driver than to take this patch.

praveenkumar:
closing this in favor of https://github.com/code-ready/machine-driver-libvirt/pull/79

Issue #1418, opened 2020-07-24
[BUG] CRC docs missing information about how to use container registry


rexroof:
Sorry if this is the wrong place for this:

minishift had pretty good documentation about how to use the internal container registry, I can't find that documentation for CRC.

here are some questions I've been trying to figure out:

should I use 
**image-registry.openshift-image-registry.svc:5000** for my registry address or should I use **default-route-openshift-image-registry.apps-crc.testing** ?
I notice that I have varying success with getting the SSL certificates on these hostnames to be recognized. 
I understand the issues around self-signed certificates but I am trying to work out the "best practices" as suggested by this project.  do I need to manually change the internal settings of my CRC cluster to allow insecure registries?  
Do I need to use one of these for docker push/pull and another for my image urls in my container manifests?
Do I need to create an imagePullSecret to use this registry?  How is that best accomplished?

I'm wondering if the lack of documentation around the internal container registry means that I shouldn't be pushing my custom containers to this registry and I should be using some other method to deploy my application on openshift 4 / CRC.

praveenkumar:
@rexroof Thanks for pointing it out, we will try to put note https://docs.okd.io/3.11/minishift/openshift/openshift-docker-registry.html like this one.

robin-owen:
@praveenkumar I'll need more information about how to document this as outlined in the original issue (great questions, by the way).

It seems to me that we could use a dedicated procedure for this task.

robin-owen:
Fixed by #1418.

PR #1741, opened 2020-12-02
Issue #1418 Document accessing the internal registry

robin-owen:
**Fixes:** Issue #1418

### DIFF
--- a/docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
+++ b/docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
@@ -1,0 +1,88 @@
+[id="accessing-the-internal-openshift-registry_{context}"]
+= Accessing the internal OpenShift registry
+
+The OpenShift cluster running in the {prod} virtual machine includes an internal container image registry by default.
+This internal container image registry can be used as a publication target for locally developed container images.
+To access the internal OpenShift registry, follow these steps.
+
+.Prerequisites
+
+* A running {prod} virtual machine.
+For more information, see link:{crc-gsg-url}#starting-the-virtual-machine_gsg[Starting the virtual machine].
+* A working [command]`oc` command.
+For more information, see link:{crc-gsg-url}#accessing-the-openshift-cluster-with-oc_gsg[Accessing the OpenShift cluster with `oc`].
+* An installation of [command]`podman` or [command]`docker`.
+** For Docker, add `default-route-openshift-image-registry.apps-crc.testing` as an insecure registry.
+For more information, see link:https://docs.docker.com/registry/insecure/[the Docker documentation].
+
+.Procedure
+
+. Check which user is logged in to the cluster:
++
+[subs="+quotes,attributes"]
+----
+$ oc whoami
+----
++
+[NOTE]
+====
+For demonstration purposes, the current user is assumed to be `kube:admin`.
+====
+
+. Log in to the registry as that user with its token:
++
+[subs="+quotes,attributes"]
+----
+$ podman login -u kubeadmin -p $(oc whoami -t) default-route-openshift-image-registry.apps-crc.testing --tls-verify=false
+----
+
+. Create a new project:
++
+[subs="+quotes,attributes"]
+----
+$ oc new-project demo
+----
+
+. Pull an example container image:
++
+[subs="+quotes,attributes"]
+----
+$ podman pull quay.io/libpod/alpine
+----
+
+. Tag the image, including namespace details:
++
+[subs="+quotes,attributes"]
+----
+$ podman tag alpine:latest default-route-openshift-image-registry.apps-crc.testing/demo/alpine:latest
+----
+
+. Push the container image to the internal registry:
++
+[subs="+quotes,attributes"]
+----
+$ podman push default-route-openshift-image-registry.apps-crc.testing/demo/alpine:latest --tls-verify=false
+----
+
+. Get imagestreams and verify that the pushed image is listed:
++
+[subs="+quotes,attributes"]
+----
+$ oc get is
+----
+
+. Enable image lookup for all tags in the imagestream:
++
+[subs="+quotes,attributes"]
+----
+$ oc set image-lookup
+----
++
+This setting allows the imagestream to be the source of images without having to provide the full URL to the internal registry.
+
+. Create a pod using the recently pushed image:
++
+[subs="+quotes,attributes"]
+----
+$ oc run demo --image=alpine --command -- sleep 600s
+----
--- a/docs/source/topics/assembly_accessing-the-openshift-cluster.adoc
+++ b/docs/source/topics/assembly_accessing-the-openshift-cluster.adoc
@@ -6,3 +6,5 @@ Access the OpenShift cluster running in the {prod} virtual machine through the O
 include::proc_accessing-the-openshift-web-console.adoc[leveloffset=+1]
 
 include::proc_accessing-the-openshift-cluster-with-oc.adoc[leveloffset=+1]
+
+include::proc_accessing-the-internal-openshift-registry.adoc[leveloffset=+1]


docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
```diff
@@ -0,0 +1,92 @@
...
+This internal container image registry can be used as a publication target for locally developed container images.
+To access the internal OpenShift registry, follow these steps.
+
+.Prerequisites
+
+* A running {prod} virtual machine.
+For more information, see link:{crc-gsg-url}#starting-the-virtual-machine_gsg[Starting the virtual machine].
+* A working [command]`oc` command.
+For more information, see link:{crc-gsg-url}#accessing-the-openshift-cluster-with-oc_gsg[Accessing the OpenShift cluster with `oc`].
+* For Docker, add `default-route-openshift-image-registry.apps-crc.testing` as an insecure registry.
```
cfergeau, line 14: what about podman ? (which is used in this doc)

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
```diff
@@ -0,0 +1,92 @@
...
+. Check which user is logged in to the cluster:
++
+[subs="+quotes,attributes"]
+----
+$ oc whoami
+----
++
+[NOTE]
+====
+For demonstration purposes, the current user is assumed to be `kube:admin`.
```
cfergeau, line 35: would that work as `developer`? or do we need elevated privileges?

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
```diff
@@ -0,0 +1,92 @@
...
+----
+
+. Get imagestreams and verify that the pushed image is listed:
++
+[subs="+quotes,attributes"]
+----
+$ oc get is
+----
+
+. Enable image lookup for all tags in the imagestream:
```
cfergeau, line 80: no idea what this means? Is there some relevant openshift documentation this could link to?

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
```diff
@@ -0,0 +1,92 @@
...
+[subs="+quotes,attributes"]
+----
+$ podman tag alpine:latest default-route-openshift-image-registry.apps-crc.testing/demo/alpine:latest
+----
+
+. Push the container image to the internal registry:
++
+[subs="+quotes,attributes"]
+----
+$ podman push default-route-openshift-image-registry.apps-crc.testing/demo/alpine:latest
```
cfergeau, line 70: when I read this, I wonder if it's possible to define short name for the registry instead of ` default-route-openshift-image-registry.apps-crc.testing` ;)

cfergeau (commented):
A few clarifications could be useful, but the step by step instructions look good!

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
robin-owen: Technically, I believe it can, but the `oc` commands would need to be run with `--as system:admin`, which is essentially the same thing if I understand it correctly.

This does make me wonder if we should simply advocate for using `kube:admin` in the docs without suggesting otherwise, though. That would be clearer for the reader in my opinion.

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
robin-owen: There is no prerequisite for using `podman` beyond having `podman` installed. It's entirely possible to add the internal registry as an insecure registry for `podman`, but this isn't strictly necessary as long as you log in via `podman` with the  `--tls-verify=false` flag (as documented).

Perhaps we could add "an installation of Docker or Podman" as a prerequisite.

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
robin-owen: Heh, same. ;)

I'm not aware of anything shorter we can use, though.

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
robin-owen: I'm not entirely certain either, haha. If I understand it correctly, this means that we can essentially just use the image name/tag instead of having to fully qualify it with the URL for the internal registry, but I may be incorrect about that.

I searched the docs for OpenShift 4.x and could only find this referenced in relation to using Kubernetes resources in OpenShift. That didn't really assist with my understanding of this option.

@praveenkumar, would you mind explaining this a bit more for us? Also, is this a necessary step for the procedure, or is it a "nice to have" for making the images easier to work with?

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
```diff
@@ -0,0 +1,92 @@
...
+* A running {prod} virtual machine.
+For more information, see link:{crc-gsg-url}#starting-the-virtual-machine_gsg[Starting the virtual machine].
+* A working [command]`oc` command.
+For more information, see link:{crc-gsg-url}#accessing-the-openshift-cluster-with-oc_gsg[Accessing the OpenShift cluster with `oc`].
+* For Docker, add `default-route-openshift-image-registry.apps-crc.testing` as an insecure registry.
+For more information, see link:https://docs.docker.com/registry/insecure/[the Docker documentation].
+
+.Procedure
+
+. Get the route to the internal OpenShift registry:
```
robin-owen, line 19: @praveenkumar On second thought, considering the URL for the internal registry is static, do we even need this step in the procedure?

praveenkumar: @robin-owen we don't need, as I said during our discussion that this url is going to static so no need to ask user to get the details about this url.

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
praveenkumar: https://docs.openshift.com/container-platform/4.6/openshift_images/using-imagestreams-with-kube-resources.html one might help

robin-owen: Found in the man page for `oc-set-image-lookup`: "Local name lookup allows an image stream to be the source of images for pods, deployments, replica sets, and other resources that reference images, without having to provide the full registry URL."

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
robin-owen: Seems I was incorrect, this procedure will work with the `developer` user without elevated privileges. :)

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
cfergeau: yes, the added `* An installation of [command]`podman` or [command]`docker`.` is what was missing.

cfergeau (approved):

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
```diff
@@ -0,0 +1,88 @@
...
+[subs="+quotes,attributes"]
+----
+$ podman tag alpine:latest default-route-openshift-image-registry.apps-crc.testing/demo/alpine:latest
+----
+
+. Push the container image to the internal registry:
++
+[subs="+quotes,attributes"]
+----
+$ podman push default-route-openshift-image-registry.apps-crc.testing/demo/alpine:latest --tls-verify=false
```
robin-owen, line 64: Updated after I noticed that `--tls-verify=false` is also required on this command.

PR #1744, opened 2020-12-03
Remove network manager and dnsmasq config in case systemd-resolved active

praveenkumar:
User who upgraded from fedora-32 to 33 and also used crc on 32 then
the networkmanager and dnsmasq configuration will be present and need
to be removed before adding the dispatcher file.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_network_linux.go
+++ b/pkg/crc/preflight/preflight_checks_network_linux.go
@@ -95,6 +95,13 @@ exit 0
 
 var systemdResolvedPreflightChecks = [...]Check{
 	{
+		configKeySuffix:  "check-dnsmasq-network-manager-config",
+		checkDescription: "Checking if dnsmasq configurations file exist for NetworkManager",
+		check:            checkCrcDnsmasqAndNetworkManagerConfigFile,
+		fixDescription:   "Removing dnsmasq configuration file for NetworkManager",
+		fix:              fixCrcDnsmasqAndNetworkManagerConfigFile,
+	},
+	{
 		configKeySuffix:  "check-systemd-resolved-running",
 		checkDescription: "Checking if the systemd-resolved service is running",
 		check:            checkSystemdResolvedIsRunning,
@@ -270,4 +277,27 @@ func fixCrcNetworkManagerDispatcherFile() error {
 
 func removeCrcNetworkManagerDispatcherFile() error {
 	return removeNetworkManagerConfigFile(crcNetworkManagerDispatcherPath)
+}
+
+func checkCrcDnsmasqAndNetworkManagerConfigFile() error {
+	// IF check return nil, which means file
+	if _, err := os.Stat(crcDnsmasqConfigPath); !os.IsNotExist(err) {
+		return fmt.Errorf("%s file exists", crcDnsmasqConfigPath)
+	}
+	if _, err := os.Stat(crcNetworkManagerConfigPath); !os.IsNotExist(err) {
+		return fmt.Errorf("%s file exists", crcNetworkManagerConfigPath)
+	}
+	return nil
+}
+
+func fixCrcDnsmasqAndNetworkManagerConfigFile() error {
+	// In case user upgrades from f-32 to f-33 the dnsmasq config for NM still
+	// exists and needs to be removed.
+	if err := removeCrcNetworkManagerConfig(); err != nil {
+		logging.Debugf("%s: not present.", crcNetworkManagerConfigPath)
+	}
+	if err := removeCrcDnsmasqConfigFile(); err != nil {
+		logging.Debugf("%s: not present.", crcDnsmasqConfigPath)
+	}
+	return nil
 }


pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -248,6 +248,15 @@ func checkSystemdResolvedIsRunning() error {
...
 func checkCrcNetworkManagerDispatcherFile() error {
+	// In case user upgrade from f-32 to f-33 the dnsmasq config for NM still
+	// exist and need to be removed.
+	if err := removeCrcNetworkManagerConfig(); err != nil {
+		logging.Debugf("%s: not present.", crcNetworkManagerConfigPath)
+	}
+	if err := removeCrcDnsmasqConfigFile(); err != nil {
+		logging.Debugf("%s: not present.", crcDnsmasqConfigPath)
+	}
+
```
cfergeau, line 259: I'd have a slight preference for doing this in a separate preflight check to make it clearer that it's cleanup for older releases ? You can indicate explicitly that this can be removed once f34 is released as f32 will be EOL'ed then. Though the same code may be needed if RHEL ever switches from NM+dnsmasq to NM+systemd-resolved.
And last but not least, this feels like a 'cleanup' thing rather than something we want to  transparently support at setup time?

praveenkumar: > this feels like a 'cleanup' thing rather than something we want to transparently support at setup time?

Most of the time user run the setup, be it a new release or running on a fresh system. cleanup tends to there to remove the changes which is done by setup side, this change is quite unusual since we are not doing it for new versions but carry forwarded from old releases. I still think having this here make more sense then putting it to cleanup side.

praveenkumar: Also when you mean by having it only a separate preflight clean up check, you mean something like below?

```patch
diff --git a/pkg/crc/preflight/preflight_checks_network_linux.go b/pkg/crc/preflight/preflight_checks_network_linux.go
index e678b278..61e237f2 100644
--- a/pkg/crc/preflight/preflight_checks_network_linux.go
+++ b/pkg/crc/preflight/preflight_checks_network_linux.go
@@ -38,6 +38,13 @@ var nmPreflightChecks = [...]Check{
        },
 }
 
+var olderReleaseCleanupChecks = [...]Check{
+       {
+               cleanupDescription: "Removing file from older os releases",
+               cleanup:            removeOlderConfigsFromOS,
+       },
+}
+
 var dnsmasqPreflightChecks = [...]Check{
        {
                configKeySuffix:    "check-network-manager-config",
@@ -248,15 +255,6 @@ func checkSystemdResolvedIsRunning() error {
 }
 
 func checkCrcNetworkManagerDispatcherFile() error {
        logging.Debug("Checking NetworkManager dispatcher file for crc network")
        err := crcos.FileContentMatches(crcNetworkManagerDispatcherPath, []byte(crcNetworkManagerDispatcherConfig))
        if err != nil {
@@ -280,3 +278,15 @@ func fixCrcNetworkManagerDispatcherFile() error {
 func removeCrcNetworkManagerDispatcherFile() error {
        return removeNetworkManagerConfigFile(crcNetworkManagerDispatcherPath)
 }
+
+func removeOlderConfigsFromOS() error{
+       // In case user upgrade from f-32 to f-33 the dnsmasq config for NM still
+       // exist and need to be removed.
+       if err := removeCrcNetworkManagerConfig(); err != nil {
+               logging.Debugf("%s: not present.", crcNetworkManagerConfigPath)
+       }
+       if err := removeCrcDnsmasqConfigFile(); err != nil {
+       logging.Debugf("%s: not present.", crcDnsmasqConfigPath)
+       }
+       return nil
+}
\ No newline at end of file
diff --git a/pkg/crc/preflight/preflight_linux.go b/pkg/crc/preflight/preflight_linux.go
index 246465c9..6ce3f71d 100644
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -180,6 +180,7 @@ func getNetworkChecksForDistro(distro *linux.OsRelease, networkMode network.Mode
                checks = append(checks, nmPreflightChecks[:]...)
                if usesSystemdResolved(distro) {
                        checks = append(checks, systemdResolvedPreflightChecks[:]...)
+                       checks = append(checks, olderReleaseCleanupChecks[:]...)
                } else {
                        checks = append(checks, dnsmasqPreflightChecks[:]...)
                }

```

gbraad: > Most of the time user run the setup, 

Setup should only be needed for the first run and after a cleanup.

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -94,6 +94,13 @@ exit 0
 )
 
 var systemdResolvedPreflightChecks = [...]Check{
+	{
+		configKeySuffix:  "check-dnsmasq-networkmanager-config-file-exist",
```
cfergeau, line 98: The similar keys are named `check-network-manager-config` and `check-crc-dnsmasq-file`, maybe go with `check-dnsmasq-network-manager-config`?

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -94,6 +94,13 @@ exit 0
...
+		checkDescription: "Checking if dnsmasq configurations file exists for NetworkManager",
```
cfergeau, line 99: `configuration files exist for` ?

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -94,6 +94,13 @@ exit 0
...
+		check:            checkCrcDnsmasqAndNetworkManagerConfigFile,
+		fixDescription:   "Removing dnsmasq configurations file for NetworkManager",
```
cfergeau, line 101: `configuration files`

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -271,3 +278,26 @@ func fixCrcNetworkManagerDispatcherFile() error {
 func removeCrcNetworkManagerDispatcherFile() error {
 	return removeNetworkManagerConfigFile(crcNetworkManagerDispatcherPath)
 }
+
+func checkCrcDnsmasqAndNetworkManagerConfigFile() error {
+	// IF check return nil, which means file
```
cfergeau, line 283: `If Stat() returns nil, this means the file exists` ?

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -271,3 +278,26 @@ func fixCrcNetworkManagerDispatcherFile() error {
...
+		return fmt.Errorf("%s file exists", crcDnsmasqConfigPath)
+	}
+	if _, err := os.Stat(crcNetworkManagerConfigPath); err == nil {
+		return fmt.Errorf("%s file exists", crcNetworkManagerConfigPath)
+	}
+	return nil
+}
+
+func fixCrcDnsmasqAndNetworkManagerConfigFile() error {
+	// In case user upgrade from f-32 to f-33 the dnsmasq config for NM still
```
cfergeau, line 294: `upgrades`, `exists`, `needs`

cfergeau (approved):

PR #1750, opened 2020-12-04
Update bundle version from 4.6.6 to 4.6.7

praveenkumar:


cfergeau:
`stable` and `latest` both point to `4.6.6`, only `candidate` is `4.6.7`, shouldn't we go with `4.6.6` for the release?

praveenkumar:
> only candidate is 4.6.7, shouldn't we go with 4.6.6 for the release?

I usually take whatever latest candidate for z stream and if that passes all the check from crc side.

cfergeau:
> > only candidate is 4.6.7, shouldn't we go with 4.6.6 for the release?
> 
> I usually take whatever latest candidate for z stream and if that passes all the check from crc side.

This feels like problems waiting to happen, isn't `candidate` something which is still being tested, and which won't get promoted to `stable`/`latest` if significant issues are found with it?

praveenkumar:
> This feels like problems waiting to happen, isn't candidate something which is still being tested, and which won't get promoted to stable/latest if significant issues are found with it?

candidate is available after the QE then only it is available for public even customer can use it, I am not sure what testing is going on from candidate to stable.

cfergeau:
> > This feels like problems waiting to happen, isn't candidate something which is still being tested, and which won't get promoted to stable/latest if significant issues are found with it?
> 
> candidate is available after the QE then only it is available for public even customer can use it, I am not sure what testing is going on from candidate to stable.

https://docs.openshift.com/container-platform/4.6/updating/updating-cluster.html#understanding-upgrade-channels_updating-cluster
> Release candidates contain all the features of the product but are not supported. Use release candidate versions to test feature acceptance and assist in qualifying the next version of OpenShift Container Platform. [...] After a version is available in the candidate channel, it goes through more quality checks. If it meets the quality standard, it is promoted to the fast-4.6 or stable-4.6 channels.

Imo `candidate` is not something we want to base official crc releases on.

Issue #1747, opened 2020-12-03
Incompatibility between the new SSH cipher policies in FCOS 33 and golang.org/x/crypto/ssh


cgruver:
It looks like the id_rsa file is not getting created in the .crc/machine directory.

### General information

  * OS: Any
  * Hypervisor: KVM / Hyper-V / hyperkit
  * Did you run `crc setup` before starting it Yes
  * Running CRC on: Laptop  & Bare Metal

## CRC version
```bash
CodeReady Containers version: 1.19.0-OKD+8d69386
OpenShift version: 4.6.0-0.okd-2020-11-27-200126 (embedded in executable)
```
 
### Steps to reproduce

1. The bundle was built with:
   ```bash
   OKD_VERSION=4.6.0-0.okd-2020-11-27-200126
   OPENSHIFT_PULL_SECRET_PATH=/tmp/pull_secret.json
   ```

   The SNC revision is from my fork: https://github.com/cgruver/snc
   The revision is: 4.1.14-157-g291ef43

   The single node cluster built successfully, and was usable.
   The bundle built successfully.

1. CRC was built with:

   ```bash
   OKD_BUILD=true
   BUNDLE_VERSION=4.6.0-0.okd-2020-11-27-200126
   BUNDLE_DIR=/crc-build/snc
   make embed_bundle
   ```

   The CRC build completed successfully.

1. The setup completes with no errors:  Tested on MacOS 11 and CentOS 8

   ```bash
   crc setup
   ```
   It unpacks the bundle and prepares the environment.

1. The cluster start fails with ssh errors:

```bash
INFO Starting CodeReady Containers VM for OpenShift 4.6.0-0.okd-2020-11-27-200126... 
DEBU Updating CRC VM configuration                
(crc) Calling .GetConfigRaw
(crc) Calling .Start
(crc) DBG | Starting VM crc
(crc) DBG | Validating network
(crc) DBG | Validating storage pool
(crc) DBG | GetIP called for crc
(crc) DBG | Getting current state...
(crc) DBG | IP address: 192.168.130.11
(crc) Found IP for machine: 192.168.130.11
(crc) Calling .GetState
(crc) DBG | Getting current state...
(crc) Calling .GetIP
(crc) DBG | GetIP called for crc
(crc) DBG | Getting current state...
(crc) DBG | IP address: 192.168.130.11
DEBU Waiting until ssh is available               
DEBU retry loop: attempt 0                        
DEBU About to run SSH command:
exit 0             
[...]

PR #1757, opened 2020-12-05
Modified to support ECDSA SSH keys.

cgruver:
**Fixes:** https://github.com/code-ready/crc/issues/1747

**Dependent on:** https://github.com/code-ready/snc/pull/289

Modified CRC to support the more secure ECDSA SSH cipher.

### DIFF
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -151,6 +151,11 @@ func GetPrivateKeyPath() string {
 	return filepath.Join(MachineInstanceDir, DefaultName, "id_ecdsa")
 }
 
+// For backward compatibility to v 1.20.0
+func GetRsaPrivateKeyPath() string {
+	return filepath.Join(MachineInstanceDir, DefaultName, "id_rsa")
+}
+
 // TODO: follow the same pattern as oc and podman above
 func GetCRCMacTrayDownloadURL() string {
 	return fmt.Sprintf(CRCMacTrayDownloadURL, version.GetCRCMacTrayVersion())
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -144,11 +144,11 @@ func contains(arr []string, str string) bool {
 }
 
 func GetPublicKeyPath() string {
-	return filepath.Join(MachineInstanceDir, DefaultName, "id_rsa.pub")
+	return filepath.Join(MachineInstanceDir, DefaultName, "id_ecdsa.pub")
 }
 
 func GetPrivateKeyPath() string {
-	return filepath.Join(MachineInstanceDir, DefaultName, "id_rsa")
+	return filepath.Join(MachineInstanceDir, DefaultName, "id_ecdsa")
 }
 
 // TODO: follow the same pattern as oc and podman above
--- a/pkg/crc/ssh/keys.go
+++ b/pkg/crc/ssh/keys.go
@@ -1,8 +1,9 @@
 package ssh
 
 import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/rsa"
 	"crypto/x509"
 	"errors"
 	"fmt"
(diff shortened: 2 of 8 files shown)


guillaumerose:
/ok-to-test

Should we change also this in the installer ? snc will also add a RSA key in the bundle if I'm correct.

guillaumerose:
If you can find a way to not break upgrade. It's maybe hard to do something backward compatible and not worth it.

cgruver:
D'oh!   Lint check failed... I'll fix that.

cgruver:
> If you can find a way to not break upgrade. It's maybe hard to do something backward compatible and not worth it.

Let's chat about that on Slack.  I'm not sure what it will take, but would be willing to take a crack at it.

gbraad (approved):

gbraad:
backward compat is not really worth it.

perhaps detecting it and suggesting to recreate the machine. we are ephemeral anyway.

cfergeau (commented):
The commit log is missing an explanation why we are doing that. It should document that this is due to the crypto policy changes in f33/fcos33 which reject connection with rsa keys, which we were using so far.

cfergeau:
Can't backward compat be dealt with by this change in machine/machine.go ?
```
- sshRunner, err := crcssh.CreateRunner(instanceIP, getSSHPort(client.useVSock()), crcBundleMetadata.GetSSHKeyPath(), constants.GetPrivateKeyPath())
+ sshRunner, err := crcssh.CreateRunner(instanceIP, getSSHPort(client.useVSock()), crcBundleMetadata.GetSSHKeyPath(), constants.GetPrivateKeyPath(), crcBundleMetadata.GetOldSSHKeyPath())
```
(`GetOldSSHKeyPath()` will return `id_rsa.pub`)

praveenkumar:
I don't understand why it is not the backward compatible (even CI is passing which is using the older bundles), initial ssh key information is part of bundle metadata, as long as it have correct details it will start the first ssh connection, now we generate the keypair (ecdsa) which is also valid for old bundles so, both the way it is going to work.

guillaumerose:
If you start an instance with crc 1.20.0 and try to do crc status with this one, you will get an error because the ssh key is not found.

praveenkumar:
Ah you mean for already existing instance (running/stopped)  not the one which you created with this changes either older or newer bundles.

cgruver:
> Ah you mean for already existing instance (running/stopped) not the one which you created with this changes either older or newer bundles.

I get it, it's simply because of the file name change...  id_rsa to id_ecdsa.

cgruver:
I think @cfergeau was correct.  Adding backward compatibility was pretty easy...  provided this worked.

pkg/crc/constants/constants.go
```diff
@@ -144,10 +144,15 @@ func contains(arr []string, str string) bool {
...
-	return filepath.Join(MachineInstanceDir, DefaultName, "id_rsa.pub")
+	return filepath.Join(MachineInstanceDir, DefaultName, "id_ecdsa.pub")
 }
 
 func GetPrivateKeyPath() string {
+	return filepath.Join(MachineInstanceDir, DefaultName, "id_ecdsa")
+}
+
+// For backward compatibility to v 1.20.0
+func GetRsaPrivateKeyPath() string {
```
guillaumerose, line 155: RSA is an acronym, so use capital letters? `GetRSAPrivateKeyPath`

pkg/crc/ssh/keys.go
```diff
@@ -26,17 +27,17 @@ type KeyPair struct {
 // NewKeyPair generates a new SSH keypair
 // This will return a private & public key encoded as DER.
 func NewKeyPair() (keyPair *KeyPair, err error) {
-	priv, err := rsa.GenerateKey(rand.Reader, 2048)
+
```
guillaumerose, line 30: extra space

guillaumerose (approved):
2 nits but otherwise lgtm.

PR #1758, opened 2020-12-07
Backup log file as part of delete command

praveenkumar:
crc log file is opened as append mode and never gets deleted until
user manually delete it or remove the base directory. Currently it
is filled with lot of message which might not be require to debug
an issue. This patch make sure every new crc cycle (setup => start => delete)
have required logs to logfile.

## Testing
- Run crc start => check the log file exist in `~/. crc/crc.log`
- Run crc delete => check the log file shouldn't present

Any other command shouldn't delete this file.


cmd/crc/cmd/delete.go
```diff
@@ -51,7 +51,12 @@ func deleteMachine(client machine.Client, clearCache bool, cacheDir string, inte
 
 	yes := input.PromptUserForYesOrNo("Do you want to delete the OpenShift cluster", force)
 	if yes {
-		return true, client.Delete()
+		err := client.Delete()
+		// Delete the log file
+		if err := os.RemoveAll(constants.LogFilePath); err != nil {
```
guillaumerose, line 56: When running unit tests, this code will delete your log folder. Pass it as a variable.

Also, maybe create a method `logging.Destroy or Clean` in the logging package. It avoids the spread of this constant.

pkg/crc/machine/fakemachine/client.go
```diff
@@ -18,8 +19,15 @@ func NewFailingClient() *Client {
 	}
 }
 
+func NewClientWithLogFile(logfile *os.File) *Client {
```
guillaumerose, line 22: Why a logFile here? Even in the real implementation we don't pass the logFile.

pkg/crc/logging/logging.go
```diff
@@ -35,6 +35,10 @@ func CloseLogging() {
 	logrus.StandardLogger().ReplaceHooks(make(logrus.LevelHooks))
 }
 
+func RemoveLogFile(path string) error {
```
guillaumerose, line 38: It would be best to not leak the log path out of this package. But this is a larger refactoring to do.

A rule I like: when you have a constant, use it only once (or at least as few as possible, only in the same package).

pkg/crc/machine/fakemachine/client.go
praveenkumar: @guillaumerose so that we can close it after delete execute, in real implementation we have `cleanup` which do the same but there the logfile path is set by the root init method.

anjannath:
Maybe this was discussed, but why are we not doing some kind of log rotation instead of deleting the log file completely, we often suggest to do `crc delete` => `crc setup` to fix some issues, in that case won't the old logs be gone?

cfergeau:
> Maybe this was discussed, but why are we not doing some kind of log rotation instead of deleting the log file completely, we often suggest to do `crc delete` => `crc setup` to fix some issues, in that case won't the old logs be gone?

Not a big fan of log rotation for our usecase, what if the latest log does not have all the info that we need? Why would 6 months old logs still be relevant? Why reinvent the wheel (log rotation) rather than integrate with pre-existing system solutions (journald, windows equivalent, ...)?

pkg/crc/logging/logging.go
cfergeau: Yes, it feels like a better API for the `logging` package would remove the need for most of the complications in this patch series (additional logfile argument to `runDelete`, `NewClientWithLogFile`, ...

Seems that having `logging.Init("/some/logfile")`, `logging.RemoveLogFile()` and `logging.Close()` would be enough? 
Looking at the code of the `logging` package, it seems `InitLogrus()` will leak a file descriptor each time it's called :-/ There's a global `logfile` variable, but `InitLogrus` initializes a different `logFile` variable.

pkg/crc/machine/fakemachine/client.go
cfergeau: Is it really important to close it in `delete`? Could it be done in the test case, or is deleting a file with an open file descriptor going to be an issue in Windows?

cmd/crc/cmd/delete.go
```diff
@@ -51,13 +52,18 @@ func deleteMachine(client machine.Client, clearCache bool, cacheDir string, inte
 
 	yes := input.PromptUserForYesOrNo("Do you want to delete the OpenShift cluster", force)
 	if yes {
-		return true, client.Delete()
+		err := client.Delete()
+		// Delete the log file
```
cfergeau, line 56: For what it's worth, since we call a method with a good descriptive name `RemoveLogFile`, I don't think this comment is needed.

cfergeau:
> > Maybe this was discussed, but why are we not doing some kind of log rotation instead of deleting the log file completely, we often suggest to do `crc delete` => `crc setup` to fix some issues, in that case won't the old logs be gone?
> 
> Not a big fan of log rotation for our usecase, what if the latest log does not have all the info that we need? Why would 6 months old logs still be relevant? Why reinvent the wheel (log rotation) rather than integrate with pre-existing system solutions (journald, windows equivalent, ...)?

Actually, I might have misunderstood your suggestion, renaming the log file on `crc delete` and creating a new one for subsequent `crc` runs could indeed be a good alternative. We could remove all logs at `crc cleanup` time to avoid keeping them indefinitely.

anjannath:
> Actually, I might have misunderstood your suggestion, renaming the log file on crc delete and creating a new one for subsequent crc runs could indeed be a good alternative. We could remove all logs at crc cleanup time to avoid keeping them indefinitely.

I was thinking of what is done in #1777 it also deletes old log files

guillaumerose:
We can go with this but it would be good in the future to have a proper struct for log functions (instead of package function).

/lgtm

Issue #1490, opened 2020-08-27
Integration test regex fails when the disk usage doesn't have decimal part.


praveenkumar:
```
13:55:58     And stdout should match ".*Disk Usage: *\d+\.\d+GB of 32.\d+GB.*"                                                                # shell.go:330 -> github.com/code-ready/clicumber/testsuite.CommandReturnShouldMatch
13:55:58     output did not match. Expected: '.*Disk Usage: *\d+\.\d+GB of 32.\d+GB.*', Actual: 'level=debug msg="CodeReady Containers version: 1.15.0+8784bec0\n"
13:55:58 level=debug msg="OpenShift version: 4.5.7 (not embedded in binary)\n"
13:55:58 level=debug msg="About to run SSH command with hidden output"
13:55:58 level=debug msg="SSH command succeeded"
13:55:58 level=debug msg="kube-apiserver operator is still progressing, Reason: NodeInstaller"
13:55:58 level=debug msg="About to run SSH command:\ndf -B1 --output=size,used,target /sysroot | tail -1"
13:55:58 level=debug msg="SSH command results: err: <nil>, output: 32720793600 13002735616 /sysroot\n"
13:55:58 CRC VM:          Running
13:55:58 OpenShift:       Running (v4.5.7)
13:55:58 Disk Usage:      13GB of 32.72GB (Inside the CRC VM)
13:55:58 Cache Usage:     10.39GB
13:55:58 Cache Directory: /home/crc_ci/.crc/cache'
```

PR #1759, opened 2020-12-07
Issue #1490: Better recognize disk usage by regex

jsliacan:
**Fixes:** Issue #1490 

Only found 1 instance in `basic.feature`. 

## Testing

Try to get your disk usage to be an integer, then run the `basic.feature` and hope it's still an integer when the check happens :). Or, here's a [go playground](https://play.golang.org/p/u0tFvgJlcT1) where the new regex recognizes both cases.

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -95,7 +95,7 @@ Feature: Basic test
     @darwin @linux @windows
     Scenario: CRC status and disk space check
         When checking that CRC is running
-        And stdout should match ".*Disk Usage: *\d+\.\d+GB of 32.\d+GB.*"
+        And stdout should match ".*Disk Usage: *\d+[\.\d]*GB of 32.\d+GB.*"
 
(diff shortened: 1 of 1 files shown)


test/integration/features/basic.feature
```diff
@@ -95,7 +95,7 @@ Feature: Basic test
     @darwin @linux @windows
     Scenario: CRC status and disk space check
         When checking that CRC is running
-        And stdout should match ".*Disk Usage: *\d+\.\d+GB of 32.\d+GB.*"
+        And stdout should match ".*Disk Usage: *\d+[\.\d+]*GB of 32.\d+GB.*"
```
guillaumerose, line 98: I guess the + between [ ] is unwanted. Otherwise, lgtm.

jsliacan: Should be paying more attention, thanks.

guillaumerose (approved):

PR #1761, opened 2020-12-08
Initial telemetry implementation

praveenkumar:
- Add config option for telemeter
- Send only error and log in case of error to segment
- No personal data collection
- Use the anonymous id based on random generated one


pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,74 @@
...
+		DefaultContext: &analytics.Context{
+			App: analytics.AppInfo{
+				Name:    constants.DefaultName,
+				Version: version.GetCRCVersion(),
+			},
+		},
+	})
+}
+
+func UploadToSegment(client analytics.Client, err error) error {
```
guillaumerose, line 25: Wrap segment library and use the config to decide wether or not send to segment:

```
type Client struct {
  segmentClient analytics.Client
  config Config // crc configuration object
}

func (c *Client) Upload() {
  if config.Get(key).AsBool() {
    // send to segment
  }
}

```

It will then be easy to add a unit test that verify no data are sent to segment if the config key is set.
A unit test that verifies what is sent to Segment is also a good thing to have.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,74 @@
...
+	"os"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/version"
+	"github.com/pborman/uuid"
+	"github.com/segmentio/analytics-go"
+)
+
+func NewSegmentClient() (analytics.Client, error) {
```
cfergeau, line 14: Imo `Segment` can be dropped from the name.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/version"
+	"github.com/pborman/uuid"
+	"github.com/segmentio/analytics-go"
```
cfergeau, line 11: Should this be versioned? `import "gopkg.in/segmentio/analytics-go.v3"`

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func NewSegmentClient() (analytics.Client, error) {
+	return analytics.NewWithConfig("cvpHsNcmGCJqVzf6YxrSnVlwFSAZaYtp", analytics.Config{
```
cfergeau, line 15: I'm not sure what are the rules for segment write keys? Is this a personal key, something owned by crc, ...? Is this supposed to be secret?

pkg/crc/segment/segment.go
cfergeau, line 25: `ToSegment` is also redundant imo. Could be `ReportError` though, as this is not uploading arbitrary data, just reporting an error to segment.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,74 @@
...
+		return "", err
+	}
+	return id, nil
+}
+
+func setError(t analytics.Traits, err error) analytics.Traits {
+	return t.Set("error", err.Error())
+}
+
+func SetLogs(t analytics.Traits, logs []byte) analytics.Traits {
```
cfergeau, line 72: This could be `setLogs`?

cfergeau (commented):
"Send only [...] log in case of error to segment" and "No personal data collection" do not go that well together.
The log file has my username, I think the AWS region can be guessed from quay.io IPs which are in the logs too, you can also find the proxy values there, ...

pkg/crc/segment/segment.go
praveenkumar: @cfergeau it is not suppose to secret, I did ask that before sending this PR

pkg/crc/segment/segment.go
praveenkumar: This contain logs and error.

pkg/crc/segment/segment.go
praveenkumar: There is issue with the library right now https://github.com/segmentio/analytics-go/issues/170 which doesn't behave correctly with go-1.15

cfergeau:
Is 'telemeter' intentional? or should it be 'telemetry'?

pkg/crc/logging/logging.go
```diff
@@ -30,6 +30,30 @@ func CloseLogFile() {
 	logfile.Close()
 }
 
+func GetLatestContentOfLogFile(path string) ([]byte, error) {
```
guillaumerose, line 33: This is where we need to create a new interface/struct otherwise it will be impossible to test.
It will also help you in the "crc delete deletes also logs" PR I guess as you will be able to mock the log system.

gbraad:
> AWS region can be guessed from quay.io

this is GeoIP and is not considered PII (Personal Identifying Information)

cmd/crc/cmd/config/config.go
```diff
@@ -24,6 +24,7 @@ const (
 	HTTPSProxy           = "https-proxy"
 	NoProxy              = "no-proxy"
 	ProxyCAFile          = "proxy-ca-file"
+	EnableTelemetry      = "enable-telemetry"
```
gbraad, line 27: we have to make sure this gets a description, as it is suggestive of something the cluster could do

pkg/crc/logging/logging.go
```diff
@@ -30,6 +30,30 @@ func CloseLogFile() {
...
+	if err != nil {
+		return nil, err
+	}
+
+	buf := make([]byte, 30000)
+	start := int64(0)
+	if fstat.Size() > 30000 {
+		start = fstat.Size() - 30000
+	}
+	_, err = f.ReadAt(buf, start)
```
gbraad, line 50: can't this be 'lines' instead?

cmd/crc/cmd/config/config.go
praveenkumar: @gbraad should we name it `EnableCRCTelemetry` ?

gbraad: I think it is more about `"enable-telemetry"`...

cfergeau:
> > AWS region can be guessed from quay.io
> 
> This is caused by GeoIP and is not considered PII (Personal Identifying Information), 

Yes, if there was only this piece of information, it would be ok. In combination with the username, it becomes one additional piece of (vague) personal data collected by crc about 'username'

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,88 @@
...
+func getUserIdentity() (string, error) {
+	if err := os.MkdirAll(constants.TelemetryDirectoryPath, 0750); err != nil {
+		return "", err
+	}
+	if _, err := os.Stat(constants.TelemetryFilePath); !os.IsNotExist(err) {
+		id, err := ioutil.ReadFile(constants.TelemetryFilePath)
+		if err != nil {
+			return "", err
+		}
+		return string(id), nil
```
fbricon, line 72: if id is empty, we should regenerate it

gbraad: right... this is a separate process/function. Here we 'detect' it, and act upon the result in the caller. Right?

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+	})
+	if err != nil {
+		return Client{}, err
+	}
+	return Client{segmentClient: client, config: config}, nil
+}
+
+func (c Client) Upload(err error) error {
+	defer c.segmentClient.Close()
```
guillaumerose, line 37: What will happen on the second call of Upload ? segmentClient will be closed.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,79 @@
...
+	if !c.config.Get(config.ConsentTelemetry).AsBool() {
+		return nil
+	}
+	logging.Info("Uploading the error to segment")
+
+	AnonymousID, uerr := getUserIdentity()
```
guillaumerose, line 43: s/AnonymousID/anonymousID/

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return c.segmentClient.Enqueue(analytics.Identify{
+		AnonymousId: AnonymousID,
+		Traits:      t,
+	})
+}
+
+func getUserIdentity() (string, error) {
+	var id []byte
+	if err := os.MkdirAll(constants.TelemetryDirectoryPath, 0750); err != nil {
```
guillaumerose, line 59: onstants.TelemetryDirectoryPath = filepath.Dir(constants.TelemetryFilePath)

It removes one constant.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,79 @@
...
+		return "", err
+	}
+	if _, err := os.Stat(constants.TelemetryFilePath); !os.IsNotExist(err) {
+		id, err = ioutil.ReadFile(constants.TelemetryFilePath)
+		if err != nil {
+			return "", err
+		}
+	}
+	if string(id) == "" {
```
guillaumerose, line 68: Maybe validate the format and remove extra whitespaces.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+	if err != nil {
+		return Client{}, err
+	}
+	return Client{segmentClient: client, config: config}, nil
+}
+
+func (c Client) Upload(err error) error {
+	defer c.segmentClient.Close()
+	if !c.config.Get(config.ConsentTelemetry).AsBool() {
```
guillaumerose, line 38: A unit test for this would be great.

cmd/crc/cmd/start.go
```diff
@@ -60,6 +60,9 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 	checkIfNewVersionAvailable(config.Get(cmdConfig.DisableUpdateCheck).AsBool())
 
 	if err := preflight.StartPreflightChecks(config); err != nil {
+		if err := segmentClient.Upload(err); err != nil {
+			return nil, err
+		}
```
cfergeau, line 65: This looks like this should not be done in `runStart()`, but in a function which calls `runStart()`, as this has to be duplicated for all error paths happening in `runStart()`

cmd/crc/cmd/config/config.go
```diff
@@ -24,6 +24,7 @@ const (
 	HTTPSProxy           = "https-proxy"
 	NoProxy              = "no-proxy"
 	ProxyCAFile          = "proxy-ca-file"
+	ConsentTelemetry     = "consent-telemetry"
```
cfergeau, line 27: not sure how well it reads in English? Should it be `telemetry-consent`, or just `enable-telemetry`?

cfergeau (commented):
I assume this needs to be documented somewhere, what data is sent, in which cases, ..? What is the overall plan btw, at the moment we only log errors from `crc start`. Do we expect to be logging  more things? like all successful invocations of `crc start`/`crc setup`/...?
In addition to the error, logging the crc version is most likely going to be useful.

pkg/crc/segment/segment.go
praveenkumar: Since we are hooking this with commands in case of error so every command will just going to run once the upload functionality.

cmd/crc/cmd/config/config.go
praveenkumar: @gbraad ^^

pkg/crc/segment/segment_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	crcConfig "github.com/code-ready/crc/pkg/crc/config"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	ConsentTelemetry = "consent-telemetry"
+)
+
+func newTestConfig(configFile string, value bool, envPrefix string) (*crcConfig.Config, error) {
```
guillaumerose, line 17: To avoid the file, you can do the same as https://github.com/code-ready/crc/blob/56356140518b158b1e6119e0f9e9598d85d2db17/pkg/crc/api/api_test.go#L200

pkg/crc/segment/segment_test.go
```diff
@@ -0,0 +1,55 @@
...
+func TestClientUploadWithoutConcent(t *testing.T) {
+	dir, err := ioutil.TempDir("", "cfg")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+	configFile := filepath.Join(dir, "crc.json")
+
+	config, err := newTestConfig(configFile, false, "CRC")
+	require.NoError(t, err)
+
+	c, err := NewClient(config)
```
guillaumerose, line 36: This client will use the real uuid and the real file. You need to use a temporary file. A test can verify the loading of the uuid or if empty, the generation of the uuid.

pkg/crc/segment/segment_test.go
```diff
@@ -0,0 +1,55 @@
...
+	require.NoError(t, err)
+	require.NoError(t, c.Upload(nil))
```
guillaumerose, line 38: This is the hard part of the test. You can change the `Endpoint` (https://github.com/segmentio/analytics-go/blob/master/analytics.go#L114) to a embedded http server and count the number of calls received. It should be 0 in this case.

When we will add more data, we can then analyze the payload and write assertions for some fields.

praveenkumar: Right now I think the only way to perform those unit test is to modify the `Upload` signature to accept the `client`, `identity_file`, `err` message because there is no way I can mock all those bit in current scenario. Then I can create a new dummy client with different endpoint and also pass it to Upload and get the result.

pkg/crc/segment/segment_test.go
```diff
@@ -0,0 +1,123 @@
...
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	config, err := newTestConfig(false)
+	require.NoError(t, err)
+
+	c, err := NewClient(config)
+	require.NoError(t, err)
+
+	require.NoError(t, Upload(c, filepath.Join(dir, "telemetry"), nil))
```
guillaumerose, line 122: You can add `require.Len(t, s.Batch, 0)` or so and it will be perfect!

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,81 @@
...
+			},
+		},
+	})
+	if err != nil {
+		return &Client{}, err
+	}
+	return &Client{segmentClient: client, config: config}, nil
+}
+
+func Upload(c *Client, telemetryFilePath string, err error) error {
```
guillaumerose, line 38: I think this signature is better:

```
type Client struct {
	segmentClient       analytics.Client
	config                    *crcConfig.Config
        telemetryFilePath string

}

func (c *Client) Upload(err error) error {
  ...
}
```

It allows mocking in cmd with:

```
type mockTelemetry struct {
}

func (m *mockTelemetry) Upload(err error) error {
  return nil // or we can collect errors in an array and write assertions on it.
}
```

guillaumerose (approved):

pkg/crc/segment/segment.go
cfergeau: This seems like a fairly strong limitation, we should at least be documented. It would be better to provide a `Close()` method, and handle that in the caller. My understanding is that long term we may want to call this for each `crc start`, `crc stop`, ... calls, even when successful, in which case this will come and byte us. We'll also want to use this in the daemon.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,82 @@
...
+	config            *crcConfig.Config
+	telemetryFilePath string
+}
+
+func NewClient(config *crcConfig.Config, telemetryFilePath string) (*Client, error) {
+	client, err := analytics.NewWithConfig("cvpHsNcmGCJqVzf6YxrSnVlwFSAZaYtp", analytics.Config{
+		DefaultContext: &analytics.Context{
+			App: analytics.AppInfo{
+				Name:    constants.DefaultName,
+				Version: version.GetCRCVersion(),
```
cfergeau, line 29: Since this is using some values from `constants`/`version`, shouldn't this also hardcode `constants.TelemetryFilePath`? I don't think the calling code cares, and this does not seem to be set to a different value in tests.

pkg/crc/segment/segment.go
```diff
@@ -0,0 +1,82 @@
...
+	segmentClient     analytics.Client
+	config            *crcConfig.Config
+	telemetryFilePath string
+}
+
+func NewClient(config *crcConfig.Config, telemetryFilePath string) (*Client, error) {
+	client, err := analytics.NewWithConfig("cvpHsNcmGCJqVzf6YxrSnVlwFSAZaYtp", analytics.Config{
+		DefaultContext: &analytics.Context{
+			App: analytics.AppInfo{
+				Name:    constants.DefaultName,
```
cfergeau, line 28: Imo `constants.DefaultName` is one of these constants that we should not be using, but try to remove.

pkg/crc/segment/segment.go
praveenkumar: @cfergeau In the test we are using a mock client with different endpoint, and we can specify different value for this and check the response also if that is the concern here?

pkg/crc/segment/segment.go
praveenkumar: @cfergeau and what do you propose, passing it's value to the NewClient side, make it more explicit ?

cfergeau: Passing it to NewClient would probably make sense yes. Maybe it can be improved in a follow-up PR.

pkg/crc/segment/segment.go
cfergeau: segment_test.go is not using `NewClient()`, so the 'path' argument is not used by tests so `NewClient()` can be changed to
```
func NewClient(config *crcConfig.Config)  (*Client, error) {
    const telemetryFilePath = filepath.Join(constants.GetHomeDir(), ".redhat", "anonymousId")

...
    return &Client{segmentClient: client, config: config, telemetryFilePath: telemetryFilePath}, nil
}
```
and the constants.go changes can be dropped.

praveenkumar: @cfergeau Alright I got what you mean now.

pkg/crc/segment/segment.go
cfergeau: `c.segmentClient.Close()` flushes all the messages which were enqueued, I suspect this could be blocking for some time in case of DNS/network issues. One thing which might need to be improved in the future :)

guillaumerose:
That's a good start! Let's iterate on this.

So far, what I see missing:
* Ensure the CI doesn't send analytics
* Close only once the segment client
* Send more metrics!
* Also send metrics when using the `crc daemon` and the tray.

Issue #1767, opened 2020-12-09
Build with -mod=vendor ?


cfergeau:
We maintain a vendor/ directory, but when experimenting with https://github.com/code-ready/crc/issues/1766 , the content of the vendor/ directory is not used, the dependencies are all fetched at build time according to go.mod. This means the crc binary that we build will not necessarily match what is in vendor/. Should make use of `-mod=vendor` to force the use of what is in vendor/?

```
diff --git a/Makefile b/Makefile
index fccb5cc8..6951b4c2 100644
--- a/Makefile
+++ b/Makefile
@@ -83,23 +83,23 @@ install: $(SOURCES)
 	go install -ldflags="$(LDFLAGS)" ./cmd/crc
 
 $(BUILD_DIR)/macos-amd64/crc: $(SOURCES)
-	GOARCH=amd64 GOOS=darwin go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/macos-amd64/crc ./cmd/crc
+	GOARCH=amd64 GOOS=darwin go build -mod=vendor -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/macos-amd64/crc ./cmd/crc
 
 $(BUILD_DIR)/linux-amd64/crc: $(SOURCES)
-	GOOS=linux GOARCH=amd64 go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/linux-amd64/crc ./cmd/crc
+	GOOS=linux GOARCH=amd64 go build -mod=vendor -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/linux-amd64/crc ./cmd/crc
 
 $(BUILD_DIR)/windows-amd64/crc.exe: $(SOURCES)
-	GOARCH=amd64 GOOS=windows go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/windows-amd64/crc.exe ./cmd/crc
+	GOARCH=amd64 GOOS=windows go build -mod=vendor -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/windows-amd64/crc.exe ./cmd/crc
 
 $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
-	go build --tags="build" -ldflags="$(LDFLAGS)" -o $(HOST_BUILD_DIR)/crc-embedder ./cmd/crc-embedder
[...]

PR #1769, opened 2020-12-10
Switch go.mod to 1.14

cfergeau:
This makes go automatically add -mod=vendor when needed so that
our vendor/ directory is used.
This fixes https://github.com/code-ready/crc/issues/1767

Some openshift components have already switched to 1.14 or newer in 4.6,
rhel8.3 has a new enough golang, so I think this can be done now without
waiting for 4.7

### DIFF
--- a/update-go-version.sh
+++ b/update-go-version.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Updates our various build files, CI configs, ... to the latest released go
+# version in a given minor release stream (1.x)
+# This requires yq and jq in addition to fairly standard shell tools (curl, grep, sed, ...)
+
+golang_base_version=$1
+latest_version=$(curl --silent  'https://golang.org/dl/?mode=json&include=all' | jq -r '.[].files[].version'  |uniq | sed -e 's/go//' |sort -V |grep ${golang_base_version}|tail -1)
+echo "Updating golang version to $latest_version"
+
+go mod edit -go ${golang_base_version}
+sed -i "s,^FROM registry.svc.ci.openshift.org/openshift/release:golang-1\... AS builder\$,FROM registry.svc.ci.openshift.org/openshift/release:golang-${golang_base_version} AS builder," images/openshift-ci/Dockerfile
+sed -i "s/GOVERSION: .*\$/GOVERSION: \"${latest_version}\"/" .circleci/config.yml
+sed -i "s/^GO_VERSION=.*$/GO_VERSION=${latest_version}/" centos_ci.sh
+yq write --inplace ./appveyor.yml  stack "go ${golang_base_version}"
+yq write --inplace ./.travis.yml go[0] ${latest_version}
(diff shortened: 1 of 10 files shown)


praveenkumar:
@cfergeau to incorporate this we also need to update our CI to use 1.14.7 (the one part of rhel right now).

cfergeau:
The last commit in this series is likely to break CI, because of missing yq/jq/... and maybe for other unforeseen reasons. This might need to be adjusted anyway, as I'm not sure we always want to set everything to the same go major version (in particular, I can see us wanting to keep the go.mod file set to 1.14, or the openshift ci Dockerfile set to something different while we switch other files to 1.15 ). This can be refined progressively :)

cfergeau:
I've removed `goversioncheck` from `check` for now, as this triggers various CI breakage, and I'd like not to hold up this for too long as it's likely to conflict with other changes.

guillaumerose:
retest this please
/lgtm

update-go-version.sh
```diff
@@ -0,0 +1,18 @@
...
+
+golang_base_version=$1
+latest_version=$(curl --silent  'https://golang.org/dl/?mode=json&include=all' | jq -r '.[].files[].version'  |uniq | sed -e 's/go//' |sort -V |grep ${golang_base_version}|tail -1)
+echo "Updating golang version to $latest_version"
+
+go mod edit -go ${golang_base_version}
+sed -i "s/^stack: go 1\...\$/stack: go ${golang_base_version}/" appveyor.yml
+sed -i "s,^FROM registry.svc.ci.openshift.org/openshift/release:golang-1\... AS builder\$,FROM registry.svc.ci.openshift.org/openshift/release:golang-${golang_base_version} AS builder," images/openshift-ci/Dockerfile
+sed -i "s/GOVERSION: .*\$/GOVERSION: \"${latest_version}\"/" .circleci/config.yml
+sed -i "s/^GO_VERSION=.*$/GO_VERSION=${latest_version}/" centos_ci.sh
```
praveenkumar, line 17: @cfergeau except centos ci all other CI config is in yaml, shouldn't we use `yq` for those instead `sed`?

cfergeau: Yup, good point. For circle-ci, I don't know how to do it cleanly, I tried `yq write --inplace ./.circleci/config.yml 'jobs[*].environment.(GOVERSION==*).GOVERSION' "${latest_version}"` but this adds empty `environment{}` fields in jobs which don't have it.

praveenkumar: @cfergeau what if we do it in 2 steps instead single line?
```
$ yq write --inplace ./.circleci/config.yml 'jobs.build.environment.GOVERSION' "${latest_version}"
$ yq write --inplace ./.circleci/config.yml 'jobs.build_mock_release.environment.GOVERSION' "${latest_version}"
```

cfergeau: This will break as soon as we rename one of these, or add a new one.

PR #1772, opened 2020-12-10
doc: Add systemd-resolved description to our DNS reference

cfergeau:
This will need to be iterated on after getting some Kevin's feedback, something similar would be needed for the remote configuration, ...

### DIFF
--- a/docs/source/topics/ref_dns-configuration.adoc
+++ b/docs/source/topics/ref_dns-configuration.adoc
@@ -17,7 +17,28 @@ Additional checks are done to verify DNS is properly configured when running [co
 [id="dns-configuration-linux_{context}"]
 == Linux
 
-On Linux, {prod} expects the following DNS configuration:
+On Linux, depending on your distribution, {prod} expects the following DNS configuration:
+
+=== NetworkManager + systemd-resolved
+
+This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions.
+
+* {prod} expects NetworkManager to manage networking.
+* {prod} configures `systemd-resolved` to forward requests for the `testing` domain to the `192.168.130.11` DNS server.
+`192.168.130.11` is the IP of the {prod} virtual machine.
+* `systemd-resolved` configuration is done through a NetworkManager dispatcher script in `/etc/NetworkManager/dispatcher.d/99-crc.sh`:
++
+----
+resolvectl domain crc ~testing
+resolvectl dns crc 192.168.130.11
+resolvectl default-route crc false
+
+exit 0
+----
+
+=== NetworkManager + dnsmasq
+
+This configuration is used on Fedora 32 or older, on {rhel}, and on {centos}.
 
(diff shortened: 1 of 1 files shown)


docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -17,7 +17,29 @@ Additional checks are done to verify DNS is properly configured when running [co
 [id="dns-configuration-linux_{context}"]
 == Linux
 
-On Linux, {prod} expects the following DNS configuration:
+On Linux, depending on your distribution, {prod} expects the following DNS configuration.
```
robin-owen, line 20: This sentence should end with a colon. :)

docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -17,7 +17,29 @@ Additional checks are done to verify DNS is properly configured when running [co
...
+
+=== NetworkManager + systemd-resolved
+
+This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions.
+
+* {prod} expects NetworkManager to manage networking.
+* {prod} configures systemd-resolved  to forward requests for the `testing` domain to the `192.168.130.11` DNS server. `192.168.130.11` is the IP of the {prod} virtual machine.
```
robin-owen, line 27: `systemd-resolved` should be formatted. This goes for the following line as well. :)

Also, could you please put each sentence on its own line? Thank you!

docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -17,7 +17,29 @@ Additional checks are done to verify DNS is properly configured when running [co
...
+        resolvectl dns "$1" 192.168.130.11
+        resolvectl default-route "$1" false
+fi
+
+exit 0
+----
+
+=== NetworkManager + dnsmasq
+
+This configuration is used on Fedora 32 or older, on {rhel} and on {centos}.
```
robin-owen, line 42: Missing a comma: "`This configuration is used on Fedora 32 or older, on {rhel}, and on {centos}.`"

robin-owen (commented):
Overall, LGTM -- I just see a few minor nitpicks I'd like to see fixed before approving. :)

Thank you! :+1:

cfergeau:
I've made a few more changes than what was requested to align this with https://github.com/code-ready/crc/pull/1819

docs/source/topics/ref_dns-configuration.adoc
cfergeau: `systemd-resolved` needs formatting, but not `NetworkManager`? I've made the change already, but looking back this seems inconsistent :)

robin-owen: NetworkManager is the name of the project, so we don't format it. You might have a point regarding systemd-resolved and whether we should be formatting that or not -- I'm not sure!

robin-owen (approved):

PR #1774, opened 2020-12-11
Add flag to crc start to automatically turn on monitoring stack

guillaumerose:
User can now use `enable-monitoring` flag with `crc start` and in
`crc config set`.

---

It can be improved by waiting for the cluster operator to be ready.

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -24,6 +24,7 @@ const (
 	HTTPSProxy           = "https-proxy"
 	NoProxy              = "no-proxy"
 	ProxyCAFile          = "proxy-ca-file"
+	EnableMonitoring     = "enable-monitoring"
 )
 
 func RegisterSettings(cfg *config.Config) {
@@ -42,6 +43,8 @@ func RegisterSettings(cfg *config.Config) {
 	cfg.AddSetting(HTTPSProxy, "", config.ValidateURI, config.SuccessfullyApplied)
 	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied)
 	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied)
+
+	cfg.AddSetting(EnableMonitoring, false, config.ValidateBool, config.SuccessfullyApplied)
 }
 
 func isPreflightKey(key string) bool {
--- a/pkg/crc/cluster/monitoring.go
+++ b/pkg/crc/cluster/monitoring.go
@@ -1,0 +1,39 @@
+package cluster
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/code-ready/crc/pkg/crc/oc"
+	v1 "github.com/openshift/api/config/v1"
+	log "github.com/sirupsen/logrus"
+)
+
+func StartMonitoring(ocConfig oc.Config) error {
+	data, _, err := ocConfig.RunOcCommand("get", "clusterversion/version", "-o", "json")
+	if err != nil {
+		return err
+	}
+
+	var cv v1.ClusterVersion
+	if err := json.Unmarshal([]byte(data), &cv); err != nil {
+		return err
+	}
+
+	pos := -1
+	for i, override := range cv.Spec.Overrides {
+		if override.Name == "cluster-monitoring-operator" {
+			pos = i
+			break
+		}
+	}
+	if pos == -1 {
+		log.Debug("monitoring operator not found in cluster version overrides")
+		return nil
+	}
+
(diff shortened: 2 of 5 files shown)


jsliacan:
I assume I can then change #1736 to a one-liner using the flag?

guillaumerose:
If it is accepted, yes ^^ 
I thought it would be a good UX enhancement for users.

cmd/crc/cmd/config/config.go
```diff
@@ -24,6 +24,7 @@ const (
 	HTTPSProxy           = "https-proxy"
 	NoProxy              = "no-proxy"
 	ProxyCAFile          = "proxy-ca-file"
+	EnableMonitoring     = "enable-monitoring"
```
gbraad, line 27: @praveenkumar enable should only be for cluster? Or should we be more descriptive here?

anjannath: You mean name it as `enable-cluster-monitoring` instead?

praveenkumar: @gbraad yes, it is cluster scope only, more descriptive mean explicitly mention cluster like @anjannath suggested?

gbraad:
Enabling monitoring might need to enforce validations to ensure the user assigns more memory than the default.

pkg/crc/machine/machine.go
```diff
@@ -40,6 +41,8 @@ import (
 	"github.com/code-ready/machine/libmachine/state"
 )
 
+const MinimumMemoryForMonitoring = 13671
```
gbraad, line 44: How did we get this value?

anjannath (approved):
looks good 👍🏽  would be nice to also have a way to disable monitoring in future.

gbraad:
> looks good 👍🏽 would be nice to also have a way to disable monitoring in future.

Would not be needed as a disable does not exist. We discussed this on the docs meeting that the user in that case has to delete the cluster; it is ephemeral anyway. besides, it is should have been a default of the cluster. on the installed cluster this is not disabled so this should not be something the user 'wishes' for.

cmd/crc/cmd/config/config.go
guillaumerose: I don't see what kind of information `cluster` is adding. Is there an other kind of monitoring we would like to add ? 

Either we go with `monitoring` or `cluster-monitoring-operator`, but not half of the name.

pkg/crc/machine/machine.go
guillaumerose: It comes from a misunderstading. Doc says 14336, I will change it.

However, there are mistakes in the code: we use `units.HumanSize` (base 1000) instead of `units.BytesSize` (base 1024). I will add a commit to fix this.

```
// HumanSize returns a human-readable approximation of a size
// capped at 4 valid numbers (eg. "2.746 MB", "796 KB").
func HumanSize(size float64) string {
	return HumanSizeWithPrecision(size, 4)
}

// BytesSize returns a human-readable size in bytes, kibibytes,
// mebibytes, gibibytes, or tebibytes (eg. "44kiB", "17MiB").
func BytesSize(size float64) string {
	return CustomSize("%.4g%s", size, 1024.0, binaryAbbrs)
}
```

pkg/crc/cluster/monitoring.go
```diff
@@ -0,0 +1,37 @@
...
+
+	var cv v1.ClusterVersion
+	if err := json.Unmarshal([]byte(data), &cv); err != nil {
+		return err
+	}
+
+	pos := -1
+	for i, override := range cv.Spec.Overrides {
+		if override.Name == "cluster-monitoring-operator" {
+			pos = i
```
praveenkumar, line 26: shouldn't we use `break` as soon as we find this operator and the position?

guillaumerose: 👍

pkg/crc/machine/machine.go
guillaumerose: It turns out to be more complicated for df. I will do that in a separated PR.

pkg/crc/machine/machine.go
```diff
@@ -441,6 +455,15 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 	}, nil
 }
 
+func validateStartConfig(startConfig StartConfig) error {
+	if startConfig.EnableMonitoring && startConfig.Memory < minimumMemoryForMonitoring {
+		return fmt.Errorf("Too few memory (%s) allocated to the virtual machine to start the monitoring stack, %s is the minimum",
```
cfergeau, line 460: "Too little memory"

cfergeau (commented):
Since this is a case when we differ from a real openshift cluster, and since re-enabling it needs non trivial steps, imo it's good to offer it as part of crc command line.
Code looks good to me.

PR #1777, opened 2020-12-14
Implement log rotation for current logger instance

praveenkumar:
With this patch the log file will autorotated when size of the log
file is reach to 1 MiB or log file is 10 days old.

## Testing

```
$ dd if=/dev/zero of=~/.crc/crc.log bs=1024 count=1024
1024+0 records in
1024+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.00477907 s, 219 MB/s

$ crc setup

$ ls ~/.crc/
crc-2020-12-14T10-51-54.748.log  crc.log
```


cfergeau:
I'd much prefer we go with https://github.com/code-ready/crc/pull/1758 and don't implement log rotation ourselves.

PR #1779, opened 2020-12-15
Return stderr in case of err for ssh commands

praveenkumar:
In case of err, we are just sending the error not the content of stderr
which makes bit hard to debug some issues.

Before this patch
```
oc get secret --context admin --cluster crc --kubeconfig /opt/kubeconfig
DEBU SSH command results: err: Process exited with status 1, output:
DEBU
command : oc get secret --context admin --cluster crc --kubeconfig /opt/kubeconfig
```

After this patch
```
oc get secret --context admin --cluster crc --kubeconfig /opt/kubeconfig
DEBU SSH command results: err: Process exited with status 1, output:
DEBU Unable to connect to the server: x509: certificate has expired or is not yet va
lid: current time 2020-12-15T11:30:30Z is after 2020-12-07T20:40:59Z
command : oc get secret --context admin --cluster crc --kubeconfig /opt/kubeconfig
```

### DIFF
--- a/pkg/crc/ssh/ssh.go
+++ b/pkg/crc/ssh/ssh.go
@@ -84,10 +84,9 @@ func (runner *Runner) runSSHCommand(command string, runPrivate bool) (string, st
 	}
 
 	if err != nil {
-		return "", "", fmt.Errorf(`ssh command error:
+		return string(stdout), string(stderr), fmt.Errorf(`ssh command error:
 command : %s
-err     : %v
-output  : %s`, command, err, string(stdout))
+err     : %v\n`, command, err)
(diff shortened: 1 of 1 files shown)


cfergeau (commented):
This seems to be a regression introduced in `ssh: Separate stderr from stdout in ssh`? Why not add stderr to `fmt.Errorf('ssh command error:
command : %s
err     : %v
output  : %s`, command, err, string(stdout))'`?

I prefer the way you do it though, but then I'd remove `output:%s` from `fmt.Errorf`

cfergeau (approved):

PR #1780, opened 2020-12-15
Add wait for apiserver function

praveenkumar:
With current scenario, we start the kubelet and then check for secret resource
(80 sec wait+retry), most of the time during this retry the apiserver come up
and start serving the requests.

Recently we found out that during openshift dev cycle the cert rotation is set for
12 hours and our generated bundles are not able to modify those to 30 days one. This
change affect the serving and aggregator certs. With this patch we are trying to add
another waiting ~2 mins with (retry) to provide time for apisever to start serving to
the incoming requests.

- https://bugzilla.redhat.com/show_bug.cgi?id=1906290
- https://bugzilla.redhat.com/show_bug.cgi?id=1883790

As per test it takes around ~100sec before apiserver become responsive.
```
oc get nodes --context admin --cluster crc --kubeconfig /opt/kubeconfig
DEBU SSH command results: err: Process exited with status 1, output:
DEBU Unable to connect to the server: x509: certificate has expired or is not yet valid: current time 2020-12-15T12:12:49Z is
after 2020-12-07T20:40:59Z
command : oc get nodes --context admin --cluster crc --kubeconfig /opt/kubeconfig
output  :  - sleeping 1s
DEBU retry loop: attempt 56
oc get nodes --context admin --cluster crc --kubeconfig /opt/kubeconfig
DEBU SSH command results: err: <nil>, output: NAME                 STATUS   ROLES           AGE   VERSION
crc-qcmcx-master-0   Ready    master,worker   9d    v1.19.2+ad738ba
DEBU NAME                 STATUS   ROLES           AGE   VERSION
crc-qcmcx-master-0   Ready    master,worker   9d    v1.19.2+ad738ba
```

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -307,6 +307,20 @@ func WaitForRequestHeaderClientCaFile(sshRunner *ssh.Runner) error {
 	return errors.RetryAfter(8*time.Minute, lookupRequestHeaderClientCa, 2*time.Second)
 }
 
+func WaitForAPIServer(ocConfig oc.Config) error {
+	logging.Debugf("Waiting for apiserver availability")
+	waitForAPIServer := func() error {
+		stdout, stderr, err := ocConfig.RunOcCommand("get", "nodes")
+		if err != nil {
+			logging.Debug(stderr)
+			return &errors.RetriableError{Err: err}
+		}
+		logging.Debug(stdout)
+		return nil
+	}
+	return errors.RetryAfter(2*time.Minute, waitForAPIServer, time.Second)
+}
+
 func DeleteOpenshiftAPIServerPods(ocConfig oc.Config) error {
 	if err := WaitForOpenshiftResource(ocConfig, "pod"); err != nil {
 		return err
--- a/pkg/crc/machine/machine.go
+++ b/pkg/crc/machine/machine.go
@@ -375,6 +375,10 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 		return nil, errors.Wrap(err, "Failed to renew TLS certificates: please check if a newer CodeReady Containers release is available")
 	}
 
+	if err := cluster.WaitForAPIServer(ocConfig); err != nil {
+		return nil, errors.Wrap(err, "Error waiting for apiserver")
+	}
+
 	if !exists {
 		logging.Info("Configuring cluster for first start")
 		if err := configProxyForCluster(ocConfig, sshRunner, sd, proxyConfig, instanceIP); err != nil {


cfergeau:
What is the impact on a non-devel openshift cluster? One where cert renewal gives us 1 month validity?

praveenkumar:
> What is the impact on a non-devel openshift cluster? One where cert renewal gives us 1 month validity?

No impact, with this patch.

guillaumerose:
I would prefer to increase WaitForOpenshiftResource to 80s+2min or so. It's more general and doesn't change the current flow.
Since we cannot recover from a failing start, I hope this will imply a higher success rate (but a longer startup time maybe?). Segment will help for that!

It will be hard to remove this wait function. In 6 months, we will come back and we will wonder: "What will break if we remove this?"

praveenkumar:
> I would prefer to increase WaitForOpenshiftResource to 80s+2min or so. It's more general and doesn't change the current flow.
Since we cannot recover from a failing start, I hope this will imply a higher success rate (but a longer startup time maybe?). Segment will help for that!

@guillaumerose this is what came to my mind first but then in long run if the rotation take more time then we need to again increase the waitForResource bit, instead of that having a separate bit for apiserver makes bit more sense because once the apiserver is available, fetching resources shouldn't take much time, even if apiserver reconcile then also putting it in 80sec make more sense then adding more time here.

guillaumerose:
Just before the wait you added, we run `ApproveCSRAndWaitForCertsRenewal`. This function already needs a running API server (oc get csr/oc adm certificate approve).

Correct me if I'm wrong, but this is the flow I see:

--- 

1. VM starts
2. kubelet starts and re-launches previously running containers
3. apiserver starts with expired certs
4. controller-manager starts and tries to connect to apiserver. It can't due to cert issue and enters master disruption mode.
5. after a period of time, controller-manager (or someone else) re-generate the apiserver cert.
6. apiserver reloads the cert

meanwhile..

4a. kubelet also tries to connect to apiserver and can't.
6a. when apiserver is back again, it creates a new csr to be approved.
6b. crc approves the csr, controller-manager generates the cert.
6c. kubelet starts to refresh what containers are running, operators wake up (and potentially reboot the apiserver couple of times).
6d. kubelet asks for a new serving cert. cluster machine approver says yes. controller-manager generates it.

7. (then, we are at the point, you add the wait apiserver).
8. crc performs some changes on the cluster, which lead to potential restart of the apiserver.
9. crc waits for a new aggregation certificate and then delete openshift-apiserver.
10. wait 3min

---

I think if the issue is the API cert, we should check explicitly for it and wait for it before the kubelet cert rotation. (use of `cluster.CheckCertsValidity` and `waitForCertRenewal`).

If we add this `WaitForAPIServer` here, can I assume that the apiserver will always be up after this point ? Can I remove all `WaitForOpenshiftResource` after this ?

WDYT?

praveenkumar:
> Correct me if I'm wrong, but this is the flow I see:

@guillaumerose The flow which you provided is correct one and this is the way things are right now.

> I think if the issue is the API cert, we should check explicitly for it and wait for it before the kubelet cert rotation. (use of cluster.CheckCertsValidity and waitForCertRenewal).

Yes this is the issue with API serving cert and if you prefer we can add that check, I still need to find the location of this cert  because I can't get it from api side since it is not working may be something like `openssl s_client -connect api.crc.testing:6443 | openssl x509 -noout -dates` ?

> If we add this WaitForAPIServer here, can I assume that the apiserver will always be up after this point ? Can I remove all WaitForOpenshiftResource after this ?

This check provide an additional cover for us in case the apiserver is reconcile so I don't want to remove it.

guillaumerose:
Let's go with this. I hope we will be able to improve that area.

/lgtm

guillaumerose (approved):

PR #1781, opened 2020-12-16
Fixes for tray and daemon installation in windows

anjannath:
Fixes: #1637 

## Solution/Idea

Instead of running the daemon as a windows service, which always runs with admin privileges if the user account is an Admin, run the daemon as a simple background process (hidden window).

## Proposed changes

Create a PowerShell script that starts the daemon process in a hidden window, then create another batch file in `"C:\Users\Anjan\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup"` folder to make that script (daemon launch PS script) auto-start on user login.

During `crc setup` we perform the above steps and also execute the PowerShell script so that we have the daemon running.
## Testing

1. `crc setup --enable-experimental-features` should successfully start the daemon and tray
2. on reboot the tray and daemon is auto-started
3. `crc cleanup` should completely remove the tray and daemon


cfergeau:
> Create a PowerShell script that starts the daemon process in a hidden window, then create another batch file in "C:\Users\Anjan\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup" folder to make that script (daemon launch PS script) auto-start on user login.

What happens if it's not a hidden window? There would be a text console appearing on the screen? Why does this need to be running for the whole user session though? Could the daemon be started automatically when the tray does an api call, and the daemon is not running already? (this probably can be improved on later).

Detailed commit logs appear to be missing.

anjannath:
>What happens if it's not a hidden window? There would be a text console appearing on the screen? Why does this need to be running for the whole user session though?

Yes, it opens a text console and shows all the o/p from the daemon, this needs to be running as we want to poll the status of the cluster, to show the user correct status on the tray icon also the vsock stuff is run alogn with the daemon and that needs to also keep running for the whole user session

> Could the daemon be started automatically when the tray does an api call, and the daemon is not running already? (this probably can be improved on later).

Yeah, right now if the daemon is not running calling the api won't start it automatically, only `crc setup` sets it up and starts it

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,126 +16,82 @@ import (
...
 
 func checkIfTrayInstalled() error {
-	/* We want to force installation whenever setup is ran
-	 * as we want the service config to point to the executable
-	 * with which the setup command was issued.
-	 */
-
-	return fmt.Errorf("Ignoring check and forcing installation of System Tray")
+	// Check if tray exists in the startup folder
+	// And version of the tray is matching with expected version
```
cfergeau, line 22: Is this a TODO? If it is, can you add `TODO`, `FIXME`,. ...before the comment to make it obvious?
Edit: the check is changed in a later commit to stop returning an error unconditionally, but I still think the comment is not 100% correct.
Edit2: this is changed again in a 3rd commit, and the comment then becomes accurate. I'd suggest only adding the comment in the last commit, or removing the comment, the code is easy enough to read (because constants are used, function name is good), so the comment is just repetition?

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,126 +16,82 @@ import (
...
-	defer func() {
-		_ = goos.RemoveAll(tempDir)
-	}()
-
-	// prepare the ps script
-	binPath, err := goos.Executable()
+		return fmt.Errorf("Failed to find current executables path: %w", err)
+	}
+
+	// Create the Posh script to launch daemon in the background with hidden window
```
cfergeau, line 41: Regarding this hidden window thing, did you try https://stackoverflow.com/questions/23250505/how-do-i-create-an-executable-from-golang-that-doesnt-open-a-console-window-whe

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,126 +16,82 @@ import (
...
+		return fmt.Errorf("Failed to install CodeReady Containers daemon: %w", err)
 	}
-	if strings.Contains(password, "\"") {
-		password = strings.ReplaceAll(password, "\"", "`\"")
+	cmd := fmt.Sprintf(`New-Item -ItemType SymbolicLink -Path "%s" -Name "%s" -Value "%s"`,
+		constants.StartUpFolder,
+		constants.TrayShortcutName,
+		constants.TrayExecutablePath,
+	)
+	if _, _, err := powershell.ExecuteAsAdmin("Copying tray and daemon executable to start-up folder", cmd); err != nil {
```
cfergeau: Does this have to be done as an admin? `err = ioutil.WriteFile(constants.DaemonBatchFilePath, []byte(daemonBatchFileContent), 0600)` earlier is in the same directory I believe, and did not require admin?

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,126 +16,82 @@ import (
...
 	}
-
-	_, _, err = powershell.ExecuteAsAdmin("Uninstalling CodeReady Containers System Tray", psFilePath)
-	// wait for the script to finish executing
-	time.Sleep(2 * time.Second)
-	if err != nil {
-		return err
+	if err := goos.Remove(constants.DaemonBatchFilePath); err != nil {
+		return fmt.Errorf("Failed to remove daemon auto start script from startup folder: %w", err)
 	}
```
cfergeau: `constants.DaemonPSscriptPath` is missing? (not removed)

cfergeau (commented):
It might make sense to squash ` Split daemon and tray preflight checks into two set of checks ` with the previous commit, you are adding new code to perform the installation of the daemon, so you could place it in its own function from the start, it does not make the diffstat much bigger:
```
 pkg/crc/constants/constants_windows.go                  |  15 ++++--
 pkg/crc/preflight/preflight_checks_tray_windows.go      | 178 +++++++++++++++++++++--------------------------------------------------
 pkg/crc/preflight/preflight_checks_tray_windows_test.go |  29 ------------
 pkg/crc/preflight/preflight_tray_powershell_windows.go  | 170 -------------------------------------------------------------------
 4 files changed, 64 insertions(+), 328 deletions(-)
```
for just `Setup daemon as a startup application instead of service`

and 

```
 pkg/crc/constants/constants_windows.go                  |  15 ++++--
 pkg/crc/preflight/preflight_checks_tray_windows.go      | 188 ++++++++++++++++++++++++-----------------------------------------------
 pkg/crc/preflight/preflight_checks_tray_windows_test.go |  29 -----------
 pkg/crc/preflight/preflight_tray_powershell_windows.go  | 170 ----------------------------------------------------------------
 pkg/crc/preflight/preflight_windows.go                  |  14 ++++++
 pkg/crc/preflight/preflight_windows_test.go             |   4 +-
 6 files changed, 89 insertions(+), 331 deletions(-)
```
for the 2 commits together. I did not check which one is easier to read, but I know that while looking at the first commit, I was thinking the checks should be split ^^

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Removed the comment!

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Strangely it needs `ExecuteAsAdmin` to copy the file, it could be that the powershell cmdlet that is being used needs an elevated shell, will try with golang `os.Symlink` and if that works we can remove the `ExecuteAsAdmin`

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Tried this, but it doesn't seem to help with our situation..
- to run the daemon we need the command `crc daemon` so, we'll still need to have a posh script and a batch file to drop in the startup folder
- after compiling with `windowsgui` it doesn't show o/p in the console, so we can't use the same binary for CLI

anjannath:
Rebased and updated, PTALA!

cfergeau:
> this needs to be running as we want to poll the status of the cluster, to show the user correct status on the tray icon also the vsock stuff is run along with the daemon and that needs to also keep running for the whole user session

The crc daemon process needs to be running:
- when the tray is running
- when the crc VM is running with vsock networking

You addressed this by making sure `crc daemon` for the whole user session, but a more fine grained approach would be possible.

pkg/crc/preflight/preflight_checks_tray_windows.go
cfergeau: Ah the second point rules this approach out :)

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,130 +17,119 @@ import (
...
-	 * this takes care of all the steps needed to have a running
-	 * tray by invoking a ps script which does the following:
-	 * a) Add logon as service permission for the user
-	 * b) Create the daemon service and start it
-	 * c) Add tray to startup folder and start it
-	 */
+func checkIfDaemonInstalled() error {
+	// We want to force the installation of daemon as we want to
+	// run the daemon from the crc executable which ran the setup
+	return fmt.Errorf("Ignoring check, forcing installation of daemon")
```
cfergeau, line 31: will that cause issue if one runs `crc setup` when the daemon is already running?

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,130 +17,119 @@ import (
...
+$psi.WindowStyle = 'Hidden'
+$newproc.StartInfo = $psi
+$newproc.Start()
+$newproc
+`, currentExecutablePath, "daemon --log-level debug")
+
+	err = ioutil.WriteFile(constants.DaemonPSscriptPath, []byte(daemonLaunchPSscriptContent), 0600)
 	if err != nil {
-		return fmt.Errorf("Unable to find the current executables location: %v", err)
+		return fmt.Errorf("Failed to create required PS script file for the daemon: %w", err)
```
cfergeau, line 54: PS is PowerShell I think? I'd expand the acronym in the error message, unless it's very common terminology in Windows.

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,130 +17,119 @@ import (
...
 	}
-	binPathWithArgs := fmt.Sprintf("%s daemon", strings.TrimSpace(binPath))
 
-	// get the password from user
-	password, err := input.PromptUserForSecret("Enter account login password for service installation", "This is the login password of your current account, needed to install the daemon service")
+	// Create the cmd file to start the Posh script at login
```
cfergeau, line 57: Posh script == PS script? Or are they different?

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,130 +17,119 @@ import (
...
 
-	psScriptContent := genTrayRemovalScript(
-		trayProcessName,
-		constants.TrayShortcutName,
-		constants.DaemonServiceName,
-		tempDir,
-	)
-	psFilePath := filepath.Join(tempDir, "trayRemoval.ps1")
+	logging.Info("Please run the following powershell command to finish the cleanup process.")
+	logging.Info("This will stop the CodeReady Containers daemon process: '(Get-Process crc).kill()'")
```
cfergeau, line 114: Why aren't we doing it ourselves?

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,130 +17,119 @@ import (
...
+	trayProcessName := constants.TrayExecutableName[:len(constants.TrayExecutableName)-4]
+	cmd := fmt.Sprintf(`Stop-Process -Name "%s"`, trayProcessName)
+	if _, _, err := powershell.Execute(cmd); err != nil {
+		return fmt.Errorf("Failed to kill tray process: %w", err)
 	}
 	return nil
 }
 
+func removeDaemon() error {
+	return goos.Remove(constants.DaemonBatchFilePath)
```
cfergeau, line 132: You already remove this in removeTray()?

cfergeau (commented):
Still some minor comments, but looks good overall.

pkg/crc/preflight/preflight_checks_tray_windows.go
cfergeau: Yup, odd :-/

pkg/crc/preflight/preflight_checks_tray_windows.go
gbraad: Posh is the otyer name, also given to the Linux binary. But both are generally OK

gbraad: Externally, to the user we should say PowerShell

pkg/crc/preflight/preflight_checks_tray_windows.go
gbraad: Expand

pkg/crc/preflight/preflight_checks_tray_windows.go
gbraad: Killing yourself? Hmmmm perhaps this was the issue?

pkg/crc/constants/constants_windows.go
```diff
@@ -1,17 +1,25 @@
...
+	TrayShortcutName            = "tray-windows.lnk"
+	DaemonBatchFileName         = "crc-daemon-autostart.bat"
+	DaemonPSscriptName          = "launch-crc-daemon.ps1"
+	DaemonBatchFileShortcutName = "crc-daemmon-autostart.bat.lnk"
 )
 
 var (
-	TrayExecutableDir  = filepath.Join(CrcBinDir, "tray-windows")
-	TrayExecutablePath = filepath.Join(TrayExecutableDir, TrayExecutableName)
+	StartUpFolder       = filepath.Join(os.Getenv("USERPROFILE"), "AppData", "Roaming", "Microsoft", "Windows", "Start Menu", "Programs", "Startup")
```
gbraad, line 20: s/StartUp/Startup

pkg/crc/constants/constants_windows.go
```diff
@@ -1,17 +1,25 @@
...
+	TrayExecutableDir   = filepath.Join(CrcBinDir, "tray-windows")
+	TrayExecutablePath  = filepath.Join(TrayExecutableDir, TrayExecutableName)
+	DaemonBatchFilePath = filepath.Join(CrcBinDir, DaemonBatchFileName)
+	DaemonPSscriptPath  = filepath.Join(CrcBinDir, DaemonPSscriptName)
```
gbraad, line 24: PSScript, as before,  although this double PS can be avoided if internally named Posh.

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,130 +19,119 @@ import (
...
-		_ = goos.RemoveAll(tempDir)
-	}()
-
-	// prepare the ps script
-	binPath, err := goos.Executable()
+		return fmt.Errorf("Failed to find current executables path: %w", err)
+	}
+
+	// Create the Posh script to launch daemon in the background with hidden window
+	daemonLaunchPSscriptContent := fmt.Sprintf(`$psi = New-Object System.Diagnostics.ProcessStartInfo
```
gbraad, line 41: PSscript

pkg/crc/preflight/preflight_checks_tray_windows.go
gbraad: You should be able to detect the other process by the identifier. Though, this needs additional code.
Refinement or essential?

gbraad: So what is stopTray?

gbraad (changes requested):
Some unclear code, especially around process stop.

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: @gbraad `stopTray` stops the tray process if one is running, it doesn't remove its symlink from the startup folder, in the `removeTray` we are removing those symlink from the startup folder

anjannath: > You should be able to detect the other process by the identifier. Though, this needs additional code.
> Refinement or essential?

Yes, was looking at this, will update with the additional code needed to handle this properly

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Sorry, will stick to a single name "PSScript" internally and for user facing text "PowerShell"

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Yup, messed up during the squash i think, will fix this

pkg/crc/preflight/preflight_checks_tray_windows.go
gbraad: You are suggesting:

```
(Get-Process crc).kill()
```
and yu have a `stopTray` function... so this puzzles me ;-)

anjannath: the tray process is different from the daemon, `stopTray`closes the tray but not the crc daemon, that command was to kill the daemon process, but in the latest commit, this is removed, and the daemon is killed during cleanup by finding its pid

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: This was causing issue, now during the fix we call `stopDaemon` which makes sure that the socket is usable and the issue is fixed.

pkg/crc/preflight/preflight_checks_tray_windows.go
cfergeau: I think this was supposed to be `daemonLaunchPSScriptContent`, in line with the others `PSScript` occurrences.

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,126 +18,139 @@ import (
...
-	// check for 'success' file
-	if _, err = goos.Stat(filepath.Join(tempDir, "success")); goos.IsNotExist(err) {
-		return fmt.Errorf("Installation script didn't execute successfully: %v", err)
+func removeTray() error {
+	_ = stopTray()
+	return goos.Remove(filepath.Join(constants.StartupFolder, constants.TrayShortcutName))
+}
+
+func stopTray() error {
+	trayProcessName := constants.TrayExecutableName[:len(constants.TrayExecutableName)-4]
```
cfergeau, line 111: This should use `TrimSuffix`

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -19,126 +18,139 @@ import (
...
-	}()
-
-	psScriptContent := genTrayRemovalScript(
-		trayProcessName,
-		constants.TrayShortcutName,
-		constants.DaemonServiceName,
-		tempDir,
+	// get the PID of the process running command 'crc daemon --log-level debug'
+	getCrcProcessCmd := `(Get-WmiObject Win32_Process -Filter "name = 'crc.exe'")`
+	cmd := fmt.Sprintf(`(%s | Where-Object -Property CommandLine -Eq '"%s" daemon --log-level debug').ProcessId`,
```
cfergeau, line 141: Hardcoding `daemon --log-level` here is fragile, this is going to silently break the day we change the command line used to spawn the daemon

cfergeau (commented):
Still a few minor comments, which can probably be addressed in followup PRs.

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Done!

pkg/crc/preflight/preflight_checks_tray_windows.go
anjannath: Added https://github.com/code-ready/crc/pull/1781/commits/8692885447bf55a292c94eb0cf59ec95574845e7 to handle this!

pkg/crc/preflight/preflight_checks_tray_windows.go
```diff
@@ -43,13 +43,13 @@ func fixDaemonInstalled() error {
...
 $psi.FileName = "%s"
-$psi.Arguments = "%s"
+$psi.Arguments = "%s %s"
 $psi.CreateNoWindow = $true
 $psi.WindowStyle = 'Hidden'
 $newproc.StartInfo = $psi
 $newproc.Start()
 $newproc
-`, currentExecutablePath, "daemon --log-level debug")
+`, currentExecutablePath, constants.DaemonCommandName, "--log-level debug")
```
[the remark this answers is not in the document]
anjannath: Ohh, sorry i interpreted it wrong then, so what we want is to keep this full command with which the pid is found same as the command in the daemon script (so the constant should be used for this) will send a follow up PR this.

PR #1783, opened 2020-12-16
Integration: Add check for enabled monitoring stack

jsliacan:
**Relates to:** PR #1774, PR #1787 and [this comment](https://github.com/code-ready/crc/pull/1774#issuecomment-744272906).

## Proposed changes

Add two scenarios to `basic.feature` (`enable-openshift-monitoring` in config settings; check if related pods are running). The check happens after `crc status` which waits for the cluster monitoring operator to come up.

This test belongs in this feature (imo) because the `enable-monitoring` functionality is "basic" in the sense that it's default on OCP and by default turned off in CRC. Plus, `basic.feature` does not deploy other resource using apps/operators/etc. So there's no worry of interfering. 

## Testing

1. `make integration` with a `@basic` tag 'on'. Expect a pass.

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -85,6 +85,14 @@ Feature: Basic test
         Then stderr should contain "Checking if the Hyper-V virtual switch exist"
 
     @darwin @linux @windows
+    Scenario: Request start with monitoring stack
+        When setting config property "enable-cluster-monitoring" to value "true" succeeds
+        And setting config property "memory" to value "14000" succeeds
+        Then starting CRC with default bundle fails
+        And stderr should contain "Too little memory"
+        And setting config property "memory" to value "16000" succeeds
+
+    @darwin @linux @windows
     Scenario: CRC start
         When starting CRC with default bundle succeeds
         Then stdout should contain "Started the OpenShift cluster"
@@ -112,6 +120,16 @@ Feature: Basic test
         When executing "crc console --credentials" succeeds
         Then stdout should contain "To login as a regular user, run 'oc login -u developer -p developer"
         And stdout should contain "To login as an admin, run 'oc login -u kubeadmin -p "
+
+    @darwin @linux @windows
+    Scenario: Monitoring stack check
+        Given checking that CRC is running
+        When executing "eval $(crc oc-env)" succeeds
+        And login to the oc cluster succeeds
+        And executing "oc get pods -n openshift-monitoring" succeeds
+        Then stdout matches ".*cluster-monitoring-operator-\w+-\w+\ *2/2\ *Running.*"
+        And unsetting config property "enable-cluster-monitoring" succeeds
+        And unsetting config property "memory" succeeds
 
     @darwin @linux @windows
     Scenario: CRC forcible stop


test/integration/features/basic.feature
```diff
@@ -113,6 +121,13 @@ Feature: Basic test
         Then stdout should contain "To login as a regular user, run 'oc login -u developer -p developer"
         And stdout should contain "To login as an admin, run 'oc login -u kubeadmin -p "
 
+    @darwin @linux @windows
+    Scenario: Monitoring stack check
+        Given checking that CRC is running
+        When login to the oc cluster succeeds
+        And executing "oc get pods -n openshift-monitoring" succeeds
```
praveenkumar, line 128: It doesn't come up as soon as the we exclude it from CVO override, it does take some time so may be you need to put some retry here.

anjannath:
retest this please

test/integration/features/basic.feature
jsliacan: Updated the description of the PR to reflect changes brought in by #1787. That also addresses your concern -- `crc status` is executed before we check the pods, and status waits for the monitoring operator. Should be OK now.

cfergeau:
> This test belongs in this feature (imo) because the enable-monitoring functionality is "basic" in the sense that it's default on OCP and by default turned off in CRC. Plus, basic.feature does not deploy other resource using apps/operators/etc. So there's no worry of interfering.

Do we still have a test that `crc start` works without enabling monitoring/with the default memory settings?

jsliacan:
> Do we still have a test that crc start works without enabling monitoring/with the default memory settings?

@cfergeau , yes, thanks for checking. It's a bit confusing, but `story_registry` should be ["all default" start](https://github.com/code-ready/crc/blob/7b098c1bcaad0422dad36a390f8b335a9554548e/test/integration/features/story_registry.feature#L10).

test/integration/features/basic.feature
```diff
@@ -84,6 +84,14 @@ Feature: Basic test
         Then stderr should contain "Checking if user is a member of the Hyper-V Administrators group"
         Then stderr should contain "Checking if the Hyper-V virtual switch exist"
 
+    @darwin @linux @windows
+    Scenario: Request start with monitoring stack
+        When setting config property "enable-openshift-monitoring" to value "true" succeeds
```
praveenkumar, line 89: @jsliacan This is changed to `enable-cluster-monitoring` so it need to be updated.

praveenkumar (changes requested):

praveenkumar:
> > This test belongs in this feature (imo) because the enable-monitoring functionality is "basic" in the sense that it's default on OCP and by default turned off in CRC. Plus, basic.feature does not deploy other resource using apps/operators/etc. So there's no worry of interfering.
> 
> Do we still have a test that `crc start` works without enabling monitoring/with the default memory settings?

@jsliacan I think cleanup doesn't remove the config file so better to reset the memory once the test execute to have other tests to run with defaults.

jsliacan:
> @jsliacan I think cleanup doesn't remove the config file so better to reset the memory once the test execute to have other tests to run with defaults.

@praveenkumar , I missed that. Thanks! Updated.

test/integration/features/basic.feature
```diff
@@ -113,6 +121,15 @@ Feature: Basic test
         Then stdout should contain "To login as a regular user, run 'oc login -u developer -p developer"
         And stdout should contain "To login as an admin, run 'oc login -u kubeadmin -p "
 
+    @darwin @linux @windows
+    Scenario: Monitoring stack check
+        Given checking that CRC is running
+        When login to the oc cluster succeeds
+        And executing "oc get pods -n openshift-monitoring" succeeds
+        Then stdout matches ".*cluster-monitoring-operator-[0-9a-z]{10}-[0-9a-z]{5}\ *2/2\ *Running.*"
```
praveenkumar, line 129: As per CI output the pod name doesn't have 10 char so better to use `.*cluster-monitoring-operator-[0-9a-z]+-[0-9a-z]+\ *2/2\ *Running.*` so we are not constraint to char number.

jsliacan: Yes, just found out that max is 10, but could be less. Thanks! Also, what I don't understand is how come `oc login` succeeded when the test does not export `oc`'s location (`eval $(crc oc-env)`)... ci.centos failed on it but the prow one didn't.

praveenkumar: @jsliacan on e2e-crc the VM have oc in the path also :(

test/integration/features/basic.feature
```diff
@@ -113,6 +121,16 @@ Feature: Basic test
...
 
+    @darwin @linux @windows
+    Scenario: Monitoring stack check
+        Given checking that CRC is running
+        When executing "eval $(crc oc-env)" succeeds
+        And login to the oc cluster succeeds
+        And executing "oc get pods -n openshift-monitoring" succeeds
+        Then stdout matches ".*cluster-monitoring-operator-\w+-\w+\ *2/2\ *Running.*"
+        And unsetting config property "enable-cluster-monitoring" succeeds
+        And unsetting config property "memory" succeeds
```
praveenkumar, line 132: 👍🏼

praveenkumar (approved):

Issue #1753, opened 2020-12-04
[Integration Test] Review story marketplace feature


adrianriobo:
Story marketplace feature covers the installation of an operator from Operator Hub Catalog. The current test uses etcd operator, at this moment the etcd operator uses an init container busybox:1.28.0-glibc from docker hub registry to install new cluster instances.

With the new policy from docker hub the execution of this test may lead to unexpected errors. 

Use a different operator with no relying on docker hub is much preferred for testing consistency

adrianriobo:
It is expected to to overcome the dependency with docker hub and maintain the small payload on top of the cluster to cover this feature. According to this there are three possible solutions:

### Override default busybox image used as init container on the etcd  operator (Proposed solution)

This is the simple solution and is compliant with the requisites. Since version [0.9.1](https://github.com/coreos/etcd-operator/releases/tag/v0.9.1) the etcd operator supports this option.

The default image uses busybox:1.28.0-glibc, busybox image is available on quay through libpod, unfortunately only latest version which is [not suitable](https://github.com/coreos/etcd-operator/blob/13a1831f9744207b39733bbe91413ca2e389aa40/pkg/util/k8sutil/k8sutil.go#L371). 

After inspect the  usage of the init container (and to maintain the smallest image possible) this init container functionality can be run on an alpine image. So the spec for the cluster can be define like this:

```bash
apiVersion: etcd.database.coreos.com/v1beta2
kind: EtcdCluster
metadata:
  name: example
  annotations:
    etcd.database.coreos.com/scope: clusterwide
  namespace: default
spec:
  size: 3
  version: 3.2.13
  pod:
    busyboxImage: quay.io/libpod/alpine:3.10.2
```

### Use an operator from Red Hat Operators

Red Hat Operators can be used as they do not require any license (like Red Hat Market operators) and their images does not depend on docker hub (like Community Operators). 

Datagrid (Infinispan) and amq-broker-lts where tested to try be as much similar as the original story. Both of them require a big amount of memory in comparison with etc instances (etcd ~ 30Mb <-> datagrid / amq ~ 300Mb), also they do not allow install mode type for all namespaces which requires to create an specific operatorgroup to test them.

Other possible option is o use any other operator which does not manage a cluster like openshift pipelines, in this case the story will change totally to test the expected functionality from this operator.

praveenkumar:
@adrianriobo Thank you for coming up with test plan, lets go with first option where we override the busybox image for init container, since this is easily doable and solve our issue.

PR #1784, opened 2020-12-16
Update story_marketplace with etcdcluster init container image from quay.io 

adrianriobo:
Remove docherhub dependecy on story marketplace 

**Fixes:** Issue #1753

## Solution/Idea

etcd operator uses as init container busybox:1.28.0-glibc, the idea is to use an equivalent image hosted at quay.io registry.

## Proposed changes

EtcdCluster allows to specify the image used for init container, the default image is busybox, as so the name on the spec is referring to it but any other small image which can be execute the init functionality  is suitable. Alpine from libpod was tested sucessfully.

Define the cluster with alpine as init container:

```bash
apiVersion: etcd.database.coreos.com/v1beta2
kind: EtcdCluster
metadata:
  name: example
  annotations:
    etcd.database.coreos.com/scope: clusterwide
  namespace: default
spec:
  size: 3
  version: 3.2.13
  pod:
    busyboxImage: quay.io/libpod/alpine:3.10.2
```

## Testing

Run e2e story_marketplace feature

### DIFF
--- a/test/testdata/etcd-cluster3.yaml
+++ b/test/testdata/etcd-cluster3.yaml
@@ -8,3 +8,5 @@ metadata:
 spec:
   size: 3
   version: 3.2.13
+  pod:
+    busyboxImage: quay.io/libpod/alpine:3.10.2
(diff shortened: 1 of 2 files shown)


praveenkumar:
ok to test

PR #1787, opened 2020-12-17
Remove start flag for monitoring, keep only config prop. Check monitoring in crc status

guillaumerose:
Follow up of the monitoring PR.

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -11,21 +11,21 @@ import (
 )
 
 const (
-	Bundle               = "bundle"
-	CPUs                 = "cpus"
-	Memory               = "memory"
-	DiskSize             = "disk-size"
-	NameServer           = "nameserver"
-	PullSecretFile       = "pull-secret-file"
-	DisableUpdateCheck   = "disable-update-check"
-	ExperimentalFeatures = "enable-experimental-features"
-	NetworkMode          = "network-mode"
-	HTTPProxy            = "http-proxy"
-	HTTPSProxy           = "https-proxy"
-	NoProxy              = "no-proxy"
-	ProxyCAFile          = "proxy-ca-file"
-	EnableMonitoring     = "enable-monitoring"
-	ConsentTelemetry     = "consent-telemetry"
+	Bundle                  = "bundle"
+	CPUs                    = "cpus"
+	Memory                  = "memory"
+	DiskSize                = "disk-size"
+	NameServer              = "nameserver"
+	PullSecretFile          = "pull-secret-file"
+	DisableUpdateCheck      = "disable-update-check"
+	ExperimentalFeatures    = "enable-experimental-features"
+	NetworkMode             = "network-mode"
+	HTTPProxy               = "http-proxy"
+	HTTPSProxy              = "https-proxy"
+	NoProxy                 = "no-proxy"
+	ProxyCAFile             = "proxy-ca-file"
+	ConsentTelemetry        = "consent-telemetry"
+	EnableClusterMonitoring = "enable-cluster-monitoring"
 )
 
 func RegisterSettings(cfg *config.Config) {
@@ -45,7 +45,7 @@ func RegisterSettings(cfg *config.Config) {
 	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied)
 	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied)
 
-	cfg.AddSetting(EnableMonitoring, false, config.ValidateBool, config.SuccessfullyApplied)
+	cfg.AddSetting(EnableClusterMonitoring, false, config.ValidateBool, config.SuccessfullyApplied)
 
 	// Telemeter Configuration
 	cfg.AddSetting(ConsentTelemetry, false, config.ValidateBool, config.SuccessfullyApplied)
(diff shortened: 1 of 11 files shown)


cmd/crc/cmd/config/config.go
```diff
@@ -11,20 +11,20 @@ import (
...
+	NameServer                = "nameserver"
+	PullSecretFile            = "pull-secret-file"
+	DisableUpdateCheck        = "disable-update-check"
+	ExperimentalFeatures      = "enable-experimental-features"
+	NetworkMode               = "network-mode"
+	HTTPProxy                 = "http-proxy"
+	HTTPSProxy                = "https-proxy"
+	NoProxy                   = "no-proxy"
+	ProxyCAFile               = "proxy-ca-file"
+	EnableOpenShiftMonitoring = "enable-openshift-monitoring"
```
cfergeau, line 27: Couldn't this get confusing on okd where we should avoid talking about openshift?

cmd/crc/cmd/root.go
```diff
@@ -146,5 +146,10 @@ func newMachine() machine.Client {
 }
 
 func newMachineWithConfig(config crcConfig.Storage) machine.Client {
-	return machine.NewClient(constants.DefaultName, isDebugLog(), network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString()))
+	return machine.NewClient(
+		constants.DefaultName,
+		isDebugLog(),
+		network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString()),
+		config.Get(cmdConfig.EnableOpenShiftMonitoring).AsBool(),
```
cfergeau, line 153: Random thought: 4 arguments is starting to be too much from a readability perspective, when a 5th one is considered, it would be nice to refactor this... For example by introducing an intermediate `ClientConfig` type, or by removing some args ,and adding some setters, `if isDebugLog() client.SetDebug(true)` and so on.

jsliacan (approved):
LGTM. I started CRC from this PR with 4.6.6 bundle and worked as expected: flag not present, config options respected. Memory requirement checked even if config contains enough but start flag `--memory` overrides it to too little. Thanks!

pkg/crc/cluster/clusteroperator.go
```diff
@@ -21,14 +21,18 @@ type Status struct {
...
+func GetClusterOperatorsStatus(ocConfig oc.Config, monitoringEnabled bool) (*Status, error) {
+	ignoredOperators := ignoreClusterOperators
+	if !monitoringEnabled {
+		ignoredOperators = append(ignoredOperators, "monitoring")
+	}
+	return getStatus(ocConfig, ignoredOperators, []string{})
 }
 
-func getStatus(ocConfig oc.Config, selector []string) (*Status, error) {
+func getStatus(ocConfig oc.Config, filter, selector []string) (*Status, error) {
```
cfergeau, line 35: maybe `ignoredOperators` rather than `filter` which is slightly more explicit?

pkg/crc/machine/machine.go
```diff
@@ -397,6 +397,13 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 		return nil, errors.Wrap(err, "Failed to update cluster ID")
 	}
 
+	if client.monitoringEnabled {
+		logging.Info("Enabling cluster monitoring operator...")
+		if err := cluster.StartMonitoring(ocConfig); err != nil {
+			return nil, errors.Wrap(err, "Cannot start monitoring stack")
+		}
+	}
+
```
cfergeau, line 406: feels weird to do that before ensuring that kubelet is running, and that the apiserver certs are good. I'd move this just before `	logging.Info("Starting OpenShift cluster ... [waiting 3m]")`

cfergeau (commented):
Looks good to me, some minor comments.

pkg/crc/machine/machine.go
guillaumerose: Yes this is what I though but this is just after an eventual `DeleteOpenShiftAPIServer`. I don't know what is the state of the cluster after this!

cmd/crc/cmd/root.go
jsliacan: That sounds like a citation from Clean Code! :)

cfergeau: Never read it, so definitely not, but could be some remnants from Code Complete which I read a long time ago :) They probably are similar books.

cmd/crc/cmd/config/config.go
guillaumerose: Good point! What about `enable-cluster-monitoring-operator` ? At least this is exactly what it does.

guillaumerose: Renamed to enable-cluster-monitoring.

pkg/crc/cluster/clusteroperator.go
guillaumerose: renamed!

pkg/crc/machine/start.go
```diff
@@ -343,6 +343,13 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 		return nil, errors.Wrap(err, "Failed to update cluster ID")
 	}
 
+	if client.monitoringEnabled {
+		logging.Info("Enabling cluster monitoring operator...")
+		if err := cluster.StartMonitoring(ocConfig); err != nil {
+			return nil, errors.Wrap(err, "Cannot start monitoring stack")
+		}
+	}
+
```
praveenkumar, line 352: 👍🏼

praveenkumar (approved):

Issue #1785, opened 2020-12-16
[Integration Test] Invalid syntax on story health for windows


adrianriobo:
There is syntax issue on windows for Scenario: Create and test app. In this scenario a rollout status is execute with:

```bash
oc rollout status dc/httpd-example || oc rollout status deployment httpd-example
```

This conditional composition of commands with || is valid syntax for bash based shells but not for powershell (5.1):

```
The token '||' is not a valid statement separator in this version
```

In these commands deploymentconfig and deployment rollouts are checked. But checking the httpd-example [template](https://github.com/sclorg/httpd-ex/blob/master/openshift/templates/httpd.json) 
only deploymentconfig is created

Before this [commit](https://github.com/code-ready/crc/commit/16d015a9e978c4a2b63b1b22b8e8677298afe52a) only deploymentconfig was checked. Then it was changed to deployment as from commits comments: newer oc versions change deploymentconfig into deployment when mapping from github. 

Then this [commit](https://github.com/code-ready/crc/commit/928a8a092e0ce2d02d5e9fa48e32ade62fb320b5) 
include both (with the current or syntax), as it seems oc creates deploymentconfig as expected. Only noticeable thing is, first commit seems based on a nightly build and all newer stable oc versions tested only creates deploymentconfig.

To solve this problem:

1. Just remove deployment rollout status part (as it should not exist any deployment). If that is the case is oc working properly?, as deployment is preferred instead of dc but their behaviour is distinct, so if the user expects a dc, does it make sense to have a deployment instead?. (Proposed solution)

2. Create an specific scenario for Windows with bot commands executed separately.

adrianriobo:
As @praveenkumar suggested other option, even better, is to move away from deploy a new app based on an external sample and deploy our own resources.

jsliacan:
That might have the downside of not being a likely user story (since the user won't deploy exactly the same thing as we would)?

adrianriobo:
Define a set of resources to deploy an app on ocp is a user story, no matter if the manifests are made by you or picked from the sample template, actually this could be as simple as grab the current template and add it as a test resource. This way we ensure the exact elements which will be created in the cluster ... in this case only the deploymentconfig

PR #1791, opened 2020-12-21
Update http example integration test

praveenkumar:
This patch will remove the check for deploymentconfig and deployment
which is causing issue for windows.

### DIFF
--- a/test/integration/features/story_health.feature
+++ b/test/integration/features/story_health.feature
@@ -45,14 +45,14 @@ Feature: End-to-end health check
 
     @darwin @linux @windows
     Scenario: Create and test app
-        When executing "oc new-app httpd-example" succeeds
-        Then stdout should contain "Creating resources"
-        And stdout should contain
-            """
-            service "httpd-example" created
-            """
-        When executing "oc rollout status dc/httpd-example || oc rollout status deployment httpd-example" succeeds
+        When executing "oc create deployment httpd-example --image=quay.io/bitnami/nginx --port=8080" succeeds
+        Then stdout should contain "deployment.apps/httpd-example created"
+        When executing "oc rollout status deployment httpd-example" succeeds
         Then stdout should contain "successfully rolled out"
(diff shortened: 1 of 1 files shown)


gbraad:
@jsliacan @adrianriobo WDYT?

adrianriobo:
This PR fix related problem described at https://github.com/code-ready/crc/issues/1785. This problem was produced due to inconsistent resources expectations, with this PR this is solved as we ensure only deployment is created and it covers the same functionality.. 

The only comment is what @guillaumerose suggested using an image based on [nginx](https://quay.io/repository/bitnami/nginx), it seems it would not change the expected http response code 403 but the image size is 1/3. 
```
IMAGE                                                   TAG                 IMAGE ID            SIZE
quay.io/bitnami/nginx                                   1.18.0              974152eb6fa49       93.3MB
registry.redhat.io/rhscl/httpd-24-rhel7                 latest              f6c37c3a78961       328MB
```

gbraad:
> The only comment is what @guillaumerose suggested using an image based on [nginx](https://quay.io/repository/bitnami/nginx), it seems it would not change the expected http response code 403 but the image size is 1/3.

In short; approved?

guillaumerose:
`quay.io/bitnami/nginx` has a default webpage, so it will be 200.

adrianriobo:
If so, I would prefer go with quay.io/bitnami/nginx

jsliacan (commented):
Others already commented on the `403` status code. Apart from that, we're using `1m` and `60s` for the same thing. This had been there before this PR, but let's align it at this opportunity - thanks!

gbraad:
do we have something that exists in an OpenShift namespace?

gbraad:
> 1m and 60s

I would prefer to see `1m`... and everything above should be possible in `1m 30` instead of `90s`... but enforcing this?
This is merely preference and more about maintainability... and should be for the whole test

PR #1793, opened 2020-12-22
Add a controller in the cluster keeping in sync routes and hosts file

guillaumerose:
* The daemon has an API available in the VM: curl http://gateway/hosts/add or remove.
* A controller is running the cluster and calls the daemon API.

For Windows, in the future, the daemon will forward request to the adminhelper daemon. Currently, a UAC prompt is shown.


Makefile
```diff
@@ -88,12 +88,19 @@ $(BUILD_DIR)/macos-amd64/crc: $(SOURCES)
 $(BUILD_DIR)/linux-amd64/crc: $(SOURCES)
 	GOOS=linux GOARCH=amd64 go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/linux-amd64/crc ./cmd/crc
 
+$(BUILD_DIR)/linux-amd64/routes-controller: $(SOURCES)
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/linux-amd64/routes-controller ./cmd/routes-controller
```
gbraad, line 92: This is a separate binary? What is it's function?
Please describe this more clearly in the description of the PR

The problem I see is that this becomes another binary we need to maintain the lifecycle of. How will this work? Who maintains this? Why can't the/a (slightly more generic) `daemon` handle this?

gbraad:
This 'routes-controller' is a separate binary? What is it's function?
Please describe this more clearly in the description of the PR

The problem I see is that this becomes another binary we need to maintain the lifecycle of. How will this work? Who maintains this? Why can't the/a (slightly more generic) daemon handle this?

Makefile
```diff
@@ -88,12 +88,19 @@ $(BUILD_DIR)/macos-amd64/crc: $(SOURCES)
...
+
 $(BUILD_DIR)/windows-amd64/crc.exe: $(SOURCES)
 	GOARCH=amd64 GOOS=windows go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/windows-amd64/crc.exe ./cmd/crc
 
 $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
 	go build --tags="build" -ldflags="$(LDFLAGS)" -o $(HOST_BUILD_DIR)/crc-embedder ./cmd/crc-embedder
 
+.PHONY: routes-controller
+routes-controller: $(BUILD_DIR)/linux-amd64/routes-controller
+	podman build -t quay.io/gurose/routes-controller:v1 -f cmd/routes-controller/Dockerfile $(BUILD_DIR)/linux-amd64/
```
gbraad, line 102: `quay.io/crcont/` ?

gbraad: `cmd/routes-controller/Dockerfile` containers need to live in their own repos as part of the security review.

Makefile
gbraad: `./cmd/routes-controller` this source code likely has to become a repo on it's own as this is part of a container.

cfergeau:
Could the controller be running on the host since we have the crc daemon which is running in the background anyway?

cmd/crc/cmd/daemon.go
```diff
@@ -159,6 +159,10 @@ func adminHelperMux() *http.ServeMux {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
+		if runtime.GOOS == "windows" { // Disable it on Windows, it avois tons of UAC prompts.
```
cfergeau, line 162: 'avoids' (twice)

guillaumerose:
I have 4 choices in fact:
* Use this controller + API
* Make crc start run in the daemon so that I can know when the VM is started and kubeconfig is here. Then I can launch the controller
* Add events to the network stack so that, when a new VM connects, I try to run the controller
* Poll the kubeconfig file. When it appears, run the controller.

I guess the good solution is to make crc start the client of the daemon.

cfergeau:
> I have 4 choices in fact:
> 
>     * Use this controller + API
> 
>     * Make crc start run in the daemon so that I can know when the VM is started and kubeconfig is here. Then I can launch the controller
> 
>     * Add events to the network stack so that, when a new VM connects, I try to run the controller
> 
>     * Poll the kubeconfig file. When it appears, run the controller.
> 
> 
> I guess the good solution is to make crc start the client of the daemon.

What I had in mind was to rework `machine.Start()` a bit:
```
        // Post-VM start
        vmState, err := host.Driver.GetState()
        if err != nil {                                                                                                                   
                return nil, errors.Wrap(err, "Error getting the state")                                                                  
        }
        if vmState != state.Running {
                return nil, errors.Wrap(err, "CodeReady Containers VM is not running")
        }
                
        instanceIP, err := getIP(host, client.useVSock())
        if err != nil {
                return nil, errors.Wrap(err, "Error getting the IP")
        }
        sshRunner, err := crcssh.CreateRunner(instanceIP, getSSHPort(client.useVSock()), crcBundleMetadata.GetSSHKeyPath(), constants.GetPrivateKeyPath(), constants.GetRsaPrivateKeyPath())
        if err != nil {
                return nil, errors.Wrap(err, "Error creating the ssh client")
        }
        defer sshRunner.Close()
        
        logging.Debug("Waiting until ssh is available")
        if err := cluster.WaitForSSH(sshRunner); err != nil {
                return nil, errors.Wrap(err, "Failed to connect to the CRC VM with SSH -- host might be unreachable")
        }
```
The `kubeconfig` file could be copied to `~/.crc/machines/crc` around that time, rather than being done later as being done now. And the host DNS configuration (`addOpenShiftHosts` / `runPostStartForOS`) could be also be done after checking the VM is up, but before waiting for SSH. In the vsock case, `addOpenShiftHosts` would talk to the daemon, starting it if necessary.

guillaumerose:
After a long discussion, it seems that embedding this controller in the daemon is really complicated as machine.Start can be called from the crc CLI and from the daemon itself. 
I propose we start with this controller in a pod and then, when crc CLI will always be a client of the daemon, move the controller to the daemon.

guillaumerose:
I extracted the controller code to https://github.com/code-ready/routes-controller and modified everything to get it working on Windows. 

UAC prompts only appear two times at `crc start` time and then each time a route is created/updated/deleted.

cc @gbraad

gbraad:
So a UAC prompt is given twice on start, and each time on a route? That sounds like a lot (too much).

guillaumerose:
> twice on start, and each time on a route

Yes twice (clean + add the default ones) and then, when you add/update/delete a route, you get a prompt. 
Compared to now, this is just one eventual prompt if you run `oc new-app`. 

It will be 0 when the adminhelper will be a daemon. 

For now:
* I can disable this for Windows and let the user do it by hand,
* or keep as it is.

gbraad:
> It will be 0 when the adminhelper will be a daemon.

That is what I am looking for. Can you outline the needed steps for this in a new issue or on the admin-helper issue?
Some estimation or timeline is appreciated. Especially to conclude if this should be merged or wait for to coincide with another possible follow-up.

guillaumerose:
This is already good to take. It makes the cluster work fully. The debate for the daemon will be more intense for sure. Let's start with this.

pkg/crc/machine/start.go
```diff
@@ -546,3 +555,37 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "routes-controller",
+			Namespace: "openshift-ingress",
+		},
+		Spec: v1.PodSpec{
+			ServiceAccountName: "router",
+			Containers: []v1.Container{
+				{
+					Name:            "routes-controller",
+					Image:           "quay.io/crcont/routes-controller:6effc2f0304d3a11c1e22c278b10630e834d0220",
```
cfergeau, line 574: one more container image we need to remember to maintain :)

pkg/crc/adminhelper/hosts.go
```diff
@@ -22,13 +24,28 @@ func UpdateHostsFile(instanceIP string, hostnames ...string) error {
 }
 
 func AddToHostsFile(instanceIP string, hostnames ...string) error {
+	if err := validateHostnames(hostnames); err != nil {
+		return err
+	}
```
cfergeau, line 29: This has been added to the admin helper code itself, do you want it here as well?

pkg/crc/adminhelper/hosts_windows.go
```diff
@@ -6,7 +6,12 @@ import (
 	"github.com/code-ready/crc/pkg/os/windows/powershell"
 )
 
-func execute(args ...string) error {
+func executePrivileged(args ...string) error {
 	_, _, err := powershell.ExecuteAsAdmin("modifying hosts file", strings.Join(append([]string{goodhostPath}, args...), " "))
 	return err
 }
```
cfergeau, line 12: Probably better to land first the admin-helper daemon work?

pkg/crc/services/dns/dns.go
```diff
@@ -124,5 +124,6 @@ func addOpenShiftHosts(serviceConfig services.ServicePostStartConfig) error {
 	return adminhelper.UpdateHostsFile(serviceConfig.IP, serviceConfig.BundleMetadata.GetAPIHostname(),
 		serviceConfig.BundleMetadata.GetAppHostname("oauth-openshift"),
 		serviceConfig.BundleMetadata.GetAppHostname("console-openshift-console"),
+		serviceConfig.BundleMetadata.GetAppHostname("downloads-openshift-console"),
 		serviceConfig.BundleMetadata.GetAppHostname("default-route-openshift-image-registry"))
```
cfergeau, line 128: This seems unrelated to `Also clean hosts file on Windows`

cfergeau (commented):
Looks good, though probably best to land 'admin-helper as a daemon' first.

cfergeau:
I rebased this in https://github.com/cfergeau/crc/tree/routescontroller but haven't tested this yet.

pkg/crc/adminhelper/hosts.go
guillaumerose: yes I dropped the commit

cmd/crc/cmd/daemon.go
```diff
@@ -109,6 +111,17 @@ func run(configuration *types.Configuration, endpoints []string) error {
 	log.Info("waiting for clients...")
 	errCh := make(chan error)
 
+	ln, err := vn.Listen("tcp", fmt.Sprintf("%s:80", configuration.GatewayIP))
```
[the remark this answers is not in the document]
guillaumerose: Yes totally virtual. It will only be reachable in the VM by `curl http://gateway.crc.testing/...`.
I added a comment near `internalAPIMux()`

pkg/crc/services/dns/dns.go
guillaumerose: Yes extracted to a new commit.

pkg/crc/adminhelper/hosts_windows.go
guillaumerose: admin-helper daemon still require some work: a preflight to install it, define the protocol, etc.

With this PR, we solve it for macOS and Linux. For Windows, the experience is not too bad. There is just one UAC prompt per new route.

This network stack could be GA on macOS/Linux after this PR (+ some minor enhancements, better errors, etc.). 
I think we have the feature parity with the traditional network stack.

cmd/crc/cmd/daemon.go
```diff
@@ -133,6 +146,44 @@ func run(configuration *types.Configuration, endpoints []string) error {
 	return <-errCh
 }
 
+// This API is only exposed in the virtual network. Any process can reach it by connecting to gateway.crc.testing.
+func internalAPIMux() *http.ServeMux {
```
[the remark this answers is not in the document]
guillaumerose: I renamed to gatewayAPIMux to indicate it's only gateway.crc.testing.

cmd/crc/cmd/daemon.go
```diff
@@ -133,6 +146,44 @@ func run(configuration *types.Configuration, endpoints []string) error {
...
+		if r.Method != http.MethodPost {
+			http.Error(w, "post only", http.StatusBadRequest)
+			return
+		}
+		var req []string
+		if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		if err := adminhelper.RemoveFromHostsFile(req...); err != nil {
```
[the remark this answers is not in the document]
guillaumerose: I don't get this. What do you mean?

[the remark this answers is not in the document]
cfergeau: The code handling `/hosts/add` and `/hosts/remove` is the same except for the line calling `adminhelper.AddToHostsFile` / `adminhelper.RemoveFromHostsFile`, dunno if it's possible to factor this somehow without making the code too complicated?

pkg/crc/adminhelper/hosts_windows.go
cfergeau: > This network stack could be GA on macOS/Linux after this PR (+ some minor enhancements, better errors, etc.).
>  I think we have the feature parity with the traditional network stack.

daemon autostart is missing on linux (and on macos setups without the tray running). And if I'm not mistaken this also does not handle routes using non-standard ports? I guess autostart falls under the 'minor enhancements category')? I don't mind the non-standard thing for now, I agree we can push for wider testing after this.

pkg/crc/adminhelper/hosts.go
```diff
@@ -22,13 +23,26 @@ func UpdateHostsFile(instanceIP string, hostnames ...string) error {
 }
 
 func AddToHostsFile(instanceIP string, hostnames ...string) error {
-	return execute(append([]string{"add", instanceIP}, hostnames...)...)
+	var filtered []string
+	for _, hostname := range hostnames {
+		stdout, err := execute("contains", instanceIP, hostname)
```
cfergeau, line 28: Have you considered using `github.com/code-ready/admin-helper/pkg/hosts` directly? `contains` implementation is:
```
func contains(args []string) error {
        hosts, err := hosts.New()
        if err != nil {
                return err
        }
        if hosts.Contains(args[0], args[1]) {
                fmt.Println("true")
        } else {
                fmt.Println("false")
        }
        return nil
}
```
which is the same amount of code, and would be more efficient.

pkg/crc/adminhelper/hosts.go
```diff
@@ -22,13 +23,26 @@ func UpdateHostsFile(instanceIP string, hostnames ...string) error {
...
 }
 
 func RemoveFromHostsFile(hostnames ...string) error {
-	return execute(append([]string{"rm"}, hostnames...)...)
+	return executePrivileged(append([]string{"rm"}, hostnames...)...)
 }
 
 func CleanHostsFile() error {
-	return execute([]string{"clean", constants.ClusterDomain, constants.AppsDomain}...)
+	return executePrivileged([]string{"clean", constants.ClusterDomain, constants.AppsDomain}...)
```
cfergeau, line 47: `return executePrivileged("clean", constants.ClusterDomain, constants.AppsDomain)` ?

cmd/crc/cmd/daemon.go
```diff
@@ -191,6 +204,45 @@ func run(configuration *types.Configuration) error {
 	}
 }
 
+// This API is only exposed in the virtual network (only the VM can reach this).
+// Any process inside the VM can reach it by connecting to gateway.crc.testing:80.
```
cfergeau, line 208: Hmm, any process in the VM can fill the host root filesystem, and maybe slow down name resolution? (dunno if huge /etc/hosts has a noticeable impact on name resolution). I did not check the admin-helper code, but we can limit the number of aliases we add to a few hundreds, and also limit their individual size (which is done already I think), which would address this concern.

cfergeau (commented):
Still a few minor comments, but overall looks good to me.

pkg/crc/adminhelper/hosts_windows.go
guillaumerose: `does not handle routes using non-standard ports`

I tried to find examples or code around that but I couldn't find it. The only route I see like this would be the one for Kubernetes API.
Router container only exposes port 80 and 443, not others. I don't really understand how an other port can be used.

cmd/crc/cmd/daemon.go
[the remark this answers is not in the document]
guillaumerose: Yes good point, done

pkg/crc/adminhelper/hosts.go
guillaumerose: Yes done. I wanted to avoid to vendor admin-helper in crc. We just need to remind to bump it in 2 places.

pkg/crc/adminhelper/hosts.go
guillaumerose: done

cmd/crc/cmd/daemon.go
guillaumerose: https://github.com/code-ready/admin-helper/issues/14 added

pkg/crc/machine/start.go
```diff
@@ -636,3 +645,37 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
...
+					ImagePullPolicy: v1.PullIfNotPresent,
+				},
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	if err := sshRunner.CopyData(bin, "/tmp/routes-controller.json", 0444); err != nil {
```
praveenkumar, line 673: @guillaumerose this is only pod definition so if somehow this pod crash then this route stuff will break, should we have deployment as `Kind` instead pod so that pod recreation happen if pod crashes?

guillaumerose: Deployment are useful version management. For crash, there is the RestartPolicy on the pod directly. It is "Always" by default so we should be good. 

I tried to kill the process by hand. It is restarted. 
I tried to start/stop/start the VM. The process is here.

praveenkumar: @guillaumerose For me it is not happening.
```
13:07 $ oc get pods -n openshift-ingress
NAME                              READY   STATUS    RESTARTS   AGE
router-default-7476bbd498-g9r7k   1/1     Running   1          13d
routes-controller                 1/1     Running   0          30m

13:08 $ oc delete pod routes-controller -n openshift-ingress
pod "routes-controller" deleted

13:11 $ oc get pods -n openshift-ingress
NAME                              READY   STATUS    RESTARTS   AGE
router-default-7476bbd498-g9r7k   1/1     Running   1          13d
```

guillaumerose: Yes sure, I mean kill with `sudo kill -9 pid`. 
If the user does this, why bother him and recreate the pod ?

praveenkumar: Since it is part of `openshift-ingress` and if  a user just delete all the pods under this namespace, all pods except this one come up without any issue, Also can we make it as static pod if we really want it to be running always?

guillaumerose: I tried to add in static pod directory but it doesn't work:

```
Apr 19 12:19:17 crc-xl2km-master-0 hyperkube[2154]: E0419 12:19:17.473167    2154 kubelet.go:1671] Failed creating a mirror pod for "routes-controller-crc-xl2km-master-0_openshift-ingress(fd89789fbe33ba7075234cb104ec8e39)": pods "routes-controller-crc-xl2km-master-0" is forbidden: a mirror pod may not reference service accounts
```

>  if a user just delete all the pods under this namespace

It should not happen or he really wants to do this?

guillaumerose:
retest this please

pkg/crc/machine/start.go
praveenkumar: This is not blocker for this PR so let's have some follow up issue in case we hit it.

pkg/crc/adminhelper/hosts_windows.go
cfergeau: I thought our own doc had such examples ( https://code-ready.github.io/crc/#deploying-sample-application-with-odo_gsg ) but after being deployed, the app is exposed on port 80/443, not on 8080 as I expected. 
https://docs.okd.io/latest/networking/configuring_ingress_cluster_traffic/configuring-ingress-cluster-traffic-load-balancer.html may be related to this, but is not working with the default networking anyway.

PR #1794, opened 2020-12-22
Intercept errors and send to segment for all commands

guillaumerose:
* Remove most of os.Exit in the middle of the code. This is needed to send errors to segment.
* Use RunE instead of Run. postRun() is now correctly called after each method. Even in case of an error.
* Introduce a segment middleware for all commands (except config because it lives in a separate pkg - to be changed)
* Log the error and the command used in segment.
* Refactor segment tests.


gbraad:
haven't looked at the PR yet, but from the discussions with management we shouldn't send stacktraces (or logs). so we need to be smarter about the error messages first and prevent being ambiguous before we can send an error message,

cmd/crc/cmd/console.go
```diff
@@ -31,11 +30,9 @@ var consoleCmd = &cobra.Command{
 	Aliases: []string{"dashboard"},
 	Short:   "Open the OpenShift Web Console in the default browser",
 	Long:    `Open the OpenShift Web Console in the default browser or print its URL or credentials`,
-	Run: func(cmd *cobra.Command, args []string) {
-		if renderErr := runConsole(os.Stdout, newMachine(), consolePrintURL, consolePrintCredentials, outputFormat); renderErr != nil {
-			exit.WithMessage(1, renderErr.Error())
-		}
-	},
+	RunE: segmentMiddleware(func(cmd *cobra.Command, args []string) error {
```
gbraad, line 33: this can be generalized as an intercept call to allow ANY kind of logging.

So perhaps it is better to use:
```
executeWithLogging(func(cmd *cobra.Command, args []string) error {
```

and make `segmentMiddleware` just an implementation detail of the execution wrapper. Currenrly it feels like this became a dependency, while it is just one of the technologies we use.

cmd/crc/cmd/root.go
```diff
@@ -154,3 +160,15 @@ func newMachine() machine.Client {
 func newMachineWithConfig(config crcConfig.Storage) machine.Client {
 	return machine.NewClient(constants.DefaultName, isDebugLog(), network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString()))
 }
+
+func segmentMiddleware(input func(cmd *cobra.Command, args []string) error) func(cmd *cobra.Command, args []string) error {
```
gbraad, line 164: As mentioned previosuly. This wrapper can be generalized `executeWithLogging` or whatever and then use 

```
if serr := segmentClient.Upload(cmd.Use, err); serr != nil {
```

also clearly behind a `consent` flag here. This is not clear from first glance.

gbraad (changes requested):
wrapper should be generalized as usage of 'segment' is only an implementation detail

cmd/crc/cmd/root.go
```diff
@@ -82,8 +85,10 @@ func runRoot() {
 
 func Execute() {
 	if err := rootCmd.Execute(); err != nil {
-		logging.Fatal(err)
+		runPostrun()
+		exit.WithMessage(1, err.Error())
```
gbraad, line 89: only here we do `exit.WithMessage` ?
it seems like all other instances have been removed... so why keep this?

praveenkumar: This is the main execute which all commands uses since all command add to the root command so this will going to execute whenever corresponding command have error.

pkg/crc/segment/segment.go
```diff
@@ -52,7 +52,7 @@ func (c *Client) Close() error {
 	return c.segmentClient.Close()
 }
 
-func (c *Client) Upload(err error) error {
+func (c *Client) Upload(action string, err error) error {
```
praveenkumar, line 55: @guillaumerose This should be part of a different commit, here we are changing this method signature.So in this commit the `segmentMiddleware` just add the error not the action and then in the next commit you can add the action.

cmd/crc/cmd/root.go
gbraad: I am more question the use of 'exit with message' here, but also perhaps possibility this hides the usage.

See my suggestion to use a generalized wrapper and perhaps don't do this. This is almost like a generic exception handler which can lead to a 'hidden' (hard to find) invocation.

praveenkumar: Ah got it, right now we are only using this with podman-env side also but I understand your point. I think it is OK to have this as generic handler as of now since there is 2 places where it's happening but in the future might be just have it either part of that generalized wrapper or have following part of `Execute` function.

```
_, _ = output.Fout(os.Stderr, fmt.Sprintf(text, args...))
os.Exit(1)
```

cmd/crc/cmd/console.go
guillaumerose: done

cmd/crc/cmd/root.go
guillaumerose: segmentClient is built with a config object. `segmentClient, err = segment.NewClient(config)`

cmd/crc/cmd/root.go
guillaumerose: I removed `exit.WithMessage`, it's clearer now. 

I also removed podman-env command. It is broken for more than 6 months now.

pkg/crc/segment/segment.go
guillaumerose: done

cmd/crc/cmd/root.go
gbraad: > I also removed podman-env command. It is broken for more than 6 months now.

oops ... waiting for release of V2

gbraad:
I see the change from the wrapper to an attachMiddleware function. somehow the clarity is lost because of this; has become very indirect like a generic exceptionhandler. this does not mean it is wrong, just 'afraid' it is harder to find ... perhaps good to understand what other use `attachMiddleware` would have. as it is now a single use function.

guillaumerose:
rebased!

guillaumerose:
All comments seem to be resolved. @praveenkumar @gbraad can you PTAL? Thanks.

PR #1797, opened 2020-12-23
Upping client,server timeouts to prevent EOFs

dofinn:
While tailing logs of pods via accessing CRC remotely, the default
HAProxy config cuts connections by sending:

```
error: unexpected EOFs
```

Raising these timeouts eleviates the problem.

### DIFF
--- a/docs/source/topics/proc_setting-up-remote-server.adoc
+++ b/docs/source/topics/proc_setting-up-remote-server.adoc
@@ -69,8 +69,8 @@ defaults
     log global
     mode http
     timeout connect 5000
-    timeout client 5000
-    timeout server 5000
+    timeout client 500000
+    timeout server 500000
 
 frontend apps
     bind 0.0.0.0:80


docs/source/topics/proc_setting-up-remote-server.adoc
```diff
@@ -69,8 +69,8 @@ defaults
     log global
     mode http
     timeout connect 5000
-    timeout client 5000
-    timeout server 5000
+    timeout client 500000
+    timeout server 500000
```
praveenkumar, line 73: Shouldn't we just document it instead making it default `5000 => 500000` ?

gbraad: this is about 8 and a half minutes to get a timeout.
Wouldn't in this case the client not just hang for a long time until it gets an answer... not sure what would be the better user experience.

gbraad (commented):
Wouldn't in this case the client not just hang for a long time until it gets an answer... not sure what would be the better user experience. WDYT?

docs/source/topics/proc_setting-up-remote-server.adoc
dofinn: The documentation could be updated to represent this, but as a user I would expect the documentation to provide an experience where I can perform "normal" operations like tailing logs without receiving an EOF.

dofinn:
> Wouldn't in this case the client not just hang for a long time until it gets an answer... not sure what would be the better user experience. WDYT?

I guess it depends on the user, if a CRC user is testing network functionality, then this setting may sway results. Regardless, HAProxy is only typical to a remote setup and anyone using it should be aware of its potential impacts.

PR #1801, opened 2020-12-24
Use udev rule to configure the permission and group for /dev/vsock

anjannath:
This would persist the permissions and group change after reboot

### DIFF
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"syscall"
 
+	crcErrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/network"
 	crcos "github.com/code-ready/crc/pkg/os"
@@ -103,10 +104,13 @@ var vsockPreflightChecks = Check{
 	fixDescription:     "Checking if vsock is correctly configured",
 	fix:                fixVsock,
 	cleanupDescription: "Removing vsock configuration",
-	cleanup:            removeVsockUdevRule,
+	cleanup:            removeVsockCrcSettings,
 }
 
-const vsockUdevRulesPath = "/usr/lib/udev/rules.d/99-crc-vsock.rules"
+const (
+	vsockUdevRulesPath          = "/usr/lib/udev/rules.d/99-crc-vsock.rules"
+	vsockModuleAutoLoadConfPath = "/etc/modules-load.d/vhost_vsock.conf"
+)
 
 func checkVsock() error {
 	executable, err := os.Executable()
@@ -156,6 +160,10 @@ func fixVsock() error {
 	if err != nil {
 		return err
 	}
+	err = crcos.WriteToFileAsRoot(fmt.Sprintf("Create file %s", vsockModuleAutoLoadConfPath), "vhost_vsock", vsockModuleAutoLoadConfPath, 0644)
+	if err != nil {
+		return err
+	}
 	_, _, err = crcos.RunWithPrivilege("modprobe vhost_vsock", "modprobe", "vhost_vsock")
 	if err != nil {
 		return err
@@ -163,9 +171,20 @@ func fixVsock() error {
 	return nil
 }
 
-func removeVsockUdevRule() error {
+func removeVsockCrcSettings() error {
+	var mErr crcErrors.MultiError
 	_, _, err := crcos.RunWithPrivilege(fmt.Sprintf("rm %s", vsockUdevRulesPath), "rm", "-f", vsockUdevRulesPath)
-	return err
+	if err != nil {
+		mErr.Collect(err)
+	}
+	_, _, err = crcos.RunWithPrivilege(fmt.Sprintf("rm %s", vsockModuleAutoLoadConfPath), "rm", "-f", vsockModuleAutoLoadConfPath)
+	if err != nil {
+		mErr.Collect(err)
+	}
+	if len(mErr.Errors) == 0 {
+		return nil
+	}
+	return mErr
 }
 
 func getAllPreflightChecks() []Check {
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -104,6 +104,8 @@ var vsockPreflightChecks = Check{
 	fix:              fixVsock,
 }
 
+const vsockUdevRulesPath = "/usr/lib/udev/rules.d/99-crc-vsock.rules"
+
 func checkVsock() error {
 	executable, err := os.Executable()
 	if err != nil {
@@ -126,7 +128,7 @@ func checkVsock() error {
 			return err
 		}
 		if group.Name != "libvirt" {
-			return errors.New("/dev/vsock is not is the right group")
+			return errors.New("/dev/vsock is not in the right group")
 		}
 	} else {
 		return errors.New("cannot cast info")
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_linux.go
```diff
@@ -146,21 +151,34 @@ func fixVsock() error {
...
+	udevRule := `KERNEL=="vsock", MODE="0060", OWNER="root", GROUP="libvirt"`
+
+	path, err := exec.LookPath("sudo")
 	if err != nil {
-		return err
+		return fmt.Errorf("Failed to find sudo on path: %w", err)
 	}
-	_, _, err = crcos.RunWithPrivilege("chown /dev/vsock", "chown", "root:libvirt", "/dev/vsock")
+	cmd := exec.Command(path, "tee", vsockUdevRulesPath)
+	cmd.Stdin = strings.NewReader(udevRule)
```
anjannath, line 162: not sure if a new method in `pkg/os/exec` for this would be better, something like `PrepareCmdWithPrivilege` that returns a `*exec.Cmd`?

cfergeau: I think you can use `WriteToFileAsRoot` here?

pkg/crc/preflight/preflight_linux.go
```diff
@@ -146,21 +151,34 @@ func fixVsock() error {
...
+	logging.Infof("Will use root access: creating %s", vsockUdevRulesPath)
+	out, err := cmd.CombinedOutput()
 	if err != nil {
+		logging.Debug("Failed to create vsock udev rule: ", out)
 		return err
 	}
-	_, _, err = crcos.RunWithPrivilege("chmod /dev/vsock", "chmod", "g+rw", "/dev/vsock")
+
+	_, _, err = crcos.RunWithPrivilege("modprobe vhost_vsock", "modprobe", "vhost_vsock")
```
cfergeau, line 170: Don't we need an additional rule (or expand the current one) to make sure the module is loaded? If you reboot the host, I don't think `vhost_vsock` will be loaded by default? 
Rule would probably look like this: (I did not test it, nor checked the pci id is correct)
```
ACTION=="add", SUBSYSTEM=="pci", ATTR{vendor}=="0x1af4", ATTR{device}=="0x1012", RUN{builtin}+="kmod load vhost_vsock"
```

anjannath: i did not check for the module being loaded on reboot, only verified the permissions were correct after a reboot, will test it out and update

anjannath: > Don't we need an additional rule (or expand the current one) to make sure the module is loaded? If you reboot the host, I don't think `vhost_vsock` will be loaded by default?
> Rule would probably look like this: (I did not test it, nor checked the pci id is correct)
> 
> ```
> ACTION=="add", SUBSYSTEM=="pci", ATTR{vendor}=="0x1af4", ATTR{device}=="0x1012", RUN{builtin}+="kmod load vhost_vsock"
> ```

Is the rule supposed to match a parent device of vsock? for /dev/vsock it doesn't have the vendor and device keys..
```
❯ udevadm info -a -n /dev/vsock                    

Udevadm info starts with the device specified by the devpath and then
walks up the chain of parent devices. It prints for every device
found, all possible attributes in the udev rules key format.
A rule to match, can be composed by the attributes of the device
and the attributes from one single parent device.

  looking at device '/devices/virtual/misc/vsock':
    KERNEL=="vsock"
    SUBSYSTEM=="misc"
    DRIVER==""
    ATTR{power/control}=="auto"
    ATTR{power/runtime_active_time}=="0"
    ATTR{power/runtime_status}=="unsupported"
    ATTR{power/runtime_suspended_time}=="0"
```

Alternatively we could create config file `/etc/modules-load.d/vhost_vsock.conf` for automatic loading of `hvsock_vsock` module?

pkg/crc/preflight/preflight_linux.go
```diff
@@ -146,21 +151,34 @@ func fixVsock() error {
 	if err != nil {
 		return err
 	}
-	_, _, err = crcos.RunWithPrivilege("modprobe vhost_vsock", "modprobe", "vhost_vsock")
+
+	udevRule := `KERNEL=="vsock", MODE="0060", OWNER="root", GROUP="libvirt"`
```
cfergeau, line 155: why '==' for kernel and '=' for the others?

pkg/crc/preflight/preflight_linux.go
cfergeau: Yeah, actually this `SUBSYSTEM=="pci", ATTR{vendor}=="0x1af4", ATTR{device}=="0x1012"` would work inside the VM where the virtio device is a PCI device. On the host, I don't know, and it's true there is no clear trigger we can hook up to in order to load the module, so yes, probably best to just drop a file in `/etc/modules-load.d`

pkg/crc/preflight/preflight_linux.go
anjannath: the kernel is a match key and  is for matching with the device name provided by the kernel, that's the criteria to apply the rule so its a comparision "==" while the others we are assigning the values to the keys so "="

pkg/crc/preflight/preflight_linux.go
```diff
@@ -97,13 +97,17 @@ var libvirtNetworkPreflightChecks = [...]Check{
...
-	checkDescription: "Checking if vsock is correctly configured",
-	check:            checkVsock,
-	fixDescription:   "Checking if vsock is correctly configured",
-	fix:              fixVsock,
+	configKeySuffix:    "check-vsock",
+	checkDescription:   "Checking if vsock is correctly configured",
+	check:              checkVsock,
+	fixDescription:     "Checking if vsock is correctly configured",
+	fix:                fixVsock,
+	cleanupDescription: "Removing vsock configuration",
```
praveenkumar, line 106: In this commit you are also adding the cleanup logic, shouldn't that part of different commit message?

anjannath: created new commit for the cleanup part, thanks!

pkg/crc/preflight/preflight_linux.go
```diff
@@ -164,8 +172,16 @@ func fixVsock() error {
...
 	_, _, err := crcos.RunWithPrivilege("rm /usr/lib/udev/rules.d/99-crc-vsock.rules", "rm", "-f", vsockUdevRulesPath)
-	return err
+	if err != nil {
+		mErr.Collect(err)
+	}
+	_, _, err = crcos.RunWithPrivilege("rm /etc/modules-load.d/vhost_vsock.conf", "rm", "-f", vsockModuleAutoLoadConfPath)
+	if err != nil {
+		mErr.Collect(err)
+	}
+	return mErr
```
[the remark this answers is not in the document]
anjannath: Yes, added check ` if len(mErr.Errors) == 0` to handle that case

pkg/crc/preflight/preflight_linux.go
```diff
@@ -146,15 +148,13 @@ func fixVsock() error {
 	if err != nil {
 		return err
 	}
-	_, _, err = crcos.RunWithPrivilege("modprobe vhost_vsock", "modprobe", "vhost_vsock")
-	if err != nil {
-		return err
-	}
-	_, _, err = crcos.RunWithPrivilege("chown /dev/vsock", "chown", "root:libvirt", "/dev/vsock")
+
+	udevRule := `KERNEL=="vsock", MODE="0060", OWNER="root", GROUP="libvirt"`
```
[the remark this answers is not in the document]
anjannath: yeah, should be `0660`, updated!

pkg/crc/preflight/preflight_linux.go
```diff
@@ -161,6 +163,11 @@ func fixVsock() error {
 	return nil
 }
 
+func removeVsockUdevRule() error {
+	_, _, err := crcos.RunWithPrivilege("rm /usr/lib/udev/rules.d/99-crc-vsock.rules", "rm", "-f", vsockUdevRulesPath)
```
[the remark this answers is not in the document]
anjannath: Done

pkg/crc/preflight/preflight_linux.go
```diff
@@ -156,6 +160,10 @@ func fixVsock() error {
 	if err != nil {
 		return err
 	}
+	err = crcos.WriteToFileAsRoot("Create file /etc/modules-load.d/vhost_vsock.conf", "vhost_vsock", vsockModuleAutoLoadConfPath, 0644)
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
anjannath: Thanks, was an oversight, changed now

pkg/crc/preflight/preflight_linux.go
```diff
@@ -164,8 +172,16 @@ func fixVsock() error {
 }
 
 func removeVsockUdevRule() error {
```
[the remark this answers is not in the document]
anjannath: renamed to `removeVsockCrcSettings` in the next commit

PR #1802, opened 2020-12-24
Import 90% of machine code in crc

guillaumerose:
Only drivers (rpc, driver interface, hyperkit and libvirt structs) and state are left in code-ready/machine.

It will be easier to refactor machine code: better tests in the future, etc. All the persistence is now done in crc side. code-ready/machine is stateless and also all drivers.

will enable https://github.com/code-ready/machine/pull/54


go.mod
```diff
@@ -99,3 +99,5 @@ replace (
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.19.0
 	k8s.io/sample-controller => k8s.io/sample-controller v0.19.0
 )
+
+replace github.com/code-ready/machine => github.com/guillaumerose/machine v0.0.0-20201224094349-747e5fa2ff83
```
gbraad, line 103: what still remains there?

pkg/drivers/fakedriver/fakedriver.go
```diff
@@ -0,0 +1,72 @@
+package fakedriver
```
gbraad, line 1: Used by `pkg/libmachine/host/host_test.go`

pkg/libmachine/host/migrate_test.go
```diff
@@ -0,0 +1,116 @@
...
+				},
+			},
+			expectedMigrationError: nil,
+		},
+		{
+			description: "Config version 4 (from the FUTURE) on disk",
+			rawData: []byte(`{
+    "ConfigVersion": 4,
+    "Driver": {"MachineName": "default"},
+    "DriverName": "virtualbox",
```
gbraad, line 64: we never had such a thing ... so let's not suggest we do ;-)

pkg/libmachine/host/migrate_test.go
```diff
@@ -0,0 +1,116 @@
...
+    "HostOptions": {
+        "Driver": "",
+        "Memory": 0,
+        "Disk": 0,
+        "AuthOptions": {
+            "StorePath": "/Users/nathanleclaire/.code-ready/machine/machines/default"
```
gbraad, line 70: `.code-ready` ?

pkg/drivers/fakedriver/fakedriver.go
guillaumerose: Removed. It was not bringing much info.

go.mod
guillaumerose: not much https://github.com/code-ready/machine/pull/54

the RPC, driver interface and some common structs. Few lines.

pkg/libmachine/host/migrate_test.go
guillaumerose: I rewrote the test, it's now gone.

pkg/libmachine/host/migrate_test.go
guillaumerose: test removed

guillaumerose:
Ready for an other round of review :)

cfergeau (commented):
I think I'd keep all of `pkg/drivers` in `machine`, but no strong preferenc either way.

guillaumerose:
I also hesitated for hyperv driver but at the end, it is much easier to have machine reduced to the strict minimum. The major benefits is the ability to share code between preflights and the driver (how we invoke powershell, etc).

guillaumerose:
rebased!

gbraad (approved):
Addreessed the issues I had, and it uses crc/logging instead of libmachine/log

pkg/crc/machine/machine.go
```diff
@@ -9,9 +9,9 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 	"github.com/code-ready/crc/pkg/crc/network"
-	"github.com/code-ready/machine/libmachine"
+	"github.com/code-ready/crc/pkg/libmachine"
+	"github.com/code-ready/crc/pkg/libmachine/host"
 	"github.com/code-ready/machine/libmachine/drivers"
-	"github.com/code-ready/machine/libmachine/host"
 	"github.com/code-ready/machine/libmachine/log"
```
gbraad, line 15: What would prevent this to become `log "github.com/code-ready/crc/pkg/crc/logging" ?

guillaumerose: Not much. I have a PR almost ready after this one to remove the machine logging and use traditional logrus.

gbraad: 👍

gbraad: Do you want to address that, or push another PR to finalize? I'd rather not keep PRs hanging around, as this causes unnecessary difficult rebases, etc.

guillaumerose: I have to change machine before propagating it here in crc :/

PR #1804, opened 2020-12-28
Issue #1796 Add a config to skip all preflight checks

### LINKED ISSUE #1796: Add single toggle for preflight checks
Have a single toggle for preflight checks to disable with single config set

anjannath:
Fixes #1796

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -59,7 +59,7 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 
 	checkIfNewVersionAvailable(config.Get(cmdConfig.DisableUpdateCheck).AsBool())
 
-	if err := preflight.StartPreflightChecks(config); err != nil {
+	if err := preflight.RunOrSkipPreflightChecks(config); err != nil {
 		return nil, err
 	}
 
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -26,6 +26,7 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
+	SkipAllPreflightChecks  = "skip-all-preflight-checks"
 )
 
 func RegisterSettings(cfg *config.Config) {
@@ -49,6 +50,8 @@ func RegisterSettings(cfg *config.Config) {
 
 	// Telemeter Configuration
 	cfg.AddSetting(ConsentTelemetry, false, config.ValidateBool, config.SuccessfullyApplied)
+	// Skip all preflight checks
+	cfg.AddSetting(SkipAllPreflightChecks, false, config.ValidateBool, config.SuccessfullyApplied)
 }
 
 func isPreflightKey(key string) bool {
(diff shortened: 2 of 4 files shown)


cmd/crc/cmd/start.go
```diff
@@ -63,8 +63,11 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 
 	checkIfNewVersionAvailable(config.Get(cmdConfig.DisableUpdateCheck).AsBool())
 
-	if err := preflight.StartPreflightChecks(config); err != nil {
-		return nil, err
+	// If skip-all-preflight-checks is true then don't run the prefligh checks
+	if !config.Get(cmdConfig.SkipAllPreflightChecks).AsBool() {
```
gbraad, line 67: add log that all preflights were explicitly skipped

gbraad (changes requested):
needs a log and/or feedback to the CLI

cmd/crc/cmd/start.go
anjannath: Added info log as..  `All preflight checks are skipped`

cmd/crc/cmd/start.go
```diff
@@ -63,8 +63,13 @@ func runStart(arguments []string) (*machine.StartResult, error) {
...
+		if err := preflight.StartPreflightChecks(config); err != nil {
+			return nil, err
+		}
+	} else {
+		logging.Info("All preflight checks are skipped")
```
gbraad, line 72: can this be inverted ?

```
if config.Get(cmdConfig.SkipAllPreflightChecks).AsBool() {
    logging.Info("All preflight checks are skipped")
} else {
    ...
}
```

Why the inversion of the value? Having the output clearly visible explains the purpose of this.

gbraad (changes requested):
invert if

cmd/crc/cmd/start.go
anjannath: Changed!

guillaumerose:
You also need to do that for the daemon API. 

Duplicated code sadely. We need to find a way to avoid that. (not here if you want to avoid a long debate :))

anjannath:
> You also need to do that for the daemon API.
> 
> Duplicated code sadely. We need to find a way to avoid that. (not here if you want to avoid a long debate :))

Thanks so much,  forgot about the daemon :( for now added the same function in the daemon, we do need to avoid duplicating this..

cmd/crc/cmd/start.go
```diff
@@ -83,6 +83,14 @@ func runStart(arguments []string) (*machine.StartResult, error) {
...
 }
 
+func runOrSkipPreflightChecks(skip bool) error {
+	if skip {
+		logging.Info("All preflight checks are skipped")
+		return nil
+	}
+	return preflight.StartPreflightChecks(config)
+}
+
```
praveenkumar, line 93: This function is duplication for start and handler, better to have it part of preflight package itself and make `StartPreflightChecks` less visible. you can take `skip and config` as parameter for this one.

anjannath: Changed!

guillaumerose (approved):

guillaumerose:
Rebased. Let's wait for the CI now.

PR #1806, opened 2020-12-29
Change the logic for embed and exclude the bundle

praveenkumar:
## Solution/Idea
Right now, because the way we are embedding the bundle and binary is causing issue for signing in the mac side, this PR will try to resolve that by not using the bundle as part of embed and also use go generate logic to embed rest of binaries.

## Proposed changes
Remove the bundle embedding logic

## Testing

$ make generate
$ make cross
$ rm -fr ~/.crc/bin
$ crc setup --log-level debug


.gitignore
```diff
@@ -1,3 +1,5 @@
 out/
 docs/build/
 release/
+data/
```
gbraad, line 4: data ?>

Makefile
```diff
@@ -172,40 +173,30 @@ gen_release_info:
...
-	@$(call check_defined, BUNDLE_DIR, "Embedding bundle requires BUNDLE_DIR set to a directory containing CRC bundles for all hypervisors")
+	pushd $(RELEASE_DIR) && sha256sum * > sha256sum.txt && popd
 
-embed_bundle: clean cross $(HOST_BUILD_DIR)/crc-embedder check_bundledir $(HYPERKIT_BUNDLENAME) $(HYPERV_BUNDLENAME) $(LIBVIRT_BUNDLENAME)
-	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=darwin --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/macos-amd64/crc
-	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=linux --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/linux-amd64/crc
-	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=windows --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/windows-amd64/crc.exe
+.PHONY: generate
+generate:
+	GOS=linux go generate ./...
```
gbraad, line 197: what does generate do?

it seems crc-embedder is not used for the bundle, but is for the other tasks?

gbraad: It will use the directive `//go:generate go run generator.go`

gbraad:
It seems the GitHub interface hides: ` pkg/embed/generator.go `.

guillaumerose:
Sounds a good solution! Why can't you do the same thing for the bundle ? Does it implode ?

Makefile
cfergeau: I was also puzzled when reading ` Use generator to create embed file/binary content `, I'd say this is missing a commit log

cfergeau:
I read the PR description, the commit logs, and I only have a vague idea what the initial problem was, but I have no clue how this was solved/what the end result is compared to what we had initially/...

praveenkumar:
> Sounds a good solution! Why can't you do the same thing for the bundle ? Does it implode ?

@guillaumerose yes it implode, since the bundle file is way to big to be part of auto generated code so it fails.

Makefile
praveenkumar: I will try to add some more doc in the commit logs.

pkg/embed/embed.go
```diff
@@ -32,19 +55,18 @@ func Extract(embedName, destFile string) error {
 
 func ExtractFromExecutable(executablePath, embedName, destFile string) error {
 	logging.Debugf("Extracting embedded '%s' from %s to %s", embedName, executablePath, destFile)
-	reader, err := openEmbeddedFile(executablePath, embedName)
-	if err != nil {
-		return err
+	reader := Get(executablePath)
```
cfergeau, line 58: This is not an `io.Reader`, would be better to rename it.

pkg/embed/generator.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		// Add file name to map
+		configs[fileName] = b
+	}
+
+	if err != nil {
+		log.Fatal("Error walking through embed directory:", err)
+	}
+
+	generatedFileName := fmt.Sprintf("%s_%s.go", blobFileName, os.Getenv("GOS"))
```
cfergeau, line 80: `GOS` or `GOOS`?

pkg/embed/generator.go
```diff
@@ -0,0 +1,138 @@
...
+	// Create blob file
+	f, err := os.Create(generatedFileName)
+	if err != nil {
+		log.Fatal("Error creating blob file:", err)
+	}
+	defer f.Close()
+
```
cfergeau, line 87: `f` does not seem to be used?

pkg/embed/embed.go
```diff
@@ -32,19 +55,18 @@ func Extract(embedName, destFile string) error {
...
 
-	defer reader.Close()
 	writer, err := os.Create(destFile)
 	if err != nil {
 		return fmt.Errorf("Could not create '%s': %v", destFile, err)
 	}
 	defer writer.Close()
 
-	_, err = io.Copy(writer, reader)
+	_, err = io.WriteString(writer, string(reader))
```
cfergeau, line 69: Looks like this could use `ioutil.WriteFile` ?

pkg/embed/embed.go
```diff
@@ -6,20 +6,43 @@ import (
...
+func (e *embedFiles) Get(file string) []byte {
+	if f, ok := e.storage[file]; ok {
+		return f
 	}
-	return reader, nil
+	return nil
+}
+
+// Embed box expose
+var box = newEmbedBox()
```
cfergeau, line 36: Not really sold on the `box` naming, is this naming commonly used in other projects?

cfergeau (commented):
> > Sounds a good solution! Why can't you do the same thing for the bundle ? Does it implode ?
> 
> @guillaumerose yes it implode, since the bundle file is way to big to be part of auto generated code so it fails.

I suspect this also has an impact on memory usage while CRC is running and that it will use about 100MB more? (size of `oc` binary + size of the machine driver + size of `goodhosts`)

cfergeau:
> I read the PR description, the commit logs, and I only have a vague idea what the initial problem was, but I have no clue how this was solved/what the end result is compared to what we had initially/...

Ideally, there would be more details about what is failing with the current `binappend` approach + codesign, the commit log is a bit too vague (is `codesign` returning an error? is everything seemingly working, but then the binary behaves as if it wasn't signed? something else?). And I'd explain in plain words which solution was picked to replace this, with an explanation why it works better, ..

pkg/embed/embed.go
praveenkumar: @cfergeau this is something which packr uses https://github.com/gobuffalo/packr#in-code so I tried to use the same.

pkg/embed/generator.go
cfergeau: Ok, I saw the updated commit log. Let's use a name for this var which does not look like a typo rather than resorting to documentation. `TARGET_OS`?

anjannath:
I tried to use codesign to sign the build from this PR and it worked!

```
$ codesign --verify --deep --verbose out/macos-amd64/crc
out/macos-amd64/crc: code object is not signed at all
In architecture: x86_64

# After signing

$ codesign -s "Anjan Nath" ./out/macos-amd64/crc                                                                                                                                                                                      1 ↵
╭─anjan@Anjans-MacBook-Pro ~/github.com/code-ready/crc ‹aa203cd4›
╰─$ codesign --verify --verbose=4 out/macos-amd64/crc
out/macos-amd64/crc: valid on disk
out/macos-amd64/crc: satisfies its Designated Requirement
```

cfergeau (commented):
In `Remove crc embedder command and sub command` log, it is not clear what exactly is failing, which would be useful to record for future reference. I think it's something like "signature process with codesign apparently works fine, but then signature verification of the binary does not work as expected because ..." ?

`Use go generator to create embed file/binary content` is really not clear on what is doing:
- this replaces the `binappend` code which causes macos code signing issues
- this uses `go generate` to automatically generate go byte[] arrays whose value is the content of the file(s) we want to embed
- since these embedded files are now part of the actual executable rather than being appended after it, this has impact on runtime memory usage. In particular, this means we can't embed our 2GB+ bundle in the binary anymore.

How the release process needs to be adjusted, how crc will be shipped to users, what changes will be visible in everyday use, ... is not detailed, which would be useful to know in order to understand better where we are going :)

`Remove mock release check for travis and circle CI` when you say 'it require much more resource to compile.', do you have an idea how much more memory we are talking about? half a gig or so, or is this more than that? (just curious, not very important)

pkg/embed/generator.go
```diff
@@ -0,0 +1,131 @@
...
+}`),
+)
+
+func fmtByteSlice(s []byte) string {
+	builder := strings.Builder{}
+
+	for _, v := range s {
+		builder.WriteString(fmt.Sprintf("%d,", int(v)))
+	}
+
```
cfergeau, line 46: This could be
```go
+       for i, v := range s {
+               if i%32 == 0 {
+                       builder.WriteString("\n")
+               }
+               builder.WriteString(fmt.Sprintf("%#02x,", int(v)))
        }
```
which gives a generated file which is much nicer to read.

pkg/crc/machine/hyperkit/constants.go
```diff
@@ -1,5 +1,3 @@
-//+build darwin build
-
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau yes and that is causing issue during the code generation as I mentioned in the commit log.

pkg/embed/generator.go
praveenkumar: Sure I just hope it doesn't break anything, will try.

centos_ci.sh
```diff
@@ -132,7 +132,7 @@ function run_tests() {
   # this is copied over using https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml#L99
   export PULL_SECRET_FILE=--pull-secret-file=$HOME/payload/crc_pull_secret
   export BUNDLE_LOCATION=--bundle-location=$HOME/Downloads/$BUNDLE 
-  make integration 
+  make integration GODOG_OPTS="--godog.tags='@basic && @linux'"
```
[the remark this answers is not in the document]
praveenkumar: done

centos_ci.sh
```diff
@@ -52,7 +52,8 @@ function install_required_packages() {
     jq \
     gcc \
     unzip \
-    podman
+    podman \
+    zip
```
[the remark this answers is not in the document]
praveenkumar: done

pkg/crc/machine/hyperkit/constants.go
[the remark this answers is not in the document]
cfergeau: Yes, I've seen this is causing an issue, the question I'm raising is whether we want to try to solve the issue differently to keep the per-platform files, or if we keep things this way.

pkg/crc/constants/constants.go
```diff
@@ -123,26 +122,26 @@ func EnsureBaseDirectoriesExist() error {
...
-	if err != nil {
-		return false
-	}
-	extractor, err := binappend.MakeExtractor(executablePath)
+// BundlePresent returns true and path of the bundle
+// if the bundle available in user's home directory.
+func BundlePresent() (bool, string) {
+	var bundlePath string
+	var present bool
+	err := filepath.Walk(GetHomeDir(),
```
cfergeau, line 130: I don't think we want to do this. Either the bundle is present in ~/.crc/cache, or it's in the same directory as the crc executable, or the user will have to specify it with `-b`.

.circleci/config.yml
```diff
@@ -37,37 +37,6 @@ jobs:
         path: ~/project/out
         destination: crc
 
-  build_mock_release:
-    machine:
```
cfergeau, line 41: The 10GB+ memory needed to build with the generated files could be mentioned in this commit log as a data point :)

pkg/embed/generator.go
```diff
@@ -0,0 +1,131 @@
...
+	defer os.RemoveAll(embedFolder)
+
+	// Download platform specific embedded files
+	downloadedFiles, err := downloadDataFiles(os.Getenv("TARGET_OS"), embedFolder)
+	if err != nil {
+		log.Fatal("Error downloading embedded files", err)
+	}
+
+	// Create map for filenames
+	configs := make(map[string][]byte)
```
cfergeau, line 64: Why is it called `configs`?

pkg/crc/machine/hyperkit/constants.go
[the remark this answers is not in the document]
cfergeau: This seems to do the trick:
```
diff --git a/pkg/embed/embed.go b/pkg/embed/embed.go
index 9cdb5218..f95ddf3f 100644
--- a/pkg/embed/embed.go
+++ b/pkg/embed/embed.go
@@ -8,7 +8,7 @@ import (
 )
 
-//go:generate go run generator.go
+//go:generate go run -tags=build generator.go
 
 type embedFiles struct {
        storage map[string][]byte
```

cfergeau:
Another data point which I forgot, RSS for `crc daemon` on linux goes from 28MB to 51MB. This is about twice the increase in binary size (26,341,376 bytes -> 38,113,280 bytes), dunno why

pkg/crc/constants/constants.go
praveenkumar: @cfergeau I am doing it to retain the existing workflow as much as possible, do you think it might have any kind of side effect except the time taken by this logic if `HOMEDIR` is huge?

pkg/embed/generator.go
cfergeau: Actually, using `[]byte("\x00\x00\x00...")` makes the build resource consumption much much better than `[]byte{ 0x00, 0x00, 0x00 ...}` so we should use this instead:
```
diff --git a/pkg/embed/generator.go b/pkg/embed/generator.go
index 8fdb1a1f..c3668445 100644
--- a/pkg/embed/generator.go
+++ b/pkg/embed/generator.go
@@ -32,7 +32,7 @@ var tmpl = template.Must(template.New("").Funcs(conv).Parse(`package embed
 
 func init() {
     {{- range $name, $file := . }}
-        Add("{{ $name }}", []byte{ {{ conv $file }} })
+        Add("{{ $name }}", []byte("{{ conv $file }}"))
     {{- end }}
 }`),
 )
@@ -41,7 +41,7 @@ func fmtByteSlice(s []byte) string {
 	builder := strings.Builder{}
 
 	for _, v := range s {
-		builder.WriteString(fmt.Sprintf("%d,", int(v)))
+		builder.WriteString(fmt.Sprintf("\\x%02x", int(v)))
 	}
 
 	return builder.String()
```

cfergeau: The corresponding golang bug i s https://github.com/golang/go/issues/6643

pkg/crc/constants/constants.go
gbraad: on our meeting I said exactly the same as Christophe. Anything else is guess work and should be avoided.

gbraad: if the bundle is NOT in the current dir, the user likely moved the binary out of place (without considering the bundle), so this ia a user error.

Fail early ... and fail with a clear error/instruction.

.circleci/config.yml
cfergeau: With the changes to use a string for the embedded data rather than a byte array, it should be possible to drop ` Remove mock release check for travis and circle CI ` ?

pkg/crc/constants/constants.go
```diff
@@ -123,26 +122,18 @@ func EnsureBaseDirectoriesExist() error {
...
-	}
-	return contains(extractor.AvalibleData(), GetDefaultBundle())
-}
-
-func contains(arr []string, str string) bool {
-	for _, a := range arr {
-		if a == str {
-			return true
+// BundlePresent returns true and path of the bundle
+// if bundle is available in current directory or
```
cfergeau, line 126: It's not the current directory we want to look into, but the directory where the `crc` executable is.
In this scenario:
```
$ mkdir foo
$ cd foo
$ ../crc start
```
you want to look for the bundle in `..`, not in the current directory, which is `foo`. `https://golang.org/pkg/os/#Executable` should  be useful for this.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -34,7 +33,7 @@ var genericPreflightChecks = [...]Check{
 		configKeySuffix:  "check-bundle-extracted",
 		checkDescription: "Checking if CRC bundle is extracted in '$HOME/.crc'",
 		check:            checkBundleExtracted,
-		fixDescription:   "Extracting bundle from the CRC executable",
+		fixDescription:   "Extracting bundle from the cache or current directory",
```
cfergeau, line 36: or just "Extracting CodeReady Containers bundle"?

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -55,35 +54,21 @@ var genericPreflightChecks = [...]Check{
 }
 
 func checkBundleExtracted() error {
-	if !constants.BundleEmbedded() {
+	if present, _ := constants.BundlePresent(); !present {
 		return nil
```
cfergeau, line 58: Logic seems wrong here, if the bundle is not present, returning no error without any further check does not seem like what we want?

pkg/crc/validation/validation.go
```diff
@@ -69,7 +69,7 @@ func ValidateBundle(bundle string) error {
 	releaseBundleVersion := version.GetBundleVersion()
 	userProvidedBundleVersion := filepath.Base(bundle)
 	if !strings.Contains(userProvidedBundleVersion, fmt.Sprintf("%s.crcbundle", releaseBundleVersion)) {
-		if !constants.BundleEmbedded() {
+		if present, _ := constants.BundlePresent(); !present {
 			logging.Warnf("Using unsupported bundle %s", userProvidedBundleVersion)
```
cfergeau, line 73: We were using `!constants.BundleEmbedded()` as a way to differentiate between releases and builds which are not release binaries. I don't think the same logic still works when the bundle is no longer embedded.

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau this is part of fix in setup so I think explicitly mention the bundle directory make more sense then just putting a general statement.

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau this is the setup part and here if we return any error that means we are failing the setup flow, this check is only to make sure if the bundle is present in current directory or cache one, if yes then does the bundle is extracted and if not then run the fix which do the extraction.

pkg/crc/validation/validation.go
praveenkumar: @cfergeau Yes we can remove this logic in that case we are not going to return the error message but only the `Warnf` message.

pkg/crc/preflight/preflight_checks_common.go
cfergeau: thinking just now that "current directory" is not accurate. And when things "just work", I don't think the user cares a lot about where things are, they only care that this is being extracted. Could be a good debug message toe report if `checkBundleExtracted` when returning nil because the bundle could not be found actually "Not extracting bundle as we couldn't find it in ~/.crc/cache nor in the same directory as the crc executable"

praveenkumar:
/close

In favor of #1834

PR #1810, opened 2021-01-04
Delete the VM directly without stopping it in most integration tests

guillaumerose:
`crc stop` takes too much time esp. when we delete the VM just after. Delete it directly and avoid a graceful shutdown. 

`crc stop` is still tested in basic (check the console is here after a restart) and health (check user app is still here after restart) integration tests.


jsliacan (approved):
Fair point. LGTM. Thanks

guillaumerose:
retest this please

PR #1811, opened 2021-01-04
Use ID_LIKE for most preflights checks

guillaumerose:
This should fix some derivative distros of Ubuntu (popos, linux mint).

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -144,10 +144,10 @@ func fixLibvirtInstalled(distro *linux.OsRelease) func() error {
 
 func installLibvirtCommand(distro *linux.OsRelease) string {
 	yumCommand := "yum install -y libvirt libvirt-daemon-kvm qemu-kvm"
-	switch distroID(distro) {
-	case linux.Ubuntu:
+	switch {
+	case distroIsLike(distro, linux.Ubuntu):
 		return "apt-get update && apt-get install -y libvirt-daemon libvirt-daemon-system libvirt-clients"
-	case linux.RHEL, linux.CentOS, linux.Fedora:
+	case distroIsLike(distro, linux.Fedora):
 		return yumCommand
 	default:
 		logging.Warnf("unsupported distribution %s, trying to install libvirt with yum", distro)
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -180,7 +180,7 @@ func getNetworkChecksForDistro(distro *linux.OsRelease, networkMode network.Mode
 
 	switch {
 	default:
-		logging.Warnf("distribution-specific preflight checks are not implemented for '%s'", distroID(distro))
+		logging.Warnf("distribution-specific preflight checks are not implemented for '%s'", distro.ID)
 		fallthrough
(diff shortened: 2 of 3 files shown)


pkg/crc/preflight/preflight_linux.go
```diff
@@ -218,14 +218,6 @@ func usesSystemdResolved(distro *linux.OsRelease) bool {
 	}
 }
 
-func distroID(osRelease *linux.OsRelease) linux.OsType {
-	if osRelease == nil {
```
cfergeau, line 222: `distro()` can return `nil` if `linux.GetOsRelease()` fails, so it's possible to get a `nil` `*linux.OsRelease` here.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -144,10 +144,10 @@ func fixLibvirtInstalled(distro *linux.OsRelease) func() error {
 
 func installLibvirtCommand(distro *linux.OsRelease) string {
 	yumCommand := "yum install -y libvirt libvirt-daemon-kvm qemu-kvm"
-	switch distroID(distro) {
-	case linux.Ubuntu:
+	switch {
+	case distroIsLike(distro, linux.Ubuntu):
 		return "apt-get update && apt-get install -y libvirt-daemon libvirt-daemon-system libvirt-clients"
-	case linux.RHEL, linux.CentOS, linux.Fedora:
+	case distroIsLike(distro, linux.Fedora), distroIsLike(distro, linux.RHEL):
```
cfergeau, line 150: fwiw, RHEL and CentOS have `ID_LIKE=fedora`

pkg/crc/preflight/preflight_linux.go
guillaumerose: outch, thanks!

guillaumerose: done

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -144,10 +144,10 @@ func fixLibvirtInstalled(distro *linux.OsRelease) func() error {
...
 func installLibvirtCommand(distro *linux.OsRelease) string {
 	yumCommand := "yum install -y libvirt libvirt-daemon-kvm qemu-kvm"
-	switch distroID(distro) {
-	case linux.Ubuntu:
+	switch {
+	case distroIsLike(distro, linux.Ubuntu):
 		return "apt-get update && apt-get install -y libvirt-daemon libvirt-daemon-system libvirt-clients"
-	case linux.RHEL, linux.CentOS, linux.Fedora:
+	case distroIsLike(distro, linux.Fedora):
 		return yumCommand
```
cfergeau, line 151: While you are changing this method, you could probably remove the `yumCommand` variable which doesn't bring much, and is inconsistent with the ubuntu case.

cfergeau (approved):
I would move the 'Always return an OsRelease struct, even if /etc/os-release is absent ' commit before `Assume systemd resolved is used by all derivative distros of Ubuntu` and `Remove distroID(), use directly distro.ID` as they both are dependent on this change.

cfergeau (approved):

pkg/crc/preflight/preflight_checks_linux.go
guillaumerose: I changed the order but for yumCommand, I would have to duplicate in the fedora check and in the default check. Perhaps you have an other idea?

PR #1814, opened 2021-01-05
Consistency on integration test for nameserver configuration

adrianriobo:
The test start the crc cluster with a nameserver for windows environment. (This is required due to some known issues with windows dns resolutions). 

During the test the cluster is stopped and then started again. The second start does not configure any nameserver. This change will add the nameserver to this second start. Also to make it consistent across all plarforms the namesever is also configured for the first crc start on linux and macos.

### DIFF
--- a/test/integration/features/story_health.feature
+++ b/test/integration/features/story_health.feature
@@ -10,7 +10,7 @@ Feature: End-to-end health check
     Scenario: Start CRC
         Given executing "crc setup" succeeds
         When setting config property "memory" to value "12000" succeeds
-        When starting CRC with default bundle succeeds
+        When starting CRC with default bundle and nameserver "10.75.5.25" succeeds
         Then stdout should contain "Started the OpenShift cluster"
         And executing "eval $(crc oc-env)" succeeds
         When checking that CRC is running
--- a/test/integration/features/story_health.feature
+++ b/test/integration/features/story_health.feature
@@ -59,7 +59,7 @@ Feature: End-to-end health check
         Given with up to "2" retries with wait period of "60s" http response from "http://httpd-example-testproj.apps-crc.testing" has status code "200"
         When executing "crc stop -f" succeeds
         Then checking that CRC is stopped
-        When starting CRC with default bundle succeeds
+        When starting CRC with default bundle and nameserver "10.75.5.25" succeeds
         Then checking that CRC is running
         And with up to "2" retries with wait period of "1m" http response from "http://httpd-example-testproj.apps-crc.testing" has status code "200"
 


test/integration/features/story_health.feature
```diff
@@ -59,7 +59,7 @@ Feature: End-to-end health check
         Given with up to "2" retries with wait period of "60s" http response from "http://httpd-example-testproj.apps-crc.testing" has status code "200"
         When executing "crc stop -f" succeeds
         Then checking that CRC is stopped
-        When starting CRC with default bundle succeeds
+        When starting CRC with default bundle and nameserver "10.75.5.25" succeeds
```
praveenkumar, line 62: `10.75.5.25` is internal dns nameserver (belong to RH), this might fail the tests on upstream CI or add a additional lag because now every request first go to `10.75.5.25` which will fail and then the default nameserver.

praveenkumar (changes requested):

test/integration/features/story_health.feature
adrianriobo: Oks, I will switch to `8.8.8.8`

praveenkumar: with `8.8.8.8` this might not work on internal network because RH network doesn't allow any public dns.

praveenkumar: If you really want this to part of test then better to have some kind of variable which value you can substitute for internal and external CI testing.

guillaumerose: Maybe we can add an env variable and the test binary will read it and use it as the nameserver. 

`CRC_NS=10.75.5.25 make integration` for the internal CI and `make integration` for the public CI.

praveenkumar: Yes, setting a env variable also good option.

adrianriobo: I will close this PR. Then I am thinking to create two different ones: 

* First PR to solve  the windows  feature story_health. As there is one specific scenario for windows to start the cluster (#L20) if I set the nameserver as config instead of passing the arg it will keep the config for the second start. The rest will remain the same.
* Second PR to read the ENV and set the nameserver. This is a good enhancement but we need to make sure when this env will be applied as this would be a bit hidden (the execution parameter wouldn't be reflect as code)

WDYT?

jsliacan: I'm not sure about the ENV var approach, exactly because of what @adrianriobo mentions: it's hidden. Ideally, all settings would be explicit in the Feature file -- that's the "test declaration".

praveenkumar: I would see this like the `bundleEmbedded` logic where we pass the `-b` argument if bundle is not present, something similar can be done.
```
if !bundleEmbedded {
		extraBundleArgs = fmt.Sprintf("-b %s", bundleName)
}
if os.Getenv("CRC_NS") != "" {
                 extraArgs = fmt.Sprintf("-n %s", os.Getenv("CRC_NS"))
}
```

adrianriobo: Kind of similar as when we set the -b we want to run all e2e with same bundle. In case for nameserver it only applies to certain scenarios where we want to start with nameserver. 

I mean having this is good as far as I think right now we are not testing nameserver on public CI, and this will allow to test that on public. But adding this just requires a bit more thinking than the initial scope for this PR.. 

On the other hand moving the nameserver to the config does not change the current behaviour and solves the windows environment execution issue.

guillaumerose: With the current crc, you can already use `CRC_NAMESERVER` env variable. Perhaps if you run the test with this env. variable, it will be propagated to crc.

adrianriobo: Thanks @guillaumerose  CRC_NAMESERVER work as you suggested. So I will close this PR

PR #1815, opened 2021-01-05
Remove kubelet status check from start

praveenkumar:
It made sense before to check the status before perfoming some of
the cluster specific tasks since we were stopping/starting kubelet
service many time but now we just start the service and then do all
the cluster specific operations. If there is any issue with kubelet
service it is failed in early stage.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -23,7 +23,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/ssh"
 	crcssh "github.com/code-ready/crc/pkg/crc/ssh"
 	"github.com/code-ready/crc/pkg/crc/systemd"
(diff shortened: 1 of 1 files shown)


guillaumerose:
Good! Should we also change the returned struct and remove the boolean KubeletStarted ?

gbraad:
> remove KubeletStarted 

most likely

cfergeau:
> > remove KubeletStarted
> 
> most likely

Wouldn't this be useful when podman support comes back?

praveenkumar:
> > > remove KubeletStarted
> > 
> > 
> > most likely
> 
> Wouldn't this be useful when podman support comes back?

Yes, I will keep it as of now and this might be required when podman support is added.

anjannath:
The `KubeletStarted` is also used in the trays to show either  _start has finished succesfully_  or _start did not finish succesfully_ pop up from the trays...

PR #1816, opened 2021-01-05
Remove warn flags from the config

praveenkumar:
`warn-<preflight-check>` is not used in crc context because we
are already capturing the checks data as part of debug logs. `skip`
is there to skip a check if a user don't want to run a check.

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -52,22 +52,15 @@ func RegisterSettings(cfg *config.Config) {
 }
 
 func isPreflightKey(key string) bool {
-	return strings.HasPrefix(key, "skip-") || strings.HasPrefix(key, "warn-")
+	return strings.HasPrefix(key, "skip-")
 }
 
 // less is used to sort the config keys. We want to sort first the regular keys, and
-// then the keys related to preflight starting with a skip- or warn- prefix. We want
-// these preflight keys to be grouped by pair: 'skip-bar', 'warn-bar', 'skip-foo', 'warn-foo'
-// would be sorted in that order.
+// then the keys related to preflight starting with a skip- prefix.
 func less(lhsKey, rhsKey string) bool {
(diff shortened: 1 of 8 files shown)


cmd/crc/cmd/config/config.go
```diff
@@ -52,22 +52,17 @@ func RegisterSettings(cfg *config.Config) {
...
 func isPreflightKey(key string) bool {
-	return strings.HasPrefix(key, "skip-") || strings.HasPrefix(key, "warn-")
+	return strings.HasPrefix(key, "skip-")
 }
 
 // less is used to sort the config keys. We want to sort first the regular keys, and
-// then the keys related to preflight starting with a skip- or warn- prefix. We want
-// these preflight keys to be grouped by pair: 'skip-bar', 'warn-bar', 'skip-foo', 'warn-foo'
+// then the keys related to preflight starting with a skip- prefix. We want
+// these preflight keys to be grouped as: 'skip-bar', 'skip-foo'
```
cfergeau, line 60: The last sentence is no longer useful ("We want these preflight keys to be grouped ..."), it can be removed.

docs/source/topics/con_about-codeready-containers-configuration.adoc
```diff
@@ -9,4 +9,5 @@ Run the [command]`{bin} config --help` command to list the available properties.
 
 You can also use the [command]`{bin} config` command to configure the behavior of the startup checks for the [command]`{bin} start` and [command]`{bin} setup` commands.
 By default, startup checks report an error and stop execution when their conditions are not met.
-Set the value of a property starting with `skip-check` or `warn-check` to `true` to skip the check or report a warning rather than an error, respectively.
+Set the value of a property starting with `skip-check` to `true` to skip the check.
+s
```
cfergeau, line 13: `s`?

cfergeau (approved):

PR #1817, opened 2021-01-05
Send more data to segment

guillaumerose:
* Use track and identify calls
* Track all commands (success = true|false)
* Track the duration of each command

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -175,13 +175,11 @@ func addForceFlag(cmd *cobra.Command) {
 func executeWithLogging(fullCmd string, input func(cmd *cobra.Command, args []string) error) func(cmd *cobra.Command, args []string) error {
 	return func(cmd *cobra.Command, args []string) error {
 		logging.Debugf("Running '%s'", fullCmd)
-		if err := input(cmd, args); err != nil {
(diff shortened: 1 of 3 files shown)


cmd/crc/cmd/root.go
```diff
@@ -174,13 +175,12 @@ func addForceFlag(cmd *cobra.Command) {
 
 func executeWithLogging(fullCmd string, input func(cmd *cobra.Command, args []string) error) func(cmd *cobra.Command, args []string) error {
 	return func(cmd *cobra.Command, args []string) error {
-		if err := input(cmd, args); err != nil {
-			if serr := segmentClient.Upload(fullCmd, err); serr != nil {
-				fmt.Println(serr.Error())
-			}
-			return err
+		startTime := time.Now()
+		err := input(cmd, args)
```
praveenkumar, line 179: The initial agreement which we had with PM about tracking only the errors not all the commands, did we get an ack for this implementation?

guillaumerose: I didn't ask (yet). I am more worried that we will have 0 data with the current implementation.

Can we add this new data @sspeiche ?

sspeiche: There wasn't a limitation on tracking just errors, so this is fine. We just have limitations on tracking personally identifiable information like IP address.

PR #1818, opened 2021-01-05
Ask the user to turn on telemetry when running crc setup

guillaumerose:
``` 
$ crc setup 
CodeReady Containers is constantly improving and we would like to know more about usage!
Would you like to contribute anonymous usage statistics (more details at https://developers.redhat.com/article/tool-data-collection)? [y/N]: y
Thanks for helping us! You can disable telemetry with the command 'crc config set consent-telemetry no'.
...

$ crc setup 
CodeReady Containers is constantly improving and we would like to know more about usage!
Would you like to contribute anonymous usage statistics (more details at https://developers.redhat.com/article/tool-data-collection)? [y/N]: 
No worry, you can still enable telemetry manually with the command 'crc config set consent-telemetry yes'.
...
```


jsliacan:
What's the idea of `-f` with `setup`? To avoid getting the prompt?

guillaumerose:
Yes, avoid the prompt when crc is used in a script on in a CI. There is a condition to detect if the session is interactive or not, but the user can still force it.

gorkem:
+1 to adding the instructions for turning off. 
We also need to refer to the privacy statement[1]
FYI @sspeiche 

[1] https://developers.redhat.com/article/tool-data-collection

cmd/crc/cmd/setup.go
```diff
@@ -31,6 +34,13 @@ var setupCmd = &cobra.Command{
 }
 
 func runSetup(arguments []string) error {
+	if !config.Get(cmdConfig.ConsentTelemetry).AsBool() && terminal.IsTerminal(int(os.Stderr.Fd())) && !globalForce {
+		if input.PromptUserForYesOrNo("Would you like to contribute anonymous usage statistics", false) {
```
praveenkumar, line 38: I would also provide https://developers.redhat.com/article/tool-data-collection in case user want to know about our policy.

gbraad (changes requested):
needs to refer to the policy

gbraad:
> There is a condition to detect if the session is interactive or not, but the user can still force it.

This has a side-effect of 'forcing' false or true ... or ignoring the no-value state. understandable for the non-interactive flow. but this sounds almost like the `skip-all` for the preflights as this has other side-effects when used.

cmd/crc/cmd/setup.go
cfergeau: Ideally we'd be able to differentiate between "ConsentTelemetry was never set", and "ConsentTelemetry has been set to false in a previous crc setup run", at the moment I think we'll be asking again and again if the user answered yes the first time around?

guillaumerose: @praveenkumar @gbraad I added the privacy notice. 

```
$ crc setup 
CodeReady Containers is constantly improving but we need your feedback.
Would you like to contribute anonymous usage statistics (more details at https://developers.redhat.com/article/tool-data-collection)? [y/N]: y
Thanks for helping us! You can disable telemetry at any time with the command 'crc config set consent-telemetry false'.
...
```

@cfergeau yes you're totally right. We have no mechanism to detect that right now. I would say that the question is not invasive. If you press directly enter, it assumes false.

anjannath: We can document that if the user doesn't want to see the prompt they can use the `-f` flag which will suppress it

cfergeau: >  We have no mechanism to detect that right now. I would say that the question is not invasive. If you press directly enter, it assumes false.

This could be perceived as nagging the user until they say yes, as when they accept, we never ask again, when they refuse, we ask every time crc setup is run.

cmd/crc/cmd/setup.go
```diff
@@ -31,6 +32,20 @@ var setupCmd = &cobra.Command{
 }
 
 func runSetup(arguments []string) error {
+	if config.Get(cmdConfig.ConsentTelemetry).AsString() == "" {
```
praveenkumar, line 35: Now we are not adding the force option, CI will fail because it need the input?

guillaumerose: Yes this is what I want to try. `input.PromptUserForYesOrNo` checks if the sessions is interactive. https://ci.centos.org/job/codeready-crc-pr/2754/console seems to pass so it correctly falls in the no.

jsliacan: That's nice, I didn't know that was a thing... So I could theoretically change `delete -f` instances to `delete` in the tests?

cfergeau (approved):

Issue #1773, opened 2020-12-11
[BUG] domain and dns disappear from resolved once connected/disconnected to vpn (Fedora-33)


praveenkumar:
On fedora-33 we are experiencing the crc domain and dns disappear from the `resolvectl --dns` or `resolvectl --domain` when connecting to vpn or even disconnecting it. Somehow the NetworkManager dispatcher file only active single time. 

```
// On vpn
$ sudo resolvectl domain
Global: 
Link 2 (enp0s31f6):
Link 4 (wlp4s0):
Link 5 (virbr3):
Link 6 (virbr3-nic):
Link 9 (virbr1):
Link 10 (virbr1-nic):
Link 11 (virbr0):
Link 12 (virbr0-nic):
Link 39 (enp0s20f0u4u1u2): ~.
Link 48 (tun0): redhat.com
Link 49 (crc): ~testing
Link 50 (crc-nic):

// Disconnect the vpn
$ sudo resolvectl domain
Global: 
Link 2 (enp0s31f6):
Link 4 (wlp4s0):
Link 5 (virbr3):
Link 6 (virbr3-nic):
Link 9 (virbr1):
Link 10 (virbr1-nic):
Link 11 (virbr0):
Link 12 (virbr0-nic):
Link 39 (enp0s20f0u4u1u2): ~.
Link 48 (tun0):
Link 49 (crc):
Link 50 (crc-nic):
```

As a time being https://github.com/code-ready/crc/issues/1624#issuecomment-721056042 can be used but it change the global setting.

cfergeau:
This is a NetworkManager issue, which was fixed in https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=ee4e679bc7479de42780ebd8e3a4d74afa2b2ebe and will be fixed in f33 once this update lands https://bodhi.fedoraproject.org/updates/FEDORA-2020-7851982ff6 (which has issues of its own :-/ )

PR #1819, opened 2021-01-06
fedora33: Workaround NetworkManager bug with systemd-resolved

cfergeau:
crc DNS configuration on systems using systemd-resolved is done through
a NetworkManager dispatcher file which gets run when the crc bridge is
created by libvirt. This file then uses resolvectl to configure the
bridge as the interface to use for resolving names in the .testing
domain.

However, NetworkManager will overwrite this resolvectl configuration
every time a network connection goes up/down. This is a NetworkManager
bug which is fixed in version 1.26.6 by this commit:
https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=ee4e679bc7479de42780ebd8e3a4d74afa2b2ebe

This commit changes the dispatcher script to run every time any network
interface changes state so that we restore crc DNS configuration after
NM overwrites it. This workaround can be removed in a few weeks/months
once the NetworkManager package is upgraded in Fedora (I did not check
if ubuntu is impacted).

This fixes https://github.com/code-ready/crc/issues/1773

## Testing

Using a fedora 33 system with NetworkManager older than 1.26.6
1. after `crc setup`, `resolvectl domain` shows Link x (crc): ~testing
2. start/stop a NM connection with a search domain set (nmcli conn, nmcli conn show $conn |grep ipv4.dns-search, nmcli conn modify $conn ipv4.dns-search ~example.com, nmcli conn down $conn, nmcli conn up $conn) - just stopping/starting the VPN does the tric on my system
3. check `resolvectl domain` again - before this patch, `~testing` is gone, after this patch `~testing` is still present

### DIFF
--- a/pkg/crc/preflight/preflight_checks_network_linux.go
+++ b/pkg/crc/preflight/preflight_checks_network_linux.go
@@ -72,22 +72,27 @@ server=/crc.testing/192.168.130.11
 dns=dnsmasq
 `
 
-	crcNetworkManagerDispatcherPath   = filepath.Join(crcNetworkManagerRootPath, "dispatcher.d", "pre-up.d", "99-crc.sh")
-	crcNetworkManagerDispatcherConfig = `#!/bin/sh
+	crcNetworkManagerOldDispatcherPath = filepath.Join(crcNetworkManagerRootPath, "dispatcher.d", "pre-up.d", "99-crc.sh")
+	crcNetworkManagerDispatcherPath    = filepath.Join(crcNetworkManagerRootPath, "dispatcher.d", "99-crc.sh")
+	crcNetworkManagerDispatcherConfig  = `#!/bin/sh
 # This is a NetworkManager dispatcher script to configure split DNS for
 # the 'crc' libvirt network.
-# The corresponding crc bridge is recreated each time the system reboots, so
-# it cannot be configured permanently through NetworkManager.
-# Changing DNS settings with nmcli requires the connection to go down/up,
-# so we directly make the change using resolvectl
+#
+# The corresponding crc bridge is not created through NetworkManager, so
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -72,22 +72,26 @@ server=/crc.testing/192.168.130.11
 dns=dnsmasq
 `
 
-	crcNetworkManagerDispatcherPath   = filepath.Join(crcNetworkManagerRootPath, "dispatcher.d", "pre-up.d", "99-crc.sh")
+	crcNetworkManagerDispatcherPath   = filepath.Join(crcNetworkManagerRootPath, "dispatcher.d", "99-crc.sh")
```
praveenkumar, line 75: Are we going to remove the old dispatcher file, in case user already using old version and if it going to have any kind of conflict with existing one.

cfergeau: Ah, good point, totally forgot this. Should be fixed now, thanks!

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -266,6 +271,10 @@ func checkCrcNetworkManagerDispatcherFile() error {
 
 func fixCrcNetworkManagerDispatcherFile() error {
 	logging.Debug("Fixing NetworkManager dispatcher configuration")
+
+	// Remove dispatcher script which was used in crc 1.20 - it's been moved to a new location
+	_ = removeNetworkManagerConfigFile(crcNetworkManagerOldDispatcherPath)
```
praveenkumar, line 276: 👍🏼

praveenkumar (approved):

PR #1821, opened 2021-01-06
Fix destination file for win/mac tray

praveenkumar:
Extract function take embedded file name and destination filename to
extract the file from the binary to destination file. We had bug till
now in tray side where we are passing the destination folder instead
destination file and extration fails.

```
DEBU Could not extract tray from crc executableCould not create '/var/folders/6q/9mnplh5s6wg__rrvzdv9nkp80000gn/T/crc094171414': open /var/folders/6q/9mnplh5s6wg__rrvzdv9nkp80000gn/T/crc094171414: is a directory
```

### DIFF
--- a/pkg/crc/preflight/preflight_checks_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_tray_darwin.go
@@ -195,7 +195,9 @@ func downloadOrExtractTrayApp() error {
 	}()
 
 	logging.Debug("Trying to extract tray from crc executable")
-	err = embed.Extract(filepath.Base(constants.GetCRCMacTrayDownloadURL()), tmpArchivePath)
+	trayFileName := filepath.Base(constants.GetCRCMacTrayDownloadURL())
+	trayDestFileName := filepath.Join(tmpArchivePath, trayFileName)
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_tray_darwin.go
```diff
@@ -195,7 +195,9 @@ func downloadOrExtractTrayApp() error {
 	}()
 
 	logging.Debug("Trying to extract tray from crc executable")
-	err = embed.Extract(filepath.Base(constants.GetCRCMacTrayDownloadURL()), tmpArchivePath)
+	trayFileName := filepath.Base(constants.GetCRCMacTrayDownloadURL())
+	trayDestFileName := filepath.Join(tmpArchivePath, trayFileName)
```
cfergeau, line 199: you could remove `archivePath` a bit further in this function. And that's quite some duplication between the windows and macos code :-/

praveenkumar: @cfergeau right now not want to refactor this part and remove duplication but will keep in mind.

anjannath (approved):

PR #1822, opened 2021-01-06
Several fixes for proxy configuration

guillaumerose:
While working with mitmproxy and crc, I've found couple of issues.

* the deployment we watched no longer exist. I updated it.
* 60s is not enough clearly. 
* I removed the restart of kubelet/crio. I only needed to change the configuration of kubelet/crio before the start of kubelet.
* default service ip range changed.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -236,7 +236,7 @@ func AddProxyToKubeletAndCriO(sshRunner *ssh.Runner, proxy *network.ProxyConfig)
 	proxyTemplate := `[Service]
 Environment=HTTP_PROXY=%s
 Environment=HTTPS_PROXY=%s
-Environment=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,%s`
+Environment=NO_PROXY=.cluster.local,.svc,10.116.0.0/14,172.25.0.0/16,192.168.126.0/24,%s`
 	p := fmt.Sprintf(proxyTemplate, proxy.HTTPProxy, proxy.HTTPSProxy, proxy.GetNoProxyString())
 	// This will create a systemd drop-in configuration for proxy (both for kubelet and crio services) on the VM.
 	err := sshRunner.CopyData([]byte(p), "/etc/systemd/system/crio.service.d/10-default-env.conf", 0644)
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -260,6 +260,7 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 		return nil, errors.Wrap(err, "Error getting proxy configuration")
 	}
 	proxyConfig.ApplyToEnvironment()
+	proxyConfig.AddNoProxy(instanceIP)
 
 	// Create servicePostStartConfig for DNS checks and DNS start.
 	servicePostStartConfig := services.ServicePostStartConfig{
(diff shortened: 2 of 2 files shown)


pkg/crc/machine/start.go
```diff
@@ -507,39 +506,27 @@ func updateSSHKeyAndCopyKubeconfig(sshRunner *crcssh.Runner, name string, crcBun
 	return nil
 }
 
-func configProxyForCluster(ocConfig oc.Config, sshRunner *crcssh.Runner, sd *systemd.Commander, proxy *network.ProxyConfig, instanceIP string) (err error) {
+func ensureKubeletAndCRIOAreConfiguredForProxy(sshRunner *crcssh.Runner, proxy *network.ProxyConfig, instanceIP string) (err error) {
```
praveenkumar, line 508: We don't stop the crio service by default so it will be running as soon as VM is up so either that is handled in snc side or here.

guillaumerose: crio is disabled by default and it is started only when we start kubelet.

```
[root@crc-rlc2h-master-0 core]# systemctl status crio
● crio.service - Open Container Initiative Daemon
   Loaded: loaded (/usr/lib/systemd/system/crio.service; disabled; vendor preset: disabled)
[root@crc-rlc2h-master-0 core]# systemctl list-dependencies kubelet
kubelet.service
● ├─crio.service
● ├─machine-config-daemon-firstboot.service
● ├─rpc-statd.service

```

praveenkumar: Ah in that case I am OK

pkg/crc/cluster/cluster.go
```diff
@@ -236,7 +236,7 @@ func AddProxyToKubeletAndCriO(sshRunner *ssh.Runner, proxy *network.ProxyConfig)
 	proxyTemplate := `[Service]
 Environment=HTTP_PROXY=%s
 Environment=HTTPS_PROXY=%s
-Environment=NO_PROXY=.cluster.local,.svc,10.128.0.0/14,172.30.0.0/16,%s`
+Environment=NO_PROXY=.cluster.local,.svc,10.116.0.0/14,172.25.0.0/16,192.168.126.0/24,%s`
```
praveenkumar, line 239: Can you add in the commit log this is because of https://github.com/code-ready/snc/blob/master/install-config.yaml file, it took me some time to get that info.

pkg/crc/machine/start.go
```diff
@@ -541,7 +541,7 @@ func configProxyForCluster(ocConfig oc.Config, sshRunner *crcssh.Runner, sd *sys
 
 func waitForProxyPropagation(ocConfig oc.Config, proxyConfig *network.ProxyConfig) {
 	checkProxySettingsForOperator := func() error {
-		proxySet, err := cluster.CheckProxySettingsForOperator(ocConfig, proxyConfig, "redhat-operators", "openshift-marketplace")
+		proxySet, err := cluster.CheckProxySettingsForOperator(ocConfig, proxyConfig, "marketplace-operator", "openshift-marketplace")
```
praveenkumar, line 530: Is it changed recently, can you also add refer doc to commit message?

pkg/crc/machine/start.go
```diff
@@ -385,13 +385,13 @@ func (client *client) Start(startConfig StartConfig) (*StartResult, error) {
 
 	time.Sleep(time.Minute * 3)
 
+	waitForProxyPropagation(ocConfig, proxyConfig)
```
praveenkumar, line 387: 👍🏼

pkg/crc/cluster/cluster.go
guillaumerose: oh great! I didnt found it yesterday. I just compared what was injected automatically and what we hardcoded here.
Thanks.

pkg/crc/machine/start.go
guillaumerose: Apparently this deployment was always named like this.
https://github.com/operator-framework/operator-marketplace/blame/e3b79ea32bbaa898b3cc90e3bb2d64907d029ce6/manifests/07_operator.yaml#L4

Maybe redhat-operators was removed?

praveenkumar: It is not removed, here it is https://github.com/operator-framework/operator-marketplace/tree/master/defaults , I am not sure how it is consumed by operator-marketplace but it is always managed by it, will need to check with latest version of bundle.

PR #1826, opened 2021-01-08
Fix bundle embed

praveenkumar:


cmd/crc-embedder/cmd/embed.go
```diff
@@ -53,7 +54,7 @@ func runEmbed(args []string) {
 		logging.Fatalf("Failed to download data files: %v", err)
 	}
 
-	bundlePath := path.Join(bundleDir, constants.GetDefaultBundleForOs(goos))
+	bundlePath := path.Join(bundleDir, constants.GetDefaultBundleForOs(goos, version.GetBundleVersion()))
```
cfergeau, line 57: When I see 'default bundle', to me it means 'version of the bundle which we hardcoded at build time'. If you can override the version in a call to `GetDefaultBundleForOs`, it's no longer a default and the name has to be changed,  it's just `GetBundleNameForOs`

pkg/crc/constants/constants.go
```diff
@@ -71,18 +71,20 @@ func GetAdminHelperURL() string {
...
+		"linux":   fmt.Sprintf("crc_libvirt_%s.crcbundle", bundleVersion),
+		"windows": fmt.Sprintf("crc_hyperv_%s.crcbundle", bundleVersion),
+	}
 }
 
-func GetDefaultBundleForOs(os string) string {
-	return defaultBundleForOs[os]
+func GetDefaultBundleForOs(os, bundleVersion string) string {
+	return defaultBundleForOs(bundleVersion)[os]
 }
```
cfergeau, line 85: Could be 
```
var defaultBundleForOs = map[string]string{
	"darwin":  GetBundleName("darwin", version.GetBundleVersion()),
	"linux":   GetBundleName("linux", version.GetBundleVersion()),
	"windows": GetBundleName("windows", version.GetBundleVersion()),
}

func GetDefaultBundleForOs(os string) string {
	return defaultBundleForOs[os]
}

func GetBundleName(os, bundleVersion string) string {
    switch os {
        case "darwin": return fmt.Sprintf("crc_hyperkit_%s.crcbundle", bundleVersion)
        ....
    }
}
```

test/integration/crcsuite/crcsuite.go
```diff
@@ -90,7 +90,7 @@ func FeatureContext(s *godog.Suite) {
 			os.Exit(1)
 		}
 
-		if bundleLocation == "embedded" {
+		if bundleLocation == "" {
```
cfergeau, line 93: Nothing in the codebase is using --bundle-location="embedded"? To see a change in this commit removing this, or changing a default value, ...

pkg/crc/constants/constants.go
praveenkumar: @cfergeau how about then doing something like?

```
func bundleForOs(bundleVersion string) map[string]string {
	return map[string]string{
		"darwin":  fmt.Sprintf("crc_hyperkit_%s.crcbundle", bundleVersion),
		"linux":   fmt.Sprintf("crc_libvirt_%s.crcbundle", bundleVersion),
		"windows": fmt.Sprintf("crc_hyperv_%s.crcbundle", bundleVersion),
	}
}

func GetDefaultBundleForOs(os) string {
	return bundleForOs(version.GetBundleVersion())[os]
}

func GetBundleName(os, bundleVersion string) string {
    return bundleForOs(bundleVersion)[os]
}
```

test/integration/crcsuite/crcsuite.go
praveenkumar: I think it is only used by the internal jenkins job where QE test the embedded release, they pass the `--bundle-location` as `embedded` which I think not required.

pkg/crc/constants/constants.go
cfergeau: Yes, sure, as long as exported method names is good, we can implement it internally as we prefer :)
Dunno if we should do
```
func GetDefaultBundleForOs(os) string {
	return GetBundleName(os, version.GetBundleVersion())
}
```
for one-line functions, I guess not very important :)

test/integration/crcsuite/crcsuite.go
jsliacan: Yes, I think it is only used in the Makefile for the purposes of the CI or local testing with embedded bundle. This is a solution from the time when there was no way to distinguish the `crc` binary cases (embedded vs not) except by size :). As you say, it might not be needed now.

test/integration/crcsuite/prepare.go
```diff
@@ -120,7 +120,7 @@ func CopyFilesToTestDir() error {
 
 func ParseFlags() {
 
-	flag.StringVar(&bundleURL, "bundle-location", "embedded", "Path to the bundle to be used in tests")
+	flag.StringVar(&bundleLocation, "bundle-location", "embedded", "Path to the bundle to be used in tests")
```
jsliacan, line 123: If we change [L93 in `crcsuite.go`](https://github.com/code-ready/crc/pull/1826/files#diff-e4fa8ddc3d7d9a4a1c20d74b5492daac1c3e3051e5e3c2d33ee189fb9b60f80bR93), then the default value here should be changed from `"embedded"` to `""` as well.

praveenkumar: @jsliacan Good catch 👍🏼

test/integration/crcsuite/crcsuite.go
praveenkumar: @jsliacan Also moving forward we are not going to use `embedded` (no bundle embed for binary) because of signing issue so I think better to remove this completely.

PR #1831, opened 2021-01-11
Use test write key for non-release builds

guillaumerose:


pkg/crc/segment/segment.go
```diff
@@ -58,6 +58,10 @@ func (c *Client) Close() error {
 }
 
 func (c *Client) Upload(action string, duration time.Duration, err error) error {
+	if version.IsOkdBuild() {
+		return nil
```
praveenkumar, line 62: Are we sure about this (cc @sspeiche), Our motive is crc usage collection, doesn't matter if it is okd or ocp right?

guillaumerose: Some arguments:
* Community doesn't have access to Segment. They can't look at statistics. 
* The ID we used is also something Red Hat related. 
* I would expect community builds to be free from all sort of external services.

sspeiche: Agree we should try to keep out off community bits, though would be interesting to know how many OKD instances there are if we can...I think we just need to be transparent we are doing it and allow for opt-out.

praveenkumar: @sspeiche As of now by default it is opt out, it is upto the user to enable it or not.

guillaumerose: OK, I removed this from the PR. OKD will send events if turned on.

Makefile
```diff
@@ -172,6 +172,7 @@ gen_release_info:
 	@sed -i s/@OPENSHIFT_VERSION@/\"$(BUNDLE_VERSION)\"/ $(RELEASE_INFO)
 
 .PHONY: release
+release: LDFLAGS := $(LDFLAGS) -X $(REPOPATH)/pkg/crc/segment.WriteKey=cvpHsNcmGCJqVzf6YxrSnVlwFSAZaYtp
```
praveenkumar, line 175: 👍🏼

praveenkumar (approved):

PR #1832, opened 2021-01-11
Add OS version in Segment

guillaumerose:


gbraad (approved):
moving from the generic `GOOS` to an actual OS identifier including the version.

PR #1834, opened 2021-01-13
simplest macOS package

guillaumerose:
This is just crc + embedded files + bundle in /Library/crc/version/
The pkg is just a vehicle to ship multiple artifacts. `crc setup` still copy most of it in ~/.crc/

There is no impact for Windows and Linux.


pkg/embed/embed.go
```diff
@@ -4,13 +4,22 @@ import (
...
 )
 
-func openEmbeddedFile(executablePath, embedName string) (*binappend.Reader, error) {
+func openEmbeddedFile(executablePath, embedName string) (io.ReadCloser, error) {
+	if runtime.GOOS == "darwin" {
+		path := filepath.Join("/Library/crc", version.GetCRCVersion(), embedName)
+		if _, err := os.Stat(path); err == nil {
+			return os.Open(path)
+		}
+	}
```
cfergeau, line 21: This probably can be handled a bit higher in the stack, the only places where it matters seem to be `embed.Extract()` calls

PR #1835, opened 2021-01-13
Update admin-helper: it is not compressed twice anymore

guillaumerose:
Also move the root check first.


praveenkumar:
Can you elaborate on https://github.com/code-ready/crc/pull/1835/commits/54dd7ce4aca4236d58f936855a2f53dd9c55a83a one, like what is current scenario and after this patch what it would be?

guillaumerose:
I added more details. Basically, if I'm root, podman-remote, etc binaries will be created in /root/.crc/bin and then crc will refuse to continue with the root check.

praveenkumar:
> I added more details. Basically, if I'm root, podman-remote, etc binaries will be created in /root/.crc/bin and then crc will refuse to continue with the root check.

This is much better than you put in the commit log :)

PR #1837, opened 2021-01-14
Use only config object and don't recreate it for each api request

guillaumerose:
Follow-up of https://github.com/code-ready/crc/pull/1558

Since #1558 we recreate the config object each time. This is painful and implies to reconstruct the API handler for each request. 
This PR fixes that by only using viper for reading configuration. Writing is done manually. 

* Get() is now called every time.
* 2 processes can now share the same configuration without reloading

The intention of this is to simplify the API in a future PR.

### DIFF
--- a/pkg/crc/config/viper_config.go
+++ b/pkg/crc/config/viper_config.go
@@ -7,43 +7,67 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
 
 type ViperStorage struct {
-	viper      *viper.Viper
+	storeLock *sync.Mutex
+
+	// store flag values
+	flagSet *pflag.FlagSet
+
 	configFile string
+	envPrefix  string
 }
 
 func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
-	if err := ensureConfigFileExists(configFile); err != nil {
+	return &ViperStorage{
+		storeLock:  &sync.Mutex{},
+		configFile: configFile,
+		envPrefix:  envPrefix,
+	}, nil
+}
+
+func (c *ViperStorage) viperInstance() (*viper.Viper, error) {
+	if err := ensureConfigFileExists(c.configFile); err != nil {
 		return nil, err
 	}
 	v := viper.New()
-	v.SetConfigFile(configFile)
+	v.SetConfigFile(c.configFile)
 	v.SetConfigType("json")
-	v.SetEnvPrefix(envPrefix)
+	v.SetEnvPrefix(c.envPrefix)
 	// Replaces '-' in flags with '_' in env variables
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	v.AutomaticEnv()
 	v.SetTypeByDefaultValue(true)
-	err := v.ReadInConfig()
-	if err != nil {
-		return nil, fmt.Errorf("error reading configuration file '%s': %v", configFile, err)
+	if err := v.ReadInConfig(); err != nil {
+		return nil, fmt.Errorf("error reading configuration file '%s': %v", c.configFile, err)
 	}
-	return &ViperStorage{
-		viper:      v,
-		configFile: configFile,
-	}, nil
+	if c.flagSet == nil {
+		return v, nil
+	}
+	return v, v.BindPFlags(c.flagSet)
 }
 
 func (c *ViperStorage) Get(key string) interface{} {
-	return c.viper.Get(key)
+	c.storeLock.Lock()
+	defer c.storeLock.Unlock()
+	viperInstance, err := c.viperInstance()
+	if err != nil {
+		return nil
+	}
+	return viperInstance.Get(key)
 }
 
 func (c *ViperStorage) Set(key string, value interface{}) error {
+	c.storeLock.Lock()
+	defer c.storeLock.Unlock()
+	if err := ensureConfigFileExists(c.configFile); err != nil {
+		return err
+	}
 	in, err := ioutil.ReadFile(c.configFile)
 	if err != nil {
 		return err
@@ -57,13 +81,15 @@ func (c *ViperStorage) Set(key string, value interface{}) error {
 	if err != nil {
 		return err
 	}
-	if err := atomicWrite(bin, c.configFile); err != nil {
-		return err
-	}
-	return c.viper.ReadInConfig()
+	return atomicWrite(bin, c.configFile)
 }
 
 func (c *ViperStorage) Unset(key string) error {
+	c.storeLock.Lock()
+	defer c.storeLock.Unlock()
+	if err := ensureConfigFileExists(c.configFile); err != nil {
+		return err
+	}
 	in, err := ioutil.ReadFile(c.configFile)
 	if err != nil {
 		return err
@@ -77,15 +103,15 @@ func (c *ViperStorage) Unset(key string) error {
 	if err != nil {
 		return err
 	}
-	if err := atomicWrite(bin, c.configFile); err != nil {
-		return err
-	}
-	return c.viper.ReadInConfig()
+	return atomicWrite(bin, c.configFile)
 }
 
 // BindFlagset binds a flagset to their respective config properties
 func (c *ViperStorage) BindFlagSet(flagSet *pflag.FlagSet) error {
-	return c.viper.BindPFlags(flagSet)
+	c.storeLock.Lock()
+	defer c.storeLock.Unlock()
+	c.flagSet = flagSet
+	return nil
 }
 
 // ensureConfigFileExists creates the viper config file if it does not exists
--- a/cmd/crc/cmd/daemon.go
+++ b/cmd/crc/cmd/daemon.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/code-ready/crc/pkg/crc/api"
-	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/gvisor-tap-vsock/pkg/transport"
 	"github.com/code-ready/gvisor-tap-vsock/pkg/types"
@@ -142,15 +141,10 @@ func run(configuration *types.Configuration, endpoints []string) error {
 	}
 }
 
-func newConfig() (crcConfig.Storage, error) {
-	config, _, err := newViperConfig()
-	return config, err
-}
-
 func runDaemon() error {
 	// Remove if an old socket is present
 	os.Remove(constants.DaemonSocketPath)
-	apiServer, err := api.CreateServer(constants.DaemonSocketPath, newConfig, newMachineWithConfig)
+	apiServer, err := api.CreateServer(constants.DaemonSocketPath, config, newMachine())
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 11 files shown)


cmd/crc/cmd/daemon_nonwin.go
```diff
@@ -13,7 +13,7 @@ import (
 func runDaemon() {
 	// Remove if an old socket is present
 	os.Remove(constants.DaemonSocketPath)
-	crcAPIServer, err := api.CreateAPIServer(constants.DaemonSocketPath, newConfig, newMachineWithConfig)
+	crcAPIServer, err := api.CreateAPIServer(constants.DaemonSocketPath, config, newMachine())
```
cfergeau, line 16: Does this impact thread safety? The current code is already a bit scary imo, and I'm worried this adds more shared state between threads?

pkg/crc/config/viper_config.go
```diff
@@ -39,34 +39,59 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
 	}, nil
 }
 
-func (c *ViperStorage) Get(key string) interface{} {
-	return c.viper.Get(key)
+func (c *ViperStorage) Get(key string) (interface{}, error) {
+	if err := c.refresh(); err != nil {
+		return nil, err
+	}
```
cfergeau: disk IO + parsinng everytime we need a config value seems bad from an efficiency point of view?

pkg/crc/config/viper_config.go
```diff
@@ -39,34 +39,59 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
...
+	return c.viper.Get(key), nil
 }
 
 func (c *ViperStorage) Set(key string, value interface{}) error {
-	c.viper.Set(key, value)
-	return c.viper.WriteConfigAs(c.configFile)
+	in, err := ioutil.ReadFile(c.configFile)
```
cfergeau, line 48: Not clear what problems this is solving, maybe `Set` was misused, but if this has been causing issues, this is not obvious at all what issues.

pkg/crc/config/viper_config.go
```diff
@@ -33,17 +32,16 @@ func NewViperStorage(configFile, envPrefix string) (*ViperStorage, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error reading configuration file '%s': %v", configFile, err)
 	}
+	v.WatchConfig()
```
cfergeau, line 35: we had issues with `WatchConfig()` in the past, see 04832bdd3

pkg/crc/config/viper_config.go
```diff
@@ -85,13 +83,6 @@ func (c *ViperStorage) BindFlagSet(flagSet *pflag.FlagSet) error {
 	return c.viper.BindPFlags(flagSet)
 }
 
-func (c *ViperStorage) refresh() error {
```
cfergeau: This was added in the previous commits, maybe the commit adding this can be dropped?

guillaumerose: Yes, I kept it to show where I come from. I will squash this when it will be ready to merge.

pkg/crc/config/viper_config.go
guillaumerose: Yes this is what I try to solve! It works now.

pkg/crc/config/viper_config.go
guillaumerose: `Set` is not made for writing a new configuration but for overriding. You can reload 10 times the configuration, the override will stay. Viper is really made for reading configuration, not writing. Here I keep viper only for Get.

cmd/crc/cmd/daemon_nonwin.go
guillaumerose: config is not thread safe at all. Not better before or after this commit. I plan to submit this after: https://github.com/code-ready/crc/compare/master...guillaumerose:syncmachine?expand=1 which helps a bit.

At least now, one config and only one machineClient are created.

cfergeau: > At least now, one config and only one machineClient are created.

This is precisely my concern ;) Before, we were constantly recreating the config object, so each thread had its own config instance that no other thread would try to access. If all threads are now sharing the same object, this commit is making things worse.

pkg/crc/config/viper_config.go
cfergeau: Could probably be expanded a bit on in the commit log, and does not hurt to quote viper documentation
```

// Set sets the value for the key in the override register.
// Set is case-insensitive for a key.
// Will be used instead of values obtained via
// flags, config file, ENV, default, or key/value store.
func Set(key string, value interface{}) { v.Set(key, value) }
```

pkg/crc/config/viper_config.go
cfergeau: Details are needed in the commit log then ;) Mentioning the issue we used to have is very useful for reviewers, as well as explaining why it's no longer a problem (but after seeing again the old log, I guess the bug we had really was because we were misusing `Set()`)

pkg/crc/config/viper_config.go
cfergeau: I feel like a broken record, but mentioning that the `refresh()` is only temporary and should be squashed before merging is really useful info to know at review time :)

pkg/crc/config/viper_config.go
cfergeau: See https://github.com/code-ready/crc/pull/1837#discussion_r557477597 , only a temporary step towards a better fix.

guillaumerose:
Still looking at it. I am testing if removing viper is better.

cmd/crc/cmd/daemon_nonwin.go
guillaumerose: This is now better for config. For machine, it will come right after.

pkg/crc/config/viper_config.go
guillaumerose: It's now removed

pkg/crc/config/viper_config.go
guillaumerose: I removed the use of Watch.

guillaumerose:
This is ready for another round. I found it better without viper.

cfergeau (changes requested):
Having troubles with the "processes"  terminology in " Add unit test proving that 2 processes cannot share same configuration", I'd expect some processes to be forked/spawned/...?

The commit log in the koanf change seems to be outdated? A detailed description of why the change, the design decisions/tradeoffs, the behaviour changes (I see some of the test cases had to be changed), ... would be really useful for the review.

pkg/crc/config/combined_config.go
```diff
@@ -0,0 +1,135 @@
...
+	flagSet *pflag.FlagSet
+
+	// protect value stores
+	storeLock *sync.Mutex
+
+	configFile string
+}
+
+func NewCombinedStorage(configFile, envPrefix string) (*CombinedStorage, error) {
+	kBase := koanf.New(".")
```
gbraad, line 32: what is `kBase` supposed to mean? the `koanf` library's name by itself is not indicative enough of intent.

pkg/crc/config/combined_config.go
```diff
@@ -0,0 +1,135 @@
...
+	if err := kBase.Load(env.Provider(envPrefix, ".", func(s string) string {
+		return strings.ReplaceAll(strings.ToLower(strings.TrimPrefix(s, envPrefix)), "_", "")
+	}), nil); err != nil {
+		return nil, err
+	}
+	storage := &CombinedStorage{
+		kEnv:       kBase,
```
gbraad, line 39: `kEnv`? an environment of what; where the shell lives? people? keys? the unnecessary abbr(eviations) aren't useful.

pkg/crc/config/combined_config.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return err
+}
+
+func (c *CombinedStorage) merged() *koanf.Koanf {
+	if err := ensureConfigFileExists(c.configFile); err != nil {
+		logrus.Errorf("cannot create config file %s: %v", c.configFile, err)
+		return nil
+	}
+	combined := koanf.New(".")
```
gbraad, line 125: earlier it was a kBase, now a combined... I only know the Combine from Half-Life... are they the same?

gbraad: Please use more descriptive names.

pkg/crc/config/combined_config_test.go
```diff
@@ -17,7 +17,7 @@ const (
 )
 
 func newTestConfig(configFile, envPrefix string) (*Config, error) {
-	storage, err := NewViperStorage(configFile, envPrefix)
+	storage, err := NewCombinedStorage(configFile, envPrefix)
```
gbraad, line 20: A storage locker for Combine soldiers? or a CombinedStorage like in a fridge; both cold storage and freezer?

pkg/crc/config/combined_config_test.go
```diff
@@ -285,13 +285,13 @@ func TestTwoProcessesWriteSameConfiguration(t *testing.T) {
 
 	bin, err := ioutil.ReadFile(configFile)
 	assert.NoError(t, err)
-	assert.JSONEq(t, `{"nameservers":"1.1.1.1"}`, string(bin)) // cpus missing
+	assert.JSONEq(t, `{"cpus":5, "nameservers":"1.1.1.1"}`, string(bin))
```
gbraad, line 288: This 'TwoProcesses' test is supposed to test mutual exclusion or conflict prevention?
and why was it necessary to add the CPU here?

gbraad (changes requested):
Just reviewed: https://github.com/code-ready/crc/pull/1837/commits/81ed281b09bc2fd05779df2ebe0effc5e6c4156d and I would like see better naming for some of the classes and variables; especially around descripriveness and consistency. Some names are too generic and others are to abstract as they seem to be abbreviated.

gbraad:
> > Still looking at it. I am testing if removing viper is better.

> This is ready for another round. I found it better without viper.

Based on which conclusions?

guillaumerose:
I simplified the change and kept viper for the moment. The major benefit of this PR is that a single config object is used everywhere (and thus a single machine client).
Follow-up will include a way to deny/order machine client calls (https://github.com/code-ready/crc/compare/master...guillaumerose:syncmachine?expand=1) and then, I would like to have an http.Handler for machine client.
It is far more easier with the same handler all the time than if I need to recreate a handler for each request.

cfergeau (commented):
With this change, each config.Get() / config.Set() will require a mutex lock + reading/writing of the whole config file. It's not really obvious why such a heavyweight approach is required so details would be welcome. I understand the goal is thread safe Get/Set, but hopefully this can be achieved with just locks?

cfergeau:
> With this change, each config.Get() / config.Set() will require a mutex lock + reading/writing of the whole config file. It's not really obvious why such a heavyweight approach is required so details would be welcome. I understand the goal is thread safe Get/Set, but hopefully this can be achieved with just locks?

I've tried to make the commit log somehow more explicit, not sure it's fully accurate
```
Since commit 04832bdd3f and because of limitations in viper API [1], for each request to the API server, we create a new Config instance. This ensures that the config is up to date, even if modifications were made by external processes.
This use/recreation of short-lived config objects is suboptimal, and complicates the API server code.

To avoid this, this commit changes the ViperStorage implementation, and re-reads or rewrites the config file every time a Get/Set/Unset operation is done.
This also adds locking so that the config API can be accessed concurrently from multiple go routines, which is needed in the api server case.

[1] viper.Set() not doing exactly what the name indicates, viper.WatchConfig() being racy)
```
This tries to add a description of what the initial problem is, and what it's doing. After doing this though, I wonder if the locking itself should be split, and if we could achieve something similar by rewriting viper.WatchConfig without the concurrency issues.

gbraad (approved):

PR #1841, opened 2021-01-15
Check for race conditions when running go test

guillaumerose:
`viper.WatchConfig` is full of race and having this race check will prevent us (and esp. me in https://github.com/code-ready/crc/pull/1837) to add it again.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ environment:
   GOPATH: c:\gopath
 stack: go 1.14
 before_test:
+  - choco install mingw
   - choco install make
   - make cross
 test_script:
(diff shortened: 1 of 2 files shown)


gbraad (approved):

appveyor.yml
```diff
@@ -4,6 +4,7 @@ environment:
   GOPATH: c:\gopath
 stack: go 1.14
 before_test:
+  - choco install mingw
```
gbraad, line 7: is this a necessity for `-race` ?

guillaumerose: yes, it needs gcc :/

guillaumerose: https://ci.appveyor.com/project/code-ready/crc/builds/37279990

PR #1842, opened 2021-01-15
Some improvements for the macos package

cfergeau:
My goal was to install in `/Library/CodeReady Containers` rather than `/Library/crc`, cleaned up a few build-related things while doing that.
Only user-visible change before/after this commit is the directory the macos package installs to, which is not even that visible as the `crc` binary is symlinked in `/usr/local/bin`


Makefile
```diff
@@ -118,12 +118,18 @@ docs_serve: build_docs
 docs_check_links:
 	${CONTAINER_RUNTIME} run -it -v $(CURDIR)/docs:/docs:Z --rm $(DOCS_BUILD_CONTAINER) docs_check_links
 
-.PHONY: clean_docs
+.PHONY: clean_docs clean_macos_package
 clean_docs:
 	rm -rf $(CURDIR)/docs/build
 
+clean_macos_package:
```
praveenkumar, line 125: Shouldn't we put `rm -fr packaging`

cfergeau: `packaging/darwin` has some source files managed in git

guillaumerose (approved):

PR #1848, opened 2021-01-18
Update to latest clicumber in go mod

praveenkumar:

### DIFF
--- a/test/integration/features/story_marketplace.feature
+++ b/test/integration/features/story_marketplace.feature
@@ -22,13 +22,13 @@ Feature: Operator from marketplace
         When checking that CRC is running
         Then login to the oc cluster succeeds
 
-    @darwin @linux @windows
+    @darwin @linux @windows @testdata
     Scenario: Install new operator
         When executing "oc apply -f etcdop-sub.yaml" succeeds
         # check if cluster operator is running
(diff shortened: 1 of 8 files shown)


test/integration/features/story_marketplace.feature
```diff
@@ -22,13 +22,13 @@ Feature: Operator from marketplace
         When checking that CRC is running
         Then login to the oc cluster succeeds
 
-    @darwin @linux @windows
+    @darwin @linux @windows @testdata
```
jsliacan, line 25: Would only 1 tag per Feature be required? The 1st one. Scenarios can't be run concurrently anyways. And once the data is copied, it's present at least until end of Feature. 

Ideally, only the first instance in the Suite would be enough, but in downstream CI, Feature files are run one by one (each in its own suite). So once per Feature is good :).

praveenkumar: > Would only 1 tag per Feature be enough? The 1st one. Scenarios can't be run concurrently anyways. And once the data is copied, it's present at least until end of Feature.

I am trying to avoid all the Feature related stuff so we can start upgrading the godog which doesn't have feature specific helper functions but only for Scenario specific.

jsliacan: Yep, I understand that. And the tags are a good idea. But the scenarios are not independent anyways, so no need to copy files again after they were already copied once.

praveenkumar:
/close 

Deleted fork accidentally :(

PR #1849, opened 2021-01-18
Align license title and remove appendix like openshift/origin

guillaumerose:
In the installer, we display the license and the title was not aligned.

I re-used https://github.com/openshift/origin/blob/master/LICENSE

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
-Apache License
+
+                                 Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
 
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -175,18 +176,7 @@ Apache License
...
-      boilerplate notice, with the fields enclosed by brackets "{}"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
-      comment syntax for the file format. We also recommend that a
-      file or class name and description of purpose be included on the
-      same "printed page" as the copyright notice for easier
-      identification within third-party archives.
-
-   Copyright {yyyy} {name of copyright owner}
+   Copyright 2019 Red Hat, Inc.
```
cfergeau, line 179: 2021?

cfergeau (approved):
> In the installer, we display the license and the title was not aligned.
> 
> I re-used https://github.com/openshift/origin/blob/master/LICENSE

Can you add this to the commit log, rather than only having it in the PR description? Slightly better to say "macOS installer, as I was initially thinking of the openshift installer, and was totally confused :)

LICENSE
guillaumerose: yes! I never know if it's the starting date or end date. I changed to 2021, Thanks.

cfergeau: I never know either in doubt, put both!

cfergeau: A quick look seems to say it's the starting date which matters ( https://softwareengineering.stackexchange.com/questions/130478/what-copyright-date-for-an-update-to-an-open-source-project-from-last-year https://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/ )

cfergeau (approved):

PR #1851, opened 2021-01-18
Make behaviour releases installed from macos package closer to the previous embedded behaviour

cfergeau:
This readds 2 things when using a binary built with 'make package':
- `crc setup` no longer instructs to use `-b $bundlename` since the bundle can be located automatically
- `crc start -b $bundlename` requires the openshift version which was set at build time

### DIFF
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -7,9 +7,9 @@ import (
 	"os"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
-	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/input"
 	"github.com/code-ready/crc/pkg/crc/preflight"
+	pkgversion "github.com/code-ready/crc/pkg/crc/version"
 	"github.com/spf13/cobra"
 )
 
@@ -68,7 +68,7 @@ func (s *setupResult) prettyPrintTo(writer io.Writer) error {
 
 func extraArguments() string {
 	var bundle string
-	if !constants.BundleEmbedded() {
+	if !pkgversion.IsRelease() {
 		bundle = " -b $bundlename"
 	}
 	return bundle
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -10,6 +10,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 	"github.com/code-ready/crc/pkg/crc/validation"
+	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/code-ready/crc/pkg/embed"
 	"github.com/docker/go-units"
 	"github.com/pkg/errors"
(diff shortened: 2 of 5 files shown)


cmd/crc/cmd/setup.go
```diff
@@ -68,7 +68,7 @@ func (s *setupResult) prettyPrintTo(writer io.Writer) error {
 
 func extraArguments() string {
 	var bundle string
-	if !constants.BundleEmbedded() {
+	if !pkgversion.IsRelease() {
```
guillaumerose, line 71: I am not really convince by this flag. Ideally, we should always build release binaries. I am worry that we will diverge even more over time and the CI will never test release binaries. 

Would it be possible to remove those ifs instead ? Should we invert and say `IsDev()` ?

cfergeau: this `extraArgument()` change could be replaced by a check that `GetDefaultBundle()` is a valid path, but I think I would just remove the `-b ...` mention, and just indicate to use `crc start`. Then `crc start` can hint that `-b ...` should be used when it fails to find the default bundle.

guillaumerose: Yes no really good alternative. Let's continue with this.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -76,7 +77,7 @@ func fixBundleExtracted() error {
 	if err := os.Chmod(constants.MachineCacheDir, 0775); err != nil {
 		logging.Debugf("Error changing %s permissions to 0775", constants.MachineCacheDir)
 	}
-	if constants.BundleEmbedded() {
+	if version.IsRelease() {
```
guillaumerose, line 80: I think this line will add an extra copy for the pkg. 
The crcbundle is already in /Library and will be now copied in ~/.crc/cache before extraction.

praveenkumar: We can only start extracting the bundle if it is embedded so may be with release, we should also use embedded logic?
```
if version.IsRelease() && constants.BundleEmbedded{
}
```

cfergeau: This indeed looks like this needs more work... I think something like this is more correct though:
```
if version.IsRelease() {
		bundleDir := filepath.Dir(constants.DefaultBundlePath)
		if err := os.MkdirAll(bundleDir, 0775); err != nil {
			return fmt.Errorf("Cannot create directory %s: %v", bundleDir, err)
		}
		if constants.BundleEmbedded {
			if err := embed.Extract(filepath.Base(constants.DefaultBundlePath), constants.DefaultBundlePath); err != nil {
				return err
			}
		}
		_, err := bundle.Extract(constants.DefaultBundlePath)
		return err
	}
```

But this makes me wonder why we have this code in the first place https://github.com/code-ready/crc/blob/de1d9f11eea9c10e822aa2dbbf040aaf722ae6de/pkg/embed/embed.go#L16-L21 if we never want to do an extra copy of the bundle.

guillaumerose: Yeah.. the copy of small binaries was less noticeable for the end user. It should be better to override binDir totally.

cfergeau: Actually with the current code, `fixBundleExtracted` will never be called in the macos package case. 
```
	if !version.IsRelease() {
		return nil
	}
	if _, err := os.Stat(constants.DefaultBundlePath); os.IsNotExist(err) {
		return err
	}
	return nil
}
```

`constants.DefaultBundlePath` points at `/Library/CodeReady  Containers/....` which will always exist after package install.

cmd/crc/cmd/setup.go
cfergeau: Thinking more about `IsRelease`, it's probably equivalent to `version.IsMacosInstallPathSet() || constants.BundleEmbedded()`

PR #1855, opened 2021-01-19
Rename integration tests to e2e tests

jsliacan:
Renaming integration to e2e will create space for actual integration tests. The Cucumber tests will be simplified to adhere to design principles of the framework and they'll serve the purpose of e2e tests. The more detailed testing will be moved to integration tests **not** written in Cucumber. 

There's a PR on the Clicumber side doing the same: https://github.com/code-ready/clicumber/pull/13. Including renaming logs from `integration_*.log` to `e2e_*.log`. 

## Testing

1. `make e2e` should behave as `make integration` did previously (including renamed dirs and files). All tests should run correctly. 

I tested this on Linux only so far.


jsliacan:
Will need to change this line as well after this goes in https://github.com/openshift/release/blob/431aac76304a21b1c290811441d8b12b9252450c/ci-operator/step-registry/openshift/e2e/gcp/crc/test/openshift-e2e-gcp-crc-test-commands.sh#L86

developing.adoc
```diff
@@ -46,40 +46,40 @@ For more information, see the following:
 . link:https://github.com/golang/go/wiki/Modules[Introduction to Go modules]
 . link:https://blog.golang.org/using-go-modules[Using Go modules]
 
-[[running-integration-tests]]
-== Running integration tests
+[[running-e2e-tests]]
+== Running e2e tests
 
-We have automated integration tests which keep CodeReady Containers in shape.
+We have automated e2e tests which keep CodeReady Containers in shape.
```
cfergeau, line 52: In the documentation, I'm not so sure about using 'e2e' as is. Maybe explicitly spell 'end-to-end (e2e)', at least the first few times it appears?

developing.adoc
```diff
@@ -46,40 +46,40 @@ For more information, see the following:
...
 
-[[intro-to-integration-testing]]
+[[intro-to-e2e-testing]]
 === Introduction
-Integration tests use the link:http://github.com/code-ready/clicumber[Clicumber] package to provide basic functionality for testing CLI binaries.
+E2e tests use the link:http://github.com/code-ready/clicumber[Clicumber] package to provide basic functionality for testing CLI binaries.
```
cfergeau, line 56: Same comment about the e2e acronym. And it looks odd written `E2e` rather than `e2e`

cfergeau (commented):
Can you add the PR rationale in the commit log? Looks good otherwise apart from a docs comment.

developing.adoc
jsliacan: I don't want to replace it completely with 'end-to-end', as in code we use `e2e`. But I did the "definition" as you said: spelled it out the first time with bracketed "from now on" form. Looks more natural now!

cfergeau:
`ci/prow/e2e-crc` will need an update after this change? I see this in the logs:
```
 make: *** No rule to make target 'integration'.  Stop. 
```

cfergeau (approved):

jsliacan:
@cfergeau yes, I sent a [PR](https://github.com/openshift/release/pull/15458) there as well. Thanks

cfergeau (approved):
Let's see how this goes :)

guillaumerose:
Is it better to keep integration target as an alias of e2e for couple of days?

cfergeau:
> Is it better to keep integration target as an alias of e2e for couple of days?

Ah yes, great suggestion, should ease the transition :)

cfergeau:
> > Is it better to keep integration target as an alias of e2e for couple of days?
> 
> Ah yes, great suggestion, should ease the transition :) Let's do that if [openshift/release#15458](https://github.com/openshift/release/pull/15458) is not merged in the next hours.

It's been merged, hopefully we won't need the `integration` alias.

PR #1857, opened 2021-01-20
Use public registry for ubi

cfergeau:
registry.redhat.io requires authentication, registry.access.redhat.com
does not as described on https://access.redhat.com/RegistryAuthentication

Since both have the image that we need, we can use the public one to
make builds easier.

### DIFF
--- a/images/dnsmasq/Dockerfile
+++ b/images/dnsmasq/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/ubi8/ubi
+FROM registry.access.redhat.com/ubi8/ubi
 MAINTAINER CodeReady Container <devtools-cdk@redhat.com>
 
 RUN yum -y install dnsmasq && \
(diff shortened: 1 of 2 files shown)


guillaumerose (approved):

praveenkumar:
@cfergeau This registry is deprecated and stay here only for legacy support https://access.redhat.com/RegistryAuthentication#comment-1991831 I am OK as as long as this provides us latest content.

cfergeau:
> @cfergeau This registry is deprecated and stay here only for legacy support https://access.redhat.com/RegistryAuthentication#comment-1991831 I am OK as as long as this provides us latest content.

These are all private comments which you can't see without proper authentication, however this is documented here as well https://www.redhat.com/en/blog/transitioning-red-hat-container-registry
I don't think for now retirement plans were announced? I have no idea how to add credentials for registry.redhat.io to quay.io  builders, which is why I went this way instead.

PR #1861, opened 2021-01-20
ubuntu: tell the user to logout, login and run crc setup

guillaumerose:
It is not obvious that the setup is not done and needs to be triggered again.


gbraad (approved):

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -19,7 +19,7 @@ var nmPreflightChecks = [...]Check{
 		configKeySuffix:  "check-systemd-networkd-running",
 		checkDescription: "Checking if systemd-networkd is running",
 		check:            checkSystemdNetworkdIsNotRunning,
-		fixDescription:   "network configuration with systemd-networkd is not supported",
+		fixDescription:   "Network configuration with systemd-networkd is not supported. Perhaps you can try this new network mode: https://github.com/code-ready/crc/wiki/VPN-support--with-an--userland-network-stack",
```
cfergeau, line 22: Imo it's a bit too early for this :)

PR #1863, opened 2021-01-21
Misc changes around command and segment client

praveenkumar:


pkg/crc/segment/segment.go
```diff
@@ -70,7 +79,7 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error
 	properties := analytics.NewProperties().
 		Set("version", version.GetCRCVersion()).
 		Set("success", err == nil).
-		Set("duration", duration.Milliseconds())
+		Set("duration", fmt.Sprintf("%.5fsec", duration.Seconds()))
```
guillaumerose, line 82: There is a type change here. It was int and now a string. I am not sure Woopra can automatically detect that.

Also, if we use seconds, most of the time we will see 0 as duration (or 1, 2) for short commands. It will be difficult to have good statistics. I am sure we can arrange Woopra and consumers to display duration/1000 if needed.

praveenkumar: @guillaumerose Ack, let's see if we can make modification on woopra side, removing this commit.

PR #1865, opened 2021-01-21
Misc changes around segment client

praveenkumar:
**Fixes:** Issue #N

## Solution/Idea

## Proposed changes

1. Add `start` command.
3. ...

## Testing

7. ...

### DIFF
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -121,7 +121,7 @@ func checkVsock() error {
 	if err != nil {
 		return err
 	}
-	if !strings.Contains(string(getcap), "cap_net_bind_service+eip") {
+	if !strings.Contains(getcap, "cap_net_bind_service+eip") {
 		return fmt.Errorf("capabilities are not correct for %s", executable)
 	}
 	info, err := os.Stat("/dev/vsock")
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -2,6 +2,7 @@ package segment
 
 import (
 	"io/ioutil"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
@@ -34,6 +35,9 @@ func newCustomClient(config *crcConfig.Config, telemetryFilePath, segmentEndpoin
 	client, err := analytics.NewWithConfig(WriteKey, analytics.Config{
 		Endpoint: segmentEndpoint,
 		Logger:   &loggingAdapter{},
+		DefaultContext: &analytics.Context{
+			IP: net.IPv4(0, 0, 0, 0),
+		},
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


pkg/os/exec.go
```diff
@@ -11,6 +11,7 @@ import (
 
 func runCmd(command string, args []string, env map[string]string) (string, string, error) {
 	cmd := exec.Command(command, args...) // #nosec G204
+	logging.Debugf("Running '%s %s'", command, strings.Join(args, " "))
```
cfergeau: nack, `runPrivate` also calls `runCmd`, and we do not want to print the command when `runPrivate` is used.

praveenkumar: Good catch, will try to change it like we do for oc, send the private flag to run command.

cfergeau (commented):
`Your preference will be part of 'crc config set consent-telemetry <yes/no>'`

I find this odd to read. Maybe `Your preference can be changed manually if desired using ``crc config set consent-telemetry <yes/no>`` ` ?

pkg/crc/segment/segment.go
```diff
@@ -34,6 +35,13 @@ func newCustomClient(config *crcConfig.Config, telemetryFilePath, segmentEndpoin
 	client, err := analytics.NewWithConfig(WriteKey, analytics.Config{
 		Endpoint: segmentEndpoint,
 		Logger:   &loggingAdapter{},
+		DefaultContext: &analytics.Context{
+			IP: net.IPv4(0, 0, 0, 0),
+			Library: analytics.LibraryInfo{
```
guillaumerose, line 40: You don't need to set the library. It is done automatically. See https://github.com/segmentio/analytics-go/blob/v3.0/config.go#L162

praveenkumar: Ah yes, I did check that code but I thought it override whatever in DefaultContext but it just add the `Library` part.

pkg/crc/segment/segment.go
```diff
@@ -34,6 +35,9 @@ func newCustomClient(config *crcConfig.Config, telemetryFilePath, segmentEndpoin
 	client, err := analytics.NewWithConfig(WriteKey, analytics.Config{
 		Endpoint: segmentEndpoint,
 		Logger:   &loggingAdapter{},
+		DefaultContext: &analytics.Context{
+			IP: net.IPv4(0, 0, 0, 0),
```
anjannath, line 39: 👍🏽

pkg/crc/preflight/preflight_linux.go
```diff
@@ -121,7 +121,7 @@ func checkVsock() error {
 	if err != nil {
 		return err
 	}
-	if !strings.Contains(string(getcap), "cap_net_bind_service+eip") {
+	if !strings.Contains(getcap, "cap_net_bind_service+eip") {
```
anjannath, line 124: 👍🏽

anjannath (approved):
LGTM

PR #1867, opened 2021-01-21
Update to latest clicumber in go mod

praveenkumar:


test/integration/crcsuite/crcsuite.go
```diff
@@ -137,9 +137,13 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeScenario(func(pickle *messages.Pickle) {
 		// copy data/config files to test dir
-		err := CopyFilesToTestDir()
-		if err != nil {
-			os.Exit(1)
+		for _, tag := range pickle.GetTags() {
```
cfergeau, line 141: Is this a 'valid' use of godog tags? (change test behaviour depending on the tag, as opposed as only a way of filtering which tests are run)

test/integration/crcsuite/crcsuite.go
```diff
@@ -80,10 +80,11 @@ func FeatureContext(s *godog.Suite) {
 			usr, _ := user.Current()
 			CRCExecutable = filepath.Join(usr.HomeDir, "go", "bin")
 		}
+		OCExecutable := filepath.Join(CRCHome, "bin", "oc")
```
cfergeau, line 83: This does not need to be exported, so `ocExecutable` should work too. However this commit makes me wonder if the clicumber changes were not too drastic, as now a clicumber test case can no longer change environment variables, which was the case before. Wondering if there could be other side effects.

test/integration/features/story_marketplace.feature
```diff
@@ -41,7 +41,6 @@ Feature: Operator from marketplace
     Scenario: Failover
         # simulate failure of 1 pod, check that it was replaced
         When executing "POD=$(oc get pod -o jsonpath="{.items[0].metadata.name}")" succeeds
-        And executing "echo $POD" succeeds
         And executing "oc delete pod $POD --now" succeeds
```
cfergeau, line 44: Is 
```
        When executing "POD=$(oc get pod -o jsonpath="{.items[0].metadata.name}")" succeeds
        And executing "oc delete pod $POD --now" succeeds
```
still working as expected with the new clicumber ?

praveenkumar:
/close 

Current clicumber update doesn't work as user expected behavior, which again need more update and testing.

PR #1869, opened 2021-01-21
Preflight checks improvements around udev/vsock code

cfergeau:
Initially, the goal was to fix some corner cases when adding the udev rule for the vsock device.
I added a bit too many cleanups before/after it, so the PR is bigger than I would have liked :)

### DIFF
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -101,7 +101,7 @@ var vsockPreflightChecks = Check{
 	configKeySuffix:    "check-vsock",
 	checkDescription:   "Checking if vsock is correctly configured",
 	check:              checkVsock,
-	fixDescription:     "Checking if vsock is correctly configured",
+	fixDescription:     "Setting up vsock support",
 	fix:                fixVsock,
 	cleanupDescription: "Removing vsock configuration",
 	cleanup:            removeVsockCrcSettings,
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -158,28 +158,33 @@ func fixVsock() error {
 		return err
 	}
 	udevRule := `KERNEL=="vsock", MODE="0660", OWNER="root", GROUP="libvirt"`
-	err = crcos.WriteToFileAsRoot("Create udev rule for /dev/vsock", udevRule, vsockUdevLocalAdminRulesPath, 0644)
-	if err != nil {
-		return err
+	if crcos.FileContentMatches(vsockUdevLocalAdminRulesPath, []byte(udevRule)) != nil {
+		err = crcos.WriteToFileAsRoot("Create udev rule for /dev/vsock", udevRule, vsockUdevLocalAdminRulesPath, 0644)
+		if err != nil {
+			return err
+		}
+		_, _, err = crcos.RunWithPrivilege("reloading udev rules database", "udevadm", "control", "--reload")
+		if err != nil {
+			return err
+		}
 	}
-	_, _, err = crcos.RunWithPrivilege("reloading udev rules database", "udevadm", "control", "--reload")
-	if err != nil {
-		return err
-	}
-	if crcos.FileExists("/dev/vsock") {
+	if crcos.FileExists("/dev/vsock") && unix.Access("/dev/vsock", unix.R_OK|unix.W_OK) != nil {
 		_, _, err = crcos.RunWithPrivilege("applying udev rule to /dev/vsock", "udevadm", "trigger", "/dev/vsock")
+		if err != nil {
+			return err
+		}
+	} else {
+		_, _, err = crcos.RunWithPrivilege("modprobe vhost_vsock", "modprobe", "vhost_vsock")
 		if err != nil {
 			return err
 		}
 	}
 
-	err = crcos.WriteToFileAsRoot(fmt.Sprintf("Create file %s", vsockModuleAutoLoadConfPath), "vhost_vsock", vsockModuleAutoLoadConfPath, 0644)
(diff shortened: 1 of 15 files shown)


anjannath (approved):
LGTM

pkg/crc/preflight/preflight_linux.go
```diff
@@ -173,7 +186,8 @@ func fixVsock() error {
 
 func removeVsockCrcSettings() error {
 	var mErr crcErrors.MultiError
-	err := crcos.RemoveFileAsRoot(fmt.Sprintf("removing udev rule in %s", vsockUdevRulesPath), vsockUdevRulesPath)
+	_ = crcos.RemoveFileAsRoot(fmt.Sprintf("removing udev rule in %s", vsockUdevSystemRulesPath), vsockUdevSystemRulesPath)
+	err := crcos.RemoveFileAsRoot(fmt.Sprintf("removing udev rule in %s", vsockUdevLocalAdminRulesPath), vsockUdevLocalAdminRulesPath)
```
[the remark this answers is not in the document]
cfergeau: Ah I was assuming `crcos.RemoveFileAsRoot("/nonexistentfile")` would print an error message, which I don't want. If that's not the case, no need to ignore the error.

pkg/os/util_nonwin.go
```diff
@@ -30,6 +30,9 @@ func WriteToFileAsRoot(reason, content, filepath string, mode os.FileMode) error
 }
 
 func RemoveFileAsRoot(reason, filepath string) error {
+	if !FileExists(filepath) {
```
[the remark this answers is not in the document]
cfergeau: Hmm yeah, the weird behaviour of `FileExists` is problematic here. I'll drop this change.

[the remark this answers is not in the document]
cfergeau: I changed `FileExists` instead, and kept the change in `RemoveFileAsRoot`

pkg/os/util.go
```diff
@@ -92,9 +92,9 @@ func WriteFileIfContentChanged(path string, newContent []byte, perm os.FileMode)
...
 
 func FileExists(path string) bool {
-	_, err := os.Stat(path)
-	if err != nil && os.IsNotExist(err) {
+	info, err := os.Lstat(path)
+	if err != nil {
 		return false
 	}
-	return true
+	return !info.IsDir()
```
praveenkumar, line 99: @cfergeau Even this change doesn't provide the correct behavior. (Trying with something which a user I don't have access to the folder and existing implementation did provide correct result.)

```
$ ls /var/lib/libvirt/qemu/
ls: cannot open directory '/var/lib/libvirt/qemu/': Permission denied

$ sudo ls /var/lib/libvirt/qemu/
channel  checkpoint  domain-2-crc-l44pc-master-0  domain-48-test1-44tjj-master-0  dump	nvram  ram  save  snapshot

$ cat fileexist.go
package main

import (
	"fmt"
	"os"
)

func FileExistsOld(path string) bool {
	_, err := os.Stat(path)
	if err != nil && os.IsNotExist(err) {
		return false
	}
	return true
}

func FileExistsNew(path string) bool {
	info, err := os.Lstat(path)
	if err != nil {
		return false
	}
	return !info.IsDir()
}

func main() {
	fmt.Println("New change in FileExist: ", FileExistsNew("/var/lib/libvirt/qemu/domain-48-test1-44tjj-master-0"))
	fmt.Println("Existing in FileExist: ", FileExistsOld("/var/lib/libvirt/qemu/domain-48-test1-44tjj-master-0"))
}

$ go run exist.go 
New change in FileExist:  false
Existing in FileExist:  true
```

cfergeau: Since you have this:
```
$ ls /var/lib/libvirt/qemu/
ls: cannot open directory '/var/lib/libvirt/qemu/': Permission denied
```
the correct answer to `FileExists("/var/lib/libvirt/qemu/...")` is "I don't know" (in other words, there was an error, so I can't tell if the file exists or not). The old implementation will always return `true` when there is an error, regardless of the actual file existence. In your test case, `FileExistsOld("/var/lib/libvirt/qemu/domain-48-test1-44tjj-master-0")` returns true, but `FileExistsOld("/var/lib/libvirt/qemu/I know there is no such file")` woul also be true. The new implementation will always return `false" in such cases, which I believe is more correct given how the rest of the code uses this method.

Old implementation can have false positives (returns true when the file does not exist), new implementation can have false negatives (returns false when the file actually exists).
The correct implementation would be `FileExists(string) (bool, error)`, but this makes it less convenient to use:
```
exists, err := FileExists(path)
if err != nil {
   // do something which can be exists = false
}
if exists {
  // do something
}
``
as opposed to
```
if FileExists(path) {
  // do something
}
```

I preferred to keep the convenient API here.

Comment (praveenkumar) [reply pkg/os/util.go:99]: @cfergeau Can you provide a sample example with this false positive, I tired it with file which I don't have permission to list and vice versa, I didn't get any false positive, may be I am missing some test case here?

Comment (cfergeau) [reply pkg/os/util.go:99]: In your previous test case, I'd expect this to return `true` even if `this is a random file` does not exist:
```
fmt.Println("Existing in FileExist: ", FileExistsOld("/var/lib/libvirt/qemu/this is a random file"))
```

Actually, I put a test case for this:

```
	err = os.Chmod(dirname, 0000)
	assert.NoError(t, err)
	assert.False(t, FileExists(filename))
	filename = filepath.Join(dirname, "nonexistent")
	assert.False(t, FileExists(filename))
```

you can try `go test ./pkg/os/` with the new/old implementation, and see the differences. With the old implementation, the 2 `assert.False(t, FileExists(...))` above will fail (ie `FileExists` returns `true` in these cases)

PR #1870, opened 2021-01-21
Remove OcBinDir if it's a file

guillaumerose:
Apparently, we removed this code too quickly. Some users still need it.
They hit the error: mkdir /Users/name/.crc/bin/oc: not a directory

Found with telemetry.


pkg/crc/machine/bundle/repository.go
```diff
@@ -83,6 +83,11 @@ func (repo *Repository) Use(bundleName string) (*CrcBundleInfo, error) {
 func (bundle *CrcBundleInfo) createSymlinkOrCopyOpenShiftClient(ocBinDir string) error {
 	ocInBundle := bundle.resolvePath(constants.OcExecutableName)
 	ocInBinDir := filepath.Join(ocBinDir, constants.OcExecutableName)
+	if info, err := os.Stat(ocBinDir); err == nil && !info.IsDir() {
```
cfergeau, line 86: Can you add a comment which says this is needed when upgrading from crc versions anterior to commit `1.11.0~5` ?

cfergeau (approved):

pkg/crc/machine/bundle/repository.go
guillaumerose: done

Issue #1876, opened 2021-01-22
Add crc setup in the error string if crc start fails during preflight


praveenkumar:
We should tell user to run `crc setup` as part of error message or separate one in case `crc start` fails during preflight check.

praveenkumar:
Resolved by #1877

PR #1877, opened 2021-01-22
Add crc setup instruction if start fail in preflight

praveenkumar:
**Fixes:** Issue #1876 

## Solution/Idea
In case of error during preflight, better to advice the user to
run crc setup once if user directly started executed `crc start`.

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -129,6 +129,10 @@ type startResult struct {
 
 func (s *startResult) prettyPrintTo(writer io.Writer) error {
 	if s.Error != nil {
+		var e *crcErrors.PreflightError
+		if errors.As(s.Error, &e) {
+			logging.Warn("Preflight checks failed during `crc start`, please try to run `crc setup` first in case you haven't done so yet")
+		}
 		return s.Error
 	}
 	if s.ClusterConfig == nil {
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -16,6 +16,18 @@ func (v vmNotExist) Error() string {
 
 const VMNotExist vmNotExist = "Machine does not exist. Use 'crc start' to create it"
 
+type PreflightError struct {
+	Err error
+}
+
+func (p *PreflightError) Error() string {
+	return p.Err.Error()
+}
+
+func (p *PreflightError) Unwrap() error {
+	return p.Err
+}
+
 type MultiError struct {
 	Errors []error
 }
--- a/pkg/crc/preflight/preflight.go
+++ b/pkg/crc/preflight/preflight.go
@@ -153,7 +153,10 @@ func doRegisterSettings(cfg config.Schema, checks []Check) {
 func StartPreflightChecks(config config.Storage) error {
 	experimentalFeatures := config.Get(cmdConfig.ExperimentalFeatures).AsBool()
 	mode := network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString())
-	return doPreflightChecks(config, getPreflightChecks(experimentalFeatures, mode))
+	if err := doPreflightChecks(config, getPreflightChecks(experimentalFeatures, mode)); err != nil {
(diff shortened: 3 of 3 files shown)


cmd/crc/cmd/start.go
```diff
@@ -60,7 +60,7 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 	checkIfNewVersionAvailable(config.Get(cmdConfig.DisableUpdateCheck).AsBool())
 
 	if err := preflight.StartPreflightChecks(config); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w\nplease try 'crc setup' first", err)
```
cfergeau, line 63: Would be best to return a specific error type/an error which can be matched against IsPreflightFailure(err) so that the upper layers can print this correctly
```
if IsPreflightFailure(err) {
    logging.Infof("Preflight checks failed during `crc start`,  please try to run `crc setup` first in case you haven't done so yet")
}
```
(main advantage is that this will get a nice INFO prefix, and won't look like it's part of thenot necessarily user-friendly error message)

praveenkumar: Here the upper layer is viper RunE method which just put the error in case any. Right now `if err := preflight.StartPreflightChecks(config)` which means the error is from the PreflightChecks. Do you want to modify the `StartPreflightCheck` signature which return the preflight failure specific error and here we just detect it ?

cfergeau: What I had in mind was (probably) to make changes in `cmd/crc/cmd/start.go.prettyPrintTo()` using 
```
var PreflightError
if errors.As(err, &e) {
     fmt.Printf("Preflight checks failed during `crc start`,  please try to run `crc setup` first in case you haven't done so yet")
   
} else {
    return ...
}
```
Sticking generic user-level instructions in an error because we know crc code prints errors using `fmt.Printf` at the end of execution seems like something which could get broken anytime, especially with no corresponding integration test.

cmd/crc/cmd/start.go
```diff
@@ -129,6 +129,10 @@ type startResult struct {
 
 func (s *startResult) prettyPrintTo(writer io.Writer) error {
 	if s.Error != nil {
+		var e *crcErrors.PreflightError
+		if errors.As(s.Error, &e) {
+			logging.Warn("Preflight checks failed during `crc start`,  please try to run `crc setup` first in case you haven't done so yet")
```
guillaumerose, line 134: extra space before `please`. 

I did a make release and run crc start on a clean installation and the message was already clear:

```
$ crc start
Run 'crc setup' to unpack the bundle to disk
```

Maybe we can rephrase the last part and say `please run `crc setup` to configure your machine.`

praveenkumar: > please run crc setup to configure your machine.

@guillaumerose I like the current one because many time it happen that as user did crc setup and still preflight check fails for whatever reason so better to put `in case you haven't done so yet`.

praveenkumar: > extra space before please.

@guillaumerose done.

guillaumerose (approved):

guillaumerose:
```
$ crc start -p ~/pull-secret.txt -b ~/.crc/cache/crc_libvirt_4.6.15.crcbundle

INFO Checking if systemd-networkd is running      
File not found: /etc/NetworkManager/dnsmasq.d/crc.conf: stat /etc/NetworkManager/dnsmasq.d/crc.conf: no such file or directory
```

Looks good maybe the log could be in preflights package after all. 
We can start with this.

praveenkumar:
> Looks good maybe the log could be in preflights package after all.
We can start with this.

@guillaumerose if we add `logging.Warn` to preflight then we don't even need the error type right?

guillaumerose:
Yes!

praveenkumar:
> Yes!

@guillaumerose :), I went with what @cfergeau suggested to pass the error to upper chain and  check the error type to have the info/warn message, If we add it to preflight then doesn't `crc setup` also show same warn message which is not what a user expect right?

PR #1880, opened 2021-01-22
Integration: pull the image instead building using podman

praveenkumar:


test/integration/features/story_registry.feature
```diff
@@ -14,15 +14,16 @@ Feature: Local image to image-registry
         And login to the oc cluster succeeds
 
     Scenario: Create local image
-        Given executing "podman build -t hello:test -f Dockerfile" succeeds
+        Given executing "podman pull quay.io/bitnami/nginx" succeeds
         When executing "podman images" succeeds
-        Then stdout should contain "localhost/hello"
+        Then stdout should contain "quay.io/bitnami/nginx"
```
cfergeau, line 19: This moves from self contained tests to adding an external dependency we don't control, this could be a bit fragile in the long run. Why is `podman build` hanging the stdin pipe of the shell instance? can this be worked around, or fixed?

test/testdata/Dockerfile
```diff
@@ -1,9 +0,0 @@
-FROM registry.centos.org/centos/centos:latest
```
cfergeau, line 1: ok, my previous comment was a bit unfair as this is also an external dependency :) This could be squashed in the commit moving to use quay.io/../nginx though.

test/integration/features/story_registry.feature
praveenkumar: @cfergeau I need more time why it is hanging the stdin pipe but meanwhile I want to fix the CI first.

test/testdata/Dockerfile
praveenkumar: Right.

test/integration/features/story_registry.feature
cfergeau: Ah, if it's meant to be a temporary fix to fix CI while a deeper investigation is ongoing, it's worth noting it in the commit log.

cfergeau (approved):

PR #1882, opened 2021-01-22
Prepare an archive with everything to build the macOS pkg. Add codesign.

guillaumerose:
Works fine. No more pop-up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,8 +34,8 @@ jobs:
         name: Make
         command: make check
     - run:
-        name: mock make package
-        command: make BUNDLE_DIR=/tmp MOCK_BUNDLE=true package
+        name: mock make out/macos-amd64/crc-macos-amd64.pkg
+        command: make NO_CODESIGN=1 BUNDLE_DIR=/tmp MOCK_BUNDLE=true out/macos-amd64/crc-macos-amd64.pkg
     - store_artifacts:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -34,8 +34,8 @@ jobs:
         name: Make
         command: make check
     - run:
-        name: mock make package
-        command: make BUNDLE_DIR=/tmp MOCK_BUNDLE=true package
+        name: mock make out/macos-amd64/crc-installer-macos-amd64.pkg
+        command: make NO_CODESIGN=1 BUNDLE_DIR=/tmp MOCK_BUNDLE=true out/macos-amd64/crc-installer-macos-amd64.pkg
```
praveenkumar, line 38: Do we really need `installer` as part of naming doesn't `crc-macos-amd64.pkg` good enough since `.pkg` format is the installer one?

guillaumerose: good point I remove it

guillaumerose:
I rebuild the pkg and will ask to test the signing again.

.circleci/config.yml
praveenkumar: @guillaumerose looks like you forgot to update it in the circle CI configuration since it still present.

guillaumerose: thanks!

PR #1884, opened 2021-01-25
ssh: log private keys used for the connection

guillaumerose:
Instead of logging missing private keys, log the private keys used.
Many user are misled by the previous error message. They think they miss
a key.

Output:
```
DEBU Waiting until ssh is available               
DEBU retry loop: attempt 0                        
exit 0             
DEBU Using ssh private keys: [/home/guillaumerose/.crc/cache/crc_libvirt_4.6.9/id_ecdsa_crc /home/guillaumerose/.crc/machines/crc/id_ecdsa] 
DEBU SSH command results: err: ssh: handshake failed: read tcp 127.0.0.1:40008->127.0.0.1:2222: read: connection reset by peer, output:  
command : exit 0
err     : ssh: handshake failed: read tcp 127.0.0.1:40008->127.0.0.1:2222: read: connection reset by peer\n - sleeping 1s 
DEBU retry loop: attempt 1                        
exit 0             
DEBU Using ssh private keys: [/home/guillaumerose/.crc/cache/crc_libvirt_4.6.9/id_ecdsa_crc /home/guillaumerose/.crc/machines/crc/id_ecdsa] 
DEBU SSH command results: err: <nil>, output:     
```

### DIFF
--- a/pkg/crc/ssh/client.go
+++ b/pkg/crc/ssh/client.go
@@ -2,13 +2,14 @@ package ssh
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"strconv"
 	"time"
 
-	"github.com/code-ready/machine/libmachine/log"
+	log "github.com/code-ready/crc/pkg/crc/logging"
 	"golang.org/x/crypto/ssh"
 )
 
@@ -38,13 +39,12 @@ func NewClient(user string, host string, port int, keys ...string) (Client, erro
 func clientConfig(user string, keys []string) (*ssh.ClientConfig, error) {
 	var (
 		privateKeys []ssh.Signer
-		authMethods []ssh.AuthMethod
+		keyPaths    []string
 	)
 
 	for _, k := range keys {
 		key, err := ioutil.ReadFile(k)
 		if err != nil {
-			log.Debugf("Cannot read private ssh key %s", k)
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


gbraad (approved):

pkg/crc/ssh/client.go
```diff
@@ -54,15 +54,17 @@ func clientConfig(user string, keys []string) (*ssh.ClientConfig, error) {
...
 		privateKeys = append(privateKeys, privateKey)
+		keyPaths = append(keyPaths, k)
 	}
 
-	if len(privateKeys) > 0 {
-		authMethods = append(authMethods, ssh.PublicKeys(privateKeys...))
+	if len(privateKeys) == 0 {
+		return nil, errors.New("no ssh private keys available")
 	}
+	log.Debugf("Using ssh private keys: %v", keyPaths)
```
cfergeau, line 63: For what it's worth, since commit 81268bf2f2, we have a single ssh session for each crc run, so during initial start, only the key from `~/.crc/cache` will be used, and shown here. During subsequent starts, only the key from `~/.crc/machines/crc` will be usable, the other one has been removed from `authorized_keys`. This code will be printing the 2 keys, but only the one from `~.crc/machines/crc` will be usable.

Mentioning this here because it confused me and I was not sure the code was working as expected or not (I thought it would be switching to the new ssh key once it's generated).

cfergeau (approved):

PR #1885, opened 2021-01-25
telemetry: send additional data (cpus, memory, network mode) 

guillaumerose:
This uses context to pass the data. This is required as we cannot easily extend or change the signature of cobra.Command RunE.

### DIFF
--- a/cmd/crc/cmd/config/get.go
+++ b/cmd/crc/cmd/config/get.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/spf13/cobra"
 )
 
@@ -18,15 +19,19 @@ func configGetCmd(config config.Storage) *cobra.Command {
 				return errors.New("Please provide a configuration property to get")
 			}
 			key := args[0]
+
 			v := config.Get(key)
-			switch {
-			case v.Invalid:
+			if v.Invalid {
 				return fmt.Errorf("Configuration property '%s' does not exist", key)
-			case v.IsDefault:
-				return fmt.Errorf("Configuration property '%s' is not set. Default value is '%s'", key, v.AsString())
-			default:
-				fmt.Println(key, ":", v.AsString())
 			}
+
+			telemetry.SetContextProperty(cmd.Context(), "key", args[0])
+
+			if v.IsDefault {
+				return fmt.Errorf("Configuration property '%s' is not set. Default value is '%s'", key, v.AsString())
+
+			}
+			fmt.Println(key, ":", v.AsString())
 			return nil
 		},
 	}
--- a/cmd/crc/cmd/config/unset.go
+++ b/cmd/crc/cmd/config/unset.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/spf13/cobra"
 )
 
@@ -21,6 +22,9 @@ func configUnsetCmd(config config.Storage) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			telemetry.SetContextProperty(cmd.Context(), "key", args[0])
+
 			if unsetMessage != "" {
 				fmt.Println(unsetMessage)
 			}
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -16,6 +17,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/preflight"
+	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/code-ready/crc/pkg/crc/validation"
 	crcversion "github.com/code-ready/crc/pkg/crc/version"
 	"github.com/spf13/cobra"
@@ -46,14 +48,14 @@ var startCmd = &cobra.Command{
 		if err := viper.BindFlagSet(cmd.Flags()); err != nil {
 			return err
 		}
-		if err := renderStartResult(runStart(args)); err != nil {
+		if err := renderStartResult(runStart(cmd.Context())); err != nil {
 			return err
 		}
 		return nil
 	},
 }
 
-func runStart(arguments []string) (*machine.StartResult, error) {
+func runStart(ctx context.Context) (*machine.StartResult, error) {
 	if err := validateStartFlags(); err != nil {
 		return nil, err
 	}
@@ -62,6 +64,10 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 
 	if err := preflight.StartPreflightChecks(config); err != nil {
 		return nil, err
+	}
+
+	for _, key := range []string{cmdConfig.CPUs, cmdConfig.Memory, cmdConfig.DiskSize} {
+		telemetry.SetContextProperty(ctx, key, config.Get(key).AsInt())
 	}
 
 	startConfig := machine.StartConfig{
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -68,7 +68,7 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 
 	if err := c.segmentClient.Enqueue(analytics.Identify{
 		AnonymousId: anonymousID,
-		Traits:      traits(),
+		Traits:      addConfigTraits(c.config, traits()),
 	}); err != nil {
 		return err
 	}
@@ -90,6 +90,18 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 		Event:       action,
 		Properties:  properties,
 	})
+}
+
+func addConfigTraits(c *crcConfig.Config, in analytics.Traits) analytics.Traits {
+	return in.
+		Set("proxy", isProxyUsed(c)).
+		Set(config.NetworkMode, c.Get(config.NetworkMode).AsString()).
+		Set(config.EnableClusterMonitoring, c.Get(config.EnableClusterMonitoring).AsBool()).
+		Set(config.ExperimentalFeatures, c.Get(config.ExperimentalFeatures).AsBool())
+}
+
+func isProxyUsed(c *crcConfig.Config) bool {
+	return c.Get(config.HTTPProxy).AsString() != "" || c.Get(config.HTTPSProxy).AsString() != ""
 }
 
 func getUserIdentity(telemetryFilePath string) (string, error) {
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -1,6 +1,7 @@
 package segment
 
 import (
+	"context"
 	"io/ioutil"
 	"net"
 	"os"
@@ -12,6 +13,7 @@ import (
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/pborman/uuid"
 	"github.com/segmentio/analytics-go"
@@ -54,7 +56,7 @@ func (c *Client) Close() error {
 	return c.segmentClient.Close()
 }
 
-func (c *Client) Upload(action string, duration time.Duration, err error) error {
+func (c *Client) Upload(ctx context.Context, action string, duration time.Duration, err error) error {
 	if c.config.Get(config.ConsentTelemetry).AsString() != "yes" {
 		return nil
 	}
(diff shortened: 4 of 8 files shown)


cmd/crc/cmd/start.go
```diff
@@ -63,6 +65,11 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 		return nil, err
 	}
 
+	properties := telemetry.PropertiesFromContext(ctx)
+	for _, key := range []string{cmdConfig.CPUs, cmdConfig.Memory, cmdConfig.DiskSize} {
+		properties.Set(key, config.Get(key).AsInt())
```
gbraad, line 70: this assumes that the CPU, Memory and Disksize are always given. For these this might be the case, as they have default values... but might be risky for other options. Just saying

pkg/crc/segment/segment.go
```diff
@@ -86,6 +94,15 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error
 	})
 }
 
+func addConfigTraits(c *crcConfig.Config, in analytics.Traits) analytics.Traits {
+	return in.
+		Set(config.NetworkMode, c.Get(config.NetworkMode).AsString()).
+		Set(config.HTTPProxy, c.Get(config.HTTPProxy).AsString() != "").
```
gbraad, line 100: We are not allowed to capture these as they can contain credentials for HTTP proxy authentication, or even the naming, like `int-proxy.[companyname]`, etc.

gbraad (changes requested):
Proxy values can not be taken... they need to be redacted.

pkg/crc/segment/segment.go
gbraad: I understand `c.Get(config.HTTPProxy).AsString() != ""` is treated as a bool... but make this more obvious from proceeding code. Values should never leak into here by someone removing this part.

pkg/crc/telemetry/telemetry.go
```diff
@@ -0,0 +1,29 @@
...
+func (p *Properties) Set(name string, value interface{}) {
+	p.Map[name] = value
+}
+
+func ContextWithEmptyProperties(ctx context.Context) context.Context {
+	return context.WithValue(ctx, key, &Properties{Map: make(map[string]interface{})})
+}
+
+func PropertiesFromContext(ctx context.Context) *Properties {
+	value := ctx.Value(key)
```
cfergeau, line 24: Quoting https://blog.golang.org/context
> Value allows a Context to carry request-scoped data. That data must be safe for simultaneous use by multiple goroutines.

I don't think `Properties` as it is now is going to be safe to be shared across multiple go routines. Overall, I'm not so convinced by the use of the `context` package here:

> Incoming requests to a server should create a Context, and outgoing calls to servers should accept a Context. 

We have no incoming/outgoing requests/calls here but we still use a `context`, not clear to me we should be using this here.

cmd/crc/cmd/config/unset.go
```diff
@@ -21,6 +22,10 @@ func configUnsetCmd(config config.Storage) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			properties := telemetry.PropertiesFromContext(cmd.Context())
+			properties.Set("key", args[0])
```
cfergeau, line 27: I'd prefer to have a `config.Watch()` method which allows to register a watcher implementing a given interface, and then to only refer to this interface in the `config` code rather than hardcoding telemetry there.

cfergeau (commented):
When you say "It uses context" in the commit log, could you word it as "It uses the `context` package to pass the extra properties", it took me a while to get there :) And I'm not sure with the current PR, using `context` brings the benefits that are described in the commit log, it almost feels like a `config` instance could be passed directly to the telemetry code, which would make for simpler code.

cmd/crc/cmd/config/unset.go
guillaumerose: `config.Watch` pattern works in the daemon perspective. In `crc config set`, the watcher doesn't exist and if it was, it will likely never see the event as the CLI will quit as soon as possible.

pkg/crc/telemetry/telemetry.go
guillaumerose: I will make `Properties` thread-safe.

We don't have many choices. Either, it is a global map in a var at package level, or it is a context.
The benefit of context is that it can be used in the daemon API quite nicely: just pass a context extended with telemetry and at the end of the request, send to segment.

pkg/crc/segment/segment.go
guillaumerose: extracted, it is more obvious now.

cmd/crc/cmd/config/unset.go
cfergeau: Imo `config.Watch(configWatcherForTelemetry)` can be called somewhere in `cmd/` to avoid having a hard dependency between config and telemetry.

pkg/crc/telemetry/telemetry.go
cfergeau: The segment package already has a `github.com/code-ready/crc/pkg/crc/config.Config` instance, really not clear to me why we need to add context + telemetry dependencies to the config package when the telemetry code already has access to it.

guillaumerose: All config keys are not relevant to all events. cpus/memory are used to crc start, not crc status. 
Also, having config in segment package is not enough: in `crc config set <key>`, it's impossible to pass key to segment without either a global map or context.

cmd/crc/cmd/config/unset.go
guillaumerose: Then, how do you ensure when the CLI quits, that config.Watch correctly saw the event ? You need an extra channel and wait for it.

```
func init() {
  config.Watch(func() {
    // do stuff with the event
    done <- struct{}{}
  })
}

func configSet() {
  config.Set(key, value)
  <-done
}
```

`<-done` cannot be added in `postRun()` since we don't know if an event will be sent. 

This is not better at the end.

cmd/crc/cmd/start.go
cfergeau: Can we convert Memory and DiskSize to bytes before sending to telemetry?

cmd/crc/cmd/config/unset.go
cfergeau: Never mind, ignore this, when I made that comment I thought there was an added dependency on the `telemetry` package in `pkg/crc/config`. But these changes are all in `cmd/` where it's going to be less of a concern.

cmd/crc/cmd/start.go
guillaumerose: Yes very good point ! Will do that

pkg/crc/telemetry/telemetry.go
```diff
@@ -0,0 +1,43 @@
...
+	ret := make(map[string]interface{})
+	for k, v := range p.storage {
+		ret[k] = v
+	}
+	return ret
+}
+
+func ContextWithEmptyProperties(ctx context.Context) context.Context {
+	return context.WithValue(ctx, key, &Properties{storage: make(map[string]interface{})})
+}
```
cfergeau, line 35: Just `NewContext()`?

pkg/crc/telemetry/telemetry.go
```diff
@@ -0,0 +1,43 @@
...
+
+func PropertiesFromContext(ctx context.Context) *Properties {
```
cfergeau, line 37: Not sure the intermediate `Properties` brings much over `telemetry.SetContextProperty(context, "key", value)` and `telemetry.GetContextProperties(context)`. This would be more concise than the current
```
			properties := telemetry.PropertiesFromContext(cmd.Context())
			properties.Set("key", args[0])
```
and all these calls to `PropertiesFromContext` are missing a `nil` check.

pkg/crc/segment/segment.go
```diff
@@ -71,8 +73,14 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error
 		return err
 	}
 
-	properties := analytics.NewProperties().
-		Set("version", version.GetCRCVersion()).
+	properties := analytics.NewProperties()
+	contextProperties := telemetry.PropertiesFromContext(ctx)
+	if contextProperties != nil {
```
praveenkumar, line 78: @guillaumerose do you think with current implementation `contextProperties` is going to `nil`? In the root we use `ContextWithEmptyProperties` where we already do `return context.WithValue(ctx, key, &Properties{storage: make(map[string]interface{})})` which means is is already assigned and not going to be nil, or this is something from the daemon side?

cmd/crc/cmd/start.go
praveenkumar: @guillaumerose @cfergeau what is the benefit of converting it to bytes, we are already sending the memory and disk as int as part of config or start option right?

guillaumerose: I was about to write it and I would have to remove the loop and convert manually here. Maybe it's better to keep what the user entered.

pkg/crc/telemetry/telemetry.go
guillaumerose: done :+1:

pkg/crc/telemetry/telemetry.go
guillaumerose: done :+1:

pkg/crc/segment/segment.go
guillaumerose: With the refactoring Christophe suggested, it's better and it avoids nil checks. PTAL :)

cmd/crc/cmd/config/unset.go
```diff
@@ -21,6 +22,9 @@ func configUnsetCmd(config config.Storage) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			telemetry.SetContextProperty(cmd.Context(), "key", args[0])
```
cfergeau, line 26: Don't we want to also know which key was attempted to be (un)set in failure cases?

cmd/crc/cmd/config/get.go
```diff
@@ -18,15 +19,19 @@ func configGetCmd(config config.Storage) *cobra.Command {
...
+
 			v := config.Get(key)
-			switch {
-			case v.Invalid:
+			if v.Invalid {
 				return fmt.Errorf("Configuration property '%s' does not exist", key)
-			case v.IsDefault:
+			}
+
+			telemetry.SetContextProperty(cmd.Context(), "key", args[0])
```
cfergeau, line 28: We send the key in the "value unset" case, but not in the "non-existing value" case, I would either send in both cases, or in neither.

pkg/crc/segment/segment.go
```diff
@@ -92,6 +92,18 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
...
+func addConfigTraits(c *crcConfig.Config, in analytics.Traits) analytics.Traits {
+	return in.
+		Set("proxy", isProxyUsed(c)).
+		Set(config.NetworkMode, c.Get(config.NetworkMode).AsString()).
+		Set(config.EnableClusterMonitoring, c.Get(config.EnableClusterMonitoring).AsBool()).
+		Set(config.ExperimentalFeatures, c.Get(config.ExperimentalFeatures).AsBool())
+}
+
+func isProxyUsed(c *crcConfig.Config) bool {
+	return c.Get(config.HTTPProxy).AsString() != "" || c.Get(config.HTTPSProxy).AsString() != ""
```
cfergeau, line 104: I forgot if when `http_proxy` is set in the environment, it's reflected in the config?

cmd/crc/cmd/start.go
```diff
@@ -64,6 +66,10 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 		return nil, err
 	}
 
+	for _, key := range []string{cmdConfig.CPUs, cmdConfig.Memory, cmdConfig.DiskSize} {
+		telemetry.SetContextProperty(ctx, key, config.Get(key).AsInt())
```
cfergeau, line 70: Seems like I can't continue the thread which was started on a previous iteration of this PR :-/

> I was about to write it and I would have to remove the loop and convert manually here. Maybe it's better to keep what the user entered.

If we want to report what the user entered, should the key sent to telemetry be named "MemoryUserInput"/..., and should this be sent as a string to catch erroneous non-int inputs?

cfergeau (approved):
At this point, only minor comments left.

cmd/crc/cmd/config/unset.go
guillaumerose: I was tempted but what if the key is the value and we collect bad data? `crc config set mypassword`. At least the key is always validated here.

pkg/crc/segment/segment.go
guillaumerose: I would say no. Viper uses CRC_HTTP_PROXY.

cmd/crc/cmd/config/unset.go
cfergeau: ah good point

cmd/crc/cmd/start.go
guillaumerose:
```
telemetry.SetContextProperty(ctx, cmdConfig.CPUs, config.Get(cmdConfig.CPUs).AsInt())
telemetry.SetContextProperty(ctx, cmdConfig.Memory, config.Get(cmdConfig.Memory).AsInt())
telemetry.SetContextProperty(ctx, cmdConfig.DiskSize, config.Get(cmdConfig.DiskSize).AsInt()*1024)
```

I don't really enjoy this code. It doesn't bring much. I guess we can stay simple.

cfergeau: "Simple" means "hard to understand" to be honest :) people seeing `40` as the disk size in segment will have to figure out if it's GB or GiB if they require a size in bytes.
More annoying is that at the moment in the go-units branch, it's possible to have disk sizes which cannot be represented as GiB, especially not as ints (40GB, 40.75GiB, ...), dunno what should be reported when this is merged? Is it acceptable at that time to start reporting this in bytes?

pkg/crc/segment/segment.go
cfergeau: Imo it's very misleading to have false negative in telemetry regarding proxy usage. If `http_proxy` is set in the env, telemetry will report "proxy: false" while the cluster will be using a proxy.

cmd/crc/cmd/start.go
guillaumerose: I recall a user asking for the units to be the same for memory and disk.

PR #1889, opened 2021-01-26
Pull Secret UX: keep asking it in case of error and better wording

guillaumerose:
Related to #1892 

Couple of UX changes:
* `crc start` now keep asking for the pull secret until a valid one is passed. 
* directly print how to get the pull secret before asking for it. No more indirection with [? for help].
* make the daemon and the cli call almost the same code for the pull secret

```
$ crc start
...
CodeReady Containers requires a pull secret to download content from Red Hat.
You can copy it from the Pull Secret section of https://cloud.redhat.com/openshift/install/crc/installer-provisioned.
? Please enter the pull secret 
X Sorry, your reply was invalid: empty pull secret
X Sorry, your reply was invalid: invalid pull secret: missing 'auths' JSON-object field
X Sorry, your reply was invalid: invalid pull secret: invalid character 'b' looking for beginning of value
X Sorry, your reply was invalid: invalid pull secret: invalid character '?' looking for beginning of value
? Please enter the pull secret ************************
INFO Loading bundle: crc_libvirt_4.6.9.crcbundle ... 
INFO Verifying bundle crc_libvirt_4.6.9.crcbundle ... 
INFO Creating CodeReady Containers VM for OpenShift 4.6.9... 
..

```

### DIFF
--- a/pkg/crc/cluster/pullsecret.go
+++ b/pkg/crc/cluster/pullsecret.go
@@ -4,15 +4,17 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/input"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/validation"
 	crcversion "github.com/code-ready/crc/pkg/crc/version"
+	terminal "golang.org/x/term"
+	"gopkg.in/AlecAivazis/survey.v1"
 )
 
 type PullSecretLoader interface {
@@ -39,7 +41,7 @@ func (loader *interactivePullSecretLoader) Value() (string, error) {
 		return fromNonInteractive, nil
 	}
 
-	fromUser, err := input.PromptUserForSecret("Image pull secret", fmt.Sprintf("Copy it from %s", constants.CrcLandingPageURL))
+	fromUser, err := promptUserForSecret("Image pull secret", fmt.Sprintf("Copy it from %s", constants.CrcLandingPageURL))
 	// This is just to provide a new line after user enter the pull secret.
 	fmt.Println()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


praveenkumar:
/lgtm

Tested the keyring stuff on the linux.

pkg/crc/cluster/pullsecret.go
```diff
@@ -0,0 +1,153 @@
...
+	terminal "golang.org/x/term"
+	"gopkg.in/AlecAivazis/survey.v1"
+)
+
+const (
+	keyringService = "crc"
+	keyringUser    = "pull-secret"
+)
+
+type PullSecretLoader interface {
```
gbraad, line 26: Why is this called `Loader` ?

guillaumerose: because it loads the pull secret from a store. It can load it from the keyring, a file, etc.

pkg/crc/cluster/pullsecret.go
```diff
@@ -0,0 +1,153 @@
...
+	return &PullSecretMemoizer{
+		Getter: &interactivePullSecretLoader{
+			nonInteractivePullSecretLoader: &nonInteractivePullSecretLoader{
+				config: config,
+			},
+		},
+	}
+}
+
+func (loader *interactivePullSecretLoader) Value() (string, error) {
```
gbraad, line 44: The loader is what retrieves the pullsecret from a keyring?

Isn't Retriever or Reader a better fit? A loader sounds more like a skidloeader, or a helper program to get your software up and going, like those for the C64... a tape loader?

Also wondering what the distinction between Interactive and Non-interactive is.

gbraad: More about what triggers the non-interactive flow and when the interactive. as you said, the daemon would not ask... not in that case it would have to error and allow the resolution to happen from where? the daemon is only used by the systray, so in that case this has to be provided beforehand anyways, but this does not support an error feedback for this to 'prompt' for this by a dialog.

gbraad: Getter is a very Java-like term, but perhaps it is more about lead-up to this.... StorageHandler... Store... KeyringLoader? Ot misses perhaps the target to indicate how it is loaded; reason why I mention it is generic. Perhaps purposely?

pkg/crc/machine/types.go
```diff
@@ -19,7 +19,7 @@ type StartConfig struct {
 	NameServer string
 
 	// User Pull secret
-	PullSecret *cluster.PullSecret
+	PullSecret cluster.PullSecretLoader
```
gbraad, line 22: Is this actually still 'cluster' logic? It concerns the cluster ...
*just wondering*

gbraad (commented):
in general LGTM...

just something about the way the pulsecret is handled as part of naming. but aso, how to go about this for non-dekstop environments?

pkg/crc/cluster/pullsecret.go
gbraad: for the interface that would make sense... but in the implementation I do not see this distinction,

guillaumerose:
> how to go about this for non-dekstop environments?

Maybe we could extend the logic with: if keyring.Set() failed, ask the user to store it in ~/.crc/pull-secret

guillaumerose:
I extended commit logs and removed the part about the keyring.

PR #1895, opened 2021-01-27
Update the cluster cidr and service network in case of proxy

praveenkumar:
Need update because of https://github.com/code-ready/snc/commit/e3ae681571cdb86ebff027c91b052b7d8eb9a5a4

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -236,7 +236,7 @@ func AddProxyToKubeletAndCriO(sshRunner *ssh.Runner, proxy *network.ProxyConfig)
 	proxyTemplate := `[Service]
 Environment=HTTP_PROXY=%s
 Environment=HTTPS_PROXY=%s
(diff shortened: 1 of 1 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -236,7 +236,7 @@ func AddProxyToKubeletAndCriO(sshRunner *ssh.Runner, proxy *network.ProxyConfig)
 	proxyTemplate := `[Service]
 Environment=HTTP_PROXY=%s
 Environment=HTTPS_PROXY=%s
-Environment=NO_PROXY=.cluster.local,.svc,10.116.0.0/14,172.25.0.0/16,192.168.126.0/24,%s`
+Environment=NO_PROXY=.cluster.local,.svc,10.217.0.0/22,10.217.4.0/23,192.168.126.0/24,%s`
```
cfergeau, line 239: Looks like these should be added to the metadata file.

cfergeau (approved):

guillaumerose:
I am wondering if we should bump the version number of the bundle. This is a (very small) breaking change.

/lgtm

PR #1896, opened 2021-01-28
Add vm not exist error type 

praveenkumar:
With recent data collection side, lots of error message is for the
crc vm is not exist and user need to start it. With this patch
`VMNotExist` error type introduce for such case so in segment side
it can be tracked and ignored.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -13,6 +13,7 @@ import (
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
+	crcErr "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/code-ready/crc/pkg/crc/network"
@@ -102,7 +103,7 @@ func checkIfMachineMissing(client machine.Client) error {
 		return err
 	}
 	if !exists {
-		return fmt.Errorf("Machine '%s' does not exist. Use 'crc start' to create it", client.GetName())
+		return crcErr.VMNotExist
 	}
 	return nil
 }
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -8,6 +8,14 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
+type vmNotExist string
+
+func (v vmNotExist) Error() string {
+	return string(v)
+}
+
+const VMNotExist vmNotExist = "Machine does not exist. Use 'crc start' to create it"
+
 type MultiError struct {
 	Errors []error
 }
--- a/pkg/crc/errors/serialerror.go
+++ b/pkg/crc/errors/serialerror.go
@@ -17,6 +17,6 @@ func (e *SerializableError) MarshalJSON() ([]byte, error) {
 	return json.Marshal(e.Error())
 }
 
-func (e *SerializableError) UnWrap() error {
+func (e *SerializableError) Unwrap() error {
 	return e.error
 }
--- a/test/e2e/features/basic.feature
+++ b/test/e2e/features/basic.feature
@@ -33,7 +33,7 @@ Feature: Basic test
         When executing "crc status" fails
         Then stderr should contain
             """
-            Machine 'crc' does not exist. Use 'crc start' to create it
+            Machine does not exist. Use 'crc start' to create it
             """
 
     @linux


pkg/crc/segment/segment.go
```diff
@@ -73,9 +74,9 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error
 
 	properties := analytics.NewProperties().
 		Set("version", version.GetCRCVersion()).
-		Set("success", err == nil).
+		Set("success", err == nil || err.Error() == crcErr.VMNotExist.Error()).
 		Set("duration", duration.Milliseconds())
-	if err != nil {
+	if err != nil && err.Error() != crcErr.VMNotExist.Error() {
```
guillaumerose, line 79: You can use https://golang.org/pkg/errors/#As

pkg/crc/errors/multierror.go
```diff
@@ -8,6 +8,14 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
+type vmNotExist string
+
+func (v vmNotExist) Error() string {
+	return string(v)
+}
+
+const VMNotExist vmNotExist = "Machine crc does not exist. Use 'crc start' to create it"
```
guillaumerose, line 17: Maybe remove `crc` and just say `Machine does not exist. Use 'crc start' to create it`

pkg/crc/segment/segment.go
```diff
@@ -73,9 +74,9 @@ func (c *Client) Upload(action string, duration time.Duration, err error) error
 
 	properties := analytics.NewProperties().
 		Set("version", version.GetCRCVersion()).
-		Set("success", err == nil).
+		Set("success", err == nil || err.Error() == crcErr.VMNotExist.Error()).
```
cfergeau, line 77: Does this have to be a string comparison? Could it be `err == crcErr.VMNotExist`? Feels a bit odd to add an error type, and not really use it, the way it's used, it could be replaced by `const VMNotExistStr = "Machine crc does not exist. Use 'crc start' to create it"`.
And it might make sense to record this separately, so that we can have reports for "command succeeded", "command did not succeed because the VM did not exist", "command failed" ?

pkg/crc/errors/multierror.go
praveenkumar: @guillaumerose Yes that what I was thinking, good that we have same thought here.

pkg/crc/segment/segment.go
praveenkumar: > Does this have to be a string comparison? Could it be `err == crcErr.VMNotExist`?

@cfergeau @guillaumerose We can't use either `err == crcErr.VMNotExist` or even `error.As()` because what we do as part of render we don't propagate same error but only propagate the error string and then check again if string is non empty, create a new one which break the error chain and show that user :(

- https://github.com/code-ready/crc/blob/master/cmd/crc/cmd/delete.go#L61
- https://github.com/code-ready/crc/blob/master/cmd/crc/cmd/delete.go#L74

>  Feels a bit odd to add an error type, and not really use it, the way it's used, it could be replaced by `const VMNotExistStr = "Machine crc does not exist. Use 'crc start' to create it"`.

I know, I am thinking a way to persist the error chain so we can do something like `error.As()` . It does feel old right now but it should be useful but if you want in I can change it to the string constant for time being.

> And it might make sense to record this separately, so that we can have reports for "command succeeded", "command did not succeed because the VM did not exist", "command failed" ?

The way I think for a user this command is failed since the vm is not exist but we really don't want to capture that data (are we?) what it will tell us in long run?

cfergeau: Yeah, ideally we would not lose the initial error so that we can use `errors.As` or such. But no idea how involved this would be :)

praveenkumar:
/hold

Depend on #1898

gbraad (approved):

pkg/crc/segment/segment.go
```diff
@@ -79,9 +81,9 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 	}
 
 	properties = properties.Set("version", version.GetCRCVersion()).
-		Set("success", err == nil).
+		Set("success", err == nil || errors.Is(err, crcErr.VMNotExist)).
```
guillaumerose, line 84: Maybe extract a function like:

```
sentError := filter(err)

func filterError(err error) error {
  if errors.Is(err, crcErr.VMNotExist) {
    return nil
  }
  return err
}
```

praveenkumar: @guillaumerose sure that can be a helper function.

gorkem:
I am not sure about this change. There are other errors that appear on data that are not really errors either and we can do their filtering on the reports easily. IMHO you should let the reporting tool take care of filtering.

cfergeau:
> I am not sure about this change. There are other errors that appear on data that are not really errors either and we can do their filtering on the reports easily. IMHO you should let the reporting tool take care of filtering.

This echoes the short discussion in https://github.com/code-ready/crc/pull/1896#discussion_r565920147

>>    And it might make sense to record this separately, so that we can have reports for "command succeeded", "command did not succeed because the VM did not exist", "command failed" ?

>  The way I think for a user this command is failed since the vm is not exist but we really don't want to capture that data (are we?) what it will tell us in long run?

praveenkumar:
I remove the commit which ignore the error to segment, will try to put separate PR for capturing the error type so we can create filter later.

cmd/crc/cmd/root.go
```diff
@@ -102,7 +103,7 @@ func checkIfMachineMissing(client machine.Client) error {
 		return err
 	}
 	if !exists {
-		return fmt.Errorf("Machine '%s' does not exist. Use 'crc start' to create it", client.GetName())
+		return crcErr.VMNotExist
 	}
```
cfergeau, line 107: Just a note for future improvement, `client.Exists()` could return a `crcErr.VMNotExist` when the VM does not exist, which would make `checkIfMachineMissing()` implementation simpler/shorter.

PR #1898, opened 2021-01-28
Implement serializable error to use with render

praveenkumar:
Right now, errors which are occurred in the pkg level and passed to
cmd are rendered and lose it's error chain. Following is happening
for most of the commands where error is passed as string and then
it again converted to new error using `errors.New`.

```
func (s *deleteResult) prettyPrintTo(writer io.Writer) error {
	if s.Error != "" {
		return errors.New(s.Error)
	}
...
```

This patch will try to have error chain back to cmd level when render
happen and output format is not json one.

### DIFF
--- a/cmd/crc/cmd/cleanup.go
+++ b/cmd/crc/cmd/cleanup.go
@@ -1,11 +1,11 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"os"
 
+	crcErrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/preflight"
 	"github.com/spf13/cobra"
 )
@@ -28,18 +28,18 @@ func runCleanup() error {
 	err := preflight.CleanUpHost()
 	return render(&cleanupResult{
 		Success: err == nil,
-		Error:   errorMessage(err),
+		Error:   crcErrors.ToSerializableError(err),
 	}, os.Stdout, outputFormat)
 }
 
 type cleanupResult struct {
-	Success bool   `json:"success"`
-	Error   string `json:"error,omitempty"`
+	Success bool                         `json:"success"`
+	Error   *crcErrors.SerializableError `json:"error,omitempty"`
 }
 
 func (s *cleanupResult) prettyPrintTo(writer io.Writer) error {
-	if s.Error != "" {
-		return errors.New(s.Error)
+	if s.Error != nil {
+		return s.Error
 	}
 	_, err := fmt.Fprintln(writer, "Cleanup finished")
 	return err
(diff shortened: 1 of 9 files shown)


cmd/crc/cmd/json.go
```diff
@@ -19,17 +19,17 @@ func addOutputFormatFlag(cmd *cobra.Command) {
 }
 
 type prettyPrintable interface {
-	prettyPrintTo(writer io.Writer) error
+	prettyPrintTo(writer io.Writer, err error) error
```
guillaumerose, line 22: This extra error feels weird. I understand it's not really convenient to pass the error. 

I made an attempt with the use of Unwrap() method: https://github.com/code-ready/crc/compare/master...guillaumerose:unwrap?expand=1#diff-b6063933a261de0116a269a05888f288ccde569b5e83c2abe9435f79a65fc46bR65

praveenkumar: @guillaumerose I will look into the suggestion tomorrow but if you see when you are using render it is having the obj which contains the error as string so passing same err as different argument shouldn't be the issue.

cfergeau: I did not look in detail at Guillaume's code, but upon seeing 
```
	if s.Error != "" {
		return errors.New(s.Error)
	}
```
I wanted to ask why `s.Error` is not a real error, which is what Guillaume's changes are doing, so imo it's worth taking a closer look ;)

cmd/crc/cmd/cleanup.go
```diff
@@ -29,18 +28,18 @@ func runCleanup() error {
...
 
 type cleanupResult struct {
 	Success bool   `json:"success"`
 	Error   string `json:"error,omitempty"`
 }
 
-func (s *cleanupResult) prettyPrintTo(writer io.Writer) error {
-	if s.Error != "" {
-		return errors.New(s.Error)
+func (s *cleanupResult) prettyPrintTo(writer io.Writer, err error) error {
```
gbraad, line 39: teh error is here just to pass it on? feels weird...

pkg/crc/errors/serialerror.go
```diff
@@ -0,0 +1,22 @@
+package errors
+
+import "encoding/json"
+
+func EventuallySerializableError(err error) *SerializableError {
```
guillaumerose, line 5: `errors.EventuallySerializableError` can be also:
* `errors.EventuallySerializable` (remove the extra Error)
* `errors.MaybeSerializable`

Otherwise, lgtm!

cfergeau: `errorMessage` naming did not indicate the `maybe`/`eventually`/... Imo it's more readable without this hint, hopefully it makes sense to assume `errors.EventuallySerializableError(nil)` returns `nil`? In which case, this could become `MakeSerializable`, `ToSerializable` (feels a bit weird without the `Error` suffix, so maybe keep it too).

praveenkumar: I changed it to `ToSerializableError` make it much readable and also provide sense that it will serialize the error.

Issue #1340, opened 2020-06-26
Bundle compression/decompression takes too much time


guillaumerose:
The crcbundle is currently compressed using `xz`. This format is not natively supported by Golang and we have to use github.com/xi2/xz library.

On my machine, it takes 3m47s to decompress a bundle using xi2/xz. With `xz` command line tool, it takes 2m20s. 
xi2/xz documentation says:
``` 
// On the author's Intel Ivybridge i5, decompression speed is about
// half that of the standard XZ Utils (tested with a recent linux
// kernel tarball).
``` 
I also tried with https://github.com/ulikunitz/xz and it doesn't finish in less than 5min.

Given that the compression step in snc takes also a huge amount of time (about 30min by bundle if I'm correct), maybe we should investigate an other format.

```
crc_libvirt_4.4.6.crcbundle
xz: 2.0GB
zip (default level): 2.8GB
gzip (default level): 2.8GB
zstd: 2.6GB
```

gbraad:
See #994

gbraad:
Decompression takes even more time on Windows

cfergeau:
On an idle 6 cores/12 threads machine, using threaded compression in snc makes a big difference:
```
$ time xz --keep --threads=0 /home/teuf/.crc/cache/crc_libvirt_4.4.8/crc.qcow2
real    4m36.255s
user    54m17.698s
sys     0m3.416s

$ time xz --keep /home/teuf/.crc/cache/crc_libvirt_4.4.8/crc.qcow2
real    34m48.959s
user    34m46.237s
sys     0m2.710s
```

anjannath:
https://github.com/code-ready/crc/issues/1340#issuecomment-653589793 Improves compression time for our bundle generation. 

The concern with the decompression time on crc side still remains.

cfergeau:
Additional data from https://github.com/code-ready/crc/issues/1893#issuecomment-769067339
`zstd --ultra -22 -T0` gives me a 2.2G tarball which decompresses in 10 seconds
The corresponding crcbundle (4.6.9) is 2.6G and takes 2 minutes to decompress.
(figures to be double checked/confirmed)

cfergeau:
A few more datapoints with zstd, default settings gave me a 3.6GB tarball, so not interesting, `zstd -19` gives a 2.7GB tarball, so it's comparable in size with xz. Decompression time is 13 seconds.

cfergeau:
[...]

PR #1905, opened 2021-01-29
extract: Add zstd support

cfergeau:
Testing[1] suggests that zstd compression might be better than xz, with
decompression being much faster than xz, so it's worth considering it
for our bundles.

There is no native support for zstd in golang, so this makes use of
https://pkg.go.dev/github.com/klauspost/compress/zstd for handling this
format. An alternative would be https://github.com/DataDog/zstd which
provides performance numbers, but which is not a pure go implementation,
which is why I picked the other one.

When compressed with `zstd --ultra -22 -T0`, the 4.6.9 bundle is 2.2G in
size, and uncompresses in about 1 minute. The xz-compressed bundle is
2.6G in size, and takes 3 minutes to uncompress.

An existing bundle can be converted to zstd for testing purposes with:
xzcat $bundle |zstd  -  -o crc_libvirt_xxx.crcbundle

This fixes https://github.com/code-ready/crc/issues/1340 for now.

## Testing

Behaviour should be unchanged, crc should now be able to use zstd-compressed
bundles when they are made available.

This can be tested by converting an existing bundle to zstd:
xzcat $bundle |zstd  -  -o crc_libvirt_xxx.crcbundle

and then cleaning up ~/.crc/cache, and running crc setup/crc start -b crc_libvirt_xxx.crcbundle


pkg/extract/extract.go
```diff
@@ -13,6 +14,7 @@ import (
 	"github.com/cheggaaa/pb/v3"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/h2non/filetype"
+	"github.com/klauspost/compress/zstd"
```
praveenkumar, line 17: Didn't go through the PR but as per document `Currently the package is heavily optimized for 64 bit processors and will be significantly slower on 32 bit processors.`  and some folks want to use crc on power-pc does it also affect performance there or they should be good using tarball of xz compression?

cfergeau: I expect most power machines to be 64 bit? Name of the arch is `ppc64le` after all. Not sure about s390, but I'd expect them to also be 64 bit.

gbraad: we do not support 32bits anyways, as Hyper-V is not available on those platforms (our image is 64bits), plus even PPC has been 64bits long before Intel+AMD started testing this.

gbraad (approved):

guillaumerose:
Great! Let's wait few days for https://github.com/h2non/filetype/pull/98 maybe.

/lgtm

cfergeau:
> Great! Let's wait few days for [h2non/filetype#98](https://github.com/h2non/filetype/pull/98) maybe.

The maintainer was very responsive, so yes we can wait for it (not merged yet as it needed a doc change which I have made now).

cfergeau:
> > Great! Let's wait few days for [h2non/filetype#98](https://github.com/h2non/filetype/pull/98) maybe.
> 
> The maintainer was very responsive, so yes we can wait for it (not merged yet as it needed a doc change which I have now made).

It's merged now, I've amended the PR accordingly, and am testing the corresponding snc changes.

guillaumerose (approved):

gbraad (approved):

PR #1910, opened 2021-02-01
Misc changes around configuration.

praveenkumar:
- Remove confusing words from config usage
- Add value type for key
- Introduce tabwriter to display properties and value type
- Use expected value from validation function

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -97,7 +97,7 @@ func GetConfigCmd(config *config.Config) *cobra.Command {
 		Use:   "config SUBCOMMAND [flags]",
 		Short: "Modify crc configuration",
 		Long: `Modifies crc configuration properties.
-Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields(config),
+Properties: ` + "\n\n" + configurableFields(config),
 		Run: func(cmd *cobra.Command, args []string) {
 			_ = cmd.Help()
 		},
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -1,9 +1,11 @@
 package config
 
 import (
+	"bytes"
 	"fmt"
 	"sort"
 	"strings"
+	"text/tabwriter"
 
 	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
@@ -79,16 +81,24 @@ func less(lhsKey, rhsKey string) bool {
 
 func configurableFields(config *config.Config) string {
 	var fields []string
-	keys := keysAndValueType(config)
+	var buf bytes.Buffer
+	writer := tabwriter.NewWriter(&buf, 0, 8, 1, ' ', tabwriter.TabIndent)
+	for _, keyAndValueType := range keysAndValueType(config) {
+		fmt.Fprintln(writer, keyAndValueType)
+	}
+	writer.Flush()
+	keys := strings.Split(buf.String(), "\n")
 	sort.Slice(keys, func(i, j int) bool {
 		return less(keys[i], keys[j])
 	})
 	for _, key := range keys {
+		if key == "" {
+			continue
+		}
 		fields = append(fields, " * "+key)
 	}
 	return strings.Join(fields, "\n")
 }
-
 func keysAndValueType(config *config.Config) []string {
 	var keyAndValueType []string
 	for key, value := range config.AllConfigs() {
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


cmd/crc/cmd/config/config.go
```diff
@@ -78,26 +81,49 @@ func less(lhsKey, rhsKey string) bool {
...
 	}
 	return strings.Join(fields, "\n")
 }
+func keysAndValueType(config *config.Config) []string {
+	var keyAndValueType []string
+	for key, value := range config.AllConfigs() {
+		var valueType string
+		switch value.Value.(type) {
+		case int:
+			valueType = "Integer"
```
gbraad, line 108: is this what we show to the user? an integer as better descirbed as a number, string as text and boolean as a yes/no

gbraad (commented):
Do we want to be so programmer specific to the user?

cmd/crc/cmd/config/config.go
jsliacan: Maybe yes, but I'd keep `Integer` instead of `Number`, as it isn't a rational or a Gaussian integer :). To be fair, it should be a `Natural number` (as in a non-negative int)  because negative integers aren't sensible in this context either.

praveenkumar: @gbraad yes this is what shown to user if you think that this should be described as number, text and true/false then I am OK, I think string is better than text but let's see what other think.

cmd/crc/cmd/config/config.go
```diff
@@ -92,6 +89,25 @@ func configurableFields(config *config.Config) string {
...
+		case string:
+			valueType = "String"
+		case bool:
+			valueType = "Boolean"
+		default:
+			valueType = ""
```
cfergeau, line 114: fmt.Sprintf("%T", value.Value)?

cmd/crc/cmd/config/config.go
cfergeau: `Integer` is not a great way of describing `Memory`, as what you want to know is that it's an integer value in MiB (this is almost solved by the go-units work though). Similarly, `Boolean` is not really useful to know which values are correct (true/false, 1/0, yes/no ?)

pkg/crc/config/config.go
```diff
@@ -55,6 +55,11 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
 		return "", fmt.Errorf(configPropDoesntExistMsg, key)
 	}
 
+	ok, expectedValue := c.settingsByName[key].validationFn(value)
+	if !ok {
+		return "", fmt.Errorf(invalidProp, value, key, expectedValue)
+	}
+
 	var castValue interface{}
```
cfergeau, line 63: Looking at `ValidateBool`, it special cases `true`/`false`, but does not make use of `strconv.ParseBool`. Should both behaviours be unified? It makes more sense to first convert the value to the correct type (casting), and then validate this value. I understand you want a better error message, but I'm not sure exchanging them is the right way of achieving that. Maybe there's a way of 'merging' both steps somehow?

praveenkumar: @cfergeau But we are casting every user input as part of validation function and if casting fails then returning expected message. in case of `ValidateBool` if a user provide something different than `true/false` string the reason is present which is easy to understand. Now with providing the valuetype of key as part of config help should remove some confusion around it.

cfergeau: `ValidateBool` only accepts "true"/"false", contrast this with `cast.ToBoolE` which uses https://golang.org/pkg/strconv/#ParseBool :
>  It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False
This is the inconsistency/change in behaviour I wanted to point out.

cfergeau: > in case of ValidateBool if a user provide something different than true/false string the reason is present which is easy to understand. Now with providing the valuetype of key as part of config help should remove some confusion around it.

What would be even nicer is to have `memory - memory size in MiB (integer, default '9000')` rather than `memory - number`. I think this could be done with the addition of a help string to `config.Setting`?

praveenkumar: > Instead of only providing all the details (minimum size, ...) when there's a failure, it would be even nicer is to have memory - memory size in MiB (integer, default '9000') rather than memory - number. I think this could be done with the addition of a help string to config.Setting?

@cfergeau Sure, but that will need bit more work, may be a follow up PR, I can create a tracking issue if you want.

praveenkumar: @cfergeau created #1933

cfergeau: > `ValidateBool` only accepts `true`/`false`, contrast this with `cast.ToBoolE` which uses https://golang.org/pkg/strconv/#ParseBool :
> 
> > It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False
> 
> This is the inconsistency/change in behaviour I wanted to point out.

This PR has been merged without further discussion on this point. I'd like to emphasize that there was a command line behaviour change before/after this PR:

Before:
```
$ crc config set skip-check-vsock 1
Successfully configured skip-check-vsock to true
```

After:
```
$ crc config set skip-check-vsock 1
Value '1' for configuration property 'skip-check-vsock' is invalid, reason: must be true or false
```

Issue #1611, opened 2020-10-22
[Epic] Basic Podman 2 support


gbraad:
* [x] deal with client download/embedding https://github.com/code-ready/snc/issues/318
  * [ ] expose from command -> this is now a `podman system connection` using `SSH`
  * ~~handle delayed startup of the OpenShift cluster -> #1097~~
     * ~~state of OpenShift needs to be handled on (re)start of cluster~~

---

Delayed startup is now part of the #2167

gbraad:
@anjannath did you have any comments?

guillaumerose:
As we do for oc client, I guess we should add podman-remote in the snc bundle.

gbraad:
Podman 2 might be planned for inclusion with 4.7

gbraad:
Podman v2 is NOT usable in the current release of RHCOS/OpenShift 4.7 as it still uses the v1 API (2.0 release). We need a v2.1 release before we can make use of the improved API.

guillaumerose:
Bundle 4.7.2 provides Podman v2.2.1. Would this work?

gbraad:
I'll have to run a few tests as https://github.com/containers/podman/issues/9098 showed that this version was inconsistently defined and used. Will let you know.

guillaumerose:
I tried and it works fine now (using podman-remote v2.2.1)

```
$ podman-remote-static system connection add myuser --identity ~/.crc/machines/crc/id_ecdsa ssh://core@192.168.130.11/run/user/1000/pod
$ podman-remote-static pull nginx
$ podman-remote-static run -p 8080:80 -it nginx 
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
[...]

Issue #1097, opened 2020-03-11
Possibility to run CRC without deploying OpenShift (for use of podman runtime)


gbraad:
Currently we deciced not to change the default behaviour of CRC, which means we deploy OpenShift and expose Podman up[on request. Details in the [spike](https://github.com/code-ready/crc/issues/874#issuecomment-575854891) (copied to end of the message)

However, people have expressed, as expected, not to start openShift when only Podman is wanted. We can expose this as a possible start option: `--no-provision` (as done in Minishift), or a specific `--podman` option?

---

[We] spoke about the possible strategies to expose this to the user as part of the user story. We identified the following 3 possiblities:

  1) We use a lean VM approach by re-using the current RHCOS and not starting the `kubelet` service. This means we can re-use the current setup (while prepared during the build phase), but might have to handle a situation to allow the start the OpenShift cluster afterwards;. This enables a scenario in which we do not need to share the resource initially between both `podman` and OpenShift. The problem is however, to handle the start situation as this impacts our current flow of interaction. While not impossible, it might look like:
```
$ crc start --podman
Starting CRC VM
Podman access available
$ crc status
CRC VM started
OpenShift cluster not available
Podman access available
$ crc start  # default --openshift is assumed
CRC VM started
Starting OpenShift cluster.
$ crc status
CRC VM started
OpenShift cluster started
...
Podman access available
```

  2) Start the full VM and allow access to `podman` after specifically requesting this. This simplfiies the start process and does not have to do existence checks, etc. But in this situation, the VM will consume resource to maintain both the OpenShift cluster and `podman`. This might be in a podman-only sitiation waste a significant amount of memory

```
$ crc start
CRC VM starting
OpenShift cluster started
$ crc podman-env
# CRC VM started, so we only need to expose access
Podman access available
$ crc status
OpenShift cluster started 
...
Podman access available
```

  3) A dedicated VM. While being a leaner aproach, this introduces additional conmplexity of maintaining another VM, allow VMs to co-exist, etc. While not impossible (our codebase can) it does introduce a resource overhead when both are used, but solves separation, etc.

```
$ crc start --podman
Podman access available
$ crc start --openshift    # default
OpenShift started
$ crc status
Podman VM started
OpenShift VM started
...
```

At the moment we decided to go for option **2**, and work on a situation as described in **1** over time.

gbraad:
the idea is to also use the systray configuration option for this, as it allows to make this stand out from a dialog

[...]

PR #1912, opened 2021-02-01
[WIP] Add toggle to not start the OpenShift cluster

gbraad:
Signed-off-by: Gerard Braad <me@gbraad.nl>

This adds a toggle to prevent starting the kubelet service for use with a podman-only target
However, some side-effects remains:

  * a running VM will report `KubeletStarted` (restart)
  * etc.

For a podman-only VM, you can start with `crc start --start-openshift=false`

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -26,6 +26,7 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
+	StartOpenShift          = "start-openshift"
 )
 
 func RegisterSettings(cfg *config.Config) {
@@ -49,6 +50,9 @@ func RegisterSettings(cfg *config.Config) {
 
 	// Telemeter Configuration
 	cfg.AddSetting(ConsentTelemetry, "", config.ValidateYesNo, config.SuccessfullyApplied)
+
+	// Change startup behaviour
+	cfg.AddSetting(StartOpenShift, true, config.ValidateYesNo, config.SuccessfullyApplied)
 }
 
 func isPreflightKey(key string) bool {
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -33,6 +33,7 @@ func init() {
 	flagSet.UintP(cmdConfig.DiskSize, "d", constants.DefaultDiskSize, "Total size in GiB of the disk used by the OpenShift cluster")
 	flagSet.StringP(cmdConfig.NameServer, "n", "", "IPv4 address of nameserver to use for the OpenShift cluster")
 	flagSet.Bool(cmdConfig.DisableUpdateCheck, false, "Don't check for update")
+	flagSet.Bool(cmdConfig.StartOpenShift, true, "Toggle to prevent starting the OpenShift cluster")
 
 	startCmd.Flags().AddFlagSet(flagSet)
 }
@@ -72,6 +73,7 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 		PullSecret: &cluster.PullSecret{
 			Getter: getPullSecretFileContent,
 		},
+		StartOpenShift: config.Get(cmdConfig.StartOpenShift).AsBool(),
 	}
 
 	client := newMachine()
@@ -80,14 +82,15 @@ func runStart(arguments []string) (*machine.StartResult, error) {
 
 func renderStartResult(result *machine.StartResult, err error) error {
 	return render(&startResult{
-		Success:       err == nil,
-		Error:         errorMessage(err),
-		ClusterConfig: toClusterConfig(result),
+		Success:          err == nil,
+		Error:            errorMessage(err),
+		ClusterConfig:    toClusterConfig(result),
+		OpenShiftStarted: result.KubeletStarted,
 	}, os.Stdout, outputFormat)
 }
 
 func toClusterConfig(result *machine.StartResult) *clusterConfig {
-	if result == nil {
+	if result == nil || result.KubeletStarted == false {
 		return nil
 	}
 	return &clusterConfig{
@@ -126,9 +129,10 @@ type credentials struct {
 }
 
 type startResult struct {
-	Success       bool           `json:"success"`
-	Error         string         `json:"error,omitempty"`
-	ClusterConfig *clusterConfig `json:"clusterConfig,omitempty"`
+	Success          bool           `json:"success"`
+	Error            string         `json:"error,omitempty"`
+	ClusterConfig    *clusterConfig `json:"clusterConfig,omitempty"`
+	OpenShiftStarted bool           `json:"openshiftStarted,omitempty"`
 }
 
 func (s *startResult) prettyPrintTo(writer io.Writer) error {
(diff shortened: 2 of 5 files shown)


cmd/crc/cmd/config/config.go
```diff
@@ -49,6 +50,9 @@ func RegisterSettings(cfg *config.Config) {
 
 	// Telemeter Configuration
 	cfg.AddSetting(ConsentTelemetry, "", config.ValidateYesNo, config.SuccessfullyApplied)
+
+	// Change startup behaviour
+	cfg.AddSetting(StartOpenShift, true, config.ValidateYesNo, config.SuccessfullyApplied)
```
cfergeau, line 55: Fwiw, it's not really intuitive to disable seemingly unrelated stuff ('openshift') in order to enable 'podman mode', maybe this should be `cluster-mode: podman|openshift` or something similar.

gbraad: Podman is not a mode; it is always existent.

guillaumerose: Would it be possible to compile crc under a new name like podman-machine and change slightly the behavior based on this ?
If I am using only podman, I use podman-machine but I can "upgrade" instantly by using "crc".

cfergeau: On linux/mac, no need for a new binary, this can be a symlink + a check of os.Args[0]. Maybe this can also work on windows as well.

cfergeau: > Podman is not a mode; it is always existent

Could be `podman-only` and `podman+openshift`. But ok, this `disable-openshift` toggle is not required for podman, it's just an optimization.

gbraad: This should be reviewed in isolation of podman
The start and stop of openshift is something that enables this, but it's not the only driver. 

Starting the vm without anything besides DHCP already is a podman vm with our image

gbraad: The idea could be to reintroduce `crc openshift [start/stop/status/...]`

guillaumerose: > Complexity of a new executable for what benefit? 

Why podman would be a second-class citizen and not openshift ? 
On Linux, users will use crc for openshift. 
On macOS, there is more value and more users to gain with podman.

For me the UX feels better with 2 commands. 

crc is a way to have a maintained docker-machine for podman nothing else.. With 2 separated binaries/or a clean way to differentiate both, I would bake a minimal bundle and create a Podman Desktop.pkg :)

gbraad: you would want to run two VMs too?

guillaumerose: In case of 2 differents apps, yes. Maybe for OKD also, we could be able to launch side by side 2 VMs.

deboer-tim: For me CRC is the 'brand' for single-user/evaluator OpenShift, so even if we agree on a parameter name to have any sort of 'crc --no-openshift' is entirely confusing. The same way that 'java --no-java' or 'chectl --no-che' would be.

It seems like the use of this would be fairly different too, so even if the implementation is shared or very similar I'd prefer to see a separate cli.

guillaumerose: The cost of 2 CLIs is really high.
Maybe we can start with this simple flag and see if people using it with the telemetry. If there is enough usage, we can turn that into a proper CLI ?

gbraad: Right,  it should not be underestimated what the cost involved is. It would be like a totally new lifecycle management tool. Even if shared in this codebase, the technical debt for that is very high.

These changes to have the cluster be restarted bring benefit by itself.

Note: this parameter we also used on minishift for a docker runtime only option.

gbraad: The question can still be; should this be part of CRC?

gbraad:
@deboer-tim this is related to the original scenarios to enable podman to be 're-used' https://github.com/code-ready/crc/issues/1097

the use would be
```
$ crc start 
$ crc podman-env
```

which will start the regular cluster and allows you to use podman alongside the cluster (shares resource usage)

or with this option:
```
$ crc start --start-openshift=false
$ crc podman-env
```

you would have a podman-only VM, but allowing to start the cluster if we add the lifecycle commands like:

```
$ crc openshift start
```

which will start the cluster on the VM if it wasn't running.

deboer-tim:
Thanks for taking the time to walk me through the use cases and background today. Is the ability to change the running CRC a core use-case, or just something that's technically possible? If it hasn't been requested then IMHO it would probably be better to leave out for now and users can just stop and restart.

I agree it's not worth creating a separate download or command line, but it's still bothering me that the cli focuses on 'not starting OpenShift' more than 'We start two things by default but you can choose to only start one of them'. Maybe instead of:
> $ crc start --start-openshift=false
it should be something like:
> $ crc start --podman
or:
> $ crc start --podman-only
?

gbraad:
> it should be something like:
> `$ crc start --podman`
> or:
> `$ crc start --podman-only`
> ?

I would like to leave out referencing `podman` in the flags we have, as they can not easily be removed ince they has once been added. we do not know if the podman support stays

cfergeau:
> > it should be something like:
> > `$ crc start --podman`
> > or:
> > `$ crc start --podman-only`
> > ?
> 
> I would like to leave out referencing `podman` in the flags we have, as they can not easily be removed ince they has once been added. we do not know if the podman support stays (as it has disappeared before :-)

But if it needs to go away, we can rename them while keeping the old option name as an alias, hide the old name, output warnings when the old name is used, before finally removing it. Preferrable not to do that, but does not seem very hard to achieve

guillaumerose:
Proposition: have different levels of crc.

4 levels:
* podman only
* slim openshift: routes and image registry only (recommended for developers - 6GB)
* openshift: all the above + samples/marketplace/olm and console (recommended for trial - 9GB)
* full openshift: all the above + monitoring - (best openshift experience - 12GB)

I imagine at `crc start` time we could prompt the question to the user. `What level of integration do you choose?`
In the tray, we could think at a slider from bottom to top with the 4 levels + description.

The major argument is that with slim openshift people can run it almost all the time.

WDYT?  @gbraad @deboer-tim

gbraad:
I believe we decided to leave CRC untouched and merely expose the `podman-env` command when possible.
We are here still waiting on a usable client as the 2.0.x release did not provide the v2 API (only v2.1).

guillaumerose:
Yes sure, I am just trying to find new ideas/solutions for the future :)

deboer-tim:
I agree it would be nice to have levels or templates, primarily since that would allow us to provide the full version and remove 'missing' capability, while still offering a developer option that's lighter than today.

Personally I'd put the (dev) console in developer, but we could close on details later. :)

guillaumerose:
I made an attempt in the tray. It fits better than a flag like `--only-podman` or `--enable-cluster-monitoring`. 

Screenshot 2021-03-11 at 21 12 51
Screenshot 2021-03-11 at 21 14 03

gbraad:
But can we move the discussion out of the PR and on to the issue perhaps?

gbraad:
Let's continue the discussion here: https://github.com/code-ready/crc/issues/1611 as we have discussed the saeme scenarios before in: https://github.com/code-ready/crc/issues/1097 with agreement of Scott. I believe we need to handle the start and detect the state of the cluster in a better fashion before continuing work. This is why I put this PR on HOLD (and closed now).

PR #1913, opened 2021-02-01
Update libvirt and hyperkit drivers

guillaumerose:
libvirt driver - https://github.com/code-ready/machine-driver-libvirt/releases/tag/0.12.14:
* Logger is now a regular logrus instance
* Redirect console to the VM log file
* Use MiB instead of MB as unit for memory size

hyperkit driver - https://github.com/code-ready/machine-driver-hyperkit/releases/tag/v0.12.11: 
* Logger is now a regular logrus instance


cfergeau (approved):
`logger is now a regular logrus instance` could be put in the full log rather than shoehorned in the one-line shortlog

Issue #1909, opened 2021-01-29
crc status -o json doesn't answer json


guillaumerose:
### General information

  * OS: Linux

## CRC version

master

### Steps to reproduce

```
$ crc status -o json   
Machine 'crc' does not exist. Use 'crc start' to create it

$ crc console -o json
{
  "success": false,
  "error": "Machine 'crc' does not exist. Use 'crc start' to create it"
}
```

First command should answers json like the second.

gbraad:
I would call this a high priority issue

Issue #1102, opened 2020-03-13
crc delete --clear-cache should work when a crc vm doesn't exist


anjannath:
```
$ tree ~/.crc/cache
/home/anjan/.crc/cache
└── crc_libvirt_4.3.0
    ├── crc-bundle-info.json
    ├── crc.qcow2
    ├── id_rsa_crc
    ├── kubeadmin-password
    └── kubeconfig

1 directory, 5 files

~/github.com/code-ready/crc master*
$ crc status
Machine 'crc' does not exist. Use 'crc start' to create it.

$ crc delete --clear-cache
Machine 'crc' does not exist. Use 'crc start' to create it.

$ tree ~/.crc/cache       
/home/anjan/.crc/cache
└── crc_libvirt_4.3.0
    ├── crc-bundle-info.json
    ├── crc.qcow2
    ├── id_rsa_crc
    ├── kubeadmin-password
    └── kubeconfig

1 directory, 5 files
```

PR #1918, opened 2021-02-01
Fix status command for json output

praveenkumar:
**Fixes:** Issue #1909 

Before
```
$ ./crc status -o json
Machine 'crc' does not exist. Use 'crc start' to create it
```

Now
```
$ ./crc status -ojson
{
  "success": false,
  "error": "Machine 'crc' does not exist. Use 'crc start' to create it"
}
```

### DIFF
--- a/cmd/crc/cmd/delete.go
+++ b/cmd/crc/cmd/delete.go
@@ -33,11 +33,10 @@ var deleteCmd = &cobra.Command{
 }
 
 func deleteMachine(client machine.Client, clearCache bool, cacheDir string, interactive, force bool) (bool, error) {
-	if !interactive && !force {
-		return false, errors.New("non-interactive deletion requires --force")
-	}
-
 	if clearCache {
+		if !interactive && !force {
+			return false, errors.New("non-interactive deletion requires --force")
+		}
 		yes := input.PromptUserForYesOrNo("Do you want to delete the OpenShift cluster cache", force)
 		if yes {
 			_ = os.RemoveAll(cacheDir)
@@ -46,6 +45,10 @@ func deleteMachine(client machine.Client, clearCache bool, cacheDir string, inte
 
 	if err := checkIfMachineMissing(client); err != nil {
 		return false, err
+	}
+
+	if !interactive && !force {
+		return false, errors.New("non-interactive deletion requires --force")
 	}
 
 	yes := input.PromptUserForYesOrNo("Do you want to delete the OpenShift cluster", force)
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -8,6 +8,7 @@ import (
 	"text/tabwriter"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	crcErrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/machine"
 	"github.com/docker/go-units"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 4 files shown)


cmd/crc/cmd/delete.go
```diff
@@ -33,10 +33,6 @@ var deleteCmd = &cobra.Command{
 }
 
 func deleteMachine(client machine.Client, clearCache bool, cacheDir string, interactive, force bool) (bool, error) {
-	if !interactive && !force {
-		return false, errors.New("non-interactive deletion requires --force")
-	}
-
 	if clearCache {
 		yes := input.PromptUserForYesOrNo("Do you want to delete the OpenShift cluster cache", force)
```
cfergeau, line 37: Before this `PromptUserForYesOrNo` line, the `!interactive && !force` check is required. It's probably the `checkIfMachineMissing(client)` check that you want to move earlier in this function.

gbraad (approved):

cfergeau (commented):
```
If VM is not exist and user try to do `crc delete -ojson`, it gets
something following which means user need to again run it with `-f`
option even the vm doesn't exist where the error message will change
to `Machine not exist, use 'crc start' to create it`.
```

would probably be easier to read if split in several sentences :)

cmd/crc/cmd/delete.go
praveenkumar: @cfergeau A user can clear the cache which is `~/.crc/cache` even the VM is not available so that's the reason I didn't move the machine check before.

cfergeau: But `clearCache` is calling `input.PromptUserForYesOrNo`. Isn't it required to have the `!interactive && !force` check **before** any call to `input.PromptUserForYesOrNo`?

praveenkumar: @cfergeau yes, looks like it is like chicken-egg problem, I will take your suggestion to move the machine check in the beginning so at least it is not broken, just a trade-off.

cfergeau: Or you could do the `!interactive && !force` check multiple times maybe?

guillaumerose (approved):

cmd/crc/cmd/delete.go
```diff
@@ -44,10 +48,6 @@ func deleteMachine(client machine.Client, clearCache bool, cacheDir string, inte
 		}
 	}
 
-	if err := checkIfMachineMissing(client); err != nil {
-		return false, err
-	}
-
```
cfergeau, line 50: You were right when pointing out that we want `crc delete --clear-cache` to work without a preexisting machine, this was fixed once in `Issue #1102 Make clear-cache work when there is no existing crc vm`

cmd/crc/cmd/delete.go
praveenkumar: @cfergeau I will try to do something like so if user provide `--clear-cache` flag we can explicit check the interactive and force flags otherwise we will do it after machine exist check.

```golang
	if clearCache {
	        if !interactive && !force {
		        return false, errors.New("non-interactive deletion requires --force")
	        }
		yes := input.PromptUserForYesOrNo("Do you want to delete the OpenShift cluster cache", force)
		if yes {
			_ = os.RemoveAll(cacheDir)
		}
	}

	if err := checkIfMachineMissing(client); err != nil {
		return false, err
	}

        if !interactive && !force {
		return false, errors.New("non-interactive deletion requires --force")
	}
       yes := input.PromptUserForYesOrNo("Do you want to delete the OpenShift cluster", force)
	if yes {
		return true, client.Delete()
	}
	return false, nil

```

cmd/crc/cmd/delete.go
guillaumerose: It's weird to have clear-cache flag under delete and not cleanup 🤔

cfergeau: `delete --clear-cache` was probably added before `cleanup`, but good point.

cfergeau (approved):
Looks good to me!

PR #1919, opened 2021-02-01
Update bundle to 4.6.13 from 4.6.12

praveenkumar:


Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION = 4.6.12
+BUNDLE_VERSION = 4.6.13
```
cfergeau, line 3: Ah, misread the commit log, I thought this was a downgrade from 4.6.13 to 4.6.12, and I was puzzled why we are doing this. But It's just my eyes not reading correctly!!

cfergeau (approved):

gbraad (approved):

PR #1921, opened 2021-02-01
Daemon should exit with an error if it can't accept connections

guillaumerose:
This is the default behavior of most servers (http.Server, etc.)

---

https://golang.org/src/net/http/server.go?s=45531:45575 look for Accept()

### DIFF
--- a/pkg/crc/api/api.go
+++ b/pkg/crc/api/api.go
@@ -41,8 +41,11 @@ func (api Server) Serve() error {
 	for {
 		conn, err := api.listener.Accept()
 		if err != nil {
-			logging.Error("Error establishing communication: ", err.Error())
-			continue
+			if neterr, ok := err.(net.Error); ok && neterr.Temporary() {
+				logging.Errorf("accept temporary error: %v", err)
+				continue
+			}
+			return err
 		}
 		api.handleConnections(conn) // handle version, status, webconsole, etc. requests
 	}


gbraad (approved):

gbraad:
We might have a problem mitigation too... like assignment of another port?

pkg/crc/api/api.go
```diff
@@ -41,8 +41,7 @@ func (api Server) Serve() error {
 	for {
 		conn, err := api.listener.Accept()
 		if err != nil {
-			logging.Error("Error establishing communication: ", err.Error())
-			continue
+			return err
```
anjannath, line 44: this is going to crash the daemon process if an error occurs here, won't this make it less fault tolerant, because maybe one `Accept` failed because it was timeout or temporary error and the next ones would succeed? looking at the `http.Serve()` they do check the done channel and if the error is a temporary one and if so adds some delay to accept the next request, they don't directly exit if `Accept()` return an error.

Is the motivation to make the daemon send this error to segment, if we want to achieve that maybe instead of logging we can push it to segment and not exit?

guillaumerose: I changed it to continue on temporary errors. 

For others errors, I would prefer a real crash: yes it will report to segment but also, it will be more obvious for the user that something is wrong. The error is not just a line in a log/stderr.

cfergeau: Feels like proper error reporting belongs on the client side, not on the daemon side. "Could not connect to api server/daemon/...", even if it's an improvement to start by catching this here.

cfergeau (approved):

Issue #1923, opened 2021-02-02
[Feature] Check if crc daemon process is reachable in case the network mode set as vsock


praveenkumar:
On all the platforms we need `crc daemon` to be running to work with vsock, It would be great if we can detect it as part preflight if the network mode is `vsock` ?

gbraad:
s/running/reachable...

gbraad:
The reason for reachable is a possible cause of it not running; at the moment we do not perform any checks if we can bind to these ports. They might be in use by something else. So we need to check if the port is reachable (and followed by a process or API check). If no expected response, we might have to start the daemon... but if this fails, we need to act on the error. The problem might be that these ports are already in use and need to be suffixed or offset for the new process.

PR #1924, opened 2021-02-02
Add check if crc daemon is running and bind to default endpoints

praveenkumar:
If user set the network mode as vsock then `crc daemon` needs to
be running as a backgroud process or in different terminal. This
is something user might forget and direct start the crc after setup.

This patch address it for windows and linux since those platform
have some preflight check for vsock side.

```
$ crc setup
[...]
INFO Setting up vsock support
'crc daemon' is not running, run it manually on different terminal/tab
```

fixes: #1923


pkg/crc/preflight/preflight_checks_nondarwin.go
```diff
@@ -0,0 +1,25 @@
...
+package preflight
+
+import (
+	"errors"
+
+	"github.com/code-ready/gvisor-tap-vsock/pkg/transport"
+)
+
+func checkIfCRCDaemonRunning() error {
+	l, err := transport.Listen(transport.DefaultURL)
```
guillaumerose, line 12: It seems a little bit brutal. Perhaps try to connect to the daemon with the crc.sock (the binary protocol) or to transport.DefaultURL with http.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -140,6 +140,10 @@ func checkVsock() error {
 }
 
 func fixVsock() error {
+	// first check if crc daemon is running
```
cfergeau, line 143: Nit: Since the next line uses a helper with a nice name (`checkIfCRCDaemonRunning`), this comment is redundant, code reading is smoother if you remove it

gbraad:
> This patch address it for windows and linux since those platform
> have some preflight check for vsock side.

Is there an issue to implement on Darwin/macOS?

gbraad (approved):

guillaumerose:
Yes it should work on darwin with the socket you use.

gbraad:
@praveenkumar would you be able to modify this PR for macOS? As suggested this should be an easy test... else, a follow issue is possible.

praveenkumar:
@guillaumerose @gbraad I changed the logic a bit so can be work on all the platform and also exit as soon as user try to use `vsock` mode.

gbraad (approved):

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -174,3 +175,19 @@ func removeOldLogs() error {
 	}
 	return nil
 }
+
+func checkIfCRCDaemonRunning() error {
+	l, err := net.Dial("unix", constants.DaemonSocketPath)
+	if err != nil {
+		// Log or report the error here
+		return errors.New("Network mode 'vsock' requires 'crc daemon' to be running, run it manually on different terminal/tab")
```
cfergeau, line 183: `in a different terminal`, not sure it's worth mentioning tabs

cfergeau (approved):

pkg/crc/preflight/preflight.go
```diff
@@ -163,6 +163,11 @@ func StartPreflightChecks(config config.Storage) error {
 func SetupHost(config config.Storage) error {
 	experimentalFeatures := config.Get(cmdConfig.ExperimentalFeatures).AsBool()
 	mode := network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString())
+	if mode == network.VSockMode {
```
guillaumerose, line 166: I don't think it should be here. Preflight checks configure the host so that the daemon can run. 
If vsock is turned off, daemon will not start. And with this if, it will be impossible to enable it with crc setup.
Perhaps just put it near crc start?

praveenkumar: @guillaumerose Does `crc daemon` also need `crc setup` before ?

guillaumerose: Yes

pkg/crc/preflight/preflight.go
```diff
@@ -153,6 +153,11 @@ func doRegisterSettings(cfg config.Schema, checks []Check) {
 func StartPreflightChecks(config config.Storage) error {
 	experimentalFeatures := config.Get(cmdConfig.ExperimentalFeatures).AsBool()
 	mode := network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString())
+	if mode == network.VSockMode {
```
guillaumerose, line 156: Is there a reason to not use a regular preflight check for this ?

praveenkumar: @guillaumerose then we need to add a network check on the mac since there is no network checks on the mac side.

cfergeau (approved):
Looks good to me, this apparently needs rebase?

PR #1925, opened 2021-02-02
Small fixes to data sent to segment

cfergeau:
This is a follow-up to https://github.com/code-ready/crc/pull/1885 to
convert Memory/DiskSize to bytes before sending them, and to avoid some
false negatives in segment.isProxySet().
   
Testing:

1. check that Memory and DiskSize are sent to segment in bytes
2. check that IsProxy is true in segment when proxy is configured through `http_proxy`/`https_proxy`

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -66,9 +66,9 @@ func runStart(ctx context.Context) (*machine.StartResult, error) {
 		return nil, err
 	}
 
-	for _, key := range []string{cmdConfig.CPUs, cmdConfig.Memory, cmdConfig.DiskSize} {
-		telemetry.SetContextProperty(ctx, key, config.Get(key).AsInt())
-	}
+	telemetry.SetContextProperty(ctx, cmdConfig.CPUs, config.Get(cmdConfig.CPUs).AsInt())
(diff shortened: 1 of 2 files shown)


cmd/crc/cmd/start.go
```diff
@@ -66,9 +66,9 @@ func runStart(ctx context.Context) (*machine.StartResult, error) {
 		return nil, err
 	}
 
-	for _, key := range []string{cmdConfig.CPUs, cmdConfig.Memory, cmdConfig.DiskSize} {
-		telemetry.SetContextProperty(ctx, key, config.Get(key).AsInt())
-	}
+	telemetry.SetContextProperty(ctx, cmdConfig.CPUs, config.Get(cmdConfig.CPUs).AsInt())
+	telemetry.SetContextProperty(ctx, cmdConfig.Memory, uint64(config.Get(cmdConfig.Memory).AsInt())*1024*104)
```
praveenkumar, line 70: @cfergeau shouldn't this `1024*1024` ?

cfergeau: Thanks! :) /me hides

guillaumerose (approved):

Issue #1903, opened 2021-01-29
Aborting bundle extraction leads to cache corruption


guillaumerose:
### General information

OS: All

## CRC version

All

### Steps to reproduce

```
$ crc setup
INFO Checking if CRC bundle is extracted in '$HOME/.crc' 
INFO Extracting bundle from the CRC executable    
crc.qcow2: 163.09 MiB / 9.85 GiB [-->______________________________________________________________________________________________________________________________________________________________________________________] 1.62%^C

$ crc setup
Setup is complete, you can now run 'crc start' to start the OpenShift cluster

$ du -h ~/.crc/cache/crc_libvirt_4.6.9
166M	/home/guillaumerose/.crc/cache/crc_libvirt_4.6.9

$ crc start -p ~/pull-secret.txt
...
INFO Checking if libvirt 'crc' network is active  
INFO Loading bundle: crc_libvirt_4.6.9.crcbundle ... 
INFO Verifying bundle crc_libvirt_4.6.9.crcbundle ... 
Invalid bundle crc_libvirt_4.6.9.crcbundle: oc not found in bundle

```

Then the VM cannot start correctly.

cfergeau:
I happen to have been working on a PR which would address this among other things.

cfergeau:
Fixed in 1.22

PR #1926, opened 2021-02-02
 #1903 aborting extract doesn't corrupt cache directory

guillaumerose:
Previously, when a user interrupted crc during the extraction, the
next crc start command failed because of corrupted bundle.
Now, the bundle is extracted to a temporary directory and at the end, if
everything is ok, it is copied to the final destination.

The directory structure of the archive is discarded. The bundle name is
built based on the metadata of the bundle.

### DIFF
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -102,8 +102,20 @@ func (bundle *CrcBundleInfo) createSymlinkOrCopyOpenShiftClient(ocBinDir string)
 }
 
 func (repo *Repository) Extract(path string) error {
-	_, err := extract.Uncompress(path, repo.CacheDir, true)
-	return err
+	bundleName := filepath.Base(path)
+
+	tmpDir := filepath.Join(repo.CacheDir, "tmp-extract")
+	_ = os.RemoveAll(tmpDir) // clean up before using it
+	defer func() {
+		_ = os.RemoveAll(tmpDir) // clean up after using it
+	}()
+
+	if _, err := extract.Uncompress(path, tmpDir, true); err != nil {
+		return err
+	}
+
+	bundleDir := strings.TrimSuffix(bundleName, bundleExtension)
+	return os.Rename(filepath.Join(tmpDir, bundleDir), filepath.Join(repo.CacheDir, bundleDir))
 }
 
 var defaultRepo = &Repository{


pkg/crc/machine/bundle/repository.go
```diff
@@ -102,10 +105,23 @@ func (bundle *CrcBundleInfo) createSymlinkOrCopyOpenShiftClient(ocBinDir string)
 }
 
 func (repo *Repository) Extract(path string) error {
-	_, err := extract.Uncompress(path, repo.CacheDir, &extract.Options{
+	dir := filepath.Join(repo.CacheDir, "tmp-extract")
+	_ = os.RemoveAll(dir) // clean up before using it
```
cfergeau, line 109: I'd use `ioutil.TempDir()` + `defer os.RemoveAll()`

pkg/crc/machine/bundle/repository.go
```diff
@@ -102,10 +105,23 @@ func (bundle *CrcBundleInfo) createSymlinkOrCopyOpenShiftClient(ocBinDir string)
...
+		return err
+	}
+	bundle, err := readBundleInfo(dir)
+	if err != nil {
+		return err
+	}
+	if err := bundle.Verify(); err != nil {
+		return err
+	}
+	return os.Rename(dir, filepath.Join(repo.CacheDir, bundle.GetName()))
```
cfergeau, line 124: > The directory structure of the archive is discarded. The bundle name is built based on the metadata of the bundle.

I don't think it's mandatory to do that in this commit, the bundle directory can be renamed to ` 	path := filepath.Join(repo.CacheDir, strings.TrimSuffix(bundleName, bundleExtension))` as was done before. The use of the bundle name can be done in a separate commit to avoid mixing bug fixes with functional changes.

pkg/extract/extract.go
```diff
@@ -24,6 +24,7 @@ const minSizeForProgressBar = 100_000_000
 type Options struct {
 	ShowProgress bool
 	FileFilter   func(string) bool
+	Flatten      bool
```
praveenkumar, line 27: @guillaumerose I am not able to get why this is required and what is the use of this (as of now I don't see it in used excepted the testing part). Can you please add bit description on the commit log around same?

pkg/crc/machine/bundle/repository.go
guillaumerose: yes good point!

pkg/crc/machine/bundle/repository.go
guillaumerose: If I use a tmp dir given by the fs, it is not guaranteed to be on the same disk as the cache directory.
If I use `ioutil.TmpDir(repo.CacheDir, "")` and hit ctrl+c, crc can't delete them as the name is random. 

I kept this :/

pkg/extract/extract.go
guillaumerose: yes I simplified it. I will bring the bundle name in the future more globally.

cfergeau (approved):

PR #1931, opened 2021-02-02
Improve bundle checks/bundle extraction

cfergeau:
This pull request fixes various things related to bundle extraction/validation.
At `setup` time, for release builds, it now checks that the default bundle is uncompressed and valid.
If not, it extracts the bundle from the binary if needed and then uncompresses it.

At `start` time, it's now checking more accurately that the bundle is uncompressed and valid before trying to use it.
This should fix some corner cases with partially uncompressed bundles/failures when the `.crcbundle` file is removed while the uncompressed bundle is present, ...

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -10,7 +10,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 	"github.com/code-ready/crc/pkg/crc/validation"
-	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/code-ready/crc/pkg/embed"
 	"github.com/docker/go-units"
 	"github.com/pkg/errors"
@@ -61,7 +60,7 @@ var genericPreflightChecks = [...]Check{
 }
 
 func checkBundleExtracted() error {
-	if !version.IsRelease() {
+	if !constants.IsRelease() {
 		return nil
 	}
 	if _, err := os.Stat(constants.DefaultBundlePath); os.IsNotExist(err) {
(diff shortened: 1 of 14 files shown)


pkg/crc/validation/validation.go
```diff
@@ -75,6 +76,16 @@ func ValidateBundlePath(bundle string) error {
 	return nil
 }
 
+func ValidateBundle(bundlePath string) error {
+	bundleName := filepath.Base(bundlePath)
+	_, err := bundle.GetCachedBundleInfo(bundleName)
```
praveenkumar, line 81: This check is just a feel good one because if it error occur we are going to run `ValidateBundlePath` so I am not able to figure out what advantage it is going to add.

cfergeau: With this change, if for some reason `~/.crc/cache/crc_libvirt_4.6.9` exists but `~/.crc/cache/crc_libvirt_4.6.9.crcbundle` does not exist, we can still successfully run `crc start`. With current crc master, `~/.crc/cache/crc_libvirt_4.6.9.crcbundle` must exist even if the uncompressed bundle is present.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -61,11 +62,15 @@ var genericPreflightChecks = [...]Check{
 
 func checkBundleExtracted() error {
 	if !constants.IsRelease() {
+		logging.Debugf("Development build, skipping check")
 		return nil
 	}
-	if _, err := os.Stat(constants.DefaultBundlePath); os.IsNotExist(err) {
+	logging.Infof("Checking if %s exists", constants.DefaultBundlePath)
```
guillaumerose, line 68: With this new check, we can now delete the temporary .crcbundle file

cfergeau:
I've added this to the commti log to try to make it clearer:
> The main benefit of this new ValidateBundle is that the `.crcbundle`  files in `~/.crc/cache` become optional and can be removed after they've been uncompressed.

PR #1932, opened 2021-02-03
Add containerized build using go-toolset

### LINKED ISSUE #1766: Build crc in a container
In order to use OpenShift CI for crc basic checks (lint, test and compilation), we need to create a Dockerfile for crc.

It will help us to move out of Travis.

gbraad:
Fixes #1766; Initial container build

Seperated the git install step as this would re-use the containers between routes-controller, gvisor-tap-vsock, etc.

---
```
$ make containerized
podman build -t crc-build .
STEP 1: FROM registry.access.redhat.com/ubi8/ubi AS build
STEP 2: WORKDIR /go/src/app
--> Using cache b1fe4c0c69acdbe1480a0430c23a457d85e5f4288d8114e44d09e13418d8c551
--> b1fe4c0c69a
STEP 3: RUN yum -y install golang make
--> Using cache 047ae8989bf2d77d6d07e2901e4f5aa3134daeeceb009aff92cff29be18e79e7
--> 047ae8989bf
STEP 4: RUN yum -y install git-core
--> Using cache 33f530f208bc13a01423b813aee9f8cd7415fdc07a24c081cef421143119fda1
--> 33f530f208b
STEP 5: COPY . .
STEP 6: COMMIT crc-build
--> cbd377eb88a
cbd377eb88ab037e34646e8e72006717436783a310b2eead654451c231a0c967
podman run --name crc-cross crc-build make cross
GOARCH=amd64 GOOS=darwin go build -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=1.21.0 -X github.com/code-ready/crc/pkg/crc/version.bundleVersion=4.6.13 -X github.com/code-ready/crc/pkg/crc/version.commitSha=df022a04 -extldflags='-static' -s -w" -o out/macos-amd64/crc ./cmd/crc
GOOS=linux GOARCH=amd64 go build -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=1.21.0 -X github.com/code-ready/crc/pkg/crc/version.bundleVersion=4.6.13 -X github.com/code-ready/crc/pkg/crc/version.commitSha=df022a04 -extldflags='-static' -s -w" -o out/linux-amd64/crc ./cmd/crc
GOARCH=amd64 GOOS=windows go build -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=1.21.0 -X github.com/code-ready/crc/pkg/crc/version.bundleVersion=4.6.13 -X github.com/code-ready/crc/pkg/crc/version.commitSha=df022a04 -extldflags='-static' -s -w" -o out/windows-amd64/crc.exe ./cmd/crc
podman cp crc-cross:/go/src/app/out ./
podman rm crc-cross
61c5095fc4502f6a2518374605234a58c7e58c2597173662856fbcd321f9495e
podman rmi crc-build
Untagged: localhost/crc-build:latest
Deleted: cbd377eb88ab037e34646e8e72006717436783a310b2eead654451c231a0c967
```

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,19 @@
+stages:
+  - build
+
+containerized-cross-build:
+  image: registry.fedoraproject.org/fedora:latest
+  stage: build
+  script:
+    # setup base image to allow Makefile to run
+    - dnf install -y make git-core golang podman
+    # The following changes are needed to allow podman to run in podman/docker
+    - cp /usr/share/containers/containers.conf /etc/containers/containers.conf 
+    - sed -i '/^# cgroup_manager = "systemd"/ a cgroup_manager = "cgroupfs"' /etc/containers/containers.conf
+    - sed -i '/^# events_logger = "journald"/ a events_logger = "file"' /etc/containers/containers.conf
+    - sed -i '/^driver = "overlay"/ c\driver = "vfs"' /etc/containers/storage.conf
+    # run the actual containerized build
+    - make containerized
+  artifacts:
+    paths:
+      - out
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,14 @@ $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
 .PHONY: cross ## Cross compiles all binaries
 cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/windows-amd64/crc.exe
 
+.PHONY: containerized ## Cross compile from container
+containerized: clean
+	${CONTAINER_RUNTIME} build -t crc-build -f images/build .
+	${CONTAINER_RUNTIME} run --name crc-cross crc-build make cross
+	${CONTAINER_RUNTIME} cp crc-cross:/opt/app-root/src/out ./
+	${CONTAINER_RUNTIME} rm crc-cross
+	${CONTAINER_RUNTIME} rmi crc-build
+
 .PHONY: test
 test:
 	go test -race --tags build -v -ldflags="$(VERSION_VARIABLES)" ./pkg/... ./cmd/...
--- a/images/build/Dockerfile
+++ b/images/build/Dockerfile
@@ -1,0 +1,5 @@
+FROM registry.access.redhat.com/ubi8/go-toolset:1.14.12
+MAINTAINER CodeReady Containers <devtools-cdk@redhat.com>
+
+WORKDIR $APP_ROOT/src
+COPY . .


gbraad:
Seperated the git install step as this would re-use the containers between routes-controller, gvisor-tap-vsock, etc. 

it however does not copy the artifacts out yet.

cfergeau:
Some explanations why we want to do this, who is the target for these container builds, ... would be nice.

gbraad:
@guillaumerose any suggestions about re-use between CI and local use?

container-build.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+CMD=${1-docker}
+
+$CMD build -t crc-build .
```
guillaumerose, line 9: docker build can return the id of the image so you don't need to name it and rmi after.

container-build.sh
```diff
@@ -0,0 +1,13 @@
...
+$CMD run --name crc-cross crc-build make cross
```
guillaumerose, line 10: docker run --rm

cfergeau (commented):
`machine-driver-libvirt` (and some OpenShift modules) use `CONTAINER_RUNTIME` to specify podman/docker/... And `podman` is a better default :stuck_out_tongue_closed_eyes:

guillaumerose:
No specific idea for a common base image. We can start with this!

gbraad:
> machine-driver-libvirt (and some OpenShift modules) use CONTAINER_RUNTIME to specify podman/docker/... And podman is a better default 😝

hadn't seen this

gbraad:
better to use the Makefile as a new target?

```
$ make containerized
```

or so...

gbraad:
Can successfully run on GitLab CI: https://gitlab.com/gbraad/crc/-/jobs/1014092836

.gitlab-ci.yml
```diff
@@ -0,0 +1,17 @@
+stages:
+  - build
+
+cross-build:
+  image: registry.fedoraproject.org/fedora:latest
+  stage: build
+  script:
+    - dnf install -y make git-core golang podman
```
gbraad, line 8: `golang` is needed for the variables that `Makefile` sets:

```
GOOS ?= $(shell go env GOOS)
HOST_BUILD_DIR=$(BUILD_DIR)/$(GOOS)-$(GOARCH)
GOPATH ?= $(shell go env GOPATH
```

`git-core` is needed for:
```
```
and
```
SOURCES := $(shell git ls-files  *.go ":^vendor"
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,17 @@
...
+    - cp /usr/share/containers/containers.conf /etc/containers/containers.conf 
+    - sed -i '/^# cgroup_manager = "systemd"/ a cgroup_manager = "cgroupfs"' /etc/containers/containers.conf
+    - sed -i '/^# events_logger = "journald"/ a events_logger = "file"' /etc/containers/containers.conf
+    - sed -i '/^mountopt =.*/d' /etc/containers/storage.conf
+    - sed -i '/^driver = "overlay"/ c\driver = "vfs"' /etc/containers/storage.conf
```
gbraad, line 13: overlay driver does not work when overlay is already being used.

.gitlab-ci.yml
```diff
@@ -0,0 +1,16 @@
+stages:
```
guillaumerose, line 1: I have a deal for you: if you add a CI, you remove one like travis. deal? :)

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM registry.access.redhat.com/ubi8/ubi AS build
+WORKDIR /go/src/app
+RUN yum -y install golang make
```
guillaumerose, line 3: I am wondering if we can be more selective on the version and take 1.14.

.gitlab-ci.yml
gbraad: This is not a public one per se, as gitlab runs internally and locally here.

Dockerfile
gbraad: Distribution provided, pretty much like the internal CI situation

gbraad: Will check the modules optiona they do provide streams for app development

gbraad: Would also apply to the other uses of the UBI image, like https://github.com/code-ready/gvisor-tap-vsock/blob/e5f886c34c9ff186c403557bad3c7533aa3275be/Dockerfile#L3

gbraad: Will have to check https://catalog.redhat.com/software/containers/ubi8/go-toolset/5ce8713aac3db925c03774d1?container-tabs=packages

gbraad: 'our use' is not the way this image is intended... as this go-toolset is for s2i.

gbraad: Also changed https://github.com/code-ready/routes-controller/pull/4/

gbraad: https://github.com/code-ready/gvisor-tap-vsock/pull/10

Makefile
```diff
@@ -100,6 +100,14 @@ $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
 .PHONY: cross ## Cross compiles all binaries
 cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/windows-amd64/crc.exe
 
+.PHONY: containerized ## Cross compile from container
+containerized: clean
+	${CONTAINER_RUNTIME} build -t crc-build .
```
cfergeau, line 105: I should have double checked before mentioning `CONTAINER_RUNTIME`, but `RUNTIME` seems more common than `CONTAINER_RUNTIME` in OpenShift components, https://github.com/openshift/enhancements/pull/638 for one such case (though the only other one is `api`)

gbraad: We already used this variable for the docs build container. just re-using it here

Dockerfile
```diff
@@ -1,5 +1,3 @@
-FROM registry.access.redhat.com/ubi8/ubi AS build
-WORKDIR /go/src/app
-RUN yum -y install golang make
-RUN yum -y install git-core
+FROM registry.access.redhat.com/ubi8/go-toolset:1.14.12
```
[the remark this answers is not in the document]
gbraad: we can not specify a 1.14 image (does not exist)

Makefile
gbraad: Also, I am not a huge fan of just calling this `${RUNTIME}` as it is too generic... I can't say `JRE` here, right? ;-)

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM registry.access.redhat.com/ubi8/ubi AS build
```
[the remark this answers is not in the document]
gbraad: That is an inconsistency that is everywhere. will have a look at this

Makefile
```diff
@@ -100,10 +100,10 @@ $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
 cross: $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/windows-amd64/crc.exe
 
 .PHONY: containerized ## Cross compile from container
-containerized:
+containerized: clean
 	${CONTAINER_RUNTIME} build -t crc-build .
```
cfergeau: Still feels that this could get squashed in the initial Dockerfile addition?

images/build/Dockerfile
```diff
@@ -1,3 +1,5 @@
 FROM registry.access.redhat.com/ubi8/go-toolset:1.14.12
+MAINTAINER CodeReady Containers <devtools-cdk@redhat.com>
+
```
cfergeau, line 3: same for this, could be squashed in initial commit?

cfergeau (approved):
I'd squash some of the commits, otherwise looks good to me.

images/build/Dockerfile
gbraad: does this prevent approval?

Makefile
gbraad: this is not squashed in that commit as it handles the `COPY` that would else push `out` into the image, but being prevents the build due to a Permission denied, as the go-toolset image uses `1001` as the yuser instead of `root`,

images/build/Dockerfile
cfergeau: I'd prefer to have these squashed, but not going to fight for it. At the moment it makes the history messy

```
Move container to images folder

+ MAINTAINER CodeReady Containers <devtools-cdk@redhat.com>
```
does not make a lot of sense.

Makefile
gbraad: https://catalog.redhat.com/software/containers/ubi8/go-toolset/5ce8713aac3db925c03774d1?container-tabs=dockerfile&gti-tabs=unauthenticated

cfergeau: But you add a new Containerfile which initially uses the ubi8 image, and subsequently moves to use the go-toolset image. I don't think it's very valuable to record that change in our history, this new file could use the go-toolset image from the start? Especially as the 'move to go-toolset' commit has no commit log, which would indicate it's a non-interesting change?

PR #1936, opened 2021-02-03
[Segment] Capture crc run on tty/non-tty environment

praveenkumar:
We already have checks around interactive and terminal execution
which help user to run the crc on a CI environment if needed.
We should capture that data as part of telemeter to identify
how many users are using CRC for non-tty case (mostly be the CI).

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -18,6 +18,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/pborman/uuid"
 	"github.com/segmentio/analytics-go"
+	terminal "golang.org/x/term"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/crc/segment/segment.go
```diff
@@ -81,7 +82,8 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 
 	properties = properties.Set("version", version.GetCRCVersion()).
 		Set("success", err == nil).
-		Set("duration", duration.Milliseconds())
+		Set("duration", duration.Milliseconds()).
+		Set("tty", terminal.IsTerminal(int(os.Stdin.Fd())))
```
guillaumerose, line 86: Elegant solution :+1:

guillaumerose (approved):

PR #1938, opened 2021-02-03
Store the pull secret manually entered by the user in the keyring

guillaumerose:
If the user has a keyring configured, crc uses it to store the pull
secret.
The user won't have to enter the pull secret for each start now.

The pull secret is stored compressed. Windows doesn't support secret
larger than 2500 chars.

## Testing

1. crc start (it asks for the pull secret)
2. crc delete
3. crc start (it doesn't ask the pull secret)


praveenkumar:
/lgtm

Since we now have #1944 as follow up issue

PR #1939, opened 2021-02-03
Fix ValidateBool function to accept valid range of boolean

praveenkumar:
Till now we are casting the user input to string and the comparing
it with `true/false` but in golang accepts 1, t, T, TRUE, true, True, 0,
f, F, FALSE, false, False as boolean when parse it so all those are
valid boolean value. With this patch `cast.ToBoolE` is used to get
the error if casting not happen properly.

### DIFF
--- a/pkg/crc/config/validations.go
+++ b/pkg/crc/config/validations.go
@@ -13,10 +13,11 @@ import (
 // ValidateBool is a fail safe in the case user
 // makes a typo for boolean config values
 func ValidateBool(value interface{}) (bool, string) {
-	if cast.ToString(value) == "true" || cast.ToString(value) == "false" {
-		return true, ""
+	if _, err := cast.ToBoolE(value); err != nil {
+		return false, "must be true or false"
 	}
-	return false, "must be true or false"
+
+	return true, ""
 }
 
 // ValidateDiskSize checks if provided disk size is valid in the config


pkg/crc/config/validations.go
```diff
@@ -13,10 +13,11 @@ import (
 // ValidateBool is a fail safe in the case user
 // makes a typo for boolean config values
 func ValidateBool(value interface{}) (bool, string) {
-	if cast.ToString(value) == "true" || cast.ToString(value) == "false" {
-		return true, ""
+	if _, err := cast.ToBoolE(value); err != nil {
+		return false, "must be 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False"
```
cfergeau, line 17: This is not very user friendly, I would only list one valid pair (`true/false`?). The fact that more values can be used is nice, can be added to our documentation if we think it's important, but here it's too much information to have everything.

cfergeau (commented):
The commit log is slightly off the mark. The problem we want to solve is a regression after commit e34584f0fe3 which moved validation before casting. For booleans, the validation code only accepts 'true/false', while the casting code accepts more values because it uses `strconv.ParseBool`. This caused a regression (`crc config set skip-xxx 1` no mlonger works for example). This commit makes `ValidateBool` use `cast.ToBoolE` so that it accepts the same inputs as the casting code.

guillaumerose (approved):

pkg/crc/config/validations.go
gbraad: the other options is an additional... and is more to be contextually correct, ike saying "yes" or "no".

gbraad (approved):

PR #1949, opened 2021-02-04
#1948 don't use viper to write the config file

### LINKED ISSUE #1948: [BUG] Unwanted config key in crc.json after crc setup
### General information

  * OS: All

### Steps to reproduce

With an empty ~/.crc directory
  1. crc setup

### Expected

```
$ cat ~/.crc/crc.json 
{
  "consent-telemetry": "no",
}          
```

### Actual

```
$ cat ~/.crc/crc.json 
{
  "consent-telemetry": "no",
  "enable-experimental-features": false,
  "help": false,
  "log-level": "info",
  "output": ""
}       
```

guillaumerose:
Viper is not made for writing config file. When using flags, it adds
them in the config file even if they are not config key we defined.
    
This change uses traditional method to edit json and then, asks viper to
reload the configuration.

Fixes #1948

### DIFF
--- a/pkg/crc/config/viper_config.go
+++ b/pkg/crc/config/viper_config.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


cfergeau:
> Viper is not made for writing config file

Given that it has API for that, plus documentation around it https://github.com/spf13/viper/#writing-config-files , I beg to disagree with that statement. After reading the next sentence, I can agree with "viper.WriteConfig is not the right API for what we are trying to do" though :)

guillaumerose:
I reworded the commit log :)

pkg/crc/config/viper_config_test.go
```diff
@@ -181,13 +181,13 @@ func TestViperConfigBindFlagSet(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Equal(t, SettingValue{
-		Value:     6,
+		Value:     5, // flag has the priority
```
praveenkumar, line 184: @guillaumerose does that mean we were ignoring the flag priority over what present in config?

guillaumerose: no this is a weird case where someone would launch `crc config set cpus 8 --cpus 10`.  This is purely theoretical. Set a key and at the same time set it with flag.

praveenkumar: @guillaumerose for me, it leads to following error which make sense.
 
```
$ ./crc config set cpus 6 --cpus 10
unknown flag: --cpus
```

But when I am looking the test case I think we are adding the flags and only for test case it might be the case. Can't we just remove it? It will be confusing over time.

guillaumerose: I removed the assertion :+1:

cfergeau:
Fwiw, I think `When using flags, it adds them in the config file even if they are not config key we defined.` used to work fine (ie crc's code was being careful so that these flags don't end up in the config file - assuming by 'flags' you mean `--memory` and such)

pkg/crc/config/viper_config.go
```diff
@@ -45,21 +44,43 @@ func (c *ViperStorage) Get(key string) interface{} {
...
+	}
+	cfg[key] = value
+	bin, err := json.MarshalIndent(cfg, "", "  ")
+	if err != nil {
+		return err
+	}
+	if err := atomicWrite(bin, c.configFile); err != nil {
+		return err
+	}
+	return c.viper.ReadInConfig()
```
cfergeau, line 63: Adding more and more IO for each `Set`/`Unset`/... does not feel like they way to go to be honest :-/

pkg/crc/config/viper_config_test.go
```diff
@@ -183,6 +184,10 @@ func TestViperConfigBindFlagSet(t *testing.T) {
 		Value:     6,
 		IsDefault: false,
 	}, config.Get(CPUs))
+
+	bin, err := ioutil.ReadFile(configFile)
+	assert.NoError(t, err)
+	assert.JSONEq(t, `{"cpus":6,"extra":"","nameservers":""}`, string(bin))
```
cfergeau: Not clear if this is a 'theoritical' issue though, or if this can happen while using crc.

guillaumerose:
Set/Unset worked fine as long as you don't use them when you BindFlags. 
If you do crc start --memory 10000 and also do Set("cpus", 10) for an other reason, you end up with both memory and cpus changed in the config file.

pkg/crc/config/viper_config.go
guillaumerose: Yes.. I added as many tests as possible so that we can refactor that someday if we have better ideas.

cfergeau: Oh I have better ideas, but this requires significant work, so not quite realistic now ;) (real IPC signaling rather than using files for config change notifications, better eventloop in crc so that we can delay writing of the config file, ...)

PR #1958, opened 2021-02-05
Use regex to replace the carriage return to '\n' char

praveenkumar:
With this patch, carriage return for windows (`\r\n`) and unix ('\n')
replaced with `\n` char. Till now it was only replacing it for unix.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -214,8 +215,9 @@ func addProxyCACertToCluster(sshRunner *ssh.Runner, ocConfig oc.Config, proxy *n
   }
 }
 `
-	// Replace the carriage return with `\n` char
-	p := fmt.Sprintf(proxyCABundleTemplate, strings.ReplaceAll(proxy.ProxyCACert, "\n", `\n`), trustedCAName)
+	// Replace the carriage return ("\n" or "\r\n") with `\n` char
```
cfergeau, line 218: I realise now the comment is confusing? This replaces the carriage return with the litteral `\n` string (aka "\\n"),  not with a single 'char'. Could be a good opportunity to fix the comment.

cfergeau (approved):

guillaumerose:
Looks good but it would be even better if we built the config object programmatically. 


```
	json.Marshal(v1.ConfigMap{
		TypeMeta: metav1.TypeMeta{
			Kind:       "ConfigMap",
			APIVersion: "v1",
		},
		ObjectMeta: metav1.ObjectMeta{
			Name:      "???",
			Namespace: "openshift-config",
		},
		Data: map[string]string{
			"ca-bundle.crt": trustedCAName,
		},
	})

```

praveenkumar:
> Looks good but it would be even better if we built the config object programmatically.

@guillaumerose sure, I can take this up in following PR/issues.

guillaumerose (approved):

PR #1960, opened 2021-02-08
Add Help parameter for config setting struct

praveenkumar:
This patch will make use of help string for key and show output
as following.
```
* bundle                          Bundle path (string, default '/home/prkumar/.crc/cache/crc_libvirt_4.6.15.crcbundle')
* consent-telemetry               Send feedback about CodeReady Containers (yes/no)
* cpus                            Number of CPU cores (integer, default '4')
* disable-update-check            Don't check for update (default false)
* disk-size                       Total size in GiB of the disk (integer, default '31')
* enable-cluster-monitoring       Enable cluster monitoring operator (default false)
```

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -95,42 +95,19 @@ func less(lhsKey, rhsKey string) bool {
 }
 
 func configurableFields(config *config.Config) string {
-	var fields []string
+	settings := config.AllSettings()
+	sort.Slice(settings, func(i, j int) bool {
+		return less(settings[i].Name, settings[j].Name)
+	})
+
 	var buf bytes.Buffer
 	writer := tabwriter.NewWriter(&buf, 0, 8, 1, ' ', tabwriter.TabIndent)
-	for _, keyAndValueType := range keysAndValueType(config) {
-		fmt.Fprintln(writer, keyAndValueType)
+	for _, cfg := range settings {
+		fmt.Fprintf(writer, "* %s\t%s\n", cfg.Name, cfg.Help)
 	}
 	writer.Flush()
-	keys := strings.Split(buf.String(), "\n")
-	sort.Slice(keys, func(i, j int) bool {
-		return less(keys[i], keys[j])
-	})
-	for _, key := range keys {
-		if key == "" {
-			continue
-		}
-		fields = append(fields, " * "+key)
-	}
-	return strings.Join(fields, "\n")
-}
-func keysAndValueType(config *config.Config) []string {
-	var keyAndValueType []string
-	for key, value := range config.AllConfigs() {
-		var valueType string
-		switch value.Value.(type) {
-		case int:
-			valueType = "Number"
-		case string:
-			valueType = "String"
-		case bool:
-			valueType = "true/false"
-		default:
-			valueType = fmt.Sprintf("%T", value.Value)
-		}
-		keyAndValueType = append(keyAndValueType, fmt.Sprintf("%s\t%s", key, valueType))
-	}
-	return keyAndValueType
+
+	return buf.String()
 }
 
 func GetConfigCmd(config *config.Config) *cobra.Command {
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -33,25 +33,40 @@ const (
 
 func RegisterSettings(cfg *config.Config) {
 	// Start command settings in config
-	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundlePath, config.SuccessfullyApplied)
-	cfg.AddSetting(CPUs, constants.DefaultCPUs, config.ValidateCPUs, config.RequiresRestartMsg)
-	cfg.AddSetting(Memory, constants.DefaultMemory, config.ValidateMemory, config.RequiresRestartMsg)
-	cfg.AddSetting(DiskSize, constants.DefaultDiskSize, config.ValidateDiskSize, config.RequiresRestartMsg)
-	cfg.AddSetting(NameServer, "", config.ValidateIPAddress, config.SuccessfullyApplied)
-	cfg.AddSetting(PullSecretFile, "", config.ValidatePath, config.SuccessfullyApplied)
-	cfg.AddSetting(DisableUpdateCheck, false, config.ValidateBool, config.SuccessfullyApplied)
-	cfg.AddSetting(ExperimentalFeatures, false, config.ValidateBool, config.SuccessfullyApplied)
-	cfg.AddSetting(NetworkMode, string(network.DefaultMode), network.ValidateMode, network.SuccessfullyAppliedMode)
+	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundlePath, config.SuccessfullyApplied,
+		fmt.Sprintf("Bundle path (string, default '%s')", constants.DefaultBundlePath))
+	cfg.AddSetting(CPUs, constants.DefaultCPUs, config.ValidateCPUs, config.RequiresRestartMsg,
+		fmt.Sprintf("Number of CPU cores (must be greater than or equal to '%d')", constants.DefaultCPUs))
+	cfg.AddSetting(Memory, constants.DefaultMemory, config.ValidateMemory, config.RequiresRestartMsg,
+		fmt.Sprintf("Memory size in MiB (must be greater than or equal to '%d')", constants.DefaultMemory))
+	cfg.AddSetting(DiskSize, constants.DefaultDiskSize, config.ValidateDiskSize, config.RequiresRestartMsg,
+		fmt.Sprintf("Total size in GiB of the disk (must be greater than or equal to '%d')", constants.DefaultDiskSize))
+	cfg.AddSetting(NameServer, "", config.ValidateIPAddress, config.SuccessfullyApplied,
+		"IPv4 address of nameserver (string, like '1.1.1.1 or 8.8.8.8')")
+	cfg.AddSetting(PullSecretFile, "", config.ValidatePath, config.SuccessfullyApplied,
+		fmt.Sprintf("Path of image pull secret (download from %s)", constants.CrcLandingPageURL))
+	cfg.AddSetting(DisableUpdateCheck, false, config.ValidateBool, config.SuccessfullyApplied,
+		"Disable update check (true/false, default: false)")
+	cfg.AddSetting(ExperimentalFeatures, false, config.ValidateBool, config.SuccessfullyApplied,
+		"Enable experimental features (true/false, default: false)")
+	cfg.AddSetting(NetworkMode, string(network.DefaultMode), network.ValidateMode, network.SuccessfullyAppliedMode,
+		"Network mode (default or vsock)")
 	// Proxy Configuration
-	cfg.AddSetting(HTTPProxy, "", config.ValidateURI, config.SuccessfullyApplied)
-	cfg.AddSetting(HTTPSProxy, "", config.ValidateURI, config.SuccessfullyApplied)
-	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied)
-	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied)
+	cfg.AddSetting(HTTPProxy, "", config.ValidateURI, config.SuccessfullyApplied,
+		"HTTP proxy URL (string, like 'http://my-proxy.com:8443')")
+	cfg.AddSetting(HTTPSProxy, "", config.ValidateURI, config.SuccessfullyApplied,
+		"HTTPS proxy URL (string, like 'https://my-proxy.com:8443')")
+	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied,
+		"Hosts, ipv4 addresses or CIDR which do not use a proxy (string, comma-separated list such as '127.0.0.1,192.168.100.1/24')")
+	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied,
+		"Path to an HTTPS proxy certificate authority (CA)")
 
-	cfg.AddSetting(EnableClusterMonitoring, false, config.ValidateBool, config.SuccessfullyApplied)
+	cfg.AddSetting(EnableClusterMonitoring, false, config.ValidateBool, config.SuccessfullyApplied,
+		"Enable cluster monitoring Operator (true/false, default: false)")
 
 	// Telemeter Configuration
-	cfg.AddSetting(ConsentTelemetry, "", config.ValidateYesNo, config.SuccessfullyApplied)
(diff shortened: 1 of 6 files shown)


guillaumerose (approved):

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
-	cfg.AddSetting(PullSecretFile, "", config.ValidatePath, config.SuccessfullyApplied)
-	cfg.AddSetting(DisableUpdateCheck, false, config.ValidateBool, config.SuccessfullyApplied)
-	cfg.AddSetting(ExperimentalFeatures, false, config.ValidateBool, config.SuccessfullyApplied)
-	cfg.AddSetting(NetworkMode, string(network.DefaultMode), network.ValidateMode, network.SuccessfullyAppliedMode)
+	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundlePath, config.SuccessfullyApplied,
+		fmt.Sprintf("Bundle path (string, default '%s')", constants.DefaultBundlePath))
+	cfg.AddSetting(CPUs, constants.DefaultCPUs, config.ValidateCPUs, config.RequiresRestartMsg,
+		fmt.Sprintf("Number of CPU cores (integer, default '%d')", constants.DefaultCPUs))
+	cfg.AddSetting(Memory, constants.DefaultMemory, config.ValidateMemory, config.RequiresRestartMsg,
+		fmt.Sprintf("Memory size in MiB (integer, default '%d')", constants.DefaultMemory))
```
cfergeau, line 41: instead of just 'integer, default ...', I'd mention it must be equal or bigger than this default ?

pkg/crc/preflight/preflight.go
```diff
@@ -144,7 +144,7 @@ func doCleanUpPreflightChecks(checks []Check) error {
 func doRegisterSettings(cfg config.Schema, checks []Check) {
 	for _, check := range checks {
 		if check.configKeySuffix != "" {
-			cfg.AddSetting(check.getSkipConfigName(), false, config.ValidateBool, config.SuccessfullyApplied)
+			cfg.AddSetting(check.getSkipConfigName(), false, config.ValidateBool, config.SuccessfullyApplied, "true/false")
```
cfergeau, line 147: should this be described as "Skip preflight check (true/false, default: false)"?

pkg/crc/config/types.go
```diff
@@ -10,20 +10,22 @@ type Storage interface {
...
 	validationFn ValidationFnType
 	callbackFn   SetFn
+	Help         string
 }
 
 type SettingValue struct {
 	Value     interface{}
 	Invalid   bool
 	IsDefault bool
+	Help      string
```
cfergeau, line 28: For what it's worth, this feels odd to have in `SettingValue`. Have you tried adding a new `AllSettings()` helper which would return a `[]Settings`?

cmd/crc/cmd/config/config.go
```diff
@@ -94,7 +94,7 @@ func less(lhsKey, rhsKey string) bool {
 	return lhsKey < rhsKey
 }
 
-func configurableFields(config *config.Config) string {
+func configurableFields(config config.Storage) string {
```
cfergeau, line 97: Actually, `configurableFields` is more about `config.Schema` than `config.Storage`. But not really convenient to pass a `config.Schema` around, so I'd keep `config *config.Config`

pkg/crc/config/types.go
cfergeau: I've done this in https://github.com/cfergeau/crc/commit/ef1ef9b5e8ef97a9d3485452973d02c12929e538 and this makes `configurableFields` much nicer to read. Feel free to squash this in the relevant commits if you'd like, or I can send this as a follow-up if you prefer.

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
+	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundlePath, config.SuccessfullyApplied,
+		fmt.Sprintf("Bundle path (string, default '%s')", constants.DefaultBundlePath))
+	cfg.AddSetting(CPUs, constants.DefaultCPUs, config.ValidateCPUs, config.RequiresRestartMsg,
+		fmt.Sprintf("Number of CPU cores (must be equal of bigger than default '%d')", constants.DefaultCPUs))
+	cfg.AddSetting(Memory, constants.DefaultMemory, config.ValidateMemory, config.RequiresRestartMsg,
+		fmt.Sprintf("Memory size in MiB (must be equal of bigger than default '%d')", constants.DefaultMemory))
+	cfg.AddSetting(DiskSize, constants.DefaultDiskSize, config.ValidateDiskSize, config.RequiresRestartMsg,
+		fmt.Sprintf("Total size in GiB of the disk (must be equal of bigger than default '%d')", constants.DefaultDiskSize))
+	cfg.AddSetting(NameServer, "", config.ValidateIPAddress, config.SuccessfullyApplied,
+		"IPv4 address of nameserver (string, like '1.1.1.1 or 8.8.8.8')")
```
cfergeau, line 45: not sure ipv4 examples are required?

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
-	cfg.AddSetting(HTTPProxy, "", config.ValidateURI, config.SuccessfullyApplied)
-	cfg.AddSetting(HTTPSProxy, "", config.ValidateURI, config.SuccessfullyApplied)
-	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied)
-	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied)
+	cfg.AddSetting(HTTPProxy, "", config.ValidateURI, config.SuccessfullyApplied,
+		"HTTP proxy URL (string, like 'http://my-proxy.com:8443')")
+	cfg.AddSetting(HTTPSProxy, "", config.ValidateURI, config.SuccessfullyApplied,
+		"HTTPS proxy URL (string, like 'https://my-proxy.com:8443')")
+	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied,
+		"No proxy (string, like '127.0.0.1,192.168.100.1/24')")
```
cfergeau, line 60: "comma-separated list of ipv4 addresses or CIDR such as ..." ?

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
+	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied,
+		"Certificate Authority (CA) path for proxy")
```
cfergeau, line 62: Maybe "Path to a file containing the Certificate Authority (CA) to use for proxy"?

pkg/crc/preflight/preflight.go
```diff
@@ -144,7 +144,8 @@ func doCleanUpPreflightChecks(checks []Check) error {
 func doRegisterSettings(cfg config.Schema, checks []Check) {
 	for _, check := range checks {
 		if check.configKeySuffix != "" {
-			cfg.AddSetting(check.getSkipConfigName(), false, config.ValidateBool, config.SuccessfullyApplied)
+			cfg.AddSetting(check.getSkipConfigName(), false, config.ValidateBool, config.SuccessfullyApplied,
+				"Skip preflight check (true/false, default: false)")
```
cfergeau, line 148: This is different from all other booleans, which don't mention `true/false`, just the default.

cfergeau (commented):
Couple of comments, mostly related to the help strings. Might be good to get a review from @robin-owen for these?

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
-	cfg.AddSetting(DiskSize, constants.DefaultDiskSize, config.ValidateDiskSize, config.RequiresRestartMsg)
-	cfg.AddSetting(NameServer, "", config.ValidateIPAddress, config.SuccessfullyApplied)
-	cfg.AddSetting(PullSecretFile, "", config.ValidatePath, config.SuccessfullyApplied)
-	cfg.AddSetting(DisableUpdateCheck, false, config.ValidateBool, config.SuccessfullyApplied)
-	cfg.AddSetting(ExperimentalFeatures, false, config.ValidateBool, config.SuccessfullyApplied)
-	cfg.AddSetting(NetworkMode, string(network.DefaultMode), network.ValidateMode, network.SuccessfullyAppliedMode)
+	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundlePath, config.SuccessfullyApplied,
+		fmt.Sprintf("Bundle path (string, default '%s')", constants.DefaultBundlePath))
+	cfg.AddSetting(CPUs, constants.DefaultCPUs, config.ValidateCPUs, config.RequiresRestartMsg,
+		fmt.Sprintf("Number of CPU cores (must be equal of bigger than default '%d')", constants.DefaultCPUs))
```
robin-owen, line 39: "must be greater than or equal to"

This applies to all of the relevant strings. :)

cmd/crc/cmd/config/config.go
robin-owen: I don't think it hurts to have the examples.

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
+	cfg.AddSetting(Memory, constants.DefaultMemory, config.ValidateMemory, config.RequiresRestartMsg,
+		fmt.Sprintf("Memory size in MiB (must be equal of bigger than default '%d')", constants.DefaultMemory))
+	cfg.AddSetting(DiskSize, constants.DefaultDiskSize, config.ValidateDiskSize, config.RequiresRestartMsg,
+		fmt.Sprintf("Total size in GiB of the disk (must be equal of bigger than default '%d')", constants.DefaultDiskSize))
+	cfg.AddSetting(NameServer, "", config.ValidateIPAddress, config.SuccessfullyApplied,
+		"IPv4 address of nameserver (string, like '1.1.1.1 or 8.8.8.8')")
+	cfg.AddSetting(PullSecretFile, "", config.ValidatePath, config.SuccessfullyApplied,
+		fmt.Sprintf("Path of image pull secret (download from %s)", constants.CrcLandingPageURL))
+	cfg.AddSetting(DisableUpdateCheck, false, config.ValidateBool, config.SuccessfullyApplied,
+		"Don't check for update (default false)")
```
robin-owen, line 49: This feels awkward to me, but part of that is due to the configuration property itself being negated. I actually think the wording of the setting itself is adequate ("Disable update check").

cmd/crc/cmd/config/config.go
robin-owen, line 60: "No proxy" seems vague to me. What does that actually mean? IP addresses not routed through a configured proxy?

cmd/crc/cmd/config/config.go
robin-owen: As I understand it, a CA _issues_ certificate files. Perhaps we should be a bit more generic in the wording?

How about "path to an HTTPS proxy certificate"?

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
+	cfg.AddSetting(HTTPSProxy, "", config.ValidateURI, config.SuccessfullyApplied,
+		"HTTPS proxy URL (string, like 'https://my-proxy.com:8443')")
+	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied,
+		"No proxy (string, like '127.0.0.1,192.168.100.1/24')")
+	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied,
+		"Certificate Authority (CA) path for proxy")
 
-	cfg.AddSetting(EnableClusterMonitoring, false, config.ValidateBool, config.SuccessfullyApplied)
+	cfg.AddSetting(EnableClusterMonitoring, false, config.ValidateBool, config.SuccessfullyApplied,
+		"Enable cluster monitoring operator (default false)")
```
robin-owen, line 65: "Operator" should be capitalized.

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
 
 	// Telemeter Configuration
-	cfg.AddSetting(ConsentTelemetry, "", config.ValidateYesNo, config.SuccessfullyApplied)
+	cfg.AddSetting(ConsentTelemetry, "", config.ValidateYesNo, config.SuccessfullyApplied,
+		"Send feedback about CodeReady Containers (yes/no)")
```
robin-owen, line 69: I feel like "feedback" doesn't describe what the consent is actually for. Maybe "Consent to collection of anonymous usage data"?

Also, is there a default here? None is specified in the help string.

robin-owen (changes requested):

cmd/crc/cmd/config/config.go
praveenkumar: yes those are IP address which are not routed through the proxy.

praveenkumar: As per curl man page 
```
--noproxy <no-proxy-list>
              Comma-separated list of hosts which do not use a proxy, if one is specified. 
```

cmd/crc/cmd/config/config.go
praveenkumar: @robin-owen It doesn't have any default, always asked to the user `yes/no`.

pkg/crc/config/config.go
```diff
@@ -37,6 +37,14 @@ func (c *Config) AllConfigs() map[string]SettingValue {
 	return allConfigs
 }
 
+func (c *Config) AllSettings() []Setting {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau there is no `Setting` interface but `Schema` interface is there.

cmd/crc/cmd/config/config.go
```diff
@@ -33,25 +33,40 @@ const (
...
-	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied)
-	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied)
+	cfg.AddSetting(HTTPProxy, "", config.ValidateURI, config.SuccessfullyApplied,
+		"HTTP proxy URL (string, like 'http://my-proxy.com:8443')")
+	cfg.AddSetting(HTTPSProxy, "", config.ValidateURI, config.SuccessfullyApplied,
+		"HTTPS proxy URL (string, like 'https://my-proxy.com:8443')")
+	cfg.AddSetting(NoProxy, "", config.ValidateNoProxy, config.SuccessfullyApplied,
+		"Hosts, ipv4 addresses or CIDR which do not use a proxy (string, comma-separated list such as '127.0.0.1,192.168.100.1/24')")
+	cfg.AddSetting(ProxyCAFile, "", config.ValidatePath, config.SuccessfullyApplied,
+		"path to an HTTPS proxy certificate authority(CA)")
```
robin-owen, line 62: For consistency, let's capitalize "Path" here. :)

PR #1961, opened 2021-02-08
add login-as flag to allow auto login

kwiesmueller:
this change should remove the need for people to copy and paste the login command

## Solution/Idea

Added `login-as` flag that can be set to `admin` or `developer` and triggers an automated run of `oc login` when the cluster is ready.
I thought about it when I copy-pasted my 100st login command.
Not sure if this change makes sense and is executed correctly. Couldn't run the local build yet, not sure why.

## Proposed changes

1. Add `login-as` flag.
2. This means users can provide the flag so they don't have to copy the login command after the cluster is up.
3. (optional) This could default to developer, but I didn't want to be that invasive as the user might not expect their Kubernetes context to change without them knowing.

## Testing

1. `start` succeeds with the `login-as` flag set 
2. `oc login` gets called with admin credentials if the flag is set to `admin`
3. `oc login` gets called with developer credentials if the flag is set to `developer`

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -29,6 +29,7 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
+	AutoLoginAs             = "login-as"
 )
 
 func RegisterSettings(cfg *config.Config) {
@@ -52,6 +53,8 @@ func RegisterSettings(cfg *config.Config) {
 
 	// Telemeter Configuration
 	cfg.AddSetting(ConsentTelemetry, "", config.ValidateYesNo, config.SuccessfullyApplied)
+
+	cfg.AddSetting(AutoLoginAs, "", config.ValidateAdminOrDeveloper, config.SuccessfullyApplied)
 }
 
 func isPreflightKey(key string) bool {
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"strings"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
@@ -34,6 +35,7 @@ func init() {
 	flagSet.UintP(cmdConfig.DiskSize, "d", constants.DefaultDiskSize, "Total size in GiB of the disk used by the OpenShift cluster")
 	flagSet.StringP(cmdConfig.NameServer, "n", "", "IPv4 address of nameserver to use for the OpenShift cluster")
 	flagSet.Bool(cmdConfig.DisableUpdateCheck, false, "Don't check for update")
+	flagSet.String(cmdConfig.AutoLoginAs, "", "Automatically run the login command when the cluster is up; Can be developer or admin")
 
 	startCmd.Flags().AddFlagSet(flagSet)
 }
@@ -156,7 +158,28 @@ func (s *startResult) prettyPrintTo(writer io.Writer) error {
 			"This cluster was built from OKD - The Community Distribution of Kubernetes that powers Red Hat OpenShift.",
 			"If you find an issue, please report it at https://github.com/openshift/okd"}, "\n"))
 	}
-	return err
+	if err != nil {
+		return err
+	}
+
+	if userType := config.Get(cmdConfig.AutoLoginAs).AsString(); len(userType) > 0 {
+		return s.runLoginCommandFor(userType)
+	}
+
+	return nil
+}
+
+func (s *startResult) runLoginCommandFor(userType string) error {
+	username := s.ClusterConfig.DeveloperCredentials.Username
+	password := s.ClusterConfig.DeveloperCredentials.Password
+	if userType == "admin" {
+		username = s.ClusterConfig.AdminCredentials.Username
+		password = s.ClusterConfig.AdminCredentials.Password
+	}
+	cmd := exec.Command("oc", "login", "-u", username, "-p", password, s.ClusterConfig.URL)
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	return cmd.Run()
 }
 
 func isDebugLog() bool {
--- a/pkg/crc/config/validations.go
+++ b/pkg/crc/config/validations.go
@@ -103,3 +103,10 @@ func ValidateYesNo(value interface{}) (bool, string) {
 	}
 	return false, "must be yes or no"
 }
+
+func ValidateAdminOrDeveloper(value interface{}) (bool, string) {
+	if cast.ToString(value) == "admin" || cast.ToString(value) == "developer" {
+		return true, ""
+	}
+	return false, "must be admin or developer"
+}


cmd/crc/cmd/config/config.go
```diff
@@ -52,6 +53,8 @@ func RegisterSettings(cfg *config.Config) {
 
 	// Telemeter Configuration
 	cfg.AddSetting(ConsentTelemetry, "", config.ValidateYesNo, config.SuccessfullyApplied)
+
+	cfg.AddSetting(AutoLoginAs, "", config.ValidateAdminOrDeveloper, config.SuccessfullyApplied)
```
gbraad, line 57: I would probably even go as far as making developer the default,and always log in as this user after a succesful deployment...

... however, at times, the cluster does not reconcile in time to allow this to happen.

cmd/crc/cmd/start.go
```diff
@@ -156,7 +158,28 @@ func (s *startResult) prettyPrintTo(writer io.Writer) error {
 			"This cluster was built from OKD - The Community Distribution of Kubernetes that powers Red Hat OpenShift.",
 			"If you find an issue, please report it at https://github.com/openshift/okd"}, "\n"))
 	}
-	return err
+	if err != nil {
+		return err
+	}
+
+	if userType := config.Get(cmdConfig.AutoLoginAs).AsString(); len(userType) > 0 {
+		return s.runLoginCommandFor(userType)
```
gbraad, line 166: No error is returned/handled, and does not wait for the cluster to be available.

gbraad (changes requested):
Too optimistic in my opinion

praveenkumar:
@ktdreyer we are already creating the `crc-developer` and `crc-admin` context in kubeconfig which anyone can use after `crc start` happen successfully using following. 

```
$ oc config use-context crc-developer
Switched to context "crc-developer".

$ oc whoami
developer

$ oc config use-context crc-admin
Switched to context "crc-admin".

$ oc whoami
kube:admin
```

guillaumerose:
There is already code dedicated for this here: https://github.com/code-ready/crc/blob/master/pkg/crc/machine/kubeconfig.go#L33

The difference with this PR is that it uses directly oc library and doesn't change the default context.
Most of the time the context is not updated because the authentication service is down even after waiting 2min.

Maybe we could force the change of the context to crc-admin in any case.

praveenkumar:
> The difference with this PR is that it uses directly oc library and doesn't change the default context.
> Most of the time the context is not updated because the authentication service is down even after waiting 2min.

Even for login the `auth` service should be running so I would say that we should use context instead of login to cluster (which is in a way same as both create an entry on kubeconfig file).

gbraad:
> Even for login the auth service should be running so I would say that we should use context instead of login to cluster

Yep, as we can not expect the reconcilation and auth service to be available in time... unless 'waited' for.

cmd/crc/cmd/start.go
kwiesmueller: I assumed the cluster is ready when this happens as it's at the same time the login info is printed.
Seems like this assumption is wrong :-/

kwiesmueller:
I didn't know the context is already present. So far all my ongoing kubectl calls (using watch) remained failing after restarting crc until I called the `oc login` snippet.

My selfish goal is allowing the user to happily recreate and restart clusters without having to think about login, context etc.
It happens too often that for some reason I have to recreate my cluster because auth doesn't let me in anymore. If you point me in the right direction I'm happy to give it a better try than this.

So far I didn't know a few things you mentioned here, that already helps.

cmd/crc/cmd/config/config.go
kwiesmueller: My reason against this was to try and be less invasive with current behavior and so users don't get their context switched while working on a different one.
Like you're just running something on a different context and suddenly your `crc` is finally finished and takes over. 
Wanted it to be a conscious decision for now, but haven't got a that strong opinion on it.

gbraad:
@code-ready/crc-devel what to do about this PR?

cfergeau:
> @code-ready/crc-team l what to do about this PR?

My take on that is that with a production cluster, you would not be logged in automatically, you'd need to do some manual steps to authenticate with it before. I prefer when crc is as close as possible to the interactions one would have with a production cluster, `--login-as` is at odds with that. We've also had requests to automatically create a new project, which falls in the same bucket.
A more automated wrapper on top of crc could be useful though?

kwiesmueller:
I don't agree @cfergeau.
As far as I understood crc is mainly intended for local development and testing. Sure the cluster itself should be close to reality (which it already isn't due to features disabled by default), but for me that shouldn't imply the user experience has to be as "bad" as in production.
Why create a wrapper around crc that improves using crc if crc is already the tool to get a cluster as easy as possible?

Creating projects and running other things I tend to agree, because they are easily done after login, but auto-login is just automation to actually let you automate things in the first place.
Also other simple kubernetes installers (like k3s) let you provide manifests and commands that get executed on cluster setup. So it's not entirely uncommon.

Right now crc can't easily be used without human intervention (as far as I know) which as a user I don't understand why.
The name code-ready-containers kinda is an unfulfilled promise if it's not ready for my code as long as I still need to do manual work (even though it's fairly small) to make it ready for my code. Unrealistic I know, but the reference was too tempting.

gbraad:
> but for me that shouldn't imply the user experience has to be as "bad" as in production.

Then the problem should be solved in the upstream... ? The reason we take this stance is not to introduce convenience features that will bite us. Especially as we would first have to solve the 'wait for auth possible' before we can handle this usecase. 

> but auto-login

is not an essential feature that is reliant on a lot of CRC specifics. it could run as:

```
$ crc start && oc login -u developer -p developer api.crc-testing:6443 
```

> is an unfulfilled promise if it's not ready for my code as long as I still need to do manual work

this would also mean actions like `oc new-app`, etc are expected to be automated? if so, this is actually what `odo` and the plugins/extensions do in union with CodeReady Containers.

> Also other simple kubernetes installers (like k3s) let you provide manifests and commands that get executed on cluster setup. So it's not entirely uncommon.

Which we provided before with add-ons for minishift. But we didn't want to create CRC specific solutions if what can be solved as part of OpenShift or devtools wide. Again, this is just running buildscrpts after the deployment of 'a cluster' which does not have to be CRC specific. We rather not devise a solution that would only work for CRC.

Note: I am concerned that this `auto login` will not function as expected in most cases as the cluster startuo phase can take longer than our timeout of several minutes. This would mean the login would error and would be a 50-50 chance of working. Those aren't great odds for a convenience feature.

cmd/crc/cmd/config/config.go
gbraad: That makes sense... so the use is:

```
$ crc config set auto-login-as developer
```

and disable as:

```
$ crc config unset auto-login-as
```

which means until `unset` this would blow away the current one... and the unset if set this way can not be overriden from the `crc start` command as it would not take an empty value, right?

so this means that beforehand you would need to see this config is set, or at least as the final output that it happened and succeeded. so at least the user gets the feedback what happened. the current code does not have this output. this means the behaviour to the user can be unpredictable. especially in regards to the possible startup timeout this would make it unclear to the user if follow-up commands are actually running in the right context.

kwiesmueller:
For developer this would work.
For admin the login is different every time :-/
But as you mentioned the context should be there, so I'll try to just chain `kubectl config use-context` and see if that's enough.

Note that I'm very new to crc and have been using it for the last two months. It's the only way for me to test on OpenShift that worked well enough so far, but still gave me problems once in a while. This being one of them.
This is my attempt of making my own life easier with the understanding I have of what's going on. I didn't even know there is an upstream that could fix this.

I can understand your points but it feels a bit like I'm pushing on a sensitive point here that I don't know about. It seemed trivial to me to add and have as a feature, sorry for the assumptions.

cmd/crc/cmd/start.go
gbraad: I am very concerned that this leads to unpreditability; the user gets no feedback about error or success in this case.

gbraad:
> For admin the login is different every time :-/

this is being addressed ...

gbraad:
> This is my attempt of making my own life easier with the understanding I have of what's going on

Much appreciated!

> I can understand your points but it feels a bit like I'm pushing on a sensitive point here that I don't know about. It
> seemed trivial to me to add and have as a feature, sorry for the assumptions.

No no no, the issue here is a technical debt on our end that has been pushed forward. ;-/

I do understand your point of view, and it does make sense ... although we also got flack when this auto login happened in the minishift time period. we'd rather not take an action if the user is clearly in control and gets feedback on this.

cmd/crc/cmd/start.go
gbraad: https://github.com/code-ready/crc/blob/d8c8b8d5b7eda770a28ce66275d6e886b1b26550/pkg/crc/machine/start.go#L370

which might not be enough...

so this is followed by

https://github.com/code-ready/crc/blob/d8c8b8d5b7eda770a28ce66275d6e886b1b26550/pkg/crc/machine/start.go#L374-L377

doing:
https://github.com/code-ready/crc/blob/d8c8b8d5b7eda770a28ce66275d6e886b1b26550/pkg/crc/machine/kubeconfig.go#L33-L49

As you can see, this is mostly based on timeouts and retry. This is the technical debt I refer to.

cmd/crc/cmd/config/config.go
kwiesmueller: Not sure if I get what you mean here.
This might be wrong, I just followed other flag implementations. It should be possible for the user to set login-as as a flag or in the config and disable it the same way (so it probably should accept an empty value on the flag to unset).

Regarding the output, I agree. Now that I know the login might not work instantly, it should write a message like "attempting to login" and either try silently until a timeout is reached or directly fail with an error afterwards.

cmd/crc/cmd/start.go
kwiesmueller: Thanks for explaining! That makes a lot of sense.
One option I could think about is to change the flag and not run login, but if the flag is set, set the context written to kubeconfig as current context at the same time?

Either way, how would you like to proceed with this? If you think having this option is not really viable feel free to close this, if you think it would be and we can get it in there good enough, I don't mind working to bring it up to standard.

gbraad: @praveenkumar WDYT?

cmd/crc/cmd/config/config.go
gbraad: Attempting suggest very clearly this might fail. Uncertainty is not suggested in communicating.

Auto login ... [failed]

Would work better.  But yeah, a possible approach is to try this several times with a timeout. But Praveen might have an idea around this

praveenkumar:
> But as you mentioned the context should be there, so I'll try to just chain kubectl config use-context and see if that's enough.

@kwiesmueller Please do try this and let us know if that is not working for you, I don't have strong opinion for this autologin feature but I am afraid it will not going to work for the user for whom `kubectl config use-context crc-admin` is not working.

praveenkumar:
@kwiesmueller Have you used `config` stuff and does it working for you?

praveenkumar:
Closing this since there is no response from the PR submitter, @kwiesmueller please open an issue if something doesn't work with you using `oc config use-context crc-admin` side.

gbraad:
Thanks for the submission. If you have any features or suggestions, please do not hesitate to file an issue to start the discussion.

gbraad:
> > For admin the login is different every time :-/

> this is being addressed ...

This came uo with the security review and we are looking into a solution around this.

kwiesmueller:
Sorry for the late reply. `use-context` worked for me the last few times thanks!
I thought I replied a few days ago, my bad.

PR #1974, opened 2021-02-10
Fix #1874 Do not run preflights if VM already running

### LINKED ISSUE #1874: VM Started state should be checked before running preflights
Starting an already running cluster does:

```
N:\go\bin\windows_amd64>.\crc.exe start
INFO Checking Windows edition
INFO Checking if user is a member of the Hyper-V Administrators group
INFO A CodeReady Containers VM for OpenShift 4.6.9 is already running


Started the OpenShift cluster
```

which ran pre-flights checks before concluding:

```
INFO A CodeReady Containers VM for OpenShift 4.6.9 is already running
```

gbraad:
Signed-off-by: Gerard Braad <me@gbraad.nl>

**Fixes:** Issue #1874

## Solution/Idea

Preflights were run for an akready running VM. This means they have been able to pass the pre-flights before.

## Proposed changes

```
PS> .\crc.exe start
INFO A CodeReady Containers VM for OpenShift 4.7.0-0.nightly-2021-01-22-134922 is already running
Started the OpenShift cluster

To access the cluster, first set up your environment by following the instructions returned by executing 'crc oc-env'.
Then you can access your cluster by running 'oc login -u developer -p developer https://api.crc.testing:6443'.
To login as a cluster admin, run 'oc login -u kubeadmin -p ex6Fy-ALyjf-kIy8N-4NXNH https://api.crc.testing:6443'.

You can also run 'crc console' and use the above credentials to access the OpenShift web console.
The console will open in your default browser.
```

## Testing

  * Initial start will run preflights
  * When start is run again, the message 'is already running' follows. No preflight checks
  * When stop is called, the preflights will run again on start
  * When delete is called, the preflights will run again on start

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -60,10 +60,6 @@ func runStart(ctx context.Context) (*machine.StartResult, error) {
 
 	checkIfNewVersionAvailable(config.Get(cmdConfig.DisableUpdateCheck).AsBool())
 
-	if err := preflight.StartPreflightChecks(config); err != nil {
-		return nil, err
-	}
-
 	telemetry.SetContextProperty(ctx, cmdConfig.CPUs, config.Get(cmdConfig.CPUs).AsInt())
 	telemetry.SetContextProperty(ctx, cmdConfig.Memory, uint64(config.Get(cmdConfig.Memory).AsInt())*1024*1024)
 	telemetry.SetContextProperty(ctx, cmdConfig.DiskSize, uint64(config.Get(cmdConfig.DiskSize).AsInt())*1024*1024*1024)
@@ -78,6 +74,14 @@ func runStart(ctx context.Context) (*machine.StartResult, error) {
 	}
 
 	client := newMachine()
+	isRunning, _ := client.IsRunning()
+
+	if !isRunning {
+		if err := preflight.StartPreflightChecks(config); err != nil {
+			return nil, err
+		}
+	}
+
 	return client.Start(startConfig)
 }
 
--- a/pkg/crc/machine/client.go
+++ b/pkg/crc/machine/client.go
@@ -16,6 +16,7 @@ type Client interface {
 	Start(startConfig StartConfig) (*StartResult, error)
 	Status() (*ClusterStatusResult, error)
 	Stop() (state.State, error)
+	IsRunning() (bool, error)
 }
 
 type client struct {
--- a/pkg/crc/machine/fakemachine/client.go
+++ b/pkg/crc/machine/fakemachine/client.go
@@ -100,3 +100,7 @@ func (c *Client) Status() (*machine.ClusterStatusResult, error) {
 func (c *Client) Exists() (bool, error) {
 	return true, nil
 }
+
+func (c *Client) IsRunning() (bool, error) {
+	return true, nil
+}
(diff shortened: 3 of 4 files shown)


gbraad:
e2e-crc reports ...
```
 OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 104
```

cmd/crc/cmd/start.go
```diff
@@ -78,6 +74,14 @@ func runStart(ctx context.Context) (*machine.StartResult, error) {
 	}
 
 	client := newMachine()
+	exists, _ := client.Exists()
+
+	if !exists {
```
guillaumerose, line 79: Fine for me. This code you wrote should also be duplicated in the daemon API.

I am tempted to move this in the machine.Start function.

cfergeau:
I'm confused by this, no idea what this means?
> When stop is called, the VM will start deployment of the cluster

This PR seems to be dropping preflights for the second start in this sequence: `crc start; crc stop; crc start`?
I'm not sure this is desirable, as it could be some days/weeks between crc runs, with changes to the system being done in the mean time (`crc` libvirt network being stopped for whatever reason, networking configuration being altered because it gets in the way, or something which happened to me yesterday, bios upgrades on ryzen-based machines disabling kvm)

guillaumerose:
So, just add a check if the VM is already running ?

cfergeau:
> So, just add a check if the VM is already running ?

Yup, no preflight when VM is already running sounds good.

cmd/crc/cmd/start.go
gbraad: The problem is that the prefflights do not run from the start... and the start already detects this situation.

gbraad:
Follow-up issue will need to be looked into, related to #1912 to check the state of OpenShift running or not:

A `start` will now not run the pre-flights, but is very optimistic about reporting:

```
PS> .\crc.exe start
INFO A CodeReady Containers VM for OpenShift 4.7.0-0.nightly-2021-01-22-134922 is already running
Started the OpenShift cluster

To access the cluster, first set up your environment by following the instructions returned by executing 'crc oc-env'.
Then you can access your cluster by running 'oc login -u developer -p developer https://api.crc.testing:6443'.
```

without actually checking if OpenShift IS actually running.

Note: not to be solved in this issue!

cfergeau (approved):

gbraad:
rebased

gbraad:
@jsliacan tested this manually on Linux. Works as expected.

jsliacan:
Tested on MacOS, basic things work:
- `start` + `start` = no preflights
- `start` + `stop` + `start` = preflights
- `start` + `delete` + `start` = preflights

jsliacan:
> without actually checking if OpenShift IS actually running.

The message only says that the VM is already running. It does not talk about the cluster. So it seems like a correct behaviour to me. Here an example:

```
$ crc status
CRC VM:          Running
OpenShift:       Starting (v4.7.0-fc.3)
Disk Usage:      12.17GB of 32.74GB (Inside the CRC VM)
Cache Usage:     22.43GB
Cache Directory: /home/jsliacan/.crc/cache
$ crc start
INFO A CodeReady Containers VM for OpenShift 4.7.0-fc.3 is already running 
Started the OpenShift cluster

To access the cluster, first set up your environment by following the instructions returned by executing 'crc oc-env'.
Then you can access your cluster by running 'oc login -u developer -p developer https://api.crc.testing:6443'.
To login as a cluster admin, run 'oc login -u kubeadmin -p cGQi7-LRGZU-HN9ZU-dsCbi https://api.crc.testing:6443'.

You can also run 'crc console' and use the above credentials to access the OpenShift web console.
The console will open in your default browser.
```

Issue #1764, opened 2020-12-09
Bring back the 'cluster started' banner/summary


gbraad:
Looking at: https://gitlab.com/redhatdemocentral/ocp-install-demo/-/blob/main/init.sh#L231-255 I noticed this is the same output Minishift returned, and to be honest this is a lot clearer. We should bring this back

```
echo
echo "======================================================"
echo "=                                                    ="
echo "=  Install complete, get ready to rock.              ="
echo "=                                                    ="
echo "=  The server is accessible via web console at:      ="
echo "=                                                    ="
echo "= ${OCP_CONSOLE} ="
echo "=                                                    ="
echo "=  Log in as admin: kubeadmin                        ="
echo "=         password: ${KUBE_PASS}          ="
echo "=                                                    ="
echo "=  Log in as dev: developer                          ="
echo "=       password: developer                          ="
echo "=                                                    ="
echo "=  Now get your Red Hat Demo Central example         ="
echo "=  projects here:                                    ="
echo "=                                                    ="
echo "=     https://github.com/redhatdemocentral           ="
echo "=                                                    ="
echo "=  To stop, restart, or delete your OCP cluster:     ="
echo "=                                                    ="
echo "=     $ crc {stop | start | delete}                  ="
echo "=                                                    ="
echo "======================================================"
```

/cc: @eschabell

PR #1975, opened 2021-02-10
Change the cluster started summary

guillaumerose:
**Fixes:** Issue #1764

* Directly add the link to the webconsole. No more indirection with crc
console.
* Better display credentials
* Include a small how-to for the oc command line

```
$ crc start

Log in as administrator:
  Username: kubeadmin
  Password: APBEh-jjrVy-hLQZX-VI9Kg

Log in as user:
  Username: developer
  Password: developer

  $ eval $(crc oc-env)
  $ oc login https://api.crc.testing:6443
```

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"strings"
+	"text/template"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	"github.com/code-ready/crc/pkg/crc/cluster"
@@ -18,6 +20,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/code-ready/crc/pkg/crc/validation"
 	crcversion "github.com/code-ready/crc/pkg/crc/version"
+	"github.com/code-ready/crc/pkg/os/shell"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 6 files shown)


cmd/crc/cmd/start.go
```diff
@@ -199,3 +195,44 @@ func checkIfNewVersionAvailable(noUpdateCheck bool) {
...
+
+Log in as administrator:
+  Username: {{ .ClusterConfig.AdminCredentials.Username }}
+  Password: {{ .ClusterConfig.AdminCredentials.Password }}
+
+Log in as user:
+  Username: {{ .ClusterConfig.DeveloperCredentials.Username }}
+  Password: {{ .ClusterConfig.DeveloperCredentials.Password }}
+
+Use directly the 'oc' command line:
```
robin-owen, line 212: This is worded incorrectly. My preference would be "Use the `oc` command line interface".

gbraad:
> Use directly the 'oc' command line:

This does not sound correctly.

`To use the 'oc' CLI tool:` ?

cmd/crc/cmd/start.go
guillaumerose: done thx

cmd/crc/cmd/start.go
```diff
@@ -203,3 +200,54 @@ func checkIfNewVersionAvailable(noUpdateCheck bool) {
...
+	}
+	return "$"
+}
+
+func shellForTemplate() string {
+	if runtime.GOOS == "windows" {
+		return "powershell"
+	}
+	return ""
+}
```
cfergeau, line 253: Wouldn't these 2 helpers be a better fit in the `shell` package? I though on Windows you could either use `powershell` or `cmd.exe`? Why `shellForTemplate()` when there is `shell.GetShell("")`?

guillaumerose: If the user launches it with cmd, the start output is really ugly: 
`@FOR /f "tokens=*" %%i IN ('crc oc-env') DO @call %%i`

I prefer to encourage the user to use PowerShell.

cfergeau: > I prefer to encourage the user to use PowerShell.

But this is not what this PR is about, and the current PR output won't work if I ran crc from cmd.exe and c&p it

cfergeau (approved):

Issue #1914, opened 2021-02-01
[refactoring] crc daemon should use http for its API, not a custom protocol


guillaumerose:
crc daemon API is using a custom protocol. It makes it "hard" to debug and use. It would be really useful to be able to use it with curl for instance.

To ease the transition with the tray, we could support both protocols and then only keep http.

Linked to https://github.com/code-ready/tray-macos/issues/53

guillaumerose:
This should prevent this kind of change: https://github.com/code-ready/tray-windows/pull/21/files#diff-7cc7f97725df2bd69b61f14bfb2b714e9d6d577e9726359b92a83c22799f2d7dR164

guillaumerose:
Only Windows side is remaining. 

See https://github.com/code-ready/tray-windows/issues/57 and https://github.com/code-ready/tray-windows/issues/84

PR #1978, opened 2021-02-11
Use HTTP to serve the daemon api

anjannath:
**Fixes:** #1914 

Needs #1993

### DIFF
--- a/pkg/crc/api/api_http.go
+++ b/pkg/crc/api/api_http.go
@@ -1,0 +1,150 @@
+package api
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	crcConfig "github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/machine"
+)
+
+func NewMux(config crcConfig.Storage, machine machine.Client) http.Handler {
+	handler := NewHandler(config, machine)
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/start", func(w http.ResponseWriter, r *http.Request) {
+		var startResult string
+		data, err := verifyRequestAndReadBody(w, r, http.MethodGet, http.MethodPost)
+		if err != nil {
+			return
+		}
+		if len(data) > 0 {
+			startResult = handler.Start(data)
+		} else {
+			startResult = handler.Start(nil)
+		}
+		sendResponse(w, startResult)
+	})
+
+	mux.HandleFunc("/stop", func(w http.ResponseWriter, r *http.Request) {
+		if wrongHTTPMethodUsed(r, w, http.MethodGet) {
+			return
+		}
+		stopResult := handler.Stop()
+		sendResponse(w, stopResult)
+	})
+
+	mux.HandleFunc("/status", func(w http.ResponseWriter, r *http.Request) {
+		if wrongHTTPMethodUsed(r, w, http.MethodGet) {
+			return
+		}
+		status := handler.Status()
+		sendResponse(w, status)
+	})
+
+	mux.HandleFunc("/delete", func(w http.ResponseWriter, r *http.Request) {
+		if wrongHTTPMethodUsed(r, w, http.MethodGet) {
+			return
+		}
+		deleteResult := handler.Delete()
+		sendResponse(w, deleteResult)
+	})
+
+	mux.HandleFunc("/version", func(w http.ResponseWriter, r *http.Request) {
+		if wrongHTTPMethodUsed(r, w, http.MethodGet) {
+			return
+		}
+		versionResult := handler.GetVersion()
+		sendResponse(w, versionResult)
+	})
+
+	mux.HandleFunc("/webconsoleurl", func(w http.ResponseWriter, r *http.Request) {
+		if wrongHTTPMethodUsed(r, w, http.MethodGet) {
+			return
+		}
+		webconsoleInfo := handler.GetWebconsoleInfo()
+		sendResponse(w, webconsoleInfo)
+	})
+
+	mux.Handle("/config/", http.StripPrefix("/config", newConfigMux(handler)))
+
+	return mux
+}
+
+func newConfigMux(handler *Handler) http.Handler {
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("/get", func(w http.ResponseWriter, r *http.Request) {
+		var configGetResult string
+		data, err := verifyRequestAndReadBody(w, r, http.MethodGet, http.MethodPost)
+		if err != nil {
+			return
+		}
+		if len(data) > 0 {
+			configGetResult = handler.GetConfig(data)
+		} else {
+			configGetResult = handler.GetConfig(nil)
+		}
+		sendResponse(w, configGetResult)
+	})
+
+	mux.HandleFunc("/set", func(w http.ResponseWriter, r *http.Request) {
+		handleConfigSetUnset(w, r, func(data []byte) string { return handler.SetConfig(data) })
+	})
+
+	mux.HandleFunc("/unset", func(w http.ResponseWriter, r *http.Request) {
+		handleConfigSetUnset(w, r, func(data []byte) string { return handler.UnsetConfig(data) })
+	})
+
+	return mux
+}
+
+func wrongHTTPMethodUsed(r *http.Request, w http.ResponseWriter, allowedMethods ...string) bool {
+	for _, method := range allowedMethods {
+		if r.Method == method {
+			return false
+		}
+	}
+	logging.Debugf("Wrong method: %s used for request to: %s", r.Method, r.RequestURI)
+	http.Error(w, fmt.Sprintf("Only %s is allowed.", strings.Join(allowedMethods, ",")), http.StatusMethodNotAllowed)
+	return true
+}
+
+func sendResponse(w http.ResponseWriter, response string) {
+	w.Header().Set("Content-Type", "application/json")
+	if _, err := io.WriteString(w, response); err != nil {
+		logging.Error("Failed to send response: ", err)
+	}
+}
+
+func handleConfigSetUnset(w http.ResponseWriter, r *http.Request, fn func([]byte) string) {
+	data, err := verifyRequestAndReadBody(w, r, http.MethodPost)
+	if err != nil {
+		return
+	}
+	if len(data) < 1 {
+		http.Error(w, "Not enough arguments for /config/(un)set", http.StatusBadRequest)
+		return
+	}
+
+	result := fn(data)
+	sendResponse(w, result)
+}
+
+func verifyRequestAndReadBody(w http.ResponseWriter, r *http.Request, method ...string) ([]byte, error) {
+	if wrongHTTPMethodUsed(r, w, method...) {
+		return nil, fmt.Errorf("Wrong HTTP method used")
+	}
+
+	data, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		logging.Debugf("Unable to read request body for URI: %s", r.RequestURI)
+		http.Error(w, "Unable to read request body", http.StatusInternalServerError)
+		return nil, err
+	}
+	return data, nil
+}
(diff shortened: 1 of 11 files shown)


pkg/crc/api/api_http.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			logging.Debugf("Unable to read request body for URI: %s", r.RequestURI)
+			http.Error(w, "Unable to read request body", http.StatusInternalServerError)
+			return
+		}
+		if len(data) < 1 {
+			http.Error(w, "Not enough arguments for /config/set", http.StatusBadRequest)
+			return
+		}
+		configSetResult := handler.SetConfig(data)
```
cfergeau, line 115: Fwiw, that's a lot of copy&pasted code when the only difference between each block of code is this line.

cfergeau (commented):
`Inform if a start, stop or delete operation is in progress` serializes some of the http handlers, but not others. Not clear if this means these other handlers are concurrency-safe, or if they were missed. This commit is imo a good opportunity to clarify that, and either add a mutex/semaphore to the handler, or a comment to document that it's safe to run this handler concurrently with itself/with other handers.

pkg/crc/api/handlers.go
```diff
@@ -15,9 +17,12 @@ import (
 	"github.com/code-ready/crc/pkg/crc/version"
 )
 
+const startStopOrDeleteInProgressMessage = "Another Start, Stop or Delete Operation is in progress. Please try again later."
```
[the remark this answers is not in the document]
anjannath: Yes, we should associate all the errors with some http status codes,  but this will be done in a follow up PR

pkg/crc/api/handlers.go
```diff
@@ -26,6 +31,7 @@ func NewHandler(config crcConfig.Storage, machine machine.Client) *Handler {
 			Underlying: machine,
 		},
 		Config: config,
+		lock:   semaphore.NewWeighted(int64(1)),
```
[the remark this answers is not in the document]
anjannath: No, functionally a mutex would be enough, but i wanted the `TryAcquire() bool` method to check if the lock can be attained, which is not there in `Mutex`

pkg/crc/api/api_http.go
anjannath: Changed, not ideal but reduced the duplicating lines a bit.

pkg/crc/api/handlers.go
```diff
@@ -35,11 +41,19 @@ func (h *Handler) Status() string {
...
+		return encodeStructToJSON(commandResult)
+	}
+	return startStopOrDeleteInProgressMessage
 }
 
 func (h *Handler) Start(args json.RawMessage) string {
+	if !h.lock.TryAcquire(int64(1)) {
+		return startStopOrDeleteInProgressMessage
+	}
+	defer h.lock.Release(int64(1))
```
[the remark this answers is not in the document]
anjannath: Done

pkg/crc/api/api_http.go
```diff
@@ -0,0 +1,167 @@
...
+
+	var result string
+
+	switch configCmd {
+	case set:
+		result = handler.SetConfig(data)
+	case unset:
+		result = handler.UnsetConfig(data)
+	default:
+		logging.Debugf("Unknown config command: %s", configCmd)
```
anjannath, line 149: Should this be `logging.Panic` instead?

cfergeau (commented):
` Restrict the API available via vsock to only the connect path` could be `... to the /connect endpoint` maybe?

What I'm having troubles understanding though is the `/network` change, this replaces an API which was accessed through a separate socket with an API which is accessible through a new http endpoint?

pkg/crc/api/handlers.go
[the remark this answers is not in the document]
cfergeau: ok!

pkg/crc/api/handlers.go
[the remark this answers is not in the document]
cfergeau: Ah right, no `TryLock()`, apparently this is on purpose. Seems like the go way of handling this would be to use `select`+go channel+ another go routine

pkg/crc/api/api_http.go
```diff
@@ -0,0 +1,167 @@
...
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	crcConfig "github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/machine"
+)
+
+var handler *Handler
```
cfergeau, line 15: Better if we can avoid this.

pkg/crc/api/api_http.go
cfergeau: You can avoid this issue if you pass a function to your method instead of `configCmd string`:
```
func handleConfigSetUnset(w http.ResponseWriter, r *http.Request, func([]byte) string)
```
and 
```
var set = func (data []byte) string {
    return handler.SetConfig(data)
}
```
(you need to define this `set` function in `newConfigMux` so that it can capture/use the `handler` local variable)

pkg/crc/api/handlers.go
```diff
@@ -26,20 +31,37 @@ func NewHandler(config crcConfig.Storage, machine machine.Client) *Handler {
...
+		lock:   semaphore.NewWeighted(int64(1)),
 	}
 }
 
+/*
+* Start, Stop and Delete  handlers are serialized with a semaphore
+* since they modify VM state  they should not be ran  concurrently
+* The other handlers viz. Status, GetVersion, GetConfig, SetConfig
+* UnsetConfig and GetWebconsoleInfo are thread safe and can be ran
+* concurrently
```
cfergeau, line 43: `viz.`? dunno if it's intentional/not sure what it means?

cfergeau: This statement seems strongly biased by the needs of the trays ;) If Start/Stop/Delete modify VM state, why are we allowing `Status` to run concurrently with them, as the `Status` we return is most likely going to be outdated very soon?
Running `SetConfig`/`UnsetConfig` concurrently can also be misleading, since we return the list of the keys which were successfully set/unset. `GetConfig/SetConfig/UnsetConfig` are only thread safe as long as no other thread calls `AddSetting` while they are running.

anjannath: `Status` can also be `Starting` or `Stopping` so if the status is queried when a `Start` `Stop` or `Delete` is happening concurrently it'll actually show the correct status (as long as it can fetch the status) even if it gets outdated its the status at that point of time.

> GetConfig/SetConfig/UnsetConfig are only concurrency safe as long as no other goroutine calls AddSetting while they are running.

yeah, but none of the API handlers will be calling it, we'll always be using the config object created in root.go

anjannath: > viz.? dunno if it's intentional/not sure what it means?

Its supposed to mean "namely" i used 'viz.' to keep the line of the same length as the one above it :)

cfergeau: ah, very first time I see this. It's supposed to be surrounded by commas, so the line lengths won't match ;)

cfergeau: >>    GetConfig/SetConfig/UnsetConfig are only concurrency safe as long as no other goroutine calls AddSetting while they are running.

> yeah, but none of the API handlers will be calling it, we'll always be using the config object created in root.go

This statement is missing a very important word, this word is "today". The code is currently written this way, and it's why it's indeed not an issue. But we now have an undocumented/unenforced requirement that no `crc` code should be calling `AddSetting` after the daemon listeners  have been started. If we forget about this, we'll start having random crashes or bugs when using the daemon.

gbraad: > If Start/Stop/Delete modify VM state, why are we allowing Status to run concurrently with them, as the Status we return is most likely going to be outdated very soon?

A status will ALWAYS reflect a moment in time, from the time it was requested. Otherwise we need to implement a streaming model... with all the complexity involved. This sounds like improvements.

pkg/crc/api/api_http.go
```diff
@@ -0,0 +1,158 @@
...
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/start", func(w http.ResponseWriter, r *http.Request) {
+		var startResult string
+		data, err := verifyRequestAndReadBody(w, r, http.MethodGet, http.MethodPost)
+		if err != nil {
+			return
+		}
+		if len(data) > 0 {
+			startResult = handler.Start(data)
```
guillaumerose, line 26: It would be better to use http error code here instead of our custom structs. But we can start with this!

pkg/crc/api/api_http.go
```diff
@@ -0,0 +1,158 @@
...
+
+	mux.Handle("/config/", http.StripPrefix("/config", newConfigMux(handler)))
+
+	return mux
+}
+
+func newConfigMux(handler *Handler) http.Handler {
+	mux := http.NewServeMux()
+
+	set := func(data []byte) string {
```
guillaumerose, line 81: inline this function?

pkg/crc/api/api_http_test.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func TestHTTPApi(t *testing.T) {
+	machine := fakemachine.NewClient()
+	config := setupNewInMemoryConfig()
+
+	ts := httptest.NewServer(NewMux(config, machine))
+
+	client := http.DefaultClient
+	res, err := client.Get(fmt.Sprintf("%s%s", ts.URL, "/version"))
```
guillaumerose, line 25: Extract this client to a real pkg that we will support like pkg/api/client

pkg/crc/api/api_http.go
anjannath: You mean the the `set` and `unset` functions? they are used in the `handleConfigSetUnset` function and if inline it in that function, then we cannot differentiate which method to call based on whether the request came to the `/set` endpoint or `/unset` endpoints. Or did i completely misunderstood this 😛?

pkg/crc/api/api_http.go
```diff
@@ -0,0 +1,158 @@
...
+		if r.Method == method {
+			return false
+		}
+	}
+	logging.Debugf("Wrong method: %s used for request to: %s", r.Method, r.RequestURI)
+	http.Error(w, fmt.Sprintf("Only %s is allowed.", strings.Join(allowedMethods, ",")), http.StatusMethodNotAllowed)
+	return true
+}
+
+func sendResponse(w http.ResponseWriter, response string) {
```
gbraad, line 125: what is the status code?

pkg/crc/api/api_http.go
gbraad: any status code, as even 200 is missing?

pkg/crc/api/api_http.go
anjannath: This 200 by default for now, if we don't explicitly set the status code, its automatically `200 OK`

pkg/crc/api/api_http.go
cfergeau: I think the suggestion is, rather than
```
	unset := func(data []byte) string {
		return handler.UnsetConfig(data)
	}
	mux.HandleFunc("/unset", func(w http.ResponseWriter, r *http.Request) {
		handleConfigSetUnset(w, r, unset)
	})
```
use
```
	mux.HandleFunc("/unset", func(w http.ResponseWriter, r *http.Request) {
		handleConfigSetUnset(w, r, func(data []byte) string { return handler.UnsetConfig(data) } )
	})
```

anjannath: Ahh, Got it, Thanks

pkg/crc/api/handlers.go
anjannath: Dropped the serialization commit as PR #2101 addresses this

pkg/crc/api/api_http.go
anjannath: Done!

pkg/crc/api/api_http_test.go
anjannath: Done, PTAL again :)

pkg/crc/api/api_http.go
anjannath: I wanted to do this as a follow up PR, fear not to break the old api with changing the handlers

guillaumerose:
Thanks! 

/lgtm

PR #1979, opened 2021-02-11
Increase apiserver wait time for a min

praveenkumar:
In openshift 4.7 looks like the moment we start the kubelet service
and kubelet being able to start the pod, have added more than a min
time. This patch will unblocks us to start the 4.7.0-rc testing.

We should revert it as soon as https://bugzilla.redhat.com/show_bug.cgi?id=1927263
is fixed.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -320,7 +320,7 @@ func WaitForAPIServer(ocConfig oc.Config) error {
 		logging.Debug(stdout)
 		return nil
 	}
-	return errors.RetryAfter(2*time.Minute, waitForAPIServer, time.Second)
+	return errors.RetryAfter(3*time.Minute, waitForAPIServer, time.Second)
 }
 
 func DeleteOpenshiftAPIServerPods(ocConfig oc.Config) error {


gbraad (approved):
Ugh... really?

this is gonna come back in some form. we let people wait too long, or later we will increase it again

cfergeau:
One thing I'm wondering is if this issue we are having with 4.7 is just some delay being moved around. Maybe we need to wait 1 minute more for the apiserver to be available, but after that, we can wait less time at the end for the rest of the cluster to come up (the 3 minutes wait we already have), because some of the initialization is now being done earlier? (not sure I'm clear ^^ )

praveenkumar:
> One thing I'm wondering is if this issue we are having with 4.7 is just some delay being moved around. Maybe we need to wait 1 minute more for the apiserver to be available, but after that, we can wait less time at the end for the rest of the cluster to come up (the 3 minutes wait we already have), because some of the initialization is now being done earlier? (not sure I'm clear ^^ )

@cfergeau no it is not like that, problem here is somehow kubelet stuck to run any pod for more than a min which is why we need to increase the time, I am not even sure why it is happening because ideally all the static pods suppose to run as soon as the kubelet service is started but it is not the case here :(

Issue #1966, opened 2021-02-09
Use `crc-developer` or `crc-admin` context for some of e2e tests


praveenkumar:
In our e2e tests we are not testing the oc context switch and it should be tested to make sure the feature is working as expected.

cfergeau:
Seems to be causing CI failures in #1980?

guillaumerose:
Should we ensure authentication is working before the end of crc start ? Right now, when crc start ends, the cluster is maybe 50% of the time for me not working (console and auth).

guillaumerose:
I added 5min timeout to try and it said ok all operators up. Just the time to type a new command, 3-4 operators were restarting..

PR #1980, opened 2021-02-11
Extract the bundle at the end of crc setup

guillaumerose:
This improves the UX by failing fast if something is wrong with the
machine or OS. For instance, it will not uncompress the bundle and then
fail because virtualization is not enabled.

We can then check if the bundle is in place to know if crc setup was run.


cfergeau (approved):

cfergeau:
CI fails with `And setting kubeconfig context to "crc-admin" succeeds` which is probably related to #1966

PR #1983, opened 2021-02-11
Update download location

gbraad:
Update README to the new download location: https://cloud.redhat.com/openshift/create/local


gbraad:
Already captured in: #1982

PR #1984, opened 2021-02-11
Add error type to segment

praveenkumar:
It will be easy to create filter using error type for telemetry side.

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -2,6 +2,8 @@ package segment
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
@@ -85,7 +87,8 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 		Set("duration", duration.Milliseconds()).
 		Set("tty", crcos.RunningInTerminal())
 	if err != nil {
-		properties = properties.Set("error", err.Error())
+		properties = properties.Set("error", err.Error()).
+			Set("error-type", errorType(err))
 	}
 
 	return c.segmentClient.Enqueue(analytics.Track{
@@ -139,4 +142,12 @@ func (l *loggingAdapter) Logf(format string, args ...interface{}) {
 
 func (l *loggingAdapter) Errorf(format string, args ...interface{}) {
 	logging.Errorf(format, args...)
+}
+
+func errorType(err error) string {
+	wrappedErr := errors.Unwrap(err)
+	if wrappedErr != nil {
+		return fmt.Sprintf("%T", wrappedErr)
+	}
+	return fmt.Sprintf("%T", err)
 }
--- a/pkg/crc/segment/segment_test.go
+++ b/pkg/crc/segment/segment_test.go
@@ -15,6 +15,7 @@ import (
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
+	crcErr "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/code-ready/crc/pkg/crc/version"
@@ -30,9 +31,10 @@ type segmentResponse struct {
 			ExperimentalFeatures bool   `json:"enable-experimental-features"`
 		} `json:"traits"`
 		Properties struct {
-			Error   string `json:"error"`
-			Version string `json:"version"`
-			CPUs    int    `json:"cpus"`
+			Error     string `json:"error"`
+			ErrorType string `json:"error-type"`
+			Version   string `json:"version"`
+			CPUs      int    `json:"cpus"`
 		} `json:"properties"`
 		Type string `json:"type"`
 	} `json:"batch"`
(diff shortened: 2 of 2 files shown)


pkg/crc/segment/segment.go
```diff
@@ -140,3 +144,15 @@ func (l *loggingAdapter) Logf(format string, args ...interface{}) {
...
+
+func errorType(err error) string {
+	var preflightError crcErr.PreflightError
+	switch {
+	case errors.Is(err, crcErr.VMNotExist):
+		return "VMNotExit"
+	case errors.Is(err, &preflightError):
+		return "PreflightError"
+	default:
+		return fmt.Sprintf("%T", err)
```
cfergeau, line 156: Isn't `"%T"` creating strings containing `VMNotExist` or `PreflightError`? A bit surprised they need to be special-cased here.

pkg/crc/segment/segment.go
```diff
@@ -140,3 +144,15 @@ func (l *loggingAdapter) Logf(format string, args ...interface{}) {
 func (l *loggingAdapter) Errorf(format string, args ...interface{}) {
 	logging.Errorf(format, args...)
 }
+
+func errorType(err error) string {
+	var preflightError crcErr.PreflightError
+	switch {
+	case errors.Is(err, crcErr.VMNotExist):
+		return "VMNotExit"
```
cfergeau, line 152: `Exit` typo.

pkg/crc/segment/segment_test.go
```diff
@@ -30,9 +30,10 @@ type segmentResponse struct {
 			ExperimentalFeatures bool   `json:"enable-experimental-features"`
 		} `json:"traits"`
 		Properties struct {
-			Error   string `json:"error"`
-			Version string `json:"version"`
-			CPUs    int    `json:"cpus"`
+			Error     string `json:"error"`
+			ErrorType string `json:"error-type"`
```
cfergeau, line 34: Wondering if in the long run we could send only `ErrorType` and not the error message. To be useful, this implies fine-grained error types rather than always returning `fmt.Errorf`, but this would remove the risk that the error message contains personal user information.

pkg/crc/segment/segment.go
praveenkumar: >  is this because they are wrapped inside another error? If yes, maybe we should do some error unwrapping? Or it does not make sense in general?

All the error types are going to be `*errors.SerializableError` ? we can have well known error using the `unwrap` which is just one level below in the error chain.

cfergeau: > we can have well known error using the unwrap which is just one level below in the error chain.

Yes, I'd test for this specific `*errors.SerializableError` type and unwrap, hopefully it's enough:

```
var serializableErr *errors.SerializableError
if errors.Is(err, &serializableErr) {
    err = serializableErr.Unwrap()
}
return fmt.Sprintf("%T", err)
```
(totally untested code ^^)

pkg/crc/segment/segment_test.go
```diff
@@ -84,7 +86,8 @@ func TestClientUploadWithConsent(t *testing.T) {
 	c, err := newCustomClient(config, filepath.Join(dir, "telemetry"), server.URL)
 	require.NoError(t, err)
 
-	require.NoError(t, c.Upload(context.Background(), "start", time.Minute, errors.New("an error occurred")))
+	err = errors.New("an error occurred")
+	require.NoError(t, c.Upload(context.Background(), "start", time.Minute, fmt.Errorf("%w", err)))
```
praveenkumar, line 90: @cfergeau so I need to change the test to create the error change so unwrap can happen. Since `errors.New` doesn't implement `Unwrap` method from the interface side but `fmt.Errorf` have it.

Also if we are returning an `errors.New` any of upper caller function then on the segment the type would be `<nil>`.

cfergeau: Checking if the error is `SerializableError` before unwrapping should be good enough for now? Or try to `Unwrap()`, and if the error is `nil`, return the initial error instead.

pkg/crc/segment/segment.go
praveenkumar: @cfergeau we can use unwarp method for any error and it will return `nil` if the `err` doesn't implement it otherwise it will return the unwrapped error.

cfergeau: Yup, sounds good as well

pkg/crc/segment/segment_test.go
```diff
@@ -84,7 +87,7 @@ func TestClientUploadWithConsent(t *testing.T) {
 	c, err := newCustomClient(config, filepath.Join(dir, "telemetry"), server.URL)
 	require.NoError(t, err)
 
-	require.NoError(t, c.Upload(context.Background(), "start", time.Minute, errors.New("an error occurred")))
+	require.NoError(t, c.Upload(context.Background(), "start", time.Minute, crcErr.ToSerializableError(crcErr.VMNotExist)))
```
cfergeau, line 90: Might be worth testing both cases, error type implementing `Unwrap()`, and error type not implementing `Unwrap()`

pkg/crc/segment/segment_test.go
```diff
@@ -95,7 +98,8 @@ func TestClientUploadWithConsent(t *testing.T) {
 		require.Equal(t, s.Batch[0].Traits.OS, runtime.GOOS)
 		require.Equal(t, s.Batch[0].Traits.ExperimentalFeatures, true)
 		require.Equal(t, s.Batch[1].Type, "track")
-		require.Equal(t, s.Batch[1].Properties.Error, "an error occurred")
+		require.Equal(t, s.Batch[1].Properties.Error, crcErr.VMNotExist.Error())
+		require.Equal(t, s.Batch[1].Properties.ErrorType, fmt.Sprintf("%T", crcErr.VMNotExist))
```
cfergeau, line 102: Might be worth hardcoding the actual string (`"github.com/code-ready/crc/pkg/errors.VMNotExist"` or whatever it is), as we don't want to inadvertantly change the value of "Properties.ErrorType" for `crcErr.VMNotExist`.

cfergeau (approved):
2 minor comments, I'm fine with the PR either way.

cfergeau (approved):

PR #1992, opened 2021-02-12
Change the error message when running as root on Windows

guillaumerose:
I hope it will give a clue to the user: use an other shell, without
admin.


cfergeau (approved):

gbraad:
@robin-owen You didn't change any comments in the documentation around this, right?

robin-owen:
@gbraad No, we have not changed the wording in the docs.

gbraad:
please don't as the message does not actually fit the use. The user should not be the Administrator user but he is allowed to have administrator role or rights. Windows calls this a Standard user whichbis a member of the Administrator group.

PR #1993, opened 2021-02-12
vsock: export ports when needed

### LINKED ISSUE #1916: crc daemon should not bind ports if not running in vsock mode
As soon as the user launches `crc daemon`, the process bind ports 2222, 6443 and 443 even if it is not using vsock.

`crc start` could ask the daemon to expose the ports right before starting the virtual machine.

guillaumerose:
Fixes #1916 

* Windows: expose the daemon API over a named pipe
* Restrict the API available via vsock
* Use the daemon API in crc start to expose ports on demand.

https://github.com/code-ready/gvisor-tap-vsock/pull/12 needs to be merged first.


anjannath:
> Restrict the API available via vsock to only the connect path

Q: There's an http server also listening on the `~/.crc/tap.sock`, is that the one serving the vsock API?

since i can still do `curl --unix-socket ~/.crc/crc-http.sock http:/unix/network/stats`

guillaumerose:
> Q: There's an http server also listening on the ~/.crc/tap.sock, is that the one serving the vsock API?

This socket only exists on macOS. On Windows, this is a hvsock listener and on Linux, this is a vsock listener. 

I separated the 2 APIs to be more clear: one for vsock where only /connect is available and an other one with everything.

anjannath:
Should we unexpose the ports when `crc stop` is called?

guillaumerose:
@anjannath yes I wanted to do it in a follow up in order to keep the PR a little bit smaller ^^

anjannath:
> @anjannath yes I wanted to do it in a follow up in order to keep the PR a little bit smaller ^^

Okay, initially i thought of the situation where `start->stop->start` would fail trying to bind to the already bound port during expose on the second start, but you already have `isOpened` check

cmd/crc/cmd/daemon.go
```diff
@@ -104,27 +92,44 @@ func captureFile() string {
...
+		mux := http.NewServeMux()
+		mux.Handle("/network/", http.StripPrefix("/network", vn.Mux()))
+		if err := http.Serve(listener, mux); err != nil {
+			errCh <- err
+		}
+	}()
+
+	go func() {
+		mux := http.NewServeMux()
+		mux.Handle("/connect", vn.Mux())
```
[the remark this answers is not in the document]
guillaumerose: Yes I need to write a proper API doc. There is beginning here: https://github.com/code-ready/gvisor-tap-vsock#services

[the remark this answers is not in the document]
guillaumerose: `/connect` is used by a tap device to connect to the network stack and be able to send/receive packets.

cmd/crc/cmd/daemon_unix.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"net"
+	"os"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+)
+
+func controlListener() (net.Listener, error) {
```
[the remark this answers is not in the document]
guillaumerose: yes changed to `vsockListener` and `httpListener`.

cmd/crc/cmd/daemon_windows.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"net"
+
+	"github.com/Microsoft/go-winio"
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+)
+
+func controlListener() (net.Listener, error) {
```
anjannath, line 11: need a `httpListener()` wrapper over this, build fails otherwise

```diff
diff --git a/cmd/crc/cmd/daemon_windows.go b/cmd/crc/cmd/daemon_windows.go
index a188d301..bbdfd313 100644
--- a/cmd/crc/cmd/daemon_windows.go
+++ b/cmd/crc/cmd/daemon_windows.go
@@ -20,3 +20,7 @@ func controlListener() (net.Listener, error) {
        }
        return ln, nil
 }
+
+func httpListener() (net.Listener, error) {
+       return controlListener()
+}
```

guillaumerose: thanks! fixed

guillaumerose:
retest this please

pkg/crc/machine/vsock.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+	return false
+}
+
+func vsockPorts() []types.ExposeRequest {
+	return []types.ExposeRequest{
+		{
+			Local:  fmt.Sprintf(":%d", constants.VsockSSHPort),
+			Remote: "192.168.127.2:22",
```
gbraad, line 47: can we name these variables?

`Remote: [remotehost]:[ssh]`

pkg/crc/machine/vsock.go
```diff
@@ -0,0 +1,58 @@
...
+		},
+		{
+			Local:  ":6443",
+			Remote: "192.168.127.2:6443",
```
gbraad, line 51: `Remote: [remotehost]:[api]`

pkg/crc/machine/vsock.go
```diff
@@ -0,0 +1,58 @@
...
+		},
+		{
+			Local:  ":443",
+			Remote: "192.168.127.2:443",
```
gbraad, line 55: `Remote: [remotehost]:[https]`

pkg/crc/daemonclient/transport_windows.go
```diff
@@ -0,0 +1,18 @@
...
+	"net/http"
+
+	"github.com/Microsoft/go-winio"
+	"github.com/code-ready/crc/pkg/crc/constants"
+)
+
+func transport() *http.Transport {
+	return &http.Transport{
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			return winio.DialPipeContext(ctx, constants.DaemonHTTPSocketPath)
```
gbraad, line 15: cryptic naming

pkg/crc/machine/vsock.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/daemonclient"
+	"github.com/code-ready/gvisor-tap-vsock/pkg/types"
+	"github.com/pkg/errors"
+)
+
+func exposePorts() error {
+	daemonClient := daemonclient.New()
+	requestedPorts := vsockPorts()
```
gbraad, line 14: they aren't requested yet; rather `portsToExpose` or similar.

pkg/crc/machine/vsock.go
```diff
@@ -0,0 +1,58 @@
...
+		return err
+	}
+	var missingPorts []types.ExposeRequest
+	for _, port := range requestedPorts {
+		if !isOpened(alreadyOpenedPorts, port) {
+			missingPorts = append(missingPorts, port)
+		}
+	}
+	for i := range missingPorts {
+		port := &requestedPorts[i]
```
gbraad, line 26: ???

as you are wrote this it looks like you are poking into requestedPorts with the indexer of the missingPorts? It would be better to call this `for p := ` as this is the integer value of the actual port?

gbraad (changes requested):
Indexer issue

cmd/crc/cmd/daemon_windows.go
```diff
@@ -2,8 +2,21 @@ package cmd
...
 	"net"
+
+	"github.com/Microsoft/go-winio"
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
 )
 
 func httpListener() (net.Listener, error) {
-	return nil, nil
+	ln, err := winio.ListenPipe(constants.DaemonHTTPSocketPath, &winio.PipeConfig{
```
gbraad, line 12: `constants.DaemonHTTPSocketPath` This name is quiet cryptic and not even true in this case;

cmd/crc/cmd/daemon.go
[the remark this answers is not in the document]
gbraad: that is not described and therefore confusing.

also, why here
```
		if vsockListener== nil {
			return
		}
```
is not checked?

cmd/crc/cmd/daemon_unix.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+)
+
+func httpListener() (net.Listener, error) {
+	_ = os.Remove(constants.DaemonHTTPSocketPath)
+	ln, err := net.Listen("unix", constants.DaemonHTTPSocketPath)
+	logging.Infof("listening %s", constants.DaemonHTTPSocketPath)
+	if err != nil {
```
gbraad, line 17: Inconsistent with how `vsocklistener` is written. there seems to be more logging here.

cmd/crc/cmd/daemon_windows.go
```diff
@@ -0,0 +1,9 @@
+package cmd
+
+import (
+	"net"
+)
+
+func httpListener() (net.Listener, error) {
+	return nil, nil
```
gbraad: log not supported? or not implemented here?

pkg/crc/constants/constants_linux.go
```diff
@@ -1,7 +1,16 @@
...
+	"path/filepath"
+
+	"github.com/code-ready/gvisor-tap-vsock/pkg/transport"
+)
+
 const (
 	OcExecutableName          = "oc"
 	PodmanExecutableName      = "podman"
 	AdminHelperExecutableName = "admin-helper-linux"
+	TapSocketPath             = transport.DefaultURL
```
gbraad, line 13: `URL` or `socketPath` ? I think the name is vsock is too generic and, and should actually be something like:

`vsockResource` or `vsockPath`... as it is not an actual url, but rather an URI

gbraad (changes requested):
Sometimes have a hard time reading the code as different terminology is used or not indicative enough about what it represents.

pkg/crc/machine/vsock.go
guillaumerose: done

pkg/crc/machine/vsock.go
guillaumerose: done

pkg/crc/machine/vsock.go
guillaumerose: done

pkg/crc/machine/vsock.go
guillaumerose: done

pkg/crc/machine/vsock.go
guillaumerose: yes good point! oups

pkg/crc/constants/constants_linux.go
guillaumerose: I simplified this.

cmd/crc/cmd/daemon_windows.go
guillaumerose: I wanted to add named pipe in a separated commit for more clarity.

cmd/crc/cmd/daemon_unix.go
guillaumerose: done

cmd/crc/cmd/daemon.go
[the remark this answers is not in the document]
guillaumerose: I wanted to introduce named pipe in a proper commit.

cmd/crc/cmd/daemon_windows.go
guillaumerose: I added a new variable for named pipe. It should be better now.

guillaumerose:
I went on all comments I guess. PTAL :)

pkg/crc/machine/vsock.go
```diff
@@ -0,0 +1,65 @@
...
+		return err
+	}
+	var missingPorts []types.ExposeRequest
+	for _, port := range portsToExpose {
+		if !isOpened(alreadyOpenedPorts, port) {
+			missingPorts = append(missingPorts, port)
+		}
+	}
+	for i := range missingPorts {
+		port := &missingPorts[i]
```
cfergeau, line 26: nit: not really consistent,
```
for _, port := range portsToExpose {
```

and a few lines below,
```
for i := range missingPorts {
		port := &missingPorts[i]
```

pkg/crc/machine/vsock.go
```diff
@@ -0,0 +1,65 @@
...
+		if err := daemonClient.Expose(port); err != nil {
+			return errors.Wrapf(err, "failed to expose port %s -> %s", port.Local, port.Remote)
+		}
```
cfergeau, line 29: Not sure failure to expose port 443 should be fatal, but this can be addressed later.

cmd/crc/cmd/daemon_windows.go
```diff
@@ -17,5 +19,14 @@ func vsockListener() (net.Listener, error) {
 }
 
 func httpListener() (net.Listener, error) {
-	return nil, nil
+	ln, err := winio.ListenPipe(constants.DaemonHTTPNamedPipe, &winio.PipeConfig{
+		MessageMode:      true,  // Use message mode so that CloseWrite() is supported
+		InputBufferSize:  65536, // Use 64KB buffers to improve performance
```
cfergeau, line 24: The unit is `kB`, not `KB` :(

cmd/crc/cmd/daemon_windows.go
```diff
@@ -17,5 +19,14 @@ func vsockListener() (net.Listener, error) {
...
+		OutputBufferSize: 65536,
+	})
+	logging.Infof("listening %s", constants.DaemonHTTPNamedPipe)
```
cfergeau, line 27: dunno if it should be `Info` or `Debug`?

cfergeau (dismissed):
Looks good to me apart from minor nits.

cfergeau:
Actually missed the first commit in my latest review, not sure ho wto cancel approval /o\

cmd/crc/cmd/daemon.go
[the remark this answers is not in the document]
cfergeau: Shouldn't `/connect` be `types.ConnectPath`?

cfergeau (commented):
Hard to get an idea about the first commit as I'm not fully sure what the initial state is, and why the final state (`/connect` and `/network`) is better. It seems by accessing `/network/connect` I could access the `/connect` path? Is this a problem?

pkg/crc/machine/vsock.go
guillaumerose: It happens very early in the process. Even before the VM is started, so it's easy to fix it and launch crc start again. 

If it doesn't work, the console or API wont be reachable, so it's mandatory.

cmd/crc/cmd/daemon_windows.go
guillaumerose: done

cmd/crc/cmd/daemon_windows.go
guillaumerose: I found it useful to see something from the daemon when starting it.

pkg/crc/machine/vsock.go
cfergeau: Being able to reach the console is not mandatory to use the cluster, is it? People need to be aware the console is broken, but maybe their usecase does not involve the console? For example https://code-ready.github.io/crc/#deploying-sample-application-with-odo_gsg

guillaumerose:
> Hard to get an idea about the first commit as I'm not fully sure what the initial state is, and why the final state (/connect and /network) is better. It seems by accessing /network/connect I could access the /connect path? Is this a problem?

Previously the whole API was available inside the VM. So if we go forward with https://github.com/code-ready/crc/pull/1978, you could start/stop/status (or maybe in the future set the pull secret). I prefer to keep the API smaller in vsock (just /connect).
With that, you get Internet access + future access to an internal API exposed on the gateway (like https://github.com/code-ready/crc/pull/1793/files#diff-6a402f241a359c13dc3183a088f87a7f0a331df8350209c3d674f40f7119cef1R150). 

Yes /network/connect is still available on the host. We could drop it.

pkg/crc/machine/vsock.go
guillaumerose: Oh yes, I see! Fair point, we can fail for apiserver and display a (very big) warning for the console port.

cfergeau:
> I separated the 2 APIs to be more clear: one for vsock where only /connect is available and an other one with everything.

What is imo confusing is that they both use `vn.Mux()`, I'd expect one mux handling only `/connect`, and a different one handling the rest of the API?

cfergeau (commented):
If I'm not mistaken, the first commit:
- introduces `tap.sock` on macOS so that there are 2 distinct sockets, one we can use for vsock, and another one which can be used for the 'http' API
- renames `network.sock` to `crc-http.sock` in preparation for using it for more generic API
- refactors the existing code so that the vsock API is exposed under `/connect` while the control API will be exposed under `/network`, freeing the 'top-level' namespace for additional http API

This commit would probably be much easier to follow and review if it had been split in smaller steps rather than doing all of this at the same time.

cfergeau:
> This commit would probably be much easier to follow and review if it had been split in smaller steps rather than doing all of this at once.

Experimented with this, maybe split a bit too much, but imo something like https://github.com/cfergeau/crc/commits/restrictvsockapi-split is much easier to review, even if the commit logs would need a bit more work.

cfergeau (approved):
First commit in the series is much harder to review than it should :-/ The current http configuration can probably be skipped when vsock is not being used?
Giving an ack so that we can move forward, and improve later.

PR #1999, opened 2021-02-15
Remove code that changes the proxy env. variable in crc status

guillaumerose:
Small cleanup

### DIFF
--- a/pkg/crc/machine/status.go
+++ b/pkg/crc/machine/status.go
@@ -39,11 +39,6 @@ func (client *client) Status() (*ClusterStatusResult, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "Error loading bundle metadata")
 	}
-	proxyConfig, err := getProxyConfig(crcBundleMetadata.ClusterInfo.BaseDomain)
-	if err != nil {
-		return nil, errors.Wrap(err, "Error getting proxy configuration")
-	}
-	proxyConfig.ApplyToEnvironment()
 
 	ip, err := getIP(host, client.useVSock())
 	if err != nil {


cmd/crc/cmd/oc_env.go
```diff
@@ -27,12 +27,6 @@ func runOcEnv(args []string) error {
 		return fmt.Errorf("Error running the oc-env command: %s", err.Error())
 	}
 
-	client := newMachine()
-	consoleResult, err := client.GetConsoleURL()
-	if err != nil {
-		return err
-	}
-	proxyConfig := consoleResult.ClusterConfig.ProxyConfig
```
cfergeau, line 35: Is `NoProxy` correct without this? `NoProxy` needs to contain `api.xxx`, in order to know what `xxx` is, we need the bundle which is being used, in order to know which bundle is currently being used, we need to look at the machine metadata.

guillaumerose: Yes this is a mistake, thanks! 
Should I drop this? Hardcode the domain? It is sad to have crc oc-env failing this often only for proxy setup (which are rare setup). 
Should this check only performed when a proxy is enabled ?

gbraad:
Oc-env is only intended for use with the CRC environment, which will only be useful when a VM with deployed bundle exists.

So, what is the idea here?

gbraad (changes requested):
See previous comments as I ma not sure about the intentions here.

cmd/crc/cmd/oc_env.go
guillaumerose: I dropped the commit. Too complicated for little result.

guillaumerose:
The idea here was to have `crc oc-env` always working. Or at least in 90% of cases.

guillaumerose:
I only kept the small cleanup commit.

cmd/crc/cmd/oc_env.go
cfergeau: > Hardcode the domain? It is sad to have crc oc-env failing this often only for proxy setup (which are rare setup).
Should this check only performed when a proxy is enabled ?

A mix of these suggestions could work? If proxy is configured, try to get proxy info through machine metadata, if this fails, fallback to hardcoded domain, and warn that NO_PROXY could be incorrect?

gbraad (approved):

cmd/crc/cmd/oc_env.go
guillaumerose: Will do that in a follow up

PR #2000, opened 2021-02-15
Distuiguish first start from others call of crc start 

guillaumerose:
With this, we will better understand the duration of the first crc start.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -24,6 +24,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/ssh"
 	crcssh "github.com/code-ready/crc/pkg/crc/ssh"
 	"github.com/code-ready/crc/pkg/crc/systemd"
+	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/code-ready/crc/pkg/libmachine"
 	"github.com/code-ready/crc/pkg/libmachine/host"
 	crcos "github.com/code-ready/crc/pkg/os"
@@ -104,7 +105,10 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 	if err != nil {
 		return nil, errors.Wrap(err, "Cannot determine if VM exists")
 	}
+
 	if !exists {
+		telemetry.SetStartType(ctx, telemetry.CreationStartType)
+
 		// Ask early for pull secret if it hasn't been requested yet
 		_, err = startConfig.PullSecret.Value()
 		if err != nil {
@@ -167,12 +171,16 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 			if err != nil {
 				return nil, errors.Wrap(err, "Cannot create cluster configuration")
 			}
+
+			telemetry.SetStartType(ctx, telemetry.AlreadyRunningStartType)
 			return &StartResult{
 				Status:         vmState,
 				ClusterConfig:  *clusterConfig,
 				KubeletStarted: true,
 			}, nil
 		}
+
+		telemetry.SetStartType(ctx, telemetry.StartStartType)
 
 		logging.Infof("Starting CodeReady Containers VM for OpenShift %s...", crcBundleMetadata.GetOpenshiftVersion())
 
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -90,6 +90,10 @@ func (client *client) updateVMConfig(startConfig StartConfig, api libmachine.API
 }
 
 func (client *client) Start(ctx context.Context, startConfig StartConfig) (*StartResult, error) {
+	telemetry.SetCPUs(ctx, startConfig.CPUs)
+	telemetry.SetMemory(ctx, uint64(startConfig.Memory)*1024*1024)
+	telemetry.SetDiskSize(ctx, uint64(startConfig.DiskSize)*1024*1024*1024)
+
 	if err := client.validateStartConfig(startConfig); err != nil {
 		return nil, err
 	}
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -1,6 +1,7 @@
 package machine
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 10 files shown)


pkg/crc/machine/start.go
```diff
@@ -104,7 +112,10 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 	if err != nil {
 		return nil, errors.Wrap(err, "Cannot determine if VM exists")
 	}
+
 	if !exists {
+		telemetry.SetContextProperty(ctx, telemetryStartType, firstStart)
+
```
cfergeau, line 118: Wondering if `firstStart` should be set in `startResult`, and then added to telemetry in the `cmd/` package? This would keep the `machine` package separate from `telemetry`/`context`. If returned as part of `startResult`, rather than `firstStart` this could be returned as `vmCreated` or similar name.

guillaumerose: startResult is not returned in case of error. It's good to know an error occurred in the context of a first start or so. 
I hope context will be also used by the daemon for cancellation. If the client aborts, then the daemon cut somewhere in a wait loop maybe.

pkg/crc/machine/start.go
```diff
@@ -167,6 +180,8 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 			if err != nil {
 				return nil, errors.Wrap(err, "Cannot create cluster configuration")
 			}
+
+			telemetry.SetContextProperty(ctx, telemetryStartType, alreadyRunning)
```
cfergeau, line 184: This one feels odd, we will be sending `start-type: first-start`, `start-type: start`, or `start-type:start` followed by `start-type: already-running`. I'd expect only the latter to be sent in the last case, not to get 2 start-type events.

pkg/crc/machine/start.go
cfergeau: > startResult is not returned in case of error. It's good to know an error occurred in the context of a first start or so.

Indeed, but this does not mean the only way to achieve that is by adding a telemetry dependency to the machine package :) I also don't really consider "this package will be using cancellation through `context`" as a valid reason for making the package dependent on `telemetry`. I'm still undecided whether the `telemetry` package is a low-level one which is fine to use in most packages, or if we should try to keep it as a high level thing. 

Thinking while writing this message, I'd say what is missing is a slightly higher level telemetry API for use in other packages, ie some `telemetry.SetStartType()`, `telemetry.SetDiskSize()`, ... so that it's harder to misuse, and we keep some central list of telemetry data which is sent.

pkg/crc/machine/start.go
```diff
@@ -33,7 +34,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-const minimumMemoryForMonitoring = 14336
+const (
+	minimumMemoryForMonitoring = 14336
+
+	telemetryStartType = "start-type"
+	alreadyRunning     = "already-running"
+	firstStart         = "first-start"
```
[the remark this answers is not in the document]
guillaumerose: Yes not really inspired by the name. creation / start / already-running then ?

pkg/crc/machine/start.go
guillaumerose: already-running is the case where the duration is almost 0. 

This call overrides the value of previously set. Not ideal.. I will remove this.

pkg/crc/machine/start.go
guillaumerose: I will do that!

pkg/crc/machine/start.go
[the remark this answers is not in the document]
cfergeau: `creation`/`created`/`vm-created`/`initial-start`/`fresh-start`. Not of these are great, but I'd say less ambiguous than `first-start`

pkg/crc/machine/start.go
cfergeau: Yeah, took me a long time to realize `SetContextProperty` overwrites the previous value, so the behaviour is fine (only one `start-type` event sent). Once wrapped in a slightly higher level API, should be all fine!

pkg/crc/telemetry/telemetry.go
```diff
@@ -50,9 +50,25 @@ func GetContextProperties(ctx context.Context) map[string]interface{} {
...
 
-func SetContextProperty(ctx context.Context, key string, value interface{}) {
+func setContextProperty(ctx context.Context, key string, value interface{}) {
 	properties := propertiesFromContext(ctx)
 	if properties != nil {
 		properties.set(key, value)
 	}
 }
+
+func SetCPUs(ctx context.Context, value interface{}) {
```
cfergeau, line 60: `value int`?

pkg/crc/telemetry/telemetry.go
```diff
@@ -50,9 +50,25 @@ func GetContextProperties(ctx context.Context) map[string]interface{} {
...
+	setContextProperty(ctx, "cpus", value)
+}
+
+func SetMemory(ctx context.Context, value interface{}) {
+	setContextProperty(ctx, "memory", value)
```
cfergeau, line 65: `value (u?)int64`

pkg/crc/telemetry/telemetry.go
```diff
@@ -50,9 +50,25 @@ func GetContextProperties(ctx context.Context) map[string]interface{} {
...
+}
+
+func SetDiskSize(ctx context.Context, value interface{}) {
```
cfergeau, line 68: same here

pkg/crc/telemetry/telemetry.go
```diff
@@ -50,9 +50,25 @@ func GetContextProperties(ctx context.Context) map[string]interface{} {
...
+	setContextProperty(ctx, "disk-size", value)
+}
+
+func SetConfigurationKey(ctx context.Context, value interface{}) {
```
cfergeau, line 72: `value string`

pkg/crc/machine/start.go
```diff
@@ -167,13 +171,17 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
...
+			telemetry.SetStartType(ctx, telemetry.AlreadyRunningStartType)
 			return &StartResult{
 				Status:         vmState,
 				ClusterConfig:  *clusterConfig,
 				KubeletStarted: true,
 			}, nil
 		}
 
+		telemetry.SetStartType(ctx, telemetry.StartStartType)
+
```
cfergeau, line 188: I think it was better when it was done earlier in one of the previous iterations.

cfergeau (approved):

PR #2001, opened 2021-02-15
Update bundle from 4.6.16 to 4.7.0

praveenkumar:


praveenkumar:
ok to test

guillaumerose:
For enabling the monitoring stack, there is now 2 objects to activate: the CO object and the deployment.

praveenkumar:
> or enabling the monitoring stack, there are now 2 objects to activate: the CO object and the deployment.

@guillaumerose Doesn't enablement of CO object also take care of the deployment or because deployment is also part of override so we need to explicit remove it?

praveenkumar:
@guillaumerose When only removing the cluster operator kind ( https://github.com/code-ready/snc/blob/master/cvo-overrides.yaml#L8-L12 ) then monitoring appears  to `oc get co` output but it doesn't have any status but when you check
the resource definition, it does have `manager: cluster-version-operator`.

```
$ oc get co
NAME                                       VERSION      AVAILABLE   PROGRESSING   DEGRADED   SINCE
[...]
machine-approver                           4.7.0-rc.2   True        False         False      7d21h
marketplace                                4.7.0-rc.2   True        False         False      4m34s
monitoring                                                                                   
network                                    4.7.0-rc.2   True        False         False      7d21h
node-tuning                                4.7.0-rc.2   True        False         False      7d21h
```

But if we remove the deployment override ( https://github.com/code-ready/snc/blob/master/cvo-overrides.yaml#L3-L6 ) which is what we are currently doing then the monitoring does appear to `oc get co` output with it status but then the resource definition have `manager: operator` instead `cluster-version-operator` so we might need to remove both from override.

gbraad:
We do not yet have a go to release with 4.7 due to https://bugzilla.redhat.com/show_bug.cgi?id=1927263

gbraad:
We have a Go!

PR #2003, opened 2021-02-15
Be sure that crc start ends up with a working cluster

guillaumerose:
Most of the time, at the end of crc start, the cluster is broken: 3min is not enough. 
@jsliacan solved this in the CI by waiting 3 consecutive running statuses (1min interval). 

This PR replaces the sleep(3 min) by a check of 3 consecutive checks (30s interval). 

It also changes the output of the crc start. It says now why it is waiting (and this is not CRC fault).

I found this PR really needed since we changed the output of crc start. Now if the user clicks on the console URL, it will simply work.

4.6.16 bundle: 

```
INFO Starting OpenShift cluster ... [waiting for the cluster to stabilize] 
INFO 2 operators are progressing: authentication, image-registry 
INFO 3 operators are progressing: authentication, image-registry, operator-lifecycle-manager-packageserver 
INFO 3 operators are progressing: authentication, dns, image-registry 
INFO Operator image-registry is progressing       
INFO 2 operators are not available: authentication, openshift-apiserver 
INFO Operator kube-apiserver is progressing       
INFO Operator kube-apiserver is progressing       
INFO Operator kube-apiserver is progressing       
INFO Operator kube-apiserver is progressing       
INFO Operator kube-apiserver is progressing       
INFO All operators are available. Ensuring stability ... 
INFO Operators are stable (2/3) ...               
INFO Operators are stable (3/3) ...    
```

4.6.15 bundle (4m10s total wait):
```
INFO Starting OpenShift cluster ... [waiting for the cluster to stabilize] 
INFO 7 operators are progressing: authentication, dns, image-registry, network, openshift-controller-manager, operator-lifecycle-manager-packageserver, service-ca 
INFO 7 operators are progressing: authentication, dns, image-registry, network, openshift-controller-manager, operator-lifecycle-manager-packageserver, service-ca 
INFO 4 operators are progressing: authentication, image-registry, openshift-controller-manager, service-ca 
INFO 4 operators are progressing: authentication, console, image-registry, service-ca 
INFO Operator ingress is degraded                 
INFO Operator authentication is progressing       
INFO All operators are available. Ensuring stability ... 
INFO Operators are stable (2/3) ...               
INFO Operators are stable (3/3) ...    
```

### DIFF
--- a/pkg/crc/cluster/status.go
+++ b/pkg/crc/cluster/status.go
@@ -24,7 +24,13 @@ func WaitForClusterStable(ocConfig oc.Config, monitoringEnabled bool) error {
 			// update counter for consecutive matches
 			if status.IsReady() {
 				count++
+				if count == 1 {
+					logging.Info("All operators are available. Ensuring stability ...")
+				} else {
+					logging.Infof("Operators are stable (%d/%d) ...", count, numConsecutive)
+				}
 			} else {
+				logging.Info(status.String())
 				count = 0
 			}
 			// break if done
--- a/pkg/crc/cluster/status.go
+++ b/pkg/crc/cluster/status.go
@@ -1,0 +1,42 @@
+package cluster
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/oc"
+)
+
+// WaitForClusterStable checks that the cluster is running a number of consecutive times
+func WaitForClusterStable(ocConfig oc.Config, monitoringEnabled bool) error {
+	startTime := time.Now()
+
+	retryDuration := 30 * time.Second
+	retryCount := 20 // 10 minutes
+
+	numConsecutive := 3
+	var count int // holds num of consecutive matches
+
+	for i := 0; i < retryCount; i++ {
+		status, err := GetClusterOperatorsStatus(ocConfig, monitoringEnabled)
+		if err == nil {
+			// update counter for consecutive matches
+			if status.IsReady() {
+				count++
+			} else {
+				count = 0
+			}
+			// break if done
+			if count == numConsecutive {
+				logging.Debugf("Cluster took %s to stabilize", time.Since(startTime))
+				return nil
+			}
+		} else {
+			count = 0
+		}
+		time.Sleep(retryDuration)
+	}
+
+	return fmt.Errorf("cluster operators are still not stable after %s", time.Since(startTime))
+}
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -20,6 +20,10 @@ type Status struct {
 	Disabled    bool
 }
 
+func (status *Status) IsReady() bool {
+	return status.Available && !status.Progressing && !status.Degraded && !status.Disabled
+}
+
 func GetClusterOperatorStatus(ocConfig oc.Config, operator string) (*Status, error) {
 	return getStatus(ocConfig, defaultIgnoredClusterOperators, []string{operator})
 }
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -3,10 +3,11 @@ package cluster
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/oc"
-
 	openshiftapi "github.com/openshift/api/config/v1"
 )
 
@@ -18,6 +19,35 @@ type Status struct {
 	Degraded    bool
 	Progressing bool
 	Disabled    bool
+
+	progressing []string
+	degraded    []string
+	unavailable []string
+}
+
+func (status *Status) String() string {
+	if len(status.progressing) == 1 {
+		return fmt.Sprintf("Operator %s is progressing", status.progressing[0])
+	} else if len(status.progressing) > 1 {
+		return fmt.Sprintf("%d operators are progressing: %s", len(status.progressing), strings.Join(status.progressing, ", "))
+	}
+
+	if len(status.degraded) == 1 {
+		return fmt.Sprintf("Operator %s is degraded", status.degraded[0])
+	} else if len(status.degraded) > 0 {
+		return fmt.Sprintf("%d operators are degraded: %s", len(status.degraded), strings.Join(status.degraded, ", "))
+	}
+
+	if len(status.unavailable) == 1 {
+		return fmt.Sprintf("Operator %s is not yet available", status.unavailable[0])
+	} else if len(status.unavailable) > 0 {
+		return fmt.Sprintf("%d operators are not available: %s", len(status.unavailable), strings.Join(status.unavailable, ", "))
+	}
+
+	if status.IsReady() {
+		return "All operators are ready"
+	}
+	return "Operators are not ready yet"
 }
 
 func (status *Status) IsReady() bool {
@@ -65,16 +95,19 @@ func getStatus(ocConfig oc.Config, ignoreClusterOperators, selector []string) (*
 			case "Available":
 				if con.Status != "True" {
 					logging.Debug(c.ObjectMeta.Name, " operator not available, Reason: ", con.Reason)
+					cs.unavailable = append(cs.unavailable, c.ObjectMeta.Name)
 					cs.Available = false
 				}
 			case "Degraded":
 				if con.Status == "True" {
 					logging.Debug(c.ObjectMeta.Name, " operator is degraded, Reason: ", con.Reason)
+					cs.degraded = append(cs.degraded, c.ObjectMeta.Name)
 					cs.Degraded = true
 				}
 			case "Progressing":
 				if con.Status == "True" {
 					logging.Debug(c.ObjectMeta.Name, " operator is still progressing, Reason: ", con.Reason)
+					cs.progressing = append(cs.progressing, c.ObjectMeta.Name)
 					cs.Progressing = true
 				}
 			case "Upgradeable":
--- a/pkg/crc/cluster/clusteroperator_test.go
+++ b/pkg/crc/cluster/clusteroperator_test.go
@@ -16,6 +16,7 @@ var (
 	progressing = &Status{
 		Available:   true,
 		Progressing: true,
+		progressing: []string{"authentication"},
 	}
 )
 
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -36,7 +36,7 @@ func eventuallyWriteKubeconfig(ocConfig oc.Config, ip string, clusterConfig *Clu
 		if err != nil {
 			return &errors.RetriableError{Err: err}
 		}
-		if isReady(status) {
+		if status.IsReady() {
 			return nil
 		}
 		return &errors.RetriableError{Err: goerrors.New("cluster operator authentication not ready")}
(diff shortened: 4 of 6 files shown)


cfergeau (changes requested):
` Wait for the cluster to stabilize instead of waiting 3min` needs to be more explicit about what "stabilize" means, about the impact on startup time, ...

pkg/crc/cluster/status.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/oc"
+)
+
+// WaitForClusterStable checks that the cluster is running a number of consecutive times
+func WaitForClusterStable(ocConfig oc.Config, monitoringEnabled bool) error {
+	startTime := time.Now()
+
+	retryDuration := 30 * time.Second
+	retryCount := 30
```
praveenkumar, line 16: So with this we are actually going to wait for around 900 sec before tell user that cluster is still in unstable state?

guillaumerose: Logs will tell him it's stuck before but yes, crc will be stuck 15min (this is what we have in the CI).

gbraad:
> Checking stability

This indicates uncertainty.  Can we use availability?

gbraad:
> Checking stability

This indicates uncertainty.  Can we use availability?

jsliacan:
> This indicates uncertainty. Can we use availability?

In this case, it is probably stability though... We want "availability over time", i.e. stability, imo. But I would maybe suggest to change `Checking stability...` to `Ensuring stability...`. But that's just being petty.

gbraad:
> Ensuring stability

yep... as 'we' can't ensure that. we can only check for it.
although where does this 'stability' come from? what do we check for?
This is why I mentioned availability; but it should include the work 'operator' as this is what is being checked.

pkg/crc/cluster/status.go
guillaumerose: I put 10min now

guillaumerose:
@gbraad @jsliacan I changed the wording with `Ensuring stability`. Check the PR description to see the final output.

guillaumerose:
@cfergeau I updated the commit log, it should be better

pkg/crc/cluster/cluster.go
```diff
@@ -287,7 +287,7 @@ func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret Pul
 	if _, _, err := sshRunner.Run(fmt.Sprintf("test -e %s", vmPullSecretPath)); err == nil {
 		return nil
 	}
-	logging.Info("Adding user's pull secret to instance disk...")
+	logging.Info("Adding user's pull secret to instance disk ...")
```
cfergeau, line 290: Really not sure what the goal is here? Consistency with the other log messages? Or something else?
```
$ git grep "[^ ]\.\.\.\"" :^vendor |wc -l
21

$ git grep " \.\.\.\"" :^vendor |wc -l
17
```

There are currently more 'no space before ...' messages than 'space before ...' messages.

pkg/crc/cluster/status.go
```diff
@@ -24,7 +24,13 @@ func WaitForClusterStable(ocConfig oc.Config, monitoringEnabled bool) error {
 			// update counter for consecutive matches
 			if status.IsReady() {
 				count++
+				if count == 1 {
+					logging.Info("All operators are available. Ensuring stability ...")
+				} else {
+					logging.Infof("Operators are stable (%d/%d) ...", count, numConsecutive)
+				}
 			} else {
+				logging.Info(status.String())
```
cfergeau, line 33: I'd need to try it to see how this looks exactly, but this might be better as `Debug`?

cfergeau (changes requested):
I still don't really understand the scope of the changes with the commit log.. This replaces the 3 minutes wait, with 'something'. I need to take a look at the code to understand that this 'something' seems to be waiting for all the operators to be in the 'available' state (ie 'oc get co' is all good), and this condition must be true 3 times in a row, with 30s wait between each check.
The implications are:
1. after the cluster is stable, we'll wait one more minute to make sure this was not a transient state
2. if for some reason the cluster never becomes fully available, we'll have waited at least 10 minutes in total compared to 3 minutes before

I don't know what the impact is on `crc start` run time? What was the extent of the issues which were caused by the 'old way' for users?

pkg/crc/cluster/cluster.go
guillaumerose: Yes the goal was consistency. I didnt check for others! Let s drop this commit.

guillaumerose:
I run this couple of times and here are some numbers I got:

```
 level=debug msg="Cluster took 5m5.77392309s to stabilize"
 level=debug msg="Cluster took 3m14.719086407s to stabilize"
 level=debug msg="Cluster took 4m54.412423508s to stabilize"
 level=debug msg="Cluster took 9m53.335788713s to stabilize"
 level=debug msg="Cluster took 7m26.386394568s to stabilize"
 level=debug msg="Cluster took 4m10.586245613s to stabilize"
 level=debug msg="Cluster took 4m4.36198784s to stabilize"
```

I can add code that checks if the state is not ready and is not moving for X minutes, then return an error. This will avoid waiting up to 10min.

pkg/crc/cluster/status.go
guillaumerose: The goal is to show the user progress. This displays the lines:
```
INFO 3 operators are progressing: authentication, dns, image-registry 
INFO Operator image-registry is progressing       
INFO 2 operators are not available: authentication, openshift-apiserver 
```

gbraad:
7 and 9? Did you use the machine in the meantime?

gbraad:
Without reading the commit logs and the description her I understand it was about the state of the operators. This is confirmed by the actual code.

So perhaps it is better not to have any commit logs in that case? And rely on the issue description.

We keep going around in circles. I would like to stress; working code is more important than extensive documentation (git descriptions). If the code I behaviour matches expectations I deem that more important.

gbraad:
The original implementation left people guessing about the state and just bailed out and saud; we don't know, we don't wait any longer, it returned to the prompt and let's the user guess if things work.

Anythibg around this change is in line with the reported issues to handle the startup; check state of the operators.

So even if it looks like the command runs longer, it is effectively a much better outcome.

gbraad (approved):

pkg/crc/cluster/status.go
gbraad: Clearliness (avoid confusion) is good. at this point, We can always decide to drop it if it becomes/feels like noise.

guillaumerose:
9min53 and 7min are made with crc_libvirt_4.6.16.crcbundle bundle.
Others are with crc_libvirt_4.6.15.crcbundle (the bundle in the release).

@praveenkumar any thoughts on this?

jsliacan:
@guillaumerose , not sure if this helps with deciding, but woopra gives this for start times (with 3min wait regardless of the state of the cluster):

| start duration  [min]    | % of actions |
| ----------- | ----------- |
| 3-4      | 1      |
| 4-5 | 45 |
| 5-6 | 23 |
| 6-7 | 12 |
| 7-8 | 6.5 |
| 8-9 | 4 |
| 9-10 | 2.2 |
| 10-11 | 1.33 |
| 11+ | 1 |

So your PR would be just at the back of the pile. I guess the question is: Those 81% (1+45+23+12) of the users who get a faster start than your new one, do they get a cluster that's not ready? If yes, then it's definitely an improvement. If not, then I guess it's back to the value judgement (better fast with fewer guarantees, or slower with more reliability)? :) Glad to be of no help whatsoever!

pkg/crc/cluster/status.go
gbraad: if people are tired of waiting, they will break out. something in that case to capture and send as telemetry

gbraad:
> 9min53 and 7min are made with crc_libvirt_4.6.16.crcbundle bundle.
> Others are with crc_libvirt_4.6.15.crcbundle (the bundle in the release).

So this is based on the bundle. Would like to see if a monitoring pipeline would catch that too.
it is concerning that there are clear jumps in startup time for certain bundles.

pkg/crc/cluster/status.go
```diff
@@ -0,0 +1,40 @@
...
+				count++
+			} else {
+				count = 0
+			}
+			// break if done
+			if count == numConsecutive {
+				logging.Debugf("Cluster took %s to stabilize", time.Since(startTime))
+				return nil
+			}
+		}
```
cfergeau, line 41: `else count = 0` ?

cfergeau (commented):
How good/bad is the cluster if `WaitForClusterStable` returns the first time `GetClusterOperatorsStatus` says it's ready? if the cluster is bad at that time, what about the 2nd time? For CI, we need to be fully sure that there won't be transient errors when interacting with the cluster, even if this means waiting a bit longer. For normal crc use though, it may be ok to wait less, and maybe have to retry once or twice what you do next if you are not lucky and the cluster was not fully ready?

cfergeau:
> I would like to stress; working code is more important than extensive documentation (git descriptions). If the code I behaviour matches expectations I deem that more important.

If the expectations are not explicitly listed in the commit log (or in the issue or in the PR description), then the developer may intend to do one thing, the code may be correct, but doing something different, and the reviewer might expect a 3rd behaviour. If you state what you want to do in the commit log, you remove this ambiguity, and allow to actually check that the code matches expectations.

> So perhaps it is better not to have any commit logs in that case? And rely on the issue description ?

Replace `git log` with numbers that you have to c&p in your browser after opening N tab and going to the issues page? This does not seem very efficient. And the outcome of issues with dozens of comments need to be summarized somewhere.

guillaumerose:
CI failed right now when I removed the check on the test side.

```
time="2021-02-16T14:28:03Z" level=info msg="2 operators are progressing: kube-apiserver, operator-lifecycle-manager-packageserver"
time="2021-02-16T14:28:33Z" level=error msg="Cluster is not ready: cluster operators are still not stable after 10m15.219159472s"
```

The cluster is unusable at all if the kube-apiserver is not ready (even after 10min!).

retest this please

jsliacan:
> CI failed right now when I removed the check on the test side.

@guillaumerose I think this was the case with one of the older bundles as well (around the time this test went in). I remember first having 3 conseq. 1min waits within 10 mins and failing CI. Then I increased it to 15min. Not sure if it's 4.6.16 specific or if it's a returning issue.

gbraad:
> If the expectations are not explicitly listed in the commit log

the expectation match the outcome of what previously existed in the testcase.

praveenkumar:
@jsliacan @adrianriobo Can you folks also test this patch to downstream CI, I am just worry about we are using `10m` time now to check if cluster is stable and the CI test used `15m` . This shouldn't impact on the internal CI pipeline which took some time now for being stabilize.

guillaumerose:
What is your final word @jsliacan ? What should be the timeout ? 15min ?

cfergeau:
I ran some testing on this, and the wait for 3 consecutive 'good' states still feels quite arbitrary :-/ On initial start, once it's stabilized once, waiting 1 more minute does not seem really useful. On the other hand, on restarts, it's needed, as it very quickly succeed once at the 'stable' check, but then some operators get degraded. There was even one time when right after 3 successful checks, some operators became degraded/restarting (authentication, ...), so `crc start` finished successfully, but `oc get co` was not 'clean'

pkg/crc/cluster/cluster.go
cfergeau: (this commit is still present)

pkg/crc/cluster/status.go
cfergeau: Still an issue in current version of the PR.

cfergeau (commented):
Still not convinced displaying the operators which are still in progress is very useful to the user (sometimes everything is fine initially, then we wait for a few operators, then some more ,the some disappear, ...) Does not really give a feeling of 'progression', just hat something is going on, which could be achieved with a spinner or such.

jsliacan:
I spent good time on this today and I'm consistently getting:
- fresh start ~7-9min to stabilize
- start from stop ~4-5min to stabilize (I didn't find any run, out of 10, that passed this and some operator wasn't running in the next 2 minutes as well)
Seems (at the very least) no worse than the original approach in terms of handing over a running cluster. @cfergeau's comment worries me a bit though ...

I don't mind all the output, but it's easy to agree with Christophe about non-monotonicity of the information displayed.

PR #2008, opened 2021-02-17
Use admin-helper daemon on Windows to avoid UAC prompts

guillaumerose:
Requires https://github.com/code-ready/admin-helper/pull/3

This change removes all UAC prompts (at crc start) on Windows by adding a small admin helper daemon. It accepts add/remove commands. 

Once the daemon PR is merged, I will do a new release and add a preflight check that install the daemon automatically.


anjannath:
@guillaumerose This is only for the vsock network mode right? for the default mode UAC prompt on `start` in windows is shown for configuring the dns server address in the vswitch.

for vsock network mode we need to keep the hosts file updated with the .apps routes, which i assume will be done by the route-controller? or do we need to update the hosts file during crc start as well?

guillaumerose:
> @guillaumerose This is only for the vsock network mode right? for the default mode UAC prompt on start in windows is shown for configuring the dns server address in the vswitch.

Yes it is only for vsock

> for vsock network mode we need to keep the hosts file updated with the .apps routes, which i assume will be done by the route-controller? or do we need to update the hosts file during crc start as well?

The route controller talks to the unprivileged crc daemon (via http://gateway.crc.testing in the VM), then the daemon talks to the admin helper daemon (via named pipe crc-admin-helper). 

All routes (except ssh) will be added and updated by the route controller, yes. We can still help him at start and setup the default ones.

pkg/crc/adminhelper/hosts_nowin.go
```diff
@@ -3,10 +3,26 @@
...
+
+func instance() helper {
+	return &cliHelper{}
+}
+
+type cliHelper struct {
+}
+
+func (c *cliHelper) Add(req *types.AddRequest) error {
+	return execute(append([]string{"add", req.IP}, req.Hosts...)...)
```
cfergeau, line 23: maybe get rid of `execute`?

cfergeau (commented):
If I'm not mistaken, `github.com/code-ready/admin-helper/pkg/client` is not merged yet, and the preflights to start admin-helper as a daemon are also missing?

cfergeau:
(rebased at https://github.com/cfergeau/crc/commits/daemonhelper )

gbraad:
Is the executable correctly named?

pkg/os/windows/powershell/powershell_windows.go
```diff
@@ -95,7 +95,13 @@ func runAsCmds(powershell string) []string {
...
 		`  $procInfo.Arguments = "-ExecutionPolicy RemoteSigned & '" + $script:MyInvocation.MyCommand.Path + "'"`,
 		`  $procInfo.Verb = "runas";`,
-		`  [System.Diagnostics.Process]::Start($procInfo);`,
+		`  $p = New-Object System.Diagnostics.Process`,
+		`  $p.StartInfo = $procInfo`,
+		`  $p.Start() | Out-Null`,
+		`  $p.WaitForExit()`,
+		`  if ($p.ExitCode -ne 0) {`,
+		`    throw "Unexpected failure";`,
+		`  }`,
```
anjannath, line 104: 👍🏽

anjannath:
shouldn't "add preflights..." commit come before the "use admin helper.." commit? 😛

pkg/crc/preflight/preflight_windows.go
```diff
@@ -136,6 +140,64 @@ var vsockChecks = []Check{
...
+			}
+			expected := filepath.Base(constants.DefaultAdminHelperCliBase)
+			if version != expected {
+				return fmt.Errorf("unexpected admin-helper daemon version (expected: %s, got: %s)", expected, version)
+			}
+			return nil
+		},
+		fixDescription: "Installing admin-helper daemon",
+		fix: func() error {
+			_, _, err := powershell.ExecuteAsAdmin("install admin-helper daemon", fmt.Sprintf("& '%s' uninstall-daemon; & '%s' install-daemon", adminhelper.BinPath, adminhelper.BinPath))
```
anjannath, line 188: 👍🏽

guillaumerose:
I added a commit to merge somehow the action of adding the current user in Hyper-V group and crc-users group.
It avoids the user to reboot 2 times.
It seems to work well for me. Let me know for you!

pkg/crc/preflight/preflight_windows.go
```diff
@@ -140,38 +174,37 @@ var vsockChecks = []Check{
...
-		},
-		fixDescription: "Creating crc-users group",
-		fix: func() error {
-			_, _, err := powershell.ExecuteAsAdmin("create crc-users group", "New-LocalGroup -Name crc-users")
-			return err
-		},
-		labels: labels{Os: Windows},
+var errReboot = errors.New("Please reboot your system and run 'crc setup' to complete the setup process")
+
+var adminHelperGroup = Check{
```
cfergeau, line 179: Does this bring any benefits to keep using a `Check` for this? I don't think `configKeySuffix` will still be used?
Looks like you could call `checkIfUserPartOfHyperVAdmins` directly, and have similar helpers for the `crc-users` group?

Concerns I have are around the `skip-` config options. I think after this `check-user-in-hyperv-group` will no longer be a valid config option? And doing 2 very different things in a single preflight check is problematic from the perspective of  these skip options, if only one of these groups is undesirable, as a user I can no longer skip the corresponding test.

cfergeau: > This line is also weird: https://github.com/code-ready/crc/blob/master/pkg/crc/preflight/preflight_checks_windows.go#L99

It's not really possible to abort preflight checks during setup, they will all run even if their `fix` method returns an error, so this `abort()` was added to stop as soon as we now a reboot will be needed.
One PR I've had around for far too long is one adding some kind of state to all preflights ( https://github.com/cfergeau/crc/commit/04ec76c308486cb4176e8cf199520380dd37d3b1 adds an 'options' parameter, but this could be a receiver instead). If we had some preflight state, we could do `SetRebootNeeded(...)` in the relevant preflight, and then have one test which aborts/fails when `RebootNeeded` is set. But we are not there yet :-/

guillaumerose: No real good solution. Can we keep this as it is for the current release ?

cfergeau: This is breaking backwards-compatibility (removal of a config option), and this ties 2 distinct things in a single preflight, which makes it harder to skip just the hyperv group if this was problematic before. So this is not as simple as "this is an improvement over the previous situation, we can refine it later", as it is now, it is both an improvement and a regression.

guillaumerose: OK so to get that in the release, I won't have time to get something proper running. Let's remove the last commit.
The user will have to reboot twice his computer for installing crc.

anjannath:
Tested, works as expected, initial and new routes are added in the hosts file👍🏽

pkg/crc/preflight/preflight_windows.go
cfergeau: Maybe for now the new user can be added as part of `check-user-in-hyperv-group` so that we at least keep the config key naem, and we improve on this next release?

pkg/crc/preflight/preflight_windows.go
```diff
@@ -141,6 +145,67 @@ var vsockChecks = []Check{
...
+		},
+		fixDescription: "Adding current user to crc-users group",
+		fix: func() error {
+			_, _, err := powershell.ExecuteAsAdmin("adding current user to crc-users group", fmt.Sprintf("Add-LocalGroupMember -Group crc-users -Member '%s'", os.Getenv("USERNAME")))
+			if err != nil {
+				return err
+			}
+			return errors.New("Please reboot your system and run 'crc setup' to complete the setup process")
+		},
+		labels: labels{Os: Windows},
```
anjannath, line 178: these two checks should also have the `NetworkMode` label?

guillaumerose: The admin helper is used in both networking mode. So there is no need for that.

pkg/crc/preflight/preflight_windows.go
guillaumerose: So this is equivalent to https://github.com/code-ready/crc/commit/3dd44303afaa85fb183200a3610da7dcf4635608 + change the config key  "check-user-in-groups" to "check-user-in-hyperv-group".
Should I push this?

cfergeau: yeah, this might be the 'less bad' option, but maybe you feel strongly otherwise?

guillaumerose: Ok done. I dont really have strong feelings about this. I just want to get the release done in time :)

cfergeau: I think this can be squashed in the previous commit?

guillaumerose: The commit itself explains why this is ugly. It also avoids a too large commit.

pkg/crc/preflight/preflight_windows.go
anjannath: i found its only used in `user` network-mode, in poststart service we have a check and for `system` network-mode we set the dns address in  vswitch only for user networking mode we use the admin helper.. or do we use it somewhere else aswell?

guillaumerose: Oh yes good catch! I was sure it was used for both cases, sorry...
I think it's something we forgot to fix. When goodhosts was first introduced, Windows implementation was supposed to be done "later".

guillaumerose: For instance, https://github.com/code-ready/crc/blob/f113feb287006253db46c647c17ed6ecc89d1d64/pkg/crc/preflight/preflight_checks_common.go#L84 has no label so runs with system network mode.
It uses the admin helper.

anjannath: ahh, yes, thanks!

anjannath (approved):

pkg/crc/preflight/preflight_windows.go
anjannath: Also later we'll know why these two checks were combined into one, (to avoid rebooting twice)

PR #2010, opened 2021-02-18
Mask username from error before sending to segment

praveenkumar:


pkg/crc/segment/segment.go
```diff
@@ -87,7 +87,7 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 		Set("duration", duration.Milliseconds()).
 		Set("tty", crcos.RunningInTerminal())
 	if err != nil {
-		properties = properties.Set("error", err.Error()).
+		properties = properties.Set("error", telemetry.SetError(err)).
```
cfergeau, line 90: I think I'd remove 'err' from `Upload` arguments, and call `telemetry.SetError()` in whatever is calling `Upload`. I did not look too closely at it, so maybe it's not so great, just opening this for discussion.

pkg/crc/telemetry/telemetry.go
```diff
@@ -84,3 +86,12 @@ const (
...
 	StartStartType          StartType = "start"
 )
+
+func SetError(err error) string {
+	// Mask username if present in the error string
+	user, err1 := user.Current()
+	if err1 != nil {
+		return err1.Error()
+	}
+	return strings.ReplaceAll(err.Error(), user.Username, "XXXXX")
```
cfergeau, line 96: `XXXX` would be more readable as `$USER` or something like this. I'd go with a (precompiled) regexp to be able to match `'\b' + user.Username + '\b'\` (only match full words, to avoid some false positives)

cfergeau (approved):

pkg/crc/telemetry/telemetry.go
jsliacan: I'd suggest something confusing like `fluffy-cat` or `blind-mole`. Would be the best thing to stumble on in Woopra :)

praveenkumar: @cfergeau regex is bit expensive than strings replace, when you say `false positive` you mean if the same username is part of the any error substring and if that occur we can always improve.

cfergeau: regex matching, especially precompiled is not going to be more expensive than the network traffic which is triggered shortly after this masking. We are already going to get some false positives when generic usernames are used (`ubuntu`). If we only match full words, this will at least avoid false positives on short usernames (made up one, but if someone picks `unn`, it's better if we don't match it in `Running`).

PR #2011, opened 2021-02-18
Add global 5s timeout to oc command

guillaumerose:
crc status can take a long time (> 20s) to  answer. If the apiserver is taking too long, then, we can say the cluster is unreachable.

For crc start, it will help retry loop: they will be more reactive and will not wait 20s before trying again. 
The issue I can see is in specific calls like "patch clusterversion", "apply specific config". If it takes too long, it can break the start process. Should the timeout limited to get requests? Should I use a 30s timeout for other requests?

### DIFF
--- a/pkg/crc/oc/oc.go
+++ b/pkg/crc/oc/oc.go
@@ -8,6 +8,8 @@ import (
 	crcos "github.com/code-ready/crc/pkg/os"
 )
 
+const timeout = "5s"
+
 type Config struct {
 	Runner           crcos.CommandRunner
 	OcExecutablePath string
@@ -39,10 +41,10 @@ func (oc Config) runCommand(isPrivate bool, args ...string) (string, string, err
 	}
 
 	if isPrivate {
-		return oc.Runner.RunPrivate(oc.OcExecutablePath, args...)
+		return oc.Runner.RunPrivate("timeout", append([]string{timeout, oc.OcExecutablePath}, args...)...)
 	}
 
-	return oc.Runner.Run(oc.OcExecutablePath, args...)
+	return oc.Runner.Run("timeout", append([]string{timeout, oc.OcExecutablePath}, args...)...)
 }
 
 func (oc Config) RunOcCommand(args ...string) (string, string, error) {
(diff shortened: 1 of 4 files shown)


pkg/crc/oc/oc.go
```diff
@@ -39,10 +41,10 @@ func (oc Config) runCommand(isPrivate bool, args ...string) (string, string, err
 	}
 
 	if isPrivate {
-		return oc.Runner.RunPrivate(oc.OcExecutablePath, args...)
+		return oc.Runner.RunPrivate("timeout", append([]string{timeout, oc.OcExecutablePath}, args...)...)
```
praveenkumar, line 44: Instead of doing this better to use the oc command timeout option.
```
$ oc options
[...]
--request-timeout='0': The length of time to wait before giving up on a single server request. Non-zero values
should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests.
```

guillaumerose: Yes I tried but it's not global timeout and even if I put 5s, sometimes oc takes 20s.

praveenkumar: which version of oc (checked with oc-4.6.x and 4.7)  you tried it with? It does works for me

```
$ oc version --client
Client Version: 4.7.0-rc.2 

$ time oc get nodes
Unable to connect to the server: dial tcp 192.168.130.11:6443: connect: no route to host

real	0m3.198s
user	0m0.166s
sys	0m0.044s

$ time oc get nodes --request-timeout='1s'
Unable to connect to the server: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)

real	0m1.105s
user	0m0.130s
sys	0m0.038s
```

guillaumerose: I am using oc 4.6.16. 

```
INFO ssh 12.971452547s sudo systemctl start kubelet 
INFO ssh 229.354148ms oc get nodes --request-timeout 5s --context admin --cluster crc --kubeconfig /opt/kubeconfig 
INFO ssh 226.358692ms oc get nodes --request-timeout 5s --context admin --cluster crc --kubeconfig /opt/kubeconfig 
INFO ssh 20.551353534s oc get nodes --request-timeout 5s --context admin --cluster crc --kubeconfig /opt/kubeconfig 
INFO ssh 2.580351017s oc get nodes --request-timeout 5s --context admin --cluster crc --kubeconfig /opt/kubeconfig 
INFO ssh 264.985137ms oc get secret --request-timeout 5s --context admin --cluster crc --kubeconfig /opt/kubeconfig 
```

praveenkumar: Alright,I also tried it and looks like when it trying the initial request, does fail for timeouts.

PR #2018, opened 2021-02-22
add details to telemetry about libvirt network failure

guillaumerose:
* Collect why crc network creation failed in telemetry

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -349,7 +349,7 @@ func fixLibvirtCrcNetworkAvailable() error {
 	err = cmd.Run()
 	if err != nil {
 		logging.Debugf("%v : %s", err, buf.String())
-		return fmt.Errorf("Failed to create libvirt 'crc' network")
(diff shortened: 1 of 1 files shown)


cmd/crc/cmd/podman_env.go
```diff
@@ -45,6 +45,5 @@ func RunPodmanEnv(args []string) error {
 }
 
 func init() {
-	rootCmd.AddCommand(podmanEnvCmd)
```
gbraad, line 48: ?

gbraad (changes requested):

cmd/crc/cmd/podman_env.go
guillaumerose: According to telemetry, lot of people try this command. This is disappointing. We should remove it if it doesn't work.

gbraad: or fix it ?

guillaumerose: I don't think we should fix it.
* It will be hard to fix it before next release
* The version of podman included with the bundle is too old to be interesting for the end user.

guillaumerose: I dropped this commit.

gbraad (approved):

Issue #1937, opened 2021-02-03
make macOS tray enabled by default on setup


anjannath:
we have the basic functionality for the tray on macOS, so we can have it enabled by default on setup, user can always exit it from the tray menu.

To uninstall user needs to run `crc cleanup`
Maybe also a config option to skip the installation of the tray during `crc setup` like `crc setup install-tray false` ?

cfergeau:
> user can always exit it from the tray menu.

But it will be restarted automatically with next reboot. I'm not clear on what the exact workflows would be, especially to reenable the tray, but in my opinion it should be possible to permanently disable the tray (with a click or so, not through several `crc` commands).

anjannath:
> but in my opinion we need a way to permanently disable the tray (with a click or so, not through several crc commands).

+1,  i was thinking of adding a button for only removing the tray related files in the `Advanced` tab of preferences window

gbraad:
it should not only be from the tray...

the action would be to allow a toggle to `start-tray` which will default to true, but will be allowed to modify.
the startup of the tray can check this and exit if the files AREN'T removed. but the cleanup could in that case also handle the removal. best to decide the flow before working on it.

anjannath:
1. User runs `crc setup` and has the tray installed, To uninstall -
- He can run `crc cleanup` but this'll also remove the crc instance.
- He can close the try from its menu, but it'll show up on next reboot.

We already have the uninstall feature in the `crc cleanup` command, what we want to control is the auto-start behaviour, for that we can introduce a new config option in crc called `auto-start-tray` (_Launch tray at login_ on the tray UI) and this would control the auto-start behavior of the trays, this is for enabling/disabling only the trays and not control the daemon.

cfergeau:
> For re-enabling the tray the user can either use the cli to do config set auto-start-tray true

followed by `crc setup` I assume? The rest of your proposal sounds good to me, and simple enough on the crc side!

anjannath:
> followed by crc setup I assume?

I was thinking to run a callback on `config set/unset` for this, It'll just be creation/removal of a  file in `~/Library/LaunchDaemons`

guillaumerose:
Do we have approval of @sspeiche for that ? 
I still can find lot of tiny bugs when I try the tray.

sspeiche:
> Do we have approval of @sspeiche for that ?
> I still can find lot of tiny bugs when I try the tray.

[...]

PR #2019, opened 2021-02-23
Install macOS tray by default during crc setup

anjannath:
Fixes: #1937 

## Proposed changes
1. Add `tray-auto-start` config key.

### First Installation
1. `crc setup` should install the tray.
2. Reboot -> tray should start on its own.

### Disable auto start of the tray
1. `crc config set auto-start-tray false`
2. reboot -> tray should not start.

### Re-Install
1. `crc config set auto-start-tray true`
2. `crc setup` -> tray should be started
3.  reboot -> tray should be started on its own

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -10,6 +10,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/network"
+	"github.com/code-ready/crc/pkg/crc/tray"
 	"github.com/spf13/cobra"
 )
 
@@ -30,6 +31,7 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
+	AutostartTray           = "autostart-tray"
 )
 
 func RegisterSettings(cfg *config.Config) {
@@ -63,6 +65,9 @@ func RegisterSettings(cfg *config.Config) {
 		"Network mode (default or vsock)")
 	cfg.AddSetting(HostNetworkAccess, false, validateHostNetworkAccess, config.SuccessfullyApplied,
 		"Allow TCP/IP connections from the CodeReday Containers VM to services running on the host (true/false, default: false)")
+	// System tray auto-start config
+	cfg.AddSetting(AutostartTray, true, tray.ValidateTrayAutostart, tray.DisableEnableTrayAutostart,
+		"Automatically start the tray (true/false, default: true)")
 	// Proxy Configuration
 	cfg.AddSetting(HTTPProxy, "", config.ValidateURI, config.SuccessfullyApplied,
 		"HTTP proxy URL (string, like 'http://my-proxy.com:8443')")
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/tray"
 	"github.com/spf13/cobra"
@@ -44,6 +45,23 @@ func RegisterSettings(cfg *config.Config) {
 		return config.ValidateBool(value)
 	}
 
+	disableEnableTrayAutostart := func(key string, value interface{}) string {
+		mode := network.ParseMode(cfg.Get(NetworkMode).AsString())
+		var mErr = errors.MultiError{}
+		if err := tray.DisableTrayAutostart(); err != nil {
+			mErr.Collect(err)
+		}
+		if mode == network.DefaultMode {
+			if err := tray.DisableDaemonAutostart(); err != nil {
+				mErr.Collect(err)
+			}
+		}
+		if len(mErr.Errors) > 0 {
+			return fmt.Sprintf("Error occurred on disabling tray autostart: %s", mErr.Error())
+		}
+		return "Successfully disabled autostart of tray at login."
+	}
+
 	// Start command settings in config
 	cfg.AddSetting(Bundle, constants.DefaultBundlePath, config.ValidateBundlePath, config.SuccessfullyApplied,
 		fmt.Sprintf("Bundle path (string, default '%s')", constants.DefaultBundlePath))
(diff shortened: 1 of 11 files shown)


pkg/crc/preflight/preflight_darwin.go
```diff
@@ -107,24 +107,24 @@ var traySetupChecks = [...]Check{
 }
 
 func getAllPreflightChecks() []Check {
-	return getPreflightChecks(true, network.DefaultMode)
+	return getPreflightChecks(true, true, network.DefaultMode)
 }
 
-func getPreflightChecks(experimentalFeatures bool, mode network.Mode) []Check {
+func getPreflightChecks(experimentalFeatures bool, trayAutostart bool, mode network.Mode) []Check {
```
praveenkumar, line 113: Here `experimentalFeatures` is no longer used so better to `_ bool` ?

cmd/crc/cmd/config/config.go
```diff
@@ -51,6 +52,9 @@ func RegisterSettings(cfg *config.Config) {
 		"Enable experimental features (true/false, default: false)")
 	cfg.AddSetting(NetworkMode, string(network.DefaultMode), network.ValidateMode, network.SuccessfullyAppliedMode,
 		"Network mode (default or vsock)")
+	// System tray auto-start config
+	cfg.AddSetting(AutostartTray, true, config.ValidateTrayAutostart, config.DisableEnableTrayAutostart,
+		"Disable tray auto-start (true/false, default: true)")
```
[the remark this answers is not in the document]
cfergeau: Or "Automatically start the tray (when the system starts)"?

pkg/crc/config/callbacks.go
```diff
@@ -15,3 +15,8 @@ func RequiresRestartMsg(key string, _ interface{}) string {
 func SuccessfullyApplied(key string, value interface{}) string {
 	return fmt.Sprintf("Successfully configured %s to %s", key, cast.ToString(value))
 }
+
+func RequiresCRCSetup(key string, _ interface{}) string {
+	return fmt.Sprintf("Changes to configuration property '%s' are only applied during 'crc setup'.\n"+
```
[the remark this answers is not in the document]
cfergeau: This is about "Changes" which is plural, so I'd say `Changes to ... are only applied..` (ie no changes).

pkg/crc/config/callbacks_darwin.go
```diff
@@ -0,0 +1,24 @@
...
+	// Enable
+	if cast.ToBool(value) {
+		return RequiresCRCSetup(key, value)
+	}
+	// Disable
+	for _, agentLabel := range agentLabels {
+		if err := launchd.RemovePlist(agentLabel); err != nil {
+			return fmt.Sprintf("Error trying to disable auto-start of tray: %s", err.Error())
+		}
+	}
```
cfergeau, line 22: Wondering if we should also leave this removal up to `crc setup` even if that's a bit odd?
Even if we decide to trigger the file removal from here, the code itself should be moved to a different package, `config` has no business knowing the tray uses 2 labels `crc.tray` `crc.daemon` which are managed using `launchd`.

pkg/crc/config/validations.go
```diff
@@ -103,3 +104,11 @@ func ValidateYesNo(value interface{}) (bool, string) {
 	}
 	return false, "must be yes or no"
 }
+
+// ValidateTrayAutostart checks tray-auto-start is used in macOS and its a bool
+func ValidateTrayAutostart(value interface{}) (bool, string) {
+	if runtime.GOOS == "darwin" {
+		return ValidateBool(value)
+	}
+	return false, "only supported in macOS"
```
cfergeau, line 113: Since you added per-os files, you could also have 2 different definitions for `ValidateTrayAutostart` in these files rather than checking the OS at runtime.

pkg/crc/config/callbacks_darwin.go
anjannath: Moved this to a separate tray package `pkg/crc/tray`

pkg/crc/preflight/preflight_darwin.go
anjannath: Done

pkg/crc/config/validations.go
anjannath: Moved this validation function also to the tray package

cmd/crc/cmd/config/config.go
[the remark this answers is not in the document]
anjannath: Changed to:
```
* auto-start-tray                      Automatically start the tray (true/false, default: true)
```

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -107,24 +107,24 @@ var traySetupChecks = [...]Check{
...
+func getPreflightChecks(_ bool, trayAutostart bool, mode network.Mode) []Check {
 	checks := []Check{}
 
 	checks = append(checks, nonWinPreflightChecks[:]...)
 	checks = append(checks, genericPreflightChecks[:]...)
 	checks = append(checks, hyperkitPreflightChecks(mode)...)
 	checks = append(checks, dnsPreflightChecks[:]...)
 
+	if trayAutostart {
+		checks = append(checks, traySetupChecks[:]...)
```
guillaumerose, line 122: Can you separate the daemon from the tray so that the daemon is always installed now ? We can do it separately (PR or commit) if you prefer also.

anjannath: Sure, will add another commit for it.

guillaumerose:
We also need https://github.com/code-ready/crc/pull/1993 for the release otherwise, even if not running in vsock mode, the daemon will bind port 2222, 443 and 6443. What should we do? We can also turn off the vsock part of the daemon if network mode != vsock.

cmd/crc/cmd/config/config.go
```diff
@@ -29,6 +30,7 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
+	AutostartTray           = "auto-start-tray"
```
gbraad, line 33: `autostart-tray` ?

gbraad (approved):
only concerned about the naming of the config option

gbraad:
This will affect documentation... @robin-owen

pkg/crc/preflight/preflight_darwin.go
cfergeau: If "daemon is always installed" means "daemon is always running" regardless of the tray being used or not, this is only needed for vsock, not with the default networking.

cmd/crc/cmd/config/config.go
anjannath: Changed to `autostart-tray`

pkg/crc/preflight/preflight_darwin.go
anjannath: Separated the daemon checks, and these are ran if network mode is vsock or if tray is set to autostart

anjannath:
> We also need #1993 for the release otherwise, even if not running in vsock mode, the daemon will bind port 2222, 443 and 6443. What should we do? We can also turn off the vsock part of the daemon if network mode != vsock.

I think the latter would be better, `network-mode` should control if the VSock part

cmd/crc/cmd/config/config.go
```diff
@@ -30,18 +32,35 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
+	AutostartTray           = "autostart-tray"
 )
 
 func RegisterSettings(cfg *config.Config) {
+	mode := network.ParseMode(cfg.Get(NetworkMode).AsString())
```
anjannath, line 39: Yikes...
```
$ crc                            
ERRO[0000] unexpected network mode , using default  
No command given
```

can't parse the mode before its Registered.

PR #2022, opened 2021-02-23
RFC: put the bundle name logic in one place

guillaumerose:
When I modified crc to use a podman_hyperkit_3.0.0 bundle, I had to change the name both in Makefile and in the Go code. This was not obvious. I try in this PR to change this.

For the e2e test, I am not sure what I should do. Should I try to remove --bundle-version?
For the OpenShiftVersion in crc version, I guess I should read it from the bundle embedded and not get it from the variable. WDYT?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,9 +231,9 @@ endif
 	@$(call check_defined, BUNDLE_DIR, "Embedding bundle requires BUNDLE_DIR set to a directory containing CRC bundles for all hypervisors")
 
 embed_bundle: clean cross $(HOST_BUILD_DIR)/crc-embedder check_bundledir $(HYPERKIT_BUNDLENAME) $(HYPERV_BUNDLENAME) $(LIBVIRT_BUNDLENAME)
-	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=darwin --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/macos-amd64/crc
-	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=linux --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/linux-amd64/crc
-	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=windows --bundle-dir=$(BUNDLE_DIR) $(BUILD_DIR)/windows-amd64/crc.exe
+	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=darwin --bundle=$(HYPERKIT_BUNDLENAME) $(BUILD_DIR)/macos-amd64/crc
+	$(HOST_BUILD_DIR)/crc-embedder embed --log-level debug --goos=linux --bundle=$(LIBVIRT_BUNDLENAME) $(BUILD_DIR)/linux-amd64/crc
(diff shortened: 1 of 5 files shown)


praveenkumar:
> When I modified crc to use a podman_hyperkit_3.0.0 bundle,

Why do we need to have different bundle for podman, am I missing something? IIRC our plan is to use same bundle for podman and openshift side.

Makefile
```diff
@@ -35,6 +35,14 @@ RELEASE_INFO := release-info.json
 
 MOCK_BUNDLE ?= false
 
+HYPERKIT_BUNDLENAME = crc_hyperkit_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
+HYPERV_BUNDLENAME = crc_hyperv_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
+LIBVIRT_BUNDLENAME = crc_libvirt_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
```
cfergeau, line 40: Fwiw, there are a few occurrences of `crc_libvirt_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)` in the Makefile which could be replaced with `LIBVIRT_BUNDLENAME`

test/e2e/crcsuite/crcsuite.go
```diff
@@ -161,6 +159,18 @@ func FeatureContext(s *godog.Suite) {
...
 
+func defaultBundleForOs(bundleVersion string) map[string]string {
+	return map[string]string{
+		"darwin":  fmt.Sprintf("crc_hyperkit_%s.crcbundle", bundleVersion),
+		"linux":   fmt.Sprintf("crc_libvirt_%s.crcbundle", bundleVersion),
+		"windows": fmt.Sprintf("crc_hyperv_%s.crcbundle", bundleVersion),
+	}
+}
+
+func getBundleFosOs(os, bundleVersion string) string {
```
cfergeau, line 170: Can you fix the `FosOs` Type while you are moving this? The intermediate helper is no longer needed as it only has one user.

Makefile
```diff
@@ -90,16 +96,16 @@ check: cross build_e2e $(HOST_BUILD_DIR)/crc-embedder test cross-lint vendorchec
 
 .PHONY: install
 install: $(SOURCES)
-	go install -ldflags="$(LDFLAGS)" ./cmd/crc
+	go install -ldflags="$($(GOOS_LDFLAGS))" ./cmd/crc
```
cfergeau, line 99: this is not working as expected, it's looking for a variable called `linux_LDFLAGS`, not `LINUX_LDFLAGS`

pkg/crc/version/version.go
```diff
@@ -21,6 +21,9 @@ var (
 	// Bundle version which used for the release.
 	bundleVersion = "0.0.0-unset"
 
+	// Bundle which used for the release.
+	bundle = ""
```
cfergeau, line 25: `bundleName` or `bundlePath` to avoid an ambiguous name?

cfergeau (commented):
> For the OpenShiftVersion in crc version, I guess I should read it from the bundle embedded and not get it from the variable. WDYT?

If the embedded bundle version does not match the version from the variable, something is wrong, I would focus on being 100% sure this is the case rather than changing where we read the version from. The advantage of getting it from the variable is that it works even if the bundle hasn't been unpacked.

guillaumerose:
Thanks for the feedback! I thought it would be useful but it doesn't worth it (esp. if podman bundle doesn't exist).

Maybe we could guess the bundle name by searching for the first .crcbundle files in embed files. For dev build, we could list bundles in ~/.crc/cache and pick the first one.

Issue #1986, opened 2021-02-11
Add `host-network-access` configuration option


cfergeau:
https://github.com/code-ready/crc/pull/1977 added unrestricted access from the VM to the host when vsock is used. This should be controlled by a configuration option which defaults to 'false' as this could be very unexpected from a security perspective.

PR #2025, opened 2021-02-24
Don't enable host network access by default

cfergeau:
Since 'vsock: allow the VM to reach host services', when vsock is in
use, any containers running in the VM can connect to services listening
on a TCP port on the host. This is similar to podman --network=host.
Since this could be unexpected, and cause security issues in some cases,
it's preferrable not to enable this by default. This commit adds
a `host-network-access` boolean to control this which is false by
default.

This fixes https://github.com/code-ready/crc/issues/1986

**Fixes:** Issue #1986

## Solution/Idea

There is now a "host-network-access" configuration option to enable-disable this behaviour of the vsock stack.

## Proposed changes

1. Try `crc config set host-network-access true/false` with the default network mode -> it should fail
2. Try `crc config set host-network-access true/false` after `crc config set network-mode vsock` -> it should not fail
2. Run `crc daemon --log-level debug` after setting `host-network-access` to true -> the log should show:
```
DEBU Running 'crc daemon'
DEBU Enabling host network access
```
3. When `host-network-access` is true, the steps from https://github.com/code-ready/crc/commit/149aedafd9c5ae0625853700abffb7d128bd9b54 should still work
4. When `host-network-access` is false, these steps should fail

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -23,6 +23,7 @@ const (
 	DisableUpdateCheck      = "disable-update-check"
 	ExperimentalFeatures    = "enable-experimental-features"
 	NetworkMode             = "network-mode"
+	HostNetworkAccess       = "host-network-access"
 	HTTPProxy               = "http-proxy"
 	HTTPSProxy              = "https-proxy"
 	NoProxy                 = "no-proxy"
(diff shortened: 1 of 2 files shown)


anjannath (approved):
Tested on macOS, works as expected

Issue #2028, opened 2021-03-01
enable-cluster-monitoring flag: remove also ClusterOperator from CVO overrides


guillaumerose:
In 4.6, in the cvo overrides, the deployment of the monitoring was disabled. crc only had to remove this object from the cvo overrides to turn the monitoring on. 

In 4.7 bundle, both the deployment and the cluster operator monitoring are disabled. We need to update crc code accordingly. 

The current code works, there is just a small difference (see https://github.com/code-ready/crc/pull/2001#issuecomment-784743822).

PR #2029, opened 2021-03-01
Remove ClusterOperator kind from CVO overrides

praveenkumar:
For 4.7 bundles we are adding the deployment kind and clusteroperator
kind for monitoring as part of overrides, better to remove both when
enabling the monitoring from CVO.

**Fixes:** #2028

### DIFF
--- a/pkg/crc/cluster/monitoring.go
+++ b/pkg/crc/cluster/monitoring.go
@@ -20,20 +20,28 @@ func StartMonitoring(ocConfig oc.Config) error {
 		return err
 	}
 
+	indexForClusterMonitoringDeploymentKind := getIndexInOverridesForObjectName(cv, "cluster-monitoring-operator")
+	indexForClusterMonitoringCVOKind := getIndexInOverridesForObjectName(cv, "monitoring")
+
+	if indexForClusterMonitoringDeploymentKind != -1 && indexForClusterMonitoringCVOKind != -1 {
+		_, _, err = ocConfig.RunOcCommand("patch", "clusterversion/version",
+			"--type", "json",
+			"--patch", fmt.Sprintf(`'[{"op":"remove", "path":"/spec/overrides/%d"},{"op":"remove", "path":"/spec/overrides/%d"}]'`,
+				indexForClusterMonitoringDeploymentKind, indexForClusterMonitoringCVOKind-1))
+	}
+	return err
+}
+
+func getIndexInOverridesForObjectName(cv v1.ClusterVersion, objectName string) int {
 	pos := -1
 	for i, override := range cv.Spec.Overrides {
-		if override.Name == "cluster-monitoring-operator" {
+		if override.Name == objectName {
 			pos = i
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/crc/cluster/monitoring.go
```diff
@@ -20,20 +20,28 @@ func StartMonitoring(ocConfig oc.Config) error {
...
 	}
 
+	indexForClusterMonitoringDeploymentKind := getIndexInOverridesForObjectName(cv, "cluster-monitoring-operator")
+	indexForClusterMonitoringCVOKind := getIndexInOverridesForObjectName(cv, "monitoring")
+
+	if indexForClusterMonitoringDeploymentKind != -1 && indexForClusterMonitoringCVOKind != -1 {
+		_, _, err = ocConfig.RunOcCommand("patch", "clusterversion/version",
+			"--type", "json",
+			"--patch", fmt.Sprintf(`'[{"op":"remove", "path":"/spec/overrides/%d"},{"op":"remove", "path":"/spec/overrides/%d"}]'`,
+				indexForClusterMonitoringDeploymentKind, indexForClusterMonitoringCVOKind-1))
```
anjannath, line 30: why `indexForClusterMonitoringCVOKind-1` doesn't `getIndexInOverridesForObjectName()` return the correct index?

pkg/crc/cluster/monitoring.go
```diff
@@ -20,20 +20,28 @@ func StartMonitoring(ocConfig oc.Config) error {
...
+	}
+	return err
```
guillaumerose, line 32: Maybe ?

```
if indexForClusterMonitoringDeploymentKind != -1 && indexForClusterMonitoringCVOKind != -1 { 
  _, _, err = ocConfig.RunOcCommand()
  return err
}
return errors.New("Unable to find monitoring objects in CVO overrides")
```

pkg/crc/cluster/monitoring.go
praveenkumar: It does return correct index but as soon as we run json patch the index change since the first patch from the list removed so it is shifted.

pkg/crc/cluster/monitoring.go
praveenkumar: @guillaumerose if we want it blocking in case we are not able to find the kind index then sure but in this PR I didn't change the existing behavior of this function which doesn't take any action in case of not finding the object index.

guillaumerose: Oh I see, we need that in case we start-stop-start again the cluster. 👍

pkg/crc/cluster/monitoring.go
cfergeau: This is assuming `indexForClusterMonitoringDeploymentKind` is smaller than `indexForClusterMonitoringCVOKind`, if this is not true, you'll be removing a random override, if `indexForClusterMonitoringCVOKind` is 0, this will fail too.
Maybe remove each override independently, ie run `oc` twice?

Issue #2039, opened 2021-03-01
Ellipsis consistency: choose between `word...` or `word ...`


guillaumerose:
### Steps to reproduce

  1. crc setup
  2. crc start

### Expected

```
INFO Adding user's pull secret to instance disk... 
INFO Verifying validity of the kubelet certificates... 
```

### Actual

```
INFO Adding user's pull secret to instance disk... 
INFO Verifying validity of the kubelet certificates ... 
```

### Description

Three dots are often used in crc: sometimes with a space before, sometimes directly next to the word. We should make a choice! :)

ping @robin-owen

robin-owen:
From a "grammatically correct" standpoint, we should not have a space before the ellipses.

cfergeau:
> From a "grammatically correct" standpoint, we should not have a space before the ellipses.

wikipedia lists various point of views on this https://en.wikipedia.org/wiki/Ellipsis#In_English :) But no spaces look nicer to me!

PR #2043, opened 2021-03-01
After waiting cluster operators, directly write kubeconfig without waiting for authentication

guillaumerose:
Since d6cc7a1, all operators should be
either up or really broken before this function call.
If it's up, then the kubeconfig will be created directly.
If it's down, we don't wait 1 extra minute, we directly skip this.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -398,7 +398,7 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 
 	waitForProxyPropagation(ocConfig, proxyConfig)
 
-	logging.Info("Updating kubeconfig")
+	logging.Info("Adding crc-admin and crc-developer contexts to kubeconfig...")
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/start.go
```diff
@@ -398,8 +398,8 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 
 	waitForProxyPropagation(ocConfig, proxyConfig)
 
-	logging.Info("Updating kubeconfig")
-	if err := eventuallyWriteKubeconfig(ocConfig, instanceIP, clusterConfig); err != nil {
+	logging.Info("Adding crc-admin and crc-developer contexts to kubeconfig ...")
```
praveenkumar, line 401: As per @robin-owen comment on the issue https://github.com/code-ready/crc/issues/2039#issuecomment-788181671 , we should not have any space in between.

guillaumerose: thanks done!

PR #2046, opened 2021-03-02
Load the kubeadmin password in a more robust way: remove extra spaces

guillaumerose:
The password is in a file that can contain extra newline at the end.
This newline breaks `oc login` when crc does it with the oc library.
Also, `crc start` output shows this extra line.


cfergeau (approved):
I thought this was generated by snc. in which case I would have suggested to also fix it there, but it's actually generated by openshift-install, so we have less control over that file.

praveenkumar:
With the current PR https://github.com/code-ready/snc/pull/348 we can control that new line.

cfergeau:
If it is only an issue when using https://github.com/code-ready/snc/pull/348, maybe better to just fix it in snc.

guillaumerose:
It doesnt hurt to sanitize this input. Esp. if we offer the user a command to repack bundles.

PR #2049, opened 2021-03-03
Revert "Remove tray and daemon preflight checks from experimental"

cfergeau:
Testing revealed last minute issues with the macOS tray, which we can't
fix in time for the 1.23 release. Let's move it back behind experimental
in the release branch while we fix these remaining issues.

### DIFF
--- a/pkg/crc/preflight/preflight_darwin.go
+++ b/pkg/crc/preflight/preflight_darwin.go
@@ -113,7 +113,7 @@ func getAllPreflightChecks() []Check {
 	return getPreflightChecks(true, true, network.DefaultMode)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/crc/preflight/preflight_darwin.go
```diff
@@ -113,26 +113,28 @@ func getAllPreflightChecks() []Check {
...
-	if trayAutostart || mode == network.VSockMode {
+	switch mode {
+	case network.DefaultMode:
+		checks = append(checks, resolverPreflightChecks[:]...)
+	case network.VSockMode:
 		checks = append(checks, daemonSetupChecks[:]...)
 	}
 
-	if trayAutostart {
+	if experimentalFeatures && trayAutostart {
```
anjannath, line 131: 👍🏽

PR #2052, opened 2021-03-03
Convert Debug to Info for wait for apiserver availablility

praveenkumar:
Right now if user doesn't start crc with debug mode then apiserver
availablity check happen just after kubelet service and user can see
this it is hang.

```
INFO Adding user's pull secret to instance disk...
INFO Verifying validity of the kubelet certificates ...
INFO Starting OpenShift kubelet service
[.. here around 2 min time before next message appear ..]
```

This patch try to tell user that `starting kubelet` service is not
blocked but we are waiting for apiserver to be ready.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -310,7 +310,7 @@ func WaitForRequestHeaderClientCaFile(sshRunner *ssh.Runner) error {
 }
 
 func WaitForAPIServer(ocConfig oc.Config) error {
-	logging.Debugf("Waiting for apiserver availability")
+	logging.Info("Waiting for kube-apiserver availability... [takes around 2min]")
 	waitForAPIServer := func() error {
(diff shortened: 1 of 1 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -310,7 +310,7 @@ func WaitForRequestHeaderClientCaFile(sshRunner *ssh.Runner) error {
 }
 
 func WaitForAPIServer(ocConfig oc.Config) error {
-	logging.Debugf("Waiting for apiserver availability")
+	logging.Info("Waiting for kube-apiserver availability(takes around 2m)...")
```
guillaumerose, line 313: A missing space before the (. 
Also, here we use [] not (): https://github.com/code-ready/crc/blob/master/pkg/crc/machine/start.go#L393

guillaumerose (approved):

cfergeau:
Why is this taking 2 minutes? I thought the 4.7 issue was adding 45s to 1min to startup?

PR #2056, opened 2021-03-03
Note systemd-resolved usage for RHEL 8

omenos:
systemd-resolved can be used on RHEL 8.3 and compatible distributions as it is available as a Tech Preview. However, there are some configurations that need to be made for this system to actually work with present releases.

This is a documentation pull request, it does not alter functionality of CRC itself. I have personally been running with this setup on RHEL 8 for several days now with no obvious issues.

### DIFF
--- a/docs/source/topics/ref_dns-configuration.adoc
+++ b/docs/source/topics/ref_dns-configuration.adoc
@@ -39,8 +39,6 @@ exit 0
 
 [NOTE]
 ====
-:rhel-resolved-docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/using-different-dns-servers-for-different-domains_configuring-and-managing-networking
-
 `systemd-resolved` is also available as an unsupported Technology Preview on {rhel} and {centos} 8.3. After {rhel-resolved-docs}[configuring the host] to use `systemd-resolved`, any {prod} `dnsmasq` configurations need to be removed and the above script needs to be created with executable permissions.
 
 Configure the {prod} binary to skip startup checks for `dnsmasq` and NetworkManager configuration:
--- a/docs/source/topics/ref_dns-configuration.adoc
+++ b/docs/source/topics/ref_dns-configuration.adoc
@@ -29,12 +29,25 @@ This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions
 * `systemd-resolved` configuration is done through a NetworkManager dispatcher script in `/etc/NetworkManager/dispatcher.d/99-crc.sh`:
 +
 ----
+#!/bin/sh
 resolvectl domain crc ~testing
 resolvectl dns crc 192.168.130.11
 resolvectl default-route crc false
 
 exit 0
 ----
+
+[NOTE]
+====
+`systemd-resolved` is also available as a Tech Preview on {rhel} and {centos} 8.3. After configuring the host to use systemd-resolved, any CRC dnsmasq configurations need to be removed and the above script needs to be created with executable permissions.
+
+The crc virtual machine presently needs to be configured to skip certain checks during startup:
+
+----
+$ crc config set skip-check-crc-dnsmasq-file true
+$ crc config set skip-check-network-manager-config true
+----
+====
 
 === NetworkManager + dnsmasq
 
--- a/docs/source/topics/ref_dns-configuration.adoc
+++ b/docs/source/topics/ref_dns-configuration.adoc
@@ -39,13 +39,16 @@ exit 0
 
 [NOTE]
 ====
-`systemd-resolved` is also available as a Tech Preview on {rhel} and {centos} 8.3. After configuring the host to use systemd-resolved, any CRC dnsmasq configurations need to be removed and the above script needs to be created with executable permissions.
+:rhel-resolved-docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/using-different-dns-servers-for-different-domains_configuring-and-managing-networking
 
-The crc virtual machine presently needs to be configured to skip certain checks during startup:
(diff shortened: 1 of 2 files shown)


cfergeau:
Thanks for the patch!

> systemd-resolved can be used on RHEL 8.3 and compatible distributions as it is available as a Tech Preview. However, there are some configurations that need to be made for this system to actually work with present releases.
> 
> This is a documentation pull request, it does not alter functionality of CRC itself. I have personally been running with this setup on RHEL 8 for several days now with no obvious issues.

Can you clarify "can be used"? Does this mean the system administrator can manually configure their rhel system to use systemd-resolved? Or is this the default after doing an install?

omenos:
@cfergeau I've updated the notes to *hopefully* clarify things a bit more about the state of systemd-resolved. Would it be a good idea to link to the RHEL docs?

cfergeau:
> @cfergeau I've updated the notes to _hopefully_ clarify things a bit more about the state of systemd-resolved in RHEL 8. Would it be a good idea to link to the RHEL docs?
> 
> It is a Tech Preview, and does need to be configured by the user as it is not a default configuration.

Thanks for clarifying. A link to RHEL docs would not hurt, though I'll let @robin-owen weight on this.
Ideally `crc` would detect whether NM or system-resolved or ... are in use, rather than hardcoding what is used depending on the distro it detects. But this is not there yet, hence the need for this kind of documentation :)

docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -29,13 +29,26 @@ This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions
 * `systemd-resolved` configuration is done through a NetworkManager dispatcher script in `/etc/NetworkManager/dispatcher.d/99-crc.sh`:
 +
 ----
+#!/bin/sh
```
robin-owen, line 32: :+1:

docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -29,13 +29,26 @@ This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions
...
 resolvectl domain crc ~testing
 resolvectl dns crc 192.168.130.11
 resolvectl default-route crc false
 
 exit 0
 ----
 
+[NOTE]
+====
+`systemd-resolved` is also available as a Tech Preview on {rhel} and {centos} 8.3. After configuring the host to use systemd-resolved, any CRC dnsmasq configurations need to be removed and the above script needs to be created with executable permissions.
```
robin-owen, line 42: I'd like to see `systemd-resolved` (and `dnsmasq`) formatted each time. Also, we should use the `{prod}` attribute instead of the acronym "CRC".

docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -29,13 +29,26 @@ This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions
...
+
+The crc virtual machine presently needs to be configured to skip certain checks during startup:
```
robin-owen, line 44:
```adoc
Configure the {prod} binary to skip startup checks for `dnsmasq` and NetworkManager configuration:
```

docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -29,13 +29,26 @@ This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions
...
+
+----
```
robin-owen, line 46: This block should enable attribute substitutions and use the `{bin}` attribute for the binary name in the commands like so:

```adoc
[subs="+quotes,attributes"]
----
$ {bin} config set skip-check-crc-dnsmasq-file true
$ {bin} config set skip-check-network-manager-config true
----
```

omenos:
Yes, if I had the technical skill right now this pull request would have been for replacing the hard-coding with detection, but I'm not quite there yet with my gophering :wink:

robin-owen (changes requested):
Thank you for submitting this!

I left a few comments which I'd like to see addressed before approving this PR, but I truly appreciate you taking the time to write this. :+1:

robin-owen:
I just saw the comments about linking to the RHEL documentation. I don't see an issue with this one way or the other, honestly, so I'm perfectly happy whether we include that or not. :)

docs/source/topics/ref_dns-configuration.adoc
omenos: Updated formatting

docs/source/topics/ref_dns-configuration.adoc
omenos: Swapped statements

docs/source/topics/ref_dns-configuration.adoc
omenos: Added the substitutions and header

omenos:
Updated to include the text revisions and variable substitutions. I should have looked at the product-attributes file containing the macros, will keep that in mind for the future.

docs/source/topics/ref_dns-configuration.adoc
```diff
@@ -29,13 +29,29 @@ This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions
...
 resolvectl domain crc ~testing
 resolvectl dns crc 192.168.130.11
 resolvectl default-route crc false
 
 exit 0
 ----
 
+[NOTE]
+====
+:rhel-resolved-docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/using-different-dns-servers-for-different-domains_configuring-and-managing-networking
```
robin-owen, line 42: Would you mind placing this attribute in the attributes file instead of here? The file is in `docs/source/meta/product-attributes.adoc`. Thank you!

Everything else LGTM. :+1:

omenos: Attribute moved to `product-attributes.adoc`.

robin-owen (approved):
LGTM. :+1:

PR #2060, opened 2021-03-04
Improve logging of daemon protocol

cfergeau:
We currently log the input data, but not the answer which is sent.
This commit ensures both are logged, and also logs the address of the
socket as multiple requests can be handled concurrently, but they use
differet sockets, so this should help identify which answer corresponds
to which request.

### DIFF
--- a/pkg/crc/api/api.go
+++ b/pkg/crc/api/api.go
@@ -90,7 +90,7 @@ func (api Server) handleConnections(conn net.Conn) {
 		logging.Error("Error reading from socket")
 		return
 	}
-	logging.Debug("Received Request:", string(inBuffer[0:numBytes]))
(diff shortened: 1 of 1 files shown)


guillaumerose:
I don't think we should log more but rather just log the command name + status code (like http). The request body and answer body can contain the proxy password for example.

cfergeau:
It seems there is no easy way to sniff the tray<->daemon protocol though as this goes through a unix socket?
I can hide this behind an env var for example if we want to avoid customers giving us passwords by mistake when they upload logs?

guillaumerose:
Let's add this and raise the log level of the daemon? Currently it's trace level. With the status polling coming from the tray, it will quickly fill the log file!

PR #2061, opened 2021-03-04
Improve status reporting

guillaumerose:
When the VM is starting (esp before ssh key rotation), these changes will help report a proper status and not just an error.


cfergeau (approved):
In 
```
 Add bundle default ssh key when connecting to the VM

Before ssh key rotation, we can still use the default key to get it and
give as much as possible status details.
```
I don't know what `to get it` refers to, apart from this looks good.

Issue #131, opened 2019-05-14
Document known issues


robin-owen:
Add a new known issues "guide" to the docs. This would be a separate title from the getting started guide. (Place in `docs/source/known-issues`, maybe?)

What persistent known issues do we know about that we should document here?

gbraad:
* The `marketplace` has been disabled which causes the OperatorHub to not function

robin-owen:
* #127: Commands to resize the allocated memory, CPUs, and storage space.

We don't currently have any configuration options to use with `crc` _directly_ for these operations.

jmazzitelli:
I was going to write this up as an individual issue (let me know if I should still do that), but right now I know of no way to push images directly to the container repository in the CRC VM. In other words, I'm thinking of something analogous to "minikube docker-env" where minikube gives you a simple environment change to make and then "docker push" simply pushes images to the minikube container repo. CRC needs this (this is what I am wondering if I should write a separate issue for).

But, since there isn't anything like that now in CRC, can we document some instructions on "This is how you can set up your host and the CRC VM to have images pushed from an external client into CRC's container repo".

This is useful for people doing development on their microservices - it would be nice to just build new images and push them directly to CRC's repo for OpenShift to pick up. The only obvious alternative is the annoying workflow of pushing to some public repo (like quay.io or docker.io) and then having CRC's OpenShift have to then pull from that public repo.

robin-owen:
@jmazzitelli Go ahead and create a separate issue -- we can link to it here for documentation in the known issues.

Essentially, I'm using this issue as a "gathering place" for all of the known issues and limitations of CRC that we should document. :slightly_smiling_face:

Once I have a good list going, I'll add in a checklist to this issue.

@gbraad @praveenkumar By the way, the more I'm thinking about these limitations and how they must be configured at the moment, I'm starting to think that we could use a dedicated "Configuring CodeReady Containers" section of the documentation. I'd like to start keeping "known issues" limited to the scope of actual _issues_, not the current _limitations_ which we're planning to address in the future.

robin-owen:
@gbraad @praveenkumar To add a little more context to what I said above, we can create an assembly for configuring CRC which will contain the needed commands **at the moment**, but in the future, will instead document how to configure CRC with the `crc` binary directly.

valdar:
[...]

PR #2062, opened 2021-03-04
#2058 Create oc symlink at crc start time

### LINKED ISSUE #2058: [BUG] crc status is replacing oc symlink
### General information

  * OS: macOS

### Steps to reproduce

  1. crc status (with the tray)

### Expected

The status

### Actual

```
Error loading bundle metadata: symlink /Users/guillaumerose/.crc/cache/crc_hyperkit_4.7.0/oc /Users/guillaumerose/.crc/bin/oc/oc: file exists
```

### Description

Each time the code loads bundle details, it tries to re-create the symlink for oc. `crc status` destroys and recreates the symlink. It shouldn't. 

I believe this error happens on my mac because I have the tray polling the status every 5s.

guillaumerose:
Activate the bundle (eg install oc symlink) only on start action

Previously almost each commands recreated the symlink.
Fixes #2058

### DIFF
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -126,7 +126,7 @@ var defaultRepo = &Repository{
 	OcBinDir: constants.CrcOcBinDir,
 }
 
-func GetCachedBundleInfo(bundleName string) (*CrcBundleInfo, error) {
+func Get(bundleName string) (*CrcBundleInfo, error) {
 	return defaultRepo.Use(bundleName)
 }
 
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -48,6 +48,12 @@ func (repo *Repository) Get(bundleName string) (*CrcBundleInfo, error) {
 	if err := bundleInfo.verify(); err != nil {
 		return nil, err
 	}
+	if fmt.Sprintf(".%s", bundleInfo.ClusterInfo.AppsDomain) != constants.AppsDomain {
+		return nil, fmt.Errorf("unexpected bundle, it must have %s apps domain", constants.AppsDomain)
+	}
+	if bundleInfo.GetAPIHostname() != fmt.Sprintf("api%s", constants.ClusterDomain) {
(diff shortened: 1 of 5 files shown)


pkg/crc/machine/bundle/repository.go
```diff
@@ -127,12 +127,17 @@ var defaultRepo = &Repository{
...
 
 func Get(bundleName string) (*CrcBundleInfo, error) {
-	return defaultRepo.Use(bundleName)
+	return defaultRepo.Get(bundleName)
+}
+
+func Use(bundleName string) error {
+	_, err := defaultRepo.Use(bundleName)
+	return err
 }
```
cfergeau, line 136: Might be slightly odd that `Use` does not have the same return values as `(*Repo) Use`

pkg/crc/machine/start.go
```diff
@@ -133,6 +133,10 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 			return nil, errors.Wrap(err, "Error getting bundle metadata")
 		}
 
+		if err := bundle.Use(filepath.Base(startConfig.BundlePath)); err != nil {
+			return nil, err
+		}
+
```
cfergeau, line 139: This could be moved to `getCrcBundleInfo()` which is only used here, or maybe `bundle.Use()` could be called a single time around the `// Post-VM start` comment? At this point we don't have ssh connectivity to the VM yet, so I don't think the code tries to use oc before that.

cfergeau (approved):
Looks good to me.

pkg/crc/machine/bundle/repository.go
praveenkumar: Agree, better to have to same return value as `(*Repo) Use` and we can ignore the no used value from the caller function.

pkg/crc/machine/start.go
praveenkumar: `getCrcBundleInfo()` is only used at #131 and we should use `bundle.Use` there instead `bundle.Get` in the `getCrcBundleInfo()` and `bundle.Use` should also return the `bundleInfo` as mentioned https://github.com/code-ready/crc/pull/2062#discussion_r588125500 here.

PR #2064, opened 2021-03-05
Issue #2054 Ask user to run setup when 'autostart-tray' set to true

### LINKED ISSUE #2054: [BUG] `crc config set autostart-tray true` not working as expected
Running `crc set autostart-tray true` does not enable the tray, it runs the same codepath as for disabling it in `disableEnableTrayAutostart`.
It should probably also be using `UnloadPlist` rather than `RemovePlist` so that the services are stopped right away.

anjannath:
When 'autostart-tray' is set to false, we can just delete the launchd
config of the tray to prevent it from starting automatically on login
but when 'autostart-tray' is set to true, we need have the daemon and
the tray executable, so better to ask the user to run 'crc setup'

Fixes #2054

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -9,9 +9,9 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/tray"
+	"github.com/spf13/cast"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


gbraad:
If this is the case... can't we ASK the user to run setup and trigger the setup functions on yes? perhaps even finegrained to ONLY these calls?

cfergeau:
> If this is the case... can't we ASK the user to run setup and trigger the setup functions on yes? perhaps even finegrained to ONLY these calls?

One more thing we need to improve in the config code so we can do this without the code getting too messy :)

cmd/crc/cmd/config/config.go
```diff
@@ -46,6 +47,13 @@ func RegisterSettings(cfg *config.Config) {
 	}
 
 	disableEnableTrayAutostart := func(key string, value interface{}) string {
+		// If autostart-tray set to 'true' show message to run setup
+		if cast.ToBool(value) {
+			return fmt.Sprintf(
```
guillaumerose, line 52: If we can't make it work in the two ways, maybe we should simplify and always ask the user to do crc setup?

anjannath: Yeah, i think i'll go with this approach, will later try to implement the flow as described in https://github.com/code-ready/crc/pull/2064#issuecomment-791247341

anjannath: and for uninstalling/disable the user will have to do `crc cleanup` then `crc setup`

Issue #2053, opened 2021-03-03
[BUG] crc start returns too early on restart


guillaumerose:
### General information

  * OS: All

### Steps to reproduce

  1. crc setup
  2. crc start
  3. crc stop
  4. crc start

### Description

The last start exits too early. In the CI, we now see a lot of failures because `kubectl get co` looks good for long enough for crc. Although, the status of cluster operators is stale as they are not started.

Example 1:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/code-ready_crc/2029/pull-ci-code-ready-crc-master-e2e-crc/1367014128643739648

```
$ kubectl get pods
NAMESPACE                                    NAME                                                    READY   STATUS              RESTARTS   AGE
openshift-apiserver-operator                 openshift-apiserver-operator-665ff47758-zx5fw           0/1     ContainerCreating   0          7d4h
openshift-apiserver                          apiserver-6d4c4f5579-g58zl                              0/2     PodInitializing     0          6d4h
openshift-authentication-operator            authentication-operator-6f87fddf67-n6djg                0/1     ContainerCreating   0          7d4h
openshift-authentication                     oauth-openshift-5fd4b98799-6bvqj                        0/1     ContainerCreating   0          6d4h
openshift-authentication                     oauth-openshift-5fd4b98799-xw4jg                        0/1     ContainerCreating   0          6d4h
openshift-cluster-machine-approver           machine-approver-577bcbcd5b-s7gtq                       2/2     Running             0          7d4h
openshift-cluster-node-tuning-operator       cluster-node-tuning-operator-77dfc8648d-x4j26           0/1     ContainerCreating   0          7d4h
openshift-cluster-node-tuning-operator       tuned-phqbf                                             1/1     Running             0          7d4h
openshift-cluster-samples-operator           cluster-samples-operator-74c6f96576-6fvz6               0/2     ContainerCreating   0          7d4h
openshift-cluster-version                    cluster-version-operator-657df76796-2x58m               1/1     Running             0          7d4h
openshift-config-operator                    openshift-config-operator-5b8f5c5689-l52bz              1/1     Running             0          7d4h
...
```

```
$ kubectl get co
NAME                                       VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE
authentication                             4.7.0     True        False         False      9m15s
config-operator                            4.7.0     True        False         False      7d4h
console                                    4.7.0     True        False         False      6d4h
[...]

PR #2067, opened 2021-03-05
When all operators are running, check that containers are also up

guillaumerose:
When restarting a cluster, cluster operators statuses are stale. They
are all available but they are not running at all.

This change verifies this by looking at all containers running on the
machine with crictl. Each operator has a container that needs to be
running.

Sadly, this adds extra time when checking the status (both crc start and
crc status).

Fixes https://github.com/code-ready/crc/issues/2053

If you have other ideas! I don't really like the new call... It's one more 
slow thing to do when checking status.

### DIFF
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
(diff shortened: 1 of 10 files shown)


praveenkumar:
Shouldn't we just run an additional check  something like https://github.com/code-ready/snc/blob/master/snc-library.sh#L302-L305 with a timeout instead checking the operator pods. Also if we check operator pods is running, it doesn't mean that it's operand pods are running and we are again in the same situation so I would add another check for all the pods. The only drawback I see for checking all the pods is if some apps are deployed and it is restart scenario then we have a false wait for those pods also. 

Another thought is just check if auth pods are up (operator and operand) most of the times those pods are the one which comes up late.

guillaumerose:
I would like to have a single waiting loop. If crc waits for operators, then pods, there is no guarantee operators are still ok. 
I will try to combine both: operators + pods in openshift-* namespaces

guillaumerose:
I tried multiple times with 4.7.0 bundle and I can't reproduce the problem anymore. Let's see if the CI still complain.

PR #2069, opened 2021-03-05
Bump macos tray to 1.0.3

cfergeau:
This should be the final build for the 1.23 release.
Cropped labels are fixed, and the 'disk size' setting is removed.


cfergeau:
This syncs master with the 1.23 branch.

Issue #1991, opened 2021-02-12
Add preflight check for `resolvectl`


cfergeau:
Ubuntu 18.04LTS, and distributions derived from it (Elementary) use systemd-resolved, but don't provide a `resolvectl` executable. I don't know if there's an easy way to configure systemd-resolved on these systems, but if not, we should at least fail when `resolvectl` is missing.
See the discussion in https://github.com/code-ready/crc/issues/1572#issuecomment-777900158 for more context.

cfergeau:
This has been fixed in https://github.com/code-ready/crc/pull/2071

PR #2071, opened 2021-03-06
Add support for dynamically checking network stack

omenos:
This is a naive attempt to alter the network stack selection on Linux from hard-coded by distribution to seeing what the host is presently running with.

I don't think any changes to the test suite are needed as I didn't add/alter any checks, but I'm not sure if there needs to be a new check for the dnsmasq list to remove any systemd-resolved related dispatch files.

None of this code has been tested, so unsure if the `systemd-resolve` command will actually work.

This looks to address notes in #1991 #2056 and #1572

(I am presently fighting with VSCode on some linting that made it into the commit)

### DIFF
--- a/pkg/crc/preflight/preflight_checks_network_linux.go
+++ b/pkg/crc/preflight/preflight_checks_network_linux.go
@@ -80,9 +80,13 @@ dns=dnsmasq
 #
 # The corresponding crc bridge is not created through NetworkManager, so
 # it cannot be configured permanently through NetworkManager. We make the
-# change directly using resolvectl instead.
+# change directly using systemd-resolve instead.
 #
-# NetworkManager will overwrite this resolvectl configuration every time a
+# systemd-resolve is used instead of resolvectl due to distributions shipping
+# systemd releases older than 239 not having the newer renamed tool. resolvectl
+# supports being called as systemd-resolve, correctly handling the old CLI.
+#
+# NetworkManager will overwrite this systemd-resolve configuration every time a
 # network connection goes up/down, so we run this script on each of these events
 # to restore our settings. This is a NetworkManager bug which is fixed in
 # version 1.26.6 by this commit:
@@ -90,9 +94,7 @@ dns=dnsmasq
 
 export LC_ALL=C
 
-resolvectl domain crc ~testing
-resolvectl dns crc 192.168.130.11
-resolvectl default-route crc false
+systemd-resolve --interface crc --set-dns 192.168.130.11 --set-domain ~testing
 
 exit 0
 `
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -228,7 +228,7 @@ func getPreflightChecks(_ bool, _ bool, networkMode network.Mode) []Check {
 	return getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil)
 }
 
-func getNetworkChecksForDistro(networkMode network.Mode, systemdResolved bool) []Check {
+func getNetworkChecks(networkMode network.Mode, systemdResolved bool) []Check {
 	var checks []Check
 
 	if networkMode == network.VSockMode {
@@ -251,7 +251,7 @@ func getPreflightChecksForDistro(distro *linux.OsRelease, networkMode network.Mo
 	checks = append(checks, nonWinPreflightChecks[:]...)
 	checks = append(checks, genericPreflightChecks[:]...)
 	checks = append(checks, libvirtPreflightChecks(distro)...)
-	networkChecks := getNetworkChecksForDistro(networkMode, systemdResolved)
+	networkChecks := getNetworkChecks(networkMode, systemdResolved)
 	checks = append(checks, networkChecks...)
 	if networkMode == network.DefaultMode {
 		checks = append(checks, libvirtNetworkPreflightChecks[:]...)
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -217,16 +217,18 @@ func removeVsockCrcSettings() error {
 }
 
 func getAllPreflightChecks() []Check {
-	checks := getPreflightChecksForDistro(distro(), network.DefaultMode)
+	usingSystemdResolved := checkSystemdResolvedIsRunning()
+	checks := getPreflightChecksForDistro(distro(), network.DefaultMode, usingSystemdResolved == nil)
 	checks = append(checks, vsockPreflightChecks)
 	return checks
 }
 
 func getPreflightChecks(_ bool, _ bool, networkMode network.Mode) []Check {
-	return getPreflightChecksForDistro(distro(), networkMode)
+	usingSystemdResolved := checkSystemdResolvedIsRunning()
+	return getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil)
 }
 
-func getNetworkChecksForDistro(distro *linux.OsRelease, networkMode network.Mode) []Check {
+func getNetworkChecksForDistro(networkMode network.Mode, systemdResolved bool) []Check {
 	var checks []Check
 
 	if networkMode == network.VSockMode {
@@ -234,45 +236,28 @@ func getNetworkChecksForDistro(distro *linux.OsRelease, networkMode network.Mode
 		return append(checks, daemonRunningChecks)
 	}
 
-	switch {
-	default:
-		logging.Warnf("distribution-specific preflight checks are not implemented for '%s'", distro.ID)
-		fallthrough
-	case distroIsLike(distro, linux.Ubuntu), distroIsLike(distro, linux.Fedora):
-		checks = append(checks, nmPreflightChecks[:]...)
-		if usesSystemdResolved(distro) {
-			checks = append(checks, systemdResolvedPreflightChecks[:]...)
-		} else {
-			checks = append(checks, dnsmasqPreflightChecks[:]...)
-		}
+	checks = append(checks, nmPreflightChecks[:]...)
+	if systemdResolved {
+		checks = append(checks, systemdResolvedPreflightChecks[:]...)
+	} else {
+		checks = append(checks, dnsmasqPreflightChecks[:]...)
 	}
 
 	return checks
 }
 
-func getPreflightChecksForDistro(distro *linux.OsRelease, networkMode network.Mode) []Check {
+func getPreflightChecksForDistro(distro *linux.OsRelease, networkMode network.Mode, systemdResolved bool) []Check {
 	var checks []Check
 	checks = append(checks, nonWinPreflightChecks[:]...)
 	checks = append(checks, genericPreflightChecks[:]...)
 	checks = append(checks, libvirtPreflightChecks(distro)...)
-	networkChecks := getNetworkChecksForDistro(distro, networkMode)
+	networkChecks := getNetworkChecksForDistro(networkMode, systemdResolved)
 	checks = append(checks, networkChecks...)
 	if networkMode == network.DefaultMode {
 		checks = append(checks, libvirtNetworkPreflightChecks[:]...)
 	}
 	checks = append(checks, bundleCheck)
 	return checks
-}
-
-func usesSystemdResolved(distro *linux.OsRelease) bool {
-	switch {
-	case distroIsLike(distro, linux.Ubuntu):
-		return true
-	case distro.ID == linux.Fedora:
-		return distro.VersionID >= "33"
-	default:
-		return false
-	}
 }
 
 func distroIsLike(osRelease *linux.OsRelease, osType linux.OsType) bool {
--- a/docs/source/topics/ref_dns-configuration.adoc
+++ b/docs/source/topics/ref_dns-configuration.adoc
@@ -11,7 +11,7 @@ The `apps-crc.testing` domain is for accessing OpenShift applications deployed o
 For example, the OpenShift API server will be exposed as `api.crc.testing` while the OpenShift console is accessed through `console-openshift-console.apps-crc.testing`.
 These DNS domains are served by a `dnsmasq` DNS container running inside the {prod} virtual machine.
 
-Running [command]`{bin} setup` will adjust your system DNS configuration so that it can resolve these domains.
+Running [command]`{bin} setup` will detect and adjust your system DNS configuration so that it can resolve these domains.
 Additional checks are done to verify DNS is properly configured when running [command]`{bin} start`.
 
 [id="dns-configuration-linux_{context}"]
@@ -21,7 +21,7 @@ On Linux, depending on your distribution, {prod} expects the following DNS confi
 
 === NetworkManager + systemd-resolved
 
-This configuration is used on Fedora 33 or newer, and on Ubuntu Desktop editions.
+This configuration is used by default on Fedora 33 or newer, and on Ubuntu Desktop editions.
 
 * {prod} expects NetworkManager to manage networking.
 * {prod} configures `systemd-resolved` to forward requests for the `testing` domain to the `192.168.130.11` DNS server.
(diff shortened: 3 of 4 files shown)


omenos:
Some of the tests for `preflight_linux_test.go` need to be redesigned, as they are writen with distro specific choices hardcoded.

guillaumerose:
Thanks for your PR! Let's focus on the feature now and fix the tests after. I am sure we can mock `os.Stat` somehow.

@cfergeau WDYT of this new way of checking network manager ?

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -72,9 +72,29 @@ server=/crc.testing/192.168.130.11
...
+#
+# NetworkManager will overwrite this systemd-resolve configuration every time a
+# network connection goes up/down, so we run this script on each of these events
+# to restore our settings. This is a NetworkManager bug which is fixed in
+# version 1.26.6 by this commit:
+# https://cgit.freedesktop.org/NetworkManager/NetworkManager/commit/?id=ee4e679bc7479de42780ebd8e3a4d74afa2b2ebe
+
+export LC_ALL=C
+
+systemd-resolve --interface crc --set-dns 192.168.130.11 --set-domain ~testing
```
cfergeau, line 93: Ideally we'd be able to use this command both on older and newer systems? Any idea if this would work? The introduction of `crcNetworkManagerDispatcherLegacyConfig` deserves to be split in its own commit though.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -295,3 +282,10 @@ func distro() *linux.OsRelease {
 	}
 	return distro
 }
+
+func resolvectlExists() bool {
+	if _, err := os.Stat("/usr/bin/resolvectl"); err != nil {
+		return false
+	}
+	return true
```
cfergeau, line 290: This could be `return crcos.FileExists("/usr/bin/resolvectl")`

cfergeau:
> @cfergeau WDYT of this new way of checking network manager ?

Slightly worried we're going to hit corner cases of `checkSystemdResolvedIsRunning`. This is not going to do the right thing on a distro where `DNSStubListener` is set to `no` in `/etc/systemd/resolved.conf`. Before, this code was mostly used as a failsafe on distros where we were sure we want to use systemd-resolved. Now, it's going to be this method which determines if we use systemd-resolved or dnsmasq, so it becomes more important to get it right.

I'll need to take a look at a few non-systemd-resolved distros to make sure they don't use this setup. We can move forward with this in the mean time.

pkg/crc/preflight/preflight_linux.go
omenos: Would it be better to replace the `resolvectlExists()` function and calls with that function completely?

pkg/crc/preflight/preflight_checks_network_linux.go
omenos: `resolvectl` supports being called as `systemd-resolve` (which on RHEL 8 is a symlink to `resolvectl`, and same with `resolvconf`) and will convert its arguments to the necessary compatibility points. The native syntax type for `resolvectl` is the new verb based CLI.

https://github.com/systemd/systemd/blob/9813d8354cc9af33b739e8d478368adfb548edc1/src/resolve/resolvectl.c#L3398-L3403

However, that only applies to 239+ where the new binary change was made. In distributions using 238- there is no translation to the older syntax, so you have to use the original CLI. The options I've used here are my best guess to what the original syntax would have been.

So, hypothetically we could use the old `systemd-resolve` as the official call, which would work (assuming my command is correct) on older distributions like Ubuntu 18.04, and on newer systems using 239+ as it'll get translated properly under the hood by `resolvectl`.

omenos: Sorry for the multi-comments, these things come to me as I look at the code. The reason I went with the multi-config approach is because I didn't find a comparable argument set for the `resolvectl default-route crc false`.

cfergeau: I'm not sure `default-route` is important, it's one of these "should not hurt to have this" addition, but I'd say having it or not should not make a difference.

cfergeau: Ubuntu 20.04 has `resolvctl` and `set-default`. The 'resolvconf' executable from the `resolvconf` package is a shell script, not sure it's related to systemd at all.

pkg/crc/preflight/preflight_linux.go
cfergeau: yep, makes sense, avoids the 1 line helper. Add a `const resolvectlPath="/usr/bin/resolvectl"`  though if you use `crcos.FileExists()` in the other code.

cfergeau:
> > @cfergeau WDYT of this new way of checking network manager ?
> 
> Slightly worried we're going to hit corner cases of `checkSystemdResolvedIsRunning`. This is not going to do the right thing on a distro where `DNSStubListener` is set to `no` in `/etc/systemd/resolved.conf`. Before, this code was mostly used as a failsafe on distros where we were sure we want to use systemd-resolved. Now, it's going to be this method which determines if we use systemd-resolved or dnsmasq, so it becomes more important to get it right.
> 
> I'll need to take a look at a few non-systemd-resolved distros to make sure they don't use this setup. We can move forward with this in the mean time.

I think what I had in mind was systemd-resolved on RHEL8 which has a default configuration where it does not bind on 127.0.0.53:53, but can still get UDP DNS queries. The RHEL doc linked in https://github.com/code-ready/crc/pull/2056 explains how to configure this properly (NM configuration change is needed). Since systemd-resolved is not running by default on RHEL (tried a fresh 8.3 install), I don't think this is an issue. So it should be fine to detect systemd-resolved at runtime instead of the distro matching we currently are doing.

pkg/crc/preflight/preflight_linux.go
omenos: I have that implemented locally, but added that const to the linux_constants file. Would you rather that be done within the scope of the preflight? Wasn’t sure if that was the appropriate place for such a variable.

pkg/crc/preflight/preflight_checks_network_linux.go
omenos: Looking at it's sources, it doesn't appear to be. `resolvectl` has the ability to act as a passthrough for `resolvconf`, which it is in the Red Hat family. Though it looks like Canonical isn't following suite on that just yet.

I'll get some eyes on my `systemd-resolve` command and check to make sure it's an accurate representation of what's the `resolvectl` runs are doing. The other option is to just not support systemd-resolved if `resolvectl` isn't available (something I was playing with originally).

omenos: I did a quick test with a Fedora 33 Server install running the aforementioned `systemd-resolve` command (swapped `--interface` with a different device) and it seemed to produce the output matching the `crc` interface managed by `resolvectl`.

pkg/crc/preflight/preflight_linux.go
omenos: This is no longer relevant with the latest commit pushed.

omenos:
If approved and merged, #2056 will likely need some changes.

gbraad (approved):

gbraad:
have not been able to verify this... but the impact seems minimal

gbraad:
> If approved and merged, #2056 will likely need some changes.

can you detail this?

pkg/crc/preflight/preflight_linux.go
```diff
@@ -227,17 +227,15 @@ func getNetworkChecksForDistro(distro *linux.OsRelease, networkMode network.Mode
...
-		} else {
-			checks = append(checks, dnsmasqPreflightChecks[:]...)
-		}
+	}
+
+	checks = append(checks, nmPreflightChecks[:]...)
+	if checkSystemdResolvedIsRunning() == nil {
+		checks = append(checks, systemdResolvedPreflightChecks[:]...)
+	} else {
+		checks = append(checks, dnsmasqPreflightChecks[:]...)
```
praveenkumar, line 238: Shouldn't this part of `else` block ? we don't want to have any nm preflight checks if distro is not like ubuntu or fedora since it's not tested and fix might not work.

omenos: > fix might not work.
Are you referring to this patch or to the NetworkManager preflight operations? 

When authoring this, I was under the strict interpretation that NetworkManager is required, so if it were missing the run should fail.

From the docs:
> CodeReady Containers requires the libvirt and NetworkManager packages to run on Linux.

Are you meaning to say that the preflights may not function correctly on another distribution using NetworkManager due to that distribution’s configuration?

praveenkumar: @omenos Ignore it, I just missed we were using `fallthrough` so the logic is still the same.

omenos:
> can you detail this?

The doc change details how to configure a RHEL 8.3 system using systemd-resolved for CRC usage. Those changes include disabling preflight checks, a step which wouldn’t be needed with this patch.

However, until this patch or a similar one is really approved/merged, I wouldn’t touch those docs. And the mock tests still need to be fixed in this PR.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -227,17 +227,15 @@ func getNetworkChecksForDistro(distro *linux.OsRelease, networkMode network.Mode
...
-		checks = append(checks, nmPreflightChecks[:]...)
-		if usesSystemdResolved(distro) {
-			checks = append(checks, systemdResolvedPreflightChecks[:]...)
-		} else {
-			checks = append(checks, dnsmasqPreflightChecks[:]...)
-		}
+	}
+
+	checks = append(checks, nmPreflightChecks[:]...)
+	if checkSystemdResolvedIsRunning() == nil {
```
praveenkumar, line 235: The integration tests are failing because of this logic now before as part of `uessSystemResolved` function we explicit mentioned if `distroIsLike` ubuntu or `fedora>= 33` then we should include this reolved preflight check but now unit test depend on the system where it is executed and `crcSystemResolvedIsRunning` deciding factor.

guillaumerose: `distro` argument is not needed anymore in this function in fact. You can remove it from here. 

For fixing the tests, I would: 
* Add a bool like `systemdResolved bool` to `getPreflightChecksForDistro` function. 
* Pass `checkSystemdResolvedIsRunning()` in  `getPreflightChecks`
* Add a new field like `systemdResolved bool` in:

```
type checkListForDistro struct {
	distro      *crcos.OsRelease
	networkMode network.Mode
	checks      []Check
}

```
* Update test cases accordingly. 

No worry, if you don't have time to do it. Let me know I can do it.

omenos: I'll take a shot at it and see what happens! (also, re: docs, the dispatcher script would also need to be modified to be the `systemd-resolve` command, rather than `resolvectl`).

cfergeau:
I've tested the current version of this PR on freshly installed ubuntu 20.04/Fedora 32/Fedora 33/RHEL8.3 and they all worked fine.

pkg/crc/preflight/preflight_linux.go
omenos: @guillaumerose I take it I'm going to have to duplicate each test case that uses `network.DefaultMode`?

Also, `checkSystemdResolvedIsRunning()` returns an `error` rather than a bool, should I do a conversion internal to the calling function, i.e. `if cSRIR() == null: otherVar := true else false`, and pass that to `getPreflightChecksForDistro` or will the `error` be fine?

Just want to be sure of what you consider best practice for this.

omenos:
@guillaumerose I took a shot at the tests, but they don't seem to have worked out. I may have been overthinking what you were suggesting. I'm a little confused why `checkSystemdResolvedIsRunning()` function isn't being recognized within the `preflight` package according to some of the tests run above. If you would like, I can revert this commit, add the small doc change that should accompany it, and let you handle the tests the way you see fit.

cfergeau:
> @guillaumerose I took a shot at the tests, but they don't seem to have worked out. I may have been overthinking what you were suggesting. I'm a little confused why `checkSystemdResolvedIsRunning()` function isn't being recognized within the `preflight` package according to some of the tests run above. If you would like, I can revert this commit, add the small doc change that should accompany it, and let you handle the tests the way you see fit.

The code in `*_linux.go` files is only built on linux. `checkSystemdResolvedIsRunning` is defined in such a file, so it can't be used in `preflight.go`: `preflight.go` is built on all 3 platforms (macOS, Windows, Linux) while `checkSystemdResolvedIsRunning` is linux-only. If you move the logic to `getPreflightChecks` in `preflight_linux.go` (which is also only built on linux), then it builds fine. I did not try to fix the test cases.

```diff
diff --git a/pkg/crc/preflight/preflight.go b/pkg/crc/preflight/preflight.go
index 5bc8414c..05686689 100644
--- a/pkg/crc/preflight/preflight.go
+++ b/pkg/crc/preflight/preflight.go
@@ -155,12 +155,7 @@ func StartPreflightChecks(config config.Storage) error {
        experimentalFeatures := config.Get(cmdConfig.ExperimentalFeatures).AsBool()
        mode := network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString())
        trayAutostart := config.Get(cmdConfig.AutostartTray).AsBool()
-       usingSystemdResolved := checkSystemdResolvedIsRunning()
-       useSystemdResolved := false
-       if usingSystemdResolved == nil {
-               useSystemdResolved = true
-       }
-       if err := doPreflightChecks(config, getPreflightChecks(experimentalFeatures, trayAutostart, mode, useSystemdResolved)); err != nil {
+       if err := doPreflightChecks(config, getPreflightChecks(experimentalFeatures, trayAutostart, mode)); err != nil {
                return &errors.PreflightError{Err: err}
        }
 
@@ -172,12 +167,7 @@ func SetupHost(config config.Storage) error {
        experimentalFeatures := config.Get(cmdConfig.ExperimentalFeatures).AsBool()
        mode := network.ParseMode(config.Get(cmdConfig.NetworkMode).AsString())
        trayAutostart := config.Get(cmdConfig.AutostartTray).AsBool()
-       usingSystemdResolved := checkSystemdResolvedIsRunning()
-       useSystemdResolved := false
-       if usingSystemdResolved == nil {
-               useSystemdResolved = true
-       }
-       return doFixPreflightChecks(config, getPreflightChecks(experimentalFeatures, trayAutostart, mode, useSystemdResolved))
+       return doFixPreflightChecks(config, getPreflightChecks(experimentalFeatures, trayAutostart, mode))
 }
 
 func RegisterSettings(config config.Schema) {
diff --git a/pkg/crc/preflight/preflight_linux.go b/pkg/crc/preflight/preflight_linux.go
index c73b34d1..25c73680 100644
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -215,8 +215,10 @@ func getAllPreflightChecks() []Check {
        return checks
 }
 
-func getPreflightChecks(_ bool, _ bool, networkMode network.Mode, systemdResolved bool) []Check {
-       return getPreflightChecksForDistro(distro(), networkMode, systemdResolved)
+func getPreflightChecks(_ bool, _ bool, networkMode network.Mode) []Check {
+       usingSystemdResolved := checkSystemdResolvedIsRunning()
+
+       return getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil)
 }
 
 func getNetworkChecksForDistro(networkMode network.Mode, systemdResolved bool) []Check {
```

omenos:
Made the requested changes and fixed issues with the tests (forgot to create a third `unexpected` test case).

Something that did seem a bit strange is that the `TestApi` and `TestHTTPApi` tests will fail if the `systemdResolved` component in the `getAllPreflightChecks` function is set to false, which is what I had it as originally. The API tests will then complain that they couldn't find the dispatcher script and bomb out.

However, with those problems out of the way, I don't know if I can solve the current e2e-crc and PR issues, as they seem (to me) completely unrelated to my code changes.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -295,3 +282,10 @@ func distro() *linux.OsRelease {
 	}
 	return distro
 }
+
+func resolvectlExists() bool {
+	if _, err := os.Stat("/usr/bin/resolvectl"); err != nil {
```
praveenkumar: Instead of hard coding the abs path of `resolvectl` shouldn't just use `exec.LookPath("resolvectl")` ?

cfergeau:
> Something that did seem a bit strange is that the `TestApi` and `TestHTTPApi` tests will fail if the `systemdResolved` component in the `getAllPreflightChecks` function is set to false, which is what I had it as originally. The API tests will then complain that they couldn't find the dispatcher script and bomb out.

This happens if `getPreflightChecks` and `getAllPreflightChecks` are out of sync, the api tests call `doRegisterSettings` which uses `getAllPreflightChecks`. Then it disables all the preflight checks which were registered through `doRegisterSettings`. The failure you saw happened because this specific tests was not disabled correctly because the 2 lists were different. If you change both `getPreflightChecks` and `getAllPreflightChecks` to use false then the error no longer occurs.

pkg/crc/preflight/preflight_linux.go
omenos: This function is no longer being used since switching away from `resolvectl` in favor of `systemd-resolve`.

cfergeau (commented):
I've squashed a few commits together, and moved all the systemd-resolve/resolvectl changes to a single commit, changes available at https://github.com/cfergeau/crc/commits/omenos (not sure I can directly edit this PR).
@omenos if that's fine with you, I'll go with the changes as they are in this branch.

omenos:
@guillaumerose Those changes lgtm. There should also be some changes to the DNS topic in the docs to correct the language and the script provided.

omenos:
@cfergeau I replaced my branch with yours and added the documentation changes.

praveenkumar:
/lgtm

/assign @robin-owen 
(for doc changes)

guillaumerose:
Thanks for the PR @omenos !

Issue #2048, opened 2021-03-03
Have different kubeadmin password for each fresh start of crc


praveenkumar:
As of now all the crc instances for specific bundle have same `kubeadmin` password. For security reason we should take another look and try to regenerate those for each fresh crc start.

PR #2072, opened 2021-03-08
Update kubeadmin password for each fresh start

praveenkumar:
This patch will create a random password and then update the cluster
secret for htpasswd. It also update the cache kubeadmin-password file
so that all the commands which rely on it doesn't break.

fixes: #2048

### DIFF
--- a/pkg/crc/cluster/kubeadminPassword.go
+++ b/pkg/crc/cluster/kubeadminPassword.go
@@ -1,0 +1,94 @@
+package cluster
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"math/big"
+
+	"github.com/code-ready/crc/pkg/crc/machine/bundle"
+	"github.com/code-ready/crc/pkg/crc/oc"
+	"golang.org/x/crypto/bcrypt"
+)
+
+// UpdateKubeAdminUserPassword does following
+// - Create and put updated kubeadmin password to ~/.crc/machine/crc/kubeadmin-password
+// - Update the htpasswd secret
+func UpdateKubeAdminUserPassword(kubeAdminPassword string, ocConfig oc.Config, bundle *bundle.CrcBundleInfo) error {
+	// In case of cluster started from stopped state.
+	if kubeAdminPassword == "" {
+		return nil
+	}
+	hashDeveloperPasswd, err := hashBcrypt("developer")
+	if err != nil {
+		return err
+	}
+
+	hashKubeAdminPasswd, err := hashBcrypt(kubeAdminPassword)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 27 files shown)


pkg/crc/cluster/kubeadminPassword.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/code-ready/crc/pkg/crc/machine/bundle"
+	"github.com/code-ready/crc/pkg/crc/oc"
+	"golang.org/x/crypto/bcrypt"
+)
+
+// UpdateKubeAdminUserPassword does following
+// - Generate a random password for kubeadmin user
+// - Update the kubeadmin password file
+// - Update the htpasswd secret
+func UpdateKubeAdminUserPassword(kubeAdminPassword string, ocConfig oc.Config, bundle *bundle.CrcBundleInfo) error {
```
guillaumerose, line 18: Is it possible to create a new file with the password in ~/.crc/machines/crc/password ? I don't really like changing a file in the cache after start. It is like "corrupting" it.

pkg/crc/cluster/kubeadminPassword.go
```diff
@@ -0,0 +1,95 @@
...
+	// In case of cluster started from stopped state.
+	if kubeAdminPassword == "" {
+		return nil
+	}
+	hashDeveloperPasswd, err := hashBcrypt("developer")
```
guillaumerose, line 23: developer user is sudoer. I guess we should also rotate the password. WDYT?

praveenkumar: we can remove developer from sudoer.

praveenkumar:
I am thinking why we have any password set from snc side? We do have kubecofig file for `system:admin` role and same is consumed by all the `oc` commands. We can consume same kubeconfig file for creating <user> - <password> from crc side?

guillaumerose:
Yes we can remove the kubeadmin file from the bundle :+1:

pkg/crc/cluster/kubeadminPassword.go
guillaumerose: Yes but we need to also to update the doc to remove this kind of commands: `oc get co --as system:admin`.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -111,13 +111,27 @@ func (bundle *CrcBundleInfo) GetInitramfsPath() string {
 }
 
 func (bundle *CrcBundleInfo) GetKubeadminPassword() (string, error) {
```
guillaumerose, line 113: You can delete this function in fact now. kubeadmin password in bundle will disppear. It removes the if in ~/.crc/machines/... below.

guillaumerose: Ok :+1:

PR #2086, opened 2021-03-11
Slightly change the end of crc start

guillaumerose:
* List only 5 operators in the status debug string
* Remove the not-so-useful warning

### DIFF
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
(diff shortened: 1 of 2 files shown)


pkg/crc/cluster/clusteroperator.go
```diff
@@ -50,6 +51,23 @@ func (status *Status) String() string {
 	return "Operators are not ready yet"
 }
 
+func joinWithLimit(names []string) string {
```
cfergeau, line 54: I'd pass the limit as an argument, but not really important :)

pkg/crc/cluster/clusteroperator.go
```diff
@@ -50,6 +51,23 @@ func (status *Status) String() string {
...
+	const maxNames = 5
+	sort.Strings(names)
+	ret := strings.Join(names[0:min(len(names), maxNames)], ", ")
+	if len(names) > maxNames {
+		ret += "..."
```
cfergeau, line 59: `, ...` ?

pkg/crc/machine/start.go
```diff
@@ -410,7 +410,7 @@ func (client *client) Start(ctx context.Context, startConfig StartConfig) (*Star
 
 	logging.Info("Adding crc-admin and crc-developer contexts to kubeconfig...")
 	if err := writeKubeconfig(instanceIP, clusterConfig); err != nil {
-		logging.Warnf("Cannot update kubeconfig: %v", err)
+		logging.Errorf("Cannot update kubeconfig: %v", err)
```
cfergeau, line 413: Slightly inaccurate commit log maybe, it says:
> The previous call ensures all operators are running

but when `WaitForClusterStable` fails, it only logs an error, it does not return, so we could perfectly get there without all operators running. But this is minor, makes sense to try to do these updates even if the operators are not all running.

cfergeau (approved):

pkg/crc/cluster/clusteroperator.go
guillaumerose: done

pkg/crc/cluster/clusteroperator.go
guillaumerose: done

PR #2088, opened 2021-03-11
Propagate context to eventually cancel the startup sequence of the VM

guillaumerose:
If the machine.Start() caller cancels the context, as soon as the code
will enter a wait loop, start will returned an error.

### DIFF
--- a/pkg/crc/cluster/status.go
+++ b/pkg/crc/cluster/status.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"context"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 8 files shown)


pkg/crc/cluster/status.go
```diff
@@ -41,7 +46,11 @@ func WaitForClusterStable(ocConfig oc.Config, monitoringEnabled bool) error {
 		} else {
 			count = 0
 		}
-		time.Sleep(retryDuration)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(retryDuration):
+		}
 	}
```
cfergeau, line 54: Maybe this could also use `RetryAfterWithContext` ?

cfergeau (approved):
Short explanation why we want to `Reduce the ssh timeout in order to trigger more retries`  would not hurt, apart from this looks good to me.

PR #2090, opened 2021-03-11
Fix #2089 Update Windows tray assembly name

### LINKED ISSUE #2089: Update names of the Windows tray binaries
The tray on Windows has been renamed to `crc-tray.exe` This however is not reflected in the preflight checks.

gbraad:
**Fixes:** Issue #2089

**Relates to:** https://github.com/code-ready/tray-windows/commit/1e4c3b1c86293aed98af725370a4c7fd62ac39d7

## Solution/Idea

The assembly for the Windows tray is now `crc-tray.exe`

## Testing

Preflight checks for setting up the tray should still work, and the tray should start


pkg/crc/constants/constants_windows.go
```diff
@@ -9,8 +9,8 @@ const (
 	OcExecutableName            = "oc.exe"
 	PodmanExecutableName        = "podman.exe"
 	AdminHelperExecutableName   = "admin-helper-windows.exe"
-	TrayExecutableName          = "tray-windows.exe"
-	TrayShortcutName            = "tray-windows.lnk"
+	TrayExecutableName          = "crc-tray.exe"
+	TrayShortcutName            = "crc-tray.lnk"
```
cfergeau, line 13: Has there been a release with these changes? Seems like this would break `crc.exe setup` with the tray enabled until there is a new release?

gbraad: hence I would not merge... but this change is needed...
... really depends on if we bump the release. I am not inclined at the moment.

cfergeau (changes requested):
Explaining why we want to change the name would ensure everyone is on the same line.

cfergeau:
Do we need to add more code to handle upgrades? (remove tray with the old name, replace with the new one)
I forgot how the tray code is on Windows.

gbraad:
> Explaining why we want to change the name would ensure everyone is on the same line.

To make sure the binary is identifiable from the Process manager as belonging to CRC. The name is otherwise to generic.

pkg/crc/constants/constants_windows.go
gbraad: @anjannath any thoughts on this?

anjannath:
> Do we need to add more code to handle upgrades? (remove tray with the old name, replace with the new one)
I forgot how the tray code is on Windows.

Yeah, this is needed, i took the commit from here and opened #2290

pkg/crc/constants/constants_windows.go
anjannath: @gbraad Close this in favor of #2290 ? that handles the upgrade scenario!

anjannath: New release has been published with the changed name: https://github.com/code-ready/tray-windows/releases/tag/v0.5.0.0

anjannath (approved):

PR #2091, opened 2021-03-11
Validate sudo before using it

guillaumerose:
Also remove home directory from telemetry errors.

### DIFF
--- a/pkg/crc/telemetry/telemetry.go
+++ b/pkg/crc/telemetry/telemetry.go
@@ -5,6 +5,8 @@ import (
 	"os/user"
 	"strings"
 	"sync"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
 )
 
 type contextKey struct{}
@@ -93,5 +95,6 @@ func SetError(err error) string {
 	if err1 != nil {
 		return err1.Error()
 	}
-	return strings.ReplaceAll(err.Error(), user.Username, "XXXXX")
+	withoutHomeDir := strings.ReplaceAll(err.Error(), constants.GetHomeDir(), "$HOME")
+	return strings.ReplaceAll(withoutHomeDir, user.Username, "$USERNAME")
 }
--- a/pkg/os/exec.go
+++ b/pkg/os/exec.go
@@ -2,6 +2,7 @@ package os
 
 import (
 	"bytes"
+	"errors"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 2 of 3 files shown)


pkg/os/exec.go
```diff
@@ -48,6 +48,9 @@ func RunPrivileged(reason string, cmdAndArgs ...string) (string, string, error)
 	if err != nil {
 		return "", "", errors.New("sudo binary not found")
 	}
+	if _, _, err := run(sudo, []string{"--validate"}, map[string]string{}); err != nil {
+		return "", "", errors.New("user may not use sudo")
+	}
```
cfergeau, line 53: This feels like something which should be part of a preflight check rather than something we run every time `RunPrivileged` is called.

cfergeau (commented):
Apart from sudo validation, looks good to me.

pkg/os/exec.go
guillaumerose: Yes but if everything is ok in crc setup, sudo is not run at all. If I change it to a preflight, it will always force the user to type his password.

pkg/os/exec.go
```diff
@@ -45,7 +46,7 @@ func runPrivate(command string, args []string, env map[string]string) (string, s
 func RunPrivileged(reason string, cmdAndArgs ...string) (string, string, error) {
 	sudo, err := exec.LookPath("sudo")
 	if err != nil {
-		return "", "", err
+		return "", "", errors.New("sudo binary not found")
```
cfergeau, line 49: `"sudo executable not found"` is more accurate.

pkg/os/exec.go
cfergeau: Since you want to make this error 'special' («It will help us group this kind of errors under the same name in the telemetry. ») , might be worth having a constant for it, or even its own error type. 
`"user is not allowed to use the 'sudo' command"` or somethnig like this rather than `may not use`?

pkg/crc/telemetry/telemetry.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os/user"
 	"strings"
 	"sync"
+
```
cfergeau, line 8: The main thing this commit achieves is that it replaces `/home/XXXXX` in telemetry errors with `$HOME`, and same for `XXXXX` as a username replacement which is replaced with `$USERNAME`.
The homedir was already sanitized before being sent to telemetry through username replacement.

guillaumerose:
retest this please
/retest

cfergeau (approved):

pkg/os/exec.go
guillaumerose: yes, all done!

PR #2096, opened 2021-03-15
Change the root check error message

guillaumerose:
Hopefully, it will increase the success rate of crc setup.

It's not perfect but I guess still good to take:

First run:
```
$ sudo crc setup
CodeReady Containers is constantly improving and we would like to know more about usage (more details at https://developers.redhat.com/article/tool-data-collection)
Your preference can be changed manually if desired using 'crc config set consent-telemetry <yes/no>'
Would you like to contribute anonymous usage statistics? [y/N]: 
No worry, you can still enable telemetry manually with the command 'crc config set consent-telemetry yes'.
INFO Checking if running as root                  
crc should not be run as root
``` 

The telemetry message is annoying here :/

### DIFF
--- a/pkg/crc/preflight/preflight_checks_nonwin.go
+++ b/pkg/crc/preflight/preflight_checks_nonwin.go
@@ -3,6 +3,7 @@
 package preflight
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"syscall"
@@ -17,7 +18,7 @@ var nonWinPreflightChecks = [...]Check{
 		configKeySuffix:  "check-root-user",
 		checkDescription: "Checking if running as non-root",
 		check:            checkIfRunningAsNormalUser,
-		fixDescription:   "crc should be ran as a normal user",
+		fixDescription:   "crc should not be run as root",
 		flags:            NoFix,
 	},
 	{
@@ -32,7 +33,7 @@ func checkIfRunningAsNormalUser() error {
 		return nil
 	}
 	logging.Debug("Ran as root")
-	return fmt.Errorf("crc should be ran as a normal user")
+	return errors.New("crc should not be run as root")
 }
 
 func setSuid(path string) error {


cfergeau:
I don't think we want to make it easier to disable this check? Either the check is needed, and we don't want to encourage people to disable it, or the check is not needed, and we should remove it.

pkg/crc/preflight/preflight_checks_nonwin.go
```diff
@@ -12,12 +13,17 @@ import (
 	crcos "github.com/code-ready/crc/pkg/os"
 )
 
+const (
+	rootCheck = "check-root-user"
+	rootError = "crc should not be ran as root. Turn off this check with 'crc config set skip-" + rootCheck + " true'"
```
jsliacan, line 18: 'ran' --> 'run'

guillaumerose:
Should we keep this check as a warning only?

cfergeau:
> Should we keep this check as a warning only?

Why? What problem is this solving? People are running `crc setup` as root and this is error report in telemetry? After getting this failure, do people successfully run `crc setup` as a regular user? Or are there cases when this *must* be run as root, and the preflight check gets in the way?

guillaumerose:
According to telemetry, 200 people over 419 people dropped using crc after this error. They downloaded crc, they encountered this error, and they dropped using crc.

After this, 119 people never reached a successful crc setup. 

100 people who encountered this error got at some point a working crc setup.

cfergeau:
> According to telemetry, 200 people over 419 people dropped using crc after this error. They downloaded crc, they encountered this error, and they dropped using crc.

I really doubt half of the people hitting this 'root' message gave up on crc because of this.  


> After this, 119 people never reached a successful crc setup.

But surely the problem was not that they were unable to switch from root to regular user?

My feeling is that with just telemetry data, we are missing the bigger picture which we could use to improve on this.

guillaumerose:
Also, minikube doesn't check for root.

guillaumerose:
I removed most of it and just kept the wording change.

praveenkumar:
with current wording.

/lgtm

PR #2101, opened 2021-03-15
#949 Add synchronisation between Start/Stop/Delete actions

guillaumerose:
Related to https://github.com/code-ready/crc/issues/949

This PR implements cancel for machine start.

If the user clicks stop or delete in the tray, it will first abort startup sequence and then perform stop/delete.

### DIFF
--- a/pkg/crc/machine/sync.go
+++ b/pkg/crc/machine/sync.go
@@ -1,0 +1,170 @@
+package machine
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/machine/libmachine/state"
+)
+
+const startCancelTimeout = 15 * time.Second
+
+type State string
+
+const (
+	Idle     State = "Idle"
+	Deleting State = "Deleting"
+	Stopping State = "Stopping"
+	Starting State = "Starting"
+)
+
+type Synchronized struct {
+	underlying Client
+
+	stateLock    sync.Mutex
+	currentState State
+	startCancel  context.CancelFunc
+
+	syncOperationDone chan State
+}
+
+func NewSynchronizedMachine(machine Client) *Synchronized {
+	return &Synchronized{
+		underlying:        machine,
+		currentState:      Idle,
+		syncOperationDone: make(chan State, 1),
+	}
+}
+
+func (s *Synchronized) CurrentState() State {
+	s.stateLock.Lock()
+	defer s.stateLock.Unlock()
+
+	return s.currentStateUnlocked()
+}
+
+func (s *Synchronized) currentStateUnlocked() State {
+	select {
+	case st := <-s.syncOperationDone:
+		if s.currentState == st {
+			s.currentState = Idle
+		}
+		if st == Starting {
+			s.startCancel = nil
+		}
+	default:
+	}
+	return s.currentState
+}
+
+func (s *Synchronized) Delete() error {
+	if err := s.prepareStopDelete(Deleting); err != nil {
+		return err
+	}
+
+	err := s.underlying.Delete()
+	s.syncOperationDone <- Deleting
+	return err
+}
+
+func (s *Synchronized) prepareStart(startCancel context.CancelFunc) error {
+	s.stateLock.Lock()
+	defer s.stateLock.Unlock()
+	if s.currentStateUnlocked() != Idle {
+		return errors.New("cluster is busy")
+	}
+	s.startCancel = startCancel
+	s.currentState = Starting
+
+	return nil
+}
+
+func (s *Synchronized) Start(ctx context.Context, startConfig StartConfig) (*StartResult, error) {
+	ctx, startCancel := context.WithCancel(ctx)
+	if err := s.prepareStart(startCancel); err != nil {
+		return nil, err
+	}
+
+	startResult, err := s.underlying.Start(ctx, startConfig)
+	s.syncOperationDone <- Starting
+	return startResult, err
+}
+
+/* cancel ongoing start, and wait until the start is fully cancelled. Time out if cancellation takes more than 'timeout'
+ * s.stateLock must be locked before calling this function
+ */
+func (s *Synchronized) cancelUnlocked(timeout time.Duration) error {
+	if s.startCancel != nil {
+		logging.Infof("Cancelling virtual machine start...")
+		s.startCancel()
+	}
+	select {
+	case <-s.syncOperationDone:
+		return nil
+	case <-time.After(timeout):
+		return errors.New("cannot abort startup sequence quickly enough")
+	}
+}
+
+func (s *Synchronized) prepareStopDelete(state State) error {
+	s.stateLock.Lock()
+	defer s.stateLock.Unlock()
+
+	switch s.currentStateUnlocked() {
+	case Starting:
+		if err := s.cancelUnlocked(startCancelTimeout); err != nil {
+			return err
+		}
+	case Idle:
+		break
+	case Deleting, Stopping:
+		return errors.New("cluster is stopping or deleting")
+	default:
+		return errors.New("invalid condition")
+	}
+
+	s.currentState = state
+	return nil
+}
+
+func (s *Synchronized) Stop() (state.State, error) {
+	if err := s.prepareStopDelete(Stopping); err != nil {
+		return state.Error, err
+	}
+
+	st, err := s.underlying.Stop()
+	s.syncOperationDone <- Stopping
+
+	return st, err
+}
+
+func (s *Synchronized) GetName() string {
+	return s.underlying.GetName()
+}
+
+func (s *Synchronized) Exists() (bool, error) {
+	return s.underlying.Exists()
+}
+
+func (s *Synchronized) GetConsoleURL() (*ConsoleResult, error) {
+	return s.underlying.GetConsoleURL()
+}
+
+func (s *Synchronized) IP() (string, error) {
+	return s.underlying.IP()
+}
+
+func (s *Synchronized) PowerOff() error {
+	return s.underlying.PowerOff()
+}
+
+func (s *Synchronized) Status() (*ClusterStatusResult, error) {
+	return s.underlying.Status()
+}
+
+func (s *Synchronized) IsRunning() (bool, error) {
+	return s.underlying.IsRunning()
+}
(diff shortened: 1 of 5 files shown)


cfergeau (commented):
Since you introduce a new struct, have you considered using a regular mutex to protect access to the struct, and either some booleans, or some state enum, `startInProgress bool` `stopInProgress bool` `deleteInProgress bool`? This would be less scary than these locks, trylock/,.... 
Hard to know the exact behaviour you are after from the one line logs, so no idea if this is workable or not ^^

guillaumerose:
> Since you introduce a new struct, have you considered using a regular mutex to protect access to the struct, and either some booleans, or some state enum, startInProgress bool stopInProgress bool deleteInProgress bool? This would be less scary than these locks, trylock/,....

It is already hard to deal with 2 locks and one cancel function. I don't want to have to deal with more objects.
The test coverage should enable refactorings/enhancements.

> Hard to know the exact behaviour you are after from the one line logs, so no idea if this is workable or not ^^

This is why I added 3 tests. It will be easy to change the implementation. If there is something missing add a test, fix the code and move on.

Hundred lines of commit log doesn't worth a single test. 
I can also add: I don't trust a commit log, I trust tests.

cfergeau:
> > Since you introduce a new struct, have you considered using a regular mutex to protect access to the struct, and either some booleans, or some state enum, startInProgress bool stopInProgress bool deleteInProgress bool? This would be less scary than these locks, trylock/,....
> 
> It is already hard to deal with 2 locks and one cancel function. I don't want to have to deal with more objects.

The point would be there would be a single lock protecting the instance, lock, read value you are interested in, unlock, act on the value that you read. No need for additonal locks, for trylock, ..

> 
> > Hard to know the exact behaviour you are after from the one line logs, so no idea if this is workable or not ^^
> 
> This is why I added 3 tests. It will be easy to change the implementation. If there is something missing add a test, fix the code and move on.
> 
> Hundred lines of commit log don't worth a single test.
> I can also add: I don't trust a commit log, I trust tests.

Yup, tests definitely will be the source of truth for the code behaviour. However, as a reviewer, I can spend time reverse engineering the intended behaviour from the test code, I might misunderstand it, or the test may not be doing what you wanted it to do, I'll have no way of knowing. And the next person interested in knowing how this code should behave will have to do exactly the same thing. A descriptive commit log will save time overall, and will make the reviews faster and better..

gbraad:
> I can also add: I don't trust a commit log, I trust tests.

If tests are defined in a readable way, with clear intentions in code, there is no commit log that would make that clearer.

I also trust tests more; working code.

cfergeau:
Not sure why the discussion seemingly turned into "it is either tests or commit logs, cannot have both" :)

pkg/crc/machine/sync.go
```diff
@@ -0,0 +1,119 @@
...
+	if !s.startLock.TryAcquire(1) {
+		return nil, errors.New("cluster is starting")
+	}
+	defer s.startLock.Release(1)
+	ctx, s.startCancel = context.WithCancel(ctx)
+	return s.underlying.Start(ctx, startConfig)
+}
+
+func (s *Synchronized) Stop() (state.State, error) {
+	cleanup, err := s.lockForStopOrDelete()
```
gbraad, line 61: I am curious if we shouldn't be more explicit that Delete can only happen AFTER a Stop has been performed. It seems here it is defined as either.

gbraad (changes requested):
/lgtm

Just a question around the `delete` command.

pkg/crc/machine/sync.go
guillaumerose: Very good question. Today, the CLI doesn't prevent you to delete right after start. 
Delete is a destructive action so, I think, there is no need to wait for the VM to be correctly stopped to delete it.

In the other way, there is no point stopping a VM that is in the process of being deleted. 

`crc delete` is perhaps the wrong word. I see it as a lightweight `crc cleanup`, more like `crc reset`

gbraad: I believe it also do this sometimes, but yeah... I do not think this should prevent the PR.
It is more a detail.

gbraad (approved):

cfergeau:
> Since you introduce a new struct, have you considered using a regular mutex to protect access to the struct, and either some booleans, or some state enum, `startInProgress bool` `stopInProgress bool` `deleteInProgress bool`? This would be less scary than these locks, trylock/,....

I've experimented with this, replacing the semaphores/trylock/... with a mutex, a channel and a state enum.
https://github.com/cfergeau/crc/tree/sync
https://github.com/cfergeau/crc/blob/sync/pkg/crc/machine/sync.go

The tests added in this PR pass after this change. It's more code than in the PR, and unfortunately there's still some magic going on (the async state updates through a go channel). I haven't carefully re-read the code looking for races. It's probably more idiomatic go code than using multiple semaphores+trylock. The 'state' this introduce could be useful as a high level state to be shown in the tray.

pkg/crc/machine/sync.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return s.underlying.Delete()
+}
+
+func (s *Synchronized) Start(ctx context.Context, startConfig StartConfig) (*StartResult, error) {
+	if !s.startLock.TryAcquire(1) {
+		return nil, errors.New("cluster is starting")
+	}
+	defer s.startLock.Release(1)
+	ctx, s.startCancel = context.WithCancel(ctx)
```
cfergeau, line 56: `s.startCancel` is never reset to nil, not sure if this matters or not.

pkg/crc/machine/sync.go
```diff
@@ -0,0 +1,119 @@
...
+		return func() {
+
+		}, errors.New("cluster is stopping or deleting")
+	}
+
+	if s.startCancel != nil {
+		logging.Infof("Cancelling virtual machine start...")
+		s.startCancel()
+	}
+
```
cfergeau, line 80: Fwiw, I think this has a slight race, if the `Start()` operation completes at about the same time as when we call `s.startCancel()`, another `Start()` might be started right away before this method acquires the `startLock` a few lines below. Probably not going to matter much though.

pkg/crc/machine/sync.go
```diff
@@ -0,0 +1,119 @@
...
+	return s.underlying.Stop()
+}
+
+func (s *Synchronized) lockForStopOrDelete() (func(), error) {
+	if !s.stopDeleteLock.TryAcquire(1) {
+		return func() {
+
+		}, errors.New("cluster is stopping or deleting")
+	}
+
```
cfergeau, line 75: a `func (s *Synchronized) cancelStart(time.Duration)` helper would probably help with code readability?

pkg/crc/machine/sync.go
guillaumerose: Yes added a missing `defer`. Also, cancelFunc can be called multiple times, so no issues.

pkg/crc/machine/sync.go
guillaumerose: done

pkg/crc/machine/sync.go
guillaumerose: Yes, I don't see how to fix it correctly. Perhaps run `startLock.Acquire` first and in parallel run `cancelStart`.

cfergeau: I think its' correct in https://github.com/cfergeau/crc/blob/sync/pkg/crc/machine/sync.go :)

guillaumerose: Ok I took your commit. I converted the enum to string as it is easier to read/debug.

pkg/crc/machine/sync.go
```diff
@@ -12,15 +12,13 @@ import (
 
 const startCancelTimeout = 15 * time.Second
 
-type State int
+type State string
```
cfergeau, line 15: With my old C programmer hat on, I prefer `type State int` + `String()` method rather than `type State string`. Should not matter much here though.

cfergeau:
> guillaumerose requested a review from cfergeau 2 hours ago

I'm not sure how to move forward here. I don't think the initial commits are that useful as most of their code is removed by `Rework sync code`. I can split that `Rework` commit along the same lines as your initial PR if needed. 
Since a good chunk of the final code was written by me, I don't think that's right that I give this a review :)

guillaumerose:
I squashed most commits. 
PTAL @code-ready/crc-team

pkg/crc/machine/sync.go
jsliacan: > Delete can only happen AFTER a Stop has been performed

@gbraad, I was asking about this some time very early on and I'm pretty sure I was told that both `start` -> `stop` -> `delete` and `start` -> `delete` were valid workflows. As @guillaumerose says, why do we care if VM is destroyed in a nice state or not?  It would seem to me that `stop` only makes sense if the user intends to restart that same VM later. Otherwise `start` -> `delete` seems like a more suitable sequence. Basically just asking out of curiosity...

pkg/crc/machine/sync.go
gbraad: can you explain more about the reasoning of this preference? The emphasize is otherwise to strongly on 'I prefer' instead of what is better from the maintenance perspective.

pkg/crc/machine/sync.go
```diff
@@ -0,0 +1,177 @@
...
+
+	switch s.currentStateUnlocked() {
+	case Starting:
+		if err := s.cancelUnlocked(startCancelTimeout); err != nil {
+			return err
+		}
+	case Idle:
+		break
+	default:
+		return errors.New("cluster is stopping or deleting")
```
gbraad, line 133: why not being explicit where with `Stopping` and `Deleting` and use Default as a fault condition ?

pkg/crc/machine/sync.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		if st == Starting {
+			s.startCancel = nil
+		}
+	default:
+	}
+	return s.currentState
+}
+
+func (s *Synchronized) IsProgressing() bool {
```
gbraad, line 63: so, `IsProgressing` indicates the busy state.
However nowhere in the implementation this is clear as this is not check or enforced using this function.

so... who uses this?

gbraad:
Do we want debug logging in the synchronization ?

pkg/crc/machine/sync.go
guillaumerose: done

guillaumerose:
> Do we want debug logging in the synchronization ?

Let's add logs if we encounter bug or someone complains. Logs are already pretty big.

gbraad:
> Let's add logs if we encounter bug or someone complains. Logs are already pretty big.

Agree.... sorry for the delay, but have been busy today.

/lgtm

... but merged anyway ;-)

Issue #2004, opened 2021-02-16
[BUG] (Windows) Access is denied on bundle extraction during setup; rename from tmp to final location


ssaranathan:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V 
  * Did you run `crc setup` before starting it : Yes
  * Running CRC on: Laptop

## CRC version
```bash
# Put `crc version` output here
CodeReady Containers version: 1.22.0+6faff76f
OpenShift version: 4.6.15 (embedded in executable)
```
  
## CRC status
```bash
# Put `crc status --log-level debug` output here

DEBU CodeReady Containers version: 1.22.0+6faff76f
DEBU OpenShift version: 4.6.15 (embedded in executable)
DEBU Running 'crc status'
DEBU Checking file: C:\Users\xxx\.crc\machines\crc\.crc-exist
Machine does not exist. Use 'crc start' to create it
```

## CRC config
```bash
# Put `crc config view` output here 
consent-telemetry                     : yes
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
# put the output of `sw_vers` in case of Mac
# Put the output of `systeminfo` in case of Windows

Host Name:                 LAPTOP-P3IVLEKN
OS Name:                   Microsoft Windows 10 Enterprise
OS Version:                10.0.19041 N/A Build 19041
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Standalone Workstation
OS Build Type:             Multiprocessor Free
Registered Owner:          Windows User
Registered Organization:
Product ID:                00330-80000-00000-AA885
Original Install Date:     9/30/2020, 10:03:41 AM
System Boot Time:          2/13/2021, 7:05:04 PM
System Manufacturer:       LENOVO
System Model:              20EQS3B400
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 94 Stepping 3 GenuineIntel ~2712 Mhz
BIOS Version:              LENOVO N1EET62W (1.35 ), 11/10/2016
Windows Directory:         C:\WINDOWS
System Directory:          C:\WINDOWS\system32
Boot Device:               \Device\HarddiskVolume1
System Locale:             en-us;English (United States)
Input Locale:              en-us;English (United States)
Time Zone:                 (UTC-05:00) Eastern Time (US & Canada)
Total Physical Memory:     32,602 MB
Available Physical Memory: 18,472 MB
[...]

PR #2102, opened 2021-03-15
#2004 Extract then rename can be too fast for Windows antivirus

guillaumerose:
Add a retry loop as a temporary fix.

In the future we should this change to:
* read the first bytes of the bundle and get the manifest,
* find the bundle name and check if this bundle already exists,
* extract the bundle directly in the right directory, skipping the
directory structure of the archive.

Related to #2004

### DIFF
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -8,9 +8,11 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/Masterminds/semver"
 	"github.com/code-ready/crc/pkg/crc/constants"
+	crcerrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/extract"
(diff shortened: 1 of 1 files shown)


cfergeau (commented):
Were you able to reproduce? Not really clear why being too fast is an issue?

gbraad (approved):
5 seconds is quite a lot between a retry. But let's see first. This is merely a workaround to begin with. You bad no comment on coffee to explain why

gbraad:
> Were you able to reproduce? Not really clear why being too fast is an issue?

Not myself as I do not really on 3rd party tools. But this kinda a known thing.

guillaumerose:
I asked users to give it a try. Let's wait a bit and see if it works.

cfergeau:
> . But this kinda a known thing

I don't know about it, so this change just looks like magic to me :) What are the antivirus expectations wrt renames/how is slowing things down helping?

guillaumerose:
@juhanevalainen told us it is working for him. Let's ship this in the next release.

PR #2103, opened 2021-03-15
Bump gvisor-tap-vsock vendoring: remove debug log

guillaumerose:


guillaumerose:
Nothing special. Moving on.

PR #2105, opened 2021-03-15
Improve launchd preflight checks and fix hyperkit startup crash

cfergeau:
This makes error reporting a bit better, and makes sure plist files are recreated when their content changed.
For testing, the main change is that the path in `~/Library/LaunchAgents/crc.daemon.plist` is updated when the absolute path of crc changes.
You can either check its content after running `crc setup` with the crc binary being in different directories, or just check `ps aux |grep 'crc daemon'` after running `crc setup`

### DIFF
--- a/pkg/crc/preflight/preflight_checks_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_tray_darwin.go
@@ -21,33 +21,35 @@ const (
 	trayAgentLabel   = "crc.tray"
 )
 
-var (
-	stdOutFilePathDaemon = filepath.Join(constants.CrcBaseDir, ".crcd-agent.log")
-	stdOutFilePathTray   = filepath.Join(constants.CrcBaseDir, ".crct-agent.log")
-)
-
 type TrayVersion struct {
 	ShortVersion string `plist:"CFBundleShortVersionString"`
 }
 
 func checkIfDaemonPlistFileExists() error {
-	// crc setup can be ran from any location in the
-	// users computer, and we need to update the plist
-	// file with the path of the crc executable which was
-	// used to run setup, to force it this check needs
-	// to always fail so the fix routine is triggered
+	// we have no way of detecting if a running crc daemon instance uses
+	// the same binary as the one running 'crc setup'
+	// Safer for now to keep the previous behaviour of always
+	// recreating the plist/restarting the daemon
+	// When we have a way of checking the version of the running
+	// daemon instance, we can recreate the plist less often.
 
-	return fmt.Errorf("Ignoring this check and triggering creation of daemon plist")
+	return fmt.Errorf("Ignoring this check and triggering re-creation of daemon plist")
+
+	/*
+		daemonConfig, err := getDaemonConfig()
+		if err != nil {
+			return err
+		}
+		return launchd.CheckPlist(*daemonConfig)
+	*/
 }
 
-func fixDaemonPlistFileExists() error {
-	// Try to remove the daemon agent from launchd
-	// and recreate its plist
-	_ = launchd.Remove(daemonAgentLabel)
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_tray_darwin.go
```diff
@@ -21,61 +21,78 @@ const (
...
 type TrayVersion struct {
 	ShortVersion string `plist:"CFBundleShortVersionString"`
 }
 
 func checkIfDaemonPlistFileExists() error {
-	// crc setup can be ran from any location in the
-	// users computer, and we need to update the plist
-	// file with the path of the crc executable which was
-	// used to run setup, to force it this check needs
-	// to always fail so the fix routine is triggered
```
cfergeau, line 38: Not sure we can remove this easily, as we also want to restart `crc daemon` when the binary changed, which will no longer be the case. Before this change, we restart the daemon too often, after this change, we won't restart it often enough.

anjannath: It should be okay, i think? as its checking the content of the plist file now and it has the path to the binary, if the crc binary changed, the generated plist content won't match with the existing one

anjannath (approved):

pkg/crc/preflight/preflight_checks_tray_darwin.go
cfergeau: When I was testing from a git tree, the path was always `$HOME/crc.git/out/macos-amd64/crc`, and the binary changed each time I rebuilt. When using the installer, I'm not sure the path we install to is versioned? And people using crc from a tarball may copy the binary to `/usr/local/bin` or similar place before running it?

anjannath: We can include custom field in the plist, if we include the version of crc in it, this could be solved, at least for different releases, i did a manual test and `launchctl` didn't complain when loading/unloading

```xml
<?xml version='1.0' encoding='UTF-8'?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version='1.0'>
 <dict>
            <key>Version</key>
            <string>1.21.1+cf45646bc</string>
            <key>Label</key>
            <string>crc.daemon</string>
[...]
``` 
There's docs about the Info.plist, but not exactly for launchctl plist files.. https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/AboutInformationPropertyListFiles.html

cfergeau: launchctl plist files are documented in the `launchd.plist` man page, but I could not find explicit mentions of custom keys in there https://manpagez.com/man/5/launchd.plist/

cfergeau: I've been trying to force restart with https://github.com/cfergeau/crc/commit/6e21b6990e4fea4b9f913bc8b36515ef397169d1 but this seems racy, daemon takes a few seconds to start, not sure why.

cfergeau: > I've been trying to force restart with [cfergeau@6e21b69](https://github.com/cfergeau/crc/commit/6e21b6990e4fea4b9f913bc8b36515ef397169d1) but this seems racy, daemon takes a few seconds to start, not sure why.

main reason is that it's not a good idea to run `launchctl stop crc.daemon && launchctl start crc.daemon && launchctl stop crc.daemon && launchctl start crc.daemon`.

In the latest iteration, I haven't really addressed this, I've kept the current behaviour with respect to plist recreation.

praveenkumar:
CI failure is not because of this PR.

PR #2114, opened 2021-03-17
Update internal registry access doc

praveenkumar:


docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
```diff
@@ -75,7 +75,7 @@ $ oc get is
 +
 [subs="+quotes,attributes"]
 ----
-$ oc set image-lookup
+$ oc set image-lookup --all
```
cfergeau, line 78: Wouldn't it be better to restrict this to `oc set image-lookup alpine` rather than `--all`, in case the user has other image streams where short name lookups would not be desirable?

cfergeau (commented):
`~/.crc/bin/oc/oc set image-lookup --help` indeed says
```
Examples:
  # Print all of the image streams and whether they resolve local names
  oc set image-lookup
    
  # Set local name lookup on all image streams
  oc set image-lookup --all
```

what a weird UI :)

docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
praveenkumar: @cfergeau yes sure, we can just use the image name which user want to add for lookup.

robin-owen (approved):

PR #2123, opened 2021-03-18
exec: Return stdout/stderr when user not in sudoers

cfergeau:
Most RunPrivileged() error paths (for example in
preflight_checks_nonwin.go) print stdout/stderr when it returns an
error. If we return an empty stderr, the error message for a failed sudo
check would be:
`Unable to set ownership of /home/teuf/.crc/bin/admin-helper-linux to root:  user is not allowed to use the 'sudo' command:`
(note the trailing `:`).

Since runCmd() print stdout/stderr in debug, we could remove stderr
from the user-visible output. However, this close to the release, let's
go with a simpler fix/less invasive fix.

### DIFF
--- a/pkg/os/exec.go
+++ b/pkg/os/exec.go
@@ -48,8 +48,8 @@ func RunPrivileged(reason string, cmdAndArgs ...string) (string, string, error) 
 	if err != nil {
 		return "", "", errors.New("sudo executable not found")
 	}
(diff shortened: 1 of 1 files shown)


cfergeau:
https://github.com/code-ready/crc/commit/e95e583f6b7502393864f3d2c300ec37a24e9078 is going to be reverted since this breaks passwordless sudo setups, so this PR can be dropped.

PR #2129, opened 2021-03-19
2 preflight fixes

cfergeau:
Details in the associated bugs/commit logs, but this fixes a bug on Ubuntu
after adding the user to the libvirt group, and tries to fix a virsh
capabilities issues on some hardened setups.

## Testing

1. run crc setup on a fresh ubuntu install
2. an error message tells the user to logout/login again
(this is currently not happening as expected)


cfergeau:
e2e seems to be failing because of `https://packages.cloud.google.com/yum/repos/google-compute-engine-el8-x86_64-stable/repodata/repomd.xml [SSL certificate problem: certificate has expired]` :-/

pkg/crc/preflight/preflight_linux.go
```diff
@@ -43,7 +43,14 @@ func libvirtPreflightChecks(distro *linux.OsRelease) []Check {
...
 			check:            checkUserPartOfLibvirtGroup,
 			fixDescription:   "Adding user to libvirt group",
-			fix:              fixUserPartOfLibvirtGroup(distro),
+			fix:              fixUserPartOfLibvirtGroup,
+		},
+		{
+			configKeySuffix:  "check-libvirt-group-active",
+			checkDescription: "Checking if active user/process is currently part of the libvirt group",
+			check:            checkCurrentGroups(distro),
+			fixDescription:   "Current user has beed added to the libvirt group, but you need to logout, login again, and run crc setup again for the addition to become effective.",
```
praveenkumar, line 52: @cfergeau this will be displayed even the `checkCurrentGroups` return the error about `User in the currently active process is not part of the libvirt group` which makes this message a bit confusing since as part of check we are saying that user is not part of libvirt group and here it is contradict with that. May be in fixDescription message we drop `Current user has beed added to the libvirt group` ?

praveenkumar:
CI is failing with following, looks like the we might need retry because we hit ` qemu:///system` and if libvirtd is socket enabled it might be taking time to activate libvirt and failed?

```
22:34:33 level=debug msg="Checking 'virsh capabilities' for libvirtd/qemu availability"
22:34:33 level=debug msg="Running 'virsh --readonly --connect qemu:///system capabilities'"
22:34:33 level=debug msg="Command failed: exit status 1"
22:34:33 level=debug msg="stdout: "
22:34:33 level=debug msg="stderr: error: failed to connect to the hypervisor\nerror: Cannot recv data: Connection reset by peer\n"
22:34:33 level=debug msg="Failed to run 'virsh capabilities': exit status 1"
22:34:33 level=warning msg="Preflight checks failed during `crc start`, please try to run `crc setup` first in case you haven't done so yet"
22:34:33 Failed to run 'virsh capabilities': exit status 1
```

cfergeau:
> CI is failing with following, looks like the we might need retry because we hit ` qemu:///system` and if libvirtd is socket enabled it might be taking time to activate libvirt and failed?

Retrying feels like a workaround, even if it takes a few seconds to restart libvirtd, it should not close the connection like this?

cfergeau:
> > CI is failing with following, looks like the we might need retry because we hit ` qemu:///system` and if libvirtd is socket enabled it might be taking time to activate libvirt and failed?
> 
> Retrying feels like a workaround, even if it takes a few seconds to restart libvirtd, it should not close the connection like this?

Phew, CI is finally green. libvirtd is confused for a few seconds after the time jump (date -s "5 minutes"). This seems to be really related to the time jump, as during testing, if I sleep 5 minutes instead, I don't see the same issue for a few seconds. I've added some wait in the clicumber test right after changing the time to ensure libvirtd is ready.

pkg/crc/preflight/preflight_linux.go
praveenkumar: Still need different team member opinion around this @gbraad @guillaumerose wdyt?

cfergeau: Ah sorry, I was so annoyed at my fight with CI/e2e tests that I totally forgot this other comment

cfergeau: I changed it now, let me know if it's better.

Issue #1698, opened 2020-11-18
Add reserved IP subnets in docs


guillaumerose:
The current setup has some limitations: some IP are reserved for CRC use. These subnets should not collide with the host network (or resources that the user wants to reach).

@adrianriobo hit this bug:
`network_controller.go:59] Error starting OpenShift Network Controller: service IP: 172.25.0.0/16 conflicts with host network: 172.25.160.0/20`

We should find all reserved subnets and add them in the doc.

gbraad:
Right. The kubernetes cluster itself uses this rnage... and it seems the default switch can assign different ranges within 172. it is nnot clear which it might be, and if this is configurable. For nnow, a reboot might rotate this again

adrianriobo:
Yes, this was on Windows machine. I will try with restart, as it was not happening before. I will check if ranges are configurable at all, and you can decide if we should inform about the reserved subnets (and configure them to avoid the collisions)

guillaumerose:
As of today, the user cannot use this network subnets:
* 10.217.0.0/22
* 10.217.4.0/23
* 192.168.126.0/24
* 192.168.130.0/24  (linux only)
* for macOS and Windows, there is also an extra subnet generated automatically. 

See https://github.com/code-ready/crc/pull/1895

robin-owen:
> for macOS and Windows, there is also an extra subnet generated automatically.

Can we document this in more detail? Is there a range that the subnet can potentially be in which we can refer to? Also, is this automatically generated subnet related to the static subnet on Linux (`192.168.130.0/24`)?

guillaumerose:
For macOS: 192.168.64.0/24 (according to /Library/Preferences/SystemConfiguration/com.apple.vmnet.plist)
For Windows, apparently it's random. I need to double check https://docs.microsoft.com/en-us/answers/questions/48268/change-hyper-v-default-switch-ip-address-range.html

robin-owen:
Fixed by #1698.

PR #2136, opened 2021-03-23
Issue #1698 docs: Add reserved IP subnets

robin-owen:
**Fixes:** Issue #1698

### DIFF
--- a/docs/source/topics/ref_reserved-ip-subnets.adoc
+++ b/docs/source/topics/ref_reserved-ip-subnets.adoc
@@ -1,0 +1,17 @@
+[id="reserved-ip-subnets_{context}"]
+= Reserved IP subnets
+
+The {prod} OpenShift cluster reserves IP subnets for internal use which should not collide with your host network.
+Ensure that the following IP subnets are available for use:
+
+.Reserved IP subnets
+* `10.217.0.0/22`
+* `10.217.4.0/23`
+* `192.168.126.0/24`
+
+Additionally, each host hypervisor reserves an additional IP subnet depending on the host operating system.
+On {msw}, the hypervisor reserves a randomly generated IP subnet that cannot be determined ahead-of-time.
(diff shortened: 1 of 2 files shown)


docs/source/topics/ref_reserved-ip-subnets.adoc
```diff
@@ -0,0 +1,17 @@
+[id="reserved-ip-subnets_{context}"]
+= Reserved IP subnets
+
+The {prod} OpenShift cluster reserves IP subnets for internal use which should not collide with your host network.
+Ensure that the following IP subnets are available for use:
+
+.Reserved IP subnets
+* `10.217.0.0/22`
+* `10.217.4.0/23`
+* `192.168.126.0/24`
```
praveenkumar, line 10: Do we really required `192.168.126.0/24` one because this is something with internal dummy etcd IP and only `192.168.126.11` is what we use.

guillaumerose: If the user starts a pod that need to reach an IP in 192.168.126.0/24, it will not work.

guillaumerose (approved):

docs/source/topics/ref_reserved-ip-subnets.adoc
praveenkumar: @guillaumerose but this is machine CIDR so why a pod need to reach to it?

guillaumerose: If I have on my LAN a MySQL server running at 192.168.126.11, I will not be able to connect to it from my CRC cluster. 

Maybe the wording is too strong:

Hypervisor networks need to be free but,
LAN networks `10.217.0.0/22`, `10.217.4.0/23`, `192.168.126.0/24` won't be reachable from the cluster.

PR #2138, opened 2021-03-23
[Integration] Testing with embedded bundle or pkg

jsliacan:
Changes to integration tests needed to run downstream CI on release binary (embedded) and pkg form of CRC. I made use of the fact that on re-start one does not need to specify bundle again. 

Additional commit to remove a bug for macos test. 

Passed tests on downstream CI.

### DIFF
--- a/test/integration/resize_test.go
+++ b/test/integration/resize_test.go
@@ -76,7 +76,11 @@ var _ = Describe("vary VM parameters: memory cpus, disk", func() {
 		It("check VM's disk size", func() {
 			out, err := SendCommandToVM("df -h | grep sysroot")
 			Expect(err).NotTo(HaveOccurred())
-			Expect(out).Should(MatchRegexp(`.*40G[\s].*[\s]/sysroot`))
+			if runtime.GOOS == "darwin" { // darwin does not support resize
+				Expect(out).Should(MatchRegexp(`.*31G[\s].*[\s]/sysroot`))
+			} else {
+				Expect(out).Should(MatchRegexp(`.*40G[\s].*[\s]/sysroot`))
+			}
 		})
 
 		It("stop CRC", func() {
--- a/test/integration/resize_test.go
+++ b/test/integration/resize_test.go
@@ -16,7 +16,12 @@ var _ = Describe("vary VM parameters: memory cpus, disk", func() {
 		})
 
 		It("start CRC", func() {
-			Expect(RunCRCExpectSuccess("start", "-b", bundlePath, "-p", pullSecretPath)).To(ContainSubstring("Started the OpenShift cluster")) // default values: "--memory", "9216", "--cpus", "4", "disk-size", "31"
+			// default values: "--memory", "9216", "--cpus", "4", "disk-size", "31"
+			if bundlePath == "embedded" {
+				Expect(RunCRCExpectSuccess("start", "-p", pullSecretPath)).To(ContainSubstring("Started the OpenShift cluster"))
+			} else {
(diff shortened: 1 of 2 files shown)


jsliacan:
Needs work to distinguish the case when the user does not define `BUNDLE_PATH` with a standalone bundle and the case when the user installs CRC from pkg (bundle not embedded but does not need to be passed with `-b` option).

test/integration/resize_test.go
```diff
@@ -48,10 +53,10 @@ var _ = Describe("vary VM parameters: memory cpus, disk", func() {
 
 		It("start CRC", func() {
 			if runtime.GOOS == "darwin" {
-				Expect(RunCRCExpectFail("start", "-b", bundlePath, "--memory", "12000", "--cpus", "6", "--disk-size", "40")).To(ContainSubstring("Disk resizing is not supported on macOS"))
-				Expect(RunCRCExpectSuccess("start", "-b", bundlePath, "--memory", "12000", "--cpus", "6")).To(ContainSubstring("Started the OpenShift cluster"))
+				Expect(RunCRCExpectFail("start", "--memory", "12000", "--cpus", "6", "--disk-size", "40")).To(ContainSubstring("Disk resizing is not supported on macOS"))
```
praveenkumar, line 56: Does it fails or we just warn the `--disk-size` part  for mac?

jsliacan: That step seems to pass, so it must fail.

praveenkumar: @jsliacan why it must fail, right now we don't have `disk-size` implemented for mac so it should success right ?

jsliacan: Yes, so the test step succeeds, but the disk-size command "fails" (doesn't only "warn"). I thought that's what you asked.

praveenkumar: @jsliacan So following is what `RunCRCExpectFail` implemented, so what we are doing is running the crc which doesn't fails because of the warn but that `RunCRCExpectFail` return the error and we check the string containing that string right?
```
func RunCRCExpectFail(args ...string) (string, error) {
	stdout, stderr, err := NewCRCCommand(args...).ExecWithFullOutput()

	if err == nil {
		err = fmt.Errorf("Expected error but exited without error")
		return stdout, err
	}

	return stderr, nil
}
```

jsliacan: OK, thanks for clarifying. Now I understand. So to answer your original question (pretty much what you wrote!): 

- `crc start` command doesn't fail, i.e. exitcode is 0
- `--disk-size` fails to be applied and a warning is issued about it. This we check for.  

Sorry for the confusion. I thought you were saying that `--disk-size` should succeed.

jsliacan: OK, so I went to a Mac and checked what happens. 

`crc start --disk-size 40` gives the following:
```
$ crc start -p ~/Downloads/crc-pull-secret --disk-size 40
Disk resizing is not supported on macOS
$ echo $?
1
```
So to answer your question:
> so what we are doing is running the crc which doesn't fails because of the warn but that RunCRCExpectFail return the error and we check the string containing that string right?

Not really. The `crc start` fails. We then check for the message `Disk resizing is not supported on macOS` in stderr. After this, we run `crc start` without `--disk-size` option and expect a success (with 31G disk, of course).

Am I closer to answering your question now? Sorry for confusion.

praveenkumar:
```
	// https://github.com/code-ready/machine-driver-hyperkit/issues/18
	if runtime.GOOS == "darwin" && value > constants.DefaultDiskSize {
		return fmt.Errorf("Disk resizing is not supported on macOS")
	}
```

Ah this is the validation code which make sure it doesn't even start so yes my understanding was not correct regarding mac for `disk-resize` option.

PR #2140, opened 2021-03-23
telemetry: enable tracking between Red Hat products

guillaumerose:
It allows us to know if a user also uses others products like vscode or
odo.
Telemetry must be enabled on both products.

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -65,14 +65,14 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 		return nil
 	}
 
-	anonymousID, uerr := getUserIdentity(c.telemetryFilePath)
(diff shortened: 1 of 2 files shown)


pkg/crc/segment/segment.go
```diff
@@ -65,14 +65,14 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
...
-	anonymousID, uerr := getUserIdentity(c.telemetryFilePath)
+	userID, uerr := getUserIdentity(c.telemetryFilePath)
 	if uerr != nil {
 		return uerr
 	}
 
 	if err := c.segmentClient.Enqueue(analytics.Identify{
-		AnonymousId: anonymousID,
-		Traits:      addConfigTraits(c.config, traits()),
+		UserId: userID,
```
praveenkumar, line 74: 👍🏼

praveenkumar (approved):

PR #2144, opened 2021-03-24
Correctly retry on error. Without the pointer, it doesn't retry at all.

guillaumerose:
RetryAfter() is running this check:
if _, ok := err.(*RetriableError); !ok {
}

### DIFF
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -77,7 +77,7 @@ type RetriableError struct {
 	Err error
 }
 
-func (r RetriableError) Error() string {
+func (r *RetriableError) Error() string {
 	return "Temporary error: " + r.Err.Error()
 }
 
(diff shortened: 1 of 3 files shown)


cfergeau (approved):
I'd suggest adding 
```
--- a/pkg/crc/errors/multierror.go
+++ b/pkg/crc/errors/multierror.go
@@ -77,7 +77,7 @@ type RetriableError struct {
        Err error
 }
 
-func (r RetriableError) Error() string {
+func (r *RetriableError) Error() string {
        return "Temporary error: " + r.Err.Error()
 }
 ```
as a follow-up so that compilation fails when this is misused.

This finds a second occurrence of the same bug:
```
diff --git a/pkg/libmachine/host/host.go b/pkg/libmachine/host/host.go
index d197a028..da8d1b37 100644
--- a/pkg/libmachine/host/host.go
+++ b/pkg/libmachine/host/host.go
@@ -97,7 +97,7 @@ func MachineInState(d drivers.Driver, desiredState state.State) func() error {
                if currentState == desiredState {
                        return nil
                }
-               return crcerrors.RetriableError{
+               return &crcerrors.RetriableError{
                        Err: fmt.Errorf("expected machine state %s, got %s", desiredState.String(), currentState.String()),
                }
        }
```

praveenkumar:
@guillaumerose let's do https://github.com/code-ready/crc/pull/2144#pullrequestreview-619695942 along with this patch so we don't have same issue in the future.

guillaumerose:
Thanks for the suggestion @cfergeau ! PR updated :+1:

cfergeau (approved):

PR #2146, opened 2021-03-24
drivers: don't start the VM when creating it

guillaumerose:
Drivers were changed to not start the VM when calling Create().
crc has to do it now.

crc can now better control what happens at creation time.

Fixes https://github.com/code-ready/machine/issues/59

--- 

It brings back the new hyperkit go bindings.

### DIFF
--- a/pkg/drivers/hyperv/hyperv.go
+++ b/pkg/drivers/hyperv/hyperv.go
@@ -203,8 +203,7 @@ func (d *Driver) Create() error {
 		return err
 	}
 
-	log.Debugf("Starting VM...")
-	return d.Start()
+	return nil
 }
 
 func (d *Driver) chooseVirtualSwitch() (string, error) {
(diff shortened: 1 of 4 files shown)


cfergeau (approved):

pkg/drivers/hyperv/hyperv.go
```diff
@@ -204,7 +204,7 @@ func (d *Driver) Create() error {
 	}
 
 	log.Debugf("Starting VM...")
-	return d.Start()
+	return nil
```
praveenkumar, line 207: Shouldn't `log.Debugf("Starting VM...")` also removed it from here and have this where we actually call the driver start?

guillaumerose: good catch! removing this.

cfergeau (approved):

PR #2147, opened 2021-03-25
Add generate command to create custom bundle from running openshift cluster

praveenkumar:
This patch implements `crc generate` command which can be used to
regenerate the bundle from the running openshift cluster.

- As of now it is tested on linux and follow-up patch will be created for
windows and mac support.
- In linux, it is depend on `qemu-img` to merge and commit the backup
  disk image (which is part of cache) with running cluster disk image
  (which is part of machine directory)
- As part of compression, we use top level directory to create same kind
  of bundles as snc repo
- Remove pull secret from the disk and from the cluster before stop it
  and creating the disk image.
- Update the bundle metadata json info (sha256sum is not updated since
  it is not used during any verification but will add it in a follow up
  PR)
- Another dependency is having around 10GB disk space in `/tmp`
  directory because we copy all the bits there even the custom disk
  image.
- $USER home should have enough free space for `qemu-img` operations
  (like rebase and commit disk image).

```
$ crc start
[...]
<-- deploy some applications -->
$ crc generate
<-- create the custom bundle on same directory -->
$ crc start -b <custom_bundle>
```

fixes: #2026


pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Get size of newly created disk
+	diskPath := filepath.Join(customBundleDir, filepath.Base(crcBundleMetadata.GetDiskImagePath()))
+	stat, err := os.Stat(diskPath)
+	if err != nil {
+		return err
+	}
+
+	// update bundle info
+	crcBundleMetadata.Type = "custom"
```
cfergeau, line 132: not sure I would change `Type`, might be best to add a new `buildInfo.Custom = "true"` field or such?

pkg/crc/machine/generate_bundle_linux.go
```diff
@@ -0,0 +1,50 @@
...
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/machine/bundle"
+	crcos "github.com/code-ready/crc/pkg/os"
+)
+
+func copyDiskImage(destDir string, crcBundleMetadata *bundle.CrcBundleInfo) error {
+	tmpDir := filepath.Join(constants.MachineCacheDir, "image-rebase")
+	_ = os.RemoveAll(tmpDir) // clean up before using it
```
cfergeau, line 15: If I somehow create a `~/.crc/cache/image-rebase` -> `~` symlink, this will wipe my homedir...

pkg/crc/machine/generate_bundle_linux.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := os.Mkdir(tmpDir, 0700); err != nil {
+		return err
+	}
+	defer func() {
+		_ = os.RemoveAll(tmpDir) // clean up after using it
+	}()
+
+	// Copy the disk image from the machine/crc folder
+	diskName := filepath.Base(crcBundleMetadata.GetDiskImagePath())
+	diskPath := filepath.Join(constants.MachineInstanceDir, constants.DefaultName, diskName)
```
cfergeau, line 25: probably better if we can avoid hardcoding `DefaultName` here? I suspect this is poking at internals of the machine driver? Or are there other places of `crc` code which know about `~/.crc/machines/crc/crc.qcow2`? At the very least, I'd move this `filepath.Join` in a `GetVMDiskImagePath()` in one of the `machine` go modules. But maybe more can be done, I did not open this code during the review?

pkg/crc/machine/generate_bundle_linux.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := crcos.CopyFileContents(crcBundleMetadata.GetDiskImagePath(),
+		filepath.Join(tmpDir, diskName), 0644); err != nil {
+		return err
+	}
+
+	// Use qemu-img commands to rebase and commit it.
+	if _, _, err := crcos.RunWithDefaultLocale("qemu-img", "rebase", "-F", "qcow2", "-b",
+		filepath.Join(tmpDir, diskName), filepath.Join(tmpDir, fmt.Sprintf("%s_back", diskName))); err != nil {
+		return err
+	}
```
cfergeau, line 40: Can we use more explicit names in this method? I'm just lost with all these filenames :) `baseImage`, `overlayImage` and `destImage` for example, or `srcFile`, `backingFile`, `destFile`.

pkg/crc/machine/generate_bundle_linux.go
```diff
@@ -0,0 +1,50 @@
...
+	}()
+
+	// Copy the disk image from the machine/crc folder
+	diskName := filepath.Base(crcBundleMetadata.GetDiskImagePath())
+	diskPath := filepath.Join(constants.MachineInstanceDir, constants.DefaultName, diskName)
+	if err := crcos.CopyFileContents(diskPath, filepath.Join(tmpDir, fmt.Sprintf("%s_back", diskName)), 0644); err != nil {
+		return err
+	}
+
+	// Copy the original image from cache folder
```
cfergeau, line 30: I'd be interested in *why* we copy it, than something repeating what the code just below says :)
We copy the base image because this is where we are going to commit the changes which were made to the VM
We copy the overlay file because before doing the commit, we need to adjust the backing file.

I'm wondering if everything (copy/rebase/commit) could be done in one go with `qemu-img dd` or `qemu-img convert`, or maybe even some `virsh vol-*` command?

cfergeau:
Is the current crc instance still functional after `crc generate`? Ie can I do
```
$ crc start
// do stuff
$ crc generate
$ crc start
```
?

We need to be more explicit during startup regarding which bundle is being used. This can be a message saying that a custom bundle is being used, or a more neutral `logging.Infof("Using bundle %s")` which does not seem to be in our logs at the moment.

pkg/compress/compress.go
```diff
@@ -0,0 +1,70 @@
...
+	currentDir, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	if err := os.Chdir(parent); err != nil {
+		return err
+	}
+	defer func() {
+		_ = os.Chdir(currentDir)
+	}()
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau yes, other way is to solve to not add the top level directory and let the extraction code take care of it but it need change in https://github.com/code-ready/crc/blob/master/pkg/extract/extract.go#L145 the directory mode from `750` to `775`.

pkg/crc/cluster/cluster.go
```diff
@@ -296,6 +308,14 @@ func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret Pul
 	return sshRunner.CopyData([]byte(content), vmPullSecretPath, 0600)
 }
 
+func RemovePullSecretFromTheInstanceDisk(sshRunner *ssh.Runner) error {
+	logging.Info("Removing user's pull secret to instance disk...")
+	if _, _, err := sshRunner.RunPrivileged("Removing user's pull secret", "rm", "-fr", vmPullSecretPath); err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I don't mind to have single helper for remove the pull secret, I think for adding it needed 2 because of the order.

pkg/crc/machine/generate_bundle_nonlinux.go
```diff
@@ -0,0 +1,11 @@
+// +build !linux
+
+package machine
+
+import (
+	"github.com/code-ready/crc/pkg/crc/machine/bundle"
+)
+
+func copyDiskImage(dirName string, crcBundleMetadata *bundle.CrcBundleInfo) error {
+	return nil
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau And then handle it as warning?

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,152 @@
...
+	crcBundleMetadata.BuildInfo.BuildTime = time.Now().String()
+	crcBundleMetadata.Storage.DiskImages[0].Size = strconv.FormatInt(stat.Size(), 10)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau yes, both provide same result.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,152 @@
...
+	// Create a tmp directory with name of <bundle>_custom
+	tmpDir, err := ioutil.TempDir("", "crc_custom_bundle")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpDir)
+
+	// Create the custom bundle directory which is used as top level directory for tarball during compression
+	currentBundleName := strings.TrimSuffix(crcBundleMetadata.GetBundleName(), filepath.Ext(crcBundleMetadata.GetBundleName()))
+	customBundleName := fmt.Sprintf("%s_custom", currentBundleName)
```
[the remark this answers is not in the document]
praveenkumar: Yes, but may be a follow up issue and PR. Not targeted for this PR.

pkg/crc/machine/generate_bundle.go
praveenkumar: Sure, anyway we are not consuming that info so I will just remove it and then if required we can add it in follow up PR's.

pkg/crc/machine/generate_bundle_linux.go
praveenkumar: This is not the first place where I am using this logic, it is already part of our codebase https://github.com/code-ready/crc/blob/master/pkg/crc/machine/bundle/repository.go#L114-L118 . The reason I am using `~` because `image-rebase` takes a lot of space and `/tmp` might not have that much but user home directory might have.

pkg/crc/machine/generate_bundle_linux.go
praveenkumar: Agree, but as of now this `DefaultName` is used lot of places and might be a refactor needed if we want to avoid that completely or you are just thinking about this specific part of code?

pkg/crc/machine/generate_bundle_linux.go
praveenkumar: Ah yes, it is confusing, let me update the variable name to make it bit more understandable .

pkg/crc/machine/generate_bundle_linux.go
praveenkumar: > I'd be interested in _why_ we copy it, than something repeating what the code just below says :)
> We copy the base image because this is where we are going to commit the changes which were made to the VM
> We copy the overlay file because before doing the commit, we need to adjust the backing file.

Done.

> I'm wondering if everything (copy/rebase/commit) could be done in one go with `qemu-img dd` or `qemu-img convert`, or maybe even some `virsh vol-*` command?

This is same steps which we perform on snc side for adding the backing file to overlay one, not sure if it can be done with single command, need to look more around it.

praveenkumar:
> Is the current crc instance still functional after `crc generate`? Ie can I do
> 
> ```
> $ crc start
> // do stuff
> $ crc generate
> $ crc start
> ```

Yes that you can do, the current instance of crc will be there until you delete it.

> 
> We need to be more explicit during startup regarding which bundle is being used. This can be a message saying that a custom bundle is being used, or a more neutral `logging.Infof("Using bundle %s")` which does not seem to be in our logs at the moment.

We do have a warning message which says an `unsupported bundle in use` but yes we do need an neutral logging around it.

cfergeau:
> Yes that you can do, the current instance of crc will be there until you delete it.

But  we ran `RemovePullSecretFromCluster` and similar methods on this instance, so I'm not sure how usable it will be?

> We do have a warning message which says an unsupported bundle in use but yes we do need an neutral logging around it.

You add a `custom` field to the bundle metadata, would also be useful to have somethnig in the logs when it is found.

pkg/compress/compress.go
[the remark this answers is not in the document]
cfergeau: I've experimented with this, https://github.com/cfergeau/crc/commit/b067dae97667a43e252cbed965ebda691630e2cd seems to work. Dunno what the initial problematic code was? Can easily be tested with `go test` now :)

pkg/crc/machine/generate_bundle.go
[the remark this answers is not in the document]
cfergeau: But one is imo easier to read as it's more usual :)

pkg/crc/machine/generate_bundle_linux.go
cfergeau: > This is not the first place where I am using this logic, it is already part of our codebase https://github.com/code-ready/crc/blob/master/pkg/crc/machine/bundle/repository.go#L114-L118 

We can try to do better in this PR rather than adding one more use of this problematic pattern?

pkg/crc/machine/generate_bundle_linux.go
cfergeau: Thinking of this PR, and this hunk in particular since this is where I noticed it.

pkg/compress/compress.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau I tried this patch with https://play.golang.org/p/A2GXsDFWx9m and looks like this is not what we want, we need to have top level directory from src not the complete tree.

Following is what we are getting
```
$ go run compress.go 
[/tmp/testdata-extracted565694984/tmp/crc_tmp_data049295885/bar /tmp/testdata-extracted565694984/tmp/crc_tmp_data049295885/foo]
```

what we want as part of list is not have `/tmp` and just have list like following.
```
$ zstdcat crc_libvirt_4.7.1_custom.zstd | tar -t
crc_tmp_data049295885
crc_tmp_data049295885/bar
crc_tmp_data049295885/foo
```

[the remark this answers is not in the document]
praveenkumar: Also with this patch the generated bundle is not able to rename and have following error.
```
Error getting bundle metadata: Temporary error: rename /home/prkumar/.crc/cache/tmp-extract/crc_libvirt_4.7.1_custom /home/prkumar/.crc/cache/crc_libvirt_4.7.1_custom: no such file or directory
```

praveenkumar:
> But we ran RemovePullSecretFromCluster and similar methods on this instance, so I'm not sure how usable it will be?

Starting the stop cluster again add the user pull secret so it doesn't have any issue.

> You add a custom field to the bundle metadata, would also be useful to have something in the logs when it is found.

Sure, this is something we can add.

pkg/compress/compress.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau I updated the PR which try to split the `src` path and use the `filepath.Rel` to generate the header Name info, this way we don't have to change the directory and  also updated the unit tests accordingly (thanks for the tests).

guillaumerose:
Is it possible to put this code and use it in snc ?
We can put the Go definition of the bundle json and this code, then vendor it here. 

The best CI we can have for this is in snc. I am afraid that with this code here, it will either extend the CI time by a good amount of time or it will never be launched/tested.

cfergeau:
> > But we ran RemovePullSecretFromCluster and similar methods on this instance, so I'm not sure how usable it will be?
> 
> Starting the stop cluster again add the user pull secret so it doesn't have any issue.

The cluster will recover, but this means the user may be asked to enter their pull secret again in some cases. Dunno if we want to document this somewhere.

cfergeau:
> Another dependency is having around 10GB disk space in /tmp directory because we copy all the bits there even the custom disk image.

`/tmp` is `tmpfs on /tmp type tmpfs` on fedora, `/var/tmp` might be a better location, as using 10GB RAM even temporarily is quite a lot.

praveenkumar:
> > Another dependency is having around 10GB disk space in /tmp directory because we copy all the bits there even the custom disk image.
> 
> `/tmp` is `tmpfs on /tmp type tmpfs` on fedora, `/var/tmp` might be a better location, as using 10GB RAM even temporarily is quite a lot.

@cfergeau This is created using `ioutil.TempDir` not something we create manually to have platform independent support.

praveenkumar:
> Is it possible to put this code and use it in snc ?
> We can put the Go definition of the bundle json and this code, then vendor it here.
> 
> The best CI we can have for this is in snc. I am afraid that with this code here, it will either extend the CI time by a good amount of time or it will never be launched/tested.

@guillaumerose We do need to have a scenario for this on CI (can be only run downstream, if timing is issue) because this is going to be part of each platform and snc mostly can test linux use case.

pkg/crc/machine/generate_bundle_linux.go
cfergeau: Actually, RemoveAll does not follow symlink to dirs, it only removes the symlink. I tested it after seeing its code:
https://golang.org/src/os/removeall_at.go
```golang
	// Simple case: if Remove works, we're done.
	err := Remove(path)
	if err == nil || IsNotExist(err) {
		return nil
	}
```

pkg/crc/machine/generate_bundle_nonlinux.go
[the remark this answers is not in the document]
cfergeau: yeah sounds good

pkg/crc/machine/generate_bundle.go
cfergeau: I'd really like to have this so that we can detect modified bundles. A file is too easy to rename, especially if this means crc picks it up automatically when it has the correct name.

pkg/crc/machine/generate_bundle_linux.go
cfergeau: I ran some tests in the morning:

```
$ time virsh -c qemu:///system vol-clone --pool crc crc.qcow2 clone.qcow2
Vol clone.qcow2 cloned from crc.qcow2

real	0m40.244s
user	0m0.009s
sys	0m0.005s

# additional runs 'real' for qemu-img convert were about the same as virsh clone (~38s)
# even if data below is showing 49s for 'real'
```

```
$ time qemu-img convert -f qcow2 -O qcow2 ~/.crc/machines/crc/crc.qcow2 convert.qcow2

real	0m49.962s
user	0m0.525s
sys	0m9.613s
```

For comparison, copy+rebase+commit gives:
```
$ time sh -x  ./img.sh 
+ mkdir -p /tmp/test.qcow/
+ cp /home/teuf/.crc/machines/crc/crc.qcow2 /tmp/test.qcow/overlayImage.qcow2
+ cp /home/teuf/.crc/cache/crc_libvirt_4.7.2/crc.qcow2 /tmp/test.qcow/crc.qcow2
+ qemu-img rebase -F qcow2 -b /tmp/test.qcow/crc.qcow2 /tmp/test.qcow/overlayImage.qcow2
+ qemu-img commit /tmp/test.qcow/overlayImage.qcow2
Image committed.
+ cp /tmp/test.qcow/crc.qcow2 /home/teuf/snc.qcow2

real	2m29.638s
user	0m1.573s
sys	0m23.620s
```

And all 3 images are about the same size (I did not look more closely if they have the same content or not)
```
$ ls -al snc.qcow2 convert.qcow2 ~/.crc/machines/crc/clone.qcow2 
-rw-r--r--. 1 teuf teuf 17041063936 Apr  6 03:33 ~.crc/machines/crc/clone.qcow2
-rw-r--r--. 1 teuf teuf 17041063936 Apr  6 03:31 convert.qcow2
-rw-r--r--. 1 teuf teuf 17045651456 Apr  6 05:04 snc.qcow2
```

cfergeau: But this code really belongs in a new machine driver API imo :)

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+	if err := copyDiskImage(customBundleDir, crcBundleMetadata); err != nil {
+		return err
+	}
+
+	// Get size of newly created disk
+	diskPath := filepath.Join(customBundleDir, filepath.Base(crcBundleMetadata.GetDiskImagePath()))
+	stat, err := os.Stat(diskPath)
+	if err != nil {
+		return err
+	}
```
cfergeau, line 88: The disk image sha256sum is not updated.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+	err = ioutil.WriteFile(kubeConfigAdminPasswordPath, []byte(kubePassword), 0600)
+	if err != nil {
+		return fmt.Errorf("error writing to %s: %v", kubeConfigFilePath, err)
+	}
+
+	// Copy ssh keyfile
+	sshKeyPath := filepath.Join(customBundleDir, "id_ecdsa_crc")
+	generatedKey := filepath.Join(constants.MachineInstanceDir, constants.DefaultName, "id_ecdsa")
```
cfergeau, line 66: This is `constants.GetPrivateKeyPath()`

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// Remove the pull secret
+	instanceIP, err := getIP(host, client.useVSock())
+	if err != nil {
+		return crcBundleMetadata, errors.Wrap(err, "Error getting the IP")
+	}
+	sshRunner, err := crcssh.CreateRunner(instanceIP, getSSHPort(client.useVSock()), crcBundleMetadata.GetSSHKeyPath(), constants.GetPrivateKeyPath(), constants.GetRsaPrivateKeyPath())
+	if err != nil {
+		return crcBundleMetadata, errors.Wrap(err, "Error creating the ssh client")
+	}
```
cfergeau, line 139: The hoops we jump through in order to get a `sshRunner` and the bundle metada still obscure what this function is doing, which starts right after this comment.
This could probably become
```
func (client *client) GenerateBundle() error {
        crcBundleMetadata, sshRunner, err := loadVM(client)
        if err != nil {
                return err
        }
        defer sshRunner.Close()
        ocConfig := oc.UseOCWithSSH(sshRunner)
        if err := cluster.RemovePullSecretFromCluster(ocConfig, sshRunner); err != nil {
                return err
        }

        if err != nil {
                return err
        }
        if currentState != state.Stopped {
                return fmt.Errorf("VM is not stopped, current state is %s", currentState.String())
        }

        return generateBundle(crcBundleMetadata)
}
```
which only shows the useful steps for the bundle generation. If we want more details, we can go look at the other helpers.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+	defer sshRunner.Close()
+	ocConfig := oc.UseOCWithSSH(sshRunner)
+	if err := cluster.RemovePullSecretFromCluster(ocConfig, sshRunner); err != nil {
+		return crcBundleMetadata, err
```
cfergeau, line 143: `return nil, err`?

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// Stop the cluster
+	currentState, err := client.Stop()
+	if err != nil {
+		return crcBundleMetadata, err
```
cfergeau, line 149: ditto, no need to return `crcBundleMetadata`

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	if currentState != state.Stopped {
+		return crcBundleMetadata, fmt.Errorf("VM is not stopped, current state is %s", currentState.String())
```
cfergeau, line 152: `return nil, ...`

cfergeau (changes requested):
`pkg/crc/machine/generate_bundle.go` has lots of knowledge it should not have (about bundle metadata, and disk images). The creation of the disk image can be improved when windows/macos support is added, but I'd really prefer that the bundle metadata is maintained in `pkg/crc/machine/bundle`, and that `pkg/crc/machine/generate_bundle.go` only uses higher level methods provided by the `bundle` package. 
I've experimented with this in https://github.com/cfergeau/crc/tree/praveen/generate , but I did not get yet to writing a unit test for the new bundle code, so it's not working :)

cfergeau:
> I've experimented with this in https://github.com/cfergeau/crc/tree/praveen/generate , but I did not get yet to writing a unit test for the new bundle code, so it's not working :)

Did more work on this (basic unit test now) in https://github.com/cfergeau/crc/tree/praveen/generate , but this is turning into quite a bit of code, not sure how to make this simpler :-/

pkg/crc/machine/bundle/metadata.go
```diff
@@ -201,3 +201,7 @@ func (bundle *CrcBundleInfo) checkDiskImageSize() error {
 	}
 	return nil
 }
+
+func GetBundleNameWithoutExtension(bundleName string) string {
```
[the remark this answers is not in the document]
gbraad: Descriptive,  no problem

gbraad:
> has lots of knowledge it should not have (about bundle metadata, and disk images).
> The creation of the disk image can be improved when windows/macos support is added,
> but I'd really prefer that the bundle metadata is maintained in ...

Can these be refactor commits as follow up, or are they blocking further progress? I would consider this more ov a priority when windows and macOS is added.

Also, Amy comments about command pattern?

pkg/crc/machine/bundle/metadata.go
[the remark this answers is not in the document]
cfergeau: It's too descriptive, it's encoding the implementation in the function name, leading to a very long name.

cfergeau:
> > has lots of knowledge it should not have (about bundle metadata, and disk images).
> > The creation of the disk image can be improved when windows/macos support is added,
> > but I'd really prefer that the bundle metadata is maintained in ...
> 
> Can these be refactor commits as follow up, or are they blocking further progress? I would consider this more ov a priority when windows and macOS is added.

The disk image generation, as I said, this is a part of the code we will have to change again for windows/macos support, so fine with me as a followup (we'll have to remember to plan for it though, and not just go for `io.Copy` in `generate_non-linux.go`)

Regarding bundle metadata, the current PR is generating invalid metadata (checksums don't match), and is hardcoding a lot of things regarding the bundle format as it is right now. For example, if we add a `podman-remote` binary to our bundles, and forget to update this code, generation will still seem to be working fine, but crc won't be able to load the generated bundle. If we change the ssh key name (this happened in the past), the code will break in the same way if we forget to update it.

> Also, any comments about command pattern?

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+
+	logging.Infof("Compressing %s...", copier.copiedBundle.cachedPath)
+	if err := compress.Compress(copier.copiedBundle.cachedPath, fmt.Sprintf("%s%s", bundleName, constants.DefaultBundleExtension)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func sha256sum(path string) (string, error) {
+	logging.Infof("Generating sha256sum for %s...", filepath.Base(path))
```
cfergeau, line 134: Maybe just `debug`?

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+
+func NewCopier(srcBundle *CrcBundleInfo, basePath string, customBundleName string) (*Copier, error) {
+	var copier Copier
+
+	bundlePath := filepath.Join(basePath, customBundleName)
+	if err := os.Mkdir(bundlePath, 0775); err != nil {
+		return nil, err
+	}
+	if err := godeepcopy.Copy(&copier.copiedBundle, srcBundle); err != nil {
+		return nil, err
```
cfergeau, line 34: I think you wanted to add a comment explaining why we do a deep copy here?

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+	copier.copiedBundle.bundleName = customBundleName
+	copier.copiedBundle.Type = "custom"
+	copier.srcBundle = srcBundle
+	copier.copiedBundle.cachedPath = bundlePath
+
+	return &copier, nil
+}
+
+func (copier *Copier) Cleanup() error {
+	return os.RemoveAll(copier.copiedBundle.cachedPath)
```
cfergeau, line 46: This is not called?

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if err := godeepcopy.Copy(&copier.copiedBundle, srcBundle); err != nil {
+		return nil, err
+	}
+
+	copier.copiedBundle.bundleName = customBundleName
+	copier.copiedBundle.Type = "custom"
+	copier.srcBundle = srcBundle
+	copier.copiedBundle.cachedPath = bundlePath
+
```
cfergeau, line 41: You can also initialize `copier` in place:
```
copier := Copier {
    copiedBundle: CrcBundleInfo {
        bundleName: customBundleName,
        ...
    }
}
```
The indentation + the removal of `copier` on each line makes it a bit easier to read imo (just a suggestion, not a required change)

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+	return copier.copiedBundle.resolvePath(filename)
+}
+
+func (copier *Copier) CopyPrivateSSHKey(srcPath string) error {
+	sshKeyFileName := filepath.Base(copier.srcBundle.GetSSHKeyPath())
+	destPath := copier.resolvePath(sshKeyFileName)
+	if err := crcos.CopyFileContents(srcPath, destPath, 0400); err != nil {
+		return err
+	}
+	// copier.srcBundle.ClusterInfo.SSHPrivateKeyFile = sshKeyFileName
```
cfergeau, line 59: you need to update the filename in `copier.copiedBundle`

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+	return nil
+}
+
+func (copier *Copier) SetKubeAdminPassword(kubePassword string) error {
+	kubeAdminPasswordFileName := copier.srcBundle.ClusterInfo.KubeadminPasswordFile
+	path := copier.resolvePath(kubeAdminPasswordFileName)
+	if err := ioutil.WriteFile(path, []byte(kubePassword), 0600); err != nil {
+		return err
+	}
+	// copier.copiedBundle.ClusterInfo.KubeadminPasswordFile = kubeAdminPasswordFileName
```
cfergeau, line 69: I'd also uncomment this.

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (copier *Copier) CopyKubeConfig() error {
+	kubeConfigFileName := filepath.Base(copier.srcBundle.GetKubeConfigPath())
+	srcPath := copier.srcBundle.GetKubeConfigPath()
+	destPath := copier.resolvePath(kubeConfigFileName)
+	if err := crcos.CopyFileContents(srcPath, destPath, 0640); err != nil {
+		return err
+	}
+	// copier.copiedBundle.ClusterInfo.KubeConfig = kubeConfigFileName
```
cfergeau, line 80: and this one. Actually you could just hardcode some names (`"kubeconfig"`)rather than using `filepath.Base(copier.srcBundle.GetKubeConfigPath())`

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (copier *Copier) SetDiskImage(path string) error {
+	fileInfo, err := getFileInfo(path)
+	if err != nil {
+		return err
+	}
+	diskImage := DiskImage{
+		File:   *fileInfo,
+		Format: copier.srcBundle.GetDiskImageFormat(),
```
cfergeau, line 100: The format should be passed to `SetDiskImage`, you cannot assume that `path` is in the same format as the one used by the source bundle.

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+	// copier.copiedBundle.ClusterInfo.KubeConfig = kubeConfigFileName
+	return nil
+}
+
+func (copier *Copier) CopyOcExecutable() error {
+	srcPath := copier.srcBundle.GetOcPath()
+	destPath := copier.resolvePath(constants.OcExecutableName)
+	if err := crcos.CopyFileContents(srcPath, destPath, 0640); err != nil {
+		return err
+	}
```
cfergeau, line 89: You need to update the name of the `oc-executable` element(s) in `Files` to `constants.OcExecutableName` as it could be different from `copier.srcBundle.GetOcPath()`. In general, it would feel more future-proof to regenerate `Storage.Files` as we copied it from the source bundle, but we don't know what is in there.

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,166 @@
...
+	fileInfo, err := getFileInfo(path)
+	if err != nil {
+		return err
+	}
+	diskImage := DiskImage{
+		File:   *fileInfo,
+		Format: copier.srcBundle.GetDiskImageFormat(),
+	}
+
+	copier.copiedBundle.Storage.DiskImages = []DiskImage{diskImage}
```
cfergeau, line 103: You also need to update ` copier.copiedBundle.Nodes[0].DiskImage`

pkg/crc/validation/validation.go
```diff
@@ -58,18 +59,20 @@ func ValidateEnoughMemory(value int) error {
...
-		return fmt.Errorf("%s not found, please provide the path to a valid bundle using the -b option", bundle)
+		return fmt.Errorf("%s not found, please provide the path to a valid bundle using the -b option", bundlePath)
 	}
 
-	userProvidedBundle := filepath.Base(bundle)
-	if userProvidedBundle != constants.GetDefaultBundle() {
+	userProvidedBundle := filepath.Base(bundlePath)
+	customBundle := fmt.Sprintf("%s_custom", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle()))
+	if userProvidedBundle != constants.GetDefaultBundle() ||
+		strings.Contains(userProvidedBundle, customBundle) {
```
cfergeau, line 73: `strings.StartsWith`? But not sure about the conditional, I'd expect `!strings.Contains(...)` ?

cfergeau (commented):
Still some details to iron out, but this overall looks fine.
It's missing a big warning / log when loading a bundle with `custom` type.

pkg/crc/validation/validation.go
cfergeau: Forgot to mention, but would be good to only generate hte custom bundle name in a single place. Instead of `GetBundleNameWithoutExtension`, this could become `GetCustomBundleName()`. This avoids duplicating this generation in 2 very separate places (validation and generate_bundle)

cmd/crc/cmd/bundle/generate.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/code-ready/crc/pkg/crc/config"
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/machine"
+	"github.com/spf13/cobra"
+)
+
+func getGenerateCmd(config *config.Config) *cobra.Command {
+	return &cobra.Command{
+		Use:   "generate",
```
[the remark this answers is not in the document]
praveenkumar: No strong opinion using `generate` or `create`  but `create` feels like we are creating something from scratch but generate means we are using something existing.

pkg/crc/machine/bundle/copier.go
praveenkumar: @cfergeau it takes fair amount of time so `Info` is preferable than `Debug`.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,128 @@
...
+	crcBundleMetadata, err := getBundleMetadataFromDriver(host.Driver)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Error loading bundle metadata")
+	}
+
+	instanceIP, err := getIP(host, client.useVSock())
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Error getting the IP")
+	}
+	sshRunner, err := crcssh.CreateRunner(instanceIP, getSSHPort(client.useVSock()), crcBundleMetadata.GetSSHKeyPath(), constants.GetPrivateKeyPath(), constants.GetRsaPrivateKeyPath())
```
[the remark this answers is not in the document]
praveenkumar: yes it is same code we use it during the status and also in start, I think this key path should be part of ssh package itself where we create the runner and those should be default one and provide a way to add more to this default.  This might be a refactor and may be take care later.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,128 @@
...
+	tmpBaseDir, err := ioutil.TempDir("", "crc_custom_bundle")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpBaseDir)
+
+	// Create the custom bundle directory which is used as top level directory for tarball during compression
+	currentBundleName := bundle.GetBundleNameWithoutExtension(bundleMetadata.GetBundleName())
+	customBundleName := fmt.Sprintf("%s_custom", currentBundleName)
+	customBundleDir := filepath.Join(tmpBaseDir, customBundleName)
```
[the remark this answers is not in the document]
praveenkumar: This is actually `copier.copiedbundle.cachePath` not the resolve path, I created a helper function for same.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,128 @@
...
+	sshRunner, err := crcssh.CreateRunner(instanceIP, getSSHPort(client.useVSock()), crcBundleMetadata.GetSSHKeyPath(), constants.GetPrivateKeyPath(), constants.GetRsaPrivateKeyPath())
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "Error creating the ssh client")
+	}
+
+	return crcBundleMetadata, sshRunner, nil
+}
+
+func removePullSecret(sshRunner *crcssh.Runner) error {
+	defer sshRunner.Close()
```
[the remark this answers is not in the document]
praveenkumar: I moved it to `loadVM` caller function.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,128 @@
...
+	ocConfig := oc.UseOCWithSSH(sshRunner)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau `cluster.RemovePullSecretFromCluster` needs `ocConfig` and `sshRunner` both because using sshRunner we need to remove the pull secret from the VM and oc config for removing it from the openshift pull secret side.

pkg/crc/machine/bundle/copier.go
praveenkumar: Yes, added.

pkg/crc/machine/bundle/copier.go
praveenkumar: Take care of from the caller.

pkg/crc/machine/bundle/copier.go
praveenkumar: I just want to have initial copy of copier bundle metadata same as source one and then make require modification.

pkg/crc/machine/bundle/copier.go
praveenkumar: @cfergeau yes but I wanted to have discussion here why it is even required because it is the fileName which we also want same as what is in the source bundle so why need to update it because copiedBundle is anyway a deep copy of source right?

pkg/crc/machine/bundle/copier.go
praveenkumar: Same https://github.com/code-ready/crc/pull/2147#discussion_r612924000

pkg/crc/machine/bundle/copier.go
praveenkumar: same https://github.com/code-ready/crc/pull/2147#discussion_r612924000 , I don't want to hard code because it will make sure we are always consuming what we have as part of source bundle.

pkg/crc/machine/bundle/copier.go
praveenkumar: @cfergeau Don't you think if it is different than `copier.srcBundle.GetOcPath()` then we have a problem? Which also means source bundle data is different than generated one which we don't want.

pkg/crc/machine/bundle/copier.go
praveenkumar: https://github.com/code-ready/crc/pull/2147#discussion_r612924000

pkg/crc/machine/bundle/copier.go
praveenkumar: https://github.com/code-ready/crc/pull/2147#discussion_r612924000

cmd/crc/cmd/bundle/generate.go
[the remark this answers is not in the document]
cfergeau: No strong opinion either, it was a suggestion from @guillaumerose  :)

pkg/crc/machine/bundle/copier.go
cfergeau: In the current implementation of `CopyKubeConfig`, it indeed does not matter.
I'd prefer to have it anyway for consistency with the rest of the code, and since that's one less thing to update if we were to change it to `CopyKubeConfig(kubeconfigPath string)` (same as the SSH key function).

pkg/crc/machine/generate_bundle.go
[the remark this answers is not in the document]
cfergeau: But couldn't `cluster.RemovePullSecretFromCluster` create its `ocConfig`?
```
func RemovePullSecretFromCluster(sshRunner) ... {
    ocConfig := oc.UseOCWithSSH(sshRunner)
    ...
}
```

pkg/crc/machine/bundle/copier.go
cfergeau: The method is `SetDiskImage(path string)`, having a hidden assumption that `path` is in the same format as `srcBundle` can be unexpected.

pkg/crc/machine/bundle/copier.go
cfergeau: > Don't you think if it is different than copier.srcBundle.GetOcPath() then we have a problem? 

Then we can check for this, and error out if this does not match your expectations ? 
```
if path.Base(copier.srcBundle.GetOcPath()) != constants.OcExecutableName {
    return fmt.Errorf("Malformed source bundle")
}
```

pkg/crc/machine/bundle/copier.go
cfergeau: The source bundle uses `copier.srcBundle.GetSSHKeyPath()` as its ssh key filename
The destination bundle uses `srcPath` as its ssh key filename
Are both names going to be the same (now and in the future)? We can't know, so if we add a ssh key named `srcPath` in the  destination bundle, then we need to update the corresponding field.

pkg/crc/machine/generate_bundle.go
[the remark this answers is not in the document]
cfergeau: Ah ok, I did not check the content of the cluster package, makes sense. Having a cluster object containing the sshrunner/occonfig/... would probably be useful rather than passing these as args to each method.

pkg/crc/constants/constants.go
```diff
@@ -21,8 +21,6 @@ const (
 
 	CrcEnvPrefix = "CRC"
 
-	DefaultBundleExtension = ".crcbundle"
-
```
cfergeau: This should be moved to `Add GetBundleNameWithoutExtension function `

pkg/crc/machine/bundle/copier.go
cfergeau: `SSHPrivateKeyFile` is still not updated

pkg/crc/machine/bundle/copier.go
cfergeau: This is also unaddressed.

test/e2e/features/basic.feature
```diff
@@ -135,6 +135,10 @@ Feature: Basic test
         And unsetting config property "enable-cluster-monitoring" succeeds
         And unsetting config property "memory" succeeds
 
+    @linux
+    Scenario: Bundle generation check
+        Given executing "crc bundle generate" succeeds
```
cfergeau, line 140: I forgot if it's possible to add comments in .feature files? I'd mention that after running `crc bundle generate`, the pull secret has been removed from the VM, as this could have impacts on other tests if this ever gets moved around?

praveenkumar: @cfergeau Sure I will add that comment here.

pkg/crc/constants/constants.go
praveenkumar: I already removed it and have it on original location, like currently on master.

praveenkumar:
> SSHPrivateKeyFile is still not updated

@cfergeau Forgot to remove the `SSHPrivateKeyFile` code, since this is again going to same as source bundle `SSHPrivateKey` file name.

pkg/crc/constants/constants.go
cfergeau: Looking at the individual commit diffs, it's added to `pkg/crc/constants/constants.go` in `Add GetBundleNameWithoutExtension function` and then removed in the `Add generate command` commit, is this intended?

pkg/crc/machine/bundle/copier.go
```diff
@@ -61,41 +61,26 @@ func (copier *Copier) CachedPath() string {
 func (copier *Copier) CopyPrivateSSHKey(srcPath string) error {
 	sshKeyFileName := filepath.Base(copier.srcBundle.GetSSHKeyPath())
 	destPath := copier.resolvePath(sshKeyFileName)
-	if err := crcos.CopyFileContents(srcPath, destPath, 0400); err != nil {
-		return err
-	}
-	// copier.srcBundle.ClusterInfo.SSHPrivateKeyFile = sshKeyFileName
-	return nil
+	return crcos.CopyFileContents(srcPath, destPath, 0400)
```
cfergeau, line 64: Ah these are in `Add e2e test for bundle creation command` rather than in the previous commit, I missed these when I looked again at the PR.

pkg/crc/machine/bundle/copier.go
```diff
@@ -61,41 +61,26 @@ func (copier *Copier) CachedPath() string {
...
-		return err
-	}
-	// copier.copiedBundle.ClusterInfo.KubeConfig = kubeConfigFileName
-	return nil
+	return crcos.CopyFileContents(srcPath, destPath, 0640)
 }
 
 func (copier *Copier) CopyOcExecutable() error {
 	srcPath := copier.srcBundle.GetOcPath()
 	destPath := copier.resolvePath(constants.OcExecutableName)
```
cfergeau, line 82: Can you use use `filepath.Base(srcPath)` instead of `constants.OcExecutableName`?

pkg/crc/machine/bundle/copier.go
praveenkumar: we discussed this over slack channel.

pkg/crc/machine/bundle/copier.go
praveenkumar: @cfergeau fixed.

test/e2e/features/basic.feature
praveenkumar: @cfergeau Added comment.

pkg/crc/constants/constants.go
cfergeau: This is still being moved to `pkg/crc/constants/constants.go` in the first commit, and then moved back in one of the later ones.

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,155 @@
...
+	kubeConfigFileName := filepath.Base(copier.srcBundle.GetKubeConfigPath())
+	srcPath := copier.srcBundle.GetKubeConfigPath()
+	destPath := copier.resolvePath(kubeConfigFileName)
+	return crcos.CopyFileContents(srcPath, destPath, 0640)
+}
+
+func (copier *Copier) CopyOcExecutable() error {
+	srcPath := copier.srcBundle.GetOcPath()
+	destPath := copier.resolvePath(filepath.Base(srcPath))
+	return crcos.CopyFileContents(srcPath, destPath, 0640)
```
cfergeau, line 82: `0755`

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,155 @@
...
+	bundleContent, err := json.MarshalIndent(copier.copiedBundle, "", " ")
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(copier.resolvePath("crc-bundle-info.json"), bundleContent, 0600)
+	if err != nil {
+		return fmt.Errorf("error copying bundle metadata  %w", err)
+	}
+
+	logging.Infof("Compressing %s...", copier.copiedBundle.cachedPath)
```
cfergeau, line 118: `copier.copiedBundle.bundelName` would be more user-friendly than `copier.copiedBundle.cachedPath`

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,121 @@
...
+	if err := copier.CopyPrivateSSHKey(constants.GetPrivateKeyPath()); err != nil {
+		return err
+	}
+
+	if err := copier.CopyOcExecutable(); err != nil {
+		return err
+	}
+
+	// Copy disk image
+	logging.Infof("Copy the disk image to %s", customBundleDir)
```
cfergeau, line 76: Here I'd also favour using `customBundleName`, the full path can be added to `logging.Debugf`

cfergeau (commented):
Some miscellaneous things  I noticed while testing:
- after bundle generation, crc could print where the generated bundle is
- after bundle generation, it's not obvious we need to run `crc delete && crc start ...`
- if `crc_libvirt_4.7.5_custom.crcbundle` already exists in the current directory, it will be silently overwritten
- if I'm running a custom bundle, and then try to create a new one, it will be named `crc_libvirt_4.7.5_custom_custom.crcbundle` (unexpected to me, but maybe you did this on purpose)
- if there's already a ~/.crc/cache/crc_libvirt_4.7.5_custom/ directory, then trying to use a different bundle with the same name will never overwrite the cache directory
- oc in the generated bundle has the wrong permissions (no executable bit set)

Most of these items can be handled incrementally/treated as known issues.

praveenkumar:
> * after bundle generation, crc could print where the generated bundle is

Can be taken care of.

> 
> * after bundle generation, it's not obvious we need to run `crc delete && crc start ...`

Sure, at the end of bundle generate command we can have some info message if that helps.

> 
> * if `crc_libvirt_4.7.5_custom.crcbundle` already exists in the current directory, it will be silently overwritten

Yes, that is expected.

> 
> * if I'm running a custom bundle, and then try to create a new one, it will be named `crc_libvirt_4.7.5_custom_custom.crcbundle` (unexpected to me, but maybe you did this on purpose)

Yes, because we shouldn't stop user to create another custom bundle from existing custom bundle. 

> 
> * if there's already a ~/.crc/cache/crc_libvirt_4.7.5_custom/ directory, then trying to use a different bundle with the same name will never overwrite the cache directory

This is even case with release bits if I have `~/.crc/cache/crc_libvirt_4.7.5` directory then using doesn't override it or you want to differ this in case of custom bundle?

cfergeau:
> > if crc_libvirt_4.7.5_custom.crcbundle already exists in the current directory, it will be silently overwritten
> Yes, that is expected.

I don't think that's  desirable, I'd require the use of `-f` to allow overwriting a file of the same name.

> > if I'm running a custom bundle, and then try to create a new one, it will be named crc_libvirt_4.7.5_custom_custom.crcbundle (unexpected to me, but maybe you did this on purpose)
> Yes, because we shouldn't stop user to create another custom bundle from existing custom bundle.

I'd keep the same name,  `crc_libvirt_4.7.5_custom.crcbundle`, or append a number to the end `crc_libvirt_4.7.5_custom_2.crcbundle`. Appending `custom` multiple times look odd 

> This is even case with release bits if I have ~/.crc/cache/crc_libvirt_4.7.5 directory then using doesn't override it or you want to differ this in case of custom bundle?

But released bundles never change, so it's not a problem that once unpacked, we never update the unpacked content. With custom bundles, the user is in control, they can recreate the custom bundle whenever they want. And this behaviour of `~/.crc/cache` will be very unexpected, and very user unfriendly (people will spend hours not understanding why an old bundle is being used, and will file bugs/ask on slack). So this will need to be addressed somehow.

cfergeau:
By the way, this code needs to be updated for bundles generated with https://github.com/code-ready/snc/commit/9a70944f5b64071d4791a5a3dd987e2a5fd66e59

gbraad:
> > if I'm running a custom bundle, and then try to create a new one, it will be named crc_libvirt_4.7.5_custom_custom.crcbundle (unexpected to me, but maybe you did this on purpose)

> Yes, because we shouldn't stop user to create another custom bundle from existing custom bundle.

actually, expected... or should become `custom_2` ;-), but that would need more code to handle ;-)

gbraad:
> I don't think that's desirable, I'd require the use of -f to allow overwriting a file with the same name.

`-f` being a `force` or `file` specification?

cfergeau:
> > I don't think that's desirable, I'd require the use of -f to allow overwriting a file with the same name.
> 
> `-f` being a `force` or `file` specification?

`--force` as already done in a few other commands, ambiguous indeed :)

gbraad:
> but I think at least this issue with ~/.crc/cache/crc_libvirt_4.7.5_custom/ never being overwritten needs to be fixed before bundle generation is released to users.

This is a difficult one to answer. This might be a 'known limitation' for now. I am not sure if this should prevent this merge... as follow-up issues related to this might be more related to how the cache code handles this situation; the metadata would need to be checked for a possible `"creation-timestamp"` or so, but this would mean an extraction of the metadata from the bundle is needed.

pkg/crc/machine/bundle/copier.go
praveenkumar: @cfergeau bundle name have `crcbundle` extension.

pkg/crc/machine/bundle/copier.go
praveenkumar: Ah so looks like when we copy the `oc` binary file then we make sure `750` as permission mode for file but that is not the case when doing symlink.

```
		return crcos.CopyFileContents(ocInBundle, ocInBinDir, 0750)
	}
```

praveenkumar:
> These can be addressed in follow-ups to this PR, but I think at least this issue with ~/.crc/cache/crc_libvirt_4.7.5_custom/ never being overwritten needs to be fixed before bundle generation is released to users.

That means even nothing change on the `custom` bundle side we have to extract it freshly for each crc start?

praveenkumar:
> By the way, this code needs to be updated for bundles generated with [code-ready/snc@9a70944](https://github.com/code-ready/snc/commit/9a70944f5b64071d4791a5a3dd987e2a5fd66e59)

Should only happen when #2234 merges or if this PR merge before then it should be handled by #2234

cfergeau:
> > By the way, this code needs to be updated for bundles generated with [code-ready/snc@9a70944](https://github.com/code-ready/snc/commit/9a70944f5b64071d4791a5a3dd987e2a5fd66e59)
> 
> Should only happen when #2234 merges or if this PR merge before then it should be handled by #2234

I did not try it, but I think `fileList` in the generated bundle metadata will have an entry for a `podman` binary which is not present in the bundle. This will be the case even if we never merge #2234.

pkg/crc/machine/start.go
```diff
@@ -44,7 +44,7 @@ const minimumMemoryForMonitoring = 14336
 func getCrcBundleInfo(bundlePath string) (*bundle.CrcBundleInfo, error) {
 	bundleName := filepath.Base(bundlePath)
 	bundleInfo, err := bundle.Use(bundleName)
-	if err == nil {
+	if err == nil && !bundle.IsCustomBundle(bundleName) {
```
cfergeau, line 47: If this bit of code is there to always unpack custom bundles, I'd prefer to see it in a separate commit which explains why we are doing this.

cfergeau:
> > These can be addressed in follow-ups to this PR, but I think at least this issue with ~/.crc/cache/crc_libvirt_4.7.5_custom/ never being overwritten needs to be fixed before bundle generation is released to users.
> 
> That means even nothing change on the `custom` bundle side we have to extract it freshly for each crc start?

Extracting custom bundle each time the user runs `crc start` is one easy way of dealing with it, but it's suboptimal :-/
A better way of dealing with this would be to read the metadata file from the archive, and if this metadata does not match what is unpacked, then we unpack it again. But that's more work, not sure it fits in this sprint :)

pkg/crc/machine/bundle/copier.go
cfergeau: Even with `crcbundle`, in my opinion this looks better than a long path referencing a directory in /tmp which will be gone as soon as the compression is done.

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if err := copier.GenerateBundle(customBundleName); err != nil {
+		return err
+	}
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	logging.Infof("Bundle is generated in %s", filepath.Join(cwd, customBundleName))
+	logging.Infof("You need to perform 'crc delete' and 'crc start -b %s' to use this bundle", filepath.Join(cwd, customBundleName))
```
cfergeau, line 96: Regarding https://github.com/code-ready/crc/pull/2147#issuecomment-824973933
Maybe `crc delete` can remove `~/.crc/cache/crc_libvirt_custom_4.7.5/` when a custom bundle is being used? This means no need to unpack the bundle with every start.

praveenkumar:
> I did not try it, but I think fileList in the generated bundle metadata will have an entry for a podman binary which is not present in the bundle. In other words, either the bundle metadata or the bundle content is invalid. This is the case even if we never merge #2234.

@cfergeau file is present but until it is part of `verify` function it will not fails and as soon as #2234 merges then the generation will automatic start failing because of this so that's the reason I didn't added it here yet.

pkg/crc/machine/generate_bundle.go
praveenkumar: >Maybe crc delete can remove ~/.crc/cache/crc_libvirt_custom_4.7.5/ when a custom bundle is being used? This means no need to unpack the bundle with every start.

@cfergeau Even then you need to unpack it every start if delete remove the custom bundle right?

pkg/crc/machine/start.go
praveenkumar: > > > These can be addressed in follow-ups to this PR, but I think at least this issue with ~/.crc/cache/crc_libvirt_4.7.5_custom/ never being overwritten needs to be fixed before bundle generation is released to users.
> > 
> > 
> > That means even nothing change on the `custom` bundle side we have to extract it freshly for each crc start?
> 
> Extracting custom bundle each time the user runs `crc start` is one easy way of dealing with it, but it's suboptimal :-/
> A better way of dealing with this would be to read the metadata file from the archive, and if this metadata does not match what is unpacked, then we unpack it again. But that's more work, not sure it fits in this sprint :)

Yes same what I discussed with @gbraad , I don't want to spend a lot of time on this if we are not getting enough user feedback if they are using bundle generation at all.

> Maybe there are other alternative ways, looking at file creation times? If the `.crcbundle` file was created after the cached directory, then recreate the directory?

It may fail on following scenario if user is not generating the bundle but consuming a generated bundle and downloaded/fetched without preserving existing timestamps. So I would go with extraction every time until we have a proper solution to detect if bundle dir need to be overwritten.

praveenkumar: > If this bit of code is there to always unpack custom bundles, I'd prefer to see it in a separate commit which explains why we are doing this.

@cfergeau Done.

pkg/crc/machine/generate_bundle.go
cfergeau: If delete removes the custom bundle directory, then `crc start -b custom_bundle; crc stop; crc start; crc stop` will only unpack once. Next time you run `crc delete` then yes the bundle will need to be unpacked again.

gbraad:
@praveenkumar does the bundle contain the podman-remote binary? if this is not the case, the bundle is invalid.

pkg/crc/machine/bundle/copier.go
```diff
@@ -0,0 +1,160 @@
...
+		File:   *fileInfo,
+		Format: format,
+	}
+
+	copier.copiedBundle.Storage.DiskImages = []DiskImage{diskImage}
+
+	return nil
+}
+
+func (copier *Copier) GenerateBundle(bundleName string) error {
```
cfergeau, line 105: You could call `CopyFilesFromFileList()` unconditionally here instead of relying on the user of `Copier` to do it? I realize the same could be said for `CopyKubeConfig()`. But maybe better to keep things consistent and leave both out of the `GenerateBundle` ?

pkg/crc/cluster/cluster.go
```diff
@@ -126,6 +126,22 @@ func EnsurePullSecretPresentInTheCluster(ocConfig oc.Config, pullSec PullSecretL
...
+		return err
+	}
+	cmdArgs := []string{"patch", "secret", "pull-secret", "-p",
+		`'{"data":{".dockerconfigjson":"e30K"}}'`,
+		"-n", "openshift-config", "--type", "merge"}
+
+	_, stderr, err := ocConfig.RunOcCommand(cmdArgs...)
+	if err != nil {
+		return fmt.Errorf("Failed to remove Pull secret %w: %s", err, stderr)
+	}
```
cfergeau, line 141: This could also do https://github.com/code-ready/snc/commit/14087055f22237b91996298275baff0e8bcbe47e

pkg/crc/validation/validation.go
```diff
@@ -58,18 +59,23 @@ func ValidateEnoughMemory(value int) error {
...
 		if constants.BundleEmbedded() {
 			return fmt.Errorf("Run 'crc setup' to unpack the bundle to disk")
 		}
-		return fmt.Errorf("%s not found, please provide the path to a valid bundle using the -b option", bundle)
+		return fmt.Errorf("%s not found, please provide the path to a valid bundle using the -b option", bundlePath)
 	}
 
-	userProvidedBundle := filepath.Base(bundle)
+	userProvidedBundle := filepath.Base(bundlePath)
+	isUsingCustomBundle := strings.Contains(userProvidedBundle, constants.GetDefaultBundle())
```
cfergeau, line 71: `strings.HasPrefix`?

pkg/crc/validation/validation.go
```diff
@@ -58,18 +59,23 @@ func ValidateEnoughMemory(value int) error {
...
 	if userProvidedBundle != constants.GetDefaultBundle() {
 		if !constants.IsRelease() {
-			logging.Warnf("Using unsupported bundle %s", userProvidedBundle)
+			logging.Warnf("Using unsupported bundlePath %s", userProvidedBundle)
```
cfergeau, line 74: `bundlePath` -> `bundle`

pkg/crc/validation/validation.go
```diff
@@ -58,18 +59,23 @@ func ValidateEnoughMemory(value int) error {
...
+			return nil
+		}
```
cfergeau, line 76: You can move  `isUsingCustomBundle := strings.Contains(userProvidedBundle, constants.GetDefaultBundle())` here as it's not used anywhere else. I'd name the variable `usingCustomBundle` or `isCustomBundle`, this is easier to read.

pkg/crc/validation/validation.go
```diff
@@ -58,18 +59,23 @@ func ValidateEnoughMemory(value int) error {
...
+		if isUsingCustomBundle {
+			logging.Warnf("Using unsupported bundlePath %s", userProvidedBundle)
```
cfergeau, line 78: `Using custom bundle %s` ?

pkg/crc/machine/bundle/copier.go
cfergeau: `Compressing /tmp/crc_custom_bundle273644518/crc_libvirt_4.7.5_1619457875...` is not easy to read, and the `/tmp/crc_custom_bundle273644518` path will go away once bundle is generated, so I don't think it's really useful to print it.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -201,5 +199,9 @@ func (bundle *CrcBundleInfo) checkDiskImageSize() error {
 }
 
 func GetBundleNameWithoutExtension(bundleName string) string {
-	return strings.TrimSuffix(bundleName, constants.DefaultBundleExtension)
+	return strings.TrimSuffix(bundleName, bundleExtension)
+}
+
+func GetCustomBundleName(bundleName string) string {
+	return fmt.Sprintf("%s_%d", GetBundleNameWithoutExtension(bundleName), time.Now().Unix())
```
cfergeau, line 206: Should this append `bundleExtension`?

cfergeau (commented):
Still some (minor) comments :( 
And 1 fixup commit I added to https://github.com/cfergeau/crc/commits/praveen/generate

pkg/crc/machine/generate_bundle.go
```diff
@@ -0,0 +1,132 @@
...
+	if err := copier.CopyPrivateSSHKey(constants.GetPrivateKeyPath()); err != nil {
+		return err
+	}
+
+	if err := copier.CopyFilesFromFileList(); err != nil {
+		return err
+	}
+
+	// Copy disk image
+	logging.Infof("Copy the disk image to %s", customBundleName)
```
cfergeau, line 77: `Copying the disk image ..`

cfergeau:
```
INFO Removing user's pull secret from instance disk and from cluster secret... 
INFO Stopping the OpenShift cluster, this may take a few minutes... 
INFO Copy the disk image to crc_libvirt_4.7.5_1619459287 
INFO Generating sha256sum for crc.qcow2...        
INFO Compressing /tmp/crc_custom_bundle642695715/crc_libvirt_4.7.5_1619459287... 
INFO Bundle is generated in /home/teuf/crc/crc/crc_libvirt_4.7.5_1619459287 
INFO You need to perform 'crc delete' and 'crc start -b /home/teuf/crc/crc/crc_libvirt_4.7.5_1619459287' to use this bundle
```
The last 2 lines are missing the `.crcbundle` extension.

pkg/crc/machine/bundle/copier.go
praveenkumar: yes I think it is better to keep those out of `GenerateBundle` so we have bit consistency in the flow (like copies all the required resource before call this function).

pkg/crc/machine/bundle/copier.go
praveenkumar: I tried to put `Compressing crc_libvirt_4.7.5_1619457875 ...` which doesnt' have long path reference.

pkg/crc/validation/validation.go
praveenkumar: Sure.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -218,3 +218,7 @@ func (bundle *CrcBundleInfo) checkDiskImageSize() error {
 func GetBundleNameWithoutExtension(bundleName string) string {
 	return strings.TrimSuffix(bundleName, bundleExtension)
 }
+
+func GetCustomBundleName(bundleName string) string {
+	return fmt.Sprintf("%s_%d", GetBundleNameWithoutExtension(bundleName), time.Now().Unix())
```
cfergeau, line 223: Nit: This should be named `GetCustomBundleNameWithoutExtension`

pkg/crc/machine/bundle/repository.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	supportedVersion = "^1.0"
-	bundleExtension  = ".crcbundle"
+	Extension        = ".crcbundle"
```
cfergeau, line 22: If `GetCustomBundleName` returns a bundle name with its extension, and with the help of `GetBundleNameWithoutExtension`, I think you can achieve the same as what you do in this commit without exporting `Extension`.

praveenkumar: @cfergeau it return name without extension, I fixed the naming of the function as mentioned https://github.com/code-ready/crc/pull/2147#discussion_r620999485 here.

cfergeau: If you don't change the naming, but return a name with the extension instead, you can reuse existing API. Here you are adding additional API which overlaps with API added in this PR.

praveenkumar: @cfergeau updated.

pkg/crc/machine/bundle/copier.go
cfergeau: This leaves a corner case as calling `CopyFilesFromFileList` is mandatory otherwise `Generate` will produce an invalid bundle, but that can be reworked later.

cfergeau (approved):

pkg/crc/validation/validation.go
```diff
@@ -58,16 +59,22 @@ func ValidateEnoughMemory(value int) error {
...
 		}
-		return fmt.Errorf("%s not found, please provide the path to a valid bundle using the -b option", bundle)
+		return fmt.Errorf("%s not found, please provide the path to a valid bundle using the -b option", bundlePath)
 	}
 
-	userProvidedBundle := filepath.Base(bundle)
+	userProvidedBundle := filepath.Base(bundlePath)
 	if userProvidedBundle != constants.GetDefaultBundle() {
+		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
+			bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle()))
```
cfergeau, line 73: I realized it would be better to append a `_` here, as we don't want `crc_libvirt_4.7.15.crcbundle` to be detected as a custom bundle for `crc_libvirt_4.7.1.crcbundle` :)

praveenkumar: @cfergeau I knew it something will be come up at the last moment in your mind so didn't merge it :P

cfergeau: I don't think it's really important to be honest, can easily be fixed afterwards :)

cfergeau (approved):

PR #2150, opened 2021-03-25
docs: Simplify haproxy configuration

cfergeau:
Instead of 'backend/frontend' blocks, we can use single 'listen' blocks
as our needs are fairly simple/basic. This also removes several
options which did not seem so useful in our case.
I've tested that this works on up to date centos7 and RHEL8

### DIFF
--- a/docs/source/topics/proc_setting-up-remote-server.adoc
+++ b/docs/source/topics/proc_setting-up-remote-server.adoc
@@ -31,7 +31,7 @@ Ensure that the cluster remains running throughout this procedure.
 . Install the [package]`haproxy` package and other utilities:
 +
 ----
-$ sudo dnf install haproxy policycoreutils-python-utils jq
+$ sudo dnf install haproxy policycoreutils-python-utils
 ----
 
 . Modify the firewall to allow communication with the cluster:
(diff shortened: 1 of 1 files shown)


docs/source/topics/proc_setting-up-remote-server.adoc
```diff
@@ -63,49 +63,28 @@ $ sudo cp /etc/haproxy/haproxy.cfg{,.bak}
 $ export CRC_IP=$({bin} ip)
 $ sudo tee /etc/haproxy/haproxy.cfg &>/dev/null <<EOF
 global
-    debug
+    log /dev/log local0
```
praveenkumar, line 66: As per https://www.haproxy.com/blog/the-four-essential-sections-of-an-haproxy-configuration/ (section log) 
```
Note that in order to read the logs, you will need to configure any of the syslog daemons, or journald, to write them to a file.
```
I am not sure if that need manual configuration or just these daemons suppose to present

cfergeau: I'm also relatively confused as to how haproxy logging works :) These days, most systems have journald configured, so should be fine.

praveenkumar: @cfergeau I checked if we add `debug` flag then logs are part of journalctl.

PR #2161, opened 2021-03-29
Ask the user to stop the application before updating

guillaumerose:
It’s better to ask the user to shutdown the application before upgrading.
    
Hyperkit or crc CLI still can be running.
    
This is important to have this when changing the path of the application from
`/Library/CodeReady Containers/1.24.0/` to `/Applications/CodeReady Containers`.
We will only have one copy of the application at the same time.

Screenshot 2021-03-29 at 09 06 19

### DIFF
--- a/packaging/darwin/Distribution.in
+++ b/packaging/darwin/Distribution.in
@@ -10,6 +10,14 @@
     <installation-check script="installCheck();"/>
     <script>
 function installCheck() {
+	var apps = system.applications.fromIdentifier('com.redhat.codeready.containers');
+	if(apps) {
+        my.result.title = 'Update failed';
+        my.result.message = 'CodeReady Containers is running. Please stop the application before updating.';
+        my.result.type = 'Fatal';
+        return false;
+	}
+
     if(!(system.compareVersions(system.version.ProductVersion, '10.14.0') >= 0)) {
         my.result.title = 'Unable to install';
         my.result.message = 'CodeReady Containers requires macOS 10.14 or later.';


cfergeau (commented):
Took me a while to get to https://developer.apple.com/documentation/installer_js/applications/1812198-fromidentifier :) 
This error will need documentation to explain the various cases when this can trigger, and what to do to solve it.
When the tray+daemon are running, won't this fail too?

guillaumerose:
> When the tray+daemon are running, won't this fail too?

This only fails when the tray is running. 
If I run crc daemon on the side, it's not failing. 

This will work better when crc daemon will be launched by the tray (soon) and crc CLI will talk to the daemon (later).

guillaumerose:
I changed upgrading to updating.

cfergeau:
Hmm I think I misunderstood what this does :) At the moment, if crc is installed and running, running the installer for a newer version will overwrite the binaries, but the running processes won't be restarted, so the upgrade will only be partial.
With this change, the installer forces the user to stop the tray before allowing the upgrade?
This sounds good, but will need to be mentioned in the documentation if not there yet
@robin-owen

cfergeau (approved):

guillaumerose:
>  At the moment, if crc is installed and running, running the installer for a newer version will overwrite the binaries, but the running processes won't be restarted, so the upgrade will only be partial.
With this change, the installer forces the user to stop the tray before allowing the upgrade?

Yes exactly!

cfergeau:
> > At the moment, if crc is installed and running, running the installer for a newer version will overwrite the binaries, but the running processes won't be restarted, so the upgrade will only be partial.
> > With this change, the installer forces the user to stop the tray before allowing the upgrade?
> 
> Yes exactly!

I thought that there was already some magic going on, and that the use of the macos installer guaranteed that whatever is needed would be restarted ^^ how naive of me!

anjannath (approved):

PR #2162, opened 2021-03-29
Daemon exits with code 2 if preflight checks failed

guillaumerose:
With this, the tray will know why the daemon couldnt start and display a nice error.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -2,13 +2,12 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
 	"time"
-
-	"github.com/code-ready/crc/pkg/crc/telemetry"
 
 	cmdConfig "github.com/code-ready/crc/cmd/crc/cmd/config"
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
@@ -19,7 +18,9 @@ import (
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/preflight"
 	"github.com/code-ready/crc/pkg/crc/segment"
+	"github.com/code-ready/crc/pkg/crc/telemetry"
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/util/exec"
 )
 
 var rootCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


cfergeau (commented):
At the moment, the tray is installed through `crc setup`. I guess the plan is to install it through the installer, in which case I'd expect the installer should also be running `crc setup`?

gbraad:
>  in which case I'd expect the installer should also be running crc setup?

yes, and there would be no tray 'checks' in the preflight to fix.

guillaumerose:
Yes this API is not meant to last long. I really hope we can move crc setup in the installer soon. 
It helps people that run the installer and directly run the app.

cfergeau:
> The same feature can also be written with a specific exit code when starting the daemon. Let me know what you prefer.

https://github.com/code-ready/crc/issues/2170 may be about the same need, but from the command line (still waiting for more details).

pkg/crc/api/api.go
```diff
@@ -117,7 +119,7 @@ func (api Server) handleConnections(conn net.Conn) {
 			conn.Close()
 		}
 
-	case "status", "version", "setconfig", "getconfig", "unsetconfig", "webconsoleurl", "logs":
+	case "status", "version", "setconfig", "getconfig", "unsetconfig", "webconsoleurl", "logs", "ready":
```
cfergeau, line 122: slightly worried at running multiple `ready` in parallel even if it should be fine.

cfergeau (commented):
The code as is looks fine to me, but I think we need to account for https://github.com/code-ready/crc/issues/2170 in its design if it's possible to deal with both needs in a similar way.

guillaumerose:
I will rewrite it with a specific exit code when setup is needed.

guillaumerose:
I rewrite it with an exit code. PTAL.

anjannath:
Tested this works as expected, LGTM

cmd/crc/cmd/root.go
```diff
@@ -90,17 +91,32 @@ func runRoot() {
...
+			os.Exit(defaultExitCode)
+		}
 	}
 	runPostrun()
 }
 
+type errorWithExitCode struct {
+	error
+	code int
+}
```
cfergeau, line 118: I have 2 conflicting suggestions here :)
- https://pkg.go.dev/k8s.io/utils/exec#CodeExitError could be reused maybe?
- or `code` could be a separate type so that it's easier to find the corresponding constants if one needs them?

cmd/crc/cmd/root.go
```diff
@@ -90,17 +91,32 @@ func runRoot() {
 	fmt.Println("No command given")
 }
 
+const (
+	defaultExitCode         = 1
```
cfergeau, line 95: This is an error code, when I read `defaultExitCode`, I assume it's going to be 0.
`generic/defaultErrorExitCode` maybe?

cfergeau (commented):
I suspect a bit of additional work will be needed for https://github.com/code-ready/crc/issues/2170 but the requirements are still not fully clear after a few days of discussion, so let's go with this for now.
Couple of questions/suggestions, but overall looks good to me.

cmd/crc/cmd/root.go
guillaumerose: done

cmd/crc/cmd/root.go
guillaumerose: I reused CodeExitError.

cfergeau (approved):

PR #2163, opened 2021-03-29
Change install path of crc to /Applications/CodeReady Containers.app

guillaumerose:
Requires https://github.com/code-ready/tray-macos/pull/91

It uses the tray as a conveyor for the full crc application.
Everything is installed in /Applications/CodeReady Containers.app.
If the user wants to uninstall the app, he just has to drag and drop the
app in the trash (after running crc cleanup).

When upgrading, all files will be replaced with newer. There is no more
one directory per version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ BUNDLE_VERSION = 4.7.4
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.24.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
-MACOS_INSTALL_PATH = /Library/CodeReady Containers
+MACOS_INSTALL_PATH = /Applications/CodeReady Containers.app/Contents/Resources/
 CONTAINER_RUNTIME ?= podman
 
 ifdef OKD_VERSION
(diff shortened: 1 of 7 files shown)


packaging/package.sh
```diff
@@ -19,17 +19,19 @@ function sign() {
   codesign --deep --sign "${CODESIGN_IDENTITY}" --options runtime --force ${opts} "$1"
 }
 
-version=$(ls "${BASEDIR}/root/Library/CodeReady Containers/")
+binDir="${BASEDIR}/root/Applications/CodeReady Containers.app/Contents/Resources"
+
+version=$("${binDir}/crc" version -o json | jq -r .version)
```
praveenkumar, line 24: We are adding `jq` as requirement which means it should be present on the RCM signing box also?

guillaumerose: Yes thanks, I removed this dependency. I added a small VERSION file in the tarball we give to RCM. The script just has to read it.

Makefile
```diff
@@ -252,20 +252,26 @@ update-go-version:
...
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/welcome.html.in >packaging/darwin/Resources/welcome.html
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/postinstall.in >packaging/darwin/scripts/postinstall
 	chmod 755 packaging/darwin/scripts/postinstall
-	mkdir -p packaging/root/"$(MACOS_INSTALL_PATH)"/$(CRC_VERSION)/
-	$(HOST_BUILD_DIR)/crc-embedder download packaging/root/"$(MACOS_INSTALL_PATH)"/$(CRC_VERSION)/
+	mkdir -p packaging/tmp/
+	$(HOST_BUILD_DIR)/crc-embedder download packaging/tmp/
+
+	mkdir -p packaging/root/Applications
+	tar -C packaging/root/Applications -xvzf $(TRAY_RELEASE)
```
praveenkumar, line 268: Doesn't `xvzf` should be only `vzf`? here we are compressing `packaging/root/Applications` to `TRAY_RELEASE` and `-x` is for extract.

guillaumerose: No, here we are uncompressing the tray in `packaging/root/Applications `. crc setup won't have to uncompress it at setup time.
The application will be directly ready in /Applications/CodeReady Containers.app right after pkg installation.

praveenkumar: Ack, I was confuse with `-C` flag and thought it is `-c` :)

Makefile
```diff
@@ -252,20 +252,26 @@ update-go-version:
...
 packagedir: LDFLAGS+= -X '$(REPOPATH)/pkg/crc/version.macosInstallPath=$(MACOS_INSTALL_PATH)' $(RELEASE_VERSION_VARIABLES)
 packagedir: clean check_bundledir $(BUILD_DIR)/macos-amd64/crc $(HOST_BUILD_DIR)/crc-embedder
+	echo -n $(CRC_VERSION) > packaging/VERSION
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/Distribution.in >packaging/darwin/Distribution
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/welcome.html.in >packaging/darwin/Resources/welcome.html
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/postinstall.in >packaging/darwin/scripts/postinstall
 	chmod 755 packaging/darwin/scripts/postinstall
-	mkdir -p packaging/root/"$(MACOS_INSTALL_PATH)"/$(CRC_VERSION)/
-	$(HOST_BUILD_DIR)/crc-embedder download packaging/root/"$(MACOS_INSTALL_PATH)"/$(CRC_VERSION)/
+	mkdir -p packaging/tmp/
```
cfergeau, line 264: `rm -rf packaging/tmp` should be added to `clean_macos_package`

Makefile
```diff
@@ -4,7 +4,7 @@ BUNDLE_VERSION = 4.7.4
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.24.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
-MACOS_INSTALL_PATH = /Library/CodeReady Containers
+MACOS_INSTALL_PATH = /Applications/CodeReady Containers.app/Contents/Resources/
```
cfergeau, line 7: This probably deserves some followup to have 
```
MACOS_INSTALL_PATH = /Applications/CodeReady Containers.app
MACOS_RESOURCES_PATH = /Applications/CodeReady Containers.app/Contents/Resources/`
```
this would make sense at least in the go codebase.

cfergeau (approved):
> When upgrading, all files will be replaced with newer. There is no more
one directory per version.

Great, I was wondering about this :)
Looks good to me, only one minor comment regarding `make clean`. I did not run it though.

PR #2165, opened 2021-03-29
Remove unnecessary second issue template

gbraad:


cfergeau (commented):
Why it's unnecessary is buried in an internal slack conversation :( Ack from me with an additional sentence in the log explaining that the 2 templates are identical, so it's a bit silly to have an extra step when creating an issue which will give you the same instructions anyway.

gbraad:
> Why it's unnecessary is buried in an internal slack conversation :(

image

PR #2172, opened 2021-03-31
Add bundle list function

guillaumerose:
I had these commits in a local branch for a long time.
This might help for a command like `crc bundle list`. 

cc @praveenkumar

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -17,6 +17,7 @@ import (
 type CrcBundleInfo struct {
 	Version     string      `json:"version"`
 	Type        string      `json:"type"`
+	Name        string      `json:"name"`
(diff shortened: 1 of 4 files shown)


cfergeau:
Not sure we want to merge this before releasing `crc bundle create`? It's really not clear how `create` and `list` should interact with the content of `~/.crc/cache`. At the moment, I think `crc bundle create` followed by `crc bundle list` would not show the bundle which was just created, which is probably not the expected result. I'd tend to land `create` in this release, and then we can figure out the exact behaviour we want to get out of `list`.

praveenkumar:
/hold

Agree with https://github.com/code-ready/crc/pull/2172#issuecomment-811053858

gbraad (approved):

guillaumerose:
reopen if needed.

PR #2177, opened 2021-04-02
Bump the tray and activate it by default

guillaumerose:
Go code only enables the tray at startup now. The installation is done by the pkg.

side-effects: 
* the macOS static crc binary doesn't install the tray anymore (but still has it in his embedded files..).
* crc setup installs the tray in the middle of other checks. tray starts directly and verifies all preflight checks are good. A race can exist between the setup and the start of the tray. I didnt face it but theorically it could happen if crc setup is too slow to end and the tray starts fast. 

I tried updating from version N to N+1 and https://github.com/code-ready/crc/pull/2161 is really needed: the tray is not stopped/restarted by N+1 crc setup, so the tray+daemon running is still the old one.


pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -1,4 +1,4 @@
-// +build !linux
+// +build windows
```
cfergeau, line 1: This can be moved to `preflight_checks_windows.go`

cfergeau (approved):

anjannath (approved):

Issue #2170, opened 2021-03-30
CodeReady Studio 'crc setup detection' algorithm broken with CRC 1.24.0


jeffmaury:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)?No
  * Running CRC on: Laptop

## CRC version
```bash
CodeReady Containers version: 1.24.0+5f06e84b
OpenShift version: 4.7.2 (embedded in executable)
```
  
## CRC status
```bash
# Put `crc status --log-level debug` output here
```

## CRC config
```bash
Machine does not exist. Use 'crc start' to create it
```

## Host Operating System
```bash

Nom de l’hôte:                              DESKTOP-U2EVGBM
Nom du système d’exploitation:              Microsoft Windows 10 Professionnel
Version du système:                         10.0.19042 N/A build 19042
Fabricant du système d’exploitation:        Microsoft Corporation
Configuration du système d’exploitation:    Station de travail autonome
Type de build du système d’exploitation:    Multiprocessor Free
Propriétaire enregistré:                    Jeff MAURY
Organisation enregistrée:                   
Identificateur de produit:                  00330-80819-37372-AA911
Date d’installation originale:              29/06/2020, 14:54:33
Heure de démarrage du système:              29/03/2021, 15:14:51
Fabricant du système:                       LENOVO
Modèle du système:                          20LAS3NJ0H
Type du système:                            x64-based PC
Processeur(s):                              1 processeur(s) installé(s).
                                            [01] : Intel64 Family 6 Model 142 Stepping 10 GenuineIntel ~1910 MHz
Version du BIOS:                            LENOVO N27ET36W (1.22 ), 04/07/2019
Répertoire Windows:                         C:\WINDOWS
Répertoire système:                         C:\WINDOWS\system32
Périphérique d’amorçage:                    \Device\HarddiskVolume1
Option régionale du système:                fr;Français (France)
Paramètres régionaux d’entrée:              fr;Français (France)
Fuseau horaire:                             (UTC+01:00) Bruxelles, Copenhague, Madrid, Paris
Mémoire physique totale:                    32 379 Mo
Mémoire physique disponible:                12 884 Mo
Mémoire virtuelle : taille maximale:        57 979 Mo
Mémoire virtuelle : disponible:             30 613 Mo
Mémoire virtuelle : en cours d’utilisation: 27 366 Mo
Emplacements des fichiers d’échange:        C:\pagefile.sys
[...]

PR #2179, opened 2021-04-02
Add --check-only to crc setup

cfergeau:
This adds a way of checking if `crc setup` has to be run before `crc
start` can be called:

$ ./crc setup --check-only
admin-helper executable is not cached

$ echo $?
2

$ ./crc setup --check-only -ojson
{
  "success": false,
  "error": "admin-helper executable is not cached"
}

$ echo $?
0

This should fix https://github.com/code-ready/crc/issues/2170

### DIFF
--- a/cmd/crc/cmd/setup_test.go
+++ b/cmd/crc/cmd/setup_test.go
@@ -14,7 +14,7 @@ func TestSetupRenderActionPlainSuccess(t *testing.T) {
 	assert.NoError(t, render(&setupResult{
 		Success: true,
 	}, out, ""))
-	assert.Equal(t, "Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster\n", out.String())
+	assert.Equal(t, "Your system is correctly setup for using CodeReady Containers, you can now run 'crc start -b $bundlename' to start the OpenShift cluster\n", out.String())
 }
 
 func TestSetupRenderActionPlainFailure(t *testing.T) {
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -10,11 +10,18 @@ import (
 	crcErrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/input"
 	"github.com/code-ready/crc/pkg/crc/preflight"
+
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/util/exec"
+)
+
+var (
+	checkOnly bool
 )
 
 func init() {
 	setupCmd.Flags().Bool(cmdConfig.ExperimentalFeatures, false, "Allow the use of experimental features")
+	setupCmd.Flags().BoolVar(&checkOnly, "check-only", false, "Only run the preflight checks, don't try to fix any misconfiguration")
 	addOutputFormatFlag(setupCmd)
 	rootCmd.AddCommand(setupCmd)
 }
@@ -47,7 +54,19 @@ func runSetup(arguments []string) error {
 			fmt.Printf("No worry, you can still enable telemetry manually with the command 'crc config set %s yes'.\n", cmdConfig.ConsentTelemetry)
 		}
 	}
-	err := preflight.SetupHost(config)
+	var err error
+	if checkOnly {
+		err = preflight.StartPreflightChecks(config)
+		if err != nil {
+			err = exec.CodeExitError{
+				Err:  err,
+				Code: preflightFailedExitCode,
+			}
+		}
+	} else {
+		err = preflight.SetupHost(config)
+	}
+
 	return render(&setupResult{
 		Success: err == nil,
 		Error:   crcErrors.ToSerializableError(err),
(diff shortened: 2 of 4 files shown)


cfergeau:
Let's make sure this fits jboss tools need before getting this in, as we don't need this option if it's not useful to them.

gbraad:
what does ` --checks-only ` only mean? There is a plural in here that just what? I believe this should be `--check-only`, but using `--only-run-checks` would be more indicative of what happens

@robin-owen Any suggestions?

cfergeau:
Could also be the opposite, something like `--no-fixes`
Just tell me what flag to use and I'll make the change.

gbraad:
the `--no-fixes` is not indicative to the user. this is more about how we implemented this.

jsliacan:
@cfergeau , so the exitcode always refers to the last option? I.e. would above example with `crc setup -ojson --checks-only` yield exitcode 2? Thanks

cfergeau:
> EDIT: So I tried it and if you add -ojson flag anywhere it returns exitcode 0 (provided all goes OK, even though crc setup --check-only would give 2). What's the rule for these things?

Yeah the return value with `-ojson` seems to always be 0, but this is already the case before this change. The json itself describes the error, so I'm not sure this is a problem.

gbraad:
Please update test description, commit description and title to reflect recent changes

gbraad (changes requested):
Commit descriptions are inconsistent with the code

cfergeau:
> Commit descriptions are inconsistent with the code

I changed the `checks-only` occurrences if this was the inconsistency you refer to.

praveenkumar:
With current state following happen if there is no issue with preflight checks. Here `Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster` output is bit confusing and doesn't make sense because we are only running the preflight checks so better not to have this message when `check-only` option in used.

```
$ ./crc setup --check-only
[...]
Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster
```

I think this should do it.

```patch
diff --git a/cmd/crc/cmd/setup.go b/cmd/crc/cmd/setup.go
index e8cf8098..1b5322ad 100644
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -82,6 +82,9 @@ func (s *setupResult) prettyPrintTo(writer io.Writer) error {
        if s.Error != nil {
                return s.Error
        }
+       if checksOnly {
+               return nil
+       }
        _, err := fmt.Fprintf(writer, "Setup is complete, you can now run 'crc start%s' to start the OpenShift cluster\n", extraArguments())
        return err
 }
```

jsliacan:
> With current state following happen if there is no issue with preflight checks. Here `Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster` output is bit confusing and doesn't make sense because we are only running the preflight checks so better not to have this message when `check-only` option in used.
> 
> ```
> $ ./crc setup --check-only
> [...]
> INFO Checking if libvirt 'crc' network is active  
> Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster
> ```
> 
> I think this should do it.
> 
> ```diff
> diff --git a/cmd/crc/cmd/setup.go b/cmd/crc/cmd/setup.go
> index e8cf8098..1b5322ad 100644
> --- a/cmd/crc/cmd/setup.go
> +++ b/cmd/crc/cmd/setup.go
> @@ -82,6 +82,9 @@ func (s *setupResult) prettyPrintTo(writer io.Writer) error {
>         if s.Error != nil {
>                 return s.Error
>         }
> +       if checksOnly {
> +               return nil
> +       }
>         _, err := fmt.Fprintf(writer, "Setup is complete, you can now run 'crc start%s' to start the OpenShift cluster\n", extraArguments())
>         return err
>  }
> ```

@praveenkumar, agree that the full message is a bit confusing. Maybe just change the message in checksOnly case? For example to "You can run 'crc start%s'..." (without the Setup is complete). The idea being that the message makes sense if all checks passed. WDYT?

cfergeau:
Maybe change the message to `Your system is correctly setup for using CodeReady Containers, you can now run 'crc start%s' to start the OpenShift cluster` which is correct in all cases (with and without `--check-only`)?

jsliacan (approved):
LGTM. Thanks!

cmd/crc/cmd/setup_test.go
```diff
@@ -14,7 +14,7 @@ func TestSetupRenderActionPlainSuccess(t *testing.T) {
 	assert.NoError(t, render(&setupResult{
 		Success: true,
 	}, out, ""))
-	assert.Equal(t, "Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster\n", out.String())
+	assert.Equal(t, "Your system is correctly setup for using CodeReady Containers, you can now run 'crc start -b $bundlename' to start the OpenShift cluster\n", out.String())
```
anjannath, line 17: 👍🏽

anjannath (approved):

Issue #2188, opened 2021-04-06
[BUG] crc cleanup error


adrianriobo:
### General information

Tested on mac and rhel

## CRC version
Pre release 1.25.0
```bash
CodeReady Containers version: 1.24.0+be6656fd
OpenShift version: 4.7.4 (not embedded in executable)
```
 
### Steps to reproduce

  1. fresh install (no ~/.crc folder) 
  2. crc cleanup 

### Expected

command should be executed without any error.

### Actual

command ends up showing an internal error and command exit code is 1

### Logs

```bash
macmini-crcqe-1:tray crcqe$ crc cleanup --log-level debug
DEBU CodeReady Containers version: 1.24.0+        
DEBU OpenShift version: 4.7.4 (not embedded in executable) 
DEBU Running 'crc cleanup'                        
INFO Unload CodeReady Containers tray             
DEBU Running 'launchctl unload /Users/crcqe/Library/LaunchAgents/crc.tray.plist' 
INFO Removing launchd configuration for tray      
INFO Removing /etc/resolver/testing file          
INFO Unload CodeReady Containers daemon           
DEBU Running 'launchctl unload /Users/crcqe/Library/LaunchAgents/crc.daemon.plist' 
INFO Stopping CRC Hyperkit process                
DEBU Running '/usr/bin/pgrep -f /Users/crcqe/.crc/bin/hyperkit' 
DEBU Command failed: exit status 1                
DEBU stdout:                                      
DEBU stderr:                                      
DEBU No running 'hyperkit' process started by crc 
INFO Removing pull secret from the keyring        
INFO Removing older logs                          
INFO Removing CRC Machine Instance directory      
DEBU Deleting machines directory                  
INFO Removing hosts file records added by CRC     
DEBU Running '/Users/crcqe/.crc/bin/admin-helper-darwin clean .crc.testing .apps-crc.testing' 
DEBU Command failed: fork/exec /Users/crcqe/.crc/bin/admin-helper-darwin: no such file or directory 
DEBU stdout:                                      
DEBU stderr:                                      
DEBU fork/exec /Users/crcqe/.crc/bin/admin-helper-darwin: no such file or directory 
fork/exec /Users/crcqe/.crc/bin/admin-helper-darwin: no such file or directory
macmini-crcqe-1:tray crcqe$ echo $?
1

```

PR #2189, opened 2021-04-06
Ignore clean host file error in case admin helper binary is not present

praveenkumar:
If a user downlond the crc and execute `crc cleanup` as first command
it will fail with following error because there is no admin-helper
binary present.
```
fork/exec Users/crcqe/.crc/bin/admin-helper-darwin: no such file or directory
fork/exec /Users/crcqe/.crc/bin/admin-helper-darwin: no such file or directory
```

This patch validate the error and if it is because of file not present
the ignore that host file cleanup.

fixes: #2188

### DIFF
--- a/pkg/crc/preflight/preflight_checks_nonwin.go
+++ b/pkg/crc/preflight/preflight_checks_nonwin.go
@@ -23,7 +23,7 @@ var nonWinPreflightChecks = [...]Check{
 	},
 	{
 		cleanupDescription: "Removing hosts file records added by CRC",
-		cleanup:            adminhelper.CleanHostsFile,
+		cleanup:            removeHostsFileEntry,
 		flags:              CleanUpOnly,
 	},
 }
@@ -66,4 +66,12 @@ func checkSuid(path string) error {
 	}
 
 	return nil
+}
+
+func removeHostsFileEntry() error {
+	err := adminhelper.CleanHostsFile()
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	return err
 }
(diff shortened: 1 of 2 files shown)


pkg/crc/adminhelper/hosts_nowin.go
```diff
@@ -3,10 +3,15 @@
 package adminhelper
 
 import (
+	"os"
+
 	crcos "github.com/code-ready/crc/pkg/os"
 )
 
 func execute(args ...string) error {
+	if _, err := os.Stat(adminHelperPath); os.IsNotExist(err) {
```
cfergeau, line 12: I don't think the `os.IsNotExist` check is useful here, if we cannot stat the helper, I doubt we'll be able to run it.

pkg/crc/preflight/preflight_checks_nonwin.go
```diff
@@ -67,3 +67,11 @@ func checkSuid(path string) error {
 
 	return nil
 }
+
+func removeHostsFileEntry() error {
+	err := adminhelper.CleanHostsFile()
+	if os.IsNotExist(err) {
```
cfergeau, line 73: This is obsolete:
https://golang.org/pkg/os/#IsNotExist
> New code should use errors.Is(err, os.ErrNotExist).

cfergeau (commented):
Minor changes to be made, but looks good to me.

cfergeau (approved):
The `Stat( )` check is gone, I assume this is intentional?

praveenkumar:
> The `Stat( )` check is gone, I assume this is intentional?

Yes, because the error type can handle that part.

Issue #2195, opened 2021-04-08
[BUG] Inconsistent cleanup logs from built artifacts for MacOS


adrianriobo:
### General information

Tested on QE Environment

## CRC version

Pre released 1.25.0
  

## Host Operating System

MacOS

### Steps to reproduce

  1. clone crc master branch
  2. make 
  3. crc cleanup
  4. make NO_CODESIGN=1 BUNDLE_DIR=./../ out/macos-amd64/crc-macos-amd64.pkg
  5. install crc from generated pkg
  6. crc cleanup

### Expected

crc cleanup command output should be the same

### Actual

Probably related to #2177, now binary does not install the tray so on cleanup messages referring to it are lost. So the output is different depending on the crc used (pkg vs binary)

### Logs

crc cleanup (from make):
```bash
redhat@blr-crc-macserver kk % ~/go/bin/crc version                                                                           
CodeReady Containers version: 1.24.0+c2886ff
OpenShift version: 4.7.5 (not embedded in executable)
redhat@blr-crc-macserver kk % ~/go/bin/crc cleanup                                                                           
INFO Removing /etc/resolver/testing file          
INFO Unload CodeReady Containers daemon           
INFO Stopping CRC Hyperkit process                
INFO Removing pull secret from the keyring        
INFO Removing older logs                          
INFO Removing CRC Machine Instance directory      
INFO Removing hosts file records added by CRC     
Cleanup finished
```

crc cleanup (from pkg build)
```bash
redhat@blr-crc-macserver kk % crc version
CodeReady Containers version: 1.24.0+c2886ff
OpenShift version: 4.7.5 (not embedded in executable)
redhat@blr-crc-macserver kk % crc cleanup
INFO Unload CodeReady Containers tray             
INFO Removing launchd configuration for tray      
INFO Removing /etc/resolver/testing file          
INFO Unload CodeReady Containers daemon           
INFO Stopping CRC Hyperkit process                
INFO Removing pull secret from the keyring        
INFO Removing older logs                          
INFO Removing CRC Machine Instance directory      
INFO Removing hosts file records added by CRC     
Cleanup finished
```

cfergeau:
Is this fixed by #2196, or is #2196 just an improvement but not a full fix? I would assume the former, and so this issue can be closed?

praveenkumar:
fixed by #2196

adrianriobo:
I think this is not fully fixed, as this is still happening, the fix catch the minimum common behavior. 

[...]

PR #2196, opened 2021-04-08
[E2E] Fix cleanup expected output on MacOS to common minimum expected information

adrianriobo:
Related to #2195

This is a patch for checking minimum expected output from cleanup command on MacOS, as described on the related issue the output differs depending on the crc distribution used on MacOS. This patch will cover the output for binary distribution and most of pkg.

Also this modification is required due to new mechanism to handle daemon process. So now this process is not handle by launchd.

Pkg clean up command will be tested from ux / installer feature and it will check beyond the output the right expected state for the host (no tray process, no hyperkit, ...) 

The [run](https://crcqe-jenkins-csb-codeready.cloud.paas.psi.redhat.com/blue/organizations/jenkins/qe%2Fbundle_baremetal_macos15-brno/detail/bundle_baremetal_macos15-brno/168/pipeline/) for this PR can be checked on our downstream CI

### DIFF
--- a/test/e2e/features/basic.feature
+++ b/test/e2e/features/basic.feature
@@ -157,13 +157,11 @@ Feature: Basic test
     @darwin
     Scenario Outline: CRC clean-up
         When executing "crc cleanup" succeeds
-        Then stderr should contain "Unload CodeReady Containers tray"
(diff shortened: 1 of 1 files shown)


anjannath:
ok to test

PR #2198, opened 2021-04-08
Fix kubeconfig update on restarts

cfergeau:
Since `Simplify kubeadmin renewal code`, running `crc start && crc stop
&& crc start` displays:

ERRO Cannot update kubeconfig: challenger chose not to retry the request

This is because on the 2nd start, we mistakenly set
`clusterConfig.KubeAdminPass` to ""  after calling
`UpdateKubeAdminUserPassword`.
This commit fixes this by adding a check that the kubeadmin password was
actually changed (its value is non-empty) before modifying its value in
`clusterConfig`


praveenkumar:
@cfergeau I think #2200 will be a better fix, do check once and we can close one of them.

anjannath:
Closed in favor of https://github.com/code-ready/crc/pull/2200

Issue #2199, opened 2021-04-09
[BUG] Stop => start fails to update the kubeconfig


praveenkumar:
```
$ crc start
$ crc stop
$ crc start
INFO Adding crc-admin and crc-developer contexts to kubeconfig... 
ERRO Cannot update kubeconfig: challenger chose not to retry the request 
```

PR #2200, opened 2021-04-09
Fix kubeconfig update on restarts

praveenkumar:
Since f974ac2f, running crc start && crc stop &&
crc start displays:

ERRO Cannot update kubeconfig: challenger chose not to retry the request

This is because on the 2nd start, we mistakenly set
clusterConfig.KubeAdminPass to "" after calling
UpdateKubeAdminUserPassword.

This patch will make sure that UpdateKubeadminUserPassword will just
do the update of password and not return the updated password instead
use 'GetKubeadminPassword' for getting the password which always
provides correct password for kubeadmin user.

fixes: #2199


cfergeau:
Looks good to me, we could add https://paste.centos.org/view/fb654679 on top of it.
https://github.com/code-ready/crc/pull/2198 has an alternative fix, I'm fine with either version.

anjannath (approved):

Issue #2202, opened 2021-04-09
[BUG] crc start stuck checking DNS on Big Sur


adrianriobo:
### General information

This error was capture on downstream environments 

## CRC version

Pre release 1.25.0
  
## Host Operating System

```bash
ProductName:	macOS
ProductVersion:	11.2.1
BuildVersion:	20D74
```
And
```bash
ProductName:	macOS
ProductVersion:	11.1
BuildVersion:	20C69
```
### Steps to reproduce

  1. crc setup 
  2. crc start

### Expected

The cluster start successfully. 

### Actual

During the crc start the log shows:

```
ERRO Host is not connected to internet            
DEBU Running SSH command: cat /etc/resolv.conf    
DEBU SSH commahttps://gist.github.com/adrianriobo/78787ee108c64db19c4c58230dfbfb14nd results: err: <nil>, output: # Generated by NetworkManager
search mw.lab.eng.rdu2.redhat.com
nameserver 192.168.64.1
nameserver fe80::10dc:b758:d0c3:4a1d%enp0s1 
DEBU Creating /etc/resolv.conf with permissions 0644 in the CRC VM 
DEBU Running SSH command: <hidden>                
DEBU SSH command succeeded                        
DEBU retry loop: attempt 0                        
DEBU Running SSH command: host -R 3 foo.apps-crc.testing 
DEBU SSH command results: err: <nil>, output: foo.apps-crc.testing has address 192.168.64.2 
INFO Check internal and public DNS query ...      
DEBU Running SSH command: host -R 3 quay.io   
```
Then it  gets stuck checking DNS resolution.

In the meantime checking the resolution directly connecting to the machine from other terminal we are success:

```bash
jbossqa@dev-platform-mac3 crc % ssh -i ~/.crc/machines/crc/id_ecdsa core@api.crc.testing host -R 3 quay.io
quay.io has address 52.72.182.67
quay.io has address 3.233.133.41
quay.io has address 54.197.99.84
quay.io has address 54.156.10.58
quay.io has address 50.16.140.223
quay.io has address 34.224.196.162
quay.io has address 34.197.63.98
quay.io has address 52.4.104.248
quay.io mail is handled by 20 alt2.aspmx.l.google.com.
quay.io mail is handled by 30 aspmx2.googlemail.com.
quay.io mail is handled by 30 aspmx3.googlemail.com.
quay.io mail is handled by 10 aspmx.l.google.com.
[...]

PR #2207, opened 2021-04-09
Rework waitForNetwork

cfergeau:
During testing, some issues were found on Big Sur, `waitForNetwork`
returns without having found network connectivity, and the subsequent
`host` call in the VM hangs.
In https://github.com/code-ready/crc/issues/2202 it seems no ping
attempts are made, so /etc/resolv.conf is empty? This means
`waitForNetwork` will return relatively quickly. Adding a small delay
fixed the observed issue.

This commit reworks waitForNetwork so that it always rereads resolv.conf
(in case it is first created empty, and then populated), and then tries
to ping the nameservers. If no nameserver was pinged, it returns an
error so that we retry, and wait a bit longer for network connectivity
to come up.

This should fix https://github.com/code-ready/crc/issues/2202

### DIFF
--- a/pkg/crc/services/dns/dns_darwin.go
+++ b/pkg/crc/services/dns/dns_darwin.go
@@ -110,31 +110,36 @@ func restartNetwork() error {
 	return nil
 }
 
+func checkNetworkConnectivity() error {
+	hostResolv, err := network.GetResolvValuesFromHost()
+	if err != nil {
+		logging.Debugf("Unable to read resolv.conf: %v", err)
+		return err
+	}
+
+	for _, ns := range hostResolv.NameServers {
+		_, _, err := crcos.RunWithDefaultLocale("ping", "-c4", ns.IPAddress)
+		if err != nil {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


gbraad (approved):

pkg/crc/services/dns/dns_darwin.go
```diff
@@ -110,31 +110,36 @@ func restartNetwork() error {
...
+// Wait for Network wait till the network is up, since it is required to resolve external dnsquery
+func waitForNetwork() error {
+	retriableConnectivityCheck := func() error {
+		err := checkNetworkConnectivity()
+		if err != nil {
+			return &crcerrors.RetriableError{Err: err}
 		}
+		return nil
+	}
+	if err := crcerrors.RetryAfter(10*time.Second, retriableConnectivityCheck, time.Second); err != nil {
```
praveenkumar, line 140: @cfergeau As per https://github.com/code-ready/crc/issues/2202#issuecomment-816643847 looks like it succeed in 9th attempt so may be better to have it `15*time.Second` so if a system is bit slow then also it works?

Issue #2190, opened 2021-04-06
[BUG] MacOS pkg installer skip step destination select


adrianriobo:
### General information

 Tested on MacOS 10.14, with self built pkg installer:

```bash
make NO_CODESIGN=1 BUNDLE_DIR=~/jenkins/bundles out/macos-amd64/crc-macos-amd64.pkg
```

## CRC version

Pre release 1.25.0

### Steps to reproduce

  1.  Double click on pkg 
  2.  Installer start 
  3.  Tab Introduction. Click Continue
  3.  Tab License. Click Continue
  4.  Pop up on License Agreement. Click Agree
  5.  Installer jumps to Tab Installation Type, skipping Destination Select tab. Click Go Back
  6.  Go back to Tab License. Click Continue
  7.  Here goes to step 5..so destination can not be selected

### Expected

Select destination folder for installation

### Actual

Occasionally the full flow is honored and the destination folder can be selected... but most of the times ends up on a wrong flow.

cfergeau:
Isn't the macOS install path hardcoded in the `crc` binary anyway?

adrianriobo:
I mean I can install it, it uses the default path, but the problem is the flow with the installer is not consistent on every run

cfergeau:
Yeah, Guillaume mentioned this inconsistency in the past, he was not sure why this is happening.

guillaumerose:
Thanks for this! 
We should disable it at least for now. The path `/Applications/CodeReady Containers.app` is set in stone in `crc` binary. Moving the app would certainly break `crc`.

PR #2209, opened 2021-04-12
pkg: remove destination select step

guillaumerose:
Application path is set at compilation time and cannot be changed. If the user
moves the application, crc won’t be able to find helper binaries, etc.

This option disables the select destination menu as per:
https://stackoverflow.com/questions/16458656/disable-change-install-location-button-in-installer-created-using-productbu

Fix https://github.com/code-ready/crc/issues/2190

### DIFF
--- a/packaging/darwin/Distribution.in
+++ b/packaging/darwin/Distribution.in
@@ -7,6 +7,7 @@
     <license file="LICENSE.txt"/>
     <options customize="never" allow-external-scripts="no"/>
     <domains enable_localSystem="true" />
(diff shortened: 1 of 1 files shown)


anjannath:
It still appears, tested the artifact from circle CI
Screenshot 2021-04-12 at 5 18 50 PM

guillaumerose:
The menu is here but it is always skipped. Is it what you see also?

anjannath:
> The menu is here but it is always skipped. Is it what you see also?

Yes, it is skipped.

anjannath (approved):

PR #2210, opened 2021-04-12
[Segment] Capture SSH_TTY env

### LINKED ISSUE #2204: Detect if CRC is running in a remote server via an ssh session
using $SSH_TTY we can detect people who ssh to linux server and start crc there

/cc @veillard

praveenkumar:
This patch captures following scenario and set true for `SSH_TTY` which
able to provide us bit detail around how many user are using crc on
remote host.
- User ssh to a remote box
- User run `crc` commands from that box

fixes #2204

### DIFF
--- a/pkg/os/util.go
+++ b/pkg/os/util.go
@@ -115,3 +115,7 @@ func RemoveFileIfExists(path string) error {
 func RunningInTerminal() bool {
 	return terminal.IsTerminal(int(os.Stdin.Fd()))
 }
+
+func RunningUsingSSH() bool {
+	return os.Getenv("SSH_TTY") != ""
+}
--- a/usage-data.adoc
+++ b/usage-data.adoc
@@ -20,6 +20,7 @@ The following table describes the opt-in usage data collected by CodeReady Conta
 |                  | Experimental features       | true/false
 |                  | tty                         | true/false
 |                  | duration                    |
+|                  | remotely connected with SSH | true/false
 
 |*Config*          | Num of CPUs                 | 
 |                  | Memory size                 |
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -85,7 +85,8 @@ func (c *Client) Upload(ctx context.Context, action string, duration time.Durati
 	properties = properties.Set("version", version.GetCRCVersion()).
 		Set("success", err == nil).
 		Set("duration", duration.Milliseconds()).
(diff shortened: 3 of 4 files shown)


jsliacan:
@praveenkumar , maybe we should also add it to the [doc](https://github.com/code-ready/crc/blob/master/usage-data.adoc)? Thanks!

pkg/os/util.go
```diff
@@ -115,3 +115,7 @@ func RemoveFileIfExists(path string) error {
 func RunningInTerminal() bool {
 	return terminal.IsTerminal(int(os.Stdin.Fd()))
 }
+
+func RunningUsingSSH() bool {
+	return os.Getenv("SSH_TTY") != ""
```
cfergeau, line 120: Or `LookupEnv`? I guess `SSH_TTY` is not going to be set but empty, but `LookupEnv` seems more correct

usage-data.adoc
```diff
@@ -20,6 +20,7 @@ The following table describes the opt-in usage data collected by CodeReady Conta
 |                  | Experimental features       | true/false
 |                  | tty                         | true/false
 |                  | duration                    |
+|                  | remote (Capture $SSH_TTY)   | true/false
```
cfergeau, line 23: It captures if it's set or not, not the value of `$SSH_TTY`. Maybe just `remotely connected with SSH`, how it's implemented does not necessarily have to be mentioned.

pkg/os/util.go
praveenkumar: @cfergeau It is going to be set
```
// On my current workstation
$ env | grep -i ssh
SSH_AUTH_SOCK=/run/user/1000/keyring/ssh
SSH_AGENT_PID=5205


// On the remote machine
$ env | grep -i ssh
SSH_CONNECTION=10.10.113.37 45514 10.0.78.252 22
SSH_CLIENT=10.10.113.37 45514 22
SSH_TTY=/dev/pts/0
```

usage-data.adoc
praveenkumar: Sure, I can change the wording.

jsliacan: Then maybe also de-mystify the `tty` above? Like @cfergeau says, let's just say what we find out with it (in plain English) -- whether CRC is used in a terminal or not?.

praveenkumar: @jsliacan not in this PR.

anjannath (approved):

cfergeau (approved):

PR #2212, opened 2021-04-12
vsock: check daemon is running in machine.Start instead of preflights

guillaumerose:
'crc daemon' needs setup to be done but also 'crc setup' requires the
daemon to be running. There is a dependency loop.

This patch removes the need of 'crc daemon' running before 'crc setup'.
'crc start' will complain before starting the VM if the daemon is not
running.

```
$ crc start -p ~/pull-secret.txt -b ~/Downloads/crc_libvirt_4.7.2.crcbundle
...
INFO Loading bundle: crc_libvirt_4.7.2.crcbundle ... 
ERRO Is 'crc daemon' running? Network mode 'vsock' requires 'crc daemon' to be running, run it manually on different terminal/tab 
Get "http://unix/network/services/forwarder/all": dial unix /home/guillaumerose/.crc/crc-http.sock: connect: connection refused

```


cfergeau (approved):
Alternatively, the preflight check could have made use of https://github.com/code-ready/crc/commit/60c286c9ecc8c2a4775afc4230c3effe7af6af7e :)

anjannath (approved):
Tested on window!

Issue #2214, opened 2021-04-12
Can't access crc from crc start on Fedora 33 - err: ssh: handshake failed


aireilly:
I can't get access to the crc VM on `crc start`.

```
time="2021-04-12T17:21:04+01:00" level=debug msg="error: Temporary error: ssh command error:\ncommand : exit 0\nerr     : ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain\\n - sleeping 1s"
time="2021-04-12T17:21:05+01:00" level=debug msg="retry loop: attempt 139"
time="2021-04-12T17:21:05+01:00" level=debug msg="Running SSH command: exit 0"
time="2021-04-12T17:21:05+01:00" level=debug msg="Using ssh private keys: [/home/aireilly/.crc/cache/crc_libvirt_4.7.2/id_ecdsa_crc]"
time="2021-04-12T17:21:05+01:00" level=debug msg="SSH command results: err: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain, output: "
``` 
I've tried: 

```
$ crc stop -f
$ crc delete -f
$ crc cleanup
$ sudo rm -rf ~/.crc
< REBOOT >
$ crc setup
$ crc start --log-level debug
```
Software version: 
CodeReady Containers version: 1.24.0+5f06e84b
OpenShift version: 4.7.2 (embedded in executable
Fedora 33 (Workstation Edition)
[crc.log](https://github.com/code-ready/crc/files/6298297/crc.log)

I'm not on VPN, and running crc on a Lenovo thinkpad t590.

What can I try?

cfergeau:
The VM has apparently failed to start, not sure why.
Can you check `virsh -c qemu:///system list` after `crc start` failed? And also look in `journalctl -u libvirtd`, and in `/var/lib/libvirt/qemu/crc.log`

aireilly:
Thanks for quick response!

```aireilly@aireilly-t590: ~ $ virsh -c qemu:///system list
 Id   Name   State
----------------------
 1    crc    running```

journalctl -u libvirtd

```
-- Reboot --
Jan 31 10:24:16 VFIEVOX3.Router systemd[1]: Starting Virtualization daemon...
Jan 31 10:24:16 VFIEVOX3.Router systemd[1]: Started Virtualization daemon.
Jan 31 10:24:17 VFIEVOX3.Router dnsmasq[2139]: started, version 2.83 cachesize 150
[...]

PR #2217, opened 2021-04-13
start: Add log with VM's IP when it started

cfergeau:
This can be useful to detect some `crc start` issues early, and is not
always present in the debug log at the moment.

This is related to https://github.com/code-ready/crc/issues/2214


pkg/crc/machine/start.go
```diff
@@ -277,6 +277,7 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 	if err != nil {
 		return nil, errors.Wrap(err, "Error getting the IP")
 	}
+	logging.Infof("CodeReady Containers instance is running with IP %s", instanceIP)
```
praveenkumar, line 280: 👍🏼

praveenkumar (approved):

PR #2234, opened 2021-04-16
Fix 'crc podman-env' command

guillaumerose:
It enables the user to use rootless podman. 

This is very early stage and tbh, it doesn't work well.
With bundle 4.7.5, podman 2.2.1, a simple 'podman ps' hangs half of the time. It also freezes the VM (cannot ssh during the podman hang). 

With podman rootless, I have this error when it hangs:
```
$ systemctl --user status podman.service
● podman.service - Podman API Service
   Loaded: loaded (/usr/lib/systemd/user/podman.service; static; vendor preset: enabled)
   Active: activating (start) since Fri 2021-04-16 20:00:56 UTC; 735ms ago
     Docs: man:podman-system-service(1)
 Main PID: 3990 (podman)
   CGroup: /user.slice/user-1000.slice/user@1000.service/podman.service
           ├─3990 /usr/bin/podman --log-level=info system service
           └─4000 /usr/bin/podman --log-level=info system service

Apr 16 20:00:56 crc-xl2km-master-0 systemd[3830]: Starting Podman API Service...
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=info msg="/usr/bin/podman filtering at log level info"
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=warning msg="Error initializing configured OCI runtime kata: no valid executable found for OCI runtime kata: invalid a>
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=info msg="Setting parallel job count to 13"
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=info msg="/usr/bin/podman filtering at log level info"
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=warning msg="Error initializing configured OCI runtime kata: no valid executable found for OCI runtime kata: invalid a>
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=info msg="Setting parallel job count to 13"
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=info msg="using API endpoint: 'unix:/run/user/1000/podman/podman.sock'"
Apr 16 20:00:56 crc-xl2km-master-0 podman[3990]: time="2021-04-16T20:00:56Z" level=info msg="API server listening on \"/run/user/1000/podman/podman.sock\""
Apr 16 20:00:56 crc-xl2km-master-0 systemd[3830]: podman.service: Got notification message from PID 4000, but reception only permitted for main PID 3990
```

With podman root, I have this error when it hangs:
```
Apr 16 19:53:04 crc-xl2km-master-0 systemd[3830]: Failed to connect to API bus: Connection refused
```

### DIFF
--- a/cmd/crc/cmd/podman_env.go
+++ b/cmd/crc/cmd/podman_env.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
@@ -14,12 +13,11 @@ var podmanEnvCmd = &cobra.Command{
 	Short: "Setup podman environment",
 	Long:  `Setup environment for 'podman' executable to access podman on CRC VM`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		// See issue #961; Currently does not work on Windows in combination with the CRC vm.
-		return errors.New("currently not supported")
+		return runPodmanEnv()
 	},
 }
 
-func RunPodmanEnv(args []string) error {
+func runPodmanEnv() error {
 	userShell, err := shell.GetShell(forceShell)
 	if err != nil {
 		return fmt.Errorf("Error running the podman-env command: %s", err.Error())
@@ -35,11 +33,13 @@ func RunPodmanEnv(args []string) error {
 		return err
 	}
 
-	fmt.Println(shell.GetPathEnvString(userShell, constants.CrcBinDir))
-	fmt.Println(shell.GetEnvString(userShell, "PODMAN_USER", connectionDetails.SSHUsername))
-	fmt.Println(shell.GetEnvString(userShell, "PODMAN_HOST", connectionDetails.IP))
-	fmt.Println(shell.GetEnvString(userShell, "PODMAN_IDENTITY_FILE", connectionDetails.SSHKeys[0]))
-	fmt.Println(shell.GetEnvString(userShell, "PODMAN_IGNORE_HOSTS", "1"))
+	fmt.Println(shell.GetPathEnvString(userShell, constants.CrcOcBinDir))
+	fmt.Println(shell.GetEnvString(userShell, "CONTAINER_SSHKEY", connectionDetails.SSHKeys[0]))
+	fmt.Println(shell.GetEnvString(userShell, "CONTAINER_HOST",
+		fmt.Sprintf("ssh://%s@%s:%d/run/user/1000/podman/podman.sock",
+			connectionDetails.SSHUsername,
+			connectionDetails.IP,
+			connectionDetails.SSHPort)))
 	fmt.Println(shell.GenerateUsageHintWithComment(userShell, "crc podman-env"))
 	return nil
 }
--- a/cmd/crc/cmd/podman_env.go
+++ b/cmd/crc/cmd/podman_env.go
@@ -30,15 +30,15 @@ func RunPodmanEnv(args []string) error {
 		return err
 	}
 
-	ip, err := client.IP()
+	connectionDetails, err := client.ConnectionDetails()
 	if err != nil {
 		return err
 	}
 
 	fmt.Println(shell.GetPathEnvString(userShell, constants.CrcBinDir))
-	fmt.Println(shell.GetEnvString(userShell, "PODMAN_USER", constants.DefaultSSHUser))
-	fmt.Println(shell.GetEnvString(userShell, "PODMAN_HOST", ip))
-	fmt.Println(shell.GetEnvString(userShell, "PODMAN_IDENTITY_FILE", constants.GetPrivateKeyPath()))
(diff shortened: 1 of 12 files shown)


guillaumerose:
Success rate is much higher with https://gist.github.com/guillaumerose/596cd7d99cbe0ad03ca7b0c9a6b20376 🤔 
I don't understand why.. Less SSH connections ?

pkg/crc/machine/bundle/repository.go
```diff
@@ -117,6 +120,23 @@ func (bundle *CrcBundleInfo) createSymlinkOrCopyOpenShiftClient(ocBinDir string)
 	return os.Symlink(ocInBundle, ocInBinDir)
 }
 
+func (bundle *CrcBundleInfo) createSymlinkOrCopyPodmanClient(ocBinDir string) error {
+	podmanInBundle := bundle.GetPodmanPath()
+	if podmanInBundle == "" {
+		return nil
+	}
+	podmanInBinDir := filepath.Join(ocBinDir, filepath.Base(podmanInBundle))
```
praveenkumar, line 128: why we want to have podman in `ocBinDir` just because symlink pupose?

pkg/crc/machine/start.go
```diff
@@ -342,6 +342,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		return nil, errors.Wrap(err, "Error running post start")
 	}
 
+	if _, ok := os.LookupEnv("UNSUPPORTED_PODMAN_ONLY"); ok {
```
praveenkumar, line 345: Souldn't this before we start the DNS server inside the VM because this is only required for resolving openshift domain names.

pkg/crc/machine/bundle/repository.go
guillaumerose: Because if user has ocBinDir in their PATH, podman will already be available for them without extra action.

We could rename oc binary directory but it would break some users, etc.

pkg/crc/machine/start.go
guillaumerose: Yes not very good.. Lots of things should be turned off with this flag. I don't know if I should keep it as it is.
It is really convienent to have it for working with podman only..

guillaumerose:
I added a proxy in the middle. It deals with ssh for podman. It's far more reliable for root mode.
For rootless mode, it is still flaky: first call fail, then it "works".

cfergeau:
> I added a proxy in the middle. It deals with ssh for podman. It's far more reliable for root mode.
> For rootless mode, it is still flaky: first call fail, then it "works".

Not so sure we should be adding workarounds for bugs which should be fixed in podman, or some place else, but not in crc.

cfergeau:
> With podman rootless, I have this error when it hangs:
```
Apr 16 20:00:56 crc-xl2km-master-0 systemd[3830]: podman.service: Got notification message from PID 4000, but reception only permitted for main PID 3990
```

Could be a bug in podman socket activation, `systemctl status` shows
```
 Main PID: 3990 (podman)
   CGroup: /user.slice/user-1000.slice/user@1000.service/podman.service
           ├─3990 /usr/bin/podman --log-level=info system service
           └─4000 /usr/bin/podman --log-level=info system service
```

The PIDs match the error message, but I'm not sure it's intentional 2 podman binaries were spawned.
The corresponding unit file is
```
[Unit]
Description=Podman API Service
Requires=podman.socket
After=podman.socket
Documentation=man:podman-system-service(1)
StartLimitIntervalSec=0

[Service]
Type=notify
KillMode=process
Environment=LOGGING="--log-level=info"
ExecStart=/usr/bin/podman $LOGGING system service
```

cmd/crc/cmd/podman_env.go
```diff
@@ -30,21 +33,27 @@ func RunPodmanEnv(args []string) error {
...
+func podmanSocket(rootless bool) string {
+	if rootless {
+		return filepath.Join(constants.CrcBaseDir, "podman-rootless.sock")
+	}
+	return filepath.Join(constants.CrcBaseDir, "podman.sock")
+}
+
 func init() {
 	rootCmd.AddCommand(podmanEnvCmd)
+	podmanEnvCmd.Flags().BoolVar(&rootless, "rootless", false, "Allow the use of experimental features")
```
cfergeau, line 57: The description does not describe `--rootless` but seems copied/pasted.

guillaumerose:
I kept only necessary commits. I tested it fine on the 3 platforms. 
It kind of works, it could be better. I see this PR as a first step before getting something really good when OpenShift 4.8 will ship also Podman 3.

Issue #2206, opened 2021-04-09
When an installer is used, use helper binaries from the installation directory


cfergeau:
Rather than copying admin-helper, ... from the installation directory to ~/.crc/bin/ when crc setup is run, crc could directly use the binaries which are installed.
This guarantees the expected binary is always used, no need for version check in preflights, ...

guillaumerose:
Admin rights (setuid bit) can also be set directly in the .pkg. It removes one more chmod.

guillaumerose:
macOS side is done.

PR #2252, opened 2021-04-20
#2206 macOS pkg: use hyperkit, driver and admin-helper directly from app dir

guillaumerose:
It removes the copy from the application directoy to ~/.crc/.
crc directly use binaries from /Application/CodeReady Containers.app.

**Fixes:** Issue #2206

---

I start to dislike the spread of `if version.IsMacosInstallPathSet() { }` blocks :)
Would it be better with a build tag ? 🤔 

Also, I need to double check code signing is ok with this.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -12,6 +12,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/network"
+	"github.com/code-ready/crc/pkg/crc/version"
 	crcos "github.com/code-ready/crc/pkg/os"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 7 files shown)


cfergeau:
> I start to dislike the spread of if version.IsMacosInstallPathSet() { } blocks :)
> Would it be better with a build tag ? thinking

I sympathize with this dislike :) No idea if a build tag would be better or worse, but hopefully it would make things clearer.

cfergeau:
> Fixes: Issue #2206

NB: this also needs to be fixed for Windows.

cfergeau (approved):
Wondering if https://github.com/code-ready/crc/blob/088d645a24f8feef5c39e7be7556509e8545f35a/pkg/embed/embed.go#L16-L21 can be removed?
Looks good to me otherwise.

guillaumerose:
I asked release team to sign it. Let's wait a bit.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -45,6 +50,10 @@ func checkHyperKitInstalled(networkMode network.Mode) func() error {
 
 func fixHyperKitInstallation(networkMode network.Mode) func() error {
 	return func() error {
+		if version.IsMacosInstallPathSet() {
+			return nil
+		}
+
```
praveenkumar, line 56: @guillaumerose Do we really need this check for this in the `fix` functions, if we have it on the `check`  then fix will not even run right?

guillaumerose:
Code signing and notarization are ok. It's good to merge now.

pkg/crc/preflight/preflight_checks_darwin.go
guillaumerose: oups sorry didnt see it. 

Yes you're right. I really hope to remove this code in two/three releases.

Issue #2181, opened 2021-04-02
Determining actions around sunsetting the single binary archive (macOS)


gbraad:
Since we now have an installer for macOS we need to determine how and when to take actions when it comes to sunsetting the single binary archive. For instance:

  * [x] how to handle the 'setup' (in relation to `crc setup`
  * [x] how to deal with providing the pull-secret 
  * [x] the integration with consumers, eg adapters (informing about timeline, so they can schedule around this)
     * [x] vscode
     * [x] eclipse
  * [ ] ...

cfergeau:
We could stop shipping crc-macos-amd64.tar.xz now, and tell our consumers that they need to adjust how they install crc, and that they can find the crc executable in /usr/local/bin.

The first 2 items seem like additional integration work which is made possible by the use of an installer, but the single binary archive could be removed before or after they are done, this seems relatively independent.

gbraad:
> We could stop shipping crc-macos-amd64.tar.xz now

I do not think this is the right approach as both the vscode adapters and the eclipse plugin use these files (download) to setup a local cluster environment.

@denisgolovin, @jeffmaury and @robstryker can likely tell you more about this.

jeffmaury:
> 
> 
> > We could stop shipping crc-macos-amd64.tar.xz now
> 
> I do not think this is the right approach as both the vscode adapters and the eclipse plugin use these files (download) to setup a local cluster environment.
> 
> @dgolovin, @jeffmaury and @robstryker can likely tell you more about this.

Yes removing the tar.xz packaging for MacOS will break the CRC integration in JBoss Tools as we provide bootstrap experience. We are evaluating integrating .pkg into the next release of JBoss Tools / CodeReady Studio scheduled early July

dgolovin:
Not breaking VSCode OpenShift Extension, bc installing the binaries is still user responsibility.
Is installer available only for macOS?

dgolovin:
NM. I see it is only 'crc-macos-amd64.pkg', which would require to run packager on macOS to install. It is still archive for everything else.

cfergeau:
> I do not think this is the right approach as both the vscode adapters and the eclipse plugin use these files (download) to setup a local cluster environment.

I was not suggesting to do it right away, I was using  'could' to express something purely hypothetical/rethoritical. All the steps that you list in this issue are not required before we can stop shipping the crc archive.

gbraad:
Windows msi is currently being tested, but awaiting the signing to move forward.

[...]

Issue #1945, opened 2021-02-03
Add more values for `network-mode` config key


cfergeau:
`network-mode` valid values are `default` or `vsock`. We need another name for the `default` mode as the current plan is that `vsock` becomes the default in the future.

gbraad:
All I could think of was `legacy` or perhaps more informative `bridged` or `switched`, but this kinda based on the hypervisor... and `native` is also not the right name.

cfergeau:
`bridged` sounds good to me, but not sure how good this name is on Windows?

gbraad:
Workable...

cfergeau:
Another possibility for the naming could be `network-mode`: `user` (=`vsock`) and `system`/`kernel` (=`default`).

PR #2253, opened 2021-04-21
Change default network mode to vsock for macOS

guillaumerose:
With the routes controller and the pkg (almost) in place, we can now change the network mode to vsock.

Maybe I should block all commands with `checkDaemonStarted` apart from setup/cleanup/version/daemon. WDYT?

---

When https://cloud.redhat.com/openshift/create/local will serve the pkg, we can merge this

### DIFF
--- a/cmd/crc/cmd/config/config.go
+++ b/cmd/crc/cmd/config/config.go
@@ -11,6 +11,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/network"
+	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/spf13/cast"
 	"github.com/spf13/cobra"
 )
@@ -82,7 +83,7 @@ func RegisterSettings(cfg *config.Config) {
 		"Disable update check (true/false, default: false)")
 	cfg.AddSetting(ExperimentalFeatures, false, config.ValidateBool, config.SuccessfullyApplied,
 		"Enable experimental features (true/false, default: false)")
-	cfg.AddSetting(NetworkMode, string(network.DefaultMode), network.ValidateMode, network.SuccessfullyAppliedMode,
+	cfg.AddSetting(NetworkMode, string(defaultNetworkMode()), network.ValidateMode, network.SuccessfullyAppliedMode,
 		"Network mode (default or vsock)")
 	cfg.AddSetting(HostNetworkAccess, false, validateHostNetworkAccess, config.SuccessfullyApplied,
 		"Allow TCP/IP connections from the CodeReady Containers VM to services running on the host (true/false, default: false)")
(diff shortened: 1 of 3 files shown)


cmd/crc/cmd/start.go
```diff
@@ -250,3 +256,25 @@ func commandLinePrefix(shell string) string {
...
 }
+
+func checkDaemonStarted() error {
+	if config.Get(cmdConfig.NetworkMode).AsString() == string(network.DefaultMode) {
+		return nil
+	}
+	daemonClient := daemonclient.New()
+	version, err := daemonClient.APIClient.Version()
+	if err != nil {
+		return fmt.Errorf(daemonStartedErrorMessage(), err)
```
cfergeau, line 267: This could start the daemon rather than returning an error, at least in the !installer case.

cfergeau:
Changing the default for the installer is fine, however we'll also still be shipping the binary with embedded bundle for a while because of https://github.com/code-ready/crc/issues/2181
Before switching to vsock by default in the fat binary, the daemon needs to be started on demand (which is hopefully not very hard to add now :)

cmd/crc/cmd/config/config.go
```diff
@@ -107,6 +108,13 @@ func RegisterSettings(cfg *config.Config) {
 		"Consent to collection of anonymous usage data (yes/no)")
 }
 
+func defaultNetworkMode() network.Mode {
+	if runtime.GOOS == "darwin" && version.IsMacosInstallPathSet() {
+		return network.VSockMode
+	}
+	return network.DefaultMode
```
gbraad, line 115: @guillaumerose what would be the value to 'revert' to the legacy (switched) network mode?

cfergeau: `crc config set network-mode default` to use the same network mode as in the previous releases.

gbraad: and that would be a 'lie', as it is not the default. We might wanna consider changing the wording of this for macOS. WDYT?

cfergeau: yes this is confusing, this is tracked in https://github.com/code-ready/crc/issues/1945
Imo, that's good enough for this release, and we improve this in the next release.

cmd/crc/cmd/start.go
```diff
@@ -250,3 +256,25 @@ func commandLinePrefix(shell string) string {
...
+	}
+	if version.CrcVersion != crcversion.GetCRCVersion() {
+		return fmt.Errorf("The executable version (%s) doesn't match the daemon version (%s)", crcversion.GetCRCVersion(), version.CrcVersion)
+	}
+	return nil
+}
+
+func daemonStartedErrorMessage() string {
+	if crcversion.IsMacosInstallPathSet() {
+		return "Is '/Application/CodeReady Containers.app' running? Cannot reach daemon API: %v"
```
cfergeau, line 277: It's `/Applications` (missing 's'). If this message is intended for users, should it be using "macOS tray" instead? But I don't think it's going to be possible to trigger this from the tray, as if the daemon is not running, we'll never reach the go code?

cfergeau (approved):
Minor typo/comment, looks good otherwise (did not test it)

cmd/crc/cmd/start.go
guillaumerose: Thanks fixed. 
This message is for people trying to run `crc` from the CLI without the tray launched. "macOS tray" doesn't really say what do to to get this working.

guillaumerose:
retest this please

PR #2272, opened 2021-04-26
Add preflight check to prevent running on Apple M1-based hardware

cfergeau:
This detects emulated x86_64 environment on M1, and returns an error when this happens.
On non-m1 systems, `crc setup` will print an additional message:
```
```

On M1 systems, this will then print:
```
INFO CodeReady Containers is unsupported on Apple M1 hardware
```


gbraad:
> On non-m1 systems, crc setup will print an additional message:

Why on non-M1 do we need to show a message?

gbraad (commented):
Merely questions around the 'why test amd64' and the description detailing a different expectation and what can be seen in code.

anjannath:
> This detects emulated x86_64 environment on M1, and returns an error when this happens.
> On non-m1 systems, `crc setup` will print an additional message:
> 
> ```
> INFO CodeReady Containers is unsupported on Apple M1 hardware
> ```

The additional message printed during setup is i think:
```
```

Also should we add this message to the integration test's expected output for `crc setup` and `crc start`?

gbraad:
It is wrongly described, as this message is intended for M1 systems. And the amd64 check is to ensure this runs on a non-Arm machine... I understand this, but it is differently described and not documented/commented as such

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -10,6 +10,13 @@ import (
 // SetupHost performs the prerequisite checks and setups the host to run the cluster
 func hyperkitPreflightChecks(networkMode network.Mode) []Check {
 	return []Check{
+		{
+			configKeySuffix:  "check-supported-cpu-arch",
+			checkDescription: "Checking if running on a supported CPU architecture",
+			check:            checkSupportedCPUArch,
+			fixDescription:   "CodeReady Containers is unsupported on Apple M1 hardware",
```
praveenkumar, line 17: @cfergeau Since `checkSupportedCPUArch` handle M1 and also any other arch so in `fixDescription` just mentioning `Apple M1` would be enough?

praveenkumar (approved):

guillaumerose (approved):
Unhappy unit tests otherwise,
/lgtm

pkg/crc/preflight/preflight_darwin.go
gbraad: Not sure what you mean

cfergeau:
> Unhappy unit tests otherwise,

Only ran `make check` on a linux box, I guess this did not run some of the tests, will fix.

cfergeau:
> You check for amd64 and this could be a generic test, as also on linux and windows we do not support anything else, right? Why this check ...

Yup, the amd64 check could be in generic code, but it's more code ;) Part of the reasoning was that we only produce x86_64 binaries anyway, so the generic check is not that important.

cfergeau:
I split the check into a generic x86_64 one, and a macos-specific 'emulated x86_64' check, and tried to adjust the various unit tests (only tested on linux and macos).

gbraad (approved):

PR #2273, opened 2021-04-27
Use a release version of the routes controller (v0.0.1)

guillaumerose:
https://github.com/code-ready/routes-controller/releases/tag/v0.0.1

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -661,7 +661,7 @@ func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Confi
 			Containers: []v1.Container{
 				{
 					Name:            "routes-controller",
-					Image:           "quay.io/crcont/routes-controller:6effc2f0304d3a11c1e22c278b10630e834d0220",
+					Image:           "quay.io/crcont/routes-controller:latest",
 					ImagePullPolicy: v1.PullIfNotPresent,
 				},
 			},


guillaumerose:
/hold
I forgot that I need to use a tag like :current to avoid the duplication of version in snc and crc.

gbraad (approved):

guillaumerose:
I tried with `:current` tag. It works but I am worry that the user might delete the image from the bundle by mistake. 
If the user does something like `podman rmi $(podman images -q)`, the pod won't be able to start anymore as the tag is only local.

/unhold

gbraad:
That would be a user error.m the vm is supposed to be tested ephemeral, so a delete and restart would resolve that

praveenkumar:
I am thinking it https://github.com/code-ready/snc/pull/398#issuecomment-827550149 this way.

pkg/crc/machine/start.go
```diff
@@ -661,7 +661,7 @@ func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Confi
 			Containers: []v1.Container{
 				{
 					Name:            "routes-controller",
-					Image:           "quay.io/crcont/routes-controller:6effc2f0304d3a11c1e22c278b10630e834d0220",
+					Image:           "quay.io/crcont/routes-controller:latest",
```
anjannath, line 664: 👍🏽

anjannath (approved):

anjannath:
retest this please

Issue #2276, opened 2021-04-27
[BUG] Don't crash the daemon if clients send unexpected data for a request


anjannath:
The daemon crashes with the following error, if a client request `SetConfig` and has empty data:

```
PS C:\Users\Anjan\github.com\code-ready\crc> crc daemon
INFO Checking if podman remote executable is cached
INFO Checking if admin-helper executable is cached
INFO Checking minimum RAM requirements
INFO Checking if running in a shell with administrator rights
INFO Checking Windows 10 release
INFO Checking Windows edition
INFO Checking if Hyper-V is installed and operational
INFO Checking if user is a member of the Hyper-V Administrators group
INFO Checking if Hyper-V service is enabled
INFO Checking if the Hyper-V virtual switch exist
INFO Found Virtual Switch to use: Default Switch
INFO listening vsock://00000400-FACB-11E6-BD58-64006A7986D3
INFO listening \\.\pipe\crc-http
panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 27 [running]:
github.com/code-ready/crc/pkg/crc/api.(*Handler).SetConfig(0xc000584ba0, 0xc00058adb0, 0x10, 0x10, 0x0, 0xc0003adf90)
        C:/Users/Anjan/github.com/code-ready/crc/pkg/crc/api/handlers.go:139 +0x717
github.com/code-ready/crc/pkg/crc/api.Server.handleRequest(0x1fb2c28, 0xc000584ba0, 0x1fa07a0, 0xc000584b70, 0xc00058ada0, 0x9, 0xc00058adb0, 0x10, 0x10, 0x1fb0f68, ...)
        C:/Users/Anjan/github.com/code-ready/crc/pkg/crc/api/api.go:65 +0x414
created by github.com/code-ready/crc/pkg/crc/api.Server.handleConnections
        C:/Users/Anjan/github.com/code-ready/crc/pkg/crc/api/api.go:94 +0x3fa
```

guillaumerose:
It would be probably good to use a more high-level framework for the HTTP API. It would avoid this kind of problems. 

https://github.com/gin-gonic/gin, https://github.com/gorilla/mux or https://echo.labstack.com/

Personally I would like a framework that do the JSON marshalling/unmarshalling for me.

PR #2281, opened 2021-04-28
Issue #2276 Check if interface is nil before trying to convert it

anjannath:
This fixes the panic during interface conversion when the daemon is provided unexpected data
like: `echo '{"command":"setconfig", "args":{}}' | nc -U ~/.crc/crc.sock`

which results in the following panic:
```
DEBU rpc:0xc000114048: Received request: {"command":"setconfig", "args":{}}
panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 88 [running]:
github.com/code-ready/crc/pkg/crc/api.(*Handler).SetConfig(0xc0005a80c0, 0xc00011a118, 0x2, 0x8, 0x0, 0x0)
	/home/anjan/github.com/code-ready/crc/pkg/crc/api/handlers.go:146 +0x717
github.com/code-ready/crc/pkg/crc/api.Server.handleRequest(0xc0005a80c0, 0x155bb60, 0xc0005a2240, 0xc00011a130, 0x9, 0xc00011a118, 0x2, 0x8, 0x156c1e0, 0xc000114048)
	/home/anjan/github.com/code-ready/crc/pkg/crc/api/api.go:61 +0x425
created by github.com/code-ready/crc/pkg/crc/api.Server.handleConnections
	/home/anjan/github.com/code-ready/crc/pkg/crc/api/api.go:92 +0x3c7

```

### DIFF
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -143,7 +143,11 @@ func (h *Handler) SetConfig(args json.RawMessage) string {
 		return encodeStructToJSON(setConfigResult)
 	}
 
-	configs := a["properties"].(map[string]interface{})
+	configs, ok := a["properties"].(map[string]interface{})
+	if !ok {
+		setConfigResult.Error = "No config keys provided"
+		return encodeStructToJSON(setConfigResult)
+	}
 
 	// successProps slice contains the properties that were successfully set
 	var successProps []string


pkg/crc/api/handlers.go
```diff
@@ -143,6 +143,11 @@ func (h *Handler) SetConfig(args json.RawMessage) string {
 		return encodeStructToJSON(setConfigResult)
 	}
 
+	if a["properties"] == nil {
+		setConfigResult.Error = "No config keys provided"
+		return encodeStructToJSON(setConfigResult)
+	}
+
 	configs := a["properties"].(map[string]interface{})
```
guillaumerose, line 151: You can also do: 

```
configs, ok := a["properties"].(map[string]interface{})
if !ok {
  setConfigResult.Error = "Invalid config"
  return encodeStructToJSON(setConfigResult)
}
``` 

It will throw an error if you pass a JSON array for instance.

gbraad: ... while this is a better fix and prevents an error, this should also be prevented to be sent from the client(s)

pkg/crc/api/handlers.go
```diff
@@ -143,7 +143,11 @@ func (h *Handler) SetConfig(args json.RawMessage) string {
 		return encodeStructToJSON(setConfigResult)
 	}
 
-	configs := a["properties"].(map[string]interface{})
+	configs, ok := a["properties"].(map[string]interface{})
+	if !ok {
+		setConfigResult.Error = "No config keys provided"
+		return encodeStructToJSON(setConfigResult)
+	}
```
praveenkumar, line 150: 👍🏼

praveenkumar (approved):

PR #2282, opened 2021-04-28
Remove upgrade code that fixes /etc/hosts permissions on macOS

guillaumerose:
This code was introduced to fix crc 1.10 and before (see
https://github.com/code-ready/crc/pull/1350).

---

It removes one more piece of code that requires sudo. It will be easier to include `crc setup` in the .pkg with fewer sudo actions.

### DIFF
--- a/pkg/crc/preflight/preflight_darwin_test.go
+++ b/pkg/crc/preflight/preflight_darwin_test.go
@@ -12,13 +12,13 @@ import (
 func TestCountConfigurationOptions(t *testing.T) {
 	cfg := config.New(config.NewEmptyInMemoryStorage())
 	RegisterSettings(cfg)
-	assert.Len(t, cfg.AllConfigs(), 10)
+	assert.Len(t, cfg.AllConfigs(), 9)
 }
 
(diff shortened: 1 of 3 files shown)


gbraad (approved):

pkg/crc/preflight/preflight_darwin_test.go
```diff
@@ -12,13 +12,13 @@ import (
...
 	cfg := config.New(config.NewEmptyInMemoryStorage())
 	RegisterSettings(cfg)
-	assert.Len(t, cfg.AllConfigs(), 11)
+	assert.Len(t, cfg.AllConfigs(), 10)
 }
 
 func TestCountPreflights(t *testing.T) {
-	assert.Len(t, getPreflightChecks(true, false, network.DefaultMode), 17)
-	assert.Len(t, getPreflightChecks(true, true, network.DefaultMode), 17)
+	assert.Len(t, getPreflightChecks(true, false, network.DefaultMode), 16)
```
gbraad, line 19: the more often I see this, the more I start to dislike this 'faux'-test

PR #2297, opened 2021-04-30
macOS pkg: set the list of supported architectures (x86_64, not arm64)

guillaumerose:
This should prevent users from installing the pkg on M1 machines.

(to be tested on a M1 machine...)


guillaumerose:
After test, this doesn't prompt an error.

cfergeau:
Maybe by the time the checks is triggered, the M1 is already running in x86_64 emulation mode, and thus the tests succeeds.

PR #2302, opened 2021-05-04
Add an enum for OpenShift status. Ensure it doesn't change.

guillaumerose:
These values are used in various clients (consumers of the json output
of the CLI, tray, etc.). It is better to highlight this with an enum.

The goal is after to consolidate the 3 enums:

``` 
// State represents the state of a host
type State int

const (
	None State = iota
	Running
	Paused
	Saved
	Stopped
	Stopping
	Starting
	Error
	Timeout
)

type State string

const (
	Idle     State = "Idle"
	Deleting State = "Deleting"
	Stopping State = "Stopping"
	Starting State = "Starting"
)

type OpenshiftStatus string

const (
	OpenshiftUnreachable OpenshiftStatus = "Unreachable"
	OpenshiftStarting    OpenshiftStatus = "Starting"
	OpenshiftRunning     OpenshiftStatus = "Running"
	OpenshiftDegraded    OpenshiftStatus = "Degraded"
	OpenshiftStopped     OpenshiftStatus = "Stopped"
)
```

### DIFF
--- a/pkg/crc/machine/types/types.go
+++ b/pkg/crc/machine/types/types.go
@@ -46,11 +46,21 @@ type StopResult struct {
 
 type ClusterStatusResult struct {
 	CrcStatus        state.State
-	OpenshiftStatus  string
+	OpenshiftStatus  OpenshiftStatus
 	OpenshiftVersion string
 	DiskUse          int64
 	DiskSize         int64
 }
+
+type OpenshiftStatus string
+
+const (
+	OpenshiftUnreachable OpenshiftStatus = "Unreachable"
+	OpenshiftStarting    OpenshiftStatus = "Starting"
+	OpenshiftRunning     OpenshiftStatus = "Running"
+	OpenshiftDegraded    OpenshiftStatus = "Degraded"
+	OpenshiftStopped     OpenshiftStatus = "Stopped"
+)
 
 type ConsoleResult struct {
 	ClusterConfig ClusterConfig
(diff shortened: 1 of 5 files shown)


pkg/crc/machine/types/types.go
```diff
@@ -46,12 +46,22 @@ type StopResult struct {
...
+type OpenshiftStatus string
+
+const (
+	OpenshiftUnreachable OpenshiftStatus = "Unreachable"
+	OpenshiftStarting    OpenshiftStatus = "Starting"
+	OpenshiftRunning     OpenshiftStatus = "Running"
+	OpenshiftDegraded    OpenshiftStatus = "Degraded"
+	OpenshiftStopped     OpenshiftStatus = "Stopped"
+)
+
```
praveenkumar, line 64: @guillaumerose  I want to know what is the benefit we are going to have with enum and  if we do something like following? We are only making sure it is only single place. Even with enum if someone want to change the value they can can in the code right?

```
const (
	OpenshiftUnreachable  = "Unreachable"
	OpenshiftStarting     = "Starting"
	OpenshiftRunning      = "Running"
	OpenshiftDegraded     = "Degraded"
	OpenshiftStopped      = "Stopped"
)
```

guillaumerose: For instance, OpenshiftStopped is used twice and it is easy to find where. 
Before this enum, I would have to search for "Stopped" string. I would get tons of result. 

It is also a good way to show consumers what are the possible values. 

The point of this PR is also to highlight these 3 enums and do something with that. 
I don't know what but there is something to do for sure!

guillaumerose:
/retest
retest this please

PR #2304, opened 2021-05-05
Add sshkey with root user

praveenkumar:
Right now https://bugzilla.redhat.com/show_bug.cgi?id=1956739 blocker
for us to enable MCO. This patch we will revert it as soon as it is
fixed on openshift side?

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -571,7 +571,7 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	}
 
 	logging.Info("Updating authorized keys ...")
-	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys; chmod 644 /home/core/.ssh/authorized_keys", publicKey)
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/start.go
```diff
@@ -571,7 +571,7 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	}
 
 	logging.Info("Updating authorized keys ...")
-	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys; chmod 644 /home/core/.ssh/authorized_keys", publicKey)
+	cmd := fmt.Sprintf("echo '%s' | sudo tee /home/core/.ssh/authorized_keys;sudo chmod 644 /home/core/.ssh/authorized_keys", publicKey)
```
cfergeau, line 574: Can we do this `chmod` in `snc` rather than in `crc`? Can you indicate in the commit log that this would fix CI against 4.8 bundles/snc master?

Issue #2182, opened 2021-04-05
Try to remove kernel args as part of hyperkit process


praveenkumar:
As of now we provide the ostree info as part of kernel argument for hyperkit which is part of crc metadata and updated during the bundle creation.  Since with 4.8 openshift going to support single node which means user able to run the machine config changes and after reboot instance should be on latest ostree layer. In case of hyperV and libvirt this is not an issue but hyperkit we have to find something so we don't relay on this kernel args.

cfergeau:
Is the issue the following?

- 4.8 crc (maybe) runs MCO
- user does configuration changes through MCO
- the `ostree=/ostree/boot.0/rhcos/bf3b38268ddb2a2070dc587b361ce45c46a6f9e3606bbd3b5e15b2e6d3d47e5d/0` boot parameter in the kernel command line changes

Since the kernel parameters are currently hardcoded in the bundle metadata on macOS, this cannot work.

praveenkumar:
@cfergeau That's right, from 4.8 side user can enable MCO and do config changes which MCO take care and update the ostree.

cfergeau:
With hyperkit, we could use UEFI boot to remove the need for kernel/initrd/kernel cmdline, but I'm not sure the same is going to be possible with virtualization.framework.

guillaumerose:
Using qemu on macOS would solve that. 

It might be a good move. I heard multiple times that it might be better than the Virtualization framework.

Also: https://docs.docker.com/docker-for-mac/apple-m1/#docker-desktop-rc-3

> In this build, we have defaulted to a qemu-based virtual machine, which we believe resolves some of the issues noted as known issues in the previous release candidate. You can switch between qemu-based and virtualization.framework-based virtual machines using the Preferences

praveenkumar:
That's good to know, we should at least evaluate it and check what issues we might face.

cfergeau:
Might become a more pressing issue with the MCO changes coming in 4.8

cfergeau:
Using UEFI is not complicated:
- download a `UEFI.fd` build (docker desktop has one, and https://freebsd.pkgs.org/12/freebsd-amd64/uefi-edk2-bhyve-0.2_1,1.txz.html worked too)
- apply this small hyperkit driver patch: https://github.com/cfergeau/crc-driver-hyperkit/commit/4e6a85f7d19725aa05bbef6468cc4dc3c531d321

cfergeau:
[...]

PR #2307, opened 2021-05-06
MCO specific changes

praveenkumar:
- Since MCO is now enabled by default, pull secret should be added
automatic to the vm when it starts.
- With MCO enabled we need to update the ssh key to machineconfig
so it is become part of cluster and also to instance.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"regexp"
 	"strconv"
 	"strings"
@@ -90,6 +91,33 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 		return 0, 0, err
 	}
 	return diskSize, diskUsage, nil
+}
+
+func UpdateSSHKeyToMachineConfig(ocConfig oc.Config, sshPublicKeyPath string) error {
+	sshPublicKeyByte, err := ioutil.ReadFile(sshPublicKeyPath)
+	if err != nil {
+		return err
+	}
+	sshPublicKey := strings.TrimRight(string(sshPublicKeyByte), "\r\n")
+	if err := WaitForOpenshiftResource(ocConfig, "machineconfigs"); err != nil {
+		return err
+	}
+	stdout, _, err := ocConfig.RunOcCommand("get", "machineconfigs", "99-master-ssh", "-o", `jsonpath='{.spec.config.passwd.users[0].sshAuthorizedKeys[0]}'`)
+	if err != nil {
+		return err
+	}
+	if stdout == string(sshPublicKey) {
+		return nil
+	}
+	logging.Info("Updating SSH key to machine config resource...")
+	cmdArgs := []string{"patch", "machineconfig", "99-master-ssh", "-p",
+		fmt.Sprintf(`'{"spec": {"config": {"passwd": {"users": [{"name": "core", "sshAuthorizedKeys": ["%s"]}]}}}}'`, sshPublicKey),
+		"--type", "merge"}
+	_, stderr, err := ocConfig.RunOcCommandPrivate(cmdArgs...)
+	if err != nil {
+		return fmt.Errorf("Failed to update ssh key %v: %s", err, stderr)
+	}
+	return nil
 }
 
 func EnsurePullSecretPresentInTheCluster(ocConfig oc.Config, pullSec PullSecretLoader) error {
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -313,18 +313,6 @@ func (p *PullSecretMemoizer) Value() (string, error) {
 	return val, err
 }
 
-func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret PullSecretLoader) error {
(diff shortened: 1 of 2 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -92,6 +93,33 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
...
+		return err
+	}
+	if stdout == string(sshPublicKey) {
+		return nil
+	}
+	logging.Info("Updating SSH key to machine config resource...")
+	cmdArgs := []string{"patch", "machineconfig", "99-master-ssh", "-p",
+		fmt.Sprintf(`'{"spec": {"config": {"passwd": {"users": [{"name": "core", "sshAuthorizedKeys": ["%s"]}]}}}}'`, sshPublicKey),
+		"--type", "merge"}
+	_, stderr, err := ocConfig.RunOcCommandPrivate(cmdArgs...)
```
cfergeau, line 116: This does not need to be private.

pkg/crc/cluster/cluster.go
```diff
@@ -92,6 +93,33 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 	return diskSize, diskUsage, nil
 }
 
+func UpdateSSHKeyToMachineConfig(ocConfig oc.Config, sshPublicKeyPath string) error {
+	sshPublicKeyByte, err := ioutil.ReadFile(sshPublicKeyPath)
```
cfergeau, line 97: `sshPublicKeyBytes`

pkg/crc/cluster/cluster.go
```diff
@@ -92,6 +93,33 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 	return diskSize, diskUsage, nil
 }
 
+func UpdateSSHKeyToMachineConfig(ocConfig oc.Config, sshPublicKeyPath string) error {
```
cfergeau, line 96: The pull secret method with a similar role is called `EnsurePullSecretPresentInTheCluster`, would be nice to have some consistency I think.

pkg/crc/cluster/cluster.go
```diff
@@ -92,6 +93,33 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
...
+		return err
+	}
+	stdout, _, err := ocConfig.RunOcCommand("get", "machineconfigs", "99-master-ssh", "-o", `jsonpath='{.spec.config.passwd.users[0].sshAuthorizedKeys[0]}'`)
+	if err != nil {
+		return err
+	}
+	if stdout == string(sshPublicKey) {
+		return nil
+	}
+	logging.Info("Updating SSH key to machine config resource...")
```
cfergeau, line 112: `Adding public SSH key to the cluster` or something similar is easier to read/understand imo.

pkg/crc/machine/start.go
```diff
@@ -394,6 +394,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		return nil, errors.Wrap(err, "Error waiting for apiserver")
 	}
 
+	if err := cluster.UpdateSSHKeyToMachineConfig(ocConfig, constants.GetPublicKeyPath()); err != nil {
+		return nil, errors.Wrap(err, "Failed to update ssh public key to machine config")
+	}
+
```
cfergeau, line 400: I'd add some details in the commit log explaining why this cant' be done in `updateSSHKeyPair`. Or maybe we'll want to run `updateSSHKeyPair` later in the future?

cfergeau (commented):
> pull secret should be added automatic to the vm when it starts.

This is misleading, since the MCO is running, after adding the pull secret `EnsurePullSecretPresentInTheCluster`, the MCO will automatically create `/var/lib/kubelet/config.json`. When I read the commit log, I get the (wrong) impression that it is always added even if we would not call `EnsurePullSecretPresentInTheCluster`

pkg/crc/machine/start.go
```diff
@@ -366,10 +366,6 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		logging.Warn(fmt.Sprintf("Failed to query DNS from host: %v", err))
 	}
 
-	if err := cluster.EnsurePullSecretPresentOnInstanceDisk(sshRunner, startConfig.PullSecret); err != nil {
```
guillaumerose, line 369: Doesn that mean the MCO will restart Kubelet to reload the pull secret ? Or the pull secret is reloading without restart Kubelet ?

pkg/crc/cluster/cluster.go
```diff
@@ -313,18 +313,6 @@ func (p *PullSecretMemoizer) Value() (string, error) {
 	return val, err
 }
 
-func EnsurePullSecretPresentOnInstanceDisk(sshRunner *ssh.Runner, pullSecret PullSecretLoader) error {
-	if _, _, err := sshRunner.Run(fmt.Sprintf("test -e %s", vmPullSecretPath)); err == nil {
```
cfergeau, line 317: Should `RemovePullSecretFromCluster` also be changed?

cfergeau:
Is this going to be impacted by https://github.com/code-ready/crc/issues/2182 on macOS?

pkg/crc/machine/start.go
guillaumerose: I also recall that some containers won't start if this file is not present. This line removed a lot of crc start failures in fact.

praveenkumar:
/close

Will try to put another PR which compare openshift version and then perform those operations.

PR #2308, opened 2021-05-06
Remove default operator ignore list for status

praveenkumar:
From 4.7 we are already removing operators from being part of
final bundle as part of snc so we don't need any list of ignore
operators. All the operators which are part of bundle should be
available and working as expected.


guillaumerose (approved):

guillaumerose:
Good catch!

guillaumerose:
/retest
retest this please

PR #2309, opened 2021-05-06
Add missing states for Hyper-V, remove state.None usage

guillaumerose:
The motivation is still to remove as much as possible the number of possible states.

I found these 2 new Hyper-V states by looking at the Hyper-V manager application and also with an infinite loop polling the crc VM status during start and stop.

### DIFF
--- a/pkg/drivers/hyperv/hyperv.go
+++ b/pkg/drivers/hyperv/hyperv.go
@@ -96,7 +96,7 @@ func (d *Driver) GetState() (state.State, error) {
 	}
 
 	switch resp[0] {
-	case "Running":
+	case "Starting", "Running", "Stopping":
 		return state.Running, nil
 	case "Off":
 		return state.Stopped, nil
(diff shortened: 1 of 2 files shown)


pkg/drivers/hyperv/hyperv.go
```diff
@@ -87,12 +87,12 @@ func (d *Driver) DriverName() string {
...
 	stdout, err := cmdOut("(", "Hyper-V\\Get-VM", d.MachineName, ").state")
 	if err != nil {
-		return state.None, fmt.Errorf("Failed to find the VM status")
+		return state.Error, fmt.Errorf("Failed to find the VM status")
 	}
 
 	resp := parseLines(stdout)
 	if len(resp) < 1 {
-		return state.None, nil
+		return state.Error, fmt.Errorf("unexpected Hyper-V state %s", resp)
```
cfergeau, line 95: Fwiw, the libvirt machine driver also uses `state.None` in similar situations, but it returns errors. The returned state should be assumed to be invalid anyway if the returned error is not `nil.

cfergeau (approved):
Looks good.

pkg/drivers/hyperv/hyperv.go
guillaumerose: yes for libvirt I have this one: https://github.com/code-ready/machine-driver-libvirt/pull/88

Issue #2154, opened 2021-03-26
[BUG] Failed to connect socket to '/var/run/libvirt/libvirt-sock-ro': No such file or directory


praveenkumar:
### General information

  * OS: Linux
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? yes

```
$ systemctl status libvirtd.socket 
● libvirtd.socket - Libvirt local socket
     Loaded: loaded (/usr/lib/systemd/system/libvirtd.socket; disabled; vendor preset: disabled)
     Active: active (listening) since Mon 2021-03-15 09:40:58 IST; 1 weeks 4 days ago
   Triggers: ● libvirtd.service

$ ./crc setup
INFO Checking if running as non-root              
INFO Checking if podman remote executable is cached 
INFO Checking if admin-helper executable is cached 
INFO Checking minimum RAM requirements            
INFO Checking if Virtualization is enabled        
INFO Checking if KVM is enabled                   
INFO Checking if libvirt is installed             
INFO Installing libvirt service and dependencies  
INFO Using root access: Installing virtualization packages 
INFO Checking if user is part of libvirt group    
INFO Checking if active user/process is currently part of the libvirt group 
INFO Checking if libvirt daemon is running        
INFO Checking if a supported libvirt version is installed 
INFO Checking if crc-driver-libvirt is installed  
INFO Checking if systemd-networkd is running      
INFO Checking if NetworkManager is installed      
INFO Checking if NetworkManager service is running 
INFO Checking if dnsmasq configurations file exist for NetworkManager 
INFO Checking if the systemd-resolved service is running 
INFO Checking if /etc/NetworkManager/dispatcher.d/99-crc.sh exists 
INFO Checking if libvirt 'crc' network is available 
INFO Checking if libvirt 'crc' network is active  
INFO Checking if CRC bundle is extracted in '$HOME/.crc' 
Setup is complete, you can now run 'crc start -b $bundlename' to start the OpenShift cluster

$ ./crc start
[...]
DEBU stdout:                                      
DEBU stderr: error: failed to connect to the hypervisor
error: Failed to connect socket to '/var/run/libvirt/libvirt-sock-ro': No such file or directory 
DEBU Failed to run 'virsh capabilities': exit status 1 
WARN Preflight checks failed during `crc start`, please try to run `crc setup` first in case you haven't done so yet 
Failed to run 'virsh capabilities': exit status 1

$ virsh --readonly --connect qemu:///system capabilities
[...]

PR #2312, opened 2021-05-06
preflight: Add virsh capabilities fallback

cfergeau:
On some setups, `virsh --read-only -c qemu:///system capabilities` may
not work out of the box.
This commit adds a fallback running
`virsh -c qemu:///session capabilities` when there is a failure. This is
the command which was used before commit 9f83839136

This fixes https://github.com/code-ready/crc/issues/2154

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -90,7 +90,10 @@ func fixKvmEnabled() error {
 func getLibvirtCapabilities() (*libvirtxml.Caps, error) {
 	stdOut, _, err := crcos.RunWithDefaultLocale("virsh", "--readonly", "--connect", "qemu:///system", "capabilities")
 	if err != nil {
-		return nil, fmt.Errorf("Failed to run 'virsh capabilities': %v", err)
+		stdOut, _, err = crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///session", "capabilities")
+		if err != nil {
+			return nil, fmt.Errorf("Failed to run 'virsh capabilities': %v", err)
+		}
 	}
 	caps := &libvirtxml.Caps{}
 	err = caps.Unmarshal(stdOut)


praveenkumar:
I didn't put the PR before because https://github.com/code-ready/crc/issues/2154#issuecomment-808151948 of this :) . Did someone also hit with same issue or we just added it for safer side?

cfergeau:
> I didn't put the PR before because [#2154 (comment)](https://github.com/code-ready/crc/issues/2154#issuecomment-808151948) of this :) . Did someone also hit with same issue or we just added it for safer side?

Just to be on the safer side, and was quick to fix. I don't mind dropping this as WONTFIX :)

praveenkumar:
> > I didn't put the PR before because [#2154 (comment)](https://github.com/code-ready/crc/issues/2154#issuecomment-808151948) of this :) . Did someone also hit with same issue or we just added it for safer side?
> 
> Just to be on the safer side, and was quick to fix. I don't mind dropping this as WONTFIX :)

No, it's good to have a safer side if we can avoid even a single user to have this kind of issues 👍🏼

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -90,7 +90,10 @@ func fixKvmEnabled() error {
 func getLibvirtCapabilities() (*libvirtxml.Caps, error) {
 	stdOut, _, err := crcos.RunWithDefaultLocale("virsh", "--readonly", "--connect", "qemu:///system", "capabilities")
 	if err != nil {
-		return nil, fmt.Errorf("Failed to run 'virsh capabilities': %v", err)
```
gbraad, line 93: Perhaps record fallback was needed?

cfergeau: The full log when the fallback codepath is triggered is 
```
DEBU Checking if 'virsh' is available             
DEBU 'virsh' was found in /usr/bin/virsh          
DEBU Checking 'virsh capabilities' for libvirtd/qemu availability 
DEBU Running 'virsh --readonly --connect qemu:///system capabilities2' 
DEBU stdout:                                      
DEBU stderr: error: unknown command: 'capabilities2' 
DEBU Running 'virsh --connect qemu:///session capabilities' 
DEBU Found x86_64 hypervisor with 'hvm' capabilities 
```
it's imo explicit enough regarding what's happening, but I can add a `logging.Debugf("Falling back to getting capabilities from qemu:///session")` if you think that would help.

PR #2320, opened 2021-05-07
Use more explicit names for the 'network-mode' options

cfergeau:
I went with `user` and `system` (`kernel`?) but I expect this is going to be discussed/changed.
The old options are still supported for backwards compatibility (in particular when they were written to the config file).
For testing, `vsock`/`user` should be usable interchangeably, and same for `default`/`system`.
Behaviour should be exactly the same regardless of what value the option is set to.

### DIFF
--- a/pkg/crc/network/types.go
+++ b/pkg/crc/network/types.go
@@ -23,15 +23,15 @@ type ResolvFileValues struct {
 type Mode string
 
 const (
-	SystemNetworkingMode Mode = "default"
-	UserNetworkingMode   Mode = "vsock"
+	SystemNetworkingMode Mode = "system"
+	UserNetworkingMode   Mode = "user"
 )
 
 func parseMode(input string) (Mode, error) {
 	switch input {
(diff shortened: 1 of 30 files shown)


guillaumerose (approved):

pkg/crc/network/types.go
```diff
@@ -23,27 +23,35 @@ type ResolvFileValues struct {
 type Mode string
 
 const (
-	DefaultMode Mode = "default"
-	VSockMode   Mode = "vsock"
+	SystemNetworkingMode Mode = "system"
+	UserNetworkingMode   Mode = "user"
```
gbraad, line 27: usermode would have made more sense

cfergeau: I opted out of `usermode` as `crc config set network-mode usermode` felt redundant.
I've renamed the options to `vsock`/`bridged` as was discussed earlier in this PR, code is at https://github.com/cfergeau/crc/tree/network-mode I can file a PR if this naming is preferred.

PR #2321, opened 2021-05-10
[msi] Don't use embedded crc binary and use helper binaries from install location

### LINKED ISSUE #2205: Don't use embedded binary with Windows installer
In the target installation directory, use separate files for crc.exe, crc bundle, helper binaries, ... rather than a crc.exe binary with everything embedded in it.
This is similar to what is done on macOS.

### LINKED ISSUE #2206: When an installer is used, use helper binaries from the installation directory

anjannath:
- Don't use embedded `crc` binary for building the msi
- use the bundle and helper binaries directly from `Program Files\CodeReady Containers` and don;t copy them to `~\.crc` 
- prompt user to exit the running tray if one is running during installation (this not similar to the macOS installer, as this would not exit the installation but presents user with a dialog to continue even if the tray is running), [screenshot below]

Fixes #2205 
Fixes #2206

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -323,7 +323,7 @@ endif
 	rm $(PACKAGE_DIR)/product.wxs
 
 $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi: msidir
-	candle.exe -arch x64 -o $(PACKAGE_DIR)/msi/ $(PACKAGE_DIR)/msi/*.wxs
+	candle.exe -arch x64 -ext WixUtilExtension -o $(PACKAGE_DIR)/msi/ $(PACKAGE_DIR)/msi/*.wxs
 	cd $(PACKAGE_DIR)/msi && light.exe -ext WixUIExtension -ext WixUtilExtension -sacl -spdb -sice:ICE61 -out ../../../$@ *.wixobj
 
 CABS_MSI = "cab1.cab,cab2.cab,cab3.cab,crc-windows-amd64.msi"
--- a/Makefile
+++ b/Makefile
@@ -301,18 +301,21 @@ $(GOPATH)/bin/gomod2rpmdeps:
 $(HOST_BUILD_DIR)/split: packaging/windows/split.go
 	go build -o $(HOST_BUILD_DIR)/split packaging/windows/split.go
 
+CRC_EXE=crc.exe
+BUNDLE_NAME=crc_hyperv_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -301,18 +301,21 @@ $(GOPATH)/bin/gomod2rpmdeps:
...
 	cp $(HOST_BUILD_DIR)/crc.exe $(PACKAGE_DIR)/msi/$(CRC_EXE)
+	pwsh -NoProfile -Command "cd $(PACKAGE_DIR)/msi; Expand-Archive crc-tray-windows.zip -DestinationPath .\; Remove-Item crc-tray-windows.zip"
 ifeq ($(MOCK_BUNDLE),true)
-	mv $(PACKAGE_DIR)/msi/$(CRC_EXE) $(PACKAGE_DIR)/msi/$(CRC_EXE).0
-	touch $(PACKAGE_DIR)/msi/$(CRC_EXE).1 $(PACKAGE_DIR)/msi/$(CRC_EXE).2
+	touch $(PACKAGE_DIR)/msi/$(BUNDLE_NAME).0 $(PACKAGE_DIR)/msi/$(BUNDLE_NAME).1 $(PACKAGE_DIR)/msi/$(BUNDLE_NAME).2
 else
-	$(HOST_BUILD_DIR)/split $(PACKAGE_DIR)/msi/$(CRC_EXE)
-	rm $(PACKAGE_DIR)/msi/$(CRC_EXE)
+	cp $(HYPERV_BUNDLENAME) $(PACKAGE_DIR)/msi
```
cfergeau, line 317: NB: given the size of the file, the copy can be a bit expensive (slow), `split` could probably be improved to accept a target directory in addition to the source file so that the copy can be avoided (but that's a different PR)

appveyor.yml
```diff
@@ -23,6 +23,6 @@ artifacts:
   - path: out\windows-amd64\crc.exe
     name: crc windows executable
     type: File
-  - path: out\windows-amd64\crc-installer.zip
+  - path: out\windows-amd64\crc-windows-installer.zip
```
cfergeau, line 26: I'd squash this commit in the one renaming the .zip file.

anjannath: Done!

cfergeau (approved):

PR #2324, opened 2021-05-10
Update ssh public key to machineconfig resource if used 4.8 cluster

praveenkumar:
With MCO enabled we need to update the ssh key to machineconfig
so it is become part of cluster and also to instance.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"regexp"
 	"strconv"
 	"strings"
@@ -90,6 +91,33 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 		return 0, 0, err
 	}
 	return diskSize, diskUsage, nil
+}
+
+func EnsureSSHKeyPresentInTheCluster(ocConfig oc.Config, sshPublicKeyPath string) error {
+	sshPublicKeyByte, err := ioutil.ReadFile(sshPublicKeyPath)
+	if err != nil {
+		return err
+	}
+	sshPublicKey := strings.TrimRight(string(sshPublicKeyByte), "\r\n")
+	if err := WaitForOpenshiftResource(ocConfig, "machineconfigs"); err != nil {
+		return err
+	}
+	stdout, _, err := ocConfig.RunOcCommand("get", "machineconfigs", "99-master-ssh", "-o", `jsonpath='{.spec.config.passwd.users[0].sshAuthorizedKeys[0]}'`)
+	if err != nil {
+		return err
+	}
+	if stdout == string(sshPublicKey) {
+		return nil
+	}
+	logging.Info("Updating SSH key to machine config resource...")
+	cmdArgs := []string{"patch", "machineconfig", "99-master-ssh", "-p",
+		fmt.Sprintf(`'{"spec": {"config": {"passwd": {"users": [{"name": "core", "sshAuthorizedKeys": ["%s"]}]}}}}'`, sshPublicKey),
+		"--type", "merge"}
+	_, stderr, err := ocConfig.RunOcCommand(cmdArgs...)
+	if err != nil {
+		return fmt.Errorf("Failed to update ssh key %v: %s", err, stderr)
+	}
+	return nil
 }
 
 func EnsurePullSecretPresentInTheCluster(ocConfig oc.Config, pullSec PullSecretLoader) error {
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -388,6 +388,13 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 		return nil, errors.Wrap(err, "Error waiting for apiserver")
 	}
 
+	// Remove this check when we have official 4.8 bundle
+	if strings.HasPrefix(crcBundleMetadata.GetOpenshiftVersion(), "4.8.") {
+		if err := cluster.EnsureSSHKeyPresentInTheCluster(ocConfig, constants.GetPublicKeyPath()); err != nil {
+			return nil, errors.Wrap(err, "Failed to update ssh public key to machine config")
+		}
+	}
+
 	if err := ensureProxyIsConfiguredInOpenShift(ocConfig, sshRunner, proxyConfig, instanceIP); err != nil {
 		return nil, errors.Wrap(err, "Failed to update cluster proxy configuration")
 	}
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -606,8 +606,9 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	}
 
 	logging.Info("Updating authorized keys...")
-	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys; chmod 644 /home/core/.ssh/authorized_keys", publicKey)
-	_, _, err = sshRunner.Run(cmd)
+	// CopyData uses sudo and we need to use it
+	// because of https://bugzilla.redhat.com/show_bug.cgi?id=1956739
+	err = sshRunner.CopyData(publicKey, "/home/core/.ssh/authorized_keys", 0644)
 	if err != nil {
 		return err
 	}
--- a/pkg/crc/machine/stop.go
+++ b/pkg/crc/machine/stop.go
@@ -1,7 +1,11 @@
 package machine
 
 import (
+	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/oc"
+	crcssh "github.com/code-ready/crc/pkg/crc/ssh"
+	"github.com/code-ready/crc/pkg/libmachine/host"
 	"github.com/code-ready/machine/libmachine/state"
 	"github.com/pkg/errors"
 )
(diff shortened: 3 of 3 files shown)


pkg/crc/machine/start.go
```diff
@@ -385,6 +385,13 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		return nil, errors.Wrap(err, "Error waiting for apiserver")
 	}
 
+	// Remove this check when we have official 4.8 bundle
+	if strings.HasPrefix(crcBundleMetadata.GetOpenshiftVersion(), "4.8.") {
+		if err := cluster.EnsureSSHKeyPresentInTheCluster(ocConfig, constants.GetPublicKeyPath()); err != nil {
```
guillaumerose, line 390: I guess the worst case is:

* VM starts,
* crc updates the ssh key manually to the new one
* kubelet starts and starts the MCO
* MCO updates the ssh key to the old one
* crc updates the ssh key config map
* MCO updates the ssh key to the new one

Could this happen? MCO is slower to start than the apiserver for sure.

praveenkumar: @guillaumerose That's the reason we do it as soon as apiserver is responding, in this case MCO will take the latest config when it start to render it. What you described only possible if MCO started with old config. I am hoping this shouldn't happen at all

anjannath: > Should the MCO be scaled down to 0 by snc, then crc updates all the resources and scale up to 1 ?
I think also at the pull secret. Could it trigger a restart of kubelet if MCO goes too fast ?

This feels like more safer to do..

guillaumerose: For ssh key, it is fine I guess but for the pull secret it is more an issue. 
If the MCO triggers a restart of the kubelet because it detects a change of pull secret, it will add extra start time.

praveenkumar:
/hold

Since we need to revert the MCO changes on snc side.

pkg/crc/machine/stop.go
```diff
@@ -15,6 +18,27 @@ func (client *client) Stop() (state.State, error) {
...
 	}
 
+	// L#21-37 should be removed after https://bugzilla.redhat.com/show_bug.cgi?id=1965992
+	// is fixed. We should also ignore the errors and just make it part of debug logs because
+	// if user start crc but cluster didn't start properly then we shouldn't block user to
+	// delete the crc instance.
+	logging.Info("Deleting the pods from openshift-machine-config-operator namespace")
+	instanceIP, err := getIP(host, client.useVSock())
+	if err != nil {
+		logging.Debugf("Error getting the IP: %v", err)
```
guillaumerose, line 28: return in case of error

pkg/crc/machine/stop.go
```diff
@@ -15,6 +18,27 @@ func (client *client) Stop() (state.State, error) {
 		return state.Error, errors.Wrap(err, "Cannot load machine")
 	}
 
+	// L#21-37 should be removed after https://bugzilla.redhat.com/show_bug.cgi?id=1965992
```
guillaumerose, line 21: Maybe extract this block in a function?

guillaumerose: Also maybe we need to do this only for 4.8 bundle. Add `if strings.HasPrefix(crcBundleMetadata.GetOpenshiftVersion(), "4.8.") {` ?

pkg/crc/machine/start.go
guillaumerose: Let's try this. We will see how it goes 👍

pkg/crc/machine/stop.go
cfergeau: Maybe better as a helper in the `bundle` package then?
```
if crcBundleMetadata.CheckOpenShiftMajorVersion("4.8") {
...
}
```

praveenkumar: > Also maybe we need to do this only for 4.8 bundle. Add `if strings.HasPrefix(crcBundleMetadata.GetOpenshiftVersion(), "4.8.") {` ?

`stop` doesn't have any idea about which bundle used to start the cluster so may be use what version of `oc` client on the host which is always up to date with bundle?

pkg/crc/machine/generate_bundle.go
```diff
@@ -31,9 +31,13 @@ func (client *client) GenerateBundle() error {
 	// Stop the cluster
 	currentState, err := client.Stop()
 	if err != nil {
-		return err
+		logging.Debug("Forcefully powering off")
+		if err := client.PowerOff(); err != nil {
+			return err
+		}
```
cfergeau, line 37: `Force poweroff the VM if not gracefully shutdown` feels like something which deserves its own issue/more discussion. If I'm reading this right, crc can report a successful bundle generation after a forced shutdown? How do we know the image is not corrupted in such a situation? I'm not necessarily against an option to allow forced shutdowns, but I'm not sure we want to do this by default. Can this be extracted to a separate PR?

pkg/crc/cluster/cluster.go
```diff
@@ -173,14 +173,24 @@ func RemovePullSecretFromCluster(ocConfig oc.Config, sshRunner *ssh.Runner) erro
...
+	// in 4.7 we disabled mco and also deleted all the machineconfig/machineconfig pools.
+	// For 4.8 we don't disable mco and it does contain the machineconfigs.
+	cmd := "oc --kubeconfig=/opt/kubeconfig get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-master"
+	_, _, err = sshRunner.Run(cmd)
+	if err != nil {
+		return nil
+	}
+
+	cmd = "oc --kubeconfig=/opt/kubeconfig get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-master | head -n -1 | xargs -t oc --kubeconfig=/opt/kubeconfig delete"
+	_, stderr, err = sshRunner.Run(cmd)
```
cfergeau, line 187: Did you try to go through an `OcRunner`? As I said before, 4.8 fixes to bundle generation can probably go in a separate PR so that we can move forward more quickly with the main 4.8 changes.

pkg/crc/cluster/cluster.go
```diff
@@ -159,6 +159,11 @@ func RemovePullSecretFromCluster(ocConfig oc.Config, sshRunner *ssh.Runner) erro
 	if _, _, err := sshRunner.RunPrivileged("Removing user's pull secret", "rm", "-fr", vmPullSecretPath); err != nil {
 		return err
 	}
+	// Change the ownership of authorized_keys file
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1956739
+	if _, _, err := sshRunner.RunPrivileged("Change the ownership of authorized_keys for core user", "chown", "core.core", "/home/core/.ssh/authorized_keys"); err != nil {
+		return err
+	}
```
cfergeau, line 166: This does not belong in a method called `RemovedPullSecretFromCluster`

cfergeau: Seems like we could change `updateSSHKeyPair` to use `sudo` to cover this MCO unexpected behaviour, and not need any changes here? We'd have all the code dealing with `authorized_keys` in a single place, which should make things slightly easier if we rework it to use the MCO to do its changes.

pkg/crc/machine/generate_bundle.go
praveenkumar: @cfergeau https://github.com/code-ready/crc/issues/2468

pkg/crc/cluster/cluster.go
```diff
@@ -92,6 +93,33 @@ func GetRootPartitionUsage(sshRunner *ssh.Runner) (int64, int64, error) {
...
+	if stdout == string(sshPublicKey) {
+		return nil
+	}
+	logging.Info("Updating SSH key to machine config resource...")
+	cmdArgs := []string{"patch", "machineconfig", "99-master-ssh", "-p",
+		fmt.Sprintf(`'{"spec": {"config": {"passwd": {"users": [{"name": "core", "sshAuthorizedKeys": ["%s"]}]}}}}'`, sshPublicKey),
+		"--type", "merge"}
+	_, stderr, err := ocConfig.RunOcCommandPrivate(cmdArgs...)
```
cfergeau, line 116: I don't this needs to be private since this only lists public keys?

pkg/crc/machine/start.go
```diff
@@ -606,7 +606,9 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	}
 
 	logging.Info("Updating authorized keys...")
-	cmd := fmt.Sprintf("echo '%s' > /home/core/.ssh/authorized_keys; chmod 644 /home/core/.ssh/authorized_keys", publicKey)
+	// We are using sudo here because of https://bugzilla.redhat.com/show_bug.cgi?id=1956739
+	// MCO change this file permission to root so better to update it with sudo.
+	cmd := fmt.Sprintf("echo '%s' | sudo tee /home/core/.ssh/authorized_keys; sudo chmod 644 /home/core/.ssh/authorized_keys", publicKey)
```
cfergeau, line 611: I think this can become `sshRunner.CopyData(publicKey, "/home/core/.ssh/authorized_keys", 0644)`

pkg/crc/machine/stop.go
```diff
@@ -25,3 +31,26 @@ func (client *client) Stop() (state.State, error) {
 	}
 	return host.Driver.GetState()
 }
+
+// This should be removed after https://bugzilla.redhat.com/show_bug.cgi?id=1965992
+// is fixed. We should also ignore the openshift specific errors because stop
+// operation shouldn't depend on the openshift side.
```
cfergeau, line 37: I'd mention somewhere that without this workaround, shutting down the VM takes more than 8 minutes.

cfergeau (approved):

Issue #2332, opened 2021-05-11
[BUG] Error during accessing the cluster when user switch/create a project and delete->start crc


praveenkumar:
```
$ crc start
$ crc whoami
kubeadmin
$ oc config get-contexts
$ oc config get-contexts
CURRENT   NAME                                                               CLUSTER                AUTHINFO                         NAMESPACE
*         crc-admin                                                          api.crc.testing:6443   kubeadmin                        default
          crc-developer                                                      api.crc.testing:6443   developer                        default
$ oc config get-clusters
NAME
api.crc.testing:6443

$ oc project openshift-machine-config-operator
$ oc config get-contexts
CURRENT   NAME                                                               CLUSTER                AUTHINFO                         NAMESPACE
          crc-admin                                                          api.crc.testing:6443   kubeadmin                        default
          crc-developer                                                      api.crc.testing:6443   developer                        default
*         openshift-machine-config-operator/api-crc-testing:6443/kubeadmin   api-crc-testing:6443   kubeadmin/api-crc-testing:6443   openshift-machine-config-operator
$ oc config get-clusters
NAME
api-crc-testing:6443
api.crc.testing:6443

$ crc delete && crc start
$ oc whoami
Unable to connect to the server: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kube-apiserver-lb-signer")
```

This is happening because when we add cluster name to kubeconfig we use hostname as it is but oc does https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/helpers/kubeconfig/smart_merge.go#L21 way, it first replace `.` with `-` and then update it. So now we have 2 different cluster which have same certificate-authority but after delete and start the `api.crc.testing` is only updated and not the `api-crc-testing` one which cause this issue. we should better to use same convention to avoid this different cluster name.

PR #2333, opened 2021-05-11
Update the cluster name as per oc norms

praveenkumar:
This patch fixes following issue
```
$ crc start
$ crc whoami
kubeadmin
$ oc config get-contexts
$ oc config get-contexts
CURRENT   NAME
CLUSTER                AUTHINFO                         NAMESPACE
*         crc-admin
	  api.crc.testing:6443   kubeadmin
default
          crc-developer
api.crc.testing:6443   developer                        default
$ oc config get-clusters
NAME
api.crc.testing:6443

$ oc project openshift-machine-config-operator
$ oc config get-contexts
CURRENT   NAME
CLUSTER                AUTHINFO                         NAMESPACE
          crc-admin
api.crc.testing:6443   kubeadmin                        default
          crc-developer
api.crc.testing:6443   developer                        default
*
openshift-machine-config-operator/api-crc-testing:6443/kubeadmin
api-crc-testing:6443   kubeadmin/api-crc-testing:6443
openshift-machine-config-operator
$ oc config get-clusters
NAME
api-crc-testing:6443
api.crc.testing:6443

$ crc delete && crc start
$ oc whoami
Unable to connect to the server: x509: certificate signed by unknown
authority (possibly because of "crypto/rsa: verification error" while
trying to verify candidate authority certificate
"kube-apiserver-lb-signer")
```

Which is happen 2 different cluster having same
certificate-authority but after delete and start the `api.crc.testing`
server is only updated and not the `api-crc-testing` one which cause this issue.

fixes: #2332

### DIFF
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -16,6 +16,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/machine/types"
 	"github.com/openshift/oc/pkg/helpers/tokencmd"
+	"k8s.io/apimachinery/third_party/forked/golang/netutil"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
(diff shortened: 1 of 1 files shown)


gbraad (approved):

praveenkumar:
> I don't even understand why oc does all of this, how this works/fits together/... :-/ Isn't this also an `oc` bug that it fails to detect there's already a context for the cluster? And will this work fine on upgrades when moving from one version with the old name to a version with the new name?

@cfergeau it is not oc bug because the way oc create the cluster name on the kubeconfig is different than what we are creating, so here we are trying to use same logic to create the cluster name and that will make sure if context is already there it should be used.

cfergeau:
> > I don't even understand why oc does all of this, how this works/fits together/... :-/ Isn't this also an `oc` bug that it fails to detect there's already a context for the cluster? And will this work fine on upgrades when moving from one version with the old name to a version with the new name?
> 
> @cfergeau it is not oc bug because the way oc create the cluster name on the kubeconfig is different than what we are creating, so here we are trying to use same logic to create the cluster name and that will make sure if context is already there it should be used.

But it seems oc/kubectl has some logic to somehow choose which context to use automatically? And the context must contain the cluster hostname/ip somewhere? So  `oc` could try harder not to recreate conflicting contexts when a valid one already exists?

praveenkumar:
> But it seems oc/kubectl has some logic to somehow choose which context to use automatically? And the context must contain the cluster hostname/ip somewhere? So `oc` could try harder not to recreate conflicting contexts when a valid one already exists?

@cfergeau yes it does contain the hostname/ip but context is not map with hostname/ip but with the name and the name we are generating was different than the oc/kubectl generate.

cfergeau:
I understand there are names and IP/hostnames. What I don't get is how `oc` decides to use a context/switch to a different one? It just picks the latest created context unless instructed otherwise?

PR #2334, opened 2021-05-11
http api: return only the error instead of a full object

guillaumerose:
This PR cleans up the http a little bit. It forces consumers to not rely on success field but use the status code. 

When an error occurs, 
* returns a 500 error code (to be refine when the binary protocol will be gone)
* returns error text (instead of a full object)

All API objects contains success and eventually error. 

Before

``` 
$ curl  --unix-socket ~/.crc/crc-http.sock http:/unix/api/start -v | jq .
< HTTP/1.1 200 OK
{
  "Name": "crc",
  "Status": "",
  "Error": "Bundle 'crc_libvirt_4.7.8.crcbundle' was requested, but the existing VM is using 'crc_libvirt_4.7.7.crcbundle'",
  "ClusterConfig": {
    "ClusterCACert": "",
    "KubeConfig": "",
    "KubeAdminPass": "",
    "ClusterAPI": "",
    "WebConsoleURL": "",
    "ProxyConfig": null
  },
  "KubeletStarted": false
}
```

After

``` 
$ curl  --unix-socket ~/.crc/crc-http.sock http:/unix/api/start -v | jq .
< HTTP/1.1 500 Internal Server Error
Bundle 'crc_libvirt_4.7.8.crcbundle' was requested, but the existing VM is using 'crc_libvirt_4.7.7.crcbundle.
```

### DIFF
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -68,7 +68,6 @@ func (h *Handler) Start(args json.RawMessage) string {
 		if err != nil {
 			startErr := &client.StartResult{
 				Success: false,
-				Name:    h.MachineClient.GetName(),
 				Error:   fmt.Sprintf("Incorrect arguments given: %s", err.Error()),
 			}
 			return encodeStructToJSON(startErr)
@@ -77,7 +76,6 @@ func (h *Handler) Start(args json.RawMessage) string {
 	if err := preflight.StartPreflightChecks(h.Config); err != nil {
 		startErr := &client.StartResult{
 			Success: false,
-			Name:    h.MachineClient.GetName(),
 			Error:   err.Error(),
 		}
 		return encodeStructToJSON(startErr)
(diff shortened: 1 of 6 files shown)


pkg/crc/api/handlers.go
```diff
@@ -68,7 +68,6 @@ func (h *Handler) Start(args json.RawMessage) string {
 		if err != nil {
 			startErr := &client.StartResult{
 				Success: false,
-				Name:    h.MachineClient.GetName(),
```
anjannath: We'll also need to update the structs on the trays..

anjannath (approved):

pkg/crc/api/handlers.go
guillaumerose: Yes I opened this one for macOS. https://github.com/code-ready/tray-macos/pull/108
For Windows, is there something needed ?

PR #2335, opened 2021-05-11
Check if running in WSL2 right after checking root

guillaumerose:
There is no point trying to install KVM, etc before checking for WSL2
installation.

Before:

$ crc setup
Your system is correctly setup for using CodeReady Containers

After:

$ crc setup
Your system is correctly setup for using CodeReady Containers

### DIFF
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -228,16 +228,13 @@ func removeVsockCrcSettings() error {
 func getAllPreflightChecks() []Check {
 	usingSystemdResolved := checkSystemdResolvedIsRunning()
 	checks := getPreflightChecksForDistro(distro(), network.SystemNetworkingMode, usingSystemdResolved == nil)
-	checks = append(checks, wsl2PreflightChecks)
 	checks = append(checks, vsockPreflightChecks)
 	return checks
 }
 
 func getPreflightChecks(_ bool, _ bool, networkMode network.Mode) []Check {
 	usingSystemdResolved := checkSystemdResolvedIsRunning()
-	checks := getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil)
-	checks = append(checks, wsl2PreflightChecks)
-	return checks
+	return getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil)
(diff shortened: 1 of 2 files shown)


cfergeau (approved):

pkg/crc/preflight/preflight_linux.go
```diff
@@ -228,16 +228,13 @@ func removeVsockCrcSettings() error {
...
 	checks = append(checks, vsockPreflightChecks)
 	return checks
 }
 
 func getPreflightChecks(_ bool, _ bool, networkMode network.Mode) []Check {
 	usingSystemdResolved := checkSystemdResolvedIsRunning()
-	checks := getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil)
-	checks = append(checks, wsl2PreflightChecks)
-	return checks
+	return getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil)
```
gbraad, line 237: Wouldn't the fix just be:
```golang
func getPreflightChecks(_ bool, _ bool, networkMode network.Mode) []Check {
	usingSystemdResolved := checkSystemdResolvedIsRunning()
	checks := wsl2PreflightChecks
	checks = append(checks, getPreflightChecksForDistro(distro(), networkMode, usingSystemdResolved == nil))
	return checks
}
```

gbraad (changes requested):

pkg/crc/preflight/preflight_linux.go
```diff
@@ -260,6 +257,7 @@ func getNetworkChecks(networkMode network.Mode, systemdResolved bool) []Check {
 func getPreflightChecksForDistro(distro *linux.OsRelease, networkMode network.Mode, systemdResolved bool) []Check {
 	var checks []Check
 	checks = append(checks, nonWinPreflightChecks[:]...)
+	checks = append(checks, wsl2PreflightChecks)
```
gbraad, line 260: This is confusing in my opinion... This is not a 'distro' check.
The naming of the preflights is very messy...

and this is not adjusted in the `GetAllPreflightChecks`?

gbraad: I do appreciate the attempt there ;-)

PR #2350, opened 2021-05-18
Always validate pull secret before storing it in keyring

cfergeau:
Better to do the validation once before saving the pull secret, rather
than expecting all callers of StoreInKeyring to do the validation
themselves.

## Proposed changes

This should not have any user visible change, except for a changed status code in the HTTP API when the pull secret is invalid.

### DIFF
--- a/pkg/crc/api/api_http.go
+++ b/pkg/crc/api/api_http.go
@@ -13,7 +13,6 @@ import (
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine"
-	"github.com/code-ready/crc/pkg/crc/validation"
 )
 
 func NewMux(config crcConfig.Storage, machine machine.Client, logger Logger, telemetry Telemetry) http.Handler {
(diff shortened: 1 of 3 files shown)


anjannath (approved):

pkg/crc/api/api_http.go
```diff
@@ -101,10 +100,6 @@ func pullSecretHandler(config crcConfig.Storage) func(w http.ResponseWriter, r *
 				http.Error(w, err.Error(), http.StatusBadRequest)
 				return
 			}
-			if err := validation.ImagePullSecret(string(bin)); err != nil {
```
anjannath, line 104: I think we should keep the check here as well, this differentiate between an invalid pull secret and failing to save the store in the keyring. The tray also depends on this to validate the pull secret and let the user know if the pull secret it invalid

anjannath: Tested, no issues in the tray, the validation is not depending on the returned status code

anjannath (approved):

pkg/crc/api/api_http.go
cfergeau: Thanks for the test, the returned error code seemed fairly arbitrary, so I was indeed hoping this would have no impact on the tray.

Issue #2346, opened 2021-05-18
[Feature] Allow config option to assign the kubeadmin/developer passwords on startup


gbraad:
Due to security concerns (see #1391) we had to implement the option to assign a random password, however this is not ideal when it comes to a mental model. While the `start` command now summarized the password(s) used (see #1240) it still be suggested to allow a configuration option to set a password the user 'knows'(wants).

cfergeau:
Could kube contexts be a way of addressing this, at least for the 'cmdline automation' use case?

praveenkumar:
> Could kube contexts be a way of addressing this, at least for the 'cmdline automation' use case?

@cfergeau kube contexts have token which expire in 24 hours, if that was not the case then it would be the best way to deal it.

cfergeau:
> kube contexts have token which expire in 24 hours

I keep forgetting that :relaxed: , thanks for the reminder!

PR #2355, opened 2021-05-19
Add config option to set kubeadmin password for cluster

praveenkumar:
This patch will provides an option to set kubeadmin password for
the user.
```
$ crc config set kubeadmin-password mypassword
$ crc start
[...]

Log in as administrator:
  Username: kubeadmin
  Password: mypassword
```

fixes: #2346

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -66,12 +66,13 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 	checkIfNewVersionAvailable(config.Get(crcConfig.DisableUpdateCheck).AsBool())
 
 	startConfig := types.StartConfig{
-		BundlePath: config.Get(crcConfig.Bundle).AsString(),
-		Memory:     config.Get(crcConfig.Memory).AsInt(),
-		DiskSize:   config.Get(crcConfig.DiskSize).AsInt(),
-		CPUs:       config.Get(crcConfig.CPUs).AsInt(),
-		NameServer: config.Get(crcConfig.NameServer).AsString(),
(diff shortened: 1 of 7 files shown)


guillaumerose:
/lgtm, just not sure about the config. key.

I don't know if it's a good idea to have this new configuration key. 
For sure, we can't use only the keychain because Linux, so we have a file on disk with the password. Now, with `crc` command line, I can directly get it.
User might be tempted to use $(crc config get kubeadmin-password) in scripts instead of $(crc console --credentials | jq ...)

praveenkumar:
> User might be tempted to use $(crc config get kubeadmin-password) in scripts instead of $(crc console --credentials | jq ...)

@guillaumerose if the user try `config get` in the scripts it is going to fail if the password is not set already `$ crc console --credentials` always be the right choice and we might publishing it more aggressive, like in the start template add this details?
```
You can also view the password for the developer and kubeadmin users by running crc console --credentials.
```

guillaumerose (approved):

guillaumerose:
I just realize the password is not loaded from the config when starting the cluster from the API. If you can do it in this PR or I can do it in a follow up. As you want.

praveenkumar:
> I just realize the password is not loaded from the config when starting the cluster from the API. If you can do it in this PR or I can do it in a follow up. As you want.

Yes, I will try to do it in same PR, this way I can have a code walk through of api :)

praveenkumar:
> I just realize the password is not loaded from the config when starting the cluster from the API. If you can do it in this PR or I can do it in a follow up. As you want.

Fixed  for api side also.

PR #2356, opened 2021-05-19
Faster crc status

guillaumerose:
`crc status` takes around 400-600ms on my machine. The macOS tray polls it every 5s. 
This PR makes crc status faster. 

It caches `df -h` output. The daemon no longer opens a ssh connection to get these numbers every 5s. The cache lasts 1 minute.
For the openshift status, the json is retrieved directly via the API server without using oc in ssh. The connection is setup to avoid proxies and use directly the VM IP.

For the CLI, crc status takes now around 200-300ms. From the daemon, 200ms without size cache, 50ms with cache.

`crc status` still hang from time to time (the bug was present before this PR - it happens mostly during crc start or stop when the connection is not yet ready). This is due to the ssh client not correctly handling timeouts. I will fix it in an other PR.


cfergeau:
Just some high-level questions before I look more closely (not asking for changes)

> It caches df -h output. The daemon no longer opens a ssh connection to get these numbers every 5s. The cache lasts 1 minute.

An alternative could be for the daemon to keep a long lasting ssh connection, so that we don't need to recreate it for every status call? Or would it be too slow anyway?

> The macOS tray polls it every 5s. This PR makes crc status faster.

But does it need the full status all the time? I don't think disk size is required for each of these polls, so there could be a 'fast status' command and a slower one?

guillaumerose:
> An alternative could be for the daemon to keep a long lasting ssh connection, so that we don't need to recreate it for every status call?

Yes, I tried this but the default Golang implementation doesn't have ssh keepalive and we have to do it by ourselves. 
Something like https://github.com/shazow/ssh-chat/pull/147/files#diff-b9630fed085ae953f0961a2b245323823c700efdbb43fe85b03bc229c1d8a5bcR83
Then we have to deal with reconnection, etc. Not really easy!

> Or would it be too slow anyway?

I think the second call to df is pretty fast. 

> But does it need the full status all the time? I don't think disk size is required for each of these polls, so there could be a 'fast status' command and a slower one?

Yes it can be a solution. The approach in this PR will also work if we change `crc status` to call the daemon.

pkg/crc/cluster/clusteroperator.go
```diff
@@ -123,10 +118,6 @@ func getStatus(ocConfig oc.Config, selector []string) (*Status, error) {
 					logging.Debug(c.ObjectMeta.Name, " operator is disabled, Reason: ", con.Reason)
 					cs.Disabled = true
 				}
-			case "ManagementStateDegraded": // only for the network operator
-				continue
-			default:
-				logging.Debugf("Unexpected operator status for %s: %s", c.ObjectMeta.Name, con.Type)
```
cfergeau, line 129: Can we keep the `default`?

pkg/crc/machine/start.go
```diff
@@ -443,7 +443,11 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 
 	logging.Info("Starting OpenShift cluster... [waiting for the cluster to stabilize]")
 
-	if err := cluster.WaitForClusterStable(ctx, ocConfig); err != nil {
+	lister, err := lister(instanceIP, crcBundleMetadata)
```
cfergeau, line 446: wondering if `kubeclient` or such would be a better name than `lister`? I'm kind of confused that you create something called `lister` here, and then have to call `lister.ConfigV1().ClusterOperators()` in order to get a `lister` which is usable by `cluster.WaitForClusterStable`.
Maybe `lister` should live in the `cluster` package with just `cluster.WaitForClusterStable(ctx, instanceIP, crcBundleMetadata)`?

pkg/crc/cluster/status.go
```diff
@@ -24,7 +23,7 @@ func WaitForClusterStable(ctx context.Context, ocConfig oc.Config) error {
 	var count int // holds num of consecutive matches
 
 	for i := 0; i < retryCount; i++ {
-		status, err := GetClusterOperatorsStatus(ocConfig)
+		status, err := GetClusterOperatorsStatus(lister)
```
cfergeau, line 26: Should this be passed `ctx` as `getStatus` needs a context too? (still not quite sure familiar with contexts :)

pkg/crc/machine/status.go
```diff
@@ -73,8 +67,13 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
 	}, nil
 }
 
-func (client *client) getDiskDetails(sshRunner *crcssh.Runner) (int64, int64) {
+func (client *client) getDiskDetails(ip string, bundle *bundle.CrcBundleInfo) (int64, int64) {
 	disk, err, _ := client.diskDetails.Memoize("disks", func() (interface{}, error) {
+		sshRunner, err := crcssh.CreateRunner(ip, getSSHPort(client.useVSock()), constants.GetPrivateKeyPath(), constants.GetRsaPrivateKeyPath(), bundle.GetSSHKeyPath())
+		if err != nil {
```
cfergeau, line 73: While looking at this code, without looking at more code, I'm not sure what the benefit is. I assume `CreateRunner` creates the TCP connection to the VM, so moving it here and running it only when needed saves some time?
I think it would be fine to squash it in the first commit.

cfergeau:
`Get the cluster operator list via the http API and not ssh` makes the crc binary 30% to 50% bigger :-/ Goes from 38MB to 52MB

pkg/crc/cluster/clusteroperator.go
guillaumerose: outch yes, a mistake.

pkg/crc/cluster/status.go
guillaumerose: yes added!

pkg/crc/machine/status.go
guillaumerose: Yes I squashed this.
The connection is created only when used, so it was just esthetic.

guillaumerose:
/retest
retest this please

pkg/crc/machine/start.go
guillaumerose: yes done

cfergeau (approved):
Looks good to me, just slightly annoyed by the binary size change, but maybe that's just me :)

guillaumerose:
retest this please
/retest

PR #2357, opened 2021-05-19
Configure ssh connection to timeout after 30s. Use 5s timeout for crc status.

guillaumerose:
I am not really a big fan of the solution but at least "it works". `crc status` should not hang anymore.

If `RunWithTimeout` times out, the next call will fail with "connection closed" error. I believe we should try to reconnect ?
Also, I realize the code with conn/sshClient is not really thread safe.. I am open to all suggestions :)

This loop below triggers the timeout (with zsh):
```
$ repeat 1000 time crc status                                                  
CRC VM:          Running
OpenShift:       Running (v4.7.11)
Disk Usage:      12.28GB of 32.74GB (Inside the CRC VM)
Cache Usage:     47.73GB
Cache Directory: /home/guillaumerose/.crc/cache
crc status  0.09s user 0.02s system 18% cpu 0.586 total
CRC VM:          Running
OpenShift:       Running (v4.7.11)
Disk Usage:      12.28GB of 32.74GB (Inside the CRC VM)
Cache Usage:     47.73GB
Cache Directory: /home/guillaumerose/.crc/cache
crc status  0.09s user 0.02s system 17% cpu 0.617 total
CRC VM:          Running
OpenShift:       Running (v4.7.11)
Disk Usage:      12.28GB of 32.74GB (Inside the CRC VM)
Cache Usage:     47.73GB
Cache Directory: /home/guillaumerose/.crc/cache
crc status  0.08s user 0.02s system 18% cpu 0.585 total
CRC VM:          Running
OpenShift:       Running (v4.7.11)
Disk Usage:      12.28GB of 32.74GB (Inside the CRC VM)
Cache Usage:     47.73GB
Cache Directory: /home/guillaumerose/.crc/cache
crc status  0.07s user 0.02s system 26% cpu 0.345 total
CRC VM:          Running
OpenShift:       Unreachable (v4.7.11)
Cache Usage:     47.73GB
Cache Directory: /home/guillaumerose/.crc/cache
crc status  0.08s user 0.02s system 1% cpu 5.122 total
CRC VM:          Running
OpenShift:       Unreachable (v4.7.11)
Cache Usage:     47.73GB
Cache Directory: /home/guillaumerose/.crc/cache
crc status  0.08s user 0.02s system 1% cpu 5.133 total

```

### DIFF
--- a/pkg/crc/ssh/client.go
+++ b/pkg/crc/ssh/client.go
@@ -13,8 +13,11 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
+const dialTimeout = 10 * time.Second
+
 type Client interface {
 	Run(command string) ([]byte, []byte, error)
+	RunWithTimeout(command string, timeout time.Duration) ([]byte, []byte, error)
 	Close()
 }
 
@@ -68,11 +71,11 @@ func clientConfig(user string, keys []string) (*ssh.ClientConfig, error) {
 		Auth: []ssh.AuthMethod{ssh.PublicKeys(privateKeys...)},
 		// #nosec G106
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
-		Timeout:         10 * time.Second,
+		Timeout:         dialTimeout,
 	}, nil
 }
 
-func (client *NativeClient) session() (*ssh.Session, error) {
+func (client *NativeClient) session(timeout time.Duration) (*ssh.Session, error) {
 	if client.sshClient == nil {
 		var err error
 		config, err := clientConfig(client.User, client.Keys)
@@ -84,12 +87,28 @@ func (client *NativeClient) session() (*ssh.Session, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		if err := client.conn.SetDeadline(time.Now().Add(timeout)); err != nil {
+			return nil, err
+		}
+		defer func() {
+			_ = client.conn.SetDeadline(time.Time{})
+		}()
+
 		c, chans, reqs, err := ssh.NewClientConn(client.conn, addr, config)
 		if err != nil {
 			return nil, err
 		}
 		client.sshClient = ssh.NewClient(c, chans, reqs)
 	}
+
+	if err := client.conn.SetDeadline(time.Now().Add(timeout)); err != nil {
+		return nil, err
+	}
+	defer func() {
+		_ = client.conn.SetDeadline(time.Time{})
+	}()
+
 	session, err := client.sshClient.NewSession()
 	if err != nil {
 		return nil, err
@@ -97,12 +116,19 @@ func (client *NativeClient) session() (*ssh.Session, error) {
 	return session, err
 }
 
-func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
-	session, err := client.session()
+func (client *NativeClient) RunWithTimeout(command string, timeout time.Duration) ([]byte, []byte, error) {
+	session, err := client.session(timeout)
 	if err != nil {
 		return nil, nil, err
 	}
 	defer session.Close()
+
+	if err := client.conn.SetDeadline(time.Now().Add(timeout)); err != nil {
+		return nil, nil, err
+	}
+	defer func() {
+		_ = client.conn.SetDeadline(time.Time{})
+	}()
 
 	var (
 		stdout bytes.Buffer
@@ -114,6 +140,10 @@ func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
 	err = session.Run(command)
 
 	return stdout.Bytes(), stderr.Bytes(), err
+}
+
+func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
+	return client.RunWithTimeout(command, 2*time.Minute)
 }
 
 func (client *NativeClient) Close() {
--- a/pkg/crc/ssh/client.go
+++ b/pkg/crc/ssh/client.go
@@ -25,6 +25,7 @@ type NativeClient struct {
 	Keys     []string
 
 	sshClient *ssh.Client
+	conn      net.Conn
 }
 
 func NewClient(user string, host string, port int, keys ...string) (Client, error) {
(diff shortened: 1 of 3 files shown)


pkg/crc/ssh/client.go
```diff
@@ -24,7 +24,7 @@ type NativeClient struct {
 	Port     int
 	Keys     []string
 
-	conn *ssh.Client
+	sshClient *ssh.Client
```
cfergeau, line 30: not sure which name is best to be honest, having something named `sshClient` in a type defined in `ssh/client.go` is imo not great.
Edit: I see there's another member named `conn` in the next commit, always useful to explain why you are doing something in the log to save some review time, and avoid this kind of comments ;)

cfergeau (commented):
Some 'why' in the commit log would definitely be useful to have, by just reading git history one will have no clue why this was done.
My main concern is that this seems to be papering over the root cause of this issue, the ssh hang will still happen, but it will 'only' last 5s/30s? Do you know why this happens?

pkg/crc/ssh/client.go
```diff
@@ -68,11 +71,11 @@ func clientConfig(user string, keys []string) (*ssh.ClientConfig, error) {
 		Auth: []ssh.AuthMethod{ssh.PublicKeys(privateKeys...)},
 		// #nosec G106
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
-		Timeout:         10 * time.Second,
+		Timeout:         dialTimeout,
```
cfergeau, line 74: this `dialTimeout` is not enough to avoid the issue, the connection is established and then it hangs?

guillaumerose:
The hang I face is apparently due to to dbus.
If I flood the VM with lot of `crc status`, at some point, it hangs on getting a `NewSession()` and I can see in syslog: `Failed to connect to API bus: Connection refused` couple of times.

It should be better in the next RHEL.

cfergeau:
> The hang I face is apparently due to to dbus.
> If I flood the VM with lot of `crc status`, at some point, it hangs on getting a `NewSession()` and I can see in syslog: `Failed to connect to API bus: Connection refused` couple of times.
> 
> It should be better in the next RHEL.

Ah right, it's that issue, thanks for the reminder! Have you tried issuing a `systemctl mask user@.service` in the VM as the `core` user as suggested in https://access.redhat.com/solutions/5211801 ?

guillaumerose:
I just tried the command, it works better but.. it breaks `podman-remote`. I guess the socket activation of podman is blocked by the mask :/

pkg/crc/ssh/client.go
cfergeau: Still wondering about this :)

pkg/crc/ssh/client.go
```diff
@@ -84,26 +87,49 @@ func (client *NativeClient) session() (*ssh.Session, error) {
...
 			return nil, err
 		}
+
+		if err := client.conn.SetDeadline(time.Now().Add(timeout)); err != nil {
+			return nil, err
+		}
+		defer func() {
+			_ = client.conn.SetDeadline(time.Time{})
+		}()
+
```
cfergeau, line 97: This does not need to be duplicated here as you are doing it after the end of the block anyway? Or can this particular deadline be triggered as part of `NewClientConn`?
Could we only call  `SetDeadline` once when creating `client.sshClient`? If a different timeout is needed, `session` can be called again with the new timeout value?

pkg/crc/ssh/client.go
```diff
@@ -84,26 +87,49 @@ func (client *NativeClient) session() (*ssh.Session, error) {
...
 	}
 	defer session.Close()
 
+	if err := client.conn.SetDeadline(time.Now().Add(timeout)); err != nil {
+		return nil, nil, err
+	}
+	defer func() {
+		_ = client.conn.SetDeadline(time.Time{})
+	}()
+
```
cfergeau, line 132: `client.session()` just did this, didn't it?

pkg/crc/ssh/client.go
cfergeau: This code is confusing me, it's written as if we could call `session()` multiple times with different timeout values, and it will create the low-level connection if needed, or reuse an existing one, and set the timeout as expected in all cases. But the only user of this method just does `session := session(); session.Close()`, so we don't have a need for something so sophisticated?

cfergeau:
> I just tried the command, it works better but.. it breaks podman-remote. I guess the socket activation of podman is blocked by the mask :/

Ah too bad :-/ I guess we'll have to go with this code for now then, we can revert it when we have 8.4 based images, which should have a fix for this.

pkg/crc/ssh/client.go
cfergeau: > `client.session()` just did this, didn't it?

Ah no, `client.session()` did this, then it undid it, then we are doing it again :-/

guillaumerose: Yes the code is really ugly. I can't find a way to do it properly. I just added the block 

```
	if err := client.conn.SetDeadline(time.Now().Add(timeout)); err != nil {
		return nil, nil, err
	}
	defer func() {
		_ = client.conn.SetDeadline(time.Time{})
	}()
```

Before all interactions with the connection.

guillaumerose:
/retest
retest this please

pkg/crc/ssh/client.go
guillaumerose: If we add the SetDeadline only here, it wont timeout if in a middle of a Read or Write the user shutdowns the VM.

pkg/crc/ssh/client.go
gbraad: > does not seem to have any impact on the hang which this PR is trying to fix

what is it trying to fix? can you explain what steps you did you play this scenario?

pkg/crc/ssh/client.go
gbraad: > add a workaround for a podman issue which hopefully will be fixed when switching to OpenShift 4.8

Did that happen, or are we still waiting for an updated release?

guillaumerose:
It seems not to be a problem anymore. At least, I don't feel this is needed.

PR #2358, opened 2021-05-19
Replace mirror link to developer.redhat.com

praveenkumar:
As of now developer.redhat.com link need to require cookie so
we create a cookieJar and pass it to the http.Client

### DIFF
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -29,7 +29,7 @@ var (
 )
 
 const (
-	releaseInfoLink = "https://mirror.openshift.com/pub/openshift-v4/clients/crc/latest/release-info.json"
+	releaseInfoLink = "https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/latest/release-info.json"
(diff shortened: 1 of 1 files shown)


cfergeau:
I'd hold this until we get confirmation we don't have a choice and cookies are required :)

cfergeau (approved):
Looks good to me

cfergeau (changes requested):
Cookies are no longer required to fetch from developers.redhat.com

pkg/crc/version/version.go
```diff
@@ -29,7 +29,7 @@ var (
 )
 
 const (
-	releaseInfoLink = "https://mirror.openshift.com/pub/openshift-v4/clients/crc/latest/release-info.json"
+	releaseInfoLink = "https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/1.26.0/release-info.json"
```
cfergeau, line 32: This has to be a `latest` link, not `1.26.0`, otherwise we won't be able to detect the newer version

guillaumerose (approved):

PR #2359, opened 2021-05-20
Remove windows tray from experimental and enable it by default when using the msi to install

anjannath:
### Proposed changes

1. User installs crc using the msi
2. runs `crc setup`
3. tray should be enabled and auto-started after reboot

The tray cannot be enabled now using the embedded `crc` binary

msi to test can be found at: https://ci.appveyor.com/api/buildjobs/t2q5ot2ep91nempt/artifacts/out%2Fwindows-amd64%2Fcrc-windows-installer.zip

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -33,8 +33,8 @@ const (
 
 func RegisterSettings(cfg *Config) {
 	validateTrayAutostart := func(value interface{}) (bool, string) {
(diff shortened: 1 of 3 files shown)


anjannath:
Not sure if we should remove the `--enable-experimental-feature` flag and config option, its not used anywhere now, but we might need it in future

gbraad:
`crc setup` works as expected, but unable to find the autostart anywhere ... ?

Issue #1642, opened 2020-11-04
Use go-units for disk size/memory size


cfergeau:
At the moment disk sizes/memory sizes are passed as unitless numbers, one has to know disk size uses GiB and memory size uses MiB (or look at `crc start --help`. We should be adding support for https://github.com/docker/go-units to the option parsing code so that the unit can be specified to make this more user friendly.

rohanKanojia:
@cfergeau : I'm trying this on CRC v2.45.0 . I'm observing that we're showing disk size memory size not in unitless numbers:

```shell
~ : $ crc status
CRC VM:          Running
OpenShift:       Running (v4.17.7)
RAM Usage:       20.06GB of 32.68GB
Disk Usage:      1.713GB of 10.95GB (Inside the CRC VM)
Cache Usage:     25.71GB
Cache Directory: /home/rohan/.crc/cache
```

cfergeau:
This is about how the units are passed around in the code, for example `pkg/crc/machine/driver.go` has `func setMemory(host *host.Host, memorySize uint)`, it's not obvious which unit the memory size is in. This is only one example among many.
However I think I reached the conclusion that go-units was not a great fit for us as it tends to confuse units in 1000 multiples and in 1024 multiples.
`github.com/containers/common/pkg/strongunits' might be better to carry around integer values together with a unit.

PR #2365, opened 2021-05-23
[WIP] feat(set): make memory a string

georgettica:
This PR is still WIP because of https://github.com/georgettica/crc/blob/700b16ceef0343f6a46ef782d4464e8f3e2ce048/pkg/crc/config/config.go#L75-L87 which makes changing it a hassle..

if you could assist me in that endevour I would be glad to keep on pushing it

this is to make 15Gi to be used instead of the MiB converted value

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,4 +12,6 @@ linters:
     - gocritic
     - deadcode
     - misspell
-    - golint
+    # replacement to golint, but it seems to be very verbose
+    # this is why it's commented but the old one is removed
+    #- revive
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -170,6 +170,7 @@ func isDebugLog() bool {
 }
 
 func validateStartFlags() error {
+	fmt.Println("im here")
 	if err := validation.ValidateMemory(config.Get(crcConfig.Memory).AsInt()); err != nil {
 		return err
 	}
--- a/pkg/crc/config/validations.go
+++ b/pkg/crc/config/validations.go
@@ -7,6 +7,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/validation"
+	"github.com/docker/go-units"
 	"github.com/spf13/cast"
 )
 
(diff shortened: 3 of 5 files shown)


georgettica:
wrote my issue in title

I need some assistance, if you have the time

cfergeau:
There was work in progress related to this in https://github.com/code-ready/crc/issues/1642 but I never finished the PR :-/ The code was at https://github.com/cfergeau/crc/tree/gounits

georgettica:
Thanks! I will.take a look :)

georgettica:
Thanks! I will see if I can reuse your changes :)

cfergeau:
Let me know if you need any help, in particular I can rebase it to the latest master if there were too many changes in the mean time.
The biggest issue with this code was that I was not sure what to store in the config file (which format)

georgettica:
I am thinking of creating a new type for memory which will be a string under the hood

This way all my code still works and we get the method for cpu cores if we so choose

Also maybe in the error for the monitoring we can make the error show the exact command to run in `crc set memory ***`
(This is written from my phone and not fact checked with actual code/commands)

cfergeau:
New type is useful to add some clarity to the code, and getting help from the compiler with code not using the right type. This may require some additional  boilerplate here and there though. And keep in mind for some of the code, you'll need a numerical value. Feel free to push some wip PRs, or to contact me privately if you need to talk it through.

gbraad:
This might add complexity that does not provide a lot in maintainability, as the simplest approach would be to force the user to specify a unit and convert that input immediately to an internal representation as a MiB or MB.

cfergeau:
> This might add complexity that does not provide a lot in maintainability, as the simplest approach would be to force the user to specify a unit and convert that input immediately to an internal representation as a MiB or MB (and use it as throughout unless needed by the driver)

Internal MiB or MB representation is not great either in my opinion, as this leads to odd situations:
If stored internally as MB, `crc config set 16384MiB` won't be read back as exactly 16384MiB but as 16383MiB (assuming int use + rounding down). This in turns means the drivers won't get exactly what was specified even if they use `MiB` values. And if I remember correctly, we can't standardize on the 'right' unit between `MiB` and `MB` as Windows and linux use different ones.

georgettica:
I would say that you all raise good issues, and I didn't have time you touch it recently, but I would really like the error in the monitoring section to not cause confusion (I needed to reverify my conversions before applying the `crc set` and the error with monitoring enabled

I'll address them next week hopefully

georgettica:
:((
But may maybe I'll come back to it

PR #2380, opened 2021-05-26
Cleanup powershell code

guillaumerose:
* Use the same code to run powershell in the Hyper-V driver and in crc preflight checks
* Add same debug log for powershell as for RunWithDefaultLocale.
* Remove the progress bar for Hyper-V actions (yay! :tada:)

----

I did that because I found very complicated to add a preflight check for Windows as there is no debug log for powershell commands by default...


pkg/os/windows/powershell/powershell_windows.go
```diff
@@ -87,27 +87,18 @@ func ExecuteAsAdmin(reason, cmd string) (string, string, error) {
...
-	_, err = psFile.Write([]byte{0xef, 0xbb, 0xbf})
-	if err != nil {
-		return "", "", err
-	}
-	_, err = psFile.WriteString(scriptContent)
-	if err != nil {
+	filename := filepath.Join(tempDir, "runAsAdmin.ps1")
+
+	// #nosec G306
+	if err := ioutil.WriteFile(filename, append([]byte{0xef, 0xbb, 0xbf}, []byte(scriptContent)...), 0666); err != nil {
```
cfergeau, line 72: Why is it important to write it with a single call? This is less readable, and does not seem to bring much?

pkg/drivers/hyperv/powershell.go
```diff
@@ -12,22 +12,16 @@ import (
 	log "github.com/code-ready/crc/pkg/crc/logging"
 )
 
-var powershell string
```
cfergeau, line 15: It's some form of memoization, `LookPath` is not 100% free. I don't mind not caching it.

pkg/os/windows/powershell/powershell_windows.go
```diff
@@ -31,7 +31,7 @@ func IsAdmin() bool {
 }
 
 func Execute(args ...string) (string, string, error) {
-	args = append([]string{"-NoProfile", "-NonInteractive", "-ExecutionPolicy", "RemoteSigned", "-Command"}, args...)
+	args = append([]string{"-NoProfile", "-NonInteractive", "-ExecutionPolicy", "RemoteSigned", "-Command", "$ProgressPreference = 'SilentlyContinue';"}, args...)
```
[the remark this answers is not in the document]
guillaumerose: This progress bar: https://robertwray.co.uk/Media/Default/Blog/PowerShell/powershell_startvm.png

When running crc start, in the middle of the call, a progress bar appears for few seconds and then goes away.
It happens for Start-VM and Stop-VM at least.

pkg/os/windows/powershell/powershell_windows.go
guillaumerose: Not really important but I found this easier to read and it is less likely to be wrong (correctly call Close, check errors, etc.).

pkg/os/windows/powershell/powershell_windows.go
```diff
@@ -4,46 +4,17 @@ import (
...
-	}
-	isAdminCmds = []string{
-		"$currentPrincipal = New-Object Security.Principal.WindowsPrincipal([Security.Principal.WindowsIdentity]::GetCurrent())",
-		"$currentPrincipal.IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)",
-	}
-
-	poshStdLocation = filepath.Join(os.Getenv("SYSTEMROOT"), "System32", "WindowsPowerShell", "v1.0", "powershell.exe")
-)
-
-func LocatePowerShell() string {
```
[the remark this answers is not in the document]
anjannath: I think this was only needed for windows 7 and got here from minishift, (or if the $PATH has been messed with), safe to remove now, since we don't support win7

pkg/drivers/hyperv/powershell_windows.go
```diff
@@ -2,40 +2,23 @@ package hyperv
...
-	log.Debugf("[executing ==>] : %v %v", powershell, strings.Join(args, " "))
-	var stdout bytes.Buffer
-	var stderr bytes.Buffer
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
-	err := cmd.Run()
-	log.Debugf("[stdout =====>] : %s", stdout.String())
-	log.Debugf("[stderr =====>] : %s", stderr.String())
-	return stdout.String(), err
+	stdout, _, err := powershell.Execute(args...)
```
gbraad, line 20: Just wondering before looking further; is everything run with Default Locale?

pkg/libmachine/load_unix.go
```diff
@@ -0,0 +1,37 @@
...
+	}, nil
+}
+
+func (api *Client) Load(name string) (*host.Host, error) {
+	h, err := api.Filestore.Load(name)
+	if err != nil {
+		return nil, err
+	}
+
+	d, err := api.clientDriverFactory.NewRPCClientDriver(h.DriverName, h.DriverPath, h.RawDriver)
```
gbraad, line 31: was curious why this was necessary, but looking at the `_unix` and followed with a new `_windows` file it is clear to break up and prevent this on another platform where it doesn't make sense.

pkg/os/windows/powershell/powershell_windows.go
```diff
@@ -61,7 +32,7 @@ func IsAdmin() bool {
 
 func Execute(args ...string) (string, string, error) {
 	args = append([]string{"-NoProfile", "-NonInteractive", "-ExecutionPolicy", "RemoteSigned", "-Command"}, args...)
-	cmd := exec.Command(LocatePowerShell(), args...) // #nosec G204
+	cmd := exec.Command("powershell.exe", args...) // #nosec G204
```
anjannath, line 35: why not use `exec.LookPath` here like in the `ExecuteAsAdmin` below

pkg/drivers/hyperv/powershell_windows.go
guillaumerose: I dont see mention of a locale.

pkg/os/windows/powershell/powershell_windows.go
guillaumerose: done!

anjannath (approved):
LGTM

pkg/drivers/hyperv/powershell_windows.go
```diff
@@ -19,17 +17,8 @@ var (
...
-	cmd := exec.Command("powershell.exe", args...)
-	log.Debugf("[executing ==>] : powershell.exe %v", strings.Join(args, " "))
-	var stdout bytes.Buffer
-	var stderr bytes.Buffer
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
-	err := cmd.Run()
-	log.Debugf("[stdout =====>] : %s", stdout.String())
-	log.Debugf("[stderr =====>] : %s", stderr.String())
-	return stdout.String(), err
```
anjannath: Sorry forgot to mention this earlier, we loose these debug statements, as `powershell.Execute` only logs when there's an error, and also the command getting executed will not be printed which is sometimes helpful to know in case of the commands being executed from the driver.

guillaumerose: Yes. Now logs are the same for the driver and crc code. I don't know which one is better. Should I log the command and stdout/stderr for all commands?

anjannath: Yes, i think log stdout/stderr for all commands, because this'll help us debug issues where there is no error but o/p is different then what we expect, but i can think of only the ipv6 scenario as of now where this helped..

guillaumerose: I added the command line log before each command. It is 100% like on Linux/macOS now.

``` 
logging.Debugf("Running '%s'", strings.Join(args, " "))
``` 

stdout/stderr are only log in case of failures.

PR #2382, opened 2021-05-26
RFC: Refactor getAllPreflightChecks/getPreflightChecks

cfergeau:
The implementation of these methods is currently a bit complicated, with the list of tests being manually generated by appending the 'right' tests depending on various parameters.
This PR tries to make this a bit easier, by adding labels to each `Check` instance in order to explicitly define when it should be used.

I tagged this as RFC to get feedback on this approach, maybe thet current way will be considered simpler, maybe this needs to be approached differently, ...

## Proposed changes

There should be no user-visible changes, exactly the same preflight as before should be run by crc setup/crc cleanup/crc start.

### DIFF
--- a/pkg/crc/preflight/labels.go
+++ b/pkg/crc/preflight/labels.go
@@ -1,0 +1,33 @@
+package preflight
+
+type LabelName uint32
+
+const (
+	Os LabelName = iota
+	NetworkMode
+
+	// Keep it last
+	lastLabelName // will be used in OS-specific go files to extend LabelName
+)
+
+type LabelValue uint32
+
+const (
+	// os
+	Darwin LabelValue = iota
+	Linux
+	Windows
+
+	// network mode
+	User
+	System
+
+	// Keep it last
+	lastLabelValue // will be used in OS-specific go files to extend LabelValue
+)
+
+var (
+	None = labels{}
+)
+
+type labels map[LabelName]LabelValue
--- a/pkg/crc/preflight/labels.go
+++ b/pkg/crc/preflight/labels.go
@@ -1,5 +1,11 @@
 package preflight
 
+import (
+	"runtime"
+
+	"github.com/code-ready/crc/pkg/crc/network"
+)
+
 type LabelName uint32
 
 const (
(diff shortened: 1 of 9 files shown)


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -29,7 +29,7 @@ var bundleCheck = Check{
 	flags:            SetupOnly,
 }
 
-var genericPreflightChecks = [...]Check{
+var genericPreflightChecks = []Check{
```
praveenkumar, line 34: This makes an array literal to slice but here it is good.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -244,26 +244,26 @@ func getNetworkChecks(networkMode network.Mode, systemdResolved bool) []Check {
...
+		checks = append(checks, dnsmasqPreflightChecks...)
 	}
 
 	return checks
 }
 
 func getPreflightChecksForDistro(distro *linux.OsRelease, networkMode network.Mode, systemdResolved bool) []Check {
 	var checks []Check
-	checks = append(checks, nonWinPreflightChecks[:]...)
+	checks = append(checks, nonWinPreflightChecks...)
```
praveenkumar, line 297: 👍🏼

pkg/crc/preflight/preflight.go
```diff
@@ -30,6 +77,8 @@ type Check struct {
 	flags              Flags
 	cleanupDescription string
 	cleanup            CleanUpFunc
+
+	labels labels
```
guillaumerose, line 81: Looks good! Maybe convert `flags` to a label also?

cfergeau: Yeah, that's next, I started here https://github.com/cfergeau/crc/commit/256eae0462dc1708cc7f0485289d1570ef5189f8 but I prefer to get this initial PR in first, and then improve on it.

pkg/crc/preflight/labels.go
```diff
@@ -0,0 +1,93 @@
...
+	Darwin LabelValue = iota
+	Linux
+	Windows
+
+	// network mode
+	User
+	System
+
+	// Keep it last
+	lastLabelValue // will be used in OS-specific go files to extend LabelName
```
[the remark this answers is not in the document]
cfergeau: Thanks, fixed.

PR #2383, opened 2021-05-27
Add method to list all bundles available

guillaumerose:
Sort them by OpenShift version.

It will be useful when upgrading crc: delete previous bundle, eventually
 be able to load it, etc.

### DIFF
--- a/pkg/crc/machine/bundle/repository_test.go
+++ b/pkg/crc/machine/bundle/repository_test.go
@@ -1,7 +1,6 @@
 package bundle
 
 import (
-	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -9,8 +8,6 @@ import (
 	"testing"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
-
-	"github.com/jinzhu/copier"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/crc/machine/bundle/repository_test.go
```diff
@@ -110,42 +107,56 @@ func TestVersionCheck(t *testing.T) {
...
+
+	ocBinDir, err := ioutil.TempDir("", "oc-bin-dir")
+	assert.NoError(t, err)
+	defer os.RemoveAll(ocBinDir)
 
-func createDummyBundleContent(t *testing.T, bundlePath string, version string) {
-	assert.NoError(t, os.Mkdir(bundlePath, 0755))
+	createDummyBundleContent(t, dir, "crc_libvirt_4.6.15", "1.0")
+	createDummyBundleContent(t, dir, "crc_libvirt_4.7.0", "1.0")
+	createDummyBundleContent(t, dir, "crc_libvirt_4.10.0", "1.0")
```
praveenkumar, line 137: As part of test can we also include the custom bundles and the nightly one to make sure it also parsed by semver?
```
$ ls ~/.crc/cache/
crc_libvirt_4.7.11_1621489482  crc_libvirt_4.8.0-0.nightly-2021-05-26-021757
```

guillaumerose: done. crc_libvirt_4.7.11_1621489482 doesnt work with semver. It is always first :/ 

```
"crc_libvirt_4.7.11_1621489482",
"crc_libvirt_4.10.0",
"crc_libvirt_4.8.0-0.nightly-2021-05-26-021757",
"crc_libvirt_4.7.0",
"crc_libvirt_4.6.15",
``` 

It is not particularly a problem: a custom bundle can be the first element of the list.

Issue #2374, opened 2021-05-25
[E2E] Fix UX experience with MacOS tray


adrianriobo:
On MacOS platform, with the new CRC version `1.27.0`. When starting the cluster from the tray without previous setting of the pull secret file, the tray is asking for the pull secret content instead of the pull secret file location (as it was on previous versions). 

This change breaks the previous UX experience, so this needs to be changed on the e2e UX Feature to accommodate the current UX flow.

guillaumerose:
Sorry, I totally forgot this.. :(

adrianriobo:
No problem fixed!

PR #2388, opened 2021-05-31
[E2E] Set pullsecret content when prompted from tray

adrianriobo:
This fixes **#2374**

Now instead of pullsecret file location the content is added when prompted during CRC start from tray on MacOS.


guillaumerose:
add to whitelist

PR #2390, opened 2021-05-31
start: Display help message when preflights fail

cfergeau:
The message instructing to run `crc setup` when one of the preflight
check failed when running `crc start` is no longer shown after commit
c1d9b48. This happens because the type of the error for preflight
failures changed. This commit adjusts the check to use the right type.

## Testing

1. Check that when one of the preflight checks is failing, crc setup fails (I used `virsh -c qemu:///system net-destroy crc && crc start`)
2. stderr contains a message instructing to run crc setup, which is not the case before this PR

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -21,10 +21,10 @@ import (
 	"github.com/code-ready/crc/pkg/crc/preflight"
 	"github.com/code-ready/crc/pkg/crc/validation"
 	crcversion "github.com/code-ready/crc/pkg/crc/version"
+	crcos "github.com/code-ready/crc/pkg/os"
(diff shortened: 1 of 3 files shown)


guillaumerose:
What if we add `Unwrap()` to CodeExitError ? Does it avoid an other if ?

cfergeau:
> What if we add `Unwrap()` to CodeExitError ? Does it avoid an other if ?

It's indeed nicer with an Unwrap(), thanks for the suggestion. However this means that for now we need to copy/paste `CodeExitError` implementation in crc, as this needs to get in upstream kubernetes, and then in the openshift fork before we can use it as an `import`.

cfergeau:
I've created https://github.com/kubernetes/kubernetes/pull/102478

guillaumerose (approved):

PR #2392, opened 2021-06-01
Create VM auth files right after driver.Create()

guillaumerose:
crc console command is available if the VM exists. 
When running crc start, there is a time frame where the VM exists but the kubeadmin password doesn't exist in the instance directory.

Now, kubeadmin password, ssh key and kubeconfig files are written right after the VM is created (driver.Create())
Right after, there is a call to `api.SetExists()`, which unlock functions like crc console.

Also,
* Refactor kubeadmin password generation to be done at creation time only
* htpasswd is changed only if needed by comparing the htpasswd file in the cluster and the wanted passwords.

### DIFF
--- a/pkg/crc/cluster/kubeadmin_password.go
+++ b/pkg/crc/cluster/kubeadmin_password.go
@@ -1,6 +1,8 @@
 package cluster
 
 import (
+	"bufio"
+	"bytes"
 	"crypto/rand"
 	"encoding/base64"
 	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/crc/cluster/kubeadmin_password.go
```diff
@@ -11,47 +13,56 @@ import (
...
+}
 
-	logging.Infof("Generating new password for the kubeadmin user")
-
-	kubeAdminPassword, err := GenerateRandomPasswordHash(23)
+// UpdateKubeAdminUserPassword updates the htpasswd secret
+func UpdateKubeAdminUserPassword(ocConfig oc.Config) error {
+	kubeAdminPassword, err := GetKubeadminPassword()
 	if err != nil {
 		return fmt.Errorf("Cannot generate the kubeadmin user password: %w", err)
```
praveenkumar, line 34: This error need to be updated, it is not any more issue with generation but reading the kubeconfig file with password.

pkg/crc/cluster/kubeadmin_password.go
```diff
@@ -11,47 +13,56 @@ import (
...
+	}
 
-	hashDeveloperPasswd, err := hashBcrypt("developer")
+	given, _, err := ocConfig.RunOcCommandPrivate("get", "secret", "htpass-secret", "-n", "openshift-config", "-o", `jsonpath="{.data.htpasswd}"`)
 	if err != nil {
 		return err
 	}
-
-	hashKubeAdminPasswd, err := hashBcrypt(kubeAdminPassword)
+	ok, err := compareHtpasswd(given, credentials)
```
praveenkumar, line 45: This check is going to break the behavior of cluster if user do something following.
- Start crc 
- Add more user to htpasswd using https://github.com/code-ready/crc/wiki/Add-another-user-to-cluster
- Stop the cluster
- Start again and here we again check len of password on cluster and what we have here which will not match
- We force again to update the password which doesn't have user's specific password.

pkg/crc/machine/start.go
```diff
@@ -524,6 +524,9 @@ func createHost(api libmachine.API, machineConfig config.MachineConfig) error {
 		return fmt.Errorf("Error in driver during machine creation: %s", err)
 	}
 
+	if err := cluster.GenerateKubeAdminUserPassword(); err != nil {
```
cfergeau, line 531: My understanding is that this kubeadmin password change will be fixing an error which was seen by some users? Can this be mentioned in the commit log?

pkg/crc/machine/start.go
```diff
@@ -524,6 +524,10 @@ func createHost(api libmachine.API, machineConfig config.MachineConfig) error {
 		return fmt.Errorf("Error in driver during machine creation: %s", err)
 	}
 
+	logging.Info("Generating new SSH Key pair...")
+	if err := crcssh.GenerateSSHKey(constants.GetPrivateKeyPath()); err != nil {
```
cfergeau, line 528: Can you add a few sentences to the commit log explaining why this ssh change is needed? As it is now, it's just code being moved around and a bit simpler code, not sure more than that is expected.

pkg/crc/machine/start.go
```diff
@@ -280,9 +281,6 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 	if err := updateSSHKeyPair(sshRunner); err != nil {
 		return nil, errors.Wrap(err, "Error updating public key")
 	}
-	if err := copyKubeconfig(client.name, crcBundleMetadata); err != nil {
-		return nil, errors.Wrap(err, "Error copying kubeconfig file")
-	}
```
cfergeau, line 284: We can do that, but why? Is this needed by one of the other commits? Is this fixing a bug? Is this just code movement? Would be easier to review if this was explicit :)

cfergeau (changes requested):

pkg/crc/cluster/kubeadmin_password.go
guillaumerose: Ok good point, I will try to write a merge function then.

praveenkumar: @guillaumerose Looking again to the code, why do you think we need that kind of complicated logic on crc side in `UpdateKubeAdminUserPassword` function? why is just checking the `kubeadmin_password` file in the `~/.crc/machine/crc/` directory is enough and return  in case it there. On openshift side it make sense because controller need to make sure the password updated in the resource should be updated but I think in crc this might not needed util you think current way have some breakage?

I do think generating this password file in early stage is a good improvement.

guillaumerose: The current mechanism differentiates how it works when creating the VM vs restarting the VM. 

```
if first run {
  change the password
} else {
  assume password is ok
} 
```

I think it is not a good thing to have lot of code only running for the first start: 
* Each crc start should ensure the cluster is a good shape. Right now, we don't know about the status of the auth at the 2nd run. Maybe the user broke it. With this change, we are sure the auth will work. 
* I did it already for the pull secret, cluster ID and I found it more robust against CTRL+C during the first start. Ideally I would like `machine.Start()` code to be idempotent (and remove the `if running { check nothing, assume it is ok, return }`.

pkg/crc/machine/start.go
guillaumerose: yes added.
```
Cannot create cluster configuration: Error reading kubeadmin password from bundle
```

pkg/crc/machine/start.go
guillaumerose: I added more context. 
Yes this is just a code move. The goal is to have the machine directory ready before the machine is marked as existing.

pkg/crc/machine/start.go
guillaumerose: I started with the kubeadmin-password but for completeness, I also did it for the ssh key and kubeconfig files.
The commit log should be better now.

cfergeau (commented):
Looks good to me, thanksfor the updated logs!

cfergeau (approved):

PR #2394, opened 2021-06-02
Display a more precise download link when a new version is available

guillaumerose:
Before:

``` 
WARN A new version (1.27.0) has been published on https://cloud.redhat.com/openshift/create/local 
```

After:

```
WARN A new version (1.27.0) has been published on https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/1.27.0/crc-linux-amd64.tar.xz 
```

The goal is to allow people to upgrade easily. Right now they have to login to the portal, etc to get the tarball. 

It is also to better control the download link:
* if someone downloads from mirror.openshift.com, we advertise the new link developers.redhat.com,
* display the link to the tarball or the installer for instance.

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/code-ready/crc/pkg/crc/cluster"
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/constants"
@@ -63,7 +64,9 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		return nil, err
 	}
 
-	checkIfNewVersionAvailable(config.Get(crcConfig.DisableUpdateCheck).AsBool())
+	if err := checkIfNewVersionAvailable(config.Get(crcConfig.DisableUpdateCheck).AsBool()); err != nil {
+		logging.Debugf("Unable to find out if a new version is available: %v", err)
+	}
 
 	startConfig := types.StartConfig{
 		BundlePath:        config.Get(crcConfig.Bundle).AsString(),
@@ -191,20 +194,42 @@ func validateStartFlags() error {
 	return nil
 }
 
-func checkIfNewVersionAvailable(noUpdateCheck bool) {
+func checkIfNewVersionAvailable(noUpdateCheck bool) error {
 	if noUpdateCheck {
-		return
+		return nil
 	}
-	isNewVersionAvailable, newVersion, err := crcversion.NewVersionAvailable()
+	isNewVersionAvailable, newVersion, link, err := newVersionAvailable()
 	if err != nil {
-		logging.Debugf("Unable to find out if a new version is available: %v", err)
-		return
+		return err
 	}
 	if isNewVersionAvailable {
-		logging.Warnf("A new version (%s) has been published on %s", newVersion, constants.CrcLandingPageURL)
-		return
+		logging.Warnf("A new version (%s) has been published on %s", newVersion, link)
+		return nil
 	}
 	logging.Debugf("No new version available. The latest version is %s", newVersion)
+	return nil
+}
+
+func newVersionAvailable() (bool, string, string, error) {
+	release, err := crcversion.GetCRCLatestVersionFromMirror()
+	if err != nil {
+		return false, "", "", err
+	}
+	currentVersion, err := semver.NewVersion(crcversion.GetCRCVersion())
+	if err != nil {
+		return false, "", "", err
+	}
+	if release.Version.CrcVersion == nil {
+		return false, "", "", errors.New("empty version")
+	}
+	return release.Version.CrcVersion.GreaterThan(currentVersion), release.Version.CrcVersion.String(), downloadLink(release), nil
+}
+
+func downloadLink(release *crcversion.CrcReleaseInfo) string {
+	if link, ok := release.Links[runtime.GOOS]; ok {
+		return link
+	}
+	return constants.CrcLandingPageURL
 }
 
 const startTemplate = `Started the OpenShift cluster.
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -83,7 +83,7 @@ func IsMsiBuild() bool {
 	return msiBuild != "false"
 }
 
-func getCRCLatestVersionFromMirror() (*semver.Version, error) {
+func GetCRCLatestVersionFromMirror() (*CrcReleaseInfo, error) {
 	client := &http.Client{
 		Timeout: 5 * time.Second,
 	}
@@ -107,17 +107,5 @@ func getCRCLatestVersionFromMirror() (*semver.Version, error) {
 		return nil, fmt.Errorf("Error unmarshaling JSON metadata: %v", err)
 	}
 
-	return releaseInfo.Version.CrcVersion, nil
-}
-
-func NewVersionAvailable() (bool, string, error) {
-	latestVersion, err := getCRCLatestVersionFromMirror()
-	if err != nil {
-		return false, "", err
-	}
-	currentVersion, err := semver.NewVersion(GetCRCVersion())
-	if err != nil {
-		return false, "", err
-	}
-	return latestVersion.GreaterThan(currentVersion), latestVersion.String(), nil
+	return &releaseInfo, nil
 }
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -37,9 +37,14 @@ const (
 )
 
 type CrcReleaseInfo struct {
-	Version struct {
-		LatestVersion string `json:"crcVersion"`
-	}
+	Version Version           `json:"version"`
+	Links   map[string]string `json:"links"`
+}
+
+type Version struct {
+	CrcVersion       *semver.Version `json:"crcVersion"`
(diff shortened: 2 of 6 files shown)


cmd/crc/cmd/start.go
```diff
@@ -190,20 +193,39 @@ func validateStartFlags() error {
...
-		logging.Warnf("A new version (%s) has been published on %s", newVersion, constants.CrcLandingPageURL)
-		return
+		logging.Warnf("A new version (%s) has been published on %s", newVersion, link)
+		return nil
 	}
 	logging.Debugf("No new version available. The latest version is %s", newVersion)
+	return nil
+}
+
+func newVersionAvailable() (bool, string, string, error) {
```
praveenkumar, line 212: All looks good but why this function moved from pkg/version to here, I think it's good that it is part of pkg/version and not scoped to `cmd` only?

praveenkumar:
The reason we had it to point to landing page because we notified that pull secret might updated for the user so always good to get latest pull secret but now we are not seeing that kind of report.

cfergeau:
The relevant link would be https://access.redhat.com/documentation/en-us/red_hat_codeready_containers/1.27/html/release_notes_and_known_issues/index or https://access.redhat.com/documentation/en-us/red_hat_codeready_containers/1.27/ , both of which do not require to login to see the content (information provided by Praveen and Anjan in an internal discussion)

cmd/crc/cmd/start.go
guillaumerose:
```

func downloadLink(release *crcversion.CrcReleaseInfo) string {
	if link, ok := release.Links[runtime.GOOS]; ok {
		return link
	}
	return constants.CrcLandingPageURL
}
```

requires a constant in constants pkg. 
But also, versions pkg in included in constants pkg. So there is cyclic dependency if I put this code in version pkg.

guillaumerose:
The version for the doc doesn't have the patch number :/

cmd/crc/cmd/start.go
praveenkumar: @guillaumerose Ah yes, ideally we shouldn't import anything related to crc pkg so it always immune to cyclic dependency but looking at the code it looks like unavoidable until we refactor around the way we pass `ldflags` during compilation. In the meanwhile I think following is also clean.
```patch
diff --git a/cmd/crc/cmd/start.go b/cmd/crc/cmd/start.go
index 3ae37714..02863baa 100644
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -10,7 +10,6 @@ import (
        "strings"
        "text/template"
 
-       "github.com/Masterminds/semver/v3"
        "github.com/code-ready/crc/pkg/crc/cluster"
        crcConfig "github.com/code-ready/crc/pkg/crc/config"
@@ -198,7 +197,7 @@ func checkIfNewVersionAvailable(noUpdateCheck bool) error {
        if noUpdateCheck {
                return nil
        }
-       isNewVersionAvailable, newVersion, link, err := newVersionAvailable()
+       isNewVersionAvailable, newVersion, link, err := crcversion.NewVersionAvailable()
        if err != nil {
                return err
        }
@@ -210,25 +209,6 @@ func checkIfNewVersionAvailable(noUpdateCheck bool) error {
        return nil
 }
 
-func newVersionAvailable() (bool, string, string, error) {
-       release, err := crcversion.GetCRCLatestVersionFromMirror()
-       if err != nil {
-               return false, "", "", err
-       }
-       currentVersion, err := semver.NewVersion(crcversion.GetCRCVersion())
-       if err != nil {
-               return false, "", "", err
-       }
-       return release.Version.CrcVersion.GreaterThan(currentVersion), release.Version.CrcVersion.String(), downloadLink(release), nil
-}
-
-func downloadLink(release *crcversion.CrcReleaseInfo) string {
-       if link, ok := release.Links[runtime.GOOS]; ok {
-               return link
-       }
-       return constants.CrcLandingPageURL
-}
-
 const startTemplate = `Started the OpenShift cluster.
 
diff --git a/pkg/crc/constants/constants.go b/pkg/crc/constants/constants.go
index 98a9d62b..3ba00977 100644
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -25,7 +25,6 @@ const (
        ConfigFile                = "crc.json"
        LogFile                   = "crc.log"
        DaemonLogFile             = "crcd.log"
-       CrcLandingPageURL         = "https://cloud.redhat.com/openshift/create/local" // #nosec G101
        DefaultPodmanURLBase      = "https://storage.googleapis.com/libpod-master-releases"
        DefaultAdminHelperCliBase = "https://github.com/code-ready/admin-helper/releases/download/0.0.4"
        CRCMacTrayDownloadURL     = "https://github.com/code-ready/tray-macos/releases/download/v%s/crc-tray-macos.tar.gz"
@@ -42,6 +41,8 @@ const (
        AppsDomain    = ".apps-crc.testing"
 )
 
+var CrcLandingPageURL = version.CrcLandingPageURL
+
 var adminHelperExecutableForOs = map[string]string{
        "darwin":  "crc-admin-helper-darwin",
        "linux":   "crc-admin-helper-linux",
diff --git a/pkg/crc/version/version.go b/pkg/crc/version/version.go
index d73b4991..cab0e293 100644
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -5,6 +5,7 @@ import (
        "fmt"
        "io/ioutil"
        "net/http"
+       "runtime"
        "time"
 
        "github.com/Masterminds/semver/v3"
@@ -34,6 +35,7 @@ const (
        crcMacTrayVersion = "1.0.9"
        // Windows forms application version type major.minor.buildnumber.revesion
        crcWindowsTrayVersion = "0.6.0.0"
+       CrcLandingPageURL     = "https://cloud.redhat.com/openshift/create/local" // #nosec G101
 )
 
 type CrcReleaseInfo struct {
@@ -109,3 +111,22 @@ func GetCRCLatestVersionFromMirror() (*CrcReleaseInfo, error) {
 
        return &releaseInfo, nil
 }
+
+func NewVersionAvailable() (bool, string, string, error) {
+       release, err := GetCRCLatestVersionFromMirror()
+       if err != nil {
+               return false, "", "", err
+       }
+       currentVersion, err := semver.NewVersion(GetCRCVersion())
+       if err != nil {
+               return false, "", "", err
+       }
+       return release.Version.CrcVersion.GreaterThan(currentVersion), release.Version.CrcVersion.String(), downloadLink(release), nil
+}
+
+func downloadLink(release *CrcReleaseInfo) string {
+       if link, ok := release.Links[runtime.GOOS]; ok {
+               return link
+       }
+       return CrcLandingPageURL
+}

```

praveenkumar:
> The version for the doc doesn't have the patch number :/

@guillaumerose you mean like `1.25.1` and it just have `1.25` ?  Docs only care about `major.minor` not the patch one since they updated asynchronously.

pkg/crc/version/version.go
```diff
@@ -88,22 +92,22 @@ func getCRCLatestVersionFromMirror() (*semver.Version, error) {
...
+	if err := json.Unmarshal(releaseMetaData, &releaseInfo); err != nil {
 		return nil, fmt.Errorf("Error unmarshaling JSON metadata: %v", err)
 	}
-	version, err := semver.NewVersion(releaseInfo.Version.LatestVersion)
-	if err != nil {
-		return nil, err
-	}
-	return version, nil
+
+	return releaseInfo.Version.CrcVersion, nil
```
cfergeau: Might be good to return an error if `Version` or `CrcVersion` are missing.

cmd/crc/cmd/start.go
```diff
@@ -191,20 +194,39 @@ func validateStartFlags() error {
...
```
cfergeau, line 213: A short comment describing the return values might be useful as `(bool, string, string, error)` is a bit complex

cfergeau (commented):
Looks good to me!

cmd/crc/cmd/start.go
guillaumerose:
```
+var CrcLandingPageURL = version.CrcLandingPageURL
```

I dont really like this. It is complicated to follow. Maybe we should merge version and constants pkg ?

pkg/crc/version/version.go
guillaumerose: Yes added a nil check, good point!

When the json was empty:
```
$ crc version
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x773e7a]

goroutine 1 [running]:
github.com/Masterminds/semver/v3.(*Version).Compare(0x0, 0xc0008f60f0, 0xc0008f60f0)
	/home/guillaumerose/gomod/crc/vendor/github.com/Masterminds/semver/v3/version.go:388 +0x3a
github.com/Masterminds/semver/v3.(*Version).GreaterThan(...)
	/home/guillaumerose/gomod/crc/vendor/github.com/Masterminds/semver/v3/version.go:368
github.com/code-ready/crc/cmd/crc/cmd.newVersionAvailable(0x1b2991c, 0xc00051bb78, 0x8017bc, 0x1815d60, 0x2be5540, 0x1815d60, 0x2be5540)
	/home/guillaumerose/gomod/crc/cmd/crc/cmd/start.go:222 +0x9a
github.com/code-ready/crc/cmd/crc/cmd.checkIfNewVersionAvailable(0x1815d00, 0x2be5540, 0x0)
	/home/guillaumerose/gomod/crc/cmd/crc/cmd/start.go:201 +0x44
```

PR #2395, opened 2021-06-02
Fix release info URI to use `rest/mirror` instead `file`

praveenkumar:


gbraad (approved):

gbraad:
I actually noticed the `rest/mirror` today upon downloading CRC. but didn't realize the 'impact'.

/lgtm

PR #2399, opened 2021-06-04
Remove system networking mode on macOS

guillaumerose:
Fixes: #2475

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -79,8 +79,12 @@ func RegisterSettings(cfg *Config) {
 		"Disable update check (true/false, default: false)")
 	cfg.AddSetting(ExperimentalFeatures, false, ValidateBool, SuccessfullyApplied,
 		"Enable experimental features (true/false, default: false)")
-	cfg.AddSetting(NetworkMode, string(defaultNetworkMode()), network.ValidateMode, network.SuccessfullyAppliedMode,
-		fmt.Sprintf("Network mode (%s or %s)", network.UserNetworkingMode, network.SystemNetworkingMode))
+
+	if !version.IsMacosInstallPathSet() {
+		cfg.AddSetting(NetworkMode, string(defaultNetworkMode()), network.ValidateMode, network.SuccessfullyAppliedMode,
+			fmt.Sprintf("Network mode (%s or %s)", network.UserNetworkingMode, network.SystemNetworkingMode))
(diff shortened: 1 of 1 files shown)


anjannath:
No changes are needed in the preflight checks?

guillaumerose:
The fat binary still exist so I can't remove old preflight checks. 
If the user was using system networking, it will be now user networking.

cfergeau:
This means there are no longer any fallbacks if the user hits usermode networking limitations. Have we gathered enough feedback already to be confident that nobody will need that fallback while we fix some usermode networking bugs/missing features?

guillaumerose:
1.26.0 was released one month ago and we had no issue about this. The release is in 1 week so I guess it is safe to merge this at the last minute if there is still no issue.

praveenkumar:
I don't understand why we want to force it without any fallback options. In case of pkg we are already setting `UserMode` for network as default. This way until user manually set `network-mode` to `system` it is by default using `user` mode. With this change he will never able to go back on `system` mode.  If the plan is to drop system mode for mac then I am OK but until then I am in favor to provide user an option to choose.

guillaumerose:
Yes the goal is to remove the system network mode. 
The system network mode is only useful for the fat binary. I keep it only because of IDEs still consume it.

When the system mode is removed we can:
* run hyperkit as normal user
* remove the driver binary and compile it directly in crc binary

cfergeau:
Right now, this PR removes a potential workaround for usermode networking issues, but as you pointed out, it cannot remove all the non-usermode yet. Since the code for non-usermode will still be there, I don't see what benefit we get from this PR, I'd favour not merging this, and just remove the old code when possible.

praveenkumar:
We are going to keep this on hold till we have proxy tests are in for vsock/userland networking side.

guillaumerose:
Proxy test OK:

* Run mitmproxy (port 8080)
* Install crc
* Run crc setup
* crc config set https-proxy http://192.168.0.21:8080 (this my network card IP)
* crc config set http-proxy http://192.168.0.21:8080
* crc config set proxy-ca-file /Users/guillaumerose/.mitmproxy/mitmproxy-ca-cert.cer
* Start the cluster as usual.

Screenshot 2021-06-24 at 13 35 12

gbraad (approved):

praveenkumar:
Follow up issue #2475

PR #2400, opened 2021-06-04
Update bundle to 4.7.13

praveenkumar:


guillaumerose:
https://github.com/openshift/cincinnati-graph-data/pull/839

4.7.14 is already a dead release :)

pkg/crc/machine/start.go
```diff
@@ -290,10 +290,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
...
 
 	// Start network time synchronization if `CRC_DEBUG_ENABLE_STOP_NTP` is not set
-	if stopNtp, _ := strconv.ParseBool(os.Getenv("CRC_DEBUG_ENABLE_STOP_NTP")); !stopNtp {
-		logging.Info("Starting network time synchronization in CodeReady Containers VM")
-		if _, _, err := sshRunner.Run("sudo timedatectl set-ntp on"); err != nil {
-			return nil, errors.Wrap(err, "Failed to start network time synchronization")
+	if stopNtp, _ := strconv.ParseBool(os.Getenv("CRC_DEBUG_ENABLE_STOP_NTP")); stopNtp {
+		logging.Info("Stopping network time synchronization in CodeReady Containers VM")
+		if _, _, err := sshRunner.Run("sudo timedatectl set-ntp off"); err != nil {
+			return nil, errors.Wrap(err, "Failed to stop network time synchronization")
```
cfergeau, line 296: Not sure this is going to be enough, `test/e2e/features/cert_rotation.feature` sets the host time 3 months ahead, and knows for sure the VM will use the host time rather than getting it from NTP since NTP was disabled at boot time. Now this has changed, and this looks like it could be racy, the test sets the host clock, the VM starts and syncs to the correct time using NTP, and then we stop NTP. We'd need an additional `ssh` command to set the VM time to be the same as the host time.

praveenkumar: @cfergeau yes, we also need to set the VM time ahead of 3 month after stopping the time sync so that when kubelet starts it sees that time and start cert rotation.

pkg/crc/machine/start.go
```diff
@@ -290,10 +290,14 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
...
-		if _, _, err := sshRunner.Run("sudo timedatectl set-ntp on"); err != nil {
-			return nil, errors.Wrap(err, "Failed to start network time synchronization")
+	if stopNtp, _ := strconv.ParseBool(os.Getenv("CRC_DEBUG_ENABLE_STOP_NTP")); stopNtp {
+		logging.Info("Stopping network time synchronization in CodeReady Containers VM")
+		if _, _, err := sshRunner.RunPrivileged("timedatectl set-ntp off"); err != nil {
+			return nil, errors.Wrap(err, "Failed to stop network time synchronization")
+		}
+		logging.Info("Setting clock to 3 months ahead")
+		if _, _, err := sshRunner.RunPrivileged("date -s '3 month'"); err != nil {
+			return nil, errors.Wrap(err, "Failed to set clock to 3 months ahead")
```
cfergeau, line 300: I'd prefer that we sync the current host time instead of hardcoding '3 months', this would make this code slightly more flexible.

cfergeau (approved):

PR #2404, opened 2021-06-07
Use user network mode by default with the MSI

guillaumerose:


cmd/crc/cmd/start.go
```diff
@@ -274,6 +274,9 @@ func checkDaemonStarted() error {
 }
 
 func daemonStartedErrorMessage() string {
+	if crcversion.IsMsiBuild() {
+		return "Is CodeReady Containers tray application is running ? Cannot reach daemon API: %v"
```
cfergeau, line 278: Extra 'is' ("is running" -> "running"), and extra space between "running" and "?", there should be no space

cfergeau (approved):
Ack with two minor string improvements.

cmd/crc/cmd/start.go
guillaumerose: fixed!

PR #2406, opened 2021-06-07
Add section on compiling CRC locally to developing.adoc

timflannagan:
Update the developing.adoc developer documentation and add a section on
how to build the `crc` binary locally.

**Fixes:** Issue #N

## Solution/Idea

## Proposed changes

1. Add `start` command.
3. ...

## Testing

7. ...

### DIFF
--- a/developing.adoc
+++ b/developing.adoc
@@ -46,6 +46,36 @@ For more information, see the following:
 . link:https://github.com/golang/go/wiki/Modules[Introduction to Go modules]
 . link:https://blog.golang.org/using-go-modules[Using Go modules]
 
+[[compiling-binaries]]
+== Compiling the CRC Binary
+
+In order to compile the crc executable for your local platform, run the following command:
+
+----
+$ make
+----
+
+By default, the above command will place the crc executable in the `$GOBIN` path.
+
+Run the following command to cross-compile the crc executable for many platforms:
+
+----
+$ make cross
+----
+
+*Note*: This command will output the cross-compiled crc executable(s) in the `out` directory by default:
+
+----
+$ tree out/
+out/
+├── linux-amd64
+│   └── crc
+├── macos-amd64
+│   └── crc
+└── windows-amd64
+    └── crc.exe
+----
+
 [[running-e2e-tests]]
 == Running e2e tests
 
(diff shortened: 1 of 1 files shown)


timflannagan:
I took a stab at adding this documentation as it wasn't immediately clear to me how to build the binaries locally when inspecting the root Makefile. I'm open to closing this if we don't think there's value in documenting this behavior here - I just had some free time while waiting for a cluster to spin up.

developing.adoc
```diff
@@ -46,6 +46,34 @@ For more information, see the following:
...
+Run the following Makefile target that will produce binaries for multiple platforms:
+
+----
+$ make cross
+----
+
+In order to compile the crc binary for a specific platform, run the following command:
+
+----
+$ make out/linux-amd64/crc
```
praveenkumar, line 61: This can be replaced with `make` because we have `default` target in the Makefile is `install` which will build the binary on `$GOPATH/bin` and also platform independent.

anjannath: For the above suggestion, the statement would also need to be changed to something like `In order to compile the crc binary for your platform, and install it in $GOBIN`...

timflannagan: That sounds reasonable to me. Any suggestions on how we'd document, or whether we see value in documenting, how to install CRC or the embedded binaries for a specific platform? It seems a bit awkward to me in the current form, but maybe that's because we'd have to specify an override variable if we're interested in overriding the default binary output path to something more traditional like `bin` for Go-based projects.

praveenkumar: > It seems a bit awkward to me in the current form, but maybe that's because we'd have to specify an override variable if we're interested in overriding the default binary output path to something more traditional like bin for Go-based projects.

@timflannagan  if a user want to use some different path then `$GOPATH/bin` then they can set `$GOPATH` accordingly. I think that's how it is for go projects.

praveenkumar:
@timflannagan Are you planning to update this PR as per review?

timflannagan:
@praveenkumar I'm fairly tied up with work atm - I can likely get back to this by Friday.

developing.adoc
timflannagan: Sounds good to me - updated.

Issue #2387, opened 2021-05-31
Cleanup Kubernetes context when deleting the VM


guillaumerose:
`crc start` automatically adds 2 contexts: crc-admin and crc-developer. 
When the user runs `oc login`, it creates even more contexts.

After VM deletion, kubeconfig is full of broken contexts. `crc delete` should delete them.

anjannath:
Done in https://github.com/code-ready/crc/pull/2407

PR #2407, opened 2021-06-08
Clean kubeconfig file after crc delete

guillaumerose:
Remove contexts, auth infos and clusters related to the crc cluster from
 the kubeconfig file.

```
$ kubectx
/api-crc-testing:6443/developer
crc-admin
crc-developer
kind-reference
minishift
myproject/192-168-42-71:8443/developer
myproject/192-168-42-71:8443/system:admin
project1/api-crc-testing:6443/developer

$ crc delete
Do you want to delete the OpenShift cluster? [y/N]: y
Deleted the OpenShift cluster

$ kubectx
kind-reference
minishift
myproject/192-168-42-71:8443/developer
myproject/192-168-42-71:8443/system:admin

```

PR #2408, opened 2021-06-08
Remove the crc cluster entry from kubeconfig

praveenkumar:
fixes: #2387

## Solution/Idea

Remove all the entries from kubeconfig file which belong to crc cluster when execute `crc delete`

## Testing
- Start cluster using `crc start`
- Check the kubeconfig file content `oc config view`
- delete crc `crc delete`
- Check again the content of kubeconfig file which shouldn't have any entry about crc.


praveenkumar:
/close 

In favor of #2407

PR #2412, opened 2021-06-09
Preliminary work in dns package for presets

guillaumerose:
This separates the code related to DNS for OpenShift and DNS for a working VM. 
The end goal is to be able to not run the OpenShift part.


guillaumerose:
/retest
retest this please

pkg/crc/machine/start.go
```diff
@@ -324,7 +324,11 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 	}
 
 	// Run the DNS server inside the VM
-	if err := dns.RunPostStart(servicePostStartConfig); err != nil {
+	if err := dns.ConfigureForOpenShift(servicePostStartConfig); err != nil {
+		return nil, errors.Wrap(err, "Error running post start")
+	}
+
+	if err := dns.CreateResolvFileOnInstance(servicePostStartConfig); err != nil {
```
praveenkumar, line 331: @guillaumerose I understand that you want to move openshift stuff separate so these changes shouldn't be part of podman specific run but I don't understand the separating of `CreateResolvFileOnInstance` doesn't changes inside only make sense for openshift (at least on system network mode) I am not sure if you want to use the predefined gateway for podman stuff?

guillaumerose: Yes this is because vsock mode doesn't use DHCP yet. The resolv config file is manually written by crc.

praveenkumar: @guillaumerose Thanks, can you add that part in commit log so that it's easier to follow in future about why we did it?

PR #2415, opened 2021-06-09
docs: Expand installation procedure, cover binary and installer

robin-owen:
## Solution/Idea

Describe the installation procedure in more detail, particularly for installing the portable binary on RHEL. This PR also briefly documents that CRC is available on macOS using a guided installer. In the future, we will likely replace the Windows instructions in this PR once the installer for Windows is the default installation method for that platform (but will retain the notice regarding installing to the `C:\` drive).

**Note:** Regarding the wording of the installation procedure, I wasn't sure if we should advocate for _copying_ the executable or _moving_ it from the extracted location. I opted for copying due to that action being non-destructive.

### DIFF
--- a/docs/source/topics/proc_installing-codeready-containers.adoc
+++ b/docs/source/topics/proc_installing-codeready-containers.adoc
@@ -1,6 +1,9 @@
 [id="installing-codeready-containers_{context}"]
 = Installing {prod}
 
+{prod} is available as a portable executable for {rhel} and {msw}.
+On {mac}, {prod} is available using a guided installer.
+
 .Prerequisites
 
 * Your host machine must meet the minimum system requirements.
@@ -8,4 +11,47 @@ For more information, see link:{crc-gsg-url}#minimum-system-requirements_gsg[Min
 
 .Procedure
 
-. Download the link:{crc-download-url}[latest release of {prod}] for your platform and extract the contents of the archive to a location in your `_PATH_`.
+. Download the link:{crc-download-url}[latest release of {prod}] for your platform.
+
+. On {mac}, run the guided installer and follow the instructions.
+On {rhel} or {msw}, assuming the archive is in the [filename]*_~/Downloads_* directory, follow the procedure for your platform.
++
+** For {rhel}:
++
+... Extract the contents of the archive:
++
+[subs="attributes"]
+----
+$ cd ~/Downloads
+$ tar xvf crc-linux-amd64.tar.xz
+----
++
+... Create the [filename]*_~/bin_* directory if it does not exist and copy the [command]`{bin}` executable to it:
++
+[subs="attributes"]
+----
+$ mkdir -p ~/bin
+$ cp ~/Downloads/crc-linux-*-amd64/{bin} ~/bin
+----
++
+... Add the [filename]*_~/bin_* directory to your `_PATH_`:
++
+[subs="attributes"]
+----
+$ export PATH=$PATH:$HOME/bin
+$ echo 'export PATH=$PATH:$HOME/bin' >> ~/.bashrc
+----
++
+** For {msw}:
++
+... Extract the contents of the archive.
+... Create the desired directory path for {prod} and copy the extracted [command]`{bin}.exe` executable to it.
+... Add the directory path to your `_PATH_` environment variable.
++
+[NOTE]
+====
+On {msw}, you must execute the {prod} executable from your local [filename]*_C:\_* drive.
+You cannot run {prod} from a network drive.
+
+If you cannot install the executable in your `_PATH_`, run {prod} from the current directory as `./{bin}.exe`.
+====


docs/source/topics/proc_installing-codeready-containers.adoc
```diff
@@ -1,11 +1,57 @@
...
++
+[subs="attributes"]
+----
+$ export PATH=$PATH:$HOME/bin
+$ echo 'export PATH=$PATH:$HOME/bin' >> ~/.bashrc
+----
++
+** For {msw}:
++
+... Extract the contents of the archive.
```
robin-owen, line 47: Note that I'd really like to have more detailed instructions for Windows, but I can't seem to find reliable information about how to perform this procedure myself across the various versions. Could anyone help me make this clearer for readers who are using Windows, or is this adequate?

docs/source/topics/proc_installing-codeready-containers.adoc
```diff
@@ -1,11 +1,57 @@
 [id="installing-codeready-containers_{context}"]
 = Installing {prod}
 
+{prod} is available as a portable executable for {rhel} and {msw}.
```
jc-berger, line 4: I'm not used to seeing `executable` as a noun. But your users might be, in which case, keep it!

docs/source/topics/proc_installing-codeready-containers.adoc
```diff
@@ -1,11 +1,57 @@
...
+
 .Prerequisites
 
 * Your host machine must meet the minimum system requirements.
 For more information, see link:{crc-gsg-url}#minimum-system-requirements_gsg[Minimum system requirements].
 
 .Procedure
 
-. Download the link:{crc-download-url}[latest release of {prod}] for your platform and extract the contents of the archive to a location in your `_PATH_`.
+. Download the link:{crc-download-url}[latest release of {prod}] for your platform.
```
jc-berger, line 14: The linked page takes me to a 404 error page. See the screenshot: 

Screen Shot 2021-06-09 at 10 43 10 AM

docs/source/topics/proc_installing-codeready-containers.adoc
```diff
@@ -1,11 +1,57 @@
...
++
+... Extract the contents of the archive:
++
+[subs="attributes"]
+----
+$ cd ~/Downloads
+$ tar xvf crc-linux-amd64.tar.xz
+----
++
+... Create the [filename]*_~/bin_* directory if it does not exist and copy the [command]`{bin}` executable to it:
```
jc-berger, line 29: Ah, it does seem like `executable` is commonly used as a noun, so your users will be familiar with it :) Disregard my earlier comment, thanks!

docs/source/topics/proc_installing-codeready-containers.adoc
```diff
@@ -1,11 +1,57 @@
...
+... Extract the contents of the archive.
+... Create the desired directory path for {prod} and copy the extracted [command]`{bin}.exe` executable to it.
+... Add the directory path to your `_PATH_` environment variable.
++
+[NOTE]
+====
+On {msw}, you must execute the {prod} executable from your local [filename]*_C:\_* drive.
+You cannot run {prod} from a network drive.
+
+{prod} can be run from the current directory as [command]`./{bin}.exe` if you cannot install the executable in your `_PATH_`.
```
jc-berger, line 56: So I don't think the IBM SG specifically dictates that subordinate clauses must begin a sentence instead of end a sentence. However, most of the SG's examples of `if` clauses are at the beginning of the sentence, not the end. So consider rephrasing to:

`If you cannot install the executable in your `_PATH_`, {prod} can be run...`

Also, for consistency, phrase the second sentence in active voice, directly talking to the user, like you do in the first sentence. So consider rephrasing to something like, 

`If you cannot install the executable in your `_PATH_`, run {prod} from your current directory as [command] `./{bin}.exec`. `

docs/source/topics/proc_installing-codeready-containers.adoc
jc-berger: So I can't give you any advice on how to expand the current instructions for Windows users, but I do think the instructions are adequate as is.

docs/source/topics/proc_installing-codeready-containers.adoc
robin-owen: This has been discussed before, actually. :)

Normally, I write "binary," but it was brought to my attention before that technically, we're discussing an _executable_ file, not just a binary one. It's a subtle difference, but an important one. :)

docs/source/topics/proc_installing-codeready-containers.adoc
robin-owen: The URL is an attribute, so GitHub may not know how to render it correctly. The link should take you to this page when properly rendered: https://cloud.redhat.com/openshift/create/local

For the record, you can also generate the documentation locally by running `make build_docs` in the repository. The generated docs are in `docs/build/index.html`. :)

docs/source/topics/proc_installing-codeready-containers.adoc
```diff
@@ -1,11 +1,57 @@
...
 * Your host machine must meet the minimum system requirements.
 For more information, see link:{crc-gsg-url}#minimum-system-requirements_gsg[Minimum system requirements].
 
 .Procedure
 
-. Download the link:{crc-download-url}[latest release of {prod}] for your platform and extract the contents of the archive to a location in your `_PATH_`.
+. Download the link:{crc-download-url}[latest release of {prod}] for your platform.
+
+. On {mac}, run the guided installer and follow the instructions.
+On {rhel} or {msw}, assuming the archive is in the [filename]*_~/Downloads_* directory, follow the procedure for your platform.
```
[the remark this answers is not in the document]
robin-owen: Good point, I was unsure of using "assuming" myself. My reasoning for writing it this way is as follows, though: technically, we cannot control where a user downloads the file, but the *_Downloads_* directory is the default on every major platform unless configured otherwise. Since we can't be sure that the archive is downloaded there, we assume that it is and inform the reader of that assumption so that they can modify the commands appropriately if they downloaded it somewhere else.

docs/source/topics/proc_installing-codeready-containers.adoc
robin-owen: Good point, fixing that now. Thanks!

docs/source/topics/proc_installing-codeready-containers.adoc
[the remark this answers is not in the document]
jc-berger: Good reasoning! I think it's reasonable to use `assuming` in this instance.

jc-berger (approved):
LGTM from doc's point of view. Great job!

guillaumerose (approved):

PR #2416, opened 2021-06-09
Remove fat macOS binary, keep only for the .pkg for release

guillaumerose:


cfergeau (approved):
This potentially allows for some additional cleanups, for example preflights with `if runtime.GOOS == "darwin" && version.IsMacosInstallPathSet() return`. However, these are convenient for testing from git, so dunno yet what to do for these :)

praveenkumar:
> This potentially allows for some additional cleanups, for example preflights with `if runtime.GOOS == "darwin" && version.IsMacosInstallPathSet() return`. However, these are convenient for testing from git, so dunno yet what to do for these :)

Yes agreed 👍🏼 

/lgtm

PR #2418, opened 2021-06-10
Make rootless as default socket for podman-env command

praveenkumar:
We thought that enabling root socket, make a step less for developers to deploy
an application on OpenShift but during testing it might only work if developer
don't tag the image with registry which have `localhost/<imageName>`. Also our
goal is to teach user about Openshift about how to use internal registry so
it does divert from the production clusters. I think if a user really want
the application directly from the dockerfile without building it localy then
`oc new-build` would be best choice.

### DIFF
--- a/cmd/crc/cmd/podman_env.go
+++ b/cmd/crc/cmd/podman_env.go
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	rootless bool
+	root bool
 )
 
 var podmanEnvCmd = &cobra.Command{
@@ -37,9 +37,9 @@ func runPodmanEnv() error {
 		return err
 	}
 
-	socket := "/run/podman/podman.sock"
-	if rootless {
-		socket = "/run/user/1000/podman/podman.sock"
+	socket := "/run/user/1000/podman/podman.sock"
+	if root {
+		socket = "/run/podman/podman.sock"
 	}
 	fmt.Println(shell.GetPathEnvString(userShell, constants.CrcOcBinDir))
 	fmt.Println(shell.GetEnvString(userShell, "CONTAINER_SSHKEY", connectionDetails.SSHKeys[0]))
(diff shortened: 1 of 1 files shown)


gbraad (approved):

gbraad:
So, how to disable this? Perhaps food to describe this in the command output as an option.

cfergeau:
> So, how to disable this? Perhaps good to describe this in the command output as an option.

"disable this"? Do you mean how to switch between `rootless`/`non rootless`? This can done with `--rootless=false` which is odd..

gbraad:
> This can done with --rootless=false which is odd..

Right, and that is what either has to be documented/added in the message ... or simple, this won't do.
Perhaps the option is entirely not useful as we target a VM, but might be a different experience from a Linux box.

gbraad (changes requested):
`--rootless=false` is not a good UX

guillaumerose:
I updated the PR:
* flag is now `root`
* by default it is false

anjannath (approved):

anjannath:
To have podman command talk to the root socket now user needs to do `crc podman-env --root=true`

cfergeau:
Wondering if it would not be simpler/safer to revert ebb0e28cb4fba89 instead while we figure out a clearer story around podman. If the goal of using `--root` is to be able to bind to port 80, setting `CAP_NET_BIND_SERVICE` on the podman binary in the VM could be an alternative.

guillaumerose:
podman-machine is also creating the root context. See https://github.com/containers/podman/blob/a634b2cd5977c60f1907733efe07b61ba36271fb/pkg/machine/qemu/machine.go#L177

praveenkumar:
> If the goal of using --root is to be able to bind to port 80, setting CAP_NET_BIND_SERVICE on the podman binary in the VM could be an alternative which does not need any additional option.

@cfergeau It is not only that but also the storage location is different for root and rootless use of podman, so even some advance user does try to build using podman and deploy the app on Openshift will directly work.

gbraad:
@guillaumerose

Also creating, but as connections it seems that is named `-root` as suffix:

https://github.com/containers/podman/blob/a634b2cd5977c60f1907733efe07b61ba36271fb/pkg/machine/qemu/machine.go#L172-L180

```go
		uri := machine.SSHRemoteConnection.MakeSSHURL("localhost", "/run/user/1000/podman/podman.sock", strconv.Itoa(v.Port), v.RemoteUsername)
		if err := machine.AddConnection(&uri, v.Name, filepath.Join(sshDir, v.Name), opts.IsDefault); err != nil {
			return err
		}


		uriRoot := machine.SSHRemoteConnection.MakeSSHURL("localhost", "/run/podman/podman.sock", strconv.Itoa(v.Port), "root")
		if err := machine.AddConnection(&uriRoot, v.Name+"-root", filepath.Join(sshDir, v.Name), opts.IsDefault); err != nil {
			return err
		}
```

PR #2436, opened 2021-06-15
`crc setup` hangs when trying to enable hyperv

cfergeau:
On my Windows 10 21h1 install, when I run
'Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V -All'
the command asks whether I want to reboot the system now or not, and
does not complete until it gets an answer.
This causes `crc setup` to hang when it calls it to setup hyper-v.
This commit uses the -NoRestart argument to make sure there is no
interactive prompt, `crc setup` will tell the user to reboot when
appropriate.

This is a pre-existing issue which was made visible by 0f044b17a6 since
we now wait until the end of execution of powershell scripts that we
spawn.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -87,7 +87,7 @@ func checkHyperVInstalled() error {
 
 //
 func fixHyperVInstalled() error {
-	enableHyperVCommand := `Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V -All`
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -87,7 +87,7 @@ func checkHyperVInstalled() error {
 
 //
 func fixHyperVInstalled() error {
-	enableHyperVCommand := `Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V -All`
+	enableHyperVCommand := `Enable-WindowsOptionalFeature -Online -FeatureName Microsoft-Hyper-V -All -NoRestart`
```
anjannath, line 90: 👍🏽

anjannath (approved):

Issue #1640, opened 2020-11-04
[BUG] crc 1.18 crc config set disk-size no respected on macos


csantanapr:
Trying to set the disk size to `50GB` with `crc config set disk-size 50` doesn't work

### General information

  * OS: macOS
  * Hypervisor: hyperkit
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop

## CRC version
```bash
❯ crc version
CodeReady Containers version: 1.18.0+bb304aa
OpenShift version: 4.6.1 (embedded in binary)
```
  
## CRC status
```bash
crc status --log-level debug
DEBU CodeReady Containers version: 1.18.0+bb304aa
DEBU OpenShift version: 4.6.1 (embedded in binary)
Checking file: /Users/csantana23/.crc/machines/crc/.crc-exist
DEBU About to run SSH command with hidden output
DEBU SSH command succeeded
DEBU image-registry operator is degraded, Reason: ImagePrunerJobFailed
DEBU About to run SSH command:
df -B1 --output=size,used,target /sysroot | tail -1
DEBU SSH command results: err: <nil>, output: 32720793600 12768727040 /sysroot
CRC VM:          Running
OpenShift:       Running (v4.6.1)
Disk Usage:      12.77GB of 32.72GB (Inside the CRC VM)
Cache Usage:     15.26GB
Cache Directory: /Users/csantana23/.crc/cache
```

## CRC config
```bash
crc config view
- cpus                                  : 6
- disk-size                             : 50
- memory                                : 20384
```

## Host Operating System
```bash
sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H2
```

### Steps to reproduce

  1. crc delete
  2. crc setup
  3. crc config set cpus 6
  4. crc config set memory 20384
  5. crc config set disk-size 50

### Expected

Expect VM disk size to be 50GB

### Actual

VM Disk size is the default 33GB

Logs
```
crc start -p ~/Downloads/pull-secret.txt
INFO Checking if oc binary is cached
INFO Checking if podman remote binary is cached
INFO Checking if goodhosts binary is cached
INFO Checking minimum RAM requirements
INFO Checking if running as non-root
INFO Checking if HyperKit is installed
INFO Checking if crc-driver-hyperkit is installed
INFO Checking file permissions for /etc/hosts
INFO Checking file permissions for /etc/resolver/testing
[...]

PR #2439, opened 2021-06-16
macos: Add disk resizing support

cfergeau:
This requires a new machine-driver-hyperkit release (+ the merge of the
corresponding PR) before it works.

**Fixes:** https://github.com/code-ready/crc/issues/1640 

## Testing

1. After this PR, and with a matching machine driver, crc start --disk-size xxx should work the same as on the other platforms

### DIFF
--- a/cmd/crc-embedder/cmd/embed.go
+++ b/cmd/crc-embedder/cmd/embed.go
@@ -94,6 +94,7 @@ var (
 		"darwin": {
 			{hyperkit.MachineDriverDownloadURL, 0755},
 			{hyperkit.HyperKitDownloadURL, 0755},
+			{hyperkit.QcowToolDownloadURL, 0755},
 			{constants.GetCRCMacTrayDownloadURL(), 0644},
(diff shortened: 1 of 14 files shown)


praveenkumar:
/hold

I am testing it as user prospective, will merge after testing.

guillaumerose:
Tested on macOS, works fine for me. 
The only issue was with the tray. The disk size parameter was not correctly passed. https://github.com/code-ready/crc/pull/2439

There is maybe an issue when upgrading: QcowToolPath won't be defined for old cluster. I didn't test that but we ask the user to delete before upgrade.

praveenkumar:
/unhold

Tested and it works as expected.

/lgtm

guillaumerose:
Can you do the driver release and include it in this PR ?

praveenkumar:
> Can you do the driver release and include it in this PR ?

https://github.com/code-ready/crc/pull/2467 I put the PR for same.

PR #2443, opened 2021-06-16
Disable update check in the CI. Add User-Agent in http.Client{}

guillaumerose:
It will help counting the number of clusters started.

### DIFF
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -87,7 +87,12 @@ func GetCRCLatestVersionFromMirror() (*CrcReleaseInfo, error) {
 	client := &http.Client{
 		Timeout: 5 * time.Second,
 	}
-	response, err := client.Get(releaseInfoLink)
+	req, err := http.NewRequest(http.MethodGet, releaseInfoLink, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("User-Agent", fmt.Sprintf("crc/%s", crcVersion))
+	response, err := client.Do(req)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/crc/version/version.go
```diff
@@ -87,7 +87,12 @@ func GetCRCLatestVersionFromMirror() (*CrcReleaseInfo, error) {
 	client := &http.Client{
 		Timeout: 5 * time.Second,
 	}
-	response, err := client.Get(releaseInfoLink)
+	req, err := http.NewRequest(http.MethodGet, releaseInfoLink, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("User-Agent", fmt.Sprintf("crc/%s", crcVersion))
```
praveenkumar, line 94: @guillaumerose this is only added to understand for which version the request made?

guillaumerose: yes, it is also a good practice to specify a proper user agent. I should also add the platform.

pkg/crc/version/version.go
```diff
@@ -87,7 +87,12 @@ func GetCRCLatestVersionFromMirror() (*CrcReleaseInfo, error) {
 	client := &http.Client{
 		Timeout: 5 * time.Second,
 	}
-	response, err := client.Get(releaseInfoLink)
+	req, err := http.NewRequest(http.MethodGet, releaseInfoLink, nil)
```
anjannath, line 90: Don't we need to use the `CRC_DISABLE_UPDATE_CHECK` env variable to stop this check?

guillaumerose: The variable is automatically bound to the config key via that config package. It is a magic feature of viper.

PR #2448, opened 2021-06-18
Add version check for golangci lint

### LINKED ISSUE #2424: Makefile: check version of golangci-lint
`make cross-lint` and `make lint` don't verify the version of golangci-lint. It can lead to inconsistencies and unexpected failures.

We hit that issue with the release 1.28.0. An old version was installed on the Jenkins slave.

praveenkumar:
This patch adds version check for golangci-lint binary and if it is
mismatched then execute `go get` otherwise use what's on the system.

fixes #2424

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ CRC_VERSION = 1.28.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
 MACOS_INSTALL_PATH = /Applications/CodeReady Containers.app/Contents/Resources/
 CONTAINER_RUNTIME ?= podman
+GOLANGCI_LINT_VERSION = v1.39.0
 
 ifdef OKD_VERSION
     BUNDLE_VERSION = $(OKD_VERSION)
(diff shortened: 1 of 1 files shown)


guillaumerose (approved):

guillaumerose:
Thanks! works great

Makefile
```diff
@@ -189,15 +190,17 @@ e2e:
...
-	pushd /tmp && GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.39.0 && popd
+golangci-lint:
+	@if $(GOPATH)/bin/golangci-lint version 2>&1 | grep -vq $(GOLANG_CI_LINT_VERSION); then\
+		pushd /tmp && GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANG_CI_LINT_VERSION) && popd; \
+	fi
 
 # Run golangci-lint against code
 .PHONY: lint cross-lint
-lint: $(GOPATH)/bin/golangci-lint
+lint: golangci-lint
```
cfergeau, line 200: `golangci-lint` is `.PHONY` now.

Makefile
```diff
@@ -6,6 +6,7 @@ CRC_VERSION = 1.28.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
 MACOS_INSTALL_PATH = /Applications/CodeReady Containers.app/Contents/Resources/
 CONTAINER_RUNTIME ?= podman
+GOLANG_CI_LINT_VERSION = v1.39.0
```
cfergeau, line 9: I'd prefer `GOLANGCI_LINT` as this closer to `golangci-lint`

Makefile
```diff
@@ -189,15 +190,17 @@ e2e:
 fmt:
 	@gofmt -l -w $(SOURCE_DIRS)
 
-$(GOPATH)/bin/golangci-lint:
-	pushd /tmp && GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.39.0 && popd
+golangci-lint:
+	@if $(GOPATH)/bin/golangci-lint version 2>&1 | grep -vq $(GOLANG_CI_LINT_VERSION); then\
```
cfergeau, line 194: I think it's probably possible to keep using a real path if you version it. Instead of checking `$(GOPATH)/bin/golangci-lint`, the Makefile checks for `$(GOPATH)/bin/golangci-lint-$GOLANGCI_LINT_VERSION`, and we create a `golangci-lint` symlink to the versioned binary. No idea if this is simpler, this is what I was planning to try if nobody fixed that bug first :)

PR #2449, opened 2021-06-18
Create crc-users group from the msi add logon user to that group and hyperv admins group

anjannath:
### Steps to test:
After Installing the msi:
1. Check that `crc-users` group is created and has the current user as a member
2. the current user is also added to the `hyper-v administrators` group
3. after finishing installation user is prompted to reboot
4. When uninstalled `crc-users` group is removed

Msi to test can be downloaded from appveyor CI artifacts at: https://ci.appveyor.com/api/buildjobs/27np6xmfa3j9dg3v/artifacts/out%2Fwindows-amd64%2Fcrc-windows-installer.zip

### DIFF
--- a/packaging/windows/WixUI_HK.wxs
+++ b/packaging/windows/WixUI_HK.wxs
@@ -11,7 +11,7 @@
          <Property Id="WixUI_Mode" Value="InstallDir" />
          <Property Id="ARPNOMODIFY" Value="1" />
          <Property Id="ARPNOREPAIR" Value="1" />
-         <Property Id="WIXUI_EXITDIALOGOPTIONALTEXT" Value="CodeReady Containers is installed in your computer. Run the command 'crc.exe setup' to setup your environment." />
+         <Property Id="WIXUI_EXITDIALOGOPTIONALTEXT" Value="To finish installation of CodeReady Containers please restart you computer. After restart Open PowerShell or Command Prompt and run the command 'crc.exe setup' to setup your environment." />
 
          <DialogRef Id="BrowseDlg" />
          <DialogRef Id="DiskCostDlg" />
--- a/packaging/windows/product.wxs.in
+++ b/packaging/windows/product.wxs.in
@@ -76,7 +76,7 @@
                     </Component>
                     <Component Id="AddToPath" Guid="09C1E713-44DE-44C3-BDAD-72BE10C10542">
                         <CreateFolder />
-                        <Environment Id="PATH" Name="PATH" Value="[INSTALLDIR]" Permanent="no" Part="last" Action="set" />
+                        <Environment Id="PATH" Name="PATH" Value="[INSTALLDIR]" Permanent="no" Part="last" Action="set" System="yes" />
                     </Component>
                 </Directory>
             </Directory>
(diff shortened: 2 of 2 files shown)


anjannath:
It doesn't remove the user from the Hyper-V admins group when uninstalling!

guillaumerose:
Does it work if Hyper-V is not installed ? I mean, maybe the group Hyper-V admin doesn't exist at all.

anjannath:
> 
> 
> Does it work if Hyper-V is not installed ? I mean, maybe the group Hyper-V admin doesn't exist at all.

I think this'll fail if the group doesn't exist, but Hyper-V admins group will always exist even if hyper-v is not enabled at least on windows 10, and you can't delete it the normal way, it shows an error saying you can't delete buil-in groups

guillaumerose:
What happens if you install crc for all users of your computer ? They would need to be added manually to the group ?

anjannath:
> What happens if you install crc for all users of your computer ? They would need to be added manually to the group ?

Yeah, they'll have to manually add themselves.

currently the msi doesn't offer an option to select between install for current user vs install for all users but we install in `C:\Program Files` and also add the start menu entry for all users and the tray is also set to auto start for all users.

we can based on the failure ask user to run `crc setup` again maybe to solve this?

guillaumerose:
> Yeah, they'll have to manually add themselves.

👍 
I think that's fine for the moment. We could add a check in `crc start`/when the tray is launching.

gbraad:
> all users
> > Yeah, they'll have to manually add themselves.

Difficult situation to handle. Current user is the most important now. We might be able to detect more as part of the tray usage and perform certain actions in future.

anjannath:
This is now in addition to adding the crc-users group also install the admin-helper service and enables hyperv.

packaging/windows/WixUI_HK.wxs
```diff
@@ -11,7 +11,7 @@
          <Property Id="WixUI_Mode" Value="InstallDir" />
          <Property Id="ARPNOMODIFY" Value="1" />
          <Property Id="ARPNOREPAIR" Value="1" />
-         <Property Id="WIXUI_EXITDIALOGOPTIONALTEXT" Value="CodeReady Containers is installed in your computer. Run the command 'crc.exe setup' to setup your environment." />
+         <Property Id="WIXUI_EXITDIALOGOPTIONALTEXT" Value="CodeReady Containers is installed in your computer. Open PowerShell or Command Prompt and run the command 'crc.exe setup' to setup your environment." />
```
guillaumerose, line 14: I would change this to always say 'Please reboot your computer'

anjannath: what about the `crc setup` message? the tray would show a message box saying this after the reboot though

anjannath: There was condition missing for the `<SheduleReboot>` element in the next commits and the prompt was not shown, fixed this in the latest push

anjannath:
Latest build of the msi is at: https://ci.appveyor.com/api/buildjobs/27np6xmfa3j9dg3v/artifacts/out%2Fwindows-amd64%2Fcrc-windows-installer.zip

also updated the first comment with this link

adrianriobo:
@anjannath tested group was created and user was added to both groups correctly with domain user. Also checked uninstall removes the crc-user group.

> and also add the start menu entry for all users

Well here the problem would be domain users, as long as a domain user is not logged in there is no home for the user, so you probably would not be able to handle this adding entries on start menu for current users on the host.

adrianriobo (approved):
Test pass correctly on domain member guest  with domain user

packaging/windows/WixUI_HK.wxs
anjannath: @guillaumerose Do you think the new message is good, with the reboot prompt at the end?

PR #2470, opened 2021-06-23
Delete outdated rendered master/worker machineconfigs and just keep the latest one

praveenkumar:
Those old rendered machine config contain the initial pull spec for
the cluster so better to remove those before bundle generation.

```
$ oc get mc --sort-by=.metadata.creationTimestamp --no-headers
99-master-ssh                                                                                 3.2.0   43h
chronyd-mask                                                                                  3.1.0   43h
99-worker-ssh                                                                                 3.2.0   43h
99-openshift-machineconfig-master-dummy-networks                                              3.2.0   43h
00-master                                          b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
00-worker                                          b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
01-master-kubelet                                  b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
01-master-container-runtime                        b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
01-worker-container-runtime                        b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
01-worker-kubelet                                  b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
rendered-master-6a52c924beeab10c4d917eb7786d8d9c   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
rendered-worker-c0d4ccd0599554ca84cdac121c99b8c0   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
99-worker-generated-registries                     b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
99-master-generated-registries                     b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   43h
rendered-worker-73ad21005c178f63a66398638aaa05ae   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   18h
rendered-master-35111bcd08dccd57f47acaa57cb1d296   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   18h
rendered-master-f5f05aad7d6a34e21fca04e799e54a44   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   3h21m
rendered-worker-24d71aa1fbe25a89567669d89d53065d   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   3h21m
rendered-master-de3a028583cb2ac28121fc8d527637a2   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   142m
rendered-worker-68ca5066a7c81354e3693a0abf8ff3df   b20ecdcd81dc3aa1269757ddfe97603918194b3b   3.2.0   142m

// All the rendered master except lastet one
$ oc get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-master | head -n -1
machineconfig.machineconfiguration.openshift.io/rendered-master-6a52c924beeab10c4d917eb7786d8d9c
machineconfig.machineconfiguration.openshift.io/rendered-master-35111bcd08dccd57f47acaa57cb1d296
machineconfig.machineconfiguration.openshift.io/rendered-master-f5f05aad7d6a34e21fca04e799e54a44
```

**Fixes:** Issue #N

## Solution/Idea

## Proposed changes

1. Add `start` command.
3. ...

## Testing

7. ...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION ?= 4.7.18
+BUNDLE_VERSION ?= 4.8.0
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.29.1
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -192,18 +192,51 @@ func RemovePullSecretFromCluster(ocConfig oc.Config, sshRunner *ssh.Runner) erro
 		return fmt.Errorf("Failed to remove Pull secret %w: %s", err, stderr)
 	}
 
-	cmdArgs = []string{"delete", "machineconfigs", "--all"}
-	_, stderr, err = ocConfig.RunOcCommand(cmdArgs...)
+	return nil
+}
+
+func RemoveOldRenderedMachineConfig(ocConfig oc.Config) error {
+	// This block (#L179-183) should be removed as soon as we start shipping with 4.8 bundle.
+	// This check is only make sure if there is any machineconfig resource or not because
+	// in 4.7 we disabled mco and also deleted all the machineconfig/machineconfig-pools.
+	// For 4.8 we don't disable mco and it does contain the machineconfigs.
+	stdout, stderr, err := ocConfig.RunOcCommand("get mc --sort-by=.metadata.creationTimestamp --no-headers -oname")
 	if err != nil {
-		return fmt.Errorf("Failed to remove machineconfigs %w: %s", err, stderr)
+		return fmt.Errorf("failed to get machineconfig resource %w: %s", err, stderr)
+	}
+	sortedMachineConfigsWithTime := strings.Split(stdout, "\n")
+	if len(sortedMachineConfigsWithTime) == 0 {
+		return nil
 	}
 
-	cmdArgs = []string{"delete", "machineconfigpools", "--all"}
-	_, stderr, err = ocConfig.RunOcCommand(cmdArgs...)
-	if err != nil {
-		return fmt.Errorf("Failed to remove machineconfigpools %w: %s", err, stderr)
+	// We need to make sure only old machine configs are deleted not the new one.
+	var (
+		renderedMaster []string
+		renderedWorker []string
+	)
+	for _, mc := range sortedMachineConfigsWithTime {
+		if strings.HasPrefix(mc, "rendered-master") {
+			renderedMaster = append(renderedMaster, mc)
+		}
+		if strings.HasPrefix(mc, "rendered-worker") {
+			renderedWorker = append(renderedWorker, mc)
+		}
 	}
 
+	var deleteRenderedMachineConfig string
+	if len(renderedMaster) > 0 {
+		deleteRenderedMachineConfig = strings.Join(renderedMaster[:len(renderedMaster)-1], " ")
+	}
+	if len(renderedWorker) > 0 {
+		deleteRenderedMachineConfig = fmt.Sprintf("%s %s", deleteRenderedMachineConfig, strings.Join(renderedWorker[:len(renderedWorker)-1], " "))
+	}
+
+	if deleteRenderedMachineConfig != "" {
+		_, stderr, err = ocConfig.RunOcCommand(fmt.Sprintf("delete %s", deleteRenderedMachineConfig))
+		if err != nil {
+			return fmt.Errorf("Failed to remove machineconfigpools %w: %s", err, stderr)
+		}
+	}
 	return nil
 }
 
--- a/pkg/crc/machine/generate_bundle.go
+++ b/pkg/crc/machine/generate_bundle.go
@@ -1,7 +1,6 @@
 package machine
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -33,8 +32,7 @@ func (client *client) GenerateBundle(forceStop bool) error {
 	}
 
 	// Stop the cluster
-	currentState, err := client.Stop()
-	if err != nil {
+	if _, err := client.Stop(); err != nil {
 		if forceStop {
 			if err := client.PowerOff(); err != nil {
 				return err
@@ -42,8 +40,12 @@ func (client *client) GenerateBundle(forceStop bool) error {
 		}
 		return err
 	}
-	if currentState != state.Stopped {
-		return fmt.Errorf("VM is not stopped, current state is %s", currentState.String())
+	running, err := client.IsRunning()
+	if err != nil {
+		return err
+	}
+	if running {
+		return errors.New("VM is still running")
 	}
 
 	tmpBaseDir, err := ioutil.TempDir(constants.MachineCacheDir, "crc_custom_bundle")
--- a/pkg/crc/machine/generate_bundle.go
+++ b/pkg/crc/machine/generate_bundle.go
@@ -37,8 +37,9 @@ func (client *client) GenerateBundle(forceStop bool) error {
 			if err := client.PowerOff(); err != nil {
 				return err
 			}
+		} else {
+			return err
 		}
-		return err
 	}
 	running, err := client.IsRunning()
 	if err != nil {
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION ?= 4.7.13
+BUNDLE_VERSION ?= 4.8.0-rc.0
```
praveenkumar, line 3: This version update is only to test if current master along with this PR is able to provide us a green CI or do we still need some more changes to incorporate 4.8 release with crc.

pkg/crc/cluster/cluster.go
```diff
@@ -168,18 +168,31 @@ func RemovePullSecretFromCluster(ocConfig oc.Config, sshRunner *ssh.Runner) erro
...
+	// in 4.7 we disabled mco and also deleted all the machineconfig/machineconfig-pools.
+	// For 4.8 we don't disable mco and it does contain the machineconfigs.
+	cmd := "oc --kubeconfig=/opt/kubeconfig get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-master"
+	_, _, err := runner.Run(cmd)
+	if err != nil {
+		return nil
+	}
+
+	cmd = "oc --kubeconfig=/opt/kubeconfig get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-master | head -n -1 | xargs -t oc --kubeconfig=/opt/kubeconfig delete"
+	_, stderr, err := runner.Run(cmd)
```
cfergeau, line 186: Could these various `oc` commandlines make use of `OcRunner`/`OcConfig`?

praveenkumar: @cfergeau `OcRunner` doesn't support pipe so if we really go with `OcRunner` then we need to do something like following.
```
mc, stderr, err := ocConfig.RunOcCommand("get mc --sort-by=.metadata.creationTimestamp --no-headers -oname")
[...]
oldmc, stderr, err := sshRunner.Run(fmt.Sprintf("%s | grep rendered-master | head -n -1", mc)
[...]
_, stderr,err := ocConfig.RunOcCommand(fmt.Sprintf("delete %s", oldmc))
```

cfergeau: We don't have to use `| grep| head` to achieve this, we could ask for json output and let go parse it (do we already vendor the module with the type definitions?), or use https://golang.org/pkg/bufio/#NewScanner to get what we are interested in?

praveenkumar: @cfergeau I will check again with json output.

praveenkumar:
CI failure on `e2e-crc` because of https://bugzilla.redhat.com/show_bug.cgi?id=1979550 one.

pkg/crc/machine/generate_bundle.go
```diff
@@ -33,17 +32,21 @@ func (client *client) GenerateBundle(forceStop bool) error {
...
+		} else {
+			return err
 		}
-		return err
 	}
-	if currentState != state.Stopped {
-		return fmt.Errorf("VM is not stopped, current state is %s", currentState.String())
+	running, err := client.IsRunning()
+	if err != nil {
+		return nil
```
cfergeau, line 46: `return err`?

pkg/crc/machine/generate_bundle.go
```diff
@@ -33,17 +32,21 @@ func (client *client) GenerateBundle(forceStop bool) error {
...
-	if err != nil {
+	if _, err := client.Stop(); err != nil {
 		if forceStop {
 			if err := client.PowerOff(); err != nil {
 				return err
 			}
+		} else {
+			return err
 		}
-		return err
```
cfergeau, line 39: The commit logs says ` Fix the state of VM when generating the bundle ` but it seems this is also fixing  the fact that we'd always error out when forced shutdown was triggered, even if it was successful?

Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /bin/bash
 
-BUNDLE_VERSION ?= 4.7.18
+BUNDLE_VERSION ?= 4.8.0
```
cfergeau, line 3: no `-rcX` in the versioning?

pkg/crc/cluster/cluster.go
```diff
@@ -192,18 +192,50 @@ func RemovePullSecretFromCluster(ocConfig oc.Config, sshRunner *ssh.Runner) erro
...
+	if len(sortedMachineConfigsWithTime) == 0 {
+		return nil
 	}
 
-	cmdArgs = []string{"delete", "machineconfigpools", "--all"}
-	_, stderr, err = ocConfig.RunOcCommand(cmdArgs...)
-	if err != nil {
-		return fmt.Errorf("Failed to remove machineconfigpools %w: %s", err, stderr)
+	var (
+		renderedMaster []string
```
cfergeau, line 213: could be `renderedMasters` and `renderedWorkers` since an array can hold multiple masters/workers.

pkg/crc/cluster/cluster.go
```diff
@@ -192,18 +192,50 @@ func RemovePullSecretFromCluster(ocConfig oc.Config, sshRunner *ssh.Runner) erro
...
+		deleteRenderedMachineConfig = fmt.Sprintf("%s %s", deleteRenderedMachineConfig, strings.Join(renderedWorker[:len(renderedWorker)-1], " "))
 	}
 
+	if deleteRenderedMachineConfig != "" {
+		_, stderr, err = ocConfig.RunOcCommand(fmt.Sprintf("delete %s", deleteRenderedMachineConfig))
+		if err != nil {
+			return fmt.Errorf("Failed to remove machineconfigpools %w: %s", err, stderr)
+		}
+	}
 	return nil
```
cfergeau, line 239: I have only build-tested this, but I think the code below does the same while being simpler (less code overall):
```
func RemoveOldRenderedMachineConfig(ocConfig oc.Config) error {
        // This block (#L179-183) should be removed as soon as we start shipping with 4.8 bundle.
        // This check is only make sure if there is any machineconfig resource or not because
        // in 4.7 we disabled mco and also deleted all the machineconfig/machineconfig-pools.
        // For 4.8 we don't disable mco and it does contain the machineconfigs.    
        stdout, stderr, err := ocConfig.RunOcCommand("get mc --no-headers -oname")
        if err != nil {
                return fmt.Errorf("failed to get machineconfig resource %w: %s", err, stderr)
        }
        machineConfigs := strings.Split(stdout, "\n")
        
        var ocDeleteArgs = []string{"delete"}
                                                            
        for _, mc := range machineConfigs {                 
                if strings.Contains(mc, "rendered-master") {   
                        ocDeleteArgs = append(ocDeleteArgs, mc)
                }                                              
                if strings.Contains(mc, "rendered-worker") {       
                        ocDeleteArgs = append(ocDeleteArgs, mc)
                }
        }
        if len(ocDeleteArgs) == 1 {
                // no machine configs to delete
                return nil
        }
                
        _, stderr, err = ocConfig.RunOcCommand(ocDeleteArgs...)
        if err != nil {
                return fmt.Errorf("Failed to remove machineconfigpools %w: %s", err, stderr)
        }
        return nil
}
```

cfergeau: (`Contains` could probably be `StartsWith`/`HasSuffix`?

cfergeau (commented):
Few more comments, but overall looking good :+1:

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau the above code will delete all the mc which have `rendered master` and `redered worker` but what we want only to delete the older one not the latest one.

pkg/crc/cluster/cluster.go
praveenkumar: In our case we have only single master/worker, config may be multiple.

Makefile
praveenkumar: @cfergeau we already have 4.8.0 bundle.

pkg/crc/machine/generate_bundle.go
praveenkumar: @cfergeau yes fixed it, created new commit to handle it.

guillaumerose (approved):

PR #2472, opened 2021-06-23
Bump golangci-lint to v1.41.1

guillaumerose:
This version is compatible with go1.15. golint linter is replaced by
revive.

### DIFF
--- a/test/extended/util/util.go
+++ b/test/extended/util/util.go
@@ -22,7 +22,8 @@ func CopyFilesToTestDir() error {
 	}
 
 	l := strings.Split(cwd, string(filepath.Separator))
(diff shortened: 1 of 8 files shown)


test/extended/util/util.go
```diff
@@ -22,7 +22,8 @@ func CopyFilesToTestDir() error {
 	}
 
 	l := strings.Split(cwd, string(filepath.Separator))
-	dataDirPieces := append(l[:len(l)-3], "testdata")
+	dataDirPieces := l[:len(l)-3]
+	dataDirPieces = append(dataDirPieces, "testdata")
```
praveenkumar, line 26: Good that it force on readability instead one liner :)

anjannath (approved):

Issue #2432, opened 2021-06-15
[RFE] Replace the admin kubeconfig  (system:admin) generated at install time


praveenkumar:
As per https://access.redhat.com/solutions/6054981 now it is possible to change the CA for kubeconfig file which is generated by the installer at installation time. 

We should try to see if this can be added as part of crc. This should also fix one of biggest security issue we are having same kubeconfig file for all the crc deployment which uses same bundle.

praveenkumar:
fixed by #2482

PR #2482, opened 2021-06-25
Replace kubeconfig file for each fresh start of crc 

praveenkumar:
**Fixes:** Issue #2432

## Solution/Idea
This patch tries to implement a KCS article about replacing the
kubeconfig file which generated during installation time.

- https://access.redhat.com/solutions/6054981

## Proposed changes
No changes from ux side.

## Testing
This patch make sure once the cluster is provisioned then old kubeconfig file which is part of bundle metadata should be outdated.
```
$ crc start 
$ oc --kubeconfig=$HOME/.crc/cache/<bundle_location>/kubeconfig whoami
error: You must be logged in to the server (Unauthorized)
```

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -11,11 +12,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/oc"
 	"github.com/code-ready/crc/pkg/crc/ssh"
+	crctls "github.com/code-ready/crc/pkg/crc/tls"
 	"github.com/code-ready/crc/pkg/crc/validation"
 	"github.com/pborman/uuid"
 )
@@ -139,6 +142,39 @@ func EnsurePullSecretPresentInTheCluster(ocConfig oc.Config, pullSec PullSecretL
 	if err != nil {
 		return fmt.Errorf("Failed to add Pull secret %v: %s", err, stderr)
 	}
+	return nil
+}
+
+func EnsureGeneratedClientCAPresentInTheCluster(ocConfig oc.Config, sshRunner *ssh.Runner, selfSignedCACert *x509.Certificate, adminCert string) error {
+	selfSignedCAPem := crctls.CertToPem(selfSignedCACert)
+	if err := WaitForOpenshiftResource(ocConfig, "configmaps"); err != nil {
+		return err
+	}
+	clusterClientCA, stderr, err := ocConfig.RunOcCommand("get", "configmaps", "admin-kubeconfig-client-ca", "-n", "openshift-config", "-o", `jsonpath="{.data.ca-bundle\.crt}"`)
+	if err != nil {
+		return fmt.Errorf("Failed to get config map %v: %s", err, stderr)
+	}
+
+	ok, err := crctls.VerifyCertificateAgainstRootCA(clusterClientCA, adminCert)
+	if err != nil {
+		return err
+	}
+	if ok {
+		return nil
+	}
+
+	logging.Info("Updating root CA cert to admine-kubeconfig-client-ca configmap...")
+	jsonPath := fmt.Sprintf(`'{"data": {"ca-bundle.crt": %q}}'`, selfSignedCAPem)
+	cmdArgs := []string{"patch", "configmap", "admin-kubeconfig-client-ca",
+		"-n", "openshift-config", "--patch", jsonPath}
+	_, stderr, err = ocConfig.RunOcCommand(cmdArgs...)
+	if err != nil {
+		return fmt.Errorf("Failed to patch admin-kubeconfig-client-ca config map with new CA` %v: %s", err, stderr)
+	}
+	if err := sshRunner.CopyFile(constants.KubeconfigFilePath, ocConfig.KubeconfigPath, 0644); err != nil {
+		return fmt.Errorf("Failed to copy generated kubeconfig file to VM: %v", err)
+	}
+
 	return nil
 }
 
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 	clientset "github.com/openshift/client-go/config/clientset/versioned"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -80,8 +79,8 @@ func (status *Status) IsReady() bool {
 	return status.Available && !status.Progressing && !status.Degraded && !status.Disabled
 }
 
-func GetClusterOperatorsStatus(ctx context.Context, ip string, bundle *bundle.CrcBundleInfo) (*Status, error) {
-	lister, err := kubernetesClient(ip, bundle)
+func GetClusterOperatorsStatus(ctx context.Context, ip string, kubeconfigFilePath string) (*Status, error) {
+	lister, err := kubernetesClient(ip, kubeconfigFilePath)
 	if err != nil {
 		return nil, err
 	}
@@ -157,16 +156,16 @@ type operatorLister interface {
 	List(ctx context.Context, opts metav1.ListOptions) (*openshiftapi.ClusterOperatorList, error)
 }
 
-func kubernetesClient(ip string, bundle *bundle.CrcBundleInfo) (*clientset.Clientset, error) {
-	config, err := kubernetesClientConfiguration(ip, bundle)
+func kubernetesClient(ip string, kubeconfigFilePath string) (*clientset.Clientset, error) {
+	config, err := kubernetesClientConfiguration(ip, kubeconfigFilePath)
 	if err != nil {
 		return nil, err
 	}
 	return clientset.NewForConfig(config)
 }
 
-func kubernetesClientConfiguration(ip string, bundle *bundle.CrcBundleInfo) (*restclient.Config, error) {
-	config, err := clientcmd.BuildConfigFromFlags("", bundle.GetKubeConfigPath())
+func kubernetesClientConfiguration(ip string, kubeconfigFilePath string) (*restclient.Config, error) {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigFilePath)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -27,6 +27,25 @@ const (
 	developerContext = "crc-developer"
 )
 
+func updateClientCrtAndKeyToKubeconfig(clientKey, clientCrt []byte, srcKubeconfigPath, destKubeconfigPath string) error {
+	cfg, err := clientcmd.LoadFromFile(srcKubeconfigPath)
+	if err != nil {
+		return err
+	}
+	var adminUserExist bool
+	for username, auth := range cfg.AuthInfos {
+		if username == "admin" {
+			adminUserExist = true
+			auth.ClientCertificateData = clientCrt
+			auth.ClientKeyData = clientKey
+		}
+	}
+	if !adminUserExist {
+		return fmt.Errorf("unable to find admin user in %s", srcKubeconfigPath)
+	}
+	return clientcmd.WriteToFile(*cfg, destKubeconfigPath)
+}
+
 func writeKubeconfig(ip string, clusterConfig *types.ClusterConfig) error {
 	kubeconfig := getGlobalKubeConfigPath()
 	dir := filepath.Dir(kubeconfig)
@@ -78,6 +97,19 @@ func certificateAuthority(kubeconfigFile string) ([]byte, error) {
 		return nil, fmt.Errorf("crc cluster not found in kubeconfig %s", kubeconfigFile)
 	}
 	return cluster.CertificateAuthorityData, nil
+}
+
+func adminClientCertificate(kubeconfigFile string) (string, error) {
+	builtin, err := clientcmd.LoadFromFile(kubeconfigFile)
+	if err != nil {
+		return "", err
+	}
+	users := builtin.AuthInfos
+	adminUser, ok := users["admin"]
+	if !ok {
+		return "", fmt.Errorf("admin user is not found in kubeconfig %s", kubeconfigFile)
+	}
+	return string(adminUser.ClientCertificateData), nil
 }
 
 // https://github.com/openshift/oc/blob/f94afb52dc8a3185b3b9eacaf92ec34d80f8708d/pkg/helpers/kubeconfig/smart_merge.go#L21
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -2,6 +2,8 @@ package machine
 
 import (
 	"context"
+	"crypto/rsa"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -26,9 +28,9 @@ import (
 	crcssh "github.com/code-ready/crc/pkg/crc/ssh"
 	"github.com/code-ready/crc/pkg/crc/systemd"
 	"github.com/code-ready/crc/pkg/crc/telemetry"
+	crctls "github.com/code-ready/crc/pkg/crc/tls"
 	"github.com/code-ready/crc/pkg/libmachine"
 	"github.com/code-ready/crc/pkg/libmachine/host"
-	crcos "github.com/code-ready/crc/pkg/os"
 	"github.com/code-ready/machine/libmachine/drivers"
 	"github.com/code-ready/machine/libmachine/state"
 	"github.com/docker/go-units"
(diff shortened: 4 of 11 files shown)


guillaumerose:
Does it change crc start duration ? Does it trigger a resync of all operators ?

cfergeau (commented):
A high level overview of what `Replace kubeconfig file for each fresh start of crc` is doing would be really useful, especially as it does not follow the KCS article it links to very closely.

When copying the kubeconfig file from the bundle to the machines directory, it uses `tls.GetClientKeyAndCert` to generate a new self-signed CA, and also generates a new client certificate for the `system:admin` user

This introduces a new `EnsureUpdatedUserCAPresentInTheCluster` to make the cluster aware of this new CA. Before that is done, `/opt/kubeconfig` must keep using the old CA, or authentication will fail. Once that is done, `/opt/kubeconfig` will be updated to use the new CA.

Together with `UpdateKubeAdminUserPassword`, this should ensure having access to one bundle does not mean one can access any cluster which was created using this bundle.

Is this about right?

pkg/crc/machine/start.go
```diff
@@ -447,6 +446,19 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		}
 	}
 
+	// Update the user cert to 'admin-kubeconfig-client-ca' configmap
+	userCA, err := userCA(constants.KubeconfigFilePath)
```
cfergeau, line 450: I'd prefer that we don't use the `user` terminology at all in this commit, this is all about changing the kubeconfig `admin` client CA both in the kubeconfig file and in the cluster. Nothing related to `user`.

pkg/crc/machine/kubeconfig.go
```diff
@@ -80,6 +99,19 @@ func certificateAuthority(kubeconfigFile string) ([]byte, error) {
...
+	builtin, err := clientcmd.LoadFromFile(kubeconfigFile)
+	if err != nil {
+		return "", err
+	}
+	users := builtin.AuthInfos
+	adminUser, ok := users["admin"]
+	if !ok {
+		return "", fmt.Errorf("crc-admin user is not found in kubeconfig %s", kubeconfigFile)
+	}
+	return string(adminUser.ClientCertificateData), nil
```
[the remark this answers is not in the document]
cfergeau: Looking again at `GetClientKeyAndCert()` and the overall commit. I'm not sure this is right.
`GetClientKeyAndCert()` generates a self-signed CA, and then uses this CA to sign an `admin` client certificate. We update the kubeconfig file with this certificate/key, this is going to be useful to authenticate with the cluster ( https://blog.cloudflare.com/introducing-tls-client-auth/ ). This is what `updateUserCaAndKeyToKubeconfig` is doing.

We also need to let the cluster know that our `admin` client certificate will now be signed with a new CA, this is what `EnsureUpdatedUserCAPresentInTheCluster` is doing by changing the `admin-kubeconfig-client-ca` configmap.
However, I'd expect that it only needs to be setting it to the self-signed CA that `GetClientKeyAndCert()` generated with `caConf` and that it does not need to know about the `admin` client certificate that `GetClientKeyAndCert()` also generated (using `adminUserConf`)?

This seems consistent with this KCS article, `${NAME}-ca.crt` is what is used in configmaps, and `${USER}.crt` is only used as a parameter to `oc config set-credentials`

pkg/crc/cluster/clusteroperator.go
```diff
@@ -157,16 +156,16 @@ type operatorLister interface {
...
 	if err != nil {
 		return nil, err
 	}
 	return clientset.NewForConfig(config)
 }
 
-func kubernetesClientConfiguration(ip string, bundle *bundle.CrcBundleInfo) (*restclient.Config, error) {
-	config, err := clientcmd.BuildConfigFromFlags("", bundle.GetKubeConfigPath())
+func kubernetesClientConfiguration(ip string, kubeconfigFilePath string) (*restclient.Config, error) {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigFilePath)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau yes, but I wanted have this function to pass kubeconfig path explicit in case if we want to add a unit test around it.

pkg/crc/machine/start.go
praveenkumar: `admin` is a user which clientCA need to updated so I can change it to `userClientCA` ?

pkg/crc/tls/tls.go
```diff
@@ -0,0 +1,234 @@
...
+
+	crt, err := SelfSignedCertificate(cfg, key)
+	if err != nil {
+		logging.Debugf("Failed to create self-signed certificate: %s", err)
+		return nil, nil, errors.Wrap(err, "failed to create self-signed certificate")
+	}
+	return key, crt, nil
+}
+
+func GetClientKeyAndCert() ([]byte, []byte, error) {
```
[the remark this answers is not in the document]
cfergeau: would be useful to mention this file/these functions in the comment so that they can be looked up as reference code if needed.

pkg/crc/machine/start.go
cfergeau: why the indirection? The CA will be used for the admin user, so let's call it `adminClientCA`? This is close to the name of the configmap: `admin-kubeconfig-client-ca`. Just `clientCA` would also work for me

pkg/crc/cluster/clusteroperator.go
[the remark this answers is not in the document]
cfergeau: This PR does not introduce a unit test, so passing this as an argument just adds unneeded complexity.

cfergeau:
> Does it change crc start duration ? Does it trigger a resync of all operators ?

@praveenkumar these questions from Guillaume are still pending.

cfergeau:
```
$ go test ./pkg/crc/machine/
--- FAIL: TestCleanKubeconfig (0.00s)
    kubeconfig_test.go:53: 
        	Error Trace:	kubeconfig_test.go:53
        	Error:      	Not equal: 
        	            	expected: map[string]interface {}{"apiVersion":"v1", "clusters":[]interface {}{map[string]interface {}{"cluster":map[string]interface {}{"certificate-authority-data":"ZW1wdHkK", "server":"https://192.168.42.71:8443"}, "name":"192-168-42-71:8443"}, map[string]interface {}{"cluster":map[string]interface {}{"certificate-authority-data":"ZW1wdHkK", "server":"https://127.0.0.1:33407"}, "name":"kind-reference"}}, "contexts":[]interface {}{map[string]interface {}{"context":map[string]interface {}{"cluster":"kind-reference", "user":"kind-reference"}, "name":"kind-reference"}, map[string]interface {}{"context":map[string]interface {}{"cluster":"192-168-42-71:8443", "namespace":"myproject", "user":"developer/192-168-42-71:8443"}, "name":"minishift"}}, "current-context":"", "kind":"Config", "preferences":map[string]interface {}{}, "users":[]interface {}{map[string]interface {}{"name":"developer/192-168-42-71:8443", "user":map[string]interface {}{"token":"token2"}}, map[string]interface {}{"name":"developer/api-crc-testing:6443", "user":map[string]interface {}{"token":"token3"}}, map[string]interface {}{"name":"kind-reference", "user":map[string]interface {}{"client-certificate-data":"ZW1wdHkK", "client-key-data":"ZW1wdHkK"}}}}
        	            	actual  : map[string]interface {}{"apiVersion":"v1", "clusters":[]interface {}{map[string]interface {}{"cluster":map[string]interface {}{"certificate-authority-data":"ZW1wdHkK", "server":"https://192.168.42.71:8443"}, "name":"192-168-42-71:8443"}, map[string]interface {}{"cluster":map[string]interface {}{"certificate-authority-data":"ZW1wdHkK", "server":"https://127.0.0.1:33407"}, "name":"kind-reference"}}, "contexts":[]interface {}{map[string]interface {}{"context":map[string]interface {}{"cluster":"kind-reference", "user":"kind-reference"}, "name":"kind-reference"}, map[string]interface {}{"context":map[string]interface {}{"cluster":"192-168-42-71:8443", "namespace":"myproject", "user":"developer/192-168-42-71:8443"}, "name":"minishift"}}, "current-context":"", "kind":"Config", "preferences":map[string]interface {}{}, "users":[]interface {}{map[string]interface {}{"name":"admin", "user":map[string]interface {}{"client-certificate-data":"ZW1wdHkK", "client-key-data":"ZW1wdHkK"}}, map[string]interface {}{"name":"developer/192-168-42-71:8443", "user":map[string]interface {}{"token":"token2"}}, map[string]interface {}{"name":"developer/api-crc-testing:6443", "user":map[string]interface {}{"token":"token3"}}, map[string]interface {}{"name":"kind-reference", "user":map[string]interface {}{"client-certificate-data":"ZW1wdHkK", "client-key-data":"ZW1wdHkK"}}}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -39,3 +39,10 @@
        	            	  },
        	            	- (string) (len=5) "users": ([]interface {}) (len=3) {
        	            	+ (string) (len=5) "users": ([]interface {}) (len=4) {
        	            	+  (map[string]interface {}) (len=2) {
        	            	+   (string) (len=4) "name": (string) (len=5) "admin",
        	            	+   (string) (len=4) "user": (map[string]interface {}) (len=2) {
        	            	+    (string) (len=23) "client-certificate-data": (string) (len=8) "ZW1wdHkK",
        	            	+    (string) (len=15) "client-key-data": (string) (len=8) "ZW1wdHkK"
        	            	+   }
        	            	+  },
        	            	   (map[string]interface {}) (len=2) {
        	Test:       	TestCleanKubeconfig
```

pkg/crc/cluster/cluster.go
```diff
@@ -142,6 +145,39 @@ func EnsurePullSecretPresentInTheCluster(ocConfig oc.Config, pullSec PullSecretL
...
+
+	ok, err := crctls.VerifyCertificateAgainstRootCA(stdout, userCA)
+	if err != nil {
+		return err
+	}
+	if ok {
+		return nil
+	}
+
+	logging.Info("Updating root CA cert to admine-kubeconfig-client-ca configmap...")
```
cfergeau, line 166: 'admine' -> 'admin'. Maybe just "Updating admin-kubeconfig-client-ca configmap"?

pkg/crc/tls/tls.go
```diff
@@ -0,0 +1,277 @@
...
+	certInPem := pem.EncodeToMemory(
+		&pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: cert.Raw,
+		},
+	)
+	return certInPem
+}
+
+func VerifyCertificateAgainstRootCA(rootPEM, clientPEM string) (bool, error) {
```
cfergeau, line 252: This function takes a CA and a certificate (its name says so), but this is not reflected in the argument names? Maybe `CA` and `cert`, and leave the fact that they must be PEM-encoded up to a comment?

praveenkumar:
> Does it change crc start duration ? Does it trigger a resync of all operators ?

@guillaumerose As per my testing it doesn't change much on crc start side. It resync the kube-apiserver and that's the reason I do it just before we start waiting for cluster to comeup.

pkg/crc/machine/start.go
```diff
@@ -548,10 +564,6 @@ func createHost(api libmachine.API, machineConfig config.MachineConfig) error {
 	if err := cluster.GenerateKubeAdminUserPassword(); err != nil {
 		return errors.Wrap(err, "Error generating new kubeadmin password")
 	}
-	if err := copyKubeconfig(machineConfig.Name, machineConfig); err != nil {
```
guillaumerose, line 551: If you remove this, there is a small amount of time during `crc start` where `crc status` will be broken.
The machine will be marked as created but the kubeconfig file will be absent.

pkg/crc/machine/start.go
```diff
@@ -447,6 +449,21 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
...
+		return nil, errors.Wrap(err, "Not able to generate root CA key and Cert")
+	}
+	if err := copyKubeconfigFileWithUpdtedUserClientCACertAndKey(rootCAKey, rootCACert, crcBundleMetadata.GetKubeConfigPath(), constants.KubeconfigFilePath); err != nil {
+		return nil, errors.Wrapf(err, "Failed to copy kubeconfig file: %s", constants.KubeconfigFilePath)
+	}
+	userClientCA, err := userClientCA(constants.KubeconfigFilePath)
+	if err != nil {
+		return nil, errors.Wrap(err, "Not able to get user CA")
+	}
+	if err := cluster.EnsureUpdatedRootCAPresentInTheCluster(ocConfig, sshRunner, rootCACert, userClientCA); err != nil {
```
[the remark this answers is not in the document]
praveenkumar: > In particular, I don't think this has any impact on the certificates used for https://api.crc.testing?

@cfergeau No, how about `EnsureGeneratedClientCAPresentInTheCluster` , I am also bit confuse about naming it?

pkg/crc/tls/tls.go
```diff
@@ -0,0 +1,277 @@
...
+		IsCA:      true,
+	}
+	return GenerateSelfSignedCertificate(rootCAConf)
+}
+
+func GetClientKeyAndCert(rootCAKey *rsa.PrivateKey, rootCACert *x509.Certificate) ([]byte, []byte, error) {
+	adminUserConf := &CertCfg{
+		Subject:      pkix.Name{CommonName: "system:admin", OrganizationalUnit: []string{"system:masters"}},
+		KeyUsages:    x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau As per https://access.redhat.com/solutions/6054981 it only need `clientAuth` but I think I will use  same extkeyusage as part of installer side.

praveenkumar:
retest this please.

pkg/crc/machine/kubeconfig.go
```diff
@@ -27,6 +27,20 @@ const (
...
+	cfg, err := clientcmd.LoadFromFile(srcKubeconfigPath)
+	if err != nil {
+		return err
+	}
+	for username, auth := range cfg.AuthInfos {
+		if username == "admin" {
+			auth.ClientCertificateData = clientCrt
+			auth.ClientKeyData = clientKey
+		}
+	}
```
cfergeau, line 40: Are you sure this works as expected? Isn't `auth` a copy of `cfg.AuthInfos["admin"]`? Changing `auth` would not modify `cfg.AuthInfos`?
https://stackoverflow.com/questions/42716852/how-to-update-map-values-in-go led me to try https://play.golang.org/p/iUGDFdnSAVr and ask these questions :)

pkg/crc/cluster/cluster.go
cfergeau: Still pending

cfergeau (commented):
` Use rootCA to generate cert and key for clientt (in our cas  `system:admin` user)`

`clientt` typo. Imo it's more accurate to say `Use root CA to generate a client certificate for the `system:admin` user

`update it's` -> `update its`


`is signed by root CA and if not then update the rootCA which is` -> Maybe explicitly say `is signed by the/this new root CA`

pkg/crc/machine/kubeconfig.go
praveenkumar: @cfergeau it is `map[string]*Auth` so https://play.golang.org/p/3Pomaiu7mfq is something which replicate it and it does change the value.

cfergeau: Ah right, forgot to check the type of the values :) A pointer is good. Nit: not sure why we iterate over the map when we could be looking up `cfg.AuthInfos["admin"]` directly.

pkg/crc/machine/start.go
```diff
@@ -713,3 +706,24 @@ func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Confi
 	}
 	return nil
 }
+
+func updateKubeconfig(ocConfig oc.Config, sshRunner *crcssh.Runner, srcKubeconfigFilePath, destKubeconfigPath string) error {
+	if _, err := os.Stat(destKubeconfigPath); err == nil {
```
cfergeau, line 711: This could use `crcos.FileExists()`

pkg/crc/machine/start.go
```diff
@@ -713,3 +706,24 @@ func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Confi
...
+		return nil
+	}
+	selfSignedCAKey, selfSignedCACert, err := crctls.GetSelfSignedCA()
+	if err != nil {
+		return errors.Wrap(err, "Not able to generate root CA key and Cert")
+	}
+	if err := copyKubeconfigFileWithUpdatedUserClientCertAndKey(selfSignedCAKey, selfSignedCACert, srcKubeconfigFilePath, destKubeconfigPath); err != nil {
+		return errors.Wrapf(err, "Failed to copy kubeconfig file: %s", constants.KubeconfigFilePath)
+	}
+	adminClientCA, err := adminClientCertificate(constants.KubeconfigFilePath)
```
cfergeau, line 721: This is not a CA, this is a certificate.

cfergeau (approved):
Still the naming of `adminClientCA`  left to be fixed, otherwise now looks good!

pkg/crc/machine/start.go
praveenkumar: @cfergeau resolved, missed this one :(

cfergeau (approved):

pkg/crc/machine/kubeconfig.go
```diff
@@ -27,6 +27,20 @@ const (
...
+	return clientcmd.WriteToFile(*cfg, destKubeconfigPath)
```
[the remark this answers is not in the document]
praveenkumar: if we don't find the admin user that means our source kubeconfig file is corrupted which is generated as part of installer, so I didn't give it much thought about not having the `admin` user but if we go with code logic it will not fail/warning or even create it. In this case it will just dump source kubeconfig to destination kubeconfig path.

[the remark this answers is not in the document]
praveenkumar: @cfergeau We should return an error in case `admin` user is not found which means that kubeconfig is not updated properly.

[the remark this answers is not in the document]
cfergeau: I did not see the latest iteration, looks like this could be simplified to:
```
auth, adminUserExists := cfg.AuthInfos["admin"]
if !adminUserExists {
    return fmt.Errorf("...")
}
auth.ClientCertificateData = clientCrt
auth.ClientKeyData = clientKey
```

Issue #2481, opened 2021-06-25
[BUG] crc setup fails on Windows with installer


jeffmaury:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? N/A
  * Running CRC on: Laptop

## CRC version
```bash
CodeReady Containers version: 1.28.0+08de64bd
OpenShift version: 4.7.13 (not embedded in executable)
```
  
## CRC status
```bash
DEBU CodeReady Containers version: 1.28.0+08de64bd
DEBU OpenShift version: 4.7.13 (not embedded in executable)
DEBU Running 'crc status'
DEBU Checking file: C:\Users\JeffMAURY\.crc\machines\crc\.crc-exist
DEBU Running '(Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion" -Name ReleaseId).ReleaseId'
DEBU Running '(Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").EditionID'
Machine does not exist. Use 'crc start' to create it
```

## CRC config
```bash
- consent-telemetry                     : yes
```

## Host Operating System
```bash

Nom de l’hôte:                              DESKTOP-U2EVGBM
Nom du système d’exploitation:              Microsoft Windows 10 Professionnel
Version du système:                         10.0.19043 N/A build 19043
Fabricant du système d’exploitation:        Microsoft Corporation
Configuration du système d’exploitation:    Station de travail autonome
Type de build du système d’exploitation:    Multiprocessor Free
Propriétaire enregistré:                    Jeff MAURY
Organisation enregistrée:                   
Identificateur de produit:                  00330-80819-37372-AA911
Date d’installation originale:              29/06/2020, 14:54:33
Heure de démarrage du système:              17/06/2021, 15:40:32
Fabricant du système:                       LENOVO
Modèle du système:                          20LAS3NJ0H
Type du système:                            x64-based PC
Processeur(s):                              1 processeur(s) installé(s).
                                            [01] : Intel64 Family 6 Model 142 Stepping 10 GenuineIntel ~1910 MHz
Version du BIOS:                            LENOVO N27ET36W (1.22 ), 04/07/2019
Répertoire Windows:                         C:\WINDOWS
Répertoire système:                         C:\WINDOWS\system32
Périphérique d’amorçage:                    \Device\HarddiskVolume1
[...]

PR #2483, opened 2021-06-25
Issue #2481 use current crc location to determine the msi installation path

anjannath:
Since both are copied to the same location by the msi it is expected to find
crc-tray in the same directory

Fixes: #2481

Relates to: PR https://github.com/code-ready/tray-windows/pull/89

## Testing

Install crc using the msi in location other then `C:\Program Files\CodeReady Containers` and then running `crc setup` should not fail.

(Note: the tray will be started but will immediately fail with an error saying unable to find crc.exe this is fixed by https://github.com/code-ready/tray-windows/pull/89)

### DIFF
--- a/cmd/crc-embedder/cmd/root.go
+++ b/cmd/crc-embedder/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
@@ -26,6 +27,11 @@ when building the crc executable for release`,
 }
 
 func init() {
+	err := constants.EnsureBaseDirectoriesExist()
+	if err != nil {
+		fmt.Println("CRC base directories are missing: ", err)
+		os.Exit(1)
+	}
 	rootCmd.PersistentFlags().StringVar(&logging.LogLevel, "log-level", constants.DefaultLogLevel, "log level (e.g. \"debug | info | warn | error\")")
 }
 
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 
 	"github.com/YourFin/binappend"
+	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/version"
 )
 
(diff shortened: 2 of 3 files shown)


pkg/crc/constants/constants.go
```diff
@@ -200,5 +200,7 @@ func GetCRCWindowsTrayDownloadURL() string {
 }
 
 func GetMsiInstallPath() string {
-	return filepath.Join(os.Getenv("PROGRAMFILES"), "CodeReady Containers")
+	// In case of error path will be empty string and upperr layer will handle
+	currentExecutablePath, _ := os.Executable()
```
guillaumerose, line 204: Maybe log the error ? Otherwise lgtm

anjannath: should we do logging in the constants package? there's an import cycle if i try to import the logging package :(

anjannath: Updated, logging the error now and fixed the import cycle in the next commit.

cmd/crc-embedder/cmd/root.go
```diff
@@ -26,6 +26,10 @@ when building the crc executable for release`,
 }
 
 func init() {
+	err := constants.EnsureBaseDirectoriesExist()
+	if err != nil {
+		fmt.Println("CRC base directories are missing: ", err)
+	}
```
anjannath, line 32: Should `return` or `os.Exit` maybe?

guillaumerose: yes as you wish. It is not production code. crc does `logging.Fatal()`.

pkg/crc/logging/logging.go
```diff
@@ -18,10 +17,6 @@ var (
 )
 
 func OpenLogFile(path string) (*os.File, error) {
-	err := constants.EnsureBaseDirectoriesExist()
```
guillaumerose, line 21: good to remove this!

cmd/crc-embedder/cmd/root.go
anjannath: added `return`

guillaumerose: If you add return here, it won't add the flag and continue the execution ? I think you should either use logging.Fatal or os.Exit but not return.

anjannath: updated, used `os.Exit(1)`

guillaumerose (approved):

guillaumerose:
/retest
retest this please

PR #2488, opened 2021-06-28
Move WaitForSSH to ssh package from cluster package

praveenkumar:


pkg/crc/cluster/cluster.go
```diff
@@ -23,18 +23,6 @@ import (
 // #nosec G101
 const vmPullSecretPath = "/var/lib/kubelet/config.json"
 
-func WaitForSSH(ctx context.Context, sshRunner *ssh.Runner) error {
-	checkSSHConnectivity := func() error {
```
cfergeau, line 27: Why move it?

pkg/crc/ssh/ssh.go
```diff
@@ -92,3 +95,15 @@ err     : %w\n`, command, err)
 
 	return string(stdout), string(stderr), nil
 }
+
+func WaitForSSH(ctx context.Context, sshRunner *Runner) error {
+	checkSSHConnectivity := func() error {
```
cfergeau, line 100: Should it be `sshRunner.WaitForConnectivity(ctx)` rather than `WaitForSSH(ctx, sshRunner)`? Maybe the timeout should be passed as a parameter?

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau because as my understanding this is more appropriate location for this function this is nothing to do with cluster side but waiting for SSH connection.

pkg/crc/ssh/ssh.go
praveenkumar: @cfergeau will check that, I just focused on the moving it to this package.

pkg/crc/cluster/cluster.go
cfergeau: Fine with me, but it's always useful to document the rationale in the commit log, both for easier reviews, and for posterity

pkg/crc/ssh/ssh.go
cfergeau: Nit: It's in the `ssh` package, so the `SSH` mention in `WaitForSSHConnectivity` is redundant.

cfergeau (approved):
Would be nice to expand the 1st commit log a bit before pushing.

Issue #2492, opened 2021-06-29
Status command over http and cli are inconsistent; Success = false, and 500


gbraad:
In the case of no VM:

```
c:\Program Files\CodeReady Containers>crc status
Machine does not exist. Use 'crc start' to create it
```

On the CLI the following can be observed:
```
c:\Program Files\CodeReady Containers>crc status -o json
```
```json
{
  "success": false,
  "error": "Machine does not exist. Use 'crc start' to create it"
}
```

However, when performing this call from HTTP, a 500 is returned; with the content body being: 

     Cannot load machine: machine crc does not exist

It would be better to return 200 and the same JSON result message.

  * 200 indicating the request was handled
  * `Success: false` to indicate a failure to perform the request action

If a 500 is considered, the message would still have to be consistent with the CLI and returns proper JSON. Currently deserialization fails and need to add an alternative to handle this error case.

gbraad:
This is a pattern that needs to be consistent throughout the protocol.

However, the No VM state is also considered as follows: #2023.

gbraad:
In https://github.com/code-ready/crc/blob/f543acc59bc2b5b721e5ac1290154596dd7c7d2d/pkg/crc/api/api_http.go#L176 we define this to be `application/json`, but the code:
https://github.com/code-ready/crc/blob/f543acc59bc2b5b721e5ac1290154596dd7c7d2d/pkg/crc/api/api_http.go#L181-L186
doesn't return this?

gbraad:
@anjannath Shouldn't this be:

```diff
	} else {
+		w.Header().Set("Content-Type", "application/json") 
		w.WriteHeader(http.StatusInternalServerError)
- 		if _, err := io.WriteString(w, result.Error); err != nil { 
+		if _, err := io.WriteString(w, result); err != nil {
			logging.Error("Failed to send response: ", err)
		}
	}
```

gbraad:
It seems the `-o json` command does nothing 'special' to get the simple error struct, but the current API returns:

```json
{
   "CrcStatus": "",
   "OpenshiftStatus": "",
   "OpenshiftVersion": "",
   "DiskUse": 0,
   "DiskSize": 0,
   "Error":"Machine doesn't exist",
   "Success":false
}
```

... so wondering where this inconsistency occurs.

Issue #2498, opened 2021-06-30
[RFE] [Windows] Allow HTTP backend also be used over standard TCP/IP (instead of named pipe)


gbraad:
Named pipes is a great way to communicate on Windows, but provides issues when wanting to forward/reroute or even debug on Windows. Tools like Thunder/Postman do not support the use of Named Pipes.

PR #2501, opened 2021-06-30
Issue #2492 return errors as JSON success=false

gbraad:
**Fixes:** Issue #2492

With this change the returned JSON is:

```json
{
   "Error":"Machine doesn't exist",
   "Success":false
}
```

### Steps to verify:

  * `crc delete`
  * verify message
    * socket mangling on macOS for `/api/status`
    * breakpoint at `DaemonCommander.cs`'s `getResultsForBasicCommand` to inspect the output on Windows
 
---
Re-added this comment:

Note: filing this to start a discussion

### DIFF
--- a/pkg/crc/api/api_http.go
+++ b/pkg/crc/api/api_http.go
@@ -179,8 +179,10 @@ func sendResponse(w http.ResponseWriter, response string) {
 			logging.Error("Failed to send response: ", err)
 		}
 	} else {
+		// Success = false, so only return the basic Result
+		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusInternalServerError)
-		if _, err := io.WriteString(w, result.Error); err != nil {
+		if _, err := io.WriteString(w, encodeStructToJSON(result)); err != nil {
 			logging.Error("Failed to send response: ", err)
 		}
 	}


gbraad:
The Windows tray experiences no issues from this change, but looking at the macOS version it seems to rely on a workaround to wrap these in a DaemonError: https://github.com/code-ready/tray-macos/blob/007e3c872462cc5dded60b638d2bef7a84b6704d/CodeReady%20Containers/DaemonCommander/DaemonCommander.swift#L64-L67

... which in this case will be the full JSON message instead.

cfergeau:
By reading this and the commit log, I'm not sure how to reproduce the initial issue, nor what is the behaviour before the change :-/ 
Does this still return an http error? Or does this now return an http 200 + success = false?

cfergeau:
Looking closer at the code, and reading carefully the commit log, before this change, an error 500 is returned followed by an error message as a string (not wrapped in json). After this change, an error 500 is still returned, but the body is now json, simillar to the one we get with `-ojson` on the commandline.

gbraad:
... as described in the issue ;-)

gbraad:
While not necessary for the Windows tray:

  * https://github.com/code-ready/tray-windows/pull/93

cfergeau:
> ... as described in the issue ;-)

There are already commit logs and a PR description, and the PR template says to describe these steps. Having to look in a 3rd location for information on a 1 commit PR seems quite heavyweight..

gbraad:
Unfortunately it is not possible on Windows to verify this output due to #2498, however on macOS something like

```
$ crc delete
$ nc socketpath/api/status
```

before the PR would show a `500` with a body content of `VM does not exist`, and after `500` and `{ Success: false, ...}`

Note: do not ask me about these macOS details as I am unable to verify this on a broken machine.

gbraad:
@anjannath has verified that I was right about https://github.com/code-ready/crc/pull/2501#issuecomment-871140517 and some changes are needed to correctly deserialize the message, by changing (or removing) the daemonerror abstraction.
screenshot_2021-06-30_at_12 17 26_pm_720

guillaumerose:
`success` field is a duplicate of the status code. It was important for the binary protocol but it is redundant with http, that's why it is not sent.

Are you planning to send more complex error ? If it is just a string (and I think it will stay a string for a long time), we don't really need this struct.

gbraad:
> success field is a duplicate of the status code.

I disagree with you, as it provides a correct implementation and consistency.

https://datatracker.ietf.org/doc/html/rfc7231#section-6.6

```
The 5xx (Server Error) class of status code indicates that the server is aware that it has erred or is incapable of performing the requested method.
```

In the case of a status, this can NOT be a 500 indicating an error. The VM is merely not existing. It would be clearer when you mention a `start` or `stop` failing, but still you need to distinguish between the status codes to determine the output message. I do not see why this is a necessity.

cfergeau:
For the "VM not existing" case, the problem is that `machine.Status()` returns an error in this case when it probably should not. Changing the api/http code to send `success true/false` is more of a workaround for this. Or we could also do some error type -> status code conversion in the api/http code so that 500 is not returned for all errors.

gbraad:
The 500 is a severe error that I am not challenging yet, but this for situations outside of the scope of normal use; unexpected or unable to further specify. A missing VM is not unexpected.

In the Windows implementation I allowed 500 to be passed on. The implementation for the clients is very simple if the 'Result' is passed on and simply checked for a boolean, but instead a try/carch block and rewrapping is not considered a workaround? The 500 should not happen unless the error is severe and the request can't be handled properly, or we allow the use and properly pass JSON data. Because if not, you will have to wrap an internal representation of a result ... I know what is the 'cleaner' solution. And that includes a revision of the VM states and how this results in an error.

guillaumerose:
+1 with @cfergeau.
I think we should either return Stopped state if the VM doesn't exist (and be careful of the behavior of the crc CLI) or introduce a new state.

> In the Windows implementation I allowed 500 to be passed on. 

This is the issue. Yes it simplifies the tray code but it pushes client behavior to the daemon.

cfergeau:
> The implementation for the clients is very simple if the 'Result' is passed on and simply checked for a boolean, but instead a try/carch block and rewrapping is not considered a workaround?

Even if you can check for result.success, don't you need a try/catch block in case something goes very wrong during the interaction with the daemon?
The discussion seems to revolve around "is our json API an additional layer on top of http? or is it at the same layer?".
If it is the former, then yes, the json API should build its own error reporting ("success true/false"). If the latter, it makes more sense to reuse http status codes for errors, and avoid a redundant "success" (which at some point will most likely result in weird corner cases, 200 + success = false, or 500 + success = true).

gbraad:
If you agree 500 is the inappropriate error message, it means the client can interpret 500 to be an exception, which means NO valid message is returned. Since in the current situation it does not provide a valid response for a common scenario.

gbraad:
> "is our json API an additional layer on top of http? or is it at the same layer?".

Shouldn't this have been answered before this got implemented. At several points in time I have raised this question and got no clear answer or description of the protocol, except for a reference to https://github.com/code-ready/machine/blob/master/libmachine/state/state.go

If `Success` is redundant, than why did it get added to the response structs?

cfergeau:
> If `Success` is redundant, than why did it get added to the response structs?

Dunno why it got added, never really tried to remove it as I was quite sure there would be some resistance to that ;)

> Shouldn't this have been answered before this got implemented.

Protocol documentation definitely would be nice to have. For now we seem to be fine changing the protocol as needed, rather maintaining compatibility, so documentation is less useful.

gbraad:
These are several examples of the explicit setting of the `Success`:
  * https://github.com/code-ready/crc/blob/1a152e21d8a5604a9c21841955285a09aebc303a/pkg/crc/api/handlers.go#L36
  * https://github.com/code-ready/crc/blob/1a152e21d8a5604a9c21841955285a09aebc303a/pkg/crc/api/handlers.go#L72
  * https://github.com/code-ready/crc/blob/1a152e21d8a5604a9c21841955285a09aebc303a/pkg/crc/api/handlers.go#L79
  * https://github.com/code-ready/crc/blob/1a152e21d8a5604a9c21841955285a09aebc303a/pkg/crc/api/handlers.go#L107

so if it means nothing, why add this to the protocol. It adds complexity and inconsistency. Either we use the `Success` state to avoid confusion, or we get rid of it. Merely a 500 status code check is not enough as can be seen here:

https://github.com/code-ready/crc/blob/1a152e21d8a5604a9c21841955285a09aebc303a/pkg/crc/api/handlers.go#L67-L88

guillaumerose:
`success` is present because the binary protocol doesn't have status code feature built-in. 
It is in the http output because we wanted to avoid duplicating too much code when transitioning from binary to http protocol. 
Now, we can clean this up!

gbraad:
See the start code. It does not error with a 500, and you mentioned the current Windows tray (HTTP implementation) does not provide feedback why an error occurred. Well, because the codebase isn't consistent and I refuse to add workaround for this.

cfergeau:
We probably all agree that http error reporting in the daemon needs improving. At this point we disagree regarding the way forward though (`success` everywhere VS more http statuses). Since the proposed change impacts the macos tray, I'd suggest to hold this, and look at a bigger overhaul of the http api ?

gbraad:
I modified the Windows tray to use Exceptions, but due to the async nature it needs handling of the AggregateException https://github.com/code-ready/tray-windows/pull/94 which can complicate the code a lot.

gbraad:
> I'd suggest to hold this, and look at a bigger overhaul of the http api ?

Merely wanted to start a discussion, as adding a half-way solution to a new codebase to 'handle' errors/failure is not ideal. For instance, config set/unset depends on the 'Error' field in report back the success state.

https://github.com/code-ready/tray-windows/blob/02e75987cd852b310726b6aa1f5dca0555967748/Forms/CrcSettingsForm.cs#L143-L165

anjannath:
`Success` and `Error` fields were needed in the old protocol, for the http protocol we need to clearly define the status codes as in which code represent what error in the daemon and document that as part of the api doc and the clients can act accordingly.

PR #2523, opened 2021-07-05
Issue 2468 Add forceStop flag for bundle generate command

praveenkumar:
Recently with 4.8 testing one of the blocking issue we found is vm taking longer time to shutdown gracefully
(~7-8 mins) and a bug is filled against it https://bugzilla.redhat.com/show_bug.cgi?id=1965992.
As part of bundle creation we first stop the cluster and that run client.Stop functionality
which waits for 2 min before error out and currently if there is any error we just error out
from the bundle generate command. This patch adds a `forceStop` flag for `bundle generate` command
to make sure if user uses it then force stop the VM otherwise error out.

### DIFF
--- a/pkg/crc/machine/generate_bundle.go
+++ b/pkg/crc/machine/generate_bundle.go
@@ -16,7 +16,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (client *client) GenerateBundle() error {
+func (client *client) GenerateBundle(forceStop bool) error {
 	bundleMetadata, sshRunner, err := loadVM(client)
 	if err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


pkg/crc/machine/generate_bundle.go
```diff
@@ -31,6 +31,11 @@ func (client *client) GenerateBundle() error {
 	// Stop the cluster
 	currentState, err := client.Stop()
 	if err != nil {
+		if forceStop {
```
anjannath, line 34: Should this branch off before we call `client.Stop()` user will still have to wait for 2mins even if `force` flag is used?

praveenkumar: @anjannath yes, this is same experience what user have for `stop` command where even using `-f` need to wait for 2 mins.

PR #2532, opened 2021-07-05
e2e crc commands with envs repect platfom syntax

### LINKED ISSUE #2531: [e2e] Fix crc command execution with envs on windows
PR #2443 disabled update check from CI, the synxtax is valid for unix based platforms but it fails on windows.

adrianriobo:
Fixes #2531

### DIFF
--- a/test/e2e/crcsuite/crcsuite.go
+++ b/test/e2e/crcsuite/crcsuite.go
@@ -18,7 +18,7 @@ import (
 
 	clicumber "github.com/code-ready/clicumber/testsuite"
 	"github.com/code-ready/crc/test/e2e/crcsuite/ux"
-	"github.com/code-ready/crc/test/extended/crc/cmd"
+	crcCmd "github.com/code-ready/crc/test/extended/crc/cmd"
 	"github.com/code-ready/crc/test/extended/util"
 )
 
@@ -49,7 +49,7 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^setting config property "(.*)" to value "(.*)" (succeeds|fails)$`,
 		SetConfigPropertyToValueSucceedsOrFails)
 	s.Step(`^unsetting config property "(.*)" (succeeds|fails)$`,
-		cmd.UnsetConfigPropertySucceedsOrFails)
+		crcCmd.UnsetConfigPropertySucceedsOrFails)
 	s.Step(`^login to the oc cluster (succeeds|fails)$`,
 		LoginToOcClusterSucceedsOrFails)
 	s.Step(`^setting kubeconfig context to "(.*)" (succeeds|fails)$`,
@@ -62,6 +62,10 @@ func FeatureContext(s *godog.Suite) {
 		CheckCRCStatus)
 	s.Step(`^(stdout|stderr) (?:should contain|contains) "(.*)" if bundle (is|is not) embedded$`,
 		CommandReturnShouldContainIfBundleEmbeddedOrNot)
+	s.Step(`^execute crc (.*) command$`,
+		ExecuteCommand)
+	s.Step(`^execute crc (.*) command (.*)$`,
+		ExecuteCommandWithExpectedExitStatus)
 
 	// CRC file operations
 	s.Step(`^file "([^"]*)" exists in CRC home folder$`,
@@ -159,7 +163,7 @@ func FeatureContext(s *godog.Suite) {
 	})
 
 	s.AfterSuite(func() {
-		err := cmd.DeleteCRC()
+		err := crcCmd.DeleteCRC()
 		if err != nil {
 			fmt.Printf("Could not delete CRC VM: %s.", err)
 		}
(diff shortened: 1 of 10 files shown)


guillaumerose:
clicumber contains 7 files. Could we include them in this repo and directly modify the env ?
It will be easier to maintain also.

adrianriobo:
About including the files in this repo seems fine to me. I will create a new PR for this. About  modify the env directly what do you mean?

guillaumerose:
It is an open question, not a specific ask. I don't know what others think about that. Don't do it now ^^

test/e2e/crcsuite/crcsuite.go
```diff
@@ -112,11 +112,11 @@ func FeatureContext(s *godog.Suite) {
 		}
 
 		// remove $HOME/.crc
-		err = util.RemoveCRCHome(CRCHome)
-		if err != nil {
-			fmt.Println(err)
-			os.Exit(1)
-		}
+		// err = util.RemoveCRCHome(CRCHome)
```
guillaumerose, line 115: Is it wanted ?

test/extended/crc/cmd/cmd.go
```diff
@@ -15,6 +15,12 @@ const (
 	CRCExecutableNotInstalled = "notInstalled"
 )
 
+type Command struct {
```
guillaumerose, line 18: Good! In the future we could have a builder like `crc("setup").WithoutNTP().WithoutUpdateCheck()`.

test/extended/crc/cmd/cmd_unix.go
```diff
@@ -0,0 +1,16 @@
+// +build !windows
+
+package cmd
+
+import "fmt"
+
+func (c Command) ToString() string {
```
guillaumerose, line 7: I played a bit with the code. Here is what I got:

```

func (c Command) ToString() string {
	cmd := append(c.env(), "crc", c.Action)
	return strings.Join(cmd, " ")
}

func (c Command) env() []string {
	var env []string
	if c.DisableUpdateCheck {
		env = append(env, envVariable("CRC_DISABLE_UPDATE_CHECK", "true"))
	}
	if c.DisableNTP {
		env = append(env, envVariable("CRC_DEBUG_ENABLE_STOP_NTP", "true"))
	}
	return env
}

func envVariable(key, value string) string {
		return fmt.Sprintf("$env:%s=%s;", key, value)
	}
	return fmt.Sprintf("%s=%s", key, value)
}

func TestCommand(t *testing.T) {
	assert.Equal(t, "crc setup", Command{
		Action: "setup",
	}.ToString())
	assert.Equal(t, "CRC_DEBUG_ENABLE_STOP_NTP=true crc start -b bundle", Command{
		Action:     "start -b bundle",
		DisableNTP: true,
	}.ToString())
	assert.Equal(t, "CRC_DISABLE_UPDATE_CHECK=true CRC_DEBUG_ENABLE_STOP_NTP=true crc start -b bundle", Command{
		Action:             "start -b bundle",
		DisableNTP:         true,
		DisableUpdateCheck: true,
	}.ToString())
}

```

It avoids to have a different file for Windows and share a bit more the code.

adrianriobo: Cool seems fine, will add it.

test/e2e/crcsuite/crcsuite.go
adrianriobo: no no I just realize, I was undoing :), but this need to be reviewed as now is failing on windows, because if crc-admin-helper process is running this remove fails (just commented to test on windows) but forget undo 

Thx though!

test/e2e/features/basic.feature
```diff
@@ -4,12 +4,18 @@ Feature: Basic test
     User explores some of the top-level CRC commands while going
     through the lifecycle of CRC.
 
-    @darwin @linux @windows
+    @darwin @linux
     Scenario: CRC version
         When executing "CRC_DISABLE_UPDATE_CHECK=true crc version" succeeds
```
praveenkumar, line 9: @adrianriobo Shouldn't we just have a version function as part of `crcsuite.go` so we don't have to maintain the env setting in the feature file?

adrianriobo: Yeah definitely, I would prefer move towards that direction instead of executing...

adrianriobo: I will change this

test/e2e/crcsuite/crcsuite.go
```diff
@@ -62,6 +62,10 @@ func FeatureContext(s *godog.Suite) {
 		CheckCRCStatus)
 	s.Step(`^(stdout|stderr) (?:should contain|contains) "(.*)" if bundle (is|is not) embedded$`,
 		CommandReturnShouldContainIfBundleEmbeddedOrNot)
+	s.Step(`^execute (.*) command$`,
```
praveenkumar, line 65: This looks very generic because it can capture `execute oc version command` which we don't want. Can't we do this something like `^execute crc (.*) coomand$` which will make sure it only capture crc specific commands?

Issue #2461, opened 2021-06-21
[qe] Improve product testability during sprint


adrianriobo:
Currently the testing during development phase of sprints is made building the crc project, but there are other components involved which should be tested as well. 

The idea would be generate a partial distributable every day which will be the base for running tests. To accomplish this several actions should be implemented. 

This issue will track the actions related to accomplish this goal

- [ ] Decouple distributable build from release process (use current component builds instead of embber)
- [ ] Use https://ci.appveyor.com/project/code-ready/crc/history to test bits on downstream
- [ ] Isolate e2e test execution (container running as monitools) #2533
- [ ] Sign develop distributables

PR #2533, opened 2021-07-06
[e2e] e2e container image

adrianriobo:
Part of #2461

Current qe environments include go to run e2e tests, this approach impacts:

* Include a go dependency on guest machines for CRC when this is not a requisite, this should not affect but best approach would be use minimal environments as guests.
* Extra effort on maintenance, as every upgrade on go version would require to upgrade on every environment.
* In case of regression this is really problematic as it would be possible to have a situation where an upgraded (go) guest requires a re run tests from previous version (with previous go version).

The solution would be create a container holding the build bits for e2e, the container will handle the logic to run the tests on a guest machine and take back the results. 

At the moment there is a limitation on e2e due to non go files which requires the container to pick some extra content from the project this will be overworked when CRC will upgrade go to [version 1.16, then the embed package](https://blog.golang.org/go1.16) could be use to embed all non go files.

### DIFF
--- a/images/build-e2e/Dockerfile
+++ b/images/build-e2e/Dockerfile
@@ -1,0 +1,29 @@
+
+FROM registry.access.redhat.com/ubi8/go-toolset:1.15.13 as builder
+
+USER root
+WORKDIR /workspace
+COPY . .
+RUN make build_e2e
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal 
+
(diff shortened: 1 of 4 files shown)


images/build-e2e/Dockerfile
```diff
@@ -0,0 +1,23 @@
+
+FROM registry.access.redhat.com/ubi8/go-toolset:1.14.12 as builder
```
guillaumerose, line 2: we are moving to go1.15 maybe this image also?

gbraad (approved):

PR #2546, opened 2021-07-07
WIP [e2e] add check for initial state shown on tray

### LINKED ISSUE #2545: [e2e] ux testing checks for stopped state as initial state on tray
Ux test definitions were reusing stop state check as the initial state of the cluster shown by the tray. Both states should have different checks (even if they re use the same state). 

A new method will be included on the tray interface to check explicitly the expected initial state.

adrianriobo:
Fixes #2545 

Added WIP until expected states are properly defined


adrianriobo:
Closed until can be implemented properly based on defined states

PR #2554, opened 2021-07-09
3 small changes

cfergeau:
3 commits I don't want to keep locally for too long.

The only user-visible changes are:
- slightly different error message for some daemon startup failures
- `capture.pcap` is no longer always created when the daemon is started with `--log-level debug`, but only when `CRC_DAEMON_PCAP_FILE` is set in the environment

### DIFF
--- a/cmd/crc/cmd/daemon.go
+++ b/cmd/crc/cmd/daemon.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -26,6 +25,7 @@ import (
 	"github.com/code-ready/gvisor-tap-vsock/pkg/virtualnetwork"
 	"github.com/docker/go-units"
 	"github.com/gorilla/handlers"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/exec"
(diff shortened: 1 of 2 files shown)


cmd/crc/cmd/daemon.go
```diff
@@ -59,7 +58,7 @@ var daemonCmd = &cobra.Command{
 
 		virtualNetworkConfig := types.Configuration{
 			Debug:             false, // never log packets
-			CaptureFile:       captureFile(),
+			CaptureFile:       os.Getenv("CRC_DAEMON_PCAP_FILE"),
```
praveenkumar, line 61: 👍🏼

praveenkumar (approved):

pkg/crc/constants/constants.go
```diff
@@ -128,20 +128,9 @@ func BinDir() string {
...
-			}
-		}
-		if userProfile := os.Getenv("USERPROFILE"); len(userProfile) > 0 {
-			if _, err := os.Stat(userProfile); err == nil {
-				return userProfile
-			}
-		}
-	}
-	return os.Getenv("HOME")
+	homeDir, _ := os.UserHomeDir()
```
guillaumerose, line 131: @praveenkumar @gbraad some thoughts here ? Maybe you have some history?

Should the code panic in case of an error ?

praveenkumar: @guillaumerose Without this patch we don't handle error here, but may be we should and panic, which can be part of another commit, since this commit is not changing the existing flow of code.

Issue #2556, opened 2021-07-09
[linux] crc daemon socket activation


cfergeau:
At the moment, `crc daemon` must be run manually before one can use its HTTP API, or when using usermode networking. Using systemd socket activation would let systemd listen for HTTP/vsock connection attempts, and tell it to automatically start the daemon when such a connection attempt is made.

The end result would be that it's possible to do
```
crc config set network-mode user
crc cleanup
crc setup
crc start
```
without having to run `crc daemon` explicitly

cfergeau:
There are still a few corner cases to handle, but this has landed several releases ago, and there are individual issues to track these remaining problems, this can be closed.

PR #2558, opened 2021-07-09
linux: Add systemd-socket activation

cfergeau:
At the moment, crc daemon must be run manually before one can use its HTTP API, or when using usermode networking. Using systemd socket activation would let systemd listen for HTTP/vsock connection attempts, and tell it to automatically start the daemon when such a connection attempt is made.

The end result is that after this PR, it's possible to do

```
crc config set network-mode user
crc cleanup
crc setup
crc start
```
without having to run crc daemon explicitly

After this PR, `systemctl --user status crc-http.socket crc-vsock.socket crc-daemon.service` knows about these 3 new units.
The socket units are listening/running, and the daemon is by  default not started. As soon as something interacts with the sockets in `~/.crc`, the daemon starts (can be tested with `nc -U ~/.crc/crc-http.sock`).

**Fixes:** Issue https://github.com/code-ready/crc/issues/2556

## Proposed changes

1. running `crc daemon` manually is no longer needed
2. if one wants to keep running `crc daemon` manually, the new preflight check for systemd socket units needs to be disabled


cfergeau:
CI is failing because the code in `preflight: Add symlink to crc binary in ~/.crc/bin` does not deal with the situation where `crc` is in the path and is started with just `crc`. https://golang.org/pkg/os/exec/#LookPath needs to be used somehow.

cfergeau:
The latest CI failure is more problematic (on `ci.centos.org`):
```
level=info msg="Setting up crc daemon systemd socket units"
Executing systemctl action failed:  exit status 1: Failed to connect to bus: No such file or directory
```

Not sure what to do about this one :-/

cfergeau:
Phew, made progress with CI. Looks like non-obvious steps are involved to use `systemctl --user` as a different user which is different from the one who logged in (ie use of `su` or `runuser`). I had to add:
```
# this creates /run/user/1000/ even if crc_ci is not logged in
$ loginctl enable-linger crc_ci
$ export XDG_RUNTIME_DIR="/run/user/$UID"
$ export DBUS_SESSION_BUS_ADDRESS="unix:path=${XDG_RUNTIME_DIR}/bus"
```
and only after this would `systemctl --user list-units` work without complaining about the bus not being found, or systemd not being found.

cfergeau:
centos CI is failing with 
```
13:39:10   Scenario: Bundle generation check # features/basic.feature:147
13:39:10     Given executing "crc bundle generate -f" succeeds # features/basic.feature:151
13:39:10       Error: command 'crc bundle generate -f', expected to succeed, exited with exit code: 1
13:39:10 Command stdout: 
13:39:10 Command stderr: level=info msg="Removing user's pull secret from instance disk and from cluster secret..."
13:39:10 Error removing pull secret from cluster: Temporary error: machine config is not rendered after removing pull secret (x22)
```
but I don't think that's related to this PR.

Makefile
```diff
@@ -179,11 +179,11 @@ ifndef BUNDLE_PATH
 export BUNDLE_PATH = $(HOME)/Downloads/crc_libvirt_$(BUNDLE_VERSION).$(BUNDLE_EXTENSION)
 endif
 integration:
-	@go test -timeout=60m $(REPOPATH)/test/integration -v
+	@go test -timeout=60m $(REPOPATH)/test/integration -v -ginkgo.noColor
```
praveenkumar, line 182: @cfergeau if you are talking about centos CI run then to get the colored output you always need to click full logs and you wlll have nice colored logs, this is really helped me to just go to error message quickly.

cfergeau: The other similar change is about CentOS CI, this one is about prow but is not working as expected.
They were mostly added as experiment/to help debug this PR, I can move them away for now. But default output of CentOS CI is not readable, clicking on "unformatted text" is also unreadable, one has to know there is a 3rd non-default link which will give readable output. Suboptimal :(

cmd/crc/cmd/daemon_linux.go
```diff
@@ -1,16 +1,88 @@
...
+	vsockUnitName = "crc-vsock.socket"
+	httpUnitName  = "crc-http.socket"
 )
 
+var systemdListeners map[string][]net.Listener
+
+func init() {
+	// listenerWithNames() cannot be called multiple times
+	systemdListeners, _ = listenersWithNames()
+}
```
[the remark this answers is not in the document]
cfergeau: > Also we are not consuming the error from listenersWithNames() so why we need to return it if it's value always going to be nil?

Since it's code c&p'ed from https://github.com/coreos/go-systemd/blob/f9f254610339c14ad7c50e9352848c7ff62fc81f/activation/listeners.go#L42 I'd prefer to avoid having too many differences between the 2 implementations

> instead of using init why don't we just use the var systemdListeners to assign return value of listenersWithNames() function?
Ah, thanks for the suggestion, I did not think/know this would compile just fine:
```
var systemdListeners, _ = listenersWithNames()
```

PR #2562, opened 2021-07-12
Introduce a new type for crc status. Distinguish it from machine status.

guillaumerose:
crc pkgs should not leak libmachine state enums. This commit converts it
to the crc status type when needed.

Starting/Stopping are proper to crc. It is not used in drivers/machine.
libmachine None, Paused and Saved states are not used anymore and will
be deleted.

---

It will help refactor statuses while not touching too much libmachine.

### DIFF
--- a/cmd/crc/cmd/console.go
+++ b/cmd/crc/cmd/console.go
@@ -8,8 +8,8 @@ import (
 
 	crcErrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/machine"
+	"github.com/code-ready/crc/pkg/crc/machine/state"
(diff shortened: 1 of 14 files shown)


guillaumerose:
/retest
retest this please

praveenkumar:
> No behavior should have changed with this PR. Where do you see that @praveenkumar ?

@guillaumerose No I am not seeing that, just curious if that is the plan?

guillaumerose:
Not my intention right now but it might be a good idea. 
Right after, I would like to delete somehow unused states in libmachine https://github.com/code-ready/machine/pull/63
Then, find a way to handle `Machine doesn't exist` state. Either add a new state? or reuse Stopped?

PR #2573, opened 2021-07-13
api: return state.Stopped when the machine doesnt exist

guillaumerose:
With this change, machine.Status() will always return a status.

crc cli still answers 'Machine does not exist. Use 'crc start' to create
 it'.
Only the http api will return Stopped status and not fail. This is
useful for showing a nice status in the tray.

### DIFF
--- a/pkg/crc/machine/status.go
+++ b/pkg/crc/machine/status.go
@@ -23,7 +23,10 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
 		return nil, errors.Wrap(err, "Cannot check if machine exists")
 	}
 	if !exists {
-		return nil, errors.New("Machine doesn't exist")
+		return &types.ClusterStatusResult{
+			CrcStatus:       state.Stopped,
+			OpenshiftStatus: types.OpenshiftStopped,
+		}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


gbraad:
In that case, can we show an actual 'meaningful' error when the Daemon is NOT reachable? because in the Windows tray this results in an `CommunicationException` that is for now ignored. Would help to handle this more graceful than concluding 'Stopped'.

anjannath:
> In that case, can we show an actual 'meaningful' error when the Daemon is NOT reachable? because in the Windows tray this results in an `CommunicationException` that is for now ignored. Would help to handle this more graceful than concluding 'Stopped'. WDYT?

i think we need to handle the `not reachable` case in the trays, this status would be received only when communication is happening but the machine doesn't exists, which is == 'Stopped'.

anjannath (approved):

guillaumerose:
> daemon not reachable

On macOS, it shows "Broken daemon?" in the menu status. The full exception is visible in the detailed status view.
The wording can be better for sure!

guillaumerose:
/retest
retest this please

guillaumerose:
/retest
retest this please

PR #2577, opened 2021-07-14
supersede: 2558 Add systemd-socket activation

praveenkumar:
At the moment, crc daemon must be run manually before one can use its HTTP API, or when using usermode networking. Using systemd socket activation would let systemd listen for HTTP/vsock connection attempts, and tell it to automatically start the daemon when such a connection attempt is made.

The end result is that after this PR, it's possible to do

```
crc config set network-mode user
crc cleanup
crc setup
crc start
```
without having to run crc daemon explicitly

After this PR, `systemctl --user status crc-http.socket crc-vsock.socket crc-daemon.service` knows about these 3 new units.
The socket units are listening/running, and the daemon is by  default not started. As soon as something interacts with the sockets in `~/.crc`, the daemon starts (can be tested with `nc -U ~/.crc/crc-http.sock`).

**Fixes:** Issue https://github.com/code-ready/crc/issues/2556

## Proposed changes

1. running `crc daemon` manually is no longer needed
2. if one wants to keep running `crc daemon` manually, the new preflight check for systemd socket units needs to be disabled


praveenkumar:
retest this please

cfergeau:
>  Add user network mode as filter for daemon socket enablement
> Daemon is only required if user try to run the crc in user network mode so better to add that filter for the preflight checks.

The daemon provides 2 things:
1) the 'vsock' API for VM networking when usermode networking is being used
2) a 'command' HTTP API which the trays use to do similar work to what can be achieved on the command line

1) is only required when using usermode networking. However, I can see 2) being used in the future even with kernel networking (assuming this is not removed). Arguably in this situation, we should not install `crc-vsock.socket`, but since the daemon is socket activated, it's not started when not needed in the kernel networking use case.

In short, I would always enable daemon socket activation, even for kernel networking.

cfergeau:
I've updated https://github.com/code-ready/crc/pull/2558 picking up the `removeCrcSymlink` fix.

praveenkumar:
@cfergeau closing in favor of #2558 

/close

PR #2580, opened 2021-07-14
Issue #2570 Use HTTP verbs for config api instead of separarte urls

### LINKED ISSUE #2570: [API] Refactor the config to use HTTP verbs
Currently the config endpoint uses `config/get`, `config/set` and `config/unset` to expose the functionality to deal with configuration. However, this seems to be a direct copy of the binary protocol and does not follow HTTP verbs as expected

anjannath:
GET /config returns the configs, body can be empty in which case all the
configs are retunred or body can specify which config options to retrieve, e.g: {"properties":["cpus"]}

POST /config sets a config options, body contains the config options to
set, e.g:  {"properties":{"cpus":5}}, {"properties":{"cpus":5,"autostart-tray": true}}

DELETE /config unsets a config option, body contains the config options
to unset, e.g: {"properties":["cpus"]}

Fixes #2570 

This requires changes on the trays.

### DIFF
--- a/pkg/crc/api/api_http.go
+++ b/pkg/crc/api/api_http.go
@@ -81,7 +81,7 @@ func NewMux(config crcConfig.Storage, machine machine.Client, logger Logger, tel
 		sendResponse(w, webconsoleInfo)
 	})
 
-	mux.Handle("/config/", http.StripPrefix("/config", newConfigMux(handler)))
+	mux.HandleFunc("/config", configHandler(handler))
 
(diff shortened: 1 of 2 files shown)


gbraad:
Perhaps better to base this on the PR from @guillaumerose; #2586

gbraad:
> This requires changes on the trays.

Can we create issues the follow this and perhaps create dependent PRs in advance?

anjannath:
Closed in favor of #2629

PR #2583, opened 2021-07-15
Handle error from os.UserHomeDir function

### LINKED ISSUE #2582: Panic if `os.UserHomeDir` return an error
Currently following code doesn't handle the error and this function is core for base dir for crc so in case there is an error better to panic so it's easy to debug.

```golang
// GetHomeDir returns the home directory for the current user
func GetHomeDir() string {
	homeDir, _ := os.UserHomeDir()

	return homeDir
}
```

praveenkumar:
We should panic in case there is an error since base dir is depend
on it.

fixes #2582

### DIFF
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -128,8 +128,10 @@ func BinDir() string {
 
 // GetHomeDir returns the home directory for the current user
 func GetHomeDir() string {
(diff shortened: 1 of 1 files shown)


guillaumerose (approved):
It's better than before :+1: 

Maybe later we could change the function to return (string, error) but it will be a larger change!

PR #2587, opened 2021-07-16
Misc 4.8 related changes which hopefully fix CI flakiness

praveenkumar:

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -390,16 +390,16 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 		return nil, errors.Wrap(err, "Error waiting for apiserver")
 	}
 
+	if err := cluster.EnsurePullSecretPresentInTheCluster(ocConfig, startConfig.PullSecret); err != nil {
+		return nil, errors.Wrap(err, "Failed to update cluster pull secret")
+	}
+
 	if err := cluster.EnsureSSHKeyPresentInTheCluster(ocConfig, constants.GetPublicKeyPath()); err != nil {
 		return nil, errors.Wrap(err, "Failed to update ssh public key to machine config")
 	}
 
 	if err := ensureProxyIsConfiguredInOpenShift(ocConfig, sshRunner, proxyConfig, instanceIP); err != nil {
 		return nil, errors.Wrap(err, "Failed to update cluster proxy configuration")
-	}
-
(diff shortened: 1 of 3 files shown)


guillaumerose:
/var/lib/kubelet/config.json file is not present in 4.7 bundle but is always {} in 4.8.
`test -e /var/lib/kubelet/config.json` always succeed and the pull secret is only set by the MCO in 4.8.

https://github.com/code-ready/crc/pull/2587/commits/826797f3df0d27253cbc48db831765f2e2d649aa this is dead code for 4.8 so looks good but we loose backward compatibility with 4.7.

guillaumerose:
https://github.com/code-ready/crc/pull/2587/commits/3d076828454b9b3c202a04dc1a8909613193728a looks good. I think it will solve the proxy failure. 

https://github.com/code-ready/crc/pull/2587/commits/a0b9c2a933fd1be1a96c54e6deb04e47df9b40d3 looks also good even for 4.7.

pkg/crc/machine/start.go
```diff
@@ -390,19 +394,20 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
...
 
-	if err := ensureProxyIsConfiguredInOpenShift(ocConfig, sshRunner, proxyConfig, instanceIP); err != nil {
-		return nil, errors.Wrap(err, "Failed to update cluster proxy configuration")
+	if err := cluster.EnsureSSHKeyPresentInTheCluster(ocConfig, constants.GetPublicKeyPath()); err != nil {
+		return nil, errors.Wrap(err, "Failed to update ssh public key to machine config")
 	}
 
-	if err := cluster.EnsurePullSecretPresentInTheCluster(ocConfig, startConfig.PullSecret); err != nil {
-		return nil, errors.Wrap(err, "Failed to update cluster pull secret")
+	if err := cluster.WaitForPullSecretPresentOnInstanceDisk(sshRunner); err != nil {
```
guillaumerose, line 405: I am wondering if we could move that just before `cluster.WaitForClusterStable`.
So that we save a bit of time: every config changes are sent, we are just waiting for them to be applied.

praveenkumar: @guillaumerose I can try but doesn't the proxy test are failing because of that, I had to put it above proxy configuration.

guillaumerose: ok np, let's start with this.

PR #2589, opened 2021-07-16
[e2e] windows ux

adrianriobo:
Fixes ##2535


test/e2e/crcsuite/ux/uxsuite.go
```diff
@@ -121,6 +129,10 @@ func guaranteeClusterState(state string) error {
 	return notificationHandler.ClearNotifications()
 }
 
+func clusterIsOnState(expectedState string) error {
```
gbraad, line 132: using `On` is probably not the right word here. preferable use `clusterHasState` or `clusterIsInState`, as `on` can mean activated.
Since you compare the cluster state with a given value, `clusterEqualsState()` might also work

test/e2e/features/ux.feature
```diff
@@ -1,23 +1,29 @@
...
         When install CRC from installer
         Then CRC is installed
-
-    @darwin
+		
+	@install @windows
+    Scenario: Install CRC 
+        Given an environment where CRC is not installed
+        When install CRC from installer
+        Then reboot is required
```
gbraad, line 18: I see what you did here...

How is the reboot and continuation actually handled?

test/extended/crc/ux/installer/installer_windows.go
```diff
@@ -4,25 +4,136 @@ package installer
...
+		logging.Error(err)
+		return err
+	}
+	<-g.Finished
+	ux.Finalize()
+	return nil
+}
+
+func runInstaller(installerPath string, finished chan bool) error {
+	// cmd := exec.Command("powershell.exe", "-c", "msiexec.exe", "/i", installerPath, "/qf")
```
gbraad, line 112: Is this needed?

test/extended/crc/ux/installer/installer_windows.go
```diff
@@ -4,25 +4,136 @@ package installer
...
+	// 	return fmt.Errorf("error starting %v with error %v", cmd, err)
+	// }
+	go openInstaller(installerPath, finished)
+	// delay to get window as active
+	time.Sleep(1 * time.Second)
+	return nil
+}
+
+func openInstaller(installerPath string, finished chan bool) {
+	// msiexecArguments := strings.Join(append([]string{"'"}, "/i", installerPath, "/qf", "'"), " ")
```
gbraad, line 123: Is this needed?

gbraad (commented):
LGTM, just a small issue with naming and some dead-code

test/e2e/features/ux.feature
adrianriobo: Basically I split the test suite based on annotations...so the ux on windows requires two suites: first @ux && @install ..this will run the installer finalizing with the reboot, second suite runs @ux && ~@install which run all other ux scenarios. 

The other option I had in mind but was not possible for the time being was running the suite from outside.. (so each step would be wrapped and executed remotely) so then you can control the reboot flow...

test/e2e/crcsuite/ux/uxsuite.go
adrianriobo: ack

test/extended/crc/ux/installer/installer_windows.go
adrianriobo: Ouch, totally forgot about it..definitely forgot cleanup code and  I will give a try without the subroutine as probably it is not required, but it was changed due to some issues I got to pick the elements which in the end was not related to this but with needed for an active X session. (Same applies to next comment)

test/extended/crc/ux/installer/installer_windows.go
adrianriobo: :+1:

test/e2e/crcsuite/ux/uxsuite.go
adrianriobo: Changed as per suggestion

test/extended/crc/ux/installer/installer_windows.go
adrianriobo: removed dead code + refactor removing the async part

test/extended/crc/ux/installer/installer_windows.go
adrianriobo: removed dead code + refactor removing the async part

gbraad (approved):

PR #2591, opened 2021-07-20
Bump trays to latest versions

guillaumerose:


gbraad (approved):

pkg/crc/version/version.go
```diff
@@ -31,9 +31,9 @@ var (
 const (
 	releaseInfoLink = "https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/crc/latest/release-info.json"
 	// Tray version to be embedded in executable
-	crcMacTrayVersion = "1.0.10"
```
gbraad, line 34: Can we start using something like:

`1.11` instead ...

pkg/crc/version/version.go
```diff
@@ -31,9 +31,9 @@ var (
...
+	crcMacTrayVersion = "1.0.11"
 	// Windows forms application version type major.minor.buildnumber.revesion
-	crcWindowsTrayVersion = "0.8.0.0"
+	crcWindowsTrayVersion = "0.9.0.0"
```
gbraad, line 36: ... and also bump the Windows tray to beyond `1.0` ?

gbraad: @anjannath What do you think of bumping the release beyond 1.0. I think we have a decent GA, right?

gbraad:
Very often you will see:

```
dial tcp 192.168.130.11:22: connect: connection refused"
```

VM does not start properly?

gbraad:
Merging. as the instability is caused by something else.

PR #2592, opened 2021-07-20
Faster crc stop

guillaumerose:
Let see how it goes on the CI.

### DIFF
--- a/pkg/crc/machine/stop.go
+++ b/pkg/crc/machine/stop.go
@@ -4,8 +4,8 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/machine/state"
-	"github.com/code-ready/crc/pkg/crc/oc"
 	crcssh "github.com/code-ready/crc/pkg/crc/ssh"
+	"github.com/code-ready/crc/pkg/crc/systemd"
 	"github.com/code-ready/crc/pkg/libmachine/host"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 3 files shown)


pkg/crc/machine/stop.go
```diff
@@ -52,10 +52,9 @@ func removeMCOPods(host *host.Host, client *client) error {
...
 
-	ocConfig := oc.UseOCWithSSH(sshRunner)
-	_, stderr, err := ocConfig.RunOcCommand("delete", "pods", "--all", "-n openshift-machine-config-operator", "--grace-period=0")
-	if err != nil {
-		logging.Debugf("Error deleting the pods from openshift-machine-config-operator namespace:%v \n StdErr: %s", err, stderr)
+	if err := systemd.NewInstanceSystemdCommander(sshRunner).Stop("kubelet"); err != nil {
+		return err
 	}
-	return nil
+	_, _, err = sshRunner.RunPrivileged("", `-- sh -c 'crictl stop $(crictl ps -q)'`)
```
praveenkumar, line 58: @guillaumerose I would like to capture the `stderr` here and return it along with `err` because sometime error message is very vague and doesn't give any context so better to add stderr also.

pkg/crc/machine/start.go
```diff
@@ -469,7 +469,7 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 
 	logging.Info("Adding crc-admin and crc-developer contexts to kubeconfig...")
 	if err := writeKubeconfig(instanceIP, clusterConfig); err != nil {
-		logging.Errorf("Cannot update kubeconfig: %v", err)
+		return nil, err
```
praveenkumar, line 472: @guillaumerose Doesn't this change will impact the ux side of message because till now these are `Error` messages but at the end user have the message that cluster is unstable but still print the accessible point which means cluster might reconcile and works properly, With this change user might have to delete and start.

```

Log in as administrator:
  Username: kubeadmin
  Password: gpxkx-4kox9-4dJvw-c4CBs

Log in as user:
  Username: developer
  Password: developer

  $ eval $(crc oc-env)
  $ oc login -u developer https://api.crc.testing:6443
```

PR #2601, opened 2021-07-23
Improve crc stop message for already stopped cluster

anjannath:
It'll check if machine is already stopped before attempting
to perform the cluster stop related tasks

If a machine is already stopped it prints the following:

```
$ crc stop
Cluster is already stopped
```
Earlier this was:
```
$ crc stop
INFO Stopping kubelet and all containers...
Error getting the IP: Host is not running
```

### DIFF
--- a/pkg/crc/machine/stop.go
+++ b/pkg/crc/machine/stop.go
@@ -11,6 +11,10 @@ import (
 )
 
 func (client *client) Stop() (state.State, error) {
+	if running, _ := client.IsRunning(); !running {
+		return state.Error, errors.New("Cluster is already stopped")
+	}
+
 	libMachineAPIClient, cleanup := createLibMachineClient()
 	defer cleanup()
 	host, err := libMachineAPIClient.Load(client.name)


cmd/crc/cmd/stop.go
```diff
@@ -32,6 +32,10 @@ func stopMachine(client machine.Client, interactive, force bool) (bool, error) {
 		return false, err
 	}
 
+	if running, _ := client.IsRunning(); !running {
+		return false, fmt.Errorf("Cluster is already stopped")
```
gbraad, line 36: what does the HTTP API reply in this case:
Will it also error or success with this message?

Because, I do not believe this should be an error, but rather info?

gbraad (dismissed):

cmd/crc/cmd/stop.go
praveenkumar: Agree, if you are running stop on an stopped cluster then it just provide info and not error. But then same with other API which query when crc vm is not exist, like ip, status .. etc. `checkIfMachineMissing` always return error.

anjannath: > what does the HTTP API reply in this case:
Will it also error or success with this message?

Error, HTTP api will respond with a 500 internal server error and no body

gbraad: I just wonder how a stopping a stopped VM can result in an error; does starting a started VM also do this? what about deleting a deleted VM?

praveenkumar: As of now `starting` an already running VM just print out all the details about openshift user messages. if we `delete` a deleted machine it through an `machine doesn't exist error`

praveenkumar: Looking at the implementation of `IsRunning`, the state of the VM is not running but it can be anything like following. So we shouldn't say it is stopped but may be `not running` ?

```
	"Paused",
	"Saved",
	"Stopped",
	"Stopping",
	"Starting",
	"Error",
	"Timeout",
```

cfergeau: > Error, HTTP api will respond with a 500 internal server error and no body.

If you put a similar check in `pkg/crc/machine/stop.go` rather than in `cmd/`,  this is going to  solve this both for HTTP and command line.

anjannath: @gbraad @cfergeau Done, moved the check to `pkg/crc/machine/stop.go`

cfergeau (approved):
Looks good to me, dunno if this should be against the `master` or the `dev` branch.

anjannath:
> Looks good to me, dunno if this should be against the `master` or the `dev` branch.

merge on `master` then cherry-pick to `dev` ? /cc @praveenkumar

gbraad:
@praveenkumar @anjannath what is the status?

PR #2608, opened 2021-07-26
mco: delete the leader lease right after start.

guillaumerose:
It allows the machine-config-controller to start fast and not wait 2min
before acquiring the lease and change the pull secret.

---

It takes 20s to change the pull secret on the disk compared to 2min+.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -394,6 +394,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 		return nil, errors.Wrap(err, "Error waiting for apiserver")
 	}
 
+	if err := cluster.DeleteMCOLeaderLease(ocConfig); err != nil {
+		return nil, err
+	}
+
 	if err := cluster.EnsurePullSecretPresentInTheCluster(ocConfig, startConfig.PullSecret); err != nil {
 		return nil, errors.Wrap(err, "Failed to update cluster pull secret")
 	}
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/start.go
```diff
@@ -482,6 +486,11 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 	}, nil
 }
 
+func deleteMCOLeaderLease(ocConfig oc.Config) error {
+	_, _, err := ocConfig.RunOcCommand("delete", "-n", "openshift-machine-config-operator", "cm", "machine-config-controller")
+	return err
+}
+
```
praveenkumar, line 493: @guillaumerose does it make sense to have it part of snc repo?

guillaumerose: yes but I think we would need to stop machine-config-controller container before doing this. 
And we still need to have this somewhere for the generated bundle. 

As you prefer.

praveenkumar: @guillaumerose yes it will need for generated bundle also.

praveenkumar: @guillaumerose can we make it part of cluster package so we can also add the `waitForResource` functionality because most of time in the CI we observed even after apiserver responding correctly, it again go back to reconcile mode and these kind of command does fails.

guillaumerose: yes :+1:

PR #2611, opened 2021-07-27
[e2e] add cleanup-home option to control remove crc home before suite

adrianriobo:
This PR adds a new option when running e2e tests to control if remove the crc home folder or not before executing the tests. The default behavior is like it was before it deletes the crc home folder (if no parameter passed)

If `--cleanup-home=false` is passed when running e2e tests it will keep the crc home folder, allowing to setup the environment for tests externally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,7 +184,7 @@ ifndef CRC_BINARY
 	CRC_BINARY = --crc-binary=$(GOPATH)/bin
 endif
 e2e:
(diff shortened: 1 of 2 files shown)


praveenkumar:
@adrianriobo With this PR you want to have a way to do the setup externally and the run the test. Can you mention this in the commit logs also so it become easier to understand why we require this flag.

adrianriobo:
> @adrianriobo With this PR you want to have a way to do the setup externally and the run the test. Can you mention this in the commit logs also so it become easier to understand why we require this flag.

Changed to:

`e2e add cleanup-home flag. This will alllow to setup the environment externally before run the tests`

PR #2616, opened 2021-07-29
Hyper-V driver improvements

guillaumerose:
After looking at telemetry, I saw couple of annoying failures:
- crc status returning `unexpected Hyper-V state []`
- crc delete dead-end when the VM is deleted in Hyper-V but not in ~/.crc/machines.

For delete error, I propose to change all drivers to return no error on Remove() when the VM doesn't exist. 
Hyperkit already does this (see https://github.com/code-ready/machine-driver-hyperkit/blob/master/pkg/hyperkit/driver.go#L143)

### DIFF
--- a/pkg/drivers/hyperv/hyperv_windows.go
+++ b/pkg/drivers/hyperv/hyperv_windows.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	log "github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/os/windows/powershell"
 	"github.com/code-ready/machine/libmachine/drivers"
 	"github.com/code-ready/machine/libmachine/state"
 )
@@ -303,6 +304,10 @@ func (d *Driver) Stop() error {
 
 // Remove removes an host
 func (d *Driver) Remove() error {
+	if _, _, err := powershell.Execute(`Hyper-V\Get-VM`, d.MachineName, "-ErrorAction", "SilentlyContinue", "-ErrorVariable", "getVmErrors"); err != nil {
+		return nil
+	}
+
 	s, err := d.GetState()
 	if err != nil {
 		return err


pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -304,6 +305,11 @@ func (d *Driver) Stop() error {
 
 // Remove removes an host
 func (d *Driver) Remove() error {
+	if _, stderr, err := powershell.Execute(`Hyper-V\Get-VM`, d.MachineName); err != nil &&
+		strings.Contains(stderr, "Hyper-V was unable to find a virtual machine") {
+		return nil
+	}
+
```
praveenkumar, line 312: @guillaumerose shouldn't this also for other driver like libvirt/hyperkit when vm is deleted from virtual manager and then try to delete it?
```
$ sudo virsh destroy crc
Domain 'crc' destroyed

$ sudo virsh  list --all
 Id   Name   State
-----------------------
 -    crc    shut off

$ sudo virsh undefine crc
Domain 'crc' has been undefined

$ sudo virsh  list --all
 Id   Name   State
--------------------

$ ./crc delete
Do you want to delete the OpenShift cluster? [y/N]: y
Driver cannot remove machine: Failed to fetch machine 'crc'
```

guillaumerose: Yes libvirt would be the missing piece. Here the line to change: https://github.com/code-ready/machine-driver-libvirt/blob/master/pkg/libvirt/libvirt.go#L437

If it's ok to do this behaviour change, I will do it 👍

praveenkumar: It should be good be consistent across all the platform.

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -303,6 +305,11 @@ func (d *Driver) Stop() error {
 
 // Remove removes an host
 func (d *Driver) Remove() error {
+	if _, stderr, err := powershell.Execute(`Hyper-V\Get-VM`, d.MachineName); err != nil &&
+		strings.Contains(stderr, "Hyper-V was unable to find a virtual machine") {
```
gbraad, line 309: are calls forced to the English locale?

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -304,6 +305,11 @@ func (d *Driver) Stop() error {
...
```
cfergeau, line 309: Can this be localized? Not clear from a quick look at  the powershell code if the locale is always going to be an English one.

pkg/drivers/hyperv/hyperv_windows.go
cfergeau: Fwiw, I'd have a slight preference for having a "VMNotFound" error which we can special case, but noop in this case will also do :)

pkg/drivers/hyperv/hyperv_windows.go
gbraad:
```
get-vm -name crc -ErrorAction SilentlyContinue 
```

would error when the VM does not exist, and else provide the info... in that case the return code can be checked for

cfergeau (commented):
>Hyperkit already does this (see https://github.com/code-ready/machine-driver-hyperkit/blob/master/pkg/hyperkit/driver.go#L143)

My reading of the code is that this line will get triggered when there is an error finding the hyperkit process. When the VM does not exist, I'd expect `findHyperkitProcess` to return `nil, nil` and `GetState` to return `stopped, nil`, and this `log.Debugf` won't get triggered (but I did not run this to make sure this is the case :)

Apart from the localisation concern, looks good to me.

pkg/drivers/hyperv/hyperv_windows.go
guillaumerose: The custom error would force us to add it in machine repo so that we can use it everywhere. A little bit painful :/

pkg/drivers/hyperv/hyperv_windows.go
guillaumerose: yes not really good. What if I dont check the error message?
If the driver can't get the VM, then Remove returns nil directly.

pkg/drivers/hyperv/hyperv_windows.go
guillaumerose: > are calls forced to the English locale?

No

> -ErrorAction SilentlyContinue 

I don't understand how it works. What do you suggest exactly?

pkg/drivers/hyperv/hyperv_windows.go
cfergeau: > The custom error would force us to add it in machine repo so that we can use it everywhere. A little bit painful :/

No worries, as I said, no strong objection to just returning `nil` in this situation :)

pkg/drivers/hyperv/hyperv_windows.go
gbraad: if not using an English locale, checking "Hyper-V was unable to find a virtual machine" might not give expected results. the a;ternative is to check only the return code with slientlycontinue alone. but might not be a necessity.

gbraad:
```
get-vm cr -ErrorAction SilentlyContinue -ErrorVariable getVmErrors
$?
```

would return `False`

```
get-vm crc -ErrorAction SilentlyContinue -ErrorVariable getVmErrors
$?
```

would return `True`

guillaumerose: Thanks will update it!

pkg/drivers/hyperv/hyperv_windows.go
guillaumerose: done!

pkg/drivers/hyperv/hyperv_windows.go
guillaumerose: done and it works perfectly. Thanks for this magic trick!

guillaumerose:
I opened the libvirt side for the Remove() change. https://github.com/code-ready/machine-driver-libvirt/pull/92

retest this please

guillaumerose:
can someone approve otherwise it can't be merged given the new rules. 
Thanks

gbraad:
> given the new rules.

???

gbraad (approved):

PR #2629, opened 2021-08-05
Use HTTP verbs for config api instead of separarte urls

### LINKED ISSUE #2570: [API] Refactor the config to use HTTP verbs

praveenkumar:
Fixes #2570

```
// Get all the configs since empty property passed
$ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/config

// Get cpus property
$ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/config?cpus
{"Success":true,"Error":"","Configs":{"cpus":5}}

// Delete cpu property
$ curl -X DELETE -d '{"properties":["cpus"]}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Success":true,"Error":"","Properties":["cpus"]}

$ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/config?cpus
{"Success":true,"Error":"","Configs":{"cpus":4}}

// Update cpu propery
$ curl -X POST -d '{"properties":{"cpus":10}}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Success":true,"Error":"","Properties":["cpus"]}

$ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/config?cpus
{"Success":true,"Error":"","Configs":{"cpus":10}}
```

### DIFF
--- a/pkg/crc/api/api_http_test.go
+++ b/pkg/crc/api/api_http_test.go
@@ -126,6 +126,88 @@ func TestHTTPApi(t *testing.T) {
 		},
 		configSetResult,
 	)
+	// Get the cpus config again after setting it to make sure it is set
+	// properly.
+	configGetAfterSetResult, err := client.GetConfig([]string{"cpus"})
+	assert.NoError(t, err)
+	assert.Equal(
+		t,
+		apiClient.GetConfigResult{
+			Success: true,
+			Error:   "",
+			Configs: map[string]interface{}{
+				"cpus": float64(5),
+			},
+		},
+		configGetAfterSetResult,
+	)
+
+	configUnsetResult, err := client.UnsetConfig([]string{"cpus"})
+	assert.NoError(t, err)
+	assert.Equal(
+		t,
+		apiClient.SetOrUnsetConfigResult{
+			Success:    true,
+			Error:      "",
+			Properties: []string{"cpus"},
+		},
+		configUnsetResult,
+	)
+
+	allConfigGetResult, err := client.GetConfig(nil)
+	assert.NoError(t, err)
+	configs := make(map[string]interface{})
+	for k, v := range config.AllConfigs() {
+		// This is required because of https://pkg.go.dev/encoding/json#Unmarshal
+		// Unmarshal stores float64 for JSON numbers in case of interface.
+		switch v := v.Value.(type) {
+		case int:
+			configs[k] = float64(v)
+		default:
+			configs[k] = v
+		}
+	}
+	assert.Equal(
+		t,
+		apiClient.GetConfigResult{
+			Success: true,
+			Error:   "",
+			Configs: configs,
+		},
+		allConfigGetResult,
+	)
+
+	// Get result of making query for multiple property
+	configGetMultiplePropertyResult, err := client.GetConfig([]string{"cpus", "memory"})
+	assert.NoError(t, err)
+	assert.Equal(
+		t,
+		apiClient.GetConfigResult{
+			Success: true,
+			Error:   "",
+			Configs: map[string]interface{}{
+				"cpus":   float64(4),
+				"memory": float64(9216),
+			},
+		},
+		configGetMultiplePropertyResult,
+	)
+
+	// Get result of special query for multiple properties
+	configGetSpecialPropertyResult, err := client.GetConfig([]string{"a&a", "b&&&b"})
+	assert.NoError(t, err)
+	assert.Equal(
+		t,
+		apiClient.GetConfigResult{
+			Success: true,
+			Error:   "",
+			Configs: map[string]interface{}{
+				"a&a":   "foo",
+				"b&&&b": "bar",
+			},
+		},
+		configGetSpecialPropertyResult,
+	)
 }
 
 func TestTelemetry(t *testing.T) {
--- a/pkg/crc/api/client/client.go
+++ b/pkg/crc/api/client/client.go
@@ -8,6 +8,8 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"strings"
 )
 
 type Client struct {
@@ -109,17 +111,12 @@ func (c *Client) WebconsoleURL() (ConsoleResult, error) {
 
 func (c *Client) GetConfig(configs []string) (GetConfigResult, error) {
 	var gcr = GetConfigResult{}
-	var data = new(bytes.Buffer)
-
-	if len(configs) > 0 {
-		cfg := GetOrUnsetConfigRequest{
-			Properties: configs,
-		}
-		if err := json.NewEncoder(data).Encode(cfg); err != nil {
-			return gcr, fmt.Errorf("Failed to encode data to JSON: %w", err)
-		}
+	var escapeConfigs []string
+	for _, v := range configs {
+		escapeConfigs = append(escapeConfigs, url.QueryEscape(v))
 	}
-	body, err := c.sendPostRequest("/config/get", data)
+	queryString := strings.Join(escapeConfigs, "&")
+	body, err := c.sendGetRequest(fmt.Sprintf("/config?%s", queryString))
 	if err != nil {
 		return gcr, err
 	}
@@ -142,7 +139,7 @@ func (c *Client) SetConfig(configs SetConfigRequest) (SetOrUnsetConfigResult, er
 		return scr, fmt.Errorf("Failed to encode data to JSON: %w", err)
 	}
 
-	body, err := c.sendPostRequest("/config/set", data)
+	body, err := c.sendPostRequest("/config", data)
 	if err != nil {
 		return scr, err
 	}
@@ -164,7 +161,7 @@ func (c *Client) UnsetConfig(configs []string) (SetOrUnsetConfigResult, error) {
 	if err := json.NewEncoder(data).Encode(cfg); err != nil {
 		return ucr, fmt.Errorf("Failed to encode data to JSON: %w", err)
 	}
-	body, err := c.sendPostRequest("/config/unset", data)
+	body, err := c.sendDeleteRequest("/config", data)
 	if err != nil {
 		return ucr, err
 	}
@@ -221,11 +218,9 @@ func (c *Client) sendGetRequest(url string) ([]byte, error) {
 		return nil, err
 	}
 	defer res.Body.Close()
-
 	if res.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("Error occurred sending GET request to : %s : %d", url, res.StatusCode)
 	}
-
 	body, err := ioutil.ReadAll(res.Body)
 	if err != nil {
 		return nil, fmt.Errorf("Unknown error reading response: %w", err)
@@ -234,7 +229,15 @@ func (c *Client) sendGetRequest(url string) ([]byte, error) {
 }
 
 func (c *Client) sendPostRequest(url string, data io.Reader) ([]byte, error) {
-	req, err := http.NewRequest("POST", fmt.Sprintf("%s%s", c.base, url), data)
+	return c.sendRequest(url, http.MethodPost, data)
+}
+
+func (c *Client) sendDeleteRequest(url string, data io.Reader) ([]byte, error) {
+	return c.sendRequest(url, http.MethodDelete, data)
+}
+
+func (c *Client) sendRequest(url string, method string, data io.Reader) ([]byte, error) {
+	req, err := http.NewRequest(method, fmt.Sprintf("%s%s", c.base, url), data)
 	if err != nil {
 		return nil, err
 	}
@@ -246,8 +249,15 @@ func (c *Client) sendPostRequest(url string, data io.Reader) ([]byte, error) {
 	}
 	defer res.Body.Close()
 
-	if res.StatusCode != http.StatusOK && res.StatusCode != http.StatusCreated {
-		return nil, fmt.Errorf("Error occurred sending POST request to : %s : %d", url, res.StatusCode)
+	switch method {
+	case http.MethodPost:
+		if res.StatusCode != http.StatusOK && res.StatusCode != http.StatusCreated {
+			return nil, fmt.Errorf("Error occurred sending POST request to : %s : %d", url, res.StatusCode)
+		}
+	case http.MethodDelete, http.MethodGet:
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("Error occurred sending %s request to : %s : %d", method, url, res.StatusCode)
+		}
 	}
 
 	body, err := ioutil.ReadAll(res.Body)
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -214,9 +214,10 @@ func (h *Handler) UnsetConfig(c *context) error {
 }
 
 func (h *Handler) GetConfig(c *context) error {
+	queries := c.url.Query()
 	var req client.GetOrUnsetConfigRequest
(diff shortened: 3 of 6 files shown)


cfergeau:
I think we also want to get rid of the `Success`/`Error` in the json answer, and use http status codes instead? I guess  behaviour is unchanged when you specify multiple properties?

cfergeau:
api_http_test.go also needs to be updated

gbraad:
I think the error status code can be a follow-up

anjannath:
I cannot get all the configs from the api, when there is no data provided it should return the complete config, the trays depend on this behavior of the api to populate the settings/preferences windows

```
$ curl -vv --unix-socket ~/.crc/crc-http.sock http:/unix/api/config
*   Trying /Users/anjan/.crc/crc-http.sock...
* Connected to unix (/Users/anjan/.crc/crc-http.sock) port 80 (#0)
> GET /api/config HTTP/1.1
> Host: unix
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Fri, 06 Aug 2021 06:49:02 GMT
< Content-Length: 29
<
unexpected end of JSON input
* Connection #0 to host unix left intact
* Closing connection 0
```

pkg/crc/api/handlers.go
```diff
@@ -215,8 +215,10 @@ func (h *Handler) UnsetConfig(c *context) error {
 
 func (h *Handler) GetConfig(c *context) error {
 	var req client.GetOrUnsetConfigRequest
-	if err := c.Bind(&req); err != nil {
-		return err
+	if len(c.requestBody) != 0 {
+		if err := c.Bind(&req); err != nil {
```
cfergeau, line 219: See the conversation in https://github.com/code-ready/crc/pull/2631

pkg/crc/api/client/client.go
```diff
@@ -164,7 +164,7 @@ func (c *Client) UnsetConfig(configs []string) (SetOrUnsetConfigResult, error) {
 	if err := json.NewEncoder(data).Encode(cfg); err != nil {
 		return ucr, fmt.Errorf("Failed to encode data to JSON: %w", err)
 	}
-	body, err := c.sendPostRequest("/config/unset", data)
+	body, err := c.sendPostRequest("/config", data)
```
cfergeau, line 167: `c.sendDeleteRequest()`?

pkg/crc/api/client/client.go
```diff
@@ -215,14 +215,20 @@ func (c *Client) SetPullSecret(data string) error {
...
+	req.Header.Set("Content-Type", "application/json")
+
+	res, err := c.client.Do(req)
 	if err != nil {
 		return nil, err
 	}
 	defer res.Body.Close()
 
-	if res.StatusCode != http.StatusOK {
+	if res.StatusCode != http.StatusOK && res.StatusCode != http.StatusCreated {
```
cfergeau, line 231: `http.StatusCreated` is unexpected on a `get` request

pkg/crc/api/client/client.go
```diff
@@ -215,14 +215,20 @@ func (c *Client) SetPullSecret(data string) error {
...
-	res, err := c.client.Get(fmt.Sprintf("%s%s", c.base, url))
+func (c *Client) sendGetRequest(url string, data io.Reader) ([]byte, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s%s", c.base, url), data)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	res, err := c.client.Do(req)
 	if err != nil {
```
cfergeau, line 226: Isn't this now the same method as `sendPostRequest`?

pkg/crc/api/handlers.go
praveenkumar: In that case it was working before, if we pass `-d "{}"` in curl command we are getting all the configs

pkg/crc/api/client/client.go
praveenkumar: @cfergeau yes, Just the method is different so we can create an helper with same functionality and pass the which method is required `GET/POST ..etc`

pkg/crc/api/client/client.go
```diff
@@ -37,7 +37,7 @@ func (c *Client) Version() (VersionResult, error) {
 
 func (c *Client) Status() (ClusterStatusResult, error) {
 	var sr = ClusterStatusResult{}
-	body, err := c.sendGetRequest("/status")
+	body, err := c.sendGetRequest("/status", nil)
```
gbraad, line 40: do we really need a `sendGetRequest(path, nil)`,

why not an override `sendGetRequest(path)` method that internally calls `sendGetRequest(path, nil)`

pkg/crc/api/client/client.go
gbraad: `Created` mostly comes from a `PUT`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

pkg/crc/api/client/client.go
praveenkumar: @gbraad No, that would be too much override, I am already adding a helper function for all type of request, adding another helper on top of Get doesn't make sense, I would be more explicit here for testing purpose also.

pkg/crc/api/client/client.go
praveenkumar: Yes, I am creating a helper method now.

pkg/crc/api/api_http_test.go
```diff
@@ -126,6 +126,21 @@ func TestHTTPApi(t *testing.T) {
 		},
 		configSetResult,
 	)
+	// Get the cpus config again after setting it to make sure it is set
+	// properly.
+	configGetAfterSetResult, err := client.GetConfig([]string{"cpus"})
```
gbraad, line 131: this is an API call that could make sense, but as mentioned in the PR description with a body message this makes not sense. This can only be coming as a querystring.

gbraad:
```
// Get cpus property
$ curl -X GET -d '{"properties":["cpus"]}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Success":true,"Error":"","Configs":{"cpus":5}}

$ curl -X GET -d '{"properties":["cpus"]}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Success":true,"Error":"","Configs":{"cpus":4}}

$ curl -X GET -d '{"properties":["cpus"]}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Success":true,"Error":"","Configs":{"cpus":10}}
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET

      Request has body        |       No

cfergeau:
This PR should fix https://github.com/code-ready/crc/issues/2570

pkg/crc/api/client/client.go
gbraad: > No, that would be too much override,

a default parameter? Not sure what you mean with too much override. it would also be clear, or could be naming... anyway, a fix is 'another helper' is what I was looking for.

pkg/crc/api/api_http_test.go
```diff
@@ -126,6 +126,33 @@ func TestHTTPApi(t *testing.T) {
...
```
anjannath, line 131: Can we also add a test case for the case of getting all the configs so that we don't have regressions in the future. This was also asked in https://github.com/code-ready/crc/pull/2631#issuecomment-894075772

praveenkumar: @anjannath Added 👍🏼

pkg/crc/api/handlers.go
```diff
@@ -214,9 +214,12 @@ func (h *Handler) UnsetConfig(c *context) error {
 }
 
 func (h *Handler) GetConfig(c *context) error {
+	queries := c.url.Query()
 	var req client.GetOrUnsetConfigRequest
-	if err := c.Bind(&req); err != nil {
-		return err
+	if len(queries) != 0 {
```
gbraad, line 219: simple huh ;-)

👍

gbraad (approved):

pkg/crc/api/handlers.go
```diff
@@ -214,9 +214,12 @@ func (h *Handler) UnsetConfig(c *context) error {
...
+		for key := range queries {
+			req.Properties = append(req.Properties, key)
+		}
 	}
```
cfergeau, line 223: You have a check for `req.Properties` length just below this, I don't think the `len(queries)` check is needed, if `queries` is empty, `for key := range queries` will loop 0 times, and `len(req.Properties)` will be 0

pkg/crc/api/handlers.go
```diff
@@ -214,9 +214,12 @@ func (h *Handler) UnsetConfig(c *context) error {
 }
 
 func (h *Handler) GetConfig(c *context) error {
+	queries := c.url.Query()
 	var req client.GetOrUnsetConfigRequest
-	if err := c.Bind(&req); err != nil {
-		return err
+	if len(queries) != 0 {
+		for key := range queries {
+			req.Properties = append(req.Properties, key)
```
cfergeau, line 221: I don't know if golang calls https://pkg.go.dev/net/url#QueryUnescape automatically?

pkg/crc/api/api_http_test.go
```diff
@@ -126,6 +126,41 @@ func TestHTTPApi(t *testing.T) {
...
+			Success: true,
+			Error:   "",
+			Configs: map[string]interface{}{
+				"cpus": float64(5),
+			},
+		},
+		configGetAfterSetResult,
+	)
+
+	configUnsetResult, err := client.UnsetConfig([]string{"cpus"})
```
cfergeau, line 145: Can there also be a test for getting multiple config values?

praveenkumar: @cfergeau we can add more test.

pkg/crc/api/client/client.go
```diff
@@ -109,17 +110,8 @@ func (c *Client) WebconsoleURL() (ConsoleResult, error) {
...
-	if len(configs) > 0 {
-		cfg := GetOrUnsetConfigRequest{
-			Properties: configs,
-		}
-		if err := json.NewEncoder(data).Encode(cfg); err != nil {
-			return gcr, fmt.Errorf("Failed to encode data to JSON: %w", err)
-		}
-	}
-	body, err := c.sendPostRequest("/config/get", data)
+	queryString := strings.Join(configs, "&")
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau didn't get you why I need to escape individual config, here we are building the query string using the `&` as separator? This should be parsed by handler like https://pkg.go.dev/net/url#URL.Query way (example)

pkg/crc/api/handlers.go
praveenkumar: @cfergeau Here we are using `url.Query` property which should handle it https://pkg.go.dev/net/url#URL.Query or you have any other specific suggestion?

cfergeau: This was the question, whether it does it automatically or not. Seems to be fine https://cs.opensource.google/go/go/+/refs/tags/go1.16.7:src/net/url/url.go;l=927-959

pkg/crc/api/client/client.go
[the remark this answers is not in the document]
cfergeau: If one calls `client.GetConfig([]string{ "a&a", "b&b", "c&c"})`, the current code generates `queryString = "a&a&b&b&c&c"`. After the call to `strings.Join`, we've lost the individual query item, so there's nothing golang URL module can do about it. If you call `url.QueryEscape` beforehand, you can generate the expected `queryString = "a%26a&b%26b&c%26c"`, which `URL.Query` will then be able to handle properly.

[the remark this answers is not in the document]
cfergeau: (this is all very theoretical as our config options don't have weird characters in them, but since we don't control the inputs, and it's not very hard to do the right thing, I think it's worthwhile to do the right thing)

pkg/crc/api/client/client.go
```diff
@@ -246,8 +249,19 @@ func (c *Client) sendPostRequest(url string, data io.Reader) ([]byte, error) {
...
+			return nil, fmt.Errorf("Error occurred sending POST request to : %s : %d", url, res.StatusCode)
+		}
+	case http.MethodDelete:
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("Error occurred sending Delete request to : %s : %d", url, res.StatusCode)
+		}
+	case http.MethodGet:
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("Error occurred sending GET request to : %s : %d", url, res.StatusCode)
+		}
```
cfergeau, line 264: You can probably group the last 2 cases:
```
case http.MethodDelete, http.MethodGet:
    if res.StatusCode != http.StatusOK {
 			return nil, fmt.Errorf("Error occurred sending %s request to : %s : %d", method, url, res.StatusCode)
    }
```

cfergeau (approved):
There's a `separate` typo in the commit log. A few minor comments, but looks good!

pkg/crc/api/api_http_test.go
```diff
@@ -126,6 +126,41 @@ func TestHTTPApi(t *testing.T) {
...
+		},
+		configUnsetResult,
+	)
+
+	allConfigGetResult, err := client.GetConfig(nil)
+	assert.NoError(t, err)
+	assert.Contains(
+		t,
+		allConfigGetResult.Configs,
+		"kubeadmin-password",
```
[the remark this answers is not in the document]
gbraad: is that really 'testing' in that case? or you mean a 'generation' step that is performed once in a while to fix something in place? still, feels like testing the API with the API

pkg/crc/api/client/client.go
```diff
@@ -246,8 +244,22 @@ func (c *Client) sendPostRequest(url string, data io.Reader) ([]byte, error) {
...
+	if method == http.MethodGet {
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("Error occurred sending GET request to : %s : %d", url, res.StatusCode)
+		}
+	}
+
+	if method == http.MethodDelete {
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("Error occurred sending Delete request to : %s : %d", url, res.StatusCode)
+		}
```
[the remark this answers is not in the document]
gbraad: that would not be more readable:

```golang
switch method {
        case http.MethodDelete:
		if res.StatusCode != http.StatusOK {
			return nil, fmt.Errorf("Error occurred sending Delete request to : %s : %d", url, res.StatusCode)
		}
		break
        case ...
```

pkg/crc/api/api_http_test.go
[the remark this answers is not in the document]
cfergeau: The test case creates a 'config' object which is initialized with a given set of settings. Then the test creates a fake http server, and passes it this `config` object. Then we run `GET /api/config` against that fake server to get all the config keys that we created. Checking that what is returned through the http API matches the `config` object we created initially is in my opinion exactly what we want to test.

PR #2631, opened 2021-08-06
[wip] Check body is not empty before decoding request to api/config/get

anjannath:
This enables us to fetch all the configs, if an empty body and GET
is sent to /api/config

```
*   Trying /Users/anjan/.crc/crc-http.sock...
* Connected to unix (/Users/anjan/.crc/crc-http.sock) port 80 (#0)
> GET /api/config/get HTTP/1.1
> Host: unix
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Date: Fri, 06 Aug 2021 07:33:04 GMT
< Content-Length: 855
< Content-Type: text/plain; charset=utf-8
<
* Connection #0 to host unix left intact
{"Success":true,"Error":"","Configs":{"autostart-tray":true,"bundle":"/Users/anjan/.crc/cache/crc_hyperkit_4.8.3.crcbundle","consent-telemetry":"no","cpus":4,"disable-update-check":false,"disk-size":66,"enable-cluster-monitoring":false,"enable-experimental-features":false,"host-network-access":false,"http-proxy":"","https-proxy":"","kubeadmin-password":"","memory":9216,"nameserver":"","network-mode":"user","no-proxy":"","proxy-ca-file":"","pull-secret-file":"","skip-check-admin-helper-cached":false,"skip-check-bundle-extracted":false,"skip-check-hyperkit-driver":false,"skip-check-hyperkit-installed":false,"skip-check-m1-cpu":false,"skip-check-obsolete-admin-helper":false,"skip-check-qcow-tool-installed":false,"skip-check-ram":false,"skip-check-resolver-file-permissions":false,"skip-check-root-user":false,"skip-check-supported-cpu-arch":false}}* Closing connection 0
```

### DIFF
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -215,8 +215,10 @@ func (h *Handler) UnsetConfig(c *context) error {
 
 func (h *Handler) GetConfig(c *context) error {
 	var req client.GetOrUnsetConfigRequest
-	if err := c.Bind(&req); err != nil {
-		return err
+	if len(c.requestBody) > 0 {
+		if err := c.Bind(&req); err != nil {
+			return err
+		}
 	}
 
 	if len(req.Properties) == 0 {


cfergeau:
Instead of special-casing this, maybe it's better to require a valid JSON body in this situation as well? Either just `{}` (forgot if this is valid json), or even something more expressive `{ all-keys }`? Is it possible to add a test case for this? (did not open api_http_test.go, so no idea if it's easily doable)

anjannath:
> Instead of special-casing this, maybe it's better to require a valid JSON body in this situation as well? Either just `{}` (forgot if this is valid json), or even something more expressive `{ all-keys }`? 

If we mandate that clients pass a valid json `{}` this patch is not needed and works as expected.

> Is it possible to add a test case for this? (did not open api_http_test.go, so no idea if it's easily doable)

Yes, its not difficult, just have to add the full config struct for assertion

anjannath:
Closing in favor of #2629 which also fixes this regression

Issue #2525, opened 2021-07-05
Add configuration option for the preset in ~/.crc/crc.json


guillaumerose:
(task from epic https://github.com/code-ready/crc/issues/2167)

anjannath:
Need to decide the following:
- [ ] Name of the config option (`preset`/`mode`) ?
- [ ] Names of the configurations

we can call the config option `mode` and the configurations `podman-mode`, `minimal-crc-mode`, `full-crc-mode`, WDYT?

guillaumerose:
I think we can start with ` preset=podman|minimal|full`  ?

anjannath:
presets have been part of crc since quite a few release now, closing

PR #2634, opened 2021-08-09
preset: add configuration key

guillaumerose:
Valid values are podman, minimal and full. The value is not used for the
moment.

**Fixes:** Issue #2525

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -11,6 +11,14 @@ import (
 	"github.com/spf13/cast"
 )
 
+type Preset string
+
+const (
+	Podman  Preset = "podman"
+	Minimal Preset = "minimal"
+	Full    Preset = "full"
+)
+
 const (
 	Bundle                  = "bundle"
 	CPUs                    = "cpus"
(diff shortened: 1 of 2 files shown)


pkg/crc/config/settings.go
```diff
@@ -109,6 +118,9 @@ func RegisterSettings(cfg *Config) {
 
 	cfg.AddSetting(KubeAdminPassword, "", ValidateString, SuccessfullyApplied,
 		"User defined kubeadmin password")
+
+	cfg.AddSetting(PresetConfigurationKey, string(Minimal), validatePreset, RequiresDeleteMsg,
```
praveenkumar, line 122: @guillaumerose here we are considering `minimal` as what we have today not the microshift one right?

guillaumerose: oh yes good point. should it be `Podman`, `Microshift`, `OpenshiftInstallerCustomized`, `OpenshiftInstallerDefault`?

praveenkumar: @guillaumerose let's discuss it tomorrow team meeting and then we will make this change.

cfergeau: `Podman`, `Microshift` (`MicroShift`?), `OpenShiftMinimal`, `OpenShiftFull`?

PR #2640, opened 2021-08-12
Remove machine config render check for bundle generation

praveenkumar:
Without debug it is hard to find why bundle generation is failing. Right
now `machine config is not rendered after removing pull secret (x22)` is
coming from the CI but need to make sure if mco really not able to
create new config after pull secret update.

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -180,42 +180,14 @@ func EnsureGeneratedClientCAPresentInTheCluster(ocConfig oc.Config, sshRunner *s
 }
 
 func RemovePullSecretFromCluster(ocConfig oc.Config, sshRunner *ssh.Runner) error {
-	mcBeforeRemovePullSecret, stderr, err := ocConfig.RunOcCommand("get mc --sort-by=.metadata.creationTimestamp --no-headers -oname")
-	if err != nil {
-		return fmt.Errorf("failed to get machineconfig resource %w: %s", err, stderr)
-	}
 	logging.Info("Removing user's pull secret from instance disk and from cluster secret...")
 	cmdArgs := []string{"patch", "secret", "pull-secret", "-p",
 		`'{"data":{".dockerconfigjson":"e30K"}}'`,
 		"-n", "openshift-config", "--type", "merge"}
 
-	_, stderr, err = ocConfig.RunOcCommand(cmdArgs...)
+	_, stderr, err := ocConfig.RunOcCommand(cmdArgs...)
 	if err != nil {
 		return fmt.Errorf("Failed to remove Pull secret %w: %s", err, stderr)
-	}
-
-	mcRenderedFunc := func() error {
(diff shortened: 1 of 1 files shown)


adrianriobo (approved):

test/e2e/features/basic.feature
```diff
@@ -148,7 +148,7 @@ Feature: Basic test
         # This will remove the pull secret from the instance and from the cluster
         # You need to provide pull secret file again if you want to start this cluster
         # from a stopped state.
-        Given executing "crc bundle generate -f" succeeds
+        Given executing "crc bundle generate -f --log-level debug" succeeds
```
cfergeau, line 151: Fwiw, rather than patching each crc execution we want debug for, we could take advantage of ` test/extended/crc/cmd/cmd.go` to force it for all `crc` use in the test suite, something like https://github.com/cfergeau/crc/commit/678cf7d909aae5c7a107d8dc0abf0857cfeed8e0 in combination with https://github.com/cfergeau/crc/commit/93c52920073aa13dcc573836dab0ee5910000ee3

praveenkumar:
retest this please

cfergeau:
One thing is not clear to me regarding the log saying if the only thing which changed is the pull secret.
Let's say I start the cluster from scratch (pull secret is initially empty, a new one is added, config is rendered). I work with the cluster, including MCO changes (waiting for the config to be rendered). Then I create a new bundle. Will it also be able to reuse a preexisting config from before?

praveenkumar:
> One thing is not clear to me regarding the log saying if the only thing which changed is the pull secret.
> Let's say I start the cluster from scratch (pull secret is initially empty, a new one is added, config is rendered). I work with the cluster, including MCO changes (waiting for the config to be rendered). Then I create a new bundle. Will it also be able to reuse a preexisting config from before?

@cfergeau It will support to render because your flow is something like
- Cluster started with empty pull secret 
-  MCO rendered because you add your pull secret.
- You again make some changes to MCO (not related to pull secret) so again rendered
- Now you are trying to create bundle but since your initial rendered config doesn't have your MCO change which you did in step 3 so it will going to render again.

If there is no step-3 (no MCO changes) then mco can reuse the initial rendered config.

cfergeau:
This means sometimes the current code is needed, sometimes not? If we remove this code, what happens if a render is needed but we don't wait for it?

guillaumerose:
Sorry I blindly merge to continue the release. I didn't see the above question.
Should I hold on a bit?

praveenkumar:
> This means sometimes the current code is needed, sometimes not? If we remove this code, what happens if a render is needed but we don't wait for it?

This doesn't break the cluster, what might be happen that pending render will then happen when user start the cluster from generated bundle if there is any. All we are trying with this render wait is to make sure the pull secret is not part of the cluster and that we also check with ssh to the VM and reading that file. So we are safe even if sometime render happen or sometime it doesn't.

PR #2648, opened 2021-08-16
Send json response for pull-secret route

praveenkumar:
This patch will change the pull-secret response from string to json

```
$ curl --unix-socket ~/.crc/crc-http.sock http:/c/api/pull-secret | jq .
{
  "Success": true,
  "Error": ""
}
```

### DIFF
--- a/pkg/crc/api/api_http.go
+++ b/pkg/crc/api/api_http.go
@@ -3,6 +3,7 @@ package api
 import (
 	"net/http"
 
+	"github.com/code-ready/crc/pkg/crc/api/client"
 	"github.com/code-ready/crc/pkg/crc/cluster"
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/machine"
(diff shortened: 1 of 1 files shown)


anjannath (dismissed):
LGTM, i was thinking maybe we could instead do `c.WriteHeader(http.StatusOK)` since the returned result is not used, but this is consistent with all the other handlers returning json reponses

gbraad:
Not yet... will do as I have a binary

gbraad:
Does not work for me with the current client code.

praveenkumar:
> Does not work for me with the current client code.

Which means we need to improve on client side.

gbraad:
> Which means we need to improve on client side.

Yes. 

```
System.Text.Json.JsonException: 'The JSON value could not be converted to System.Boolean. Path: $ | LineNumber: 0 | BytePositionInLine: 1.'
```

The Windows codebase is forced to a typed result of type `<bool>`

https://github.com/code-ready/tray-windows/blob/482f4dafec482add25cbdbd92567521426807745/Communication/DaemonCommander.cs#L90

This would need to change for this to work

pkg/crc/api/api_http.go
```diff
@@ -49,15 +50,22 @@ func setPullSecret() func(c *context) error {
 		if err := cluster.StoreInKeyring(string(c.requestBody)); err != nil {
 			return err
 		}
-		return c.String(http.StatusCreated, "")
+		return c.JSON(http.StatusCreated, client.Result{
+			Success: true,
```
guillaumerose, line 54: I feel Success is a bit redundant with the status code. Is it needed for POST?

pkg/crc/api/api_http.go
```diff
@@ -49,15 +50,22 @@ func setPullSecret() func(c *context) error {
...
+		})
 	}
 }
 
 func getPullSecret(config crcConfig.Storage) func(c *context) error {
 	return func(c *context) error {
 		if _, err := cluster.NewNonInteractivePullSecretLoader(config, "").Value(); err == nil {
-			return c.String(http.StatusOK, "")
+			return c.JSON(http.StatusOK, client.Result{
+				Success: true,
```
guillaumerose, line 63: Maybe we can send a useful information about the pull secret here? Like the username of the user, the size of the pull secret, the pull secret is valid: true|false. Success seems redundant.

It is just a proposal ^^

gbraad: Would not be a good idea. This can be regarded of leaking PII.

I would either, if an actual JSON object is requested, to just return `{ }`

PR #2651, opened 2021-08-16
Create sha256sum for windows and mac release artifacts

praveenkumar:


Makefile
```diff
@@ -285,8 +285,8 @@ packagedir: clean check_bundledir $(BUILD_DIR)/macos-amd64/crc $(HOST_BUILD_DIR)
 $(BUILD_DIR)/macos-amd64/crc-macos-amd64.pkg: packagedir
 	./packaging/package.sh $(BUILD_DIR)/macos-amd64
 
-$(BUILD_DIR)/macos-amd64/crc-installer.tar: packagedir
-	tar -cvf $(BUILD_DIR)/macos-amd64/crc-installer.tar ./packaging
+$(BUILD_DIR)/macos-amd64/crc-macos-installer.tar: packagedir
+	tar -cvf $(BUILD_DIR)/macos-amd64/crc-macos-installer.tar ./packaging
```
cfergeau, line 289: What is the impact for people consuming this tarball? Do they need to adjust some scripts?

cfergeau: `tar -cvf $(BUILD_DIR)/macos-amd64/crc-macos-installer.tar ./packaging` could be `tar -cvf $@ ./packaging` I think

Makefile
```diff
@@ -329,7 +329,7 @@ endif
 
 $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi: msidir
 	candle.exe -arch x64 -ext WixUtilExtension -o $(PACKAGE_DIR)/msi/ $(PACKAGE_DIR)/msi/*.wxs
-	cd $(PACKAGE_DIR)/msi && light.exe -ext WixUIExtension -ext WixUtilExtension -sacl -spdb -sice:ICE61 -sice:ICE69 -out ../../../$@ *.wixobj
+	cd $(PACKAGE_DIR)/msi && light.exe -ext WixUIExtension -ext WixUtilExtension -sacl -spdb -sice:ICE61 -sice:ICE69 -out $(BUILD_DIR)/windows-amd64/$@ *.wixobj
```
cfergeau, line 333: I'm a bit confused by this change, I'd expect `$@` to expand to the target name, which is `$(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi` here

```
$ cat Makefile 
all: /tmp/makefile-test

/tmp/makefile-test:
	echo "$@"
	touch "$@"

$ make
echo "/tmp/makefile-test"
/tmp/makefile-test
touch "/tmp/makefile-test"

File: Makefile [hunk 3]
```diff
@@ -287,6 +287,7 @@ $(BUILD_DIR)/macos-amd64/crc-macos-amd64.pkg: packagedir
 
 $(BUILD_DIR)/macos-amd64/crc-macos-installer.tar: packagedir
 	tar -cvf $(BUILD_DIR)/macos-amd64/crc-macos-installer.tar ./packaging
+	pushd $(BUILD_DIR)/macos-amd64 && sha256sum crc-macos-installer.tar > crc-macos-installer.tar.sha256sum && popd
```
cfergeau, line 290: `sha256sum crc-macos-installer.tar > $(BUILD_DIR)/macos-amd64/crc-macos-installer.tar.sha256sum` ?

Makefile
```diff
@@ -330,6 +331,7 @@ endif
 $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi: msidir
 	candle.exe -arch x64 -ext WixUtilExtension -o $(PACKAGE_DIR)/msi/ $(PACKAGE_DIR)/msi/*.wxs
 	cd $(PACKAGE_DIR)/msi && light.exe -ext WixUIExtension -ext WixUtilExtension -sacl -spdb -sice:ICE61 -sice:ICE69 -out $(BUILD_DIR)/windows-amd64/$@ *.wixobj
+	pushd $(BUILD_DIR)/windows-amd64 && sha256sum crc-windows-amd64.msi > crc-windows-amd64.msi.sha256sum && popd
```
cfergeau, line 334: same comment

Makefile
praveenkumar: `crc-macos-installer.tar` file not on the top level dir so we do need to cd to that directory and execute right?

Makefile
praveenkumar: @cfergeau this is something I asked to @anjannath, will look more closely

Makefile
praveenkumar: >What is the impact for people consuming this tarball? Do they need to adjust some scripts?

I don't think anyone except us is consuming this tarball because this is provided to RCM team to sign it.

cfergeau: Then the RCM team is consuming the tarball ;) Do they need to adjust their scripts? Can we rename it to `crc-macos-amd64-unsigned.tar` ?

Makefile
praveenkumar: @cfergeau If I use that then following will happen which means if we need to verify the checksum either user manually have to cat the file and check with existing or need to create same directory structure and put the target file there.
```
$ sha256sum /tmp/test/bar.txt > /tmp/test/bar.txt.sha256sum
$ cat /tmp/test/bar.txt.sha256sum 
a254679f30b349b670a5098d98e2b47f4a09af80291a81e4eb148e392a52ba82  /tmp/bar.txt
$ mv /tmp/test/* /tmp
$ rm -fr /tmp/test
$ $ sha256sum -c /tmp/bar.txt.sha256sum 
sha256sum: /tmp/test/bar.txt: No such file or directory
/tmp/test/bar.txt: FAILED open or read
sha256sum: WARNING: 1 listed file could not be read
```

cfergeau: Ah yes, did not think this would end up adding relative paths to the sha256sum file :)

Makefile
praveenkumar: > Then the RCM team is consuming the tarball ;) Do they need to adjust their scripts?

@cfergeau  I don't think they consume it as part of script, they first download this artifact using wget and then run what is required.

> Can we rename it to crc-macos-amd64-unsigned.tar ?

This is the tarball not the `.pkg` file which require signed so I am not sure why we have to put `unsigned` on the tarball side?

Makefile
anjannath: Ahh, my bad, i thought its only the last part of the target and not the whole target, and since $(BUILD_DIR) is `out/`, to have the full path we needed the `../../`

Makefile
```diff
@@ -330,6 +331,7 @@ endif
 $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi: msidir
 	candle.exe -arch x64 -ext WixUtilExtension -o $(PACKAGE_DIR)/msi/ $(PACKAGE_DIR)/msi/*.wxs
 	cd $(PACKAGE_DIR)/msi && light.exe -ext WixUIExtension -ext WixUtilExtension -sacl -spdb -sice:ICE61 -sice:ICE69 -out ../../../$@ *.wixobj
+	pushd $(BUILD_DIR)/windows-amd64 && sha256sum crc-windows-amd64.msi > crc-windows-amd64.msi.sha256sum && popd
```
cfergeau, line 334: Fwiw, looks like this can be:
```
pushd $(@D) && sha256sum $(@F)>$(@F).sha256sum && popd
```
see https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html#Automatic-Variables
You don't really need `pushd`/`popd`, this can be just `cd $(@D); sha256sum ...`. I think this runs in a subshell, so the directory is unchanged for the following commands.

This could even go to its own Makefile rule:
```
%.sha256sum: %
    cd $(@D); sha256sum $(^F)>$(@F)

msi: $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi.sha256sum
```

praveenkumar:
> They are using a script.

@cfergeau In this case we will ask them to update the script, on the ticket I see they run the commands (may be those from the script), if we merge this soon then we should able to communicate them early to fix it.

> Just for clarity, when looking over a file list, so that it's very obvious that this is not the final archive. Not really important.
But I'd add an amd64 suffix as aarch64 is around the corner.

Yes will add `amd64`.

Makefile
```diff
@@ -285,8 +285,8 @@ packagedir: clean check_bundledir $(BUILD_DIR)/macos-amd64/crc $(HOST_BUILD_DIR)
 $(BUILD_DIR)/macos-amd64/crc-macos-amd64.pkg: packagedir
 	./packaging/package.sh $(BUILD_DIR)/macos-amd64
 
-$(BUILD_DIR)/macos-amd64/crc-installer.tar: packagedir
-	tar -cvf $(BUILD_DIR)/macos-amd64/crc-installer.tar ./packaging
+$(BUILD_DIR)/macos-amd64/crc-macos-installer-amd64.tar: packagedir
+	tar -cvf $(BUILD_DIR)/macos-amd64/crc-macos-installer-amd64.tar ./packaging
```
cfergeau, line 289: This one could also use `$@`, this makes name changes easier.

cfergeau (approved):

Makefile
praveenkumar: @cfergeau good catch, I did thought it but looks like missed it :(

Makefile
```diff
@@ -330,6 +331,7 @@ endif
 $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi: msidir
 	candle.exe -arch x64 -ext WixUtilExtension -o $(PACKAGE_DIR)/msi/ $(PACKAGE_DIR)/msi/*.wxs
 	cd $(PACKAGE_DIR)/msi && light.exe -ext WixUIExtension -ext WixUtilExtension -sacl -spdb -sice:ICE61 -sice:ICE69 -out ../../../$@ *.wixobj
+	cd $(@D) && sha256sum $(@F)>$(@F).sha256sum
 
 CABS_MSI = "cab1.cab,cab2.cab,cab3.cab,crc-windows-amd64.msi"
 $(BUILD_DIR)/windows-amd64/crc-windows-installer.zip: $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi
```
anjannath, line 337: i think it'll be better to generate the checksum for the zip, this is the artifact we upload in the build server.

PR #2657, opened 2021-08-19
Use go template to generate the msi wxs file

### LINKED ISSUE #2645: Have logic to dynamic create bundle split for windows msi
As of now we are using a kind of hard coded way to split the bundle files to accommodate it for windows msi but this is depend on bundle size so if bundle size reduced/increased then msi creation will fail. Since we don't have any CI in place to capture that on bundle bases, it is identified during the release and which make use to complete respin the release artifacts after adjusting the bundle split. It would be better if we can have a logic to dynamically handle it so we don't have to adjust it as per bundle size.

anjannath:
Should not change anything on the UX, apart from the fact that, now the last cab file is always embedded in the msi so in the result .zip archive we'll have one less cab file then before and a larger size for `crc-windows-amd64.msi` file

#### Testing
- Generated msi should successfully install and setup crc on windows 

Fixes #2645


Makefile
```diff
@@ -334,5 +334,5 @@ CABS_MSI = "*.cab,crc-windows-amd64.msi"
 $(BUILD_DIR)/windows-amd64/crc-windows-installer.zip: $(BUILD_DIR)/windows-amd64/crc-windows-amd64.msi
 	rm -f $(HOST_BUILD_DIR)/crc.exe
 	rm -f $(HOST_BUILD_DIR)/crc-embedder
-	rm -f $(HOST_BUILD_DIR)/split
+	rm -f $(HOST_BUILD_DIR)/GenMsiWxs
```
praveenkumar, line 337: @anjannath I don't have issue with tool name `GenMsiWxs` but the file name should be `gen_msi_wxs.go` , In go variable name should follow mixedCaps not the filenames.

anjannath: Right, will change the name to `gen_msi_wxs.go`, thanks

Makefile
```diff
@@ -304,15 +304,15 @@ $(GOPATH)/bin/gomod2rpmdeps:
 	     -e 's/__OPENSHIFT_VERSION__/'$(BUNDLE_VERSION)'/g' \
 	     $< >$@
 
-$(HOST_BUILD_DIR)/split: packaging/windows/split.go
-	go build -o $(HOST_BUILD_DIR)/split -ldflags="-X main.crcVersion=$(CRC_VERSION)" packaging/windows/split.go
+$(HOST_BUILD_DIR)/GenMsiWxs: packaging/windows/gen_msi_wxs.go
+	go build -o $@ -ldflags="-X main.crcVersion=$(CRC_VERSION)" $<
```
praveenkumar, line 308: 👍🏼

praveenkumar (approved):

PR #2659, opened 2021-08-19
api: Split tests in separate functions

cfergeau:
It's more convenient to have more focused tests for each API endpoint
rather than one big "test all" method.

### DIFF
--- a/pkg/crc/api/api_http_test.go
+++ b/pkg/crc/api/api_http_test.go
@@ -17,14 +17,32 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestHTTPApi(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/crc/api/api_http_test.go
```diff
@@ -17,14 +17,32 @@ import (
...
-func TestHTTPApi(t *testing.T) {
+type testClient struct {
+	*apiClient.Client
+	config     crcConfig.Storage
+	httpServer *httptest.Server
+}
+
+func (client *testClient) Close() {
+	client.httpServer.Close()
+}
```
praveenkumar, line 28: lgtm

praveenkumar (approved):

Issue #2656, opened 2021-08-18
Review configs around version variables


adrianriobo:
Currently this variable is used to set the default network mode on macos:

https://github.com/code-ready/crc/blob/e50d8813a5570eeb924742c4bd1ea472f09222a6/pkg/crc/config/settings.go#L124

In the past when two different distributables for macos with different defaults networking modes, that variable make sense, now there is only one default mode for networking on macos so probably this should not be dependent on this variable.

One of the issues with this variable appears on integration tests, which make use of machine Client class, so this means integration tests should be build with `macosInstallPath` set.

adrianriobo:
Just find out same can be applied on `windows` to variable `msiBuild`

https://github.com/code-ready/crc/blob/e50d8813a5570eeb924742c4bd1ea472f09222a6/pkg/crc/version/version.go#L28

cfergeau:
Couple of thoughts/questions on this:
- any reasons the integration test Makefile target does not set this variable? (not necessarily the best long term solution, but could help for now)
- I started https://github.com/cfergeau/crc/tree/deadcode to rework the code around `msiBuild`, `macosInstallPath`, ... I think it should not be too hard to replace `macosInstallPath` with just a boolean (the same as on Windows), and to default to this boolean being true, we'd require its value to be changed when building from source (just thinking out loud for now though :)
- at this point, I think we can bite the bullet and just change the default networking mode for all platforms, not have some installer/not installer special casing in `defaultNetworkMode`. Everything that we hand out to our customers should be able to deal with usermode networking now (macOS and Windows through their respective installers, linux through vsock socket activation of the daemon)

adrianriobo:
> any reasons the integration test Makefile target does not set this variable? (not necessarily the best long term solution, but could help for now)

Well right now integration goal on Makefile only run tests and does not distinguish on platforms...I am working towards build the integration container which will hold the integration binary , at that point if still existing may make more sense to add those. Otherwise as you said is not a long term solution so I wouild prefer no add nothing about this right now. 
Instead I am doing that customization manually before running those tests, not best option but is short term solution.

> I started https://github.com/cfergeau/crc/tree/deadcode to rework the code around msiBuild, macosInstallPath, ... I think it should not be too hard to replace macosInstallPath with just a boolean (the same as on Windows), and to default to this boolean being true, we'd require its value to be changed when building from source (just thinking out loud for now though :)

At fisrst thought seems fine to me

[...]

PR #2661, opened 2021-08-19
Rework macosInstallPath/msiBuild linker variable

cfergeau:
crc uses 2 different variables for macos/msi releases, with subtly different behaviour on the 2 platforms when the installer is used, and some code duplication.
This PR tries to address this.

**Relates to:** Issue #2656

## Solution/Idea

This will introduce a single `isInstaller` linker variable, which will be used on the 2 platforms.

## Proposed changes

1. **The default networking mode is changed to be user mode on all platforms** (it was already this way on macOS and Windows), see commit `settings: Use user-mode networking by default on all platforms`
2. **It's no longer possible to change the networking mode on Windows installer builds**. This is similar to what is done for macOS installer builds, see commit `settings: Force use of user networking on Windows`
3. `IsMacosInstallPathSet()`/`IsMsiBuild()` are replaced with `IsInstaller()`
4. GetMsiInstallPath()/GetMacosInstallPath() are replaced with `InstallPath()`

Items 1. and/or 2. can easily be removed without impacting the rest of the patch series if they are not desirable right now.

## Testing

1. check that after installation, crc is still able to properly use its misc binaries (helpers, trays, ..), and still work properly.


praveenkumar:
@cfergeau we need to update the e2e tests if we are switching the network mode to usermode for linux also.

praveenkumar:
@cfergeau I should separate https://github.com/code-ready/crc/pull/2661/commits/24a4e79e7c4219b11d504bdbf933bb2017b5d4f7 one to a different PR because here we are changing the network mode to user for Linux and that means our documents around remote https://crc.dev/crc/#setting-up-remote-server_gsg need to be updated.

cfergeau:
> @cfergeau we need to update the e2e tests if we are switching the network mode to usermode for linux also.

Yep, was looking into this. I'll first fix CI, then I'll see what needs to be done for default networking, don't be surprised if the follow-up iterations do nothing for default networking at first.

pkg/crc/constants/constants_darwin.go
```diff
@@ -12,7 +12,7 @@ const (
 )
 
 var (
-	TrayExecutablePath   = filepath.Join(filepath.Dir(filepath.Dir(version.GetMacosInstallPath())), "MacOS", "CodeReady Containers")
+	TrayExecutablePath   = filepath.Clean(filepath.Join(version.GetMacosInstallPath(), "..", "MacOS", "CodeReady Containers"))
```
praveenkumar: https://play.golang.org/p/IOyL1dQXi5A looks like this will change the path of installer side since we are using `/Applications/CodeReady Containers.app/Contents/Resources/` In the makefile and result for both implementation is different.

cfergeau: > https://play.golang.org/p/IOyL1dQXi5A looks like this will change the path of installer side

The link you shared is missing one call to `filepath.Dir()`, `filepath.Dir()` is called twice: https://play.golang.org/p/Q3aDTmkM0uY

pkg/crc/constants/constants_darwin.go
```diff
@@ -12,7 +12,14 @@ const (
...
 	DaemonHTTPSocketPath = filepath.Join(CrcBaseDir, "crc-http.sock")
 )
+
+func TrayExecutablePath() string {
+	if version.IsInstaller() {
+		return filepath.Clean(filepath.Join(version.InstallPath(), "..", "MacOS", "CodeReady Containers"))
+	}
+	// Should not be reached, tray is only supported on installer builds
+	return filepath.Clean(filepath.Join(BinDir(), "CodeReady Containers"))
+}
```
praveenkumar, line 25: @cfergeau As per commit message `The code is calling InstallPath() even on crc versions where it was not
set at build time. Since InstallPath() is printing a warning when this occurs, it's better if we can avoid this.`  Does it mean that when this PR is in this can still occur because we are already setting it in Makefile right?

cfergeau: If you build crc from git on macos, `TrayExecutablePath` was initialized using `InstallPath()` unconditionally even if the tray code can only be triggered in "install" builds. So a warning was printed each time `crc` was started. By moving it to a function, this will only be called when needed, which should be never on git builds, so it's only called when building an installer.

PR #2663, opened 2021-08-20
Update bundle version to 4.8.5 (candidate channel)

praveenkumar:


cfergeau:
4.8.5 is in the `fast` channel, `candidate` has 4.8.6, but this version won't be promoted to `fast`/`stable`, so 4.8.5 as picked by this PR is better. ACK.

cfergeau (approved):

Issue #2598, opened 2021-07-22
[BUG] upgrade unexpected behavior


adrianriobo:
Testing pre release bits `1.30.0` on windows we got unexpected behavior:

1. Got version `1.29.1` installed with msi (install and restart the computer)

2. Try to upgrade to `1.30.0` with `1.29.1` running, we want to avoid restart the computer. As hyper-v and user and groups are already set up from previous installation the restart should not be necessary.

3. When `1.30,0` installation finishes it asks for reboot. Is this necessary?. We do not reboot.

4. Then `crc setup` finish successfully telling us we can start crc as usual. But tray is always reporting error and exiting. If we check the `crc setup --check-only` logs we can see:

```bash
INFO Checking if admin-helper daemon is installed
DEBU Get http://unix/version: open \\.\pipe\crc-admin-helper: Access is denied.
Get http://unix/version: open \\.\pipe\crc-admin-helper: Access is denied.
```

adrianriobo:
From `crc setup`:

```powershell
PS C:\Users\crcqe> crc setup
INFO Checking if crc-admin-helper executable is cached
INFO Checking for obsolete admin-helper executable
INFO Checking if running on a supported CPU architecture
INFO Checking minimum RAM requirements
INFO Checking if running in a shell with administrator rights
INFO Checking Windows 10 release
INFO Checking Windows edition
INFO Checking if Hyper-V is installed and operational
INFO Checking if crc-users group exists
INFO Checking if current user is in Hyper-V group and crc-users group
INFO Checking if Hyper-V service is enabled
INFO Checking if the Hyper-V virtual switch exists
INFO Found Virtual Switch to use: Default Switch
INFO Checking if admin-helper daemon is installed
INFO Installing admin-helper daemon
INFO Will run as admin: install admin-helper daemon
INFO Checking if vsock is correctly configured
INFO Checking if CodeReady Containers daemon is installed
INFO Checking if tray executable is present
INFO Checking if tray is running
INFO Starting CodeReady Containers tray
INFO Checking if CRC bundle is extracted in '$HOME/.crc'
INFO Checking if C:\Program Files\CodeReady Containers\crc_hyperv_4.8.0.crcbundle exists
INFO Extracting bundle from the CRC executable
INFO Ensuring directory C:\Program Files\CodeReady Containers exists
INFO Uncompressing crc_hyperv_4.8.0.crcbundle
[...]

PR #2668, opened 2021-08-25
Don't remove crc-users group and current user from it during upgrade

anjannath:
We want to avoid restart when crc is upgraded using the msi installer.

the installer does a uninstall in the RemoveExistingProducts action
when performing upgrades, to avoid a reboot we need to skip the custom
action that removes the crc-users group when the msi detects an upgrade

FIrst part for fixing #2598 (more details in: https://github.com/code-ready/crc/issues/2598#issuecomment-905325117)

### How to test
1. Download msi from appveyor ci https://ci.appveyor.com/api/buildjobs/i6lo0x4r0l8xqlby/artifacts/out%2Fwindows-amd64%2Fcrc-windows-installer.zip
2. Installation should succeed without any errors and after reboot the tray should be started
3. Uninstallation should succeed (verify that the `crc-users` group is deleted, we want to keep it only during upgrades)


guillaumerose (approved):
works for me

Issue #2468, opened 2021-06-22
Force shutdown VM during the bundle generation


praveenkumar:
Recently with 4.8 testing one of the blocking issue we found is vm taking longer time to shutdown gracefully (~7-8 mins) and a bug is filled against it https://bugzilla.redhat.com/show_bug.cgi?id=1965992. As part of bundle creation we first stop the cluster and that run `client.Stop` functionality which waits for 2 min before error out and currently if there is any error we just error out from the bundle generate command.  Should we prompt user for forcefully shutdown after 2min like we do for `crc stop` command or we just ourself perform the force shutdown? Also what if disk corruption happen due to forcefully shutdown of VM?

This is follow up issue of the comment https://github.com/code-ready/crc/pull/2324#discussion_r655196827

cfergeau:
I would follow exactly the same behaviour as `stop`, including having a `-f` flag. Or are there important reasons forcing us to differ from `stop` behaviour in the case of bundle generation?

praveenkumar:
> I would follow exactly the same behaviour as `stop`, including having a `-f` flag. Or are there important reasons forcing us to differ from `stop` behaviour in the case of bundle generation?

How about just having `-f` option and if user don't provide it then we error out for stop otherwise forcefully shutdown?

cfergeau:
Could work... If stop is too slow, behaviour will not be great anyway, as `crc generate-bundle` will fail, but if make sure the VM is stopped, and re-run immediatly `crc generate-bundle`, the bundle generation will most likely also fail because the first steps require the VM to be up :(

cfergeau:
A `--forceStop` option was added in 02a30bf8bbf239703d

PR #2675, opened 2021-08-27
Issue #2468 remove Error and Success from response structs 

### LINKED ISSUE #2512: [Task] Remove `Error` and `Success` as HTTP return values
The values for Error and Success should be removed from the HTTP protocol as they are misleading.
Ref: #2492 #2501 https://github.com/code-ready/tray-windows/pull/93

Note: This work needs to precede work on a mock of the daemon.

anjannath:
Fixes #2512 

Should not change anything on the ux side, however needs to be tested with both the trays to make sure there's no regression

### DIFF
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -29,9 +29,12 @@ type Telemetry interface {
 	UploadAction(action, source, status string) error
 }
 
+type loggerResult struct {
+	Messages []string
+}
+
 func (h *Handler) Logs(c *context) error {
 	return c.JSON(http.StatusOK, &loggerResult{
-		Success:  true,
 		Messages: h.Logger.Messages(),
 	})
 }
@@ -57,7 +60,6 @@ func (h *Handler) Status(c *context) error {
 		PodmanVersion:    res.PodmanVersion,
 		DiskUse:          res.DiskUse,
 		DiskSize:         res.DiskSize,
-		Success:          true,
 		Preset:           res.Preset,
 	})
 }
@@ -102,7 +104,6 @@ func (h *Handler) Start(c *context) error {
 		return err
 	}
 	return c.JSON(http.StatusOK, client.StartResult{
-		Success:        true,
 		Status:         string(res.Status),
 		ClusterConfig:  res.ClusterConfig,
 		KubeletStarted: res.KubeletStarted,
@@ -127,7 +128,6 @@ func (h *Handler) GetVersion(c *context) error {
 		CrcVersion:       version.GetCRCVersion(),
 		CommitSha:        version.GetCommitSha(),
 		OpenshiftVersion: version.GetBundleVersion(),
-		Success:          true,
 	})
 }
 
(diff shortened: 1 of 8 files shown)


pkg/crc/api/helpers.go
```diff
@@ -40,6 +40,12 @@ func (c *context) String(code int, r string) error {
 	return err
 }
 
+func (c *context) Code(code int) error {
+	var err error
+	c.code = code
+	return err
```
cfergeau, line 46: could be just `return nil`?

anjannath: yeah, i just followed the same pattern as the other methods

anjannath:
Tested with both the trays and they are not depending on the `Success` and `Error` so doesn't need any changes there.
/unhold

pkg/crc/api/handlers.go
```diff
@@ -29,9 +29,12 @@ type Telemetry interface {
 	UploadAction(action, source, status string) error
 }
 
+type loggerResult struct {
```
praveenkumar, line 32: This is part of `types.go` so why don't we update it there instead of creating here again?

pkg/crc/api/helpers.go
```diff
@@ -40,6 +40,11 @@ func (c *context) String(code int, r string) error {
 	return err
 }
 
+func (c *context) Code(code int) error {
+	c.code = code
+	return nil
+}
+
```
praveenkumar, line 46: This helper function is always going to send `nil` error and what it actually doing is just assigning the `c.code = code` we can just make `code` exportable from the struct and have use it instead creating a helper function just to do that or better name this helper to `SetCode`?

praveenkumar:
Also if we are removing the error then how do we let user know what cause the non expected status code, Are we going to document that or user have to check the `crcd` logs?

gbraad:
> Also if we are removing the error then how do we let user know what cause the non expected status code, Are we going to document that or user have to check the crcd logs?

an error is always a `500` with the error message in the body

pkg/crc/api/helpers.go
gbraad: it makes no sense to have an error that always returns nil; unexpected side-effect.

pkg/crc/api/handlers.go
anjannath: All the types in the `types.go` were already moved to `client/types.go` since most were relevant for the client and the responses from the handlers, this `loggerResult` was the only type left there and is only used in the handlers. its just a refactor to bring it here where its used, but i can keep the `types.go` file and update the struct there if you prefer that way

pkg/crc/api/helpers.go
anjannath: > This helper function is always going to send `nil` error and what it actually doing is just assigning the `c.code = code` we can just make `code` exportable from the struct and have use it instead creating a helper function just to do that or better name this helper to `SetCode`?

Yeah, `Code` is not a good name, i went with it to be consistent since other methods are named like this `c.JSON`, `c.String` to write json and text as response. 

about always returning `nil`, if we don't return an error from this function we have to return a nil from the handlers.. doing it in this function allows us to write `return c.Code(httpOK)` in the handlers.

PR #2681, opened 2021-08-30
Updated bundle to 4.8.7 (candidate channel)

praveenkumar:


cfergeau (changes requested):
4.8.7 will never ship, so I would skip this release and drop this PR.

Issue #2676, opened 2021-08-27
[BUG] `crc delete` warning


cfergeau:
With crc 1.31.2, run `crc delete` after a successful `crc start`:
```
$ ./crc  delete
Do you want to delete the OpenShift cluster? [y/N]: y
WARN open /home/teuf/.crc/machines/crc/kubeconfig: no such file or directory 
Deleted the OpenShift cluster
```
The `kubeconfig` warning is unexpected, did not check where it's coming from.

praveenkumar:
@cfergeau this should be coming when we are clearing the crc cluster entry of kubeconfig file and you might be getting it because you have `KUBECONFIG` variable set to `/home/teuf/.crc/machines/crc/kubeconfig` which is deleted as soon as delete hits.

cfergeau:
Very good guess, this is indeed the issue :)

Issue #2653, opened 2021-08-17
[BUG] System socket activation fails on RHEL 7.9 


adrianriobo:
### General information

Tested on QE Environments on RHEL 7.9 with pre release bits `1.31.2`

### Logs

```bash
evel=info msg="Checking if running as non-root"
02:34:17.428767  stderr: level=info msg="Checking if running inside WSL2"
02:34:17.428946  stderr: level=info msg="Checking if crc-admin-helper executable is cached"
02:34:17.429132  stderr: level=info msg="Caching crc-admin-helper executable"
02:34:17.488319  stderr: level=info msg="Using root access: Changing ownership of /home/cloud-user/.crc/bin/crc-admin-helper-linux"
02:34:17.515416  stderr: level=info msg="Using root access: Setting suid for /home/cloud-user/.crc/bin/crc-admin-helper-linux"
02:34:17.535492  stderr: level=info msg="Checking for obsolete admin-helper executable"
02:34:17.535712  stderr: level=info msg="Checking if running on a supported CPU architecture"
02:34:17.535738  stderr: level=info msg="Checking minimum RAM requirements"
02:34:17.535907  stderr: level=info msg="Checking if crc executable symlink exists"
02:34:17.536024  stderr: level=info msg="Creating symlink for crc executable"
02:34:17.536167  stderr: level=info msg="Checking if Virtualization is enabled"
02:34:17.536849  stderr: level=info msg="Checking if KVM is enabled"
02:34:17.537064  stderr: level=info msg="Checking if libvirt is installed"
02:34:17.611709  stderr: level=info msg="Checking if user is part of libvirt group"
02:34:17.613041  stderr: level=info msg="Checking if active user/process is currently part of the libvirt group"
02:34:17.613089  stderr: level=info msg="Checking if libvirt daemon is running"
02:34:17.639959  stderr: level=info msg="Checking if a supported libvirt version is installed"
02:34:17.663809  stderr: level=info msg="Checking if crc-driver-libvirt is installed"
02:34:17.664007  stderr: level=info msg="Installing crc-driver-libvirt"
02:34:17.728401  stderr: level=info msg="Checking crc daemon systemd service"
02:34:17.731349  stderr: level=info msg="Setting up crc daemon systemd service"
02:34:17.736079  stderr: level=info msg="Checking crc daemon systemd socket units"
02:34:17.738120  stderr: level=info msg="Setting up crc daemon systemd socket units"
[...]

Issue #1824, opened 2021-01-07
Remove embedded bundle status from version


praveenkumar:
This is the follow up issue for https://github.com/code-ready/crc/pull/1806

We changed the way embedded works and excluded the bundle from the binary.

- It means now shipped tarball consist `crc` binary and the bundle file separately.
- Setup command now don't able to extract the bundle from the binary

With current workflow user (like dev) can use `-b` option and provide the bundle location with crc and it works but this will change the normal crc user habit to just use single binary (embedded everything) and not worry about bundle location.

To overcome this user workflow we can do following and then ask user to use `-b` option.
- If bundle and crc binary is in same directory then use bundle from this directory.
- If user moved the binary but bundle is still in current USER directories we can do a find and try to get the bundle path and use it.

If both doesn't work we prompt user to use the `-b` option.

cfergeau:
> If user moved the binary but bundle is still in current USER directories we can do a find and try to get the bundle path and use it.

I'd prefer not to do this, as this is going to be slow in some cases, and can lead to unexpected results. It's also quite unusual for apps to do this.

guillaumerose:
We should display where the default bundle will be loaded. With the msi and the pkg, it is not obvious.

PR #2684, opened 2021-08-31
Several bug fixes

cfergeau:
This PR fixes 4 unrelated issues, I can file 4 different PRs if that's preferred.

The first 2 commits fix socket activation on el7 by disabling the checks, and warning the user they need to run the daemon manually. 
The CSR commits are more of a code cleanup than a bug fix, but I suspect they fix a potential issue in a corner case.
The 'version' commit improves `crc version` output regarding the bundle when using an installer, see the commit log for details.
The 'delete' commit fixes a warning on `crc delete` when `KUBECONFIG` is set to `~/.crc/machines/crc/kubeconfig.

**Fixes:** Issues #2653, #1824, #2676

### DIFF
--- a/cmd/crc/cmd/version.go
+++ b/cmd/crc/cmd/version.go
@@ -34,18 +34,24 @@ func runPrintVersion(writer io.Writer, version *version, outputFormat string) er
 }
 
 type version struct {
-	Version          string `json:"version"`
-	Commit           string `json:"commit"`
-	OpenshiftVersion string `json:"openshiftVersion"`
-	Embedded         bool   `json:"embedded"`
+	Version             string `json:"version"`
+	Commit              string `json:"commit"`
+	OpenshiftVersion    string `json:"openshiftVersion"`
+	Embedded            bool   `json:"embedded"`
+	InstalledBundlePath string `json:"installedBundlePath,omitempty"`
 }
 
 func defaultVersion() *version {
+	var installedBundlePath string
+	if crcversion.IsInstaller() {
+		installedBundlePath = constants.DefaultBundlePath
+	}
 	return &version{
-		Version:          crcversion.GetCRCVersion(),
-		Commit:           crcversion.GetCommitSha(),
-		OpenshiftVersion: crcversion.GetBundleVersion(),
-		Embedded:         constants.BundleEmbedded(),
+		Version:             crcversion.GetCRCVersion(),
+		Commit:              crcversion.GetCommitSha(),
+		OpenshiftVersion:    crcversion.GetBundleVersion(),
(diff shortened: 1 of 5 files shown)


pkg/crc/cluster/cert-renewal.go
```diff
@@ -2,25 +2,43 @@ package cluster
...
+				return fmt.Errorf("Not able to approve csr (%v : %s)", err, stderr)
+			}
+			csrsApproved = true
+		}
+		if !csrsApproved {
+			return &crcerrors.RetriableError{Err: fmt.Errorf("No Pending CSR with signerName %s", expectedSignerName)}
 		}
-		return &crcerrors.RetriableError{Err: fmt.Errorf("No Pending CSR with signerName %s", signerName)}
+		return nil
 	}, time.Second*5)
```
praveenkumar, line 42: I think this retry option should shift to `getCSRList` function because this function only provide the csr list and to get the list we need to make sure  to wait till `csr` resource is available. With this path now we are retry on the error coming from the `getCSRList` which might be a valid error not because of  of resource not available.

cfergeau: The `retry` needs to wait until:
1) a list of CSRs is returned
2) there is at least one pending CSR

If we move the `retry` to `getCSRList`, we can do 1) but not 2). I agree adding a short retry loop in `getCSRList`would probably be good though.

praveenkumar: @cfergeau Yes I am talking about the 1 where we do get the list of csr. pending logic should  be here because here we are approving the pending csr.

cfergeau: I dropped the 'csr' patches from this PR for now so that we can move forward with the other unrelated changes.

cmd/crc/cmd/version.go
```diff
@@ -59,12 +65,17 @@ func (v *version) prettyPrintTo(writer io.Writer) error {
...
+		bundleStatus = fmt.Sprintf("bundle installed at %s", v.InstalledBundlePath)
+	case v.Embedded:
+		bundleStatus = "embedded in executable"
+	default:
+		bundleStatus = "not embedded in executable"
 	}
 	return []string{
 		fmt.Sprintf("CodeReady Containers version: %s+%s\n", v.Version, v.Commit),
-		fmt.Sprintf("OpenShift version: %s (%sembedded in executable)\n", v.OpenshiftVersion, embedded),
+		fmt.Sprintf("OpenShift version: %s (%s)\n", v.OpenshiftVersion, bundleStatus),
```
praveenkumar, line 79: 👍🏼

praveenkumar (approved):

PR #2685, opened 2021-08-31
Updated bundle to 4.8.9 (candidate channel)

praveenkumar:


praveenkumar:
/hold

With proxy setting cluster have some issues.

cfergeau:
Might be a good opportunity to change the proxy code to use normal OpenShift procedures now that the MCO is enabled?

praveenkumar:
> Might be a good opportunity to change the proxy code to use normal OpenShift procedures now that the MCO is enabled?

@cfergeau I did try that also, having same issue, we do still need some proxy info to part of CRC where we make connection to segment and `oc-env` side to add no-proxy in the env. Only thing we can remove is applying proxy to the cluster and let user do it as per documents but that is also failing for us same way :(

cfergeau:
Ah, I was not suggesting to remove it, just to apply it following the normal steps. Removing the OpenShift configuration would be more consistent with what we do for everything else, but a regression from a user perspective, so not clear what is best :-/

praveenkumar:
/close

4.8.9 also not going to be part of stable channel.

PR #2688, opened 2021-09-01
ci: Always run 'crc' with '--log-level debug'

cfergeau:
Rather than manually adding '--log-level debug' here and there in feature
files, this PR adds it unconditionally for all crc command runs

### DIFF
--- a/test/extended/crc/cmd/cmd.go
+++ b/test/extended/crc/cmd/cmd.go
@@ -93,8 +93,12 @@ func envVariable(key, value string) string {
 }
 
 func (c Command) validate() error {
-	if _, ok := commands[c.command]; !ok {
-		return fmt.Errorf("%s is not a supported command", c.command)
+	cmdline := strings.Fields(c.command)
+	if len(cmdline) < 1 {
+		return fmt.Errorf("empty command? %s", c.command)
+	}
+	if _, ok := commands[cmdline[0]]; !ok {
+		return fmt.Errorf("%s is not a supported command", cmdline[0])
 	}
 	return nil
 }
--- a/test/e2e/crcsuite/crcsuite.go
+++ b/test/e2e/crcsuite/crcsuite.go
@@ -63,9 +63,9 @@ func FeatureContext(s *godog.Suite) {
 		CheckCRCStatus)
 	s.Step(`^(stdout|stderr) (?:should contain|contains) "(.*)" if bundle (is|is not) embedded$`,
 		CommandReturnShouldContainIfBundleEmbeddedOrNot)
-	s.Step(`^execute crc (.*) command$`,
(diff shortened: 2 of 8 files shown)


cfergeau:
Let this run through CI first as this is untested.

test/extended/crc/cmd/cmd.go
```diff
@@ -53,7 +53,7 @@ func (c Command) WithdisableNTP() Command {
 }
 
 func (c Command) ToString() string {
-	cmd := append(c.env(), "crc", c.command)
+	cmd := append(c.env(), "crc", "--log-level", "debug", c.command)
```
adrianriobo, line 56: I would prefer have that log-level as a property on Command, with debug as default if you want, but at least have the change to decide the level when build the command

cfergeau: I tried to do this, let me know if this is what you had in mind, or if you were suggesting something else.

adrianriobo: mostly yes, but even prefer not to limit to debug or default log level.... I mean: 
```
```

so would prefer logLevel: debug on the Command and then just use logLevel property on ToString

adrianriobo: maybe it will never be used, but in case it is needed somehow we have all the options there

adrianriobo (approved):

praveenkumar:
CI need to be fixed.

cfergeau:
Usual `crc bundle generate` failure on centos CI, so these changes are probably fine from a ci.centos.org perspective.

cfergeau:
Actually this will also happen in the 'remote crc' usecase:
- ssh into your remote machine, `crc setup`, do stuff with crc, log out when you are done
- next time you log into the machine, you'll need to rerun `crc setup` before being able to start your crc instance

Starting the `.socket` units automatically with the user session would do the trick (`systemctl enable --user crc-http.socket`, does this work with user sessions?)

cfergeau:
`e2e-crc` timed out, which I think is not unusual:
```
 *** Test killed with quit: ran too long (3h1m0s). 
```

praveenkumar:
@cfergeau I am not sure why but this pr is creating a segfault on the e2e-ci side.

cfergeau:
> @cfergeau I am not sure why but this pr is creating a segfault on the e2e-ci side. I tried 2 times of e2e test.

I don't think it's a segfault, just ` *** Test killed with quit: ran too long (3h1m0s).`

praveenkumar:
> I don't think it's a segfault, just ` *** Test killed with quit: ran too long (3h1m0s).`

@cfergeau yes right but that is only happening for this PR for each test run so some change making the test run stuck and it timeout after stuck.

PR #2689, opened 2021-09-02
Improve CSR-related code in cluster package

cfergeau:
This PR refactors a bit the CSR code to use `getCSRList` more.
This fixes 2 bugs on the way, one bug while waiting for the `kubernetes.io/kube-apiserver-client-kubelet` which probably is only theoritecal, and the other one which is a fix for the PR #2673

## Testing

1. trigger renewal of certificates (for example using the bundle from 1.30.1)
2. renewal works fine
3. it's possible to login as developer/developer after cert renewal


pkg/crc/cluster/csr.go
```diff
@@ -33,7 +33,7 @@ func deleteCSR(ocConfig oc.Config, expectedSignerName string) error {
 	}
 	for _, csr := range csrs.Items {
 		logging.Debugf("Deleting csr %s (signerName: %s)", csr.ObjectMeta.Name, expectedSignerName)
-		_, stderr, err := ocConfig.RunOcCommand("adm", "certificate", "delete", csr.ObjectMeta.Name)
+		_, stderr, err := ocConfig.RunOcCommand("delete", "csr", csr.ObjectMeta.Name)
```
praveenkumar, line 36: 👍🏼

praveenkumar (approved):

PR #2691, opened 2021-09-03
Bump OpenShift to 4.8.10

guillaumerose:


cfergeau:
4.8.10 is in stable now http://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.8/ 
We can  /unhold this I guess

praveenkumar:
bumping 4.8.11 (candidate channel) with #2698 

/close

PR #2692, opened 2021-09-03
Add .testing to default no-proxy list

praveenkumar:
This should fix the issue where auth operator is failing because of
apps-crc.testing goes through the proxy.

```
OAuthServerRouteEndpointAccessibleControllerDegraded: Get "https://oauth-openshift.apps-crc.testing/healthz": Service Unavailable
      ProxyConfigControllerDegraded: failed to reach endpoint("https://oauth-openshift.apps-crc.testing/healthz") missing in NO_PROXY(".cluster.local,.svc,10.217.0.0/22,10.217.4.0/23,127.0.0.1,192.168.126.0/24,api-int.crc.testing,localhost") with error: Get "https://oauth-openshift.apps-crc.testing/healthz": Service Unavailable
```

Note: We still need to stop and start to have everything working but
this patch is required even for this workaroud.

### DIFF
--- a/pkg/crc/network/proxy.go
+++ b/pkg/crc/network/proxy.go
@@ -10,7 +10,7 @@ import (
 
 var (
 	DefaultProxy     ProxyConfig
-	defaultNoProxies = []string{"127.0.0.1", "localhost"}
+	defaultNoProxies = []string{"127.0.0.1", "localhost", ".testing"}
 )
 
 // ProxyConfig keeps the proxy configuration for the current environment


guillaumerose (approved):

cfergeau (changes requested):
`.testing` was intentionally not added here as before loading a bundle, we can't know what domain should be `noproxy`.

cfergeau:
The relevant code is https://github.com/code-ready/crc/blob/894598107592c6316eed1af6359c7fba1e98203e/pkg/crc/machine/start.go#L316-L323
```

	proxyConfig, err := getProxyConfig(crcBundleMetadata.ClusterInfo.BaseDomain)
	if err != nil {
		return nil, errors.Wrap(err, "Error getting proxy configuration")
	}
	proxyConfig.ApplyToEnvironment()
	proxyConfig.AddNoProxy(instanceIP)
```

praveenkumar:
> `.testing` was intentionally not added here as before loading a bundle, we can't know what domain should be `noproxy`. This looks like this is papering over a regression somewhere else?

@cfergeau you are right, the way I was testing to try out MCO way of handling the proxy so I start cluster without proxy and then manually edited the proxy resource and checked and it didn't have domain entry but our codebase handle it properly.

```
WARN Proxy &network.ProxyConfig{HTTPProxy:"http://xxxxx:3128", HTTPSProxy:"http://xxxxx:3128", noProxy:[]string{"127.0.0.1", "localhost", ".testing", "192.168.130.11"}, ProxyCACert:""} 
```

I am still trying to figure out what is causing it not to work. Closing this one.

/close

PR #2698, opened 2021-09-06
Misc Proxy fixes

praveenkumar:
- No need to remove CNI network definition forcefully
- Rename dnsmasq container to dnsmasq-crc
- Create systemd unit file for dnsmasq container

How to test this PR:
- Download the latest bundle 4.8.11
- Set the proxy setting 
- Set proxy ca file
- Run crc without this patch and it will fails to stabilize the cluster after the failure you can try to export the kubeconfig file from `~/.crc/machines/crc/kubeconfig` and check `oc get nodes` which will be `Nonscheduled` because of dnsmasq container is not running (can be checked after ssh to the vm).
- Run crc with this patch and it succeed

you can use virt-viewer or virt-manager to open the crc instance to keep an eye when it rebooted by MCO.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -58,6 +58,11 @@ func init() {
 	if err != nil {
 		logging.Fatal(err.Error())
 	}
+
+	if err := setProxyDefaults(); err != nil {
+		logging.Fatal(err.Error())
+	}
+
 	// Initiate segment client
 	if segmentClient, err = segment.NewClient(config, httpTransport()); err != nil {
 		logging.Fatal(err.Error())
@@ -77,9 +82,6 @@ func runPrerun(cmd *cobra.Command) error {
 		logFile = constants.DaemonLogFilePath
 	}
 	logging.InitLogrus(logging.LogLevel, logFile)
-	if err := setProxyDefaults(); err != nil {
-		return err
-	}
 
 	for _, str := range defaultVersion().lines() {
 		logging.Debugf(str)
@@ -211,22 +213,33 @@ func attachMiddleware(names []string, cmd *cobra.Command) {
 	}
 }
 
+func defaultTransport() *http.Transport {
+	transport := http.DefaultTransport.(*http.Transport)
+	proxyConfig, err := network.NewProxyConfig()
+	if err != nil {
+		return transport
+	}
+
+	transport = transport.Clone()
+	transport.Proxy = proxyConfig.ProxyFunc()
+	return transport
+}
+
 func httpTransport() http.RoundTripper {
 	if config.Get(crcConfig.ProxyCAFile).IsDefault {
-		return http.DefaultTransport
+		return defaultTransport()
 	}
 	caCert, err := ioutil.ReadFile(config.Get(crcConfig.ProxyCAFile).AsString())
 	if err != nil {
 		logging.Errorf("Cannot read proxy-ca-file, using default http transport: %v", err)
-		return http.DefaultTransport
+		return defaultTransport()
 	}
 	caCertPool := x509.NewCertPool()
 	caCertPool.AppendCertsFromPEM(caCert)
-	return &http.Transport{
-		Proxy: http.ProxyFromEnvironment,
-		TLSClientConfig: &tls.Config{
-			MinVersion: tls.VersionTLS12,
-			RootCAs:    caCertPool,
-		},
+	transport := defaultTransport()
+	transport.TLSClientConfig = &tls.Config{
+		MinVersion: tls.VersionTLS12,
+		RootCAs:    caCertPool,
 	}
+	return transport
 }
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -216,7 +216,7 @@ func attachMiddleware(names []string, cmd *cobra.Command) {
 func defaultTransport() *http.Transport {
 	transport := http.DefaultTransport.(*http.Transport)
 	proxyConfig, err := network.NewProxyConfig()
-	if err != nil {
+	if err != nil || !proxyConfig.IsEnabled() {
 		return transport
 	}
 
--- a/go.mod
+++ b/go.mod
@@ -77,4 +77,4 @@ replace (
 	k8s.io/kubectl => github.com/openshift/kubernetes-kubectl v0.0.0-20210521075729-633333dfccda
 )
 
-replace github.com/mdlayher/vsock => github.com/cfergeau/vsock v0.0.0-20210707084117-4d87f8a20ba8
+replace github.com/mdlayher/vsock => github.com/cfergeau/vsock v0.0.0-20210707100525-4def293f672e
--- a/go.mod
+++ b/go.mod
@@ -59,7 +59,7 @@ require (
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8
 	github.com/zalando/go-keyring v0.1.1
 	golang.org/x/crypto v0.0.0-20210813211128-0a44fdfbc16e
-	golang.org/x/net v0.0.0-20210813160813-60bc85c4be6d // indirect
+	golang.org/x/net v0.0.0-20210913180222-943fd674d43e
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 	golang.org/x/sys v0.0.0-20210816074244-15123e1e1f71
 	golang.org/x/term v0.0.0-20210615171337-6886f2dfbf5b
--- a/pkg/crc/cluster/status.go
+++ b/pkg/crc/cluster/status.go
@@ -6,10 +6,11 @@ import (
 	"time"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/network"
 )
 
 // WaitForClusterStable checks that the cluster is running a number of consecutive times
-func WaitForClusterStable(ctx context.Context, ip string, kubeconfigFilePath string) error {
+func WaitForClusterStable(ctx context.Context, ip string, kubeconfigFilePath string, proxy *network.ProxyConfig) error {
 	if ctx.Err() != nil {
 		return ctx.Err()
 	}
@@ -18,6 +19,12 @@ func WaitForClusterStable(ctx context.Context, ip string, kubeconfigFilePath str
 
 	retryDuration := 30 * time.Second
 	retryCount := 20 // 10 minutes
+
+	if proxy.IsEnabled() {
+		// In case proxy is enabled increase the retry count
+		// to 10 and this will add addition 5 mins.
+		retryCount += 10
+	}
 
 	numConsecutive := 3
 	var count int // holds num of consecutive matches
--- a/pkg/crc/services/dns/dns.go
+++ b/pkg/crc/services/dns/dns.go
@@ -52,11 +52,6 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
 	// Remove the dnsmasq container if it exists during the VM stop cycle
 	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq")
 
-	// Remove the CNI network definition forcefully
-	// https://github.com/containers/libpod/issues/2767
-	// TODO: We need to revisit it once podman update the CNI plugins.
-	_, _, _ = serviceConfig.SSHRunner.Run(fmt.Sprintf("sudo rm -f /var/lib/cni/networks/podman/%s", dnsContainerIP))
-
 	// Start the dnsmasq container
 	dnsServerRunCmd := fmt.Sprintf("sudo podman run  --ip %s --name dnsmasq -v %s:/etc/dnsmasq.conf -p 53:%d/udp --privileged -d %s",
 		dnsContainerIP, dnsConfigFilePathInInstance, dnsServicePort, dnsContainerImage)
--- a/pkg/crc/services/dns/dns.go
+++ b/pkg/crc/services/dns/dns.go
@@ -10,14 +10,15 @@ import (
 	"github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/services"
+	"github.com/code-ready/crc/pkg/crc/systemd"
+	"github.com/code-ready/crc/pkg/crc/systemd/states"
 )
 
 const (
-	dnsServicePort              = 53
-	dnsConfigFilePathInInstance = "/var/srv/dnsmasq.conf"
-	dnsContainerIP              = "10.88.0.8"
-	dnsContainerImage           = "quay.io/crcont/dnsmasq:latest"
-	publicDNSQueryURI           = "quay.io"
+	dnsServicePort    = 53
+	dnsContainerIP    = "10.88.0.8"
+	publicDNSQueryURI = "quay.io"
+	crcDnsmasqService = "crc-dnsmasq.service"
 )
 
 func init() {
@@ -48,17 +49,13 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
 	if err := createDnsmasqDNSConfig(serviceConfig); err != nil {
 		return err
 	}
-
-	// Remove the dnsmasq container if it exists during the VM stop cycle
-	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq")
-
-	// Start the dnsmasq container
-	dnsServerRunCmd := fmt.Sprintf("sudo podman run  --ip %s --name dnsmasq -v %s:/etc/dnsmasq.conf -p 53:%d/udp --privileged -d %s",
-		dnsContainerIP, dnsConfigFilePathInInstance, dnsServicePort, dnsContainerImage)
-	if _, _, err := serviceConfig.SSHRunner.Run(dnsServerRunCmd); err != nil {
-		return err
+	sd := systemd.NewInstanceSystemdCommander(serviceConfig.SSHRunner)
+	if state, err := sd.Status(crcDnsmasqService); err != nil || state != states.Running {
+		if err := sd.Enable(crcDnsmasqService); err != nil {
+			return err
+		}
 	}
-	return nil
+	return sd.Start(crcDnsmasqService)
 }
 
 func getResolvFileValues(serviceConfig services.ServicePostStartConfig) (network.ResolvFileValues, error) {
--- a/pkg/crc/ssh/client.go
+++ b/pkg/crc/ssh/client.go
@@ -93,6 +93,11 @@ func (client *NativeClient) session() (*ssh.Session, error) {
 func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
 	session, err := client.session()
 	if err != nil {
+		if client.conn != nil {
+			log.Debugf("Failed to create new ssh session: %s", err)
+			client.conn.Close()
+			client.conn = nil
+		}
 		return nil, nil, err
 	}
 	defer session.Close()
--- a/pkg/crc/ssh/ssh.go
+++ b/pkg/crc/ssh/ssh.go
@@ -90,7 +90,7 @@ func (runner *Runner) runSSHCommand(command string, runPrivate bool) (string, st
 	if err != nil {
 		return string(stdout), string(stderr), fmt.Errorf(`ssh command error:
 command : %s
-err     : %w\n`, command, err)
+err     : %w`+"\n", command, err)
 	}
 
 	return string(stdout), string(stderr), nil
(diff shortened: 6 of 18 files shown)


pkg/crc/services/dns/dns.go
```diff
@@ -49,10 +49,10 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
 	}
 
 	// Remove the dnsmasq container if it exists during the VM stop cycle
-	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq")
+	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq-crc")
```
cfergeau, line 54: Maybe `crc-dnsmasq`? must the container name match the systemd service file name? snc/the bundle most likely needs an update?

pkg/crc/services/dns/dns.go
```diff
@@ -51,13 +53,27 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
...
+		// Generate the systemd unit file
+		if _, _, err := serviceConfig.SSHRunner.Run("sudo podman generate systemd --restart-policy=no dnsmasq-crc | sudo tee /etc/systemd/system/dnsmasq-crc.service"); err != nil {
+			return err
+		}
+		if err := sd.DaemonReload(); err != nil {
+			return err
+		}
+		if err := sd.Enable("dnsmasq-crc.service"); err != nil {
+			return err
+		}
```
cfergeau, line 74: Shouldn't this be all done by snc with this commit only removing the code manually starting the container?

praveenkumar: @cfergeau we can create systemd unit file as part of snc but since we update the config (which need instance IP ) as part of crc start so we might need to restart the service after updating the config file. In the snc side we can have dummy file.

pkg/crc/ssh/client.go
```diff
@@ -72,16 +72,14 @@ func clientConfig(user string, keys []string) (*ssh.ClientConfig, error) {
...
-			return nil, err
-		}
+	var err error
+	config, err := clientConfig(client.User, client.Keys)
+	if err != nil {
+		return nil, fmt.Errorf("Error getting config for native Go SSH: %s", err)
+	}
+	client.conn, err = ssh.Dial("tcp", net.JoinHostPort(client.Hostname, strconv.Itoa(client.Port)), config)
+	if err != nil {
+		return nil, err
```
cfergeau, line 82: I guess this reverts `81268bf2 Connect ssh only once, and reuse the same connection`, not clear from the log why the error handling can't be fixed instead of a revert.

pkg/crc/services/dns/dns.go
```diff
@@ -51,13 +53,27 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
...
-	if _, _, err := serviceConfig.SSHRunner.Run(dnsServerRunCmd); err != nil {
-		return err
+	sd := systemd.NewInstanceSystemdCommander(serviceConfig.SSHRunner)
+	if state, err := sd.Status("dnsmasq-crc"); err != nil || state == states.NotFound {
+		// Create the dnsmasq container
+		dnsServerCreateCmd := fmt.Sprintf("sudo podman create  --ip %s --name dnsmasq-crc -v %s:/etc/dnsmasq.conf -p 53:%d/udp --privileged %s",
+			dnsContainerIP, dnsConfigFilePathInInstance, dnsServicePort, dnsContainerImage)
+		// Create the systemd file
+		if _, _, err := serviceConfig.SSHRunner.Run(dnsServerCreateCmd); err != nil {
+			return err
```
cfergeau, line 63: I think we have `RunPrivileged`

pkg/crc/services/dns/dns.go
```diff
@@ -51,13 +53,27 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
...
+		}
+		// Generate the systemd unit file
+		if _, _, err := serviceConfig.SSHRunner.Run("sudo podman generate systemd --restart-policy=no dnsmasq-crc | sudo tee /etc/systemd/system/dnsmasq-crc.service"); err != nil {
```
cfergeau, line 66: Maybe use `serviceConfig.SSHRunner.CopyData()`

pkg/crc/services/dns/dns.go
cfergeau: Looks like the commit log is only telling half of the story then ;) 
From `man systemd.unit` and https://kinvolk.io/docs/flatcar-container-linux/latest/setup/systemd/environment-variables/ I think we can use a `/etc/systemd/system/dnsmasq-crc.service.d/30-ip.conf` file with the content below. Then we'd only need to overwrite that file for each `crc start` and restart the dnsmasq container rather than recreating the whole systemd unit. 

```
$ cat /etc/systemd/system/dnsmasq-crc.service.d/30-ip.conf

[Service]
Environment=DNSMASQ_LISTEN_IP=...
```

praveenkumar: @cfergeau it is not only one place where this IP is needed and listening to but this template https://github.com/code-ready/crc/blob/master/pkg/crc/services/dns/template.go#L11-L21 need to be updated.

cfergeau: During `crc start`, update `/etc/systemd/system/dnsmasq-crc.service.d/30-ip.conf`, update `/var/srv/dnsmasq.conf`, `systemctl restart crc-dnsmasq-container`, is this a problem?

cfergeau: I suspect this could even be done VM-side, with some `ExecPre` in the `crc-dnsmasq-container` unit reading the environment file with the IPs and doing some `sed` in the config file, but probably better to keep code that is known to work for now :)

praveenkumar:
/hold

CI will failed till https://github.com/code-ready/snc/pull/466 is merged and we have bundle using this.

pkg/crc/services/dns/dns.go
```diff
@@ -51,11 +51,6 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
 	// Remove the dnsmasq container if it exists during the VM stop cycle
 	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq")
 
-	// Remove the CNI network definition forcefully
-	// https://github.com/containers/libpod/issues/2767
-	// TODO: We need to revisit it once podman update the CNI plugins.
-	_, _, _ = serviceConfig.SSHRunner.Run(fmt.Sprintf("sudo rm -f /var/lib/cni/networks/podman/%s", dnsContainerIP))
-
```
cfergeau, line 58: The link suggests the CNI update is a 3.2 change, which would make it something fairly new as the current 4.8 bundles only have podman 3.0.1. It's been backported in https://github.com/containers/podman/commit/ce4265af83aca3a30909476b21b8f000a90011e3 aka `v3.0.0-rc3~3^2` so this should be fixed in at least all 4.8 bundles.

pkg/crc/services/dns/dns.go
```diff
@@ -49,10 +49,10 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
 	}
 
 	// Remove the dnsmasq container if it exists during the VM stop cycle
-	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq")
+	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f crc-dnsmasq")
 
 	// Start the dnsmasq container
-	dnsServerRunCmd := fmt.Sprintf("sudo podman run  --ip %s --name dnsmasq -v %s:/etc/dnsmasq.conf -p 53:%d/udp --privileged -d %s",
+	dnsServerRunCmd := fmt.Sprintf("sudo podman run  --ip %s --name crc-dnsmasq -v %s:/etc/dnsmasq.conf -p 53:%d/udp --privileged -d %s",
```
cfergeau: this commit (` Rename dnsmasq container to crc-dnsmasq `)  can be removed (squashed with the next commit) as its code will be gone after next commit.

pkg/crc/services/dns/dns.go
```diff
@@ -49,15 +49,14 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
...
-		return err
+	sd := systemd.NewInstanceSystemdCommander(serviceConfig.SSHRunner)
+	if state, err := sd.Status("crc-dnsmasq"); err != nil || state != states.Running {
+		// Create the dnsmasq container
+		if err := sd.Enable("crc-dnsmasq.service"); err != nil {
+			return err
+		}
 	}
-	return nil
+	return sd.Start("crc-dnsmasq.service")
```
cfergeau, line 59: Looks like this no longer needs the external VM IP, so snc could also take care of `systemd enable`? `crc` would only need to make sure the service is actually running to let the user know something is wrong if it's not running.

pkg/crc/ssh/ssh_test.go
```diff
@@ -88,6 +102,7 @@ func createSSHServer(t *testing.T, listener net.Listener, clientKey *ecdsa.Priva
 
 	go func() {
 		for {
+			fmt.Printf("#goroutines default channel: %d\n", runtime.NumGoroutine())
```
cfergeau, line 105: Not sure this `Printf` and the similar one below is needed?

pkg/crc/ssh/client.go
```diff
@@ -105,6 +106,10 @@ func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
 	session.Stderr = &stderr
 
 	err = session.Run(command)
+	if err != nil {
```
cfergeau, line 109: Can we/do we need to differentiate between "The command that we ran returned an error" and "There was an error in the ssh communication"?

pkg/crc/ssh/client.go
```diff
@@ -105,6 +106,10 @@ func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
...
+		client.stale = true
+		client.conn.Close()
```
cfergeau, line 111: With `client.conn = nil` after this, it looks like `client.stale` is not needed?

pkg/crc/services/dns/dns.go
praveenkumar: @cfergeau if we enable it that means it is going to run for all the platform bundle which we don't want for windows/mac where this is taken care by usermode networking.

pkg/crc/ssh/ssh_test.go
praveenkumar: Ah that was for debug purposes :) , will remove

pkg/crc/ssh/client.go
praveenkumar: @cfergeau we do need to catch specific ssh type error and see if this can be used.

pkg/crc/services/dns/dns.go
cfergeau: Probably worth mentioning in the commit log that we don't want it always enabled because it's unneeded for user mode networking. Is it harmful to have it running when usermode networking is used? Because it's missing a `sd.Stop("crc-dnsmasq")` + `sd.Disable("crc-dnsmasq.service")` when switching from system networking to user networking.

cfergeau: (maybe `crc-dnsmasq.service` and `gvisor-tap-vsock.service` can be made mutually exclusive at the systemd level, but that's just a long term cleanup/improvement, not something required right now).

praveenkumar: > Because it's missing a sd.Stop("crc-dnsmasq") + sd.Disable("crc-dnsmasq.service") when switching from system networking to user networking.

@cfergeau when you  run it using user mode networking then this block not even run so we don't need to stop/disable the service.

pkg/crc/services/dns/dns.go
praveenkumar: Yes it is backported till  v3.0 so all bundles from 4.8 side should work because those are using 3.0.x or higher.

pkg/crc/services/dns/dns.go
cfergeau: We would need to stop/disable if we were reusing the same VM. But I forgot that it's required to recreate the VM when switching from one mode to the other, so that indeed takes care of it.

pkg/crc/ssh/client.go
guillaumerose, line 111: I am curious: what happens for instance when crc start polls for the pull secret file on disk. Does it reconnect ssh for each try?

praveenkumar: @guillaumerose with this patch yer, but I am looking into specific error to filter this out so that we only perform this action when connection reset happen.

cfergeau:
By the way, is this an issue with 4.8.5, ie latest crc release, in which case we'd need to document this in known issues? Or did it start being a problem in later OpenShift releases?

praveenkumar:
> By the way, is this an issue with 4.8.5, ie latest crc release, in which case we'd need to document this in known issues? Or did it start being a problem in later OpenShift releases?

@cfergeau Our CI able to detect this in later version of 4.8, even @adrianriobo tested proxy for 4.8.5 bundle and it works so I think it is problem for later OpenShift release.

pkg/crc/services/dns/dns.go
```diff
@@ -47,17 +47,14 @@ func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
...
-	_, _, _ = serviceConfig.SSHRunner.Run("sudo podman rm -f dnsmasq")
-
-	// Start the dnsmasq container
-	dnsServerRunCmd := fmt.Sprintf("sudo podman run  --ip %s --name dnsmasq -v %s:/etc/dnsmasq.conf -p 53:%d/udp --privileged -d %s",
-		dnsContainerIP, dnsConfigFilePathInInstance, dnsServicePort, dnsContainerImage)
-	if _, _, err := serviceConfig.SSHRunner.Run(dnsServerRunCmd); err != nil {
-		return err
+	sd := systemd.NewInstanceSystemdCommander(serviceConfig.SSHRunner)
+	if state, err := sd.Status("crc-dnsmasq"); err != nil || state != states.Running {
+		// Create the dnsmasq container
```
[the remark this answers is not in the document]
cfergeau: The commit log is missing a description of the concrete problem this is solving. Some proxy-related issue, but the steps to trigger it, and what part of the code triggers the reboot is never explicitly described.

pkg/crc/ssh/ssh_test.go
cfergeau: It still seems to be present?

pkg/crc/ssh/client.go
```diff
@@ -93,6 +93,10 @@ func (client *NativeClient) session() (*ssh.Session, error) {
 func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
 	session, err := client.session()
 	if err != nil {
+		if client.conn != nil {
+			client.conn.Close()
+			client.conn = nil
+		}
```
cfergeau, line 99: What is a good way to test this code path with a crc VM? When is this needed?

cfergeau (commented):
Looks good to me.

pkg/crc/ssh/client.go
praveenkumar: @cfergeau Good way to test this code path with crc VM is during the pull secret update and manually reboot the machine, I think without this patch it will not work at all.

cfergeau: I would put this in the commit log :)

cfergeau: But is this what happens when configuring a proxy? The exact sequence of events which happen with 4.8 when configuring a proxy, and which you are fixing with this PR is very unclear for me

cfergeau:
>  If proxy is set then MCO reboots the system because proxy resource changes
> ...
> - As soon as machine config operator identify there is an update to  proxy resource
> - MCO reboots the node

I haven't seen this happen with git master and a 4.8.5 bundle. I've applied proxy changes, but the node was not automatically rebooted.

praveenkumar:
> > If proxy is set then MCO reboots the system because proxy resource changes
> > ...
> > 
> > * As soon as machine config operator identify there is an update to  proxy resource
> > * MCO reboots the node
> 
> I haven't seen this happen with git master and a 4.8.5 bundle. I've applied proxy changes, but the node was not automatically rebooted. How would I trigger this?

@cfergeau Try it with 4.8.11 bundle you will notice the reboot.

cfergeau:
> @cfergeau Try it with 4.8.11 bundle you will notice the reboot.

Could you add a "steps to reproduce" section to the PR to save time on the back and forth questions trying to find a reproducer? :)

praveenkumar:
> > @cfergeau Try it with 4.8.11 bundle you will notice the reboot.
> 
> Could you add a "steps to reproduce" section to the PR to save time on the back and forth questions trying to find a reproducer? :)

Added

cfergeau:
> > > @cfergeau Try it with 4.8.11 bundle you will notice the reboot.
> > 
> > 
> > Could you add a "steps to reproduce" section to the PR to save time on the back and forth questions trying to find a reproducer? :)
> 
> Added

"Set proxy settings" is not specific enough. Especially when setting a CA cert matters for one of the commits.
And "it fails" is also not a great description. What should I/QE/... be looking for?

pkg/crc/ssh/client.go
```diff
@@ -93,6 +93,11 @@ func (client *NativeClient) session() (*ssh.Session, error) {
 func (client *NativeClient) Run(command string) ([]byte, []byte, error) {
 	session, err := client.session()
 	if err != nil {
+		if client.conn != nil {
```
cfergeau, line 96: This is still resetting the connection even on expected errors (for example busy loops until a given oc command succeeds) as was pointed out by Guillaume.

pkg/crc/ssh/ssh.go
```diff
@@ -90,7 +90,7 @@ func (runner *Runner) runSSHCommand(command string, runPrivate bool) (string, st
 	if err != nil {
 		return string(stdout), string(stderr), fmt.Errorf(`ssh command error:
 command : %s
-err     : %w\n`, command, err)
+err     : %w`+"\n", command, err)
```
cfergeau, line 93: or just add a carriage return since this is already a multi-line string?
```
fmt.Errorf(`ssh command error:
command : %s
err     : %w
`, command, err)
```

Maybe the short log can be reworded to `ssh: Append new line instead of '\n' to error string` ?

go.mod
```diff
@@ -77,4 +77,4 @@ replace (
 	k8s.io/kubectl => github.com/openshift/kubernetes-kubectl v0.0.0-20210521075729-633333dfccda
 )
 
-replace github.com/mdlayher/vsock => github.com/cfergeau/vsock v0.0.0-20210707084117-4d87f8a20ba8
+replace github.com/mdlayher/vsock => github.com/cfergeau/vsock v0.0.0-20210707100525-4def293f672e
```
cfergeau, line 80: Too bad the upstream PR is not merged yet:(

pkg/crc/cluster/status.go
```diff
@@ -17,7 +17,7 @@ func WaitForClusterStable(ctx context.Context, ip string, kubeconfigFilePath str
 	startTime := time.Now()
 
 	retryDuration := 30 * time.Second
-	retryCount := 20 // 10 minutes
+	retryCount := 30 // 15 minutes
```
cfergeau, line 20: 10 minutes is already a very long time to wait for failures, I'd rather not add 5 more minutes to it. Maybe make this conditional on proxy usage?

cmd/crc/cmd/root.go
```diff
@@ -216,7 +216,7 @@ func attachMiddleware(names []string, cmd *cobra.Command) {
 func defaultTransport() *http.Transport {
 	transport := http.DefaultTransport.(*http.Transport)
 	proxyConfig, err := network.NewProxyConfig()
-	if err != nil {
+	if err != nil || !proxyConfig.IsEnabled() {
```
cfergeau, line 219: Can you add a `proxy:` prefix to the shortlog? And maybe `if proxy is disabled` instead of `in case proxy is not enable` to keep it short.
 For what it's worth, I'll rework this code some more in a follow-up PR.

pkg/crc/ssh/client.go
praveenkumar: @cfergeau earlier it used to but with this code shouldn't because now we are tapping the session creation error not the command execution error so it shouldn't reset connection on expected errors for oc commands.

pkg/crc/ssh/ssh.go
praveenkumar: Sure we can do that.

pkg/crc/cluster/status.go
praveenkumar: Done

go.mod
praveenkumar: Nothing we can do here :(

cfergeau (approved):

PR #2702, opened 2021-09-07
Use context to eventually cancel waiting loops

guillaumerose:
When the user hit the stop or delete button while the VM is still starting, it 
allows a fast abort.
The daemon quicky stops the start sequence and can run stop or delete function.

---

I saw this when clicking on start in the tray with an old bundle. I couldnt stop or delete while kubelet was starting.

### DIFF
--- a/pkg/crc/cluster/cert_renewal.go
+++ b/pkg/crc/cluster/cert_renewal.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"context"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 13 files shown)


cfergeau (commented):
Should we just rename `RetryAfterWithContext` to `RetryAfter` and kill the non-context version? (I'd also change the name to something without `After` which I don't understand)

guillaumerose:
Yes done!

cfergeau (approved):

PR #2703, opened 2021-09-07
Increase the resource on circle CI

praveenkumar:
/hold


praveenkumar:
So by default we can have only medium resource not the large one which is part of paid plan.

/close

PR #2706, opened 2021-09-07
Add unit test for the HTTP API

cfergeau:
This adds a unit test for the HTTP API which checks the protocol similarly to what could be done manually with `curl`
This tries to be as extensive as possible even if some things are currently missing (thorough testing for `/api/config`, lack of useful tests for resources using `POST`, ...)
This should already help detect protocol changes which could impact the trays. They are not written in golang, so they can't use the go client library we provide for the API.


cfergeau:
Reverting `a23b12832401e76a21c36c6b1608ce8f589f097f` causes this unit test to fail, which is good as this means it detects the protocol change which caused unexpected issues with the trays.
```
--- FAIL: TestRequests (0.04s)
    api_http_test.go:469: 
                Error Trace:    api_http_test.go:469
                                                        api_http_test.go:479
                Error:          Not equal: 
                                expected: ""
                                actual  : "pull secret exists"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -
                                +pull secret exists
                Test:           TestRequests
                Messages:       GET /pull-secret HTTP/1.1
FAIL
FAIL    github.com/code-ready/crc/pkg/crc/api   0.199s
FAIL
```

guillaumerose (approved):

pkg/crc/api/api_http_test.go
```diff
@@ -0,0 +1,267 @@
...
+	} else {
+		data = nil
+	}
+
+	req := httptest.NewRequest(request.httpMethod, url, data)
+	req.Header.Set("Content-Type", "application/json")
+	{
+		requestDump, _ := httputil.DumpRequest(req, true)
+		fmt.Println(string(requestDump))
+	}
```
praveenkumar, line 73: @cfergeau why we need this in a separate block?

cfergeau: Not sure, iirc I wanted to isolate visually this code during development in case I want to move it someplace else or something like this ^^ Definitely not needed, I can remove the extra `{}`

pkg/crc/api/api_http_test.go
```diff
@@ -1,314 +1,507 @@
...
+func (req request) withBody(data string) request {
+	req.data = data
+	return req
+}
+
+func jSon(data string) response {
+	return response{
+		statusCode: 200,
+		protoMajor: 1,
+		protoMinor: 1,
```
praveenkumar, line 140: @cfergeau what does this protoMajor/protoMinor for? It is the api version scheme?

pkg/crc/api/api_http_test.go
praveenkumar: @cfergeau I figrued, it just break the flow during review (I thought you only need for debug purpose and not need when put the PR). when I run the test cases looks like it is required because the level of info it add to test.

pkg/crc/api/api_http_test.go
cfergeau: It is http protocol version, the `HTTP/1.1` at the end of each requests. Might change to 2.0 some day. 
Reminds me at least one thing which is missing, it is a check for the `Content-Type` header.

pkg/crc/api/api_http_test.go
cfergeau: It may fit better in the main api/ code rather than in tests (enabled by CRC_DEBUG=http-api or such), but I forgot to look into that :)

pkg/crc/api/api_http_test.go
praveenkumar: @cfergeau that we can improve later also.

PR #2713, opened 2021-09-09
preflight: Fix rhel7 detection

cfergeau:
The check added in 'preflight: Don't try to setup daemon socket
activation on el7' works on centos7 but not on rhel7.
While centos7 uses "7" as VersionID for all its releases, rhel7 uses a
different VersionID for each minor release ("7.9" for the current one).

This commit checks for both and should fix https://github.com/code-ready/crc/issues/2653

**Fixes:** Issue #2653

### DIFF
--- a/pkg/crc/preflight/preflight_linux.go
+++ b/pkg/crc/preflight/preflight_linux.go
@@ -326,7 +326,7 @@ func (filter preflightFilter) SetDistro(distro *linux.OsRelease) {
 
 func (filter preflightFilter) SetSystemdUser(distro *linux.OsRelease) {
 	switch {
-	case distroIsLike(distro, linux.RHEL) && distro.VersionID == "7":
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_linux.go
```diff
@@ -326,7 +326,7 @@ func (filter preflightFilter) SetDistro(distro *linux.OsRelease) {
 
 func (filter preflightFilter) SetSystemdUser(distro *linux.OsRelease) {
 	switch {
-	case distroIsLike(distro, linux.RHEL) && distro.VersionID == "7":
+	case distroIsLike(distro, linux.RHEL) && (distro.VersionID == "7" || strings.HasPrefix(distro.VersionID, "7.")):
```
praveenkumar, line 329: 👍🏼

praveenkumar (approved):

PR #2719, opened 2021-09-13
Remove story marketplace from centos CI

praveenkumar:
Centos CI is not green for sometime because test run takes more than
120mins and we don't have a way now to increase the time on CI side
so better to disable some tests on centos CI since same tests are
executing on the openshift ci also.

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -134,7 +134,7 @@ function run_tests() {
   # this is copied over using https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml#L99
   export PULL_SECRET_FILE=--pull-secret-file=$HOME/payload/crc_pull_secret
   export BUNDLE_LOCATION=--bundle-location=$HOME/Downloads/$BUNDLE 
-  make e2e 
+  make e2e GODOG_OPTS="--godog.tags='~@story_marketplace && @linux'"
   if [[ $? -ne 0 ]]; then
     upload_logs $1
     exit 1


cfergeau:
>  same tests are executing on the openshift ci also.

Not clear what this means exactly? Are they run using crc? When are they triggered?
An alternative may be to split our tests in 2, one shorter and mandatory, and one longer test which includes 'less important' tests?

praveenkumar:
> > same tests are executing on the openshift ci also.
> 
> Not clear what this means exactly? Are they run using crc? When are they triggered?

Yes it is our e2e-crc tests https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/code-ready_crc/2719/pull-ci-code-ready-crc-master-e2e-crc/1437285455916699648 which are running for every PR.

> An alternative may be to split our tests in 2, one shorter and mandatory, and one longer test which includes 'less important' tests?

It would be good to have less test cases on the centos CI  if we already running same suite on openshift ci side.

cfergeau:
> > > same tests are executing on the openshift ci also.
> > 
> > 
> > Not clear what this means exactly? Are they run using crc? When are they triggered?
> 
> Yes it is our e2e-crc tests https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/code-ready_crc/2719/pull-ci-code-ready-crc-master-e2e-crc/1437285455916699648 which are running for every PR.

Ah ok, these e2e-crc tests which are started for each crc PR. I was not 100% sure you were referring to this or to something different ^^ The thing is, the e2e tests also tend to get killed because they run for too long. These e2e-crc tests have been at least as reliable as centos CI in the last months, so we can definitely start moving away from centos CI. However in this case both CI seems to have similar issues.

gbraad:
Marketplae and possibly Coolstore are tests that need to run, but perhaps they should be run as part of a nightly (allowing 'infinite time').

jsliacan:
I think I saw @cfergeau's PR somewhere about running all the stories on the same cluster (i.e. *not* stop/start CRC inbetween them). That could be a way to run tests against PRs?

Issue #2687, opened 2021-09-01
[BUG] Daemon not properly initialized after crc setup


adrianriobo:
The daemon process on macos is started from the tray process. The tray is installed before extracting and uncompressing the bundle, so when this long process happens during the setup the daemon is properly initialized. 

When we start crc on a previous environment, and all the commands are included on a script (no delay between setup without require bundle management) the setup ends up successfully but the daemon is not properly initialized, and the start ends up failing.

In this case the setup probably should check the state of the daemon to ack that your system is ready to start crc.

Notice that in this case skipping the update check on the start will produce the error, with the check in place the delay is enough for the daemon.  

To reproduce:

1. Create a pre existing environment 

2. Create a script 
```bash
crc cleanup
crc setup
CRC_DISABLE_UPDATE_CHECK=true crc start -p PULL_SECRET  --log-level debug
```
3. Run the script

Logs from previous script
```bash
INFO Unload CodeReady Containers tray             
INFO Removing launchd configuration for tray      
INFO Removing /etc/resolver/testing file          
INFO Unload CodeReady Containers daemon           
INFO Stopping CRC Hyperkit process                
INFO Removing hosts file records added by CRC     
INFO Removing pull secret from the keyring        
INFO Removing older logs                          
INFO Removing CRC Machine Instance directory      
Cleanup finished
INFO Checking if running as non-root              
INFO Checking if crc-admin-helper executable is cached 
INFO Checking for obsolete admin-helper executable 
INFO Checking if running on a supported CPU architecture 
INFO Checking minimum RAM requirements            
INFO Checking if running emulated on a M1 CPU     
INFO Checking if HyperKit is installed            
INFO Checking if qcow-tool is installed           
INFO Checking if crc-driver-hyperkit is installed 
INFO Checking if CodeReady Containers daemon is running 
INFO Checking if launchd configuration for tray exists 
INFO Creating launchd configuration for tray      
INFO Check if CodeReady Containers tray is running 
INFO Checking if CRC bundle is extracted in '$HOME/.crc' 
INFO Checking if /Applications/CodeReady Containers.app/Contents/Resources/crc_hyperkit_4.8.4.crcbundle exists 
Your system is correctly setup for using CodeReady Containers, you can now run 'crc start' to start the OpenShift cluster
DEBU CodeReady Containers version: 1.31.2+19538dab 
DEBU OpenShift version: 4.8.4 (not embedded in executable) 
DEBU Running 'crc start'                          
[...]

PR #2722, opened 2021-09-13
Issue #2687 Modify tray started check to verify if daemon is running

anjannath:
the daemon might take a few seconds to come up and make the api endpoint available
it is better to check that the daemon is up by trying to access this endpoint  to
make sure user is successfully able to do `crc start` after `crc setup`

### DIFF
--- a/pkg/crc/preflight/preflight_checks_tray_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_tray_darwin.go
@@ -1,11 +1,15 @@
 package preflight
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/daemonclient"
+	crcerrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/os/launchd"
 )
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_checks_tray_darwin.go
```diff
@@ -99,3 +103,12 @@ func fixPlistFileExists(agentConfig launchd.AgentConfig) error {
 	}
 	return nil
 }
+
+func checkIfDaemonRunning() error {
+	return crcerrors.Retry(context.Background(), time.Second*20, func() error {
```
cfergeau, line 108: 20 seconds seem very long?

anjannath: yes, it needs to be changed, in the on macOS tray we only wait 1 second before making `/status` calls, so 1 second interval and the limit we can set to 3? worst case 3 seconds wait time

pkg/crc/preflight/preflight_checks_tray_darwin.go
```diff
@@ -99,3 +103,12 @@ func fixPlistFileExists(agentConfig launchd.AgentConfig) error {
...
 	return nil
 }
+
+func checkIfDaemonRunning() error {
+	return crcerrors.Retry(context.Background(), time.Second*3, func() error {
+		if _, err := daemonclient.New().APIClient.Version(); err != nil {
+			return &crcerrors.RetriableError{Err: fmt.Errorf("Daemon is not yet running: %w", err)}
+		}
+		return nil
+	}, time.Second*1)
```
cfergeau, line 113: the `*1` is not needed :)

cfergeau (approved):

praveenkumar:
Why this is marked as `wip` ?

anjannath:
> Why this is marked as wip ?

Its removed now, the bot took sometime to remove the label :)

PR #2737, opened 2021-09-17
[wip] Issue #2626 Validate pull-secret by trying to login to registries

### LINKED ISSUE #2626: Validate user pull secret by trying to pull image content
Current pull secret validation only validate the expected filed on the json data not the content of it. There might be a case where user already have pull secret downloaded and his/her account is export hold https://access.redhat.com/articles/1340183 in that case user can use that downloaded pull secret to start the crc but not able to pull the images from the registries (like quay, registry.redhat.io ...etc.) 

This issue is to track if crc can improve the pull validation by any means, like extract the registry auth from pull secret and try to login to that registry.

```
$ jq -r '.auths | keys[]' p.yaml
cloud.openshift.com
quay.io
registry.connect.redhat.com
registry.redhat.io
```

anjannath:
Fixes #2626

## Solution/Idea

As mentioned in https://github.com/code-ready/crc/issues/2626#issue-959852675 we extract the registries in the pull secret and try to login to the registries with the provided auth to validate the pull secret.
it uses the github.com/containers/common go package to login to the registry, since we don't always have access to the VM to use podman when we need to validate the pull secret


gbraad:
* How much time does this add to the startup?
  * Can this be a preflight that is skippable?

praveenkumar:
There is some open questions around this PR.

/hold

gbraad:
> There is some open questions around this PR.

what is going to be done about these questions?

anjannath:
>     * How much time does this add to the startup?
> 
It does add more time for startup around ~30 secs

>     * Can this be a preflight that is skippable?

For this we need to get the pull secret earlier, now we do it in machine.Start

cfergeau:
> It does add more time for startup around ~30 secs

Imo that's far too long to be something that we always do when creating a cluster. Maybe something to be done asynchronously? But then the error reporting is not great. It also seems to be called twice during cluster creation, and it's also called when removing the pull secret from the cluster?

anjannath:
> > It does add more time for startup around ~30 secs
> 
> Imo that's far too long to be something that we always do when creating a cluster. Maybe something to be done asynchronously? But then the error reporting is not great. It also seems to be called twice during cluster creation, and it's also called when removing the pull secret from the cluster?

Its checked once during cluster start, basically whenever the pullsecret is loaded from keyring or config this checks is performed, as you suggested it'll be better to change it to only check when the pull secret is stored (for both keyring and crc config).

I don't think this is called also when removing the pull secret from the cluster..

anjannath:
@cfergeau Now the validation check is only performed at the time of storing the pull-secret, that means no validation checks during cluster start if the pull-secret is already present.

cfergeau:
/retest
retest this please

anjannath:
Will reopen once updated and fixed, keeping the PR queue clean :)

Issue #2739, opened 2021-09-20
[BUG] host system reboot needs crc setup again on linux


praveenkumar:
As of now we don't enable the crc-daemon.socket but just start it so in case host system reboot then user need to perform the setup again.

PR #2740, opened 2021-09-20
Enable and start vsock and http crc sockets 

praveenkumar:
**Fixes:** Issue #2739 

## Solution/Idea
This patch enable the http/vsock crc sockets from systemd side.

## Testing
Without this patch (on linux)
- `crc setup`
- Reboot the host
- `crc start` => gives the error 
```
INFO Checking crc daemon systemd socket units
```

With this patch after rebooting the host, user doesn't need to do `setup` again for an already configured crc.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -379,6 +379,9 @@ func fixSystemdUnit(unitName string, unitContent string, shouldBeRunning bool) e
 	running := systemdUnitRunning(sd, unitName)
 	if !running && shouldBeRunning {
 		logging.Debugf("Starting %s", unitName)
+		if err := sd.Enable(unitName); err != nil {
+			return err
+		}
 		return sd.Start(unitName)
 	} else if running && !shouldBeRunning {
 		logging.Debugf("Stopping %s", unitName)
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -290,6 +290,9 @@ Description=CodeReady Containers vsock socket
 [Socket]
 ListenStream=vsock::1024
 Service=crc-daemon.service
+
+[Install]
+WantedBy=default.target
 `
 
 	httpUnitName = "crc-http.socket"
@@ -299,6 +302,9 @@ Description=CodeReady Containers HTTP socket
 [Socket]
 ListenStream=%h/.crc/crc-http.sock
 Service=crc-daemon.service
+
+[Install]
+WantedBy=default.target
 `
 
 	daemonUnitName     = "crc-daemon.service"


pkg/crc/systemd/systemd.go
```diff
@@ -70,7 +70,6 @@ func (c Commander) EnableAndStart(name string) error {
 	return err
 }
 
-
```
cfergeau: This belongs in the previous commit.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -373,7 +373,7 @@ func fixSystemdUnit(unitName string, unitContent string, shouldBeRunning bool) e
 	running := systemdUnitRunning(sd, unitName)
 	if !running && shouldBeRunning {
 		logging.Debugf("Starting %s", unitName)
-		return sd.Start(unitName)
+		return sd.EnableAndStart(unitName)
```
cfergeau, line 376: This will also enable `crc-daemon.service`, but this service is socket activated, so should not be enabled.
Is this working without additional changes? When I looked into this, I had to add this to the unit files I wanted to enable:
```
[Install]
WantedBy=default.target
```

pkg/crc/systemd/actions/action.go
```diff
@@ -23,6 +24,7 @@ func (action Action) String() string {
 		"disable",
 		"status",
 		"daemon-reload",
+		"enable --now",
```
cfergeau, line 27: Not a big fan of pretending `enable --now` is an 'action'. This is leaking at the wrong level that we use actions as command line arguments. 
In theory, `stdOut, stdErr, err = c.commandRunner.Run("systemctl", "enable --now", name)` should be `stdOut, stdErr, err = c.commandRunner.Run("systemctl", "enable", "--now", name)`

pkg/crc/preflight/preflight_checks_linux.go
praveenkumar: @cfergeau yes I added that in one of the commit.

cfergeau: > This will also enable crc-daemon.service, but this service is socket activated, so should not be enabled.

Never mind, this is wrapped in `if shouldBeRunning` which is there for this purpose, to differentiate between the `socket` units and the `service` unit.
One note, on upgrades this won't enable the socket units if they are running already. When the user hits the bug and reruns `setup`, then they will be upgraded as expected.

pkg/crc/systemd/systemd.go
```diff
@@ -65,6 +65,14 @@ func (c Commander) DaemonReload() error {
 	return nil
 }
 
+func (c Commander) EnableAndStart(name string) error {
+	if _, err := c.service(name, actions.Enable); err != nil {
+		return err
+	}
+	_, err := c.service(name, actions.Start)
+	return err
+}
```
cfergeau, line 74: Why? Chaining 2 methods from the public API in a method which will be used once does not feel like something which deserves its own helper

praveenkumar: @cfergeau I wanted to use it for podman systemd unit and libvirtd unit also but looks like in case of libvirtd we don't enable the socket/service. For podman side we we just need to enable it once and run if not running, will drop this one.

gbraad:
Is any of this reverted on 'uninstall' ?

praveenkumar:
> Is any of this reverted on 'uninstall' (cleanup)?

@gbraad on cleanup we remove those unit files (for socket and service).

gbraad (approved):

PR #2748, opened 2021-09-22
Add support for https_proxy=https://...

cfergeau:
At the moment, crc rejects use of `https_proxy=https://...` because it was thought that OpenShift does not support it.
This is not correct, it's only `http_proxy=https://...` which is not supported by OpenShift.
This PR will fix usage of a proxy over https.

## Testing

Before this PR, using `crc config set https-proxy https://...` fails with an error, after this PR, this is allowed, and a cluster can be successfully started using such a proxy.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -142,12 +142,7 @@ func setProxyDefaults() error {
 	}
 
 	if proxyConfig.IsEnabled() {
-		var caFileForDisplay string
-		if proxyCAFile != "" {
-			caFileForDisplay = fmt.Sprintf(", proxyCAFile: %s", proxyCAFile)
-		}
-		logging.Debugf("HTTP-PROXY: %s, HTTPS-PROXY: %s, NO-PROXY: %s%s", proxyConfig.HTTPProxyForDisplay(),
-			proxyConfig.HTTPSProxyForDisplay(), proxyConfig.GetNoProxyString(), caFileForDisplay)
+		logging.Debugf(proxyConfig.String())
 		proxyConfig.ApplyToEnvironment()
 	}
 	return nil
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -2,11 +2,8 @@ package cmd
 
 import (
 	"context"
-	"crypto/tls"
-	"crypto/x509"
 	"errors"
 	"fmt"
-	"net/http"
 	"os"
 	"strings"
 	"time"
@@ -63,7 +60,7 @@ func init() {
 	}
 
 	// Initiate segment client
-	if segmentClient, err = segment.NewClient(config, httpTransport()); err != nil {
+	if segmentClient, err = segment.NewClient(config, network.HTTPTransport()); err != nil {
 		logging.Fatal(err.Error())
 	}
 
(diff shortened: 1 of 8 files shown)


praveenkumar:
retest this please

pkg/crc/network/proxy.go
```diff
@@ -200,9 +200,14 @@ func (p *ProxyConfig) tlsConfig() *tls.Config {
 	if p.ProxyCACert == "" {
 		return nil
 	}
-	caCertPool := x509.NewCertPool()
-	caCertPool.AppendCertsFromPEM([]byte(p.ProxyCACert))
-
+	caCertPool, err := x509.SystemCertPool()
+	if err != nil {
```
praveenkumar, line 210: If we are not sending error to upper layer (caller function) so better to put the error in debug log so we know what caused this error.

pkg/crc/services/dns/dns.go
```diff
@@ -109,7 +109,27 @@ func CheckCRCLocalDNSReachable(ctx context.Context, serviceConfig services.Servi
...
+		}
+		if proxyConfig.HTTPSProxy != "" {
+			curlArgs = append(curlArgs, "--proxy", proxyConfig.HTTPSProxy)
+		}
+		curlArgs = append(curlArgs, "--noproxy", proxyConfig.GetNoProxyString())
+		if proxyConfig.ProxyCAFile != "" {
+			curlArgs = append(curlArgs, "--proxy-cacert", proxyConfig.HTTPProxy)
+		}
+	}
+	stdout, _, err := serviceConfig.SSHRunner.Run("curl", curlArgs...)
```
praveenkumar, line 132: Can we not ignore `stderr` and club it with error because sometime stderr provide some context behind the error?

```
return stdout, fmt.Errorf("stderr:%s, err: %v", stderr, err)
```

pkg/crc/network/proxy.go
praveenkumar: will take of this in some following PRs.

pkg/crc/services/dns/dns.go
praveenkumar: not blocker but will take care of in following PR

cfergeau: I still haven't merged this branch https://github.com/cfergeau/crc/tree/exec-error which would solve this once and for all, but it itches me every time I see the addition of `stderr := ...` :)

Instead of copying the same code everywhere (`stdout, stderr, err := Run(...)` followed by `logging.Warnf("%s: %v", stderr, err)`), we can save the duplication by using a specialized error type: https://github.com/cfergeau/crc/commit/c6e3953afa2e532f96237030e09651ead03d0770

I keep rebasing that branch, cleaning it up a little bit, but still missing testing and the last bit of polish before filing a PR :)

pkg/crc/network/proxy.go
cfergeau: Should be fine now. `proxy.go` did not include `logging`, this is the main/bad reason for not adding logging in the first place.

cmd/crc/cmd/root.go
```diff
@@ -136,40 +132,18 @@ func setProxyDefaults() error {
...
-	proxyConfig, err := network.NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAData)
+	proxyConfig, err := network.NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAFile)
 	if err != nil {
 		return err
 	}
 
 	if proxyConfig.IsEnabled() {
-		logging.Debugf("HTTP-PROXY: %s, HTTPS-PROXY: %s, NO-PROXY: %s, proxyCAFile: %s", proxyConfig.HTTPProxyForDisplay(),
-			proxyConfig.HTTPSProxyForDisplay(), proxyConfig.GetNoProxyString(), proxyCAFile)
+		logging.Debugf(proxyConfig.String())
```
praveenkumar, line 141: 👍🏼

praveenkumar (approved):

PR #2757, opened 2021-09-24
Add https support to our squid image

cfergeau:
We have a squid image which is used for testing. This commit adds support for
accessing squid over https, the bulk of the changes are in the addition of a
script to generate the certificates.

The last commit dumping the root CA when something connects to port 9999 can be
omitted, it's half something useful, half some fun systemd hack I tried while
doing this.

## Solution/Idea

The behaviour of the squid proxy over port 3128 should be unchanged. After this commit squid also listens on port 3129 over https.
This can be used with `crc config set https-proxy https://localhost:3129` and `crc config set proxy-ca-file rootCA.crt`


images/squid/Dockerfile
```diff
@@ -51,6 +51,14 @@ RUN systemctl enable squid.service
...
 RUN usermod -a -G tty squid
 
+# Make rootCA.crt easy to access externally with:
+# nc localhost 9999
+COPY squid-root-ca.socket /etc/systemd/system/
+COPY squid-root-ca@.service /etc/systemd/system/
+RUN systemctl enable squid-root-ca.socket
+
+EXPOSE 9999/tcp
+
```
praveenkumar, line 61: Since this is to get the rootCA and user need to also add `-p 9999:9999` so better to use `podman exec` command.

```
$ podman run -d -p 3128:3128 -p 3129:3129 --name=squid 95605b5c1e47
$ podman exec squid cat /etc/pki/squid/rootCA.crt
-----BEGIN CERTIFICATE-----
[....]
-----END CERTIFICATE-----
```

cfergeau: "better" depends on the usecase, you don't necessarily have easy access to the machine where the podman container is running (which is needed to use `podman exec`), while connecting to port 9999 can be done from anywhere. But I agree this port 9999 thing is just a "convenience hack", I don't mind dropping it.

images/squid/Dockerfile
```diff
@@ -1,22 +1,64 @@
...
+# sudo firewall-cmd  --direct --add-rule ipv4 filter OUTPUT 1 -j ACCEPT
+#
+# To allow traffic on port 3128/3129 from the VM to the host, run this on the host:
+# sudo firewall-cmd --zone=libvirt --add-port=3128/tcp
+# sudo firewall-cmd --zone=libvirt --add-port=3129/tcp
+#
+# After this, running this command in a VM hangs:
+# curl -L -I https://gandi.net
+# and this command succeds:
+# https_proxy=https://192.168.122.1:3129 curl  --proxy-cacert ./rootCA.crt  -L  https://gandi.net
```
praveenkumar, line 29: 👍🏼

praveenkumar (approved):

Issue #2763, opened 2021-09-27
[BUG] Routes are not synced to hosts file


adrianriobo:
Tested on qe environment with pre released version `1.33.0`

After starting a fresh crc installation when deploying a sample application and try to reach the service, it fails  with `The remote name could not be resolved:...` 

Checking the hosts file the route was not added.

Admin-helper service is running and logs from routes-controller reports timeout while trying to reach routes gateway:

```
PS C:\Users\crcqe> oc logs routes-controller -n openshift-ingress
W0927 08:34:25.592870       1 client_config.go:614] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
E0927 08:34:25.604006       1 reflector.go:138] /remote-source/app/main.go:82: Failed to watch *v1.Route: failed to list *v1.Route: the server is currently unable to handle the request (get routes.route.openshift.io)
E0927 08:34:26.889288       1 reflector.go:138] /remote-source/app/main.go:82: Failed to watch *v1.Route: failed to list *v1.Route: the server is currently unable to handle the request (get routes.route.openshift.io)
time="2021-09-27T08:34:30Z" level=info msg="added: oauth-openshift oauth-openshift.apps-crc.testing"
time="2021-09-27T08:34:40Z" level=error msg="Post \"http://gateway/hosts/add\": dial tcp: lookup gateway on 10.217.4.10:53: read udp 10.217.0.47:40083->10.217.4.10:53: i/o timeout"
time="2021-09-27T08:34:40Z" level=info msg="added: console console-openshift-console.apps-crc.testing"
time="2021-09-27T08:34:50Z" level=error msg="Post \"http://gateway/hosts/add\": dial tcp: lookup gateway on 10.217.4.10:53: read udp 10.217.0.47:39025->10.217.4.10:53: i/o timeout"
time="2021-09-27T08:34:50Z" level=info msg="added: downloads downloads-openshift-console.apps-crc.testing"
time="2021-09-27T08:35:00Z" level=error msg="Post \"http://gateway/hosts/add\": dial tcp: lookup gateway on 10.217.4.10:53: read udp 10.217.0.47:40894->10.217.4.10:53: i/o timeout"
time="2021-09-27T08:35:00Z" level=info msg="added: default-route default-route-openshift-image-registry.apps-crc.testing"
[...]

PR #2765, opened 2021-09-27
Revert "Use DHCP to configure DNS with user network mode"

gbraad:
This reverts commit f901599920e94e83a8de04052428543c7332c5b1.

### DIFF
--- a/pkg/crc/services/dns/dns.go
+++ b/pkg/crc/services/dns/dns.go
@@ -25,10 +25,6 @@ func init() {
 }
 
 func RunPostStart(serviceConfig services.ServicePostStartConfig) error {
-	if serviceConfig.NetworkMode == network.UserNetworkingMode {
-		return addOpenShiftHosts(serviceConfig)
-	}
-
 	if err := setupDnsmasq(serviceConfig); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/crc/services/dns/dns.go
```diff
@@ -46,6 +42,10 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) error {
 }
 
 func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {
+	if serviceConfig.NetworkMode == network.UserNetworkingMode {
+		return nil
+	}
+
```
praveenkumar, line 48: 👍🏼

praveenkumar (approved):

cfergeau:
This was reverted because of https://github.com/code-ready/crc/issues/2763
This was not caught before because only CI tests app deployment, and this runs on linux where system networking is the default.
What caused the regression is that the routes-controller tries to connect to `gateway`, which needs to be turned into `gateway.crc.testing` through `resolv.conf` search, but this was removed in `Use DHCP to configure DNS with user network mode`.
Another option would be to get the search domain to be used from the dhcp request.

Issue #2770, opened 2021-09-28
[Proxy] system root pool is not available for windows


praveenkumar:
As per https://golang.org/src/crypto/x509/cert_pool.go looks like there is no implementation of  `SystemCertPool` for windows platform and it just return the error.
```
func SystemCertPool() (*CertPool, error) {
	if runtime.GOOS == "windows" {
		// Issue 16736, 18609:
		return nil, errors.New("crypto/x509: system root pool is not available on Windows")
	}
....
```

This will affect windows user who are behind proxy and using cert file to access https proxy. Current code will log the warning and return the http transport without tls config.

```
func (p *ProxyConfig) HTTPTransport() http.RoundTripper {
...
	transport.Proxy = p.ProxyFunc()
	tlsConfig, err := p.tlsConfig()
	if err != nil {
		logging.Warnf("Failed to add proxy CA to crc http transport")
		return transport
	}
....
```

- https://github.com/golang/go/issues/16736
- https://github.com/golang/go/issues/46287

PR #2771, opened 2021-09-28
proxy: Add fallback to x509.NewCertPool() on Windows

cfergeau:
On Windows, x509.SystemCertPool returns an error:
https://github.com/golang/go/issues/16736

This commit reverts to the behaviour before commit b50dc99 when catching
such an error. This means https_proxy=https://... will be broken for
non-mitm https proxies. Such proxies were not usable before the PR
adding b50dc99, so this should not have much impact for our existing
users.

These CAs are used:
- when accessing telemetry
- when checking for a new crc version
- when downloading binaries (only happens with git builds)

**Fixes:** https://github.com/code-ready/crc/issues/2770

### DIFF
--- a/pkg/crc/network/proxy.go
+++ b/pkg/crc/network/proxy.go
@@ -210,7 +210,8 @@ func (p *ProxyConfig) tlsConfig() (*tls.Config, error) {
 	}
 	caCertPool, err := x509.SystemCertPool()
 	if err != nil {
-		return nil, err
+		logging.Warnf("Could not load system CA pool: %v", err)
+		caCertPool = x509.NewCertPool()
 	}
 	ok := caCertPool.AppendCertsFromPEM([]byte(p.ProxyCACert))
 	if !ok {


pkg/crc/network/proxy.go
```diff
@@ -210,7 +210,8 @@ func (p *ProxyConfig) tlsConfig() (*tls.Config, error) {
 	}
 	caCertPool, err := x509.SystemCertPool()
 	if err != nil {
-		return nil, err
+		logging.Warnf("Could not load system CA pool")
```
praveenkumar, line 213: logging.Warnf("Could not load system CA pool: %v", err)

pkg/crc/network/proxy.go
```diff
@@ -210,7 +210,8 @@ func (p *ProxyConfig) tlsConfig() (*tls.Config, error) {
 	}
 	caCertPool, err := x509.SystemCertPool()
 	if err != nil {
-		return nil, err
+		logging.Warnf("Could not load system CA pool: %v", err)
+		caCertPool = x509.NewCertPool()
```
praveenkumar, line 214: /lgtm

praveenkumar (approved):

Issue #2773, opened 2021-09-28
Add an easy way to create a new application/... in a running cluster


cfergeau:
I realized I almost never test application creation/... during development.
The story_marketplace and story_registry would be useful to run during development, but they start/stop the cluster, making them not very convenient to use.
In my opinion it would be beneficial to have a way to run them against a running cluster, this would provide an easy way for some testing of the cluster functionality.

praveenkumar:
Putting the command here also in case I forget :) 
```
$ oc new-app <sample_app> && oc rollout status <sample_app> && oc get routes -ojsonpath='{.items[].spec.host}' | xargs curl -Ik 
```

PR #2776, opened 2021-09-29
e2e: Add 'e2e-stories' Makefile target

cfergeau:
`make e2e-stories` is meant to be run against an already running cluster, and it won't run `crc start` or `crc stop` which are very slow operations.
`e2e-stories` only takes a few minutes to run on my test machine, with `e2e-story-health` and `e2e-story-registry` only taking less than 20 seconds each. This is useful for quick sanity checks during development. 
There are also individual `e2e-story-health`, `e2e-story-marketplace` and `e2e-story-registry` to run each of these stories separately.

This introduces a @startstop godog tag which is used to disable the parts of theses tests which would start or stop the CRC VM.
I could not come up with a more elegant way of dealing with this.

This fixes https://github.com/code-ready/crc/issues/2773


Makefile
```diff
@@ -212,6 +212,18 @@ endif
...
+.PHONY: e2e-stories e2e-story-health e2e-story-marketplace e2e-story-registry
+# cluster must already be running, crc must be in the path
+e2e-stories: install e2e-story-health e2e-story-marketplace e2e-story-registry
+
+e2e-story-health: install
+	@go test --timeout=180m $(REPOPATH)/test/e2e -v --godog.tags="$(GOOS) && ~@startstop && @story_health" --cleanup-home=false
+e2e-story-marketplace: install
+	@go test --timeout=180m $(REPOPATH)/test/e2e -v --godog.tags="$(GOOS) && ~@startstop && @story_marketplace" --cleanup-home=false
+e2e-story-registry: install
+	@go test --timeout=180m $(REPOPATH)/test/e2e -v --godog.tags="$(GOOS) && ~@startstop && @story_registry" --cleanup-home=false
```
praveenkumar, line 224: `180m` timeout for this? Also since this is run by dev and not going to part of CI so we can remove the timeout completely.

cfergeau: ah this is just c&p, I'll remove

PR #2779, opened 2021-09-30
Cleanup use of `CrcBinDir` vs `BinDir()`

cfergeau:
This PR is just some code cleanup which should have no user-visible impact.
The main goal is to improve a bit the cache code, and to remove an installer-specific hack from the embed package.


pkg/embed/embed.go
```diff
@@ -7,9 +7,14 @@ import (
 
 	"github.com/YourFin/binappend"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/version"
 )
 
 func openEmbeddedFile(executablePath, embedName string) (io.ReadCloser, error) {
+	if version.IsInstaller() {
+		panic("should no longer happen")
+	}
```
praveenkumar, line 16: @cfergeau do we really need those panic , shouldn't we just send error back to caller function.

cfergeau: This `panic` commit was not meant to be pushed to this PR, dunno how it ended up there :)

anjannath (approved):

PR #2792, opened 2021-10-05
[WIP] Decouple bundle

praveenkumar:


pkg/crc/machine/bundle/metadata.go
```diff
@@ -217,3 +219,40 @@ func GetCustomBundleName(bundleFilename string) string {
...
+			} `json:"darwin"`
+			Windows struct {
+				URL    string `json:"url"`
+				Sha256 string `json:"sha256"`
+			} `json:"windows"`
+		} `json:"amd64"`
+	} `json:"openshift"`
+}
+
+func GetBundleURLAndSha256sum(data []byte) (string, string, error) {
```
cfergeau, line 242: With the comment above, this can become `GetBundleDownloadDetails()` to avoid exposing url and sha256sum too much.
`data []byte` in the args is not very nice. What should I pass to it? I'd expect to either be able to pass it the parsed data, or to not pass it any args, and it would be doing everything by itself.

pkg/embed/embed.go
```diff
@@ -38,6 +39,25 @@ func Extract(embedName, destFile string) error {
 	return ExtractFromExecutable(executablePath, embedName, destFile)
 }
 
+func ReadEmbededFile(embedName string) ([]byte, error) {
```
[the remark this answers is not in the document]
cfergeau: Have you considered returning an `io.Reader`? This is more usual in go APIs, and unmarshalling the json is not much more code: `json.NewDecoder(reader).Decode(&m)`

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -88,14 +89,23 @@ func fixBundleExtracted() error {
...
+		if _, err := os.Stat(constants.DefaultBundlePath); os.IsNotExist(err) {
+			data, err := embed.ReadEmbededFile("bundle-info.json")
+			if err != nil {
+				return err
+			}
+			url, sha256sum, err := bundle.GetBundleURLAndSha256sum(data)
+			if err != nil {
+				return err
+			}
+			if _, err := download.Download(url, constants.DefaultBundlePath, 0664, []byte(sha256sum)); err != nil {
```
cfergeau, line 103: I'd go with `0444`, should be at most `0644`

cfergeau, line 103: `GetBundleURLAndsha256sum` could return a []byte directly to avoid the cast. A bit worried there's going to be some mismatch between `"0123ABCD"` read as a string from the json file, and `[]uint8{0x01, 0x23, 0xab, 0xcd}` expected by the download code?

centos_ci.sh
```diff
@@ -158,7 +159,7 @@ else
 
 	# setup to run e2e tests
 	cd payload
-	make
+	make release
```
cfergeau, line 162: Should we be making releases by default? Is there any differences left between release/non-release bundles?

centos_ci.sh
```diff
@@ -133,7 +133,8 @@ function run_tests() {
   # In Jenkins slave we have pull secret file in the $HOME/payload/crc_pull_secret
   # this is copied over using https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml#L99
   export PULL_SECRET_FILE=--pull-secret-file=$HOME/payload/crc_pull_secret
-  export BUNDLE_LOCATION=--bundle-location=$HOME/Downloads/$BUNDLE 
+  export BUNDLE_LOCATION=--bundle-location=$HOME/Downloads/$BUNDLE
```
cfergeau, line 136: > Since bundle is not embeded so it is easy to do make release and test the release bits instead intermidate one.

I'd change the log to something like this:

-> Since the bundle is no longer embedded, `make release` can be used so that the CI tests builds similar to the ones which would get released

cfergeau:
One thing I forgot to ask, if I already have a downloaded bundle, how can I get `crc` to use it rather than downloading the bundle? I don't think `copy it to ~/.crc/cache` is a great answer, as it's not discoverable at all.

cfergeau:
I wonder if introducing an intermediate json file is really needed. If we set the values in go code, QE could directly change these in the code, it's not much more difficultt than changing a json file? See https://github.com/cfergeau/crc/commit/99d5e583f8586301788b3fee1add8e9d79806d53 for example:
```golang
pkg/crc/machine/bundle/constants.go:

package bundle

import (
	"github.com/code-ready/crc/pkg/download"
)

var bundleLocations = map[string]bundlesDownloadInfo{
	"amd64": bundlesDownloadInfo{
		"darwin": download.NewRemoteFile("", ""),
		"linux": download.NewRemoteFile(
			"https://storage.googleapis.com/crc-bundle-github-ci/4.8.12/crc_libvirt_4.8.12.crcbundle",
			"6283a7037c050494f1c06f78b06d2a54b133dc73941cc929d66bbee877ee3254",
		),
		"windows": download.NewRemoteFile("", ""),
	},
}
```

praveenkumar:
> I wonder if introducing an intermediate json file is really needed. If we set the values in go code, QE could directly change these in the code, it's not much more difficultt than changing a json file? See [cfergeau@99d5e58](https://github.com/cfergeau/crc/commit/99d5e583f8586301788b3fee1add8e9d79806d53) for example:
> 
> ```go
> pkg/crc/machine/bundle/constants.go:
> 
> package bundle
> 
> import (
> 	"github.com/code-ready/crc/pkg/download"
> )
> 
> var bundleLocations = map[string]bundlesDownloadInfo{
> 	"amd64": bundlesDownloadInfo{
> 		"darwin": download.NewRemoteFile("", ""),
> 		"linux": download.NewRemoteFile(
> 			"https://storage.googleapis.com/crc-bundle-github-ci/4.8.12/crc_libvirt_4.8.12.crcbundle",
> 			"6283a7037c050494f1c06f78b06d2a54b133dc73941cc929d66bbee877ee3254",
> 		),
> 		"windows": download.NewRemoteFile("", ""),
> 	},
> }
> ```

@cfergeau my plan is to have a hack script to which you provide the bundle download location and it automatic create this json file for us. Another reason is keeping separate is in long term I will not going to use `pkg/embed` but directly use the go embed functionality which make this whole embed bit more cleaner.

PR #2796, opened 2021-10-07
Bump OpenShift to 4.8.13 (stable)

cfergeau:


cfergeau:
Won't have time to properly finalize this, and probably outdated by now ;)

Issue #2723, opened 2021-09-13
[BUG] Don't send telemetry identify event with every track event


deboer-tim:
It looks like CRC is sending an identify event with each track event (or at some frequent interval - I can see 9 from one user in the top 25 events at the time I looked). This uses up our segment allocation twice as fast and is unnecessary. The identify event only needs to be sent once, and then again only if something (e.g. os version) changes. For instance, a hash of the event could be cached in a file and only sent if it changes, or at least only send once every couple weeks.

cfergeau:
https://segment.com/docs/connections/spec/identify/ has some info about when to send `identify`

cfergeau:
At the moment, we are indeed sending `identify` and `track` together:
https://github.com/code-ready/crc/blob/320c4d37196804e92f2ebf80a5583cb322b36d98/pkg/crc/segment/segment.go#L73-L96

I guess we could send `identify` once at `crc` daemon startup,

cfergeau:
https://github.com/redhat-developer/vscode-redhat-telemetry/commit/47e9015e20f2b4f374e0239747df545a5997f849 did something similar

cfergeau:
I've started working on this in https://github.com/cfergeau/crc/tree/segment
With these changes, `identify` will only be sent once each time the crc process is started. On macOS and Windows, this means it will be sent once per user session since we have a long running daemon on these platforms. On linux, this means it will be sent once every time the user runs `crc ...`.

Is this good enough? Or do we have to implement on disk caching of this to only send `identify` when any of its components change?

deboer-tim:
The goal was to reduce the events sent down from 2x to 1x. Ideally we'd probably disk cache and go all the way, but if you think based on usage patterns this would get it down to (say) 1.2x with a smaller/safer change then I'm ok with this - we can always confirm the reduction via telemetry and see if it is worth making a further change in a couple months.

PR #2802, opened 2021-10-14
telemetry: Don't send 'identify' events too often

cfergeau:
This calculates a hash of the content of the identify event, and only resend such identify events when something changes.
It's apparently harmful (read 'costs money') to send too many identify events.
The hash is stored in `~/.crc` so on a typical system, identify events will be sent once after this PR, and never again (until ~/.crc is removed).

**Fixes:** https://github.com/code-ready/crc/issues/2723

## Testing

There's a debug log line printed when an identify event is sent, so this can be tested by changing `enable-experimental-features` or by changing `network-mode`.
The user-id can also be changed by removing it from `~/.redhat`, even if I did not test this. Changing it would also trigger the sending of an identify event.

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -2,6 +2,7 @@ package segment
 
 import (
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"hash/fnv"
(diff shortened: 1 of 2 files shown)


praveenkumar:
Tested with segment debugger and caching on disk does work and we are not sending the identity now more than once as long as it doesn't change.

pkg/crc/segment/segment.go
```diff
@@ -70,25 +91,93 @@ func (c *Client) UploadCmd(ctx context.Context, action string, duration time.Dur
...
+
+	cachedHashBytes, err := ioutil.ReadFile(identifyHashPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return 0, nil
+		}
+		return 0, err
+	}
+
+	return binary.LittleEndian.Uint64(cachedHashBytes), nil
```
gbraad, line 142: This fails when the file is empty

PR #2824, opened 2021-11-01
Add podman bundle detection and update start command template

praveenkumar:


anjannath:
Few log messages are not modified according to the preset set, apart from that LGTM 👍🏽 

- During `crc start` it still says about Genereting new kubeadmin password, which is not required in the podman preset. 
<details>
<summary>crc start logs</summary>

```PS> .\crc.exe start -b C:\Users\anath\Downloads\Persepolis\Others\crc_podman_hyperv_3.3.1.crcbundle
INFO Checking Windows edition
INFO Checking if the Hyper-V virtual switch exists
INFO Loading bundle: crc_podman_hyperv_3.3.1...
INFO Creating CodeReady Containers VM for Podman 3.3.1...
INFO Generating new SSH Key pair...
INFO Generating new password for the kubeadmin user -- not required for podman preset?
INFO CodeReady Containers instance is running with IP 172.29.218.251
Enabled the podman socket.

  PS> & crc podman-env | Invoke-Expression
```
</details>

- During `crc stop` its trying to kill the kubelet process
<details>
<summary>crc stop logs</summary>

```PS> .\crc.exe stop
INFO Stopping kubelet and all containers...
WARN Executing systemctl action failed:  ssh command error:
command : sudo systemctl stop kubelet
err     : Process exited with status 5
: Failed to stop kubelet.service: Unit kubelet.service not loaded.
INFO Stopping the OpenShift cluster, this may take a few minutes...
Stopped the OpenShift cluster
```
</details>

- During `crc setup` the finale message talks about OpenShift instead of Podman
<details>
<summary> crc setup logs</summary>

```PS> .\crc.exe setup -b C:\Users\anath\Downloads\Persepolis\Others\crc_podman_hyperv_3.3.1.crcbundle
INFO Using bundle path C:\Users\anath\Downloads\Persepolis\Others\crc_podman_hyperv_3.3.1.crcbundle
INFO Checking if C:\Users\anath\Downloads\Persepolis\Others\crc_podman_hyperv_3.3.1.crcbundle exists
INFO Ensuring directory C:\Users\anath\.crc\cache exists
INFO Uncompressing C:\Users\anath\Downloads\Persepolis\Others\crc_podman_hyperv_3.3.1.crcbundle
crc-podman.vhdx: 2.15 GiB / 2.15 GiB [---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
Your system is correctly setup for using CodeReady Containers, you can now run 'crc start -b $bundlename' to start the OpenShift cluster
```
</details>

anjannath (approved):

pkg/crc/machine/bundle/metadata.go
```diff
@@ -15,6 +15,13 @@ import (
 	"github.com/code-ready/crc/pkg/download"
 )
 
+type Type string
+
+const (
+	Openshift Type = "snc"
```
cfergeau, line 21: Maybe `OpenShift` rather than `Openshift`?

pkg/crc/machine/bundle/repository.go
```diff
@@ -82,8 +85,10 @@ func (repo *Repository) Use(bundleName string) (*CrcBundleInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := bundleInfo.createSymlinkOrCopyOpenShiftClient(repo.OcBinDir); err != nil {
-		return nil, err
+	if bundleInfo.GetBundleType() == Openshift {
+		if err := bundleInfo.createSymlinkOrCopyOpenShiftClient(repo.OcBinDir); err != nil {
+			return nil, err
+		}
 	}
```
cfergeau, line 92: Could this be done with 2 distinct types, PodmanBundle and OpenShiftBundle implementing the same interface, with OpenShiftBundle adding a few more checks in addition to the ones done by the PodmanBundle? Imo the less `if bundleInfo.GetBundleType() ==...` we have, the better.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -173,6 +174,10 @@ func (bundle *CrcBundleInfo) GetOpenshiftVersion() string {
 	return bundle.ClusterInfo.OpenShiftVersion.String()
 }
 
+func (bundle *CrcBundleInfo) GetPodmanVersion() string {
+	return bundle.Nodes[0].PodmanVersion
```
cfergeau, line 178: A bit surprising to see `PodmanVersion` depending on the node while `OpenShiftVersion` is part of ClusterInfo?

pkg/crc/machine/stop.go
```diff
@@ -19,7 +19,7 @@ func (client *client) Stop() (state.State, error) {
 		return state.Error, errors.Wrap(err, "Cannot load machine")
 	}
 	if err := stopAllContainers(host, client); err != nil {
-		return state.Error, err
+		logging.Warnf("%v", err)
```
cfergeau, line 22: This is not going to be a user-friendly message, since it's a warning, a nice message would be better, with the error being printed to debug.

pkg/crc/machine/bundle/metadata.go
praveenkumar: @cfergeau As of now we have single node cluster so podman version can be part of node/cluster but the way our bundle info sample json file is makes it bit confusing when it comes to openshift version because we still have node as list instead having it a single one so in future (which may never happen) when worker and master are part of different node then openshift version still make sense to part of cluster info but podman version is node specific.

pkg/crc/machine/stop.go
praveenkumar: @cfergeau yes, I agree any specific wording you have in mind?

pkg/crc/machine/bundle/repository.go
praveenkumar: @cfergeau We can have bundleInfo as interface and have OpenShift/Podman bundle struct which can satisfy this interface but it depend on how much refactor we want to do at this moment since `dev` branch is currently experimental and once we have have our podman/openshift POC ready then all refactor can follow.

praveenkumar: We can have condition simple something like `bundleInfo.IsPodman` which just return bool and then instead of using `bundleInfo.GetBundleType() == Openshift` we can just use that helper.

cfergeau: >  since dev branch is currently experimental and once we have have our podman/openshift POC ready then all refactor can follow

Does this mean once the POC is working, we rebase/force push to `dev` in order to get in a nice state before merging in `master`? Or that we merge the POC in `master`, and then we clean things up? The crc `dev` branch is quite new to me ;)

cfergeau: > We can have condition simple something like bundleInfo.IsPodman which just return bool and then instead of using bundleInfo.GetBundleType() == Openshift we can just use that helper.

Yup, probably better than the comparisons, even if it's even better if we can also get rid of the helper ;)

pkg/crc/machine/bundle/metadata.go
cfergeau: > when worker and master are part of different node then openshift version still make sense to part of cluster info but podman version is node specific.

Not sure we want to support this without a concrete usecase. If I'm not mistaken, it's also possible to have worker/master nodes with different openshift versions. I'd keep things simple and consistent, and have a single podman version for the whole cluster. When there are multiple nodes, and we need to support different podman versions for each node, then we can revisit this.

pkg/crc/machine/stop.go
```diff
@@ -19,7 +19,7 @@ func (client *client) Stop() (state.State, error) {
 		return state.Error, errors.Wrap(err, "Cannot load machine")
 	}
 	if err := stopAllContainers(host, client); err != nil {
-		return state.Error, err
+		logging.Warnf("Failed to stop all containers: %v\nProceeding with stop...", err)
```
cfergeau, line 22: maybe `Failed to stop OpenShift containers` (or cluster containers). `Proceeding with VM shutdown`, this is more accurate about what is being stopped, and what these containers are. I'd favour only having `logging.Debugf("%v", err)` for the error printing, and not include it in the `logging.Warnf`.

pkg/crc/constants/constants.go
```diff
@@ -155,3 +153,17 @@ func GetCRCMacTrayDownloadURL() string {
...
+	}
+	return 2
+}
+
+func GetDefaultMemory(openshift bool) int {
+	if openshift {
+		return 9216
+	}
+	return 2048
+}
```
cfergeau, line 169: This change also feels like it should be done the opposite way, first instantiate a `PresetDefaults`, and then call `presetDefaults.GetMemory()` and such?

pkg/crc/preflight/preflight.go
```diff
@@ -18,6 +18,10 @@ const (
 	StartUpOnly
 )
 
+var (
+	preset string
+)
+
```
cfergeau, line 24: We've avoided global variables in the `preflight` package so far, would be better to keep it this way.

pkg/crc/machine/status.go
```diff
@@ -44,11 +44,16 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
...
-		}, nil
+		clusterStatusResult := &types.ClusterStatusResult{
+			CrcStatus: state.FromMachine(vmStatus),
+		}
+		if crcBundleMetadata.GetBundleType() == bundle.OpenShift {
+			clusterStatusResult.OpenshiftStatus = types.OpenshiftStopped
+			clusterStatusResult.OpenshiftVersion = crcBundleMetadata.GetOpenshiftVersion()
+		} else {
+			clusterStatusResult.PodmanVersion = crcBundleMetadata.GetPodmanVersion()
+		}
```
cfergeau, line 55: If we add more bundle types, this will quickly become messy, I've tried to address this differently in https://github.com/cfergeau/crc/commits/machine-refactor2

pkg/crc/machine/stop.go
cfergeau: This comment seems to have been missed before this PR got merged.

cmd/crc/cmd/start.go
```diff
@@ -243,6 +244,12 @@ Use the 'oc' command line interface:
   {{ .CommandLinePrefix }} {{ .EvalCommandLine }}
   {{ .CommandLinePrefix }} oc login -u {{ .ClusterConfig.DeveloperCredentials.Username }} {{ .ClusterConfig.URL }}
 `
+	startTemplateForPodman = `Enabled the podman socket.
```
cfergeau, line 248: I don't think this is a useful message for the user. What does this mean? What matters is that there's a podman runtime (dunno about the correct name) that is available and ready to interact with.

cmd/crc/cmd/start.go
```diff
@@ -251,18 +258,32 @@ type templateVariables struct {
...
 func writeTemplatedMessage(writer io.Writer, s *startResult) error {
-	parsed, err := template.New("template").Parse(startTemplate)
+	parsed, err := template.New("template").Parse(startTemplateForPodman)
 	if err != nil {
 		return err
 	}
-
+	// This should be replaced with the config preset once
+	// we start using preset as part of config.
+	if s.ClusterConfig.ClusterCACert != "" {
```
cfergeau, line 268: I think this is added later in the PR? Why wasn't this commit moved after the introduction of presets or at least updated after presets are added?

pkg/crc/machine/bundle/metadata.go
cfergeau: Still pending..

pkg/crc/preflight/preflight_linux.go
```diff
@@ -351,26 +351,26 @@ func getAllPreflightChecks() []Check {
 	filter.SetDistro(distro())
 	filter.SetSystemdUser(distro())
 
-	return filter.Apply(getChecks(distro(), ""))
+	return filter.Apply(getChecks(distro(), "", ""))
 }
 
-func getPreflightChecks(_ bool, _ bool, networkMode network.Mode, bundlePath string) []Check {
+func getPreflightChecks(_ bool, _ bool, networkMode network.Mode, bundlePath, preset string) []Check {
```
cfergeau, line 357: This method already has too many args, adding more should be avoided :-/

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -58,7 +58,7 @@ var genericCleanupChecks = []Check{
 	},
 }
 
-func checkBundleExtracted(bundlePath string) func() error {
+func checkBundleExtracted(bundlePath, preset string) func() error {
```
cfergeau, line 61: nothing seems to be making use of this new arg?

pkg/crc/machine/start.go
```diff
@@ -719,3 +723,13 @@ func updateKubeconfig(ctx context.Context, ocConfig oc.Config, sshRunner *crcssh
 	}
 	return nil
 }
+
+func bundleMismatchWithPreset(preset string, bundleMetadata *bundle.CrcBundleInfo) error {
+	if preset != string(crcPreset.OpenShift) && bundleMetadata.IsOpenShift() {
+		return errors.Errorf("Preset %s is used but bundle is provided for %s preset", crcPreset.Podman, crcPreset.OpenShift)
+	}
+	if preset != string(crcPreset.Podman) && !bundleMetadata.IsOpenShift() {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau we can but we shouldn't because our openshift bundle is based on RHCOS which have older version of podman and podman bundle is based on FCOS so better to differ those use cases.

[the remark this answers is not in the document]
cfergeau: Actually, it might still be worth to explore, last time we discussed this, it was not clear if there were any known issues for our use cases, so even if the podman version is slightly old, it might still do the trick.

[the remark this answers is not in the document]
gbraad: Continue conversation on: https://github.com/code-ready/crc/issues/2867

For now we tie this choice to a bundle, as also on the welcome screen it will be a choice to use Podman or OpenShift

PR #2828, opened 2021-11-04
Improve usage hint of crc podman-env --root

matskiv:
## Solution/Idea
When `crc podman-env --root` is executed the "hint" should contain `--root` too:
```
...
# eval $(crc podman-env --root)
```

## Testing
Running `crc podman-env --root` should print `# eval $(crc podman-env --root)` on the last line of the output.

### DIFF
--- a/cmd/crc/cmd/podman_env.go
+++ b/cmd/crc/cmd/podman_env.go
@@ -49,7 +49,11 @@ func runPodmanEnv() error {
 			connectionDetails.IP,
 			connectionDetails.SSHPort,
 			socket)))
-	fmt.Println(shell.GenerateUsageHintWithComment(userShell, "crc podman-env"))
+	cmdLine := "crc podman-env"
+	if root {
+		cmdLine += " --root"
+	}
+	fmt.Println(shell.GenerateUsageHintWithComment(userShell, cmdLine))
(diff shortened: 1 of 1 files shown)


gbraad (approved):

gbraad:
@praveenkumar this change is small enough to avoid the use of the development branch?

cfergeau:
> @praveenkumar this change is small enough to avoid the use of the development branch?

This can be seen as a bug fix, and yes, it's small, it belongs to `master` in my opinion.

cmd/crc/cmd/podman_env.go
```diff
@@ -49,7 +49,11 @@ func runPodmanEnv() error {
 			connectionDetails.IP,
 			connectionDetails.SSHPort,
 			socket)))
-	fmt.Println(shell.GenerateUsageHintWithComment(userShell, "crc podman-env"))
+	cmdLine := "crc podman-env"
+	if root {
+		cmdLine += " --root"
+	}
+	fmt.Println(shell.GenerateUsageHintWithComment(userShell, cmdLine))
```
praveenkumar, line 56: 👍🏼

praveenkumar (approved):

PR #2832, opened 2021-11-09
dev: Various improvements to bundle downloading

cfergeau:
The initial goal was to use the same progress bar as the archive extraction
code, which this PR does. But it also tries to improve the decisions around
when to download the bundle (do it in !release builds).


praveenkumar:
Rebasing and testing it for dev branch.

PR #2840, opened 2021-11-16
Use the 'Preset' type more extensively, and rework the preset-related code in config and prefligts

cfergeau:
My initial goal was to do some cleanups in preflight, there was both a 'preset' global variable, and a 'preset' argument passed to several methods which is never used.
I ended up doing more changes along the way. There should be no behaviour change after this PR.


praveenkumar:
These changes look good to me. We can merge this as soon CI passes.

/lgtm

PR #2845, opened 2021-11-22
Detect default bundle according to preset

praveenkumar:


cmd/crc/cmd/setup.go
```diff
@@ -22,7 +22,7 @@ var (
 
 func init() {
 	setupCmd.Flags().Bool(crcConfig.ExperimentalFeatures, false, "Allow the use of experimental features")
-	setupCmd.Flags().StringP(crcConfig.Bundle, "b", constants.DefaultBundlePath(), "Bundle to use for VM")
+	setupCmd.Flags().StringP(crcConfig.Bundle, "b", constants.GetDefaultBundlePath(), "Bundle to use for VM")
```
cfergeau: I think this can be merged with the previous commit, I expect they are changing exactly the same lines of code, and the intermediate `DefaultBundlePath()` state is not really useful to have.

praveenkumar: @cfergeau done.

cfergeau: The commit log would deserve slightly more rewording, it's really not important to explicitly say that it was renamed and turned into a function. Just `constanst: Change DefaultBundlePath to GetDefaultBundlePath()` as the shortlog does the trick, together with a longer explanation in the long log

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -90,15 +91,15 @@ func fixBundleExtracted(bundlePath string) func() error {
 			if !errors.As(err, &e) {
 				return err
 			}
-			if bundlePath != constants.GetDefaultBundlePath() {
+			if bundlePath != constants.GetDefaultBundlePath(preset.OpenShift) {
```
cfergeau: Not clear why this is hardcoded to `preset.OpenShift`?

pkg/crc/validation/validation.go
```diff
@@ -61,11 +61,11 @@ func ValidateBundlePath(bundlePath string) error {
 	}
 
 	userProvidedBundle := filepath.Base(bundlePath)
-	if userProvidedBundle != constants.GetDefaultBundle() {
+	if userProvidedBundle != constants.GetDefaultBundle(crcpreset.OpenShift) {
```
cfergeau, line 64: Some more hardcoding of OpenShift, also not clear why.

pkg/crc/constants/constants.go
```diff
@@ -61,24 +61,31 @@ func GetAdminHelperURL() string {
...
 
-func GetDefaultBundleForOs(os string) string {
-	return GetBundleFosOs(os, version.GetBundleVersion())
+func GetDefaultBundleForOs(os string, preset crcpreset.Preset) string {
+	return GetBundleFosOs(os, version.GetBundleVersion(), version.GetPodmanVersion(), preset)
 }
 
-func GetBundleFosOs(os, bundleVersion string) string {
-	return defaultBundleForOs(bundleVersion)[os]
+func GetBundleFosOs(os, bundleVersion, podmanVersion string, preset crcpreset.Preset) string {
```
cfergeau, line 83: Can you fix the `FosOs` typo while you are changing this code?

pkg/crc/constants/constants.go
```diff
@@ -61,24 +61,31 @@ func GetAdminHelperURL() string {
 	return GetAdminHelperURLForOs(runtime.GOOS)
 }
 
-func defaultBundleForOs(bundleVersion string) map[string]string {
+func defaultBundleForOs(bundleVersion, podmanVersion string, preset crcpreset.Preset) map[string]string {
```
cfergeau, line 64: I don't think you need to pass `bundleVersion`/`podmanVersion`, the preset is going to be enough since you have a switch on the preset value anyway.
Have you explored adding a
```
type PresetConstants  interface {
    GetDefaultBundle() string
    GetDefaultBundlePath() string
    GetMemory() int
    GetCPUs() int
}

func newPresetConstants(preset preset.Preset) PresetConstants
```
I wonder if this would help makes things slightly cleaner by not having to pass a 'preset' arg to everything, and by avoiding some `if podman ... else ...`. But maybe this won't make much difference, I haven't tried it :)

pkg/crc/constants/constants.go
```diff
@@ -61,24 +61,31 @@ func GetAdminHelperURL() string {
...
+	if preset == crcpreset.Podman {
+		return map[string]string{
+			"darwin":  fmt.Sprintf("crc_podman_hyperkit_%s.crcbundle", podmanVersion),
+			"linux":   fmt.Sprintf("crc_podman_libvirt_%s.crcbundle", podmanVersion),
+			"windows": fmt.Sprintf("crc_podman_hyperv_%s.crcbundle", podmanVersion),
```
cfergeau, line 69: Wondering how redundant this is with the table we now have with the bundle URLs?

pkg/crc/machine/bundle/metadata.go
```diff
@@ -256,12 +256,12 @@ func getBundleDownloadInfo() (*download.RemoteFile, error) {
...
 
-func Download() error {
-	downloadInfo, err := getBundleDownloadInfo()
+func Download(preset preset.Preset) error {
+	downloadInfo, err := getBundleDownloadInfo(preset)
 	if err != nil {
 		return err
 	}
-	if _, err := downloadInfo.Download(constants.GetDefaultBundlePath(preset.OpenShift), 0664); err != nil {
+	if _, err := downloadInfo.Download(constants.GetDefaultBundlePath(preset), 0664); err != nil {
```
cfergeau, line 269: Looks like this could be merged with the previous commit? This does not seem to add a lot more new lines, it's mostly changing lines which were already changed in the previous patch.

pkg/crc/machine/bundle/constants.go
```diff
@@ -6,11 +6,21 @@ import (
...
 	"amd64": {
-		"darwin": download.NewRemoteFile("", ""),
-		"linux": download.NewRemoteFile(
-			"https://storage.googleapis.com/crc-bundle-github-ci/4.9.5/crc_libvirt_4.9.5.crcbundle",
-			"603d87599f4c129683d3a8a11bcd41f240a529b2dd81075755eebf3d2b8cacc6",
-		),
-		"windows": download.NewRemoteFile("", ""),
+		"darwin": {
+			"openshift": download.NewRemoteFile("", ""),
+			"podman":    download.NewRemoteFile("", ""),
```
cfergeau, line 11: Would it be better to not add these dummy files and make sure a good error is returned? At the moment this gives:
```
INFO Downloading crc_hyperkit_4.9.5.crcbundle     
Get "": unsupported protocol scheme ""
```

cfergeau (changes requested):

pkg/crc/api/api_client_test.go
```diff
@@ -213,6 +214,8 @@ func TestConfigGetAll(t *testing.T) {
 		switch v := v.Value.(type) {
 		case int:
 			configs[k] = float64(v)
+		case preset.Preset:
+			configs[k] = float64(v)
```
cfergeau, line 218: cast to `float64`? This indicates a bug introduced by 80c53b7, the json serialization of the configuration should be `preset: "podman"`
I'd revert this part of 80c53b7:
```
diff --git a/pkg/crc/preset/preset.go b/pkg/crc/preset/preset.go
index 57cc3883..b038dda3 100644
--- a/pkg/crc/preset/preset.go
+++ b/pkg/crc/preset/preset.go
@@ -6,11 +6,11 @@ import (
 )
 
-type Preset int
+type Preset string
 
 const (
-       Podman Preset = iota
-       OpenShift
+       Podman    Preset = "podman"
+       OpenShift Preset = "openshift"
 )
 
 func (preset Preset) String() string {
```

Makefile
```diff
@@ -53,6 +54,7 @@ __check_defined = \
 # Linker flags
 VERSION_VARIABLES := -X $(REPOPATH)/pkg/crc/version.crcVersion=$(CRC_VERSION) \
 	-X $(REPOPATH)/pkg/crc/version.bundleVersion=$(BUNDLE_VERSION) \
```
cfergeau, line 56: `bundleVersion` could be renamed to `openshiftVersion` for more clarity in a follow-up commit.

pkg/crc/constants/constants.go
cfergeau: > preset value doesn't contain the required version as part of switch so this is still required.

But `defaultBundleForOs(bundleVersion, podmanVersion string, preset crcpreset.Preset)` is not a nice API at all, depending on the value of `preset`, either `bundleVersion` or `podmanVersion` is going to be ignored, `test/e2e/crcsuite/crcsuite.go ` even ends up not needing one of the args, ...
One way to clean it up is to remove `GetBundleForOs`/`GetDefaultBundleForOs`, which allows to make the whole API simpler. |crcsuite.go` should be able to directly use `bundleName = constants.GetDefaultBundle(preset.OpenShift)`

pkg/crc/validation/validation.go
cfergeau: This concern is still pending

Makefile
praveenkumar: @cfergeau yes that something I can do a follow up commit

pkg/crc/constants/constants.go
praveenkumar: @cfergeau Sure but some follow up commit.

pkg/crc/preset/preset.go
```diff
@@ -6,35 +6,23 @@ import (
...
-}
-
 func ParsePresetE(input string) (Preset, error) {
 	switch input {
-	case Podman.String():
+	case string(Podman):
 		return Podman, nil
-	case OpenShift.String():
+	case string(OpenShift):
 		return OpenShift, nil
```
cfergeau: I'd squash `Preset: Implement stringer interface for preset type` in this commit instead of partially reverting this revert with a separate commit ^^
The main reason for reverting is that we don't want to get `preset: 1` when interacting with the config in the HTTP API, but `preset: "podman"`

pkg/crc/validation/validation.go
```diff
@@ -55,17 +55,17 @@ func ValidateEnoughMemory(value int) error {
...
 
 	userProvidedBundle := filepath.Base(bundlePath)
-	if userProvidedBundle != constants.GetDefaultBundle() {
+	if userProvidedBundle != constants.GetDefaultBundle(preset) {
 		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
 		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
 		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
-			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle())))
+			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(crcpreset.OpenShift))))
 		if usingCustomBundle {
```
cfergeau, line 69: `preset`?

pkg/crc/validation/validation.go
```diff
@@ -74,16 +74,16 @@ func ValidateBundlePath(bundlePath string) error {
 			logging.Warnf("Using unsupported bundle %s", userProvidedBundle)
 			return nil
 		}
-		return fmt.Errorf("%s is not supported by this crc executable, please use %s", userProvidedBundle, constants.GetDefaultBundle())
+		return fmt.Errorf("%s is not supported by this crc executable, please use %s", userProvidedBundle, constants.GetDefaultBundle(crcpreset.OpenShift))
```
cfergeau, line 77: `preset`?

pkg/crc/config/settings.go
```diff
@@ -72,12 +72,16 @@ func RegisterSettings(cfg *Config) {
 		return ValidateMemory(value, GetPreset(cfg))
 	}
 
+	validateBundle := func(value interface{}) (bool, string) {
```
cfergeau, line 75: `validateBundlePath` maybe?

pkg/crc/constants/constants.go
cfergeau: Feels like a missed opportunity to first make the code more complicated/harder to review/.., and then on a second pass clean it up, when it can be first cleaned up, and then we never add this hard to follow code.

cfergeau (approved):

PR #2842, opened 2021-11-17
Update go modules we depend on to their latest versions

cfergeau:
There should be no functional/user-visible changes.

PR #2847, opened 2021-11-23
dev: Update go modules we depend on to their latest

cfergeau:
This is the same PR as #2842 but for the dev branch


cfergeau:
centos CI has `18:04:55 + git rebase origin/master`, this does not look correct when running against the `dev` branch.

PR #2858, opened 2021-11-30
RFC: Hide libMachineAPIClient/CrcBundleMetadata/libmachine.Host from `pkg/crc/machine`

cfergeau:
At the moment, a lot of the code in `pkg/crc/machine` is doing:
```
       libMachineAPIClient, cleanup := createLibMachineClient()
       defer cleanup()

       host, err := libMachineAPIClient.Load(client.name)
       if err != nil {
               return nil, nil, errors.Wrap(err, "Cannot load machine")
       }

       crcBundleMetadata, err := getBundleMetadataFromDriver(host.Driver)
       if err != nil {
               return nil, nil, errors.Wrap(err, "Error loading bundle metadata")
       }
```

Sometimes this code only needs `crcBundelMetadata`, sometimes it also needs `host` (to get the VM IP for example), and some other times, `libMachineAPIClient` is also needed in order to create/remove the VM
This is all quite noisy, and exposes some relatively low-level details to code more focused on crc/OpenShift logic.
This commit tries to hide this code from `machine` by adding a `virtualMachine` concept which can be seen as abstracting the `~/.crc/machines/crc` directory.
Apart from making code in `pkg/crc/machine` easier to follow, this should also make it easier to split some methods in this package, which could be useful for the podman work.

### DIFF
--- a/pkg/crc/machine/console.go
+++ b/pkg/crc/machine/console.go
@@ -10,7 +10,7 @@ func (client *client) GetConsoleURL() (*types.ConsoleResult, error) {
 	// Here we are only checking if the VM exist and not the status of the VM.
 	// We might need to improve and use crc status logic, only
 	// return if the Openshift is running as part of status.
-	vm, err := loadVirtualMachine(client.name)
+	vm, err := loadVirtualMachine(client.name, client.useVSock())
 	if err != nil {
 		return nil, errors.Wrap(err, "Cannot load machine")
 	}
--- a/pkg/crc/machine/console.go
+++ b/pkg/crc/machine/console.go
@@ -11,24 +11,18 @@ func (client *client) GetConsoleURL() (*types.ConsoleResult, error) {
 	// Here we are only checking if the VM exist and not the status of the VM.
 	// We might need to improve and use crc status logic, only
 	// return if the Openshift is running as part of status.
-	libMachineAPIClient, cleanup := createLibMachineClient()
-	defer cleanup()
-	host, err := libMachineAPIClient.Load(client.name)
+	vm, err := loadVirtualMachine(client.name)
 	if err != nil {
 		return nil, errors.Wrap(err, "Cannot load machine")
 	}
+	defer vm.Close()
 
-	vmState, err := host.Driver.GetState()
+	vmState, err := vm.Driver.GetState()
 	if err != nil {
-		return nil, errors.Wrap(err, "Error getting the state for host")
+		return nil, errors.Wrap(err, "Error getting the state of the virtual machine")
 	}
 
-	crcBundleMetadata, err := getBundleMetadataFromDriver(host.Driver)
-	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/crc/machine/host.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/machine/bundle"
+	"github.com/code-ready/crc/pkg/libmachine"
+	libmachinehost "github.com/code-ready/crc/pkg/libmachine/host"
+	"github.com/pkg/errors"
+)
+
+type virtualMachine struct {
+	name string
+	*libmachinehost.Host
```
praveenkumar, line 17: Why this doesn't have any name association something like `host *libmachinehost.Host` ?

cfergeau: This allows to use the methods from `libmachinehost.Host` on a `virtualMachine` instance, even if I don't remember if I make use of this.
```
  vm := virtualMachine{}
  vm.Stop()
  vm.Kill()
```
(the last 2 methods are from `pkg/libmachine/host/`)

praveenkumar: So if we put `host *libmachinehost.Host` then we need to do `vm.host.Stop()` which is kind of cryptic.

pkg/crc/machine/host.go
```diff
@@ -72,3 +73,11 @@ func (vm *virtualMachine) Remove() error {
 
 	return nil
 }
+
+func (vm *virtualMachine) State() (state.State, error) {
```
praveenkumar, line 80: 👍🏼

praveenkumar:
Right now we are creating the `virtualMachine` instance and assign it to `host` which again make code bit less intuitive it would be better to have it assign to `vm` and then something like `vmState, err := host.State()` should become `vmState, err := vm.State()`. 

Also we should cherry pick this one to master right?

pkg/crc/machine/console.go
```diff
@@ -17,9 +16,9 @@ func (client *client) GetConsoleURL() (*types.ConsoleResult, error) {
 	}
 	defer vm.Close()
 
-	vmState, err := vm.Driver.GetState()
+	vmState, err := vm.State()
 	if err != nil {
-		return nil, errors.Wrap(err, "Error getting the state of the virtual machine")
+		return nil, errors.Wrap(err, "Error getting the state for host")
```
praveenkumar, line 21: We are getting state of the VM so why error message is changed? I think initial one is more correct.

cfergeau: Some small rebase issue, I'll fix it, thanks for catching it!

cfergeau:
> Also we should cherry pick this one to master right?

Depends how careful we want to be with master. It could go there yeah as it's not changing any behaviour. But since it's changing code, there could be unexpected side effects causing bugs ^^

PR #2874, opened 2021-12-03
Remove obsolete code/factor some machine/bundle code a bit

cfergeau:
This should not cause any behaviour change, just removing some no longer useful code, or some c&p'ed code.

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -118,9 +118,9 @@ func (bundle *CrcBundleInfo) GetKubeConfigPath() string {
 	return bundle.resolvePath(bundle.ClusterInfo.KubeConfig)
 }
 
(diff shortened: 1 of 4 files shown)


praveenkumar:
Except the first commit we can have it in the dev branch which can be cherry-pick on the master also.

cfergeau:
> Except the first commit 

Any suggestions how to improve it?

cfergeau:
CI failures seem to be infra-related:
```
ERROR: gcloud crashed (SystemError): ffi_prep_closure(): bad user_data (it seems that the version of the libffi library seen at runtime is different from the 'ffi.h' file seen at compile-time)
If you would like to report this issue, please run the following command:
  gcloud feedback
To check gcloud for common problems, please run the following command:
  gcloud info --run-diagnostics 
```

PR #2887, opened 2021-12-16
Remove bundle from mac and windows installer

praveenkumar:
This patch remove the bundle as part of installer for mac/windows
platform and now it will be downloaded as part of `crc setup` command.

This patch also remove the default bundle path from installed directory
to crc cache directory so user not face permission issue to download it.

PR #2888, opened 2021-12-16
Daemon API: Reregister the settings as part of start/get api call

praveenkumar:
As of now daemon use the initial config settings during the initial
start and then when user try to update the config which need reregister
to get default values (in case of preset), it doesn't updated in config
Storage API side. User get different result for `config view` from cli
and daemon side.

```
// Initial start of the tray with following config
✗ crc config view

// Check the defaults using api request
✗ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/config | jq .
{
  "Success": true,
  "Error": "",
  "Configs": {
    "autostart-tray": true,
    "bundle": "/Users/prkumar/.crc/cache/crc_hyperkit_4.9.10.crcbundle",
    "consent-telemetry": "no",
    "cpus": 4,
    "disable-update-check": false,
    "disk-size": 31,
    "memory": 9216,
    "preset": "openshift",
    ...
  }
}

// Use podman preset
✗ crc config set preset podman
✗ crc config view
- preset                                : podman

// Again check the defaults using api
✗ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/config | jq .
{
  "Success": true,
  "Error": "",
  "Configs": {
    "autostart-tray": true,
    "bundle": "/Users/prkumar/.crc/cache/crc_hyperkit_4.9.10.crcbundle",   => no change
    "consent-telemetry": "no",
    "cpus": 4,  => no change
    "disable-update-check": false,
    "disk-size": 31,
    "memory": 9216, => no change
    "preset": "podman",
    ...
  }
}
```

This patch is going to use `Config` type and then reregister the setting
before start or get call from the API.

fixes: #2887

### DIFF
--- a/pkg/crc/api/api_http.go
+++ b/pkg/crc/api/api_http.go
@@ -8,7 +8,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/machine"
 )
 
-func NewMux(config crcConfig.Storage, machine machine.Client, logger Logger, telemetry Telemetry) http.Handler {
+func NewMux(config *crcConfig.Config, machine machine.Client, logger Logger, telemetry Telemetry) http.Handler {
 	handler := NewHandler(config, machine, logger, telemetry)
 
 	server := newServerWithRoutes(handler)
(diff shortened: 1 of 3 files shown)


anjannath (approved):
LGTM

cfergeau (commented):
Reviewed this yesterday, but forgot to submit /o\

PR #2899, opened 2021-12-21
Expose cockpit port(9090) in case of podman preset

praveenkumar:
This patch open the port as per preset. In case of Openshift it is
unchanged and in case of podman it expose cockpit port 9090.

### DIFF
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -6,13 +6,14 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/daemonclient"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/preset"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
 	"github.com/pkg/errors"
 )
 
-func exposePorts() error {
+func exposePorts(preset preset.Preset) error {
 	daemonClient := daemonclient.New()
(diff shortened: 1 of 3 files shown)


pkg/crc/machine/vsock.go
```diff
@@ -48,25 +49,36 @@ const (
...
+			{
+				Local:  fmt.Sprintf(":%d", httpsPort),
+				Remote: fmt.Sprintf("%s:%d", virtualMachineIP, httpsPort),
+			},
+			{
+				Local:  fmt.Sprintf(":%d", httpPort),
+				Remote: fmt.Sprintf("%s:%d", virtualMachineIP, httpPort),
+			},
+		}...)
+	} else {
```
gbraad, line 77: please make the if checked named too and error when preset is not matched.
use the same pattern in the client.

gbraad (changes requested):

pkg/crc/machine/vsock.go
```diff
@@ -48,25 +49,37 @@ const (
...
+				Local:  fmt.Sprintf(":%d", httpsPort),
+				Remote: fmt.Sprintf("%s:%d", virtualMachineIP, httpsPort),
+			},
+			{
+				Local:  fmt.Sprintf(":%d", httpPort),
+				Remote: fmt.Sprintf("%s:%d", virtualMachineIP, httpPort),
+			},
+		}...)
+	}
+	if ps == preset.Podman {
```
gbraad, line 78: 👍 
intent is now much clearer

gbraad (approved):

PR #2901, opened 2021-12-22
e2e default bundle path for installers

### LINKED ISSUE #2712: [QE] change current bundle path on e2e
Current e2e specs uses the old path based on crc home folder. Now on macos and windows the default path has changed so this need to be updated on crcsuite for e2e tests.

adrianriobo:
Fixes #2712

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,8 +157,8 @@ clean: clean_docs clean_macos_package clean_windows_msi
 .PHONY: build_e2e
 build_e2e: $(SOURCES)
 	GOOS=linux   go test ./test/e2e/ -c -o $(BUILD_DIR)/linux-amd64/e2e.test
-	GOOS=windows go test ./test/e2e/ -c -o $(BUILD_DIR)/windows-amd64/e2e.test.exe
-	GOOS=darwin  go test ./test/e2e/ -c -o $(BUILD_DIR)/macos-amd64/e2e.test
+	GOOS=windows go test --ldflags="-X $(REPOPATH)/test/e2e/crcsuite.installerBuild=true" ./test/e2e/ -c -o $(BUILD_DIR)/windows-amd64/e2e.test.exe
+	GOOS=darwin  go test --ldflags="-X $(REPOPATH)/test/e2e/crcsuite.installerBuild=true" ./test/e2e/ -c -o $(BUILD_DIR)/macos-amd64/e2e.test
 
 .PHONY: build_integration
 build_integration: $(SOURCES)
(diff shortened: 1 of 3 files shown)


test/e2e/crcsuite/crcsuite.go
```diff
@@ -433,7 +435,11 @@ func CommandReturnShouldContainIfBundleEmbeddedOrNot(commandField string, value
 func SetConfigPropertyToValueSucceedsOrFails(property string, value string, expected string) error {
 	if value == "current bundle" {
 		if bundleEmbedded {
-			value = filepath.Join(CRCHome, "cache", bundleName)
+			if installerBuild != "false" {
+				value = filepath.Join(cmd.GetCRCBinaryPath(), bundleName)
```
praveenkumar, line 439: That mean when it is not the installer build then the bundle location should be same as binary location? Also I am not sure why we are making this change on master now (this should be just maintenance branch for time being).

adrianriobo: Well this is the current behavior managed not based on platform but on the installer control variable, But basically installers uses its own path (os dependant) in the meantime linux uses old home folder...so just fixing this as this is creating a false postive on e2e tests.

THis is done on master as it is fixing a bug on e2e.

praveenkumar: Ack

Makefile
```diff
@@ -157,8 +157,8 @@ clean: clean_docs clean_macos_package clean_windows_msi
 .PHONY: build_e2e
 build_e2e: $(SOURCES)
 	GOOS=linux   go test ./test/e2e/ -c -o $(BUILD_DIR)/linux-amd64/e2e.test
-	GOOS=windows go test ./test/e2e/ -c -o $(BUILD_DIR)/windows-amd64/e2e.test.exe
-	GOOS=darwin  go test ./test/e2e/ -c -o $(BUILD_DIR)/macos-amd64/e2e.test
+	GOOS=windows go test --ldflags="-X $(REPOPATH)/test/e2e/crcsuite.installerBuild=true" ./test/e2e/ -c -o $(BUILD_DIR)/windows-amd64/e2e.test.exe
+	GOOS=darwin  go test --ldflags="-X $(REPOPATH)/test/e2e/crcsuite.installerBuild=true" ./test/e2e/ -c -o $(BUILD_DIR)/macos-amd64/e2e.test
```
praveenkumar, line 161: 👍🏼

praveenkumar (approved):

Issue #2890, opened 2021-12-18
[Task] Unique bearer token to allow cockpit login (podman preset)


gbraad:
When the podman preset is used, the user should be able to login using a unique bearer token. At the moment the token has been statically defined as mentioned in https://github.com/code-ready/snc/pull/494/files#r771795733. Once the image is started, a unique `base64` encoded value should be used; this would be generated by `crc` and stored inside the VM, and used by tray-electron as the bearer authentication method.

Issue #2667, opened 2021-08-24
Use firewall-cmd instead of haproxy?


cfergeau:
Following an internal discussion, an alternative to configuring haproxy to forward requests to the VM is to use firewall-cmd to forward external requests on port 443/80/.. to the VM:

```
firewall-cmd --add-forward-port=port=443:proto=tcp:toaddr=$(crc ip):toport=443
firewall-cmd --add-forward-port=port=6443:proto=tcp:toaddr=$(crc ip):toport=6443
firewall-cmd --add-forward-port=port=80:proto=tcp:toaddr=$(crc ip):toport=80
firewall-cmd --direct --passthrough ipv4 -I FORWARD -i crc -j ACCEPT
firewall-cmd --direct --passthrough ipv4 -I FORWARD -o crc -j ACCEPT
```

Not implying this is better, but recording this here so that this is not forgotten and can be considered.

jmazzitelli:
I tested this and it appears to work for me. I take no credit for this, as these are [the same basic rules](https://github.com/kxr/ocp4_setup_upi_kvm/blob/d97e566ac9b8dea6a39d2f494f3380a09e35f7b6/.post_scripts/expose_cluster.sh#L128-L137) that are used by the [ocp4_setup_upi_kvm script](https://github.com/kxr/ocp4_setup_upi_kvm).

praveenkumar:
This might allow us to use user network mode on Linux also, only problem is does this firewall cmd available to all distro or we need iptables cli along with firewall command.

cfergeau:
With usermode networking, the daemon by default listens on all interfaces. On my RHEL box it's firewalld which prevents external access. These commands are enough to allow external access:
```
firewall-cmd --zone=public --add-port=2222/tcp
firewall-cmd --zone=public --add-service=https
firewall-cmd --zone=public --add-service=http
```
(to be adjusted depending on what you want to access :)

Kalinins93:
> With usermode networking, the daemon by default listens on all interfaces. On my RHEL box it's firewalld which prevents external access. These commands are enough to allow external access:
> 
> ```
> firewall-cmd --zone=public --add-port=2222/tcp
> firewall-cmd --zone=public --add-service=https
> firewall-cmd --zone=public --add-service=http
> ```
> 
> (to be adjusted depending on what you want to access :)

How property and config?. CRC view and host? i did it, but my crc remote don't work.

cmoulliard:
[...]

PR #2903, opened 2021-12-24
Vsock: Add ip to local port mapping

praveenkumar:
As of now, port expose happen on all the interface `:<port>`, with
this PR it happen only on the localhost/127.0.0.1 so other interface
are free to use for same port if required.

### DIFF
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -46,6 +46,7 @@ func isOpened(exposed []types.ExposeRequest, port types.ExposeRequest) bool {
 const (
 	virtualMachineIP = "192.168.127.2"
 	internalSSHPort  = 22
+	localIP          = "127.0.0.1"
 	httpPort         = 80
 	httpsPort        = 443
 	apiPort          = 6443
@@ -55,7 +56,7 @@ const (
 func vsockPorts(preset crcPreset.Preset) []types.ExposeRequest {
 	exposeRequest := []types.ExposeRequest{
 		{
-			Local:  fmt.Sprintf(":%d", constants.VsockSSHPort),
+			Local:  fmt.Sprintf("%s:%d", localIP, constants.VsockSSHPort),
 			Remote: fmt.Sprintf("%s:%d", virtualMachineIP, internalSSHPort),
 		},
 	}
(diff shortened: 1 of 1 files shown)


gbraad:
Will have to test this

pkg/crc/machine/vsock.go
```diff
@@ -46,6 +46,7 @@ func isOpened(exposed []types.ExposeRequest, port types.ExposeRequest) bool {
 const (
 	virtualMachineIP = "192.168.127.2"
 	internalSSHPort  = 22
+	hostInterfaceIP  = "127.0.0.1"
```
cfergeau, line 49: I would drop the `Interface` from the name. Dunno if the right name is `host`, or just `localIP`, `loopbackIP`

cfergeau (commented):
Not 100% sure what the usecase is for this/what concrete problems this will solve, could you detail a bit more in the commit log?
The daemon would bind only on 127.0.0.1:80, leaving 192.168.x.y:80 free for use by another webserver, but is this really something we want to encourage? Similarly, we could instruct users to configure sshd to listen on all IPs except localhost, and use 22 to access the VM instead of 2222, but I'm not sure it's going to be that convenient to configure over using 2222?

A side effect of this change is that https://github.com/code-ready/crc/issues/2667#issuecomment-915257243 would probably no longer work.

gbraad:
> A side effect of this change is that #2667 (comment) would probably no longer work.

Can you confirm this?

praveenkumar:
/close

For cockpit following issues are created.
- https://github.com/code-ready/tray-electron/issues/30
- https://github.com/code-ready/crc/issues/2890

praveenkumar:
> > A side effect of this change is that #2667 (comment) would probably no longer work.
> 
> Can you confirm this?

It will not work as it is but with documents, user can use something like following to make it work.

```
$ curl -X POST -d '{"local": "<external_ip>:6443", "remote": "192.168.127.2:6443"}'  --unix-socket ~/.crc/crc-http.sock http:/unix/network/services/forwarder/expose
```

cfergeau:
> It will not work as it is but with documents, user can use something like following to make it work.

But this is not going to be persistent, so will have to be redone after each VM reboot, or even more often than that when https://github.com/code-ready/crc/issues/2923 is solved.
The commit log does not reflect the most recent conversation we had about why it's desirable to make this change.

cfergeau (changes requested):
https://github.com/code-ready/crc/pull/2903#issuecomment-1009717556

praveenkumar:
> > It will not work as it is but with documents, user can use something like following to make it work.
> 
> But this is not going to be persistent, so will have to be redone after each VM reboot, or even more often than that when #2923 is solved. 

Yes, that would be the limitation at the moment but it makes apiserver/cockpit more secure for unknown external access.

> The commit log does not reflect the most recent conversation we had about why it's desirable to make this change.

@cfergeau Updated the comment with recent discussion.

cfergeau (approved):

cfergeau:
Totally forgot a very important detail, on macOS, a regular user cannot bind on localhost-only, you need additional privileges for that. This broke `crc start` until I reverted this.

PR #2909, opened 2021-12-28
Embed electron based mac tray to installer

praveenkumar:


Makefile
```diff
@@ -5,7 +5,7 @@ PODMAN_VERSION ?= 3.4.2
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 1.34.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)
-MACOS_INSTALL_PATH = /Applications/CodeReady Containers.app/Contents/Resources/
+MACOS_INSTALL_PATH = /Applications/CodeReady Containers.app/Contents/Resources/app
```
gbraad, line 8: * https://github.com/code-ready/tray-electron/pull/48

It seems for `isWin` we need `'..', '..'` and for `isMac` we need `'..'` ?
I prefer to place the binary back to the place where it is supposed to be.

gbraad (changes requested):

gbraad (approved):

Makefile
gbraad: Only Windows will expect the binary in the root of the application folder.

Issue #2905, opened 2021-12-27
[BUG] Status API does not include preset when state is 'starting', 'stopping'.


gbraad:
```json
{
  CrcStatus: 'Stopping',
  OpenshiftStatus: 'Stopping',
  OpenshiftVersion: '',
  PodmanVersion: '',
  DiskUse: 0,
  DiskSize: 0,
  Error: '',
  Success: true,
  Preset: ''     <=   should be known
}
```

I have worked around this for now... but on starting, the actual presets and possible values for podman version, etc exist... just not the disk space usage.

praveenkumar:
This is fixed.

PR #2918, opened 2022-01-04
API: Add preset info as part of machine client

praveenkumar:
Using this patch, status api should now return the preset info
as part of starting/stopping/deleting state.

```
{
  CrcStatus: 'Stopping',
  OpenshiftStatus: 'Stopping',
  OpenshiftVersion: '',
  PodmanVersion: '',
  DiskUse: 0,
  DiskSize: 0,
  Error: '',
  Success: true,
  Preset: 'openshift'
}
```

fixes: #2905

### DIFF
--- a/pkg/crc/machine/client.go
+++ b/pkg/crc/machine/client.go
@@ -8,6 +8,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/machine/state"
 	"github.com/code-ready/crc/pkg/crc/machine/types"
 	"github.com/code-ready/crc/pkg/crc/network"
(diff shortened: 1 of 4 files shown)


gbraad:
> Preset: 'Openshift'

presets are case-insensitive, right?

praveenkumar:
> > Preset: 'Openshift'
> 
> presets are case-insensitive, right?

@gbraad it's in lower case.

gbraad:
Works as expected:
image

Note: the `NaN%` for division by zero is now obvious ;-)

gbraad (approved):

Issue #2891, opened 2021-12-20
[Dev] crc delete require bundle metadata


praveenkumar:
With https://github.com/code-ready/crc/pull/2858 in looks like delete now require to have bundle metadata. if user manually delete `cache` folder and then try to delete the vm, following error happen. 

```
$ mv ~/.crc/cache/ ~/.crc/cache_bak

$ ./crc delete --log-level debug
DEBU CodeReady Containers version: 1.34.0+28526d06 
DEBU OpenShift version: 4.9.10 (not embedded in executable) 
DEBU Running 'crc delete'                         
DEBU Checking file: /home/prkumar/.crc/machines/crc/.crc-exist 
Do you want to delete the OpenShift cluster? [y/N]: y
DEBU Checking file: /home/prkumar/.crc/machines/crc/.crc-exist 
DEBU Found binary path at /home/prkumar/.crc/bin/crc-driver-libvirt 
DEBU Launching plugin server for driver libvirt   
DEBU Plugin server listening at address 127.0.0.1:42149 
DEBU () Calling .GetVersion                       
DEBU Using API Version 1                          
DEBU () Calling .SetConfigRaw                     
DEBU () Calling .GetMachineName                   
DEBU (crc) Calling .GetBundleName                 
Cannot load machine: Error loading bundle metadata: could not find cached bundle info in /home/prkumar/.crc/cache/crc_podman_libvirt_3.4.1: stat /home/prkumar/.crc/cache/crc_podman_libvirt_3.4.1: no such file or directory
```

PR #2921, opened 2022-01-04
delete: Ignore errors loading bundle metadata

cfergeau:
commit 60dcd9ffa8446 introduced a regression with `crc delete`, if a
crc VM exists, but ~/.crc/cache is removed, then `crc delete` will error
out [1] instead of ignoring the error and sucessfully deleting the VM.

This commit adds a specific error for bundle loading issues so that
Delete() can ignore these errors.

This fixes https://github.com/code-ready/crc/issues/2891

[1] `Cannot load machine: Error loading bundle metadata: could not find
cached bundle info in /home/prkumar/.crc/cache/crc_podman_libvirt_3.4.1:
stat /home/prkumar/.crc/cache/crc_podman_libvirt_3.4.1: no such file or
directory`

### DIFF
--- a/pkg/crc/machine/delete.go
+++ b/pkg/crc/machine/delete.go
@@ -9,7 +9,7 @@ import (
 
 func (client *client) Delete() error {
 	vm, err := loadVirtualMachine(client.name, client.useVSock())
-	if err != nil {
+	if err != nil && !errors.Is(err, errInvalidBundleMetadata) {
 		return errors.Wrap(err, "Cannot load machine")
 	}
 	defer vm.Close()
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/delete.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func (client *client) Delete() error {
 	vm, err := loadVirtualMachine(client.name, client.useVSock())
-	if err != nil {
+	if err != nil && !errors.Is(err, errInvalidBundleMetadata) {
```
praveenkumar, line 12: 👍🏼

praveenkumar (approved):

PR #2943, opened 2022-01-17
Release ports during delete operation

praveenkumar:
Daemon is long running process and with usermode networking we expose
some of the ports to host and same happen when user uses podman bundle
and expose port for a container. Now those ports are remain exposed until
daemon process is stopped and started again. This PR try to use `unexpose`
api from the usermode networking and release all the ports which is exposed
by the process once user execute delete.

```
➜  macos-amd64 git:(dev) crc start
➜  macos-amd64 git:(dev) sudo netstat -p tcp -van | grep '^Proto\|LISTEN'
Password:
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)     rhiwat shiwat    pid   epid  state    options
tcp46      0      0  *.9090                 *.*                    LISTEN      131072 131072  67452      0 0x0100 0x00000006
tcp46      0      0  *.2222                 *.*                    LISTEN      131072 131072  67452      0 0x0100 0x00000006
[..]
➜  macos-amd64 git:(dev) podman run --log-level=debug -d -p 8081:80 docker.io/httpd:2.4
2849e69f4798613aa1aec8b8ce5d8481fd636ddfb3be6df125e184ae64c29a5c
➜  macos-amd64 git:(dev) sudo netstat -p tcp -van | grep '^Proto\|LISTEN'
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)     rhiwat shiwat    pid   epid  state    options
tcp46      0      0  *.8081                 *.*                    LISTEN      131072 131072  67452      0 0x0100 0x00000006
tcp46      0      0  *.9090                 *.*                    LISTEN      131072 131072  67452      0 0x0100 0x00000006
tcp46      0      0  *.2222                 *.*                    LISTEN      131072 131072  67452      0 0x0100 0x00000006
[...]
➜  macos-amd64 git:(dev) ./crc delete
➜  macos-amd64 git:(dev) sudo netstat -p tcp -van | grep '^Proto\|LISTEN'
<no_port_bind_to_crc_daemon_process>
```

### DIFF
--- a/pkg/crc/machine/delete.go
+++ b/pkg/crc/machine/delete.go
@@ -18,6 +18,13 @@ func (client *client) Delete() error {
 		return errors.Wrap(err, "Cannot remove machine")
 	}
 
+	// In case usermode networking make sure all the port bind on host should be released
+	if client.useVSock() {
+		if err := unexposePorts(); err != nil {
+			return err
+		}
+	}
+
 	if err := cleanKubeconfig(getGlobalKubeConfigPath(), getGlobalKubeConfigPath()); err != nil {
 		if !errors.Is(err, os.ErrNotExist) {
 			logging.Warnf("Failed to remove crc contexts from kubeconfig: %v", err)
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/daemonclient"
+	crcErrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	crcPreset "github.com/code-ready/crc/pkg/crc/preset"
 	"github.com/containers/gvisor-tap-vsock/pkg/types"
@@ -12,11 +13,10 @@ import (
 )
 
 func exposePorts(preset crcPreset.Preset) error {
-	daemonClient := daemonclient.New()
 	portsToExpose := vsockPorts(preset)
-	alreadyOpenedPorts, err := daemonClient.NetworkClient.List()
+	daemonClient := daemonclient.New()
+	alreadyOpenedPorts, err := listOpenPorts(daemonClient)
 	if err != nil {
-		logging.Error("Is 'crc daemon' running? Network mode 'vsock' requires 'crc daemon' to be running, run it manually on different terminal/tab")
 		return err
 	}
 	var missingPorts []types.ExposeRequest
@@ -41,6 +41,33 @@ func isOpened(exposed []types.ExposeRequest, port types.ExposeRequest) bool {
 		}
 	}
 	return false
+}
+
+func unexposePorts() error {
+	var mErr crcErrors.MultiError
+	daemonClient := daemonclient.New()
+	alreadyOpenedPorts, err := listOpenPorts(daemonClient)
+	if err != nil {
+		return err
+	}
+	for _, port := range alreadyOpenedPorts {
+		if err := daemonClient.NetworkClient.Unexpose(&types.UnexposeRequest{Local: port.Local}); err != nil {
+			mErr.Collect(errors.Wrapf(err, "failed to unexpose port %s ", port.Local))
+		}
+	}
+	if len(mErr.Errors) == 0 {
+		return nil
+	}
+	return mErr
+}
+
+func listOpenPorts(daemonClient *daemonclient.Client) ([]types.ExposeRequest, error) {
+	alreadyOpenedPorts, err := daemonClient.NetworkClient.List()
+	if err != nil {
+		logging.Error("Is 'crc daemon' running? Network mode 'vsock' requires 'crc daemon' to be running, run it manually on different terminal/tab")
+		return nil, err
+	}
+	return alreadyOpenedPorts, nil
 }
 
 const (


pkg/crc/machine/vsock.go
```diff
@@ -12,11 +12,9 @@ import (
 )
 
 func exposePorts(preset crcPreset.Preset) error {
-	daemonClient := daemonclient.New()
 	portsToExpose := vsockPorts(preset)
-	alreadyOpenedPorts, err := daemonClient.NetworkClient.List()
+	daemonClient, alreadyOpenedPorts, err := getClientAndOpenPorts()
```
cfergeau, line 16: functions returning multiple values, and listing them all in their names just confuse me :) Maybe this can be:
```
daemonClient := daemonclient.New()
alreadyOpenedPorts, err := listOpenPorts(daemonClient)
```
?

pkg/crc/machine/vsock.go
```diff
@@ -43,6 +41,29 @@ func isOpened(exposed []types.ExposeRequest, port types.ExposeRequest) bool {
...
 }
 
+func unExposePorts() error {
+	daemonClient, alreadyOpenedPorts, err := getClientAndOpenPorts()
+	if err != nil {
+		return err
+	}
+	for _, port := range alreadyOpenedPorts {
+		if err := daemonClient.NetworkClient.Unexpose(&types.UnexposeRequest{Local: port.Local}); err != nil {
+			return errors.Wrapf(err, "failed to unexpose port %s ", port.Local)
```
cfergeau, line 51: Might be a good opportunity for using `Multierror`, we want to unexpose as many ports as possible rather than stop at the first failure.

pkg/crc/machine/vsock.go
```diff
@@ -43,6 +41,29 @@ func isOpened(exposed []types.ExposeRequest, port types.ExposeRequest) bool {
...
+	}
+	return nil
+}
+
+func getClientAndOpenPorts() (*daemonclient.Client, []types.ExposeRequest, error) {
+	daemonClient := daemonclient.New()
+	alreadyOpenedPorts, err := daemonClient.NetworkClient.List()
+	if err != nil {
+		logging.Error("Is 'crc daemon' running? Network mode 'vsock' requires 'crc daemon' to be running, run it manually on different terminal/tab")
+		return daemonClient, nil, err
```
cfergeau, line 62: `return nil, nil ,err`, though I'd prefer to only return `([]types.ExposeRequest, error)`
It might even be simpler to just add a constant for the `logging.Error()` parameter, and to not add a helper for this.

cfergeau (commented):
Looks like something which should be done on stop as well?

pkg/crc/machine/delete.go
```diff
@@ -18,6 +18,13 @@ func (client *client) Delete() error {
 		return errors.Wrap(err, "Cannot remove machine")
 	}
 
+	// In case usermode networking make sure all the port bind on host should be released
+	if vm.vsock {
+		if err := unExposePorts(); err != nil {
```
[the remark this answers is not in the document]
praveenkumar: I tired unexposePorts before but it was showing as lint error in IDE so I need to use this way.

pkg/crc/machine/delete.go
```diff
@@ -18,6 +18,13 @@ func (client *client) Delete() error {
 		return errors.Wrap(err, "Cannot remove machine")
 	}
 
+	// In case usermode networking make sure all the port bind on host should be released
+	if vm.vsock {
```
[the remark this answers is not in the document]
praveenkumar: I thought we are now consuming `virtualmachine` instance which also have bool for vsock.

pkg/crc/machine/delete.go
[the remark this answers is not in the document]
cfergeau: I'm not getting any errors with `make cross-lint` after making this change to the dev branch
```
diff --git a/pkg/crc/machine/delete.go b/pkg/crc/machine/delete.go
index 2682efc4..9689c159 100644
--- a/pkg/crc/machine/delete.go
+++ b/pkg/crc/machine/delete.go
@@ -7,7 +7,12 @@ import (
 	"github.com/pkg/errors"
 )
 
+func unexposePort() error {
+	return nil
+}
+
 func (client *client) Delete() error {
+	_ = unexposePort()
 	vm, err := loadVirtualMachine(client.name, client.useVSock())
 	if err != nil && !errors.Is(err, errInvalidBundleMetadata) {
 		return errors.Wrap(err, "Cannot load machine")
```
Not sure what was failing on your side? If it's an IDE specific lint, maybe it can be fixed there/ignored?

pkg/crc/machine/vsock.go
praveenkumar: @cfergeau I think we should return as soon as any port which is exposed and api not able to unexpose it. Even if we use Multierror, at the end we are failing the delete operation because of an error occurred.

pkg/crc/machine/delete.go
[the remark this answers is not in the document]
praveenkumar: It was IDE specific not `make cross-lint` side, need to ignore it.

praveenkumar:
> Looks like something which should be done on stop as well?

@cfergeau yes, I do need to check the behavior of podman-machine in this regard and can create a separate PR.

pkg/crc/machine/vsock.go
cfergeau: After a `crc delete`, we want to be able to run `crc start`. Even there are some failures, we should try to delete as much as possible. If a port unexpose fails, then we should still try to unexpose as many ports as possible. If there was a failure during port unexpose, imo we should still do the rest of the delete process and remove the VM/...
Especially as there seems to be a small race we cannot do much about (`crc delete` starts unexposing ports, but podman-machine unexposed one of the ports while `crc delete` was running).

pkg/crc/machine/vsock.go
```diff
@@ -5,18 +5,18 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/daemonclient"
+	errors2 "github.com/code-ready/crc/pkg/crc/errors"
```
cfergeau, line 8: crcErrors?

cfergeau (commented):
One small nit, apart from this, ACK.

PR #2944, opened 2022-01-17
Centos CI: Remove bundle download step

praveenkumar:
In dev branch we don't consume downloaded bundle but download the
bundle as part of `crc setup`. This PR remove those steps from centos
CI to avoid confusion.

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-# bundle location
-BUNDLE_VERSION=4.9.10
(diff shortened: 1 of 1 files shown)


praveenkumar:
Will need to do same for openshift CI but that PR would be in openshift/release side.

cfergeau (approved):

praveenkumar:
/close

Looks like we still need to download the bundle but then use it with `setup -b` so this can work with installer e2e side also and here also.

Issue #2866, opened 2021-11-30
Issues with podman + bundle generation?


cfergeau:
There appeared to be potential issues during code review
https://github.com/code-ready/crc/pull/2824#discussion_r749463659
https://github.com/code-ready/crc/pull/2824#discussion_r749462731

praveenkumar:
This is fixed.

Issue #2865, opened 2021-11-30
bundleMismatchWithPreset check is done too late


cfergeau:
https://github.com/code-ready/crc/pull/2824#discussion_r749470003

Issue #2864, opened 2021-11-30
Don't update kubeadmin password when using podman


cfergeau:
With the podman preset:
```
INFO Generating new password for the kubeadmin user 
```
This is not needed.

Issue #2859, opened 2021-11-30
Misleading message after changing preset


cfergeau:
It currently says:
```
$ crc config set preset podman
"Changes to configuration property 'preset' are only applied when the CRC instance is created."
"If you already have a running CRC instance, then for this configuration change to take effect, delete the CRC instance with 'crc delete' and start it with 'crc start'."
```

but if the podman bundle is not available yet, `crc setup` will be needed.

gbraad:
> but if the podman bundle is not available yet, crc setup will be needed.

Interesting to know how this is going to be messaged to the user. The tray/app needs to do the right thing too.

PR #2946, opened 2022-01-18
Fixes Misc issue for dev branch

praveenkumar:
Try to fix following issues 

- Issues with podman + bundle generation? #2866
- bundleMismatchWithPreset check is done too late #2865
- Don't update kubeadmin password when using podman #2864
- Active preset should be sent through telemetry #2861
- Misleading message after changing preset #2859

### DIFF
--- a/pkg/crc/machine/bundle/copier.go
+++ b/pkg/crc/machine/bundle/copier.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/code-ready/crc/pkg/compress"
@@ -38,11 +39,17 @@ func NewCopier(srcBundle *CrcBundleInfo, basePath string, customBundleName strin
 	}
 
 	copier.copiedBundle.Name = customBundleName
-	copier.copiedBundle.Type = fmt.Sprintf("%s_custom", srcBundle.Type)
+	copier.copiedBundle.Type = getType(srcBundle.Type)
 	copier.srcBundle = srcBundle
 	copier.copiedBundle.cachedPath = bundlePath
 
 	return &copier, nil
+}
+
+func getType(srcType string) string {
+	customType := "_custom"
+	srcType = strings.TrimSuffix(srcType, customType)
+	return fmt.Sprintf("%s%s", srcType, customType)
 }
 
 func (copier *Copier) Cleanup() error {
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -189,7 +189,7 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 		if crcBundleMetadata.IsOpenShift() {
 			machineConfig.KubeConfig = crcBundleMetadata.GetKubeConfigPath()
 		}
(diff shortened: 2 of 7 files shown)


gbraad (approved):

pkg/crc/machine/bundle/copier.go
```diff
@@ -38,13 +39,21 @@ func NewCopier(srcBundle *CrcBundleInfo, basePath string, customBundleName strin
...
+	copier.copiedBundle.Type = getType(srcBundle.Type)
 	copier.srcBundle = srcBundle
 	copier.copiedBundle.cachedPath = bundlePath
 
 	return &copier, nil
 }
 
+func getType(srcType string) string {
+	customType := "_custom"
+	if strings.Contains(srcType, customType) {
```
cfergeau, line 51: `strings.HasSuffix`?

pkg/crc/machine/start.go
```diff
@@ -163,6 +163,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 			return nil, errors.Wrap(err, "Error getting bundle metadata")
 		}
 
+		if err := bundleMismatchWithPreset(startConfig.Preset, crcBundleMetadata); err != nil {
+			return nil, err
+		}
+
```
cfergeau, line 170: Isn't this only triggered at VM creation time though? And now for example `crc start; crc stop; crc config set preset podman; crc start` will not return an error?

pkg/crc/machine/start.go
```diff
@@ -520,7 +520,7 @@ func (client *client) validateStartConfig(startConfig types.StartConfig) error {
 	return nil
 }
 
-func createHost(machineConfig config.MachineConfig) error {
+func createHost(machineConfig config.MachineConfig, isOpenShift bool) error {
```
cfergeau, line 524: Maybe this should be part of `MachineConfig`? Not fully sure what this type represents exactly :)

cfergeau (commented):
A few minor comments but looks good overall

pkg/crc/machine/start.go
praveenkumar: @cfergeau In that scenario as soon as user do `crc config set preset podman`, we show the warning about `crc delete => crc setup => crc start` as callback. If user ignore this message then it should start the vm with same preset which user set on the fresh start. I think that is expected behavior?

pkg/crc/machine/start.go
praveenkumar: @cfergeau I can pass the preset if that make it more clear but I don't think we should have it part of MachineConfig just for this one.

pkg/crc/machine/start.go
cfergeau: > If user ignore this message then it should start the vm with same preset which user set on the fresh start. I think that is expected behavior?

Well, this is a change of behaviour introduced by this commit which was not mentioned in the commit log, so I pointed it out. You are right that if the user does not run delete, reusing the same preset as what the VM was created with is the right thing to do. However, won't the preset get in the way, as they'll be done for the new preset rather than the previous ones?

praveenkumar: @cfergeau Agreed, I moved this bit very early on the start method so now it should handle both the cases.

PR #2953, opened 2022-01-19
Vsock: Don't bind port < 1024 with localhost

praveenkumar:
With dcae4544 all the exposed ports are bind to localhost but in case
of MacOS to bind a port <1024 requires sudo permission which cause
the failure. This PR make sure only ports which are > 1024 are bind
to localhost and other are bind to all interfaces.

This will still help use to expose apiserver and cockpit to all the
interfaces.

```
$ sudo netstat -p tcp -van | grep '^Proto\|LISTEN'
Password:
Proto Recv-Q Send-Q  Local Address          Foreign Address        (state)     rhiwat shiwat    pid   epid  state    options
tcp46      0      0  *.80                   *.*                    LISTEN      131072 131072  68606      0 0x0100 0x00000006
tcp46      0      0  *.443                  *.*                    LISTEN      131072 131072  68606      0 0x0100 0x00000006
tcp4       0      0  127.0.0.1.6443         *.*                    LISTEN      131072 131072  68606      0 0x0100 0x00000006
tcp4       0      0  127.0.0.1.2222         *.*                    LISTEN      131072 131072  68606      0 0x0100 0x00000006
```

### DIFF
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -95,11 +95,11 @@ func vsockPorts(preset crcPreset.Preset) []types.ExposeRequest {
 				Remote: fmt.Sprintf("%s:%d", virtualMachineIP, apiPort),
 			},
 			types.ExposeRequest{
-				Local:  fmt.Sprintf("%s:%d", localIP, httpsPort),
+				Local:  fmt.Sprintf(":%d", httpsPort),
 				Remote: fmt.Sprintf("%s:%d", virtualMachineIP, httpsPort),
 			},
 			types.ExposeRequest{
-				Local:  fmt.Sprintf("%s:%d", localIP, httpPort),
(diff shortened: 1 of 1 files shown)


anjannath:
Just wondering if there is any specific macOS entitlement available that can be granted to still bind to only localhost?

praveenkumar:
> Just wondering if there is any specific macOS entitlement available that can be granted to still bind to only localhost?

@anjannath https://developer.apple.com/documentation/bundleresources/entitlements nothing as per official doc. Also https://github.com/code-ready/admin-helper/pull/17 was created long before to achieve same with admin helper.

cfergeau (commented):
The commit log seems misleading?

> This will still help use to expose apiserver and cockpit to all the
interfaces.

This is the opposite, we do not expose the apiserver and cockpit on all network interfaces.
Apart from this, looks good, and works fine on macOS.
ACK

Issue #2949, opened 2022-01-18
[BUG] segment issue shown on delete; segment failure are not supposed to be user visible


gbraad:
```
C:\Users\gbraad>crc delete
Do you want to delete the OpenShift cluster? [y/N]: y
Deleted the OpenShift cluster
ERRO sending request - Post "https://api.segment.io/v1/batch": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
ERRO 1 messages dropped because they failed to be sent and the client was closed
```

![image](https://user-images.githubusercontent.com/1894/149924023-a756617a-937b-497c-9a85-704debae8b95.png)

cfergeau:
Not clear why this gets triggered? is this expected that segment requests fail at this point? Did you do anything special to trigger this? Or does this always happen?

PR #2954, opened 2022-01-19
Segment: Send error to debug log instead stderr

praveenkumar:
Currently segment logger which we attach is sending error to stderr
using `logging.Errorf`. With this PR it is send to debug using `logging.Debugf`
which will avoid confusion for the users in case there is issue to accessing
the segment api due to network connectivity or in general.

```
---- Current Scenario -----
ERRO sending request - Post "https://api.segment.io/v1/batch": dial tcp: lookup api.segment.io: no such host
ERRO 1 messages dropped because they failed to be sent and the client was closed

--- With this PR ----
DEBU sending request - Post "https://api.segment.io/v1/batch": dial tcp: lookup api.segment.io: no such host
DEBU 1 messages dropped because they failed to be sent and the client was closed
```

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -250,7 +250,7 @@ func (l *loggingAdapter) Logf(format string, args ...interface{}) {
 }
 
 func (l *loggingAdapter) Errorf(format string, args ...interface{}) {
-	logging.Errorf(format, args...)
+	logging.Debugf(format, args...)
 }
 
 func errorType(err error) string {


cfergeau:
I'd rather we don't push this without also making sure telemetry errors are not fatal, as was apparently the case in #2949

praveenkumar:
> I'd rather we don't push this without also making sure telemetry errors are not fatal, as was apparently the case in #2949

@cfergeau In #2949 also the issue with connection timeout as @gbraad was not able to reach segment.io same fatal can happen if user is not connected to internet so I think it is not good idea to show user these kind of error directly to stderr which can be ignored.

gbraad:
No telemetry error should ever be shown to the user. It gives the wrong idea / confusing if the command failed or not. It is not relevant for them.

cfergeau:
If a command can fail because of a telemetry error, it's a step in the wrong direction to make these errors invisible. My understanding of issue #2949 is that the telemetry error is preventing the instance from being fully deleted? Or is this just reporting a useless error, which is harmless apart from the confusing messages?
I'd rather we start by ensuring telemetry errors  don't cause commands to fail, and once that's done, hiding these logs makes a lot of sense.

praveenkumar:
> My understanding of issue #2949 is that the telemetry error is preventing the instance from being fully deleted? Or is this just reporting a useless error, which is harmless apart from the confusing messages?

@cfergeau no it is not preventing the instance from being deleted. From user side it is useless error because user should not care about if data is uploaded to segment or not.

cfergeau:
> @cfergeau no it is not preventing the instance from being deleted.

grmble, so much for placeholder issues with missing information. No objections from my side then.

cfergeau (approved):

Issue #2860, opened 2021-11-30
Improvements to bundle downloading


cfergeau:
Miscellaneous things I wrote down when initially testing this:
- it's hard to recover from ctrl+c during initial bundle download, might be better to first download to a temp file, and then move it in place
- sha256sum is not checked if the file is already present
- when failing to unpack the bundle, we should fallback to trying to download it
- the bundle URL version and the version expected by crc are disconnected at the moment, would be nice to ensure they are consistent somehow

cfergeau:
steps to reproduce the first item:
```
$ rm -rf ~/.crc/cache
$ crc setup
...
INFO Downloading crc_podman_libvirt_3.4.2.crcbundle 
...
<press ctrl+C during the download>

$ crc setup
...
INFO Extracting bundle: crc_podman_libvirt_3.4.2... 
crc-podman.qcow2: 34.00 MiB / 1.62 GiB Error getting bundle metadata: unexpected EOF
```
One need to manually remove the partially downloaded file to fix this.

gbraad:
this has always been an issue. when the tray app is closed, this also results in partial files. no resume

PR #2955, opened 2022-01-19
Preflight: Improve bundle download to recover from ctrl+c and resume

praveenkumar:
Currently during the bundle download if user press `ctrl+c` or network
drop happen then `crc setup` fails to extract the bundle because downloaded
bundle is corrupt one and there is no sha256sum check in place. This PR adjust
the bundle download such that it always run in case user doesn't specify custom
bundle as part of setup command and download function have sha256sum verification
in place which take care of resuming the download in case it corrupted.

```
$ crc setup
...
INFO Downloading crc_podman_hyperkit_3.4.2.crcbundle
47.23 MiB / 683.84 MiB [-------->_________________________________________________________________________________________________________________] 6.91% 32.63 MiB p/s^C

$ crc setup
INFO Downloading crc_podman_hyperkit_3.4.2.crcbundle
122.70 MiB / 683.84 MiB [--------------------->__________________________________________________________________________________________________] 17.94% 26.95 MiB p/s^C

$ crc setup
INFO Downloading crc_podman_hyperkit_3.4.2.crcbundle
0 B / 683.84 MiB [________________________________________________________________________________________________________________________________________] 0.00% ? p/s
INFO Uncompressing /Users/prkumar/.crc/cache/crc_podman_hyperkit_3.4.2.crcbundle
crc-podman.qcow2: 515.87 MiB / 1.62 GiB [------------------------------------>_________________________________________________________________________________] 31.15%^C
```

fixes: #2860

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -95,6 +95,8 @@ func fixBundleExtracted(bundlePath string, preset preset.Preset) func() error {
 				/* This message needs to be improved when the bundle has been set in crc config for example */
 				return fmt.Errorf("%s is invalid or missing, run 'crc setup' to download the bundle", bundlePath)
 			}
+		}
+		if bundlePath == constants.GetDefaultBundlePath(preset) {
 			logging.Infof("Downloading %s", constants.GetDefaultBundle(preset))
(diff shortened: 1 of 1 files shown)


cfergeau (approved):
Nice trick, I was expecting something more complicated :)

PR #2960, opened 2022-01-20
Bump tray and misc changes

praveenkumar:


cfergeau (commented):
Does the build fail on 1.17 without the `go:build` lines? Or is this change at the moment purely proactive to prepare for the switch which will happen with some future go version?

Looks good apart from this question

praveenkumar:
> Does the build fail on 1.17 without the go:build lines? 

@cfergeau  Linter is failing without this if you are using go1.17 and do `make cross-lint`.

cfergeau:
> > Does the build fail on 1.17 without the go:build lines?
> 
> @cfergeau Linter is failing without this if you are using go1.17 and do `make cross-lint`.

yeah ok, this makes sense, I thought the commit log was saying the *build* was breaking.

PR #2961, opened 2022-01-20
Bump podman bundle to 3.4.4

praveenkumar:


cfergeau (approved):
an alternative implementation of `stopAllContainers()` using podman would probably make sense though when stopping the VM.

PR #2966, opened 2022-01-25
Start: Switch between preset without a delete

praveenkumar:
This PR adds a symlink creation function which create the symlink
to respective preset before creating the instance. It will allow
user to switch preset without deleting the existing vm but it has
to be stopped first.

/hold

This need to be tested on windows and some more checks are required before it become mergeable. 
- We do need to check if any of preset in running state before starting the preset.
- Handle the situation if user already have `~/.crc/machine/crc` folder and didn't deleted before this change.

### DIFF
--- a/pkg/crc/machine/delete.go
+++ b/pkg/crc/machine/delete.go
@@ -2,7 +2,9 @@ package machine
 
 import (
 	"os"
+	"path/filepath"
 
+	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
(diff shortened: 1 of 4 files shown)


anjannath:
Maybe make this a linux and macOS only feature for now?

gbraad:
@praveenkumar @cfergeau anything we still wanna consider?

gbraad:
> Maybe make this a linux and macOS only feature for now?

due to symlinks? you can use something similar on Windows using junctions, so that should not be limitation.
Perhaps better to describe what we want to do.

PR #2986, opened 2022-02-03
docs: Update terminology, use "instance" and "container runtime"

robin-owen:
Avoid mentioning virtual machines directly. Only one instance of the term "virtual machine" remains in the description of differences between the CRC OpenShift cluster and a production cluster.

This commit also introduces some rewording of generic procedures which work regardless of the selected container runtime preset. Further work on incorporating the idea of presets into the documentation is required; this represents a starting point.


gbraad (approved):

gbraad:
Note: this is merged into master. the change is generic enough.

PR #2991, opened 2022-02-04
Version cmd/api improvements

praveenkumar:
- Version: Add podman bundle version to version command/api
- Version: Remove bundle path detail from version command

### DIFF
--- a/pkg/crc/api/api_http_test.go
+++ b/pkg/crc/api/api_http_test.go
@@ -265,7 +265,7 @@ var testCases = []testCase{
 	// version
 	{
 		request:  get("version"),
-		response: jSon(fmt.Sprintf(`{"CrcVersion":"%s","CommitSha":"%s","OpenshiftVersion":"%s"}`, version.GetCRCVersion(), version.GetCommitSha(), version.GetBundleVersion())),
+		response: jSon(fmt.Sprintf(`{"CrcVersion":"%s","CommitSha":"%s","OpenshiftVersion":"%s","PodmanVersion":"%s"}`, version.GetCRCVersion(), version.GetCommitSha(), version.GetBundleVersion(), version.GetPodmanVersion())),
 	},
 
 	// version never fails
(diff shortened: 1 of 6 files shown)


pkg/crc/api/api_http_test.go
```diff
@@ -265,7 +265,7 @@ var testCases = []testCase{
 	// version
 	{
 		request:  get("version"),
-		response: jSon(fmt.Sprintf(`{"CrcVersion":"%s","CommitSha":"%s","OpenshiftVersion":"%s"}`, version.GetCRCVersion(), version.GetCommitSha(), version.GetBundleVersion())),
+		response: jSon(fmt.Sprintf(`{"CrcVersion":"%s","CommitSha":"%s","OpenshiftVersion":"%s","PodmanVersion":"%s"}`, version.GetCRCVersion(), version.GetCommitSha(), version.GetBundleVersion(), version.GetPodmanVersion())),
```
cfergeau, line 268: There were some discussions to give `OpenShiftVersion`/`PodmanVersion` more descriptive names, as these are the versions of the bundles supported by the binary, not necessarily the version currently being used.

anjannath (approved):

pkg/crc/api/api_http_test.go
anjannath: it'll make more sense when multiple bundles can be used with `crc` for now only the supported bundle can be used with the released `crc` binary, or did we remove that restriction?

cfergeau: The restriction is still in place

gbraad (approved):

praveenkumar:
/hold

More clarification require around what we really need from the `crc version` side. During slack conversation looks like we have to have array of supported versions but not sure what it would be.

cfergeau:
> /hold
> 
> More clarification require around what we really need from the `crc version` side. During slack conversation looks like we have to have array of supported versions but not sure what it would be.

@gbraad  ?

gbraad:
You can't include versions of future bundles, only older versions or a range.


It might even thwr fore suggest that this information is actually irrelevant here as the versions of podman and OpenShift are based on the actual bundle that runs.

I think a discussion around this is in place. As to me, the version command is more about the daemon/crc and versions of a bundle to run is decided somewhere else.

gbraad (approved):

PR #3008, opened 2022-02-15
Update openshift bundle to 4.9.18 (stable channel)

praveenkumar:


cfergeau (approved):
/lgtm
Checked that the base layer of the cvo image matches the base layer of the dnsmasq image we use, so both are based on the same ubi8 image.

gbraad:
This is the release bundle for sprint 214s deliverable?
/cc @praveenkumar @robin-owen

PR #3011, opened 2022-02-15
[cherry-pick] Improve crc stop message for already stopped cluster

anjannath:
It'll check if machine is already stopped before attempting
to perform the cluster stop related tasks

If a machine is already stopped it prints the following:

```
$ crc stop
Instance is already stopped
```

### DIFF
--- a/pkg/crc/machine/stop.go
+++ b/pkg/crc/machine/stop.go
@@ -9,6 +9,9 @@ import (
 )
 
 func (client *client) Stop() (state.State, error) {
+	if running, _ := client.IsRunning(); !running {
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/stop.go
```diff
@@ -9,6 +9,9 @@ import (
 )
 
 func (client *client) Stop() (state.State, error) {
+	if running, _ := client.IsRunning(); !running {
+		return state.Error, errors.New("Instance is already stopped")
```
praveenkumar, line 13: 👍🏼

praveenkumar (approved):

PR #3015, opened 2022-02-16
Update golangci-lint to latest version

cfergeau:
This also changes the way golangci-lint is installed when it's not available.


cfergeau:
`ci/prow/images` fails with
```
 go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.44.0: cannot query module due to -mod=vendor 
```
I'll need to investigate this ^^

praveenkumar:
> I'll need to investigate this ^^

May be an older version of go in prow CI side.

PR #3016, opened 2022-02-16
rpm: Add support for building crc linux releases using rpmbuild

cfergeau:
At the moment, linux releases are built on a crc-specific jenkins instance. This PR improves the rpm spec file in crc source so that it can be used to generate a linux release tarball.
It also builds windows/macos binaries with the proper release flags, which could then be reused in the build process for their respective installers.

PR #3022, opened 2022-02-18
Supersede for PR #3016

praveenkumar:
This PR supersede #3016 to check if openshift CI now successfully build.

/hold


cfergeau (commented):
For the record, I don't think `Testdata: Change the directory permission to 775`  has an impact on the CI issue, as it only changes file permissions from 0644 to 0755, it does not change the permissions of the testdata directory.
The next commit changing `release` to `linux-release` is the actual fix. `go test ./pkg/compress` causes the permissions of `testdata` to change to 0700, and switching from `release` to `linux-release` removes a `make check`, which would trigger the permission changes.

PR #3025, opened 2022-02-21
CI: Add integration test binary to final image

praveenkumar:
We should directly use integration test binary against the built
crc binary to perform the integration test on the openshift CI. Instead
of dumping the complete crc directory to final image and running `make
release`.

With this commit I am not removing the old crc directory dump to make
sure CI is green and then on openshift/release repo we can use this to
make sure it works with e2e test binary.


praveenkumar:
/close

looks like the binary still needs feature files.

PR #3027, opened 2022-02-22
RFC: Switch from appveyor/circleci to github actions

cfergeau:
This PR replaces the appveyor windows builds and circle ci macos/docs builds with github actions which provide builders for all 3 platforms.
Sample run can be found at https://github.com/cfergeau/crc/actions
There are some follow-up improvements which can be made (see if common yaml between the multiple action files can be shared, export the linux rpms as artifacts, ...), but this PR is already provided enough functionality to replace circleci/appveyor.


cfergeau:
The Windows failures are caused by a mismatch between \ and / during `make lint`, `$GOPATH` contains `\`, when running `golangci-lint`, we append `/bin/golangci-lint`, and the added `/` are not parsed as a path separator, so the path points to a non-existent path, and the build errors out.

anjannath:
i was thinking maybe drop the removal commit (https://github.com/code-ready/crc/pull/3027/commits/74b631d7377567b4101f12d1f7b0ea4544f387f2) from this PR and create another PR for the removal of appveyor/circle ci, that way, we merge this PR give it some time then remove the existing ones ?

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,28 @@
...
+      matrix:
+        os:
+          - ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Build docs
+        run: CONTAINER_RUNTIME=docker make build_docs
+      - name: Check links in docs
+        run: CONTAINER_RUNTIME=docker make  docs_check_links
```
anjannath, line 22: nit: extra space after `make `

anjannath (approved):
LGTM 👍🏽

.github/workflows/publish-docs.yml
cfergeau: fixed, thanks!

cfergeau:
> i was thinking maybe drop the removal commit ([74b631d](https://github.com/code-ready/crc/commit/74b631d7377567b4101f12d1f7b0ea4544f387f2)) from this PR and create another PR for the removal of appveyor/circle ci, that way, we merge this PR give it some time then remove the existing ones ?

I've moved away the circleci/appveyor removal, I'll file a PR with just that commit once this is merged so that we don't forget about it :)

anjannath (approved):

Issue #2480, opened 2021-06-25
[Epic] User can run crc on Apple silicon (M1/M2)


guillaumerose:
@gbraad here:

Our first target is to enable this for the Podman bundle.

Support to generate SNC bundles has been merged; https://github.com/code-ready/snc/pull/454, though as mentioned in https://github.com/code-ready/snc/pull/454#issuecomment-904835166 more work is needed in regards to the driver.

  * #1850
  * https://github.com/code-ready/crc/issues/2881#issuecomment-992269314

---

Currently, crc  is not running on Apple M1. The goal is to fully support this platform, like Apple and amd64 arch.

denismaggior8:
Any update on this?

guillaumerose:
crc currently uses hyperkit as the hypervisor. This is not possible anymore with M1. The easiest thing to do is to use qemu.

First step is to adapt https://github.com/code-ready/machine-driver-hyperkit/ for qemu.
We also need to update https://github.com/containers/gvisor-tap-vsock to the latest version.

gbraad:
We are currently working on a new driver for the Virtualization Framework that is part of Catalina and Big Sur.
More details around this will follow.

gbraad:
Support to generate SNC bundles has been merged; https://github.com/code-ready/snc/pull/454

cfergeau:
I made some progress on issue #1850, I can run crc using virtualization.framework for the VM. However, this is currently only working on x86_64, I haven't been able yet to start a VM on a M1 machine :-/

gbraad:
To time-line is to offer

* [x] fcos image for podman
* [ ] run this on arm/m1
* [ ] add microshift
* [ ] test microshift on arm/m1

praveenkumar:
I created the fcos bundle on `aarch64` and tried on m1 using vfkit (work which @cfergeau did) and able to start the podman bundle. 
https://github.com/code-ready/snc/issues/502

gbraad:
A dev preview for M1 is available: https://github.com/code-ready/crc/discussions/2988 Things might or might not work, so be sure to add. Comm ets to the discussion. Especially the UI is still changing.

mwitteveen:
Same as [raisingdibar](https://github.com/raisingdibar), I'm developing an application for the Dutch government and just bought a new Macbook m1 max to develop this on.

Is there any time indication for when the M1 Openshift release for CRC is expected?

At this moment, I am walking around with a separate Linux laptop just to run the cluster on.....

[...]

Issue #1850, opened 2021-01-18
Investigate making use of the new macos Virtualization framework instead of hyperkit


cfergeau:
See https://developer.apple.com/documentation/virtualization which provides a native virtualization layer in recent macos version (Big Sur and newer).

Miscellaneous projects making use of it:
 * https://github.com/evansm7/vftool
 * https://github.com/brucehoult/SimpleVM
 * https://github.com/Code-Hex/vz
 * https://github.com/gyf304/vmcli

This will need to be built on a native macOS machine (not different from the current hyperkit machine driver), but hopefully it would be possible to get rid of the out-of-process hyperkit driver, and have something more similar to the hyperv driver.

cfergeau:
Current status: https://github.com/cfergeau/crc/tree/macos-vf can run a podman/openshift VM with user or system networking using virtualization framework. The hyperkit code is removed from this branch.

There is no longer an external machine driver, it's been merged in crc (similar to the hyperv machine driver). This however depends on https://github.com/cfergeau/vfkit which is a very small command line wrapper on top of https://github.com/Code-Hex/vz which are go bindings for the virtualization framework.
vfkit+vz is less than 1000 lines of code but which must still be built on a macOS machine. I don't expect this code to change very often though. crc with the vfkit machine driver code can be cross-built.

This currently depends on qemu-img to be able to reuse the hyperkit bundles. Virtualization framework only supports raw images, so we need to convert the qcow2 in the bundle to a raw image file. Praveen pointed out that `qcow-tool decode` might be usable for this. We already ship `qcow-tool` with hyperkit, so this would be a better alternative.

Using Apple's virtualization framework means that macOS 11 or newer is required.

Next steps for that code in no particular order:
- test/fix on m1 machines
- remove the qemu-img/qcow-tool dependency. This means putting a raw disk image in our bundle, and making sure compression/decompression, file copies, ... all handle sparse files correctly as we don't want to allocate the full 31GB on disk when the space is not used. Disk resizing will need to be reimplemented on top of this
- some more refinement can be done in vfkit (some VM startup errors may not be caught, it should catch signals and then call into virtualization.framework to stop the VM, we can even imagine using a  management socket rather than relying on pid matching and signals, ...)

gbraad:
![image](https://user-images.githubusercontent.com/1894/151661415-6a579be6-5991-47a7-b716-08723ffe2fe5.png)

cfergeau:
[...]

PR #3035, opened 2022-02-23
mac: Switch from hyperkit to virtualization framework

cfergeau:
This PR removes the hyperkit driver from crc, and switches to [vfkit](https://github.com/code-ready/vfkit) instead, which makes use of macOS native virtualization framework to handle VMs.
There are various benefits to this:
- hyperkit is quite a bit of code, including ocaml code, and is not actively maintained lately, vfkit is a thin commandline wrapper over virtualization framework go bindings, and a lot less code
- vfkit does support M1 machines, while hyperkit does not
- the separate machine-driver-hyperkit binary is no longer needed

Some downsides:
- vpnkit support has been reverted for now, even if it probably could be readded
- vfkit has been a lot less tested than hyperkit
- some bundle generation changes are needed (no qcow2 support, only raw image support)

This PR adds x86_64 vfkit support, additional patches (which can be found in https://github.com/cfergeau/crc/tree/macos-vf ) are needed to enable M1 support.

**Fixes:** https://github.com/code-ready/crc/issues/1850

**Relates to:** https://github.com/code-ready/crc/issues/2480

## Testing

At this time, testing is a bit tricky as this needs a special bundle :-/ Apart from that, all which needs to be tested is if crc still works as usual on a mac.


cfergeau:
/hold
(it's ready for review, even if it looks like I did not run `make check` on a mac, but is not to be merged right now)

cfergeau:
> At this time, testing is a bit tricky as this needs a special bundle

Tentative vfkit bundle generation PR at https://github.com/code-ready/snc/pull/512 but I did not have time to test the generated bundle with the code in this PR.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -10,7 +10,7 @@ assignees: ""
 ### General information
 
   * OS: Linux / macOS / Windows
-  * Hypervisor: KVM / Hyper-V / hyperkit
+  * Hypervisor: KVM / Hyper-V / hyperkit / vfkit
```
gbraad, line 13:
```diff
-  * Hypervisor: KVM / Hyper-V / hyperkit
+  * Hypervisor: KVM / Hyper-V / vfkit
```

cmd/crc/cmd/daemon.go
```diff
@@ -182,26 +171,10 @@ func run(configuration *types.Configuration) error {
...
-					}
-				}()
-			}
-		} else {
-			mux := http.NewServeMux()
-			mux.Handle(types.ConnectPath, vn.Mux())
-			if err := http.Serve(vsockListener, mux); err != nil {
-				errCh <- errors.Wrap(err, "virtualnetwork http.Serve failed")
-			}
+		mux := http.NewServeMux()
```
gbraad, line 174: simplicity !

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -13,52 +13,32 @@ import (
...
-			checkDescription: "Checking if qcow-tool is installed",
-			check:            checkQcowToolInstalled,
-			fixDescription:   "Installing qcow-tool",
-			fix:              fixQcowToolInstalled,
+var vfkitPreflightChecks = []Check{
+	{
+		configKeySuffix:  "check-m1-cpu",
+		checkDescription: "Checking if running emulated on a M1 CPU",
+		check:            checkM1CPU,
+		fixDescription:   "CodeReady Containers is unsupported on Apple M1 hardware",
```
gbraad, line 21: `is currently unsupported`

gbraad (changes requested):
Minor issues or suggestions

.github/ISSUE_TEMPLATE/bug_report.md
cfergeau: People might still report bugs against older releases, so I'd keep this for a while

cmd/crc/cmd/daemon.go
cfergeau: In a way, yes, the VM networking is probably more complicated without vpnkit :)

pkg/crc/preflight/preflight_darwin.go
cfergeau: It would be better to reword this, this check only triggers when trying to run a x86_64 executable on a m1 cpu, which we'll never support.

cfergeau:
One thing I forgot to mention is that the part of the PR using sparse files when uncompressing the bundle does not work well with the uncompression progress bar, it gets stuck for a bit.

cmd/crc/cmd/daemon.go
gbraad: > networking is probably more complicated without vpnkit :)

can this be implemented?

pkg/crc/preflight/preflight_darwin.go
gbraad: `This version of CodeReady Containers for x86_64 is unsupported ...`

cmd/crc/cmd/daemon.go
cfergeau: This looked doable using https://github.com/Code-Hex/vz/blob/a2ebc854a26128c56da3cbeea653eb73e6c145c8/network.go#L95-L101 but I did not try to implement it at all.

cfergeau:
> One thing I forgot to mention is that the part of the PR using sparse files when uncompressing the bundle does not work well with the uncompression progress bar, it gets stuck for a bit.

I retested this, and the progress bar looked fine, it gets stuck for several seconds at some point, but I think that's because it's uncompressing or flushing data to disk, as when ti resumes, there is no big jump in the progress bar, so I don't think it's related to the sparse code.

cfergeau:
The vfkit bundle at https://mirror.openshift.com/pub/openshift-v4/clients/crc/bundles/openshift/4.10.9 was generated without this fix https://github.com/code-ready/snc/commit/dc615596befb69a320008e2c486b7cf07531882d
This means it's non functional as it's missing the ssh key to connect to the VM. vfkit support can't be merged for this release.

praveenkumar:
/unhold

Since we are done with 2.2.2 release, it should be good to make this PR in and have extensive tests around it.

cmd/crc/cmd/daemon.go
cfergeau: Added.

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -17,7 +17,7 @@ func hyperkitPreflightChecks(networkMode network.Mode) []Check {
 			configKeySuffix:  "check-m1-cpu",
 			checkDescription: "Checking if running emulated on a M1 CPU",
 			check:            checkM1CPU,
-			fixDescription:   "CRC is unsupported on Apple M1 hardware",
+			fixDescription:   "This version of CRC for AMD64/Intel64 is unsupported on Apple M1 hardware",
```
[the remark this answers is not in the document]
cfergeau: Fixed, thanks!

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -0,0 +1,430 @@
...
+// inherited from  libmachine.BaseDriver
+// func (d *Driver) GetIP() (string, error)
+
+// GetMachineName returns the name of the machine
+// inherited from  libmachine.BaseDriver
+// func (d *Driver) GetMachineName() string
+
+// GetBundleName() Returns the name of the unpacked bundle which was used to create this machine
+// inherited from  libmachine.BaseDriver
+// func (d *Driver) GetBundleName() (string, error)
```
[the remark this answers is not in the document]
cfergeau: Yes, this is in the  comment: `// inherited from  libmachine.BaseDriver`

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -0,0 +1,430 @@
...
+	cmd := exec.Command(vfkitPath, args...)
+	cmd.Stdout = log.StandardLogger().WriterLevel(log.DebugLevel)
+	cmd.Stderr = cmd.Stdout
+	cmd.Env = os.Environ()
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+	// cmd.Process.Pid
+	// cmd.Process
+	// cmd.Process.Kill()
```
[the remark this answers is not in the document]
cfergeau: More like implementation notes if I remember correctly, I removed these.

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -0,0 +1,430 @@
...
+		return state.Stopped, nil
+	}
+	return state.Running, nil
+	/*
+			if d.vzVirtualMachine == nil {
+				return state.Stopped, nil
+			}
+			return vzStateToState(d.vzVirtualMachine.State()), nil
+		return state.Stopped, nil
+	*/
```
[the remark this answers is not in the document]
cfergeau: Some remnants from a previous iteration of the driver, removed now.

PR #3047, opened 2022-03-04
[Integration] Remove unzip and also make use of bundle parameter for setup command

praveenkumar:


test/e2e/crcsuite/crcsuite.go
```diff
@@ -23,13 +23,13 @@ import (
...
-	CRCHome        string
-	CRCExecutable  string
-	bundleEmbedded bool
-	bundleName     string
-	bundleLocation string
-	pullSecretFile string
-	cleanupHome    bool
+	CRCHome                   string
+	CRCExecutable             string
+	bundleDownloadedAutomatic bool
```
jsliacan, line 28: Not sure about the name of this variable. `bundleDownloadAutomatic`, or `defaultBundle` or inverting true/false values with `userProvidedBundle` might be more intuitive.

test/e2e/crcsuite/crcsuite.go
```diff
@@ -114,10 +116,10 @@ func FeatureContext(s *godog.Suite) {
 
 		if bundleLocation == "" {
 			fmt.Println("Expecting the bundle to be embedded in the CRC executable.")
```
jsliacan, line 118: We should change this printout too (or remove it, although it's useful if stuff goes wrong to know what was "expected" without reading any code).

test/e2e/crcsuite/crcsuite.go
```diff
@@ -375,12 +377,27 @@ func StartCRCWithDefaultBundleSucceedsOrFails(expected string) error {
 	return err
 }
 
+func RunCRCSetupSucceedsOrFails(expected string) error {
```
jsliacan, line 380: Wouldn't it be simpler to, instead of this new function and a pattern matching cucumber expression, just insert something like the following 
```go
if !bundleDownloadedAutomatic && (command == "setup" || command == "start") {
	command = fmt.Sprintf("%s -b %s", command, bundleLocation)
}
```
above this line: https://github.com/code-ready/crc/blob/ff5d17656d1677301b42b78297e18e98f4a4618a/test/e2e/crcsuite/crcsuite.go#L427
Then we'd require only a handful of 1-word changes in the feature files.

test/integration/testsuite_test.go
```diff
@@ -54,13 +53,9 @@ var _ = BeforeSuite(func() {
...
 	Expect(err).NotTo(HaveOccurred())
 
-	// bundle location
-	bundlePath = "embedded"
-	if !versionInfo.Embedded {
-		bundlePath = os.Getenv("BUNDLE_PATH") // this env var should contain location of bundle or string "embedded"
-		if bundlePath != "embedded" {         // if real bundle
-			Expect(bundlePath).To(BeAnExistingFile())
-		}
+	bundlePath = os.Getenv("BUNDLE_PATH") // this env var should contain location of bundle or string "embedded"
```
jsliacan, line 56: Comment should end with `or empty string ""`?

test/integration/testsuite_test.go
```diff
@@ -54,13 +53,9 @@ var _ = BeforeSuite(func() {
...
+	if bundlePath != "" {
```
jsliacan, line 57: Which bundle is going to be automatically downloaded if `bundlepath == ""` is true?

jsliacan (commented):
It's good to remove all mentions of "embedded" - thanks.

test/integration/testsuite_test.go
praveenkumar: @jsliacan the one which is in the https://github.com/code-ready/crc/blob/main/pkg/crc/machine/bundle/constants.go here.

test/e2e/crcsuite/crcsuite.go
praveenkumar: @jsliacan I did try that but something was not working locally so I went with this path but I did try again and it is working.

jsliacan (approved):
Looks good, thanks!

PR #3053, opened 2022-03-07
docs: Add documentation for runtime presets

robin-owen:
## Solution/Idea

Add documentation for the use and configuration of container runtime presets.

Note that this PR also contains the introduction of the `{openshift}` and `{ocp}` attributes and fixes a few small errors I encountered while updating the content. These changes have all been isolated to their own commits.

## Proposed changes

1. Add a new module to conceptually define presets.
2. Add a new module for configuring the desired preset.
3. Modify the minimum system requirements to reflect the resources needed by each preset.
4. Modify `crc setup` documentation to include information about selecting the `podman` preset for Linux users.

### DIFF
--- a/docs/source/topics/con_about-presets.adoc
+++ b/docs/source/topics/con_about-presets.adoc
@@ -1,0 +1,18 @@
+[id="about-presets_{context}"]
+= About presets
+
+[role="_abstract"]
+{prod} presets represent a managed container runtime and the lower bounds of system resources required by the instance to run it.
+{prod} offers presets for {ocp} and the Podman container runtime.
+
+On {msw} and {mac}, the {prod} guided installer prompts you for your desired preset.
+On Linux, the {ocp} preset is selected by default.
+You can change this selection using the [command]`{bin} config` command before running the [command]`{bin} setup` command.
+You can change your selected preset from the system tray on {msw} and {mac} or from the command line on all supported operating systems.
+Only one preset can be active at a time.
+
+[role="_additional-resources"]
+.Additional resources
+
+* For more information about the minimum system requirements for each preset, see link:{crc-gsg-url}#minimum-system-requirements_gsg[Minimum system requirements].
+* For more information on changing the selected preset, see link:{crc-gsg-url}#changing-the-selected-preset_gsg[Changing the selected preset].
--- a/docs/source/topics/proc_changing-the-selected-preset.adoc
+++ b/docs/source/topics/proc_changing-the-selected-preset.adoc
@@ -1,0 +1,31 @@
+[id="changing-the-selected-preset_{context}"]
+= Changing the selected preset
+
+[role="_abstract"]
+You can change the container runtime used for the {prod} instance by selecting the desired preset.
+
+On {msw} and {mac}, you can change the selected preset using the system tray or command line interface.
+On Linux, use the command line interface.
+
+[IMPORTANT]
+====
+You cannot change the preset of an existing {prod} instance.
+Preset changes are only applied when a {prod} instance is created.
+To enable preset changes, you must delete the existing instance and start a new one.
+====
+
+.Procedure
+
+* Change the selected preset from the command line:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config preset __<name>__
+----
++
+Valid preset names are `openshift` for {ocp} and `podman` for the Podman container runtime.
+
+[role="_additional-resources"]
+.Additional resources
+
+* For more information about the minimum system requirements for each preset, see link:{crc-gsg-url}#minimum-system-requirements_gsg[Minimum system requirements].
--- a/docs/source/topics/ref_minimum-system-requirements.adoc
+++ b/docs/source/topics/ref_minimum-system-requirements.adoc
@@ -1,27 +1,36 @@
 [id="minimum-system-requirements_{context}"]
 = Minimum system requirements
 
+[role="_abstract"]
 {prod} has the following minimum hardware and operating system requirements.
 
 [id="minimum-system-requirements-hardware_{context}"]
 == Hardware requirements
 
-{prod} requires the following system resources:
+{prod} is supported only on AMD64 and Intel 64 processor architectures.
+{prod} does not support the ARM-based M1 architecture.
+{prod} does not support nested virtualization.
+
+Depending on the desired container runtime, {prod} requires the following system resources:
+
+=== For {ocp}
 
 * 4 physical CPU cores
 * 9 GB of free memory
 * 35 GB of storage space
 
-{prod} is supported only on AMD64 and Intel 64 processor architectures.
-{prod} does not support the ARM-based M1 architecture.
-{prod} does not support nested virtualization.
-
 [NOTE]
 ====
-The OpenShift cluster requires these minimum resources to run in the {prod} instance.
+The {ocp} cluster requires these minimum resources to run in the {prod} instance.
 Some workloads may require more resources.
 To assign more resources to the {prod} instance, see link:{crc-gsg-url}#configuring-the-instance_gsg[Configuring the instance].
 ====
+
+=== For the Podman container runtime
+
+* 2 physical CPU cores
+* 2 GB of free memory
+* 35 GB of storage space
 
 [id="minimum-system-requirements-operating-system_{context}"]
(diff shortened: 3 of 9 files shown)


docs/source/topics/ref_minimum-system-requirements.adoc
```diff
@@ -1,28 +1,37 @@
...
 
+[role="_abstract"]
 {prod} has the following minimum hardware and operating system requirements.
 
 [id="minimum-system-requirements-hardware_{context}"]
 == Hardware requirements
 
-{prod} requires the following system resources:
+{prod} is supported only on AMD64 and Intel 64 processor architectures.
+Support for the ARM-based M1 architecture is in development.
```
robin-owen, line 11: @code-ready/crc-team Please double-check the wording here for me -- I'm not particularly aware of the developments in this area. Should we mention it at all at this point?

docs/source/topics/con_about-presets.adoc
```diff
@@ -0,0 +1,18 @@
+[id="about-presets_{context}"]
+= About presets
+
+[role="_abstract"]
+{prod} presets represent a managed container runtime and the lower bounds of system resources required by the instance to run it.
+{prod} offers presets for {ocp} and the Podman container runtime.
```
gbraad, line 6: no variable for Podman?

docs/source/topics/proc_changing-the-selected-preset.adoc
```diff
@@ -0,0 +1,31 @@
...
+You can change the container runtime used for the {prod} instance by selecting the desired preset.
+
+On {msw} and {mac}, you can change the selected preset using the system tray or command line interface.
+On Linux, use the command line interface.
+
+[IMPORTANT]
+====
+You cannot change the preset of an existing {prod} instance.
+Preset changes are only applied when a {prod} instance is created.
+To enable preset changes, you must delete the existing instance and start a new one.
```
gbraad, line 14: Not sure how this will stand out in the documentation. It briefly touches on 'delete'. Will have to check the rendered version.

docs/source/topics/ref_minimum-system-requirements.adoc
gbraad: We have a Q2 goal to deliver M1 support, but this might only be possible for the `{podman}` preset.
Not sure we should make a 'promise' here as the `{ocp}` options depends on availability of `{rhcos}` that can run on Aarch64.

gbraad (dismissed):
Overall OK, but perhaps the M1 support should not be mentioned
This is better suited for a blog post.

docs/source/topics/ref_minimum-system-requirements.adoc
praveenkumar: I think we should avoid M1 arch as of now since vfkit part is still not in the branch and also we might need some condition for avoid user to run openshift on M1.

robin-owen: Okay, I reverted the change to that line. We're back to the original docs statement of "not supported" for the time being.

docs/source/topics/con_about-presets.adoc
robin-owen: I didn't see the need to create one. The attributes for OpenShift largely exist to avoid unfortunate typos when the "f" key is missed. ;)

If we foresee the name of Podman itself (or the `podman` preset) to change, we should probably create an attribute for it.

docs/source/topics/con_about-presets.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+On {msw} and {mac}, the {prod} guided installer prompts you for your desired preset.
+On Linux, the {ocp} preset is selected by default.
+You can change this selection using the [command]`{bin} config` command before running the [command]`{bin} setup` command.
+You can change your selected preset from the system tray or command line.
```
bburt-rh, line 11: The system tray doesn't exist on macOS, and I believe refers to a MS Windows-specific UI element. Is there a way to specify this fact here? Maybe something like `from the system tray on {msw} or from the command line on any supported operating system`?

docs/source/topics/proc_changing-the-selected-preset.adoc
```diff
@@ -0,0 +1,31 @@
+[id="changing-the-selected-preset_{context}"]
+= Changing the selected preset
+
+[role="_abstract"]
+You can change the container runtime used for the {prod} instance by selecting the desired preset.
+
+On {msw} and {mac}, you can change the selected preset using the system tray or command line interface.
```
bburt-rh, line 7: See comment above about OS-specific UI elements.

docs/source/topics/proc_changing-the-selected-preset.adoc
```diff
@@ -0,0 +1,31 @@
...
+To enable preset changes, you must delete the existing instance and start a new one.
+====
+
+.Procedure
+
+* Change the selected preset from the command line:
++
+[subs="+quotes,attributes"]
+----
+$ {bin} config preset _<name>_
```
bburt-rh, line 23:
```suggestion
$ {bin} config preset __<name>__
```
Per our SSG, I think that variables need two underline characters to render them as italics. At least that's how I've done it. Does a single underline character work?

docs/source/topics/ref_minimum-system-requirements.adoc
```diff
@@ -1,28 +1,37 @@
...
 {prod} has the following minimum hardware and operating system requirements.
 
 [id="minimum-system-requirements-hardware_{context}"]
 == Hardware requirements
 
-{prod} requires the following system resources:
+{prod} is supported only on AMD64 and Intel 64 processor architectures.
+{prod} does not support the ARM-based M1 architecture.
+{prod} does not support nested virtualization.
+
```
bburt-rh, line 13: Consider reformatting the requirements using a bulleted list.

docs/source/topics/ref_minimum-system-requirements.adoc
```diff
@@ -1,28 +1,37 @@
...
+{prod} requires the following system resources depending on the desired container runtime:
```
bburt-rh, line 14:
```suggestion
{prod} requires the following system resources, depending on the desired container runtime:
```

docs/source/topics/proc_changing-the-selected-preset.adoc
robin-owen: Ah, good catch! Yes, the single underlines do work, but doubling up on them is actually much more robust and should be the general practice. I just forgot. ;)

docs/source/topics/ref_minimum-system-requirements.adoc
robin-owen: I prefer the latter, it sounds a bit cleaner to me. Thank you!

docs/source/topics/con_about-presets.adoc
robin-owen: Does it have another name on macOS other than "system tray"? CodeReady Containers has GUI applications which sit in the "tray" area of both Windows and macOS. The only system that does not have this functionality right now is Linux.

docs/source/topics/proc_changing-the-selected-preset.adoc
robin-owen: I placed the content under an "important" admonition to (hopefully) encourage careful reading of it. Do you think we should make this more visible somehow?

docs/source/topics/con_about-presets.adoc
robin-owen: Aha, it's called the Menu Bar in macOS. The question now is whether we go for complete technical accuracy or if we target the reader's understanding. I'm fairly confident that macOS users will know what "system tray" refers to, but I may be incorrect.

bburt-rh: Per our Slack discussion, this is OK to leave as-is

docs/source/topics/proc_changing-the-selected-preset.adoc
bburt-rh: Per our Slack discussion, this is OK to leave as-is

bburt-rh:
Approved

gbraad (approved):

PR #3059, opened 2022-03-10
Make user mode networking default for dev build

praveenkumar:
As of now we support system mode and user mode networking for all platform and in
windows/mac user mode is default for releases and system mode for dev builds. This
create bit confusion and also dev forget to test what is going to delivered with release
so this PR will make it default for both release and dev side.

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"runtime"
 
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/network"
@@ -119,7 +120,7 @@ func GetPreset(config Storage) preset.Preset {
 }
 
 func defaultNetworkMode() network.Mode {
-	if version.IsInstaller() {
+	if runtime.GOOS != "linux" {
 		return network.UserNetworkingMode
 	}
 	return network.SystemNetworkingMode


pkg/crc/config/settings.go
```diff
@@ -96,12 +94,8 @@ func RegisterSettings(cfg *Config) {
...
 	cfg.AddSetting(ExperimentalFeatures, false, ValidateBool, SuccessfullyApplied,
 		"Enable experimental features (true/false, default: false)")
-
-	if !version.IsInstaller() {
-		cfg.AddSetting(NetworkMode, string(defaultNetworkMode()), network.ValidateMode, network.SuccessfullyAppliedMode,
-			fmt.Sprintf("Network mode (%s or %s)", network.UserNetworkingMode, network.SystemNetworkingMode))
-	}
-
+	cfg.AddSetting(NetworkMode, string(defaultNetworkMode()), network.ValidateMode, network.SuccessfullyAppliedMode,
+		fmt.Sprintf("Network mode (%s or %s)", network.UserNetworkingMode, network.SystemNetworkingMode))
```
cfergeau, line 98: The `network-mode` option has been removed from macOS/Windows releases for a while now. This hunk readds the option to releases, I don't think that's desirable.

pkg/crc/config/settings.go
```diff
@@ -145,14 +139,14 @@ func GetPreset(config Storage) preset.Preset {
 }
 
 func defaultNetworkMode() network.Mode {
-	if version.IsInstaller() {
+	if runtime.GOOS != "linux" {
 		return network.UserNetworkingMode
 	}
```
cfergeau, line 144: This is the hunk which is doing what the commit log describes (using the same default networking mode on release/non-release builds).

pkg/crc/config/settings.go
```diff
@@ -145,14 +139,14 @@ func GetPreset(config Storage) preset.Preset {
...
 	return network.SystemNetworkingMode
 }
 
 func GetNetworkMode(config Storage) network.Mode {
-	if version.IsInstaller() {
+	if runtime.GOOS != "linux" {
 		return network.UserNetworkingMode
```
cfergeau, line 150: This hunk can be dropped together with the first one. If you change this, this means non-release builds will always be using usermode networking, even if there is a configuration option for it.

cfergeau (changes requested):
Re-reading the commit log, I'm not 100% sure what the goal of the commit is:

> Remove system mode networking from main build

or 

> this PR will make it default for both release and dev side

This PR does a bit of both, it forces usermode networking on macos/windows, but also readds the configuration option which was removed, even if setting it will be a no-op.

If you remove system networking from macos/windows, there are more changes to add on top of that PR to remove the related code (at least the preflights).

cfergeau (approved):

PR #3066, opened 2022-03-14
Mac Preflight: Check if plist file exist before removing it

praveenkumar:
This patch will first check if the plist file exist before error out
for check function and also remove it in fix function. This will fix
following error for preflight on mac.
```
INFO Checking if old launchd config for tray autostart exists
DEBU force trigger cleanup to remove old launchd config for tray
force trigger cleanup to remove old launchd config for tray
```

### DIFF
--- a/pkg/crc/preflight/preflight_darwin.go
+++ b/pkg/crc/preflight/preflight_darwin.go
@@ -9,6 +9,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/preset"
 	crcpreset "github.com/code-ready/crc/pkg/crc/preset"
+	crcos "github.com/code-ready/crc/pkg/os"
 )
 
 // SetupHost performs the prerequisite checks and setups the host to run the cluster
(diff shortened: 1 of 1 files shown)


anjannath (approved):

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -70,13 +71,23 @@ var trayLaunchdCleanupChecks = []Check{
...
+		fixDescription: "Removing old launchd config for tray autostart",
+		fix: func() error {
+			var err error
+			if crcos.FileExists(filepath.Join(constants.GetHomeDir(), "Library", "LaunchAgents", "crc.tray.plist")) {
+				err = os.Remove(filepath.Join(constants.GetHomeDir(), "Library", "LaunchAgents", "crc.tray.plist"))
+			}
+			if crcos.FileExists(filepath.Join(constants.GetHomeDir(), "Library", "LaunchAgents", "crc.daemon.plist")) {
+				err = os.Remove(filepath.Join(constants.GetHomeDir(), "Library", "LaunchAgents", "crc.daemon.plist"))
+			}
+			return err
```
cfergeau, line 89: I can't follow what is expected from 'err' here?
If removal of `crc.tray.plist` fails but removal of `crc.daemon.plist` succeeds, we ignore the error.
If removal of `crc.tray.list` succeeds but removal of `crc.daemon.plist` fails, we return the error.
Is this the intended behaviour?

PR #3068, opened 2022-03-16
Initial rename

gbraad:


README.adoc
```diff
@@ -1,4 +1,4 @@
-= Red Hat CodeReady Containers - OpenShift 4 on your Laptop
+= Container Runtime Control - OpenShift 4 and Podman on your Laptop
```
cfergeau, line 1: The new naming feels contrived and artificial to me :) Why not just call it `CRC` without having an official long name? 
`Running OpenShift 4 and Podman containers on your Laptop` reads better if it's not too long.

praveenkumar: To me `Containter Runtime Control` is better than `CodeReady Containers` but what are we controlling here since we only ship single container runtime for both offering (openshift/podman). If we can make `crc` as it without any abbreviation then it would be great.

gbraad: We settled on referring to `crc` and the ability to `crc runs containers`. It will not have any direct meaning to the name.

Note: the CodeReady branding is to be removed and can not be referred to anymore.

README.adoc
```diff
@@ -13,16 +13,17 @@ image:https://ci.centos.org/buildStatus/icon?job=codeready-crc-master["CentOS CI
...
+This project is focused on bringing a minimal http://github.com/openshift/origin[OpenShift 4.x] cluster or Podman VM to your local laptop or desktop computer
+for development and testing purposes.
 
 If you are looking for a solution for running OpenShift 3.x, you will need tools such as `oc cluster up`, http://github.com/minishift/minishift[Minishift] or https://developers.redhat.com/products/cdk/overview/[CDK].
 
 [[usage-data]]
 == Usage data
 
-The first time CodeReady Containers is run, you will be asked to opt-in to Red Hat's telemetry collection program.
+The first time the tool is run, you will be asked to opt-in to Red Hat's telemetry collection program.
```
gbraad, line 24: s/the tool/`crc`

PR #3072, opened 2022-03-16
[WIP] Mac: Use launchd to start the daemon as part of setup

praveenkumar:
This PR adds a launchd plist file which use socket activation to
start the daemon process. This plist file created as part of setup
step and loaded using `launchd load`. When a process require the
`~/.crc/crc-http.sock` then launchd should start the daemon process.


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -700,50 +700,3 @@ func getCPUFlags() (string, error) {
 	}
 	return flags, nil
 }
-
-func runtimeExecutablePath() (string, error) {
```
cfergeau, line 704: Can you split the code movement to its own commit?

cmd/crc/cmd/daemon_darwin.go
```diff
@@ -7,8 +7,11 @@ import (
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	crcversion "github.com/code-ready/crc/pkg/crc/version"
+	"github.com/code-ready/crc/pkg/os/launchd"
 )
 
+const daemonAgentLabel = "com.github.crc.daemon"
```
cfergeau, line 13: Not sure why the `const` is duplicated twice. `dev.crc.daemon` or `com.redhat.crc.daemon` would be better labels in my opinion.

cfergeau (commented):
Is this using socket activation as written in the commit log? Or did you switch to autostarting the daemon with the user session?

praveenkumar:
> Is this using socket activation as written in the commit log? Or did you switch to autostarting the daemon with the user session?

@cfergeau as of now it is socket activation and it does hang first time, I just put WIP to get some initial feedback and also looking into a way to get ownership of socket from launchd process to daemon.

cfergeau:
> > Is this using socket activation as written in the commit log? Or did you switch to autostarting the daemon with the user session?
> 
> @cfergeau as of now it is socket activation and it does hang first time, I just put WIP to get some initial feedback and also looking into a way to get ownership of socket from launchd process to daemon.

Ok, was not clear from the PR description that it has a known blocker, and that it's only preliminary. So I assumed you dropped the problematic parts of the PR for now :)

cmd/crc/cmd/daemon_darwin.go
```diff
@@ -30,13 +30,20 @@ func httpListener() (net.Listener, error) {
 }
 
 func checkIfDaemonIsRunning() (bool, error) {
+	if launchd.PlistExists(constants.DaemonAgentLabel) {
+		/* detect if the daemon is being started by launchd,
+		* and socket activation is in use. In this scenario,
```
cfergeau, line 35: Since socket activation is not being used with launchd, we don't need this code?

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -242,3 +243,68 @@ func stopCRCHyperkitProcess() error {
...
+func checkIfDaemonPlistFileExists() error {
+	daemonConfig, err := getDaemonConfig()
+	if err != nil {
+		return err
+	}
+	// If plist file content match with available plist file
+	if err := launchd.CheckPlist(*daemonConfig); err != nil {
+		return err
+	}
+	// If agent is running
```
cfergeau, line 271: This is also unneeded.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -242,3 +243,68 @@ func stopCRCHyperkitProcess() error {
...
+	if !launchd.AgentRunning(daemonConfig.Label) {
+		return fmt.Errorf("agent for %s label not running", daemonConfig.Label)
```
cfergeau, line 273: `launchd agent '%s' is not running"?

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -242,3 +243,68 @@ func stopCRCHyperkitProcess() error {
...
+	return launchd.RemovePlist(constants.DaemonAgentLabel)
+}
+
+func fixPlistFileExists(agentConfig launchd.AgentConfig) error {
+	logging.Debugf("Creating plist for %s", agentConfig.Label)
+	err := launchd.CreatePlist(agentConfig)
+	if err != nil {
+		return err
+	}
+	// load plist
```
cfergeau, line 299: redundant comment.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -242,3 +243,68 @@ func stopCRCHyperkitProcess() error {
...
+	if err := launchd.LoadPlist(agentConfig.Label); err != nil {
+		logging.Debug("failed while loading plist:", err.Error())
+		return err
+	}
+	// restart agent
```
cfergeau, line 304: and this is also redundant, this can be removed.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -242,3 +243,68 @@ func stopCRCHyperkitProcess() error {
...
+	if err := launchd.RestartAgent(agentConfig.Label); err != nil {
+		logging.Debugf("failed while restarting plist: %s", err.Error())
```
cfergeau, line 306: `failed to restart launch agent '%s': %v` ?

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -106,6 +106,18 @@ var resolverPreflightChecks = []Check{
 	},
 }
 
+var daemonLaunchdChecks = []Check{
+	{
+		configKeySuffix:    "check-daemon-launchd-plist",
+		checkDescription:   "Checking if crc daemon plist file present and loaded",
```
cfergeau, line 112: `*is* present and loaded`

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -106,6 +106,18 @@ var resolverPreflightChecks = []Check{
...
+		check:              checkIfDaemonPlistFileExists,
+		fixDescription:     "Adding crc daemon plist file and loading it",
+		fix:                fixDaemonPlistFileExists,
+		cleanupDescription: "Unloading and removing the daemom plist file",
+		cleanup:            removeDaemonPlistFile,
```
cfergeau, line 117: Not sure we should be emphasizing the plist file in these messages. What matters is that we are setting up launchd to start the daemon with the session?

pkg/os/launchd/launchd_darwin.go
```diff
@@ -17,36 +17,49 @@ import (
...
-			<key>ProcessType</key>
-			<string>Interactive</string>
+			<key>SockFamily</key>
+			<string>Unix</string>
+			<key>SockPathName</key>
+			<string>{{ .SocketPath }}</string>
+			<key>SockPathMode</key>
+			<integer>493</integer>
+			<key>SockType</key>
+			<string>stream</string>
```
cfergeau, line 48: Isn't this configuring socket activation for the daemon, which we don't want to use yet?

praveenkumar:
@cfergeau why `has-to-be-in-release` label removed?

cfergeau:
> @cfergeau why `has-to-be-in-release` label removed?

PR seems not fully finished yet, if it's ready in time for the release, then it gets in, otherwise I don't know? And this label has always been weird to me, naming is very pushy, and it's odd that it's the person filing the PR who makes the decision about it.

praveenkumar:
> > @cfergeau why `has-to-be-in-release` label removed?
> 
> PR seems not fully finished yet, if it's ready in time for the release, then it gets in, otherwise I don't know? 

@cfergeau It is not a WIP anymore, latest suggestions will be worked on and we have around 2 weeks from now so should have enough time. We reduce the scope of this PR so instead of socket activation, it will be start the agent as part of setup.

> And this label has always been weird to me, naming is very pushy, and it's odd that it's the person filing the PR who makes the decision about it.

I think person who fill the PR makes the decision around it because this is what committed for up coming release and assigned to that person. It will also make sure required folks do the review since it will be part of the release. Who would be the best to take this decision if not the person filing the PR?

pkg/crc/preflight/preflight_darwin.go
praveenkumar: @cfergeau yes we are using launchd to start the daemon but plist is driving force something like in linux the unit file for systemd operations.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -242,3 +243,64 @@ func stopCRCHyperkitProcess() error {
...
+	daemonConfig, err := getDaemonConfig()
+	if err != nil {
+		return err
+	}
+	if err := launchd.CheckPlist(*daemonConfig); err != nil {
+		return err
+	}
+	if !launchd.AgentRunning(daemonConfig.Label) {
+		return fmt.Errorf("launchd agent '%s' is not running", daemonConfig.Label)
+	}
```
cfergeau, line 272: We also need to restart the daemon on upgrades, I don't know if it's possible to add version information to the launchd plist? Otherwise we could check the running daemon version, or maybe even always restart it during setup?

cfergeau (commented):
Could you rework the commti log a bit, to start by explaining what this commit is doing first (installs a plist file so that the daemon is started at session startup (I guess?)), and then move the explanations why we aren't using socket activation after this?

Now it's a bit hard to read as there's a long paragraph explaining what we don't do, but there is no clear/easy to find explanation what it's doing :)

pkg/crc/preflight/preflight_checks_darwin.go
praveenkumar: @cfergeau we can use `EnvironmentVariable` key to do `CRC_VERSION` as env which always make sure specific version of daemon running for specific crc version during setup.

```
     EnvironmentVariables <dictionary of strings>
     This optional key is used to specify additional environmental variables
     to be set before running the job.
```

praveenkumar:
@cfergeau This is also updated ( commit log also changed as requested).

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -242,3 +243,65 @@ func stopCRCHyperkitProcess() error {
...
+}
+
+func removeDaemonPlistFile() error {
+	if err := launchd.UnloadPlist(constants.DaemonAgentLabel); err != nil {
+		return err
+	}
+	return launchd.RemovePlist(constants.DaemonAgentLabel)
+}
+
+func fixPlistFileExists(agentConfig launchd.AgentConfig) error {
```
cfergeau, line 292: Not sure it's worth its own helper since it is used only once, and `fixDaemonPlistFileExists` is short.

pkg/os/launchd/launchd_darwin.go
```diff
@@ -18,36 +18,42 @@ import (
...
+</plist>
+`
 )
 
 // AgentConfig is struct to contain configuration for agent plist file
 type AgentConfig struct {
 	Label          string
 	ExecutablePath string
 	StdOutFilePath string
+	StdErrFilePath string
```
cfergeau, line 54: I would just reuse StdOutFilePath since we are setting both to the same file anyway.

cfergeau (approved):

PR #3075, opened 2022-03-21
Use podman system connection to add remote uri to podman config

praveenkumar:
As of now user have to do `eval $(crc podman-env)` to export podman
connection specific environment variable to current shell session.
Podman cli have `system connection` command which manage the ssh
destination information for podman configuration which consumed
by podman client.

With this PR now if a user have podman-remote client binary to path
then after `crc start` they can directly use podman commands without
executing current `eval` command. `podman-env` command option still
there to support backward compatibility and also exposing moby specific
environment variable. It will still make sense if user don't have podman
remote binary in the path.

```
✗ ./crc start
[...]
podman runtime is currently configured in rootless mode. If your containers
require root permissions, or if you run into compatibility
issues with non-podman clients, you can switch using the following commands:

  $ eval $(crc podman-env --root)
  $ podman COMMAND

✗ podman system connection ls
Name                         URI                                                         Identity                                    Default
crc                          ssh://core@127.0.0.1:2222/run/user/1000/podman/podman.sock  /Users/prkumar/.crc/machines/crc/id_ecdsa   true
crc-root                     ssh://core@127.0.0.1:2222/run/podman/podman.sock            /Users/prkumar/.crc/machines/crc/id_ecdsa   false

✗ ./crc podman-env --root
export PATH="/Users/prkumar/.crc/bin/oc:$PATH"
export DOCKER_HOST="unix:///Users/prkumar/.crc/machines/crc/docker.sock"

✗ podman system connection ls
Name                         URI                                                         Identity                                    Default
crc                          ssh://core@127.0.0.1:2222/run/user/1000/podman/podman.sock  /Users/prkumar/.crc/machines/crc/id_ecdsa   false
crc-root                     ssh://core@127.0.0.1:2222/run/podman/podman.sock            /Users/prkumar/.crc/machines/crc/id_ecdsa   true

✗ podman info
host:
  arch: amd64
  buildahVersion: 1.23.1
  cgroupControllers:
```

### DIFF
--- a/cmd/crc/cmd/podman_env.go
+++ b/cmd/crc/cmd/podman_env.go
@@ -38,9 +38,9 @@ func runPodmanEnv() error {
 		return err
 	}
 
-	socket := "/run/user/1000/podman/podman.sock"
+	socket := constants.RootlessPodmanSocket
 	if root {
-		socket = "/run/podman/podman.sock"
+		socket = constants.RootfulPodmanSocket
 	}
 	fmt.Println(shell.GetPathEnvString(userShell, constants.CrcOcBinDir))
 	fmt.Println(shell.GetEnvString(userShell, "CONTAINER_SSHKEY", connectionDetails.SSHKeys[0]))
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -23,6 +23,7 @@ import (
 	"github.com/code-ready/crc/pkg/crc/machine/types"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/oc"
+	"github.com/code-ready/crc/pkg/crc/podman"
 	crcPreset "github.com/code-ready/crc/pkg/crc/preset"
 	"github.com/code-ready/crc/pkg/crc/services"
 	"github.com/code-ready/crc/pkg/crc/services/dns"
@@ -329,6 +330,14 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 
 		if err := updateCockpitConsoleBearerToken(sshRunner); err != nil {
 			return nil, fmt.Errorf("Failed to rotate bearer token for cockpit webconsole: %w", err)
+		}
+
+		c, err := client.ConnectionDetails()
+		if err != nil {
+			return nil, err
+		}
+		if err := addPodmanSystemConnections(c); err != nil {
+			return nil, err
 		}
 
 		return &types.StartResult{
@@ -755,5 +764,20 @@ func updateCockpitConsoleBearerToken(sshRunner *crcssh.Runner) error {
 		return fmt.Errorf("failed to change ownership of cockpit-bearer-token to core user: %w", err)
 	}
 
+	return nil
+}
+
+func addPodmanSystemConnections(c *types.ConnectionDetails) error {
+	rootlessURI := fmt.Sprintf("ssh://%s@%s:%d%s", c.SSHUsername, c.IP, c.SSHPort, constants.RootlessPodmanSocket)
+	rootfulURI := fmt.Sprintf("ssh://%s@%s:%d%s", c.SSHUsername, c.IP, c.SSHPort, constants.RootfulPodmanSocket)
+	if err := podman.AddRootlessSystemConnection(c.SSHKeys[0], rootlessURI); err != nil {
+		return err
+	}
+	if err := podman.MakeRootlessSystemConnectionDefault(); err != nil {
+		return err
+	}
+	if err := podman.AddRootfulSystemConnection(c.SSHKeys[0], rootfulURI); err != nil {
+		return err
+	}
 	return nil
 }
--- a/pkg/crc/podman/podman.go
+++ b/pkg/crc/podman/podman.go
@@ -1,0 +1,65 @@
+package podman
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	crcos "github.com/code-ready/crc/pkg/os"
+)
+
+const (
+	rootlessConn = "crc"
+	rootfulConn  = "crc-root"
+)
+
+var podmanExecutablePath = filepath.Join(constants.CrcOcBinDir, constants.PodmanRemoteExecutableName)
+
+func run(args ...string) (string, string, error) {
+	return crcos.RunWithDefaultLocale(podmanExecutablePath, args...)
+}
+
+func addSystemConnection(name, identity, uri string) error {
+	if _, stderr, err := run("system", "connection", "add", "--identity", identity, name, uri); err != nil {
+		return fmt.Errorf("failed to add podman system connection %v: %s", err, stderr)
+	}
+	return nil
+}
+
+func AddRootlessSystemConnection(sshIdentity, uri string) error {
+	return addSystemConnection(rootlessConn, sshIdentity, uri)
+}
+
+func AddRootfulSystemConnection(sshIdentity, uri string) error {
+	return addSystemConnection(rootfulConn, sshIdentity, uri)
+}
+
+func removeSystemConnection(name string) error {
+	if _, stderr, err := run("system", "connection", "remove", name); err != nil {
+		return fmt.Errorf("failed to remove podman system connection %v: %s", err, stderr)
+	}
+	return nil
+}
+
+func RemoveRootlessSystemConnection() error {
+	return removeSystemConnection(rootlessConn)
+}
+
+func RemoveRootfulSystemConnection() error {
+	return removeSystemConnection(rootfulConn)
+}
+
+func makeSystemConnectionDefault(name string) error {
+	if _, stderr, err := run("system", "connection", "default", name); err != nil {
+		return fmt.Errorf("failed to make %s podman system connection as default %v: %s", name, err, stderr)
+	}
+	return nil
+}
+
+func MakeRootlessSystemConnectionDefault() error {
+	return makeSystemConnectionDefault(rootlessConn)
+}
+
+func MakeRootfulSystemConnectionDefault() error {
+	return makeSystemConnectionDefault(rootfulConn)
+}
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -30,6 +30,8 @@ const (
 	DefaultContext            = "admin"
 	DaemonHTTPEndpoint        = "http://unix/api"
 	DefaultPodmanNamedPipe    = `\\.\pipe\crc-podman`
+	RootlessPodmanSocket      = "/run/user/1000/podman/podman.sock"
+	RootfulPodmanSocket       = "/run/podman/podman.sock"
 
 	VSockGateway = "192.168.127.1"
 	VsockSSHPort = 2222
--- a/pkg/crc/machine/delete.go
+++ b/pkg/crc/machine/delete.go
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/code-ready/crc/pkg/crc/podman"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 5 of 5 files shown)


pkg/crc/podman/podman.go
```diff
@@ -0,0 +1 @@
+package podman
```
cfergeau, line 1: This belongs in the next commit

cmd/crc/cmd/podman_env.go
```diff
@@ -32,24 +33,16 @@ func runPodmanEnv() error {
 	if err := checkIfMachineMissing(client); err != nil {
 		return err
 	}
-
-	connectionDetails, err := client.ConnectionDetails()
-	if err != nil {
+	p := podman.UsePodman()
```
cfergeau, line 36: The naming `UsePodman()` is weird (I know it's used elsewhere for OC/...), I was thinking this was testing if something is using podman. Maybe just `podman.Config()`? And one letter variable are quite bad for readability.

cmd/crc/cmd/podman_env.go
```diff
@@ -32,24 +33,16 @@ func runPodmanEnv() error {
...
+	if err := p.MakeRootlessSystemConnectionDefault(); err != nil {
```
cfergeau, line 37: I don't expect `podman-env` to change my system configuration, I only expect it to set some environment variable and that's it. Feels like we are doing too much (setting both env vars and calling `podman system config` when only one of these is enough to achieve what we need).

cmd/crc/cmd/start.go
```diff
@@ -248,9 +248,10 @@ Use the 'oc' command line interface:
...
   {{ .CommandLinePrefix }} oc login -u {{ .ClusterConfig.DeveloperCredentials.Username }} {{ .ClusterConfig.URL }}
 `
-	startTemplateForPodman = `podman runtime is now running.
+	startTemplateForPodman = `podman runtime is currently configured in rootless mode. If your containers
+require root permissions, or if you run into compatibility
+issues with non-podman clients, you can switch using the following commands:
 
-Use the 'podman' command line interface:
   {{ .CommandLinePrefix }} {{ .EvalCommandLine }}
   {{ .CommandLinePrefix }} {{ .PodmanRemote }} COMMAND
```
cfergeau, line 256: By default, we don't need to do anything  because we configured a podman connection, but to change to root, we don't document how to change the podman connection as described in the commit log, but we document something entirely different? How do I switch back to the rootless mode (I mean, if I don't know how this works at all, ,and only followed the instructions from this template?)

pkg/crc/machine/start.go
```diff
@@ -331,6 +332,20 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
...
+		}
+		rootlessURI := fmt.Sprintf("ssh://%s@%s:%d%s", c.SSHUsername, c.IP, c.SSHPort, constants.RootlessPodmanSocket)
+		rootfulURI := fmt.Sprintf("ssh://%s@%s:%d%s", c.SSHUsername, c.IP, c.SSHPort, constants.RootfulPodmanSocket)
+		if err := p.AddRootlessSystemConnection(c.SSHKeys[0], rootlessURI); err != nil {
+			return nil, err
+		}
+		if err := p.AddRootfulSystemConnection(c.SSHKeys[0], rootfulURI); err != nil {
+			return nil, err
+		}
+
```
cfergeau, line 348: This deserves its own helper (`addPodmanSystemConnections()` or something like this) to make it obvious to the reader what this is about without overwhelming it with implementation details.

pkg/crc/podman/podman.go
```diff
@@ -1 +1,96 @@
...
+)
+
+type Config struct {
+	Runner               crcos.CommandRunner
+	PodmanExecutablePath string
+}
+
+func UsePodman() Config {
+	return Config{
+		Runner:               crcos.NewLocalCommandRunner(),
```
cfergeau, line 23: Not sure why we need a runner, or to kind of implement the `Runner` interface in here? Looks like it would be good enough to use directly `RunWithDefaultLocale`? I don't mind using a `Runner`, but I'm not sure why we have public `RunPodmanCommand/RunPodmanCommandPrivate/...?

cfergeau (commented):
The commit log for the 2nd commit is missing a shortlog.
What happens on linux when this is used, does this affect only podman-remote, or is `podman` itself forcefully switched to use a remote runtime? I assume when the crc VM is not running this errors out?

pkg/crc/podman/podman.go
praveenkumar: Done

pkg/crc/machine/start.go
praveenkumar: Done.

praveenkumar:
> What happens on linux when this is used, does this affect only podman-remote, or is `podman` itself forcefully switched to use a remote runtime? I assume when the crc VM is not running this errors out?

In linux it is only affect `podman-remote` nothing around `podman` change with this config.

praveenkumar:
@cfergeau I removed start message and `podman-env` from this PR, I think we need to discuss what it should be for end user to educate them how to make `rootful` one as default for system connection setting (A doc or start message template or something else).

cfergeau:
> @cfergeau I removed start message and `podman-env` from this PR, I think we need to discuss what it should be for end user to educate them how to make `rootful` one as default for system connection setting (A doc or start message template or something else).

Yup, need to decide how we want the cmdline UI to look like. Do we need to keep `podman-env` at all? Or could we just tell the user at the end of start that rootless podman is ready for use by calling `podman-remote`, and if they want to switch to rootful podman, they can use `podman-remote system connection ...` ?

praveenkumar:
> > @cfergeau I removed start message and `podman-env` from this PR, I think we need to discuss what it should be for end user to educate them how to make `rootful` one as default for system connection setting (A doc or start message template or something else).
> 
> Yup, need to decide how we want the cmdline UI to look like. Do we need to keep `podman-env` at all? Or could we just tell the user at the end of start that rootless podman is ready for use by calling `podman-remote`, and if they want to switch to rootful podman, they can use `podman-remote system connection ...` ?

@cfergeau Yes on the UI side I will try to create new PR once we decide what we actually want. We still need `podman-env` to show docker host env but again this is again can be part of `crc start` message where we tell user around rootless/rootful` ?

cfergeau:
> We still need podman-env to show docker host env 

If `podman-env` now is docker-only, its name is confusing :(

praveenkumar:
> > We still need podman-env to show docker host env
> 
> If `podman-env` now is docker-only, its name is confusing :(

@cfergeau Right now we are not changing anything around `podman-env` so as part of new PR/issue we can check what should be the name. Also `podman-env` still set the podman binary to the path.

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/code-ready/crc/pkg/crc/constants"
+	crcos "github.com/code-ready/crc/pkg/os"
+)
+
+const (
+	rootless = "crc"
```
cfergeau, line 12: `rootlessConn` or `rootlessConnectionName` or ...

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,82 @@
...
+
+type Config struct {
+	PodmanExecutablePath string
+}
+
+func NewConfig() Config {
+	return Config{
+		PodmanExecutablePath: filepath.Join(constants.CrcOcBinDir, constants.PodmanRemoteExecutableName),
+	}
+}
```
cfergeau, line 24: At this point, `Config` is not needed, whatever needs to execute podman can build the path as required (or this can be a global variable in this module if you prefer). This should make the code simpler.

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (podman Config) AddRootfulSystemConnection(identity, uri string) error {
+	return podman.addSystemConnection(false, rootful, identity, uri)
+}
+
+func (podman Config) removeSystemConnection(name string) error {
+	// podman system connection remove crc
+	args := []string{"system", "connection", "remove", name}
+	if _, stderr, err := podman.runCommand(args...); err != nil {
```
cfergeau, line 54: Could be `podman.runCommand("system", "connection", "remove", name)` , in which case it's paraphrasing the comment which can be dropped? I'm not even sure the intermediate helper is needed, to make it even more obvious what's going on.

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,82 @@
...
+	PodmanExecutablePath string
+}
+
+func NewConfig() Config {
+	return Config{
+		PodmanExecutablePath: filepath.Join(constants.CrcOcBinDir, constants.PodmanRemoteExecutableName),
+	}
+}
+
+func (podman Config) runCommand(args ...string) (string, string, error) {
```
cfergeau, line 26: or just `run(...)`?

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,82 @@
...
+	return podman.removeSystemConnection(rootless)
+}
+
+func (podman Config) RemoveRootfulSystemConnection() error {
+	return podman.removeSystemConnection(rootful)
+}
+
+func (podman Config) makeSystemConnectionDefault(name string) error {
+	args := []string{"system", "connection", "default", name}
+	if _, stderr, err := podman.runCommand(args...); err != nil {
```
cfergeau, line 70: same comment, the intermediate `args` is not really needed.

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,82 @@
...
+		return fmt.Errorf("failed to make root podman system connection %v: %s", err, stderr)
```
cfergeau, line 71: `root` is inaccurate here.

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,82 @@
...
+		args = append(args, "-d")
+	}
+	if _, stderr, err := podman.runCommand(args...); err != nil {
+		return fmt.Errorf("failed to add podman system connection %v: %s", err, stderr)
+	}
+	return nil
+}
+
+// AddRootlessSystemConnection and makes it default
+func (podman Config) AddRootlessSystemConnection(identity, uri string) error {
```
cfergeau, line 43: maybe rename `identity` to `sshIdentity` over this file to make it more obvious what `identity` is?

cfergeau (commented):
some more comments about the overall API, but looks good.

praveenkumar:
@cfergeau updated PR with all the suggestions

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,72 @@
...
+	if defaultCon {
+		args = append(args, "-d")
+	}
+	if _, stderr, err := run(args...); err != nil {
+		return fmt.Errorf("failed to add podman system connection %v: %s", err, stderr)
+	}
+	return nil
+}
+
+// AddRootlessSystemConnection and makes it default
```
cfergeau, line 34: A part of the  sentence seems to be missing. Not sure `AddRootlessSystemConnection` and `AddRootfulSystemConnection` should behave differently wrt to default. There are `MakexxxSystemConnectionDefault`  methods already, `addPodmanSystemConnections` could just call these, this way it is very clear what the default is, the API is symmetrical, ...

pkg/crc/podman/podman.go
```diff
@@ -0,0 +1,72 @@
...
+func AddRootlessSystemConnection(sshIdentity, uri string) error {
+	return addSystemConnection(true, rootlessConn, sshIdentity, uri)
+}
+
+func AddRootfulSystemConnection(sshIdentity, uri string) error {
+	return addSystemConnection(false, rootfulConn, sshIdentity, uri)
+}
+
+func removeSystemConnection(name string) error {
+	// podman system connection remove crc
```
cfergeau, line 44: This is paraphrasing the line just below.

cfergeau (commented):
Looks good, some minor discussion points.

pkg/crc/podman/podman.go
praveenkumar: @cfergeau I was doing 2 separate thing with single function :(

Issue #2920, opened 2022-01-04
CI: Add podman bundle job on openshift CI


praveenkumar:
As of now the CI only test the openshift bundle not podman bundle, we should have the CI job in place to test the podman bundles also which are created and going to part of dev releases.

cfergeau:
On a related note, it would be nice to add more unit tests to the dev branch as new features are added/bugs fixed/...

PR #3081, opened 2022-03-22
[Integration] Podman preset tests

jsliacan:
**Relates to:** Issue #2920 

This PR sets up integration tests to work with podman preset on all 3 platforms. The test itself is barely populated, but this will be remedied when we pull in some tests from the podman e2e suite.

### DIFF
--- a/test/integration/resize_test.go
+++ b/test/integration/resize_test.go
@@ -7,7 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("vary VM parameters: memory cpus, disk", func() {
+var _ = Describe("vary VM parameters: memory cpus, disk", Label("openshift-preset", "vm-resize"), func() {
 
 	Describe("use default values", func() {
 
--- a/test/integration/testsuite_test.go
+++ b/test/integration/testsuite_test.go
@@ -23,6 +23,7 @@ var userHome string
 var versionInfo VersionAnswer
 
 var bundlePath string
+var ginkgoOpts string
 var pullSecretPath string
 
 func TestTest(t *testing.T) {
@@ -30,14 +31,15 @@ func TestTest(t *testing.T) {
 	RegisterFailHandler(Fail)
 
 	// fetch the current (reporter) config
-	_, reporterConfig := GinkgoConfiguration()
+	suiteConfig, reporterConfig := GinkgoConfiguration()
+
 	err := os.MkdirAll("out", 0775)
 	if err != nil {
 		logrus.Infof("failed to create directory: %v", err)
 	}
 	reporterConfig.JUnitReport = filepath.Join("out", "integration.xml")
 
-	RunSpecs(t, "Test Suite", reporterConfig)
+	RunSpecs(t, "Test Suite", suiteConfig, reporterConfig)
 
 }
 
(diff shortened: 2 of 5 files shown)


anjannath:
@jsliacan Looking at the job logs, the CI failures seems to be because it can't find the correct `podman` bundle and trying to run setup with the `libvirt` bundle.

```
level=info msg="Getting bundle for the CRC executable"
[387](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/code-ready_crc/3081/pull-ci-code-ready-crc-main-integration-crc/1514175911950290944#1:build-log.txt%3A387)
        level=debug msg="Ensuring directory /home/packer/.crc/cache exists"
[388](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/code-ready_crc/3081/pull-ci-code-ready-crc-main-integration-crc/1514175911950290944#1:build-log.txt%3A388)
        crc_libvirt_4.10.6_amd64.crcbundle is not supported by this crc executable, please use crc_podman_libvirt_3.4.4_amd64.crcbundle
[389](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/code-ready_crc/3081/pull-ci-code-ready-crc-main-integration-crc/1514175911950290944#1:build-log.txt%3A389)
```

Currently we can't set the "bundle" in the testing code right? its configurable using an `env` variable, which is fixed for the whole test run, but now since we have tests that expects two (and more..) kinds of bundles, maybe we should make this configurable from the testing code?

jsliacan:
@anjannath yes, you're right for one PR check, both bundles will be needed. I'll have a look and suggest some changes. I'll make it a separate commit. Thanks!

anjannath:
@jsliacan are these tests supposed to exists along and separate from the upstream tests from `podman` as i think the upstream tests will be stand alone and we won't pull in actual test code to our repo?

test/integration/resize_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("vary VM parameters: memory cpus, disk", func() {
+var _ = Describe("vary VM parameters: memory cpus, disk", Label("openshift-preset"), func() {
```
anjannath, line 10: I'd use `disk-resize` as the label here, even though this performs the test with `openshift` preset

test/integration/testsuite_test.go
```diff
@@ -23,21 +23,23 @@ var userHome string
...
+	suiteConfig, reporterConfig := GinkgoConfiguration()
+
 	err := os.MkdirAll("out", 0775)
 	if err != nil {
 		logrus.Infof("failed to create directory: %v", err)
 	}
 	reporterConfig.JUnitReport = filepath.Join("out", "integration.xml")
 
-	RunSpecs(t, "Test Suite", reporterConfig)
+	RunSpecs(t, "Test Suite", suiteConfig, reporterConfig)
```
anjannath, line 42: Is this change needed to use the labels? maybe add a comment or explain in the commit log why this is needed

jsliacan: `suiteConfig.LabelFilter` is a suite-wide variable and it might be useful (e.g. when I thought I'd use it to enforce the use of PULL_SECRET_PATH in case filter is either empty or contains `openshift-preset`. I did not end up doing it, but I think having `suiteConfig` out in the open is better than hiding it behind `_` as I had done before. So I kept it there. Can it stay or should it be removed? It's not needed per se.

test/integration/resize_test.go
jsliacan: I added `disk-resize` label (unused as of now). You seem to be assuming we can only have 1 label. Since we can have many, do you still want to get rid of the `openshift-preset` label? I wouldn't be in favor of that.

anjannath: yes right, assumed we can only add one label.. having both is better, thanks

test/integration/testsuite_test.go
anjannath: thanks for explaining,  its just that its not clear how this change relates to the overall "adding of podman preset test" change. so I was trying to guess! :)

anjannath (approved):
LGTM

PR #3088, opened 2022-03-28
Windows: Run 'crc daemon' as task for 'crc setup'

praveenkumar:
This PR is uses windows powershell to create `crc daemon` as task
and start it on demand for the current user. This doesn't need any
elevated privillages to run `crc daemon` for user context. It try to
perform following tasks
- Check if there is already crcDaemon task
    + If present does it the latest one using version info
- Install the task using predefined xml definition
- Start the task which run the `crc daemon` in background.

There is a known issue with windows powershell when you start a task
in background using `powershell.exe` then for a second the windows pops
up and disappear, https://github.com/PowerShell/PowerShell/issues/3028
have more details around it but as of now this is only the way for
interactive user context otherwise the task will need the admin
privillages which we want to avoid.


cfergeau (commented):
> There is a known issue with windows powershell when you start a task in background using `powershell.exe` then for a second the windows pops up and disappear, https://github.com/PowerShell/PowerShell/issues/3028

Looks like this can be worked around though?

pkg/crc/preflight/preflight_daemon_powershell_windows.go
```diff
@@ -0,0 +1,44 @@
...
+    <DisallowStartIfOnBatteries>false</DisallowStartIfOnBatteries>
+    <StopIfGoingOnBatteries>false</StopIfGoingOnBatteries>
+    <Hidden>true</Hidden>
+    <MultipleInstancesPolicy>IgnoreNew</MultipleInstancesPolicy>
+    <IdleSettings>
+      <Duration>PT10M</Duration>
+      <WaitTimeout>PT1H</WaitTimeout>
+      <StopOnIdleEnd>true</StopOnIdleEnd>
+      <RestartOnIdle>false</RestartOnIdle>
+    </IdleSettings>
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau no it is stopping it intantly. This template is exported from a task and didn't modified.

pkg/crc/preflight/preflight_daemon_service_check_windows.go
```diff
@@ -0,0 +1,93 @@
...
+	return nil
+}
+
+func fixDaemonTaskInstalled() error {
+	// prepare the task script
+	binPath, err := goos.Executable()
+	if err != nil {
+		return fmt.Errorf("unable to find the current executables location: %v", err)
+	}
+	binPathWithArgs := fmt.Sprintf("%s daemon", strings.TrimSpace(binPath))
```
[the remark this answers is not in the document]
praveenkumar: In case output have spaces in between.

pkg/crc/preflight/preflight_daemon_service_check_windows.go
```diff
@@ -0,0 +1,93 @@
...
+	return nil
+}
+
+func checkIfDaemonTaskRunning() error {
+	stdout, stderr, err := powershell.Execute(fmt.Sprintf(`(Get-ScheduledTask -TaskName "%s").State`, constants.DaemonTaskName))
+	if err != nil {
+		logging.Debugf("%s task is not running: %v : %s", constants.DaemonTaskName, err, stderr)
+		return err
+	}
+	if strings.TrimSpace(stdout) != "Running" {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau That I didn't check and love to know if that works with localized version.

pkg/crc/preflight/preflight_windows.go
```diff
@@ -112,6 +112,29 @@ var vsockChecks = []Check{
 	},
 }
 
+var daemonServiceChecks = []Check{
+	{
+		configKeySuffix:    "check-daemon-task-install",
+		checkDescription:   "Check if daemon task is installed",
```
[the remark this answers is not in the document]
praveenkumar: Ah yes, I will improve that.

praveenkumar:
> Looks like this can be worked around though?

The workaround use `vbs` script which is detected as virus ( @anjannath told me) I don't find any other workaround which would work without installing any different tool.

cfergeau:
> > Looks like this can be worked around though?
> 
> The workaround use `vbs` script which is detected as virus ( @anjannath told me) I don't find any other workaround which would work without installing any different tool.

Has it already been tried to have our own small helper non-console binary which would just run the appropriate powershell comnad? or would it also be detected as a virus?

pkg/crc/preflight/preflight_daemon_service_check_windows.go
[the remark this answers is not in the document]
cfergeau: But are you getting spaces when you test it? Or are there concrete cases when you know for sure there will be spaces that we need to remove? This is what is not clear to me :)

praveenkumar:
> Has it already been tried to have our own small helper non-console binary which would just run the appropriate powershell comnad? or would it also be detected as a virus?

@cfergeau own small helper you mean create a small go binary which run `crc daemon`  like we create for file embed?

cfergeau:
> > Has it already been tried to have our own small helper non-console binary which would just run the appropriate powershell comnad? or would it also be detected as a virus?
> 
> @cfergeau own small helper you mean create a small go binary which run `crc daemon` like we create for file embed?

I think it's `powershell.Execute("Register-ScheduledTask", ...)` which causes the popup? I'm thinking of a non-console binary which would run this command for us, in the hope that it would allow us to hide the window.

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -0,0 +1,133 @@
...
+      <StopOnIdleEnd>true</StopOnIdleEnd>
+      <RestartOnIdle>false</RestartOnIdle>
+    </IdleSettings>
+    <UseUnifiedSchedulingEngine>true</UseUnifiedSchedulingEngine>
+  </Settings>
+  <Triggers />
+  <Actions Context="Author">
+    <Exec>
+      <Command>powershell.exe</Command>
+      <Arguments>-WindowStyle Hidden -Command %s</Arguments>
```
cfergeau, line 39: If what %s is substituted for contains `<`, `>`, `&`, ... this will fail badly. If you want to be 100% safe, you probably need https://pkg.go.dev/encoding/xml#EscapeText

praveenkumar:
> I think it's powershell.Execute("Register-ScheduledTask", ...) which causes the popup? I'm thinking of a non-console binary which would run this command for us, in the hope that it would allow us to hide the window.

It is not the `Register-ScheduledTask` but the command which use to run the crc daemon using `powershell.exe -WindowStyle Hidden -Command crc.exe daemon` if you don't use the `powershell.exe -WindowStyle Hidden` and direct use `crc.exe daemon` then user will see a powershell window will pop up and stay there which execute `crc.exe daemon` using `-WindowStyle Hidden` is kind of a neat way to don't show that window but as discribed it in the issue it is because `powershell.exe` is console application and there is nothing exist as  a win32 host application.

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -183,3 +183,14 @@ func removeCrcVM() (err error) {
 	logging.Debug("'crc' VM is removed")
 	return nil
 }
+
+func checkIfAdminHelperService() error {
```
anjannath, line 187: nit: rename to `checkIfAdminHelperServiceRunning()`

pkg/crc/preflight/preflight_daemon_task_check_windows.go
praveenkumar: Done

pkg/crc/preflight/preflight_daemon_service_check_windows.go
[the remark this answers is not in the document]
praveenkumar: Removed, because `binPath` string is what we are creating so should be done it properly there.

gbraad (approved):
Approved

@jsliacan have you been able to test this in an update scenario?

praveenkumar:
> > `` but the command which use to run the crc daemon using `powershell.exe -WindowStyle Hidden -Command crc.exe daemon` if you don't use the `powershell.exe -WindowStyle Hidden` and direct use `crc.exe daemon` then user will see a powershell window will pop up and stay there which execute `crc.exe daemon` using `-WindowStyle Hidden` is kind of a neat way to don't show that window but as discribed it in the issue it is because `powershell.exe` is console application and there is nothing exist as a win32 host application.
> 
> Ah, so each time the daemon is started, a console window flashes on the screen. Have you tried what is suggested in https://stackoverflow.com/questions/23250505/how-do-i-create-an-executable-from-golang-that-doesnt-open-a-console-window-whe instead of using powershell.exe?

Yes, if I use this during build of `crc` binary then there is no terminal output for any of the commands not only for `crc side`. I created following test script to experiment with it but this is not what we can consumed directly. With other helper command which can execute `crc daemon` command . But this can be part of improvement, I can create a follow up issue for this and have this discussion moved there.

```golang
package main

import (
	"fmt"
	"time"
)

func main() {
	for i := 0; i < 120; i++ {
		fmt.Println("This is for testing purpose")
		time.Sleep(time.Second)
	}
}
```

```
> go build -ldflags -H=windowsgui -o wintest.exe main.go
PS C:\Users\prkumar\work\github\practise\test> .\wintest.exe 
PS C:\Users\prkumar\work\github\practise\test> 
PS C:\Users\prkumar\work\github\practise\test> Get-Process wintest

Handles  NPM(K)    PM(K)      WS(K)     CPU(s)     Id  SI ProcessName
-------  ------    -----      -----     ------     --  -- -----------
     82       8    12472       5232       0.03  54140   3 wintest
```

gbraad:
`.\crc.exe setup`
```
...
INFO Checking if daemon task is installed
INFO Installing the daemon task
INFO Checking if daemon task is running
INFO Running the daemon task
...
```

`.\crc.exe start`
```
INFO Checking Windows edition
INFO Checking if the Hyper-V virtual switch exists
INFO Checking if daemon task is installed
INFO Checking if daemon task is running
INFO Checking admin helper service is running
INFO Loading bundle: crc_podman_hyperv_3.4.4_amd64...
INFO Creating CodeReady Containers VM for Podman 3.4.4...
INFO Generating new SSH Key pair...
INFO CodeReady Containers instance is running with IP 127.0.0.1
INFO Adding new bearer token for cockpit webconsole
podman runtime is now running.

  PS> & crc podman-env | Invoke-Expression
  PS> podman.exe COMMAND
```

gbraad:
@praveenkumar @anjannath should we modify the tray to check for this scheduled task and task accordingly if not running?
At the moment we rely on the daemon to be started from the tray... this might lead to a conflict.

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -0,0 +1,138 @@
...
+	// prepare the task script
+	binPath, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("unable to find the current executable location: %v", err)
+	}
+	binPathWithArgs := fmt.Sprintf("%s daemon", binPath)
+	var buf bytes.Buffer
+	if err := xml.EscapeText(&buf, []byte(binPathWithArgs)); err != nil {
+		return err
+	}
```
cfergeau, line 77: This belongs in `genDaemonTaskInstallTemplate`, not here. If you do it here, this means all callers of `genDaemonTaskInstallTemplate` have to be aware of it. This also kind of hardcodes in `fixDaemonTaskInstalled` that `genDaemonTaskInstallTemplate` will be doing `fmt.Sprintf(xml, ...)`. `fixDaemonTaskInstalled` has no need to know that, this just muddles unrelated abstraction layers

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -0,0 +1,138 @@
...
+		logging.Debugf("unable to run the %s task: %v : %s", constants.DaemonTaskName, err, stderr)
+		return err
+	}
+	return nil
+}
+
+func checkIfOlderTask() error {
+	stdout, stderr, err := powershell.Execute(fmt.Sprintf(`(Get-ScheduledTask -TaskName "%s").Version`, constants.DaemonTaskName))
+	if err != nil {
+		return fmt.Errorf("%s task is not running: %v : %s", constants.DaemonTaskName, err, stderr)
```
anjannath, line 132: Looking at the other `Get-ScheduledTask` calls the error message should be modified, its actually checking the `Version` here, and if this returns error then actually task is not installed ?

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		logging.Debugf("%s task is not installed: %v : %s", constants.DaemonTaskName, err, stderr)
+		return err
+	}
+	if err := checkIfOlderTask(); err != nil {
+		return err
+	}
+	return nil
+}
+
```
anjannath, line 66: here `checkIfDaemonTaskInstalled` can be simplified to just `return checkIfOlderTask()`? because that already returns an error if the call to `Get-ScheduledTask` returns an error

pkg/crc/constants/constants_windows.go
```diff
@@ -5,4 +5,6 @@ const (
 	PodmanRemoteExecutableName = "podman.exe"
 	TapSocketPath              = ""
 	DaemonHTTPNamedPipe        = `\\.\pipe\crc-http`
+	DaemonTaskName             = "crcDaemon"
+	AdminHelperServiceName     = "CodeReadyContainersAdminHelper"
```
gbraad, line 9: can we rename this to just `crcAdminHelper` ?

gbraad (changes requested):
simplify naming ?

pkg/crc/constants/constants_windows.go
praveenkumar: It would be changed in the installer also, this service name is part of installer https://github.com/code-ready/crc/blob/main/packaging/windows/product.wxs.template#L43 , Also if we change the name we have to make sure older service shouldn't exist.

pkg/crc/preflight/preflight_daemon_task_check_windows.go
praveenkumar: @anjannath No, `Get-ScheduledTask` first check if the task is present and then this check if present task if running and if it is running then does the version is same.

pkg/crc/preflight/preflight_daemon_task_check_windows.go
praveenkumar: I think it is much clean this way 
- Check if task installed using `Get-ScheduledTask -TaskName crcDaemon` (first block of Get-ScheduledTask) 
- Check if task is running using `$(Get-ScheduledTask -TaskName crcDaemon).Version` because if it is not running the it fail to fetch the version info
- Finally check the version is same what it should be

pkg/crc/preflight/preflight_daemon_task_check_windows.go
cfergeau: Not very easy to read with the XML embedded in the function, but I had in mind something like this:
```
func genDaemonTaskInstallTemplate(crcVersion, daemonServiceName string) (string, error) {
       var escapedName bytes.Buffer
	if err := xml.EscapeText(&escapedName, []byte(daemonServiceName)); err != nil {
		return "", err
	}
	return fmt.Sprintf(daemonTaskTemplate,
		crcVersion,
		escapedName.String(),
	), nil
}
```

praveenkumar:
> Sure, but this begs a few questions:
> 
>     * is it ok to merge this PR now, and possibly have the flashing window in the next release? Or do we hold this PR for a bit?

@cfergeau I think it should be OK to take this PR as of now with flashing window.
 
>     * if this PR goes in, and we eventually realize there is no alternative to the flashing window, do we revert this? or will we live with the flashing window?

@cfergeau No, because this flashing window is not something user even notice and if even they do we have it part of the known issue. we only improve it, revert only take place if there is an alternative to task manager which can run `daemon` in user context without Admin priviliages.

cfergeau:
> @cfergeau No, because this flashing window is not something user even notice and if even they do we have it part of the known issue.

Err, sorry, but I'd definitely notice, and find this ugly and unpolished from whatever application did this. Let's not make blanket statements regarding what users would or would not notice.

pkg/crc/constants/constants_windows.go
praveenkumar: @gbraad Can we have an follow up issue for name change and not block this PR for this one?

praveenkumar: @gbraad #3092

praveenkumar:
> > @cfergeau No, because this flashing window is not something user even notice and if even they do we have it part of the known issue.
> 
> Err, sorry, but I'd definitely notice, and find this ugly and unpolished from whatever application did this. Let's not make blanket statements regarding what users would or would not notice.

@cfergeau Wrong wording in the sentense. `No` was for the revert query. I am all up for making it hidden from the start but as of now we have to use this workaround.

gbraad (approved):

anjannath (approved):

anjannath:
Tested the end to end flow (both from cli and tray) and its working as expected

PR #3090, opened 2022-03-29
docs: Add OpenShift preset as a prerequisite to applicable procedures

robin-owen:
This is a simple PR to add use of the `openshift` preset as a prerequisite to certain procedures which depend on the use of an OpenShift cluster.


gbraad (approved):

gbraad:
This makes things clearer and a good start for documenting things related to the podman preset

PR #3095, opened 2022-03-31
Mac: Enable daemon service to run as soon as user login

praveenkumar:
Without this PR following scenario happen
- User installed the crc first time and do the setup
- As part of the setup crc daemon is started by launchd
- User reboot
- Tray autostart with reboot but `crc setup --check-only` fails
because `launchd agent 'com.redhat.crc.daemon' is not running`.

With this PR, check for daemon task running should first validate
if an existing process of `crc daemon` is running and the version
it get from the api should be same as the binary then task running
check succeed. So this PR will makes `crc setup --check-only` work
again with tray autostart.

### DIFF
--- a/pkg/os/launchd/launchd_darwin.go
+++ b/pkg/os/launchd/launchd_darwin.go
@@ -41,6 +41,8 @@ const (
 		<string>{{ $value }}</string>
 		{{ end }}
 	</dict>
+	<key>RunAtLoad</key>
+	<true/>
 </dict>
 </plist>
 `
@@ -112,12 +114,12 @@ func CheckPlist(config AgentConfig) error {
 
 // LoadPlist loads a launchd agents' plist file
 func LoadPlist(label string) error {
-	return runLaunchCtl("load", getPlistPath(label))
+	return runLaunchCtl("load", "-w", getPlistPath(label))
 }
 
 // UnloadPlist Unloads a launchd agent's service
 func UnloadPlist(label string) error {
-	return runLaunchCtl("unload", getPlistPath(label))
+	return runLaunchCtl("unload", "-w", getPlistPath(label))
 }
 
 // RemovePlist removes a launchd agent plist config file


cfergeau:
I thought launchd would ensure the daemon is started when the user opens a session on their mac? The launchd changes don't guarantee the daemon will be running during startup?

praveenkumar:
> I thought launchd would ensure the daemon is started when the user opens a session on their mac? The launchd changes don't guarantee the daemon will be running during startup?

@cfergeau As of now launchd is working same way as we are doing in linux except the socket activation part (in linux our socket activation service is enabled but that doesn't start daemon until something query to that socket) here also I think we don't want to start the daemon by default by launchd, right now it is required because of tray but in future tray is not part of crc.

cfergeau:
> @cfergeau As of now launchd is working same way as we are doing in linux except the socket activation part (in linux our socket activation service is enabled but that doesn't start daemon until something query to that socket) here also I think we don't want to start the daemon by default by launchd, right now it is required because of tray but in future tray is not part of crc.

On linux, the daemon is not automatically started on login because we can avoid it thanks to the socket activation files. If socket activation did not exist, we would have to start the daemon automatically on login.
If the daemon is not autostarted, I don't understand what launchd is used for? What does this bring over just running `crc daemon` manually?

praveenkumar:
> If the daemon is not autostarted, I don't understand what launchd is used for? What does this bring over just running `crc daemon` manually.

Running daemon manually means you have to keep that terminal always open or manually run it in background, as user you have to make sure stop it when not required `crc cleanup` not remove that process also, so even with current state it is better than running manually but our plan is to have it auto start every restart that we can improve over time.

cfergeau:
>  even with current state it is better than running manually but our plan is to have it auto start every restart that we can improve over time.

I am really unclear on what "the current state" is. I'm totally lost as to what happens if I run `crc setup`, and then reboot the machine. How will the daemon be started?

praveenkumar:
> > even with current state it is better than running manually but our plan is to have it auto start every restart that we can improve over time.
> 
> I am really unclear on what "the current state" is. I'm totally lost as to what happens if I run `crc setup`, and then reboot the machine. How will the daemon be started? Isn't autostarting the daemon on session logging one or 2 lines to change in the launchd plist file?

@cfergeau I thought we don't want that but yes it is easy and I updated the PR accordingly.

pkg/os/launchd/launchd_darwin.go
```diff
@@ -112,12 +114,12 @@ func CheckPlist(config AgentConfig) error {
 
 // LoadPlist loads a launchd agents' plist file
 func LoadPlist(label string) error {
-	return runLaunchCtl("load", getPlistPath(label))
+	return runLaunchCtl("load", "-w", getPlistPath(label))
```
anjannath, line 117: aha, does the `-w` mean its going to wait till the operation returns?

praveenkumar: No as per launchctl man page this suppose to enable the service 
```
              -w       Overrides the Disabled key and sets it to false or true for the load and unload subcommands respectively. In previous versions, this option would modify the
                       configuration file. Now the state of the Disabled key is stored elsewhere on- disk in a location that may not be directly manipulated by any process other than
                       launchd.

```

cfergeau:
One limitation to the current approach is if the daemon was started by the tray in a previous version. Starting the daemon with launchd will silently fail, and nothing detects this old daemon version.
One way to solve this is to check the version of the running daemon as was done in an earlier iteration of this code, and kill any old version before restarting it in the proper version.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -8,13 +8,19 @@ import (
...
 	"github.com/code-ready/crc/pkg/crc/machine/bundle"
 	"github.com/code-ready/crc/pkg/crc/preset"
 	"github.com/code-ready/crc/pkg/crc/validation"
+	"github.com/code-ready/crc/pkg/crc/version"
 	"github.com/pkg/errors"
 )
 
+// This make sure golint doesn't error out because not
+// using it on all the platforms
+var _ = isDaemonRunningWithReleasedVersion()
```
cfergeau, line 25: `isDaemonRunningWithReleasedVersion` is almost the same code as `checkDaemonStarted` which is used on all platforms, so this could be factored.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -158,7 +161,57 @@ func isDaemonRunningWithReleasedVersion() bool {
 	}
 	if ver.CrcVersion != version.GetCRCVersion() {
 		logging.Debugf("Daemon is running with %s version but binary version is %s", ver.CrcVersion, version.GetCRCVersion())
+		if err := killExistingDaemonProcess(); err != nil {
+			logging.Debugf("Failed to kill the existing crc daemon process: %v", err)
+		}
```
cfergeau, line 166: This function is called `isDaemonRunningWithReleasedVersion`, it's not supposed to stop/kill the daemon as a side effect.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -158,7 +161,57 @@ func isDaemonRunningWithReleasedVersion() bool {
...
+				if err := process.Kill(); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// isDaemonProcess return true if following conditions met
```
cfergeau, line 202: if any of the following conditions are met? if all of the following conditions are met? if none of the following conditions are met?

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -268,8 +268,10 @@ func checkIfDaemonPlistFileExists() error {
 	if err := launchd.CheckPlist(*daemonConfig); err != nil {
 		return err
 	}
-	if !launchd.AgentRunning(daemonConfig.Label) {
-		return fmt.Errorf("launchd agent '%s' is not running", daemonConfig.Label)
+	if !isDaemonRunningWithReleasedVersion() {
+		if !launchd.AgentRunning(daemonConfig.Label) {
+			return fmt.Errorf("launchd agent '%s' is not running", daemonConfig.Label)
+		}
```
cfergeau, line 274: What about writing it like this? This seems clearer regarding what the code wants to do.

```
if olderDaemonVersionRunning() {
   if err := killDaemonProcess(); err != nil {
       return err
   }
}
if !launchd.AgentRunning(daemonConfig.Label) {
}
```

pkg/crc/preflight/preflight_checks_common.go
anjannath: The killing of the existing daemon process can be part of the fix function for this check?

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau Handled in #3099 , will hold this until that merge and then rebase it

pkg/crc/preflight/preflight_checks_darwin.go
praveenkumar: This will be also easy as once #3099 in.

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau updated to `any of the following`.

pkg/crc/preflight/preflight_checks_darwin.go
praveenkumar: @cfergeau It is now suggested way.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -350,7 +350,16 @@ func checkDaemonSystemdSockets() error {
...
+	if err := checkSystemdUnit(vsockUnitName, vsockUnit, true); err != nil {
+		return err
+	}
+
+	if olderDaemonVersionRunning() {
+		if err := killDaemonProcess(); err != nil {
+			return err
+		}
+	}
+	return nil
```
praveenkumar, line 362: In linux I have to add it in this socket file check otherwise `cross-lint` error out not using `killDaemonProcess` function to linux. If we don't want to do that then might be we need to create a different file `preflight_check_nonlinux` and move this there.

gbraad:
> [praveenkumar](https://github.com/praveenkumar) added the [has-to-be-in-release](https://github.com/code-ready/crc/labels/has-to-be-in-release) label [4 hours ago](https://github.com/code-ready/crc/pull/3095#event-6393606320)

Why was this tag added? Is everything lined up correctly to allow this?

praveenkumar:
> > [praveenkumar](https://github.com/praveenkumar) added the [has-to-be-in-release](https://github.com/code-ready/crc/labels/has-to-be-in-release) label [4 hours ago](https://github.com/code-ready/crc/pull/3095#event-6393606320)
> 
> Why was this tag added? 

@gbraad because I want this to be part of 2.1.0 release (upcoming one).

> Is everything lined up correctly to allow this? 

Yes, tray side changes were already there, when I filed this PR initially I thought it would be a trival but @cfergeau suggestions around having launchd service as auto reload and also detecting the old version of daemon and kill it, took
some time to implement and getting it reviewed. 

> Which issue does this solve or related to?

This will solve the issue for a user who just want to do `crc setup` and `crc start` on mac with or without tray.

gbraad:
I am just concerned about how this was planned. Feels a little last minute.

We have to better describe dependency to merge, or accept a known issue?

anjannath:
> 

The review comments are mostly addressed, i'll test this now e2e with the tray and upgrade scenario

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: Anjan had a good point though that the killing fits better in a `fix` method rather than in `check`.

cfergeau (approved):
Can you add a description of the bug this fixes, or a link to the relevant issue? Apart  from this, looks good to me.

PR #3097, opened 2022-04-05
Update openshift bundle to 4.10.6 (stable channel)

praveenkumar:


praveenkumar:
> Not sure what happens when crc setup tries to download the bundles?

@cfergeau In the tests `crc setup` is not trying to download the bundle. Bundle download happen before and then `crc setup -b <bundle_path>` is used so that we can test the bundles without uploading it to mirror.openshift. This also allows us to run e2e on installer side where bundle is created for the PR and then used it with `crc setup -b` option.

cfergeau:
I was not specifically asking about the tests, but something more general. Once this is pushed to the main branch, crc will be in a weird state wrt bundle downloading?

praveenkumar:
> I was not specifically asking about the tests, but something more general. Once this is pushed to the main branch, crc will be in a weird state wrt bundle downloading?

@cfergeau from master branch prospective right it will be in weird state because we specify in Makefile 4.10.6 and bundle which part of constant have 4.10.3 

We can hold this until 4.10.6 bundles are available on mirror and then update on both places.

cfergeau:
I did not mean we must wait, but I wanted at least to check how this is going to be handled

PR #3099, opened 2022-04-06
Daemon: Move daemon running check to daemon client package

praveenkumar:
This move will help us to consume, daemon running check in the
preflight to make decision around if we need to start the daemon
as part of preflight or use the existing one.

### DIFF
--- a/pkg/crc/daemonclient/client.go
+++ b/pkg/crc/daemonclient/client.go
@@ -1,11 +1,16 @@
 package daemonclient
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/code-ready/crc/pkg/crc/api/client"
+	crcversion "github.com/code-ready/crc/pkg/crc/version"
 	networkclient "github.com/containers/gvisor-tap-vsock/pkg/client"
+	pkgerrors "github.com/pkg/errors"
 )
+
+const genericDaemonNotRunningMessage = "Is 'crc daemon' running? Cannot reach daemon API"
 
 type Client struct {
 	NetworkClient *networkclient.Client
@@ -21,4 +26,20 @@ func New() *Client {
 			Transport: transport(),
 		}, "http://unix/api"),
 	}
+}
+
+func GetVersionFromDaemonAPI() (*client.VersionResult, error) {
+	apiClient := client.New(&http.Client{Transport: transport()}, "http://unix/api")
+	version, err := apiClient.Version()
+	if err != nil {
+		return nil, pkgerrors.Wrap(err, genericDaemonNotRunningMessage)
+	}
+	return &version, nil
+}
+
+func CheckIfOlderVersion(version *client.VersionResult) error {
+	if version.CrcVersion != crcversion.GetCRCVersion() {
+		return fmt.Errorf("The executable version (%s) doesn't match the daemon version (%s)", crcversion.GetCRCVersion(), version.CrcVersion)
+	}
+	return nil
 }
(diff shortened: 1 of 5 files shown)


pkg/crc/daemonclient/client.go
```diff
@@ -22,3 +27,19 @@ func New() *Client {
 		}, "http://unix/api"),
 	}
 }
+
+func GetVersionFromDaemonAPI() (*client.VersionResult, error) {
+	daemonClient := New()
+	version, err := daemonClient.APIClient.Version()
+	if err != nil {
+		return nil, pkgerrors.Wrap(err, genericDaemonNotRunningMessage)
```
anjannath, line 35: `fmt.Errorf("%w")` can be used to wrap the error instead of the pkg/errors package

pkg/crc/daemonclient/client.go
```diff
@@ -22,3 +27,19 @@ func New() *Client {
 		}, "http://unix/api"),
 	}
 }
+
+func GetVersionFromDaemonAPI() (*client.VersionResult, error) {
+	daemonClient := New()
```
anjannath, line 32: We only want the `APIClient` for getting the version, we can avoid initializing the NetworkClient and just use:
```go
apiClient := client.New(&http.Client{Transport: transport()}, "http://unix/api")
version, err := apiClient.Version()
```

pkg/crc/daemonclient/client.go
praveenkumar: @anjannath I just moved this piece without change.

pkg/crc/daemonclient/client.go
praveenkumar: @anjannath  Sure, but with `New()` also in the background it is doing the same (network and api). Is it better someway doing this?

cfergeau (commented):
I would squash both commits, it's a bit odd to first move code to a single helper function, and then split again this helper and readd some of the code which was moved to the single helper. 
The removal of 
`return "Is CodeReady Containers tray application running? Cannot reach daemon API"` could go in its own commit though to explain why we no longer need this.
Dunno what to think about the end result, the 2 helpers are not adding a lot over just duplicating the code :( They may make more sense in the context of the other PR, not sure why this was split in its own PR?

praveenkumar:
> I would squash both commits, it's a bit odd to first move code to a single helper function, and then split again this helper and readd some of the code which was moved to the single helper. 

@cfergeau This is what confuse me, sometime we want to record every small code change is separate commit and something we think it is OK to have this in a single commit :( , First commit is just the move the bits to daemon client and second commit is to further break it the functionality. 

> The removal of `return "Is CodeReady Containers tray application running? Cannot reach daemon API"` could go in its own commit though to explain why we no longer need this. 

Yes, I think this should be first commit and then code move and then further break functionality.

>Dunno what to think about the end result, the 2 helpers are not adding a lot over just duplicating the code :( They may make more sense in the context of the other PR, not sure why this was split in its own PR?

I put it in separate PR so that I get it reviewed separately, At the end my plan to just cherry pick those commit to other PR.

pkg/crc/daemonclient/client.go
anjannath: the `New` function from `pkg/api/client` doesn't create the network client, and this package is already imported
https://github.com/code-ready/crc/blob/88a0aeeb5d89af1f37b4f8000a67b5c517a1a404/pkg/crc/api/client/client.go#L19-L25

anjannath: Thanks

praveenkumar:
@cfergeau @anjannath suggestions for this PR incorporated to #3095  . Closing this now.

PR #3102, opened 2022-04-07
extend the response from API to also send default value for config 

### LINKED ISSUE #2869: [Task] Add a way to distinguish if a value is default or what the default value is
### General information

  * OS: Linux

## CRC version
Current dev and master branch
  
## CRC config
```bash
<none>
```

### Steps to reproduce

  1. `GET /api/config`

### Issue
The current API will return all settings that possible, but does not indicate if the config is a default or not. 

` { autostart-tray: true, bundle: '/home/gbraad/.crc/cache/crc_libvirt_4.9.0.crcbundle', ... , nameserver: "", `

This means empty options are given or bundle, which is not explicitly set. It would be better to return settings that also includes what the default is or even if the default is used.

```json
   { nameserver: "",
       default: "" }, ...
```

```json
    { bundle: '/home/gbraad/.crc/cache/crc_libvirt_4.9.0.crcbundle', isDefault: true }, ...
```

This way we can also print this as a description or information in the React components for Electron and Cockpit module.

anjannath:
Fixes #2869 
e.g response for a single config `cpus`:
```json
{
  "Configs": {
    "cpus": {
      "value": 2,
      "isDefault": true,
      "defaultValue": 2
    }
  }
}
```

**Note: This will also need changes on the tray**

### DIFF
--- a/pkg/crc/api/api_client_test.go
+++ b/pkg/crc/api/api_client_test.go
@@ -18,9 +18,14 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type testConfig interface {
+	crcConfig.Storage
+	AllSettings() []crcConfig.Setting
+}
+
 type testClient struct {
 	*apiClient.Client
-	config     crcConfig.Storage
+	config     testConfig
 	httpServer *httptest.Server
 }
 
@@ -125,8 +130,12 @@ func TestConfigGet(t *testing.T) {
 	assert.Equal(
 		t,
 		apiClient.GetConfigResult{
-			Configs: map[string]interface{}{
-				"cpus": float64(4),
+			Configs: map[string]apiClient.SettingValue{
+				"cpus": {
+					Value:        float64(4),
+					IsDefault:    true,
+					DefaultValue: float64(4),
+				},
 			},
 		},
 		configGetResult,
@@ -156,8 +165,12 @@ func TestConfigSet(t *testing.T) {
 	assert.Equal(
 		t,
 		apiClient.GetConfigResult{
-			Configs: map[string]interface{}{
-				"cpus": float64(5),
+			Configs: map[string]apiClient.SettingValue{
+				"cpus": {
+					IsDefault:    false,
+					DefaultValue: float64(4),
+					Value:        float64(5),
+				},
 			},
 		},
 		configGetAfterSetResult,
@@ -182,16 +195,25 @@ func TestConfigGetAll(t *testing.T) {
 	client := newTestClient()
 	defer client.Close()
 	allConfigGetResult, err := client.GetConfig(nil)
+	allConfigs := client.config.AllConfigs()
 	assert.NoError(t, err)
-	configs := make(map[string]interface{})
-	for k, v := range client.config.AllConfigs() {
+	configs := make(map[string]apiClient.SettingValue)
+	for _, setting := range client.config.AllSettings() {
 		// This is required because of https://pkg.go.dev/encoding/json#Unmarshal
 		// Unmarshal stores float64 for JSON numbers in case of interface.
-		switch v := v.Value.(type) {
+		switch v := setting.DefaultValue.(type) {
 		case int:
-			configs[k] = float64(v)
+			configs[setting.Name] = apiClient.SettingValue{
+				Value:        float64(allConfigs[setting.Name].Value.(int)),
+				DefaultValue: float64(v),
+				IsDefault:    allConfigs[setting.Name].IsDefault,
+			}
 		default:
-			configs[k] = v
+			configs[setting.Name] = apiClient.SettingValue{
+				Value:        allConfigs[setting.Name].Value,
+				DefaultValue: v,
+				IsDefault:    allConfigs[setting.Name].IsDefault,
+			}
 		}
 	}
 	assert.Equal(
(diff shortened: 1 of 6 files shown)


pkg/crc/api/api_client_test.go
```diff
@@ -20,7 +20,7 @@ import (
 
 type testClient struct {
 	*apiClient.Client
-	config     crcConfig.Storage
+	config     *crcConfig.Config
```
anjannath, line 23: since this is just for the unit test, i change the type, but might be better to compose a new interface that embeds the `crcConfig.Storage` and adds `AllSettings()` method

pkg/crc/config/config.go
```diff
@@ -59,6 +59,15 @@ func (c *Config) AddSetting(name string, defValue interface{}, validationFn Vali
 	}
 }
 
+func (c *Config) GetSetting(name string) (Setting, error) {
+	s, ok := c.settingsByName[name]
+	if ok {
+		return s, nil
+	}
```
praveenkumar, line 66:
```
if s, ok := c.settingsByName[name]; ok {
      return s, nil
}
```

anjannath: done

pkg/crc/api/api_client_test.go
cfergeau: I would add it to `crcConfig.Schema` rather than `crcConfig.Storage`, which probably means adding a `schema crcConfig.Schema` field, which would be the same value as `config` (hopefully that's not too odd? :-/ )

pkg/crc/api/handlers.go
```diff
@@ -213,22 +214,37 @@ func (h *Handler) GetConfig(c *context) error {
...
+		if len(allConfigs) != len(allSettings) {
+			return fmt.Errorf("settings and configs have different length")
+		}
+
+		configs := make(map[string]client.SettingValue)
+		for _, setting := range allSettings {
+			configs[setting.Name] = client.SettingValue{
+				Value:        allConfigs[setting.Name].Value,
+				IsDefault:    allConfigs[setting.Name].IsDefault,
+				DefaultValue: setting.GetDefaultValue(),
```
cfergeau, line 228: instead of the `if len(allConfigs) != len(allSettings) {` check, it's probably more robust to do:
```
for _, setting := range allSettings {
	settingValue, settingValueFound := allConfigs[setting.Name]
	if !settingValueFound {
		return fmt.Errorf(...)
	}                       
	configs[setting.Name] = client.SettingValue{
				Value:        settingValue.Value,
				IsDefault:    settingValue.IsDefault,
	...
```

pkg/crc/api/handlers.go
```diff
@@ -213,22 +214,37 @@ func (h *Handler) GetConfig(c *context) error {
...
 			Configs: configs,
 		})
 	}
 
-	configs := make(map[string]interface{})
+	configs := make(map[string]client.SettingValue)
 	for _, key := range req.Properties {
 		v := h.Config.Get(key)
-		if v.Invalid {
+		s, err := h.Config.GetSetting(key)
```
cfergeau, line 239: `GetSetting` could be added to the `Schema` interface if needed.

pkg/crc/api/handlers.go
```diff
@@ -213,22 +214,37 @@ func (h *Handler) GetConfig(c *context) error {
...
+		if v.Invalid || err != nil {
 			continue
 		}
-		configs[key] = v.Value
+		configs[key] = client.SettingValue{
+			Value:        v.Value,
+			IsDefault:    v.IsDefault,
+			DefaultValue: s.GetDefaultValue(),
```
cfergeau, line 246: I was the one suggesting adding `GetDefaultValue()` rather than exporting the `DefaultValue` field, but now that I look again at the PR, this seems inconsistent with the rest of the code, maybe it's better to export the field instead.

cfergeau (commented):
Overall looks good, more eyes looking at the http API would be good.

appveyor.yml
```diff
@@ -7,8 +7,11 @@ before_test:
...
   - choco install wixtoolset
   - set PATH=C:\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin;%PATH%
-  - set PATH=C:\go116;%WIX%bin;%PATH%
+  - set PATH=C:\go116\bin;%WIX%bin;%PATH%
   - make out/windows-amd64/crc-windows-installer.zip
   - make cross
 test_script:
+  - echo %PATH%
+  - dir C:\go116
+  - go version
```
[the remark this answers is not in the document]
anjannath: this commit i'll drop from the PR, it was just to check the appveyor issue

pkg/crc/api/handlers.go
anjannath: Its not strictly needed, and adds another method that'd be needed to implement if we want to have another implementation of `Schema`

pkg/crc/api/handlers.go
anjannath: yeah on single getter method, i'll revert back to exporting the `DefaultValue`

pkg/crc/api/api_client_test.go
anjannath: @cfergeau updated as i mentioned in the first comment, we only need it for the test, so do not want to change the interface in the main code, PTALA

anjannath:
/hold
(needs changes on the tray side, before merging need to decide how we'll handle that)

PR #3109, opened 2022-04-11
Kill daemon process if running with a different version 

praveenkumar:
## Solution/Idea

This pr try to kill an existing daemon process if it is running with different version than expected. It will allow setup to run
required version of daemon.

## Testing

- Start `crc daemon` with older version
- Try `crc setup` with this PR (bump the version to something like 2.2.0)
- `crc setup` should kill the existing `crc daemon` process and start the daemon with correct version.

Also
- Start `crc daemon` with same version
- Try `crc setup` with this PR
- `crc setup --check-only` should pass with using existing daemon process.

This will be useful for upcoming releases where user try to upgrade crc but daemon might be running with older version of crc so that case `crc setup` should kill that existing process and start daemon process with launchd/task manager.


pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -261,6 +261,9 @@ func getDaemonConfig() (*launchd.AgentConfig, error) {
 }
 
 func checkIfDaemonPlistFileExists() error {
+	if err := olderDaemonVersionRunning(); err != nil {
```
anjannath, line 264: The last if clause will make this check return error if the daemon is not running via `launchd` and `crc setup --check-only` will also return error, it should be modified to consider the case when `daemon` started from the `tray`, similar to windows. something like:
```go
if !launchd.AgentRunning(daemonConfig.Label) || !daemonRunning() {
}
```

however, we have enabled the `daemon` to auto-start at boot on macOS also, and while testing `daemon` starts before the `tray`, but in the case where the tray starts first and `launchd` silently fails to start the daemon, we might see the above happening

cfergeau: I'm not in favour of doing something similar to what is done on Windows. Preflight checks usually check that the system is in the state we expect. On macOS we expect the daemon to be started with launchd, so it is an error if it's not launched this way, even if the daemon was started some other way. If one does not want to use launchd to start the daemon, then they can set `skip-check-launchd` in their config.

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -0,0 +1,73 @@
...
+func olderDaemonVersionRunning() error {
+	// Here daemonclient.GetVersionFromDaemonAPI() can return the error
+	// if the daemon is not running or daemon version API is not responding
+	// in both situation we can't check if daemon is running with an older
+	// version of crc or not, so we are just using the `err == nil` for condition.
+	if v, err := daemonclient.GetVersionFromDaemonAPI(); err == nil {
+		if err := daemonclient.CheckIfOlderVersion(v); err != nil {
+			return err
+		}
+	}
```
cfergeau, line 71:
```
v, err := daemonclient.GetVersionFromDaemonAPI()
if err != nil {
    return nil
}
if err := daemonclient.CheckIfOlderVersion(v); err != nil {
	return err
}
```
I'd still write it this way, more obvious we are ignoring errors.

pkg/crc/preflight/preflight_checks_darwin.go
praveenkumar: @anjannath So plan is to have something similar for windows task scheduler so that task should start as soon as user logs in. From tray side then we can completely remove the daemon starting part.

pkg/crc/preflight/preflight_checks_nonlinux.go
praveenkumar: Done.

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -0,0 +1,72 @@
...
+
+package preflight
+
+import (
+	"runtime"
+	"strings"
+
+	"github.com/code-ready/crc/pkg/crc/daemonclient"
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/shirou/gopsutil/process"
```
cfergeau, line 12: Would be better to use `github.com/shirou/gopsutil/v3/process`

pkg/crc/preflight/preflight_checks_darwin.go
anjannath: > @anjannath So plan is to have something similar for windows task scheduler so that task should start as soon as user logs in. From tray side then we can completely remove the daemon starting part.

yep, that'll solve the issue i mentioned

anjannath:
LGTM 👍🏽

pkg/crc/preflight/preflight_checks_nonlinux.go
praveenkumar: @cfergeau I tired following and looks like it have some issue.
```golang
package main

import (
	"fmt"
	"github.com/shirou/gopsutil/v3/process"
	"log"
	"strings"
)

func main() {
		processes, err := process.Processes()
		if err != nil {
			log.Fatal(err)
		}
		for _ , process := range processes {
			name, err := process.Name()
			if err != nil {
				fmt.Println("Not able to get process Name")
				log.Fatal(err)
			}
			if strings.TrimSpace(name) == "crc" {
				fmt.Println("Got the process: ", name)
				cmdLine, err := process.CmdlineSlice()
				if err != nil {
					log.Fatal(err)
				}
				if cmdLine[1] == "daemon" {
					fmt.Println("Got the pid: ", process.Pid)
				}
			}
		}
}
```

Run it
```
$ go run proc.go
Not able to get process Name
2022/04/19 13:04:04 invalid argument
exit status 1
```

cfergeau: > I tried the following and looks like it has some issue.

Your example works fine on my system, no errors when calling `process.Name()`, no idea what's going on with yours :-/

cfergeau: I've been able to reproduce on macOS, and this is indeed behaves differently between v3 and !v3 :-/

cfergeau: That's a bit of a mess:
https://github.com/shirou/gopsutil/issues/1288
It's even worse as Exe() does always for me, and is broken on !CGO_ENABLED builds, so we cannot fallback to this. Maybe the best for now is to ignore Name() errors and assume this was not `crc` when we have an error.

praveenkumar: > Maybe the best for now is to ignore Name() errors and assume this was not crc when we have an error.

@cfergeau as per https://github.com/shirou/gopsutil/issues/1288#issuecomment-1103214757 looks like https://github.com/shirou/gopsutil/pull/1229#issuecomment-1012408022 is what the issue. I am wondering why then `process.Processes()` have list of the process which is not owned by current user, should we first filter those process with current user and the check `process.Name()` ?

praveenkumar: Something like
```golang
package main

import (
	"fmt"
	"github.com/shirou/gopsutil/v3/process"
	"log"
	"os/user"
	"strings"
)

func main() {
		processes, err := process.Processes()
		if err != nil {
			log.Fatal(err)
		}
		for _ , process := range processes {
			usr, err := process.Username()
			if err != nil {
				log.Fatal(err)
			}
			currentUser, err := user.Current()
			if err != nil {
				log.Fatal(err)
			}
			if currentUser.Username != usr {
				continue
			}
			name, err := process.Name()
			if err != nil {
				log.Fatal(err)
			}
			if strings.TrimSpace(name) == "crc" {
				fmt.Println("Got the process: ", name)
				cmdLine, err := process.CmdlineSlice()
				if err != nil {
					log.Fatal(err)
				}
				if cmdLine[1] == "daemon" {
					fmt.Println("Got the pid: ", process.Pid)
				}
			}
		}
}
```

praveenkumar: @cfergeau ignored the error for `process.Name()`

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -112,8 +111,8 @@ func removeDaemonTask() error {
 }
 
 func checkIfDaemonTaskRunning() error {
-	if isDaemonRunningWithReleasedVersion() {
-		return nil
+	if err := olderDaemonVersionRunning(); err != nil {
+		return err
```
cfergeau, line 115: This feels redundant with https://github.com/code-ready/crc/pull/3133 though. Maybe https://github.com/code-ready/crc/pull/3133 should remove that daemon version check, and then this PR becomes simpler?

cfergeau (commented):
Apart from the redundancy comment, looks good to me.

pkg/crc/preflight/preflight_daemon_task_check_windows.go
praveenkumar: @cfergeau In windows we don't have autostart of task so that's why we have this redundant code, I can re-evaluate it after the release.

cfergeau: Ah right still no autostart :( Imo this has to be cleaned up as soon as possible (but I agree not for this release), as we are handling too many things (daemon started through a task, daemon started manually, daemon running with out of date task file, out of date daemon running, ...)

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	processes, err := process.Processes()
+	if err != nil {
+		return err
+	}
+	for _, process := range processes {
+		// Ignore the error from process.Name() because errors
+		// come from processes which are not owned by the current user
+		// on Mac - https://github.com/shirou/gopsutil/issues/1288
+		name, _ := process.Name()
```
cfergeau, line 28: Missed that, really should be 
```
name, err := process.Name()
if err != nil {
    // Ignore the error from process.Name() because errors
    // come from processes which are not owned by the current user
    // on Mac - https://github.com/shirou/gopsutil/issues/1288
    continue
}

PR #3111, opened 2022-04-12
Fix cluster connectivity check when a proxy CA cert is used

cfergeau:
The code in git has been broken since its introduction in v1.33.0~6 due to a
c&p typo. This PR improves on what was added in this commit so that this works
both with MITM proxies and https:// proxies


pkg/crc/services/dns/dns.go
```diff
@@ -128,7 +128,15 @@ func CheckCRCPublicDNSReachable(serviceConfig services.ServicePostStartConfig) (
...
-			curlArgs = append(curlArgs, "--proxy-cacert", proxyConfig.HTTPProxy)
+			// --proxy-cacert/--cacert replaces the system CAs with the specified one.
+			// If not using MITM proxy, --cacert must *not* be used, and if not using
+			// https:// proxy, --proxy-cacert must *not* be used
+			// ProxyCAFile is ambiguous, we cannot know if it's set because of MITM proxy,
+			// because of https:// proxy, or because of both
+			// We do not really care about transport security for this test, all that
+			// matters is whether or not we can resolve the hostname, so we can
+			// workaround this ambiguity by using an insecure connection
+			curlArgs = append(curlArgs, "--insecure", "--proxy-insecure")
```
praveenkumar, line 139: @cfergeau I was reading `man curl` and also searched bit on internet but didn't get this info which you put as comment. Can you provide a link to read around it and add it in the comment also?

pkg/crc/services/dns/dns.go
```diff
@@ -122,11 +122,21 @@ func CheckCRCPublicDNSReachable(serviceConfig services.ServicePostStartConfig) (
...
```
jsliacan, line 139: @cfergeau with these changes, I still get the same error:

```
WARN Failed public DNS query from the cluster: ssh command error:
command : curl --head quay.io --proxy http://localhost:8080 --proxy http://localhost:8080 --noproxy 127.0.0.1,localhost,.testing --insecure --proxy-insecure
err     : Process exited with status 7
 :  
```

1. There are still two `--proxy` options in the command. Not sure why, maybe one of them can be removed with as part of this PR? (this does not cause the error though)
2. When I run the command manually, I see it go through the proxy and the exit code is not 7, but 0. 
```
$ curl --head quay.io --proxy http://localhost:8080 --proxy http://localhost:8080 --noproxy 127.0.0.1,localhost,.testing --insecure --proxy-insecure
HTTP/1.1 301 Moved Permanently
Connection: keep-alive
Content-Length: 134
Content-Type: text/html
Date: Tue, 12 Apr 2022 20:18:45 GMT
Location: https://quay.io:443/
Server: awselb/2.0

$ echo $?
0
```
And on the proxy I see this:
```
2022/04/12 22:18:45 [026] INFO: Got request / quay.io HEAD http://quay.io/
2022/04/12 22:18:45 [026] INFO: Sending request HEAD http://quay.io/
2022/04/12 22:18:45 [026] INFO: Received response 301 Moved Permanently
2022/04/12 22:18:45 [026] INFO: Copying response to client 301 Moved Permanently [301]
2022/04/12 22:18:45 [026] INFO: Copied 0 bytes to client error=<nil>
```
So it seems that something else is going on here... When you run your PR, do you *not* get the warning?

pkg/crc/services/dns/dns.go
cfergeau: "this info", "around it"? I'm not sure which part of the comment is not clear/needs a link? I did not use more than the man page.
Short list of steps performed by `curl quay.io`:
1. if a proxy is in use, connect to it. If this is a https proxy, this involves creating a tls connection. cert checks/... can be bypassed with --proxy-insecure
2. DNS resolution of quay.io (will be performed either locally, or by the proxy)
3. connect to quay.io. If the connection is done to a https URL, this involves creating a tls connection. cert checks/... can be bypassed with --insecure

The step we want to test is 2., TLS failures at step 1. or 3. are not particularly worrying/useful for this test.

cfergeau:
> There are still two --proxy options in the command. Not sure why, maybe one of them can be removed with as part of this PR? (this does not cause the error though)

I added a comment for this in this PR. Code could probably be reworked to only have one occurrence
> ```
WARN Failed public DNS query from the cluster: ssh command error:
command : curl --head quay.io --proxy http://localhost:8080 --proxy http://localhost:8080 --noproxy 127.0.0.1,localhost,.testing --insecure --proxy-insecure
err     : Process exited with status 7
```

proxy URL must not use `localhost` as the curl command is going to be run in the VM. If you use 192.168.122.1 instead, and allow traffic over ports 3128/3129 between VM and host (not 100% sure this is needed, but see the comment in images/squid/Dockerfile), then the check should work.

File: pkg/crc/services/dns/dns.go [hunk 3]
```diff
@@ -118,10 +118,11 @@ func CheckCRCPublicDNSReachable(serviceConfig services.ServicePostStartConfig) (
...
 	if proxyConfig.IsEnabled() {
-		if proxyConfig.HTTPProxy != "" {
-			curlArgs = append(curlArgs, "--proxy", proxyConfig.HTTPProxy)
-		}
+		proxyHost := proxyConfig.HTTPProxy
 		if proxyConfig.HTTPSProxy != "" {
+			proxyHost = proxyConfig.HTTPSProxy
+		}
+		if proxyHost != "" {
 			curlArgs = append(curlArgs, "--proxy", proxyConfig.HTTPSProxy)
```
praveenkumar, line 126: @cfergeau shouldn't it would be `curlArgs = append(curlArgs, "--proxy", proxyHost)`

cfergeau: yes it should, thanks!

Issue #3116, opened 2022-04-19
[BUG] setup fails on mac


adrianriobo:
### General information

  Tested on macOS 

### CRC version
Pre released 2.1.0
  
### Steps to reproduce

  1. Install crc on a fresh environment
  2. crc setup

### Expected

Setup command set the machine to start the preset.

### Actual

Setup exit with error. When trying to create the symlink as ~/.crc/bin folder does not exist

### Logs

```bash
crc setup --log-level debug
DEBU CodeReady Containers version: 2.1.0+81a5f334 
DEBU OpenShift version: 4.10.6                    
DEBU Podman version: 3.4.4                        
DEBU Running 'crc setup'                          
INFO Using bundle path /Users/crcqe/.crc/cache/crc_hyperkit_4.10.6_amd64.crcbundle 
DEBU Error changing /Users/crcqe/.crc/cache permissions to 0775 
INFO Checking if running as non-root              
INFO Checking if crc-admin-helper executable is cached 
INFO Checking for obsolete admin-helper executable 
DEBU Checking if an older admin-helper executable is installed 
DEBU No older admin-helper executable found       
INFO Checking if running on a supported CPU architecture 
INFO Checking minimum RAM requirements            
DEBU Total memory of system is 34359738368 bytes  
INFO Checking if crc executable symlink exists    
DEBU lstat /Users/crcqe/.crc/bin: no such file or directory 
INFO Creating symlink for crc executable          
DEBU symlinking /Applications/CodeReady Containers.app/Contents/Resources/crc to /Users/crcqe/.crc/bin/crc 
symlink /Applications/CodeReady Containers.app/Contents/Resources/crc /Users/crcqe/.crc/bin/crc: no such file or directory
```

PR #3117, opened 2022-04-20
Constants: Creates cache and bin directories as part of EnsureBaseDirectoriesExist

praveenkumar:
Fixes: #3116 

As of now we only create `$HOME/.crc` directory as part of root init function and
other child directories are created on demand like cache is created when bundle is
download and extracted. Recently we observed that required child directory (bin) is not
exist when creating a symlink. It is easy to miss to check for well known crc clild
directory when writing code and assumed it should be always there to consume.

We have identified `$HOME/.crc`, `$HOME/.crc/bin` and `$HOME/.crc/cache` are needed
directories to `crc` to work so this commit create those as part of init function from
root.

## Testing
* rm -fr ~/.crc
* install crc using installer
* Try to run the setup (either from cli or from tray)
* setup should succeed

Note:
While this issue is macOS specific, this needs to be run on all platforms

### DIFF
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -119,9 +119,16 @@ func GetHomeDir() string {
 	return homeDir
 }
 
-// EnsureBaseDirectoryExists create the ~/.crc directory if it is not present
+// EnsureBaseDirectoriesExist creates ~/.crc, ~/.crc/bin and ~/.crc/cache directories if it is not present
 func EnsureBaseDirectoriesExist() error {
-	return os.MkdirAll(CrcBaseDir, 0750)
+	baseDirectories := []string{CrcBaseDir, MachineCacheDir, crcBinDir}
+	for _, baseDir := range baseDirectories {
+		err := os.MkdirAll(baseDir, 0750)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func IsRelease() bool {
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ release/
 *.crcbundle
 packaging/tmp
 packaging/*.pkg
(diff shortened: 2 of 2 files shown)


gbraad (approved):

anjannath (approved):

pkg/crc/constants/constants.go
```diff
@@ -119,9 +119,14 @@ func GetHomeDir() string {
...
-// EnsureBaseDirectoryExists create the ~/.crc directory if it is not present
+// EnsureBaseDirectoriesExist creates ~/.crc, ~/.crc/bin and ~/.crc/cache directories if it is not present
 func EnsureBaseDirectoriesExist() error {
-	return os.MkdirAll(CrcBaseDir, 0750)
+	var err error
+	baseDirectories := []string{CrcBaseDir, MachineCacheDir, crcBinDir}
+	for _, baseDir := range baseDirectories {
+		err = os.MkdirAll(baseDir, 0750)
+	}
+	return err
```
cfergeau, line 129: `err` does not seem very useful. This is ignoring all errors except the last one, and this returns nil if the last Mkdir is successful. Why is the last error special compared to the others?

cfergeau: When an installer is used, `crcBinDir` will be in a location where the user can't write. The directory already exists, so I don't think this will cause a problem, but did you test this scenario?

praveenkumar: @cfergeau I tested this Patch with installer and it works now. `crcBinDir` is used for identify other executable and here we are using `CrcSymlinkPath = filepath.Join(crcBinDir, "crc")`  which is not depend on `BinDir()` function.

cfergeau (approved):

PR #3118, opened 2022-04-20
[integration] fix default bundle location value on integration container

adrianriobo:
This PR is a following from #3047 

Now bundle location is not embedded anymore, the entrypoint for the integration container still pass that value as default parameter. 

This PR pass bundle location in case is specified through ENV on the container. Otherwise no value is passed.


cfergeau (approved):
Looks good to me!

PR #3120, opened 2022-04-20
Replace CodeReady Containers to CRC

praveenkumar:
This PR replace CodeReady Containers to CRC in cmd, pkg, images
directories and also update the README.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,4 +1,4 @@
-= Red Hat CodeReady Containers - OpenShift 4 on your Laptop
+= CRC - OpenShift 4 on your Laptop
 :icons:
 :toc: macro
 :toc-title:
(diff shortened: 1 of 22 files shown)


praveenkumar:
This is mostly find and replace, if some of the string doesn't make sense we can modify accordingly.

pkg/crc/constants/constants_darwin.go
```diff
@@ -9,7 +9,7 @@ import (
 const (
 	OcExecutableName           = "oc"
 	PodmanRemoteExecutableName = "podman"
-	TrayExecutableName         = "CodeReady Containers.app"
+	TrayExecutableName         = "CRC.app"
```
praveenkumar, line 12: @anjannath @gbraad I think we go with `crc.app` here?

pkg/crc/constants/constants_darwin.go
```diff
@@ -23,5 +23,5 @@ func TrayExecutablePath() string {
 		return filepath.Clean(filepath.Join(version.InstallPath(), "..", "MacOS", "crc-tray"))
 	}
 	// Should not be reached, tray is only supported on installer builds
-	return filepath.Clean(filepath.Join(BinDir(), "CodeReady Containers"))
+	return filepath.Clean(filepath.Join(BinDir(), "CRC"))
```
praveenkumar, line 26: @anjannath This should also change as per tray side changes..

gbraad (approved):

pkg/crc/constants/constants_darwin.go
gbraad: This should be `OpenShift Local.app`

pkg/crc/constants/constants_darwin.go
gbraad: The name tray remains `crc-tray.exe`, but internally named "Red Hat OpenShift Local"

gbraad (changes requested):
Application name should be `OpenShift Local`

pkg/crc/constants/constants_darwin.go
praveenkumar: @gbraad This is not the file name but folder name.

gbraad: yep... changed

gbraad (approved):
This is OK for the CLI

PR #3121, opened 2022-04-20
docs: Rename CodeReady Containers to "CRC"

robin-owen:
This is quite possibly the smallest pull request I have ever submitted. :wink: 

I double-checked the rendered content with the new name and discovered that no further disambiguation was necessary. Simply changing the related attribute provides us with an effective rename in the documentation without any rewriting.

### DIFF
--- a/docs/source/meta/product-attributes.adoc
+++ b/docs/source/meta/product-attributes.adoc
@@ -22,7 +22,7 @@
 :ocp: {openshift} Container Platform
 
 // Product naming
-:prod: CodeReady Containers
+:prod: CRC
 :rh-prod: {rh} {prod}
 :bin: crc
 


gbraad (approved):

gbraad:
for downstream this is "Red Hat OpenShift Local"

praveenkumar:
> for downstream this is "Red Hat OpenShift Local"

@robin-owen This you should able to do downstream right, changing `CRC` to `Red Hat OpenShift Local` ?

robin-owen:
@praveenkumar Correct. We have a separate `docs/source/meta/product-attributes.adoc` file for the downstream documentation. Other than that file, the documentation should be identical. This is how we can easily switch names between the two sources.

PR #3127, opened 2022-04-21
Windows: rename to Red Hat OpenShift Local

gbraad:


gbraad:
@anjannath please take a look

anjannath (approved):
LGTM

praveenkumar:
/hold

I am checking the adminhelper service part which can be broken with this change.

packaging/windows/product.wxs.template
```diff
@@ -13,35 +13,35 @@
...
+                <Directory Id="INSTALLDIR" Name="Red Hat OpenShift Local">
                     <Component Id="CrcExe" Guid="*">
                         <File Id="CrcExe" Source="SourceDir\crc.exe" KeyPath="yes" DiskId="1" />
                         <RemoveFile Id="RemoveInstallFiles" Name="*.*" On="uninstall" />
                     </Component>
                     <Component Id="AdminHelper" Guid="*">
                         <File Id="AdminHelper" Source="SourceDir\crc-admin-helper-windows.exe" KeyPath="yes" DiskId="1" />
-                        <ServiceInstall Name="CodeReadyContainersAdminHelper" Description="Perform administrative tasks for the user" Arguments="daemon" DisplayName="CodeReady Containers Admin Helper" ErrorControl="normal" Start="auto" Type="ownProcess" />
-                        <ServiceControl Id="StartAdminHelperService" Name="CodeReadyContainersAdminHelper" Start='install' Stop='both' Remove='uninstall' />
+                        <ServiceInstall Name="crcAdminHelper" Description="Perform administrative tasks for the user" Arguments="daemon" DisplayName="Red Hat OpenShift Local Admin Helper" ErrorControl="normal" Start="auto" Type="ownProcess" />
```
praveenkumar, line 43: https://github.com/code-ready/admin-helper/blob/5e6bd7a6fad7c738ec2d90a151b58d52c2c6e0c8/cmd/admin-helper/daemon.go#L59-L68 this is hard coded in the admin-helper repo but I think if we create the service ourself then we can name it as we want (do need testing on windows)

Also https://github.com/code-ready/crc/blob/2ca23cf0c1ad9be74a581b4184b443dd8a1f2b77/pkg/crc/constants/constants_windows.go#L9 needs to change it is used as part of preflight to make sure if service is running.

praveenkumar:
Once https://github.com/code-ready/crc/pull/3127#discussion_r855162936 is resolved we can unhold and merge

packaging/windows/product.wxs.template
anjannath: > https://github.com/code-ready/admin-helper/blob/5e6bd7a6fad7c738ec2d90a151b58d52c2c6e0c8/cmd/admin-helper/daemon.go#L59-L68 this is hard coded in the admin-helper repo but I think if we create the service ourself then we can name it as we want (do need testing on windows)

yeah, this is fine, we install the service in the msi so it doesn't depend on the one defined in admin-helper repo

gbraad: will make this change.

praveenkumar:
> finale msi is named `crc-windows-installer.msi` is that fine, or that should also be renamed?

@anjannath we shouldn't change it otherwise it will break the cloud.redhat download links

gbraad:
> .msi

if they need to be changed, this is something for later. I did raise it, but wasn't a concern.

gbraad:
@adrianriobo @jsliacan This needs some attention while testing

pkg/crc/constants/constants_windows.go
```diff
@@ -6,5 +6,5 @@ const (
 	TapSocketPath              = ""
 	DaemonHTTPNamedPipe        = `\\.\pipe\crc-http`
 	DaemonTaskName             = "crcDaemon"
-	AdminHelperServiceName     = "CodeReadyContainersAdminHelper"
+	AdminHelperServiceName     = "crcAdminHelper"
```
anjannath, line 9: 👍🏽

Issue #3123, opened 2022-04-21
[BUG] Error deleting CRC cluster on Windows


jsliacan:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V

## CRC version
```bash
PS C:\Users\crcqe> crc version
CodeReady Containers version: 2.1.0+81a5f334
OpenShift version: 4.10.6
Podman version: 3.4.4
```

## CRC config
```bash
PS C:\Users\crcqe> crc config view
- consent-telemetry                     : no
```

### Steps to reproduce

  1. Start CRC from powershell
  2. Issue `crc delete -f` or `crc delete` on a running cluster.

### Expected

Cluster deleted without any errors being reported.

### Actual
```
PS C:\Users\crcqe> crc delete
Do you want to delete the instance? [y/N]: y
failed to remove podman system connection exit status 1: time="2022-04-21T11:42:34+02:00" level=error msg="The handle is invalid."
```

gbraad:
can you put your `crc config view1 here too?
preset looks to be `podman` ?

jsliacan:
@gbraad preset was `openshift`. Will log into the machine & get the config now.

PR #3135, opened 2022-04-22
Delete: Send podman connection remove error to debug log

praveenkumar:
**Fixes:** Issue #3123 

In case of openshift preset there is no crc system connections are
created so during delete following error appear in windows.
```
DEBU Running 'Hyper-V\Get-VM crc | Select-Object -ExpandProperty State'
DEBU Running 'Hyper-V\Remove-VM crc -Force'
DEBU Running 'C:\Users\prkumar\.crc\bin\oc\podman.exe system connection
remove crc'
DEBU stdout:
DEBU stderr: time="2022-04-22T16:29:27+05:30" level=error msg="The
handle is invalid."
failed to remove podman system connection exit status 1:
time="2022-04-22T16:29:27+05:30" level=error msg="The handle is
invalid."
```

This pr send the error to debug logs and not block the delete
operation.

### DIFF
--- a/pkg/crc/machine/delete.go
+++ b/pkg/crc/machine/delete.go
@@ -28,10 +28,10 @@ func (client *client) Delete() error {
 
 	// Remove the podman system connection for crc
 	if err := podman.RemoveRootlessSystemConnection(); err != nil {
-		return err
+		logging.Debugf("Failed to remove podman rootless system connection: %v", err)
 	}
 	if err := podman.RemoveRootfulSystemConnection(); err != nil {
-		return err
+		logging.Debugf("Failed to remove podman rootful system connection: %v", err)
 	}
 
 	if err := cleanKubeconfig(getGlobalKubeConfigPath(), getGlobalKubeConfigPath()); err != nil {


anjannath (approved):

pkg/crc/machine/delete.go
```diff
@@ -28,10 +28,10 @@ func (client *client) Delete() error {
 
 	// Remove the podman system connection for crc
 	if err := podman.RemoveRootlessSystemConnection(); err != nil {
-		return err
+		logging.Debugf("Failed to remove podman rootless system connection: %v", err)
 	}
 	if err := podman.RemoveRootfulSystemConnection(); err != nil {
-		return err
+		logging.Debugf("Failed to remove podman rootful system connection: %v", err)
```
cfergeau, line 34: but now we are hiding errors in the podman preset case.

praveenkumar: @cfergeau yes, we shouldn't block delete operation, if you look https://github.com/code-ready/crc/blob/main/pkg/crc/machine/delete.go#L37-L41 we also ignore here but warn it, do you think warn is better than `Debug`, I can change that also?

cfergeau: The ideal situation would be:
```
func (client *podmanClient) Delete() {
    commonDelete()
    if err := podman.RemoveRootfulSystemConnection(); err != nil {
        ....
    }
}

func (client *openshiftClient) Delete() {
    commonDelete()
    if err := cleanKubeconfig(getGlobalKubeConfigPath(), getGlobalKubeConfigPath()); err != nil {
		if !errors.Is(err, os.ErrNotExist) {
			logging.Warnf("Failed to remove crc contexts from kubeconfig: %v", err)
		}
	}
}
```

This way we never run unneeded code. But we are far from that at the moment.

gbraad:
@cfergeau can you make this change?

cfergeau:
> @cfergeau can you make this change?

Praveen wants to get this in the next release, so no.

anjannath:
create https://github.com/code-ready/crc/issues/3139 to track this and added to next sprint

PR #3136, opened 2022-04-22
Integration: Replace 'CodeReady Containers' to CRC

praveenkumar:
This is missed during 6ef156b1 commit

PR #2982, opened 2022-02-01
[WIP] e2e basic scenario for electron tray

adrianriobo:
Fixes #2979

PR #3138, opened 2022-04-22
Change CodeReady Containers to CRC in extended test folder

jsliacan:
Follow-up to PR #3136

Before:
```bash
$ grep -nr "CodeReady" ./test/
./test/extended/crc/ux/installer/applescripts/install.applescript:3:    set installerWindow to "Install CodeReady Containers"
./test/extended/crc/ux/installer/installer_windows.go:16:	installerWindowTitle string = "CodeReady Containers Setup"
./test/extended/crc/ux/notification/notification_windows.go:18:	startMessage  string = "CodeReady Containers Cluster has started"
./test/extended/crc/ux/notification/notification_windows.go:22:	notificationGroupName string = "CodeReady Containers"
./test/extended/crc/ux/tray/tray_windows.go:28:	notificationIcon string = "CodeReady Containers"
```
After:
```bash
$ grep -nr "CodeReady" ./test/
$
```

### DIFF
--- a/test/extended/crc/ux/installer/installer_windows.go
+++ b/test/extended/crc/ux/installer/installer_windows.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	installerWindowTitle string = "CodeReady Containers Setup"
+	installerWindowTitle string = "CRC Setup"
 
 	installerStartTime time.Duration = 20 * time.Second
 	elementClickTime   time.Duration = 2 * time.Second
--- a/test/extended/crc/ux/installer/installer_windows.go
+++ b/test/extended/crc/ux/installer/installer_windows.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	installerWindowTitle string = "CRC Setup"
+	installerWindowTitle string = "OpenShift Local Setup"
 
 	installerStartTime time.Duration = 20 * time.Second
 	elementClickTime   time.Duration = 2 * time.Second
--- a/test/extended/crc/ux/notification/notification_windows.go
+++ b/test/extended/crc/ux/notification/notification_windows.go
@@ -15,11 +15,11 @@ type gowinxHandler struct {
 }
 
 const (
-	startMessage  string = "CodeReady Containers Cluster has started"
+	startMessage  string = "CRC Cluster has started"
 	stopMessage   string = "Cluster stopped"
 	deleteMessage string = "Cluster deleted"
 
-	notificationGroupName string = "CodeReady Containers"
+	notificationGroupName string = "CRC"
 )
 
 func NewNotification() Notification {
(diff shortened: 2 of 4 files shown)


test/extended/crc/ux/notification/notification_windows.go
```diff
@@ -15,11 +15,11 @@ type gowinxHandler struct {
 }
 
 const (
-	startMessage  string = "CodeReady Containers Cluster has started"
+	startMessage  string = "CRC Cluster has started"
```
cfergeau, line 18: This seems to be coming from `tray-windows`, not `tray-electron`, do we still need to test this?

test/extended/crc/ux/installer/installer_windows.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	installerWindowTitle string = "CodeReady Containers Setup"
+	installerWindowTitle string = "CRC Setup"
```
cfergeau, line 16: Not sure where this string comes from, could not find it by grepping around, have you checked that it's `CRC Setup` and not `OpenShift Local setup`? (I agree CRC is more likely)

anjannath: i think this is for testing the msi, in that case it'd be `Red Hat OpenShift Local setup`

test/extended/crc/ux/notification/notification_windows.go
gbraad: tray notifications show "Red Hat OpenShift Local" if this is still tested.

jsliacan: Thanks everyone, this is then probably true for all 4 files in here, including the macos installer window I assume? I'll see if we still run these (but CodeReady Containers should probably not appear in our code whether we run it or not...)

jsliacan: @gbraad @cfergeau @anjannath I changed them all to _OpenShift Local_. Not sure if we'll be using this code, but changing anyways to not have "CodeReady Containers" in the repo.

cfergeau:
I rebased/force-pushed this to trigger a re-run of CI

cfergeau:
e2e failure looks like the cluster being too slow to start, and unrelated to the changes in this PR:
```
ready/clicumber/testsuite.ExecuteCommandSucceedsOrFails
    When checking that CRC is running                                              # crcsuite.go:271 -> github.com/code-ready/crc/test/e2e/crcsuite.CheckCRCStatus
    output did not match. Expected: '.*OpenShift: .*Running \(v\d+\.\d+\.\d+.*\).*', Actual: 'CRC VM:          Running
OpenShift:       Starting (v4.10.9)
```

cfergeau:
CI is green, so ACK!

adrianriobo:
Not required any manual intervention...`extended` is reused by `e2e` (cli and ux) and `integration`, changes for ux related items are not longer required as they should be handle on #2982 , but as it was discussed we will no longer support those tests.

So yeah this can be merged

anjannath (approved):

PR #3141, opened 2022-04-25
Check for daemonTaskRunning returns nil if correct version of daemon is already running

anjannath:


gbraad (commented):
This makes sense to allow, as it also allows the user to run it manually.
this can happen when scheduled tasks are disabled for the user or otherwise unavailable.

gbraad:
no WIP, just needs testing

/hold

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -117,9 +118,16 @@ func removeDaemonTask() error {
...
+	// First check if a daemon process is running and then check if
+	// the running daemon process is of older version, this way  we
+	// return nil if a daemon process is running that is not old
+	if _, err := daemonclient.GetVersionFromDaemonAPI(); err != nil {
+		if err := olderDaemonVersionRunning(); err != nil {
+			return err
+		}
+		return nil
 	}
+
```
praveenkumar, line 130: lgtm 👍🏼

praveenkumar (approved):

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -117,9 +118,16 @@ func removeDaemonTask() error {
 }
 
 func checkIfDaemonTaskRunning() error {
-	if err := olderDaemonVersionRunning(); err != nil {
-		return err
+	// First check if a daemon process is running and then check if
+	// the running daemon process is of older version, this way  we
+	// return nil if a daemon process is running that is not old
+	if _, err := daemonclient.GetVersionFromDaemonAPI(); err != nil {
```
praveenkumar, line 124: @anjannath if daemonAPI is not running then you will get the err I think condition should be 
```
	if _, err := daemonclient.GetVersionFromDaemonAPI(); err == nil {
		if err := olderDaemonVersionRunning(); err != nil {
			return err
		}
		return nil
	}

```

praveenkumar (changes requested):

pkg/crc/preflight/preflight_daemon_task_check_windows.go
anjannath: yes, thanks, i was pulling my hair why it wasn't working as expected.

anjannath: @praveenkumar Updated

cfergeau: Fwiw, this could be refactored some more if this code stays longer term, `olderDaemonVersionRunning()` is just `GetVersionFromDaemonAPI` + `CheckIfOlderVersion`, so it's suboptimal to wrap it in a call to `GetVersionFromDaemonAPI`. But good enough for now.

Hopefully this won't stay long term anyway, and will only be a temporary fix to get the release out, as I said before, I think we should only support:
- daemon configured to autostart with `crc setup`
- daemon started manually after setting the appropriate `skip-xxx` config options.

Right now, the daemon is not autostarted on Windows, so we have a gap to fill. Checking if the daemon is already running is what was done before, so we can readd this as a stop gap, especially as this is in Windows-specific code.

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -117,9 +118,16 @@ func removeDaemonTask() error {
 }
 
 func checkIfDaemonTaskRunning() error {
-	if err := olderDaemonVersionRunning(); err != nil {
-		return err
+	// First check if a daemon process is running and then check if
+	// the running daemon process is of older version, this way  we
+	// return nil if a daemon process is running that is not old
+	if _, err := daemonclient.GetVersionFromDaemonAPI(); err == nil {
```
praveenkumar, line 124: 👍🏼

praveenkumar (approved):

cfergeau:
> This makes sense to allow, as it also allows the user to run it manually. this can happen when scheduled tasks are disabled for the user or otherwise unavailable.

For what it's worth, I disagree with this as detailed a bit in this comment
https://github.com/code-ready/crc/pull/3141#discussion_r857383185

Preflights check that the system is *exactly* in the state we expect, for example, when there are config files involved, we check that the content of the config file matches exactly what crc expects, we don't limit the check to verifying if the system behaviour matches what we tried to configure. For more concrete examples: on linux, we check precisely the dnsmasq/systemd-resolved configuration, we do not limit the check to "is wildcard DNS working?". On macOS, we check the launchd configuration files to exactly match what we expect, we don't limit the check to "do we have a crc daemon running?"

I don't think the windows daemon should be special here. If scheduled tasks are undesirable, then the user should disable them through `crc config set skip-xxx`. When the daemon is started manually and this config option is set, then I agree crc must be able to start correctly.

Issue #3105, opened 2022-04-08
[WIndows] Explore a way for task scheduler to start a task for user session


praveenkumar:
In windows `crc daemon` process is maintained by task scheduler but when system reboots or user logout-> login then this task is not autostart and user need to again do `crc setup`. Better to find a way on windows also to have it for user session.

PR #3142, opened 2022-04-25
Windows: Add task trigger to start daemon task when user logs in

praveenkumar:
**Fixes:** Issue #3105 

This PR add a task tigger for crcDaemon task to make sure the task
runs as soon as user login after reboot/logout.

## Testing
- crc cleanup
- crc setup
 
=== Reboot the system===
- crc setup --check-only  => this shouldn't fail

### DIFF
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -174,9 +174,6 @@ func getDaemonConfig() (*launchd.AgentConfig, error) {
 }
 
 func checkIfDaemonPlistFileExists() error {
-	if err := olderDaemonVersionRunning(); err != nil {
-		return err
-	}
 	daemonConfig, err := getDaemonConfig()
 	if err != nil {
 		return err
@@ -191,7 +188,7 @@ func checkIfDaemonPlistFileExists() error {
 }
 
 func fixDaemonPlistFileExists() error {
-	if err := olderDaemonVersionRunning(); err != nil {
+	if daemonRunning() {
 		if err := killDaemonProcess(); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


cfergeau (changes requested):
This is a good opportunity to cleanup all the redundancy/workarounds/... that we have accumulated in a short time in the task preflight code.
After this, we only need to handle
- daemon already running on upgrades (unless the daemon is already killed/stopped by the installer during the upgrade)
- the preflights can then only handle the scenario where the daemon was started as a scheduled task
- If the scheduled task preflights are disabled through `skip-xxx` in the config, running crc daemon manually should be possible, and the tray/the command line should work fine (if at all possible, maybe more manual steps would be needed for that to work without impacting too much the rest of the code)

cfergeau:
> daemon already running on upgrades

If the daemon was started manually (either by the user, or by the tray), then kill the instance (can be detected by checking if there's a running daemon task, if not, then check if thedaemon is running, if it is, kill it). If the running daemon is too old, it needs to be stopped and restarted. We previously killed daemon started manually, so this restart can be handled through scheduled task API.

anjannath:
@praveenkumar Can you please fix the CI errors, there's an unused function which is causing the failure
```
pkg/crc/preflight/preflight_checks_nonlinux.go#L66
`olderDaemonVersionRunning` is unused (deadcode)
```

anjannath:
Tested e2e and all of the comments/concerns are addressed:

- daemon task is auto-started 
- upgrade scenario handles old version of the daemon task and is replaced with new one

anjannath (approved):

cfergeau:
This could be slightly polished by stopping the daemon cleanly when it's running already (ie by stopping the task/the launchd agent), and only if this fails, by killing it.

PR #3144, opened 2022-04-25
[wip] [test] move the code checking if scheduled task exists in a separate function

anjannath:
this allows to only check for the scheduled task and return nil
if a daemon process is running

this prevents an error during setup where it was trying to remove
a non existing crcDaemon task as checkIfDaemonTaskRunning fnction
was wrongly returning nil because of a running daemon process

**Fixes:** Issue #N

## Solution/Idea

## Proposed changes

1. Add `start` command.
3. ...

## Testing

7. ...

### DIFF
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -100,7 +100,7 @@ func fixDaemonTaskInstalled() error {
 }
 
 func removeDaemonTask() error {
-	if err := checkIfDaemonTaskRunning(); err == nil {
+	if err := checkIfDaemonScheduledTaskRunning(); err == nil {
 		_, stderr, err := powershell.Execute("Stop-ScheduledTask", "-TaskName", constants.DaemonTaskName)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cfergeau (commented):
Can you add more details about 
> this prevents an error during setup where it was trying to remove a non existing crcDaemon task as checkIfDaemonTaskRunning function was wrongly returning nil because of a running daemon process
I assume this "running daemon process" was not started as a scheduled task, but someone or the tray manually ran `crc daemon`?
The commit looks good to me, though the difference between `checkIfDaemonScheduledTaskRunning` and `checkIfDaemonTaskRunning` is subtle (should `Task` be in the name of the 2nd method?), and we'll likely get it wrong again in the future if we change again this code.
Hopefully this can be all cleaned up after #3142

anjannath:
> I assume this "running daemon process" was not started as a scheduled task, but someone or the tray manually ran crc daemon?

yes, other then a scheduled task, most of the time we were hitting this issue when the tray was running (which started the daemon as a child process)

However closing this in favour of #3146

adrianriobo:
Run basic scenarios with tray and cmdline. Tests were ok

anjannath:
Since #3146 is already merged and this PR also tries to fixe the same issue, this is not needed anymore

Issue #3148, opened 2022-04-26
[Mac] Delay between crc setup and actual daemon start


praveenkumar:
Using launchd we start `crc daemon` process as part of setup but there is bit delay around daemon process is actually running and serving to the requests so in scripted way user might face issues.
```
 ./crc setup && ./crc start
[...]
INFO Checking if crc daemon plist file is present and loaded 
INFO Adding crc daemon plist file and loading it  
Your system is correctly setup for using CRC. Use 'crc start' to start the instance
Is 'crc daemon' running? Cannot reach daemon API: Get "http://unix/api/version": dial unix /Users/prkumar/.crc/crc-http.sock: connect: connection refused
```

```
./crc setup && sleep 10 && ./crc start
[...]
INFO Checking if crc daemon plist file is present and loaded 
INFO Adding crc daemon plist file and loading it  
Your system is correctly setup for using CRC. Use 'crc start' to start the instance
INFO Checking if running as non-root              
INFO Checking if crc-admin-helper executable is cached 
INFO Checking for obsolete admin-helper executable 
INFO Checking if running on a supported CPU architecture 
[...]
```

At the end of setup we should have a check for daemon availability to fulfill the requests.

adrianriobo:
This behavior should be fixed based on #2687 with #2722

praveenkumar:
> This behavior should be fixed based on #2687 with #2722

Both were the tray issue.

adrianriobo:
Well the issue was solved adding the check on the tray but the #2687 was not talking about the tray it said `setup` does not guarantee the daemon valid state..

Now you are properly extended the check :+1:

PR #3150, opened 2022-04-26
Mac/Windows: Check if daemon able to serve after it starts

praveenkumar:
This PR add time limit (15 sec) to make sure daemon able to serve
the requests after it is start because sometimes just after start
it takes time to serve.

fixes: #3148

### DIFF
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/cache"
 	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/daemonclient"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/version"
@@ -278,13 +277,6 @@ func checkIfDaemonPlistFileExists() error {
 	return nil
 }
 
-func daemonRunning() bool {
-	if _, err := daemonclient.GetVersionFromDaemonAPI(); err != nil {
-		return false
-	}
-	return true
-}
-
 func fixDaemonPlistFileExists() error {
 	if err := olderDaemonVersionRunning(); err != nil {
 		if err := killDaemonProcess(); err != nil {
@@ -319,5 +311,5 @@ func fixPlistFileExists(agentConfig launchd.AgentConfig) error {
 		logging.Debugf("failed to restart launchd agent '%s': %v", agentConfig.Label, err.Error())
 		return err
 	}
-	return nil
+	return waitForDaemonRunning()
 }
--- a/pkg/crc/preflight/preflight_checks_nonlinux.go
+++ b/pkg/crc/preflight/preflight_checks_nonlinux.go
@@ -4,10 +4,14 @@
 package preflight
 
 import (
+	"context"
+	"fmt"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/code-ready/crc/pkg/crc/daemonclient"
+	crcerrors "github.com/code-ready/crc/pkg/crc/errors"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/shirou/gopsutil/v3/process"
 )
@@ -69,4 +73,20 @@ func olderDaemonVersionRunning() error {
 		return nil
 	}
 	return daemonclient.CheckIfOlderVersion(v)
+}
+
+func daemonRunning() bool {
+	if _, err := daemonclient.GetVersionFromDaemonAPI(); err != nil {
+		return false
+	}
+	return true
+}
+
+func waitForDaemonRunning() error {
+	return crcerrors.Retry(context.Background(), 15*time.Second, func() error {
+		if !daemonRunning() {
+			return &crcerrors.RetriableError{Err: fmt.Errorf("daemon is not running yet")}
+		}
+		return nil
+	}, 2*time.Second)
 }
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -156,7 +156,7 @@ func fixDaemonTaskRunning() error {
 		logging.Debugf("unable to run the %s task: %v : %s", constants.DaemonTaskName, err, stderr)
 		return err
 	}
-	return nil
+	return waitForDaemonRunning()
 }
 
 func checkIfOlderTask() error {


pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -70,3 +73,13 @@ func olderDaemonVersionRunning() error {
 	}
 	return daemonclient.CheckIfOlderVersion(v)
 }
+
+func checkIfDaemonIsServing() error {
+	return crcerrors.Retry(context.Background(), 15*time.Second, func() error {
```
cfergeau, line 78: 15 seconds is a very long time, if it's not ready to serve requests after a few seconds (3?), I'm not sure it's worth waiting 15 seconds.
Does this have to be done in a separate preflight? Or could this wait be added right after the daemon start, with a big warning if it fails?

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -109,6 +109,13 @@ var daemonLaunchdChecks = []Check{
 		cleanupDescription: "Unloading and removing the daemon plist file",
 		cleanup:            removeDaemonPlistFile,
 	},
+	{
+		configKeySuffix:  "check-daemon-is-serving",
+		checkDescription: "Checking if crc daemon is serving the requests",
+		check:            checkIfDaemonIsServing,
+		fixDescription:   "crc daemon is not started (try running manually `crc daemon` command in different terminal)",
```
cfergeau, line 116: We should not instruct people to run `crc daemon` manually when we are in the process of getting rid of this requirement, especially with all the issues we have when the daemon was already started manually.

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -70,3 +73,13 @@ func olderDaemonVersionRunning() error {
...
+		_, err := daemonclient.GetVersionFromDaemonAPI()
```
cfergeau, line 79: This can reuse `daemonRunning` from #3151.
And I forgot to mention, but I'm not sure it's worth introducing a `serving` state, so far we've always used `running` when checking if the daemon replies to http requests.

pkg/crc/preflight/preflight_checks_nonlinux.go
praveenkumar: 3 sec is not enough, I checked on my machine in the loop and looks like 15 sec is best bet sometime it starts in 7-8 secs but sometime it takes around 10-11 secs when I have lot of I/O happening.

> Does this have to be done in a separate preflight? Or could this wait be added right after the daemon start, with a big warning if it fails?

I find it more clear as part of preflight and that's the reason I named it serving so that make sure after start it serve the requests.

pkg/crc/preflight/preflight_darwin.go
praveenkumar: I can remove that part, my intention was here to get debug info if something really blocked the `crc daemon` which we are not getting with preflight checks.

pkg/crc/preflight/preflight_checks_nonlinux.go
cfergeau: >  it starts in 7-8 secs but sometime it takes around 10-11 secs when I have lot of I/O happening.

What is taking so long before the http endpoint is available? This definitely would be useful data to provide in the commit log.

pkg/crc/preflight/preflight_checks_nonlinux.go
praveenkumar: yes I can reuse the #3151 but as of that that piece is only in `_darwin` preflight check file so either I have to wait it to merge and then relocate this in `_nonlinux` file.

pkg/crc/preflight/preflight_checks_nonlinux.go
cfergeau: > @cfergeau I think till now checking crc daemon process is always worked with running functionality because it is started by some other process (tray) or manually , in both situation this other process took the burden of serving (in case of tray we already waited 8-10 sec) and in case of manual we wait until it start serving.

All I mean, is that (for example) the helper introduced in #3151 is named `daemonRunning`, not `daemonServing`, and if you look at other code doing similar checks, it's always using `running` to describe the check being done (`olderDaemonRunning` for example). We never used `serving` before, and I don't think we need the distinction between `running` and `serving`. So introducing `serving` in this PR is just making things harder to understand.

cfergeau: > I am intend to look into launchd stuff if even after launchd load the process take some time to even start with launchd but it might take some more time and may be for upcoming release.

Are you saying that these 7/8 seconds are caused by launchd?

praveenkumar: @cfergeau I gave it a bit thought and right now I think going with serving make sense (for release/blocker) and after that we can put needed effort to pin point this issue (why it takes more than 3-5 secs, is it because of launchd or something else) . Also we can improve around #3151 helper to use it with this preflight. But as of now I just want to go with the workable solution even if is not best. I am happy to take PR if you put something fast what you suggested and also test it. Tomorrow morning I am going to build final bits for release.

pkg/crc/preflight/preflight_darwin.go
praveenkumar: https://github.com/code-ready/crc/pull/3150#discussion_r858899083 same

pkg/crc/preflight/preflight_checks_nonlinux.go
praveenkumar: https://github.com/code-ready/crc/pull/3150#discussion_r858899083 same here

pkg/crc/preflight/preflight_checks_nonlinux.go
cfergeau: No going with serving does not make sense as all the other recent PRs have been using `running` while doing almost the same tests. And since this adds an additional preflight, this adds 'serving' in user-visible places, so harder to change later. 
To be frank, for something last minute, this PR is quite big (additional preflight, changes in the unit tests, ...). If you want something minimal which we can refine later, then adding a small retry loop right after starting the daemon with launchd/as a scheduled task is a much smaller PR (and thus easier to review, hopefully less risky, ...)

praveenkumar: @cfergeau removed the user facing preflight check.

cfergeau (commented):
Very similar to what I was about to suggest ( https://github.com/cfergeau/crc/commit/cd495eb6fd71d1aff24a0bdb509dfbcc58fabd5d ), so overall looks good to me

pkg/crc/preflight/preflight_checks_nonlinux.go
praveenkumar: @cfergeau this shouldn't be block the PR atm (it's improvement to reuse the code and we can do it later also).

praveenkumar:
/hold

Not considered as release blocker during discussion.

anjannath:
`e2e` failure is not related to the changes in this PR

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -319,5 +319,5 @@ func fixPlistFileExists(agentConfig launchd.AgentConfig) error {
 		logging.Debugf("failed to restart launchd agent '%s': %v", agentConfig.Label, err.Error())
 		return err
 	}
-	return nil
+	return checkIfDaemonIsServing()
```
anjannath, line 322: This function `checkIfDaemonIsServing()` should be called from `fixDaemonPlistFileExists` instead of `fixPlistFileExists` 

its more related to the daemon plist file then a generic plist file

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -142,7 +142,7 @@ func checkIfDaemonTaskRunning() error {
 	if strings.TrimSpace(stdout) != "Running" {
 		return fmt.Errorf("expected %s task to be in 'Running' but got '%s'", constants.DaemonTaskName, stdout)
 	}
-	return nil
+	return checkIfDaemonIsServing()
```
anjannath, line 145: @praveenkumar I think this should be also called from the `fix` function, `fixDaemonTaskRunning` no?

anjannath: just noticed, its done in this way in the commit mentioned here: https://github.com/code-ready/crc/pull/3150#pullrequestreview-954479941

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -142,7 +142,7 @@ func checkIfDaemonTaskRunning() error {
 	if strings.TrimSpace(stdout) != "Running" {
 		return fmt.Errorf("expected %s task to be in 'Running' but got '%s'", constants.DaemonTaskName, stdout)
 	}
-	return nil
+	return waitForDaemonRunning()
```
anjannath, line 145: on a clean system this is always going to fail since we are calling `waitForDaemonRunning` in the `check` itself, and we'll be waiting for the retries to finish.

this `waitForDaemonRunning()` should be called from the `fixDaemonTaskRunning` where after starting the daemon, we retry and block till its able to serve

anjannath (approved):

cfergeau:
Do we know why the 'daemon running' check needs to wait up to 15s? What is taking so long before it's ready?

PR #3151, opened 2022-04-26
Issue #3149 check for existence of daemon process in checkIfDaemonPlistFileExists

### LINKED ISSUE #3149: [BUG] Unable to start crc from tray due to failing preflight check
### General information

  * OS: macOS

### Steps to reproduce

  1. install crc by using the pkg installer 
  2. complete the steps of the on-boarding wizard
  3. the tray icon appears, now click on start
  
Nothing will happen, as in the status on the tray doesn't change to `Starting`, if you check the `About` page or the `Configurations` page, it all works as expected and communicates with the daemon properly

Upon checking the logs in `~/.crc/crcd.log` and following the lines after the `start` request was sent to daemon, you can see the following failing preflight check, which prevents the start (tray doesn't show an error about this)

```bash
ime="2022-04-26T11:32:20+02:00" level=debug msg="Running 'bash -c launchctl list | grep com.redhat.crc.daemon | awk '{print $1}''"
time="2022-04-26T11:32:20+02:00" level=debug msg="launchd agent 'com.redhat.crc.daemon' is not running"
```

anjannath:
Fixes #3149 

the daemon process can be started by the user or the tray, we consider those
processes also to not prevent start if daemon is not started from launchd

### DIFF
--- a/pkg/crc/preflight/preflight_checks_darwin.go
+++ b/pkg/crc/preflight/preflight_checks_darwin.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/code-ready/crc/pkg/crc/cache"
 	"github.com/code-ready/crc/pkg/crc/constants"
+	"github.com/code-ready/crc/pkg/crc/daemonclient"
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/crc/version"
@@ -271,10 +272,17 @@ func checkIfDaemonPlistFileExists() error {
 	if err := launchd.CheckPlist(*daemonConfig); err != nil {
 		return err
 	}
-	if !launchd.AgentRunning(daemonConfig.Label) {
+	if !launchd.AgentRunning(daemonConfig.Label) && !daemonRunning() {
 		return fmt.Errorf("launchd agent '%s' is not running", daemonConfig.Label)
 	}
 	return nil
+}
+
+func daemonRunning() bool {
+	if _, err := daemonclient.GetVersionFromDaemonAPI(); err != nil {
+		return false
+	}
+	return true
 }
 
 func fixDaemonPlistFileExists() error {


pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -271,7 +272,9 @@ func checkIfDaemonPlistFileExists() error {
 	if err := launchd.CheckPlist(*daemonConfig); err != nil {
 		return err
 	}
-	if !launchd.AgentRunning(daemonConfig.Label) {
+	_, err = daemonclient.GetVersionFromDaemonAPI()
+
+	if !launchd.AgentRunning(daemonConfig.Label) || err != nil {
```
praveenkumar, line 277: @anjannath conditions should be `if !launchd.AgentRunning(daemonConfig.Label) && err != nil` so that both need to be true before returning this error.

praveenkumar (changes requested):

pkg/crc/preflight/preflight_checks_darwin.go
cfergeau: Might be worth moving this to a small `func daemonRunning() bool` helper to make this more readable/less error-prone

cfergeau:
Just noticed this code:
https://github.com/code-ready/crc/blob/ac9efa78250c677c1da56fe924f2d6bcc7219a70/pkg/crc/preflight/preflight_checks_darwin.go#L310-L313
so we always restart the launchd service if there were some changes.
It might be better to rework the macos `olderDaemonVersionRunning` check to always kill the daemon (regardless of version) if it was not started by launchd

pkg/crc/preflight/preflight_checks_darwin.go
anjannath: @praveenkumar @cfergeau updated

anjannath:
> It might be better to rework the macos olderDaemonVersionRunning check to always kill the daemon (regardless of version) if it was not started by launchd

https://github.com/code-ready/crc/blob/ac9efa78250c677c1da56fe924f2d6bcc7219a70/pkg/crc/preflight/preflight_checks_darwin.go#L280-L285

so in the above code, instead of doing `olderDaemonVersionRunning` we check `daemonRunning`, and then we don't need to have this 'short-circuit' in the `checkDaemonPlistFlieExists`

cfergeau:
> > It might be better to rework the macos olderDaemonVersionRunning check to always kill the daemon (regardless of version) if it was not started by launchd
> 
> https://github.com/code-ready/crc/blob/ac9efa78250c677c1da56fe924f2d6bcc7219a70/pkg/crc/preflight/preflight_checks_darwin.go#L280-L285
> 
> so in the above code, instead of doing `olderDaemonVersionRunning` we check `daemonRunning`, and then we don't need to have this 'short-circuit' in the `checkDaemonPlistFlieExists`

Yeah, something like this, for post-release cleanup :)

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -271,12 +272,19 @@ func checkIfDaemonPlistFileExists() error {
...
 		return fmt.Errorf("launchd agent '%s' is not running", daemonConfig.Label)
 	}
 	return nil
 }
 
+func daemonRunning() bool {
+	if _, err := daemonclient.GetVersionFromDaemonAPI(); err != nil {
+		return false
+	}
+	return true
```
cfergeau, line 285: A shorter alternative is 
```
_, err := daemonclient.GetVersionFromDaemonAPI()
return (err == nil)
```
but your version is fine with me.

cfergeau (approved):

PR #3155, opened 2022-04-29
Fix help.github.com link checks

cfergeau:
For some reason, help.github.com will only serve compressed content, and
will reply with a 403 error without any "Accept-encoding" headers.
This causes 'make docs_check_links' to report help.github.com links as being
dead.
This issue is tracked in https://github.com/github/docs/issues/17358

For this fix to work, quay.io/crcont/docs-builder:latest must be rebuilt
and pushed to quay.
This can be rebuilt and tested locally with:

$ (cd images/docs-builder && podman build -t quay.io/crcont/docs-builder:latest .)
$ make docs_check_links

### DIFF
--- a/images/docs-builder/Dockerfile
+++ b/images/docs-builder/Dockerfile
@@ -12,7 +12,7 @@ RUN gem install asciidoctor && \
 
 COPY entrypoint.sh /sbin/entrypoint.sh
 RUN chmod 755 /sbin/entrypoint.sh
-COPY links_ignorelist.json /links_ignorelist.json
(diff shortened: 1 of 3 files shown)


cfergeau:
> Why no github action run on this PR?

Maybe because it was initially based on `origin/master` which has no github actions instead of `origin/main`

cfergeau:
(NB: the "build and publish docs" failure is expected as the quay.io docs-build image hasn't been updated yet with the changes from this PR)

PR #3160, opened 2022-05-04
macos-pkg: apply required entitlements to vfkit binary during pkg build

anjannath:
although the required entitlements are already applied to the released vfkit
binary, they gets overwritten when codesiging again during the pkg build
process, so we download the entitlements file from code-ready/vfkit and re-apply

### DIFF
--- a/packaging/package.sh
+++ b/packaging/package.sh
@@ -3,10 +3,22 @@ set -euxo pipefail
 
 BASEDIR=$(dirname "$0")
 OUTPUT=$1
-GOARCH=$(go env GOARCH)
 CODESIGN_IDENTITY=${CODESIGN_IDENTITY:-mock}
 PRODUCTSIGN_IDENTITY=${PRODUCTSIGN_IDENTITY:-mock}
 NO_CODESIGN=${NO_CODESIGN:-0}
+
+case "$(uname -m)" in
+    "x86_64")
+        GOARCH="amd64"
+        ;;
+    "arm64")
+        GOARCH="arm64"
+        ;;
+    *)
+        echo "Unknown arch, exiting"
+        exit 1
+        ;;
+esac
 
 function sign() {
   if [ "${NO_CODESIGN}" -eq "1" ]; then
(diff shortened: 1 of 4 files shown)


packaging/package.sh
```diff
@@ -3,11 +3,15 @@ set -euxo pipefail
 
 BASEDIR=$(dirname "$0")
 OUTPUT=$1
-GOARCH=$(go env GOARCH)
+GOARCH=amd64
```
anjannath, line 6: i kept the name as `GOARCH`. but if others think its better to change then i'll update to `MARCH` or just `ARCH`

anjannath:
Note: waiting for a signed pkg to test from rcm

anjannath:
> The biggest question is why are we always resigning everything? Ideally, we'd only sign vfkit once per vfkit release, and reuse the signed binary in crc, same for admin-helper.

indeed, It'd be better if while doing the `vfkit` release we can publish signed binaries, (its just that we have to follow the whole rcm process) provided the releases of `vfkit`, `admin-helper` will be of lesser frequency this is something we should look into

packaging/package.sh
anjannath: Updated! PTALA

packaging/package.sh
```diff
@@ -3,11 +3,23 @@ set -euxo pipefail
...
 OUTPUT=$1
-GOARCH=$(go env GOARCH)
 CODESIGN_IDENTITY=${CODESIGN_IDENTITY:-mock}
 PRODUCTSIGN_IDENTITY=${PRODUCTSIGN_IDENTITY:-mock}
 NO_CODESIGN=${NO_CODESIGN:-0}
 
+case "$(uname -m)" in
+    "x86_64")
+         GOARCH="amd64"
+        ;;
```
cfergeau, line 13: Indentation might be a little bit off here?

cfergeau (approved):

packaging/package.sh
anjannath: yeah, will fix this and merge

Issue #2538, opened 2021-07-06
[e2e] integrate clicumber as part of extended test 


adrianriobo:
As per  https://github.com/code-ready/crc/pull/2532#issuecomment-874308677 may it is worth to consider integrate clicumber code on extended test code of crc project.

PR #3167, opened 2022-05-09
[QE] Compactify e2e tests

jsliacan:
This PR contains the following changes
- include the relevant clicumber code in this codebase, as suggested in #2538 and #3210 
- update dependencies (cucumber/godog/...) 
- compactify features/scenarios to fit the cucumber paradigms
- config.feature retains its old form for now (should be replaced by an integration test of similar extent #3249)


adrianriobo:
Does it make sense keep the clicumber integrated code under e2e? 
Or may keep e2e only with `suites` and `features` and move the clicumber code to `extended`?

WDYT?

jsliacan:
@adrianriobo , I moved it directly into `testsuite` because testsuite was "extending" the clicumber suite. That said, I am happy to put it wherever you think is better, I have no real opinions either way.

anjannath:
@jsliacan pr needs rebase

anjannath:
nice speedup for integration tests!

| test time  | e2e  | integration  |
|---|---|---|
| this pr  | 1h52m46s  | 1h12m18s  |
| in main  | 2h31m8s  | 1h23m38s  |

anjannath:
> [Include clicumber code, update dependencies, compactify feature files](https://github.com/code-ready/crc/pull/3167/commits/fa373ddb6068e949feebd66a142e5da9b01df5dd)

i guess these three things are tightly coupled and hard to split into individual commits without rewriting the `feature` files multiple times for each commit.. better to also include the information in the [pr description] as commit log

anjannath (approved):

anjannath:
> leave out config.feature (should be replaced by an integration test of similar extent https://github.com/code-ready/crc/issues/3249)

@jsliacan maybe keep the `config.feature` for now (if it works with the current changes) and remove it when we have something similar in the `integration` tests?

developing.adoc
```diff
@@ -83,9 +83,7 @@ We have automated e2e tests which keep CodeReady Containers in shape.
 
 [[intro-to-e2e-testing]]
 === Introduction
-End-to-end (e2e) tests use the link:http://github.com/code-ready/clicumber[Clicumber] package to provide basic functionality for testing CLI binaries.
-
-Clicumber allows running commands in a persistent shell instance (`bash`, `tcsh`, `zsh`, Command Prompt, or PowerShell), assert its outputs (standard output, standard error, or exit code), check configuration files, and so on. The general functionality of Clicumber is then extended by CodeReady Containers specific test code to cover the whole functionality of CodeReady Containers.
+End-to-end (e2e) tests borrow code from link:http://github.com/code-ready/clicumber[Clicumber] package to provide basic functionality for testing CLI binaries. This facilitates running commands in a persistent shell instance (`bash`, `tcsh`, `zsh`, Command Prompt, or PowerShell), assert its outputs (standard output, standard error, or exit code), check configuration files, and so on. The general functionality of Clicumber is then extended by CodeReady Containers specific test code to cover the whole functionality of CodeReady Containers.
```
praveenkumar, line 86: Shouldn't we update https://github.com/code-ready/clicumber codebase to use latest version and then consume it as we are doing now?

jsliacan: For easier maintenance etc., the idea was to include it on our side instead. I mention #3210 in the description above to hint at this.

anjannath:
better to exclude the `config.feature` changes from the first commit itself than removing it and re-adding

anjannath (approved):

developing.adoc
praveenkumar: @jsliacan which mean we are not using  https://github.com/code-ready/clicumber anymore?

jsliacan: Correct, after this PR we won't be importing clicumber project.

test/e2e/features/basic.feature
```diff
@@ -22,165 +22,69 @@ Feature: Basic test
...
+        Then stderr should contain "Machine does not exist. Use 'crc start' to create it"
 
     @linux
-    Scenario: CRC setup on Linux
+    Scenario: CRC setup
         When executing "crc setup --check-only" fails
         Then executing crc start command fails
         And executing single crc setup command succeeds
 
-    @linux
```
praveenkumar, line 44: Why we are removing this test?

test/e2e/features/basic.feature
```diff
@@ -22,165 +22,69 @@ Feature: Basic test
...
-        And stderr should contain "Too little memory"
-        And setting config property "memory" to value "16000" succeeds
-
-    @darwin @linux @windows
-    Scenario: CRC start
+    Scenario: CRC start usecase
+        # Request start with monitoring stack
+        * setting config property "enable-cluster-monitoring" to value "true" succeeds
+        * setting config property "memory" to value "16000" succeeds
+        Given executing crc setup command succeeds
```
praveenkumar, line 51: Also now we are only testing if user provide correct mem value what about the error it display if user doesn't provide the correct mem like it use to be?

test/e2e/features/basic.feature
```diff
@@ -22,165 +22,69 @@ Feature: Basic test
...
         When starting CRC with default bundle succeeds
         Then stdout should contain "Started the OpenShift cluster"
         # Check if user can copy-paste login details for developer and kubeadmin users
-        And stdout should match "(?s)(.*)oc login -u developer https:\/\/api\.crc\.testing:6443(.*)$"
-        And stdout should match "(?s)(.*)https:\/\/console-openshift-console\.apps-crc\.testing(.*)$"
-
-    @darwin @linux @windows
-    Scenario: CRC status and disk space check
+        * stdout should match "(?s)(.*)oc login -u developer https:\/\/api\.crc\.testing:6443(.*)$"
```
praveenkumar, line 55: what does `*` means because some place you used `And` and some place it is replaced by `*` ?

test/e2e/features/basic.feature
anjannath: there's a separate `config.feature` and its covered there, so should be good i think

test/e2e/features/basic.feature
anjannath: > Gherkin also supports using an asterisk (*) in place of any of the normal step keywords. This can be helpful when you have some steps that are effectively a list of things, so you can express it more like bullet points where otherwise the natural language of And etc might not read so elegantly.

from: https://cucumber.io/docs/gherkin/reference/#Asterisk

test/e2e/features/basic.feature
praveenkumar: @anjannath I don't see any `config.feature` file, I think @jsliacan mentioned he will create separate task issue for those missing stuff. so once that in place we should take this PR in and iterate over.

test/e2e/features/basic.feature
jsliacan: Yes yes, it's just a bullet point. I use them somewhere when it's an obvious "enumeration" of things to check. But it's just for fun in a way, `And` would've worked just as well.

test/e2e/features/basic.feature
anjannath: `config.feature` exists from before, this was removed in an earlier iteration of the PR but the current PR branch doesn't remove that or am i missing something?

jsliacan: Yes, I'll make an issue and it'll go into the config integration test. I don't think it's in the current `config.feature`. It's not the most important testcase though and the minimum we declare should work is 14GiB ([docs](https://access.redhat.com/documentation/en-us/red_hat_openshift_local/2.5/html/getting_started_guide/administrative-tasks_gsg#starting-monitoring_gsg)), so maybe I'll use this value instead of 16000 that we had.

test/e2e/features/basic.feature
jsliacan: I removed it because it cannot run in parallel with the rest of the scenarios in the Feature. Tagging would make it hard to read & cluttered if I included it in the next Scenario. So it's going to go in the integration test. I'll list it in the issue of things to move to integration.

test/e2e/features/basic.feature
jsliacan: I updated the issue with both things @praveenkumar noticed: https://github.com/code-ready/crc/issues/3249 and altered the basic.feature and config.feature a little. I moved the "missing setup" scenario to config.feature (which will be replaced by an integration test) and cleaned up scenarios a little more in `basic.feature` which will stay in e2e.

PR #3176, opened 2022-05-12
Improve a few user-visible messages

cfergeau:
They should flow slightly better this way.


pkg/crc/preflight/preflight_windows.go
```diff
@@ -140,7 +140,7 @@ var adminHelperServiceCheks = []Check{
 		configKeySuffix:  "check-admin-helper-service-running",
 		checkDescription: "Checking admin helper service is running",
 		check:            checkIfAdminHelperServiceRunning,
-		fixDescription:   "Make sure you installed the crc using installer",
+		fixDescription:   "Make sure you installed crc using its Windows installer",
```
praveenkumar, line 143: 👍🏼

praveenkumar (approved):

PR #3177, opened 2022-05-16
Daemon: Add containers.internal DNS record for podman-4.x to work

praveenkumar:
With podman 4.x when user run the container with exposed port on mac/windows
then internally port is exposed to virtual network (using gvproxy) and for podman
3.x it was `host.crc.testing:7777` but now changed to `gateway.containers.internal`.

This PR is add the `gateway.containers.internal` DNS record so running the containers
with exposed port on crc also works as expected for podman 4.x

```
macos-amd64 git:(podman_4.x) ✗ podman run -d -p 8080:80 docker.io/httpd:2.4
766bb336d71402d4747d658f17f139f6d60b8fde238bb4aedd1aa45d96ae24de

[core@podman ~]$ curl gateway.containers.internal/services/forwarder/all | jq .
[
  {
    "local": "/Users/prkumar/.crc/machines/crc/docker.sock",
    "remote": "ssh-tunnel://core@192.168.127.2:22/run/podman/podman.sock?key=%2FUsers%2Fprkumar%2F.crc%2Fmachines%2Fcrc%2Fid_ecdsa",
    "protocol": "unix"
  },
  {
    "local": "127.0.0.1:2222",
    "remote": "192.168.127.2:22",
    "protocol": "tcp"
  },
  {
    "local": "127.0.0.1:9090",
    "remote": "192.168.127.2:9090",
    "protocol": "tcp"
  },
  {
    "local": ":8080",
    "remote": "192.168.127.2:8080",
    "protocol": "tcp"
  }
]
```

### DIFF
--- a/cmd/crc/cmd/daemon.go
+++ b/cmd/crc/cmd/daemon.go
@@ -93,6 +93,15 @@ var daemonCmd = &cobra.Command{
 						},
 					},
 				},
+				{
+					Name: "containers.internal.",
+					Records: []types.Record{
+						{
+							Name: "gateway",
+							IP:   net.ParseIP(hostVirtualIP),
+						},
+					},
+				},
 			},
 			Protocol: types.HyperKitProtocol,
 		}
@@ -150,7 +159,7 @@ func run(configuration *types.Configuration) error {
 		}
 	}()
 
-	networkListener, err := vn.Listen("tcp", fmt.Sprintf("%s:7777", hostVirtualIP))
+	networkListener, err := vn.Listen("tcp", fmt.Sprintf("%s:80", hostVirtualIP))
 	if err != nil {
 		return err
 	}


cmd/crc/cmd/daemon.go
```diff
@@ -150,7 +159,7 @@ func run(configuration *types.Configuration) error {
 		}
 	}()
 
-	networkListener, err := vn.Listen("tcp", fmt.Sprintf("%s:7777", hostVirtualIP))
+	networkListener, err := vn.Listen("tcp", fmt.Sprintf("%s:80", hostVirtualIP))
```
cfergeau, line 162: If I read the PR description correctly, this will break podman 3 support? At this point I'm missing the big picture regarding when we are going to start using podman 4, so it's hard to give meaningful reviews.

cfergeau (changes requested):
Please clarify if it's fine to replace podman 3 support with podman 4.

cmd/crc/cmd/daemon.go
cfergeau: My expectation is that we'll be getting podman4 with OpenShift 4.11 and with fcos 36?

gbraad: Podman 4.x only

gbraad (approved):

cmd/crc/cmd/daemon.go
praveenkumar: @cfergeau It will not break anything in respect to openshift bundle and add support for fcos36 which going to be in the stable channel soon (as of now available in testing/next stream).

praveenkumar:
/hold

Till fcos-36 bundles are generated and mirrored.

cmd/crc/cmd/daemon.go
cfergeau: > It will not break anything in respect to openshift bundle

Why? podman 3 from OpenShift 4.10 bundles is fine with `vn.Listen("tcp", fmt.Sprintf("%s:80", hostVirtualIP))` instead of `vn.Listen("tcp", fmt.Sprintf("%s:7777", hostVirtualIP))`?

praveenkumar: @cfergeau yes, because podman 3  which is part of openshift-4.x doesn't even support gvproxy bits so if you are using openshift bundle and then follow something below doesn't work and this change doesn't change it.

```
<-- start crc with openshift bundle -->
$ eval $(crc podman-env)
$ podman run -d -p 8080:80 docker.io/httpd:2.4
$ curl -IK localhost:8080
will show error.
```

praveenkumar:
Depend on https://github.com/code-ready/snc/pull/528 one.

anjannath:
Tested locally with 4.0.2 podman and works as expected

/lgtm /approve

PR #3178, opened 2022-05-16
Update golang version to 1.17

praveenkumar:


.github/workflows/publish-docs.yml
```diff
@@ -13,6 +13,8 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
+        go:
+          - 1.17
```
cfergeau, line 17: Is this using go?

images/build-e2e/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/go-toolset:1.16.7 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.17.10 as builder
```
cfergeau, line 1: Only go-toolset:1.17.7 is available.

images/build-integration/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/go-toolset:1.16.7 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.17.10 as builder
```
cfergeau, line 1: Same here

.github/workflows/publish-docs.yml
praveenkumar: I used the `update-go-version.sh` and looks like https://github.com/code-ready/crc/blob/main/update-go-version.sh#L18-L20 is culprit to adding it.

images/build-e2e/Dockerfile
praveenkumar: Yeah it is again https://github.com/code-ready/crc/blob/main/update-go-version.sh#L15-L16, looks like after executing this script we have to manually check or if there is way to get latest toolset version from registry.

cfergeau: Yeah, it had the same issue with 1.16, the version needed to be manually adjusted.

praveenkumar: @cfergeau Updated.

.github/workflows/publish-docs.yml
praveenkumar: updated

images/build-integration/Dockerfile
praveenkumar: updated manually

cfergeau (approved):

PR #3179, opened 2022-05-17
Image: Add docker file to run vm on azure using container

praveenkumar:
This is consumed by openshift/release repo to create a job to test
crc on windows, using azure.


images/openshift-ci-azure/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# This Dockerfile is used by openshift CI
+
+FROM registry.ci.openshift.org/openshift/release:golang-1.16 AS builder
```
cfergeau, line 3: 1.17?

images/openshift-ci-azure/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+FROM quay.io/centos/centos:stream8
+COPY --from=builder /go/src/github.com/code-ready/crc /opt/crc
+COPY --from=builder /go/src/github.com/code-ready/crc/out/windows-amd64/crc.exe /bin/crc.exe
+COPY --from=builder /go/src/github.com/code-ready/crc/images/openshift-ci/mock-nss.sh /bin/mock-nss.sh
+COPY --from=builder /go/src/github.com/code-ready/crc/images/openshift-ci-azure/azure-cli.repo /etc/yum.repos.d/azure-cli.repo
+
+
+RUN yum update -y && \
+    yum install --setopt=tsflags=nodocs -y \
```
cfergeau, line 16: Why the `tsflags=nodocs`?

images/openshift-ci-azure/azure-cli.repo
```diff
@@ -0,0 +1,6 @@
+[azure-cli]
+name=Azure CLI
+baseurl=https://packages.microsoft.com/yumrepos/azure-cli
+enabled=1
+gpgcheck=1
+gpgkey=https://packages.microsoft.com/keys/microsoft.asc
```
cfergeau, line 6: I'd favour importing this gpg key in our repo. If we import the gpg key from the same location where we download the packages, it does not add a lot of trust.

images/openshift-ci-azure/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+    nss_wrapper \
+    sshpass \
+    unzip \
+    azure-cli \
+    jq \
+    openssh-clients && \
```
cfergeau, line 22: Not clear why unzip/jq are needed here, is it possible to add a link to what will be using this image to get the full picture?

images/openshift-ci-azure/Dockerfile
praveenkumar: To make sure we don't install the doc for the packages we install to keep the image size minimal.

images/openshift-ci-azure/azure-cli.repo
praveenkumar: @cfergeau we are using same https://github.com/code-ready/crc/blob/main/images/openshift-ci/google-cloud-sdk.repo#L7-L8 for gcp but may be a separate PR to do it for both place?

images/openshift-ci-azure/Dockerfile
cfergeau: But is the image size that constrained? This feels like premature optimization :)

images/openshift-ci-azure/Dockerfile
cfergeau:
```
AZURE_AUTH_TENANT_ID=$(cat $AZURE_AUTH_LOCATION | jq -r .tenantId)
AZURE_SUBSCRIPTION_ID=$(cat $AZURE_AUTH_LOCATION | jq -r .subscriptionId)
```

But this is not part of this PR, again, very hard to review this image generation without having any clue how it's going to be used.

praveenkumar: @cfergeau yes it is not part of this PR because this image is going to used in the openshift-CI side so once this is merge then only I can create a PR on openshift/release side which will have those steps.

cfergeau: I don't know what I'm expected to do during this review then. "Yes this is a Dockerfile which will create an image with the azure client, a crc binary, crc sources, and a few random binaries.", which is not that useful.

cfergeau: By the way, maybe it would be simpler to have a single image for both gcp and azure?

praveenkumar: > By the way, maybe it would be simpler to have a single image for both gcp and azure?

@cfergeau yes I did think so but then we are increasing image size unnecessary to have additional packages, that's why I have it separate.

> "Yes this is a Dockerfile which will create an image with the azure client, a crc binary, crc sources, and a few random binaries.", which is not that useful.

To be honest that is only I want as review because use for this image is not in the repo so I can't provide much info.

cfergeau: > then we are increasing image size unnecessary

What are the exact requirements we have on image size? If it is so important, why are we using centos/stream8 when ubi8 is about twice smaller? Why isn't there an additional base image with all the common packages between gcp and azure, and then azure and gcp reuse this base image, this would also save some space on the image repository, and on systems installing both.

praveenkumar: > What are the exact requirements we have on image size? I

@cfergeau I don't know if it is mentioned somewhere, I was just going with container best practice to have only stuff in single image which is required. But I think if we use same then it will also remove our maintenance burden so made the change.

cfergeau (approved):

Issue #3174, opened 2022-05-11
Hard to diagnose failure if OPENSHIFT_VERSION and bundleLocations are out of sync


cfergeau:
If `OPENSHIFT_VERSION` in `Makefile` is changed to say 4.10.12, and `bundleLocations` in pkg/crc/machine/bundle/constants.go is not changed to also point at 4.10.12 bundles, a weird failure happens:
- `crc setup` downloads a bundle named ***-4.10.12.crcbundle
- it unpacks it
- but then it fails with `Temporary error: rename /home/teuf/.crc/cache/tmp-extract/crc_libvirt_4.10.12_amd64 /home/teuf/.crc/cache/crc_libvirt_4.10.12_amd64: no such file or directory (x12)`

This happens because it downloads an older bundle as instructed by `bundleLocations`, but it names it `xxx_4.10.12` instead of keeping the older version.

Not fully sure about the best way to fix this, maybe add a 'version' field to `bundleLocations` so that we can sanity check the expected version against what we can download.

PR #3189, opened 2022-05-24
Detect mismatch of bundle version from Makefile and bundle/constant.go

praveenkumar:
This PR detect sync of the `OPENSHIFT_VERSION` and `PODMAN_VERSION` from the
makefile to bundle locations which is part of `bundle/constants.go` and
error out when it is out of sync.

If the Makefile have 4.10.13 openshift version but this is not
synced to mirror then on main branch following will happen.

```
$ crc setup
[...]
INFO Checking if /home/prkumar/.crc/cache/crc_libvirt_4.10.13_amd64.crcbundle exists
INFO Downloading crc_libvirt_4.10.13_amd64.crcbundle
expected crc_libvirt_4.10.12_amd64.crcbundle but found crc_libvirt_4.10.13_amd64.crcbundle in Makefile
```

**Fixes:** Issue #3174

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -268,6 +268,13 @@ func Download(preset preset.Preset) error {
 	if err != nil {
 		return err
 	}
+	bundleNameFromURI, err := downloadInfo.GetSourceFilename()
+	if err != nil {
+		return err
+	}
+	if constants.GetDefaultBundle(preset) != bundleNameFromURI {
+		return fmt.Errorf("expected %s but found %s in Makefile", bundleNameFromURI, constants.GetDefaultBundle(preset))
+	}
 	if _, err := downloadInfo.Download(constants.GetDefaultBundlePath(preset), 0664); err != nil {
 		return err
 	}
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -4,7 +4,9 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"net/http"
+	"net/url"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
@@ -95,4 +97,12 @@ func (r *RemoteFile) Download(bundlePath string, mode os.FileMode) (string, erro
 
 func (r *RemoteFile) GetSha256Sum() string {
 	return r.sha256sum
+}
+
+func (r *RemoteFile) GetSourceFilename() (string, error) {
+	u, err := url.Parse(r.uri)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Base(u.Path), nil
 }


cfergeau (changes requested):
This can't work this way, updating the makefile will silently break the sha256sum which will no longer match.
This needs to be handled somehow, but not sure how. Maybe by adding a 'version' field to `bundlesDownloadInfo` and doing 'something' if this does not match the version specified in the Makefile?

praveenkumar:
> This can't work this way, updating the makefile will silently break the sha256sum which will no longer match. This needs to be handled somehow, but not sure how. Maybe by adding a 'version' field to `bundlesDownloadInfo` and doing 'something' if this does not match the version specified in the Makefile?

@cfergeau Doesn't it responsibility of the dev who is filling the PR after bundle is sync on mirror to update the sha256sum also? This also doesn't break silently as a dev/user you get error around sha256sum mismatching which is much better than current situation.

cfergeau:
> > This can't work this way, updating the makefile will silently break the sha256sum which will no longer match. This needs to be handled somehow, but not sure how. Maybe by adding a 'version' field to `bundlesDownloadInfo` and doing 'something' if this does not match the version specified in the Makefile?
> 
> @cfergeau Doesn't it responsibility of the dev who is filling the PR after bundle is sync on mirror to update the sha256sum also? This also doesn't break silently as a dev/user you get error around sha256sum mismatching which is much better than current situation.

My understanding is that the issue that caused you to file this PR is because the version in the Makefile was updated, but `bundle/constants.go` was not updated at the same time. Since we are already failing to keep the 2 in sync, I don't see why after this change we would always remember to update it?

> This also doesn't break silently as a dev/user you get error around sha256sum mismatching which is much better than current situation.

No it's not, it's as broken. You wait for quite some time to get the bundle, and then the sha256sum does not match and you have no clue why unless you know of this PR and its caveat. Maybe it's the download which was corrupted?

I agree we need to improve/fix what you described in this PR, but not by breaking something else instead.

cfergeau:
> I agree we need to improve/fix what you described in this PR, but not by breaking something else instead.

Ah, my bad, the PR description does **not** describe what issue it is fixing, it only describes the behaviour *after* the PR. It's useful to have both...

> No it's not, it's as broken. You wait for quite some time to get the bundle, and then the sha256sum does not match and you have no clue why unless you know of this PR and its caveat. Maybe it's the download which was corrupted?

After this PR, this won't be happening, only a `404` error will be shown which is better than the initial error (for other readers, it's described in https://github.com/code-ready/crc/issues/3174 ). But as soon as an updated version is uploaded, it will start failing with what I described unless the uploader remembers to update crc's code, which may or may not happen. So I still don't think this is great to do.

praveenkumar:
@cfergeau what if we use https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/bundles/openshift/4.10.12/sha256sum.txt this as first level of checking and warn developer to update the sha256sum of the bundles if it is not matched `sha256sum.txt` is small file and can be fetched very fast?

praveenkumar:
> https://github.com/cfergeau/crc/commits/bundle-version-check adds a unit test to verify that Makefile and pkg/crc/machine/bundle/constants.go are in sync. You could add the sha256sum checking that you mention to it.

@cfergeau this unit test will fail in case of add-hoc bundle testing as part of CI ( crc side the way we get the bundle is https://github.com/openshift/release/blob/master/ci-operator/step-registry/code-ready/crc/e2e/test/code-ready-crc-e2e-test-commands.sh#L16) so till now as soon as we create the bundle we upload it to GCP and put a PR here with Makefile change after this change it will fail during the `make test`.

cfergeau:
If we have to update `Makefile` and `bundle/constants.go` in separate PRs, then maybe we can rework the bundle download information to be flat and with more metadata?
```
var bundleLocations = []bundleDownloadInfo {
  {
    Arch: "amd64",
    OS: "linux",
    Preset: preset.Podman,
    Version: "3.4.4",
    URL: "https://mirror.openshift.com/pub/openshift-v4/clients/crc/bundles/podman/3.4.4/crc_podman_libvirt_3.4.4_amd64.crcbundle"
    Sha256Sum: "8e8d150dfbefcec93639df53e6181377d1bfc0df3c8719a6fedf0929779f6f63"
  },
  {
    ...
  },
}
```
Then when we try to download the bundle/check its sha256sum/... and Makefile says podman version in 4.0.2, then we know we only have information for podman version 3.4.4, and we can return a proper error.

praveenkumar:
> I did not reopen the code, but maybe we can check that the end of the URL matches the filename we want to download to, and error out if they are not the same?

@cfergeau I think that is much simple and effective so implemented this one.

cfergeau:
> `GetFilenameFromURI`?

Even `GetFilename`, we are operating on a `RemoteFile` so the  fact that the method is dealing with an URI is implicit

cfergeau (approved):
Looks good, though I'm still not sure whether the unit test I suggested would be workable or not.

praveenkumar:
> Looks good, though I'm still not sure whether the unit test I suggested would be workable or not.

@cfergeau The steps you put https://github.com/code-ready/crc/pull/3189#issuecomment-1135637356 around bundle testing are correct one. To make unit test to work we need to update the bundle version in makefile and constants.go as single PR but most of the time we create intermediate bundles also which might be not part of release so I think it might be overkill just to test bundle in CI every time we need to  update url and sha hash.

cfergeau:
> I think it might be overkill just to test bundle in CI every time we need to update url and sha hash.

On the other hand, it's questionable that we break `crc setup` for manual testing when we want to get bundle automated testing. Dunno if there would be a way of making all of this better?

PR #3191, opened 2022-05-26
introduce api versioning and move current apiset to v1

### LINKED ISSUE #2543: Add versionning for http API
Now that we can remove the binary protocol and the http API is used by both tray, it might be a good idea to introduce versionned API. 

The proposition is to copy the current http API under /v1 for instance. The next breaking change could use /v2, and so on.

anjannath:
fixes #2543

### DIFF
--- a/pkg/crc/api/helpers.go
+++ b/pkg/crc/api/helpers.go
@@ -5,9 +5,22 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 	"sync"
 
 	"github.com/code-ready/crc/pkg/crc/logging"
+)
+
+
+const (
+	// AcceptVersionHeaderKey is the header key of "Accept-Version".
+	AcceptVersionHeaderKey = "Accept-Version"
+	// AcceptHeaderKey is the header key of "Accept".
+	AcceptHeaderKey = "Accept"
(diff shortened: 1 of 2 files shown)


pkg/crc/api/helpers.go
```diff
@@ -45,53 +66,81 @@ func (c *context) Code(code int) error {
...
 	return &server{
-		routes: make(map[string]map[string]func(*context) error),
+		routes: make(map[string]map[string]map[apiVersion]func(*context) error),
 	}
 }
 
-func (s *server) GET(pattern string, handler func(c *context) error) {
+func (s *server) GET(pattern string, handler func(c *context) error, version apiVersion) {
+	if !version.isValid() {
+		panic("Invalid version used to register handler")
```
anjannath, line 81: have to remove these panics

gbraad: panics are bad. throw errors, log, but panic is too strong

gbraad:
how is this used? such as, what does the URL look like for a V2 api?

pkg/crc/api/helpers.go
anjannath: to give more context: the reason i went with panic is because this catches it while running the unit tests if a wrong version is used to register the handler

PR #3192, opened 2022-05-30
Add M1 support

cfergeau:
This PR adds podman-only M1 support to crc.
After the switch to vfkit, only minor adjustments are needed (add path to arm64 bundle, reject use of openshift bundle on arm64, ...)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,9 @@ install: $(SOURCES)
 $(BUILD_DIR)/macos-amd64/crc: $(SOURCES)
 	GOARCH=amd64 GOOS=darwin go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/macos-amd64/crc $(GO_EXTRA_BUILDFLAGS) ./cmd/crc
 
+$(BUILD_DIR)/macos-arm64/crc: $(SOURCES)
+	GOARCH=arm64 GOOS=darwin go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/macos-arm64/crc $(GO_EXTRA_BUILDFLAGS) ./cmd/crc
+
 $(BUILD_DIR)/linux-amd64/crc: $(SOURCES)
 	GOOS=linux GOARCH=amd64 go build -ldflags="$(LDFLAGS)" -o $(BUILD_DIR)/linux-amd64/crc $(GO_EXTRA_BUILDFLAGS) ./cmd/crc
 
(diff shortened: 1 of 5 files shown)


pkg/crc/config/settings.go
```diff
@@ -55,8 +55,13 @@ func RegisterSettings(cfg *Config) {
 	}
 
 	// Preset setting should be on top because CPUs/Memory config depend on it.
-	cfg.AddSetting(Preset, string(preset.OpenShift), validatePreset, RequiresDeleteAndSetupMsg,
-		fmt.Sprintf("Virtual machine preset (alpha feature - valid values are: %s or %s)", preset.Podman, preset.OpenShift))
+	if runtime.GOOS == "darwin" && runtime.GOARCH == "arm64" {
+		cfg.AddSetting(Preset, string(preset.Podman), validatePreset, RequiresDeleteAndSetupMsg,
+			fmt.Sprintf("Virtual machine preset (alpha feature - valid values are: %s or %s)", preset.Podman, preset.OpenShift))
```
gbraad, line 60: so this means `OpenShift` is a valid option, as that is reported in the valid values.

praveenkumar: validatePreset actually parse it so we might change it to just having podman, even on validatePreset side we already make aware to the user in same PR.

Makefile
```diff
@@ -319,12 +321,12 @@ packagedir: clean embed-download macos-release-binary packaging/vfkit-$(GOARCH).
 	mkdir -p packaging/root/Applications
 	tar -C packaging/root/Applications -xvzf $(TRAY_RELEASE)
 	rm packaging/tmp/crc-tray-macos.tar.gz
-	mv packaging/root/Applications/crc-tray-darwin-x64/crc-tray.app packaging/root/Applications/Red\ Hat\ OpenShift\ Local.app
-	rm -fr packaging/root/Applications/crc-tray-darwin-x64
+	mv packaging/root/Applications/crc-tray-darwin-$(TRAY_ARCH)/crc-tray.app packaging/root/Applications/Red\ Hat\ OpenShift\ Local.app
+	rm -fr packaging/root/Applications/crc-tray-darwin-$(TRAY_ARCH)
```
anjannath, line 325: currently we are not doing arm builds for the tray releases, maybe we can depend on rosetta to do the right thing here for the initial release, as we are also not using arm builds of `admin-helper`

am not able to build the pkg after this commit, since we don't have the tray arm bits:
```
mv packaging/root/Applications/crc-tray-darwin-arm64/crc-tray.app packaging/root/Applications/Red\ Hat\ OpenShift\ Local.app
mv: rename packaging/root/Applications/crc-tray-darwin-arm64/crc-tray.app to packaging/root/Applications/Red Hat OpenShift Local.app: No such file or directory
make: *** [packagedir] Error 1
```

we need a target for generating arm `pkg` as well `$(BUILD_DIR)/macos-amd64/crc-macos-arm64.pkg`

gbraad: can we easily make arm builds of the tray? 
admin helper should not be that difficult

pkg/crc/config/settings.go
cfergeau: yeah, not sure why I implemented this this way. Not showing the option at all sounds better.

Makefile
cfergeau: I keep forgetting about `admin-helper` because the m1 can run the x86_64 binary, so it was not a blocking issue for m1 testing.

anjannath (dismissed):

anjannath:
Created https://github.com/code-ready/tray-electron/pull/219 for the tray error

PR #3195, opened 2022-05-31
Use segmentio/analytics-go/v3 instead segmentio/analytics-go

praveenkumar:
We didn't able to use v3 because initially because it is fixed late last
year https://github.com/segmentio/analytics-go/pull/178


cfergeau (approved):
Commit log could be changed to 
```
Use segmentio/analytics-go/v3 instead of segmentio/analytics-go

We were not able to use v3 initially, but this was fixed late last
year https://github.com/segmentio/analytics-go/pull/178
```

PR #3201, opened 2022-06-06
build: fix signing arm64 pkg on an amd64 mac

anjannath:
signing can be done on both variants of mac
currently the script fails due to depending
on host arch for forming names of vfkit and
the entitlement file and the finale pkg

instead of depending on host architecure we
write the target arch to a file and provide
it inside the tarball for the script to ref
and construct these file names

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -312,6 +312,7 @@ packaging/vfkit-$(GOARCH).entitlements:
 
 packagedir: clean embed-download macos-release-binary packaging/vfkit-$(GOARCH).entitlements
 	echo -n $(CRC_VERSION) > packaging/VERSION
+	echo -n $(GOARCH) > packaging/ARCH
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/Distribution.in >packaging/darwin/Distribution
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -312,6 +312,7 @@ packaging/vfkit-$(GOARCH).entitlements:
 
 packagedir: clean embed-download macos-release-binary packaging/vfkit-$(GOARCH).entitlements
 	echo -n $(CRC_VERSION) > packaging/VERSION
+	echo -n $(GOARCH) > packaging/ARCH
```
praveenkumar, line 315: 👍🏼

packaging/package.sh
```diff
@@ -73,7 +61,7 @@ productbuild --distribution "${BASEDIR}/darwin/Distribution" \
 rm "${OUTPUT}/crc.pkg"
 
 if [ ! "${NO_CODESIGN}" -eq "1" ]; then
-  productsign --sign "${PRODUCTSIGN_IDENTITY}" "${OUTPUT}/crc-unsigned.pkg" "${OUTPUT}/crc-macos-${GOARCH}.pkg"
+  productsign --sign "${PRODUCTSIGN_IDENTITY}" "${OUTPUT}/crc-unsigned.pkg" "${OUTPUT}/crc-macos-${arch}.pkg"
 else
-  mv "${OUTPUT}/crc-unsigned.pkg" "${OUTPUT}/crc-macos-${GOARCH}.pkg"
+  mv "${OUTPUT}/crc-unsigned.pkg" "${OUTPUT}/crc-macos-${arch}.pkg"
```
praveenkumar, line 66: 👍🏼

praveenkumar (approved):
Since we have to build the M1 bits on Intel platform so we have store the arch variable and use it for signing also.

PR #3212, opened 2022-06-07
Update go version and remove centos CI script

praveenkumar:


praveenkumar:
> Just to be sure, removing this was intentional?

Centos infra is updated to use openshift-4 and we were using the old infra which is not EOL from centos side.

PR #3216, opened 2022-06-13
SSH key replacement improvements

cfergeau:
This removes a no longer needed workaround for MCO bug when we change ~/.ssh/authorized_keys, and also removes the default ssh key when using the podman bundle.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -626,9 +626,7 @@ func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
 	}
 
 	logging.Info("Updating authorized keys...")
-	// CopyData uses sudo and we need to use it
-	// because of https://bugzilla.redhat.com/show_bug.cgi?id=1956739
-	err = sshRunner.CopyDataPrivileged(publicKey, "/home/core/.ssh/authorized_keys", 0644)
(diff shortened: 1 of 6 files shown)


pkg/crc/machine/start.go
```diff
@@ -811,7 +811,7 @@ func updateCockpitConsoleBearerToken(sshRunner *crcssh.Runner) error {
 		return fmt.Errorf("failed to write cockpit bearer token: %w", err)
 	}
 
-	if err := sshRunner.CopyData([]byte(token), "/home/core/cockpit-bearer-token", 0600); err != nil {
+	if err := sshRunner.CopyDataPrivileged([]byte(token), "/home/core/cockpit-bearer-token", 0600); err != nil {
```
anjannath, line 815: This can be changed to use `CopyData`, in the next lines we are again changing the ownership of this file to `core`

https://github.com/cfergeau/crc/blob/728f60ae630ef6a72eaa98a9d61c00e8f5d314af/pkg/crc/machine/start.go#L815-L822

anjannath (approved):

pkg/crc/machine/start.go
cfergeau: Oh nice, happy that there's a second user for CopyData/CopyDataPrivileged!

PR #3218, opened 2022-06-14
pkg: build single pkg installer using fat binaries

anjannath:
instead of building two pkg installers for arm and x86
this create one universal pkg installer using macos universal
binaries for tray, admin-helper and vfkit

we expect admin-helper, tray and vfkit universal binaries are
avaiable to download on their respective github releases

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,6 @@ DOCS_BUILD_TARGET ?= /docs/source/getting_started/master.adoc
 
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-TRAY_ARCH = $(subst amd64,x64,$(GOARCH))
 
 HOST_BUILD_DIR=$(BUILD_DIR)/$(GOOS)-$(GOARCH)
 GOPATH ?= $(shell go env GOPATH)
@@ -299,7 +298,7 @@ update-go-version:
 goversioncheck:
 	./verify-go-version.sh
 
-TRAY_RELEASE ?= packaging/tmp/crc-tray-macos-$(GOARCH).tar.gz
+TRAY_RELEASE ?= packaging/tmp/crc-tray-macos.tar.gz
 
 embed-download: $(HOST_BUILD_DIR)/crc-embedder
 ifeq ($(CUSTOM_EMBED),false)
@@ -307,12 +306,15 @@ ifeq ($(CUSTOM_EMBED),false)
 	$(HOST_BUILD_DIR)/crc-embedder download $(EMBED_DOWNLOAD_DIR)
 endif
 
-packaging/vfkit-$(GOARCH).entitlements:
+packaging/vfkit.entitlements:
 	curl -sL https://raw.githubusercontent.com/code-ready/vfkit/main/vf.entitlements -o $@
 
-packagedir: clean embed-download macos-release-binary packaging/vfkit-$(GOARCH).entitlements
+macos-universal-binary: macos-release-binary
+	mkdir -p out/macos-universal
+	lipo -create out/macos-amd64/crc out/macos-arm64/crc -output out/macos-universal/crc
+
+packagedir: clean embed-download macos-universal-binary packaging/vfkit.entitlements
 	echo -n $(CRC_VERSION) > packaging/VERSION
-	echo -n $(GOARCH) > packaging/ARCH
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/Distribution.in >packaging/darwin/Distribution
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/welcome.html.in >packaging/darwin/Resources/welcome.html
 	sed -e 's/__VERSION__/'$(CRC_VERSION)'/g' -e 's@__INSTALL_PATH__@$(MACOS_INSTALL_PATH)@g' packaging/darwin/postinstall.in >packaging/darwin/scripts/postinstall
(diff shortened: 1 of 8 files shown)


anjannath:
Test is failing because universal helper binaries are not available to download, yet

Makefile
```diff
@@ -299,20 +298,23 @@ update-go-version:
...
 	mkdir -p $(EMBED_DOWNLOAD_DIR)
 	$(HOST_BUILD_DIR)/crc-embedder download $(EMBED_DOWNLOAD_DIR)
 endif
 
-packaging/vfkit-$(GOARCH).entitlements:
+packaging/vfkit-universal.entitlements:
 	curl -sL https://raw.githubusercontent.com/code-ready/vfkit/main/vf.entitlements -o $@
 
-packagedir: clean embed-download macos-release-binary packaging/vfkit-$(GOARCH).entitlements
+prepare_crc_fat_binary: macos-release-binary
```
cfergeau, line 312: `macos-universal-binary` sounds nicer/more neutral

Makefile
```diff
@@ -299,20 +298,23 @@ update-go-version:
...
+	mkdir -p out/macos-universal
+	lipo -create out/macos-amd64/crc out/macos-arm64/crc -output out/macos-universal/crc
```
cfergeau, line 314: `lipo` is a macos binary. I think there's a reimplementation of it in go, dunno if you tried it? I'm wondering whether we should be generating this universal binary as part of the signing process or during the cross build on linux?

cfergeau: This can be adjusted later though

Makefile
```diff
@@ -322,20 +324,20 @@ packagedir: clean embed-download macos-release-binary packaging/vfkit-$(GOARCH).
 	mkdir -p packaging/root/Applications
 	tar -C packaging/root/Applications -xvzf $(TRAY_RELEASE)
 	rm $(TRAY_RELEASE)
-	mv packaging/root/Applications/crc-tray-darwin-$(TRAY_ARCH)/crc-tray.app packaging/root/Applications/Red\ Hat\ OpenShift\ Local.app
-	rm -fr packaging/root/Applications/crc-tray-darwin-$(TRAY_ARCH)
+	mv packaging/root/Applications/crc-tray-darwin-universal/crc-tray.app packaging/root/Applications/Red\ Hat\ OpenShift\ Local.app
+	rm -fr packaging/root/Applications/crc-tray-darwin-universal
```
cfergeau, line 328: This needs universal tray binaries, and universal vfkit binaries, I don't know if these are ready yet?

pkg/crc/machine/vfkit/constants.go
```diff
@@ -5,14 +5,13 @@ package vfkit
 
 import (
 	"fmt"
-	"runtime"
 )
 
 const (
 	VfkitVersion = "0.0.1"
+	VfkitCommand = "vfkit-universal"
```
cfergeau, line 12: We added `-${arch}` suffixes to our macos binaries as otherwise there would have been a name conflict. Isn't this a good opportunity to rename them back to just `vfkit` and `crc-admin-helper-darwin`? We don't add the arch on the other aplatforms.

packaging/package.sh
```diff
@@ -30,8 +30,14 @@ function signAppBundle() {
   fi
 
   frameworks=$(find "$1"/Contents/Frameworks -depth -type d -name "*.framework" -or -name "*.dylib" -or -type f -perm +111)
-  echo "${frameworks}" | xargs -t -I % codesign --deep --sign "${CODESIGN_IDENTITY}" --options runtime % || true
-  codesign --deep --sign "${CODESIGN_IDENTITY}" --options runtime --force --entitlements "${entitlements}" "$1"
+  echo "${frameworks}" | xargs -t -I % codesign --deep --sign "${CODESIGN_IDENTITY}" --options runtime --timestamp % || true
```
cfergeau, line 33: This adds `--timestamp`, this should be part of the previous commit.

packaging/package.sh
```diff
@@ -30,8 +30,14 @@ function signAppBundle() {
...
+
+  # sign the .app bundles inside $1/Contents/Frameworks
+  frameworks=$(find "$1"/Contents/Frameworks -depth -type d -name "*.app" -perm +111)
```
cfergeau, line 36: We have `-name "*.framework" -or -name "*.dylib"` in the first `frameworks` assignment, I don't know if the `--force` is required here, but maybe we could change the first line to `-name "*.framework" -or -name "*.dylib" -or -name "*.app"` ? More details in the log of this commit explaining why this is needed / what issue it's causing if we don't do that would be useful.

cfergeau (commented):
Overall looks good, a bunch of question but nothing fundamental.

cfergeau:
Would be useful to document the size impact btw

Makefile
anjannath: They are not released in the github releases, for tray we have a PR which generates universal artifact, for admin-helper and vfkit have to manually build and upload the binaries to github releases

packaging/package.sh
anjannath: Fixed

pkg/crc/machine/vfkit/constants.go
anjannath: yes, makes sense to do this instead of adding `-universal` for these

packaging/package.sh
anjannath: We have to sign inside out, so first the `*.framework` and `*.dylib` inside of the `*.app` and then at the end sign the outermost thing, that is the `.app` which is the reason they are separately done and also we don't want to rewrite signatures on the `*.framework` and `*.dylib` files

anjannath:
@cfergeau addressed most of the comments, please take a look again!

anjannath:
One thing to note is that, users upgrading, will have to do a `crc delete` since the path to the `vfkit` is part of the machine config at `~/.crc/machines/crc/config.json` and we change the name from `vfkit-<arch>` to just `vfkit`

```json
# snip from config.json
"InitrdPath": "/Users/anjan/.crc/cache/crc_podman_vfkit_4.0.2_arm64/initramfs-5.17.5-300.fc36.aarch64.img",
    "VfkitPath": "/Applications/Red Hat OpenShift Local.app/Contents/Resources/vfkit-arm64",
    "VirtioNet": false,
```

Makefile
praveenkumar: @anjannath this commit still using the `universal`, shouldn't this removed?

anjannath: This should be fine, its ony the dir that is named `out/macos-universal` and the tray directory is also after extracting the tray archive, the released executables/archives don't contain the `-universal`

PR #3219, opened 2022-06-14
e2e: fix httpd test case

cfergeau:
quay.io/bitnami/nginx is no longer available, which breaks the test case. 
The changes are more involved than just changing the image name, as 
registry.access.redhat.com/ubi8/httpd-24 needs html data to be provided in
/var/www/html. This is done using a pvc (PersistentVolumeClaim) and using
it with our httpd-example deployment. Then `oc exec` is used to create an
empty `index.html` file at that location.

### DIFF
--- a/test/e2e/features/story_health.feature
+++ b/test/e2e/features/story_health.feature
@@ -41,13 +41,13 @@ Feature: End-to-end health check
             """
         And stdout should contain "You can add applications to this project with the 'new-app' command."
 
-    @darwin @linux @windows
+    @darwin @linux @windows @testdata
     Scenario: Create and test app
         When executing "oc create deployment httpd-example --image=registry.access.redhat.com/ubi8/httpd-24 --port=8080" succeeds
         Then stdout should contain "deployment.apps/httpd-example created"
         When executing "oc rollout status deployment httpd-example" succeeds
         Then stdout should contain "successfully rolled out"
-        When executing "oc create configmap www-content --from-literal=index.html=" succeeds
+        When executing "oc create configmap www-content --from-file=index.html=httpd-example-index.html" succeeds
         Then stdout should contain "configmap/www-content created"
         When executing "oc set volume deployment/httpd-example --add --type configmap --configmap-name www-content --name www --mount-path /var/www/html" succeeds
         Then stdout should contain "deployment.apps/httpd-example volume updated"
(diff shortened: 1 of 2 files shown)


test/testdata/httpd-deployment.yaml
```diff
@@ -1,14 +1,10 @@
...
-  name: www-claim
-spec:
-  accessModes:
-    - ReadWriteOnce
-  resources:
-    requests:
-      storage: 1k
+  name: www-content
+data:
+  index.html:
```
praveenkumar, line 7: We can have this like following in case checked using `curl <route_url>` and validating the correct content. 
```
data:
  index.html: "<html>Hello CRC</html>"
```

test/e2e/features/story_health.feature
```diff
@@ -41,16 +41,13 @@ Feature: End-to-end health check
...
-        When executing "oc create deployment httpd-example --image=quay.io/bitnami/nginx --port=8080" succeeds
+        When executing "oc create -f httpd-deployment.yaml" succeeds
         Then stdout should contain "deployment.apps/httpd-example created"
         When executing "oc rollout status deployment httpd-example" succeeds
         Then stdout should contain "successfully rolled out"
-        When executing "oc expose deployment httpd-example --port 8080" succeeds
-        Then stdout should contain "httpd-example exposed"
-        When executing "oc expose svc httpd-example" succeeds
-        Then stdout should contain "httpd-example exposed"
+        And with up to "2" retries with wait period of "60s" http response from "http://httpd-example-testproj.apps-crc.testing" has status code "200"
```
gbraad, line 50: See: https://github.com/code-ready/crc/pull/3219/files#r897518410. Can content be checked? How easy or difficult would this be?

jsliacan: I think something as simple as this would work: 

```
When executing "curl http://httpd-example-testproj.apps-crc.testing" succeeds
Then stdout should contain "Hello CRC"
```
The string is reasonably unique, so that should be OK. Not sure how much value this additional check brings, but it's basically for free so it doesn't matter I guess.

cfergeau: > I think something as simple as this would work:
```
When executing "curl http://httpd-example-testproj.apps-crc.testing" succeeds
Then stdout should contain "Hello CRC"
```

I suspect the retry is needed in case the pod is not fully up yet when we start trying curl, in which case it's not going to be as simple.

jsliacan: In that case the following might work out of the box
```
Then with up to "20" retries with wait period of "30s" command "curl http://httpd-example-testproj.apps-crc.testing" output matches ".*Hello CRC.*"
```
Not saying it needs to go in, just that some templates exist for these things. But there could be other issues that I'm missing and why it won't work.

cfergeau: Tried what you suggested, and a few variations around it, but for some reason, `Hello World` on stdout is not captured correctly even if it shows up as expected in the logs.

jsliacan: Hmmm. Good investigation, I don't think I've seen this before... Thanks

cfergeau:
The e2e test deploying httpd succeeds, the e2e failure comes from
```
     Then login to the oc cluster succeeds                                          # crcsuite.go:353 -> github.com/code-ready/crc/test/e2e/crcsuite.LoginToOcClusterSucceedsOrFails
    command 'oc config use-context crc-admin', expected to succeed, exited with exit code: 1
Command stdout: 
Command stderr: error: no context exists with the name: "crc-admin"
```
which should be unrelated to this PR

praveenkumar (approved):

Issue #3217, opened 2022-06-14
[Task] Add logic to pull bundles from quay.io


praveenkumar:
This is first task from the epic #3206 where we want to pull the bundles from quay.io instead of mirror.openshift.com

anjannath:
Fixed in #3221

PR #3221, opened 2022-06-16
Logic to pull bundle from quay.io

praveenkumar:
This is initial pr to switch bundle download from mirror to quay.io
End user shouldn't affect with this change except there is no download
resume functionality.

The bundle images are pushed to quay.io/crcont with following name
- ocp-\<hypervisor>-bundle
- podman-\<hypervisor>-bundle

This PR doesn't remove existing codebase to download bundles from mirror
as of now but it should be done in a separate commit so we can easily
revert back.

```
$ crc setup
[...]
INFO Downloading crc_libvirt_4.10.14_amd64.crcbundle
Copying blob 7e0dbb47dc47 done
Copying config 4f32695e8f done
Storing signatures
crc_libvirt_4.10.14_amd64.crcbundle: 3.15 GiB / 3.15 GiB [---------------------------------------------------------------------------------------------------------------------------] 100.00%
INFO Uncompressing /home/prkumar/.crc/cache/crc_libvirt_4.10.14_amd64.crcbundle
crc.qcow2: 12.61 GiB / 12.61 GiB [---------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
oc: 117.14 MiB / 117.14 MiB [--------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
```

**Fixes:** #3217


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -144,3 +148,30 @@ func removeOldLogs() error {
 	}
 	return nil
 }
+
+func pullBundleImage(preset preset.Preset) error {
+	imgHandler := image.OCIImageHandler{
+		ImageName: image.GetImageName(preset),
+		ImageTag:  version.GetBundleVersion(),
```
anjannath, line 155: we can modify `GetBundleVersion` to also take a preset and return the podman version in case of podman preset
```
DEBU Accessing "quay.io/crcont/podman-vfkit-bundle:4.10.14" failed: reading manifest 4.10.14 in quay.io/crcont/podman-vfkit-bundle: manifest unknown: manifest unknown
initializing source docker://quay.io/crcont/podman-vfkit-bundle:4.10.14: reading manifest 4.10.14 in quay.io/crcont/podman-vfkit-bundle: manifest unknown: manifest unknown
```

cfergeau:
One initial comment, this makes the binaries 10MB/20% bigger

praveenkumar:
> One initial comment, this makes the binaries 10MB/20% bigger

Yes, because of c/image dependency, it is making the binary bit more bigger.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/containers/image/v5/transports/alltransports"
+	"github.com/containers/image/v5/types"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+var (
+	registryURI = "quay.io"
+	orgName     = "crcont"
+)
+
```
anjannath, line 23: To make this easily work with another registry, we could include the `registryURI` and `orgName` also inside the `OCIImageHandler` struct and create a constructor like function:
```go
func NewOCIImageHandler(orgName, registryURI) *OCIImageHandler {
    return &OCIImageHandler{orgName: orgName, registryURI: registryURI}
}

``` 

but probably we'll only need to use quay for our usecase

praveenkumar: Yes I thought about it but initially I am just going with quay. These improvements we can have after first iteration.

pkg/crc/preflight/preflight_checks_common.go
cfergeau: Pushing both bundle version and crc version as tags for the same image is quite confusing, there is no hint why we have `podman-vfkit-bundle:2.5.0` being newer than `podman-vfkit-bundle:4.0.2`. Might be an indication that crc version -> bundle version should not be handled as tags on the image.
Also, what happens the day we release `crc` version `4.0.2`?

pkg/crc/image/image.go
```diff
@@ -0,0 +1,113 @@
...
+type OCIImageHandler struct {
+	ImageName string
+	ImageTag  string
```
cfergeau, line 26: Not clear to me why we hardcode `quay.io` and `crcont` but we let external code specify which `image`/`tag` should be downloaded, especially with `GetImageName` implemented in this file.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,113 @@
...
+func (img *OCIImageHandler) uri() string {
+	return strings.Join([]string{registryURI, orgName, img.ImageName}, "/")
+}
+
+func (img *OCIImageHandler) parse(image string) (types.ImageReference, error) {
+	return alltransports.ParseImageName(image)
+}
+
+func (img *OCIImageHandler) policyContext() (*signature.PolicyContext, error) {
+	policy := &signature.Policy{Default: []signature.PolicyRequirement{signature.NewPRInsecureAcceptAnything()}}
```
cfergeau, line 39: `signature.NewPRInsecureAcceptAnything()` hopefully this is less bad than the name implies, or we can do without this?

pkg/crc/image/image.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, fmt.Errorf("error creating security context: %s", err.Error())
+	}
+
+	return policyContext, nil
+}
+
+func (img *OCIImageHandler) getSystemContext() *types.SystemContext {
+	return &types.SystemContext{
+		OSChoice:           "linux",
+		ArchitectureChoice: runtime.GOOS,
```
cfergeau: not `GOARCH`?

pkg/crc/image/image.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating security context: %s", err.Error())
+	}
+
+	return policyContext, nil
+}
+
+func (img *OCIImageHandler) getSystemContext() *types.SystemContext {
+	return &types.SystemContext{
+		OSChoice:           "linux",
```
cfergeau: Can we make use of `OSChoice` to have a single `bundle-openshift` image for `macos`/`linux`/`windows` OS instead of `bundle-vfkit-openshift`/`bundle-libvirt-openshift`/`bundle-hyperv-openshift`?

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+	policyContext, err := img.policyContext()
+	if err != nil {
+		return nil, err
+	}
+
+	return copy.Image(context.Background(), policyContext,
+		destRef, srcRef, &copy.Options{
+			ReportWriter: img.Progress,
+			SourceCtx:    img.getSystemContext(),
+		})
```
cfergeau, line 78: I would pass `Progress` as an argument to this method rather than having it in `OCIImageHandler` where it feels out of place. Also, `Progress` is slightly misleading, `ReportWriter` is more accurate.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (img *OCIImageHandler) getSystemContext() *types.SystemContext {
+	return &types.SystemContext{
+		OSChoice: "linux",
+	}
+}
+
+// CopyImage pull the image from the registry and put it to local storage
+func (img *OCIImageHandler) CopyImage(destPath string) ([]byte, error) {
```
cfergeau, line 55: This should be returning `(*v1.Manifest, error)` rather than raw bytes.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+	return policyContext, nil
+}
+
+func (img *OCIImageHandler) getSystemContext() *types.SystemContext {
+	return &types.SystemContext{
+		OSChoice: "linux",
+	}
+}
+
+// CopyImage pull the image from the registry and put it to local storage
```
cfergeau, line 54: `local storage` means `destPath`? It could be misread as `podman local image s torage`, so I'd mention destpath in the doc.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return copy.Image(context.Background(), policyContext,
+		destRef, srcRef, &copy.Options{
+			ReportWriter: img.Progress,
+			SourceCtx:    img.getSystemContext(),
+		})
+}
+
+func GetBundleImageLayerFromManifest(manifestData []byte) (string, error) {
```
cfergeau, line 81: If you return a `v1.Manifest` from `CopyImage`, then you don't need the very long name, just `GetBundleImageLayer` works, or `GetLayerPath` or so

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+	manifest := &v1.Manifest{}
+	// Get the file layer sha from manifest
+	if err := json.Unmarshal(manifestData, manifest); err != nil {
+		return "", err
+	}
+	return removeShaFromDigestString(manifest.Layers[0].Digest.String()), nil
```
cfergeau, line 87: not sure this deserves a separate helper since it's going to be a one-liner used once, and this method is also not very long.
I tried this code with multi-layer images, and it did not work at all (expected, and not saying it has to be fixed). I was wondering if there's a way of detecting this though, maybe disallow multi-layer images, and also check the `MediaType` to make sure it's the expected one?

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+	return strings.TrimPrefix(digestString, "sha256:")
+}
+
+func GetImageName(preset crcpreset.Preset) string {
+	img := imageNameForOs(preset)
+	return img[runtime.GOOS]
+}
+
+func imageNameForOs(preset crcpreset.Preset) map[string]string {
+	if preset == crcpreset.Podman {
```
cfergeau, line 100: Dunno if this method should be here or in `pkg/crc/constants/constants.go` with other similar code?

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -144,3 +148,30 @@ func removeOldLogs() error {
...
+	bundleLayerPath, err := image.GetBundleImageLayerFromManifest(imgDigest)
+	if err != nil {
+		return err
+	}
+	logging.Info("Extracting the image bundle layer...")
+	_, err = extract.Uncompress(filepath.Join(destDir, bundleLayerPath), constants.MachineCacheDir, true)
+	if err != nil {
+		return err
+	}
+	return nil
```
cfergeau, line 176: Wondering if this deserves to be a helper in the `image` package as this is quite a few low-level steps that needs to be done, not sure preflight should have that much knowledge about how a container image is stored.

cfergeau (commented):
I'm find the log confusing

> This is initial pr to switch bundle download from mirror to quay.io
End user shouldn't affect with this change except there is no download
resume functionality.

After reading this, I don't know if bundle download is fully moved to getting it from quay.io or not ("initial PR", "end user shouldn't affect with this change"). I would say explicitly that after this commit, bundles are downloaded from quay.io even if the old download code is still there.

Fwiw, I find the name `pkg/crc/image` very (too) generic, is this about picture, an ISO image, a qcow2 image, ... ? Might be worth naming this `container-image`, or maybe just `container` for the package name since the type used is then `OCIImage..` ?

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/containers/image/v5/types"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+var (
+	registryURI = "quay.io"
+	orgName     = "crcont"
+)
+
+type OCIImageHandler struct {
```
cfergeau, line 24: Not sure `OCI` is really useful here, we hardcode `docker:` for fetching from the remote repository, and use `dir:` for local storage. Maybe this is still stored in an OCI compliant way, but this really is not the most important part of the implementation, especially as we don't support many layer formats.

Makefile
```diff
@@ -68,6 +68,8 @@ endif
 
 # https://golang.org/cmd/link/
 LDFLAGS := $(VERSION_VARIABLES) ${GO_EXTRA_LDFLAGS}
+# Same build flags are used in the podman remote to cross build it https://github.com/containers/podman/blob/main/Makefile
+BUILDTAGS := remote exclude_graphdriver_btrfs btrfs_noversion exclude_graphdriver_devicemapper containers_image_openpgp
```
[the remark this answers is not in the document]
praveenkumar: Nothing much I can do :)

pkg/crc/image/image.go
praveenkumar:
```
	prTypeInsecureAcceptAnything prTypeIdentifier = "insecureAcceptAnything"
	prTypeReject                 prTypeIdentifier = "reject"
	prTypeSignedBy               prTypeIdentifier = "signedBy"
	prTypeSignedBaseLayer        prTypeIdentifier = "signedBaseLayer"
```

is what all we can have here as of now our images are not signed one so using insecureAcceptAnything, as soon as we start securing we will change the policy.

pkg/crc/image/image.go
praveenkumar: Arch is taken as default of running host so we don't need to explicit mention it. `OSChoice` we need because all the images are linux based.

pkg/crc/image/image.go
praveenkumar: @cfergeau I thing the OS choice is what base image should be, need to see if we can use scratch for all the different platform.

pkg/crc/image/image.go
praveenkumar: I don't see why if it feel out of place, we create imagehandler and run this method. I will change the Progress to `ReportWriter` if that make it more accurate. I used progress because it show the progress of image layer download/pull.

pkg/crc/image/image.go
praveenkumar: @cfergeau you mean `podman load <image.tar>` ?

praveenkumar: Also I will put destpath.

pkg/crc/image/image.go
praveenkumar: As long as it is easy to read or understand I don't mind to have a very long function. if we return `v1.Manifest` from image we need to process the byte there instead here.

pkg/crc/image/image.go
praveenkumar: > I tried this code with multi-layer images, and it did not work at all (expected, and not saying it has to be fixed). I was wondering if there's a way of detecting this though, maybe disallow multi-layer images, and also check the `MediaType` to make sure it's the expected one?

I am not sure why it shouldn't work with multi layer, the logic here is just return the first layer digest, in your test does the first layer didn't have any digest?

pkg/crc/image/image.go
praveenkumar: I prefer it here but don't have any strong opinion to move (if you really think)

pkg/crc/image/image.go
praveenkumar: I think it is good to be explicit around what kind of image format we are storing to dir and right now `quay.io` pulled images are in oci image format so better to have it.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,112 @@
...
+	ImageTag  string
+	Progress  io.Writer
+}
+
+func (img *OCIImageHandler) uri() string {
+	return strings.Join([]string{registryURI, orgName, img.ImageName}, "/")
+}
+
+func (img *OCIImageHandler) parse(image string) (types.ImageReference, error) {
+	return alltransports.ParseImageName(image)
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I find `alltransports.ParseImageName` much better to not to worry about which transport is in use as long it is parse correctly but if you insist I don't have issue to make transport specific parse.

pkg/crc/image/image.go
praveenkumar: Didn't get what you mean about external code as of now `GetImageName` only provide the image name for preset and then we define this struct and use the method to copy it.

pkg/crc/image/image.go
cfergeau: But this is `ArchitectureChoice` that you are setting, not `OSChoice`, `ArchitectureChoice: runtime.GOARCH` makes sense to me, so does `OSChoice: runtime.GOOS`. Here we have `ArchitectureChoice: runtime.GOOS` which looks unexpected?

pkg/crc/image/image.go
cfergeau: The []byte is useless in itself and will need to be converted anyway, and it makes it a lot more clear what the function returns (I initially though this was returning the layer content, not the manifest), ...

pkg/crc/image/image.go
cfergeau: It will do "something" for multilayer images, but will it be doing the right thing, or something unexpected to the user? Imo it's not very well defined what will happen, so it can just error out so that we catch immediatly that something was off.

pkg/crc/image/image.go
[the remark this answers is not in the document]
cfergeau: If we could use arbitrary transports, alltransports.ParseImageName would be better, but here the transports are hardcoded in image.go, so it does not add much apart from a few MB to the binary size, and more complexity.

pkg/crc/image/image.go
cfergeau: Ah indeed, does not look like there's a better choice. Hopefully `AcceptAnything` will reject images with bad signatures (which is what I was worried about, that it will silently ignore invalid sigs). We'll see!

pkg/crc/image/image.go
cfergeau: `preflight_checks_common.go` has code to fill ImageName/ImageTag:

```
imgHandler := image.OCIImageHandler{
		ImageName: image.GetImageName(preset),
		ImageTag:  version.GetCRCVersion(),
		Progress:  os.Stdout,
	}
```
but it calls into `image` to fill this.  Then `image.go` hardcodes `quay.io/crcont`.  Half of what is needed to fetch the image is set in `preflight_checks_common.go` and the other half in `image.go`. `preflight_checks_common.go` does not need to set this imo, ImageName/ImageTag could both be set internally in `image.go`

pkg/crc/image/image.go
praveenkumar: @cfergeau yes I remove that `ArchitectureChoice: runtime.GOOS` also I can't set `OSChoice: runtime.GOOS` because our images are only linux based not `darwin` or `window` based which cause issue during the image pull.
```
choosing an image from manifest list docker://quay.io/crcont/podman-vfkit-bundle:2.4.1: no image found in image index for architecture arm64, variant "", OS darwin
```

pkg/crc/image/image.go
praveenkumar: > Progress is only useful to the Copy method, so it's not clear why we need to store it in the OCIImage type which can be used over other methods? Isn't it clearer to pass it to the only method which needs it?

@cfergeau make sense, will update this.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/containers/image/v5/copy"
+	"github.com/containers/image/v5/directory"
+	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/signature"
+	"github.com/containers/image/v5/types"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const (
+	registryURI = "quay.io"
```
gbraad, line 29: can these be defined in a `constants.go` file instead?

pkg/crc/image/image.go
```diff
@@ -0,0 +1,129 @@
...
+	return manifest, nil
+}
+
+func getLayerPath(m *v1.Manifest) string {
+	return strings.TrimPrefix(m.Layers[0].Digest.String(), "sha256:")
+}
+
+func getImageName(preset crcpreset.Preset) string {
+	if preset == crcpreset.Podman {
+		return "podman-bundle"
```
gbraad, line 103: Can these be extracted as part of a constants file, like named `podmanBundleName` and `openshiftBundleName`. Pretty sure these will be somehow reused of perhaps even overwritten at some point.

pkg/crc/image/image.go
praveenkumar: @gbraad yes we can move it to constants.go, As of now it is here just to have all image related stuff in same package.

pkg/crc/image/image.go
praveenkumar: These are images name not the bundle so not sure if that is reused anywhere except in image package.

cfergeau (commented):
Still some comments/questions, but looking good overall

pkg/crc/image/image.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	return "openshift-bundle"
+}
+
+func PullBundleImage(preset crcpreset.Preset, reportWriter io.Writer) error {
+	imgHandler := imageHandler{
+		imageName: getImageName(preset),
+		imageTag:  version.GetCRCVersion(),
+	}
+	destDir, err := ioutil.TempDir(constants.MachineCacheDir, "")
```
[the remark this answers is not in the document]
praveenkumar: yes even ioutil.TempDir just call `os.MkdirTemp` :)

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -98,7 +99,7 @@ func fixBundleExtracted(bundlePath string, preset preset.Preset) func() error {
 		}
 		if bundlePath == constants.GetDefaultBundlePath(preset) {
 			logging.Infof("Downloading %s", constants.GetDefaultBundle(preset))
-			if err := bundle.Download(preset); err != nil {
+			if err := image.PullBundleImage(preset, os.Stdout); err != nil {
```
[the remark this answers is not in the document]
praveenkumar: > This could be checking the bundle sha256sum for now.

After this initial PR next would be verify using signed bits so it will be taken care of.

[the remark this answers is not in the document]
cfergeau: But we don't want to wait for this 'signature' PR before merging this one, do we? Given our fast release cycles, I'm a bit uncomfortable merging a PR which depends on a non-existent PR which has to be in before our next release.

pkg/crc/image/image.go
praveenkumar: Only for 2-3 constants which also used for same package I don't feel create a different constants file.

pkg/crc/gpg/gpg.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return signature, nil
+}
+
+func getPubicKey() (*packet.PublicKey, error) {
+	pubKeyBlock, err := armor.Decode(bytes.NewReader([]byte(constants.GPGPublicKey)))
+	if err != nil {
+		return nil, fmt.Errorf("error decoding public key: %s", err)
+	}
+	if pubKeyBlock.Type != "PGP PUBLIC KEY BLOCK" {
```
gbraad, line 50: there is no constant defined in the library for this?

praveenkumar: @gbraad No it is not I did check that first :)

gbraad: very weird. very prone to error

pkg/crc/constants/constants.go
```diff
@@ -41,6 +41,21 @@ const (
...
+QGNyYy5kZXY+iJkEExYKAEEWIQS4RlW/rByOBn/ZyofcfqxAChv9+wUCYrvgDQIb
+AwUJEswDAAULCQgHAgIiAgYVCgkICwIEFgIDAQIeBwIXgAAKCRDcfqxAChv9+/ep
+APwISi03R7npwimqdL7NYKDGMO8ikOwmmPkqh9CKwt4CdwD8Cc6HNcZumHDpJ4gH
+x7FXxIS9KLwDihpm1Gxr4t1t5Qy4OARiu+ANEgorBgEEAZdVAQUBAQdA/w7pM7hf
+bxZ2qwSuoBuhcA1sAlPSb3NrIZf3CceoqzQDAQgHiH4EGBYKACYWIQS4RlW/rByO
+Bn/ZyofcfqxAChv9+wUCYrvgDQIbDAUJEswDAAAKCRDcfqxAChv9+2UkAQCNCdaf
+vnhbvfPHDltmwDZ3aD4l3jjSKpeySeKQocgjQAD6A7kawst/50k4wb+vUDUnEoYo
+9Ix7lKfKWCXil/z0vg4=
+=lmb/
+-----END PGP PUBLIC KEY BLOCK-----`
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I already created this for `crc` purpose so we don't need to update.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/containers/image/v5/signature"
+	"github.com/containers/image/v5/types"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type imageHandler struct {
+	imageURI string
+}
+
+func uri(preset crcpreset.Preset, registryURI string) string {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau external arg is there so that it can take custom registry

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -98,7 +99,7 @@ func fixBundleExtracted(bundlePath string, preset preset.Preset) func() error {
 		}
 		if bundlePath == constants.GetDefaultBundlePath(preset) {
 			logging.Infof("Downloading %s", constants.GetDefaultBundle(preset))
-			if err := bundle.Download(preset); err != nil {
+			if err := image.PullBundle(preset, constants.RegistryURI); err != nil {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau it is expose here because in next commit we are sending user specified registry url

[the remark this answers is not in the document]
cfergeau: Well, I'm inclined to nack the next commit for now, so I'd move all of this to the next commit rather than making this one needlessly complex.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,134 @@
...
+	r := strings.SplitN(registryURI, "/", 2)
+	registry, orgName := r[0], r[1]
+	uri := url.URL{
+		Host: registry,
+		Path: orgName + "/" + getImageName(preset) + ":" + version.GetCRCVersion(),
+	}
+	return uri.String()
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau `fmt.Sprintf("%s/%s:%s", constants.RegistryURI, getImageName(preset), version.GetCRCVersion()` will not provide the uri and `strings.SplitN` can't have more that 2 elements because we put limit to 2 here.

pkg/crc/config/settings.go
```diff
@@ -103,6 +104,8 @@ func RegisterSettings(cfg *Config) {
 
 	cfg.AddSetting(KubeAdminPassword, "", ValidateString, SuccessfullyApplied,
 		"User defined kubeadmin password")
+	cfg.AddSetting(RegistryURI, constants.RegistryURI, ValidateString, SuccessfullyApplied,
+		fmt.Sprintf("Registry URI which contain bundle image (default: %s)", constants.RegistryURI))
```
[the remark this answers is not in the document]
praveenkumar: So you are saying to have `--bundle` flags works for local bundle file and also for registry? As of now `--bundle` flag is only support the bundle path locally available and we should keep that support also in case someone doesn't have local registry so they can get the bundle from image tar (using skopeo) and then use it with this option.

pkg/crc/image/image.go
```diff
@@ -126,9 +127,26 @@ func PullBundle(preset crcpreset.Preset, registryURI string) error {
 	if err != nil {
 		return err
 	}
-	_, err = extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
+	bundleExtractPath, err := extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
 	if err != nil {
 		return err
 	}
+
+	logging.Info("Extracting bundle gpg signature...")
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I thought it will take more time (~30sec) so better to let user know what's happening.

pkg/crc/image/image.go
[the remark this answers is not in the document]
cfergeau: but what is the point of splitting it since `uri.String()` will anyway concatenate it again? The interesting part of the URI construction is `orgName + "/" + getImageName(preset) + ":" + version.GetCRCVersion()` which we are already doing by hand. So why not just directly build the URI we want instead of hiding this behind `SplitN` followed by `String()` ?

pkg/crc/image/image.go
```diff
@@ -126,9 +127,26 @@ func PullBundle(preset crcpreset.Preset, registryURI string) error {
 	if err != nil {
 		return err
 	}
-	_, err = extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
+	bundleExtractPath, err := extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
```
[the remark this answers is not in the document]
cfergeau: We use `fileList` in other parts of the code I think

pkg/crc/image/image.go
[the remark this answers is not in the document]
cfergeau: Yes, but a single message about signature verification should be enough? They don't need to know that first we extract the signature, and then we verify it?

pkg/crc/gpg/gpg.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+func getSignature(signatureFilePath string) (*packet.Signature, error) {
+	sigFile, err := os.Open(signatureFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err := sigFile.Close(); err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
praveenkumar: Ah I was testing that so put panic but yes error return would be much prefer

pkg/crc/image/image.go
```diff
@@ -126,9 +127,26 @@ func PullBundle(preset crcpreset.Preset, registryURI string) error {
...
 	if err != nil {
 		return err
 	}
+
+	logging.Info("Extracting bundle gpg signature...")
+	sigImgLayer, err := getLayerPath(imgDigest, 1, "application/vnd.oci.image.layer.v1.tar+gzip")
+	if err != nil {
+		return err
+	}
+	signExtractPath, err := extract.Uncompress(filepath.Join(destDir, sigImgLayer), constants.MachineCacheDir, true)
```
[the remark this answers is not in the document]
cfergeau: > So you are trying to say that both the image tarball and signature should be part of single layer?

It feels simpler to me to have a single layer, and error out if we get something with more layers from the registry? Or will it add complications?

cfergeau:
Something which is not clear to me is how/when we want to ship this? Do we switch all bundles to quay.io? Do we want to keep shipping OpenShift bundles on mirror.openshift.com, while moving the more community-oriented ones to quay.io? Or do we ship everything on quay?
If we want to be able to have at the same time bundles in quay.io, and bundles on mirror.openshift.com, we could extend https://github.com/code-ready/crc/blob/main/pkg/crc/machine/bundle/constants.go to deal with both. This would also allow a progressive switch to using quay.io while we refine the release process (for example we could decide to only put one or 2 bundles there initially).

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/h2non/filetype v1.1.3
 	github.com/hectane/go-acl v0.0.0-20190604041725-da78bae5fc95
 	github.com/jinzhu/copier v0.3.5
+	github.com/keybase/go-crypto v0.0.0-20200123153347-de78d2cb44f4
```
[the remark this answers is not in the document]
praveenkumar: https://pkg.go.dev/golang.org/x/crypto@v0.0.0-20220622213112-05595931fe9d/openpgp is deprecated and not maintained https://github.com/golang/go/issues/44226 so plan is to use a fork which is actively maintained keybase fork is used keybase org and have  patches to support `ed25519` . I thought I put that depreciation link somewhere, will put that as comment.

praveenkumar:
> Something which is not clear to me is how/when we want to ship this? 

End of this sprint.

> Do we switch all bundles to quay.io? 

Yes, plan is to switch all bundles to quay.io 

> Do we want to keep shipping OpenShift bundles on mirror.openshift.com, while moving the more community-oriented ones to quay.io? Or do we ship everything on quay?

Ship everything on quay (also update the bundle on mirror for time being (1-2 sprint) in case outage on quay end) so user can download it manually and use `--bundle` option (as fallback).

gbraad:
> Yes, plan is to switch all bundles to quay.io

can we allow to use mirror first for this? For this we have been signed off. other locations I am not sure how they will respond.

gbraad:
> a bit odd if at a later point we ship some bundles only through mirror.openshift.com or only through an image registry.

Therefore more of a preference option and not conclusive.

go.mod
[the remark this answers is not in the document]
praveenkumar: > It's deprecated, but they still commit security fixes to it.

@cfergeau it doesn't have support for `ed25519` and I don't have issue using ProtonMail , on one the issue keybase/go-crypto was mentioned so used it.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return policyContext, nil
+}
+
+func (img *imageHandler) getSystemContext() *types.SystemContext {
+	return &types.SystemContext{
+		OSChoice:           runtime.GOOS,
+		ArchitectureChoice: runtime.GOARCH,
+	}
+}
```
cfergeau, line 50: fwiw, these are the default values
```
        // If not "", overrides the use of platform.GOARCH when choosing an image or verifying architecture match.
        ArchitectureChoice string
        // If not "", overrides the use of platform.GOOS when choosing an image or verifying OS match.
        OSChoice string
```
probably not worth the helper/..., we can just use an empty context.

pkg/crc/image/image.go
```diff
@@ -112,9 +113,16 @@ func PullBundle(preset crcpreset.Preset) error {
...
 	}
-	_, err = extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
+	fileList, err := extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
 	if err != nil {
 		return err
 	}
+	logging.Debugf("Bundle and sign path %v", fileList)
+
+	logging.Info("Verifying the bundle signature...")
+	if err := gpg.Verify(fileList[0], fileList[1]); err != nil {
```
cfergeau, line 123: I would not hardcode that first file in the archive will always be the bundle, and second file its signature.

pkg/crc/gpg/gpg.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/ProtonMail/go-crypto/openpgp/packet"
+	"github.com/code-ready/crc/pkg/crc/constants"
+)
+
+func readSignature(signatureFilePath string) (*packet.Signature, error) {
+	sigFile, err := os.Open(signatureFilePath)
+	if err != nil {
+		return nil, err
+	}
+	sigFileBlock, err := armor.Decode(sigFile)
```
cfergeau, line 21: https://github.com/code-ready/crc/pull/3221#discussion_r913570665

praveenkumar: @cfergeau I have updated to `ProtonMail/go-crypto` what is missing now?

pkg/crc/image/image.go
praveenkumar: @cfergeau this is the way our script will create bundle, should we compare the size or with the extention?

praveenkumar: It is now compare with extension so should be good.

go.mod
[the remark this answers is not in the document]
gbraad: I lost track of the issues with the different libraries.

You need `ed25519` support?

go.mod
```diff
@@ -31,6 +32,7 @@ require (
 	github.com/mdlayher/vsock v1.1.1
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.19.0
+	github.com/opencontainers/image-spec v1.0.2-0.20190823105129-775207bd45b6
```
cfergeau, line 36: Why such an old version? podman uses `go.mod: github.com/opencontainers/image-spec v1.0.3-0.20220114050600-8b9d41f48198`, `go get -u github.com/opencontainers/image-spec` switches to the `v1.0.2` tag which was created in 2021.

pkg/crc/image/image.go
```diff
@@ -0,0 +1,120 @@
...
+func PullBundle(preset crcpreset.Preset) error {
+	imgHandler := imageHandler{
+		imageURI: uri(preset),
+	}
+	destDir, err := os.MkdirTemp(constants.MachineCacheDir, "tmpBundleImage")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(destDir)
+	imgDigest, err := imgHandler.copyImage(destDir, os.Stdout)
```
cfergeau, line 106: `imgManifest`

pkg/crc/image/image.go
```diff
@@ -0,0 +1,120 @@
...
+	logging.Info("Extracting the image bundle layer...")
+	imgLayer, err := getLayerPath(imgDigest, 0, "application/vnd.oci.image.layer.v1.tar+gzip")
+	if err != nil {
+		return err
+	}
+	_, err = extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
+	if err != nil {
+		return err
+	}
+	return nil
```
cfergeau, line 135: This can be `return err`

pkg/crc/constants/constants.go
```diff
@@ -41,6 +41,21 @@ const (
 	RegistryURI   = "quay.io/crcont"
 	ClusterDomain = ".crc.testing"
 	AppsDomain    = ".apps-crc.testing"
+
+	GPGPublicKey = `-----BEGIN PGP PUBLIC KEY BLOCK-----
```
cfergeau, line 45: Might be worth a comment explaining the identity this key corresponds to, link to a public keyserver, ...?

pkg/crc/gpg/gpg.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+	}()
+
+	signature, err := os.Open(signatureFilePath)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err1 := signature.Close(); err1 != nil {
+			err = err1
```
cfergeau, line 29: Hmm if `err` is already set, won't this hide the initial error? (same question for the other occurrence in this function)

go.mod
```diff
@@ -192,6 +193,7 @@ require (
 
 replace (
 	github.com/apcera/gssapi => github.com/openshift/gssapi v0.0.0-20161010215902-5fb4217df13b
+	golang.org/x/crypto/openpgp => github.com/ProtonMail/go-crypto/openpgp v0.0.0-20220623141421-5afb4c282135
```
cfergeau, line 196: This PR directly imports `github.com/ProtonMail/go-crypto/openpgp`, is the replace required?

pkg/crc/image/image.go
```diff
@@ -112,9 +113,24 @@ func PullBundle(preset crcpreset.Preset) error {
...
-	_, err = extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
+	fileList, err := extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
 	if err != nil {
 		return err
 	}
+	logging.Debugf("Bundle and sign path %v", fileList)
+
+	logging.Info("Verifying the bundle signature...")
+	if len(fileList) != 2 {
+		return fmt.Errorf("image layer contains more files than expected %v", fileList)
```
cfergeau, line 124: Nit: `: %v` rather than just ` %v`

pkg/crc/gpg/gpg.go
```diff
@@ -0,0 +1,42 @@
...
+	"bytes"
+	"fmt"
+	"os"
+
+	"github.com/ProtonMail/go-crypto/openpgp"
+	"github.com/code-ready/crc/pkg/crc/constants"
+)
+
+func Verify(filePath, signatureFilePath string) (err error) {
+	signed, err := os.Open(filePath)
```
cfergeau, line 13: Might be just me, but `signed` feels a bit unspecific. `signedData`, `signedFile`, ... ? Or even omit the `signed` part of it?

cfergeau (approved):
Some minor comments left, but overall looks good!

go.mod
praveenkumar: @cfergeau I am not sure https://github.com/opencontainers/image-spec here I don't even find v1.0.3 tag or branch.

praveenkumar: https://github.com/opencontainers/image-spec/issues/918 there is request for release.

pkg/crc/constants/constants.go
praveenkumar: @cfergeau yes I will to public keyserver and add link here.

pkg/crc/gpg/gpg.go
praveenkumar: yes https://trstringer.com/golang-deferred-function-error-handling/ looks like we need multi error in this case to append the error.

praveenkumar: Since we are open file in reading mode so in defer I think we are OK to just use `f.Close()` without handling the error. if we open file in write mode then it is more appropriate to handle defer function error.

go.mod
cfergeau: > I don't even find v1.0.3 tag or branch

```
$ go get -u github.com/opencontainers/image-spec@main
go: downloading github.com/opencontainers/image-spec v1.0.3-0.20220512140940-7b36cea86235
go: upgraded github.com/opencontainers/image-spec v1.0.2 => v1.0.3-0.20220512140940-7b36cea86235
```
It's go which refers to a non existent 1.0.3 version when we ask it to use the `main` branch. But as I said in my earlier comment, using the latest release (`1.0.2`) is also fine with me.

Looking again at this, it seems the 20190823105129 is misleading, and that `v1.0.2-0.20190823105129-775207bd45b6` might be newer than v1.0.2, but history is complicated with a lot of merges, so not sure :-/

PR #3225, opened 2022-06-20
[Doc] Remove podman/docker dependency to access internal registry

praveenkumar:
This patch uses `oc mirror` command to push image from remote container
registry to openshift internal registry and make it available to
imagestream. Recently user observed if ubi images are pulled using
podman then pushing it to internal registry have signature verification
error (https://github.com/containers/podman/issues/14558) for
mac/windows. This issue is recently fixed on podman remote clients but
still not released so better not to use podman/docker steps just to show
user around internal registry usage.

### DIFF
--- a/docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
+++ b/docs/source/topics/proc_accessing-the-internal-openshift-registry.adoc
@@ -13,9 +13,6 @@ For more information, see link:{crc-gsg-url}#changing-the-selected-preset_gsg[Ch
 For more information, see link:{crc-gsg-url}#starting-the-instance_gsg[Starting the instance].
 * A working {openshift} CLI ([command]`oc`) command.
 For more information, see link:{crc-gsg-url}#accessing-the-openshift-cluster-with-oc_gsg[Accessing the {openshift} cluster with the {openshift} CLI].
-* An installation of [command]`podman` or [command]`docker`.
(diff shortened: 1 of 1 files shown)


cfergeau (commented):
This also switches from alpine to ubi8, this is not mentioned at all in the commit log.
There was one good thing about the previous example based on podman: this was showing that the openshift registry and regular podman/docker were interoperable, something we lost here.
However, not sure crc's doc are thebest place to showcase/document this :)

praveenkumar:
> This also switches from alpine to ubi8, this is not mentioned at all in the commit log. 

Mentioned in the commit log.

> There was one good thing about the previous example based on podman: this was showing that the openshift registry and regular podman/docker were interoperable, something we lost here. However, not sure crc's doc are thebest place to showcase/document this :)

I am also not sure may be a different section in the doc around this interaction of podman/docker with cluster?

robin-owen (approved):
The changes look good to me. I assume they've been tested across the supported platforms, correct?

Also, agreed, we could use a more dedicated section to document interaction between the cluster and Podman.

praveenkumar:
> The changes look good to me. I assume they've been tested across the supported platforms, correct?

Yes, they are tested on all the platform.

anjannath (approved):

anjannath:
> Also, agreed, we could use a more dedicated section to document interaction between the cluster and Podman.

@robin-owen created https://github.com/code-ready/crc/issues/3227 if you want to look at this

PR #3228, opened 2022-06-22
[e2e] Remove podman dependency on story_registry feature

adrianriobo:
Based on #3225 the internal registry story can be checked without rely on a podman installation on the host...this will allows to extend the test to all platforms


test/e2e/features/story_registry.feature
```diff
@@ -1,4 +1,4 @@
-@story_registry @linux
+@story_registry @darwin @linux @windows
```
praveenkumar, line 1: 👍🏼

praveenkumar (approved):

PR #3231, opened 2022-06-23
correct the admin-helper base url

anjannath:
initially the admin-helper release used the v0.0.11 tag
but the admin-helper binary had the version as '0.0.11'
which is set at build time and did not match the git tag

created a new tag 0.0.11 on admin-helper as this is what
the preflight check code expects, which gets the version
from the download URL

this fixes the following error on linux

```
unexpected version of the crc-admin-helper executable: crc-admin-helper-linux version mismatch: v0.0.11 expected but 0.0.11 found in the cache
```


pkg/crc/constants/constants.go
```diff
@@ -24,7 +24,7 @@ const (
 	DaemonLogFile             = "crcd.log"
 	CrcLandingPageURL         = "https://console.redhat.com/openshift/create/local" // #nosec G101
 	DefaultPodmanURLBase      = "https://storage.googleapis.com/libpod-master-releases"
-	DefaultAdminHelperCliBase = "https://github.com/code-ready/admin-helper/releases/download/v0.0.11"
+	DefaultAdminHelperCliBase = "https://github.com/code-ready/admin-helper/releases/download/v%s/%s"
```
cfergeau, line 27: nit: `s/Cli/URL`?

cfergeau (approved):
ACK, one minor comment, feel free to ignore :)

PR #3234, opened 2022-06-24
cut 2.5.0

anjannath:


Makefile
```diff
@@ -3,7 +3,7 @@ SHELL := /bin/bash
 OPENSHIFT_VERSION ?= 4.10.18
 PODMAN_VERSION ?= 4.0.2
 BUNDLE_EXTENSION = crcbundle
-CRC_VERSION = 2.4.1
+CRC_VERSION = 2.5.0
```
praveenkumar, line 6: 👍🏼

praveenkumar (approved):

praveenkumar:
Wait for CI run

cfergeau:
Are we going to generate new podman bundles ?

anjannath:
> Are we going to generate new podman bundles ?

since the issue with `podman-remote` binary is only on linux, and although i don't like it that it crashes `crc start` but assuming `podman` preset on linux users are very minimal and fedora coreos stream not being available with podman `4.1.1` binaries out of the box, i don't know when we can generate a podman bundle.

i'd prefer to have new podman bundle in the next release

praveenkumar:
> Are we going to generate new podman bundles ?

This is known issue even with last release only 4.1.1 release of podman have static bit correctly available.

cfergeau:
> This is known issue even with last release only 4.1.1 release of podman have static bit correctly available.

Yes, which means we could for example generate a 4.1.0 podman bundle, and use the 4.1.1 linux binary in order to avoid a bad looking bug two releases in a row.

PR #3240, opened 2022-06-28
ubuntu: Fix apparmor rule for podman preset

cfergeau:
On Ubuntu, crc's preflights add an apparmor rule to allow the use of
cacheDir/*/crc.qcow2 by libvirt. With podman bundles, the file is named
`crc-podman.qcow2`, which causes `crc start` to fail with permission
errors.


praveenkumar:
@cfergeau did you tested this on ubuntu or it need to be tested?

cfergeau:
> @cfergeau did you tested this on ubuntu or it need to be tested?

I've tested it on Ubuntu 20.04

praveenkumar (approved):

PR #3260, opened 2022-07-14
Fix AddUserToHypervAdminGroup custom action 

### LINKED ISSUE #3254: [BUG] windows msi installer fail to add local user to Hyper-V Administrators group
Testing on a fresh environment:

1. run msi installer 
2. reboot computer
3. run `crc setup`

Expected 

System fully configured and ready to `start crc`

Actual

User is added to Hyper-V Administrator group during `crc setup`, so after finish it asks for a second reboot

adrianriobo:
Fixes #3254


adrianriobo:
Hold as I have tested this with local user, but would like to give a try with AD user

praveenkumar:
Tested it on a clean VM and it solve one of the issue where user have to reboot the system again after rebooting it for installer.

Issue #3012, opened 2022-02-16
More 'OpenShift cluster/instance'/'virtual machine' occurrences


cfergeau:
This issue is a followup to https://github.com/code-ready/crc/commit/e2a76bbba5653a2c9046efce6753f075057cda06
The conversation below happened in https://github.com/code-ready/crc/pull/3007

> git grep -i 'openshift instance' and git grep -i 'virtual machine' still return a few results so it's not just e2e which was missed.

On pkg and test side only those places are remaining but there is lot of place in the doc which mention virtual machine so @robin-owen should let us know if we need to changes those with instance.

```
cmd/crc/cmd/podman_env.go:      podmanEnvCmd.Flags().BoolVar(&root, "root", false, "Use root podman in the virtual machine")
pkg/crc/config/settings.go:             fmt.Sprintf("Virtual machine preset (alpha feature - valid values are: %s or %s)", preset.Podman, preset.OpenShift))
pkg/crc/machine/config/config.go:       // Virtual machine configuration
pkg/crc/machine/console.go:             return nil, errors.Wrap(err, "Error getting the state for virtual machine")
pkg/crc/machine/start.go:               return nil, errors.Wrap(err, "Failed to connect to the CRC VM with SSH -- virtual machine might be unreachable")
pkg/crc/machine/start.go:               return fmt.Errorf("Too little memory (%s) allocated to the virtual machine to start the monitoring stack, %s is the minimum",
pkg/crc/machine/start.go:               return fmt.Errorf("Error saving virtual machine to store after attempting creation: %s", err)
pkg/crc/machine/status.go:              return nil, errors.Wrap(err, fmt.Sprintf("Cannot load '%s' virtual machine", client.name))
pkg/crc/machine/sync.go:                logging.Infof("Cancelling virtual machine start...")
pkg/crc/preflight/preflight_checks_windows.go:  // Check if Hyper-V's Virtual Machine Management Service is running
pkg/crc/preflight/preflight_checks_windows.go:          return fmt.Errorf("Hyper-V Virtual Machine Management service not running")
pkg/crc/preflight/preflight_linux.go:                   cleanupDescription: "Removing crc's virtual machine",
pkg/crc/preflight/preflight_windows.go:         cleanupDescription: "Removing crc's virtual machine",
test/e2e/features/basic.feature:        Then stderr should contain "Removing crc's virtual machine"
test/e2e/features/basic.feature:        And stderr should contain "Removing crc's virtual machine"
```

[...]

PR #3272, opened 2022-07-25
Add OKD as preset

praveenkumar:
## Solution/Idea

This PR adds okd as preset so now with same crc binary user can choose to run ocp or okd. Default preset still be Openshift.

## Proposed changes

- Add okd as preset

## Testing
- crc config set preset okd
- crc setup => should download the okd bundle
- crc start => should provision the okd cluster.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -79,7 +79,7 @@ func runPrerun(cmd *cobra.Command) error {
 	}
 	logging.InitLogrus(logFile)
 
-	for _, str := range defaultVersion().lines() {
+	for _, str := range defaultVersion(crcConfig.GetPreset(config)).lines() {
 		logging.Debugf(str)
 	}
 	return nil
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os"
 	"runtime"
-	"strings"
 	"text/template"
 
 	"github.com/Masterminds/semver/v3"
@@ -156,18 +155,7 @@ func (s *startResult) prettyPrintTo(writer io.Writer) error {
 		return errors.New("either Error or ClusterConfig is needed")
 	}
 
-	if err := writeTemplatedMessage(writer, s); err != nil {
-		return err
-	}
-	if crcversion.IsOkdBuild() {
-		_, err := fmt.Fprintln(writer, strings.Join([]string{
-			"",
-			"NOTE:",
-			"This cluster was built from OKD - The Community Distribution of Kubernetes that powers Red Hat OpenShift.",
-			"If you find an issue, please report it at https://github.com/openshift/okd"}, "\n"))
-		return err
-	}
-	return nil
+	return writeTemplatedMessage(writer, s)
 }
 
 func validateStartFlags() error {
@@ -253,6 +241,10 @@ Use the 'podman' command line interface:
   {{ .CommandLinePrefix }} {{ .EvalCommandLine }}
   {{ .CommandLinePrefix }} {{ .PodmanRemote }} COMMAND
 `
+	startTemplateForOKD = `NOTE:
+This cluster was built from OKD - The Community Distribution of Kubernetes that powers Red Hat OpenShift.
+If you find an issue, please report it at https://github.com/openshift/okd
+`
 )
 
 type templateVariables struct {
@@ -263,7 +255,7 @@ type templateVariables struct {
 }
 
 func writeTemplatedMessage(writer io.Writer, s *startResult) error {
-	if s.ClusterConfig.ClusterType == preset.OpenShift {
+	if s.ClusterConfig.ClusterType == preset.OpenShift || s.ClusterConfig.ClusterType == preset.OKD {
 		return writeOpenShiftTemplatedMessage(writer, s)
 	}
 
@@ -271,7 +263,11 @@ func writeTemplatedMessage(writer io.Writer, s *startResult) error {
 }
 
 func writeOpenShiftTemplatedMessage(writer io.Writer, s *startResult) error {
-	parsed, err := template.New("template").Parse(startTemplateForOpenshift)
+	tmpl := startTemplateForOpenshift
+	if s.ClusterConfig.ClusterType == preset.OKD {
+		tmpl = fmt.Sprintf("%s\n\n%s", startTemplateForOpenshift, startTemplateForOKD)
+	}
+	parsed, err := template.New("template").Parse(tmpl)
 	if err != nil {
 		return err
 	}
--- a/cmd/crc/cmd/version.go
+++ b/cmd/crc/cmd/version.go
@@ -7,6 +7,7 @@ import (
 
 	crcConfig "github.com/code-ready/crc/pkg/crc/config"
 	"github.com/code-ready/crc/pkg/crc/logging"
+	crcPreset "github.com/code-ready/crc/pkg/crc/preset"
 	crcversion "github.com/code-ready/crc/pkg/crc/version"
 	"github.com/spf13/cobra"
 )
@@ -21,7 +22,7 @@ var versionCmd = &cobra.Command{
 	Short: "Print version information",
 	Long:  "Print version information",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runPrintVersion(os.Stdout, defaultVersion(), outputFormat)
+		return runPrintVersion(os.Stdout, defaultVersion(crcConfig.GetPreset(config)), outputFormat)
 	},
 }
 
@@ -39,11 +40,11 @@ type version struct {
 	PodmanVersion    string `json:"podmanVersion"`
 }
 
-func defaultVersion() *version {
+func defaultVersion(preset crcPreset.Preset) *version {
 	return &version{
 		Version:          crcversion.GetCRCVersion(),
 		Commit:           crcversion.GetCommitSha(),
-		OpenshiftVersion: crcversion.GetBundleVersion(),
+		OpenshiftVersion: crcversion.GetBundleVersion(preset),
 		PodmanVersion:    crcversion.GetPodmanVersion(),
 	}
 }
--- a/pkg/crc/cluster/pullsecret.go
+++ b/pkg/crc/cluster/pullsecret.go
@@ -15,7 +15,6 @@ import (
 	"github.com/code-ready/crc/pkg/crc/logging"
 	"github.com/code-ready/crc/pkg/crc/preset"
 	"github.com/code-ready/crc/pkg/crc/validation"
-	crcversion "github.com/code-ready/crc/pkg/crc/version"
 	crcos "github.com/code-ready/crc/pkg/os"
 
 	"github.com/AlecAivazis/survey/v2"
@@ -77,8 +76,8 @@ func NewNonInteractivePullSecretLoader(config crcConfig.Storage, path string) Pu
 }
 
 func (loader *nonInteractivePullSecretLoader) Value() (string, error) {
-	// If crc is built from an OKD bundle or podman bundle is used, then use the fake pull secret in contants.
-	if crcversion.IsOkdBuild() || crcConfig.GetPreset(loader.config) == preset.Podman {
+	// If crc is built from an OKD bundle or podman bundle is used, then use the fake pull secret in constants.
+	if crcConfig.GetPreset(loader.config) == preset.OKD || crcConfig.GetPreset(loader.config) == preset.Podman {
 		return constants.OkdPullSecret, nil
 	}
 
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -92,9 +92,9 @@ func defaultBundleForOs(preset crcpreset.Preset) map[string]string {
 		}
 	}
 	return map[string]string{
-		"darwin":  fmt.Sprintf("crc_vfkit_%s_%s.crcbundle", version.GetBundleVersion(), runtime.GOARCH),
-		"linux":   fmt.Sprintf("crc_libvirt_%s_%s.crcbundle", version.GetBundleVersion(), runtime.GOARCH),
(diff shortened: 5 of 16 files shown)


cmd/crc/cmd/version.go
```diff
@@ -39,11 +40,11 @@ type version struct {
...
 }
 
-func defaultVersion() *version {
+func defaultVersion(preset crcPreset.Preset) *version {
 	return &version{
 		Version:          crcversion.GetCRCVersion(),
 		Commit:           crcversion.GetCommitSha(),
-		OpenshiftVersion: crcversion.GetBundleVersion(),
+		OpenshiftVersion: crcversion.GetBundleVersion(preset),
 		PodmanVersion:    crcversion.GetPodmanVersion(),
```
cfergeau, line 48: Looks like after this PR we'll be reporting either (OpenShiftVersion, PodmanVersion) or (OKDVersion, PodmanVersion) depending on the preset? Wondering if we should always be showing all 3 versions, OpenShiftVersion, PodmanVersion, OKDVersion?

pkg/crc/constants/constants.go
```diff
@@ -92,9 +92,9 @@ func defaultBundleForOs(preset crcpreset.Preset) map[string]string {
 		}
 	}
 	return map[string]string{
-		"darwin":  fmt.Sprintf("crc_vfkit_%s_%s.crcbundle", version.GetBundleVersion(), runtime.GOARCH),
-		"linux":   fmt.Sprintf("crc_libvirt_%s_%s.crcbundle", version.GetBundleVersion(), runtime.GOARCH),
-		"windows": fmt.Sprintf("crc_hyperv_%s_%s.crcbundle", version.GetBundleVersion(), runtime.GOARCH),
+		"darwin":  fmt.Sprintf("crc_vfkit_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
+		"linux":   fmt.Sprintf("crc_libvirt_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
+		"windows": fmt.Sprintf("crc_hyperv_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
```
cfergeau, line 97: Similarly to the previous comment, why not `crc_okd_vfkit_%s_%s.crcbundle` in line with the podman naming?

pkg/crc/version/version.go
```diff
@@ -67,8 +68,11 @@ func GetCommitSha() string {
...
 }
 
-func GetBundleVersion(_ crcPreset.Preset) string {
-	return bundleVersion
+func GetBundleVersion(preset crcPreset.Preset) string {
+	if preset == crcPreset.OpenShift {
+		return bundleVersion
+	}
+	return okdVersion
 }
```
cfergeau, line 74: Not fully clear why we have a GetBundleVersion depending on preset, but only for OpenShift/OKD. For Podman, one has to use GetPodmanVersion instead :-/

cmd/crc/cmd/start.go
```diff
@@ -263,15 +255,19 @@ type templateVariables struct {
...
 
 	return writePodmanTemplatedMessage(writer, s)
 }
 
 func writeOpenShiftTemplatedMessage(writer io.Writer, s *startResult) error {
-	parsed, err := template.New("template").Parse(startTemplateForOpenshift)
+	tmpl := startTemplateForOpenshift
+	if s.ClusterConfig.ClusterType == preset.OKD {
+		tmpl = startTemplateForOKD
+	}
```
cfergeau, line 269: The previous code was first printing the OpenShift start template, and then appending a warning after it when OKD is in use. This seems to have been lost here, and to only print the warning for OKD?

pkg/crc/cluster/pullsecret.go
```diff
@@ -78,7 +77,7 @@ func NewNonInteractivePullSecretLoader(config crcConfig.Storage, path string) Pu
 
 func (loader *nonInteractivePullSecretLoader) Value() (string, error) {
 	// If crc is built from an OKD bundle or podman bundle is used, then use the fake pull secret in contants.
```
cfergeau, line 79: Can you fix the 'contants' typo while you are changing code in that area?

pkg/crc/machine/start.go
```diff
@@ -793,7 +793,7 @@ func updateKubeconfig(ctx context.Context, ocConfig oc.Config, sshRunner *crcssh
 }
 
 func bundleMismatchWithPreset(preset crcPreset.Preset, bundleMetadata *bundle.CrcBundleInfo) error {
-	if preset != crcPreset.OpenShift && bundleMetadata.IsOpenShift() {
+	if preset == crcPreset.Podman && bundleMetadata.IsOpenShift() {
```
cfergeau, line 796: Arguably that's `bundleMetadata.IsOpenShift()` which is buggy, it should return false for an OKD bundle. 
I think we have no way of detecting this mistake:
```
crc config set preset okd
crc start --bundle crc_libvirt_4.10.14.crcbundle
```
with `crc_libvirt_4.10.14.crcbundle` being an OpenShift (not OKD) bundle. This will skip the pull secret and will fail to start.
I'd add a few more details to the commit log as this is not fully obvious what's broken/why we don't add more mistmatch detection now that we have more presets.

cfergeau (commented):
Looks like we could use the okd preset to make the distinction between okd and openshift bundles even clearer/more separate in the codebase, I'm thinking for example of GetBundleVersion vs GetPodmanVersion, and the bundle naming. Dunno if you considered this a bit while working on that PR?

cmd/crc/cmd/version.go
praveenkumar: @cfergeau Moving forward, we don't want to show any preset version as part of `crc version` command.  As of now it make sense because our crc release are bound to a specific preset bundle which will change. `crc status` should be source of truth around what preset running and which version. So as of now I don't want to change it so you are right currently it  (OpenShiftVersion, PodmanVersion) or (OKDVersion, PodmanVersion).

cmd/crc/cmd/root.go
```diff
@@ -79,7 +79,7 @@ func runPrerun(cmd *cobra.Command) error {
 	}
 	logging.InitLogrus(logFile)
 
-	for _, str := range defaultVersion().lines() {
+	for _, str := range defaultVersion(crcConfig.GetPreset(config)).lines() {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau good catch, fixing it.

pkg/crc/version/version.go
praveenkumar: @cfergeau I will take care of that also, ideally `GetBundleVersion` should return any preset bundle version (okd/ocp/podman or microshift ..etc.). Not with this PR as of now.

cmd/crc/cmd/start.go
praveenkumar: Ah right, will update 👍🏼

pkg/crc/cluster/pullsecret.go
praveenkumar: Ack, will do.

pkg/crc/machine/start.go
praveenkumar: Right, as of now bundle created from snc (ocp/okd) have type as `snc` which is correct for (single node cluster) both the cases. We do need to figure out best way to detect if bundle is OCP/OKD and print out the mismatch here. May be add another field in bundle-info data (snc side) as `preset` and put `ocp/okd/podman` according to bundle generation. As of now it is half backed solution (buggy one) but good as long as user manually download bundle and use it. I will put this details in the commit log and then start putting stuff on snc side.

pkg/crc/constants/constants.go
praveenkumar: @cfergeau This change should come from snc side. I can create issue on snc and then have bundle names as `crc_okd_..` but as of now it should be an improvement not blocker for PR.

cfergeau:
I had to add
```
diff --git a/pkg/crc/machine/vsock.go b/pkg/crc/machine/vsock.go
index 9d2e4a9f..d6a60e71 100644
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -94,6 +94,7 @@ func vsockPorts(preset crcPreset.Preset) []types.ExposeRequest {
        }
        switch preset {
        case crcPreset.OpenShift:
+       case crcPreset.OKD:
                exposeRequest = append(exposeRequest,
                        types.ExposeRequest{
                                Protocol: "tcp",
```
for testing

cfergeau:
Fwiw, the stable okd version you picked is now red in https://amd64.origin.releases.ci.openshift.org/#4-stable :( But it's because of upgrade failures apparently, which is not a problem for us.

praveenkumar:
> I had to add
> 
> ```
> diff --git a/pkg/crc/machine/vsock.go b/pkg/crc/machine/vsock.go
> index 9d2e4a9f..d6a60e71 100644
> --- a/pkg/crc/machine/vsock.go
> +++ b/pkg/crc/machine/vsock.go
> @@ -94,6 +94,7 @@ func vsockPorts(preset crcPreset.Preset) []types.ExposeRequest {
>         }
>         switch preset {
>         case crcPreset.OpenShift:
> +       case crcPreset.OKD:
>                 exposeRequest = append(exposeRequest,
>                         types.ExposeRequest{
>                                 Protocol: "tcp",
> ```
> 
> for testing

@cfergeau ah yes because as of now we are only exposing it for Openshift preset. will add it 👍🏼

praveenkumar:
> Fwiw, the stable okd version you picked is now red in https://amd64.origin.releases.ci.openshift.org/#4-stable :( But it's because of upgrade failures apparently, which is not a problem for us.

Ah it is one from the github releases but yes upgrade failure is not issue for us.

cmd/crc/cmd/version.go
cfergeau: I know we have plans to change this in the future. However, this is not done yet, and right now we're getting:
```
$ ./out/linux-amd64/crc version
CRC version: 2.6.0+f1c25cbc
OpenShift version: 4.10.0-0.okd-2022-06-24-212905
Podman version: 4.1.0
```
which is misleading and wrong. I don't mind if this becomes
```
$ ./out/linux-amd64/crc version
CRC version: 2.6.0+f1c25cbc
OKD version: 4.10.0-0.okd-2022-06-24-212905
Podman version: 4.1.0
```
or 
```
$ ./out/linux-amd64/crc version
CRC version: 2.6.0+f1c25cbc
OpenShift version: 4.10.22
OKD version: 4.10.0-0.okd-2022-06-24-212905
Podman version: 4.1.0
```
even though I have a preference for the latter as this means `crc version` is not going to change output depending on the preset.

cfergeau:
Another missing update: https://github.com/code-ready/crc/blob/b91afcd751135be1d3955f1c714d70b99bd4db93/pkg/crc/constants/constants.go#L186-L209

pkg/crc/version/version.go
cfergeau: I'd just go with `GetOKDVersion()` for now, this way there's no way to misuse/misunderstand `GetBundleVersion(preset)`

cfergeau (commented):
I tested this and it's working fine. We need to clear up the versioning ("OpenShift: 4.10.0-okd") before making a release with this preset (ideally before merging this).
There's also a set of misleading messages during cluster startup
```
INFO Creating CRC VM for OpenShift 4.10.0-0.okd-2022-06-24-212905...
INFO Starting CRC VM for OpenShift 4.10.0-0.okd-2022-06-24-212905...
INFO Starting OpenShift cluster... [waiting for the cluster to stabilize]
```

The last line is related to https://github.com/code-ready/crc/issues/3012

praveenkumar:
All the suggestions are now incorporated.

pkg/crc/constants/constants.go
```diff
@@ -200,6 +202,8 @@ func GetDefaultMemory(preset crcpreset.Preset) int {
 	switch preset {
 	case crcpreset.OpenShift:
 		return 9216
+	case crcpreset.OKD:
```
anjannath, line 205: could've been 
`case crcpreset.OpenShift, crcpreset.OKD` since both have the same CPU and memory values

pkg/crc/version/version.go
anjannath: adding a `GetOKDVersion()` helper and not importing the presets on the version package will be good, the user of the version package can then decide which bundle version (okd/ocp/podman) to use depending on the preset

pkg/crc/constants/constants.go
praveenkumar: @anjannath ye we can do that also, I just wanted to explicit

anjannath:
There's another two instance of `OpenShift` on the start logs, that we might want to change to `OKD`

```
INFO Starting OpenShift kubelet service
```

and at the end before the oc login templates..
```
```

anjannath:
cert rotation test is failing, re-running the `e2e` tests:
```
 --- Failed steps:
  Scenario: Start CRC # features/cert_rotation.feature:16
    When starting CRC with default bundle along with stopped network time synchronization succeeds # features/cert_rotation.feature:17
      Error: command 'CRC_DISABLE_UPDATE_CHECK=true CRC_DEBUG_ENABLE_STOP_NTP=true crc start -p '/home/packer/pull-secret' -b /home/packer/crc_libvirt_4.10.22_amd64.crcbundle', expected to succeed, exited with exit code: 1 
```

/test e2e-crc

praveenkumar:
> There's another two instance of `OpenShift` on the start logs, that we might want to change to `OKD`
> 
> ```
> INFO Starting OpenShift kubelet service

I think here we should remove `OpenShift` all together. It should be just `INFO Starting kubelet service`

> ```
> 
> and at the end before the oc login templates..
> 
> ```
> Started the OpenShift cluster.
> ```

This is coming from start template and at the end we are already having `Note` section for `okd` preset.

praveenkumar:
@anjannath CI failure is not because of this PR.

anjannath (approved):
LGTM

PR #3273, opened 2022-07-25
Fix several windows github actions failures

cfergeau:
The bigger issue is some golang/mingw bug described in https://github.com/golang/go/issues/46099
This is worked around by downgrading mingw. But I've been hitting quite a few download issues when trying to downgrade mingw, so let's see how reliable this is.
Then this fixes some issue with golangci-lint and GOBIN not being set to an absolute path
The last commit is a tentative fix for some intermittent issues when uploading the installer artifacts. I'm not sure this will fix anything at all, but it's worth trying and see if the failures still happen.

### DIFF
--- a/.github/workflows/macos-installer.yml
+++ b/.github/workflows/macos-installer.yml
@@ -50,5 +50,5 @@ jobs:
       - name: Upload macOS installer artifact
         uses: actions/upload-artifact@v2
         with:
-          name: macOS Installer
+          name: macOS Installer (${{ matrix.os }})
           path: "./out/macos-amd64/crc-macos-amd64.pkg"
--- a/.github/workflows/make-check-win.yml
+++ b/.github/workflows/make-check-win.yml
@@ -17,6 +17,8 @@ jobs:
         go:
           - 1.17
     steps:
+      - name: Downgrade mingw to workaround https://github.com/golang/go/issues/46099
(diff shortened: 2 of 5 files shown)


cfergeau:
/hold
This needs some more work to solve this CI failure
```
 cd tools && GOBIN="D:\a\crc\crc\tools\bin" go install github.com/golangci/golangci-lint/cmd/golangci-lint
GOOS=darwin D:\a\crc\crc\tools\bin/golangci-lint run
/usr/bin/bash: D:acrccrctoolsbin/golangci-lint: No such file or directory
```

gbraad:
We have always had issues with building on Windows...
... I do not even rely on it; cross-compiling has worked for me reliably.

Maintaining this seems to be a PITA

cfergeau:
> We have always had issues with building on Windows... ... I do not even rely on it; cross-compiling has worked for me reliably.
> 
> Maintaining this seems to be a PITA

Hopefully at some point it will work and will keep working forever :) Having `make check` run on Windows might be exercising some codepath/test cases we don't run at all otherwise, but I would not object too much if we prefer to drop it. Building the installer for every PR has its use though, to make sure we don't break this when making changes.

praveenkumar:
These changes are valid and hopefully provide us consistent green bits for windows.

Issue #3258, opened 2022-07-13
Migrate to ginko v2 as v1 is deprecated


anjannath:
> With the release of Ginkgo 2.0 the 1.x version is formally deprecated and no longer supported. All future development will occur on version 2.

From https://onsi.github.io/ginkgo/MIGRATING_TO_V2

anjannath:
fixed in #3274

PR #3274, opened 2022-07-27
Upgrade ginkgo to v2

jsliacan:
**Fixes:** Issue #3258 

Integration tests should run as before.

Output file seems to have changed structure, might need to specify some options in reporter config.


Makefile
```diff
@@ -203,7 +203,7 @@ ifndef PULL_SECRET_PATH
 export PULL_SECRET_PATH = $(HOME)/Downloads/crc-pull-secret
 endif
 ifndef BUNDLE_PATH
-export BUNDLE_PATH = $(HOME)/Downloads/crc_libvirt_$(OPENSHIFT_VERSION).$(BUNDLE_EXTENSION)
+export BUNDLE_PATH = $(HOME)/Downloads/crc_libvirt_$(OPENSHIFT_VERSION)_amd64.$(BUNDLE_EXTENSION)
```
praveenkumar, line 206: @jsliacan This should be part of different commit since here you are adding the arch explicit.

jsliacan: @praveenkumar Thanks for noticing. Should I remove it or create an `ARCH` var in expectation of arm or keep it hardcoded like it is?

praveenkumar: @jsliacan better to use `GOARCH` env from the Makefile.

praveenkumar:
@jsliacan integration tests failing on the CI with following error, which means it is not able to generate the Junit files.
```
 ------------------------------
[ReportAfterSuite] [FAILED] [0.001 seconds]
[ReportAfterSuite] Autogenerated ReportAfterSuite for --junit-report
autogenerated by Ginkgo
  Failed to generate JUnit report:
  open out/integration.xml: no such file or directory
  In [ReportAfterSuite] at: /home/packer/crc/vendor/github.com/onsi/ginkgo/v2/reporting_dsl.go:127 
```

test/integration/testsuite_test.go
```diff
@@ -31,6 +31,10 @@ func TestTest(t *testing.T) {
 
 	// fetch the current (reporter) config
 	_, reporterConfig := GinkgoConfiguration()
+	err := os.Mkdir("out", 0775)
```
praveenkumar, line 34: I should use `os.MkdirAll` because in case `out` directory already exist then this will through an error but `MkdirAll` will not.
```
// If path is already a directory, MkdirAll does nothing
// and returns nil.
```

PR #3275, opened 2022-07-27
update Podman bundle to 4.1.1

praveenkumar:


cfergeau (changes requested):
The rationale for the 2.7.0-dev change is buried in slack, it would be useful to have it in the commit log as well

> Reason behind it, I can push bundles to quay and tag the dev release so that it doesn't affect the existing release bundles.

I assume we want podman 4.1.1 because this is what the latest fcos36 images have, it would also be useful to have this in the commit log rather than leaving that up for guessing.

praveenkumar:
@cfergeau Reason I didn't put that in commit log because that was something we were doing before and then stopped. Now we need it again because of this reason. if you want it I will update the commit for for me it is just adding same which was there before.

cfergeau:
It was apparently important enough to write it on slack, from my perspective this means this deserved to be in the commit log from the start so that everyone looking at this understands why this is important/needed/...

cfergeau:
Both rpm and Windows installer builds are broken by this change though.

cfergeau:
Not a huge fan of the 2.6.0.999 versioning :-/
Regarding the upload of container images tagged with this version, are they going to be removed when a newer development image is uploaded/when an official release is made which supersedes this image? I think they should be removed in both cases, otherwise we'll have a hard time remembering that some of these images are not official ones.

cfergeau (approved):

praveenkumar:
> Regarding the upload of container images tagged with this version, are they going to be removed when a newer development image is uploaded/when an official release is made which supersedes this image? 

In case of podman it depends, if a new version of podman hit to fedora stable or there is some CVE which need us to rebuild the same version of podman image then we should remove the old one otherwise this image remapped with the crc release version and ship as it is.

> I think they should be removed in both cases, otherwise we'll have a hard time remembering that some of these images are not official ones. Or we could use a separate namespace for these (quay.io/crcont/podman-bundle-dev) to make the distinction very clear

In case of OKD or OCP (when quay is used) we do need to regularly prune those images from registry as soon as new images are available and tested during dev cycle but images which mapped with a specific crc release shouldn't removed.

cfergeau:
> In case of podman it depends, if a new version of podman hit to fedora stable or there is some CVE which need us to rebuild the same version of podman image then we should remove the old one otherwise this image remapped with the crc release version and ship as it is.

I don't think we should ever remove released images/overwrite them with new images/... It's always good for users to be able to roll back to the image they were using before if the new one has critical problems for them.

> In case of OKD or OCP (when quay is used) we do need to regularly prune those images from registry as soon as new images are available and tested during dev cycle but images which mapped with a specific crc release shouldn't removed.

I'm only asking about the -dev images, which are uploaded to the same image registry as the released images, use versioning which might be mistaken for an official release, ...  If they stay around, this could get coonfusing.

PR #3283, opened 2022-08-03
[WIP] Create Preset interface

praveenkumar:


praveenkumar:
The unit tests around api server fails because we are using now preset as interface for the struct and it is not able to parse it. I put a sample code to pin point the issue so looks like we need a custom unmarshal which can add proper preset or we need to change the config to `string` and use `preset.String()` but that means we need more changes.

```golang
package main

import (
	"bytes"
	"encoding/json"
	"fmt"
)

type preset interface {
	fmt.Stringer
	json.Marshaler
}

type OpenShiftPreset struct{}

func (preset OpenShiftPreset) String() string {
	return "openshift"
}

func (preset OpenShiftPreset) MarshalJSON() ([]byte, error) {
	return json.Marshal(preset.String())
}

var OpenShift = OpenShiftPreset{}

type clusterConfig struct {
	Preset        preset
	ClusterCACert string
	WebConsoleURL string
	URL           string
}

var DummyClusterConfig = clusterConfig{
	Preset:        OpenShift,
	ClusterCACert: "MIIDODCCAiCgAwIBAgIIRVfCKNUa1wIwDQYJ",
	URL:           "https://foo.testing:6443",
	WebConsoleURL: "https://console.foo.testing:6443",
}

func main() {
	var data = new(bytes.Buffer)
	var conf = clusterConfig{}
	if err := json.NewEncoder(data).Encode(DummyClusterConfig); err != nil {
		fmt.Errorf("Failed to encode data to JSON: %w", err)
	}
	fmt.Printf("Data: %s\n", data)
	if err := json.Unmarshal(data.Bytes(), &conf); err != nil {
		fmt.Errorf("Failed to unmarshal data to JSON: %w", err)
	}
	fmt.Printf("Conf: %#v\n", conf)
}
```

- Output
```
$ go run js.go 
Data: {"Preset":"openshift","ClusterCACert":"MIIDODCCAiCgAwIBAgIIRVfCKNUa1wIwDQYJ","WebConsoleURL":"https://console.foo.testing:6443","URL":"https://foo.testing:6443"}

Conf: main.clusterConfig{Preset:main.preset(nil), ClusterCACert:"MIIDODCCAiCgAwIBAgIIRVfCKNUa1wIwDQYJ", WebConsoleURL:"https://console.foo.testing:6443", URL:"https://foo.testing:6443"}
```

Playground: https://go.dev/play/p/Y9XTHlBV44U

PR #3286, opened 2022-08-04
Add config option to enable or disable shared directories 

### LINKED ISSUE #3285: Add a config option to enable/disable directory sharing
While testing on RHEL 8.5 and libvirt 8.0.0 `crc start` ends up with error `Failed to create the VM: virError(Code=67, Domain=10, Message='unsupported configuration: 'virtiofs' requires shared memory')"`

but the podman preset on macOS where this is most beneficial works without any issues we can add a config option `enable-shared-dirs` to provide an option to the user to disable it if they face any issues

anjannath:
This adds a new config option 'enable-shared-dirs' for linux and
macOS which will be used to control the behavior of automounting
the user's $HOME dir on the crc vm during start, config type  is
bool and the default value is true

We are adding this to provide the user an option to disable this
feature if they face any issues, in our testing we found  issues
with RHEL 8.5 and libvirt 8.0.0 where 'crc start' errors out as:

```
level=warning msg="Failed to create the VM: virError(Code=67, Domain=10, Message='unsupported configuration: 'virtiofs' requires shared memory')"
```

Fixes #3285

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -74,6 +74,7 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		PullSecret:        cluster.NewInteractivePullSecretLoader(config),
 		KubeAdminPassword: config.Get(crcConfig.KubeAdminPassword).AsString(),
 		Preset:            crcConfig.GetPreset(config),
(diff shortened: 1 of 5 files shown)


praveenkumar:
This PR is doing what is suppose to do but the RHEL issue is still there. I am merging this.

PR #3289, opened 2022-08-05
Revert "config: Force use of podman preset on M1 CPUs"

praveenkumar:
We now have a way to create ocp bundles for M1 so reverting this
PR.
This reverts commit b14fb9dd247e8860daa848e5966ca0aff1af46ec.


anjannath (approved):
tested on M1 and able to change preset and start an openshift cluster

PR #3308, opened 2022-08-22
Add CIFS/SMB based file sharing for windows 

anjannath:
Needs https://github.com/code-ready/machine/pull/66

## Solution/Idea

the default CIFS/SMB feature provided from windows
can be used to share directories over the  network
we can use it to share files between host and  crc
vm

Adds the following config options:
- `enable-shared-dirs` (was present on macOS and linux now added to windows)
- `smb-share-password` 

## Testing
0. use the `msi` from this PR to install `crc`
1. run `crc config set enable-shared-dirs true` and follow the instructions from the validation msg (it should ask to configure `smb-share-password` first)
2. run `crc config set preset podman` and `crc setup`
3. `crc start` with podman preset and bundle from https://github.com/code-ready/snc/pull/565 should succeed (download from openshift CI artifacts)
4. container with a volume mount should work, e.g: `podman run -v /mnt/home/crc/conf:/etc/http/conf httpd`

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -75,9 +75,9 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		PullSecret:        cluster.NewInteractivePullSecretLoader(config),
 		KubeAdminPassword: config.Get(crcConfig.KubeAdminPassword).AsString(),
 		Preset:            crcConfig.GetPreset(config),
-		EnableSharedDirs:  crcConfig.ShouldEnableSharedDirs(config),
 		IngressHTTPPort:   config.Get(crcConfig.IngressHTTPPort).AsUInt(),
 		IngressHTTPSPort:  config.Get(crcConfig.IngressHTTPSPort).AsUInt(),
+		EnableSharedDirs:  config.Get(crcConfig.EnableSharedDirs).AsBool(),
 	}
 
 	client := newMachine()
@@ -94,6 +94,17 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 				Code: preflightFailedExitCode,
 			}
 		}
+	}
+
+	if runtime.GOOS == "windows" {
+		username, err := crcos.GetCurrentUsername()
+		if err != nil {
+			return nil, err
+		}
+
+		// config SharedDirPassword ('shared-dir-password') only exists in windows
+		startConfig.SharedDirPassword = config.Get(crcConfig.SharedDirPassword).AsString()
+		startConfig.SharedDirUsername = username
 	}
 
 	return client.Start(ctx, startConfig)
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/cavaliergopher/grab/v3 v3.0.1
 	github.com/cheggaaa/pb/v3 v3.0.8
 	github.com/code-ready/admin-helper v0.0.10
-	github.com/code-ready/machine v0.0.0-20220727160217-7bf0dd8c2d7b
+	github.com/code-ready/machine v0.0.0-20220927132822-3408fdecc41c
 	github.com/code-ready/vfkit v0.0.2
 	github.com/containers/gvisor-tap-vsock v0.4.1-0.20220908123107-8d91f6d62de2
 	github.com/containers/image/v5 v5.15.0
--- a/packaging/windows/product.wxs.template
+++ b/packaging/windows/product.wxs.template
@@ -38,6 +38,12 @@
                 Name="Path" />
         </Property>
 
+        <Property Id="USERFOLDER">
+            <DirectorySearch Id="userProfileSearch" Depth="0" Path="[%USERPROFILE]" />
+        </Property>
+
+        <Property Id="SHAREDDIRNAME" Secure="yes">crc-dir0</Property>
+
         <util:Group Id="CrcUsersGroup" Name="crc-users" />
         <Directory Id="TARGETDIR" Name="SourceDir">
             <Directory Id="ProgramFiles64Folder">
@@ -101,6 +107,12 @@
             Before="AddUserToHypervAdminGroup" 
             Sequence="execute"/>
         <CustomAction Id="AddUserToHypervAdminGroup" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
+        <SetProperty Action="CACreateSMBShare"
+            Id="CreateSMBShare"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;New-SmbShare -Name '[SHAREDDIRNAME]' -Path '[USERFOLDER]' -FullAccess '[LogonUser]'&quot;"
+            Before="CreateSMBShare"
+            Sequence="execute"/>
+        <CustomAction Id="CreateSMBShare" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="check" />
         <SetProperty Action="CAEnableFileAndPrinterSharing"
             Id="EnableFileAndPrinterSharing"
             Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Set-NetFirewallRule -DisplayGroup 'File And Printer Sharing' -Enabled True -Profile 'Private,Public'&quot;"
@@ -118,6 +130,7 @@
             <Custom Action="RemoveCrcGroupRollback" After="CreateCrcGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <Custom Action="RemoveCrcDaemonTask" Before='RemoveFiles'>Installed AND NOT UPGRADINGPRODUCTCODE</Custom>
             <Custom Action="EnableFileAndPrinterSharing" After="AddUserToHypervAdminGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
+            <Custom Action="CreateSMBShare" After="EnableFileAndPrinterSharing"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <ScheduleReboot After="InstallFinalize"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</ScheduleReboot>
         </InstallExecuteSequence>
         <Feature Id="DefaultFeature" Level="1">
@@ -138,6 +151,7 @@
             <ProgressText Action="InstallHyperv">Installing Hyper-V</ProgressText>
             <ProgressText Action="AddUserToHypervAdminGroup">Adding user: [LogonUser] to Hyper-V Administrators group</ProgressText>
             <ProgressText Action="RemoveCrcDaemonTask">Removing crcDaemon task</ProgressText>
+            <ProgressText Action="CreateSMBShare">Creating share named: [SHAREDDIRNAME] for folder: [USERFOLDER]</ProgressText>
             <ProgressText Action="EnableFileAndPrinterSharing">Enabling file and printer Sharing</ProgressText>
         </UI>
         <Property Id="WIXUI_INSTALLDIR" Value="INSTALLDIR" />
--- a/packaging/windows/product.wxs.template
+++ b/packaging/windows/product.wxs.template
@@ -78,10 +78,10 @@
             </Directory>
             <Directory Id="ProgramMenuFolder">
                 <Component Id="StartMenuEntry" Guid="*">
-                    <Shortcut Id="TrayStartMenuEntry" 
-                            Name="Red Hat OpenShift Local" 
-                            Target="[INSTALLDIR]crc-tray.exe" 
-                            Icon="crcicon.ico" 
+                    <Shortcut Id="TrayStartMenuEntry"
+                            Name="Red Hat OpenShift Local"
+                            Target="[INSTALLDIR]crc-tray.exe"
+                            Icon="crcicon.ico"
                             WorkingDirectory="AppDataFolder">
                         <ShortcutProperty Key="System.AppUserModel.ID" Value="redhat.codereadycontainers.tray"/>
                     </Shortcut>
@@ -101,10 +101,10 @@
         <CustomAction Id="RemoveCrcDaemonTask" BinaryKey="WixCA" DllEntry="WixQuietExec" Execute="deferred" Impersonate="no" Return="ignore" />
         <SetProperty Action="CAInstallHyperv" Id="InstallHyperv" Value='"[System64Folder]dism.exe" /online /enable-feature /featureName:microsoft-hyper-v-all /NoRestart /quiet' Before="InstallHyperv" Sequence="execute"/>
         <CustomAction Id="InstallHyperv" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
-        <SetProperty Action="CAAddUserToHypervAdminGroup" 
-            Id="AddUserToHypervAdminGroup" 
-            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Add-LocalGroupMember -Member [LogonUser] -SID S-1-5-32-578&quot;" 
-            Before="AddUserToHypervAdminGroup" 
+        <SetProperty Action="CAAddUserToHypervAdminGroup"
+            Id="AddUserToHypervAdminGroup"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Add-LocalGroupMember -Member [LogonUser] -SID S-1-5-32-578&quot;"
+            Before="AddUserToHypervAdminGroup"
             Sequence="execute"/>
         <CustomAction Id="AddUserToHypervAdminGroup" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
         <SetProperty Action="CACreateSMBShare"
--- a/packaging/windows/product.wxs.template
+++ b/packaging/windows/product.wxs.template
@@ -101,6 +101,12 @@
             Before="AddUserToHypervAdminGroup" 
             Sequence="execute"/>
         <CustomAction Id="AddUserToHypervAdminGroup" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
+        <SetProperty Action="CAEnableFileAndPrinterSharing"
+            Id="EnableFileAndPrinterSharing"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Set-NetFirewallRule -DisplayGroup 'File And Printer Sharing' -Enabled True -Profile 'Private,Public'&quot;"
+            Before="EnableFileAndPrinterSharing"
+            Sequence="execute"/>
+        <CustomAction Id="EnableFileAndPrinterSharing" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="check" />
 
         <util:CloseApplication Id = "TrayRunning" Description="Please exit Red Hat OpenShift Local from tray and run the installation again." Target="crc-tray.exe" RebootPrompt="no" PromptToContinue="yes" />
 
@@ -111,6 +117,7 @@
             <Custom Action="InstallHyperv" Before="CreateCrcGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <Custom Action="RemoveCrcGroupRollback" After="CreateCrcGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <Custom Action="RemoveCrcDaemonTask" Before='RemoveFiles'>Installed AND NOT UPGRADINGPRODUCTCODE</Custom>
+            <Custom Action="EnableFileAndPrinterSharing" After="AddUserToHypervAdminGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <ScheduleReboot After="InstallFinalize"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</ScheduleReboot>
         </InstallExecuteSequence>
         <Feature Id="DefaultFeature" Level="1">
@@ -131,6 +138,7 @@
             <ProgressText Action="InstallHyperv">Installing Hyper-V</ProgressText>
             <ProgressText Action="AddUserToHypervAdminGroup">Adding user: [LogonUser] to Hyper-V Administrators group</ProgressText>
             <ProgressText Action="RemoveCrcDaemonTask">Removing crcDaemon task</ProgressText>
+            <ProgressText Action="EnableFileAndPrinterSharing">Enabling file and printer Sharing</ProgressText>
         </UI>
         <Property Id="WIXUI_INSTALLDIR" Value="INSTALLDIR" />
         <!-- this should help to propagate env var changes -->
--- a/pkg/crc/config/callbacks.go
+++ b/pkg/crc/config/callbacks.go
@@ -33,6 +33,11 @@ func RequiresCRCSetup(key string, _ interface{}) string {
 		"Please run 'crc setup' for this configuration to take effect.", key)
 }
 
+func RequiresCleanupAndSetupMsg(key string, _ interface{}) string {
+	return fmt.Sprintf("Changes to configuration property '%s' are only applied during 'crc setup'.\n"+
+		"Please run 'crc cleanup' followed by 'crc setup' for this configuration to take effect.", key)
+}
+
 func RequiresHTTPPortChangeWarning(key string, value interface{}) string {
 	return fmt.Sprintf("Changes to configuration property '%s' will break OpenShift HTTP routes.\n"+
 		"In order to access OpenShift applications through HTTP URLs "+
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -30,6 +30,7 @@ const (
 	KubeAdminPassword       = "kubeadmin-password"
 	Preset                  = "preset"
 	EnableSharedDirs        = "enable-shared-dirs"
+	SharedDirPassword       = "shared-dir-password" // #nosec G101
 	IngressHTTPPort         = "ingress-http-port"
 	IngressHTTPSPort        = "ingress-https-port"
 )
@@ -56,6 +57,17 @@ func RegisterSettings(cfg *Config) {
 		return ValidateBundlePath(value, GetPreset(cfg))
 	}
 
+	validateSmbSharedDirs := func(value interface{}) (bool, string) {
+		if !cfg.Get(HostNetworkAccess).AsBool() {
+			return false, fmt.Sprintf("%s can only be used with %s set to 'true'",
+				EnableSharedDirs, HostNetworkAccess)
+		}
+		if cfg.Get(SharedDirPassword).IsDefault {
+			return false, fmt.Sprintf("Please set '%s' first to enable shared directories", SharedDirPassword)
+		}
+		return ValidateBool(value)
+	}
+
 	// Preset setting should be on top because CPUs/Memory config depend on it.
 	cfg.AddSetting(Preset, string(preset.OpenShift), validatePreset, RequiresDeleteAndSetupMsg,
 		fmt.Sprintf("Virtual machine preset (valid values are: %s, %s and %s)", preset.Podman, preset.OpenShift, preset.OKD))
@@ -77,10 +89,17 @@ func RegisterSettings(cfg *Config) {
 		"Disable update check (true/false, default: false)")
 	cfg.AddSetting(ExperimentalFeatures, false, ValidateBool, SuccessfullyApplied,
 		"Enable experimental features (true/false, default: false)")
-	// shared dir is not implemented for windows yet
-	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+
+	// Shared directories configs
+	if runtime.GOOS == "windows" {
+		cfg.AddSetting(SharedDirPassword, Secret(""), ValidateString, SuccessfullyApplied,
+			"Password used while using CIFS/SMB file sharing (It is the password for the current logged in user)")
+
+		cfg.AddSetting(EnableSharedDirs, false, validateSmbSharedDirs, SuccessfullyApplied,
+			"Mounts host's user profile folder at '/' in the CRC VM (true/false, default: false)")
+	} else {
 		cfg.AddSetting(EnableSharedDirs, true, ValidateBool, SuccessfullyApplied,
-			"Enable shared directories which mounts host's $HOME at /home in the CRC VM (true/false, default: true)")
+			"Mounts host's home directory at '/' in the CRC VM (true/false, default: true)")
 	}
 
 	if !version.IsInstaller() {
@@ -143,14 +162,6 @@ func GetNetworkMode(config Storage) network.Mode {
 		return network.UserNetworkingMode
 	}
 	return network.ParseMode(config.Get(NetworkMode).AsString())
-}
-
-func ShouldEnableSharedDirs(config Storage) bool {
-	// Shared dirs are not implemented for windows
-	if runtime.GOOS == "windows" {
-		return false
-	}
-	return config.Get(EnableSharedDirs).AsBool()
 }
 
 func UpdateDefaults(cfg *Config) {
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -107,7 +107,7 @@ func RegisterSettings(cfg *Config) {
 			fmt.Sprintf("Network mode (%s or %s)", network.UserNetworkingMode, network.SystemNetworkingMode))
 	}
 
-	cfg.AddSetting(HostNetworkAccess, false, validateHostNetworkAccess, SuccessfullyApplied,
+	cfg.AddSetting(HostNetworkAccess, false, validateHostNetworkAccess, RequiresCleanupAndSetupMsg,
 		"Allow TCP/IP connections from the CRC VM to services running on the host (true/false, default: false)")
 	// Proxy Configuration
 	cfg.AddSetting(HTTPProxy, "", ValidateHTTPProxy, SuccessfullyApplied,
--- a/pkg/crc/errors/maskederror.go
+++ b/pkg/crc/errors/maskederror.go
@@ -1,0 +1,16 @@
+package errors
+
+import (
+	"strings"
+)
+
+const mask = "*****"
+
+type MaskedSecretError struct {
+	Err    error
+	Secret string
+}
+
+func (err *MaskedSecretError) Error() string {
+	return strings.ReplaceAll(err.Err.Error(), err.Secret, mask)
+}
--- a/pkg/crc/machine/config/config.go
+++ b/pkg/crc/machine/config/config.go
@@ -7,15 +7,17 @@ type MachineConfig struct {
 	BundleName string
 
 	// Virtual machine configuration
-	Name            string
-	Memory          int
-	CPUs            int
-	DiskSize        int
-	ImageSourcePath string
-	ImageFormat     string
-	SSHKeyPath      string
-	KubeConfig      string
-	SharedDirs      []string
+	Name              string
+	Memory            int
+	CPUs              int
+	DiskSize          int
+	ImageSourcePath   string
+	ImageFormat       string
+	SSHKeyPath        string
+	KubeConfig        string
+	SharedDirs        []string
+	SharedDirPassword string
+	SharedDirUsername string
 
 	// macOS specific configuration
 	KernelCmdLine string
--- a/pkg/crc/machine/driver.go
+++ b/pkg/crc/machine/driver.go
@@ -57,3 +57,19 @@ func setDiskSize(host *host.Host, diskSizeGiB int) error {
 
 	return updateDriverValue(host, diskSizeSetter)
 }
+
+func setSharedDirPassword(host *host.Host, password string) error {
+	driver, err := loadDriverConfig(host)
+	if err != nil {
+		return err
+	}
+
+	if len(driver.SharedDirs) == 0 {
+		return nil
+	}
+
+	for i := range driver.SharedDirs {
+		driver.SharedDirs[i].Password = password
+	}
+	return updateDriverStruct(host, driver)
+}
--- a/pkg/crc/machine/driver_linux.go
+++ b/pkg/crc/machine/driver_linux.go
@@ -10,6 +10,7 @@ import (
 	"github.com/code-ready/crc/pkg/libmachine"
 	"github.com/code-ready/crc/pkg/libmachine/host"
 	machineLibvirt "github.com/code-ready/machine/drivers/libvirt"
+	"github.com/code-ready/machine/libmachine/drivers"
 )
 
 func newHost(api libmachine.API, machineConfig config.MachineConfig) (*host.Host, error) {
@@ -47,3 +48,7 @@ func (r *RPCServerDriver) SetConfigRaw(data []byte, _ *struct{}) error {
 	return json.Unmarshal(data, &r.ActualDriver)
 }
 */
+
+func updateDriverStruct(host *host.Host, driver *machineLibvirt.Driver) error {
+	return drivers.ErrNotImplemented
+}
--- a/pkg/crc/machine/hyperv/driver_windows.go
+++ b/pkg/crc/machine/hyperv/driver_windows.go
@@ -1,11 +1,15 @@
 package hyperv
 
 import (
+	"path/filepath"
+	"strings"
+
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/machine/config"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/crc/pkg/drivers/hyperv"
 	winnet "github.com/code-ready/crc/pkg/os/windows/network"
+	"github.com/code-ready/machine/libmachine/drivers"
 )
 
 func CreateHost(machineConfig config.MachineConfig) *hyperv.Driver {
@@ -23,5 +27,35 @@ func CreateHost(machineConfig config.MachineConfig) *hyperv.Driver {
 		hypervDriver.VirtualSwitch = switchName
 	}
 
+	hypervDriver.SharedDirs = configureShareDirs(machineConfig)
 	return hypervDriver
+}
+
+// converts a path like c:\users\crc to /mnt/c/users/crc
+func convertToUnixPath(path string) string {
+	/* podman internally converts windows style paths like C:\Users\crc  to
+	 * /mnt/c/Users/crc so it expects the shared folder to be mounted under
+	 * '/mnt' instead of '/' like in the case of macOS and linux
+	 * see: https://github.com/containers/podman/blob/468aa6478c73e4acd8708ce8bb0bb5a056f329c2/pkg/specgen/winpath.go#L24-L59
+	 */
+	path = filepath.ToSlash(path)
+	if len(path) > 1 && path[1] == ':' {
+		return ("/mnt/" + strings.ToLower(path[0:1]) + path[2:])
+	}
+	return path
+}
+
+func configureShareDirs(machineConfig config.MachineConfig) []drivers.SharedDir {
+	var sharedDirs []drivers.SharedDir
+	for _, dir := range machineConfig.SharedDirs {
+		sharedDir := drivers.SharedDir{
+			Source:   dir,
+			Target:   convertToUnixPath(dir),
+			Tag:      "crc-dir0", // smb share 'crc-dir0' is created in the msi
+			Type:     "cifs",
+			Username: machineConfig.SharedDirUsername,
+		}
+		sharedDirs = append(sharedDirs, sharedDir)
+	}
+	return sharedDirs
 }
--- a/pkg/crc/machine/libvirt/driver_linux.go
+++ b/pkg/crc/machine/libvirt/driver_linux.go
@@ -1,10 +1,13 @@
 package libvirt
 
 import (
+	"fmt"
+
 	"github.com/code-ready/crc/pkg/crc/constants"
 	"github.com/code-ready/crc/pkg/crc/machine/config"
 	"github.com/code-ready/crc/pkg/crc/network"
 	"github.com/code-ready/machine/drivers/libvirt"
+	"github.com/code-ready/machine/libmachine/drivers"
 )
 
 func CreateHost(machineConfig config.MachineConfig) *libvirt.Driver {
@@ -20,5 +23,21 @@ func CreateHost(machineConfig config.MachineConfig) *libvirt.Driver {
 	}
 
 	libvirtDriver.StoragePool = DefaultStoragePool
+	libvirtDriver.SharedDirs = configureShareDirs(machineConfig)
+
 	return libvirtDriver
+}
+
+func configureShareDirs(machineConfig config.MachineConfig) []drivers.SharedDir {
+	var sharedDirs []drivers.SharedDir
+	for i, dir := range machineConfig.SharedDirs {
+		sharedDir := drivers.SharedDir{
+			Source: dir,
+			Target: dir,
+			Tag:    fmt.Sprintf("dir%d", i),
+			Type:   "virtiofs",
+		}
+		sharedDirs = append(sharedDirs, sharedDir)
+	}
+	return sharedDirs
 }
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -99,6 +99,14 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
 		}
 	}
 
+	// we want to set the shared dir password on-the-fly to be used
+	// we do not want this value to be persisted to disk
+	if startConfig.SharedDirPassword != "" {
+		if err := setSharedDirPassword(vm.Host, startConfig.SharedDirPassword); err != nil {
+			return fmt.Errorf("Failed to set shared dir password: %w", err)
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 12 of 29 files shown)


packaging/windows/scripts/passwordless_share_disable.ps1
```diff
@@ -0,0 +1,50 @@
...
+        Start-ScheduledTask -TaskName "$taskname"
+        Sleep 5
+        Unregister-ScheduledTask -TaskName $taskname -Confirm:$false
+    }
+    catch
+    {
+        Write-Host "Exception $_"
+        Get-Variable -Scope Script
+    }
+}
```
praveenkumar, line 50: Not reviewed and tested but I am wondering, can we make this as part of `crc setup` and try to avoid put more stuff to installer side?

anjannath: we can put them in `crc setup` but these commands need admin privileges to run so to avoid the UAC prompts its added to the msi

anjannath:
still need to test on a clean vm if the msi is able to set everything up, and also verify the assumption that `Private` network profile has "File and Printer Sharing" enabled by default on window 10

anjannath:
@gbraad PTAL

pkg/crc/config/settings.go
```diff
@@ -75,8 +76,21 @@ func RegisterSettings(cfg *Config) {
...
+			if cfg.Get(SMBSharedDirPassword).IsDefault {
+				return false, fmt.Sprintf("Please set '%s' first to enable shared directories", SMBSharedDirPassword)
+			}
+			return true, ""
+		}
+		cfg.AddSetting(SMBSharedDirPassword, "", ValidateString, SuccessfullyApplied,
+			"Password used while using CIFS/SMB file sharing (Its the password for the current logon user)")
+
+		cfg.AddSetting(EnableSharedDirs, false, validateSmbPasswordSet, SuccessfullyApplied,
+			"Enable shared directories which mounts host's $HOME at /home in the CRC VM (true/false, default: true)")
```
cfergeau, line 92: "default: false".

pkg/crc/machine/types/types.go
```diff
@@ -30,7 +30,8 @@ type StartConfig struct {
 	Preset crcpreset.Preset
 
 	// Shared dirs
-	EnableSharedDirs bool
+	EnableSharedDirs     bool
+	SMBSharedDirPassword string
```
cfergeau, line 34: Can the password be set per directory share? Or is it always going to be the same? It's odd to have password there, and to guess username later on.

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -373,3 +376,34 @@ func (d *Driver) GetIP() (string, error) {
...
+		dir.Type = "cifs"
+		dir.Tag = smbShareUNCPathFromTag(dir.Tag)
+
+		dirs = append(dirs, dir)
+	}
+	return dirs, nil
+}
+
+func convertToUnixPath(path string) string {
+	// converts a path like c:\users\crc to /mnt/c/users/crc
```
cfergeau, line 397: A comment explaining why we do this would be very useful, as it's non obvious why we don't convert this to /users/crc as for macOS. This comment could link to https://github.com/containers/podman/blob/67c4068bb35fd1aad95b1701c94ed11183d0fd66/pkg/specgen/winpath.go#L23-L41

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -373,3 +376,34 @@ func (d *Driver) GetIP() (string, error) {
...
+	return fmt.Sprintf("/mnt/c%s", strings.TrimLeftFunc(filepath.ToSlash(path), func(r rune) bool {
+		return (r == ':' || unicode.IsLetter(r))
+	}))
+}
+
+func smbShareUNCPathFromTag(tag string) string {
+	hostname, err := cmdOut("hostname")
+	if err == nil {
+		return fmt.Sprintf("//%s.mshome.net/%s", strings.TrimSpace(hostname), tag)
```
cfergeau, line 406: Wondering if we could use an IP there instead of hostname.mshome.net? Not sure it's better, maybe it would make the code simpler?

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -373,3 +376,34 @@ func (d *Driver) GetIP() (string, error) {
 
 	return resp[0], nil
 }
+
+func (d *Driver) GetSharedDirs() ([]drivers.SharedDir, error) {
+	dirs := []drivers.SharedDir{}
+	for _, dir := range d.SharedDirs {
+		if !smbShareExists(dir.Tag) {
+			return []drivers.SharedDir{}, drivers.ErrNotSupported
```
cfergeau, line 384: Why "ErrNotSupported"? Could just return an empty list of shares if we don't have the share we expect?

packaging/windows/product.wxs.template
```diff
@@ -107,6 +107,18 @@
             Before="EnableSMBFeature"
             Sequence="execute"/>
         <CustomAction Id="EnableSMBFeature" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
+        <SetProperty Action="CAEnableFileAndPrinterSharing"
```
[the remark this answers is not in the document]
anjannath: i think its needed to enable for both, because if we don't enable for both then we have a pre-condition that users need to be using the profile for which we enabled this to use shared dirs which i wanted avoid, 

the way the profiles are set is If user's don't select a profile (when first time connecting to a new network) then windows will set the "Public" profile

pkg/crc/config/settings.go
```diff
@@ -75,8 +76,21 @@ func RegisterSettings(cfg *Config) {
...
+	// Shared directories configs
+	if runtime.GOOS == "windows" {
+		validateSmbPasswordSet := func(value interface{}) (bool, string) {
+			if cfg.Get(SMBSharedDirPassword).IsDefault {
+				return false, fmt.Sprintf("Please set '%s' first to enable shared directories", SMBSharedDirPassword)
+			}
+			return true, ""
+		}
+		cfg.AddSetting(SMBSharedDirPassword, "", ValidateString, SuccessfullyApplied,
+			"Password used while using CIFS/SMB file sharing (Its the password for the current logon user)")
```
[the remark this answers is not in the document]
anjannath: i guess it'll just fail authentication, because there's another setting for enabling password less share

pkg/crc/machine/types/types.go
anjannath: its the same password always, when the shares are created access controls are based on windows users and groups 

one way to do this would be to create a new user and only allow that user to access, but then it appears in the login, there's some way to hide the user from the login screen by editing registry keys, but seem too hacky and haven't tried it yet https://www.windowscentral.com/how-hide-specific-user-accounts-sign-screen-windows-10

pkg/drivers/hyperv/hyperv_windows.go
anjannath: we have to figure out the ip address on one of the interfaces of the host or the "Default Switch" the default switch name might be localized, but this gives the ip address for default switch `(Get-NetIPAddress -InterfaceAlias "vEthernet (Default Switch)").IPv4Address`

pkg/drivers/hyperv/hyperv_windows.go
anjannath: Thanks, i did not look at this code, only noticed the behaviour that it appends the `/mnt`

packaging/windows/product.wxs.template
```diff
@@ -101,6 +101,12 @@
             Before="AddUserToHypervAdminGroup" 
             Sequence="execute"/>
         <CustomAction Id="AddUserToHypervAdminGroup" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
+        <SetProperty Action="CAEnableSMBFeature"
+            Id="EnableSMBFeature"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Enable-WindowsOptionalFeature -FeatureName 'SMB1Protocol' -Online -All -NoRestart&quot;"
```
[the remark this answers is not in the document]
anjannath: actually checking again, it seems we don't even need to enable any optional features, smbv3 is enabled on windows by default so we can remove this custom actions enabling smbv1 :)

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -373,3 +376,34 @@ func (d *Driver) GetIP() (string, error) {
...
+		}
+
+		dir.Target = convertToUnixPath(dir.Target)
+		dir.Type = "cifs"
+		dir.Tag = smbShareUNCPathFromTag(dir.Tag)
+
```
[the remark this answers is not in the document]
anjannath: removed the existing `SharedDirs` struct setup from `pkg/crc/machine/config/driver.go`  and  added to `vfkit/driver_darwin.go` and `libvirt/driver_linux.go`  its the same code duplicated  https://github.com/code-ready/crc/pull/3308/commits/cba3a6cef2c5072506f3bf7e978f8694cecac441

in the next commit followed the same pattern and configured shared dirs for hyperv in `hyperv/driver_windows.go`

pkg/crc/machine/types/types.go
anjannath: >  It's odd to have password there, and to guess username later on.

yeah, the username should also be part of `StartConfig` tthen, but would prefer not to expose it as a config for the user to set (less things to configure is nicer imo)

pkg/crc/machine/start.go
```diff
@@ -169,10 +170,21 @@ func configureSharedDirs(vm *virtualMachine, sshRunner *crcssh.Runner) error {
...
-		mount.Type = "virtiofs"
-		if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Mounting %s", mount.Target), "mount", "-o", "context=\"system_u:object_r:container_file_t:s0\"", "-t", mount.Type, mount.Tag, mount.Target); err != nil {
-			return err
+		switch mount.Type {
+		case "virtiofs":
+			if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Mounting %s", mount.Target), "mount", "-o", "context=\"system_u:object_r:container_file_t:s0\"", "-t", mount.Type, mount.Tag, mount.Target); err != nil {
+				return err
+			}
+		case "cifs":
+			smbUser, err := user.Current()
```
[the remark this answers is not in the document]
anjannath: @cfergeau will it be acceptable to just store it base64 encoded? its rare we ask users to share the machine config!

or to have it stored properly encrypted, we could create a secret (long living) and store it in the OS password store and use that to encrypt the password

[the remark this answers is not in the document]
anjannath: another approach could be to not persist the `Password` filed to disk and populate it from config after reading and Unmarshaling the json, but we'll have to call into crc cmd level configs from the libmachine package

packaging/windows/product.wxs.template
```diff
@@ -107,6 +111,18 @@
             Before="EnableSMBFeature"
             Sequence="execute"/>
         <CustomAction Id="EnableSMBFeature" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
+        <SetProperty Action="CACreateSMBShare"
+            Id="CreateSMBShare"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;New-SmbShare -Name 'dir0' -Path '[USERFOLDER]' -FullAccess 'Everyone'&quot;"
```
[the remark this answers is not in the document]
anjannath: changed to `crc-dir0`

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -373,3 +376,34 @@ func (d *Driver) GetIP() (string, error) {
...
+		dirs = append(dirs, dir)
+	}
+	return dirs, nil
+}
+
+func convertToUnixPath(path string) string {
+	// converts a path like c:\users\crc to /mnt/c/users/crc
+	return fmt.Sprintf("/mnt/c%s", strings.TrimLeftFunc(filepath.ToSlash(path), func(r rune) bool {
```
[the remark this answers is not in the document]
anjannath: There's a 9MB increase in the output binary which is not a lot, but the amount of code that is pulled in is huge! thinking to just c&p the relevent part from that function

pkg/crc/config/settings.go
```diff
@@ -31,6 +31,7 @@ const (
 	KubeAdminPassword       = "kubeadmin-password"
 	Preset                  = "preset"
 	EnableSharedDirs        = "enable-shared-dirs"
+	SharedDirPassword       = "shared-dir-password" // #nosec G101
```
anjannath, line 34: @gbraad Changed this to `shared-dir-password` instead of `smb-share-password` more generic, can be reused when we do the sshfs

packaging/windows/product.wxs.template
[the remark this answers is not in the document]
gbraad: which is a security feature. you don't wanna bypass this

packaging/windows/product.wxs.template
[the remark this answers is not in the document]
gbraad: smbv1 was disabled for reasons...

Users share should always be available IIRC. This is what we used with minishift

pkg/drivers/hyperv/hyperv_windows.go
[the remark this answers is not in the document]
gbraad: I'd rather see us re-use code. c&p is not just unethical, we are supposed to share and combine most of the codebase over time anyway.

pkg/drivers/hyperv/hyperv_windows.go
gbraad: In minishift we extracted the IP address, verified reachability before allowing it to be used.

gbraad:
Just based on some of the comments from Christophe I say there is still some open issues; code re-use, IP address for use, etc.

packaging/windows/product.wxs.template
```diff
@@ -111,6 +141,10 @@
             <Custom Action="InstallHyperv" Before="CreateCrcGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <Custom Action="RemoveCrcGroupRollback" After="CreateCrcGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <Custom Action="RemoveCrcDaemonTask" Before='RemoveFiles'>Installed AND NOT UPGRADINGPRODUCTCODE</Custom>
+            <Custom Action="EnableFileAndPrinterSharing" After="AddUserToHypervAdminGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
+            <Custom Action="CreateSMBShare" After="EnableFileAndPrinterSharing"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
+            <Custom Action="RemoveSMBShare" Before="InstallFinalize"> Installed AND NOT PATCH AND REMOVE~="ALL" AND NOT UPGRADINGPRODUCTCODE</Custom>
+            <Custom Action="DisableFileAndPrinterSharing" Before="RemoveFiles"> Installed AND NOT PATCH AND REMOVE~="ALL" AND NOT UPGRADINGPRODUCTCODE</Custom>
```
gbraad, line 147: you can't assume 'we' need to disable this feature.

packaging/windows/product.wxs.template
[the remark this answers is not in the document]
anjannath: Its not a bypass, its just that I want the shared folders to be accessible in both "Private" and "Public" network profiles as we cannot predict what profile will be active on the user's machine

also the share is created giving access to only the login/current user and needs a password to access

pkg/crc/config/settings.go
```diff
@@ -75,8 +76,21 @@ func RegisterSettings(cfg *Config) {
 		"Disable update check (true/false, default: false)")
 	cfg.AddSetting(ExperimentalFeatures, false, ValidateBool, SuccessfullyApplied,
 		"Enable experimental features (true/false, default: false)")
-	// shared dir is not implemented for windows yet
-	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+
+	// Shared directories configs
+	if runtime.GOOS == "windows" {
+		validateSmbPasswordSet := func(value interface{}) (bool, string) {
```
gbraad, line 82: Why is this not a regular named function inside a `_windows.go` file?

pkg/crc/machine/config/config.go
```diff
@@ -7,15 +7,17 @@ type MachineConfig struct {
...
+	Name              string
+	Memory            int
+	CPUs              int
+	DiskSize          int
+	ImageSourcePath   string
+	ImageFormat       string
+	SSHKeyPath        string
+	KubeConfig        string
+	SharedDirs        []string
+	SharedDirPassword string
```
gbraad, line 19: so when we deal with sshfs, you are only able to use a single password for each of them?

pkg/crc/machine/config/config.go
```diff
@@ -7,15 +7,17 @@ type MachineConfig struct {
...
+	SharedDirUsername string
```
gbraad, line 20: ditto for the username?

pkg/crc/machine/driver.go
```diff
@@ -57,3 +57,22 @@ func setDiskSize(host *host.Host, diskSizeGiB int) error {
...
+		return err
+	}
+
+	if len(driver.SharedDirs) == 0 {
+		return nil
+	}
+
+	sharedDir := []libmachine.SharedDir{}
+	for _, dir := range driver.SharedDirs {
+		dir.Password = password
```
gbraad, line 73: so we assume all shares share the same password.

pkg/crc/config/settings.go
anjannath: Its following the same pattern that's in the config package.. and we can't use a `_windows.go` in there without a lot of refactor because how we are registering the settings

pkg/crc/machine/hyperv/driver_windows.go
```diff
@@ -23,5 +29,32 @@ func CreateHost(machineConfig config.MachineConfig) *hyperv.Driver {
 		hypervDriver.VirtualSwitch = switchName
 	}
 
+	for _, dir := range machineConfig.SharedDirs {
+		sharedDir := drivers.SharedDir{
+			Source:   dir,
+			Target:   convertToUnixPath(dir),
+			Tag:      smbShareUNCPathFromTag("crc-dir0"), // smb share 'crc-dir0' is created in the msi
+			Type:     "cifs",
+			Username: machineConfig.SharedDirUsername,
```
gbraad, line 38: so we associate the username at this point, but not the password? Why do we need to do this in the previous part in an update function?

anjannath: because we don't want to store the password in the machine config file, so that's handled by the update function

pkg/crc/machine/hyperv/driver_windows.go
```diff
@@ -23,5 +29,32 @@ func CreateHost(machineConfig config.MachineConfig) *hyperv.Driver {
...
+	if len(path) > 1 && path[1] == ':' {
+		return ("/mnt/" + strings.ToLower(path[0:1]) + path[2:])
+	}
+	return path
+}
+
+func smbShareUNCPathFromTag(tag string) string {
+	hostname, _, err := powershell.Execute("hostname")
+	if err == nil {
+		return fmt.Sprintf("//%s.mshome.net/%s", strings.TrimSpace(hostname), tag)
```
gbraad, line 57: This is an assumption that will only work for the default switch if not mistaken.
better to use the IP address or actual hostname for this. This can be extracted but is tedious

pkg/crc/machine/driver.go
anjannath: yes, for windows its based on the user account, so shares themselves don't have passwords i am not considering about sshfs here, when we implement that we might have to change this

pkg/crc/machine/hyperv/driver_windows.go
anjannath: well, the IP address also we have to use from the `Default Switch` no? or you mean enumerate all the interfaces and pick an ip address from there?

packaging/windows/product.wxs.template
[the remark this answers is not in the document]
gbraad: It is a bypass as you do not need 'public' access, as that is based on what the user has chosen.
If the users clicks this away, that is a 'user error'.

This is why you need to check for reachability of the share before using.

Password or not, that is a different problem

pkg/crc/machine/config/config.go
anjannath: similar to https://github.com/code-ready/crc/pull/3308#discussion_r957988343 i was not thinking about `sshfs` internally that much, was only thinking about the user facing config options as of now, when implementing `sshfs` we might have to change this

packaging/windows/product.wxs.template
anjannath: yeah, right, we don't disable other features we enable, will remove this, thanks!

pkg/crc/config/settings.go
gbraad: so it is impossible to create a file named `settings_windows.go` and place this inside as a named function ?

pkg/crc/machine/hyperv/driver_windows.go
gbraad: If you are concerned about the password, this is an association as a credential (username + password). It would make more sense to place the username in the update function too.

pkg/crc/machine/hyperv/driver_windows.go
anjannath: btw, this hostname.mshome.net also works with the user-mode networking, i  think in that case the VM doesn't any ip assigned or connection with the default switch

gbraad: We do not bind the default switch to the VM, as we use the usermode networking.
We can not actually assume the default switch is there and enabled. Our preflight is erroneous, as we do not need it anymore... and does not check if enabled or even reachable.

pkg/drivers/hyperv/hyperv_windows.go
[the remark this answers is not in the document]
anjannath: Okay, apart from the extra code in vendor, we don't need the full functionality of that Function in https://github.com/containers/podman/blob/67c4068bb35fd1aad95b1701c94ed11183d0fd66/pkg/specgen/winpath.go#L23-L41

i still think just c&p the relevant line is better then vendoring the whole podman codebase!

pkg/crc/machine/hyperv/driver_windows.go
anjannath: Its not as much about leaking information but storing a secret on disk unencrypted which we want to prevent, on the user's machine any non-privilege process can just run `whoami` and get that username, so it should be fine to store in the machine config?

btw this was suggested in https://github.com/code-ready/crc/pull/3308#discussion_r953798210

pkg/crc/machine/hyperv/driver_windows.go
gbraad: Isn't the IP not JUST the external IP of our usermode p2p ?

gbraad: `mshome.net` does not work for me. I do not have ICS enabled and DNS priorities wouldn't have allowed it anyway.

pkg/crc/machine/hyperv/driver_windows.go
gbraad: > any non-privilege process can just run whoami and get that username, 

People share the virtual machine folder as we do not offer an export function.

pkg/crc/machine/hyperv/driver_windows.go
anjannath: okay, in that case we need to switch to using IP address, thanks for testing out.

> Isn't the IP not JUST the external IP of our usermode p2p ?

no able to understand this, what is that external IP of usermode networking? when i have the daemon running, it doesn't create any interfaces

pkg/crc/machine/hyperv/driver_windows.go
anjannath: ahh, but that'd be a very special unsupported use case, assume most of our users don't do this, also we don't test it

pkg/crc/config/settings.go
anjannath: yeah looks like, if we do that we have to duplicate all of the common settings in `settings_windows.go`

packaging/windows/product.wxs.template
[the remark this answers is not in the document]
anjannath: > It is a bypass as you do not need 'public' access,

The network profiles in windows are about what type of network you are connected to (home aka private network vs cafe wifi aka public network) and not in the sense of if something is publicly accessible, based on the user's selection of the network profile some default settings are applied
 
Enabling it for both the profiles, lets the user have shared folders without worrying about what network profile is currently active, otherwise if we only enable it for one profile, "Private" network profile, as said we need the pre-condition "To use shared folders in windows users' need to set the Private profile as the active network profile"

> This is why you need to check for reachability of the share before using.
If we have properly setup the SMB share, there won't be any need to do this check, if we only enable sharing for "Private" network we can just check if the current active network profile is "Private"

pkg/drivers/hyperv/hyperv_windows.go
anjannath: changed to use an ip address instead of hostname.mshome.net for the UNC path

packaging/windows/product.wxs.template
anjannath: removed the custom action that was removing file sharing on uninstall

pkg/crc/machine/start.go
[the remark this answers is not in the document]
anjannath: @cfergeau Updated PTALA :)

pkg/crc/machine/hyperv/driver_windows.go
cfergeau: In usermode networking, I'd expect 127.0.0.1?

anjannath: `gvproxy` also has a "host-access" feature, where in the VM can reach the host using a dns name, by default we disabled it (can be enabled from config), I think that is exactly what we need

```
* host-network-access                     Allow TCP/IP connections from the CRC VM to services running on the host (true/false, default: false)
```

pkg/crc/machine/hyperv/driver_windows.go
cfergeau: It turns out to be different from what I suggested earlier, but maybe this should all be coming from the keyring then?

cmd/crc/cmd/start.go
```diff
@@ -93,6 +95,19 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
...
+	if runtime.GOOS == "windows" {
+		// config SharedDirPassword ('shared-dir-password') only exists in windows
+		startConfig.SharedDirPassword = config.Get(crcConfig.SharedDirPassword).AsString()
+		startConfig.SharedDirUsername = func() string {
+			u, err := user.Current()
+			if err != nil {
+				logging.Debugf("Error getting username needed for shared dirs: %v", err)
+				return ""
+			}
+			return strings.Split(u.Username, "\\")[1]
```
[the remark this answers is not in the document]
anjannath: the username is of the form `hostname\username`. i was hoping that it should be fine, because we can depend on std library that the o/p of `u.Username` will not change? but doesn't hurt to add a length check for the o/p of strings.Split

cmd/crc/cmd/start.go
```diff
@@ -93,6 +95,19 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		}
 	}
 
+	if runtime.GOOS == "windows" {
+		// config SharedDirPassword ('shared-dir-password') only exists in windows
+		startConfig.SharedDirPassword = config.Get(crcConfig.SharedDirPassword).AsString()
+		startConfig.SharedDirUsername = func() string {
```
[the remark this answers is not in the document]
anjannath: we need it only for windows, if we add a `getUser` helper we'd have to add this helper or stubs for unixes where we don't need it at the moment

pkg/crc/config/settings.go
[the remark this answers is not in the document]
anjannath: yeah its stored in plaintext as of now, need to work on improving the config for storing secrets ()

the pull secret ends up in the os keyring only when its entered by the user interactively during start, for this password we want it to be stored in the keyring via a `crc config set` command

pkg/crc/machine/hyperv/driver_windows.go
anjannath: only the password needs to be securely stored, username we can derive during start, we just skip writing it to machine config json same as the password

packaging/windows/product.wxs.template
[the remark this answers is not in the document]
gbraad: > based on the user's selection of the network profile some default settings are applied

And that is the point. I am not unfamiliar to these settings:

image

It is based on the user's selection, with the default being 'Public', which prevents access to these services, unless correctly set.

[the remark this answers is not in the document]
gbraad: > as said we need the pre-condition "To use shared folders in windows users need to set the Private profile as the active network profile"

Can you check for this?

pkg/crc/config/settings.go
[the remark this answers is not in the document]
gbraad: > yeah its stored in plaintext as of now,

No go until resolved

pkg/crc/config/settings.go
gbraad: password aren't shared among dirs, as they apply to each individual share.

pkg/crc/machine/hyperv/driver_windows.go
gbraad: host access might be able to expose port `445` IIRC, but that would make this available on the VM's localhost interface to be the SMB endpoint. Don't think this is what you want. Somehow you want to forward the SMB functionality to be resolvable.

pkg/crc/machine/start.go
```diff
@@ -169,10 +169,13 @@ func configureSharedDirs(vm *virtualMachine, sshRunner *crcssh.Runner) error {
...
-		mount.Type = "virtiofs"
-		if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Mounting %s", mount.Target), "mount", "-o", "context=\"system_u:object_r:container_file_t:s0\"", "-t", mount.Type, mount.Tag, mount.Target); err != nil {
-			return err
+		switch mount.Type {
+		case "virtiofs":
+			if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Mounting %s", mount.Target), "mount", "-o", "context=\"system_u:object_r:container_file_t:s0\"", "-t", mount.Type, mount.Tag, mount.Target); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("Unknown Share dir type requested: %s", mount.Type)
```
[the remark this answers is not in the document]
gbraad: agree. sounds like leaking the internal names to the outside. I don't think 'share dir' is even a great description for the user.

pkg/crc/config/settings.go
anjannath: ahh, but i couldn't find a way to set password per share in windows, either from the GUI or from the the smbshare powerhsell cmdlets

pkg/crc/config/settings.go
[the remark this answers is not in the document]
anjannath: working on #3333 to solve this

packaging/windows/product.wxs.template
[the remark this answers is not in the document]
anjannath: yes there's a cmd `Get-NetConnectionProfile` but it also needs an interface (seems profiles are per interface, so we need to first find the active interface)
https://docs.microsoft.com/en-us/powershell/module/netconnection/get-netconnectionprofile?view=windowsserver2022-ps#example-1-get-a-connection-profile

packaging/windows/product.wxs.template
```diff
@@ -95,12 +95,18 @@
...
         <CustomAction Id="InstallHyperv" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
-        <SetProperty Action="CAAddUserToHypervAdminGroup" 
-            Id="AddUserToHypervAdminGroup" 
-            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Add-LocalGroupMember -Member [LogonUser] -SID S-1-5-32-578&quot;" 
-            Before="AddUserToHypervAdminGroup" 
+        <SetProperty Action="CAAddUserToHypervAdminGroup"
+            Id="AddUserToHypervAdminGroup"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Add-LocalGroupMember -Member [LogonUser] -SID S-1-5-32-578&quot;"
+            Before="AddUserToHypervAdminGroup"
             Sequence="execute"/>
```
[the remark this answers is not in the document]
anjannath: done

pkg/crc/config/settings.go
anjannath: would prefer to do this refactor separate from this PR :) moved the function definition near the other similar functions, hopefully the code looks a bit better then before

cmd/crc/cmd/start.go
```diff
@@ -93,6 +95,19 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		}
 	}
 
+	if runtime.GOOS == "windows" {
+		// config SharedDirPassword ('shared-dir-password') only exists in windows
```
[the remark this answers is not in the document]
anjannath: done

packaging/windows/product.wxs.template
```diff
@@ -111,6 +117,9 @@
             <Custom Action="InstallHyperv" Before="CreateCrcGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <Custom Action="RemoveCrcGroupRollback" After="CreateCrcGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
             <Custom Action="RemoveCrcDaemonTask" Before='RemoveFiles'>Installed AND NOT UPGRADINGPRODUCTCODE</Custom>
+            <Custom Action="EnableFileAndPrinterSharing" After="AddUserToHypervAdminGroup"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
+            <Custom Action="CreateSMBShare" After="EnableFileAndPrinterSharing"> NOT Installed AND NOT REMOVE~="ALL" AND NOT WIX_UPGRADE_DETECTED</Custom>
+            <Custom Action="RemoveSMBShare" Before="InstallFinalize"> Installed AND NOT PATCH AND REMOVE~="ALL" AND NOT UPGRADINGPRODUCTCODE</Custom>
```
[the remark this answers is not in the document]
anjannath: next one, fixed it now thanks!

pkg/crc/machine/libvirt/driver_linux.go
```diff
@@ -20,5 +23,17 @@ func CreateHost(machineConfig config.MachineConfig) *libvirt.Driver {
 	}
 
 	libvirtDriver.StoragePool = DefaultStoragePool
+
+	// configure shared dirs
```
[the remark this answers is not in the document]
anjannath: done

pkg/crc/machine/start.go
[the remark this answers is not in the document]
anjannath: renamed to `Shared dir`

pkg/os/windows/network/vswitch_windows.go
```diff
@@ -45,3 +46,27 @@ func GetDefaultSwitchName() (bool, string) {
 
 	return true, strings.TrimSpace(stdOut)
 }
+
+// returns the IP from the first connected interface that has an IP address
+func GetActiveIP() (string, error) {
+	getInterfaceAliasCmd := `(Get-NetIPInterface -ConnectionState Connected -AddressFamily IPv4).InterfaceAlias`
```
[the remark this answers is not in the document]
anjannath: dropped this commit, we are using the `hostVirtualIP` now for the UNC path

cmd/crc/cmd/start.go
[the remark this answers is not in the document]
anjannath: added new helper `GetCurrentUsername`

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -373,3 +374,12 @@ func (d *Driver) GetIP() (string, error) {
 
 	return resp[0], nil
 }
+
+func (d *Driver) GetSharedDirs() ([]drivers.SharedDir, error) {
+	for _, dir := range d.SharedDirs {
+		if !smbShareExists(strings.Split(strings.TrimLeft(dir.Tag, "/"), "/")[1]) {
```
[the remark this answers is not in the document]
anjannath: updated, leaving the tag as `crc-dir0` and forming the UNC path at mount time

gbraad:
Unable to build:

```
$ make cross
GOARCH=arm64 GOOS=darwin go build -tags "containers_image_openpgp" -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=2.8.0 -X github.com/code-ready/crc/pkg/crc/version.oc
pVersion=4.11.1 -X github.com/code-ready/crc/pkg/crc/version.okdVersion=4.11.0-0.okd-2022-08-20-022919 -X github.com/code-ready/crc/pkg/crc/version.podmanVersion=4.1.1 -X github.com/code-ready/crc/pkg/crc/version.commitSha=758f4f3bb " -o out/macos-arm64/crc  ./cmd/crc                                                                                                        package github.com/code-ready/crc/cmd/crc
        imports github.com/code-ready/crc/cmd/crc/cmd
        imports github.com/containers/gvisor-tap-vsock/pkg/virtualnetwork
        imports github.com/containers/gvisor-tap-vsock/pkg/services/dhcp
        imports github.com/containers/gvisor-tap-vsock/pkg/tap
        imports gvisor.dev/gvisor/pkg/tcpip
        imports gvisor.dev/gvisor/pkg/atomicbitops
        imports gvisor.dev/gvisor/pkg/state
        imports gvisor.dev/gvisor/pkg/state/wire
        imports gvisor.dev/gvisor/pkg/gohacks: build constraints exclude all Go files in /workspaces/crc/vendor/gvisor.dev/gvisor/pkg/gohacks
make: *** [Makefile:94: out/macos-arm64/crc] Error 1
```

anjannath:
> on go version go1.19.1 linux/amd64 on Ubuntu it seems. Reverting to go version go1.18.3 linux/amd64 on Fedora

yes, i think `1.18` will also have issues, we are still on go `1.17` because of https://github.com/code-ready/crc/pull/3343#issuecomment-1245099348

> RHEL-8 still have 1.17.12 version of golang in official repo and also as part of go toolset images https://catalog.redhat.com/software/containers/ubi8/go-toolset/5ce8713aac3db925c03774d1 . Internally brew build might fails with those changes.

gbraad:
> [temp: replace code-ready/machine with anjannath/machine](https://github.com/code-ready/crc/pull/3308/commits/9ea4955b555010fd8f00b1be7925ce4cafb3e402)

why?

anjannath:
> > [temp: replace code-ready/machine with anjannath/machine](https://github.com/code-ready/crc/pull/3308/commits/9ea4955b555010fd8f00b1be7925ce4cafb3e402)
> 
> why?

we need to modify the `SharedDir` struct, the PR: https://github.com/code-ready/machine/pull/66 on `machine` repo needs to be merged and then we can remove this commit

cfergeau (commented):
I'm still not very good with the Windows side of things. I need to spend more time on ` add CIFS/SMB based file sharing for windows `
Apart from this, a few minor comments, but the way it's added looks fine.

go.mod
```diff
@@ -190,3 +190,5 @@ replace (
 
 	k8s.io/kubectl => github.com/openshift/kubernetes-kubectl v0.0.0-20220124124213-df787b5b471c
 )
+
+replace github.com/code-ready/machine v0.0.0-20220727160217-7bf0dd8c2d7b => github.com/anjannath/machine v0.7.1-0.20220829061903-fb47540d2400
```
gbraad, line 194: TODO: this needs to be changed

packaging/windows/product.wxs.template
```diff
@@ -95,12 +101,24 @@
...
-            Id="AddUserToHypervAdminGroup" 
-            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Add-LocalGroupMember -Member [LogonUser] -SID S-1-5-32-578&quot;" 
-            Before="AddUserToHypervAdminGroup" 
+        <SetProperty Action="CAAddUserToHypervAdminGroup"
+            Id="AddUserToHypervAdminGroup"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Add-LocalGroupMember -Member [LogonUser] -SID S-1-5-32-578&quot;"
+            Before="AddUserToHypervAdminGroup"
             Sequence="execute"/>
         <CustomAction Id="AddUserToHypervAdminGroup" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
+        <SetProperty Action="CACreateSMBShare"
```
gbraad, line 110: Just wondering; can this be optional? As in certain environment the use of workstations running print/filesharing is not allowed. Not sure how the installer handles this or how admins would appreciate this. The problem might be that the tooling side expects this; I do not remember seeing a check for this on startup.

We might have to create a follow-up, as when no setup/installer is delivered this becomes a manual step or something that needs to be checked for.

pkg/crc/config/settings.go
gbraad: Windows bases that on the UNC path; they can be assigned a different user and password combination. In the current case of CIFS use you only deal with a USERFOLDER and not 'arbitrary' options given by the user.

pkg/crc/config/settings.go
```diff
@@ -77,8 +89,15 @@ func RegisterSettings(cfg *Config) {
...
-	// shared dir is not implemented for windows yet
-	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+
+	// Shared directories configs
+	if runtime.GOOS == "windows" {
+		cfg.AddSetting(SharedDirPassword, "", ValidateString, SuccessfullyApplied,
+			"Password used while using CIFS/SMB file sharing (Its the password for the current logon user)")
+
+		cfg.AddSetting(EnableSharedDirs, false, validateSmbSharedDirs, SuccessfullyApplied,
+			"Enable shared directories which mounts host's $HOME at /home in the CRC VM (true/false, default: false)")
```
gbraad, line 99: why is `$HOME` used here, in Windows terminology this is:

 * [%USERPROFILE%](https://en.wikipedia.org/wiki/%25USERPROFILE%25)
 * [%HOMEDRIVE%](https://en.wikipedia.org/wiki/%25HOMEDRIVE%25)[%HOMEPATH%](https://en.wikipedia.org/wiki/%25HOMEPATH%25)

It would be best to use something familiar and accurate to the user; `$HOME` only exists as a derived variable in PowerShell. WDYT?

pkg/crc/config/settings.go
```diff
@@ -145,14 +164,6 @@ func GetNetworkMode(config Storage) network.Mode {
 	return network.ParseMode(config.Get(NetworkMode).AsString())
 }
 
-func ShouldEnableSharedDirs(config Storage) bool {
```
gbraad, line 148: always bothered about `should` here as from the functionality it means `can`

pkg/os/util_windows.go
```diff
@@ -28,3 +31,15 @@ func ReadFileUTF16LE(filename string) ([]byte, error) {
...
 	return decoded, err
 }
+
+func GetCurrentUsername() (string, error) {
+	u, err := user.Current()
+	if err != nil {
+		return "", err
+	}
+	userAndDomain := strings.Split(u.Username, "\\")
+	if len(userAndDomain) > 0 {
```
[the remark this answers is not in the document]
gbraad: actually `= 2` is more correct in that case. it would never be above this value, right?

pkg/crc/machine/vsock.go
```diff
@@ -76,6 +76,7 @@ func listOpenPorts(daemonClient *daemonclient.Client) ([]types.ExposeRequest, er
 
 const (
 	virtualMachineIP = "192.168.127.2"
+	hostVirtualIP    = "192.168.127.254"
```
gbraad, line 79: :+1

pkg/crc/config/settings.go
anjannath: yeah, help message is not accurate, would be better to just avoid using `env` variables and have the message in plain english..

how about ?
>Mounts host user's home directory or user profile folder in the CRC VM (true/false, default: false)

gbraad:
> use the msi from this PR to install crc

can you describe steps needed to test this without the MSI too? This is also suggested for: https://github.com/code-ready/crc/pull/3308#discussion_r980743590

go.mod
anjannath: will drop this commit and vendor updated `code-ready/machine` once the PR https://github.com/code-ready/machine/pull/66 is merged

packaging/windows/product.wxs.template
```diff
@@ -95,12 +101,24 @@
...
+            Id="CreateSMBShare"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;New-SmbShare -Name '[SHAREDDIRNAME]' -Path '[USERFOLDER]' -FullAccess '[LogonUser]'&quot;"
```
gbraad, line 112: to test without MSI:

```
New-SmbShare -Name 'crc-dir0' -Path '$ENV:USERPROFILE' -FullAccess '$ENV:LOGONUSER'
```

packaging/windows/product.wxs.template
```diff
@@ -95,12 +101,24 @@
...
+            Before="CreateSMBShare"
+            Sequence="execute"/>
+        <CustomAction Id="CreateSMBShare" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="check" />
+        <SetProperty Action="CAEnableFileAndPrinterSharing"
+            Id="EnableFileAndPrinterSharing"
+            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;Set-NetFirewallRule -DisplayGroup 'File And Printer Sharing' -Enabled True -Profile 'Private,Public'&quot;"
```
gbraad, line 118: to test without MSI (and verify):

```
Set-NetFirewallRule -DisplayGroup 'File And Printer Sharing' -Enabled True -Profile 'Private,Public
```

packaging/windows/product.wxs.template
gbraad: This was failing for me

```
New-SmbShare : No mapping between account names and security IDs was done.
```

@anjannath how is your account set up? I use a Microsoft account.

anjannath: Its a local account, did not test with an online microsoft account? is this an warning or error as in does it create the share when running this command or completely fails?

anjannath: @gbraad have you tried the username without the domain part, this is suggested in https://sharepoint.stackexchange.com/a/158471

gbraad: I had to remove the quotes.

packaging/windows/product.wxs.template
anjannath: this custom actions return value is ignored so even if this fails the installation will succeed to take care of situations where this is disabled using group or domain policies or any other errors, and during start when mounting the shared directory there's a check to see if the 'crc-dir0' share exists before attempting to mount

we need to have a generic todo to remove all the privileged setup operations from the msi and bring them back to `crc setup` again, i think this'll be best handled with that task once we are finally removing the msi

pkg/crc/errors/maskederror.go
```diff
@@ -0,0 +1,13 @@
...
+	"errors"
+	"strings"
+)
+
+const mask = "*****"
+
+func MaskSecretInError(err error, secret string) error {
+	errMsg := strings.ReplaceAll(err.Error(), secret, mask)
+	return errors.New(errMsg)
+}
```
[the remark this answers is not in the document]
anjannath: thanks, updated

go.mod
anjannath: this is taken care of now, PR on machine side is merged and this PR upgrades the code-ready/machine module

pkg/crc/config/settings.go
```diff
@@ -77,8 +89,15 @@ func RegisterSettings(cfg *Config) {
 		"Disable update check (true/false, default: false)")
 	cfg.AddSetting(ExperimentalFeatures, false, ValidateBool, SuccessfullyApplied,
 		"Enable experimental features (true/false, default: false)")
-	// shared dir is not implemented for windows yet
-	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+
+	// Shared directories configs
+	if runtime.GOOS == "windows" {
+		cfg.AddSetting(SharedDirPassword, "", ValidateString, SuccessfullyApplied,
+			"Password used while using CIFS/SMB file sharing (Its the password for the current logon user)")
```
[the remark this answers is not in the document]
anjannath: fixed

anjannath:
@cfergeau @gbraad All the review comments are addressed and changes have been made PTALA, thanks!!

pkg/crc/machine/driver_linux.go
```diff
@@ -47,3 +47,7 @@ func (r *RPCServerDriver) SetConfigRaw(data []byte, _ *struct{}) error {
 	return json.Unmarshal(data, &r.ActualDriver)
 }
 */
+
+func updateDriverStruct(host *host.Host, driver *machineLibvirt.Driver) error {
+	return nil
```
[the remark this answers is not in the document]
anjannath: now returning `drivers.ErrNotImplemented` instead of `nil` for darwin and linux

anjannath:
PR rebased on top of #3333

gbraad (approved):
Approved for podman bundle
Unable to test for the OpenShift bundle

Follow-up issues needed for the initial flow. Can't be enabled by default as an explicit password is needed.

anjannath:
> Approved for podman bundle Unable to test for the OpenShift bundle
> 

tested with the new `4.11.7` openshift bundle and working as expected

pkg/crc/config/settings.go
```diff
@@ -56,6 +57,17 @@ func RegisterSettings(cfg *Config) {
 		return ValidateBundlePath(value, GetPreset(cfg))
 	}
 
+	validateSmbSharedDirs := func(value interface{}) (bool, string) {
+		if !cfg.Get(HostNetworkAccess).AsBool() {
+			return false, fmt.Sprintf("%s can only be used with %s set to 'true'",
+				EnableSharedDirs, HostNetworkAccess)
+		}
```
cfergeau, line 64: Is there any way to restrict this to only the ports needed for file sharing?
The VM having access to all the services running on the host, and also knowing the user login/password does not sound great from a security perspective. We've removed a big part of the isolation between host and guest imo.

pkg/crc/config/callbacks.go
```diff
@@ -33,6 +33,11 @@ func RequiresCRCSetup(key string, _ interface{}) string {
 		"Please run 'crc setup' for this configuration to take effect.", key)
 }
 
+func RequiresCleanupAndSetupMsg(key string, _ interface{}) string {
+	return fmt.Sprintf("Changes to configuration property '%s' are only applied during 'crc setup'.\n"+
+		"Please run 'crc cleanup' followed by 'crc setup' for this configuration to take effect.", key)
+}
```
cfergeau, line 39: Follow-up issue to make this more dynamic would be useful to have.

cfergeau (approved):
Was this waiting for a second ACK? Gerard approved this a while ago.

pkg/crc/config/settings.go
anjannath: i don't know if its possible to filter for ports, i guess that will also need changes to `gvisor-tap-vsock` 

maybe we could try adding a `:port` to the NAT table entries setup during daemon start and see if that works!

pkg/crc/config/callbacks.go
anjannath: https://github.com/code-ready/crc/issues/3367

PR #3314, opened 2022-08-24
Version: use GetBundleVersion for any type of preset

praveenkumar:
This pr is a refactor to use `GetBundleVersion` for any type of
preset bundle and return the correct bundle version. After this
PR we don't need `GetPodmanVersion()` and it removes it.

### DIFF
--- a/pkg/crc/version/version.go
+++ b/pkg/crc/version/version.go
@@ -22,8 +22,8 @@ var (
 	// The SHA-1 of the commit this executable is build off
 	commitSha = "sha-unset"
 
-	// Bundle version which used for the release.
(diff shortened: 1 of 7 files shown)


pkg/crc/version/version.go
```diff
@@ -22,8 +22,8 @@ var (
 	// The SHA-1 of the commit this executable is build off
 	commitSha = "sha-unset"
 
-	// Bundle version which used for the release.
-	bundleVersion = "0.0.0-unset"
+	// OCP version which used for the release.
```
cfergeau, line 25: "which is used for the release", not "which used"

cfergeau (approved):

Issue #3313, opened 2022-08-24
Remove Validation for using a different version of bundle for a specific CRC release


praveenkumar:
As of now, each CRC release is tied to a specific version of bundle and if user try to use different version of bundle with release bit then we don't allow it (error out) using https://github.com/code-ready/crc/blob/main/pkg/crc/validation/validation.go#L58-L80 block of code.

With one of the discussion with @veillard, he prefer we should remove that restriction and show a user warning if using a different version of bundle then what we ship for a specific release, something like our dev bits. This issue is to discuss it and prioritize the required work. 

It has been around 3 years since we are shipping crc with pinned version and also we create bundles during dev cycle but most of the time we dont' upload all of them but only the one which would be part of crc release so with this suggested change we are able to only release the bundle once it is tested with specific version of crc even delaying a crc release. During this period we also didn't a lot of user creating their own bundle and using it with CRC so I am assuming 95% user only consume the bundles which we deliver and go with it.

cfergeau:
This will also be required by the work being tracked in https://github.com/code-ready/crc/issues/3206
Fwiw, https://github.com/code-ready/crc/blob/main/pkg/crc/validation/validation.go#L58-L80 has been partially broken for a long while, this block:
```
if err := ValidatePath(bundlePath); err != nil {
	return err
}
```
means that if the requested bundle version is already unpacked in `~/.crc/cache/` then we'll consider it valid. In upgrade scenarios, this means a newer crc release can start older bundles. And the only step required for an older release to use a newer bundle is to unpack the bundle manually.

praveenkumar:
> Fwiw, https://github.com/code-ready/crc/blob/main/pkg/crc/validation/validation.go#L58-L80 has been partially broken for a long while, if the user request a bundle version which is already unpacked in `~/.crc/cache/` then we'll consider it valid. In upgrade scenarios, this means a newer crc release can start older bundles. And the only step required for an older release to use a newer bundle is to unpack the bundle manually.

@cfergeau you mean something like following works in release bits?

- Manually download 4.11.1 bundle and extract it to `~/.crc/cache` 
- Then use crc 2.7.1 release bits to start it with `-b <path_to_4.11.1>`

cfergeau:
[...]

PR #3315, opened 2022-08-25
Remove Validation for using a different bundle for release bits

praveenkumar:
This PR remove the validation of bundle version check and error out
in case it doesn't match with crc release bit bundle version.

Just a warn message displayed to user in case use of different bundle.

**Fixes:** Issue #3313

### DIFF
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -58,7 +58,7 @@ func runSetup(arguments []string) error {
 	}
 
 	if !config.Get(crcConfig.Bundle).IsDefault {
-		if err := validation.ValidatePath(config.Get(crcConfig.Bundle).AsString()); err != nil {
+		if err := validation.ValidateBundlePath(config.Get(crcConfig.Bundle).AsString(), crcConfig.GetPreset(config)); err != nil {
 			return err
 		}
 	}
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -237,6 +237,13 @@ func GetBundleNameWithoutExtension(bundleName string) string {
 	return strings.TrimSuffix(bundleName, bundleExtension)
 }
 
+func GetBundleNameWithExtension(bundleName string) string {
+	if strings.HasSuffix(bundleName, bundleExtension) {
+		return bundleName
+	}
+	return fmt.Sprintf("%s%s", bundleName, bundleExtension)
+}
+
 func GetCustomBundleName(bundleFilename string) string {
 	re := regexp.MustCompile(`(?:_[0-9]+)*.crcbundle$`)
 	baseName := re.ReplaceAllLiteralString(bundleFilename, "")
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -67,10 +67,11 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 
 func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 	bundleName := filepath.Base(bundlePath)
-	_, err := bundle.Get(bundleName)
+	bundleMetadata, err := bundle.Get(bundleName)
 	if err != nil {
 		return ValidateBundlePath(bundlePath, preset)
 	}
+	bundleMismatchWarning(bundleMetadata.GetBundleName(), preset)
 	/* 'bundle' is already unpacked in ~/.crc/cache */
 	return nil
 }
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -70,11 +70,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 			logging.Warnf("Using custom bundle %s", userProvidedBundle)
 			return nil
 		}
-		if !constants.IsRelease() {
-			logging.Warnf("Using unsupported bundle %s", userProvidedBundle)
-			return nil
-		}
-		return fmt.Errorf("%s is not supported by this crc executable, please use %s", userProvidedBundle, constants.GetDefaultBundle(preset))
+		logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
 	}
 	return nil
 }
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -61,6 +61,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 	}
 
 	userProvidedBundle := filepath.Base(bundlePath)
+	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)
 	if userProvidedBundle != constants.GetDefaultBundle(preset) {
 		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
 		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -61,18 +61,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 	}
 
 	userProvidedBundle := filepath.Base(bundlePath)
-	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)
-	if userProvidedBundle != constants.GetDefaultBundle(preset) {
-		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
-		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
-		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
-			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset))))
-		if usingCustomBundle {
-			logging.Warnf("Using custom bundle %s", userProvidedBundle)
-			return nil
-		}
-		logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
-	}
+	bundleMismatchWarning(userProvidedBundle, preset)
 	return nil
 }
 
@@ -84,6 +73,21 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 	}
 	/* 'bundle' is already unpacked in ~/.crc/cache */
 	return nil
+}
+
+func bundleMismatchWarning(userProvidedBundle string, preset crcpreset.Preset) {
+	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)
+	if userProvidedBundle != constants.GetDefaultBundle(preset) {
+		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
+		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
+		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
+			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset))))
+		if usingCustomBundle {
+			logging.Warnf("Using custom bundle %s", userProvidedBundle)
+		} else {
+			logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
+		}
+	}
 }
 
 // ValidateIPAddress checks if provided IP is valid


cfergeau:
> **Fixes:** Issue #3313

The current consensus in #3313 seems to have a slightly stricter check, so I would not call it 'fixed'

Tested the PR on my test machine, but not seeing a warning:
```
teuf@ramen:~/crc/crc (issue_3313)$ make linux-release-binary
GOOS=linux GOARCH=amd64 go build -tags "containers_image_openpgp" -ldflags="-X github.com/code-ready/crc/pkg/crc/version.crcVersion=2.7.1 -X github.com/code-ready/crc/pkg/crc/version.ocpVersion=4.11.0 -X github.com/code-ready/crc/pkg/crc/version.okdVersion=4.11.0-0.okd-2022-07-29-154152 -X github.com/code-ready/crc/pkg/crc/version.podmanVersion=4.1.1 -X github.com/code-ready/crc/pkg/crc/version.commitSha=e7d0d07a  -X 'github.com/code-ready/crc/pkg/crc/version.linuxReleaseBuild=true' -X github.com/code-ready/crc/pkg/crc/segment.WriteKey=cvpHsNcmGCJqVzf6YxrSnVlwFSAZaYtp" -o out/linux-amd64/crc  ./cmd/crc

$ ./out/linux-amd64/crc setup  -b ~/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle 
INFO Using bundle path /home/teuf/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle 
INFO Checking for obsolete admin-helper executable 
INFO Checking if active user/process is currently part of the libvirt group 
INFO Checking crc daemon systemd service          
INFO Setting up crc daemon systemd service        
INFO Checking crc daemon systemd socket units     
INFO Checking if /home/teuf/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle exists 


$ ./out/linux-amd64/crc start -b ~/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle 
INFO Checking for obsolete admin-helper executable 
INFO Checking if active user/process is currently part of the libvirt group 
INFO Checking crc daemon systemd socket units     
INFO Loading bundle: crc_libvirt_4.10.22_amd64... 
CRC requires a pull secret to download content from Red Hat.

**Comment by praveenkumar** [comment]
> > **Fixes:** Issue #3313
> 
> The current consensus in #3313 seems to have a slightly stricter check, so I would not call it 'fixed'
> 
> Tested the PR on my test machine, but not seeing a warning:
> 

@cfergeau This is something you mentioned https://github.com/code-ready/crc/issues/3313#issuecomment-1225349342 here, will create another commit to fix that also.

```
$ ./crc setup -b ~/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle
INFO Using bundle path /home/prkumar/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle 
INFO Checking for obsolete admin-helper executable 
INFO Checking if active user/process is currently part of the libvirt group 
INFO Checking crc daemon systemd service          
INFO Checking crc daemon systemd socket units     
INFO Checking if /home/prkumar/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle exists 

 $ ./crc config set bundle ~/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle
WARN Using crc_libvirt_4.10.22_amd64.crcbundle bundle, but crc_libvirt_4.11.0_amd64.crcbundle is expected for this release 
Successfully configured bundle to /home/prkumar/.crc/cache/crc_libvirt_4.10.22_amd64.crcbundle
```

File: cmd/crc/cmd/start.go [hunk 1]
```diff
@@ -169,7 +169,7 @@ func validateStartFlags() error {
 	if err := validation.ValidateDiskSize(config.Get(crcConfig.DiskSize).AsInt()); err != nil {
 		return err
 	}
-	if err := validation.ValidateBundle(config.Get(crcConfig.Bundle).AsString(), crcConfig.GetPreset(config)); err != nil {
+	if err := validation.ValidateBundlePath(config.Get(crcConfig.Bundle).AsString(), crcConfig.GetPreset(config)); err != nil {
```
cfergeau, line 172: This reverts https://github.com/code-ready/crc/commit/f148ddc8687689a97c9b7e6baac4697c17b3547b and it looks like it will make the .crcbundle file required again even if it's already uncompressed.

praveenkumar: @cfergeau This function actually don't run `ValidateBundlePath(bundlePath, preset)` in case bundle is already extracted, no matter which version of bundle even it is supported by crc or not and we are not able to provide early warning to the users in case different bundle in use. Either we have to add another logic here where after getting `bundlemetadata` we also check if the bundle is supported on or not and show the user warning or use it in current form. Yes it will be require `.crcbundle` to be available on `~./cache` folder even it is uncompressed.


```golang
func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
	_, err := bundle.Get(bundleName)
	if err != nil {
		return ValidateBundlePath(bundlePath, preset)
	}
	/* 'bundle' is already unpacked in ~/.crc/cache */
	return nil
}
```

cfergeau: Yes more code will be needed/some logic will need to be reworked. Fixing one bug by reintroducing another does not sound like a great deal.

praveenkumar: @cfergeau Added that logic for `ValidateBundle` .

pkg/crc/validation/validation.go
```diff
@@ -77,10 +77,13 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
...
 func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 	bundleName := filepath.Base(bundlePath)
-	_, err := bundle.Get(bundleName)
+	bundleMetadata, err := bundle.Get(bundleName)
 	if err != nil {
 		return ValidateBundlePath(bundlePath, preset)
 	}
+	if bundleMetadata.GetBundleName() != bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset)) {
+		logging.Warnf("Using %s bundle, but %s is expected for this release", bundleName, constants.GetDefaultBundle(preset))
+	}
```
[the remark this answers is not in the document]
praveenkumar: > Why is this duplicating code in `ValidateBundlePath` but not taking into account custom bundles?

@cfergeau it is duplicate because when bundle is extracted then `ValidateBundlePath` not even run so the warning is ignored but now it is not. This warning works with custom bundle also but it will show same message then custom bundle one, for `start` command do we really required to put custom bundle warning? May be I just put those warnings to a helper function and consume it for both functions?


> Why is the validation for `-b` different between the start and setup commands?

I don't know, may be we are assuming that during setup we always use the bundle which is not extracted so why get the bundle metadata logic?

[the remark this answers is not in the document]
cfergeau: > May be I just put those warnings to a helper function and consume it for both functions?

Yes, this was going to be my suggestion, sharing common code is better than partially duplicating code with no clear reason for the difference in behaviour.

pkg/crc/validation/validation.go
```diff
@@ -85,6 +75,21 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 	return nil
 }
 
+func bundleMismatchWarning(userProvidedBundle string, preset crcpreset.Preset) {
+	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)
```
cfergeau, line 80: I'd prefer if this `GetBundleNameWithExtension` change was done in the commit introducing GetBundleNameWithExtension, preferrably with an explanation why it's needed. Hard to understand why it's needed now when it's done silently in a commit which is supposed to be only code movement.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -237,6 +237,13 @@ func GetBundleNameWithoutExtension(bundleName string) string {
 	return strings.TrimSuffix(bundleName, bundleExtension)
 }
 
+func GetBundleNameWithExtension(bundleName string) string {
```
cfergeau, line 240: Terminology note for later: Ideally it would be possible to use `BundleName` instead of `BundleNameWithoutExtension` and `BundleFilename` instead of `BundleNameWithExtension` but this would require some extensive review of the code base to make it all consistent.

pkg/crc/validation/validation.go
```diff
@@ -85,6 +75,21 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
...
+	if userProvidedBundle != constants.GetDefaultBundle(preset) {
+		warnMsg := fmt.Sprintf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau if I don't use `warnMsg` then I need to use `if <-> else` block because for custom bundle we don't want to put the warning that it is not supported but just show it is custom bundle.
```golang
	if userProvidedBundle != constants.GetDefaultBundle(preset) {
		constants.GetDefaultBundle(preset))
		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset))))
		if usingCustomBundle {
			logging.Warnf("Using custom bundle %s", userProvidedBundle)
		} else {
                        logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, 
                }
	}
```

pkg/crc/validation/validation.go
praveenkumar: @cfergeau you mean squashing those 2 commits or update the current commit by saying it is used in following commit message?

pkg/crc/validation/validation.go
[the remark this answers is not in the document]
cfergeau: Yes, this `if/else` block looks easy enough to read?

pkg/crc/validation/validation.go
cfergeau:
```
commit 494a512d4223f96a99619996dec3a834f2a1efe4
Author: Praveen Kumar <kumarpraveen.nitdgp@gmail.com>
Date:   Thu Sep 1 16:12:12 2022 +0530

    Bundle: Add helper function to get bundle name with extension
    
    This new method is used in pkg/crc/validation/validation.go in order to
    solve XXX

diff --git a/pkg/crc/machine/bundle/metadata.go b/pkg/crc/machine/bundle/metadata.go
index 2fd8f1b6..754fb991 100644
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -237,6 +237,13 @@ func GetBundleNameWithoutExtension(bundleName string) string {
 	return strings.TrimSuffix(bundleName, bundleExtension)
 }
 
+func GetBundleNameWithExtension(bundleName string) string {
+	if strings.HasSuffix(bundleName, bundleExtension) {
+		return bundleName
+	}
+	return fmt.Sprintf("%s.%s", bundleName, bundleExtension)
+}
+
 func GetCustomBundleName(bundleFilename string) string {
 	re := regexp.MustCompile(`(?:_[0-9]+)*.crcbundle$`)
 	baseName := re.ReplaceAllLiteralString(bundleFilename, "")
diff --git a/pkg/crc/validation/validation.go b/pkg/crc/validation/validation.go
index 8993e975..4139ccae 100644
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -61,6 +61,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 	}
 
 	userProvidedBundle := filepath.Base(bundlePath)
+	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)
 	if userProvidedBundle != constants.GetDefaultBundle(preset) {
 		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
 		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
```

gbraad:
Why in total 6 commits for just this few amount of lines of actual code that has been changed.

praveenkumar:
> Why in total 6 commits for just this few amount of lines of actual code that has been changed.
> 
> Two helper commits? Warning addition? Can we squash some of this?

Each commit describe the change which it do independently.

pkg/crc/validation/validation.go
```diff
@@ -61,6 +61,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 	}
 
 	userProvidedBundle := filepath.Base(bundlePath)
+	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)
```
cfergeau: I have a hard time understanding why this change is needed from the commit log :-/

pkg/crc/validation/validation.go
```diff
@@ -61,18 +61,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
...
-		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
-		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
-			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset))))
-		if usingCustomBundle {
-			logging.Warnf("Using custom bundle %s", userProvidedBundle)
-			return nil
-		}
-		logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
-	}
+	bundleMismatchWarning(userProvidedBundle, preset)
```
cfergeau, line 64: `warnIfBundleMismatch(ed)`?

pkg/crc/validation/validation.go
```diff
@@ -79,14 +79,15 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
...
 		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
 		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
 		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
 			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset))))
 		if usingCustomBundle {
-			logging.Warnf("Using custom bundle %s", userProvidedBundle)
+			warnMsg = fmt.Sprintf("Using custom bundle %s", userProvidedBundle)
 		}
-		logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
+		logging.Warn(warnMsg)
```
cfergeau, line 90: Why not if/else? This feels easier to read rather than having to look up what `warnMsg` is set to after reaching the end of the code?

```
if usingCustomBundle {
	logging.Warnf("Using custom bundle %s", userProvidedBundle)
} else {
	logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
}
```

cfergeau (commented):
While reading the commit logs, the reason for `Use ValidateBundlePath for setup command` was not fully clear. Is this causing user-visiblle changes? fixing a bug?

pkg/crc/machine/bundle/metadata.go
anjannath, line 240: do we need this helper, `-b` currently accepts only the path to the bundle, and that'll always have the extension?

pkg/crc/validation/validation.go
praveenkumar: @cfergeau it is there to make use of `GetBundleNameWithExtension` function, Removing it from this commit.

pkg/crc/machine/bundle/metadata.go
praveenkumar: @anjannath This is needed because if user already have a bundle extracted in the local directory then  as part of `crc start` we need to identify if the bundle is same what we released with, in subsequent commit you will understand why this helper required.

pkg/crc/validation/validation.go
```diff
@@ -61,6 +61,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 	}
 
 	userProvidedBundle := filepath.Base(bundlePath)
+	userProvidedBundle = filepath.Base(userProvidedBundle)
```
cfergeau: Why is this no longer `	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)`?

pkg/crc/machine/bundle/metadata.go
cfergeau: I don't see anything in the subsequent commit *explaining* why it's needed. I've also asked several times to clarify this in the commit log.

cfergeau (changes requested):

pkg/crc/validation/validation.go
praveenkumar: @cfergeau https://github.com/code-ready/crc/pull/3315#discussion_r962922556 I think I just don't need this line in this commit itself.

cmd/crc/cmd/setup.go
```diff
@@ -58,7 +58,7 @@ func runSetup(arguments []string) error {
 	}
 
 	if !config.Get(crcConfig.Bundle).IsDefault {
-		if err := validation.ValidatePath(config.Get(crcConfig.Bundle).AsString()); err != nil {
+		if err := validation.ValidateBundlePath(config.Get(crcConfig.Bundle).AsString(), crcConfig.GetPreset(config)); err != nil {
```
cfergeau, line 61: Before this change, there was no warning displayed at all? commit log is not clear about that.

pkg/crc/validation/validation.go
cfergeau: > Hard to understand why it's needed now when it's done silently in a commit which is supposed to be only code movement.

Unresolving this comment... :( The previous commit has an explanation now, not fully clear if this added call will change behaviour right away or not. I still feel this belongs in the previous commit, with a clear statement whether `userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)` changes the current behaviour, or not.

pkg/crc/validation/validation.go
```diff
@@ -67,10 +67,11 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
 
 func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 	bundleName := filepath.Base(bundlePath)
-	_, err := bundle.Get(bundleName)
+	bundleMetadata, err := bundle.Get(bundleName)
 	if err != nil {
 		return ValidateBundlePath(bundlePath, preset)
 	}
+	bundleMismatchWarning(bundleMetadata.GetBundleName(), preset)
```
cfergeau, line 74: Hmm or is the `userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)` call needed by this commit and only this one?

cfergeau:
While testing this PR, I noticed an extra `.`:
```
$ ./out/linux-amd64/crc start  -b ~/.crc/cache/crc_libvirt_4.9.10
WARN Using crc_libvirt_4.9.10..crcbundle bundle, but crc_podman_libvirt_4.1.1_amd64.crcbundle is expected for this release 
```

`crc_libvirt_4.9.10..crcbundle` has 2 `.` before the extension.

cfergeau:
Fwiw, after the 2nd commit in this series, 2 warnings are displayed with `$ ./out/linux-amd64/crc  setup -b ~/crc/bundles/crc_libvirt_4.9.0.crcbundle` when `crc_libvirt_4.9.0.crcbundle` is not in the cache.

cmd/crc/cmd/setup.go
cfergeau: Answering  myself, yes, no warnings before this commit, so switching to `ValidateBundlePath` allows to show a warning when the specified bundle does not correspond to the expected one. Would be good to explicitly mention in the commit log as this explains what is being done, describes how to test it.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -237,6 +237,13 @@ func GetBundleNameWithoutExtension(bundleName string) string {
...
+	if strings.HasSuffix(bundleName, bundleExtension) {
+		return bundleName
+	}
+	return fmt.Sprintf("%s.%s", bundleName, bundleExtension)
```
cfergeau, line 244: `bundleExtension` already has the `.`, it needs to be removed.
And I would add the ` userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)` in this commit, and change the commit log to something like:
> Silence incorrect bundle mismatch warning
>
> The previous commit introduced a warning when `crc setup` is used with its `-b` option, and when the bundle version does not correspond to the one hardcoded in crc's Makefile. However, if the bundle is unpacked and one uses `crc setup -b ~/.crc/cache/crc_libvirt_4.11.1_amd64`, the warning will also be printed because `crc_libvirt_4.11.1_amd64` is different from the expected `crc_libvirt_4.11.1_amd64.crcbundle`. While it's questionable if `-b` should accept a path or not, this has been working until now.
> This commit introduces a `GetBundleNameWithExtension()` helper which will be used to silence this warning.

pkg/crc/validation/validation.go
cfergeau: Answered elsewhere, it's not related to this commit.

pkg/crc/machine/bundle/metadata.go
praveenkumar: @cfergeau This helper function is not silence incorrect bundle mismatch warning but it actually enable to show that warning. Existing commit message have info why we are putting that helper function.

cfergeau: > In following commits we are going to get the bundleName from bundle metadata which doesn't have `.crcbundle` extension and this helper provide bundle name with extension to compare with the release bundle version so we can provide the bundle-mismatch warning.

If I go with just this commit log, then imo the whole commit should be squashed in ` Validate: Add warning in case bundle metadata have different bundle name` and `bundle.GetBundleNameWithExtension(userProvidedBundle)` can be moved to `ValidateBundle` to make it clear when it's required to add the extension.

However, this is not the only case when it's needed, if I take the current PR and remove the `userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)` line, then `crc setup -b ~/.crc/cache/crc_libvirt_4.11.1_amd64` will print an unwanted warning.
If I checkout `Bundle: Add helper function to get bundle name with extension` and remove the `userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)` I also get this warning. With this line added, then the warning is gone.
So at the point you add `Bundle: Add helper function to get bundle name with extension` to the PR, I would say it's fixing an incorrect warning. That it's also useful when you add `GetBundleName` later on is a nice plus, but imo not the main reason we need it.

praveenkumar: @cfergeau I misunderstood the part of using same bundle but without extension is also create warning for the user and that is an incorrect/unwanted warning and this is what is silence through it. To be honest I didn't thought of it when implemented this part as my initial thinking was just using path always with extension. But now it clear it so updated the commit message.

cfergeau: Yes I only realized this morning that you had not noticed this side-effect and added it for a different reason, it's also clear now on my side! :)

pkg/crc/validation/validation.go
```diff
@@ -86,6 +75,20 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
...
 
+func bundleMismatchWarning(userProvidedBundle string, preset crcpreset.Preset) {
+	userProvidedBundle = bundle.GetBundleNameWithExtension(userProvidedBundle)
+	if userProvidedBundle != constants.GetDefaultBundle(preset) {
+		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
+		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
+		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
+			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset))))
+		if usingCustomBundle {
+			logging.Warnf("Using custom bundle %s", userProvidedBundle)
```
cfergeau, line 87: Oh, I hadn't paid attention until now, but by removing the early `return` which was present before `Validation: add bundleMismatchWarning helper to warn user`, you need to add ` Validation: separate custom bundle and mismatch bundle warning` later on. Would be worth squashing  both together.

cfergeau (commented):
One last comment :(

pkg/crc/validation/validation.go
```diff
@@ -61,18 +61,7 @@ func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
...
-		// Should append underscore (_) here, as we don't want crc_libvirt_4.7.15.crcbundle
-		// to be detected as a custom bundle for crc_libvirt_4.7.1.crcbundle
-		usingCustomBundle := strings.HasPrefix(bundle.GetBundleNameWithoutExtension(userProvidedBundle),
-			fmt.Sprintf("%s_", bundle.GetBundleNameWithoutExtension(constants.GetDefaultBundle(preset))))
-		if usingCustomBundle {
-			logging.Warnf("Using custom bundle %s", userProvidedBundle)
-			return nil
-		}
-		logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
-	}
```
cfergeau, line 78: > We also don't want user to show both warning in case user uses custom bundle, so this patch also logs only respective warning as per bundle filename or metadata.

The code before adding the helper was not printed 2 warnings, but just one. The quoted part of the commit log seems to emphasize that there is no behaviour change, but I don't think it's very important to have it.

cfergeau (approved):

PR #3302, opened 2022-08-15
Update the kernel args to instance config during stop

praveenkumar:
This PR is going to update the kernel paramater to instance config
when stop operation happen.

In mac we use kernel parameter as part of vfkit driver to start
the instance but during the reboot it changes because of MCO and
we are getting error during `start => stop => start` operation.

```
systemctl[787]: Failed to switch root: Specified switch root path '/sysroot' does not seem to be an OS tree. os-release file is missing.
```

ostree changes from (`boot.0` => `boot.1`)
```
ostree=/ostree/boot.0/rhcos/b6f1ba590ef2df8e30bd88833fd145bfbbd424eca8ad235f573f2a136f09b0d6/0
```
to
```
ostree=/ostree/boot.1/rhcos/b6f1ba590ef2df8e30bd88833fd145bfbbd424eca8ad235f573f2a136f09b0d6/0
```

As part of the MCO logs
```
Bootloader updated; bootconfig swap: yes; bootversion: boot.1.1, deployment count change: -1
```

PR #3318, opened 2022-08-25
Small code simplifications

cfergeau:
This should make GetDefaultBundle a bit simpler, and this also removes one not useful(?) WaitForSSH call from the vfkit machine driver.

### DIFF
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -83,32 +83,30 @@ func GetAdminHelperURL() string {
 	return GetAdminHelperURLForOs(runtime.GOOS)
 }
 
-func defaultBundleForOs(preset crcpreset.Preset) map[string]string {
+func GetDefaultBundle(preset crcpreset.Preset) string {
+	var bundleMap map[string]string
 	switch preset {
 	case crcpreset.Podman:
-		return map[string]string{
+		bundleMap = map[string]string{
 			"darwin":  fmt.Sprintf("crc_podman_vfkit_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
 			"linux":   fmt.Sprintf("crc_podman_libvirt_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
 			"windows": fmt.Sprintf("crc_podman_hyperv_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
 		}
 	case crcpreset.OKD:
-		return map[string]string{
+		bundleMap = map[string]string{
 			"darwin":  fmt.Sprintf("crc_okd_vfkit_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
 			"linux":   fmt.Sprintf("crc_okd_libvirt_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
 			"windows": fmt.Sprintf("crc_okd_hyperv_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH),
 		}
 	default:
-		return map[string]string{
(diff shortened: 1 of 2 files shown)


pkg/crc/constants/constants.go
```diff
@@ -83,32 +84,29 @@ func GetAdminHelperURL() string {
...
+	case "darwin":
+		bundleName.WriteString("_vfkit")
+	case "linux":
+		bundleName.WriteString("_libvirt")
+	case "windows":
+		bundleName.WriteString("_hyperv")
+	}
+
+	fmt.Fprintf(&bundleName, "_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH)
+	return bundleName.String()
```
anjannath, line 109: Nice 👍🏽

pkg/crc/machine/driver_darwin.go
```diff
@@ -47,10 +45,6 @@ func updateKernelArgs(vm *virtualMachine) error {
 	}
 	defer sshRunner.Close()
 
-	if err := sshRunner.WaitForConnectivity(context.Background(), 20*time.Second); err != nil {
```
anjannath, line 50: this was added because we observed some errors related to ssh connectivity during testing, see https://github.com/code-ready/crc/pull/3302/files#r945923907

pkg/crc/constants/constants.go
```diff
@@ -83,32 +83,30 @@ func GetAdminHelperURL() string {
 	return GetAdminHelperURLForOs(runtime.GOOS)
 }
 
-func defaultBundleForOs(preset crcpreset.Preset) map[string]string {
```
praveenkumar, line 86: I am going to use this helper function to get bundle name when provided by http/https or docker scheme. I am changing the signature of this `func defaultBundleForOs(preset crcpreset.Preset, version string) map[string]string` so that I can add another function in constant something like. As of now our images are like `quay.io/crcont/ocp-bundle:4.11.0` so only `tag` is useful to create the bundle name for respective preset.

```golang
func GetBundleNameFromBundlePath(preset crcpreset.Preset, bundlePath string) string {
	switch {
	case strings.HasPrefix(bundlePath, "http://"), strings.HasPrefix(bundlePath, "https://"):
		return filepath.Base(bundlePath)
	case strings.HasPrefix(bundlePath, "docker://"):
		tag := strings.Split(bundlePath, ":")[2]
		bundles := defaultBundleForOs(preset, tag)
		return bundles[runtime.GOOS]
	default:
		return filepath.Base(bundlePath)
	}
}
```

pkg/crc/constants/constants.go
praveenkumar: This is really nice, I can cherry pick it 👍🏼

pkg/crc/machine/driver_darwin.go
praveenkumar: yes, it was added because of ssh connection delay happen sometime and observed by QE also iirc.

cfergeau: Added questions to https://github.com/code-ready/crc/pull/3302/files#r955160463 to get a better understanding of why this was added, and rework this patch accordingly.

anjannath: I replied to the comments in the other PR, but could not reproduce it again with both 4.11.0 or 4.11.1 bundles

praveenkumar: Tried 3 times. and didn't get error so now I don't have any strong opinion and just see this wait as safe option only. (in case there is some ssh connection timeout so retry will work). I am OK both options (remove or have it with comment).

praveenkumar (approved):

anjannath (approved):
LGTM we'll see if we need a retry/waitforconnectivity again, so far tests show its not needed

PR #3319, opened 2022-08-25
Extend -b option to accept docker/http/https scheme

praveenkumar:
This patch enable user to use `-b` option in different ways for
`crc config`, `crc setup` and `crc start`.

```
$ crc config set bundle <local_bundle_path or http://foo.com/bundle or docker://registry.com/bundle:version>

$ crc setup -b <local_bundle_path or http://foo.com/bundle or docker://registry.com/bundle:version>

$ crc start -b <local_bundle_path or http://foo.com/bundle or docker://registry.com/bundle:version>
```

Assumption here is user doesn't change the bundle name during upload to
different location (like fileshare or registry) since we are building
the bundleName using those info.

## Testing

To test different image location try following
```
$ crc setup -b docker://quay.io/praveenkumar/podman-bundle:4.1.1  => this will download and extract image 
$ crc start  -b docker://quay.io/praveenkumar/podman-bundle:4.1.1 => this will no do anything and consume it.

If you don't use `setup` and directly use `start` then it will download and extract as part of start command.
```

To test with http/https location
```
$ crc setup -b https://storage.googleapis.com/crc-bundle-github-ci/crc_podman_libvirt_4.1.1_amd64.crcbundle
$ crc start -b https://storage.googleapis.com/crc-bundle-github-ci/crc_podman_libvirt_4.1.1_amd64.crcbundle

Same as  image testing.
```

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -61,7 +61,8 @@ func RegisterSettings(cfg *Config) {
 		fmt.Sprintf("Virtual machine preset (valid values are: %s, %s and %s)", preset.Podman, preset.OpenShift, preset.OKD))
 	// Start command settings in config
 	cfg.AddSetting(Bundle, defaultBundlePath(cfg), validateBundlePath, SuccessfullyApplied,
-		fmt.Sprintf("Bundle path (string, default '%s')", defaultBundlePath(cfg)))
+		fmt.Sprintf("Bundle path/URI - absolute or local path, http, https or docker URI (string, like 'https://foo.com/%s', 'docker://quay.io/myorg/%s:%s' default '%s' )",
+			constants.GetDefaultBundle(GetPreset(cfg)), constants.GetDefaultBundle(GetPreset(cfg)), version.GetCRCVersion(), defaultBundlePath(cfg)))
 	cfg.AddSetting(CPUs, defaultCPUs(cfg), validateCPUs, RequiresRestartMsg,
 		fmt.Sprintf("Number of CPU cores (must be greater than or equal to '%d')", defaultCPUs(cfg)))
 	cfg.AddSetting(Memory, defaultMemory(cfg), validateMemory, RequiresRestartMsg,
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -27,7 +27,6 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
-	AutostartTray           = "autostart-tray"
 	KubeAdminPassword       = "kubeadmin-password"
 	Preset                  = "preset"
 	EnableSharedDirs        = "enable-shared-dirs"
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -86,7 +86,7 @@ func GetAdminHelperURL() string {
 	return GetAdminHelperURLForOs(runtime.GOOS)
 }
 
-func GetDefaultBundle(preset crcpreset.Preset) string {
+func BundleForPreset(preset crcpreset.Preset, version string) string {
 	var bundleName strings.Builder
 
 	bundleName.WriteString("crc")
@@ -107,8 +107,12 @@ func GetDefaultBundle(preset crcpreset.Preset) string {
 		bundleName.WriteString("_hyperv")
 	}
 
-	fmt.Fprintf(&bundleName, "_%s_%s.crcbundle", version.GetBundleVersion(preset), runtime.GOARCH)
+	fmt.Fprintf(&bundleName, "_%s_%s.crcbundle", version, runtime.GOARCH)
 	return bundleName.String()
+}
+
+func GetDefaultBundle(preset crcpreset.Preset) string {
+	return BundleForPreset(preset, version.GetBundleVersion(preset))
 }
 
 var (
--- a/pkg/crc/image/image.go
+++ b/pkg/crc/image/image.go
@@ -26,7 +26,7 @@ type imageHandler struct {
 	imageURI string
 }
 
-func uri(preset crcpreset.Preset) string {
+func defaultURI(preset crcpreset.Preset) string {
 	return fmt.Sprintf("//%s/%s:%s", constants.RegistryURI, getImageName(preset), version.GetCRCVersion())
 }
 
@@ -95,9 +95,12 @@ func getImageName(preset crcpreset.Preset) string {
 	}
 }
 
-func PullBundle(preset crcpreset.Preset) error {
+func PullBundle(preset crcpreset.Preset, imageURI string) error {
+	if imageURI == "" {
+		imageURI = defaultURI(preset)
+	}
 	imgHandler := imageHandler{
-		imageURI: uri(preset),
+		imageURI: strings.TrimPrefix(imageURI, "docker:"),
 	}
 	destDir, err := os.MkdirTemp(constants.MachineCacheDir, "tmpBundleImage")
 	if err != nil {
--- a/pkg/crc/image/image.go
+++ b/pkg/crc/image/image.go
@@ -129,7 +129,7 @@ func GetPresetName(imageName string) crcpreset.Preset {
 	return preset
 }
 
-func PullBundle(preset crcpreset.Preset, imageURI string) error {
+func PullBundle(preset crcpreset.Preset, imageURI string) (string, error) {
 	if imageURI == "" {
 		imageURI = defaultURI(preset)
 	}
@@ -138,33 +138,37 @@ func PullBundle(preset crcpreset.Preset, imageURI string) error {
 	}
 	destDir, err := os.MkdirTemp(constants.MachineCacheDir, "tmpBundleImage")
 	if err != nil {
-		return err
+		return "", err
 	}
 	defer os.RemoveAll(destDir)
 	imgManifest, err := imgHandler.copyImage(destDir, os.Stdout)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	logging.Info("Extracting the image bundle layer...")
 	imgLayer, err := getLayerPath(imgManifest, 0, "application/vnd.oci.image.layer.v1.tar+gzip")
 	if err != nil {
-		return err
+		return "", err
 	}
 	fileList, err := extract.Uncompress(filepath.Join(destDir, imgLayer), constants.MachineCacheDir, true)
 	if err != nil {
-		return err
+		return "", err
 	}
 	logging.Debugf("Bundle and sign path: %v", fileList)
 
 	logging.Info("Verifying the bundle signature...")
 	if len(fileList) != 2 {
-		return fmt.Errorf("image layer contains more files than expected: %v", fileList)
+		return "", fmt.Errorf("image layer contains more files than expected: %v", fileList)
 	}
 	bundleFilePath, sigFilePath := fileList[0], fileList[1]
 	if !strings.HasSuffix(sigFilePath, ".crcbundle.sig") {
 		sigFilePath, bundleFilePath = fileList[0], fileList[1]
 	}
 
-	return gpg.Verify(bundleFilePath, sigFilePath)
+	if err := gpg.Verify(bundleFilePath, sigFilePath); err != nil {
+		return "", err
+	}
+
+	return bundleFilePath, nil
 }
--- a/pkg/crc/image/image.go
+++ b/pkg/crc/image/image.go
@@ -5,7 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -28,6 +30,21 @@ type imageHandler struct {
 
 func defaultURI(preset crcpreset.Preset) string {
 	return fmt.Sprintf("//%s/%s:%s", constants.RegistryURI, getImageName(preset), version.GetCRCVersion())
+}
+
+func ValidateURI(uri *url.URL) error {
+	/* For now we are very restrictive on the docker:// URLs we accept, it
+	 * has to contain the bundle version as the tag, and it has to use the
+	 * same image names as the official registry (openshift-bundle,
+	 * okd-bundle, podman-bundle). In future releases, we'll make this more
+	 * flexible
+	 */
+	imageAndTag := strings.Split(path.Base(uri.Path), ":")
+	if len(imageAndTag) != 2 {
+		return fmt.Errorf("invalid %s registry URL, tag is required (such as docker://quay.io/crcont/openshift-bundle:4.11.0)", uri)
+	}
+	_, err := getPresetNameE(imageAndTag[0])
+	return err
 }
 
 func (img *imageHandler) policyContext() (*signature.PolicyContext, error) {
@@ -93,6 +110,23 @@ func getImageName(preset crcpreset.Preset) string {
 	default:
 		return "openshift-bundle"
 	}
+}
+
+func getPresetNameE(imageName string) (crcpreset.Preset, error) {
+	switch imageName {
+	case "openshift-bundle":
+		return crcpreset.OpenShift, nil
+	case "okd-bundle":
+		return crcpreset.OKD, nil
+	case "podman-bundle":
+		return crcpreset.Podman, nil
+	default:
+		return crcpreset.OpenShift, fmt.Errorf("invalid image name '%s' (Should be openshift-bundle, okd-bundle or podman-bundle)", imageName)
+	}
+}
+func GetPresetName(imageName string) crcpreset.Preset {
+	preset, _ := getPresetNameE(imageName)
+	return preset
 }
 
 func PullBundle(preset crcpreset.Preset, imageURI string) error {
--- a/pkg/crc/image/image.go
+++ b/pkg/crc/image/image.go
@@ -29,7 +29,7 @@ type imageHandler struct {
 }
 
 func defaultURI(preset crcpreset.Preset) string {
-	return fmt.Sprintf("//%s/%s:%s", constants.RegistryURI, getImageName(preset), version.GetCRCVersion())
+	return fmt.Sprintf("//%s/%s:%s", constants.RegistryURI, getImageName(preset), version.GetBundleVersion(preset))
 }
 
 func ValidateURI(uri *url.URL) error {
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -286,23 +286,20 @@ func getBundleDownloadInfo(preset crcPreset.Preset) (*download.RemoteFile, error
 	return downloadInfo, nil
 }
 
-func DownloadDefault(preset crcPreset.Preset) error {
+func DownloadDefault(preset crcPreset.Preset) (string, error) {
 	downloadInfo, err := getBundleDownloadInfo(preset)
 	if err != nil {
-		return err
+		return "", err
 	}
 	bundleNameFromURI, err := downloadInfo.GetSourceFilename()
 	if err != nil {
-		return err
+		return "", err
 	}
 	if constants.GetDefaultBundle(preset) != bundleNameFromURI {
-		return fmt.Errorf("expected %s but found %s in Makefile", bundleNameFromURI, constants.GetDefaultBundle(preset))
-	}
-	if _, err := downloadInfo.Download(constants.GetDefaultBundlePath(preset), 0664); err != nil {
-		return err
+		return "", fmt.Errorf("expected %s but found %s in Makefile", bundleNameFromURI, constants.GetDefaultBundle(preset))
 	}
 
-	return nil
+	return downloadInfo.Download(constants.GetDefaultBundlePath(preset), 0664)
 }
 
 func Download(preset crcPreset.Preset, bundleURI string) (string, error) {
@@ -313,19 +310,15 @@ func Download(preset crcPreset.Preset, bundleURI string) (string, error) {
 	// bundles, their sha256sums are known and can be checked.
 	if bundleURI == constants.GetDefaultBundlePath(preset) {
 		if preset == crcPreset.OpenShift {
-			return bundleURI, DownloadDefault(preset)
+			return DownloadDefault(preset)
 		}
-		return bundleURI, image.PullBundle(preset, "")
+		return image.PullBundle(preset, "")
 	}
 	switch {
 	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
 		return download.Download(bundleURI, constants.MachineCacheDir, 0644, nil)
 	case strings.HasPrefix(bundleURI, "docker://"):
-		if err := image.PullBundle(preset, bundleURI); err != nil {
-			return "", err
-		}
-		bundlePath := filepath.Join(constants.MachineCacheDir, GetBundleNameFromURI(bundleURI))
-		return bundlePath, nil
+		return image.PullBundle(preset, bundleURI)
 	}
 	// the `bundleURI` parameter turned out to be a local path
 	return bundleURI, nil
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -270,7 +270,7 @@ func getBundleDownloadInfo(preset preset.Preset) (*download.RemoteFile, error) {
 	return downloadInfo, nil
 }
 
-func Download(preset preset.Preset) error {
+func DownloadDefault(preset preset.Preset) error {
 	downloadInfo, err := getBundleDownloadInfo(preset)
 	if err != nil {
 		return err
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -3,6 +3,7 @@ package bundle
 import (
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
@@ -12,7 +13,8 @@ import (
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/code-ready/crc/pkg/crc/constants"
-	"github.com/code-ready/crc/pkg/crc/preset"
+	"github.com/code-ready/crc/pkg/crc/image"
+	crcPreset "github.com/code-ready/crc/pkg/crc/preset"
 	"github.com/code-ready/crc/pkg/download"
 )
 
@@ -174,19 +176,19 @@ func (bundle *CrcBundleInfo) GetBundleNameWithoutExtension() string {
 	return GetBundleNameWithoutExtension(bundle.GetBundleName())
 }
 
-func (bundle *CrcBundleInfo) GetBundleType() preset.Preset {
+func (bundle *CrcBundleInfo) GetBundleType() crcPreset.Preset {
 	switch bundle.Type {
 	case "snc", "snc_custom":
-		return preset.OpenShift
+		return crcPreset.OpenShift
 	case "podman", "podman_custom":
-		return preset.Podman
+		return crcPreset.Podman
 	default:
-		return preset.OpenShift
+		return crcPreset.OpenShift
 	}
 }
 
 func (bundle *CrcBundleInfo) IsOpenShift() bool {
-	return bundle.GetBundleType() == preset.OpenShift
+	return bundle.GetBundleType() == crcPreset.OpenShift
 }
 
 func (bundle *CrcBundleInfo) verify() error {
@@ -250,10 +252,24 @@ func GetCustomBundleName(bundleFilename string) string {
 	return fmt.Sprintf("%s_%d%s", baseName, time.Now().Unix(), bundleExtension)
 }
 
-type presetDownloadInfo map[preset.Preset]*download.RemoteFile
+func GetBundleNameFromURI(bundleURI string) string {
+	// the URI is expected to have been validated by validation.ValidateBundlePath first
+	switch {
+	case strings.HasPrefix(bundleURI, "docker://"):
+		imageAndTag := strings.Split(path.Base(bundleURI), ":")
+		return constants.BundleForPreset(image.GetPresetName(imageAndTag[0]), imageAndTag[1])
+	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
+		return path.Base(bundleURI)
+	default:
+		// local path
+		return filepath.Base(bundleURI)
+	}
+}
+
+type presetDownloadInfo map[crcPreset.Preset]*download.RemoteFile
 type bundlesDownloadInfo map[string]presetDownloadInfo
 
-func getBundleDownloadInfo(preset preset.Preset) (*download.RemoteFile, error) {
+func getBundleDownloadInfo(preset crcPreset.Preset) (*download.RemoteFile, error) {
 	bundles, ok := bundleLocations[runtime.GOARCH]
 	if !ok {
 		return nil, fmt.Errorf("Unsupported architecture: %s", runtime.GOARCH)
@@ -270,7 +286,7 @@ func getBundleDownloadInfo(preset preset.Preset) (*download.RemoteFile, error) {
 	return downloadInfo, nil
 }
 
-func DownloadDefault(preset preset.Preset) error {
+func DownloadDefault(preset crcPreset.Preset) error {
 	downloadInfo, err := getBundleDownloadInfo(preset)
 	if err != nil {
 		return err
@@ -287,4 +303,30 @@ func DownloadDefault(preset preset.Preset) error {
 	}
 
 	return nil
+}
+
+func Download(preset crcPreset.Preset, bundleURI string) (string, error) {
+	// If we are asked to download
+	// ~/.crc/cache/crc_podman_libvirt_4.1.1.crcbundle, this means we want
+	// are downloading the default bundle for this release. This uses a
+	// different codepath from user-specified URIs as for the default
+	// bundles, their sha256sums are known and can be checked.
+	if bundleURI == constants.GetDefaultBundlePath(preset) {
+		if preset == crcPreset.OpenShift {
+			return bundleURI, DownloadDefault(preset)
+		}
+		return bundleURI, image.PullBundle(preset, "")
+	}
+	switch {
+	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
+		return download.Download(bundleURI, constants.MachineCacheDir, 0644, nil)
+	case strings.HasPrefix(bundleURI, "docker://"):
+		if err := image.PullBundle(preset, bundleURI); err != nil {
+			return "", err
+		}
+		bundlePath := filepath.Join(constants.MachineCacheDir, GetBundleNameFromURI(bundleURI))
+		return bundlePath, nil
+	}
+	// the `bundleURI` parameter turned out to be a local path
+	return bundleURI, nil
 }
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -43,13 +43,17 @@ import (
 
 const minimumMemoryForMonitoring = 14336
 
-func getCrcBundleInfo(bundleName, bundlePath string) (*bundle.CrcBundleInfo, error) {
+func getCrcBundleInfo(preset crcPreset.Preset, bundleName, bundlePath string) (*bundle.CrcBundleInfo, error) {
 	bundleInfo, err := bundle.Use(bundleName)
 	if err == nil {
 		logging.Infof("Loading bundle: %s...", bundleName)
 		return bundleInfo, nil
 	}
 	logging.Debugf("Failed to load bundle %s: %v", bundleName, err)
+	bundlePath, err = bundle.Download(preset, bundlePath)
+	if err != nil {
+		return nil, err
+	}
 	logging.Infof("Extracting bundle: %s...", bundleName)
 	if _, err := bundle.Extract(bundlePath); err != nil {
 		return nil, err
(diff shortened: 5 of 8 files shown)


cmd/crc/cmd/setup.go
```diff
@@ -58,7 +58,7 @@ func runSetup(arguments []string) error {
 	}
 
 	if !config.Get(crcConfig.Bundle).IsDefault {
-		if err := validation.ValidatePath(config.Get(crcConfig.Bundle).AsString()); err != nil {
+		if err := validation.ValidateBundlePath(config.Get(crcConfig.Bundle).AsString(), crcConfig.GetPreset(config)); err != nil {
```
praveenkumar, line 61: this is already part of https://github.com/code-ready/crc/pull/3315/commits/030b0275998fb7f545aa2476095790d9bcc51d28 and will be removed one other merge or I will cherry pick it.

pkg/crc/validation/validation.go
```diff
@@ -98,6 +101,14 @@ func ValidateIPAddress(ipAddress string) error {
 	return nil
 }
 
+func ValidateURL(inputString string) bool {
+	u, err := url.ParseRequestURI(inputString)
+	if u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "docker" {
+		return false
+	}
+	return err == nil
```
cfergeau, line 109: If `err` is non-`nil`, `u` is invalid, so you can't check `u.Scheme` before checking `err` value

pkg/crc/constants/constants.go
```diff
@@ -111,6 +112,19 @@ func GetDefaultBundle(preset crcpreset.Preset) string {
 	return bundles[runtime.GOOS]
 }
 
+func GetBundleNameFromBundlePath(preset crcpreset.Preset, bundlePath string) string {
+	switch {
+	case strings.HasPrefix(bundlePath, "http://"), strings.HasPrefix(bundlePath, "https://"):
+		return filepath.Base(bundlePath)
+	case strings.HasPrefix(bundlePath, "docker://"):
+		tag := strings.Split(bundlePath, ":")[2]
+		bundles := defaultBundleForOs(preset, tag)
```
cfergeau, line 121: This does not seem very flexible, this means I can't use `docker://quay.io/crcont/podman-bundle` (it might even crash?), or `docker://quay.io/crcont/openshift-bundle:latest-4.10`. https://github.com/code-ready/snc/pull/569 or https://github.com/code-ready/snc/issues/567 might be better than hardcoding this.
If for now we want to be very strict on the image tag and don't want to add labels, we can also say that the registry name must match the ones on `crcont` and do a registry name -> bundle name conversion rather than using `preset`

pkg/crc/constants/constants.go
```diff
@@ -83,31 +83,31 @@ func GetAdminHelperURL() string {
 	return GetAdminHelperURLForOs(runtime.GOOS)
 }
 
-func defaultBundleForOs(preset crcpreset.Preset) map[string]string {
+func defaultBundleForOs(preset crcpreset.Preset, bundleVersion string) map[string]string {
```
cfergeau, line 87: As soon as you pass it a version, it's no longer the `default` bundle for the OS.

pkg/crc/constants/constants.go
praveenkumar: > This does not seem very flexible, this means I can't use docker://quay.io/crcont/podman-bundle (it might even crash?)

This we can put part of validation when user provide an image without tag, we error out and tell user to use a tag for image. Will add that.

> or docker://quay.io/crcont/openshift-bundle:latest-4.10. 
This doesn't crash but then bundle extraction fail but this can happen with http location where user change the bundle name. Our first goal is to provide an opinionated solution where we tell user what not to use (initially with doc and then with automation). 

> If for now we want to be very strict on the image tag and don't want to add labels, we can also say that the registry name must match the ones on crcont and do a registry name -> bundle name conversion rather than using preset

@cfergeau yes, at this point we are to strict with image tag. I think initially we should go with that and once https://github.com/code-ready/snc/pull/569 merge then we can revisit again.

pkg/crc/constants/constants.go
praveenkumar: I will create an helper  `bundleForOs`  so that can be used with `defaultBundleForOs`.

pkg/crc/constants/constants.go
```diff
@@ -111,6 +112,19 @@ func GetDefaultBundle(preset crcpreset.Preset) string {
 	return bundles[runtime.GOOS]
 }
 
+func GetBundleNameFromBundlePath(preset crcpreset.Preset, bundlePath string) string {
```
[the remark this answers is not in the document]
praveenkumar: To be very generic this function dealing with `BundlePath` string which can be URI or local path and return the `BundleName` May be I will put `GetBundleNameFromBundlePathString` to make it more clear? Otherwise every place we have to first check if it use URI or path and then have to condition at every place to get the bundleName.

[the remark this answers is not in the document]
cfergeau: I'm only suggesting to replace 'path' by 'uri' in this function as it feels a bit more representative of the kind of data it handles.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -110,7 +110,11 @@ func fixBundleExtracted(bundlePath string, preset crcpreset.Preset) func() error
 					return err
 				}
 			}
-			bundlePath = constants.GetDefaultBundlePath(preset)
+		} else {
+			var err error
+			if bundlePath, err = bundle.Download(preset, bundlePath); err != nil {
+				return err
+			}
```
cfergeau, line 117: A bit confused that we still have 
```
			logging.Infof("Downloading %s", constants.GetDefaultBundle(preset))
			// In case of OKD or podman bundle then pull the bundle image from quay
			// otherwise use mirror location to download the bundle.
			if preset == crcpreset.OKD || preset == crcpreset.Podman {
				if err := image.PullBundle(preset, ""); err != nil {
					return err
				}
			} else {
				if err := bundle.DownloadDefault(preset); err != nil {
					return err
				}
			}
```
on top of this, this looks redundant? (as in, could these 2 codepaths become a single one?)

pkg/crc/validation/validation.go
```diff
@@ -98,6 +105,30 @@ func ValidateIPAddress(ipAddress string) error {
...
+	if err != nil {
+		return false
+	}
+	if u.Scheme != "http" && u.Scheme != "https" && u.Scheme != "docker" {
+		return false
+	}
+	return true
+}
+
+func validRegistryURI(inputString string) error {
```
cfergeau, line 119: Why is it not part of `validateURL`?

pkg/crc/validation/validation.go
```diff
@@ -98,6 +105,30 @@ func ValidateIPAddress(ipAddress string) error {
...
+	u, err := url.ParseRequestURI(inputString)
+	if err != nil {
+		return err
+	}
+	if u.Scheme == "docker" {
+		if len(strings.Split(inputString, ":")) != 3 {
+			return fmt.Errorf("%s doesn't contain the tag it should be like docker://quay.io/crcont/ocp-bundle:4.11.0", inputString)
```
cfergeau, line 126: maybe `Invalid docker:// registry URL, tag is required (such as docker://quay.io/crcont/ocp-bundle:4.11.0)`

cfergeau (commented):
How are we checking the integrity of images fetched over http/https? For the releases, we check the hardcoded sha256sum of the bundle, I don't think I've seen something to replace that with this new code?

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: @cfergeau yes, in next iteration we should have same codepath. Just using `bundle.Download`

pkg/crc/validation/validation.go
praveenkumar: @cfergeau As of now I am using `ValidateURL` only for scheme validation and if not vaildURI then I check for validpath.

pkg/crc/validation/validation.go
praveenkumar: Sure 👍🏼

pkg/crc/validation/validation.go
cfergeau: Yes, but why do it this way? Isn't it more convenient to do it all in one go in `ValidateURL`? Is there a drawback to doing this?

praveenkumar: @cfergeau No but then I have to check the validpath also as part of validURL function something like so need to change the function name something like `ValidateUrlOrPath` ?
```golang
func ValidateURL(inputString string) error {
	u, err := url.ParseRequestURI(inputString)
	if err != nil {
		return ValidatePath(inputString)
	}
	switch {
	case u.Scheme == "http", u.Scheme == "https":
		return nil
	case u.Scheme == "docker":
		if len(strings.Split(inputString, ":")) != 3 {
			return fmt.Errorf("invalid %s registry URL, tag is required (such as docker://quay.io/crcont/ocp-bundle:4.11.0)", inputString)
		}
		return nil
	default:
		return fmt.Errorf("invalid %s format (only supported http, https or docker)", inputString)
	}
	return nil
}
```

praveenkumar:
> How are we checking the integrity of images fetched over http/https? For the releases, we check the hardcoded sha256sum of the bundle, I don't think I've seen something to replace that with this new code?

No, We are expecting user are going to put the bundle on their file server and don't know if they are going to put the sha256sum or not. May be create an issue to discuss what we should do in term of user upload bundles?

cfergeau:
> We are expecting user are going to put the bundle on their file server

In some distant future yes, at this point, our goal is to be able to use different bundle versions hosted on mirror.openshift.com or quay.io/crcont/, or on exact mirrors of these locations. Once this is in place and works fine, we can extend it to make it more flexible.

pkg/crc/constants/constants.go
```diff
@@ -115,6 +116,19 @@ func GetDefaultBundle(preset crcpreset.Preset) string {
 	return bundles[runtime.GOOS]
 }
 
+func GetBundleNameFromBundlePathString(preset crcpreset.Preset, bundlePath string) string {
+	switch {
+	case strings.HasPrefix(bundlePath, "http://"), strings.HasPrefix(bundlePath, "https://"):
+		return filepath.Base(bundlePath)
+	case strings.HasPrefix(bundlePath, "docker://"):
+		tag := strings.Split(bundlePath, ":")[2]
+		bundles := bundleForOs(preset, tag)
```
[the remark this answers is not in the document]
praveenkumar: > If we are requiring the tag to match what we use on quay, then I'd prefer that we also require for now that the repo name matches the ones we use on quay

@cfergeau  you mean the org name `crcont` or only the container image name `ocp-bundle`/`okd-bundle`/`podman-bundle` ? Also even using those how do we able to build the bundle name 
`ocp-bundle` => bundle name should be `crc_<hypervisor>.<tag>.crcbundle` ?

Shouldn't current method is better in case user choose the preset `openshift` but chose image of `podman` then in current scenario we fail in case there is no podman image or pull the correct one but if we don't consider the `preset` then we are going to pull wrong image for wrong preset and error out after extraction.

praveenkumar:
> > We are expecting user are going to put the bundle on their file server
> 
> In some distant future yes, at this point, our goal is to be able to use different bundle versions hosted on mirror.openshift.com or quay.io/crcont/, or on exact mirrors of these locations. Once this is in place and works fine, we can extend it to make it more flexible.

I think we are going to do the release 1-2 more OCP bundle version to mirror and later just put it on quay and for images we already have signature checking in place. So I am not sure if we really want to solve that now for our own mirror side or just work on it distant future when user uses the file server.

pkg/crc/constants/constants.go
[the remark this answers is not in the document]
cfergeau: > Shouldn't current method is better in case user choose the preset openshift but chose image of podman then in current scenario we fail in case there is no podman image or pull the correct one but if we don't consider the preset then we are going to pull wrong image for wrong preset and error out after extraction.

What bothers me is that the behaviour is totally different from the behaviour of the other 2 methods (https and file). I'd prefer to have something consistent for all 3 methods, bundle filename comes from the URL, and nothing else.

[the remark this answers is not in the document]
praveenkumar: > What bothers me is that the behaviour is totally different from the behaviour of the other 2 methods (https and file). I'd prefer to have something consistent for all 3 methods, bundle filename comes from the URL, and nothing else.

I don't know how we can do with quay, in that case we explicitly need to ask user to have the image name as bundle filename on the mirror. Or we can use preset for https and file also for consistency to make sure user use the correct bundle of preset as part of validation. This function still be inconsistent but validation side we able to validate `https/file` method also depend on preset.

```
func validBundleForPreset(inputString string, preset crcpreset.Preset) error {
	if strings.HasPrefix(inputString, "docker://") || preset == crcpreset.OpenShift {
		return nil
	}
	if !strings.Contains(filepath.Base(inputString), preset.String()) {
		return fmt.Errorf("invalid %s bundle for %s preset", filepath.Base(inputString), preset)
	}
	return nil
}
```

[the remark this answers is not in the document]
cfergeau: Could be something like:
```
switch repoName {
    case "podman-bundle: preset = preset.Podman
    case "ocp-bundle": preset = preset.OpenShift
    ...
}
```
We already put requirements on the version tag, so we could also do this on repo name. This would then be made more flexible at a later point.

pkg/crc/validation/validation.go
cfergeau: You don't necessarily have to put `ValidatePath` in it, but I'd prefer to avoid having something which looks like `if ValidateURL() { if reallyValidateURL() {}}`, better to do all the URL validation at once.

pkg/crc/constants/constants.go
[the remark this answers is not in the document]
praveenkumar: @cfergeau Just for sake of consistency which is not even correct because as par current behavior user can choose different bundle with different preset and only get to know this metadata (after bundle extraction) I would better use preset validation for even http/file and tell that early on to user then wait for extraction.

pkg/crc/image/image.go
```diff
@@ -95,9 +95,12 @@ func getImageName(preset crcpreset.Preset) string {
 	}
 }
 
-func PullBundle(preset crcpreset.Preset) error {
+func PullBundle(preset crcpreset.Preset, imageURI string) error {
+	if imageURI == "" {
+		imageURI = uri(preset)
```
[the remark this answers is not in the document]
praveenkumar: you mean `imageURI` to defaultURI ? how does it default if user provide the value?

gbraad:
is there an alternative to using `docker://` as this does not seem to be an 'accepted' scheme. why not, like podman, docker, etc assume the address is a registry, unless http(s) is specified?

praveenkumar:
> is there an alternative to using `docker://` as this does not seem to be an 'accepted' scheme. why not, like podman, docker, etc assume the address is a registry, unless http(s) is specified?

This is something we are taking from `man containers-transports` which skopeo also uses.

> 
> Local can be enforced by `file://`

I don't want to force user to use `file://` for local now because then it will change the current behavior where user can just provide bundle path without `file://` and it just works.

cfergeau:
> > Local can be enforced by `file://`
> 
> I don't want to force user to use `file://` for local now because then it will change the current behavior where user can just provide bundle path without `file://` and it just works.

Yes, mandating the use of file:// would break backward compat. We should accept it though if it's specified, did not check if that's the case. We probably can use some heuristic though when the arg to -b does not start with `xxxx://`. If it starts with `/` or `./` then it's a path, if it starts with a letter then if it's only composed of letters/dots, assume it's a `docker://` URL, otherwise treat it as a path. When the heuristic is not correct, it's always possible to use a full URI to get the wanted behaviour.

pkg/crc/config/settings.go
```diff
@@ -27,7 +27,6 @@ const (
 	ProxyCAFile             = "proxy-ca-file"
 	ConsentTelemetry        = "consent-telemetry"
 	EnableClusterMonitoring = "enable-cluster-monitoring"
-	AutostartTray           = "autostart-tray"
```
gbraad, line 30: ?

looks unrelated

gbraad (changes requested):

pkg/crc/config/settings.go
praveenkumar: @gbraad different commit.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -281,3 +282,23 @@ func DownloadDefault(preset preset.Preset) error {
...
 	return nil
 }
+
+func Download(preset crcPreset.Preset, bundlePath string, isDefault bool) (string, error) {
+	if isDefault {
+		if preset == crcPreset.OpenShift {
+			return bundlePath, DownloadDefault(preset)
+		}
+		return bundlePath, image.PullBundle(preset, "")
+	}
```
cfergeau, line 292: The `isDefault` code was moved here, but this still duplicates the code in the `switch` below, so this still has the same issue as the previous code.

praveenkumar:
/hold 

Looks like during the PR we are again changing the requirement  like https://github.com/code-ready/crc/pull/3319#issuecomment-1230009359 and https://github.com/code-ready/crc/pull/3319#discussion_r957105779 . Do we really want to have `file://` support and also extracting the bundle metadata before pull bundle or download it as part of initial PR ? Should we limit the scope of this PR and then subsequently put other patches.

cfergeau:
> /hold
> 
> Looks like during the PR we are again changing the requirement like [#3319 (comment)](https://github.com/code-ready/crc/pull/3319#issuecomment-1230009359) and [#3319 (comment)](https://github.com/code-ready/crc/pull/3319#discussion_r957105779) . Do we really want to have `file://` support and also extracting the bundle metadata before pull bundle or download it as part of initial PR ? Should we limit the scope of this PR and then subsequently put other patches.

I did not want to imply these should be part of this PR, I was trying to shape what this feature could look like when it's complete. Imo the addition of `validBundleForPreset` is also a change from the initial requirements

praveenkumar:
> > /hold
> > Looks like during the PR we are again changing the requirement like [#3319 (comment)](https://github.com/code-ready/crc/pull/3319#issuecomment-1230009359) and [#3319 (comment)](https://github.com/code-ready/crc/pull/3319#discussion_r957105779) . Do we really want to have `file://` support and also extracting the bundle metadata before pull bundle or download it as part of initial PR ? Should we limit the scope of this PR and then subsequently put other patches.
> 
> I did not want to imply these should be part of this PR, I was trying to shape what this feature could look like when it's complete. 

I think we should discuss that on the feature issue or create new issue if something is not present. May be I should've created tasks for this feature and implemented it PR by PR.

> Imo the addition of `validBundleForPreset` is also a change from the initial requirements

https://github.com/code-ready/crc/pull/3319#discussion_r955824926 I only added because of this discussion, I thought it might be blocker for the PR.

> (on the other hand, the code in this PR might be easier to write/read if `-b` was handling _only_ URIs, with a small `if isLocalPath(bundlePath) { bundlePath = "file://"+filepath.Abs(bundlePath) }` adjustment during arg parsing)

If we use `"file://"+filepath.Abs(bundlePath` then are we again using `Download` method to copy/download it or just remove `file://` and return back the Abs path for extraction?

pkg/crc/machine/bundle/metadata.go
praveenkumar: @cfergeau Yes, May be this is something we can take in a follow up PR?

pkg/crc/validation/validation.go
```diff
@@ -98,6 +110,35 @@ func ValidateIPAddress(ipAddress string) error {
 	return nil
 }
 
+func ValidateURL(inputString string) error {
+	u, err := url.ParseRequestURI(inputString)
+	if err != nil {
+		logging.Debugf("Failed to parse url: %v", err)
+		return err
```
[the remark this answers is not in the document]
praveenkumar: Fixed now.

pkg/crc/validation/validation.go
```diff
@@ -98,6 +110,35 @@ func ValidateIPAddress(ipAddress string) error {
...
+	}
+}
+
+func validBundleForPreset(inputString string, preset crcpreset.Preset) error {
+	if strings.HasPrefix(inputString, "docker://") || preset == crcpreset.OpenShift {
+		return nil
+	}
+	if !strings.Contains(filepath.Base(inputString), preset.String()) {
+		return fmt.Errorf("invalid %s bundle for %s preset", filepath.Base(inputString), preset)
+	}
```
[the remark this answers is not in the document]
praveenkumar: > I'd prefer not to rush this and to take time to do it properly. I don't know why with an OpenShift preset it's ok to use `crc_okd_....crcbundle` or `crc_podman_....crcbundle`, 

This is handled, now we are checking the bundle prefix not just the preset string so that even hypervisor side also taken care.

> I don't know why `https://internalmirror/podman-dev-env/crc_okd....crcbundle` is acceptable, ...

I don't think we should force our user to tell what `org` they want to use what if the use existing org to push bundle there. What we can do is to just make sure the `image` name should be something we have `okd-bundle`, `openshift-bundle` ..etc.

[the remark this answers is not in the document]
cfergeau: > This is handled, now we are checking the bundle prefix not just the preset string so that even hypervisor side also taken care.

It's still quite inconsistent between  docker:// and the other schemes, which I don't like. For https/file, we trust the URL to get the bundle name, for docker://, we do something complicated, we infer the bundle name from the preset, totally ignoring the docker image name, but we still trust the version from it. I'd prefer if in this case to we also trusted the URL and went from `podman-bundle:4.1.1` to `crc_podman_libvirt_4.1.1.crcbundle` rather than `crc_$currentPreset_libvirt_4.1.1.crcbundle` I've tried implementing this in https://github.com/cfergeau/crc/commits/issue_3277

Then we can add a function checking if the name we built from the URL matches the name we expect given the current preset, and it will now be done in a similar way for docker/http/file. I've dropped it for now as imo it can be a followup.

pkg/crc/validation/validation.go
```diff
@@ -98,6 +110,35 @@ func ValidateIPAddress(ipAddress string) error {
...
+	u, err := url.ParseRequestURI(inputString)
+	if err != nil {
+		logging.Debugf("Failed to parse url: %v", err)
+		return err
+	}
+	switch {
+	case u.Scheme == "http", u.Scheme == "https":
+		return nil
+	case u.Scheme == "docker":
+		if len(strings.Split(inputString, ":")) != 3 {
```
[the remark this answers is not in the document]
cfergeau: I commented elsewhere, but is there any guarantee that filepath.Base() works on an URI? Looking at https://pkg.go.dev/path , you need `path.Base()` and not `filepath.Base()`, the latter seems like it won't do the right thing on Windows.
And https://pkg.go.dev/path says:
> The path package should only be used for paths separated by forward slashes, such as the paths in URLs

so I read it as something to use on just the path part of a parsed URL, not on the full URL.

pkg/crc/machine/bundle/metadata.go
cfergeau: I did some work on this in https://github.com/cfergeau/crc/commits/issue_3277 to try to remove at least the `isDefault` parameter which seems redundant.
Not sure there's a lot more we can rework, maybe some small `downloaderFromURI()` helper which would return opaque types with just a `Download()` method and a `HasIntegrityCheck()`, and these opaque types would wrap either a DownloadInfo for default bundles, or `PullBundle` for container images, or `Download` for http URLs.
That might also be a good opportunity to push the default container registry URLs to `bundleLocations` rather than maintaining it in a different part of the code.

pkg/crc/validation/validation.go
```diff
@@ -101,6 +114,39 @@ func ValidateIPAddress(ipAddress string) error {
...
 
+func ValidateURL(uri string) error {
+	u, err := url.ParseRequestURI(uri)
+	if err != nil {
+		logging.Debugf("Failed to parse url: %v", err)
+		return err
+	}
+	// If uri string is without scheme then check if it is valid path
+	if !u.IsAbs() {
+		return ValidatePath(uri)
```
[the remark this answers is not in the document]
praveenkumar: https://go.dev/play/p/oCboAo7HPBI as per `url.ParseRequestURI()` it consider `/tmp/bundle` as valid URI but `bundle` as invalid so with `u.IsAbs()` I can check if is abs path and then validate the path but in case user have bundle at the same location as crc binary and don't use absolute path then URI validation fails and I have to again check it with `ValidatePath`. With current logic both required.

pkg/crc/validation/validation.go
```diff
@@ -56,11 +57,23 @@ func ValidateEnoughMemory(value int) error {
 
 // ValidateBundlePath checks if the provided bundle path exist
 func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
-	if err := ValidatePath(bundlePath); err != nil {
+	logging.Debugf("Got bundle path: %s", bundlePath)
+	if err := ValidateURL(bundlePath); err != nil {
+		var urlError *url.Error
+		if errors.As(err, &urlError) {
+			if err1 := ValidatePath(bundlePath); err1 != nil {
+				return err1
```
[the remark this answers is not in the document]
praveenkumar: same as https://github.com/code-ready/crc/pull/3319#discussion_r960284696

pkg/crc/validation/validation.go
```diff
@@ -56,11 +57,23 @@ func ValidateEnoughMemory(value int) error {
 
 // ValidateBundlePath checks if the provided bundle path exist
 func ValidateBundlePath(bundlePath string, preset crcpreset.Preset) error {
```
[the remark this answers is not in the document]
praveenkumar: May be another PR, not changing to much around naming also it can be a path to disk or URI.

pkg/crc/validation/validation.go
```diff
@@ -101,6 +114,39 @@ func ValidateIPAddress(ipAddress string) error {
...
+	}
+	// If uri string is without scheme then check if it is valid path
+	if !u.IsAbs() {
+		return ValidatePath(uri)
+	}
+	switch {
+	case u.Scheme == "http", u.Scheme == "https":
+		return nil
+	case u.Scheme == "docker":
+		if len(strings.Split(filepath.Base(uri), ":")) != 2 {
```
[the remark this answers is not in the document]
praveenkumar: https://go.dev/play/p/OvulM6Hnbfb `filepath.Base()` is always return last element of the path with path separator, all uri methods doesn't have it.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -281,3 +282,24 @@ func DownloadDefault(preset preset.Preset) error {
...
+		if preset == crcPreset.OpenShift {
+			return bundleURI, DownloadDefault(preset)
+		}
+		return bundleURI, image.PullBundle(preset, "")
+	}
+	switch {
+	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
+		return download.Download(bundleURI, constants.MachineCacheDir, 0644, nil)
+	case strings.HasPrefix(bundleURI, "docker://"):
+		if err := image.PullBundle(preset, strings.TrimPrefix(bundleURI, "docker:")); err != nil {
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau doing it here because we are already validating if `docker://` is prefix, if we do it in the `PullBundle` then we again need to verify it. also `PullBundle` have defaultURI which doesn't contain `docker:` as prefix.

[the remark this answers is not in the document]
cfergeau: code/API would look nicer if `image.PullBundle` directly accepted `docker://`  URL. Yes some code would likely be a bit redundant, but trimming `docker://` feels like a better fit in `PullBundle` rather than expecting all callers to do it themselves.

pkg/crc/config/settings.go
gbraad: Still, this could be a PR that would have been merged instantly

pkg/crc/image/image.go
```diff
@@ -95,39 +112,63 @@ func getImageName(preset crcpreset.Preset) string {
 	}
 }
 
-func PullBundle(preset crcpreset.Preset) error {
+func getPresetNameE(imageName string) (crcpreset.Preset, error) {
```
gbraad, line 115: `getPresetNameE`; What does the `E` suffix mean here?

pkg/crc/image/image.go
```diff
@@ -95,39 +112,63 @@ func getImageName(preset crcpreset.Preset) string {
...
+	default:
+		return crcpreset.OpenShift, fmt.Errorf("invalid image name '%s' (Should be openshift-bundle, okd-bundle or podman-bundle)", imageName)
+	}
+}
+func GetPresetName(imageName string) crcpreset.Preset {
+	preset, _ := getPresetNameE(imageName)
+	return preset
+}
+
+func PullBundle(preset crcpreset.Preset, imageURI string) (string, error) {
```
gbraad, line 132: this function does not only pull, the bundle, but handles pull, uncompress, verify, etc. The naming is not clear

pkg/crc/machine/bundle/metadata.go
gbraad: should download not just pull the bundle? or perhaps pull the verification of the gpg key out and place this more logically as part of last step here?

pkg/crc/machine/bundle/metadata.go
```diff
@@ -263,21 +275,40 @@ func getBundleDownloadInfo(preset preset.Preset) (*download.RemoteFile, error) {
...
+	// bundles, their sha256sums are known and can be checked.
+	if bundleURI == constants.GetDefaultBundlePath(preset) {
+		if preset == crcPreset.OpenShift {
+			return DownloadDefault(preset)
+		}
+		return image.PullBundle(preset, "")
+	}
+	switch {
+	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
+		return download.Download(bundleURI, constants.MachineCacheDir, 0644, nil)
```
gbraad, line 308: `download.Download`, sounds more like `http.DownloadBundle`

pkg/crc/machine/bundle/metadata.go
```diff
@@ -263,21 +275,40 @@ func getBundleDownloadInfo(preset preset.Preset) (*download.RemoteFile, error) {
...
+	case strings.HasPrefix(bundleURI, "docker://"):
+		return image.PullBundle(preset, strings.TrimPrefix(bundleURI, "docker:"))
```
gbraad, line 310: ... as this also `image.PullBundle` sounds more like `registry` or even `docker.DownloadBundle`.

Not sure if the package separation follows logical naming, especially as 'download' is now a scheme-specific solution to 'download'.

pkg/crc/machine/start.go
```diff
@@ -43,13 +43,17 @@ import (
...
 
-func getCrcBundleInfo(bundleName, bundlePath string) (*bundle.CrcBundleInfo, error) {
+func getCrcBundleInfo(preset crcPreset.Preset, bundleName, bundlePath string) (*bundle.CrcBundleInfo, error) {
 	bundleInfo, err := bundle.Use(bundleName)
 	if err == nil {
 		logging.Infof("Loading bundle: %s...", bundleName)
 		return bundleInfo, nil
 	}
 	logging.Debugf("Failed to load bundle %s: %v", bundleName, err)
+	bundlePath, err = bundle.Download(preset, bundlePath)
```
gbraad, line 53: I am not sure if a download should be started from a `getBundleInfo` function. Sounds like a responsibility it shouldn't have. It fails, as there is no info available...

in that case download and try again... but perhaps the check of the bundle exists should be done before you check for the bundle info.

pkg/crc/image/image.go
praveenkumar: @gbraad `E` means with error. This naming format is heavily used in spf13/cast/caste.go and also some of our codebase.

pkg/crc/machine/bundle/metadata.go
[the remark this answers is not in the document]
praveenkumar: Done

pkg/crc/machine/start.go
praveenkumar: @gbraad Our assumption is user run the `crc setup` and that will automatic download the bundle and use but in case user want to try out different bundle without setup since other per-requisite are already in place then we use this to provide similar experience. Validation happen on cmd/start.go side where we show the warning but once we are in the pkg level we don't validate again and try to use what user provided and extract if it is a path or download it.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -250,10 +251,21 @@ func GetCustomBundleName(bundleFilename string) string {
 	return fmt.Sprintf("%s_%d%s", baseName, time.Now().Unix(), bundleExtension)
 }
 
-type presetDownloadInfo map[preset.Preset]*download.RemoteFile
+func GetBundleNameFromURI(bundleURI string) string {
+	// the URI is expected to have been validated by validation.ValidateBundlePath first
+	if strings.HasPrefix(bundleURI, "docker://") {
+		imageAndTag := strings.Split(filepath.Base(bundleURI), ":")
+		return constants.BundleForPreset(image.GetPresetName(imageAndTag[0]), imageAndTag[1])
+	}
```
cfergeau, line 259: the `strings.Split` might deserver a helper in `image`, but that can be done next time someone touches this code.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -97,20 +96,14 @@ func fixBundleExtracted(bundlePath string, preset crcpreset.Preset) func() error
...
-				if err := bundle.DownloadDefault(preset); err != nil {
-					return err
-				}
-			}
-			bundlePath = constants.GetDefaultBundlePath(preset)
+			isDefaultBundle = true
+		}
+		var err error
+		if bundlePath, err = bundle.Download(preset, bundlePath, isDefaultBundle); err != nil {
+			return err
```
cfergeau, line 102: `Remove isDefault argument from bundle.Download`  and ` Extend -b option to accept docker/http/https scheme ` can be squashed together imo.

pkg/crc/validation/validation.go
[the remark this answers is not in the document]
cfergeau: Seems to be another case where normalizing the `-b` argument value early one would let us simplify the rest of the code. Early on, we could make sure the `-b` arg is an absolute path, and turn it into a file:// URI. This way we no longer have to have multiple code paths for URI, absolute paths, relative paths.

In the mean time, a comment would be useful such as
```
// Some local paths (for example relative paths) can't be parsed/validated by `ValidateURL` and will be validated here
```

pkg/crc/validation/validation.go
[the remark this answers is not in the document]
cfergeau: I'd add this to the comment 
```
// If uri string is without scheme then check if it is a valid absolute path.
// Relative paths will cause `ParseRequestURI` to error out, and will be handled in `ValidateBundlePath`
```

What about Windows paths, `C:\bundles\...`?

cfergeau (commented):
A few Windows corner cases/questions left, but looking fine overall!

pkg/crc/machine/bundle/metadata.go
cfergeau: When developing some new code, it's useful to cleanup preexisting code which is related to the code we are changing while this code path is fresh in our minds :)

pkg/crc/image/image.go
```diff
@@ -29,7 +29,7 @@ type imageHandler struct {
 }
 
 func defaultURI(preset crcpreset.Preset) string {
-	return fmt.Sprintf("//%s/%s:%s", constants.RegistryURI, getImageName(preset), version.GetCRCVersion())
+	return fmt.Sprintf("//%s/%s:%s", constants.RegistryURI, getImageName(preset), version.GetBundleVersion(preset))
```
cfergeau, line 32: Nit: There's an unwanted leftover in the commit log `#	изменено:      pkg/crc/image/image.go`

pkg/crc/config/settings.go
```diff
@@ -59,7 +59,8 @@ func RegisterSettings(cfg *Config) {
 		fmt.Sprintf("Virtual machine preset (valid values are: %s, %s and %s)", preset.Podman, preset.OpenShift, preset.OKD))
 	// Start command settings in config
 	cfg.AddSetting(Bundle, defaultBundlePath(cfg), validateBundlePath, SuccessfullyApplied,
-		fmt.Sprintf("Bundle path (string, default '%s')", defaultBundlePath(cfg)))
+		fmt.Sprintf("Bundle path (string, like 'https://foo.com/%s', 'docker://quay.io/myorg/%s:%s' default '%s' )",
```
cfergeau, line 64: It's called 'bundle path' and the example make it look as if it only accepts URIs :-/ Not sure it's worth having the example, 
`Bundle path/URI (absolute or local path, http, https or docker URI)`

pkg/crc/machine/bundle/metadata.go
```diff
@@ -250,10 +252,23 @@ func GetCustomBundleName(bundleFilename string) string {
...
-type presetDownloadInfo map[preset.Preset]*download.RemoteFile
+func GetBundleNameFromURI(bundleURI string) string {
+	// the URI is expected to have been validated by validation.ValidateBundlePath first
+	switch {
+	case strings.HasPrefix(bundleURI, "docker://"):
+		imageAndTag := strings.Split(path.Base(bundleURI), ":")
+		return constants.BundleForPreset(image.GetPresetName(imageAndTag[0]), imageAndTag[1])
+	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
+		return path.Base(bundleURI)
+	default:
```
cfergeau, line 263: A `// local path` comment would ease reading.

pkg/crc/validation/validation.go
```diff
@@ -100,6 +110,27 @@ func ValidateIPAddress(ipAddress string) error {
...
+func ValidateURL(uri string) error {
+	u, err := url.ParseRequestURI(uri)
+	if err != nil {
+		logging.Debugf("Failed to parse url: %v", err)
+		return err
+	}
+	// If uri string is without scheme then check if it is a valid absolute path.
+	// Relative paths will cause `ParseRequestURI` to error out, and will be handled in `ValidateBundlePath`
+	if !u.IsAbs() {
+		return ValidatePath(uri)
```
cfergeau, line 122: This could/should be part of the `switch` to keep everything in the same place.

cfergeau (approved):

pkg/crc/image/image.go
praveenkumar: removed

pkg/crc/config/settings.go
praveenkumar: Added

pkg/crc/machine/bundle/metadata.go
praveenkumar: Added

PR #3325, opened 2022-08-29
doc: Clarify when no_proxy=.testing needs to be set

### LINKED ISSUE #2726: docs: Document that `no_proxy` needs to be set for `.testing` not just for `oc` use
After configuring my cluster with an external proxy, I wanted to run `curl -k -I api.crc.testing`, but this failed as I had forgotten to add `crc.testing` to `no_proxy` first.

Our documentation only says:
> To use an existing oc executable on your host machine, export the .testing domain as part of the no_proxy environment variable.

It is not needed only for `oc`, but for any cluster interaction.

cfergeau:
The current documentation can be misleading and make people think it's
only needed when using `oc`

This fixes #2726

### DIFF
--- a/docs/source/topics/proc_starting-behind-proxy.adoc
+++ b/docs/source/topics/proc_starting-behind-proxy.adoc
@@ -10,7 +10,7 @@ SOCKS proxies are not supported by {ocp}.
 
 .Prerequisites
 
-* To use an existing {openshift} CLI ([command]`oc`) executable on your host machine, export the `.testing` domain as part of the `no_proxy` environment variable.
+* If you are not using [command]`crc oc-env`, when interacting with the cluster, export the `.testing` domain as part of the `no_proxy` environment variable.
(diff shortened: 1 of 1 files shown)


docs/source/topics/proc_starting-behind-proxy.adoc
```diff
@@ -10,7 +10,7 @@ SOCKS proxies are not supported by {ocp}.
 
 .Prerequisites
 
-* To use an existing {openshift} CLI ([command]`oc`) executable on your host machine, export the `.testing` domain as part of the `no_proxy` environment variable.
+* If you are not using [command]`crc oc-env`, when interacting with the cluster, export the `.testing` domain as part of the `no_proxy` environment variable.
```
praveenkumar, line 13: 👍🏼

praveenkumar (approved):

PR #3329, opened 2022-08-31
Rename *_nonwin.go to *_unix.go

cfergeau:
This makes us consistent with the naming used in the go codebase,
removes a _nonwin.go/_nowin.go inconsistency, and is a step towards the
new 'unix' build constraint in go 1.19


praveenkumar:
@cfergeau I thought we are changing that once switch to go-1.19 does it supported same way for 1.17 (which we currently use)? I can see the CI is green but when it is become supported ( link to commit log might be good).

cfergeau:
> @cfergeau I thought we are changing that once switch to go-1.19 does it supported same way for 1.17 (which we currently use)? I can see the CI is green but when it is become supported ( link to commit log might be good).

The current file naming is not interpreted at all by go, we could append `_foobar` instead of `_unix` and it would still build/run fine. The build tags at the beginning of the file are what golang uses to know what to build/not to build:
```
//go:build !windows
// +build !windows
```

praveenkumar:
> > @cfergeau I thought we are changing that once switch to go-1.19 does it supported same way for 1.17 (which we currently use)? I can see the CI is green but when it is become supported ( link to commit log might be good).
> 
> The current file naming is not interpreted at all by go, we could append `_foobar` instead of `_unix` and it would still build/run fine. The build tags at the beginning of the file are what golang uses to know what to build/not to build:
> 
> ```
> //go:build !windows
> // +build !windows
> ```

And that's the actual reason why CI didn't fail :), looks like we were using file extension and also build tags and in long run we might need to use build tags.

cfergeau:
> looks like we were using file extension

These file extensions are not used, it's just that some name was needed when you have `file_windows.go` with Windows-specific code, and a file for the non-Windows code was needed, so we named it `file_nonwin.go` to be similar to `file_windows.go`. However, go introduced the `_unix` suffix, so we can as well  go with this one.

Issue #3331, opened 2022-09-02
Strategy to implement a different http/https port for user mode networking


praveenkumar:
This issue is to brainstorm how should we go to resolve https://github.com/code-ready/crc/issues/2760 and provide a way to have different http/https port mapped to the host system when `80` and `443` is in use. 

1. Ask user to provide it: In this case we provide config options `http-port/https-port` and using `crc config set` user can chose the port which they want to use.

2. Assign random port in case required port in use:  We first try to use `80/443` but if mapping fail because these ports are used by other process/program then create a random port and assign it. Also put a Info/Warn message that instead of default one now this random port is mapped.

Any other thoughts?

cfergeau:
We can start with 2., and add a config option (1.) later if 2. is not enough. And I would not use 'random' port numbers, but try a selection of 'readable' ones (8000, 8080, ...). Regarding port 80, I'm not sure how important it is, I was under the impression most of the services are (also?) exposed on port 443? Maybe we can make exposing port 80 optional?

One thing to keep in mind is that all the URLs given by oc (oc expose, oc get routes, ...) will need to be manually adjusted before being usable when the port is remapped.

praveenkumar:
> We can start with 2., and add a config option (1.) later if 2. is not enough. And I would not use 'random' port numbers, but try a selection of 'readable' ones (8000, 8080, ...).

Yes, `80 => 8080` and `443 => 8443` on host.

> Regarding port 80, I'm not sure how important it is, I was under the impression most of the services are (also?) exposed on port 443? Maybe we can make exposing port 80 optional?

I am also not sure if exposing is 80 really required, need to dig bit more. When you say optional, you mean to have config option or using some different method?

> 
> One thing to keep in mind is that all the URLs given by oc (oc expose, oc get routes, ...) will need to be manually adjusted before being usable when the port is remapped.

Yes that should be there. We  should put a `warn` message during the start if we change the port or may be in the documents?

cfergeau:
> When you say optional, you mean to have config option or using some different method?

If we fail to expose port 80 (even as 8080, ...), we could ignore the error, and continue startup instead of stopping and reporting a failure.

>  We should put a warn message during the start if we change the port or may be in the documents?

We'll need to do both, but I'm quite sure it's not going to be enough, and we'll get questions about it

praveenkumar:
[...]

PR #3332, opened 2022-09-05
Vsock: Add alternate http/https port in case 80/443 in use

praveenkumar:
This patch try to solve user issue where port 80 or 443 already in
use and `crc start` fails by having alternate port 8080 and 8443.

Note: Now user have to add those ports to application routes to access
it. Like `https://myapp.apps-crc.testing:8443` or `http://myapp.apps-crc.testing:8080`

```
$ ./crc start
[...]
WARN Port 80 is already in use. Going to use 8080

$ curl --unix-socket ~/.crc/crc-http.sock http:/unix/network/services/forwarder/all | jq .
[
  {
    "local": "127.0.0.1:2222",
    "remote": "192.168.127.2:22",
    "protocol": "tcp"
  },
  {
    "local": "127.0.0.1:6443",
    "remote": "192.168.127.2:6443",
    "protocol": "tcp"
  },
  {
    "local": ":443",
    "remote": "192.168.127.2:443",
    "protocol": "tcp"
  },
  {
    "local": ":8080",
    "remote": "192.168.127.2:80",
    "protocol": "tcp"
  }
]
```

**Fixes:** Issue #3331

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"runtime"
+	"strings"
 	"text/template"
 
 	"github.com/Masterminds/semver/v3"
@@ -75,6 +76,8 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		KubeAdminPassword: config.Get(crcConfig.KubeAdminPassword).AsString(),
 		Preset:            crcConfig.GetPreset(config),
 		EnableSharedDirs:  crcConfig.ShouldEnableSharedDirs(config),
+		IngressHTTPPort:   config.Get(crcConfig.IngressHTTPPort).AsUInt(),
+		IngressHTTPSPort:  config.Get(crcConfig.IngressHTTPSPort).AsUInt(),
 	}
 
 	client := newMachine()
@@ -249,10 +252,11 @@ If you find an issue, please report it at https://github.com/openshift/okd
 )
 
 type templateVariables struct {
-	ClusterConfig     *clusterConfig
-	EvalCommandLine   string
-	CommandLinePrefix string
-	PodmanRemote      string
+	ClusterConfig       *clusterConfig
+	EvalCommandLine     string
+	CommandLinePrefix   string
+	PodmanRemote        string
+	FallbackPortWarning string
 }
 
 func writeTemplatedMessage(writer io.Writer, s *startResult) error {
@@ -267,6 +271,10 @@ func writeOpenShiftTemplatedMessage(writer io.Writer, s *startResult) error {
 	tmpl := startTemplateForOpenshift
 	if s.ClusterConfig.ClusterType == preset.OKD {
 		tmpl = fmt.Sprintf("%s\n\n%s", startTemplateForOpenshift, startTemplateForOKD)
+	}
+	fallbackPortWarning := portFallbackWarning()
+	if fallbackPortWarning != "" {
+		tmpl = fmt.Sprintf("%s\n%s", tmpl, fallbackPortWarning)
 	}
 	parsed, err := template.New("template").Parse(tmpl)
 	if err != nil {
@@ -320,4 +328,27 @@ func checkDaemonStarted() error {
 		return err
 	}
 	return daemonclient.CheckIfOlderVersion(v)
+}
+
+func portFallbackWarning() string {
+	var fallbackPortWarning string
+	httpScheme := "http"
+	httpsScheme := "https"
+	fallbackPortWarningTmpl := `
+Warning: Port %d is used for OpenShift %s routes instead of the default
+You have to add this port to %s URLs when accessing OpenShift application,
+such as %s://myapp.apps-crc.testing:%d
+`
+	ingressHTTPPort := config.Get(crcConfig.IngressHTTPPort).AsUInt()
+	ingressHTTPSPort := config.Get(crcConfig.IngressHTTPSPort).AsUInt()
+
+	if ingressHTTPSPort != constants.OpenShiftIngressHTTPSPort {
+		fallbackPortWarning += fmt.Sprintf(fallbackPortWarningTmpl,
+			ingressHTTPSPort, strings.ToUpper(httpsScheme), strings.ToUpper(httpsScheme), httpsScheme, ingressHTTPSPort)
+	}
+	if ingressHTTPPort != constants.OpenShiftIngressHTTPPort {
+		fallbackPortWarning += fmt.Sprintf(fallbackPortWarningTmpl,
+			ingressHTTPPort, strings.ToUpper(httpScheme), strings.ToUpper(httpScheme), httpScheme, ingressHTTPPort)
+	}
+	return fallbackPortWarning
 }
--- a/pkg/crc/config/callbacks.go
+++ b/pkg/crc/config/callbacks.go
@@ -32,3 +32,16 @@ func RequiresCRCSetup(key string, _ interface{}) string {
 	return fmt.Sprintf("Changes to configuration property '%s' are only applied during 'crc setup'.\n"+
 		"Please run 'crc setup' for this configuration to take effect.", key)
 }
+
+func RequiresHTTPPortChangeWarning(key string, value interface{}) string {
+	return fmt.Sprintf("Changes to configuration property '%s' will break OpenShift HTTP routes.\n"+
+		"In order to access OpenShift applications through HTTP URLs "+
+		"the %d port must be manually specified, such as http://myapp.apps-crc.testing:%d", key, value, value)
+}
+
+func RequiresHTTPSPortChangeWarning(key string, value interface{}) string {
+	return fmt.Sprintf("Changes to configuration property '%s' will break OpenShift HTTPS routes.\n"+
+		"In order to access OpenShift applications through HTTPS URLs "+
+		"the %d port must be manually specified, such as https://myapp.apps-crc.testing:%d\n"+
+		"After this change, the OpenShift console will be non-functional because of OpenShift limitations", key, value, value)
+}
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -31,6 +31,8 @@ const (
 	KubeAdminPassword       = "kubeadmin-password"
 	Preset                  = "preset"
 	EnableSharedDirs        = "enable-shared-dirs"
+	IngressHTTPPort         = "ingress-http-port"
+	IngressHTTPSPort        = "ingress-https-port"
 )
 
 func RegisterSettings(cfg *Config) {
@@ -107,6 +109,10 @@ func RegisterSettings(cfg *Config) {
 
 	cfg.AddSetting(KubeAdminPassword, "", ValidateString, SuccessfullyApplied,
 		"User defined kubeadmin password")
+	cfg.AddSetting(IngressHTTPPort, constants.OpenShiftIngressHTTPPort, ValidatePort, RequiresHTTPPortChangeWarning,
+		fmt.Sprintf("HTTP port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPPort))
+	cfg.AddSetting(IngressHTTPSPort, constants.OpenShiftIngressHTTPSPort, ValidatePort, RequiresHTTPSPortChangeWarning,
+		fmt.Sprintf("HTTPS port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPSPort))
 }
 
 func defaultCPUs(cfg Storage) int {
--- a/pkg/crc/config/types.go
+++ b/pkg/crc/config/types.go
@@ -39,6 +39,10 @@ func (v SettingValue) AsInt() int {
 	return cast.ToInt(v.Value)
 }
 
+func (v SettingValue) AsUInt() uint {
+	return cast.ToUint(v.Value)
+}
+
 // validationFnType takes the key, value as args and checks if valid
 type ValidationFnType func(interface{}) (bool, string)
 type SetFn func(string, interface{}) string
--- a/pkg/crc/config/validations.go
+++ b/pkg/crc/config/validations.go
@@ -127,3 +127,14 @@ func validatePreset(value interface{}) (bool, string) {
 	}
 	return true, ""
 }
+
+func ValidatePort(value interface{}) (bool, string) {
+	port, err := cast.ToUintE(value)
+	if err != nil {
+		return false, "Requires integer value in range of 1024-65535"
+	}
+	if port < 1024 || port > 65535 {
+		return false, fmt.Sprintf("Provided %d but requires value in range of 1024-65535", port)
+	}
+	return true, ""
+}
--- a/pkg/crc/machine/types/types.go
+++ b/pkg/crc/machine/types/types.go
@@ -31,6 +31,10 @@ type StartConfig struct {
 
 	// Shared dirs
 	EnableSharedDirs bool
+
+	// Ports to access openshift routes
+	IngressHTTPPort  uint
+	IngressHTTPSPort uint
 }
 
 type ClusterConfig struct {
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -16,8 +16,8 @@ import (
 	"github.com/pkg/errors"
 )
 
-func exposePorts(preset crcPreset.Preset) error {
-	portsToExpose := vsockPorts(preset)
+func exposePorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint) error {
+	portsToExpose := vsockPorts(preset, ingressHTTPPort, ingressHTTPSPort)
 	daemonClient := daemonclient.New()
 	alreadyOpenedPorts, err := listOpenPorts(daemonClient)
 	if err != nil {
@@ -78,13 +78,13 @@ const (
 	virtualMachineIP = "192.168.127.2"
 	internalSSHPort  = "22"
 	localIP          = "127.0.0.1"
-	httpPort         = "80"
-	httpsPort        = "443"
+	remoteHTTPPort   = "80"
+	remoteHTTPSPort  = "443"
 	apiPort          = "6443"
 	cockpitPort      = "9090"
 )
 
-func vsockPorts(preset crcPreset.Preset) []types.ExposeRequest {
+func vsockPorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint) []types.ExposeRequest {
 	exposeRequest := []types.ExposeRequest{
 		{
 			Protocol: "tcp",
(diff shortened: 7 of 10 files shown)


pkg/crc/machine/vsock.go
```diff
@@ -146,3 +139,39 @@ func getSSHTunnelURI() string {
 	}
 	return u.String()
 }
+
+func HTTPAndHTTPS() []types.ExposeRequest {
+	hPort, hsPort := httpPort, httpsPort
+	if isPortInUse(hPort) {
+		logging.Warnf("Port %s is already in use. Going to use %s", httpPort, alternateHTTPPort)
```
cfergeau, line 146: `Using port %s instead` might sound nicer.

pkg/crc/machine/vsock.go
```diff
@@ -146,3 +139,39 @@ func getSSHTunnelURI() string {
...
+		{
+			Protocol: "tcp",
+			Local:    fmt.Sprintf(":%s", hPort),
+			Remote:   net.JoinHostPort(virtualMachineIP, httpPort),
+		},
+	}
+}
+
+func isPortInUse(port string) bool {
+	ln, err := net.Listen("tcp", ":"+port)
```
cfergeau, line 170: This could use `fmt.Sprintf` for consistency with the code before it

pkg/crc/machine/vsock.go
cfergeau: Is this the very last warning which is printed? Or is this going to be lost in the middle of the startup logs of crc?

pkg/crc/machine/vsock.go
```diff
@@ -146,3 +139,39 @@ func getSSHTunnelURI() string {
...
+		{
+			Protocol: "tcp",
+			Local:    fmt.Sprintf(":%s", hsPort),
+			Remote:   net.JoinHostPort(virtualMachineIP, httpsPort),
+		},
+		{
+			Protocol: "tcp",
+			Local:    fmt.Sprintf(":%s", hPort),
+			Remote:   net.JoinHostPort(virtualMachineIP, httpPort),
+		},
```
cfergeau, line 165: I'd have a preference for *not* putting http/https together to avoid code duplication.
```
exposeRequest = append(exposeRequest, requestWithAlternatePort(httpPort, alternateHTTPPort)
exposeRequest = append(exposeRequest, requestWithAlternatePort(httpsPort, alternateHTTPSPort)
```

pkg/crc/machine/vsock.go
cfergeau: Fwiw, listening on port 80/443 is a privileged operation on some OS, I don't know if `isPortInUse` could fail because of this.

cfergeau (commented):
```
Note: Now user have to add those ports to application routes to access
it. Like https://myapp.apps-crc.testing:8443 or http://myapp.apps-crc.testing:8080
```
This should be more than a note in the commit log, but part of the warning which is displayed to the user. This also needs to be mentioned in documentation.

gbraad:
> WARN Port 80 is already in use. Going to use 8080

`WARN Port 80 is already in use. Assigned 8080 as alternative.`

pkg/crc/machine/vsock.go
```diff
@@ -75,13 +75,15 @@ func listOpenPorts(daemonClient *daemonclient.Client) ([]types.ExposeRequest, er
...
-	localIP          = "127.0.0.1"
-	httpPort         = "80"
-	httpsPort        = "443"
-	apiPort          = "6443"
-	cockpitPort      = "9090"
+	virtualMachineIP   = "192.168.127.2"
+	internalSSHPort    = "22"
+	localIP            = "127.0.0.1"
+	httpPort           = "80"
+	alternateHTTPPort  = "8080"
```
gbraad, line 82: most common use of alternate is 'every other'. better use `alternative`

pkg/crc/machine/vsock.go
gbraad: `Going to use` does not sound conclusive. `Assigned %s instead`, as `use` followed by `using` sounds repetitive.

pkg/crc/machine/vsock.go
gbraad: `Alternate` is not the right wording as that means that either can be assigned consecutively.

gbraad (changes requested):
Wording

pkg/crc/machine/vsock.go
cfergeau: `fallback` would work.

pkg/crc/machine/vsock.go
praveenkumar: > Is this the very last warning which is printed? Or is this going to be lost in the middle of the startup logs of crc?

@cfergeau It is not the last warning, this is something part of info messages (Since it is warning so should've the `WARN` and different color)

pkg/crc/machine/vsock.go
praveenkumar: > Fwiw, listening on port 80/443 is a privileged operation on some OS, I don't know if `isPortInUse` could fail because of this.

@cfergeau I checked with linux/mac running a python simple http server on port 80 and this is identified by this function.

praveenkumar:
> ```
> Note: Now user have to add those ports to application routes to access
> it. Like https://myapp.apps-crc.testing:8443 or http://myapp.apps-crc.testing:8080
> ```
> 
> This should be more than a note in the commit log, but part of the warning which is displayed to the user. This also needs to be mentioned in documentation.

@cfergeau Now this warning propagated to start. Let me know if we want this change in separate commit (like now) or you want to squash the last 2 commits.

anjannath:
when a fallback port is chosen for HTTPS then the `crc console` command won't be giving the correct URL for the console, it'd be nice if  `crc console` knew which port to use

something we can document or need to solve in a follow up pr

gbraad:
> knew which port to use

would need to be stored as part of the machine config as state :-/

anjannath:
> > knew which port to use
> 
> would need to be stored as part of the machine config as state :-/

we can also get this information from the daemon, listing the exposed ports

praveenkumar:
> > > knew which port to use
> > 
> > 
> > would need to be stored as part of the machine config as state :-/
> 
> we can also get this information from the daemon, listing the exposed ports

yes I was going to use this solution but https://github.com/code-ready/crc/issues/3331#issuecomment-1240404591 need discussion so putting that to hold.

/hold

pkg/crc/config/types.go
```diff
@@ -39,6 +39,10 @@ func (v SettingValue) AsInt() int {
 	return cast.ToInt(v.Value)
 }
 
+func (v SettingValue) AsUInt() uint {
+	return cast.ToUint(v.Value)
+}
```
cfergeau, line 44: `AsInt()` may be good enough instead of adding a new helper?

pkg/crc/config/validations.go
```diff
@@ -127,3 +127,14 @@ func validatePreset(value interface{}) (bool, string) {
 	}
 	return true, ""
 }
+
+func ValidatePort(value interface{}) (bool, string) {
+	port, err := cast.ToUintE(value)
+	if err != nil {
+		return false, "Requires integer value in range of 1024-65535"
```
cfergeau, line 134: below 1024 is now allowed?

pkg/crc/machine/types/types.go
```diff
@@ -31,6 +31,10 @@ type StartConfig struct {
 
 	// Shared dirs
 	EnableSharedDirs bool
+
+	// Port to access openshift route
```
cfergeau, line 35: `Ports to access OpenShift routes`?

pkg/crc/machine/vsock.go
```diff
@@ -102,12 +102,12 @@ func vsockPorts(preset crcPreset.Preset) []types.ExposeRequest {
 			},
 			types.ExposeRequest{
 				Protocol: "tcp",
-				Local:    fmt.Sprintf(":%s", httpsPort),
+				Local:    fmt.Sprintf(":%s", strconv.FormatUint(uint64(ingressHTTPSPort), 10)),
```
cfergeau, line 105: `:%d` should work?

pkg/crc/machine/vsock.go
```diff
@@ -146,3 +146,21 @@ func getSSHTunnelURI() string {
...
+func portFallbackWarning(ingressHTTPPort, ingressHTTPSPort uint) string {
+	var fallbackPortWarning string
+	userAssignedHTTPPort := strconv.FormatUint(uint64(ingressHTTPPort), 10)
+	userAssignedHTTPSPort := strconv.FormatUint(uint64(ingressHTTPSPort), 10)
+	fallbackPortWarningTmpl := `
+Note: Port %s is Assigned.
+You have to add this port to the application route to access
+it. like %s://myapp.apps-crc.testing:%s
+`
+	if userAssignedHTTPSPort != httpsPort {
```
cfergeau, line 159: `httpsPort` can be renamed to `remoteHTTPSPort` for more clarity

pkg/crc/machine/vsock.go
```diff
@@ -146,3 +146,21 @@ func getSSHTunnelURI() string {
...
+		fallbackPortWarning += fmt.Sprintf(fallbackPortWarningTmpl, userAssignedHTTPSPort, "https", userAssignedHTTPSPort)
+	}
+	if userAssignedHTTPPort != httpPort {
+		fallbackPortWarning += fmt.Sprintf(fallbackPortWarningTmpl, userAssignedHTTPPort, "http", userAssignedHTTPPort)
+	}
+	return fallbackPortWarning
```
cfergeau, line 165: I don't really see the point of carrying this warning string all the way from `pkg/crc/machine/vsock.go` so that it can be displayed in `cmd/crc/cmd/start.go`. We could just check if ingress-http-port/ingress-https-port are set to their default values, and print the warning if not?

pkg/crc/config/callbacks.go
```diff
@@ -32,3 +32,13 @@ func RequiresCRCSetup(key string, _ interface{}) string {
 	return fmt.Sprintf("Changes to configuration property '%s' are only applied during 'crc setup'.\n"+
 		"Please run 'crc setup' for this configuration to take effect.", key)
 }
+
+func RequiresHTTPPortChangeWarning(key string, value interface{}) string {
+	return fmt.Sprintf("Changes to configuration property '%s' will break the openshift route.\n"+
```
cfergeau, line 37: `... will break OpenShift HTTP routes. In order to access  OpenShift applications through HTTP URLs, the %d port must be manually specified, such as http://myapp.apps-crc.testing:%d`

pkg/crc/config/callbacks.go
```diff
@@ -32,3 +32,13 @@ func RequiresCRCSetup(key string, _ interface{}) string {
...
+		"User need to add %d port to application route to access it. like http://myapp.apps-crc.testing:%d", key, value, value)
+}
+
+func RequiresHTTPSPortChangeWarning(key string, value interface{}) string {
+	return fmt.Sprintf("Changes to configuration property '%s' will break the openshift webconsole.\n"+
```
cfergeau, line 42: We need the same warning as in the HTTP case, plus something for the web console. `After this change, the OpenShift console will be non-functional because of OpenShift limitations` (dunno if it's possible to insert the URL of the web console in this message?)

pkg/crc/config/settings.go
```diff
@@ -107,6 +109,10 @@ func RegisterSettings(cfg *Config) {
 
 	cfg.AddSetting(KubeAdminPassword, "", ValidateString, SuccessfullyApplied,
 		"User defined kubeadmin password")
+	cfg.AddSetting(IngressHTTPPort, 80, ValidatePort, RequiresHTTPPortChangeWarning,
+		"Consent to use alternative http port in case 80 is in use (1024-65535, default: 80)")
```
cfergeau, line 113: `Consent to use`? "Alternative HTTP port to use for OpenShift ingress/routes in case port 80 is already in use on the host" ? But this could be too long.

pkg/crc/machine/vsock.go
```diff
@@ -146,3 +146,21 @@ func getSSHTunnelURI() string {
...
 }
+
+func portFallbackWarning(ingressHTTPPort, ingressHTTPSPort uint) string {
+	var fallbackPortWarning string
+	userAssignedHTTPPort := strconv.FormatUint(uint64(ingressHTTPPort), 10)
+	userAssignedHTTPSPort := strconv.FormatUint(uint64(ingressHTTPSPort), 10)
+	fallbackPortWarningTmpl := `
+Note: Port %s is Assigned.
+You have to add this port to the application route to access
+it. like %s://myapp.apps-crc.testing:%s
```
cfergeau, line 157: `to acces it. like %s ..` -> `to access it, such as %s://myapp.apps-crc.testing:%s`

pkg/crc/config/types.go
praveenkumar: @cfergeau int can take nagative value as well so better to use `uint` just to safer side.

pkg/crc/config/validations.go
praveenkumar: I would avoid using something below 1024 because most of those ports are used by some services.

pkg/crc/config/types.go
cfergeau: I know, but most go code does not really care, see https://pkg.go.dev/net#LookupPort for example.

pkg/crc/machine/vsock.go
praveenkumar: > We could just check if ingress-http-port/ingress-https-port are set to their default values, and print the warning if not?

@cfergeau where you want to check it, in `cmd/start.go` ? If I do that I can't have it at the end as what we discussed previously so current implementation seemed clean to me.

pkg/crc/config/callbacks.go
praveenkumar: > We need the same warning as in the HTTP case, plus something for the web console. After this change, the OpenShift console will be non-functional because of OpenShift limitations

Sure, that we can do.

> dunno if it's possible to insert the URL of the web console in this message?

Not at this moment because web console url is created from bundle metadata side.

pkg/crc/config/types.go
cfergeau: https://pkg.go.dev/net/netip#AddrPortFrom uses uint16, not uint.

pkg/crc/config/settings.go
praveenkumar: Yeah I can put it, too long shouldn't be issue if it make more clear message to user.

pkg/crc/machine/vsock.go
cfergeau: The current implementation prints the message in `cmd/start.go` and manages to print it at the end, so I don't understand your objection?

praveenkumar: > The current implementation prints the message in `cmd/start.go` and manages to print it at the end, so I don't understand your objection?

@cfergeau current implementation works because we carry this warning from `pkg/crc/machine/vsock.go` or you are saying we just should should check that in `pkg/crc/machine/start.go` and return from there?

pkg/crc/config/types.go
praveenkumar: `LookupPort` is returning the port number `AddrPortForm` is more accurate, I can use uint16 if that makes it more accurate.

cfergeau: I'd be fine with just `int`, so pick whatever you prefer :)

pkg/crc/machine/vsock.go
cfergeau: If done in `pkg/crc/machine/start.go`, I'd prefer to see expose it as `fallbackHTTPPortInUse bool; fallbackHTTPSPortInUse bool;`, and convert this to a warning in `cmd/start.go`. It's probably fine to not even do that, and check the config in `cmd/start.go`? Or does it have downsides?

pkg/crc/config/types.go
praveenkumar: I am not going to change it in that case.

cmd/crc/cmd/start.go
```diff
@@ -75,32 +75,38 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
...
 				Code: preflightFailedExitCode,
 			}
 		}
 	}
+	portFallBackWarning := portFallbackWarning(config.Get(crcConfig.IngressHTTPPort).AsUInt(),
+		config.Get(crcConfig.IngressHTTPSPort).AsUInt())
 
-	return client.Start(ctx, startConfig)
+	result, err := client.Start(ctx, startConfig)
+	return result, portFallBackWarning, err
```
anjannath, line 101: nit: `portFallBackWarning` can directly be called inside `renderStartResult` the `runStart` function doesn't necessarily have to generate and return it

praveenkumar: @anjannath we don't have config option in `renderStartResult` which is used to get what http/https port user specified before generating the warning.

anjannath: you mean `config`? it should be accessible to the whole `cmd` package

praveenkumar: Ah yes, then it should be much simpler, Thanks 👍🏼

cmd/crc/cmd/start.go
```diff
@@ -268,6 +276,9 @@ func writeOpenShiftTemplatedMessage(writer io.Writer, s *startResult) error {
 	if s.ClusterConfig.ClusterType == preset.OKD {
 		tmpl = fmt.Sprintf("%s\n\n%s", startTemplateForOpenshift, startTemplateForOKD)
 	}
+	if s.FallbackPortWarning != "" {
+		tmpl = fmt.Sprintf("%s\n%s", tmpl, fallbackPortWarning)
+	}
```
cfergeau, line 281: Do we get any benefits from having `FallbackPortWarning` in `startResult` rather than calling `portFallbackWarning()` here?

cmd/crc/cmd/start.go
```diff
@@ -321,3 +333,22 @@ func checkDaemonStarted() error {
 	}
 	return daemonclient.CheckIfOlderVersion(v)
 }
+
+func portFallbackWarning() string {
+	var fallbackPortWarning string
+	fallbackPortWarningTmpl := `
+Note: Port %d is Assigned.
```
cfergeau, line 340: Why is the method called 'Warning' but the string is `Note:`? :)

praveenkumar: @cfergeau I put the `Warning` instead `Note`.

cmd/crc/cmd/start.go
praveenkumar: If it is not part of result then you will not get it as part of `crc start -ojson`, in case some other app want to make use of this warning to show then it can be used.
```
$ crc start -o json
[...]
   "developerCredentials": {
      "username": "developer",
      "password": "developer"
    }
  },
  "fallbackPortWarning": "\nNote: Port 8080 is Assigned.\nYou have to add this port to the application route to access\nit, such as http://myapp.apps-crc.testing:8080\n"
}

```

cfergeau: If  we want to expose this in json,  this would look more useful than an arbitrary string:
```
[...]
   "developerCredentials": {
      "username": "developer",
      "password": "developer"
    }
  },
  "alternativeIngressPorts": {
      "http": "8080",
  }
}
```

praveenkumar: What we want to expose is debatable, warning can be useful if the app just need to popup or even show in the terminal. Also this is the same warning what crc start cmd shows. If we just put alternativeIngressPorts then user of it need to put own interpretation/warning around it.

cfergeau: In my opinion, a preformatted string is always never the right thing to have in an API, be it a json struct, a function return value, ... As soon as you need something different from the hardcoded string, you have to add awkward parsing code to get back the original data, and rebuild what you need from it.
One way forward is to not add anything to the json output for now, and to add 'something' when we have clear requirements for it.

praveenkumar: @cfergeau Ack, I remove the warn string from json output now.

cmd/crc/cmd/start.go
```diff
@@ -321,3 +329,26 @@ func checkDaemonStarted() error {
...
 }
+
+func portFallbackWarning() string {
+	var fallbackPortWarning string
+	httpScheme := "http"
+	httpsScheme := "https"
+	fallbackPortWarningTmpl := `
+Warning: Port %d is used for OpenShift %s routes instead of the default
+You have to add this port to %s URLs when accessing OpenShift application,
+such as %s://myapp.apps-crc.testing:%d
```
cfergeau, line 340: Warning that the console will be non-functional would be useful here as well. Especially as one message just before the warning is 
```
  {{ .ClusterConfig.WebConsoleURL }}
```
A follow-up would be to remove this message when the https is overridden.

cfergeau (approved):

cmd/crc/cmd/start.go
gbraad: Address as a follow-up

PR #3333, opened 2022-09-05
Add storage backend for configs based on OS provided secret store

### LINKED ISSUE #3312: Add a secret config type to store secrets in the OS provided secret store
It'll be useful to have a way to store a password/pull-secret in the OS key store directly using the `crc config` commands.

The end result would be:

- e.g `crc config set smb-share-password crcpass` (its stored in the OS secret store and 'crcpass' is not written to ~/.crc/crc.json)
- `crc config view` should mask the value of `smb-share-password` in its output

anjannath:
Fixes #3312

### DIFF
--- a/cmd/crc/cmd/config/view.go
+++ b/cmd/crc/cmd/config/view.go
@@ -16,7 +16,10 @@ const (
 	DefaultConfigViewFormat = "- {{.ConfigKey | printf \"%-38s\"}}: {{.ConfigValue}}"
 )
 
-var configViewFormat string
+var (
+	configViewFormat string
+	showSecrets      bool
+)
 
 type configViewTemplate struct {
 	ConfigKey   string
@@ -38,6 +41,7 @@ func configViewCmd(config config.Storage) *cobra.Command {
 	}
 	configViewCmd.Flags().StringVar(&configViewFormat, "format", DefaultConfigViewFormat,
 		`Go template format to apply to the configuration file. For more information about Go templates, see: https://golang.org/pkg/text/template/`)
+	configViewCmd.Flags().BoolVar(&showSecrets, "show-secrets", false, "Show values of secret config properties")
 	return configViewCmd
 }
 
@@ -55,7 +59,7 @@ func runConfigView(cfg map[string]config.SettingValue, tmpl *template.Template, 
 		if v.IsDefault {
 			continue
 		}
-		if v.IsSecret {
+		if v.IsSecret && !showSecrets {
 			continue
 		}
 		viewTmplt := configViewTemplate{k, v.AsString()}
--- a/cmd/crc/cmd/config/view.go
+++ b/cmd/crc/cmd/config/view.go
@@ -55,6 +55,9 @@ func runConfigView(cfg map[string]config.SettingValue, tmpl *template.Template, 
 		if v.IsDefault {
 			continue
 		}
+		if v.IsSecret {
+			continue
+		}
 		viewTmplt := configViewTemplate{k, v.AsString()}
 		var buffer bytes.Buffer
 		if err := tmpl.Execute(&buffer, viewTmplt); err != nil {
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -50,7 +50,7 @@ func init() {
 		logging.Fatal(err.Error())
 	}
 	var err error
-	config, viper, err = newViperConfig()
+	config, viper, err = newConfig()
 	if err != nil {
 		logging.Fatal(err.Error())
 	}
@@ -144,12 +144,12 @@ func setProxyDefaults() error {
 	return nil
 }
 
-func newViperConfig() (*crcConfig.Config, *crcConfig.ViperStorage, error) {
+func newConfig() (*crcConfig.Config, *crcConfig.ViperStorage, error) {
 	viper, err := crcConfig.NewViperStorage(constants.ConfigPath, constants.CrcEnvPrefix)
 	if err != nil {
 		return nil, nil, err
 	}
-	cfg := crcConfig.New(viper)
+	cfg := crcConfig.New(viper, crcConfig.NewSecretStorage())
 	crcConfig.RegisterSettings(cfg)
 	preflight.RegisterSettings(cfg)
 	return cfg, viper, nil
--- a/pkg/crc/config/config.go
+++ b/pkg/crc/config/config.go
@@ -182,6 +182,7 @@ func (c *Config) Get(key string) SettingValue {
 	return SettingValue{
 		Value:     value,
 		IsDefault: reflect.DeepEqual(setting.defaultValue, value),
+		IsSecret:  isUnderlyingTypeSecret(value),
 	}
 }
 
--- a/pkg/crc/config/config.go
+++ b/pkg/crc/config/config.go
@@ -16,12 +16,14 @@ const (
 
 type Config struct {
 	storage        RawStorage
+	secretStorage  RawStorage
 	settingsByName map[string]Setting
 }
 
-func New(storage RawStorage) *Config {
+func New(storage, secretStorage RawStorage) *Config {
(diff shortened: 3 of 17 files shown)


cfergeau (commented):
Adding a second implementation of the `Config` interface to handle secrets is an interesting approach! One thing to have in mind though is that this 'secret' keyring will not always be available, I don't have access to gnome-keyring when I ssh into a linux box (but maybe this is a matter of configuration?)

anjannath:
> Adding a second implementation of the `Config` interface to handle secrets is an interesting approach! One thing to have in mind though is that this 'secret' keyring will not always be available, I don't have access to gnome-keyring when I ssh into a linux box (but maybe this is a matter of configuration?)

ahh, didn't know, i don't think i ever tried this, but always assumed the cli `secret-tool` works via ssh

cfergeau:
> ahh, didn't know, i don't think i ever tried this, but always assumed the cli `secret-tool` works via ssh

My observations are only based on the pull secret keyring code, it always outputs a warning on my test machine, I never checked if it can be fixed.

anjannath:
> > ahh, didn't know, i don't think i ever tried this, but always assumed the cli `secret-tool` works via ssh
> 
> My observations are only based on the pull secret keyring code, it always outputs a warning on my test machine, I never checked if it can be fixed.

seems like the issue is with dbus not being accessible/available, its failing in CI for ubuntu :(

anjannath:
@cfergeau i've tried to add a check to see if the keyring is accessible at the beginning when the secret store is initialised and not error out and crash the whole thing when we try to access a secret config store that doesn't have access to the keychain, PTALA :)

cmd/crc/cmd/root.go
```diff
@@ -55,6 +57,8 @@ func init() {
 		logging.Fatal(err.Error())
 	}
 
+	secretConfig = newSecretConfig()
```
anjannath, line 60: looking again at this today, this can be improved, we don't need to expose the storage side of things, we can instead extend the `Config` objects `Set`, `Get`, and `Unset` methods to take care of an additional `Secret` config type like strings, bools etc. and hide the implementation detail of the secret config store, will rework this part of the PR again.

cc @cfergeau

cmd/crc/cmd/config/config.go
```diff
@@ -36,8 +36,9 @@ func less(lhsKey, rhsKey string) bool {
 	return lhsKey < rhsKey
 }
 
-func configurableFields(config *config.Config) string {
+func configurableFields(config, secretConfig *config.Config) string {
```
anjannath, line 43: with https://github.com/code-ready/crc/pull/3333/commits/fe25223cb8d7369954bffc4d34c2d88e5c4a863a#r969320847 the goal is to remove having to pass in two config structs

pkg/crc/config/secret_config_test.go
```diff
@@ -0,0 +1,68 @@
...
+	assert.NoError(t, err)
+
+	assert.Equal(t, SettingValue{
+		Value:     Secret("apples"),
+		IsDefault: true,
+		IsSecret:  true,
+	}, cfg.Get(secret))
+}
+
+func TestSecretCannotSetWithWrongType(t *testing.T) {
```
anjannath, line 62: this test is not very useful and can be removed, inherited from the `viper_config_test.go`

pkg/crc/config/secret_config.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/code-ready/crc/pkg/crc/logging"
+	"github.com/spf13/cast"
+	"github.com/zalando/go-keyring"
+)
+
+const CRCSecretServiceName = "crc"
```
cfergeau, line 12: I would drop `CRC` from the name, and this can be private to this module, `secretServiceName`

pkg/crc/config/secret_config.go
```diff
@@ -0,0 +1,69 @@
...
+
+var ErrSecretsNotAccesible = errors.New("secret store is not accesible")
```
cfergeau, line 14: "Accessible" takes 2 's'. This happens multiple times in this file.

cmd/crc/cmd/config/view.go
```diff
@@ -38,6 +42,7 @@ func configViewCmd(config config.Storage) *cobra.Command {
 	}
 	configViewCmd.Flags().StringVar(&configViewFormat, "format", DefaultConfigViewFormat,
 		`Go template format to apply to the configuration file. For more information about Go templates, see: https://golang.org/pkg/text/template/`)
+	configViewCmd.Flags().BoolVarP(getShowSecretsFlag())
```
cfergeau, line 45: Can you define this in the same way as `format` just above?
```
	configViewCmd.Flags().BoolVarP(&showSecrets, "show-secrets", "s", false, showSecretFlagHelpMsg)
```

cmd/crc/cmd/config/view.go
```diff
@@ -55,6 +60,9 @@ func runConfigView(cfg map[string]config.SettingValue, tmpl *template.Template,
 		if v.IsDefault {
 			continue
 		}
+		if v.IsSecret && !showSecrets {
+			continue
+		}
```
cfergeau, line 65: The previous commit could possibly have `if v.IsSecret { continue }`, and this commit would only add the `&& !showSecret` part of the test.

cfergeau (commented):
I need to take a closer look at  `Add a Secret type to register secret settings`, but overall this looks nice!

pkg/crc/config/secret_config.go
anjannath: fixed now, thanks :)

pkg/crc/config/config.go
```diff
@@ -90,6 +90,8 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
 		}
 	case preset.Preset:
 		castValue = cast.ToString(value)
+	case Secret:
+		return c.setSecret(key, value)
```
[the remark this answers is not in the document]
anjannath: yeah, only difference is choosing which _storage_ is used to set the config

pkg/crc/config/secret_config_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	password = "password"
+	secret   = "secret"
+)
+
+func newTestConfigSecret() (*Config, error) {
+	cfg := New(NewEmptyInMemoryStorage(), NewEmptyInMemorySecretStorage())
```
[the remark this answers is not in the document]
anjannath: also thanks to @guillaumerose for adding the viper_config_tests these are based on that :)

pkg/crc/config/config.go
```diff
@@ -123,6 +130,9 @@ func (c *Config) Get(key string) SettingValue {
 		}
 	}
 	value := c.storage.Get(key)
+	if value == nil {
+		value = c.secretStorage.Get(key)
+	}
```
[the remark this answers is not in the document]
anjannath: yeah this'd be better, thanks, tried this and its working

pkg/crc/config/config.go
```diff
@@ -94,24 +95,35 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
...
 func (c *Config) Unset(key string) (string, error) {
-	_, ok := c.settingsByName[key]
+	setting, ok := c.settingsByName[key]
 	if !ok {
 		return "", fmt.Errorf(configPropDoesntExistMsg, key)
 	}
-
-	if err := c.storage.Unset(key); err != nil {
-		return "", err
+	switch setting.defaultValue.(type) {
```
cfergeau, line 117: Maybe use `setting.isSecret` here too?

cfergeau (approved):

cfergeau (approved):

Issue #374, opened 2019-08-03
[Epic] Add functionality to communicate from WSL(2) to the CRC VM on Windows


gbraad:
Although it is early day, WSL2 uses Hyper-V and provides a nice experience when it comes to coding (like remote-WSL from VSCode). Although for WSL2 there is no direct way to communicate from the `WSL` Virtual Switch to the `Default Switch`. A possibility would be to add the `WSL` switch tothe CRC VM once the environment is enabled, and configure accordingly so `oc` from WSL can communicate with the CRC VM.

Note: this behaviour is different from WSL(v1) where the Windows networking stack is shared between Windows and the Linux susbsystem

---

Enable forwarding for the vSwitch allows traffic https://github.com/code-ready/crc/issues/374#issuecomment-803580645 however, this only works with the legacy vSwitch mode, not vsock.

nevermosby:
Hi there, currently I am trying the provision based on wsl2 ubuntu box(using the  crc binary of linux version). When I run `crc setup` inside wsl2 box, it shows some error messages:
```shell
david@XXXX:/mnt/c/WINDOWS/system32$ crc setup --log-level debug
INFO Checking if NetworkManager is installed
DEBU Checking if 'nmcli' is available
DEBU 'nmcli' was found in /usr/bin/nmcli
INFO Checking if NetworkManager service is running
DEBU Checking if NetworkManager.service is running
DEBU exit status 1 : System has not been booted with systemd as init system (PID 1). Can't operate.
FATA NetworkManager is required. Please make sure it is installed and running manually
```

I have done `apt install network-manager` and started the service manually to make sure it is running. It turns out that in the source code:
```golang
// https://github.com/code-ready/crc/blob/master/pkg/crc/preflight/preflight_checks_linux.go#L458
......
path, err := exec.LookPath("systemctl")
...
stdOut, stdErr, err := crcos.RunWithDefaultLocale(path, "is-active", "NetworkManager")
...
```
Currently the code is using the command `systemctl is-active NetworkManager`, which is not working now in WSL2 box, it should be like `service network-manager status` and check the result for "running".

Is it worthy of making a PR to cover the provision for wsl2 box? If so, I am glad to do it.
Looking forward to ur idea.

gbraad:
The issue here talks about communication between WSL(2) and CRC-VM, like using `oc`, but actually targeting the underlying Hyper-V to host the CRC-vm.

[...]

PR #3337, opened 2022-09-07
Support WSL2 distros accessing private (apiserver, ssh, cockpit) services

GingerGeek:
**Addresses:** Issue #374

## Solution/Idea

Bind Windows host-side vSock ports to the Window's machines IP on the WSL2 network, allowing both Windows host and WSL2 to connect to VM services that aren't shared publically.

## Proposed changes

Adds decision-making on the "local" address for API Server, SSH server and DNS resolution of routes when these are using vsock as transport on windows. 

In addition, the injected DNS routes point to the WSL2 host address, as these additions are automatically* (see caveats) added into WSL2 so that the domains work across both Windows and WSL2.

## Testing

After successfully running `start` and `setup` you are able to run tooling from within WSL 2.

```shell
zed@THINKZED:~/development/oss/crc$ uname -a
Linux THINKZED 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 GNU/Linux
zed@THINKZED:~/development/oss/crc$ oc status
In project default on server https://api.crc.testing:6443

svc/openshift - kubernetes.default.svc.cluster.local
svc/kubernetes - 10.217.4.1:443 -> 6443

View details with 'oc describe <resource>/<name>' or list resources with 'oc get all'.
zed@THINKZED:~/development/oss/crc$ kubectl cluster-info
Kubernetes control plane is running at https://api.crc.testing:6443

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
```

## Caveats

Changes to the Window's hosts files are not always automatically picked up by WSL2 distributions. After a `crc start`, if the DNS entries were not already present, WSL2 may need to be restarted with `wsl --shutdown`.

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -37,6 +37,7 @@ const (
 	EmergencyLogin           = "enable-emergency-login"
 	PersistentVolumeSize     = "persistent-volume-size"
 	EnableBundleQuayFallback = "enable-bundle-quay-fallback"
+	WSLNetworkAccess         = "wsl-network-access"
 )
 
 func RegisterSettings(cfg *Config) {
@@ -96,13 +97,15 @@ func RegisterSettings(cfg *Config) {
 	cfg.AddSetting(PersistentVolumeSize, constants.DefaultPersistentVolumeSize, validatePersistentVolumeSize, SuccessfullyApplied,
 		fmt.Sprintf("Total size in GiB of the persistent volume used by the CSI driver for %s preset (must be greater than or equal to '%d')", preset.Microshift, constants.DefaultPersistentVolumeSize))
 
-	// Shared directories configs
 	if runtime.GOOS == "windows" {
+		// Shared directories configs
 		cfg.AddSetting(SharedDirPassword, Secret(""), validateString, SuccessfullyApplied,
 			"Password used while using CIFS/SMB file sharing (It is the password for the current logged in user)")
-
 		cfg.AddSetting(EnableSharedDirs, false, validateSmbSharedDirs, SuccessfullyApplied,
 			"Mounts host's user profile folder at '/' in the CRC VM (true/false, default: false)")
+		// Setting to enable access to CRC vm from wsl
+		cfg.AddSetting(WSLNetworkAccess, false, ValidateBool, SuccessfullyApplied,
+			"Enable access to CRC VM within WSL environment (true/false, default: false)")
 	} else {
 		cfg.AddSetting(EnableSharedDirs, true, ValidateBool, SuccessfullyApplied,
 			"Mounts host's home directory at '/' in the CRC VM (true/false, default: true)")
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -9,6 +9,7 @@ import (
 	"github.com/crc-org/crc/v2/pkg/crc/network"
 	"github.com/crc-org/crc/v2/pkg/crc/preset"
 	"github.com/crc-org/crc/v2/pkg/crc/version"
+	"github.com/spf13/cast"
 )
 
 const (
@@ -148,10 +149,17 @@ func RegisterSettings(cfg *Config) {
 	if err := cfg.RegisterNotifier(Preset, presetChanged); err != nil {
 		logging.Debugf("Failed to register notifier for Preset: %v", err)
 	}
+	if err := cfg.RegisterNotifier(WSLNetworkAccess, wslNetworkAccessChanged); err != nil {
+		logging.Debugf("Failed to register notifier for wsl-network-access: %v", err)
+	}
 }
 
 func presetChanged(cfg *Config, _ string, _ interface{}) {
 	UpdateDefaults(cfg)
+}
+
+func wslNetworkAccessChanged(_ *Config, _ string, value interface{}) {
+	constants.WSLNetworkAccess = cast.ToBool(value)
 }
 
 func defaultCPUs(cfg Storage) int {
--- a/pkg/crc/machine/virtualmachine.go
+++ b/pkg/crc/machine/virtualmachine.go
@@ -89,7 +89,7 @@ func (vm *virtualMachine) State() (state.State, error) {
 
 func (vm *virtualMachine) IP() (string, error) {
 	if vm.vsock {
-		return "127.0.0.1", nil
+		return VsockPrivateAddress(), nil
 	}
 	return vm.Driver.GetIP()
 }
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -92,10 +92,15 @@ func vsockPorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint)
 		socketProtocol = types.NPIPE
 		socketLocal = constants.DefaultPodmanNamedPipe
 	}
+
+	// API, Cockpit and Internal SSH are all bound to a private address, usually 127.0.0.1
+	// On Windows, where WSL2 is active, it's bound to the Windows machine's address within the private WSL2 network
+	privateIP := VsockPrivateAddress()
+
 	exposeRequest := []types.ExposeRequest{
 		{
 			Protocol: "tcp",
-			Local:    net.JoinHostPort(localIP, strconv.Itoa(constants.VsockSSHPort)),
+			Local:    net.JoinHostPort(privateIP, strconv.Itoa(constants.VsockSSHPort)),
 			Remote:   net.JoinHostPort(virtualMachineIP, internalSSHPort),
 		},
 		{
@@ -110,7 +115,7 @@ func vsockPorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint)
 		exposeRequest = append(exposeRequest,
 			types.ExposeRequest{
 				Protocol: "tcp",
-				Local:    net.JoinHostPort(localIP, apiPort),
+				Local:    net.JoinHostPort(privateIP, apiPort),
 				Remote:   net.JoinHostPort(virtualMachineIP, apiPort),
 			},
 			types.ExposeRequest{
@@ -127,7 +132,7 @@ func vsockPorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint)
 		exposeRequest = append(exposeRequest,
 			types.ExposeRequest{
 				Protocol: "tcp",
-				Local:    net.JoinHostPort(localIP, cockpitPort),
+				Local:    net.JoinHostPort(privateIP, cockpitPort),
 				Remote:   net.JoinHostPort(virtualMachineIP, cockpitPort),
 			})
 	default:
--- a/pkg/crc/machine/vsockprivateaddress_windows.go
+++ b/pkg/crc/machine/vsockprivateaddress_windows.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/crc-org/crc/v2/pkg/crc/constants"
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
 	"github.com/crc-org/crc/v2/pkg/os/windows/powershell"
 )
@@ -16,6 +17,9 @@ const (
 // On Windows it would be useful if WSL2 can access VSock resources as well
 // This address should work on both the windows side and the WSL2 side
 func VsockPrivateAddress() (addrStr string) {
+	if !constants.WSLNetworkAccess {
+		return fallbackPrivateAddress
+	}
 	cmd := fmt.Sprintf(`(Get-NetIPAddress -AddressFamily IPv4 -InterfaceAlias "%s").IPAddress`, adapterName)
 	stdout, stderr, err := powershell.Execute(cmd)
 	if err != nil {
--- a/pkg/crc/machine/vsockprivateaddress_windows.go
+++ b/pkg/crc/machine/vsockprivateaddress_windows.go
@@ -1,0 +1,29 @@
+package machine
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/crc-org/crc/v2/pkg/crc/logging"
+	"github.com/crc-org/crc/v2/pkg/os/windows/powershell"
+)
+
+const (
+	fallbackPrivateAddress = "127.0.0.1"
+	adapterName            = "vEthernet (WSL)"
+)
+
+// On Windows it would be useful if WSL2 can access VSock resources as well
+// This address should work on both the windows side and the WSL2 side
+func VsockPrivateAddress() (addrStr string) {
+	cmd := fmt.Sprintf(`(Get-NetIPAddress -AddressFamily IPv4 -InterfaceAlias "%s").IPAddress`, adapterName)
+	stdout, stderr, err := powershell.Execute(cmd)
+	if err != nil {
+		logging.Debugf("Unable to find IP address for WSL vswitch: %v: %s", err, stderr)
+		return fallbackPrivateAddress
+	}
+	if strings.TrimSpace(stdout) == "" {
+		return fallbackPrivateAddress
+	}
+	return strings.TrimSpace(stdout)
+}
--- a/cmd/crc/cmd/daemon.go
+++ b/cmd/crc/cmd/daemon.go
@@ -21,6 +21,7 @@ import (
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
 	"github.com/crc-org/crc/v2/pkg/crc/daemonclient"
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
+	"github.com/crc-org/crc/v2/pkg/crc/machine"
 	"github.com/docker/go-units"
 	"github.com/gorilla/handlers"
 	"github.com/pkg/errors"
@@ -238,7 +239,10 @@ func gatewayAPIMux() *http.ServeMux {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/hosts/add", func(w http.ResponseWriter, r *http.Request) {
 		acceptJSONStringArray(w, r, func(hostnames []string) error {
-			return adminhelper.AddToHostsFile("127.0.0.1", hostnames...)
+			// Most of the time we will add a hosts entry to resolve this to 127.0.0.1
+			// On a Windows machine running WSL this is the "IP" of the machine within the WSL2 network,
+			// so it is accessible from both WSL2 and Windows
+			return adminhelper.AddToHostsFile(machine.VsockPrivateAddress(), hostnames...)
 		})
 	})
 	mux.HandleFunc("/hosts/remove", func(w http.ResponseWriter, r *http.Request) {
--- a/cmd/crc/cmd/daemon.go
+++ b/cmd/crc/cmd/daemon.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"regexp"
+	"runtime"
 	"syscall"
 	"time"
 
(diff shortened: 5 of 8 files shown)


gbraad:
This assumes WSL2 is installed on a target machine. This is not the case in a regular install of CRC as we only enable the Hyper-V service. WSL2 is not a necessity to operate.

I would say this should NOT be the default option to run with as this changes behaviour, especially in environment where WSL2 is not enabled.

pkg/crc/machine/vsockprivateaddress_windows.go
```diff
@@ -0,0 +1,65 @@
...
+			if errAddrs != nil {
+				return fallbackPrivateAddress, errAddrs
+			}
+			for _, address := range addresses {
+				addr := address.String()
+				if strings.HasPrefix(addr, ipv4Prefix) {
+					indexOfSlash := strings.IndexAny(addr, "/")
+					if indexOfSlash > 0 {
+						var addrIPOnly = addr[0:indexOfSlash] // We can simply strip the / as WSL2 sits at the start of the network
+						logging.Infof("private vsock binds will be to the WSL2 host address %s", addrIPOnly)
```
anjannath, line 55: I think this log can be a `debug` log, during start we already notify which IP address is assigned to the vm

pkg/crc/machine/vsockprivateaddress_windows.go
```diff
@@ -0,0 +1,65 @@
...
+	addrResult, err := localhostOrWslhost()
+	if err != nil {
+		logging.Warnf("Encountered error trying to resolve WSL host IP: %s", err)
+	}
+	vsockCheckedWsl, vsockPrivateAddressToUse = true, addrResult
+	return addrResult
+}
+
+func localhostOrWslhost() (addrStr string, err error) {
+	interfaces, errIf := net.Interfaces()
```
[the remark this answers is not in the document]
GingerGeek: ah ok, I'm historically a Java dev so these multi-returns are eye-opening! Will amend

GingerGeek:
@gbraad From a "work-outs-the-box" perspective I would suggest this being enabled by default without an additional flag, with potentially an override flag to disable it (although I can't think of why you would need to).

Using WSL2 for development, and having all code repos, toolings, etc., within it, is becoming a more typical pattern for developers using Windows as a daily driver. Especially with integrations of VSCode etc.

From a usability perspective, I would think a developer trying out OKD would want to be able `crc setup` and `crc start` and by default have that accessible from their WSL2 environment without needing to get into the weeds of WSL2's or CRC's netstack.

Although there is a behaviour change, it's not a "visible" one to a user since you can still access all services as you would before from Windows. The change in January (dcae4544d6ee482e776071258d0bd71b4410ff4a) was made to ensure apiserver, cockpit and SSH were not exposed publically; this effectively retains that as the WSL2 LAN is private to the machine.

The only potential break would be if someone is referencing the cluster directly via 127.0.0.1 within their application or scripts, but given the default configuration of Openshift/OKD (ie routes based off hostname), I think that's unlikely (all docs also refer to hostnames).

On the DNS caveat... To clarify this only exists within WSL2. The behaviour from the Windows perspective remains stable. 

I've been looking into this and trying to figure out when/how WSL2 updates DNS configuration. I perhaps was being too pessimistic in my original PR. 

WSL2 relies on a service on the Windows machine for DNS, this of course picks up changes instantly from the Windows hosts file. However, the changes won't get reflected in `/etc/hosts` until a WSL restart.

This would mean there is a potential for removed routes' DNS records to linger within WSL2, but I don't think that's a show-stopping issue.

gbraad:
> From a "work-outs-the-box" perspective I would suggest this being enabled by default without an additional flag, with potentially an override flag to disable it (although I can't think of why you would need to).

> The only potential break would be if someone is referencing the cluster directly via 127.0.0.1

We can't change behaviour as people might have functionality that relies on this. This is practice that people use when DNS is not working as expected.

gbraad:
> On the DNS caveat... To clarify this only exists within WSL2. 

I understand, but this is something that needs additional messaging/feedback or documentation, otherwise we will see people file issues about broken DNS when using this.

gbraad:
> DNS records to linger within WSL2, 

Not ideal... we are called out for doing this on `system32/drivers/etc/` too.

gbraad:
Note: flag can be a config option too... as long as this can be 'enabled'/'disabled', though default might no be the right approach now. Perhaps at a later time. We are still working on some issues around WSL2 and Podman before we commit to this.

gbraad:
@code-ready/crc-team adding peer review required as this needs to be tested with a route-all VPN setup. I am sure this will fail as addresses will be resolved on a non-localhost address, and this will not be possible in that case.

GingerGeek:
I've done some more testing on DNS, this seems to be the state of play:

For static hosts declaration within the distribution:
- The `/etc/hosts` file within the Linux distributions is updated by WSL when the distribution starts
- This update means the current state of `C:\WINDOWS\Systems32\drivers\etc\hosts` is included in the `/etc/hosts` file when the distro starts
- This behaviour can be disabled by the user via [`wsl.conf`](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configuration-settings-for-wslconf)

For DNS resolution within the distribution:
- The `/etc/resolv.conf` file within the Linux distribution is updated by WSL when the distribution starts is set by WSL 
- This update sets the nameserver to the windows machine, where the ICS (Internet Connection Services) responds to DNS queries and proxies configured DNS within Windows
- This behaviour can be disabled by the user via [`wsl.conf`](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#configuration-settings-for-wslconf)

This leads to the following behaviour under default settings:
- New hosts added will be immediately picked up by the WSL2 distro, there is no restart required as the Windows DNS services picks up the alterations to `C:\WINDOWS\Systems32\drivers\etc\hosts` 
- Hosts that are removed will also stop resolving from WSL2, same as above
- EXCEPTION: Hosts that are removed after the distribution has started, which were present when the WSL2 distribution started, will still resolve within the WSL2 environment. 
  - When the distribution is restarted, these hosts will be removed due to the syncing of `/etc/hosts`

If static hosts updating/syncing is disabled, then the resolution will still function within the WSL2 environment as long as the ICS DNS server is still in use. If a user configures custom nameservers within the WSL2 distribution, then DNS resolution for CRC may fail - although I think that's out of scope to support.

I think a WSL health check would be a good addition, I note currently when I run `crc start`  (even without my changes) I see the following log line, is this an intended failure since it's not a valid apps route?

`WARN Failed to query DNS from host: lookup foo.apps-crc.testing: no such host`

I will look to implement some basic health checks

GingerGeek:
On the route-all VPN... I think this will vary from vendor to vendor, however many do now have specific support for the internal WSL2 network (e.g [Cisco AnyConnect](https://www.cisco.com/c/en/us/td/docs/security/vpn_client/anyconnect/anyconnect410/release/notes/release-notes-anyconnect-4-10.html#Cisco_Reference.dita_fa8bc69c-3e95-487c-9ba2-4b5d6b78e118)).

At the very least, since the IP we bind to is the machine's own, I believe most VPN clients would allow connections to it (but not the wider network)? This is not really something I have too much experience with.

The exception of course is if there is a routing overlap between VPN networks and WSL2 internal network. There are several open issues around this within microsoft/WSL (e.g microsoft/WSL#5782 microsoft/WSL#4210). I found [this blog](https://janovesk.com/wsl/2022/01/21/wsl2-and-vpn-routing.html) around it as well

I've tested this using Tailscale and an exit node-enabled (with local network access disabled). It's working fine from both the WSL2 side and Windows side. I can probably also have a test done with an OpenVPN setup but I don't have easy access to the more corporate-y VPNs like AnyConnect etc

GingerGeek:
General todo/issues/thoughts:

- The WSL interface is only created once a distribution has been started. This means the interface doesn't exist on a fresh boot and won't be detected with the current method until you open a distribution. A better way of detecting WSL 2 support is probably by parsing the output of wsl --list.

- I should probably do a more sensible check for a valid ipv4 address other than the current hack of .startsWith("172."). The range used for WSL2 has changed historically (used to be in 192.168 range) so I would probably just confirm that's IPv4 and that it's a private range.

- One thought, is that we could open the vsock listener on both localhost addresses and on the WSL2 address. This would mean localhost connections to apiserver etc would still work on Windows. To ensure cross compatibility, the resolved DNS addresses would need to remain as the WSL2 address.

GingerGeek:
So is the consensus I should pull this out to be behind a config option? Something along the lines of 

`crc config set wsl2-netcompat true/false`

I would then suggest
- If `crc start` is run with wsl2-netcompat false, and WSL2 is detected, log a warning and suggest enabling (link to docs)?
- If `crc start` is run with wsl2-netcompat true, and WSL2 is not detected, fatally error?

gbraad:
> One thought, is that we could open the vsock listener on both localhost addresses and on the WSL2 address.

This would now change behaviour for localhost, but essentially means there is a `127.0.0.1` address in `hosts`, but that also needs to be available correctly on WSL with a `172/8` IP. Would not need a config option as behaviour remains same, but augments (though needs additional logic for this IP).


> crc config set wsl2-netcompat true/false

Very cryptic :-s. ... not very descriptive of the usage. Other suggestion?

GingerGeek:
> > One thought, is that we could open the vsock listener on both localhost addresses and on the WSL2 address.
> 
> This would now change behaviour for localhost, but essentially means there is a `127.0.0.1` address in `hosts`, but that also needs to be available correctly on WSL with a `172/8` IP. Would not need a config option as behaviour remains same, but augments (though needs additional logic for this IP).

To clarify, I would maintain the behaviour of keeping the DNS records in the hosts file pointing to the WSL private address, just also bind the listener to 127.0.0.1 for anyone who is referencing that address directly. Tbh, I'm not a fan of the idea and would only be to maintain compatibility with those users who have been forced to reference 127.0.0.1 directly.. which shouldn't be necessary.

> 
> > crc config set wsl2-netcompat true/false
> 
> Very cryptic :-s. ... not very descriptive of the usage. Other suggestion?

wsl2-access ?

pkg/crc/machine/vsockprivateaddress_windows.go
```diff
@@ -0,0 +1,65 @@
...
+// On Windows it would be useful if WSL2 can access VSock resources as well
+// This address should work on both the windows side and the WSL2 side
+
+var (
+	vsockCheckedWsl          = false
+	vsockPrivateAddressToUse = "127.0.0.1"
+)
+
+func VsockPrivateAddress() (addrStr string) {
+	if vsockCheckedWsl {
```
praveenkumar, line 26: This really looks cryptic, in first glance it looks like in case vsock checked with wsl and it true then return `127.0.0.1` . I think this function need to re looked around what it try to do.

```golang
func VsockPrivateAddress() string {
        addrResult, err := localhostOrWslhost()
	if err != nil {
		logging.Warnf("Encountered error trying to resolve WSL host IP: %s", err)
	}
        return addrResult
}
```

praveenkumar:
> wsl2-access ?

IIUC, then this is about accessing the wsl2 network then should it be `wsl2-network-access`?

pkg/crc/machine/vsockprivateaddress_windows.go
gbraad: `fallbackPrivateAddress` need to be returned here, when `wslhostAddress()` (current: `localhostOrWslhost()`) would return an `err`.

This will also make the toggle much easier in a single place. No need to jump into `wslhostAddress` when the config is not set.

praveenkumar:
@GingerGeek ping, Are you still interested to make this PR in with provided suggestions?

gbraad:
if no time @GingerGeek, just let us know.
glad it sees 'real world use' as part of testing :-D

mmurhamm:
Has this config option actually been implemented? That would be a big help.
Please see my comment to https://github.com/crc-org/crc/issues/374.
Thanks.

GingerGeek:
Hi, sorry I'm afraid I have not had time to further look at this

gbraad:
I understand the situation has been worse on Win 11 users. Did you see this too?

> I have not had time to further look at this

Would you still want to look at this? If not, I might have a look instead.

GingerGeek:
I'm a Windows 11 user and was frustrated enough to make the changes in CRC! I haven't tried it on other windows versions. Doubt I'm going to be able to give this the time it requires anytime soon I'm afraid :/

gbraad:
@anjannath can you try this once more?

pkg/crc/machine/vsockprivateaddress_windows.go
anjannath: IIUC these two functions are added to find out the ip address of the WSL specific virtual switch named `vEthernet (WSL)` if that is correct, then we can simplify it a bit by using powershell commands to get the IP address, e.g. `(Get-NetIPAddress -AddressFamily IPv4 -InterfaceAlias "vEthernet (WSL)").IPAddress`

anjannath:
after this PR  the communication to CRC resource from WSL would be possible but another thing to note is that the commands like `crc oc-env` or `crc podman-env` which configures the shell doesn't detect the WSL env properly and reports as cmd env. instead of bash/zsh.

so along with enabling this communication, we have to also improve this detection for WSL

```
annant@DESKTOP-J8H86J0:~$ uname -a
Linux DESKTOP-J8H86J0 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 GNU/Linux
annant@DESKTOP-J8H86J0:~$ crc.exe podman-env
SET PATH=C:\Users\anath\.crc\bin\oc;%PATH%
SET CONTAINER_SSHKEY=C:\Users\anath\.crc\machines\crc\id_ecdsa
SET CONTAINER_HOST=ssh://core@127.0.0.1:2222/run/user/1000/podman/podman.sock
SET DOCKER_HOST=npipe:////./pipe/crc-podman
REM @FOR /f "tokens=*" %i IN ('crc podman-env') DO @call %i
```

anjannath:
i've rebased the PR branch and made some modifications to the way the WSL vswitch ip is fetched here: https://github.com/anjannath/crc/commit/bedcc6f22d43b2c65ec3ef0cb3eea183a6d0e45a

tested it with the podman preset and its working as expected, need to also test with a VPN and the openshift and microshift presets

gbraad:
@anjannath 

`Maintainers are allowed to edit this pull request.` so I would suggest to do so. The IP determine function as I said was way too complicted. All is needed is to get the Host IP address, as our stack binds on the 'same' localhost.

gbraad:
> the shell doesn't detect the WSL env properly
> `crc.exe podman-env`

the detection is not an issue, but the `-s` option is restricted to the 'supported' shells for the platform. I believe it should be possible to force this

anjannath:
@gbraad @praveenkumar Updated the PR:
- rebased and simplified the function fetching the IP of the WSL vswitch
- added config option ~`wsl-network-access`~ `wsl-host-access` which defaults to `false`

GingerGeek:
Thanks for picking this up, sorry I've not been able to give this any more time

gbraad:
@GingerGeek no worries. life can get in the way of things; it happens to all of us.
it is appreciated if you still use CRC to give this a quick glance or a try. Otherwise, thank you regardless for the contribution.

pkg/crc/machine/vsock.go
```diff
@@ -92,10 +92,15 @@ func vsockPorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint)
 		socketProtocol = types.NPIPE
 		socketLocal = constants.DefaultPodmanNamedPipe
 	}
+
+	// API, Cockpit and Internal SSH are all bound to a private address, usually 127.0.0.1
+	// On Windows, where WSL2 is active, it's bound to the Windows machine's address within the private WSL2 network
+	privateIP := VsockPrivateAddress()
```
anjannath, line 98: this should also be behind the config option, now on windows it is always binding the port forwards to the WSL vswitch ip address

pkg/crc/machine/virtualmachine.go
```diff
@@ -89,7 +89,7 @@ func (vm *virtualMachine) State() (state.State, error) {
 
 func (vm *virtualMachine) IP() (string, error) {
 	if vm.vsock {
-		return "127.0.0.1", nil
+		return VsockPrivateAddress(), nil
```
anjannath, line 92: similar to https://github.com/crc-org/crc/pull/3337/files#r1302624090 this should also report the WSL vswitch ip address only when `wsl-host-access` is enabled, but currently its not considering the config option

anjannath: It'd be easier to toggle this with an env variable instead of the config option, so i suggest instead of having a config option `wsl-host-access` we can depend on a env variable `CRC_WSL_HOST_ACCESS` which should toggle this behavior, wdyt @gbraad ?

cfergeau: I don't know from where to where the connections would be allowed, but crc has a `host-network-access                        Allow TCP/IP connections from the CRC VM to services running on the host` config option, let's try to get the naming of the new option consistent/not confusing with this preexisting option.

anjannath: The comm. is to be allowed from WSL to crc VM, so that the `oc` commands work from the WSL terminal

i mistakenly made the config key `wsl-host-access` but previously `wsl2-network-access` was suggested in https://github.com/crc-org/crc/pull/3337#issuecomment-1248904591

pkg/crc/config/settings.go
```diff
@@ -92,13 +93,15 @@ func RegisterSettings(cfg *Config) {
...
 	if runtime.GOOS == "windows" {
+		// Shared directories configs
 		cfg.AddSetting(SharedDirPassword, Secret(""), validateString, SuccessfullyApplied,
 			"Password used while using CIFS/SMB file sharing (It is the password for the current logged in user)")
-
 		cfg.AddSetting(EnableSharedDirs, false, validateSmbSharedDirs, SuccessfullyApplied,
 			"Mounts host's user profile folder at '/' in the CRC VM (true/false, default: false)")
+		// Setting to enable access to CRC vm from wsl
+		cfg.AddSetting(WSLHostAccess, false, ValidateBool, SuccessfullyApplied,
+			"Enable acess to CRC VM within WSL environment (true/false, default: false)")
```
gbraad: access

pkg/crc/config/settings.go
```diff
@@ -35,6 +35,7 @@ const (
 	IngressHTTPPort         = "ingress-http-port"
 	IngressHTTPSPort        = "ingress-https-port"
 	EmergencyLogin          = "enable-emergency-login"
+	WSLHostAccess           = "wsl-host-access"
```
gbraad: The title is: "add 'wsl-network-access'" ?

pkg/crc/config/settings.go
```diff
@@ -92,13 +93,15 @@ func RegisterSettings(cfg *Config) {
...
-	// Shared directories configs
 	if runtime.GOOS == "windows" {
+		// Shared directories configs
 		cfg.AddSetting(SharedDirPassword, Secret(""), validateString, SuccessfullyApplied,
 			"Password used while using CIFS/SMB file sharing (It is the password for the current logged in user)")
-
 		cfg.AddSetting(EnableSharedDirs, false, validateSmbSharedDirs, SuccessfullyApplied,
 			"Mounts host's user profile folder at '/' in the CRC VM (true/false, default: false)")
+		// Setting to enable access to CRC vm from wsl
+		cfg.AddSetting(WSLHostAccess, false, ValidateBool, SuccessfullyApplied,
```
gbraad: is it network access or host access?

anjannath: should be `WSLNetworkAccess` updated now

pkg/crc/machine/vsock.go
anjannath: this is handled now in https://github.com/crc-org/crc/pull/3337/commits/e0ef98d82a0f21a65ee8b8adb1bd320ac621944e

anjannath:
@gbraad should we close this now, we haven't seen users requesting this feature much, can reopen again later if needed

gbraad:
There is a `networkingMode`you can set for WSL2, like bridged and mirrored. These can be a possible solution.

Closing it is OK

anjannath:
> There is a `networkingMode`you can set for WSL2, like bridged and mirrored. These can be a possible solution.
> 
> Closing it is OK

this is the link to the article about mirrored `networkingMode` on WSL2 https://learn.microsoft.com/en-us/windows/wsl/networking#mirrored-mode-networking

PR #3339, opened 2022-09-09
e2e: execute crc setup as an isolated command

adrianriobo:
During e2e executions from time to time crc setup got freeze giving false positives. This is due to a wrong behavior on clicumber shell utility within the downloader progress bar to avoid this conflict crc setup will be executed as a single command with os.exec also we remove checks for setup outputs

### DIFF
--- a/test/e2e/crcsuite/crcsuite.go
+++ b/test/e2e/crcsuite/crcsuite.go
@@ -68,6 +68,8 @@ func FeatureContext(s *godog.Suite) {
 		ExecuteCommand)
 	s.Step(`^execut(?:e|ing) crc (.*) command (.*)$`,
 		ExecuteCommandWithExpectedExitStatus)
+	s.Step(`^execut(?:e|ing) single crc (.*) command (.*)$`,
+		ExecuteSingleCommandWithExpectedExitStatus)
 
 	// CRC file operations
 	s.Step(`^file "([^"]*)" exists in CRC home folder$`,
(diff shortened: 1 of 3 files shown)


jsliacan (approved):
LGTM.

praveenkumar:
Have we tried https://github.com/code-ready/crc/pull/3167 since it have latest version of clicumber so might be some bug fixes around freeze?

adrianriobo:
That PR among other things upgrade godog version, unfortunately the issue it is not related with godog but with the [shell class](https://github.com/code-ready/crc/pull/3167/files#diff-702f0e6aa26aafd3c3ebbdd45c23e4b86bbc34a3579847ef09c5d7e42bffc378) which actually has no dependecy with godog at all.
@praveenkumar may you remember back then the issue with podman images output freeze (you were trying to refactor the shell class) issue is hitting again with the progress bar :( 

shell class emulates a single session for a user ...which make sense when settings envs (+ to emulate a real user interaction which typically would use a single terminal) ...on the case for the `crc setup` command we do not need to keep the session. 

May in future we can spend someting on a following PR trying again to adapt the shell go but for the time being and to allow keep going with the automation this will help us a lot

praveenkumar:
> That PR (#3167) among other things upgrade godog version, unfortunately the issue it is not related with godog but with the [shell class](https://github.com/code-ready/crc/pull/3167/files#diff-702f0e6aa26aafd3c3ebbdd45c23e4b86bbc34a3579847ef09c5d7e42bffc378) which actually has no dependecy with godog at all.
> 
> @praveenkumar may you remember back then the issue with podman images output freeze (you were trying to refactor the shell class) issue is hitting again with the progress bar :(
> 
> shell class emulates a single session for a user ...which make sense when settings envs (+ to emulate a real user interaction which typically would use a single terminal) ...on the case for the `crc setup` command we do not need to keep the session.
> 

Ah I remember now I tried https://github.com/code-ready/clicumber/commit/b1002c1e583dc0ca23f1bbb79b1dbcc5eec5e827 but end up reverting back because user flow, may be in future we really need to look into this again or other solution.

> May in future we can spend some time on a following PR trying again to adapt the shell class but for the time being and to allow keep going with the automation this will help us a lot

anjannath (approved):

PR #3341, opened 2022-09-12
Two build fixes

cfergeau:
This fixes a regression with the default Makefile target, and update gvisor-tap-vsock in order to fix build errors if one were to use golang 1.19.


evidolob (approved):
`make` now build crc binary

PR #3342, opened 2022-09-12
Update openshift version to 4.11.3 (stable channel)

praveenkumar:


pkg/crc/machine/bundle/constants.go
```diff
@@ -10,29 +10,21 @@ var bundleLocations = map[string]bundlesDownloadInfo{
 		"darwin": {
 			preset.OpenShift: download.NewRemoteFile("https://mirror.openshift.com/pub/openshift-v4/clients/crc/bundles/openshift/4.11.3/crc_vfkit_4.11.3_amd64.crcbundle",
 				"f912ca39c5243f0e089ba6a25273f4d50c8891fdf43f1f5333ff098022952472"),
-			preset.Podman: download.NewRemoteFile("https://mirror.openshift.com/pub/openshift-v4/clients/crc/bundles/podman/4.1.0/crc_podman_vfkit_4.1.0_amd64.crcbundle",
-				"81028e2373ba4d541ef926d2ba819b39b79d7c7993d7e66010fa0b2bec270670"),
```
cfergeau, line 14: Would be nice to have a single location to define either http downloads or registry downloads instead of having http here, and registry in https://github.com/code-ready/crc/blob/main/pkg/crc/image/image.go#L28-L32

cfergeau (approved):

pkg/crc/machine/bundle/constants.go
praveenkumar: Once we have everything on quay then we might not even need those download location for mirror. It might be there as fallback for some releases.

PR #3347, opened 2022-09-15
e2e: support for testing mac m1

adrianriobo:
This PR will add support for remote run e2e testing on mac m1. To accomplish this:

* An additional build for arm64 arch of e2e binary
* Include parameters to set the target arch on the e2e container
* Remove unused ux testing assets and functionality (As this require some extra setups for mac machines due to protection on `/` path + ux testing is not executed anymore)


images/build-e2e/entrypoint.sh
```diff
@@ -128,7 +128,11 @@ if [ -n "${USER_PASSWORD}" ]; then
     OPTIONS+="--user-password=${USER_PASSWORD} "
 fi
 OPTIONS+="--godog.tags=\"@${PLATFORM} && ${E2E_TAG_EXPRESSION}\" --godog.format=junit "
-BINARY_EXEC+="cd ${EXECUTION_FOLDER}/bin && ./${BINARY} ${OPTIONS} > ${RESULTS_FILE}.results"
+if [[ ${PLATFORM} == 'windows' ]]; then
+    BINARY_EXEC+="cd ${EXECUTION_FOLDER}/bin && ./${BINARY} ${OPTIONS} > ${RESULTS_FILE}.results"
+else
+    BINARY_EXEC+="sudo su - ${TARGET_HOST_USERNAME} -c \"cd ${EXECUTION_FOLDER}/bin && ./${BINARY} ${OPTIONS} > ${RESULTS_FILE}.results\""
+fi
```
praveenkumar, line 122: When we are ssh to the machine/host are we ssh-ing as different user and that's the reason we need to use `sudo su - <username>`?

praveenkumar: Also in commit log `changin the user session` => `changing the user's session`

images/build-e2e/Dockerfile
```diff
@@ -14,10 +14,6 @@ COPY --from=builder /workspace/out /opt/crc/bin
 # Review this when go 1.16 with embed support
 COPY --from=builder /workspace/test/e2e/features /opt/crc/features
 COPY --from=builder /workspace/test/testdata /opt/crc/testdata
-COPY --from=builder /workspace/test/extended/crc/ux/installer/applescripts /opt/crc/ux/installer/applescripts
-COPY --from=builder /workspace/test/extended/crc/ux/notification/applescripts /opt/crc/ux/notification/applescripts
-COPY --from=builder /workspace/test/extended/crc/ux/tray/applescripts /opt/crc/ux/tray/applescripts
-
```
praveenkumar, line 20: In the commit logs, we should have an explanation around why we are removing those.

go.mod
```diff
@@ -7,7 +7,6 @@ require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/Microsoft/go-winio v0.5.2
 	github.com/ProtonMail/go-crypto v0.0.0-20220623141421-5afb4c282135
-	github.com/RedHatQE/gowinx v0.0.3
```
praveenkumar, line 10: Is it a step back from our automation effort? You folks effort to put the UX test as part of automation and now we are removing those because of it is outdated or not works as expected or have some kind of limitation? Can it be fixed instead removing the functionality?

images/build-e2e/entrypoint.sh
adrianriobo: No we are ssh-ing as same user, but doing the `sudo su -` we are changing the envs of the session from a remote ssh type of session to a local session (which is more aligned with a typical user case).

images/build-e2e/Dockerfile
adrianriobo: well those are assets used for ux testing, as we are removing that part we need to remove them from the container image..the commit says _e2e container remove test ux assets_

I can split this into a different PR and move the commits related to remove ux testing from here.

images/build-e2e/entrypoint.sh
praveenkumar: Please mention that in the commit log, it now explains why we are doing it 👍🏼

go.mod
praveenkumar: @adrianriobo you can have it here but in the commit message put those intentions which allow us (reviewers) to understand why it is done.

images/build-e2e/entrypoint.sh
adrianriobo: done

go.mod
adrianriobo: done

PR #3351, opened 2022-09-19
Update podman version to 4.2.0

praveenkumar:


cfergeau (approved):
Reason for this is that in the upcoming release, we'll have an updated podman bundle, and fcos 36 has updated it to 4.2.0.

Issue #2898, opened 2021-12-21
[Task] Remove unnecessary information that is not part of the preset from `status` command(s)


gbraad:
Remove OpenShift specific information from the `status` and API `status` command when the `preset` == "podman".

PR #3364, opened 2022-09-27
Update 'status' command to use daemon API

evidolob:
**Fixes:** Issue #2898

## Solution/Idea

`status` command will filter empty values for `Openshift version` or `Podman version` to not print them.

## Proposed changes

1. `status` command now use daemon API to get status data
2. `crc/api/client/Client` becomes an interface, to enhance testability.
3.  https://github.com/stretchr/testify#mock-package and https://github.com/vektra/mockery is used as mock library

## Testing

1. `status`  should print `Machine does not exist. Use 'crc start' to create it` if no machine exists.(old behaviour)
2.  `status`  should not contain `OpenShift` version if `podman` preset selected.
3. `status`  should not contain `Podman` version if `openshift` preset selected.


.gitignore
```diff
@@ -6,3 +6,4 @@ release/
 packaging/tmp
 packaging/*.pkg
 tmp-embed/
+.idea/
```
gbraad, line 9: :-O

cmd/crc/cmd/status.go
```diff
@@ -89,19 +86,29 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
...
-	}{
+	}
+	lines := []line{
 		{"CRC VM", s.CrcStatus},
-		{"OpenShift", openshiftStatus(s)},
-		{"Podman", s.PodmanVersion},
-		{"Disk Usage", fmt.Sprintf(
+	}
+
+	if s.OpenShiftVersion != "" {
```
jsliacan, line 97: I heard rumors that with openshift preset you should ideally be able to do everything you can do with podman preset. In other words, I'd probably suggest keeping podman version in the status output under openshift preset. Maybe others can comment on this too.

evidolob: I remove version only if it empty, which means that podman/openshift not present in current preset.

cfergeau:
In the CRC project, we usually try to split PR/commits to only have one 'logical' change per-commit. Smaller commits are easier to review, they can have a detailed log focused on why the change is being made, how it's being made if it's complicated, ... (longer read on this topic, https://www.berrange.com/posts/2012/06/27/thoughts-on-improving-openstack-git-commit-practicehistory/ )

In this PR for example, it has a single commit, but it seems that it contains several changes:
- introducing a `Client` interface
- making use of a mock generator
- not printing podman/openshift versions if they are empty
- and the main goal of this PR, going through the daemon for `crc status`
- (and maybe I missed more small changes)

Could you try to split it along those lines? I can help if needed.

evidolob:
@cfergeau Done

praveenkumar:
With this commit following UX is going to change.
- User download crc on a fresh system/or used `crc cleanup`
- execute  `crc status` command

Since there is no daemon is running the status command fails with following
```
$ ./crc status
Get "http://unix/api/status": dial unix /home/prkumar/.crc/crc-http.sock: connect: connection refused
```

With current master it fails with and that's the reason current e2e test is failing.
```
$ ./crc status
```

evidolob:
@praveenkumar Thx, that it what @gbraad mentioned, we need to make commands work in two modes with and without running daemon. Going to return back local version of status command.

cfergeau:
In this scenario, failing with `crc does not seem to be setup correctly, have you run 'crc setup'?` instead of `Get "http://unix/api/status": dial unix /home/prkumar/.crc/crc-http.sock: connect: connection refused` would be fine with me

praveenkumar:
> In this scenario, failing with `crc does not seem to be setup correctly, have you run 'crc setup'?` instead of `Get "http://unix/api/status": dial unix /home/prkumar/.crc/crc-http.sock: connect: connection refused` would be fine with me

Yes, this is less confusing. Also I am more tend to have same kind of experience like `podman` have which can use API to run commands or directly from command.

cmd/crc/cmd/status.go
```diff
@@ -89,19 +86,29 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
...
+		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
+	}
+	if s.PodmanVersion != "" {
+		lines = append(lines, line{"Podman", s.PodmanVersion})
+	}
+
```
cfergeau, line 103: Fwiw, I would keep the change modifying `crc status` output separate from the change making use of the daemon for the status command. It's easier to revert one without reverting the other if one of these changes proved problematic. 
One thing which could be added is the OKD version (can be a follow-up if you want to focus on the current PR without adding more changes), it's currently missing.

anjannath (commented):
Tried on macOS and working as expected, @evidolob need to update the last fixup commit message or squash it to the original commit!

cmd/crc/cmd/status.go
```diff
@@ -44,18 +44,18 @@ type status struct {
...
-		return &status{Success: false, Error: crcErrors.ToSerializableError(err)}
-	}
+func getStatus(client *daemonclient.Client, cacheDir string) *status {
 
-	clusterStatus, err := client.Status()
+	clusterStatus, err := client.APIClient.Status()
 	if err != nil {
+		if strings.HasSuffix(err.Error(), "connect: connection refused") || strings.HasSuffix(err.Error(), "connect: no such file or directory") {
+			return &status{Success: false, Error: crcErrors.ToSerializableError(crcErrors.DaemonNotRunning)}
+		}
```
praveenkumar, line 59: You should replace this with something like because the error you will get error as `*url.Error` in case of daemon service is not running or the sock file doesn't exist.
```golang
var urlError *url.Error
if errors.As(err, &urlError) {
	return &status{Success: false, Error: crcErrors.ToSerializableError(crcErrors.DaemonNotRunning)}
}
```

evidolob: Done, thx!

cfergeau: Fwiw, there's a `checkDaemonStarted` in `cmd/crc/cmd/start.go`, we may want to use it more extensively in the future instead of assuming that all `url.Error` mean that the daemon is not running. Just something to have in mind, does not require changes in this PR.

cfergeau (commented):
Still looks fine to me, especially with the improved error when daemon is missing. May eb better to wait after the release to merge this though!

praveenkumar:
/hold

Hold till the 2.10.0 release

Also e2e-crc test need to be fixed https://github.com/code-ready/crc/blob/main/test/e2e/features/basic.feature#L29 it should contain `crc does not seem to be setup correctly, have you run 'crc setup'?` message.

praveenkumar:
Since release tag is created, I am going to merge this tomorrow.

PR #3379, opened 2022-10-11
[e2e] Patch little issues in e2e tests

jsliacan:
Things that I missed in the big PR:
- execute `crc setup` in a separate session (not to get stuck on progress bar), at least the 1st instance in each scenario
- remove bits of old & unused config skip checks on Windows
- split `registry` scenario in `story_openshfit` into 2: using `podman login`, using `oc login`
  - use podman-remote to make the scenario suitable for all platforms (requires moving code from integration utils to `crc/test/extended` - so these changes affect integration tests as well. 
- remove `story_registry.feature` as it is included in `story_openshift.feature` now
- update `cert_renewal.feature` to check after 13mo instead of 3mo (following cert validity being extended to 1y now).
- update the other `story_registry` Scenario to avoid the SCC issue, like so: https://github.com/crc-org/crc/pull/3424
- do not use crc-admin context to log in as sometimes this isn't available and it compromises the tests (set up a separate integration test for this)
- move timesync and cleanup actions from several Scenarios to Before/After hooks so they aren't skipped if a Step fails in front of them.


test/e2e/features/config.feature
```diff
@@ -97,19 +97,20 @@ Feature: Test configuration settings
...
             | skip-check-bundle-extracted     | true   | false  |
-            | skip-check-hyperv-installed     | true   | false  |
-            | skip-check-hyperv-switch        | true   | false  |
             | skip-check-user-in-hyperv-group | true   | false  |
-            | skip-check-windows-version      | true   | false  |
+
+    #| skip-check-hyperv-installed     | true   | false  |
+    #| skip-check-hyperv-switch        | true   | false  |
+    #| skip-check-administrator-user   | true   | false  |
+    #| skip-check-windows-version      | true   | false  |
```
praveenkumar, line 106: Why are those commented out?

test/e2e/features/story_openshift.feature
```diff
@@ -36,31 +33,30 @@ Feature: 3 Openshift stories
...
+	@darwin @linux @windows
+	Scenario: Mirror image to OpenShift image registry
+		# mirror
 		When executing "oc new-project testproj-img" succeeds
-		And executing "podman login -u kubeadmin -p $(oc whoami -t) default-route-openshift-image-registry.apps-crc.testing --tls-verify=false" succeeds
-		And executing "podman tag quay.io/centos7/httpd-24-centos7 default-route-openshift-image-registry.apps-crc.testing/testproj-img/hello:test" succeeds
-		And executing "podman push default-route-openshift-image-registry.apps-crc.testing/testproj-img/hello:test --tls-verify=false" succeeds
-		And executing "oc new-app testproj-img/hello:test" succeeds
-		When executing "oc rollout status deployment hello" succeeds
+		And  executing "oc registry login --insecure=true" succeeds
```
praveenkumar: `oc registry login` and `podman login` both have different usecase. A user might build the image locally before pushing to internal registry and run it instead just pulling it from remote and pushing to internal registry (where `oc image mirror` is much suitable). I think we still need `podman login` test to make sure user can able to login to internal registry without any issue.

adrianriobo: yeah we can keep both scnearios...also relying on podman remote the local build now could be covered on all 3 platforms (need to run `crc podman-evn` for mac and windows)

test/e2e/features/config.feature
adrianriobo: I think those are now handle by the installer, so `setup` does not take care of them. Not sure due to expected roadmap around future on installer if they may go back to `setup` or they should disappear as config options? @anjannath ^^

anjannath: I don't think they'll disappear as config options, these are still present in current `crc`

Only the `fix` part of the preflight checks were moved to the MSI, `crc start` and `crc setup` still runs these checks and having these config options allows users to skip the checks

adrianriobo: It will check and...does it fix them in case of fails?...on the other hand skipping them...does not show the skipping message anymore

anjannath: > It will check and...does it fix them in case of fails?

my mistake, only `crc start` runs these preflight checks, `crc setup` doesn't 

> on the other hand skipping them...does not show the skipping message anymore

for `crc start` its shown, but in this case we are only testing `crc setup` with these config options?

```
PS > crc config view
- skip-check-hyperv-installed           : true

PS > crc start
INFO Checking Windows release
INFO Checking Windows edition
WARN Skipping above check...                <---- here
INFO Checking if the daemon task is installed
INFO Checking if the daemon task is running
INFO Checking admin helper service is running
INFO Loading bundle: crc_hyperv_4.11.7_amd64..
[...]
```

adrianriobo: yeah only on `setup`...so now make sense why they are not showing the skipping message anymore

anjannath: @praveenkumar @adrianriobo just to be clear, this doesn't need any changes for the PR right?

test/e2e/features/story_openshift.feature
jsliacan: So to keep this here in e2e, the suggestion is that I clone the Scenario into two and one will be with `podman login` and the other with `oc registry login`? Using httpd-24 in both cases. Correct?

test/e2e/features/config.feature
adrianriobo: no as far as I understood config will be moved to integration anyway, and from there it would be decide if exec `start` to check these skips

test/e2e/features/story_openshift.feature
adrianriobo: yeah that will cover both scenarios  (+ `crc podman-env`) :) , which I think it is not covered at all either on e2e

jsliacan: Hmmm, I returned the `podman login` scenario, but to use `podman-remote` I'll have to wrap all the steps because podman is called a different name on each of the platforms :(. Also, `eval $(podman-env)` is platform specific too.

adrianriobo:
Just checking we got this targets on [Makefile](https://github.com/crc-org/crc/blob/main/Makefile#L230) so with new naming and removing of story this should be reviewed too

test/e2e/features/story_openshift.feature
```diff
@@ -36,31 +36,48 @@ Feature: 3 Openshift stories
...
+		When executing "oc logs deployment/httpd-24" succeeds
+		Then stdout should contain "httpd"
+		# cleanup
+		And executing "oc delete project testproj-img" succeeds
+
+	@darwin @linux @windows @local-img-to-registry @testdata @story_registry
+	Scenario: Pull image locally, push to registry, deploy
+		Given podman command is available
+		And executing "oc new-project testproj-img" succeeds
+		When pulling image "quay.io/centos7/httpd-24-centos7", logging in, and pushing image to registry succeeds
```
adrianriobo: May reword like:

_"When pushing local image "IMAGE_NAME" to internal registry"_

This will include all steps: login internal, pull image to local, tag and push to internal registry

test/e2e/testsuite/testsuite.go
```diff
@@ -324,12 +325,18 @@ func InitializeScenario(s *godog.ScenarioContext) {
 		ExecuteCommandWithExpectedExitStatus)
 	s.Step(`^execut(?:e|ing) single crc (.*) command (.*)$`,
 		ExecuteSingleCommandWithExpectedExitStatus)
+	s.Step(`^executing podman command (.*) (succeeds|fails)$`,
+		ExecutingPodmanCommandSucceedsFails)
 	s.Step(`^ensuring CRC cluster is running (succeeds|fails)$`,
 		EnsureCRCIsRunningSucceedsOrFails)
```
adrianriobo, line 380: Exec as single command on `setup`

adrianriobo (changes requested):

test/e2e/testsuite/testsuite.go
jsliacan: Podman commands are all "single". Maybe it'd have been good to call the function `ExecutingSinglePodmanCommandSucceedsFails`, but it seemed better to omit it.

test/e2e/features/story_openshift.feature
jsliacan: You could also create your local image instead of pulling it from somewhere, right? I don't mind changing it, just checking. For now, I've added `local` and `internal` to clarify what image and what registry.

test/e2e/features/story_openshift.feature
```diff
@@ -56,14 +56,15 @@ Feature: 3 Openshift stories
...
-		And executing "podman pull quay.io/centos7/httpd-24-centos7" succeeds
-		When executing "oc new-project testproj-img" succeeds
-		And executing "podman login -u kubeadmin -p $(oc whoami -t) default-route-openshift-image-registry.apps-crc.testing --tls-verify=false" succeeds
-		And executing "podman tag quay.io/centos7/httpd-24-centos7 default-route-openshift-image-registry.apps-crc.testing/testproj-img/hello:test" succeeds
-		And executing "podman push default-route-openshift-image-registry.apps-crc.testing/testproj-img/hello:test --tls-verify=false" succeeds
+	@darwin @linux @windows @local-img-to-registry @testdata @story_registry
+	Scenario: Pull image locally, push to registry, deploy
+		Given podman command is available
+		And executing "oc new-project testproj-img" succeeds
+		When pulling image "quay.io/centos7/httpd-24-centos7", logging in, and pushing local image to internal registry succeeds
```
praveenkumar, line 66: instead of this which is more messy why we can't just use `eval $(crc podman-env)` command direct from shell which will anyway add those env variables and then we should go as expected?

test/e2e/features/config.feature
```diff
@@ -41,7 +41,7 @@ Feature: Test configuration settings
 
     @linux @darwin @windows
     Scenario: CRC config checks (bundle version)
-        Given executing crc setup command succeeds
+        Given executing single crc setup command succeeds
```
praveenkumar, line 44: These changes are not part of this commit.

test/e2e/testsuite/testsuite.go
```diff
@@ -663,3 +670,84 @@ func DeletingPodSucceedsOrFails(expected string) error {
...
+		var err error
+
+		if runtime.GOOS == "windows" {
+			err = util.ExecuteCommandSucceedsOrFails("podman oc-env | Invoke-Expression", "succeeds")
+		} else {
+			err = util.ExecuteCommandSucceedsOrFails("eval $(podman oc-env)", "succeeds")
+		}
+
+		return err
+	*/
```
praveenkumar: These commented code should be removed if there is no use

test/e2e/features/story_openshift.feature
```diff
@@ -56,14 +56,15 @@ Feature: 3 Openshift stories
...
+		# And executing podman command "pull quay.io/centos7/httpd-24-centos7" succeeds
+		# And executing podman command "login -u kubeadmin -p $(oc whoami -t) default-route-openshift-image-registry.apps-crc.testing --tls-verify=false" succeeds
+		# And executing podman command "tag quay.io/centos7/httpd-24-centos7 default-route-openshift-image-registry.apps-crc.testing/testproj-img/hello:test" succeeds
+		# And executing podman command "push default-route-openshift-image-registry.apps-crc.testing/testproj-img/hello:test --tls-verify=false" succeeds
```
praveenkumar: We should remove all those comments if those are not required anymore.

adrianriobo:
There is one issue with the `Mirror image to OpenShift image registry` scenario pod is crashing exact same reason as #3424.

I think the easier would be to match the name of the image in the internal registry as on `Pull image locally, push to registry, deploy` scenario and the reuse the [`hello.yaml`](https://github.com/crc-org/crc/blob/e31372985dad1e72db38ab575cb34364b516a04c/test/e2e/features/story_openshift.feature#L64) in here too

adrianriobo:
we may can change using ["oc config use-context crc-admin"](https://github.com/crc-org/crc/blob/e31372985dad1e72db38ab575cb34364b516a04c/test/e2e/testsuite/testsuite.go#L533) as @praveenkumar mention on QE sync the addition of the context does not have any retry mechanism... not sure if we want to create an issue around it but using it when the context has not been added is creating false positives...anyway at least if we want we can isolate that command for testing it and change here to do a real `oc login` as that should work and avoid those false positives WDYT?

adrianriobo:
It is required to improve the hooks around scenarios, specially to clean them up:

* for cert rotation we need to ensure the clock is in sync after the scenario, we saw some cases where the scenario fail, then the clock is not synced up b/c that command is on a step which is skipped due to previous step failure.

* for basic scenario (with monitoring) we have similar [problem](https://s3.console.aws.amazon.com/s3/object/crcqe-asia?region=ap-south-1&prefix=nightly/ocp/4.12.0-rc.5/qe-results/test-results/e2e_2022-12-20_08-50-43.log) the cleanup of scneario should be on a hook, otherwise we could end up on a situation like the one from the lgos, the scenario for cluster with monitoring enabled fails, but cluster is not cleaned up (as that step is skipped) then next scneario cert rotation .. expects to start a cluster 13 months ahead to test the cert roration...but `crc start` is using previous cluster, so no cert rotation is being tested

jsliacan:
@adrianriobo I added another commit on top to address the issues you pointed out in your last comment:

I used tags `@timesync` and `@cleanup` to label Scenarios for which (respectively):
- time de-sync should happen in Before hook and time re-sync should happen in the After hook
- `crc cleanup` should be run in the After hook.

This lead to some changes in the Feature files (I removed the steps where this was done if the action was not part of the "story"). 

One warning is that in `story_openshift.feature`, we are treating Scenarios as ordered: `@cleanup` tag is only on the last one. This is incorrect design-wise, but I can't think of a quick way to avoid it (except introducing `@leave-running` tag and check for it in the `cleanup` After hook).

test/e2e/testsuite/testsuite.go
```diff
@@ -179,15 +179,64 @@ func InitializeScenario(s *godog.ScenarioContext) {
...
+				if err != nil {
+					fmt.Println(err)
+					os.Exit(1)
+				}
+				err = util.ExecuteCommand("sudo date -s '13 month'")
+				if err != nil {
+					fmt.Println(err)
+					os.Exit(1)
+				}
+				err = util.ExecuteCommandWithRetry(10, "1s", "virsh --readonly -c qemu:///system capabilities", "^<capabilities>", "succeeds")
```
adrianriobo: This is breaking the tests, command parameters are wrong, correct syntax would be:

```go
err = util.ExecuteCommandWithRetry(10, "1s", "virsh --readonly -c qemu:///system capabilities", "contains", "<capabilities>")
```

adrianriobo (changes requested):

test/e2e/testsuite/testsuite.go
adrianriobo: although not sure why are we exactly checking this here?

jsliacan: Thanks, I've changed it now. Didn't have time to run it locally yesterday, so didn't check. I'll leave it as `matches` and `^<capabilities>` to have it as it was [originally](https://github.com/crc-org/crc/blob/76d07a985c9a7be95beba9b9c796760b6c258817/test/e2e/features/cert_rotation.feature#L12).

I was asking the same thing the other day (why is it there?) and I was pointed to git blame :). I think the reason is here:

```
After the reordering, a short wait is needed for libvirt to become
available again, as `date -s "5 minutes" && virsh hostname` is failing
for a few seconds after the time jump.
```

test/e2e/testsuite/testsuite.go
```diff
@@ -178,20 +179,69 @@ func InitializeScenario(s *godog.ScenarioContext) {
...
 		return ctx, nil
 	})
 
+	s.After(func(ctx context.Context, sc *godog.Scenario, err error) (context.Context, error) {
+
+		for _, tag := range sc.Tags {
+
+			// move host's date 13 months back and turn timesync on
+			if tag.Name == "@timesync" {
+				err := util.ExecuteCommand("sudo date -s '-13 month")
```
adrianriobo: missing `'` after `month`

adrianriobo (changes requested):

test/e2e/features/story_openshift.feature
jsliacan: Done.

PR #3380, opened 2022-10-11
msi: ignore errors from custom action creating the smb shared dir

anjannath:
if an smb share with the same name ('crc-dir0') already exits the custom action fails marking the whole installation as failed

we can safely ignore this error as crc also checks that the share exists before trying to use it


adrianriobo:
Tests results for this PR on windows can be checked at [tests results](https://crcqe-asia.s3.amazonaws.com/nightly/crc/20221011/10-enterprise/qe-results/e2e-non-ux.xml)

anjannath:
there should also be a cleanup custom action here removing the shared folder name `crc-dir0` 

/hold

Issue #2876, opened 2021-12-09
[Task] Add CPU (cores/load) and Memory to `/api/status`


gbraad:
This can be a detailed path `status?detailed` or otherwise, but this can help providing a more detailed view of the state of the VM

![image](https://user-images.githubusercontent.com/1894/145342231-b998a16c-6244-4ab0-b82d-85fca1e612d9.png)

^^^ this is as shown in Cockpit.

PR #3384, opened 2022-10-13
Show memory usage in 'status' command output

evidolob:
**Relates to:** https://github.com/code-ready/crc/issues/2876

## Solution/Idea
The solution is pretty straightforward, PR is add two field for 'total' and 'used' RAM in `status` daemon API and print them.
Daemon implements getting RAM with execution and parsing output of `free -b` command. 

## Testing

1.  call `status` command while you have any crc vm running.
2. stdout should contains `RAM Usage` row, with number of bytes use and total for your vm, for example:

```
CRC VM:          Running
Podman:          4.2.0
RAM Usage:       168.4MB of 9.373GB
Disk Usage:      1.769GB of 32.74GB (Inside the CRC VM)
Cache Usage:     34.23GB
Cache Directory: /Users/some/.crc/cache
```

### DIFF
--- a/cmd/crc/cmd/status_test.go
+++ b/cmd/crc/cmd/status_test.go
@@ -64,6 +64,7 @@ func TestPlainStatus(t *testing.T) {
 	expected := `CRC VM:          Running
 OpenShift:       Running (v4.5.1)
 Podman:          3.3.1
+RAM Usage:       0B of 0B
 Disk Usage:      10GB of 20GB (Inside the CRC VM)
 Cache Usage:     10kB
 Cache Directory: %s
@@ -95,6 +96,7 @@ func TestStatusWithoutPodman(t *testing.T) {
 
 	expected := `CRC VM:          Running
 OpenShift:       Running (v4.5.1)
+RAM Usage:       0B of 0B
 Disk Usage:      10GB of 20GB (Inside the CRC VM)
 Cache Usage:     10kB
 Cache Directory: %s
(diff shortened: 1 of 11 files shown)


anjannath:
@evidolob have you run this with the tray? since the response from `/status` api is changed the tray might complain, or since its not removing any of the existing fields, it should be ok?

cmd/crc/cmd/status_test.go
```diff
@@ -170,3 +172,37 @@ func TestJsonStatusWithError(t *testing.T) {
...
+
+	client.On("Status").Return(apiClient.ClusterStatusResult{
+		CrcStatus:        string(state.Running),
+		OpenshiftStatus:  string(types.OpenshiftRunning),
+		OpenshiftVersion: "4.5.1",
+		DiskUse:          10_000_000_000,
+		DiskSize:         20_000_000_000,
+		RAMSize:          1_000_000,
+		RAMUse:           900_000,
+		Preset:           preset.OpenShift,
```
anjannath, line 192: nit: is this should be `preset.Podman`?

evidolob:
@anjannath Just test tray app, it is working as is, with this changes.
@gbraad Should I add RAM stats in to tray app?

gbraad:
Unclear for now. Let's do such modifications later.
The fate of the tray is still undecided.

cmd/crc/cmd/status_test.go
anjannath: ohh, i asked simply because the function was named `TestStatusWithMemoryPodman` but the preset was openshift

PR #3392, opened 2022-10-20
Fetch console URL form daemon for 'console' cmd

evidolob:
## Solution/Idea
As part of moving all commands to daemon, this PR move `console` command to daemon.
It also add new API method `Exists` which detects if VM exists, to keep an old behaviour/output

## Testing

`console` command should behave and print output exactly the same as before this PR


cmd/crc/cmd/root.go
```diff
@@ -115,6 +115,8 @@ func Execute() {
 	runPostrun()
 }
 
+//Deprecated: checkIfMachineMissing is deprecated.
+// checkIfMachineMissingWithDaemon should be used
```
praveenkumar, line 119: Some of the commands are still using it so not deprecated completely right?

cmd/crc/cmd/util.go
```diff
@@ -0,0 +1,17 @@
+package cmd
+
+import (
+	"github.com/code-ready/crc/pkg/crc/daemonclient"
+	crcErr "github.com/code-ready/crc/pkg/crc/errors"
+)
+
+func checkIfMachineMissingWithDaemon(client *daemonclient.Client) error {
```
praveenkumar, line 8: if every command is going to use than shouldn't this part of root.go

praveenkumar (commented):
I think this require 2 different commit
- Adding the vm-exists api
- moving console cmd to daemon side.

cmd/crc/cmd/util.go
```diff
@@ -0,0 +1,17 @@
...
+	exists, err := client.APIClient.Exists()
```
gbraad, line 9: I am not sure if we can assume `client *daemonclient.Client` has a valid value.
Perhaps safeguard this with an additional check?

pkg/crc/api/api_http_test.go
```diff
@@ -369,6 +369,12 @@ var testCases = []testCase{
 		request:  get("config?cpus"),
 		response: jSon(`{"Configs":{"cpus":4}}`),
 	},
+
+	//
+	{
+		request:  get("wm-exists"),
```
praveenkumar, line 375: Should it be `vm-exists` instead `wm-exists` ?

cmd/crc/cmd/root.go
evidolob: Yes, I'm going to delete `checkIfMachineMissing` after finishing moving cmd to daemon

evidolob:
I have a doubt about implementation of this, to be more clear, my doubt is about of having `checkIfMachineMissingWithDaemon` as replacing of `checkIfMachineMissing` on cmd side.

I think, it may be better if I move that check to daemon, but it may change the behaviour of daemon API, which now can provide more errors. 

@gbraad @praveenkumar WDYT?

cmd/crc/cmd/console.go
```diff
@@ -33,21 +33,22 @@ var consoleCmd = &cobra.Command{
...
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return runConsole(os.Stdout, newMachine(), consolePrintURL, consolePrintCredentials, outputFormat)
+		return runConsole(os.Stdout, daemonclient.New(), consolePrintURL, consolePrintCredentials, outputFormat)
 	},
 }
 
-func showConsole(client machine.Client) (*types.ConsoleResult, error) {
-	if err := checkIfMachineMissing(client); err != nil {
+func showConsole(client *daemonclient.Client) (*client.ConsoleResult, error) {
+	if err := checkIfMachineMissingWithDaemon(client); err != nil {
```
cfergeau: Instead of adding API to the daemon to check for the machine existence, have you considered moving the VM existence check to the remote `WebconsoleURL` method, and return a `ErrMachineNotExist` from these methods when the VM is not present?

evidolob: Yes, that what I mentioned here https://github.com/crc-org/crc/pull/3392#issuecomment-1291788816

Going to redo this PR

pkg/crc/api/handlers.go
```diff
@@ -151,12 +151,19 @@ func (h *Handler) Delete(c *context) error {
...
+		// error details also.
+		return err
+	}
 	res, err := h.Client.GetConsoleURL()
 	if err != nil {
 		return err
 	}
 	return c.JSON(http.StatusOK, client.ConsoleResult{
 		ClusterConfig: res.ClusterConfig,
+		State:         res.State,
```
cfergeau, line 166: If I understand correctly, this adds a new field to the http protocol? Not a problem at the moment, unless this can confuse the tray. If one day we allow a crc binary for version N to talk to a daemon from version X then we'll need to  be more careful with http protocol changes. But once again, not a problem in this PR :)

pkg/crc/api/handlers.go
```diff
@@ -151,12 +151,19 @@ func (h *Handler) Delete(c *context) error {
 }
 
 func (h *Handler) GetWebconsoleInfo(c *context) error {
+	if err := machine.CheckIfMachineMissing(h.Client); err != nil {
+		// In case of machine doesn't exist then consoleResult error
+		// should be updated so that when rendering the result it have
+		// error details also.
```
cfergeau, line 157: I tested `crc console` before/after this change and this does not seem to change its behaviour, it prints the same thing when a machine is missing?
There is one difference, it's the error message when the daemon is not running.

evidolob: Yes that only change that introduces by this PR is new error message in case if no daemon running.

gbraad: > it's the error message when the daemon is not running.

Is that the message your are expecting? If so, please "resolve conversation"

pkg/crc/api/handlers.go
gbraad: the tray will be deprecated over time. it does not see any active development, so I am not expecting this would be an issue going forward.

gbraad (approved):

PR #3403, opened 2022-11-02
test: use `T.TempDir` to create temporary test directory

Juneezee:
## Solution/Idea

A testing cleanup. 

We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	require.NoError(t, err)
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

## Proposed changes

Replace `ioutil.TempDir` with `t.TempDir` in all `*_test.go` files.

### DIFF
--- a/pkg/crc/ssh/keys_unix.go
+++ b/pkg/crc/ssh/keys_unix.go
@@ -30,6 +30,7 @@ func (kp *KeyPair) WriteToFile(privateKeyPath string, publicKeyPath string) erro
 		if err != nil {
 			return ErrUnableToWriteFile
 		}
+		defer f.Close()
 
 		if _, err := f.Write(v.Value); err != nil {
 			return ErrUnableToWriteFile
--- a/pkg/crc/ssh/keys_windows.go
+++ b/pkg/crc/ssh/keys_windows.go
@@ -29,6 +29,7 @@ func (kp *KeyPair) WriteToFile(privateKeyPath string, publicKeyPath string) erro
 		if err != nil {
 			return ErrUnableToWriteFile
 		}
+		defer f.Close()
 
 		if _, err := f.Write(v.Value); err != nil {
 			return ErrUnableToWriteFile
@@ -37,7 +38,6 @@ func (kp *KeyPair) WriteToFile(privateKeyPath string, publicKeyPath string) erro
 		if err := acl.Chmod(v.File, 0600); err != nil {
 			return err
 		}
-
 	}
 
 	return nil
--- a/pkg/os/util_test.go
+++ b/pkg/os/util_test.go
@@ -1,13 +1,11 @@
 package os
 
 import (
-	"io/ioutil"
 	"os"
+	"os/user"
 	"path/filepath"
 	"runtime"
 	"testing"
-
-	"os/user"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -30,9 +28,7 @@ func TestAddEnv(t *testing.T) {
 }
 
 func TestFileContentFuncs(t *testing.T) {
-	dir, err := ioutil.TempDir("", "filecontent")
-	assert.NoError(t, err)
-	defer os.RemoveAll(dir)
+	dir := t.TempDir()
 
 	filename := filepath.Join(dir, "testfile")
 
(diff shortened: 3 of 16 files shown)


pkg/os/util_test.go
```diff
@@ -81,6 +75,10 @@ func TestFileExists(t *testing.T) {
 
 	err = os.Chmod(dirname, 0000)
 	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, os.Chmod(dirname, 0777))
+	}()
+
```
cfergeau, line 81: Not clear if this change is related to this commit?

cfergeau, line 81: Not clear if this change is related to this commit?

pkg/crc/ssh/keys_unix.go
```diff
@@ -32,13 +32,16 @@ func (kp *KeyPair) WriteToFile(privateKeyPath string, publicKeyPath string) erro
...
+			_ = f.Close()
 			return ErrUnableToWriteFile
 		}
 
 		if err := f.Chmod(0600); err != nil {
+			_ = f.Close()
 			return err
 		}
 
+		_ = f.Close()
```
cfergeau: These changes are unexpected in this commit, they probably should be moved to the next commit (" test: fix failing TestGenerateSSHKey on Windows ")

pkg/crc/ssh/keys_windows.go
```diff
@@ -31,13 +31,16 @@ func (kp *KeyPair) WriteToFile(privateKeyPath string, publicKeyPath string) erro
 		}
 
 		if _, err := f.Write(v.Value); err != nil {
+			_ = f.Close()
```
cfergeau: You could use `defer` to avoid the need to put this in all code paths, especially as you are ignoring the error from `Close`:
```
		f, err := os.Create(v.File)
		if err != nil {
			return ErrUnableToWriteFile
		}
		defer f.Close()
```

However, see https://www.sobyte.net/post/2022-01/golang-defer-file-close/ regarding `Close()` errors.
If I remember correctly, what we do in other parts of crc is:
```
		f, err := os.Create(v.File)
		if err != nil {
			return ErrUnableToWriteFile
		}
		defer f.Close()
 		//more code
		return f.Close() // instead of return nil
```

cfergeau (commented):
Thanks  for the PR, these are nice cleanups/fixes!

pkg/os/util_test.go
Juneezee: This change is required because we have changed the permission of `dirname` to `000`. We need to revert it back for `t.TempDir` to perform the cleanup successfully.

https://github.com/crc-org/crc/blob/1425c16d6ca9ae043f4ddcb817f7317443aa70d2/pkg/os/util_test.go#L82-L83

pkg/crc/ssh/keys_unix.go
Juneezee: Changes moved to next commit

pkg/crc/ssh/keys_windows.go
Juneezee: Amended the commit to use `defer f.Close()`

pkg/crc/ssh/keys_unix.go
```diff
@@ -38,7 +38,6 @@ func (kp *KeyPair) WriteToFile(privateKeyPath string, publicKeyPath string) erro
 		if err := f.Chmod(0600); err != nil {
 			return err
 		}
-
```
cfergeau: Could you also drop this blank line from the first commit?

cfergeau (approved):
Looks good to me, just one nit regarding a blank line removal in the first commit.
I'll let CI run, and then this can be merged.

pkg/crc/ssh/keys_unix.go
Juneezee: Yup, no problem

PR #3409, opened 2022-11-09
Allow to use file sharing for mac 13.x

praveenkumar:


pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -471,5 +471,5 @@ func (d *Driver) supportsVirtiofs() bool {
 	if err != nil {
 		return false
 	}
-	return strings.HasPrefix(macosVersion, "12.")
+	return strings.HasPrefix(macosVersion, "12.") || strings.HasPrefix(macosVersion, "13.")
```
cfergeau: It would be more future proof to use something like https://github.com/Code-Hex/vz/blob/6e20b48bc30d888e13576559826867b8ff4222c0/osversion.go
If we add 13 now, we'll need to add 14 in a year, ...

praveenkumar: @cfergeau sure, I was just checking what would be the best way to do it for future proof. Will update the PR 👍🏼

anjannath:
need to run `make fmt`, the CI fails with the following error:

```
pkg/drivers/vfkit/driver_darwin.go:22: File is not `goimports`-ed (goimports)
[1400](https://github.com/crc-org/crc/actions/runs/3439366462/jobs/5736575892#step:4:1401)
	"github.com/crc-org/crc/pkg/crc/logging"
[1401](https://github.com/crc-org/crc/actions/runs/3439366462/jobs/5736575892#step:4:1402)
make: *** [cross-lint] Error 1
```

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -467,9 +468,24 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
-	return strings.HasPrefix(macosVersion, "12.")
+	return supportsVirtioFS
+}
+
+func macosNewerThan(targetVersion float64) (bool, error) {
+	osVer, err := syscall.Sysctl("kern.osproductversion")
+	if err != nil {
+		return false, err
+	}
+	logging.Debugf("Mac version is: %s", osVer)
```
cfergeau: `kern.osproductversion is: %s` might be better?

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -467,9 +468,24 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
+	osVerArray := strings.SplitAfterN(osVer, ".", 1)
+	logging.Debugf("Mac major version is: %s", osVerArray)
+	version, err := strconv.ParseFloat(osVerArray[0], 64)
```
cfergeau: This fails to parse `12.6.1` (and most likely `13.0.1` will have a similar failure). I'm unconvinced a float is a good way to represent the macOS version, I'd either go with an int for just the major version, or with `semver.NewVersion()` if we want the full version.

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -467,9 +468,24 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
+	return supportsVirtioFS
+}
+
+func macosNewerThan(targetVersion float64) (bool, error) {
+	osVer, err := syscall.Sysctl("kern.osproductversion")
+	if err != nil {
+		return false, err
+	}
+	logging.Debugf("Mac version is: %s", osVer)
+	osVerArray := strings.SplitAfterN(osVer, ".", 1)
```
cfergeau: My understanding of `SplitAfterN( ..., 1)` is that it will return a single value array with the full string, so the log saying `osVerArray` contains the major version is incorrect.

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -467,9 +468,24 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
 	}
-	return strings.HasPrefix(macosVersion, "12.")
+	return supportsVirtioFS
+}
+
+func macosNewerThan(targetVersion float64) (bool, error) {
+	osVer, err := syscall.Sysctl("kern.osproductversion")
+	if err != nil {
+		return false, err
+	}
```
cfergeau, line 493: Fwiw, this call could be wrapped in a `var osproductversionOnce = ...`. This would ensure it's only called once for the lifetime of a crc process, and it also makes it easy to override in unit tests for this `macosNewerThan` helper.

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -467,9 +468,24 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
+	supportsVirtioFS, err := macosNewerThan(11.0)
 	if err != nil {
-		return false
+		log.Debugf("Not able to compare version: %v", err)
 	}
-	return strings.HasPrefix(macosVersion, "12.")
+	return supportsVirtioFS
+}
+
+func macosNewerThan(targetVersion float64) (bool, error) {
```
gbraad: don't think this is specific for the driver. sounds more like OS level utils stuff.

pkg/drivers/vfkit/driver_darwin.go
gbraad: Agree. A float sounds awkward

pkg/drivers/vfkit/driver_darwin.go
praveenkumar: @cfergeau Done

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -466,10 +468,29 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
 	return proc.SendSignal(s)
 }
 
+var curVersionOnce = func() (string, error) { return syscall.Sysctl("kern.osproductversion") }
+
 func (d *Driver) supportsVirtiofs() bool {
-	macosVersion, _, err := crcos.RunWithDefaultLocale("sw_vers", "-productVersion")
+	supportsVirtioFS, err := macosNewerThan(curVersionOnce, "11.0.0")
```
cfergeau: When talking about 'once', I was referring to https://pkg.go.dev/sync#Once (sorry for not being explicit about it, I had this in mind because of other code I worked on, and was far too quick when mentioning it in the review :( ).
I hinted at this so that a test case can be added at the same time ;) I don't think it's required to pass `curVersionOnce` as an arg, since it's a global var we could override it directly in the tests. 
If you don't add a unit test at the same time, I would just remove this for onw.

cfergeau: I don' think `macosNewerThan` is the right API here, `macosAtLeast` should be more correct.
macosNewerThan("11.0.0") means that "11.0.0" is to be excluded, which is not obvious/natural to me, I need to think twice about it. And my understanding is that `11.0.1` would pass this check, which is incorrect, the file sharing API is macOS12+

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -466,10 +469,35 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
+)
+
 func (d *Driver) supportsVirtiofs() bool {
-	macosVersion, _, err := crcos.RunWithDefaultLocale("sw_vers", "-productVersion")
-	if err != nil {
+	once.Do(func() {
+		macCurrentVersion, errGettingVersion = syscall.Sysctl("kern.osproductversion")
+		logging.Debugf("kern.osproductversion is: %s", macCurrentVersion)
+	})
+	if errGettingVersion != nil {
```
cfergeau, line 491: This belongs in `macosAtLeast`

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -466,10 +469,35 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
+	}
+	return supportsVirtioFS
+}
+
+func macosAtLeast(curVersion string, targetVersion string) (bool, error) {
+	cVersion, err := semver.NewVersion(curVersion)
+	if err != nil {
+		return false, errors.Wrap(err, fmt.Sprintf("cannot parse %s", curVersion))
+	}
+	tarVersion, err := semver.NewVersion(targetVersion)
```
cfergeau: `tarVersion` makes me think of the `tar` archiving format, not of target. You could go with `targetVersionStr` and `targetVersion`.

cfergeau (approved):

PR #3412, opened 2022-11-10
cmd/version: OpenShift version should not be preset dependent

cfergeau:
This fixes this:
$ crc config set preset podman
$ crc version
CRC version: 2.10.2+e04852790
OpenShift version: 4.2.0
Podman version: 4.2.0

Even taking into account OKD, an OKD version is not a valid OpenShift
version, so it's misleading to show it as such.

### DIFF
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -79,7 +79,7 @@ func runPrerun(cmd *cobra.Command) error {
 	}
 	logging.InitLogrus(logFile)
 
-	for _, str := range defaultVersion(crcConfig.GetPreset(config)).lines() {
+	for _, str := range defaultVersion().lines() {
 		logging.Debugf(str)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


gbraad:
confused by the title: "should not be preset dependent" ?

It should be dependent, as in: "Only show for bundles that support this". Podman should never show this field.

praveenkumar:
> confused by the title: "should not be preset dependent" ?
> 
> It should be dependent, as in: "Only show for bundles that support this". Podman should never show this field.

I think openshift version which is part of version shouldn't be preset dependent.

cfergeau:
> confused by the title: "should not be preset dependent" ?
> 
> It should be dependent, as in: "Only show for bundles that support this". Podman should never show this field.

At the moment, the behaviour is to show `OpenShift Version: ` followed by the default bundle version for the active preset, so it will show podman bundle version or okd bundle version as `OpenShift Version`.

PR #3416, opened 2022-11-15
vfkit: Enable timesync

cfergeau:
On macOS, after a host suspend, the guest clock will be out of sync.
 vfkit 0.0.4 has a --timesync feature, which resyncs the guest clock when the
host comes out of sleep. This PR makes use of it.

### DIFF
--- a/pkg/crc/constants/constants_darwin.go
+++ b/pkg/crc/constants/constants_darwin.go
@@ -11,6 +11,7 @@ const (
 	PodmanRemoteExecutableName = "podman"
 	TrayExecutableName         = "Red Hat OpenShift Local.app"
 	DaemonAgentLabel           = "com.redhat.crc.daemon"
+	QemuGuestAgentPort         = 1234
 )
 
 var (
--- a/pkg/crc/machine/vfkit/driver_darwin.go
+++ b/pkg/crc/machine/vfkit/driver_darwin.go
@@ -22,7 +22,9 @@ func CreateHost(machineConfig config.MachineConfig) *vfkit.Driver {
 	vfDriver.VfkitPath = filepath.Join(constants.BinDir(), VfkitCommand)
 
 	vfDriver.VirtioNet = machineConfig.NetworkMode == network.SystemNetworkingMode
+
 	vfDriver.VsockPath = constants.TapSocketPath
+	vfDriver.DaemonVsockPort = constants.DaemonVsockPort
 
 	vfDriver.SharedDirs = configureShareDirs(machineConfig)
 
--- a/pkg/crc/machine/vfkit/driver_darwin.go
+++ b/pkg/crc/machine/vfkit/driver_darwin.go
@@ -26,6 +26,8 @@ func CreateHost(machineConfig config.MachineConfig) *vfkit.Driver {
 	vfDriver.VsockPath = constants.TapSocketPath
 	vfDriver.DaemonVsockPort = constants.DaemonVsockPort
 
+	vfDriver.QemuGAVsockPort = constants.QemuGuestAgentPort
+
 	vfDriver.SharedDirs = configureShareDirs(machineConfig)
 
 	return vfDriver
--- a/pkg/drivers/vfkit/driver_darwin.go
+++ b/pkg/drivers/vfkit/driver_darwin.go
@@ -242,7 +242,7 @@ func (d *Driver) Start() error {
 
 	// virtio-vsock device
 	const vsockPort = 1024
-	dev, err = client.VirtioVsockNew(vsockPort, d.VsockPath)
+	dev, err = client.VirtioVsockNew(vsockPort, d.VsockPath, true)
 	if err != nil {
 		return err
 	}
--- a/pkg/drivers/vfkit/driver_darwin.go
+++ b/pkg/drivers/vfkit/driver_darwin.go
@@ -50,6 +50,7 @@ type Driver struct {
 
 	VsockPath       string
 	DaemonVsockPort uint
+	QemuGAVsockPort uint
 }
 
 func NewDriver(hostName, storePath string) *Driver {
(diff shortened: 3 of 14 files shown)


cfergeau:
To test this, you need these podman bundles https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/crc-org_snc/595/pull-ci-crc-org-snc-podman-e2e-snc/1588460383008460800/artifacts/e2e-snc/gather-snc/artifacts/bundles.txt 
If testing on a macbook, it's easier to test if it's not plugged in, otherwise I never know when it sleeps or not. Suspend the laptop for a few minutes, connect to the crc VM, and check `date`. It should be up to date, not lagging by the amount of time the mac slept.

pkg/crc/constants/constants_darwin.go
```diff
@@ -11,6 +11,7 @@ const (
 	PodmanRemoteExecutableName = "podman"
 	TrayExecutableName         = "Red Hat OpenShift Local.app"
 	DaemonAgentLabel           = "com.redhat.crc.daemon"
+	QemuGAPort                 = 1234
```
gbraad: Can't we just call it `QemuGuestAgentPort` ?
Which will still be shorter than `PodmanRemoteExecutableName`...

pkg/crc/machine/vfkit/driver_darwin.go
```diff
@@ -22,7 +22,11 @@ func CreateHost(machineConfig config.MachineConfig) *vfkit.Driver {
 	vfDriver.VfkitPath = filepath.Join(constants.BinDir(), VfkitCommand)
 
 	vfDriver.VirtioNet = machineConfig.NetworkMode == network.SystemNetworkingMode
+
 	vfDriver.VsockPath = constants.TapSocketPath
+	vfDriver.VsockDaemonPort = constants.DaemonVsockPort
```
gbraad: inconsistent use of `VsockDaemonPort` and `DaemonVsockPort`

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -248,6 +254,20 @@ func (d *Driver) Start() error {
 		return err
 	}
 
+	// when loading a VM created by a crc version predating this commit,
+	// d.VsockQemuGAPort will be missing from ~/.crc/machines/crc/config.json
+	// In such a case, assume the VM will not support time sync
+	if d.VsockQemuGAPort != 0 {
```
gbraad: makes sense...

perhaps something that can also be detected in the main binary ?

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -248,6 +254,20 @@ func (d *Driver) Start() error {
...
+		timesync, err := client.TimeSyncNew(d.VsockQemuGAPort)
```
gbraad: the device we add is not really timesync, but the guest agent. timesyncing is just one (only) of the features used.

anjannath:
> To test this, you need these podman bundles https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/crc-org_snc/595/pull-ci-crc-org-snc-podman-e2e-snc/1588460383008460800/artifacts/e2e-snc/gather-snc/artifacts/bundles.txt

this is not available for downloading anymore it seems, getting error  when visiting the bundle links from the file

```
Anonymous caller does not have storage.objects.get access to the Google Cloud Storage object. Permission 'storage.objects.get' denied on resource (or it may not exist)
```

pkg/crc/constants/constants_darwin.go
cfergeau: I changed it.

pkg/drivers/vfkit/driver_darwin.go
cfergeau: I prefer to detect it there, having a workaround for crc's behaviour in vfkit's code would be odd.

pkg/drivers/vfkit/driver_darwin.go
cfergeau: it's more 'timesync' than just guest agent. The guest agent connection is not exposed externally, it's a net.Conn instance internal to vfkit. And vfkit monitors for wake up events, and sends a `set-time $current_time` command to the guest agent if present.

pkg/crc/machine/vfkit/driver_darwin.go
cfergeau: Fixed.

pkg/drivers/vfkit/driver_darwin.go
gbraad: If that is the case, naming should represent that; `GuestAgentForTimesync` in some form inside the codebase.

pkg/drivers/vfkit/driver_darwin.go
gbraad: The problem is that you can't report back from the driver that this isn't possible.
We would like to inform the user about this error in some form... or we assume this will never happen?

cfergeau: It's something specific to the machine driver. With an older crc, you run `crc start; crc stop`, the machine driver state is serialized to json, and does not contain a `VsockQemuGAPort` field because this did not exist when it was built.
Then `crc start` is ran with a new crc build which looks for `VsockQemuGAPort` in the serialized json. It will get 0 instead of the expected value since the field was not present.

pkg/drivers/vfkit/driver_darwin.go
cfergeau: Or maybe `VsockQemuGAPort` -> `VsockTimesyncPort`?

praveenkumar:
Created bundle manually and tested it, works as expected.

Without this PR  (current main), time skew happens
```
== Before putting laptop to sleep==
$ date -u
Thu Dec  1 11:43:59 UTC 2022
$ ssh crc -- date  
Thu Dec  1 11:44:04 UTC 2022

==After waking from sleep==
$  date -u        
Thu Dec  1 13:04:37 UTC 2022
$ ssh crc -- date
Thu Dec  1 11:45:08 UTC 2022
```

With this PR
```
== Before putting laptop to sleep==
$ date -u
Thu Dec  1 10:49:02 UTC 2022
$ ssh crc -- date
Thu Dec  1 10:49:06 UTC 2022

==After waking from sleep==
$ date -u
Thu Dec  1 11:33:51 UTC 2022
$ ssh crc -- date
Thu Dec  1 11:33:55 UTC 2022
```

Issue #3354, opened 2022-09-20
Proxy support to pull the bundle image from quay


praveenkumar:
We don't have any test or even implementation around pulling the bundle images (podman/okd) from the quay in case user is behind the proxy. This will help us to move one step closer to have our OCP bundles also as part of quay.

anjannath:
If system wide proxy is configured either using the OS default network manager or using the env variables `http_proxy` and `https_proxy` `crc setup` will be able to pull in the bundle image.

containers/image doesn't support customizing the `httpClient` used, it uses the system proxy configs, see: https://github.com/containers/image/issues/1194 and https://github.com/containers/image/pull/537

anjannath:
If we add a scenario like below will it be enough? or maybe we should check the logs in the `squid` container as well to make sure the request actually went through it and not directly to the internet

```cucumber
Background: Setup the proxy container using podman
         * executing "podman run --name squid -d -p 3128:3128 quay.io/crcont/squid" succeeds

    Scenario: Setup CRC behind proxy
        * executing "export http_proxy='http://192.168.130.1:3128'" succeeds
        * executing "export https_proxy='http://192.168.130.1:3128'" succeeds
        Given executing "crc config set preset podman" succeeds
        Then executing crc setup command succeeds
        When executing "crc config get preset" succeeds
        Then stdout should contain "podman"
        # cleanup proxy and preset settings from config
        And executing "crc config unset http-proxy" succeeds
        And executing "crc config unset https-proxy" succeeds
        And executing "crc config unset preset" succeeds
```
cc @jsliacan

jsliacan:
Sounds great, just a couple of comments.

- If we add this Scenario to `crc/test/e2e/features/proxy.feature`, then we can skip `Background` in your snippet.
- Maybe we need to make sure that there's no podman bundle in `~/.crc/cache` or wherever it lives. So the download happens for sure.
- Should we also do `crc start`? It takes no time with podman bundle anyways. If it exits with 0, it's a good confirmation. Probably not needed, but it's a one liner `And executing "crc start" succeeds`. 

Also, I need to check what `podman run <img>` does if that was done once already (return 1, 0, do I have 2 containers? etc.) The `Background` block in `proxy.feature` might not be robust to this and we'll have to add a check. 

Thanks!

anjannath:
> If we add this Scenario to crc/test/e2e/features/proxy.feature, then we can skip Background in your snippet.

[...]

PR #3417, opened 2022-11-16
[e2e] Test fetching podman bundle when behind proxy

jsliacan:
**Fixes:** Issue #3354 

- Should run as part of usual e2e tests without issues. 
- 2nd commit cleans up the Feature file and moves setup/cleanup to Before/After Scenario hooks

**Note**: there is no check that proxy isn't circumvented. This will be fine if proxy will be enforced through environment setup.

### DIFF
--- a/test/e2e/features/proxy.feature
+++ b/test/e2e/features/proxy.feature
@@ -1,14 +1,13 @@
 @proxy @linux
 Feature: Behind proxy test
 
-    User starts CRC behind a proxy. They expect a successful start
-    and to be able to deploy an app and check its accessibility.
+    Check CRC use behind proxy
 
     Background: Setup the proxy container using podman
         * executing "podman run --name squid -d -p 3128:3128 quay.io/crcont/squid" succeeds
 
     @cleanup
-    Scenario: Start CRC behind proxy
+    Scenario: Start CRC behind proxy under openshift preset
         Given executing single crc setup command succeeds
         And  executing "crc config set http-proxy http://192.168.130.1:3128" succeeds
         Then executing "crc config set https-proxy http://192.168.130.1:3128" succeeds
@@ -29,3 +28,22 @@ Feature: Behind proxy test
         And executing crc cleanup command succeeds
 
 
+    Scenario: Cache podman bundle behind proxy under podman preset
+        * executing "crc config set http-proxy http://192.168.130.1:3128" succeeds
+        * executing "crc config set https-proxy http://192.168.130.1:3128" succeeds
+        * removing podman bundle from cache succeeds
+<<<<<<< HEAD
+        Given executing "crc config set preset podman" succeeds
+        Then executing single crc setup command succeeds
+        And podman bundle is cached
+        # cleanup proxy and preset settings from config
+        * executing "crc config unset http-proxy" succeeds
+        * executing "crc config unset https-proxy" succeeds
+        * executing "crc config unset preset" succeeds
+        * executing "podman stop squid" succeeds
+        * executing "podman rm squid" succeeds
+=======
+        * executing "crc config set preset podman" succeeds
+        When executing single crc setup command succeeds
+        Then executing "crc start" succeeds
+>>>>>>> 5c6d0b94 (Trying a pure start for podman preset)
--- a/test/e2e/features/proxy.feature
+++ b/test/e2e/features/proxy.feature
@@ -3,47 +3,20 @@ Feature: Behind proxy test
 
     Check CRC use behind proxy
 
-    Background: Setup the proxy container using podman
-        * executing "podman run --name squid -d -p 3128:3128 quay.io/crcont/squid" succeeds
-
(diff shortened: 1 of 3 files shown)


test/e2e/features/proxy.feature
```diff
@@ -29,3 +28,22 @@ Feature: Behind proxy test
         And executing crc cleanup command succeeds
 
 
+    Scenario: Cache podman bundle behind proxy under podman preset
+        * executing "crc config set http-proxy http://192.168.130.1:3128" succeeds
+        * executing "crc config set https-proxy http://192.168.130.1:3128" succeeds
+        * removing podman bundle from cache succeeds
+<<<<<<< HEAD
```
cfergeau: Left-over conflict markers.

test/e2e/testsuite/testsuite.go
```diff
@@ -539,6 +549,54 @@ func FileExistsInCRCHome(fileName string) error {
...
+func BundleIsCachedOrNot(presetName string, isOrNot string) error {
+
+	bundle := ""
+	if presetName == "podman" {
+		bundle = constants.GetDefaultBundle(preset.Podman)
+	} else {
+		bundle = constants.GetDefaultBundle(preset.OpenShift)
+	}
+
+	theFile := filepath.Join(CRCHome, "cache", bundle)
```
cfergeau: This might deserve a `getBundlePath(presetName)` helper.

test/e2e/testsuite/testsuite.go
```diff
@@ -539,6 +549,54 @@ func FileExistsInCRCHome(fileName string) error {
...
+
+	return nil
+}
+
+func RemoveBundleFromCache(presetName string) error {
+
+	bundle := ""
+	if presetName == "podman" {
+		bundle := constants.GetDefaultBundle(preset.Podman)
+		dirName := bundle[:len(bundle)-10] // remove `.crcbundle` suffix
```
cfergeau: `dirName = strings.TrimSuffix(bundle, ".crcbundle")`
This can be moved outside of the `if{}` block as you have the same code in botoh branches.

test/e2e/testsuite/testsuite.go
```diff
@@ -539,6 +549,54 @@ func FileExistsInCRCHome(fileName string) error {
...
+		theBundle = filepath.Join(CRCHome, "cache", bundle)
+	} else {
+		bundle := constants.GetDefaultBundle(preset.OpenShift)
+		dirName := bundle[:len(bundle)-10] // remove `.crcbundle` suffix
+		theFolder = filepath.Join(CRCHome, "cache", dirName)
+		theBundle = filepath.Join(CRCHome, "cache", bundle)
+	}
+
+	theFile := filepath.Join(CRCHome, "cache", bundle)
+	err := os.RemoveAll(theFile)
```
cfergeau: `theFolder`/`theBundle` are unused.

test/e2e/testsuite/testsuite.go
```diff
@@ -253,6 +279,46 @@ func InitializeScenario(s *godog.ScenarioContext) {
...
+				}
+
+				err = crcCmd.UnsetConfigPropertySucceedsOrFails("https-proxy", "succeeds") // unsetting property that is not set gives 0 exitcode, so this works
+				if err != nil {
+					fmt.Println(err)
+					os.Exit(1)
+				}
+
+				// crc oc-env sets these behind quietly if http(s)-proxy is set in crc config
+				err = util.ExecuteCommand("unset HTTP_PROXY HTTPS_PROXY NO_PROXY") // unsetting property that is not set gives 0 exitcode, so this works
```
cfergeau: Could this use https://pkg.go.dev/os#Unsetenv instead?

test/e2e/features/proxy.feature
jsliacan: Oops, sorry.

test/e2e/testsuite/testsuite.go
jsliacan: I put that function in the `util` package in `extended` folder. I then needed to redefine `CRCHome` variable whose scope didn't span to that package. Hopefully that's OK.

test/e2e/testsuite/testsuite.go
jsliacan: `TrimSuffix`, good idea - it looks much less offensive that way! Reorganized that `if`  as you suggested (I think).

test/e2e/testsuite/testsuite.go
jsliacan: Not sure about this one, but I think as it's now after changes there shouldn't be any unused vars.

test/e2e/testsuite/testsuite.go
jsliacan: I changed it as you suggest, but then I have to do it 3x (`os.Unsetenv` seems to take 1 arg only, and it cannot be multiple vars in 1 string). It functions OK though, I think.

PR #3422, opened 2022-11-18
Update okd bundle to 4.11.0-0.okd-2022-11-05-030711

praveenkumar:


cfergeau (approved):
More than 10% growth in bundle size in what should be a minor update :-/

PR #3424, opened 2022-11-18
[e2e] fix SCC constraints on deployments 

adrianriobo:
previously e2e samples use default templates to create deployments, now with new [SCC contraints](https://cloud.redhat.com/blog/pod-security-admission-in-openshift-4.11) there are some extra information on security context which can not be added from oc cli. To be able to add the required security context requirements this PR switches from using oc templating mechanism to fixed specs.

### DIFF
--- a/test/e2e/features/story_openshift.feature
+++ b/test/e2e/features/story_openshift.feature
@@ -13,7 +13,7 @@ Feature: 3 Openshift stories
 		#And ensuring user is logged in succeeds
 		Given checking that CRC is running
 		Then executing "oc new-project testproj" succeeds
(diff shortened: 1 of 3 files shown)


cfergeau:
What is surprising is that this broke between OpenShift 4.11.7 and 4.11.13. I would expect such changes to happen on a .0 release :-/

adrianriobo:
yeah that is really annoying...as they claim it should only warning on 4.11.x, but definitely do not change the behavior with the uid

adrianriobo:
hold need to add some custom SCC for allowing that uid, otherwise default SCCs are failing as uid should be on a higher range of ids.

anjannath (approved):

PR #3431, opened 2022-11-28
[e2e] e2e container fix ssh cmd for linux platform

adrianriobo:
this is a following PR for #3347 where ssh command tried to emulate a local session with `su`, that approach is working fine on darwin platform but is not working for linux. This PR will use that approach only for darwin and leave windows and linux as they were before PR #3347

### DIFF
--- a/images/build-e2e/entrypoint.sh
+++ b/images/build-e2e/entrypoint.sh
@@ -108,17 +108,17 @@ fi
 if [ "${CLEANUP_HOME}" = "false" ]; then
     OPTIONS+="--cleanup-home=false "
 fi
-if [ -n "${INSTALLER_PATH}" ]; then
-    OPTIONS+="--installer-path=${INSTALLER_PATH} "
+
+TAGS="@${PLATFORM}"
+if [ -n "${E2E_TAG_EXPRESSION}" ]; then
+    TAGS+=" && ${E2E_TAG_EXPRESSION}"
 fi
-if [ -n "${USER_PASSWORD}" ]; then
-    OPTIONS+="--user-password=${USER_PASSWORD} "
-fi
-OPTIONS+="--godog.tags=\"@${PLATFORM} && ${E2E_TAG_EXPRESSION}\" --godog.format=junit "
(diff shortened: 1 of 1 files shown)


images/build-e2e/entrypoint.sh
```diff
@@ -114,11 +114,16 @@ fi
...
-    BINARY_EXEC+="cd ${EXECUTION_FOLDER}/bin && ./${BINARY} ${OPTIONS} > ${RESULTS_FILE}.results"
+TAGS="@${PLATFORM}"
+if [ -n "${E2E_TAG_EXPRESSION}" ]; then
+    TAGS+=" && ${E2E_TAG_EXPRESSION}"
+fi
+if [[ ${PLATFORM} == 'darwin' ]]; then
+    OPTIONS+="--godog.tags=\\\"${TAGS}\\\" --godog.format=junit "
+    BINARY_EXEC+="sudo su - ${TARGET_HOST_USERNAME} -c \"PATH=$PATH:/usr/local/bin cd ${EXECUTION_FOLDER}/bin && ./${BINARY} ${OPTIONS} > ${RESULTS_FILE}.results\""
 else
-    BINARY_EXEC+="sudo su - ${TARGET_HOST_USERNAME} -c \"cd ${EXECUTION_FOLDER}/bin && ./${BINARY} ${OPTIONS} > ${RESULTS_FILE}.results\""
```
praveenkumar, line 121: When this fails what is the error you are getting?

adrianriobo: On linux I think the issue is related with systemd user services and how session is managed with `su`

adrianriobo: on darwin I still would not be able to find the exact setup which is causing this requirement. On m1 there is no security profile or param which is shrinking the remote session..and the only way I found out to overcome it is using the `su` to emulate a local session but again depending on macos version default bash is different so again envs are not set the same :(, this is why I have to force the `/url/local/bin` as part of $PATH

PR #3432, opened 2022-11-28
doc: update odo commands used in the docs

### LINKED ISSUE #3418: [doc] Use `odo` commands that works with latest `odo` release
### General information

The [Deploying sample application with `odo`](https://crc.dev/crc/#deploying-sample-application-with-odo_gsg) section of the _Getting started Guide_ uses commands that does not work with latest release of `odo`

the guide asks users to download and setup the latest version of `odo` but the commands used in the examples only work for v2 odo and doesn't work with v3 odo, see: https://github.com/redhat-developer/odo/issues/6312

We need to review this section and update the commands to work with the latest `odo`

anjannath:
since odo v3 there are changes in the CLI of `odo` this uses the new commands to deploy a sample application using odo

ref: https://github.com/redhat-developer/odo/issues/6312

Fixes: #3418

### DIFF
--- a/docs/source/topics/proc_deploying-sample-application-with-odo.adoc
+++ b/docs/source/topics/proc_deploying-sample-application-with-odo.adoc
@@ -26,7 +26,7 @@ $ odo login -u developer -p developer
 +
 [subs="+quotes,attributes"]
 ----
-$ odo project create sample-app
+$ odo create project sample-app
 ----
 
 . Create a directory for your components:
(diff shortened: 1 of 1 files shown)


docs/source/topics/proc_deploying-sample-application-with-odo.adoc
```diff
@@ -37,45 +37,22 @@ $ mkdir sample-app
...
-. List the URLs and check the desired URL for the component:
 +
-[subs="+quotes,attributes"]
-----
-$ odo url list
-----
-
-. View the deployed application using the generated URL.
+You can now access the application in your local browser using the URL from the output of the above command and start your development loop.
+[command]`odo` will watch for changes and push the code for real-time updates.
```
cfergeau, line 55: imo "developing continuously", "start your development loop" is a bit mysterious/begs for more details. Maybe we should just link to https://odo.dev/docs/user-guides/quickstart/nodejs and only detail the "Login to OpenShift Cluster" step as it's a bit more crc specific?

anjannath:
(will reopen when i update the docs)

PR #3433, opened 2022-11-28
Remove "Wildcard DNS check failed" warning when it is expected

cfergeau:
Most of this PR is about moving CheckCRCLocalDNSReachableFromHost to the `dns` package and slightly simplifying its code.
This removes the "Wildcard DNS check failed" warning from user mode networking since it is expected to fail in this scenario.
This also adds a "Downloading ..." info line before starting a lengthy bundle download.


cfergeau:
The easiest way to test is probably on linux with user-mode networking. 
Before This PR, usermode networking would print a warning. After this PR it no longer prints it.

anjannath (approved):

PR #3445, opened 2022-12-07
Add user provided timeouts for slow systems

praveenkumar:
## Solution/Idea

Enable a way for user to provide different timeouts in case system is slow and crc takes more time than default.

## Proposed changes

Introduced following env variables
- CRC_OC_TIMEOUT
- CRC_SSH_TIMEOUT
- CRC_CLUSTER_STABLE_DURATION

## Testing

To increase `CRC_OC_TIMEOUT` use following 
```
$ CRC_OC_TIMEOUT=40s crc start
```

### DIFF
--- a/pkg/os/util.go
+++ b/pkg/os/util.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/crc-org/crc/pkg/crc/logging"
 
@@ -118,4 +119,15 @@ func RunningInTerminal() bool {
 
 func RunningUsingSSH() bool {
 	return os.Getenv("SSH_TTY") != ""
+}
+
+func GetTimeOutEnvOrDefValue(env string, defValue time.Duration) time.Duration {
+	usrDefinedValue, err := time.ParseDuration(os.Getenv(env))
+	if err != nil {
+		return defValue
+	}
+	if usrDefinedValue != 0 {
+		return usrDefinedValue
+	}
+	return defValue
 }
--- a/pkg/crc/cluster/status.go
+++ b/pkg/crc/cluster/status.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/crc-org/crc/pkg/crc/logging"
 	"github.com/crc-org/crc/pkg/crc/network"
+	crcos "github.com/crc-org/crc/pkg/os"
 )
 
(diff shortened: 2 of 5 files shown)


cfergeau:
> Enable a way for user to provide different timeouts in case system is slow and crc takes more time than default.

Honestly, this is not very enticing to me. Either crc works out of the box, or it does not. If it does not, we need to fix it, or to inform nicely the user that their system is not good enough for crc. These env vars would be useful for testing/debugging purpose, but I'd rather not consider them as something some users should always set themselves.

pkg/os/util.go
```diff
@@ -119,3 +120,14 @@ func RunningInTerminal() bool {
 func RunningUsingSSH() bool {
 	return os.Getenv("SSH_TTY") != ""
 }
+
+func GetTimeOutEnvOrDefValue(env string, defValue time.Duration) time.Duration {
```
cfergeau, line 124: I'd name this just `GetTimeOutFromEnv`

cfergeau:
A test build with this patch would be useful in https://github.com/crc-org/crc/issues/3442 to see if it helps with the issue they are having.

praveenkumar:
> Honestly, this is not very enticing to me. Either crc works out of the box, or it does not. If it does not, we need to fix it, or to inform nicely the user that their system is not good enough for crc. These env vars would be useful for testing/debugging purpose, but I'd rather not consider them as something some users should always set themselves.

Yes it is not enticing or a very clean solution but if there is a dirty way to enable user to atleast start the crc I would prefer that instead completely blocking.

praveenkumar:
> A test build with this patch would be useful in #3442 to see if it helps with the issue they are having.

Added, let's wait for feedback.

cfergeau:
> > Honestly, this is not very enticing to me. Either crc works out of the box, or it does not. If it does not, we need to fix it, or to inform nicely the user that their system is not good enough for crc. These env vars would be useful for testing/debugging purpose, but I'd rather not consider them as something some users should always set themselves.
> 
> Yes it is not enticing or a very clean solution but if there is a dirty way to enable user to atleast start the crc I would prefer that instead completely blocking.

I prefer that we are fully sure what their issue is, and this PR is very useful for that, and then we decide about the best way to help them.

praveenkumar:
closing this since it didn't helped the user to run the crc on a slow hardware.

PR #3448, opened 2022-12-07
Misc CI fixes for github actions

praveenkumar:


.github/workflows/make-check.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       matrix:
         os:
           - macOS-11
+          - macOS-12
```
cfergeau: there's also `macOS-latest` if we don't care too much about the macOS version which is used.
macos-installer.yml also needs to be updated.

.github/workflows/make-check.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         os:
           - macOS-11
           - ubuntu-latest
-          - ubuntu-18.04
+          - ubuntu-20.04
```
cfergeau, line 18: or maybe drop it and only keep ubuntu-latest?

.github/workflows/make-check.yml
```diff
@@ -27,3 +27,8 @@ jobs:
           go-version: ${{ matrix.go }}
       - name: Build
         run: make check
+      - name: Upload artifacts
```
cfergeau: This might fit better in make-rpm.yml, for other platforms we ship installers, not raw binaries.

.github/workflows/make-check.yml
praveenkumar: @cfergeau `macOS-latest` still point to `macOS-11` :(

.github/workflows/make-check.yml
praveenkumar: ubuntu-latest point to 22.04 so plan is to have it latest 2 version

.github/workflows/make-check.yml
praveenkumar: I can try it to `make-rpm` action.

.github/workflows/make-check.yml
cfergeau: It's being moved to macOS-12
See this post from October 3, 2022 https://github.blog/changelog/2022-10-03-github-actions-jobs-running-on-macos-latest-are-now-running-on-macos-12/

> Over the next 10 weeks, jobs using the macos-latest runner label will migrate from macOS 11 to 12.

.github/workflows/make-check.yml
cfergeau: Part of the suggestion was to upload the RPM, not the binary :)

.github/workflows/make-check.yml
cfergeau: >  `macos-installer.yml` also needs to be updated.

This is still missing

.github/workflows/make-check.yml
praveenkumar: @cfergeau I prefer binary instead of rpm if we need to share someone who is using non-rpm base systems.

.github/workflows/make-check.yml
praveenkumar: updated

.github/workflows/make-check.yml
cfergeau: Let's keep it in `make check` then even if none of the places are great for generating/exporting a `crc` binary artifact.

praveenkumar: @cfergeau `make check` run on each runner not separate like installer so have to check if we can have a condition that only use linux binary for linux runner system.

praveenkumar:
@cfergeau do you have some more suggestion/review for this PR or should we take this in.

Makefile
```diff
@@ -130,6 +130,11 @@ spec: $(GENERATED_RPM_FILES)
 	
 test-rpmbuild: spec
 	${CONTAINER_RUNTIME} build -f images/rpmbuild/Containerfile .
+	${CONTAINER_RUNTIME} build -t test-rpmbuild-img -f images/rpmbuild/Containerfile .
+	${CONTAINER_RUNTIME} create --name test-rpmbuild test-rpmbuild-img
```
cfergeau, line 133: I would go with `${CONTAINER_RUNTIME} build -v $(pwd)/images/rpmbuild/rpms/:/root/rpmbuild/:Z`, then you have `/root/rpmbuild/RPMS` readily available without going through hoops, and you also have `images/rpmbuild/rpms/BUILD/crc-2.12.0-4.11.18/release/crc-linux-amd64.tar.xz` which removes the need for an additional build of the crc binary.

I'm not sure the 2 `${CONTAINER_RUNTIME} build` lines are intentional?

praveenkumar: @cfergeau `-v` options for build doesn't mount host to container so this not going work. 

> I'm not sure the 2 ${CONTAINER_RUNTIME} build lines are intentional?

ah, L#132 should be removed.

cfergeau: > @cfergeau -v options for build doesn't mount host to container so this not going work.

? I tested it before posting my comment and it worked as expected

cfergeau: podman VS docker difference maybe?

praveenkumar: @cfergeau docker doesn't have this capability to volume mount during build, with latest version of docker it is possible but the using buildkit, but then the command for build change for docker and podman so I think better to have it this way. I can put that as comment or in commit message it that would make sense about why we are doing this way.  https://stackoverflow.com/questions/26050899/how-to-mount-host-volumes-into-docker-containers-in-dockerfile-during-build

.github/workflows/make-rpm.yml
```diff
@@ -24,3 +24,12 @@ jobs:
           go-version: ${{ matrix.go }}
       - name: Build rpm
         run: CONTAINER_RUNTIME=docker make test-rpmbuild
+      - name: Create linux binary
+        run: make out/linux-amd64/crc
```
cfergeau: There is already a binary built as part of `make test-rpmbuild`, you could copy it from the container image at the same time as the rpms
https://github.com/crc-org/crc/pull/3448/commits/8fdbf0b1f6790d82d66aec33886220ba4fa7b393#r1053324914

.gitignore
```diff
@@ -6,3 +6,4 @@ release/
 packaging/tmp
 packaging/*.pkg
 tmp-embed/
+RPMS/
```
cfergeau, line 9: is `crc` needed as well?

cfergeau (approved):

.github/workflows/make-rpm.yml
```diff
@@ -24,3 +24,10 @@ jobs:
           go-version: ${{ matrix.go }}
       - name: Build rpm
         run: CONTAINER_RUNTIME=docker make test-rpmbuild
+      - name: Upload linux binary and rpms
+        uses: actions/upload-artifact@v3
+        with:
+          name: linux binary and rpm (${{ matrix.os }}
```
cfergeau: this is not exactly `matrix.os` as this was built inside a container image (forgot if the image is centos, ubi, ..)

Issue #3452, opened 2022-12-12
[BUG] crc setup fails with  yum: command not found


T0MASD:
### General information

  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / Hyper-V / hyperkit / vfkit
  * Did you run `crc setup` before starting it (Yes/No)?
  * Running CRC on: Laptop / Baremetal-Server / VM

## CRC version
```bash
$ crc version
CRC version: 2.11.0+823e40d
OpenShift version: 4.11.13
Podman version: 4.2.0
```
  

## Host Operating System
```bash
NAME="Fedora Linux"
VERSION="38 (Rawhide Prerelease)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (Rawhide Prerelease)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/rawhide/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=rawhide
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=rawhide
```

### Steps to reproduce

  1. crc setup
  
### Expected

crc detects dnf package manager on fedora and ues it

### Actual
```
INFO Using root access: Installing virtualization packages 
Could not install required packages:  exit status 127: /bin/sh: line 1: yum: command not found
```

praveenkumar:
@T0MASD does fedora going to drop yum => dnf script which as of now call dnf if user execute `yum` command?

anjannath:
Fixed via #3454

PR #3454, opened 2022-12-13
Replace yum with dnf for package installation

praveenkumar:
Since we now only support latest 2 releases of RHEL/Centos/Fedora and all distro now have dnf as default installed. We only needed yum for RHEL-7 which is now not supported as per our support matrics.

**Fixes:** Issue #3452

### DIFF
--- a/docs/source/topics/ref_required-software-packages.adoc
+++ b/docs/source/topics/ref_required-software-packages.adoc
@@ -8,7 +8,6 @@ Consult the following table to find the command used to install these packages f
 [options="header"]
 |====
 |Linux Distribution|Installation command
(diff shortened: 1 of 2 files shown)


anjannath (approved):

anjannath:
from your comment on the issue.. 

> does fedora going to drop yum => dnf script which as of now call dnf if user execute yum command?

if this is the case you can also mention this in the commit log that newer fedora 38 doesn't have the `yum` anymore, which was the case till fedora 37

gbraad (approved):

cfergeau (approved):

PR #3457, opened 2022-12-16
RFC: command-line only macOS pkg

cfergeau:
This reworks the macos package to only install our commandline tools (crc/vfkit/admin-helper).
Most of the changes are preparatory work/refactoring to have all the installer files in packaging/darwin.
Currently some of them are in packaging/ and others in packaging/darwin.

One thing I'm unsure of is whether we need to keep the tray installation a
while longer or not, in which case `macos: pkg: Remove electron tray from
installerr` will need to be adjusted to provide a way to build both installers.


packaging/darwin/macos-pkg-build-and-sign.sh
```diff
@@ -46,7 +46,7 @@ sign "${binDir}/crc"
 sign "${binDir}/crc-admin-helper-darwin"
 sign "${binDir}/vfkit"
 
-signAppBundle "${BASEDIR}/root/Applications/Red Hat OpenShift Local.app"
+signAppBundle "${binDir}"
```
anjannath: this can be removed along with the `signAppBundle` function, we don't have the tray `.app` anymore, only signing the `crc`, `vfkit` and `admin-helper` binaries is enough i think

anjannath:
Thanks for cleaning up the `packaging` dir 👍🏽 

LGTM, not sure when we want to remove the tray from the pkg, maybe split the rest of the commits into a separate PR?

cfergeau:
> Thanks for cleaning up the `packaging` dir 👍🏽
> 
> LGTM, not sure when we want to remove the tray from the pkg, maybe split the rest of the commits into a separate PR?

Yup, I can do that to avoid having to carry all the cleanup/refactoring patches for too long. One thing I'm not sure is whether the exd team rely on the `macos-pkg-build-and-sign.sh` script for their signing, I think not. We should check with them if the renaming/moving around of files will impact them.

praveenkumar:
https://github.com/crc-org/crc/pull/3457/commits/9f9acc30f902b42c2103ca399526a5d7c63a8361 which means if we take all those commits and perform a crc release then tray would not be part of it and as of now we still have to support tray as part of our installation.

praveenkumar:
> One thing I'm not sure is whether the exd team rely on the macos-pkg-build-and-sign.sh script for their signing, I think not. We should check with them if the renaming/moving around of files will impact them.

That's a good point, we should check with them before merge this PR.

/hold

anjannath:
> > Thanks for cleaning up the `packaging` dir 👍🏽
> > LGTM, not sure when we want to remove the tray from the pkg, maybe split the rest of the commits into a separate PR?
> 
> Yup, I can do that to avoid having to carry all the cleanup/refactoring patches for too long. One thing I'm not sure is whether the exd team rely on the `macos-pkg-build-and-sign.sh` script for their signing, I think not. We should check with them if the renaming/moving around of files will impact them.

they do use it since the signing is handled by that script, and we'll need to communicate this with them well before the next release so they can adjust their playbooks

update: informed the team about the coming changes

> 1. the tarball used to contain packaging dir but after the refactor extracted tarball will create a dir named `darwin`
> 2. the signing script used to be called packaging.sh which will be renamed to `macos-pkg-build-and-sign.sh`

cfergeau:
> update: informed the team about the coming changes
> 
> > 1. the tarball used to contain packaging dir but after the refactor extracted tarball will create a dir named `darwin`
> > 2. the signing script used to be called packaging.sh which will be renamed to `macos-pkg-build-and-sign.sh`

And their answer was:
> Hey, thanks for the heads up! 🙂 I'll prepare an internal PR that we will merge once your changes are live. 🙂

cfergeau:
> [9f9acc3](https://github.com/crc-org/crc/commit/9f9acc30f902b42c2103ca399526a5d7c63a8361) which means if we take all those commits and perform a crc release then tray would not be part of it and as of now we still have to support tray as part of our installation.

I've removed this commit for now as the other changes are good to have even without this final commit.

PR #3461, opened 2022-12-20
[integration]  timesync test

### LINKED ISSUE #3441: [Integration] Add test around timesync for podman bundle
Recently we added support for time sync for vfkit and we should have some integration tests around it to make sure we don't break it moving forward.  We still need to wait for new podman bundle to have those time sync bits in place.

adrianriobo:
This test ensures the timesync feature works as expected for darwin platform

### DIFF
--- a/test/integration/timesync_test.go
+++ b/test/integration/timesync_test.go
@@ -1,0 +1,53 @@
+package test_test
+
+import (
+	"os/exec"
+	"strconv"
+
+	// "github.com/crc-org/crc/test/extended/crc/cmd".
+	"github.com/crc-org/crc/test/extended/crc/cmd"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	sleepSeconds         = "30"
+	maxDateDiffInSeconds = 10
+)
+
+var _ = Describe("Guest machine time should be in sync with host", Label("guest-timesync", "darwin"), func() {
+
+	BeforeSuite(func() {
+		It("Ensuring instance is running")
+		if err := cmd.CheckCRCStatus("running"); err != nil {
+			// cluster is not running
+			startInstance()
+		}
+	})
+
+	It("checks instance date is in sync with host after host suspension", func() {
+		By("scheduling a suspensions and resume of the host")
+		Expect(exec.Command("pmset", "relative", sleepSeconds).Run()).To(Succeed())
+		Expect(exec.Command("pmset", "sleepnow").Run()).To(Succeed())
+		By("getting dates from instance and host")
+		out, err := exec.Command("date", "-u", "+%s").Output()
+		Expect(err).To(BeEmpty())
+		hostDate := string(out)
+		instanceDate, err := SendCommandToVM("date -u +%s")
+		Expect(err).To(BeEmpty())
+		By("ensuring date matches on host and instance")
+		hostDateAsInt, err := strconv.Atoi(hostDate)
+		Expect(err).To(BeEmpty())
+		instanceDateAsInt, err := strconv.Atoi(instanceDate)
+		Expect(err).To(BeEmpty())
+		dateDiff := instanceDateAsInt - hostDateAsInt
+		Ω(dateDiff).Should(BeNumerically("<=", maxDateDiffInSeconds))
+	})
+
+	AfterSuite(func() {
+		// delete instance cleanup...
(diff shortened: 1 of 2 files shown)


test/integration/timesync_test.go
```diff
@@ -0,0 +1,53 @@
+package test_test
+
+import (
+	"os/exec"
+	"strconv"
+
+	// "github.com/crc-org/crc/test/extended/crc/cmd".
```
jsliacan, line 7: I guess this line can be removed?

test/integration/util_test.go
```diff
@@ -0,0 +1,18 @@
+package test_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func startInstance() {
```
jsliacan, line 8: I'd maybe call this `startCRC()` (inside it we can distinguish which bundle we start with) as the other functions are called `CheckCRCStatus`, etc. Not a strong suggestion, just mentioning it.

test/integration/timesync_test.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+	})
+
+	It("checks instance date is in sync with host after host suspension", func() {
+		By("scheduling a suspensions and resume of the host")
+		Expect(exec.Command("pmset", "relative", sleepSeconds).Run()).To(Succeed())
+		Expect(exec.Command("pmset", "sleepnow").Run()).To(Succeed())
+		By("getting dates from instance and host")
+		out, err := exec.Command("date", "-u", "+%s").Output()
+		Expect(err).To(BeEmpty())
```
jsliacan, line 34: I think elsewhere I'm using `Expect(err).NotTo(HaveOccurred())`. Can stay as it is if you prefer it.

test/integration/timesync_test.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+const (
+	sleepSeconds         = "30"
+	maxDateDiffInSeconds = 10
+)
+
+var _ = Describe("Guest machine time should be in sync with host", Label("guest-timesync", "darwin"), func() {
+
+	BeforeSuite(func() {
```
jsliacan, line 20: The way I understand `BeforeSuite` is that it acts globally (before each `Describe` block?), so it should probably be defined outside of it too, e.g. in `testsuite_test.go`. Here, maybe `BeforeEach` would work better? 

That said, I am not sure if the `It` block wouldn't be better split into a few of them, in which case `BeforeEach` wouldn't work so well.

test/integration/timesync_test.go
```diff
@@ -0,0 +1,53 @@
...
+		It("Ensuring instance is running")
+		if err := cmd.CheckCRCStatus("running"); err != nil {
+			// cluster is not running
+			startInstance()
+		}
+	})
+
+	It("checks instance date is in sync with host after host suspension", func() {
```
jsliacan, line 28: I'm not sure what good practice is in Ginkgo, but I imagine `It` blocks to be smaller. One of the reasons for this is that the results file is iterated by `It` blocks. If only one block present, then only one pass/fail result. Which is probably good for e2e test (?) but not sure about integration... More detail in the results might be easier to debug too. But I'm not an expert here, just suggesting!

test/integration/util_test.go
jsliacan: BTW, there's a file called `utilities_test.go` where I put code like this: https://github.com/crc-org/crc/blob/e3054a158d9c6fad2e5641b7d5207a951287b576/test/integration/utilities_test.go. That said, recently we talked about moving this kind of code to a place which will be shared with e2e. I think I'm doing the same "start if not running" check here: https://github.com/crc-org/crc/blob/e3054a158d9c6fad2e5641b7d5207a951287b576/test/e2e/testsuite/testsuite.go#L583 Maybe if you put this in `extended` we can later refactor it to be used in multiple places. Again, up to you.

jsliacan (commented):
I haven't actually checked if this test is doing what it should be doing :).

adrianriobo:
I am keeping this PR on hold as I have been trying to ensure we can mimic the sleep mode on our qe envs, but I was not able for the moment. On `AWS` even scheduling with pmset the system seems to go for sleep but only for a second so tyme will not be out of sync for testing the feature

adrianriobo:
Close as I am not able to reproduce the scenario on remote machines

PR #3462, opened 2022-12-20
[e2e] added repetitions to ensure the state of the cluster

adrianriobo:
During e2e executions eventually we see executions when the cluster state is showing running and then is degraded right away, if the subsequent steps on an scneario depends on that state we need to ensure the stability of the cluster. This PR introduces a new param `repetitions` to define the number of times that the state of the cluster should keep stable for a period of time

### DIFF
--- a/test/extended/crc/cmd/cmd.go
+++ b/test/extended/crc/cmd/cmd.go
@@ -12,8 +12,10 @@ import (
 
 const (
 	// timeout to wait for cluster to change its state
-	clusterStateRetryCount    = 15
-	clusterStateTimeout       = 600
+	clusterStateRetryCount = 25
+	clusterStateTimeout    = 600
+	// defines the number of times the state should be matches in a row
+	clusterStateRepetition    = 3
 	CRCExecutableInstalled    = "installed"
 	CRCExecutableNotInstalled = "notInstalled"
 )
@@ -119,7 +121,7 @@ func UnsetConfigPropertySucceedsOrFails(property string, expected string) error 
 }
 
 func WaitForClusterInState(state string) error {
-	return util.MatchWithRetry(state, CheckCRCStatus,
+	return util.MatchRepetitionsWithRetry(state, CheckCRCStatus, clusterStateRepetition,
 		clusterStateRetryCount, clusterStateTimeout)
 }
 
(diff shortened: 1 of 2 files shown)


test/extended/util/util.go
```diff
@@ -144,41 +143,34 @@ func RemoveCRCHome(crcHome string) error {
...
-	iterationDuration, extraDuration, err :=
-		GetRetryParametersFromTimeoutInSeconds(retryCount, timeout)
-	if err != nil {
-		return err
-	}
-	for i := 0; i < retryCount; i++ {
-		err := match(expression)
-		if err == nil {
-			return nil
+// MatchWithRetry will execute match function with expression as arg
```
praveenkumar: it should be `MatchRepetitionsWithRetry` instead `MatchWitRetry` for this function.

adrianriobo: done

test/extended/crc/cmd/cmd.go
```diff
@@ -12,8 +12,10 @@ import (
 
 const (
 	// timeout to wait for cluster to change its state
-	clusterStateRetryCount    = 15
-	clusterStateTimeout       = 600
+	clusterStateRetryCount = 25
+	clusterStateTimeout    = 600
+	// defines the number of times the state should be matches in a row
```
gbraad, line 17: matched

gbraad (approved):

test/extended/util/util.go
```diff
@@ -144,41 +143,34 @@ func RemoveCRCHome(crcHome string) error {
...
+	matchRepetition := 0
+	for {
+		select {
+		case <-timeout:
+			tick.Stop()
+			return fmt.Errorf("not found: %s. Timeout", expression)
+		case <-tick.C:
+			if err := match(expression); err == nil {
+				matchRepetition++
+				if matchRepetition == matchRepetitions {
```
praveenkumar, line 166: @adrianriobo so with this logic we want to make sure atleast 3 time in a row the cluster state should be stable?

adrianriobo: yeah...basically the point is ..we want to ensure the cluster is healthy in less than 20 minutes...but healthy does not mean one hit on `state=running`. As I found cases where tests failed b/c cluster state was running and right away went to starting again...

so e2e step (checking the state of cluster was passed) and the next one...lets say `oc login` failed... 

So to prevent to situation...and based on the logic from `crc start` I wanna ensure the health of the cluster based on this 3 in a row running state checks

Issue #3458, opened 2022-12-16
[BUG] crc setup fails on macOS Ventura (Intel & M1)


What-Do-I-Know:
### General information

  * OS: macOS
  * Hypervisor: vfkit
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop

## CRC version
```CRC version: 2.12.0+ea98bb41
OpenShift version: 4.11.18
Podman version: 4.2.0
```
  
## CRC status
```DEBU CRC version: 2.12.0+ea98bb41
DEBU OpenShift version: 4.11.18
DEBU Podman version: 4.2.0
DEBU Running 'crc status'
DEBU Running 'sw_vers -productVersion'
DEBU sending request - Post "https://api.segment.io/v1/batch": dial tcp 0.0.0.0:443: connect: connection refused
DEBU 1 messages dropped because they failed to be sent and the client was closed
crc does not seem to be setup correctly, have you run 'crc setup'?
```

## CRC config
```- consent-telemetry                     : yes
```

## Host Operating System
```ProductName:		macOS
ProductVersion:		13.1
BuildVersion:		22C65
```

### Steps to reproduce

  1. Install CRC
  2. UI wizard fails
  3. Running crc setup manually does not succeed with the above mentioned error

### Expected
crc setup completes with success.

### Actual
crc setup fails with error: "Temporary error: daemon is not running yet (x8)"

### Logs

Before gather the logs try following if that fix your issue
```bash
$ crc delete -f
Machine does not exist. Use 'crc start' to create it
$ crc cleanup
INFO Unloading and removing the daemon plist file
INFO Removing /etc/resolver/testing file
INFO Stopping CRC vfkit process
INFO Removing hosts file records added by CRC
INFO Removing pull secret from the keyring
INFO Removing older logs
INFO Removing CRC Machine Instance directory
INFO Removing crc executable symlink
Cleanup finished
```
$ crc stetup --log-level debug
```
DEBU CRC version: 2.12.0+ea98bb41                 
DEBU OpenShift version: 4.11.18                   
DEBU Podman version: 4.2.0                        
DEBU Running 'crc setup'                          
INFO Using bundle path /Users/jp/.crc/cache/crc_vfkit_4.11.18_arm64.crcbundle 
INFO Checking if running as non-root              
INFO Checking if crc-admin-helper executable is cached 
INFO Checking for obsolete admin-helper executable 
DEBU Checking if an older admin-helper executable is installed 
[...]

PR #3464, opened 2022-12-21
[test] Issue #3458 Remove keyring accessible check from secret config store

anjannath:
this check prevents the daemon from running on macOS when the Login keychain is locked the user gets a pop-up to unlock the keychain, if that is not done on time the launchd daemon service is not started by `crc setup` command

### DIFF
--- a/pkg/crc/config/secret_config.go
+++ b/pkg/crc/config/secret_config.go
@@ -20,32 +20,25 @@ func (s Secret) String() string {
 }
 
 type SecretStorage struct {
-	secretService   string
-	storeAccessible bool
+	secretService string
 }
 
 func NewSecretStorage() *SecretStorage {
 	return &SecretStorage{
-		secretService:   secretServiceName,
-		storeAccessible: keyringAccessible(),
+		secretService: secretServiceName,
 	}
 }
 
 func (c *SecretStorage) Get(key string) interface{} {
-	if !c.storeAccessible {
-		return nil
-	}
 	secret, err := keyring.Get(c.secretService, key)
-	if errors.Is(err, keyring.ErrNotFound) {
+	if err != nil || errors.Is(err, keyring.ErrNotFound) {
+		logging.Debugf("error while getting config from secret store: %v", err)
 		return nil
 	}
 	return secret
 }
 
 func (c *SecretStorage) Set(key string, value interface{}) error {
-	if !c.storeAccessible {
-		return ErrSecretsNotAccessible
-	}
 	secret, err := cast.ToStringE(value)
 	if err != nil {
 		return fmt.Errorf("Failed to cast secret value to string: %w", err)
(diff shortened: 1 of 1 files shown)


pkg/crc/config/secret_config.go
```diff
@@ -56,7 +56,7 @@ func (c *SecretStorage) Unset(key string) error {
 }
 
 func NewEmptyInMemorySecretStorage() *SecretStorage {
-	keyring.MockInit()
+	// keyring.MockInit()
```
cfergeau, line 59: This looks like a WIP commit? There's a CI failure because dbus-launch is missing, I guess it's related to this change?

pkg/crc/config/secret_config.go
```diff
@@ -20,32 +20,25 @@ func (s Secret) String() string {
...
 	}
 }
 
 func (c *SecretStorage) Get(key string) interface{} {
-	if !c.storeAccessible {
-		return nil
-	}
 	secret, err := keyring.Get(c.secretService, key)
-	if errors.Is(err, keyring.ErrNotFound) {
+	if err != nil || errors.Is(err, keyring.ErrNotFound) {
```
cfergeau, line 34: The `errors.Is()` check seems to be dead code? It will only be reached when `err == nil`, and a `nil` error won't match `keyring.ErrNotFound`?

pkg/crc/config/secret_config.go
```diff
@@ -54,30 +47,17 @@ func (c *SecretStorage) Set(key string, value interface{}) error {
...
 	err := keyring.Delete(c.secretService, key)
-	if errors.Is(err, keyring.ErrNotFound) {
+	if err != nil || errors.Is(err, keyring.ErrNotFound) {
+		logging.Debugf("error while getting config from secret store: %v", err)
 		return nil
 	}
 	return err
 }
 
-func keyringAccessible() bool {
```
cfergeau, line 67: I'm not 100% clear what changes this removal causes? If the keyring is not accessible, instead of caching that it's not accessible, we will be calling the `keyring.*` functions, which will in turn tell us they cannot access teh keyring? So it's just a loss of efficiency, and we no longer have a specific ErrSecretsNotAccessible error?

pkg/crc/config/secret_config.go
anjannath: yeah, opened this PR to reproduce/demo the issue with the keychain not being accessible

pkg/crc/config/secret_config.go
anjannath: its an attempt to address https://github.com/crc-org/crc/issues/3458 , currently this `keyringAccessible()` check is causing `crc setup` to be stuck when the keychain is not accessible, by removing the check and caching of the result during initialization `crc setup` will be able to proceed and daemon will be started in those scenarios

(looking at it again, i think we should keep the check and the error and move it to body of `keychain.Set`, `keychain.Unset` etc. so its called when we want to access the keychain and if not accessible we get a proper error reported)

PR #3466, opened 2022-12-21
Add 'watch' argument to status command

evidolob:
**Fixes:** Issue #N

**Relates to:** #2876

## Solution/Idea

This will display CPU and RAM load in real time.
To achieve this PR add websocket implementation and daemon send data through websocket `/status` endpoint.

It is use https://github.com/nhooyr/websocket as websocket implementation
and upgrade https://github.com/cheggaaa/pb to `v3` to render load bars in terminal.

Demo:

https://user-images.githubusercontent.com/929743/208875792-e665602a-548d-4f71-bda9-fd435bfc70ac.mov

## Testing

1. run any preset
2. call `crc status -w`
3.  look on output, it should contains bar for RAM and bars for CPU's load


pkg/crc/api/websocket/status.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewStatusListener(machine machine.Client) ConnectionListener {
+	return &StatusConnectionListener{
+		machine: machine,
+		done:    make(chan bool),
+	}
+}
+
+func (s StatusConnectionListener) onClientConnected(count int, sendData SendData) {
+	if count == 1 {
```
anjannath: does it mean, only one ws `client` will be getting the status?

pkg/crc/api/websocket/handler.go
```diff
@@ -0,0 +1,38 @@
...
+type SendData func([]byte)
+
+type ConnectionListener interface {
+	onClientConnected(count int, sendData SendData)
+	onClientDisconnected(count int, sendData SendData)
+}
+
+func NewEndpointHandler(data SendData) *EndpointHandler {
+	handler := &EndpointHandler{
+		listeners: make(map[ConnectionListener]struct{}),
```
anjannath: maybe it would be better if we can avoid using these empty maps with only map keys but an empty `struct`?

cmd/crc/cmd/daemon.go
```diff
@@ -142,7 +143,9 @@ func run(configuration *types.Configuration) error {
 		}
 		mux := http.NewServeMux()
 		mux.Handle("/network/", http.StripPrefix("/network", vn.Mux()))
-		mux.Handle("/api/", http.StripPrefix("/api", api.NewMux(config, newMachine(), logging.Memory, segmentClient)))
+		machineClient := newMachine()
+		mux.Handle("/api/", http.StripPrefix("/api", api.NewMux(config, machineClient, logging.Memory, segmentClient)))
+		mux.Handle("/socket/", http.StripPrefix("/socket", websocket.NewWebsocketServer(machineClient)))
```
anjannath, line 148: wondering if this could've been part of the same `/api` path, something like `/api/statuswatch` ?

pkg/crc/api/websocket/status.go
evidolob: No, that mean that collection of CPU/RAM usage start only when first client connected, but all connected clients will receive info.

cmd/crc/cmd/daemon.go
evidolob: Hm, It may, let me try.

I prefer to separate HTTP API and WebSocket API, in URL. We may have only one websocket endpoint, and add some soer of PRC API or PubSub  API, and handle all events/query’s in single websocket connection. But this approach requires more complex client implementation.

anjannath: okay, don't change things yet if it complicates the client implementations! i was just asking to know if it was possible or there was some limitations to using the same HTTP server for websocket as well as normal HTTP

pkg/crc/api/websocket/status.go
anjannath: got it, and we don't want to start another go routine if one is already started and `count > 0` would start it for every client.

cmd/crc/cmd/status.go
```diff
@@ -48,11 +58,97 @@ type status struct {
...
+	status.RAMSize = -1
+	status.RAMUsage = -1
+	renderError := render(status, writer, outputFormat)
+	if renderError != nil {
+		return renderError
+	}
+
+	ctx := context.Background()
+
+	con, _, err := websocket.Dial(ctx, "ws://unix/socket/status", &websocket.DialOptions{HTTPClient: &http.Client{
```
cfergeau: Have you considered moving this + the json unmarshalling to `daemonclient`?  It might fit better there.
`con` is usually spelt `conn` if it's a `net.Conn`

cmd/crc/cmd/status.go
```diff
@@ -48,11 +58,97 @@ type status struct {
...
+
+	isPullInit := false
+	for {
+		_, msg, err := con.Read(ctx)
+		if err != nil {
+			return err
+		}
+
+		wmState := &types.WmStatusResult{}
+		err = json.Unmarshal(msg, wmState)
```
cfergeau: I think you could use `Read`  from https://pkg.go.dev/nhooyr.io/websocket/wsjson and skip the unmarshalling, but not sure it's going to bring us much.

cmd/crc/cmd/status.go
```diff
@@ -48,11 +58,97 @@ type status struct {
...
+		if !isPullInit {
+			bars = createBars(wmState.CPUUse, writer)
+			barPull = pb.NewPool(bars...)
+			isPullInit = true
+			err = barPull.Start()
+			if err != nil {
+				return err
+			}
+		}
+		// '0' bar is RAM
```
cfergeau: I'd use multiple go return values instead of stuffing everything in an array:
```
func createBars(cpuUse []int64, writer io.Writer) (*pb.ProgressBar, []*pb.ProgressBar) {
   [...]
    return memoryBar, result
}
[...]
    ramBar, cpuBars := createBars(...)
[...]
```
You can also name the return values in the function prototype if you want to add a bit of self-documentation.

cmd/crc/cmd/status.go
```diff
@@ -48,11 +58,97 @@ type status struct {
...
+
+	for i := 0; i < len(cpuUse); i++ {
+		bar := pb.New(101)
+		bar.SetWriter(writer)
+		bar.Set(pb.Static, true)
+		tmpl := fmt.Sprintf(`{{ green "CPU%d:" }} {{percent .}} {{string . "my_green_string" | green}} {{ bar . "[" "\u2588" "\u2588" " " "]"}}`, i)
+		bar.SetTemplateString(tmpl)
+		bar.SetMaxWidth(150)
+
+		result[i+1] = bar
```
cfergeau: I'd just use 
```
    result := []*pb.ProgressBar{} // or var result []*pb.ProgressBar if this is favoured by goland/linting tools
    [...]
    result = append(result, memoryBar)
    for i := 0; i < len(cpuUse); i++ {
        [...]
        result = append(result, bar)
    }
```
This way you don't need to precompute the length of `result` or to think about the index you are putting the element at.

cmd/crc/cmd/status.go
```diff
@@ -48,11 +58,97 @@ type status struct {
...
+
+}
+
+func createBars(cpuUse []int64, writer io.Writer) []*pb.ProgressBar {
+	result := make([]*pb.ProgressBar, len(cpuUse)+1)
+	memoryBar := pb.New(101)
+	memoryBar.SetWriter(writer)
+	memoryBar.Set(pb.Bytes, true)
+	memoryBar.Set(pb.Static, true)
+	tmpl := `{{ red "RAM:" }} {{counters . }} {{percent .}} {{string . "my_green_string" | green}} {{ bar . "[" "\u2588" "\u2588" " " "]"}} `
```
cfergeau, line 114: On my test machine, the closing `]` for the memory bar was not aligned with the other ones , it was one char before.

pkg/crc/cluster/cluster.go
```diff
@@ -104,6 +105,29 @@ func GetRAMUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 	return ramSize, ramUsage, nil
 }
 
+func GetCPUUsage(sshRunner *ssh.Runner) ([]int64, error) {
+	cmd := "top -n1 -1 -b -w 79  | grep %Cpu | awk '{print $3}'"
+	out, _, err := sshRunner.Run(cmd)
```
cfergeau, line 111: I was a bit worried that the output would be locale-dependant, but this seems ok, if I set `LC_ALL`/`LANG`, manually start the daemon, `sshRunner.Run("printenv")` does not inherit the locale changes, which is good (floats in French locale are "6,2", not "6.2")

pkg/crc/cluster/cluster.go
```diff
@@ -104,6 +105,29 @@ func GetRAMUsage(sshRunner *ssh.Runner) (int64, int64, error) {
...
+
+	cpuUsage := strings.Split(strings.TrimSpace(out), "\n")
+	result := make([]int64, len(cpuUsage))
+	for i, s := range cpuUsage {
+		c, err := strconv.ParseFloat(s, 64)
+		if err != nil {
+			return nil, err
+		}
+
+		result[i] = int64(math.Round(c))
```
cfergeau: similarly to one of my previous suggestions ,you could use `result = append(result, int64(...))` which removes the need for allocating the correct amount of memory and using the right index.

pkg/crc/cluster/cluster.go
```diff
@@ -104,6 +105,29 @@ func GetRAMUsage(sshRunner *ssh.Runner) (int64, int64, error) {
 	return ramSize, ramUsage, nil
 }
 
+func GetCPUUsage(sshRunner *ssh.Runner) ([]int64, error) {
```
cfergeau, line 109: Can you add a short comment about the content of this `[]int64`?

pkg/crc/machine/client.go
```diff
@@ -22,6 +22,7 @@ type Client interface {
 	PowerOff() error
 	Start(ctx context.Context, startConfig types.StartConfig) (*types.StartResult, error)
 	Status() (*types.ClusterStatusResult, error)
+	WmStatus() (*types.WmStatusResult, error)
```
cfergeau: Not sure what `Wm` stands for?

cfergeau (commented):
First round of comments, I haven't looked at `api/websocket` yet.

pkg/crc/api/websocket/endpoint.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"sync"
+)
+
+type Endpoint interface {
+	addClient(client *wsClient)
+	deleteClient(client *wsClient)
+	setHandler(handler *EndpointHandler)
+	sendData([]byte)
```
cfergeau: This probably be replaced this with `io.Writer` which provides almost the same method, `Write(p []byte) (n int, err error)` and will be familiar to go developers.

pkg/crc/api/websocket/endpoint.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+type endpoint struct {
+	clients   map[*wsClient]struct{}
+	clientsMu sync.Mutex
```
cfergeau: Can you name this `clientsMutex` or `clientLock`?

pkg/crc/api/websocket/endpoint.go
```diff
@@ -0,0 +1,59 @@
...
+		clients: make(map[*wsClient]struct{}),
+	}
+}
+
+func (e *endpoint) addClient(client *wsClient) {
+	e.clientsMu.Lock()
+	e.clients[client] = struct{}{}
+	e.clientsMu.Unlock()
+
+	e.handler.onClientConnected(len(e.clients))
```
cfergeau: `len(e.clients)` is unprotected by the mutex, my feeling is that this could lead to various (unlikely) races. Not sure something bad could happen apart from having 2 goroutines from `StatusConnectionListener.onClientConnected` running briefly at the same time before one of them is terminated.

pkg/crc/api/websocket/endpoint.go
```diff
@@ -0,0 +1,59 @@
...
+
+	e.handler.onClientDisconnected(len(e.clients))
+}
+
+// send data bytes to clients
+func (e *endpoint) sendData(data []byte) {
+	e.clientsMu.Lock()
+	defer e.clientsMu.Unlock()
+
+	for s := range e.clients {
```
cfergeau: `s`? Why not `client`? I get confused very quickly with one letter variable names :(

pkg/crc/api/websocket/endpoint.go
```diff
@@ -0,0 +1,59 @@
...
+		select {
+		case s.msgs <- data:
```
cfergeau: This forwards data from the goroutine created in StatusConnectionListener.onClientConnected to the goroutines Server.subscribe runs in. This runs in the StatusConnectionListener.onClientConnected goroutine.
(It took me some code reading to get to this understanding which I hope is correct, it would not hurt to describe this somewhere imo, either in code comments or in commit logs).

I'm wondering if the client could embed its `*websocket.Conn` and we'd call directly the `writeTimeout` method  (or equivalent) from here to avoid an indirection? There is no code making use of the connection in the other goroutine.

If you prefer to keep things this way, I'd at least add a comment to explicitly say we forward data to the s.msgs channel, as it's a bit hidden in the `case`, and it's easy to think execution will fallthrough to the `default` case.

pkg/crc/api/websocket/server.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func NewWebsocketServer(machine machine.Client) *Server {
+	ws := &Server{
+		clientMessageBuffer: 16,
+		endpoints:           map[string]Endpoint{},
+	}
+
+	ws.serveMux.HandleFunc("/status", func(writer http.ResponseWriter, request *http.Request) {
+		ws.endpoints["/status"] = createStatusEndpoint(machine)
```
cfergeau: Shouldn't this be done outside of `HandleFunc`? Won't this create a new status endpoint each time we get a request on `/socket/status`?

pkg/crc/api/websocket/server.go
```diff
@@ -0,0 +1,126 @@
...
+	defer cancel()
+
+	return c.Write(ctx, websocket.MessageText, msg)
+}
+
+func createStatusEndpoint(machine machine.Client) Endpoint {
+	statusEndpoint := NewEndpoint()
+	statusEndpointHandler := NewEndpointHandler(statusEndpoint.sendData)
+	statusEndpoint.setHandler(statusEndpointHandler)
+	statusEndpointHandler.addListener(NewStatusListener(machine))
```
cfergeau, line 126: A bit difficult for me to wrap my head around endpoint/handler/listener, it seems only 2 would be needed? Or is endpoint for sending data on the websocket, handler for client connection/disconnection, and listener to periodically get the status data?

pkg/crc/api/websocket/status.go
cfergeau: Might be clearer as `Start`/`Stop` or similar naming? I find it a bit artificial to have a generic method which gets passed the number of connected client which then is a noop unless there's only 1 connected client.
(and this "count" can be misleading, there is no lock taken while it's called, so it means "around the time onClientConnected() was called, there was 1 client connected, but by the time you get to `if count==1`, any number of clients may have connected/disconnected")

pkg/crc/api/websocket/status.go
```diff
@@ -0,0 +1,52 @@
...
+					ticker.Stop()
+					logging.Debug("stop fetching machine info")
+					return
+				case <-ticker.C:
+					status, err := s.machine.WmStatus()
+					if err != nil {
+						logging.Errorf("unexpected error during getting machine status: %v", err)
+					}
+
+					bytes, _ := json.Marshal(status)
```
cfergeau: error handling is missing

pkg/crc/api/websocket/server.go
```diff
@@ -0,0 +1,126 @@
...
+	logging.Debugf("handle %s request", endpointPath)
+	c, err := websocket.Accept(w, r, nil)
+	if err != nil {
+		logging.Infof("%v", err)
+		return
+	}
+
+	defer c.Close(websocket.StatusInternalError, "")
+
+	// TODO add reading messages form client
```
cfergeau: minor "form" typo

cmd/crc/cmd/status.go
evidolob: Just fix that:
Screenshot 2023-01-23 at 16 46 27

cmd/crc/cmd/status.go
```diff
@@ -106,47 +107,45 @@ func runWatchStatus(writer io.Writer, client *daemonclient.Client, cacheDir stri
...
+	ramBar.Set(pb.Static, true)
 	tmpl := `{{ red "RAM:" }} {{counters . }} {{percent .}} {{string . "my_green_string" | green}} {{ bar . "[" "\u2588" "\u2588" " " "]"}} `
-	memoryBar.SetMaxWidth(150)
-	memoryBar.SetTemplateString(tmpl)
-	result[0] = memoryBar
+	ramBar.SetMaxWidth(150)
+	ramBar.SetTemplateString(tmpl)
 
-	for i := 0; i < len(cpuUse); i++ {
+	for i, _ := range cpuUse {
```
cfergeau: `, _` is not needed

pkg/crc/api/client/websocket_client.go
```diff
@@ -0,0 +1,38 @@
...
+
+type StatusCallback func(*types.WmStatusResult, error)
+
+func NewWebSocketClient(httpClient *http.Client) *WebSocketClient {
+	return &WebSocketClient{
+		httpClient,
+	}
+}
+
+func (c *WebSocketClient) Status(callback StatusCallback) error {
```
cfergeau: `func (c *WebSocketClient) Status(callback StatusCallback) (types.WmStatusResult, error)`  without the `for {}` loop would be more consistent with the other 'client' APIs  and would avoid the need for the callback. The `for{}` loop could stay in the `cmd/` package.

pkg/crc/api/websocket/server.go
evidolob: >Or is endpoint for sending data on the websocket, handler for client connection/disconnection, and listener to periodically get the status data?

Yes, I try to create general websocket API, which will not bonded to `status` implementation.
In general we need to set up this once per endpoint(websocket url)

cfergeau:
I've done a few commits in https://github.com/cfergeau/crc/commits/status-watch on top of your branch instead of proper review comments /o\ `websocket: Use io.Writer interface` `Make StatusConnectionListener generic` are mere suggestions, feel free to take them or ignore them as you prefer.
`Should invalid RAMUse/RamSize be set to 0 or -1?` is a question regarding error values in the loadresult structure, maybe 0 is correct, it was not clear to me. And the added API doc may need some fixing, this is my understanding of the split, but maybe some of the naming I use is incorrect.
Last commit is `s/Transport/transport` and should be fairly self-explanatory.

Apart from this, this looks good to me!

evidolob:
@cfergeau I'm OK with your changes, could you add them to this PR?

evidolob:
@cfergeau I’ve cherry-picked your commits

cfergeau (approved):
I squashed a number of commits together and force pushed them.  Let's get this in!

PR #3475, opened 2023-01-11
Add flag `--components` to `crc-embedder` tool

anjannath:
this is needed for the chocolatey package generation (#3444) where we want to select only the `admin-helper` to download current crc-embedder downloads all the data files needed for a platform

### demo run

```
❯ ./crc-embedder download --components=vfkit-driver . --log-level debug
DEBU Downloading https://github.com/crc-org/vfkit/releases/download/v0.0.4/vfkit to .
DEBU Download saved to vfkit

❯ ./crc-embedder download --goos=windows --components=vfkit-driver,tray --log-level debug /tmp
WARN 'vfkit-driver' is not available, ensure goos flag is used. Supported component names are: vfkit-driver, vfkit-entitlement, libvirt-driver, admin-helper, tray
DEBU Downloading https://github.com/crc-org/tray-electron/releases/download/1.2.9/crc-tray-windows.zip to /tmp
67.28 MiB / 97.26 MiB [--------------------------------------------------------------------------------------->_______________________________________] 69.18% 3.87 MiB p/s

❯ ./crc-embedder download -h
Download data files to embed in the crc executable

Usage:
  crc-embedder download [destination directory] [flags]

Flags:
      --components strings   Needed component(s) (vfkit-driver, vfkit-entitlement, libvirt-driver, admin-helper, tray)
      --goos string          Target platform (darwin, linux or windows) (default "darwin")
  -h, --help                 help for download

Global Flags:

```

### DIFF
--- a/cmd/crc-embedder/cmd/download.go
+++ b/cmd/crc-embedder/cmd/download.go
@@ -1,13 +1,16 @@
 package cmd
 
 import (
+	"fmt"
 	"runtime"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
 
 func init() {
 	downloadCmd.Flags().StringVar(&goos, "goos", runtime.GOOS, "Target platform (darwin, linux or windows)")
+	downloadCmd.Flags().StringSliceVar(&components, "components", []string{}, fmt.Sprintf("List of component(s) to download (%s)", strings.Join(getAllComponentNames(goos), ", ")))
 	rootCmd.AddCommand(downloadCmd)
 }
 
@@ -21,6 +24,6 @@ var downloadCmd = &cobra.Command{
 }
 
 func runDownload(args []string) error {
-	_, err := downloadDataFiles(goos, args[0])
+	_, err := downloadDataFiles(goos, components, args[0])
 	return err
 }
--- a/cmd/crc-embedder/cmd/embed.go
+++ b/cmd/crc-embedder/cmd/embed.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/crc-org/crc/pkg/crc/constants"
 	"github.com/crc-org/crc/pkg/crc/logging"
@@ -21,13 +22,23 @@ import (
 var (
 	goos       string
 	cacheDir   string
+	components []string
 	noDownload bool
+)
+
+const (
+	vfkitDriver      = "vfkit-driver"
+	vfkitEntitlement = "vfkit-entitlement"
+	libvirtDriver    = "libvirt-driver"
+	adminHelper      = "admin-helper"
+	tray             = "tray"
 )
 
 func init() {
 	embedCmd.Flags().StringVar(&goos, "goos", runtime.GOOS, "Target platform (darwin, linux or windows)")
 	embedCmd.Flags().StringVar(&cacheDir, "cache-dir", "", "Destination directory for the downloaded files")
 	embedCmd.Flags().BoolVar(&noDownload, "no-download", false, "Only embed files, don't download")
+	embedCmd.Flags().StringSliceVar(&components, "components", []string{}, fmt.Sprintf("List of component(s) to download (%s)", strings.Join(getAllComponentNames(goos), ", ")))
 	rootCmd.AddCommand(embedCmd)
 }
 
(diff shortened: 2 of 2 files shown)


cmd/crc-embedder/cmd/embed.go
```diff
@@ -124,9 +143,25 @@ func getEmbedFileList(goos string, destDir string) []string {
...
+				}
+				downloadedFiles = append(downloadedFiles, filename)
+			} else {
+				logging.Warnf("'%s' is not available, ensure goos flag is used. Supported component names are: %s", component, strings.Join(componentNames, ", "))
+			}
+		}
+		return downloadedFiles, nil
+	}
+
 	for _, dl := range downloads {
```
praveenkumar: doesn't `downloads` now is the map instead of slice so you need to have `dl, ok` logic here.

cfergeau: `for i, dl := range slice` will return index in the array  + array element, `for key, dl := range map` will return map key and map value. `for _, dl := range map` and `for _dl := range slice` should have the same behaviour (but it also got me thinking for a bit)

cmd/crc-embedder/cmd/embed.go
```diff
@@ -21,13 +22,31 @@ import (
...
+)
+
+var componentNames = []string{
+	vfkitDriver,
+	vfkitEntitlement,
+	libvirtDriver,
+	adminHelper,
+	tray,
+}
+
```
cfergeau: Might be more extensible in the future to build the list of allowed components directly from the `dataFileUrls` structure? This would allow you to take into account `goos` when showing the list of allowed components
```
func allComponentNames() []string {
  var components []string
  for _, dls := range datafileUrls {
    for component, _ := range dls {
        components = append(components, component)
    }
  }
  return components
}
```

and something similar but limited to the current goos

cmd/crc-embedder/cmd/download.go
```diff
@@ -1,13 +1,18 @@
...
 
 import (
+	"fmt"
 	"runtime"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
 
+var helpMsgComponent = fmt.Sprintf("Needed component(s) (%s)", strings.Join(componentNames, ", "))
```
cfergeau: "needed" sounds odd to me here, maybe `List of component(s) to download` ?

cmd/crc-embedder/cmd/embed.go
```diff
@@ -124,9 +143,25 @@ func getEmbedFileList(goos string, destDir string) []string {
...
+				filename, err := download.Download(dl.url, destDir, dl.permissions, nil)
+				if err != nil {
+					return nil, err
+				}
+				downloadedFiles = append(downloadedFiles, filename)
+			} else {
+				logging.Warnf("'%s' is not available, ensure goos flag is used. Supported component names are: %s", component, strings.Join(componentNames, ", "))
+			}
+		}
+		return downloadedFiles, nil
```
cfergeau: I'd prefer if we had a single loop for both `len(components) == 0` and `len(components != 0`.
This could take the form of:
```
for component, dl := range downloads {
    if !shouldDownload(components, component) {
        continue
    }
    downloadedFiles = append(downloadedFiles, filename)
}
```
If you want to keep some error reporting for bad component names, you can either check this before running downloadDataFiles, or you can remove the downloaded components from the `components` map, and check if it's empty or not at the end of the loop.

cfergeau (commented):
Can you mention https://github.com/crc-org/crc/pull/3444#discussion_r1066970300 ? It was not immediatly obvious why we need this.

cmd/crc-embedder/cmd/embed.go
anjannath: i thought about this then preferred two loops to avoid an `if` check for the component name on every iteration, will update!

cmd/crc-embedder/cmd/embed.go
```diff
@@ -124,16 +135,44 @@ func getEmbedFileList(goos string, destDir string) []string {
...
+			continue
+		}
 		filename, err := download.Download(dl.url, destDir, dl.permissions, nil)
 		if err != nil {
 			return nil, err
 		}
 		downloadedFiles = append(downloadedFiles, filename)
 	}
 
+	if len(downloadedFiles) == 0 {
```
cfergeau: This could even be stricter, such as `if len(components) != 0 && len(components) != len(downloadedFiles)`

cfergeau (approved):

praveenkumar:
CI has `lint` error.

PR #3476, opened 2023-01-11
Make paths to vfkit/crc-admin-helper configurable

cfergeau:
This PR will be needed for [homebrew](https://brew.sh/) builds. In this scenario, `crc-admin-helper` and `vfkit` will be installed in 2 different locations.
The crc codebase is not prepared for this, it assumes all binaries are in the `constants.BinDir()` directory.
This PR does 2 main things:
- it adds separate `vfkit.Executable()`, `constants.AdminHelperPath()` and `libvirt.MachineDriverPath` functions so that it's explicit which binary we want the path of. This allows to remove most of `BinDir()` usage
- it uses `go generate` to allow to provide a json file at compile time which will specify where the various binaries are located. This will be used by the brew formulas.


praveenkumar:
I might have a very native question but why do we want to have admin-helper and vfkit be in a different directory? Can't we use `make out/macos-${arch}/crc` as part of brew formula so that it will just have the `crc` binary and other dependencies are downloaded as part of `crc setup` command?

After looking https://github.com/cfergeau/homebrew-crc/blob/main/crc.rb looks like we are executing `make macos-release-binary` and that's the reason we are doing all those refactor and changes. If I am not wrong then `admin-helper` and `vfkit` binary which are available on github are signed ( if not then can't we now have the key to do it) then we can directly download those binary as part of `setup` instead building as part of brew formula. The only difference between the `make macos-release-binary` and `make out/macos-${arch}/crc` is `$(REPOPATH)/pkg/crc/version.installerBuild=true` one and I am not sure why this is important for brew build because this logic is only used for installer where the files are placed by installer and we needed logic to get it.

cfergeau:
> Can't we use `make out/macos-${arch}/crc` as part of brew formula so that it will just have the `crc` binary and other dependencies are downloaded as part of `crc setup` command?

It makes more sense to me to have all the binaries that we need to be built using brew rather than downloading them on demand.

> I am not sure why this is important for brew build because this logic is only used for installer where the files are placed by installer and we needed logic to get it.

Imo brew is closer to an 'install' build than a build from git. We can ensure crc, crc-admin-helper and vfkit are all present at install time rather than relying on `crc setup` to do it for us.

praveenkumar:
> > Can't we use `make out/macos-${arch}/crc` as part of brew formula so that it will just have the `crc` binary and other dependencies are downloaded as part of `crc setup` command?
> 
> It makes more sense to me to have all the binaries that we need to be built using brew rather than downloading them on demand.

If we are building it on the brew then we have to maintain also so we are end up maintaining 3 formula instead of just 1 which is just for crc.

> 
> > I am not sure why this is important for brew build because this logic is only used for installer where the files are placed by installer and we needed logic to get it.
> 
> Imo brew is closer to an 'install' build than a build from git. We can ensure crc, crc-admin-helper and vfkit are all present at install time rather than relying on `crc setup` to do it for us.

`crc setup` is created for that kind of usecase where it make sure all the prerequisites are satisfied. In case of installer we did it different because we were shipping everything as part of tarball before and then with pkg mostly if someone want distribute internally. In case of brew we don't need to worry about it and just pull the required stuff from github (single source for setup and also for installer).

gbraad:
> It makes more sense to me to have all the binaries that we need to be built using brew rather than downloading them on demand.

Brew does not prefer to use downloaded binaries. if that would be the case, we might have to deliver the admin-helper as a dependent cask instead.

> Imo brew is closer to an 'install' build than a build from git. 

Not is not comletely true. For an install a cask should be considered.

praveenkumar:
> > It makes more sense to me to have all the binaries that we need to be built using brew rather than downloading them on demand.
> 
> Brew does not prefer to use downloaded binaries. if that would be the case, we might have to deliver the admin-helper as a dependent cask instead.

@gbraad I am not saying to use something like `curl` as part of brew formula to download the crc binary. I am just saying just build `crc` binary as part of brew formula and once user have `crc` install then they suppose to follow `crc setup` and `crc start` . During `crc setup` admin-helper/router would be downloaded without any issue.

cfergeau:
> > > It makes more sense to me to have all the binaries that we need to be built using brew rather than downloading them on demand.
> > 
> > 
> > Brew does not prefer to use downloaded binaries. if that would be the case, we might have to deliver the admin-helper as a dependent cask instead.
> 
> @gbraad I am not saying to use something like `curl` as part of brew formula to download the crc binary. I am just saying just build `crc` binary as part of brew formula and once user have `crc` install then they suppose to follow `crc setup` and `crc start` . During `crc setup` admin-helper/router would be downloaded without any issue.

This still means we are providing required binaries which were not built as a brew formula. Fwiw, the chocolatey package also ships its own admin-helper binary, it does not rely on crc setup to download/install it.

praveenkumar:
> the chocolatey package also ships its own admin-helper binary, it does not rely on crc setup to download/install it.

@cfergeau I think it embed/download admin-helper from our git release but not build from source, it is just adding the service file for admin helper so that it doesn't prompt user for UAC  , @anjannath can confirm. PR which just merge about downloading specific component as part of embed binary help to do it.

cfergeau:
> > the chocolatey package also ships its own admin-helper binary, it does not rely on crc setup to download/install it.
> 
> @cfergeau I think it embed/download admin-helper from our git release but not build from source, it is just adding the service file for admin helper so that it doesn't prompt user for UAC , @anjannath can confirm. PR which just merge about downloading specific component as part of embed binary help to do it.

It doesn't rebuild it (I guess it could if this was some kind of requirement), but when you download the chocolatey package, you get all the needed binaries. No binary is downloaded through `crc setup`.

Here it's similar, all the needed binaries are downloaded/installed after `brew install crc` completes. No binary is downloaded through `crc setup`. As a sidenote, `brew` will also provide pre-built binaries, they won't be rebuilt every time someone runs `brew install`.

praveenkumar:
>   It doesn't rebuild it (I guess it could if this was some kind of requirement), but when you download the chocolatey package, you get all the needed binaries. No binary is downloaded through crc setup.

We are doing it for the windows for a complete different reason because in the windows as part of `crc setup` there is no way to get admin-helper binary and add as part of service as of now, we tried in the past and failed because it need user to prompt multiple UAC prompt for each time a command run as admin mode. These limitation is not on the mac side so I think we shouldn't compare the flow here.

cfergeau:
CI fails with:
```
 Unknown file format when trying to uncompress /home/packer/crc_libvirt_4.11.18_amd64.crcbundle
```
I don't know if there's an issue with the bundle or if this PR is causing the breakage.

cfergeau:
Regardless of what Windows is doing or not, I still favor building as much as possible using brew instead of having crc download them. Are there big downsides to building everything on brew?

Using `make macos-release-binary` is also needed to use the 'production' segment key, which is the one we want to use for brew binaries as well.

praveenkumar:
> Regardless of what Windows is doing or not, I still favor building as much as possible using brew instead of having crc download them. Are there big downsides to building everything on brew?

The only downside I see is maintainability 
1. Added one more way to install the dependencies which mean brew formula need to be maintained
2. Added codebase to have those logic in place

Since we already have signed bits for those dependencies so I am in favor to consume those directly for all the installation method. 

> 
> Using `make macos-release-binary` is also needed to use the 'production' segment key, which is the one we want to use for brew binaries as well.

For this we can just create another target `make macos-brew-binary` and use the `production` segment key (hopefully that should be trivial).

cfergeau:
> > Regardless of what Windows is doing or not, I still favor building as much as possible using brew instead of having crc download them. Are there big downsides to building everything on brew?
> 
> The only downside I see is maintainability
> 
>     1. Added one more way to install the dependencies which mean brew formula need to be maintained

Ideally, someone from the brew community would take the maintainability over. The update of the brew formula can be automated.
 
>     2. Added codebase to have those logic in place

Regardless of brew, most commits in this PR are in my opinion good to have, no more magic `BinDir()` but more fine-grained `XxxPath()` helpers.


> Since we already have signed bits for those dependencies so I am in favor to consume those directly for all the installation method.

We also have signed crc binaries, but we build it in brew, we don't reuse the existing one. I don't think it's a matter of signed/not signed, but of 'ecosystem consistency'. If we use brew, there is an expectation that everything will be coming from brew.

anjannath:
> > the chocolatey package also ships its own admin-helper binary, it does not rely on crc setup to download/install it.
> 
> @cfergeau I think it embed/download admin-helper from our git release but not build from source, it is just adding the service file for admin helper so that it doesn't prompt user for UAC , @anjannath can confirm. PR which just merge about downloading specific component as part of embed binary help to do it.

yep `admin-helper` binary is included in the chocolatey package that is downloaded from releases in the admin-helper repo, choco doesn't force us to build it from source, and `crc setup` used to contain the `fix` for windows as well which would download it, but was removed when we moved to the msi since `admin-helper` was contained within it and became redundant.

the choco package uses the contained binary to create a service during `choco install crc`, this is the part if done from `crc setup` that'd create UAC prompt.

pkg/crc/machine/libvirt/constants.go
```diff
@@ -21,14 +20,14 @@ const (
...
 
 const (
-	MachineDriverCommand = "crc-driver-libvirt"
+	machineDriverCommand = "crc-driver-libvirt"
 	MachineDriverVersion = "0.13.5"
 )
 
 var (
-	MachineDriverDownloadURL = fmt.Sprintf("https://github.com/crc-org/machine-driver-libvirt/releases/download/%s/%s", MachineDriverVersion, MachineDriverCommand)
+	MachineDriverDownloadURL = fmt.Sprintf("https://github.com/crc-org/machine-driver-libvirt/releases/download/%s/%s", MachineDriverVersion, machineDriverCommand)
```
praveenkumar, line 28: 👍🏼

praveenkumar (approved):

PR #3485, opened 2023-01-20
tools: add script to publish github releases

anjannath:
this adds a bash script named 'gh-release.sh' which can be used to publish the github releases of CRC

it needs to be ran from the commit which is tagged with the current desired release we want to publish

(need to also add a github actions job to run this)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,17 @@
+name: Publish release on github
+on:
+  workflow_dispatch:
+jobs:
+  publish_release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Prepare empty notable changes
+        run: touch notable_changes.txt
+      - name: Run gh-release.sh
+        run: NONINTERACTIVE=1 ./gh-release.sh
+        env:
+           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


gh-release.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
+set -xeuo pipefail
+
+lastReleasedVersion="$(gh release view -R crc-org/crc --json tagName --jq .tagName)"
```
praveenkumar, line 5: So this script add `gh` as dependency?

.github/workflows/release.yml
```diff
@@ -0,0 +1,17 @@
+name: Publish release on github
+on:
+  workflow_dispatch:
```
praveenkumar, line 3: Does this workflow need to be run manually?

praveenkumar (approved):

gh-release.sh
anjannath: yeah, alternatively we could directly talk to the github api using curl or something but with `gh` its much easier to do

.github/workflows/release.yml
anjannath: yes, once we tag the release after that we have to run this manually

praveenkumar:
@anjannath Can you also update the README about how to run this manually and then we will merge it.

praveenkumar:
I am going to create follow up issue for update the README around how to use it.

.github/workflows/release.yml
cfergeau: can't this be triggered to run automatically when a tag is pushed?

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

PR #3486, opened 2023-01-23
bump openshift version to 4.12.0 in Makefile

anjannath:
this bumps CRC version to 2.13.1 and adds the missing version bump of OpenShift which was missed in 2242f7fa


cfergeau (approved):
ACK for this change.

For a longer term solution, we could actually use `go generate` to fetch https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/bundles/openshift/4.12.0/sha256sum.txt and generate https://github.com/anjannath/crc/blob/8a5421cad4eb6bc3c2bf1f9bc6250490c245335f/pkg/crc/machine/bundle/constants.go from it. This way we'd only update `CRC _VERSION` in the Makefile, and run `make update-bundle` or such, and everything would be up to date. We could have something in `make check` verifying that `CRC_VERSION` and `constants.go` are in sync.

Or we can remove the checksums altogether when the bundles are signed on mirror.openshift.com (which hopefully will happen soon)

anjannath:
Merging as the certificate rotation test failure is inconsistent

PR #3500, opened 2023-02-06
Powershell Core Dectection

stjeffrey:
This will allow Powershell Core to be detected on Windows as Powershell instead of falling back to default cmd which causes errors when running commands like crc oc-env.

### DIFF
--- a/pkg/os/shell/shell_windows.go
+++ b/pkg/os/shell/shell_windows.go
@@ -64,6 +64,8 @@ func detect() (string, error) {
 		switch {
 		case strings.Contains(strings.ToLower(shell), "powershell"):
 			return "powershell", nil
+		case strings.Contains(strings.ToLower(shell), "pwsh"):
+			return "powershell", nil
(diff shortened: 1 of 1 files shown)


gbraad:
This addresses an old discussion we (@adrianriobo and I) had https://github.com/crc-org/crc/discussions/2353

@stjeffrey does everything work as expected? That was my only concern at that time.

gbraad (approved):

stjeffrey:
> This addresses an old discussion we (@adrianriobo and I) had #2353
> 
> @stjeffrey does everything work as expected? That was my only concern at that time.

Looks like the PR was already merged but I just wanted to reply to your question.
I must admint I was mostly concerned about using oc and podman commands with this patch and did not consider if it would work for things like crc setup.  Given the existing codebase though with the fallback being CMD, in the context of Powershell Core I can't see how it could possibly make it any worse.

anjannath (approved):

PR #3506, opened 2023-02-09
cut v2.14.0

anjannath:


anjannath:
openshift CI failed again on the cert rotation scenario

PR #3511, opened 2023-02-13
resize disk also during creation of the VM when using hyper-v driver

anjannath:
in case of hyper-v driver the config value of disk-size was not applied to the VM disk image during creation of the instance

this adds a new private helper `resizeDisk` and calls it during creation as well as vm config Update during start

## Testing

1. Ensure a clean env and no non-default config values are set
2. `crc config set disk-size 80`
3. `crc start --log-level debug -p ~/pull-secret`
4. verify that the disk size of the VM has changed, either by using `crc status` or manually by ssh-ing into the VM

### DIFF
--- a/pkg/drivers/hyperv/hyperv_windows.go
+++ b/pkg/drivers/hyperv/hyperv_windows.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"strconv"
+	"strings"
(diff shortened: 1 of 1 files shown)


cfergeau (approved):
vfkit and libvirt are both calling a `resize` method from their `Create()` implementation, I don't know why this was missing from the hyperv driver.

PR #3519, opened 2023-02-15
[e2e] Add test: OpenShfit story behind proxy

jsliacan:
**Depends on:** PR #3417 

We start proxy in Before hook of each scenario and remove it in the After hook of each scenario. This way all openshift stories should directly work behind proxy. 

## Testing

Run e2e tests (this is part of them). Or run only this feature with `@linux && @story_openshift && @proxy` tag.

### DIFF
--- a/test/e2e/features/story_openshift.feature
+++ b/test/e2e/features/story_openshift.feature
@@ -1,4 +1,4 @@
-@story_openshift
+@story_openshift @darwin @linux @windows
 Feature: 4 Openshift stories
 
 	Background:
@@ -7,7 +7,7 @@ Feature: 4 Openshift stories
 
 	# End-to-end health check
 
-	@darwin @linux @windows @startstop @testdata @story_health
+	@startstop @testdata @story_health
 	Scenario: Overall cluster health
 		Then executing "oc create namespace testproj" succeeds
 		And executing "oc project testproj" succeeds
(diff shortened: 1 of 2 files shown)


test/e2e/testsuite/testsuite.go
```diff
@@ -238,7 +238,6 @@ func InitializeScenario(s *godog.ScenarioContext) {
 					fmt.Println(err)
 					os.Exit(1)
 				}
-
```
praveenkumar: @jsliacan I don't think just removing this line is doing what you described in the commit log.

jsliacan: Oh great, I must've mingled the commits when rearranging after last changes. Thanks for noticing.

cfergeau:
This is almost a copy&paste from story_openshift.feature. It would be useful to have an indication of this in the commit log/in the file itself (as a reminder both have to be kept in sync). And some explanation why this is duplicated would not hurt (I guess reuse is hard with what we use for testing?)

adrianriobo:
Indeed @cfergeau I think we should avoid the duplication, already discussed with @jsliacan, so we probably switch a bit the approach:

* We keep the simple proxy tests, so without knowing nothing about options on how to run tests, the default flow will cover the proxy setup.

* We will add a parameter on how to run the suite, so basically it will tell if we want to run everything with proxy, which basically means there will be an underneath logic on adding the proxy setup to the cluster on each spin of the cluster (this will allow us to run an in deep test of a cluster with a proxy setup)

jsliacan:
Also, I asked on cucumber/godog about Examples (Outline) for the `Background` directive. We'll see what they say, maybe there's a way to I'm not seeing that they'll suggest and we can keep the setup transparent to the user and not duplicate things. I don't think they'll actually implement the feature. 

https://github.com/cucumber/godog/issues/567

Additionally, if you have suggestions about how to re-word my issue over there, I'm all ears. Thanks!

gbraad:
how does this relate to #3684?

jsliacan:
I'm closing this for now as the changes in here aren't substantial to keep and we anyways agreed on a different approach than duplicating a feature file.

PR #3521, opened 2023-02-16
Remove lease resource from openshift-machine-config-operator namespace

praveenkumar:
With 4.12.3 we found out that MCO is block till
machine-config-controller and machine-config lease is not acquired
and that takes around 5 mins. In this patch we are removing all the
lease (as of now these 2 are present) from this namespace so that MCO
is not block for that long time.

We still don't know the root cause so a Jira issue is created against
MCO to track it: https://issues.redhat.com/browse/OCPBUGS-7583

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -483,6 +483,13 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
 	if err := WaitForOpenshiftResource(ctx, ocConfig, "cm"); err != nil {
 		return err
 	}
-	_, _, err := ocConfig.RunOcCommand("delete", "-n", "openshift-machine-config-operator", "cm", "machine-config-controller")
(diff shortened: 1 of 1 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -483,7 +483,7 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
 	if err := WaitForOpenshiftResource(ctx, ocConfig, "cm"); err != nil {
```
cfergeau, line 483: There's another `cm` occurrence here.

cfergeau (commented):
Regarding the log of the first commit, the issue is not limited to 4.12.3. It also happened with 4.12.1 even though a different bug was hiding it.

cfergeau:
I've cherry-picked these commits with the additional `cm` -> `configmap` change in https://github.com/crc-org/crc/pull/3522, we can close this.

anjannath:
closing as changes were merged in https://github.com/crc-org/crc/pull/3522

PR #3523, opened 2023-02-18
Update preflight_windows.go

hurzelpurzel:
// simple additional domain check: a local User got the computername = domain

**Fixes:** Issue #3282

## Solution/Idea

Compare USERDOMAIN against COMPUTERNAME to detect a local usage

## Proposed changes

Add a second layer of test by checking content of USERDOMAIN and do not trust the win32 DomainJoin test only.

## Testing

See my comment in Issue #3282. I' m not totally sure, but i think it fails and in the error case it assumes it is a local user.
This might be a problem with permissions.

### DIFF
--- a/pkg/crc/preflight/preflight_windows.go
+++ b/pkg/crc/preflight/preflight_windows.go
@@ -155,7 +155,7 @@ var userPartOfCrcUsersAndHypervAdminsGroupCheck = Check{
 var errReboot = errors.New("Please reboot your system and run 'crc setup' to complete the setup process")
 
 func username() string {
-	if ok, _ := win32.DomainJoined(); ok {
+	if ok := win32.DomainJoined(); ok {
 		return fmt.Sprintf(`%s\%s`, os.Getenv("USERDOMAIN"), os.Getenv("USERNAME"))
 	}
 	return os.Getenv("USERNAME")
--- a/pkg/os/windows/win32/domains_windows.go
+++ b/pkg/os/windows/win32/domains_windows.go
@@ -18,17 +18,28 @@ package win32
 
 import (
 	"errors"
+	"os"
 
 	"github.com/yusufpapurcu/wmi"
 )
 
-//nolint
+// nolint
 type Win32_ComputerSystem struct {
 	Partofdomain bool
 }
 
 // DomainJoined attempts to determine whether the machine is actively domain joined.
-func DomainJoined() (bool, error) {
+
+func DomainJoined() bool {
+	domainJoined, err := domainJoinedWmi()
+	if err != nil {
+		// simple additional domain check: a local User got the computername = domain
+		return os.Getenv("USERDOMAIN") != os.Getenv("COMPUTERNAME")
+	}
+	return domainJoined
+}
+
+func domainJoinedWmi() (bool, error) {
 	var c []Win32_ComputerSystem
 	q := wmi.CreateQuery(&c, "")
 


anjannath:
from https://github.com/crc-org/crc/issues/3282#issuecomment-1435589774
> I got a similar issue and I found out that crc using the USERNAME env variable to resolve the Group membership.
> In my case I got a domain user but USERNAME only contains the USER part and the domain was in a USERDOMAIN env variable. After setting the env variable USERNAME=%USERDOMAIN%%USERNAME% before starting crc setup and crc start, this solved my problem: crc found the hyper-v group membership and all went well.

You mentioned that you got a domain user but CRC was only using the `USERNAME` to find the group membership instead of `USERDOMAIN \ USERNAME`? so it seems the check `win32.DomainJoined()`  is not working as expected then..

cfergeau:
We have no logging of errors from DomainJoined(), so hard to tell what happened. We could add the fallback there though:
```
diff --git a/pkg/crc/preflight/preflight_windows.go b/pkg/crc/preflight/preflight_windows.go
index 14b369f8b..887127bf4 100644
--- a/pkg/crc/preflight/preflight_windows.go
+++ b/pkg/crc/preflight/preflight_windows.go
@@ -155,7 +155,7 @@ var userPartOfCrcUsersAndHypervAdminsGroupCheck = Check{
 var errReboot = errors.New("Please reboot your system and run 'crc setup' to complete the setup process")
 
 func username() string {
-	if ok, _ := win32.DomainJoined(); ok {
+	if win32.DomainJoined() {
 		return fmt.Sprintf(`%s\%s`, os.Getenv("USERDOMAIN"), os.Getenv("USERNAME"))
 	}
 	return os.Getenv("USERNAME")
diff --git a/pkg/os/windows/win32/domains_windows.go b/pkg/os/windows/win32/domains_windows.go
index 9c9fb1cb5..b09d0e8a7 100644
--- a/pkg/os/windows/win32/domains_windows.go
+++ b/pkg/os/windows/win32/domains_windows.go
@@ -18,6 +18,7 @@ package win32
 
 import (
 	"errors"
+	"os"
 
 	"github.com/yusufpapurcu/wmi"
 )
@@ -28,7 +29,7 @@ type Win32_ComputerSystem struct {
 }
 
 // DomainJoined attempts to determine whether the machine is actively domain joined.
-func DomainJoined() (bool, error) {
+func domainJoinedWmi() (bool, error) {
 	var c []Win32_ComputerSystem
 	q := wmi.CreateQuery(&c, "")
 
@@ -40,3 +41,12 @@ func DomainJoined() (bool, error) {
 	}
 	return c[0].Partofdomain, nil
 }
+
+// DomainJoined attempts to determine whether the machine is actively domain joined.
+func DomainJoined() bool {
+	domainJoined, err := domainJoinedWmi()
+	if err != nil {
+		return os.Getenv("USERDOMAIN") != os.Getenv("COMPUTERNAME")
+	}
+	return domainJoined
+}
```

anjannath:
@hurzelpurzel could you please update the PR to make the addition in the `win32.DomainJoined()` function itself, as suggested in https://github.com/crc-org/crc/pull/3523#issuecomment-1436792984

anjannath:
@LPottmeier thanks for updating the PR, could you please squash the two commits and update the commit message:

```
update check for user in hyper-v admins group

this adds additional check to figure out if the user is a domained joined
by looking at the value of env variables COMPUTERNAME and DOMAINJOINED
since the values are equal for a local user
```

anjannath:
@hurzelpurzel linter is complaining about:
```
pkg/os/windows/win32/domains_windows.go:21: File is not `gofmt`-ed with `-s` (gofmt)
	"os"
```

please run `make fmt` to fix it and re-push!

anjannath (approved):

anjannath:
Thanks for your contribution @hurzelpurzel ! ❤️

PR #3535, opened 2023-03-08
e2e: Check that we are testing the correct version

jsliacan:
Make use of `crc version` output to cross check that we are indeed testing the expected CRC version. Because version is so unique, it is enough to check for containment. If we start releasing `*-ec` and similar variations, this might need to be changed to matching instead of containing.  

In the case of this CI, this PR has no effect -- if `VERSION_TO_TEST` is set to `--crc-version=""` (default), then version is **not** checked. I manually checked the first Scenario in `basic.feature` which is affected if version is passed:

```
Log successfully started, logging into: /home/jsliacan/Github/crc/test/e2e/out/test-results/e2e_2023-3-8_15-21-12.log
Running e2e test in: /home/jsliacan/Github/crc/test/e2e/out/test-run
Working directory set to: /home/jsliacan/Github/crc/test/e2e/out/test-run
Using existing bundle: /home/jsliacan/Downloads/crc_libvirt_4.12.5_amd64.crcbundle
Feature: Basic test
  User explores some of the top-level CRC commands while going
  through the lifecycle of CRC.

  Scenario: CRC version                        # features/test.feature:8
    When executing crc version command         # testsuite.go:868 -> github.com/crc-org/crc/test/e2e/testsuite.ExecuteCRCCommand
    Then stdout should contain correct version # testsuite.go:600 -> github.com/crc-org/crc/test/e2e/testsuite.CommandREturnShouldContainCorrectVersion
Deleted CRC instance (if one existed).

1 scenarios (1 passed)
2 steps (2 passed)
58.890846ms
ok  	github.com/crc-org/crc/test/e2e	0.085s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ OKD_VERSION ?= 4.12.0-0.okd-2023-02-18-033438
 MICROSHIFT_VERSION ?= 4.12.9
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 2.17.0
-COMMIT_SHA?=$(shell git rev-parse --short HEAD)
+COMMIT_SHA?=$(shell git rev-parse --short=6 HEAD)
 MACOS_INSTALL_PATH = /usr/local/crc
 CONTAINER_RUNTIME ?= podman
 
--- a/Makefile
+++ b/Makefile
@@ -233,8 +233,11 @@ endif
 ifndef CRC_BINARY
 	CRC_BINARY = --crc-binary=$(GOPATH)/bin
 endif
+ifndef VERSION_TO_TEST
+	VERSION_TO_TEST = --crc-version=$(CRC_VERSION)+$(COMMIT_SHA)
+endif
 e2e:
-	@go test --timeout=180m $(REPOPATH)/test/e2e -tags "$(BUILDTAGS)" --ldflags="$(VERSION_VARIABLES)" -v $(PULL_SECRET_FILE) $(BUNDLE_LOCATION) $(CRC_BINARY) $(GODOG_OPTS) $(CLEANUP_HOME) $(INSTALLER_PATH) $(USER_PASSWORD)
+	@go test --timeout=180m $(REPOPATH)/test/e2e -tags "$(BUILDTAGS)" --ldflags="$(VERSION_VARIABLES)" -v $(PULL_SECRET_FILE) $(BUNDLE_LOCATION) $(CRC_BINARY) $(GODOG_OPTS) $(CLEANUP_HOME) $(VERSION_TO_TEST) $(INSTALLER_PATH) $(USER_PASSWORD)
 
 .PHONY: e2e-stories e2e-story-health e2e-story-marketplace e2e-story-registry
 # cluster must already be running, crc must be in the path
--- a/test/e2e/features/basic.feature
+++ b/test/e2e/features/basic.feature
@@ -7,7 +7,7 @@ Feature: Basic test
     @darwin @linux @windows
     Scenario: CRC version
         When executing crc version command
-        Then stdout should contain "version:"
+        Then stdout should contain correct version
 
     @darwin @linux @windows
     Scenario: CRC help
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -29,6 +29,7 @@ var (
 	pullSecretFile     string
 	cleanupHome        bool
 	testWithShell      string
+	CRCVersion         string
 
 	GodogFormat              string
 	GodogTags                string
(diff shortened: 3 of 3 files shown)


praveenkumar:
>Make use of crc version output to cross check that we are indeed testing the expected CRC version. Because version is so unique, it is enough to check for containment. If we start releasing *-ec and similar variations, this might need to be changed to matching instead of containing.

@jsliacan I don't see how it will used to check for containment . Current `crc version` output bunch of things like openshift/podman version. and if we want to check a released binary vs nightly bits then you first need to know the version of nightly bit and figure out. I think what we should do as part of makefile `v$(CRC_VERSION)-$(COMMIT_SHA)` as default `version` like we are doing for `CRC_E2E_IMG_VERSION` ?

Makefile
```diff
@@ -230,8 +230,11 @@ endif
 ifndef CRC_BINARY
 	CRC_BINARY = --crc-binary=$(GOPATH)/bin
 endif
+ifndef VERSION_TO_TEST
+	VERSION_TO_TEST = --crc-version=$(CRC_VERSION)
```
praveenkumar: should be `VERSION_TO_TEST = --crc-version=$(CRC_VERSION)-$(COMMIT_SHA)` so we are sure that same commit hash is part of version of the binary we are testing.

jsliacan: Added the `COMMIT_SHA` to `--crc-version`. So default behavior now includes commit check. 

Originally, it was only intended to tell apart a binary from the past release on the host from the current one, and similar "big" issues. That's why the check was only rough.

jsliacan:
I retested it, because even though the last commit is `a5e3f27`, I was seeing this failure:
```
Then stdout should contain correct version # testsuite.go:600 -> github.com/crc-org/crc/test/e2e/testsuite.CommandREturnShouldContainCorrectVersion
output did not match. Expected: '2.15.0+8729c83b', Actual: 'CRC version: 2.15.0+8729c83
OpenShift version: 4.12.5
Podman version: 4.3.1' 
```
Not only do I see [a5e3f27](https://github.com/crc-org/crc/pull/3535/commits/a5e3f271a71c0bd693ad884981f3262ad4ebc3b4) as the last commit, but I can't even find `8729c83b` among commits. And I most definitely don't know why there's `b` at the end of it compared to the output of `crc version`: `8729c83`.

When I run it locally, it passes fine...

praveenkumar:
@jsliacan CI is still failing with  which means as part of `crc version` returning 6char `68a59fa`  of git commit hash and our CI version generation is 7char `68a59fac`.
```
 time="2023-03-14T10:41:36Z" level=info msg="diagnose took 24.019925ms"
    Then stdout should contain correct version # testsuite.go:600 -> github.com/crc-org/crc/test/e2e/testsuite.CommandReturnShouldContainCorrectVersion
    output did not match. Expected: '2.15.0+68a59fac', Actual: 'CRC version: 2.15.0+68a59fa
OpenShift version: 4.12.5
Podman version: 4.3.1' 
```

jsliacan:
@praveenkumar yes, the error complains that expected: `26fb4fed` and actual: `26fb4fe` don't match. However, the last commit on my branch is: `2381736`. So I have to investigate a bit further.

Makefile
```diff
@@ -8,7 +8,7 @@ OKD_VERSION ?= 4.12.0-0.okd-2023-02-18-033438
 MICROSHIFT_VERSION ?= 4.12.5
 BUNDLE_EXTENSION = crcbundle
 CRC_VERSION = 2.15.0
-COMMIT_SHA?=$(shell git rev-parse --short HEAD)
+COMMIT_SHA?=$(shell git rev-parse --short=6 HEAD)
```
jsliacan, line 11: @praveenkumar , I couldn't figure out why the commit hash lengths differed, so I forced it here to be length 6 everywhere. If that looks OK to you, I've cleaned up the rest of the PR and it can go in. The CI passes now with this change.

praveenkumar: As per man page of `git-rev-parse` looks like this can be vary for different host, so I think the binary created in different container and then test execution happend in different container.
```
--short[=length]
           Same as --verify but shortens the object name to a unique prefix with at least length characters. The minimum length is 4, the default is the effective value of the core.abbrev configuration variable (see git-config(1)).
```

praveenkumar: Also in the commit log mention why this change is required for future reference.

jsliacan: > this can be vary for different host

Yes, that's why I made it fixed lenght. But there may be a better solution, e.g. adjust git settings on the hosts directly and leave this at default short? I don't know... 

Anyways, I rebased it as it was and included reasoning in the commit log. Should be OK now.

Issue #3518, opened 2023-02-15
[QE] Add custom disk size on basic scneario for e2e


adrianriobo:
Previously `e2e` tested disk resize after the cluster was started with default values. Recently a [bug](https://github.com/crc-org/crc/issues/3496) was discovered and solved, as the disk resize did not take effect with hyper-v on first cluster start.

To cover this test case and avoid overtime running `e2e` we decide to setup a custom disk size for the initial start of the cluster in one of the scenarios.

praveenkumar:
closing this issue since it is already fixed.

Issue #3496, opened 2023-02-03
[BUG] CRC not respecing disk-size config for hyper-v


gshipley:
### General information

  * OS: Windows 11
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Desktop / 64gb ram (dedicated to vm, 128gb total), 10 cored dedicated

## CRC version
```bash
C:\Users\gshipley>crc version
CRC version: 2.13.1+b5b864fd
OpenShift version: 4.12.0
Podman version: 4.3.1
```
  
## CRC status
This is after manually editing the vm setting in hyperv manager, otherwise disk usage will always show 32gb
```bash
CRC VM:          Running
OpenShift:       Starting (v4.12.0)
RAM Usage:       6.381GB of 66.98GB
Disk Usage:      19.77GB of 549.2GB (Inside the CRC VM)
Cache Usage:     35.07GB
Cache Directory: C:\Users\gshipley\.crc\cache
```

## CRC config
```bash

C:\Users\gshipley>crc config view
- consent-telemetry                     : yes
- cpus                                  : 10
- disk-size                             : 512
- enable-cluster-monitoring             : true
- enable-experimental-features          : true
- kubeadmin-password                    : xxxxxxx
- memory                                : 65000
- nameserver                            : 8.8.8.8
```

## Host Operating System
```bash
Host Name:                 GAMEBOX
OS Name:                   Microsoft Windows 11 Pro
OS Version:                10.0.22621 N/A Build 22621
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Standalone Workstation
OS Build Type:             Multiprocessor Free
Registered Owner:          gshipley
Registered Organization:
Product ID:                00330-80000-00000-AA423
Original Install Date:     11/11/2022, 11:23:33 AM
System Boot Time:          2/2/2023, 6:52:50 PM
System Manufacturer:       ASUS
System Model:              System Product Name
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 165 Stepping 5 GenuineIntel ~3696 Mhz
BIOS Version:              American Megatrends Inc. 0607, 5/29/2020
Windows Directory:         C:\WINDOWS
System Directory:          C:\WINDOWS\system32
Boot Device:               \Device\HarddiskVolume1
System Locale:             en-us;English (United States)
Input Locale:              en-us;English (United States)
[...]

PR #3536, opened 2023-03-08
Integration: Add checks for the initial start of the VM with custom values (memory, cpus, disk)

jsliacan:
Re: Issue #3496 and solution in #3511. Also, serves the same purpose as the check in `basic.feature` in e2e as asked for in #3518. 

Replacing default values in initial start by custom values and corresponding checks. The test already exists as integration test `resize_test.go`, so I just tweaked that one. I used this occasion to do the same for `memory` and `cpus` parameters.

### DIFF
--- a/test/integration/resize_test.go
+++ b/test/integration/resize_test.go
@@ -22,29 +22,29 @@ var _ = Describe("vary VM parameters: memory cpus, disk", Label("openshift-prese
 		It("start CRC", func() {
 			// default values: "--memory", "9216", "--cpus", "4", "disk-size", "31"
 			if bundlePath == "" {
-				Expect(RunCRCExpectSuccess("start", "-p", pullSecretPath)).To(ContainSubstring("Started the OpenShift cluster"))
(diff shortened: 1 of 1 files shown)


test/integration/resize_test.go
```diff
@@ -22,29 +22,29 @@ var _ = Describe("vary VM parameters: memory cpus, disk", Label("openshift-prese
...
 
 		It("check VM's memory size", func() {
 			out, err := SendCommandToVM("cat /proc/meminfo")
 			Expect(err).NotTo(HaveOccurred())
-			Expect(out).Should(MatchRegexp(`MemTotal:[\s]*9\d{6}`))
+			Expect(out).Should(MatchRegexp(`MemTotal:[\s]*11\d{6}`))
 		})
 
 		It("check VM's number of cpus", func() {
 			out, err := SendCommandToVM("cat /proc/cpuinfo")
```
praveenkumar, line 38: not a blocker but just a suggestion , why are we not using `lscpu` and then use `CPU\(s\):[\s]*5` which provide us correct cpus instead one less because `/proc/cpuinfo` start it from `0`? May be another PR to try and fix it?

praveenkumar (approved):

Issue #3456, opened 2022-12-15
[BUG] crc-admin context not added 


adrianriobo:
When the cluster is started it tries to add the `crc-admin` context to the kubeconfig file, under some conditions (specially on nested environments) the context is not added, this create some false positives on tests for cluster functionality.

I think the context add has no retry in case of failure. It would be great to check it and add a retry mechanism in there to ensure the context is successfully added.

gbraad:
> I think the context add has no retry in case of failure.

This might be true...

adrianriobo:
> I think then context add has no retry in case of failure.

Based on last conversation with @praveenkumar during QE sync, sorry did not check the code so can not ensure this is the root cause :(

gbraad:
> the root cause :(

It is unlikely the root cause. a retry-mechanism is merely a workaround for another problem; unable to write context. is this due to the the file permissions? what else?

adrianriobo:
yeah as I did not analyze the code I added the description based on conversations during qe sync, may the root cause for all the problem is not the addition of the context but a sync issue? , meaning right after the cluster has been started we try to use the context, on nested virtualization this may require a delay?

and yeah the workaround solution would not be the root cause of course :)

cfergeau:
What platform are you seeing this on? I assume linux since you mention nested virt? I assume you don't have logs when this happens?

gbraad:
> meaning right after the cluster has been started we try to use the context, on nested virtualization this may require a delay?

Does the test progress and use the context already? the startup hasn't finished, so it should not just continue, right? Hmmmm...

Perhaps a snip of the actual test might help.

adrianriobo:
> What platform are you seeing this on?

Yeah mostly on linux, but I think I saw it on other platforms as well, can not ensure...but this thing has been around for a really long time. Most of "false" positives on `openshift-ci` are due to this, as an example:

[#3444 failing](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/crc-org_crc/3444/pull-ci-crc-org-crc-main-e2e-crc/1602178874215501824)

and from the [logs of this run](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/crc-org_crc/3444/pull-ci-crc-org-crc-main-e2e-crc/1602178874215501824/artifacts/e2e-crc/gather-crc/artifacts/crc.log) 

```
[...]

PR #3537, opened 2023-03-08
Integration: Add testcase for login using crc-admin context

jsliacan:
#3456 and [this comment](https://github.com/crc-org/crc/issues/3456#issuecomment-1354482871) specifically.

Adding a testcase into existing test to login to the cluster using `crc-admin` context. No additional `crc start`. The issue above is still open so I assume this will fail for now, as it should until it's resolved. However, the way I added it here does not affect the rest of the tests - so the testcase will fail on its own.

### DIFF
--- a/test/integration/resize_test.go
+++ b/test/integration/resize_test.go
@@ -1,6 +1,7 @@
 package test_test
 
 import (
+	"os/exec"
 	"runtime"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


test/integration/resize_test.go
```diff
@@ -28,6 +29,21 @@ var _ = Describe("vary VM parameters: memory cpus, disk", Label("openshift-prese
 			}
 		})
 
+		It("login to cluster using crc-admin context", func() {
+			cmd := exec.Command("oc", "login", "use-context", "crc-admin")
```
praveenkumar: without the eval does it able to detect the `oc` binary? As per `exec.Command` doc 
```
Command uses LookPath to resolve name to a complete path if possible. Otherwise it uses name directly as Path.
```
which means if `oc` is not present in the os.Path then this might even break which might be happening as part of CI failure.

jsliacan: Sorry, that was me not checking after myself! Another mistakes was using `oc login` instead of `oc config`. Should be corrected now. I checked locally after removing `oc` from PATH.

jsliacan:
This should be good to go now, if you have time to glance over it @praveenkumar - thanks!

PR #3547, opened 2023-03-16
Fixes #3546 #3550 Add option to allow progressbars to be forced during download and uncompression of the bundle (setup phase)

### LINKED ISSUE #3546: Always show uncompress progress for use with the podman-desktop Extension
When we download we will show the download progress, but for uncompress we suppress this behavior. For the PD extension we need this infornation in the UI and it is the easiest way for now to expose it.

https://github.com/crc-org/crc/blob/1425c16d6ca9ae043f4ddcb817f7317443aa70d2/pkg/extract/extract.go#L25-L31

We originally suppressed this due to the use in non-terminal environments, although this is now less relevant.

### LINKED ISSUE #3550: [Enhancement] Allow progressbars to be hidden when not using a terminal
This is a follow-up to #3547 to allow the progressbars to be hidden when no terminal is used. This is for the case when a redirection to a file is used.

gbraad:
Fixes #3546 #3550

As part of `crc setup` we show the progress of the download and
uncompression. However, the download progress is always shown, while
the uncompression is suppressed when this is not a standard terninal.
For consistency we remove the suppression as this adds unnecessary
complexity.

### DIFF
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -11,19 +11,22 @@ import (
 	"github.com/crc-org/crc/pkg/crc/input"
 	"github.com/crc-org/crc/pkg/crc/preflight"
 	"github.com/crc-org/crc/pkg/crc/validation"
+	crcTerminal "github.com/crc-org/crc/pkg/os/terminal"
 
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/exec"
 )
 
 var (
-	checkOnly bool
+	checkOnly             bool
+	forceShowProgressbars bool
 )
 
 func init() {
 	setupCmd.Flags().Bool(crcConfig.ExperimentalFeatures, false, "Allow the use of experimental features")
 	setupCmd.Flags().StringP(crcConfig.Bundle, "b", constants.GetDefaultBundlePath(crcConfig.GetPreset(config)), "Bundle to use for instance")
 	setupCmd.Flags().BoolVar(&checkOnly, "check-only", false, "Only run the preflight checks, don't try to fix any misconfiguration")
+	setupCmd.Flags().BoolVar(&forceShowProgressbars, "show-progressbars", false, "Always show the progress bars for download and extraction")
 	addOutputFormatFlag(setupCmd)
 	rootCmd.AddCommand(setupCmd)
 }
@@ -63,6 +66,8 @@ func runSetup(arguments []string) error {
 		}
 	}
 
+	// set global variable to force terminal output
+	crcTerminal.ForceShowOutput = forceShowProgressbars
 	err := preflight.SetupHost(config, checkOnly)
 	if err != nil && checkOnly {
 		err = exec.CodeExitError{
--- a/pkg/extract/extract.go
+++ b/pkg/extract/extract.go
@@ -13,28 +13,29 @@ import (
 	"github.com/cheggaaa/pb/v3"
 	"github.com/crc-org/crc/pkg/crc/logging"
 	crcos "github.com/crc-org/crc/pkg/os"
+	"github.com/crc-org/crc/pkg/os/terminal"
+
 	"github.com/h2non/filetype"
 	"github.com/klauspost/compress/zstd"
 	"github.com/pkg/errors"
 	"github.com/xi2/xz"
-	terminal "golang.org/x/term"
 )
 
 const minSizeForProgressBar = 100_000_000
 
-func UncompressWithFilter(tarball, targetDir string, showProgress bool, fileFilter func(string) bool) ([]string, error) {
-	return uncompress(tarball, targetDir, fileFilter, showProgress && terminal.IsTerminal(int(os.Stdout.Fd())))
+func UncompressWithFilter(tarball, targetDir string, fileFilter func(string) bool) ([]string, error) {
+	return uncompress(tarball, targetDir, fileFilter, false) // never show detailed output
 }
 
 func Uncompress(tarball, targetDir string, showProgress bool) ([]string, error) {
-	return uncompress(tarball, targetDir, nil, showProgress && terminal.IsTerminal(int(os.Stdout.Fd())))
+	return uncompress(tarball, targetDir, nil, terminal.IsShowTerminalOutput())
 }
 
 func uncompress(tarball, targetDir string, fileFilter func(string) bool, showProgress bool) ([]string, error) {
 	logging.Debugf("Uncompressing %s to %s", tarball, targetDir)
 
 	if strings.HasSuffix(tarball, ".zip") {
-		return unzip(tarball, targetDir, fileFilter, showProgress)
+		return unzip(tarball, targetDir, fileFilter, terminal.IsShowTerminalOutput())
 	}
 
 	file, err := os.Open(filepath.Clean(tarball))
--- a/pkg/os/terminal/terminal.go
+++ b/pkg/os/terminal/terminal.go
@@ -1,0 +1,21 @@
+package terminal
+
+import (
+	"os"
+
+	xterminal "golang.org/x/term"
+)
+
+var (
+	// Global variable to force output regardless if terninal
+	ForceShowOutput = false
+)
+
+func IsShowTerminalOutput() bool {
+	// if this is a terminal or set to force output
+	return IsRunningInTerminal() || ForceShowOutput
+}
+
+func IsRunningInTerminal() bool {
+	return xterminal.IsTerminal(int(os.Stdout.Fd()))
+}
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -120,7 +120,7 @@ func (c *Cache) cacheExecutable() error {
 	// Check the file is tarball or not
 	if isTarball(assetTmpFile) {
 		// Extract the tarball and put it the cache directory.
-		extractedFiles, err = extract.UncompressWithFilter(assetTmpFile, tmpDir, false,
+		extractedFiles, err = extract.UncompressWithFilter(assetTmpFile, tmpDir,
 			func(filename string) bool { return filepath.Base(filename) == c.GetExecutableName() })
 		if err != nil {
 			return errors.Wrapf(err, "Cannot uncompress '%s'", assetTmpFile)
--- a/pkg/crc/cluster/pullsecret.go
+++ b/pkg/crc/cluster/pullsecret.go
@@ -15,7 +15,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/logging"
 	"github.com/crc-org/crc/pkg/crc/preset"
 	"github.com/crc-org/crc/pkg/crc/validation"
-	crcos "github.com/crc-org/crc/pkg/os"
+	crcTerminal "github.com/crc-org/crc/pkg/os/terminal"
 
 	"github.com/AlecAivazis/survey/v2"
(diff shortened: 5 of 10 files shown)


cfergeau:
Imo, this makes things worse for :
```
% crc setup &>foo.log

% cat foo.log 
level=info msg="Using bundle path /Users/teuf/.crc/cache/crc_vfkit_4.12.5_arm64.crcbundle"
level=info msg="Checking if running as non-root"
level=info msg="Checking if crc-admin-helper executable is cached"
level=info msg="Checking if running on a supported CPU architecture"
level=info msg="Checking minimum RAM requirements"
level=info msg="Checking if crc executable symlink exists"
level=info msg="Checking if running emulated on a M1 CPU"
level=info msg="Checking if vfkit is installed"
level=info msg="Checking if CRC bundle is extracted in '$HOME/.crc'"
level=info msg="Checking if /Users/teuf/.crc/cache/crc_vfkit_4.12.5_arm64.crcbundle exists"
level=info msg="Getting bundle for the CRC executable"
level=info msg="Downloading bundle: /Users/teuf/.crc/cache/crc_vfkit_4.12.5_arm64.crcbundle..."
0 B / 3.15 GiB [___________________________________________________] 0.00% ? p/s0 B / 3.15 GiB [___________________________________________________] 0.00% ? p/s339.45 MiB / 3.15 GiB [---->______________________________________] 10.51% ? p/s339.45 MiB / 3.15 GiB [--->______________________________] 10.51% 565.86 MiB p/s339.45 MiB / 3.15 GiB [--->______________________________] 10.51% 565.86 MiB p/s347.05 MiB / 3.15 GiB [--->______________________________] 10.75% 565.86 MiB p/s347.05 MiB / 3.15 GiB [--->______________________________] 10.75% 530.17 MiB p/s354.38 MiB / 3.15 GiB [--->______________________________] 10.97% 530.17 MiB p/s354.38 MiB / 3.15 GiB [--->______________________________] 10.97% 530.17 MiB p/s362.77 MiB / 3.15 GiB [--->______________________________] 11.23% 497.66 MiB p/s362.77 MiB / 3.15 GiB [--->______________________________] 11.23% 497.66 MiB p/s
```

praveenkumar:
> Imo, this makes things worse for :

@cfergeau which is the current behavior right even without this PR? Also isn't it expected even with our current terminal check since `crc` is going to detect it as terminal and show the progress bar but here we are just sending it to a log file?

cfergeau:
> @cfergeau which is the current behavior right even without this PR? 

Even if it's the current behaviour, it does not make it correct, especially when other parts of the code don't have the same behaviour.

`IsTerminal` will detect when there's a redirection to a file, with the quick hack below, the log file looks like this during the download:
```
level=info msg="Checking if CRC bundle is extracted in '$HOME/.crc'"
level=info msg="Checking if /home/teuf/.crc/cache/crc_libvirt_4.12.5_amd64.crcbundle exists"
level=info msg="Getting bundle for the CRC executable"
level=info msg="Downloading bundle: /home/teuf/.crc/cache/crc_libvirt_4.12.5_amd64.crcbundle..."
```

```
diff --git a/pkg/download/download.go b/pkg/download/download.go
index 532b5f4da..1a0ecdfef 100644
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -15,6 +15,7 @@ import (
        "github.com/cavaliergopher/grab/v3"
        "github.com/cheggaaa/pb/v3"
        "github.com/pkg/errors"
+       terminal "golang.org/x/term"
 )
 
 func doRequest(client *grab.Client, req *grab.Request) (string, error) {
@@ -28,15 +29,20 @@ func doRequest(client *grab.Client, req *grab.Request) (string, error) {
 
        t := time.NewTicker(500 * time.Millisecond)
        defer t.Stop()
-       bar := pb.Start64(resp.Size())
-       bar.Set(pb.Bytes, true)
-       defer bar.Finish()
+       var bar *pb.ProgressBar
+       if terminal.IsTerminal(int(os.Stdout.Fd())) {
+               bar = pb.Start64(resp.Size())
+               bar.Set(pb.Bytes, true)
+               defer bar.Finish()
+       }
 
 loop:
        for {
                select {
                case <-t.C:
-                       bar.SetCurrent(resp.BytesComplete())
+                       if terminal.IsTerminal(int(os.Stdout.Fd())) {
+                               bar.SetCurrent(resp.BytesComplete())
+                       }
                case <-resp.Done:
                        break loop
                }
```

praveenkumar:
> IsTerminal will detect when there's a redirection to a file, with the quick hack below, the log file looks like this during the download

yes but what we want to do is to show the progress bar in every case for PD extension usecase or you are saying we should use this and add a bool flag `--show-progressbar` at top level and then use it with `setup` and `start` ?

cfergeau:
> yes but what we want to do is to show the progress bar in every case for PD extension usecase or you are saying we should use this and add a bool flag --show-progressbar at top level and then use it with setup and start ?

I'd prefer this yes, by default, we have `--show-progressbar=auto`, and whether to show a progress bar or not will depend on `IsTerminal(stdout)`. The podman-desktop extension can use `--show-progressbar=yes` to get it shown even if it's not a terminal.

gbraad:
I think the argument you give is a little moot... This is not behaviour we normally suggest to use; file redirection for the setup flow.

> Imo, this makes things worse for :

Besides, you are judging about DOWNLOADING that has been untouched in this PR, so this is existing behaviour.

> Even if it's the current behaviour, it does not make it correct

However, this makes every comment justified and PRs linger for unnecessary long times due to enlarged scope.

---

I suggest a follow-up issue to 'fix' this for downloading, as this is not part of this issue/PR : #3550 The suggested fixes will not be part of the current PR.

praveenkumar:
Current PR also need to be fixed to run unit tests which is failing because of https://github.com/crc-org/crc/pull/3547/files#diff-2f2c46f41491d3331c52245381a6bf0a66bf5b8d708bb3018157764e7b744e50R20 one.

gbraad:
I am looking at adding the requested complexity to handle 'intelligent' showing of the progress.

cfergeau:
> This is not behaviour we normally suggest to use; file redirection for the setup flow.

We ask for logs for most issues filed, file redirection is one way of capturing these. If file redirection is never something we want the user to do, let's document this, and remove all usage of `IsTerminal`/`RunningInTerminal` from crc.

> Besides, you are judging about DOWNLOADING that has been untouched in this PR, so this is existing behaviour.

I know, but downloading and uncompressing currently behave differently, we disagree on which one need to be changed and fixed, hence the mention of the one *I* would change.

gbraad:
> currently behave differently,

Hence, addressed to remove complexity

> hence the mention of the one I would change.

Unfortunately, the suggestion is not in line with the schedule we have. We should be more pragmatic and consider breaking down issues and work on smaller increment by improvements issues instead of waxing and polishing PRs into perfection.


> We ask for logs for most issues filed

I look forward to the modification of the issue template to also include:

```
$ crc setup --log-level debug
```

cmd/crc/cmd/setup.go
```diff
@@ -11,19 +11,22 @@ import (
...
-	checkOnly bool
+	checkOnly             bool
+	forceShowProgressbars bool
 )
 
 func init() {
 	setupCmd.Flags().Bool(crcConfig.ExperimentalFeatures, false, "Allow the use of experimental features")
 	setupCmd.Flags().StringP(crcConfig.Bundle, "b", constants.GetDefaultBundlePath(crcConfig.GetPreset(config)), "Bundle to use for instance")
 	setupCmd.Flags().BoolVar(&checkOnly, "check-only", false, "Only run the preflight checks, don't try to fix any misconfiguration")
+	setupCmd.Flags().BoolVar(&forceShowProgressbars, "show-progressbars", false, "Allows show the porgress bars for download and extraction")
```
cfergeau: I'd phrase this as `Always show the progress bars for ...`

pkg/os/terminal/terminal.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"os"
+
+	xterminal "golang.org/x/term"
+)
+
+var (
+	// Global variable to force output regardless if terninal
+	ForceShowOutput bool = false
```
cfergeau: `bool` can be removed, see the comments from github actions:
```


var-declaration: should omit type bool from declaration of var ForceShowOutput; it will be inferred from the right-hand side (revive)

```

pkg/os/terminal/terminal.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func IsShowTerminalOutput() bool {
+	// if this is a terminal or set to force output
+	return IsTerminal() || ForceShowOutput
+}
+
+func IsTerminal() bool {
+	return xterminal.IsTerminal(int(os.Stdout.Fd()))
```
cfergeau, line 20: There's already a `RunningInTerminal` helper in `pkg/os/util.go`

pkg/os/terminal/terminal.go
```diff
@@ -0,0 +1,21 @@
...
+
+	xterminal "golang.org/x/term"
+)
+
+var (
+	// Global variable to force output regardless if terninal
+	ForceShowOutput bool = false
+)
+
+func IsShowTerminalOutput() bool {
```
cfergeau, line 14: `IsShowTerminalOutput` does not read great. Just `ShowTerminalOutput`? or `ShouldShowTerminalOutput`?

pkg/extract/extract.go
```diff
@@ -13,28 +13,29 @@ import (
...
 	"github.com/xi2/xz"
-	terminal "golang.org/x/term"
 )
 
 const minSizeForProgressBar = 100_000_000
 
-func UncompressWithFilter(tarball, targetDir string, showProgress bool, fileFilter func(string) bool) ([]string, error) {
-	return uncompress(tarball, targetDir, fileFilter, showProgress && terminal.IsTerminal(int(os.Stdout.Fd())))
+func UncompressWithFilter(tarball, targetDir string, fileFilter func(string) bool) ([]string, error) {
+	return uncompress(tarball, targetDir, fileFilter, false) // never show detailed output
```
cfergeau, line 27: Arguably this could go in a separate commit to make it very clear that the arg is removed because all the callers call `UncompressWithFilter` with `false`.

cfergeau (commented):
Thanks for making this configurable!

pkg/os/terminal/terminal.go
gbraad: already dealt with

pkg/os/terminal/terminal.go
gbraad: Already dealt with

gbraad:
Not sure why the test fails with:

```
pkg/extract/extract.go:20:2: "golang.org/x/term" imported but not used as terminal (typecheck)
	terminal "golang.org/x/term"
```

as this file does not contain a reference to `x/term`

pkg/os/terminal/terminal.go
gbraad: Merely following the `bool` pattern to prefix with `Is`. Minor concern.

pkg/extract/extract.go
gbraad: Comment added

cfergeau (approved):

gbraad:
has been verified by @evidolob to provide the progressbars with `crc setup --show-progressbars`

PR #3549, opened 2023-03-17
build: replace gofmt with goimports

anjannath:


Makefile
```diff
@@ -247,8 +247,8 @@ e2e-story-registry: install
 
 
 .PHONY: fmt
-fmt:
-	@gofmt -l -w $(SOURCE_DIRS)
+fmt: $(TOOLS_BINDIR)/goimports
+	@goimports -l -w $(SOURCE_DIRS)
```
cfergeau: needs to be `$(TOOLS_BINDIR)/goimports`

tools/tools.mk
```diff
@@ -11,3 +11,6 @@ $(TOOLS_BINDIR)/gomod2rpmdeps: $(TOOLS_DIR)/go.mod
 
 $(TOOLS_BINDIR)/mockery: $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR) && GOBIN="$(TOOLS_BINDIR)" go install github.com/vektra/mockery/v2@latest
+
+$(TOOLS_BINDIR)/goimports: $(TOOLS_DIR)/go.mod
+	cd $(TOOLS_DIR) && GOBIN="$(TOOLS_BINDIR)" go install golang.org/x/tools/cmd/goimports@latest
```
cfergeau, line 16: you need to add an import in https://github.com/crc-org/crc/blob/main/tools/dummy.go and update vendoring

cfergeau (commented):
A few things to fix, but looks good.

tools/tools.mk
anjannath: ahh yes, updated now, not very clear this needs to be done, maybe we should add another target to the `tools.mk` to run `go mod tidy and go mod vendor`

cfergeau: Without the addition to `dummy.go`, `go mod tidy && go mod vendor` will not vendor the newly added code :-/ We could check that `golang.org/x/tools/cmd/goimports` is in `dummy.go` somehow, but not worth spending a lot  of time on this.

cfergeau (approved):

PR #3559, opened 2023-03-24
update windows preflight check for user in required groups

anjannath:
remove the username from `Get-LocalGroupMember` cmdlet and get the list of users
belonging to a group and perform string comparison to determine if current user
is part of the group

this should work around the need to supply the username with or without the domain
part depending on whether the machine was currently domain joined

### DIFF
--- a/pkg/crc/preflight/preflight_checks_windows.go
+++ b/pkg/crc/preflight/preflight_checks_windows.go
@@ -100,15 +100,60 @@ func checkHyperVServiceRunning() error {
 }
 
 func checkUserPartOfCrcUsersAndHypervAdminsGroup() error {
-	_, _, err := powershell.Execute(fmt.Sprintf("Get-LocalGroupMember -Group 'crc-users' -Member '%s'", username()))
+	groupMembers, _, err := powershell.Execute(`(Get-LocalGroupMember -Group 'crc-users').Name`)
 	if err != nil {
 		return err
 	}
-
+	logging.Debug("Checking current user is in the 'crc-user' group")
+	if !usernameInMembersList(username(), groupMembers) {
+		return fmt.Errorf("Could not find: %s in the 'crc-users' group", username())
+	}
 	// https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems
 	// BUILTIN\Hyper-V Administrators => S-1-5-32-578
-	_, _, err = powershell.Execute(fmt.Sprintf("Get-LocalGroupMember -SID 'S-1-5-32-578' -Member '%s'", username()))
-	return err
+	groupMembers, _, err = powershell.Execute(`(Get-LocalGroupMember -SID 'S-1-5-32-578').Name`)
+	if err != nil {
+		return err
+	}
+	logging.Debug("Checking current user is in the 'Hyper-v Administrators' group")
+	if !usernameInMembersList(username(), groupMembers) {
+		return fmt.Errorf("Could not find: %s in the 'Hyper-v Administrators' group", username())
+	}
+	return nil
+}
+
+func usernameInMembersList(username, members string) bool {
+	m := strings.Split(members, "\n")
+	var memberList []string
+	// remove any empty elements
+	for _, elem := range m {
+		if strings.TrimSpace(elem) != "" {
+			memberList = append(memberList, strings.TrimSpace(elem))
+		}
+	}
+	logging.Debugf("group members: %s", strings.Join(memberList, ","))
+	for _, member := range memberList {
+		// we get the members of a group in the form domain\username
+		// if the current username is also returned in the same form
+		// check if that full domain\username is present in the list
+		if strings.Contains(username, "\\") {
+			if username == member {
+				return true
+			}
+			continue
+		}
+
+		// if we get only the username without domain part then we
+		// compare it with the username part of the group members
+		m := strings.Split(member, "\\")
+		if len(m) == 2 {
+			if m[1] == username {
+				return true
+			}
+		} else {
+			logging.Warnf("Got group member's name in unexpected format: %s", member)
+		}
+	}
+	return false
 }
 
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -100,15 +100,43 @@ func checkHyperVServiceRunning() error {
...
+}
+
+func usernameInMembersList(username, members string) bool {
+	m := strings.Split(members, "\n")
+	var memberList []string
+	// remove any empty elements
+	for _, elem := range m {
+		if strings.TrimSpace(elem) != "" {
+			memberList = append(memberList, strings.TrimSpace(elem))
+		}
```
cfergeau, line 131: This does not seem strictly needed, `strings.Contains()` would do the right thing even on an empty line?

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -100,15 +100,43 @@ func checkHyperVServiceRunning() error {
...
+	}
+	logging.Debugf("group members: %s", strings.Join(memberList, ","))
+	for _, member := range memberList {
+		if strings.Contains(member, username) {
```
cfergeau: Couldn't this give false positives?
https://stackoverflow.com/questions/4548476/powershell-list-local-users-and-their-groups has this example:
```
PS C:\WINDOWS\system32> Get-LocalGroupMember -name users

ObjectClass Name                             PrincipalSource 
----------- ----                             --------------- 
User        DESKTOP-R05QDNL\someUser1        Local           
User        DESKTOP-R05QDNL\someUser2        MicrosoftAccount
Group       NT AUTHORITY\INTERACTIVE         Unknown  
```

I'm guessing the list we'll get in CRC is `DESKTOP-R05QDNL\someUser1`, `DESKTOP-R05QDNL\someUser2`?
If `username()` is `DESK` or `some`, `strings.Contains()` will succeed for both even if none of these users are in the list. I think you should be splitting the string on `\` and check the last part for equality?
Is there any go module/any Windows API we could use to get this information in an easier to consume format rather than spawning powershell commands to do it?

anjannath: > I'm guessing the list we'll get in CRC is DESKTOP-R05QDNL\someUser1, DESKTOP-R05QDNL\someUser2?
If username() is DESK or some, strings.Contains() will succeed for both even if none of these users are in the list. I think you should be splitting the string on \ and check the last part for equality?

yes yes, i missed this.

> Is there any go module/any Windows API we could use to get this information in an easier to consume format rather than spawning powershell commands to do it?

the golang `os/user` or `os` packages don't have anything to get the members of a group, i think we can get this information with the wmi library we already use (but that seems more complicated.. https://stackoverflow.com/a/26384427), similarly using win32 api from go would be complicated, powershell seems like the simplest

pkg/crc/preflight/preflight_checks_windows.go
anjannath: yeah, not strictly needed, but it messes up the debug log line when we print the members

pkg/crc/preflight/preflight_checks_windows.go
anjannath: @cfergeau updated the PR, PTALA :)

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -100,15 +100,58 @@ func checkHyperVServiceRunning() error {
 }
 
 func checkUserPartOfCrcUsersAndHypervAdminsGroup() error {
-	_, _, err := powershell.Execute(fmt.Sprintf("Get-LocalGroupMember -Group 'crc-users' -Member '%s'", username()))
+	groupMembers, _, err := powershell.Execute(`(Get-LocalGroupMember -Group 'crc-users').Name`)
 	if err != nil {
 		return err
 	}
-
+	logging.Debug("Checking current user part of 'crc-user' group")
```
cfergeau: `Checking if the current user is in the 'crc-user' group` or `is part of the 'crc-user' group`

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -100,15 +100,58 @@ func checkHyperVServiceRunning() error {
...
+	}
 	// https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems
 	// BUILTIN\Hyper-V Administrators => S-1-5-32-578
-	_, _, err = powershell.Execute(fmt.Sprintf("Get-LocalGroupMember -SID 'S-1-5-32-578' -Member '%s'", username()))
-	return err
+	groupMembers, _, err = powershell.Execute(`(Get-LocalGroupMember -SID 'S-1-5-32-578').Name`)
+	if err != nil {
+		return err
+	}
+	logging.Debug("Checking current user part of 'Hyper-v Administrators' group")
```
cfergeau: Same comment as before

pkg/crc/preflight/preflight_checks_windows.go
```diff
@@ -100,15 +100,58 @@ func checkHyperVServiceRunning() error {
...
+		}
+
+		// if we get only the username without domain part then we
+		// compare it with the username part of the group members
+		m := strings.Split(member, "\\")
+		if len(m) == 2 {
+			if m[1] == username {
+				return true
+			}
+		}
```
cfergeau, line 154: What if len(m) != 2 ? Can it happen? Should we warn?

cfergeau (commented):
A few minor comments but overall looks good.

pkg/crc/preflight/preflight_checks_windows.go
anjannath: it can happen if the response from `Get-LocalGroupMember` is not what we expect, (i.e without the domain part), would be better if we log this

PR #3562, opened 2023-03-27
Update logging when podman or other preset in use

praveenkumar:
With this patch you shouldn't get following logging message in case of microshift bundle.
```
INFO Loading bundle: crc_microshift_hyperv_4.12.6_amd64...
INFO Creating CRC VM for Podman ...
```

**Fixes:** Issue #3557

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -242,11 +242,7 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 			return nil, errors.Wrap(err, "Failed to ask for pull secret")
 		}
 
-		if crcBundleMetadata.IsOpenShift() {
-			logging.Infof("Creating CRC VM for %s %s...", startConfig.Preset, crcBundleMetadata.GetOpenshiftVersion())
-		} else {
-			logging.Infof("Creating CRC VM for Podman %s...", crcBundleMetadata.GetPodmanVersion())
-		}
+		logging.Infof("Creating CRC VM for %s %s...", startConfig.Preset.ForDisplay(), crcBundleMetadata.GetVersion())
 
 		sharedDirs := []string{}
 		if homeDir, err := os.UserHomeDir(); err == nil {
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -24,7 +24,7 @@ var (
 
 func init() {
 	setupCmd.Flags().Bool(crcConfig.ExperimentalFeatures, false, "Allow the use of experimental features")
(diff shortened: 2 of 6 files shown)


anjannath (approved):

cfergeau (commented):
Correct spellings are OKD, MicroShift, OpenShift and Podman. I don't think it's correct to special case Podman so that we use the correct capitalization, while for the others we are fine with printing the preset name as a string which will be lower case.
You could add a `startConfig.Preset.ForDisplay()` helper or such to solve this.

praveenkumar:
>Correct spellings are OKD, MicroShift, OpenShift and Podman. I don't think it's correct to special case Podman so that we use the correct capitalization, while for the others we are fine with printing the preset name as a string which will be lower case.

@cfergeau I changed `Podman` string also to use preset's String method.

> You could add a startConfig.Preset.ForDisplay() helper or such to solve this.

Is there any reason why don't we change the const string value for preset to have correct spellings?
```
const (
	Podman     Preset = "Podman"
	OpenShift  Preset = "OpenShift"
	OKD        Preset = "OKD"
	Microshift Preset = "MicroShift"
)
```

cfergeau:
> Is there any reason why don't we change the const string value for preset to have correct spellings?

At this point, it would break users' configurations. A config file with `preset: "podman"` would no longer be valid as the preset is `Podman`. Same for existing scripts, ...
More generally, I prefer to keep the enum internal representation separate from its string/user-visible representation. "Separate" as in, it should be opaque to code using it in our code base. If the code needs a string, then I prefer an explicit function call to get it as a string.

pkg/crc/machine/start.go
```diff
@@ -242,10 +242,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 			return nil, errors.Wrap(err, "Failed to ask for pull secret")
 		}
 
-		if crcBundleMetadata.IsOpenShift() {
-			logging.Infof("Creating CRC VM for %s %s...", startConfig.Preset, crcBundleMetadata.GetOpenshiftVersion())
+		if crcBundleMetadata.IsPodman() {
+			logging.Infof("Creating CRC VM for %s %s...", startConfig.Preset, crcBundleMetadata.GetPodmanVersion())
 		} else {
-			logging.Infof("Creating CRC VM for Podman %s...", crcBundleMetadata.GetPodmanVersion())
+			logging.Infof("Creating CRC VM for %s %s...", startConfig.Preset, crcBundleMetadata.GetOpenshiftVersion())
```
cfergeau: Untested for now, but something like this should work too:
https://github.com/cfergeau/crc/commit/fcd0de25f6e8bea6414798e73f127badcee84cd5

praveenkumar: tested and that works, cherry-picked this commit

cfergeau (approved):

Issue #3514, opened 2023-02-14
[BUG] Podman preset is broken for linux if known_hosts file contain an existing pub key for 192.168.130.11


praveenkumar:
We have to check how to avoid podman remote command to add key to the `.ssh/known_hosts` file or may be remove it before starting the VM.

```
$ crc config set preset podman
$ crc setup && crc start
[...]
podman runtime is now running.

Use the 'podman' command line interface:
  $ eval $(crc podman-env)
  $ podman-remote COMMAND

$ which podman-remote
/home/prkumar/.crc/bin/oc/podman-remote

$ podman-remote system connection ls
Name        URI                                                            Identity                                  Default
crc         ssh://core@192.168.130.11:22/run/user/1000/podman/podman.sock  /home/prkumar/.crc/machines/crc/id_ecdsa  true
crc-root    ssh://core@192.168.130.11:22/run/podman/podman.sock            /home/prkumar/.crc/machines/crc/id_ecdsa  false

$ podman-remote --log-level=debug ps  
INFO[0000] podman-remote filtering at log level debug   
DEBU[0000] Called ps.PersistentPreRunE(podman-remote --log-level=debug ps) 
DEBU[0000] SSH Ident Key "/home/prkumar/.crc/machines/crc/id_ecdsa" SHA256:kFXwhQAuHL9azolLI2Y5nrW+N3eRcmsn+/VaiBcooDc ecdsa-sha2-nistp521 
WARN[0000] ssh host key mismatch for host 192.168.130.11:22, got key SHA256:jsWTDSKOVr55u+q5PSmEZjoqLfSvNXLFmUq6hjkiXxg of type ecdsa-sha2-nistp256 
Error: failed to connect: ssh: handshake failed: knownhosts: key mismatch

$ ssh-keygen -R 192.168.130.11

$ podman-remote --log-level=debug ps  
INFO[0000] podman-remote filtering at log level debug   
DEBU[0000] Called ps.PersistentPreRunE(podman-remote --log-level=debug ps) 
DEBU[0000] SSH Ident Key "/home/prkumar/.crc/machines/crc/id_ecdsa" SHA256:kFXwhQAuHL9azolLI2Y5nrW+N3eRcmsn+/VaiBcooDc ecdsa-sha2-nistp521 
INFO[0000] key SHA256:jsWTDSKOVr55u+q5PSmEZjoqLfSvNXLFmUq6hjkiXxg added to /home/prkumar/.ssh/known_hosts 
[...]

PR #3564, opened 2023-03-29
Issue #3514 Remove crc instance entries from known_hosts file

anjannath:


cmd/crc/cmd/delete.go
```diff
@@ -87,3 +93,49 @@ func (s *deleteResult) prettyPrintTo(writer io.Writer) error {
...
+	}
+	defer f.Close()
+
+	tempHostsFile, err := os.CreateTemp("", "crc")
+	if err != nil {
+		return fmt.Errorf("Unable to create temp file: %w", err)
+	}
+	defer func() {
+		os.Remove(tempHostsFile.Name())
+		tempHostsFile.Close()
```
praveenkumar: shouldn't we first close the file and the remove?

cmd/crc/cmd/delete.go
```diff
@@ -87,3 +93,49 @@ func (s *deleteResult) prettyPrintTo(writer io.Writer) error {
...
+		return fmt.Errorf("Error trying to change permissions for temp file: %w", err)
+	}
+
+	var foundCRCEntries bool
+	scanner := bufio.NewScanner(f)
+	writer := bufio.NewWriter(tempHostsFile)
+	for scanner.Scan() {
+		if strings.Contains(scanner.Text(), "[127.0.0.1]:2222") {
+			foundCRCEntries = true
+			continue
```
praveenkumar: This is only for user networking mode we have to consider system network mode in case of Linux

cmd/crc/cmd/delete.go
```diff
@@ -87,3 +93,49 @@ func (s *deleteResult) prettyPrintTo(writer io.Writer) error {
...
+			return fmt.Errorf("Error while writing hostsfile content to temp file: %w", err)
+		}
+	}
+
+	if err := writer.Flush(); err != nil {
+		return fmt.Errorf("Error while flushing buffered content to temp file: %w", err)
+	}
+
+	if foundCRCEntries {
+		return os.Rename(tempHostsFile.Name(), knownHostsPath)
```
praveenkumar: The only thing is if a user perform `crc delete` and then in different terminal ssh to some server around same time then it might be possible to have a inconsistent state of this file, may we should use some locking mechanism to overcome it?

cmd/crc/cmd/delete.go
anjannath: right, the order is wrong, thanks

cmd/crc/cmd/delete.go
anjannath: yes but since system-mode networking is only on linux, and there i think less people should be using the podman preset.. and after we upgrade podman bundle again from 4.4.1 to the next version this will not be required even for user-mode networking

cmd/crc/cmd/delete.go
anjannath: rename will be atomic, i don't think we need a locking mechanism since that other process is outside of CRC i think the os should take care of this automatically?

anjannath: we can also do something like --  keep a backup of the original known_hosts in the same location and if there's an error from `os.Rename` we can restore (rename) the backed-up version

cmd/crc/cmd/delete.go
cfergeau: For what it's worth, I don't think the order matters. 
After `os.Remove(tempHostsFile.Name())` I'd expect that it's still possible to even read/write to the file even if it's no longer visible in the file system. Read/Write/Close act on an open file descriptor, which is valid until it's closed. Remove acts on a filename, which is not directly related to the file descriptor for this file.

anjannath: hmm, i was also thinking if the last `os.Rename()` would be problematic as that also happens before the close, good to know!

praveenkumar: On windows this order matter, In the past we had issue when we tried to delete a file without closing it file descriptor and getting error. On unix/linux side order doesn't matter.

cmd/crc/cmd/delete.go
praveenkumar: I don't think it is fixed on 4.4.1 side also so we do need to have a long term solution for it.

cmd/crc/cmd/delete.go
```diff
@@ -87,3 +93,49 @@ func (s *deleteResult) prettyPrintTo(writer io.Writer) error {
 	}
 	return nil
 }
+
+// TODO: remove it after we do one more upgrade of the podman bundle from 4.4.1
+func removeCRCHostEntriesFromKnownHosts() error {
```
praveenkumar: This function should be part of `/pkg/crc/machine/delete.go` so that same can be consumed by API endpoints otherwise cmd and API will be different.

praveenkumar: remove the `TODO` also since it is going to be long lived.

anjannath: fixed, thanks

cmd/crc/cmd/delete.go
anjannath: ~updated, it should remove system-mode entries also now~

edit: removed the call `vm.IP()` to get the IP address to remove the entries in system mode networking, as that breaks `crc delete` when instance is not running, it only removes the user-mode networking entries starting with `[127.0.0.1]:2222`

cmd/crc/cmd/delete.go
anjannath: yep, on windows got the error, now closing the files before the rename, thx

pkg/crc/machine/delete.go
```diff
@@ -39,5 +43,55 @@ func (client *client) Delete() error {
...
+
+	if err := tempHostsFile.Chmod(0600); err != nil {
+		return fmt.Errorf("Error trying to change permissions for temp file: %w", err)
+	}
+
+	var foundCRCEntries bool
+	scanner := bufio.NewScanner(f)
+	writer := bufio.NewWriter(tempHostsFile)
+	for scanner.Scan() {
+		if strings.Contains(scanner.Text(), "[127.0.0.1]:2222") {
```
praveenkumar: So we are not considering system mode networking?

anjannath: no only user-mode networking, is there another alternative to get the VM ip in system-mode networking to remove the entries, other then `vm.IP()` method?

praveenkumar: @anjannath I would just use hard coded IP (in system mode) which is `192.168.130.11` in case of linux, on mac and windows we are anyway using usermode by default.

cfergeau: If we want to avoid hardcoded IPs and not have to rely on defaults being in use, maybe we can add the host key in the bundle metadata file, and remove all instances of this key from the known_hosts file.

anjannath: right, this should handle the most-likely scenarios, updated!

anjannath: > If we want to avoid hardcoded IPs and not have to rely on defaults being in use, maybe we can add the host key in the bundle metadata file, and remove all instances of this key from the known_hosts file.

(sorry, didn't see this before) this'd be a better solution, i'd say we take this PR for 2.16 release and work on this approach for the next sprint

cfergeau: Yeah, definitely not an approach we can take right now

anjannath:
@cfergeau @praveenkumar addressed all the review comments so far, do you guys have any further comments or good to merge?

pkg/crc/machine/delete.go
```diff
@@ -39,5 +44,55 @@ func (client *client) Delete() error {
...
+		}
+	}
+
+	if err := writer.Flush(); err != nil {
+		return fmt.Errorf("Error while flushing buffered content to temp file: %w", err)
+	}
+
+	if foundCRCEntries {
+		f.Close()
+		tempHostsFile.Close()
```
cfergeau: You could also catch the errors from `Close()`

pkg/crc/machine/delete.go
```diff
@@ -39,5 +44,55 @@ func (client *client) Delete() error {
 			logging.Warnf("Failed to remove crc contexts from kubeconfig: %v", err)
 		}
 	}
+	return removeCRCHostEntriesFromKnownHosts()
+}
+
+func removeCRCHostEntriesFromKnownHosts() error {
+	knownHostsPath := filepath.Join(constants.GetHomeDir(), ".ssh", "known_hosts")
```
cfergeau, line 51: Fwiw, I would have added this as a generic helper in `pkg/crc/ssh`, `ssh.RemoveIPFromKnownHosts("192.168.130.11")` or such. This can be changed later if desired.

cfergeau (commented):
Hadn't looked at the code until now, but nothing much to say as this has already gone through multiple rounds of reviews :)

cfergeau (approved):

Issue #3515, opened 2023-02-14
[BUG] Running a container using `system` mode networking with exposed port failed in linux


praveenkumar:
In linux default network-mode is `system` so if a user try to run the podman preset and then try to create/run a container which expose the port it fails with following error.
```
$ podman-remote run -d -p 8080:80 docker.io/httpd:2.4 --log-level debug
Trying to pull docker.io/library/httpd:2.4...
Getting image source signatures
Copying blob sha256:9e8776e4b876795dc00ce93ee6409bc492f17894f9e3b61aeaf9a00c610c703c
Copying blob sha256:b7f64f2f874701178f37872cfc13dbf12adbcfda23ed90d73b8df12421a851e4
Copying blob sha256:f506d7aab6524182207e901fa77ff8ed100bf402fe4c7a8ad6fe178b8d8aa311
Copying blob sha256:05289ee4f2842c404631e433b1bbb9251cc3262bd0942b0c4be974061c5fa12e
Copying blob sha256:bb263680fed18eecdc67f885094df6f589bafc19004839d7fdf141df236a61aa
Copying config sha256:3a4ea134cf8e081516a776ce184dedc28986f941ed214b9012dc888049480f5a
Writing manifest to image destination
Storing signatures
Error: Post "http://gateway.containers.internal/services/forwarder/expose": dial tcp: lookup gateway.containers.internal: no such host
```

because podman bundle expect that `gateway.containers.internal` should be resolve able since it is podman machine requirement when exposing the port and for `system` mode networking we don't have it ( no gvsior stack running ). So either we need to check we can make it work for `system` mode networking or switch to user-mode networking for podman preset.

cfergeau:
The use of gateway.containers.internal does not seem to be possible to disable from podman-machine without code changes:
https://github.com/containers/podman/blob/main/libpod/networking_machine.go#L117-L132
https://github.com/containers/common/blob/main/pkg/machine/machine.go#L66-L70

praveenkumar:
https://github.com/crc-org/snc/pull/675 and https://github.com/crc-org/snc/pull/676 adds a marker file (podman/microshift bundle) and as workaround we do need to delete it during system mode networking.

praveenkumar:
Try to make `usermode` networking as default for linux

jsliacan:
I am not reproducing this locally (on f38), using default `network-mode`:
```
$ podman-remote --log-level debug run -d -p 8080:80 docker.io/httpd:2.4
[...]

PR #3568, opened 2023-03-30
Remove container.conf file if system mode networking in use

praveenkumar:
The container.conf file has a setting called machine_enabled in the engine section, which lets the podman client know that a command is running on an instance created with the podman machine command. This allows the use of gvisor-tap-vsock when a container is created with an exposed port. However, this setting should be disabled for system mode networking, so that it doesn't prevent the creation of containers that need to expose a port.

With this patch user can able to use microshift/podman preset with system mode networking and start containers with exposed port but not able to access that service from the host.

workaround for  https://github.com/crc-org/crc/issues/3515

**Fixes:** Issue #3515 

## Testing

Without this PR try to start the microshift preset with system mode networking and it will fail with following error
```
Error running post start: Executing systemctl action failed:  ssh command error:
command : sudo systemctl start crc-dnsmasq.service
: Job for crc-dnsmasq.service failed because the control process exited with error code.
See "systemctl status crc-dnsmasq.service" and "journalctl -xe" for details.

<crc_vm> $  journalctl -u crc-dnsmasq

-- Logs begin at Thu 2023-03-30 07:17:40 EDT, end at Thu 2023-03-30 07:18:26 EDT. --
Mar 30 07:17:44 api.crc.testing systemd[1]: Starting Podman container-7a7c65fb2a35b8ff0b943aab307c11755cf09f3fbe5f4444e27caff129f7d276.service...
Mar 30 07:17:54 api.crc.testing podman[1626]: Error: unable to start container "7a7c65fb2a35b8ff0b943aab307c11755cf09f3fbe5f4444e27caff129f7d276": " POST http://gateway.containers.internal/services/forwarder/expose": dial tcp: lookup gateway.containers.internal: no such host
```

with this PR microshift preset can successfully start using system mode networking. Even with podman preset after running it with system mode networking we can start containers which need port exposed.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -399,6 +399,12 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 		}
 	}
 
+	if !client.useVSock() {
+		if err := removeContainerConfFile(sshRunner); err != nil {
+			return nil, err
+		}
+	}
+
 	if _, _, err := sshRunner.RunPrivileged("make root Podman socket accessible", "chmod 777 /run/podman/ /run/podman/podman.sock"); err != nil {
 		return nil, errors.Wrap(err, "Failed to change permissions to root podman socket")
 	}
@@ -971,5 +977,19 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 		return err
 	}
 
+	return nil
+}
+
+// The containers.conf file has a setting called machine_enabled in the engine section,
+// which lets the podman client know that a command is running on an instance created
+// with the podman machine command. This allows the use of gvisor-tap-vsock when a
+// container is created with an exposed port. However, this setting should be disabled
+// for system mode networking, so that it doesn't prevent the creation of containers that
+// need to expose a port.
+// - https://github.com/crc-org/crc/issues/3515
+func removeContainerConfFile(sshRunner *crcssh.Runner) error {
+	if _, _, err := sshRunner.RunPrivileged("remove /etc/containers/containers.conf to disable machine_enabled", "rm -fr /etc/containers/containers.conf"); err != nil {
+		return errors.Wrap(err, "Failed to remove /etc/containers/containers.conf")
+	}
 	return nil
 }


pkg/crc/machine/start.go
```diff
@@ -973,3 +979,17 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 
 	return nil
 }
+
+// The container.conf file has a setting called machine_enabled in the engine section,
```
cfergeau: The name is `containers.conf`

pkg/crc/machine/start.go
```diff
@@ -973,3 +979,17 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
...
+// which lets the podman client know that a command is running on an instance created
+// with the podman machine command. This allows the use of gvisor-tap-vsock when a
+// container is created with an exposed port. However, this setting should be disabled
+// for system mode networking, so that it doesn't prevent the creation of containers that
+// need to expose a port.
+// - https://github.com/crc-org/crc/issues/3515
+func removeContainerConfFile(sshRunner *crcssh.Runner) error {
+	if _, _, err := sshRunner.RunPrivileged("remove machine_enabled check", "rm -fr /etc/containers/containers.conf"); err != nil {
```
cfergeau: If what we want to do is to set `machine_enabled` to false, why do we remove the whole file? It might even be possible to drop a file in ` /etc/containers/containers.conf.d/` to do this for us instead of changing `containers.conf`.

Note that this setting is deprecated (but still parsed) https://github.com/containers/common/blob/main/pkg/config/config.go#L343-L346

It uses a `/etc/containers/podman-machine` file now https://github.com/containers/common/blob/main/pkg/machine/machine.go
Maybe machine_enabled can always be set to false.

praveenkumar: >If what we want to do is to set machine_enabled to false, why do we remove the whole file? It might even be possible to drop a file in  /etc/containers/containers.conf.d/ to do this for us instead of changing containers.conf.

@cfergeau  There only one config in this file which we are creating as part of snc so  instead parsing and updating it or even dropping file to `/containers.conf.d` just to remove our changes in case system mode networking.

For microshift
https://github.com/crc-org/snc/blob/master/podman_changes.ks#L18-L21

For podman
https://github.com/crc-org/snc/blob/podman/fcos-config.yaml#L115-L121

> It uses a /etc/containers/podman-machine file now https://github.com/containers/common/blob/main/pkg/machine/machine.go

not aware of those changes but might be with latest podman, for microshift it is still 4.2.0 , in that case we need to make change in the snc side to have that marker file.

cfergeau: >  There only one config in this file which we are creating as part of snc so instead parsing and updating it or even dropping file to /containers.conf.d just to remove our changes in case system mode networking.

And the day we want to add a new entry to this file, first it's going to break because we'll have forgotten about this, then we'll need to fix both snc and crc for something which should have been a snc-only change. Can we change our bundle creation process to create `/etc/containers/containers.conf.d/10-podman-machine-enabled.conf`, and then create/delete this file as needed?

praveenkumar: >Can we change our bundle creation process to create /etc/containers/containers.conf.d/10-podman-machine-enabled.conf, and then create/delete this file as needed?

Instead of that I will use the marker file which is part of https://github.com/containers/common/blob/main/pkg/machine/machine.go since the machine-enabled config is deprecated soon.

praveenkumar: I have created https://github.com/crc-org/snc/pull/675 and https://github.com/crc-org/snc/pull/676 on snc side so meanwhile we can use this hack (at least till bundle is updated with marker file) and then put another PR to remove the maker file as needed. At this moment this will atleast unblock microshift bundle on system mode networking side.

cfergeau: > meanwhile we can use this hack (at least till bundle is updated with marker file) and then put another PR to remove the maker file as needed.

Fine with me!

>  At this moment this will atleast unblock microshift bundle on system mode networking side.

This could also be documented as a known issue, `podman run -p 8080:80` failing by default on linux with the microshift bundle is imo not blocking, but I understand others might have different opinions on this matter

cfergeau: `remove /etc/containers/containers.conf to disable machine_enabled`, or just `remove /etc/containers/containers.conf` would be more explicit.

pkg/crc/machine/start.go
```diff
@@ -973,3 +979,17 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
...
+		return errors.Wrap(err, "Failed to remove machine_enabled check")
```
cfergeau: I'd say `Failed to remove /etc/containers/containers.conf ` as this what actually failed.

cfergeau (commented):
See https://github.com/crc-org/crc/pull/3568/commits/1611a418cfc339039b923987f48a323ae8bf1924#r1154203126 , but I'm fine with moving forward with this PR.

PR #3574, opened 2023-04-04
CI: add microshift preset for workflow

praveenkumar:
As of now this workflow make sure `crc start` works without any error for microshift preset. We will add some workload on this flow later to make sure all k8s component also working as expected.

### DIFF
--- a/.github/workflows/macos-microshift.yml
+++ b/.github/workflows/macos-microshift.yml
@@ -1,0 +1,49 @@
+name: Build macOS installer
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


gbraad (approved):

.github/workflows/macos-microshift.yml
```diff
@@ -0,0 +1,49 @@
+name: Build macOS installer
```
anjannath, line 1: change the name to something micorshift, `Test microhift preset`

PR #3575, opened 2023-04-04
Update microshift flow with secret and name update

praveenkumar:


.github/workflows/macos-microshift.yml
```diff
@@ -39,11 +39,21 @@ jobs:
         run: make NO_CODESIGN=1 out/macos-universal/crc-macos-installer.pkg
       - name: Install crc pkg
         run: sudo installer -pkg out/macos-universal/crc-macos-installer.pkg -target /
+      - name: check if CRC_PULL_SECRET exists
+        env:
+          super_secret: ${{ secrets.CRC_PULL_SECRET }}
+        if: ${{ env.super_secret == '' }}
+        run: 'echo the secret \"CRC_PULL_SECRET\" has not been made; echo please go to \"settings \> secrets \> actions\" to create it'
```
cfergeau: `made`? set?

.github/workflows/macos-microshift.yml
```diff
@@ -13,6 +13,8 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    env:
+      SHELL: /bin/bash
```
cfergeau: why do we need to use bash?
there's a 'standard' way of selecting the shell, `shell: bash` should work.

cfergeau (approved):

praveenkumar:
/hold

Need to make sure microshift workflow is green. There is some issue wit workflow as it is not taking the pull secret from env.

.github/workflows/macos-microshift.yml
```diff
@@ -35,6 +35,18 @@ jobs:
...
+      - name: check if CRC_PULL_SECRET exists
+        shell: bash
+        run: |
+          if [ "${{ secrets.CRC_PULL_SECRET }}" != '' ]; then
+             echo "the secret \"CRC_PULL_SECRET\" has not been set, please go to \"settings \> secrets \> actions\" to create it"
+             exit 1
+          fi
+      - name: Write the pull secret to json file
+        run: |
+          echo $CRC_PULL_SECRET > temp-ps.json
```
cfergeau: Could this be ` echo ${{ secrets.CRC_PULL_SECRET }} > temp-ps.json` without the intermediate environment variable?

praveenkumar: As per https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow it should be set as env or STDIN before using it from `run`

praveenkumar:
Closing it from the fork and creating another PR from the branch.

PR #3578, opened 2023-04-04
Update microshift flow with pull secret

praveenkumar:
Created this PR from the branch instead from a fork because as per https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow the runner not able to pass the secret from forked repo.

### DIFF
--- a/.github/workflows/macos-microshift.yml
+++ b/.github/workflows/macos-microshift.yml
@@ -35,6 +35,21 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go }}
+      - name: check if OCP_PULL_SECRET exists
+        shell: bash
+        run: |
+          if [ ${PULL_SECRET} == "" ]; then
+            echo "The secret \"OCP_PULL_SECRET\" is not set; echo please go to \"settings > secrets > actions\" to create it"
+            exit 1
+          fi
+        env:
+          PULL_SECRET: ${{ secrets.OCP_PULL_SECRET }}
+      - name: Write the pull secret to json file
+        shell: bash
+        run: |
+          echo $PULL_SECRET > temp-ps.json
+        env:
+          PULL_SECRET: ${{ secrets.OCP_PULL_SECRET }}
       - name: Build macOS installer
         run: make NO_CODESIGN=1 out/macos-universal/crc-macos-installer.pkg
       - name: Install crc pkg
@@ -46,4 +61,4 @@ jobs:
       - name: Wait 10 sec for the daemon to serve
         run: sleep 10
       - name: Run crc with microshift preset
-        run: crc start
+        run: crc start --pull-secret-file temp-ps.json
--- a/.github/workflows/macos-microshift.yml
+++ b/.github/workflows/macos-microshift.yml
@@ -1,4 +1,4 @@
-name: Build macOS installer
+name: Test MicroShift preset
 on:
   push:
     branches:


.github/workflows/macos-microshift.yml
```diff
@@ -35,6 +35,18 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go }}
+      - name: check if OSP_PULL_SECRET exists
+        shell: bash
+        run: |
+          if [ "${{ secrets.OSP_PULL_SECRET }}" == '' ]; then
```
cfergeau: You said in the other PR that this is not valid, that this has to be an env var?

praveenkumar: @cfergeau I was looking at the docs https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-secrets but when I tried out this does let me access the secrets context. I updated the `echo` also let's hope that also works.

praveenkumar: And workflow job is failing so I think it is better to use documented flow instead using secret context directly from if or from echo command.

.github/workflows/macos-microshift.yml
```diff
@@ -35,6 +35,18 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go }}
+      - name: check if OSP_PULL_SECRET exists
+        shell: bash
+        if: ${{ env.PULL_SECRET == '' }}
+        run: 'echo the secret \"OCP_PULL_SECRET\" has not been made; echo please go to \"settings \> secrets \> actions\" to create it'
+        env:
+          PULL_SECRET: ${{ secrets.OSP_PULL_SECRET }}
```
cfergeau: Not clear why this is named `OSP_PULL_SECRET`, and why the error message mentions `OCP_PULL_SECRET`

.github/workflows/macos-microshift.yml
```diff
@@ -35,6 +35,18 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go }}
+      - name: check if OSP_PULL_SECRET exists
+        shell: bash
+        if: ${{ env.PULL_SECRET == '' }}
+        run: 'echo the secret \"OCP_PULL_SECRET\" has not been made; echo please go to \"settings \> secrets \> actions\" to create it'
```
cfergeau: `has not been made` -> `is not set/available/..` ?
This won't stop the execution of the action with a failure though?

.github/workflows/macos-microshift.yml
praveenkumar: OSP => openshift platform will change error message.

.github/workflows/macos-microshift.yml
praveenkumar: > This won't stop the execution of the action with a failure though?

yes it won't stop the action. if we want to stop the execution then we need to use `run` with bash script.

praveenkumar: but the best thing here is this task not run if the pull secret is set.

cfergeau: The test is not really useful if it does not stop execution of the whole workflow. I'd remove it if it's only informative

.github/workflows/macos-microshift.yml
cfergeau: RHOSP is OpenStack, OpenShift is RHOCP.

.github/workflows/macos-microshift.yml
praveenkumar: It is still better to be informative but I will make it blocker execution.

.github/workflows/macos-microshift.yml
praveenkumar: In my brain somehow OSP stuck that time what I really meant OCP => Openshift container platform :(

cfergeau (approved):

cfergeau (approved):

Issue #3576, opened 2023-04-04
[BUG] missing preset info for microshift on config command


adrianriobo:
On pre-released version `CRC version: 2.16.0+05b62a75` when getting information for config options on presets, there is no value for microshift:

```
preset                               Virtual machine preset (valid values are: podman, openshift and okd)
```

anjannath:
fixed in #3580

PR #3580, opened 2023-04-05
config/settings: Add microshift preset to helper

praveenkumar:
**Fixes:** Issue #3576

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -70,7 +70,7 @@ func RegisterSettings(cfg *Config) {
 
 	// Preset setting should be on top because CPUs/Memory config depend on it.
 	cfg.AddSetting(Preset, string(preset.OpenShift), validatePreset, RequiresDeleteAndSetupMsg,
-		fmt.Sprintf("Virtual machine preset (valid values are: %s, %s and %s)", preset.Podman, preset.OpenShift, preset.OKD))
+		fmt.Sprintf("Virtual machine preset (valid values are: %s)", preset.AllPresets()))
 	// Start command settings in config
 	cfg.AddSetting(Bundle, defaultBundlePath(cfg), validateBundlePath, SuccessfullyApplied, BundleHelpMsg(cfg))
 	cfg.AddSetting(CPUs, defaultCPUs(cfg), validateCPUs, RequiresRestartMsg,
--- a/pkg/crc/preset/preset.go
+++ b/pkg/crc/preset/preset.go
@@ -22,6 +22,14 @@ var presetMap = map[Preset]string{
 	Microshift: string(Microshift),
 }
 
+func AllPresets() []Preset {
+	var keys []Preset
+	for k := range presetMap {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
 func (preset Preset) String() string {
 	presetStr, presetExists := presetMap[preset]
 	if !presetExists {
--- a/pkg/crc/preset/preset.go
+++ b/pkg/crc/preset/preset.go
@@ -15,18 +15,19 @@ const (
 	Microshift Preset = "microshift"
 )
 
+var presetMap = map[Preset]string{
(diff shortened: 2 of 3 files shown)


pkg/crc/config/settings.go
```diff
@@ -70,7 +70,7 @@ func RegisterSettings(cfg *Config) {
 
 	// Preset setting should be on top because CPUs/Memory config depend on it.
 	cfg.AddSetting(Preset, string(preset.OpenShift), validatePreset, RequiresDeleteAndSetupMsg,
-		fmt.Sprintf("Virtual machine preset (valid values are: %s, %s and %s)", preset.Podman, preset.OpenShift, preset.OKD))
+		fmt.Sprintf("Virtual machine preset (valid values are: %s, %s, %s and %s)", preset.Podman, preset.OpenShift, preset.Microshift, preset.OKD))
```
cfergeau: A more future-proof alternative is to add `func preset.AllPresets() []preset.Preset` and use this in `RegisterSettings` and `validatePreset`

pkg/crc/preset/preset.go
```diff
@@ -15,16 +15,20 @@ const (
...
-		return string(OpenShift)
-	case OKD:
-		return string(OKD)
-	case Microshift:
-		return string(Microshift)
+	for pSet, pString := range presetMap() {
+		if pSet == preset {
+			return pString
+		}
 	}
```
cfergeau, line 37: you have a map, you don't need to iterate over it
```
presetStr, presetExists := presetMap()[preset]
if !presetExists {
    return "invalid"
}
return presetStr
```

pkg/crc/preset/preset.go
```diff
@@ -15,16 +15,20 @@ const (
...
 )
 
+func presetMap() map[Preset]string {
+	return map[Preset]string{
+		Podman:     string(Podman),
+		OpenShift:  string(OpenShift),
+		OKD:        string(OKD),
+		Microshift: string(Microshift),
+	}
+}
```
cfergeau, line 31:
```
var presetMap = map[Preset]string {
...
}
```
should be enough, and allow us to create the map only once?

pkg/crc/config/settings.go
```diff
@@ -70,7 +70,7 @@ func RegisterSettings(cfg *Config) {
 
 	// Preset setting should be on top because CPUs/Memory config depend on it.
 	cfg.AddSetting(Preset, string(preset.OpenShift), validatePreset, RequiresDeleteAndSetupMsg,
-		fmt.Sprintf("Virtual machine preset (valid values are: %s, %s and %s)", preset.Podman, preset.OpenShift, preset.OKD))
+		fmt.Sprintf("Virtual machine preset (valid values are: %s)", preset.AllPresets()))
```
cfergeau, line 73: What does the string looks like ? I don't know how go converts an array to a string.

praveenkumar: @cfergeau it have `Virtual machine preset (valid values are: [openshift okd podman microshift])` if we really want to have as string then another helper is required.

cfergeau: `strings.Join(preset.AllPresets, ", ")` might work.

praveenkumar: > strings.Join(preset.AllPresets, ", ") might work

No, beacuse it complain `preset.AllPresets` is not `[]string` type so not able to use.

cfergeau: Let's say `[openshift okd podman microshift]` is good enough if you don't want the extra code.

pkg/crc/preset/preset.go
praveenkumar: updated.

pkg/crc/preset/preset.go
praveenkumar: updated

cfergeau (approved):

PR #3585, opened 2023-04-06
bundle/repository: Create oc symlink/copy in case of all preset except podman

praveenkumar:


pkg/crc/machine/bundle/repository.go
```diff
@@ -84,7 +84,7 @@ func (repo *Repository) Use(bundleName string) (*CrcBundleInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	if bundleInfo.IsOpenShift() {
+	if !bundleInfo.IsPodman() {
```
praveenkumar: I though of making change in the `createSymlinkOrCopyOpenShiftClient` and only have this operation in case the `oc` binary present in the bundle otherwise don't do anything but then it means if we miss the `oc` binary in the bundle then there is no error out.

cfergeau: There will be an error in `verify()`, I don't think we should be hardcoding "bundle X must have oc" in too many places to avoid this bug happening again in the future.

cfergeau (changes requested):
https://github.com/crc-org/crc/pull/3585#discussion_r1159395115

pkg/crc/machine/bundle/repository.go
```diff
@@ -97,7 +95,7 @@ func (repo *Repository) Use(bundleName string) (*CrcBundleInfo, error) {
 
 func (bundle *CrcBundleInfo) copyExecutableFromBundle(destDir string, fileType FileType, executableName string) error {
 	srcPath := bundle.getHelperPath(fileType)
-	if srcPath == "" {
+	if srcPath == "" || !crcos.FileExists(srcPath) {
```
cfergeau: I don't think the `crcos.FileExists(srcPath)` check is needed, if the file is listed in the metadata, then `verify()` checked that the file exists. I think initially we did not have the preset checks, and we relied on `srcPath==""` to do the right thing for podman/openshift bundles. Then we added the IsPodman/IsOpenShift checks, but we kept the`srcPath==""` checks anyway, even if they are redundant.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -217,7 +217,7 @@ func (bundle *CrcBundleInfo) verify() error {
 		bundle.GetKernelPath(),
 		bundle.GetInitramfsPath(),
 	}
-	if bundle.IsOpenShift() {
+	if !bundle.IsPodman() {
```
cfergeau, line 220: It seems the `verify()` method is not exactly doing what we expect. If we want an error if the metadata file does not contain data for `oc` or for a `kubeconfig` file, then we should remove `if file == "" { continue }` from the loop. 
Having both this check and changing the list of files which are checked according to the preset is odd.

cfergeau (commented):
Commit log: ` bundle/repository: Create oci/podman symlink/copy in case of it preset in bundle`

oci -> oc, preset -> present. I'd replace "in case of it" with "if" to make the commit log shorter, and maybe also drop the "repository" prefix to also make it shorter.

Shorter sentences in the full log would also be nice as it's currently very hard to read :(

pkg/crc/machine/bundle/metadata.go
praveenkumar: In case of linux and windows `bundle.GetKernelPath()` and `bundle.GetInitramfsPath()` are going to be empty string so this `if file == "" { contunue }` block is use to ignore. either we make the filelist as per platform so we can remove that block or have it as it is for general purpose.

cfergeau: The commit log is still a bit misleading, with this change, we will get a complaint with the microshift bundle if its metadata specifies oc/kubeconfig files. But if the bundle metadata is missing oc/kubeconfig files, I don't think this will be caught.

cfergeau (commented):
verify() still has some caveats, but this should fix the bug we've been seeing with the previous release.

pkg/crc/machine/bundle/metadata.go
cfergeau: Hmm, I don't know why I did not see your message from 3 days ago before posting my message from today.

cfergeau (approved):

PR #3586, opened 2023-04-06
remove CRC specific entries from known_hosts file also during cleanup

anjannath:


cfergeau (approved):

praveenkumar:
/hold

testing this on mac

praveenkumar:
@anjannath since you are on to it can you also add that because with current state when I delete or cleanup then it always add a new line at the end of known_host file.

```patch
$ git diff
diff --git a/pkg/crc/ssh/keys.go b/pkg/crc/ssh/keys.go
index b09e4885b..1b391a6b0 100644
--- a/pkg/crc/ssh/keys.go
+++ b/pkg/crc/ssh/keys.go
@@ -102,11 +102,15 @@ func RemoveCRCHostEntriesFromKnownHosts() error {
        scanner := bufio.NewScanner(f)
        writer := bufio.NewWriter(tempHostsFile)
        for scanner.Scan() {
-               if strings.Contains(scanner.Text(), "[127.0.0.1]:2222") || strings.Contains(scanner.Text(), "192.168.130.11") {
+               readLine := strings.TrimSpace(scanner.Text())
+               if readLine == "" {
+                       continue
+               }
+               if strings.Contains(readLine, "[127.0.0.1]:2222") || strings.Contains(readLine, "192.168.130.11") {
                        foundCRCEntries = true
                        continue
                }
-               if _, err := writer.WriteString(fmt.Sprintf("%s\n", scanner.Text())); err != nil {
+               if _, err := writer.WriteString(fmt.Sprintf("%s\n", readLine)); err != nil {
                        return fmt.Errorf("Error while writing hostsfile content to temp file: %w", err)
                }
        }

```

anjannath:
I think this is because of the `writer.WriteString(fmt.Sprintf("%s\n", scanner.Text()))` the last line also adds an extra `\n` and hence the newline after delete or cleanup

praveenkumar:
> I think this is because of the `writer.WriteString(fmt.Sprintf("%s\n", scanner.Text()))` the last line also adds an extra `\n` and hence the newline after delete or cleanup

so in case the entry is in between of `known_hosts` file then after removing it also add a empty line there so when it is at the end, everytime it remove and add the empty line so better to ignore the empty line and continue.

anjannath:
> @anjannath since you are on to it can you also add that because with current state when I delete or cleanup then it always add a new line at the end of known_host file.
> 
> ```diff
> $ git diff
> diff --git a/pkg/crc/ssh/keys.go b/pkg/crc/ssh/keys.go
> index b09e4885b..1b391a6b0 100644
> --- a/pkg/crc/ssh/keys.go
> +++ b/pkg/crc/ssh/keys.go
> @@ -102,11 +102,15 @@ func RemoveCRCHostEntriesFromKnownHosts() error {
>         scanner := bufio.NewScanner(f)
>         writer := bufio.NewWriter(tempHostsFile)
>         for scanner.Scan() {
> -               if strings.Contains(scanner.Text(), "[127.0.0.1]:2222") || strings.Contains(scanner.Text(), "192.168.130.11") {
> +               readLine := strings.TrimSpace(scanner.Text())
> +               if readLine == "" {
> +                       continue
> +               }
> +               if strings.Contains(readLine, "[127.0.0.1]:2222") || strings.Contains(readLine, "192.168.130.11") {
>                         foundCRCEntries = true
>                         continue
>                 }
> -               if _, err := writer.WriteString(fmt.Sprintf("%s\n", scanner.Text())); err != nil {
> +               if _, err := writer.WriteString(fmt.Sprintf("%s\n", readLine)); err != nil {
>                         return fmt.Errorf("Error while writing hostsfile content to temp file: %w", err)
>                 }
>         }
> ```

with this change also the last newline is added, and its also removing pre-existing empty lines in the file, not sure if we should remove those as that might be added deliberately by the user

if we want to avoid repeateadly adding the newline at the end, then we need to switch the splitFunc that scanner is using to return the full line with the `\n` character, so that we don't have to add it, and as a result the last line will not add an additional newline

anjannath:
@praveenkumar updated, PTAL

PR #3594, opened 2023-04-12
Status: Add microshift as part of status

### LINKED ISSUE #3592: [BUG] Status reports podman preset while microshift is running
### General information

  * OS: macOS
  * Hypervisor: hyperkit 
  * Did you run `crc setup` before starting it: Yes
  * Running CRC on: Laptop

## CRC version
```bash
CRC version: 2.16.0+05b62a75
OpenShift version: 4.12.9
Podman version: 4.4.1
```
  
## CRC status
```bash
DEBU CRC version: 2.16.0+05b62a75
DEBU OpenShift version: 4.12.9
DEBU Podman version: 4.4.1
DEBU Running 'crc status'
CRC VM:          Running
RAM Usage:       1.196GB of 3.913GB
Disk Usage:      4.746GB of 16.1GB (Inside the CRC VM)
Cache Usage:     72.04GB
Cache Directory: /Users/yevhen/.crc/cache
```

## CRC config
```bash
```

## Host Operating System
```bash
ProductName:		macOS
ProductVersion:		13.3
BuildVersion:		22E252
```

### Steps to reproduce

  1.  Configure crc to use `microshift` bundle
  2. Start crc with `microshift` bundle
  3. Call `curl -GET -i --unix-socket <home>/.crc/crc-http.sock http://localhost/api/status` to get status in json 

### Expected
```json
{"CrcStatus":"Running", "Preset":"microshift", ...}
```

### Actual
```json
{"CrcStatus":"Running", "Preset":"podman",  ...}
```

praveenkumar:
With this patch now for microshift preset user will get following
output.
```
$ ./crc status
CRC VM:          Running
MicroShift:      Running (v4.12.9)
RAM Usage:       1.144GB of 3.913GB
Disk Usage:      4.647GB of 16.1GB (Inside the CRC VM)
Cache Usage:     208.9GB

$ ./crc status -ojson
{
  "success": true,
  "crcStatus": "Running",
  "microshiftStatus": "Running",
  "microshiftVersion": "4.12.9",
  "diskUsage": 4646580224,
  "diskSize": 16095641600,
  "cacheUsage": 208867411129,
  "cacheDir": "/home/prkumar/.crc/cache",
  "ramSize": 3912945664,
  "ramUsage": 1144094720,
  "preset": "microshift"
}
```

Fixes: #3592

### DIFF
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -180,10 +180,10 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
 		{"CRC VM", s.CrcStatus},
 	}
 
-	if s.OpenShiftVersion != "" {
+	if s.Preset == preset.OpenShift {
 		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
 	}
-	if s.PodmanVersion != "" {
+	if s.Preset == preset.Podman {
 		lines = append(lines, line{"Podman", s.PodmanVersion})
 	}
 	if s.Preset == preset.Microshift {
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -186,6 +186,9 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
 	if s.PodmanVersion != "" {
 		lines = append(lines, line{"Podman", s.PodmanVersion})
 	}
+	if s.Preset == preset.Microshift {
+		lines = append(lines, line{"MicroShift", openshiftStatus(s)})
+	}
 
 	if s.RAMSize != -1 && s.RAMUsage != -1 {
 		lines = append(lines, line{"RAM Usage", fmt.Sprintf(
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -80,7 +80,7 @@ func (status *Status) IsReady() bool {
 }
 
 func GetClusterOperatorsStatus(ctx context.Context, ip string, kubeconfigFilePath string) (*Status, error) {
-	lister, err := kubernetesClient(ip, kubeconfigFilePath)
+	lister, err := openshiftClient(ip, kubeconfigFilePath)
 	if err != nil {
 		return nil, err
 	}
@@ -158,7 +158,7 @@ type operatorLister interface {
 	List(ctx context.Context, opts metav1.ListOptions) (*openshiftapi.ClusterOperatorList, error)
 }
 
-func kubernetesClient(ip string, kubeconfigFilePath string) (*clientset.Clientset, error) {
+func openshiftClient(ip string, kubeconfigFilePath string) (*clientset.Clientset, error) {
 	config, err := kubernetesClientConfiguration(ip, kubeconfigFilePath)
 	if err != nil {
 		return nil, err
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -295,13 +295,12 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 		return nil, errors.Wrap(err, "Error getting the machine state")
 	}
 	if vmState == state.Running {
-		if !vm.bundle.IsOpenShift() {
-			logging.Infof("A CRC VM for Podman %s is already running", vm.bundle.GetPodmanVersion())
+		logging.Infof("A CRC VM for %s %s is already running", startConfig.Preset.ForDisplay(), vm.bundle.GetVersion())
+		if vm.bundle.IsPodman() {
 			return &types.StartResult{
 				Status: vmState,
 			}, nil
 		}
-		logging.Infof("A CRC VM for OpenShift %s is already running", vm.bundle.GetOpenshiftVersion())
 		clusterConfig, err := getClusterConfig(vm.bundle)
 		if err != nil {
 			return nil, errors.Wrap(err, "Cannot create cluster configuration")
--- a/pkg/crc/machine/status.go
+++ b/pkg/crc/machine/status.go
@@ -10,6 +10,8 @@ import (
 	"github.com/crc-org/crc/pkg/crc/machine/state"
 	"github.com/crc-org/crc/pkg/crc/machine/types"
 	"github.com/crc-org/crc/pkg/crc/preset"
+	"github.com/crc-org/crc/pkg/crc/systemd"
+	sdState "github.com/crc-org/crc/pkg/crc/systemd/states"
 	"github.com/pkg/errors"
 )
 
@@ -31,17 +33,26 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
 		return nil, errors.Wrap(err, "Cannot get machine state")
 	}
 
+	clusterStatusResult := &types.ClusterStatusResult{
+		CrcStatus: vmStatus,
+	}
+	switch {
+	case vm.bundle.IsPodman():
+		clusterStatusResult.PodmanVersion = vm.bundle.GetPodmanVersion()
+		clusterStatusResult.Preset = preset.Podman
+	case vm.bundle.IsMicroshift():
+		clusterStatusResult.OpenshiftStatus = types.OpenshiftStopped
+		clusterStatusResult.OpenshiftVersion = vm.bundle.GetOpenshiftVersion()
+		clusterStatusResult.Preset = preset.Microshift
+	default:
+		clusterStatusResult.OpenshiftStatus = types.OpenshiftStopped
+		clusterStatusResult.OpenshiftVersion = vm.bundle.GetOpenshiftVersion()
+		clusterStatusResult.Preset = preset.OpenShift
+	}
+
 	if vmStatus != state.Running {
 		clusterStatusResult := &types.ClusterStatusResult{
 			CrcStatus: vmStatus,
-		}
-		if vm.bundle.IsOpenShift() {
-			clusterStatusResult.OpenshiftStatus = types.OpenshiftStopped
-			clusterStatusResult.OpenshiftVersion = vm.bundle.GetOpenshiftVersion()
-			clusterStatusResult.Preset = preset.OpenShift
-		} else {
-			clusterStatusResult.PodmanVersion = vm.bundle.GetPodmanVersion()
-			clusterStatusResult.Preset = preset.Podman
 		}
 		return clusterStatusResult, nil
 	}
@@ -52,18 +63,15 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
 	}
 
 	diskSize, diskUse := client.getDiskDetails(vm)
-	clusterStatusResult := &types.ClusterStatusResult{
-		CrcStatus: state.Running,
-		DiskUse:   diskUse,
-		DiskSize:  diskSize,
-	}
-	if vm.bundle.IsOpenShift() {
+	clusterStatusResult.CrcStatus = state.Running
+	clusterStatusResult.DiskUse = diskUse
+	clusterStatusResult.DiskSize = diskSize
+
+	switch {
+	case vm.bundle.IsMicroshift():
+		clusterStatusResult.OpenshiftStatus = getMicroShiftStatus(vm)
+	case vm.bundle.IsOpenShift():
 		clusterStatusResult.OpenshiftStatus = getOpenShiftStatus(context.Background(), ip)
-		clusterStatusResult.OpenshiftVersion = vm.bundle.GetOpenshiftVersion()
-		clusterStatusResult.Preset = preset.OpenShift
-	} else {
-		clusterStatusResult.PodmanVersion = vm.bundle.GetPodmanVersion()
-		clusterStatusResult.Preset = preset.Podman
 	}
 
 	ramSize, ramUse := client.getRAMStatus(vm)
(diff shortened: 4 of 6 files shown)


evidolob (approved):
This work as expected:

```json
{"CrcStatus":"Running",
"OpenshiftStatus":"",
"MicroShiftStatus":"Running",
"MicroShiftVersion":"4.12.9",
"DiskUse":4739710976,
"DiskSize":16095641600,
"RAMUse":979648512,
"RAMSize":3913109504,
"Preset":"microshift"
}
```

pkg/crc/machine/status.go
```diff
@@ -57,13 +64,20 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
...
 		DiskSize:  diskSize,
 	}
-	if vm.bundle.IsOpenShift() {
+	switch {
+	case vm.bundle.IsPodman():
+		logging.Info("podman")
+		clusterStatusResult.PodmanVersion = vm.bundle.GetPodmanVersion()
+		clusterStatusResult.Preset = preset.Podman
+	case vm.bundle.IsMicroshift():
+		logging.Info("microshift")
```
anjannath: i think this `log.Info` is unintended and left over from debugging

praveenkumar: @anjannath ah yes, I will remove those.

praveenkumar: @anjannath Updated.

anjannath (approved):

pkg/crc/machine/start.go
```diff
@@ -295,13 +295,13 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		return nil, errors.Wrap(err, "Error getting the machine state")
 	}
 	if vmState == state.Running {
-		if !vm.bundle.IsOpenShift() {
-			logging.Infof("A CRC VM for Podman %s is already running", vm.bundle.GetPodmanVersion())
+		if vm.bundle.IsPodman() {
+			logging.Infof("A CRC VM for %s %s is already running", startConfig.Preset.ForDisplay(), vm.bundle.GetPodmanVersion())
```
cfergeau: This can be changed to `logging.Infof("A CRC VM for %s %s is already running", startConfig.Preset.ForDisplay(), vm.bundle.GetVersion())` and be moved out of the `if {}` block.

```
  logging.Infof("A CRC VM for %s %s is already running", startConfig.Preset.ForDisplay(), vm.bundle.GetVersion())
  if vm.bundle.IsPodman() {
    return &types.StartResult{
				Status: vmState,
			}, nil
  }
  clusterConfig, err := getClusterConfig(vm.bundle)
  if err != nil {
    return nil, errors.Wrap(err, "Cannot create cluster configuration")
```

pkg/crc/cluster/clusteroperator.go
```diff
@@ -80,7 +80,7 @@ func (status *Status) IsReady() bool {
 }
 
 func GetClusterOperatorsStatus(ctx context.Context, ip string, kubeconfigFilePath string) (*Status, error) {
-	lister, err := kubernetesClient(ip, kubeconfigFilePath)
+	lister, err := openShiftClient(ip, kubeconfigFilePath)
```
cfergeau: fwiw, `openshiftClient` would be easier to read

cmd/crc/cmd/status.go
```diff
@@ -132,6 +134,7 @@ func createBars(cpuUse []int64, writer io.Writer) (ramBar *pb.ProgressBar, cpuBa
 func getStatus(client *daemonclient.Client, cacheDir string) *status {
 
 	clusterStatus, err := client.APIClient.Status()
+
```
cfergeau: unneeded change

cmd/crc/cmd/status.go
```diff
@@ -217,6 +225,13 @@ func openshiftStatus(status *status) string {
 	return string(status.OpenShiftStatus)
 }
 
+func microShiftStatus(status *status) string {
+	if status.MicroShiftVersion != "" {
+		return fmt.Sprintf("%s (v%s)", status.MicroShiftStatus, status.MicroShiftVersion)
+	}
+	return string(status.OpenShiftStatus)
```
cfergeau: MicroShiftStatus?

pkg/crc/machine/status.go
```diff
@@ -146,6 +158,24 @@ func getOpenShiftStatus(ctx context.Context, ip string) types.OpenshiftStatus {
...
+	sd := systemd.NewInstanceSystemdCommander(sshRunner)
+	microShiftServiceState, err := sd.Status("microshift")
+	if err != nil {
+		logging.Debugf("failed to get microshift service status: %v", err)
+		return types.OpenshiftUnreachable
+	}
+	if microShiftServiceState == sdState.Running {
+		return types.OpenshiftRunning
+	}
+	return types.OpenshiftStopped
```
cfergeau, line 172: fwiw, `crc start` waits for `oc get nodes` to be successful before deciding the cluster was started successfully, maybe `status` should use the same criterion.

cmd/crc/cmd/status.go
```diff
@@ -39,19 +39,21 @@ var statusCmd = &cobra.Command{
...
-	RAMUsage         int64                        `json:"ramUsage,omitempty"`
-	Preset           preset.Preset                `json:"preset"`
+	Success           bool                         `json:"success"`
+	Error             *crcErrors.SerializableError `json:"error,omitempty"`
+	CrcStatus         string                       `json:"crcStatus,omitempty"`
+	OpenShiftStatus   types.OpenshiftStatus        `json:"openshiftStatus,omitempty"`
+	OpenShiftVersion  string                       `json:"openshiftVersion,omitempty"`
+	MicroShiftStatus  types.OpenshiftStatus        `json:"microshiftStatus,omitempty"`
+	MicroShiftVersion string                       `json:"microshiftVersion,omitempty"`
+	PodmanVersion     string                       `json:"podmanVersion,omitempty"`
```
cfergeau: To be honest, this feels complex and error prone. `OpenShiftStatus`/`MicroShiftStatus` and `OpenShiftVersion`/`MicroShiftVersion` are mutually exclusive, and `OKDStatus`/`OKDVersion` would be missing. Maybe you could reuse `OpenShiftStatus`/`OpenShiftVersion` actually, and do something different when pretty-printing depending on the preset.
#3592 only mentions the `Preset` field being wrong, so we have some latitude with these fields.

cmd/crc/cmd/status.go
```diff
@@ -217,6 +225,13 @@ func openshiftStatus(status *status) string {
 	return string(status.OpenShiftStatus)
 }
 
+func microShiftStatus(status *status) string {
+	if status.MicroShiftVersion != "" {
```
cfergeau: This is called as 
```
if s.MicroShiftVersion != "" {
		lines = append(lines, line{"MicroShift", microShiftStatus(s)})
}
```
so this test is unneeded. I would change the test above to 
```
if preset == Microshift {
		lines = append(lines, line{"MicroShift", microShiftStatus(s)})
}
```

pkg/crc/machine/status.go
praveenkumar: `crc start` wait for cluster operator status, for `oc get nodes` it just wait for API endpoint. so if microshift service is running that means API is already up.

cmd/crc/cmd/status.go
praveenkumar: it is not only around prettyprint but also json side.

cmd/crc/cmd/status.go
praveenkumar: yes I didn't change because same what we used for `Openshift` and `podman`, may be another commit to use the preset instead those.

cfergeau: well, existing code is buggy, good opportunity to fix it.

cmd/crc/cmd/status.go
cfergeau: Even on the json side, if `Preset` is correctly set, I don't think we need redundant/mutually exclusive fields.

pkg/crc/machine/status.go
cfergeau: I am talking about `crc start` with the microshift preset, which only checks `oc get nodes` (this calls `waitForApiServer`)

cmd/crc/cmd/status.go
gbraad: Additional fields (or those that change names), ahs to be checked in the receiving end for `undefined` or existence. Which also complicates the receiving end a lot.

pkg/crc/api/client/types.go
```diff
@@ -20,15 +20,17 @@ type StartResult struct {
...
-	DiskSize         int64
-	RAMUse           int64
-	RAMSize          int64
-	Preset           preset.Preset
+	CrcStatus         string
+	OpenshiftStatus   string
+	OpenshiftVersion  string `json:"OpenshiftVersion,omitempty"`
+	MicroShiftStatus  string
+	MicroShiftVersion string `json:"MicroShiftVersion,omitempty"`
+	PodmanVersion     string `json:"PodmanVersion,omitempty"`
```
gbraad: This becomes as mess on the receiving end.

Just a `Version` and `Status` will do, as the `Preset` is the key value to determine the meaning

gbraad (changes requested):
This will complicate the logic on both ends and maybe even impede future changes. Each new type of bundle/preset will need new fields?

```json
{
  "success": true,    // <!--- ???
  "crcStatus": "Running",
  "microshiftStatus": "Running",
  "microshiftVersion": "4.12.9",
  ...
  "preset": "microshift"
}
```
could be:

```json
{
  "crcStatus": "Running",
  ...
  "preset": "microshift"
  "status": "Running",
  "version": "4.12.9",
}
```

pkg/crc/machine/status.go
praveenkumar: Created a follow up https://github.com/crc-org/crc/issues/3599 on this.

cmd/crc/cmd/status.go
```diff
@@ -180,12 +180,15 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
 		{"CRC VM", s.CrcStatus},
 	}
 
-	if s.OpenShiftVersion != "" {
+	if s.Preset == preset.OpenShift {
 		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
 	}
-	if s.PodmanVersion != "" {
+	if s.Preset == preset.Podman {
```
cfergeau, line 186: With openshift/microshift, we also expose podman, so I think we always want to print this if podman version is set.

cmd/crc/cmd/status.go
```diff
@@ -217,6 +220,13 @@ func openshiftStatus(status *status) string {
 	return string(status.OpenShiftStatus)
 }
 
+func microShiftStatus(status *status) string {
+	if status.Preset == preset.Microshift {
```
cfergeau: Shouldn't this be `if status.OpenShiftVersion != ""`? This makes the method identical to `openshiftStatus`?

pkg/crc/machine/status.go
```diff
@@ -57,13 +64,18 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
...
-	if vm.bundle.IsOpenShift() {
+	switch {
+	case vm.bundle.IsPodman():
+		clusterStatusResult.PodmanVersion = vm.bundle.GetPodmanVersion()
+		clusterStatusResult.Preset = preset.Podman
+	case vm.bundle.IsMicroshift():
+		clusterStatusResult.OpenshiftStatus = getMicroShiftStatus(vm)
+		clusterStatusResult.OpenshiftVersion = vm.bundle.GetOpenshiftVersion()
+		clusterStatusResult.Preset = preset.Microshift
+	default:
```
cfergeau: The code is starting to get quite repetitive, can you first create a `clusterStatusResult` with with the fields Preset and xxxVersion set. Then you can use this variable in the 2 code paths (cluster is stopped or running) and only set the correct Status  field. This can even be moved to different helpers for improved readabililty.

cmd/crc/cmd/status.go
praveenkumar: @cfergeau I discussed it with @gbraad about the same thing and he don't want to have podman version in case of openshift/microshift preset otherwise some user might think that this cluster running on podman.

cfergeau: Nobody can know this except you and Gerard. Please split this behaviour change to its own commit with a proper rationale so that it's obvious to everyone that the change is intentional, and why it's being done.

praveenkumar:
> This will complicate the logic on both ends and maybe even impede future changes. Each new type of bundle/preset will need new fields?

This should be now changed.

cmd/crc/cmd/status.go
```diff
@@ -180,10 +180,10 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
 		{"CRC VM", s.CrcStatus},
 	}
 
-	if s.OpenShiftVersion != "" {
+	if s.Preset == preset.OpenShift {
 		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
 	}
-	if s.PodmanVersion != "" {
+	if s.Preset == preset.Podman {
```
cfergeau, line 186: This could become a `switch`.

cfergeau (approved):

PR #3596, opened 2023-04-12
Start: use the router image as per openshift version for microshift preset

praveenkumar:
router controller images are cached in bundle as openshift version if it available for that specific version so better to use it.

Also we need to remove that hack by putting this resource in bundle.
- https://github.com/crc-org/snc/issues/660


pkg/crc/machine/start.go
```diff
@@ -888,7 +888,7 @@ func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Confi
 }
 
 func getRouterControllerImage(preset crcPreset.Preset, bundleInfo *bundle.CrcBundleInfo) string {
```
cfergeau, line 890: Should be `getRoutesControllerImage`, not `getRouterControllerImage`

cfergeau (approved):

PR #3605, opened 2023-04-18
Verify default bundles downloaded from mirror.openshift.com

### LINKED ISSUE #3493: verify signature of bundles downloaded from mirror.openshift.com
Currently we manually update the bundle hashes, which is error prone. two things needs to happen when bundle version is updated: `OPENSHIFT_VERSION` in the `Makefile` needs to be updated and the hashes in the following file needs to be updated.
https://github.com/crc-org/crc/blob/8c4fdb5227e772f6be984aa0c83a4598369a11a5/pkg/crc/machine/bundle/constants.go#L8-L30 

this was suggested in: https://github.com/crc-org/crc/pull/3486#pullrequestreview-1265185221

anjannath:
Fixes #3493


pkg/crc/gpg/gpg.go
```diff
@@ -32,3 +37,48 @@ func Verify(filePath, signatureFilePath string) error {
...
+	clearTextMsg := make([]byte, len(block.Bytes))
+	copy(clearTextMsg, block.Bytes)
+	canonicalizedMsgText := canonicalize(trimEachLine(string(clearTextMsg)))
+
+	id, err := goOpenpgp.CheckDetachedSignature(k, bytes.NewBufferString(canonicalizedMsgText), bytes.NewBuffer(sig))
+	if err != nil {
+		return "", fmt.Errorf("Invalid signature: %w", err)
+	}
+	logging.Debugf("Got valid signature from key id: %s", id.PrimaryKey.KeyIdString())
+	return trimEachLine(string(clearTextMsg)), nil
```
cfergeau, line 67: `return canonicalizedMsgText, nil` maybe?

pkg/crc/constants/keys.go
```diff
@@ -0,0 +1,50 @@
+package constants
+
+const (
+	// This public key is owned by the CRC team (crc@crc.dev), and is used
+	// to sign bundles uploaded to an image registry.
+	// It can be fetched with: `gpg --recv-key DC7EAC400A1BFDFB`
+	GPGPublicKey = `-----BEGIN PGP PUBLIC KEY BLOCK-----
```
cfergeau: This could be renamed to `CrcOrgPublicKey`  or `CrcOrgGpgKey`

pkg/crc/machine/bundle/metadata.go
```diff
@@ -305,6 +311,43 @@ func getBundleDownloadInfo(preset crcPreset.Preset) (*download.RemoteFile, error
 	return downloadInfo, nil
 }
 
+func getDefaultBundleVerifiedHash(preset crcPreset.Preset) (string, error) {
+	client := &http.Client{
+		Timeout:   5 * time.Second,
+		Transport: network.HTTPTransport(),
+	}
+	res, err := client.Get(constants.GetDefaultBundleSignedHashURL(preset))
```
cfergeau, line 305: I'd add a comment saying that this downloads `sha256sum.txt.sig` to help understand what this method is all about.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -305,6 +311,43 @@ func getBundleDownloadInfo(preset crcPreset.Preset) (*download.RemoteFile, error
...
+
+	lines := strings.Split(verifiedHashes, "\n")
+	for _, line := range lines {
+		if strings.Contains(line, constants.GetDefaultBundle(preset)) {
+			l := strings.Split(line, "  ")
+			if len(l) != 2 {
+				return "", errors.New("unable to find hash for default bundle")
+			}
+			return l[0], nil
+		}
```
cfergeau, line 329: I'd be more strict about the expect format, 
```
if strings.HasSuffix(line, constants.GetDefaultBundle(preset) {
    sha256sum := strings.TrimSuffix(line, "  "+constants.GetDefaultBundle(preset))
    return sha256sum, nil
}
```

cfergeau (commented):
A few comments, but looks good. Nice to be able to remove all these hashes we have to update for every releases!!

pkg/crc/machine/bundle/metadata.go
anjannath: thanks

pkg/crc/gpg/gpg.go
anjannath: `canonicalizedMsgText` will have `\r\n` line endings
 it feels a bit easier to deal with lines ending in `\n` especially in the unit test

praveenkumar (approved):

PR #3609, opened 2023-04-20
build: Use release binary for 'packagedir' target

### LINKED ISSUE #3608: [BUG] [Mac] 2.17.0 release does not use admin-helper and vfkit binary from installer
admin-helper and vfkit binary is downloading as part of setup instead using from the installer on Mac. Looks like we introduce some regression when we separated tray from the binary as part of package.

cfergeau:
This fixes https://github.com/crc-org/crc/issues/3608
Release binaries use installed vfkit/crc-admin-helper instead of
downloading them, and they use the correct telemetry key.


anjannath (approved):

evidolob (approved):
That work for me, `crc setup` runs without sudo prompt

Issue #3595, opened 2023-04-12
[BUG] Bundle logic not working as expected for the OKD preset


praveenkumar:
As reported in https://github.com/okd-project/okd/discussions/1556, we see errors like:

```
PS C:\Users\USER> crc config set preset okd
PS C:\Users\USER> crc setup
...
INFO Loading bundle: crc_okd_hyperv_4.12.0-0.okd-2023-02-18-033438_amd64...
Preset okd is used but bundle is provided for openshift preset
```

praveenkumar:
Issue is around https://github.com/crc-org/crc/blob/main/pkg/crc/machine/bundle/metadata.go#L188-L199 logic which return `OpenShift` preset for both `OKD` and `Openshift` side

```golang
func (bundle *CrcBundleInfo) GetBundleType() crcPreset.Preset {
	switch bundle.Type {
	case "snc", "snc_custom":
		return crcPreset.OpenShift
	case "podman", "podman_custom":
		return crcPreset.Podman
	case "microshift", "microshift_custom":
		return crcPreset.Microshift
	default:
		return crcPreset.OpenShift     // <!--- drops here
	}
}
```

and consuming it https://github.com/crc-org/crc/blob/main/pkg/crc/machine/start.go#L915-L920 here so in `start.go` logic need to be change something like below.

```golang
func bundleMismatchWithPreset(preset crcPreset.Preset, bundleMetadata *bundle.CrcBundleInfo) error {
       bundlePreset := bundleMetadata.GetBundleType()
        if preset == crcPreset.OKD && bundlePreset == crcPreset.OpenShift {
               return nil
        }
	if preset != bundleMetadata.GetBundleType() {
		return errors.Errorf("Preset %s is used but bundle is provided for %s preset", preset, bundleMetadata.GetBundleType())
	}
	return nil
}
```

A proper fix would be in the metadata the type should be changed to:
  * `okd` for OKD
  * `snc`/`openshift` for openshift preset.

Maurycy-Krzeminski:
On windows to make OKD work i created my fork with proposed fix:
https://github.com/Maurycy-Krzeminski/crc/commit/609507f062bff21c9ff4adc049ccdbcf2cc269a3
I built it and this was enough for crc to work with OKD bundl
There was no need to change bundleMismatchWithPreset function, i only changed GetBundleType function
After that i had to set kubeadmin-password because there appeared some other problems:
crc config  set  kubeadmin-password ChangeMe
[...]

PR #3610, opened 2023-04-20
OKD preset added to GetBundleType

Maurycy-Krzeminski:
**Fixes:** Issue #3595 [BUG] Bundle logic not working as expected for the OKD preset

## Solution/Idea

Fix based on: https://github.com/crc-org/crc/issues/3595#issuecomment-1505078190.  
After testing locally on Windows 10 there are no needs to change function bundleMismatchWithPreset().

## Proposed changes

1. Added posibility to get crcPreset OKD

## Testing
Was only tested by 'hand'.

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -186,16 +186,11 @@ func (bundle *CrcBundleInfo) GetBundleNameWithoutExtension() string {
 }
 
 func (bundle *CrcBundleInfo) GetBundleType() crcPreset.Preset {
-	switch bundle.Type {
-	case "snc", "snc_custom":
-		return crcPreset.OpenShift
-	case "podman", "podman_custom":
-		return crcPreset.Podman
-	case "microshift", "microshift_custom":
-		return crcPreset.Microshift
-	default:
-		return crcPreset.OpenShift
+	bundleType := strings.TrimSuffix(bundle.Type, "_custom")
+	if bundleType == "snc" {
+		bundleType = "openshift"
 	}
+	return crcPreset.ParsePreset(bundleType)
 }
 
 func (bundle *CrcBundleInfo) IsOpenShift() bool {
--- a/pkg/crc/machine/bundle/metadata_test.go
+++ b/pkg/crc/machine/bundle/metadata_test.go
@@ -164,3 +164,25 @@ func TestCustomBundleName(t *testing.T) {
 	customBundleName = GetCustomBundleName(customBundleName)
 	checkBundleName(t, customBundleName)
 }
+
+func TestGetBundleType(t *testing.T) {
+	var bundle CrcBundleInfo
+	bundle.Type = "okd"
+	require.Equal(t, preset.OKD, bundle.GetBundleType())
+	bundle.Type = "okd_custom"
+	require.Equal(t, preset.OKD, bundle.GetBundleType())
+	bundle.Type = "microshift"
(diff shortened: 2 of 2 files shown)


pkg/crc/machine/bundle/metadata.go
```diff
@@ -193,6 +193,8 @@ func (bundle *CrcBundleInfo) GetBundleType() crcPreset.Preset {
 		return crcPreset.Podman
 	case "microshift", "microshift_custom":
 		return crcPreset.Microshift
+	case "okd", "okd_custom":
+		return crcPreset.OKD
```
cfergeau: Untested, but I'd rewrite this method as:
```
diff --git a/pkg/crc/machine/bundle/metadata.go b/pkg/crc/machine/bundle/metadata.go
index c06484bb0..a73519e07 100644
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -186,16 +186,11 @@ func (bundle *CrcBundleInfo) GetBundleNameWithoutExtension() string {
 }
 
 func (bundle *CrcBundleInfo) GetBundleType() crcPreset.Preset {
-       switch bundle.Type {
-       case "snc", "snc_custom":
-               return crcPreset.OpenShift
-       case "podman", "podman_custom":
-               return crcPreset.Podman
-       case "microshift", "microshift_custom":
-               return crcPreset.Microshift
-       default:
-               return crcPreset.OpenShift
+       bundleType := strings.TrimSuffix(bundle.Type, "_custom")
+       if bundleType == "snc" {
+               bundleType = "openshift"
        }
+       return crcPreset.ParsePreset(bundleType)
 }
 
 func (bundle *CrcBundleInfo) IsOpenShift() bool {
```

Maurycy-Krzeminski: I changed for it to look like it, and added tests.
I'm not sure if i did them correctly, this is the first time i do anything for such a big Golang project.

praveenkumar: @Maurycy-Krzeminski This looks good, can you also squash all those commit to a single one?

gbraad: @praveenkumar perhaps provide instructions around rebase or a link to a doc for this:

In short it is something like
```
$ git rebase -i HEAD~3
```

your editor starts

```
p 609507f0
s bcd5fae1
s 9a36ee18
```

Which means to `p`ick the top one, and `s`quash the other commits in. When you close the editor, it will create a squashed commit that needs a new title and description.

Maurycy-Krzeminski: @praveenkumar  sure, squashed all into one commit  and forced pushed. 

In txt file there is how i locally started crc with okd preset 

[log.txt](https://github.com/crc-org/crc/files/11297855/log.txt)

the only error was:
ERRO Cannot update kubeconfig: open: Cannot find file

There was also a problem when kubeadmin-password was not set

Maurycy-Krzeminski: But after coppying kubeconfig from C:\Users\----\.crc\machines\crc to C:\Users\----\.kube\config 
then everything works with oc, and kubectl

PS E:\programming\forks\crc\out\windows-amd64> oc get nodes
NAME                 STATUS   ROLES                         AGE   VERSION
crc-dp64w-master-0   Ready    control-plane,master,worker   51d   v1.25.4+a34b9e9

gbraad:
@Maurycy-Krzeminski Thank you for the contribution!

PR #3614, opened 2023-04-25
vsock: expose remote podman ssh-tunnel connection to local unix socket

praveenkumar:
We are exposing it for podman preset but not for microshift and it is important for the tools which connect using unix socket. Also we are trying out syncing image from podman-machine instance to microshift instance using `skopeo` and that is required to make it work.

After this patch
```
$  curl --unix-socket ~/.crc/crc-http.sock http:/unix/network/services/forwarder/all | jq .
[
  {
    "local": "/home/prkumar/.crc/machines/crc/docker.sock",
    "remote": "ssh-tunnel://core@192.168.127.2:22/run/podman/podman.sock?key=%2Fhome%2Fprkumar%2F.crc%2Fmachines%2Fcrc%2Fid_ecdsa",
    "protocol": "unix"
  },
[...]
]
```

### DIFF
--- a/pkg/crc/machine/vsock.go
+++ b/pkg/crc/machine/vsock.go
@@ -86,13 +86,25 @@ const (
 )
 
 func vsockPorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint) []types.ExposeRequest {
+	socketProtocol := types.UNIX
+	socketLocal := constants.GetHostDockerSocketPath()
+	if runtime.GOOS == "windows" {
+		socketProtocol = types.NPIPE
+		socketLocal = constants.DefaultPodmanNamedPipe
+	}
 	exposeRequest := []types.ExposeRequest{
 		{
 			Protocol: "tcp",
 			Local:    net.JoinHostPort(localIP, strconv.Itoa(constants.VsockSSHPort)),
 			Remote:   net.JoinHostPort(virtualMachineIP, internalSSHPort),
 		},
+		{
+			Protocol: socketProtocol,
+			Local:    socketLocal,
+			Remote:   getSSHTunnelURI(),
+		},
 	}
+
 	switch preset {
 	case crcPreset.OpenShift, crcPreset.OKD, crcPreset.Microshift:
 		exposeRequest = append(exposeRequest,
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/vsock.go
```diff
@@ -93,6 +93,22 @@ func vsockPorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint)
 			Remote:   net.JoinHostPort(virtualMachineIP, internalSSHPort),
 		},
 	}
+
+	if preset == crcPreset.Podman || preset == crcPreset.Microshift {
```
cfergeau: We explicitly don't want this for the openshift/okd presets?
Apart from this, looks ok to me

praveenkumar: The more I think about it, looks like we can have it for each preset.

anjannath:
not related to this PR directly..

> Also we are trying out syncing image from podman-machine instance to microshift instance using skopeo and that is required to make it work.

we can also make `podman-machine`, crc's 'podman' and 'microshift' preset, all store the images in a shared dir on the host, 
the location to use for storing the container images can be configured in the (`/etc/storage.conf`, `~/.config/containers/storage.conf`) , or we'll face some filesystem limitations because of the host<->vm shared dir?

I had tried this previously to preserve the already pulled images after `crc delete` with 'podman' preset and that seems to work mostly

praveenkumar:
> we can also make podman-machine, crc's 'podman' and 'microshift' preset, all store the images in a shared dir on the host, the location to use for storing the container images can be configured in the (/etc/storage.conf, ~/.config/containers/storage.conf) , or we'll face some filesystem limitations because of the host<->vm shared dir? I had tried this previously to preserve the already pulled images after crc delete with 'podman' preset and that seems to work mostly

@anjannath Yes, same what I purpose but shared dir on the host is configured by user and if user choose different location then it fails, either we need to have a hard coded location which shouldn't change by user or use the podman sync mechanism. Also from podman desktop side for kind cluster they are storing the image on the disk and then loading it to the kind cluster using `kind load` command and it is easy to extend for CRC use case.

anjannath:
> but shared dir on the host is configured by user and if user choose different location then it fails, either we need to have a hard coded location which shouldn't change by user or use the podman sync mechanism

at least for 'crc' this is hard-coded, i mean not user configurable, its always the user's full home path that is recreated in the VM

> Also from podman desktop side for kind cluster they are storing the image on the disk and then loading it to the kind cluster using kind load command and it is easy to extend for CRC use case.

i'd have to look how its handled but i guess since kind is already running on top of the `podman` in podman-machine, it might not actually be making duplicates, but for CRC we'd be duplicating the same images inside the VMs for each preset

PR #3616, opened 2023-04-26
bump microshift openshift to 4.12.13 and update bundle hashes

anjannath:


cfergeau:
/lgtm

There are several failures in the github actions checks, better to make sure they are temporary issues before pushing this.

cfergeau:
/approve
(GH CI is green now)

praveenkumar:
@anjannath we might need to push that bundle to GCP for integration and e2e-crc test.

anjannath:
> @anjannath we might need to push that bundle to GCP for integration and e2e-crc test.

ahh, right, the extraction was failing for `crc setup` (and i verified the bundles on mirror are okay..) let me push the bundles to GCP then, is it using the bundle with `-b` flag for the tests then?

PR #3636, opened 2023-05-08
Config: Make preset config aware of cpu/memory 

praveenkumar:
Preset property is bound with cpu/memory/bundle and have default
values for it but our `set` function doesn't have map around those
settings and a user can do following without any error message.
```
$ crc config set preset microshift
$ crc config set cpus 2
$ crc config set preset openshift
```
and when user try to perform `crc start` then error message shown.

With this PR we are making the set function bit self aware of the
preset and it's bound properties (cpu/memory) and make sure if those
property doesn't validated for the preset then unset those so default
values going to take over.

with this PR following now works as expected.
```
$ curl -X POST -d '{"properties":{"preset": "microshift"}}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Properties":["preset"]}
$ curl -X POST -d '{"properties":{"cpus": "3"}}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Properties":["cpus"]}
$ curl -X POST -d '{"properties":{"preset": "openshift"}}' --unix-socket ~/.crc/crc-http.sock http:/c/api/config
{"Properties":["preset"]}
$ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/config | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   894  100   894    0     0  56924      0 --:--:-- --:--:-- --:--:-- 59600
{
  "Configs": {
    "bundle": "/Users/prkumar/.crc/cache/crc_vfkit_4.12.13_arm64.crcbundle",
    "consent-telemetry": "no",
    "cpus": 4,
    "disable-update-check": true,
    "disk-size": 31,
    "enable-cluster-monitoring": false,
    "enable-experimental-features": false,
    "enable-shared-dirs": true,
    "host-network-access": false,
    "http-proxy": "",
    "https-proxy": "",
    "ingress-http-port": 80,
    "ingress-https-port": 443,
    "kubeadmin-password": "",
    "memory": 9216,

[...]
```

### DIFF
--- a/pkg/crc/config/config.go
+++ b/pkg/crc/config/config.go
@@ -95,6 +95,17 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
 		return "", fmt.Errorf(invalidType, value, key)
 	}
 
+	// Make sure if user try to set same value which
+	// is default then just unset the value which
+	// anyway make it default and don't update it
+	// ~/.crc/crc.json (viper config) file.
+	if setting.defaultValue == castValue {
+		if _, err := c.Unset(key); err != nil {
+			return "", err
+		}
+		return "", nil
+	}
+
 	if setting.isSecret {
 		if err := c.secretStorage.Set(key, castValue); err != nil {
 			return "", err
--- a/pkg/crc/config/config.go
+++ b/pkg/crc/config/config.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 
 	"github.com/crc-org/crc/pkg/crc/preset"
+	"github.com/crc-org/crc/pkg/crc/validation"
 	"github.com/spf13/cast"
 )
 
@@ -93,6 +94,25 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
 		castValue = cast.ToString(value)
 	default:
 		return "", fmt.Errorf(invalidType, value, key)
+	}
+
+	// Preset is mapped with `memory`, `cpus` and `bundle` and
+	// we want to make sure if cpu or memory is less for a preset
+	// then default is set automatic.
+	if setting.Name == Preset {
+		preset := preset.ParsePreset(value.(string))
+		mem := c.Get(Memory)
+		if err := validation.ValidateMemory(mem.AsInt(), preset); err != nil {
+			if _, err := c.Unset(Memory); err != nil {
+				return "", err
+			}
+		}
+		cpu := c.Get(CPUs)
+		if err := validation.ValidateCPUs(cpu.AsInt(), preset); err != nil {
+			if _, err := c.Unset(CPUs); err != nil {
+				return "", err
+			}
+		}
 	}
 
 	// Make sure if user try to set same value which
--- a/pkg/crc/config/viper_config_test.go
+++ b/pkg/crc/config/viper_config_test.go
@@ -115,13 +115,14 @@ func TestViperConfigLoadDefaultValue(t *testing.T) {
 		Value:     4,
 		IsDefault: true,
 	}, config.Get(cpus))
-
 	_, err = config.Set(cpus, 4)
 	assert.NoError(t, err)
 
 	bin, err := os.ReadFile(configFile)
 	assert.NoError(t, err)
-	assert.JSONEq(t, `{"cpus":4}`, string(bin))
+	// Setting default value will not update
+	// write to the config so expected would be {}
+	assert.JSONEq(t, `{}`, string(bin))
 
 	assert.Equal(t, SettingValue{
 		Value:     4,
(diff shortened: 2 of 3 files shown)


pkg/crc/config/config.go
```diff
@@ -95,6 +96,36 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
...
+
+	// Make sure if user try to set same value which
+	// is default then just unset the value which
+	// anyway make it default and don't update it
+	// ~/.crc/crc.json (viper config) file.
+	if setting.defaultValue == castValue {
+		if _, err := c.Unset(key); err != nil {
+			return "", err
+		}
+		return "", nil
```
anjannath, line 126: i think we should print a message letting the user know the config is restored to its default, since we won't  see the config option in `config view` ?

pkg/crc/config/viper_config_test.go
```diff
@@ -121,7 +121,7 @@ func TestViperConfigLoadDefaultValue(t *testing.T) {
 
 	bin, err := os.ReadFile(configFile)
 	assert.NoError(t, err)
-	assert.JSONEq(t, `{"cpus":4}`, string(bin))
+	assert.JSONEq(t, `{}`, string(bin))
```
anjannath, line 125: maybe also add a comment here, `// default values are not written to config file`

gbraad:
Started with:

```json 
{
  "consent-telemetry": "yes",
  "cpus": 2,
  "preset": "openshift"
}
```

```powershell
PS> crc config set cpus 4
```

cleared the default:
```json
{
  "consent-telemetry": "yes",
  "preset": "openshift"
}
```
Note: the default preset remains (?)


```powershell
PS> crc config set preset microshift
```

changed the preset:
```json
{
  "consent-telemetry": "yes",
  "preset": "microshift"
}
```

```powershell
PS> crc config set cpus 1
```

```
Value '1' for configuration property 'cpus' is invalid, reason: requires CPUs >= 2
```

```powershell
PS> crc config set cpus 2
```

does not add the value (as matches default)


```powershell
PS> crc config set cpus 2
```

```json
{
  "consent-telemetry": "yes",
  "cpus": 3,
  "preset": "microshift"
}
```

Now changing to OpenShift again:
```powershell
PS> crc config set preset openshift
```

```json
{
  "consent-telemetry": "yes"
}
```

SO this looks good... though here the default preset was cleared. Earlier it didn't.
Feedback this happened would be appreciated, not high priority... we have #3636.

pkg/crc/config/config.go
```diff
@@ -95,6 +96,36 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
...
+	if setting.Name == Preset {
+		preset := preset.ParsePreset(value.(string))
+		mem := c.Get(Memory)
+		if err := validation.ValidateMemory(mem.AsInt(), preset); err != nil {
+			if _, err := c.Unset(Memory); err != nil {
+				return "", err
+			}
+		}
+		cpu := c.Get(CPUs)
+		if err := validation.ValidateCPUs(cpu.AsInt(), preset); err != nil {
```
gbraad, line 111: shouldn't we also use disk-size? This is now identical, but would suggest to create an issue to follow-up.

pkg/crc/config/viper_config_test.go
gbraad: 👍

gbraad (approved):
I agree to add the 'default values aren't stored' in the test

pkg/crc/config/config.go
praveenkumar: @anjannath Currently also we are not seeing the default configs as part of `config view`.

anjannath: right, do you prefer to keep it that way, or show a message saying `Configuration property '%s' restored to default value '%s'` (since you are in the file ;) )

anjannath (approved):

pkg/crc/config/config.go
praveenkumar: @anjannath better to keep it that way.

gbraad:
Found this reference to some interesting historical fact:
https://github.com/crc-org/crc/pull/2888/files#r770370496

pkg/crc/config/config.go
cfergeau: I also think we need to inform the user that we changed their configuration settings. I'll add that.

cfergeau:
Will need some follow-up work https://github.com/crc-org/crc/issues/3652

PR #3642, opened 2023-05-09
preflight: add check to see if admin-helper namedpipe is accessible

anjannath:
currently we only check if the admin-helper service is running, but it could happen that the service is running but current user is not in the 'crc-users' group, in this case directly trying to access the admin-helper api is better

will be helpful for https://github.com/crc-org/crc-extension/issues/93

### DIFF
--- a/pkg/crc/preflight/preflight_windows.go
+++ b/pkg/crc/preflight/preflight_windows.go
@@ -118,7 +118,7 @@ var adminHelperServiceCheks = []Check{
 		configKeySuffix:  "check-admin-helper-service-running",
 		checkDescription: "Checking admin helper service is running",
 		check:            checkIfAdminHelperServiceRunning,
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_windows.go
```diff
@@ -118,7 +118,7 @@ var adminHelperServiceCheks = []Check{
 		configKeySuffix:  "check-admin-helper-service-running",
 		checkDescription: "Checking admin helper service is running",
 		check:            checkIfAdminHelperServiceRunning,
-		fixDescription:   "Make sure you installed crc using the Windows installer",
+		fixDescription:   "Make sure you installed crc using the Windows installer and performed required reboot",
```
praveenkumar, line 121: 👍🏼

praveenkumar (approved):

PR #3649, opened 2023-05-11
e2e: response to recent config changes

jsliacan:
E2e config feature was failing on Linux and MacOS after recent changes where default values aren't written to config file -- this [commit](https://github.com/crc-org/crc/commit/f9a0a4a830a3fea39cffc2610efde4d21f6868bf) I assume.

### DIFF
--- a/test/e2e/features/config.feature
+++ b/test/e2e/features/config.feature
@@ -18,24 +18,24 @@ Feature: Test configuration settings
         @darwin
         Examples: Config settings on Mac
             | property         | value1    | value2            |
-            | cpus             | 4         | 3                 |
-            | memory           | 9216      | 4096              |
+            | cpus             | 5         | 3                 |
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -223,7 +223,7 @@ integration:
 
 .PHONY: e2e ## Run e2e tests
 e2e:
-GODOG_OPTS = --godog.tags=$(GOOS)
+GODOG_OPTS = --godog.tags="linux && config"
```
anjannath: this needs to be changed back i guess?

anjannath:
/hold

the bot will auto-merge it since all the CI are green

Makefile
jsliacan: Yes, thanks!

anjannath (approved):

Issue #3635, opened 2023-05-05
[BUG] OKD 4.12 does not start because of missing kubeadmin-password file


VGerris:
### General information

  * OS: Linux 
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes)?
  * Running CRC on: Laptop

## CRC version
```bash
crc version
CRC version: 2.18.0+4ea3a1
OpenShift version: 4.12.13
Podman version: 4.4.1
```
  
## CRC status
```bash
crc status --log-level debug
DEBU CRC version: 2.18.0+4ea3a1                   
DEBU OpenShift version: 4.12.13                   
DEBU Podman version: 4.4.1                        
DEBU Running 'crc status'                         
CRC VM:          Running
OpenShift:       Stopped (v4.12.0-0.okd-2023-02-18-033438)
RAM Usage:       5.292GB of 25.22GB
Disk Usage:      16.23GB of 32.68GB (Inside the CRC VM)
Cache Usage:     38.11GB
Cache Directory: /home/user/.crc/cache
```

## CRC config
```bash
crc status --log-level debug
DEBU CRC version: 2.18.0+4ea3a1                   
DEBU OpenShift version: 4.12.13                   
DEBU Podman version: 4.4.1                        
DEBU Running 'crc status'                         
CRC VM:          Running
OpenShift:       Stopped (v4.12.0-0.okd-2023-02-18-033438)
RAM Usage:       5.292GB of 25.22GB
Disk Usage:      16.23GB of 32.68GB (Inside the CRC VM)
Cache Usage:     38.11GB
Cache Directory: /home/user/.crc/cache
```

## Host Operating System
```bash
cat /etc/os-release
NAME="Fedora Linux"
VERSION="38 (Workstation Edition)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
[...]

PR #3657, opened 2023-05-12
CrcBundleInfo's IsOpenShift() now considers OKD to also be OpenShift

bergner:
There are numerous places in the CRC code base that calls IsOpenShift() to run logic that is not intended for Podman setups or other more limited setups. Using the "okd" preset gives you something that is very close to the "openshift" preset.

CrcBundleInfo's IsOpenShift() now treats the "okd" preset as if it was OpenShift. This means we can avoid littering the code with checks for both OpenShift and OKD in a bunch of places, which also reduces the risk of forgetting to check for OKD.

**Fixes:** Issue #3635 

## Solution/Idea

Treat the "okd" preset as equivalent to "openshift" in CrcBundleInfo. This means the various bundle.IsOpenShift() checks throughout the CRC code base will also apply to OKD.

## Testing

Before the fix `crc start` fails with errors described in #3635. Run `stop` and `delete` on previous CRC installation, then `setup` and `start` using the newly compiled CRC.

1. `crc stop`
2. `crc delete`
3.  `~/go/bin/crc setup`
4. `~/go/bin/crc start --log-level debug`
5. Takes quite a while to start but does so eventually
6. `~/go/bin/crc status` shows both CRC VM and OpenShift as "Running"
7. `~/go/bin/crc console --url` and `~/go/bin/crc console --credentials` show expected information

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -199,7 +199,8 @@ func (bundle *CrcBundleInfo) GetBundleType() crcPreset.Preset {
 }
 
 func (bundle *CrcBundleInfo) IsOpenShift() bool {
-	return bundle.GetBundleType() == crcPreset.OpenShift
+	preset := bundle.GetBundleType()
+	return preset == crcPreset.OpenShift || preset == crcPreset.OKD
 }
 
 func (bundle *CrcBundleInfo) IsMicroshift() bool {


pkg/crc/machine/bundle/metadata.go
```diff
@@ -199,7 +199,8 @@ func (bundle *CrcBundleInfo) GetBundleType() crcPreset.Preset {
 }
 
 func (bundle *CrcBundleInfo) IsOpenShift() bool {
```
cfergeau, line 201: This will have to be renamed to something else, or it's bound to cause confusion in the future (`IsPodman()` checks for the `podman` preset, `IsMicroShift()` checks for the `microshift` preset, `IsOpenShift()` checks for the `openshift` **and** the `okd` preset - not consistent).
This can be done as a follow-up PR though, as I can't think of a very good name (`IsOpenShiftLike()`? `IsOKDBased()`?)

cfergeau (approved):

cfergeau:
Thanks for the PR, this should fix most of the recent OKD issues!

pkg/crc/machine/bundle/metadata.go
bergner: Yes I definitely also felt the naming feels a bit odd here, but to me it is a slightly wider topic about what contexts the code  cares about what preset is used vs. what capabilities the CRC setup has. It would perhaps help if key capabilities were identified and you could call `SupportsSomeCapability()` rather than `IsPodman()`, `IsOpenShift()` etc. when implementing conditional logic that depends on the cluster having that capability.

cfergeau: We are definitely on the same line, I was suggesting the same in a private conversation yesterday ;) :
> it might help to move to some capability-based description of the bundle `NeedsPullSecret()`, `NeedsKubeadmin()`, `HasKubeconfig()`

Issue #3663, opened 2023-05-16
[BUG] `crc config set` regression after commit f9a0a4a83


cfergeau:
Before commit f9a0a4a83:
```
$ crc config set preset openshift
$ crc config set cpus 5
Changes to configuration property 'cpus' are only applied when the CRC instance is started.
If you already have a running CRC instance, then for this configuration change to take effect, stop the CRC instance with 'crc stop' and restart it with 'crc start'.
$ crc config set cpus 4
Changes to configuration property 'cpus' are only applied when the CRC instance is started.
If you already have a running CRC instance, then for this configuration change to take effect, stop the CRC instance with 'crc stop' and restart it with 'crc start'.
$
```

After commit f9a0a4a83, when setting `cpus` to its default value we no longer get the warning:
```
$ crc config set cpus 5
Changes to configuration property 'cpus' are only applied when the CRC instance is started.
If you already have a running CRC instance, then for this configuration change to take effect, stop the CRC instance with 'crc stop' and restart it with 'crc start'.
$ crc config set cpus 4
$
```

gbraad:
has the value been set (removed from the config/reset to default)?

cfergeau:
> has the value been set (removed from the config/reset to default)?

Yes, it's only the message which is missing:
```
$ crc config get cpus 
Configuration property 'cpus' is not set. Default value is '4'
```

jsliacan:
Not sure if this can go under the same issue (new default values behavior).

```
$ crc config set preset openshift
$ echo $?
0
$ crc config get preset
Configuration property 'preset' is not set. Default value is 'openshift'
$ echo $?
1
```
I'm doing some checks in tests and exitcode 1 is not what I would expect given that I just successfully set a preset.

cfergeau:
> Not sure if this can go under the same issue (new default values behavior).
> 
> ```
> $ crc config set preset openshift
> $ echo $?
> 0
> $ crc config get preset
> Configuration property 'preset' is not set. Default value is 'openshift'
> $ echo $?
> 1
> ```
> 
> I'm doing some checks in tests and exitcode 1 is not what I would expect given that I just successfully set a preset.

This specific issue is caused by https://github.com/crc-org/crc/commit/f9a0a4a830a3fea39cffc2610efde4d21f6868bf , but the root cause is what is being discussed here.

anjannath:
(reopened to continue the conversation around exit codes for `crc config` command)

praveenkumar:
[...]

Issue #3652, opened 2023-05-11
cpu/memory configuration not reset after `crc config unset preset`


cfergeau:
We recently added code to reset cpu/memory after a preset change, but at the moment this is only done for `crc config set preset ....`
These steps are not working as expected:

```
    crc config set preset podman
    crc config set cpus 3
    crc config unset preset
    crc config get cpus
```

This returns 3 instead of the expected 4 (default preset is openshift with a minimum requirement of 4 cpus).
This works as expected if I use `crc config set preset okd` instead of `crc config unset preset`.

I've started fixing this in https://github.com/cfergeau/crc/tree/config but this needs more testing (next week hopefully).

Issue #3637, opened 2023-05-08
[RFC] Way to switch the default preset at build time


anjannath:
This'll allow us to make a release with the default preset set to `okd`

praveenkumar:
This is something for upstream releases?

anjannath:
> This is something for upstream releases?

yeah, releases on choco, brew and on github (someday)

anjannath:
done in https://github.com/crc-org/crc/commit/ab9136012d4ebe8bd2f9c9d55edcde0bab59dd46

PR #3665, opened 2023-05-17
config: Follow-ups to code resetting cpu/memory after preset change

cfergeau:
The main goal of this patch series is to fix 2 bugs related to https://github.com/crc-org/crc/pull/3636 :
- cpu/memory are not checked when the preset is unset, only when they are set 
- the messages instructing the user to restart the VM are no longer printed if cpu/memory are set to their default value

This starts by refactoring the code a bit, and by adding a new RegisterNotifier API, and then this fixes the bugs.
This also adds unit tests for this new code/these bug fixes.

**Fixes:** Issue #3652, Issue #3663

**Relates to:** Issue #3637

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -5,6 +5,7 @@ import (
 	"runtime"
 
 	"github.com/crc-org/crc/pkg/crc/constants"
+	"github.com/crc-org/crc/pkg/crc/logging"
 	"github.com/crc-org/crc/pkg/crc/network"
 	"github.com/crc-org/crc/pkg/crc/preset"
 	"github.com/crc-org/crc/pkg/crc/version"
@@ -130,6 +131,22 @@ func RegisterSettings(cfg *Config) {
 		fmt.Sprintf("HTTP port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPPort))
 	cfg.AddSetting(IngressHTTPSPort, constants.OpenShiftIngressHTTPSPort, ValidatePort, RequiresHTTPSPortChangeWarning,
 		fmt.Sprintf("HTTPS port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPSPort))
+
+	if err := cfg.RegisterNotifier(Preset, presetChanged); err != nil {
+		logging.Debugf("Failed to register notifier for Preset: %v", err)
+	}
+}
+
+func presetChanged(cfg *Config, _ string, _ interface{}) {
+	// The `memory` and `cpus` values are preset-dependent.
+	// When they are lower than the preset requirements, this code
+	// automatically resets them to their default value.
+	if err := revalidateSettingsValue(cfg, Memory); err != nil {
+		logging.Debugf("error validating Memory config value: %v", err)
+	}
+	if err := revalidateSettingsValue(cfg, CPUs); err != nil {
+		logging.Debugf("error validating CPUs config value: %v", err)
+	}
 }
 
 func defaultCPUs(cfg Storage) int {
@@ -160,6 +177,18 @@ func GetNetworkMode(config Storage) network.Mode {
 		return network.UserNetworkingMode
 	}
 	return network.ParseMode(config.Get(NetworkMode).AsString())
+}
+
+func revalidateSettingsValue(cfg *Config, key string) error {
+	if err := cfg.validate(key, cfg.Get(key).Value); err != nil {
+		logging.Debugf("'%s' value is invalid: %v", key, err)
+		if _, err := cfg.Unset(key); err != nil {
+			return err
+		}
+		logging.Warnf("Value for config setting '%s' is invalid, resetting it to its default value", key)
+	}
+
+	return nil
 }
 
 func UpdateDefaults(cfg *Config) {
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -138,15 +138,7 @@ func RegisterSettings(cfg *Config) {
 }
 
 func presetChanged(cfg *Config, _ string, _ interface{}) {
-	// The `memory` and `cpus` values are preset-dependent.
-	// When they are lower than the preset requirements, this code
-	// automatically resets them to their default value.
-	if err := revalidateSettingsValue(cfg, Memory); err != nil {
-		logging.Debugf("error validating Memory config value: %v", err)
-	}
-	if err := revalidateSettingsValue(cfg, CPUs); err != nil {
-		logging.Debugf("error validating CPUs config value: %v", err)
-	}
+	UpdateDefaults(cfg)
 }
 
 func defaultCPUs(cfg Storage) int {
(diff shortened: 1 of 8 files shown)


pkg/crc/config/settings_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}, cfg.Get(CPUs))
+}
+
+func TestSetPreset(t *testing.T) {
+	cfg, err := newInMemoryConfig()
+	require.NoError(t, err)
+
+	_, err = cfg.Set(Preset, preset.Podman)
+	require.NoError(t, err)
+	_, err = cfg.Set(Memory, 10000)
```
cfergeau, line 80: CI/tests are failing because of this, when we set memory, the memory of the system is checked, and the runners only have 8GB. Hard to workaround in the tests as the minimum memory amount for OpenShift is 9GB. I guess I'll remove this part of the test, it's not the key part.
This will have to wait until next week though.

```
=== RUN   TestSetPreset
    settings_test.go:63: 
        	Error Trace:	/root/rpmbuild/BUILD/crc-2.19.0-4.12.13/pkg/crc/config/settings_test.go:63
        	Error:      	Received unexpected error:
        	            	Value '10000' for configuration property 'memory' is invalid, reason: only 7.281GB of memory found (10.49GB required)
        	Test:       	TestSetPreset
--- FAIL: TestSetPreset (0.00s)
```

praveenkumar: Changing the cpu/memory to default we should use `podman` and `microshift` preset instead using `openshift` which fails because of high memory requirement.

gbraad:
I am confused about the naming `notifier` and the actual use. The commit says:

```
and to to some higher level actions in response to that.
```

so this is not to 'notify' the user, but an internal action. These concern ChangedValue functions, so can this be ChangeNotifier or similar... or as ChangedValueHandler? Also `notify` could be `notifyOfChange` or `handleValueChange`.

pkg/crc/config/settings_test.go
gbraad: I agree. Public CI's should only concern with the smaller presets we offer.

cfergeau:
> I am confused about the naming `notifier` and the actual use. The commit says:
> 
> ```
> and to to some higher level actions in response to that.
> ```
> 
> so this is not to 'notify' the user, but an internal action. These concern ChangedValue functions, so can this be ChangeNotifier or similar... or as ChangedValueHandler? Also `notify` could be `notifyOfChange` or `handleValueChange`.

I'll rename to ChangeNotifier.

pkg/crc/config/settings_test.go
cfergeau: Using microshift will paper over this problem for now, but it will break in the same way the day we get a worker with less than 4G, or the day we raise the microshift requirements.

cfergeau: I started changing the test to use the microshift preset instead of openshift, but this does not allow me to test `crc config unset preset`, which is one of the bugs I'm fixing. I'll try to handle this differently.

pkg/crc/config/settings.go
```diff
@@ -130,6 +131,20 @@ func RegisterSettings(cfg *Config) {
 		fmt.Sprintf("HTTP port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPPort))
 	cfg.AddSetting(IngressHTTPSPort, constants.OpenShiftIngressHTTPSPort, ValidatePort, RequiresHTTPSPortChangeWarning,
 		fmt.Sprintf("HTTPS port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPSPort))
+
+	_ = cfg.RegisterNotifier(Preset, presetChanged)
```
anjannath: nit: did you consider logging the error in `cfg.RegisterNotifier()` instead of returning it, might lose some testability but we can get rid of the _ignore return value_

pkg/crc/config/settings_test.go
```diff
@@ -0,0 +1,163 @@
...
+		Value:     3,
+		Invalid:   false,
+		IsDefault: false,
+		IsSecret:  false,
+	}, cfg.Get(CPUs))
+
+	// Changing the preset to 'openshift' should reset the CPUs config value
+	logging.Warnf("Setting preset to OpenShift")
+	_, err = cfg.Set(Preset, crcpreset.OpenShift)
+	logging.Warnf("Preset set to OpenShift")
```
anjannath: are these two `logging.Warnf` left over from testing?

pkg/crc/config/viper_config_test.go
```diff
@@ -327,3 +327,27 @@ func TestNotifier(t *testing.T) {
...
+
+	callback, err := config.Set(cpus, 5)
+	require.NoError(t, err)
+	require.NotEmpty(t, callback)
+
+	callback, err = config.Unset(cpus)
+	require.NoError(t, err)
+	require.NotEmpty(t, callback)
+
+	callback, err = config.Set(cpus, 4)
```
anjannath, line 346: can also directly use the `constants.GetDefaultCPUs()` like the unit tests in the previous commit

pkg/crc/config/settings.go
```diff
@@ -46,16 +46,16 @@ func RegisterSettings(cfg *Config) {
 		return ValidateBool(value)
 	}
 
-	validateCPUs := func(value interface{}) (bool, string) {
-		return ValidateCPUs(value, GetPreset(cfg))
+	validCPUs := func(value interface{}) (bool, string) {
```
anjannath, line 49: should be `validateCPUs` ?

anjannath (commented):
like the addition of the change notifiers 👍🏽 just some small questions and comments.

also some of the commit msgs have small typos, listing the commits:

- 71c92b1cb7a016f7fa9d48af237d213a67df1c6f — s/is/in
- 574a883e240336518acf290ec1669a9ada12a130 — s/to/do
- 2120cdcb222755c7956dbbceb5c455f88ceca445 — s/save/except

cfergeau:
> also some of the commit msgs have small typos, listing the commits:
> [...]
>     * [2120cdc](https://github.com/crc-org/crc/commit/2120cdcb222755c7956dbbceb5c455f88ceca445) — s/save/except

"save for" is not a typo, see https://en.wiktionary.org/wiki/save_for
I've changed it to "except for" as you suggested, this will make the log easier to read

pkg/crc/config/settings_test.go
cfergeau: Yes, that's likely, I removed them, thanks!

pkg/crc/config/viper_config_test.go
cfergeau: The rest of this file uses 4, 5 directly, I prefer to keep the 4 here as well.

pkg/crc/config/settings.go
cfergeau: this can't be `validateCPUs` as the name is already used by the function in validations.go. It's the same for `validMemory`/.. The `validCPUs` name is not great, open to suggestions! Maybe rename `validateCPUs` to `validateCPUsForPreset` ?

pkg/crc/config/settings.go
cfergeau: hmm `config.go` does not log anything at the moment, I prefer to keep it this way (even though some low level debug logs there could be useful).
I'll log the error in `settings.go` though ;)

pkg/crc/config/settings.go
anjannath: eh, no need, `validCPUs` is also okay :)

anjannath (approved):

anjannath:
proxy tests are failing in `e2e` are unrelated to the changes in the PR

PR #3666, opened 2023-05-18
Microshift: add disk resize functionality

praveenkumar:
The way this PR works is in following steps
- It grows the disk size usign growpart command
- Resize the physical volume
- Make sure that volume group where `/sysroot` mounted and free space which is going to used by LVMS for persistant volume (pv) creation use equal disk size during the resizing because after `pvresize` all the space goes as free space.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -15,7 +15,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/cluster"
 	"github.com/crc-org/crc/pkg/crc/constants"
 	crcerrors "github.com/crc-org/crc/pkg/crc/errors"
-	"github.com/crc-org/crc/pkg/crc/logging"
+	logging "github.com/crc-org/crc/pkg/crc/logging"
 	"github.com/crc-org/crc/pkg/crc/machine/bundle"
 	"github.com/crc-org/crc/pkg/crc/machine/config"
 	"github.com/crc-org/crc/pkg/crc/machine/state"
@@ -117,26 +117,13 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error
 	}
 	// With 4.7, this is quite a manual process until https://github.com/openshift/installer/pull/4746 gets fixed
 	// See https://github.com/crc-org/crc/issues/2104 for details
-	rootPart, _, err := sshRunner.Run("realpath", "/dev/disk/by-label/root")
+	rootPart, err := getrootPartition(sshRunner, "/dev/disk/by-label/root")
 	if err != nil {
 		return err
 	}
-	rootPart = strings.TrimSpace(rootPart)
-	if !strings.HasPrefix(rootPart, "/dev/vda") && !strings.HasPrefix(rootPart, "/dev/sda") {
-		return fmt.Errorf("Unexpected root device: %s", rootPart)
-	}
-	// with '/dev/[sv]da4' as input, run 'growpart /dev/[sv]da 4'
-	if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Growing %s partition", rootPart), "/usr/bin/growpart", rootPart[:len("/dev/.da")], rootPart[len(rootPart)-1:]); err != nil {
-		var exitErr *ssh.ExitError
-		if !errors.As(err, &exitErr) {
-			return err
-		}
-		if exitErr.ExitStatus() != 1 {
-			return err
-		}
 
-		logging.Debugf("No free space after %s, nothing to do", rootPart)
-		return nil
+	if err := runGrowpart(sshRunner, rootPart); err != nil {
+		return err
 	}
 
 	logging.Infof("Resizing %s filesystem", rootPart)
@@ -147,6 +134,33 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error
 		return err
 	}
 
+	return nil
+}
+
+func getrootPartition(sshRunner *crcssh.Runner, label string) (string, error) {
+	rootPart, _, err := sshRunner.Run("realpath", label)
+	if err != nil {
+		return "", err
+	}
+	rootPart = strings.TrimSpace(rootPart)
+	if !strings.HasPrefix(rootPart, "/dev/vda") && !strings.HasPrefix(rootPart, "/dev/sda") {
+		return "", fmt.Errorf("Unexpected root device: %s", rootPart)
+	}
+	return rootPart, nil
+}
+
+func runGrowpart(sshRunner *crcssh.Runner, rootPart string) error {
+	// with '/dev/[sv]da4' as input, run 'growpart /dev/[sv]da 4'
+	if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Growing %s partition", rootPart), "/usr/bin/growpart", rootPart[:len("/dev/.da")], rootPart[len(rootPart)-1:]); err != nil {
+		var exitErr *ssh.ExitError
+		if !errors.As(err, &exitErr) {
+			return err
+		}
+		if exitErr.ExitStatus() != 1 {
+			return err
+		}
+		logging.Debugf("No free space after %s, nothing to do", rootPart)
+	}
 	return nil
 }
 
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -112,8 +112,7 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
 
 func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error {
 	if preset == crcPreset.Microshift {
-		logging.Debugf("growRootFileSystem does not support LVM which is used by %s images", preset)
-		return nil
+		return growLVForMicroshift(sshRunner)
 	}
 	// With 4.7, this is quite a manual process until https://github.com/openshift/installer/pull/4746 gets fixed
 	// See https://github.com/crc-org/crc/issues/2104 for details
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
 	// with '/dev/[sv]da4' as input, run 'growpart /dev/[sv]da 4'
-	if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Growing %s partition", rootPart), "/usr/bin/growpart", rootPart[:len("/dev/.da")], rootPart[len(rootPart)-1:]); err != nil {
-		var exitErr *ssh.ExitError
-		if !errors.As(err, &exitErr) {
-			return err
-		}
-		if exitErr.ExitStatus() != 1 {
-			return err
-		}
+	_, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Growing %s partition", rootPart), "/usr/bin/growpart", rootPart[:len("/dev/.da")], rootPart[len(rootPart)-1:])
```
cfergeau: Can you split this in 2 commits, first the introduction of `getRootPartition` and `runGrowpart`, and then the addition of microshift disk resizing? The refactoring introduces a minor regression which is very hard to catch as the diff is hard to read here.
I've done this in https://github.com/cfergeau/crc/commits/issue_3624

pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	if !errors.As(err, &exitErr) {
+		return err
+	}
+	if exitErr.ExitStatus() != 1 {
+		return err
+	}
 
-		logging.Debugf("No free space after %s, nothing to do", rootPart)
-		return nil
+	logging.Debugf("No free space after %s, nothing to do", rootPart)
```
cfergeau: This was only printed when `sshRunner.RunPrivileged("growpart")` returned an error, and if the exit code was 1. After your changes it's also printed when `err` is nil.

pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	return nil
+}
+
+func growLVForMicroshift(sshRunner *crcssh.Runner) error {
+	rootPart, err := getrootPartition(sshRunner, "/dev/disk/by-id/lvm-pv-*")
```
cfergeau: I'd have a strong preference for, for example, getting the device to use using `sudo pvs -S vg_name=rhel -o pv_name --noheadings` or some similar command.
Alternatively, you can change https://github.com/openshift/microshift/blob/bac892b5586edff5687b399475f8ad0e8ec01994/scripts/image-builder/config/kickstart.ks.template#L28 to set a well-known label (see [the --label option to part](https://docs.fedoraproject.org/en-US/fedora/f36/install-guide/appendixes/Kickstart_Syntax_Reference/#sect-kickstart-commands-part)). Then you can use this label to `getRootPartition`.

Passing a wildcard to `getRootPartition` only works because we get the root partition through SSH, which I'd rather not leak outside of `getRootPartition`. And there's nothing in the code erroring out if we are unlucky and there are multiple matches for lvm-pv-*.

pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
-	logging.Infof("Resizing %s filesystem", rootPart)
-	if _, _, err := sshRunner.RunPrivileged("Remounting /sysroot read/write", "mount -o remount,rw /sysroot"); err != nil {
+	if _, _, err := sshRunner.RunPrivileged("Resizing the physical volume(PV)", "/usr/sbin/pvresize", rootPart); err != nil {
 		return err
 	}
-	if _, _, err = sshRunner.RunPrivileged(fmt.Sprintf("Growing %s filesystem", rootPart), "xfs_growfs", rootPart); err != nil {
+
+	// Get the size of volume group
+	// $ sudo vgs --noheadings --nosuffix --units g -o vg_size
+	//   39.02
```
cfergeau: I'd remove these last 2 lines of comments.

pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	// $ sudo vgs --noheadings --nosuffix --units g -o vg_free
+	//   7.51
+	freeVG, _, err := sshRunner.RunPrivileged("Get the free space for volume group", "/usr/sbin/vgs", "--noheadings", "--nosuffix", "--units", "g", "-o", "vg_free")
+	if err != nil {
+		return err
+	}
+	vgFree, err := strconv.ParseFloat(strings.TrimSpace(freeVG), 64)
+	if err != nil {
+		return err
+	}
```
cfergeau, line 198: Please add a helper, `func getVGSize(kind string) (float, error)` which will be called with `vg_size` and `vg_free`.
Might be better to work in bytes directly?

pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+
+	// We want to expand the `/dev/rhel/root` vg in case the free size of volume group is greater than half of total size
```
cfergeau: My understanding is that `/dev/rhel` refers to a VG, while `/dev/rhel/root` is a LV. I don't know which of the 2 you want to refer to when you say `the '/dev/rhel/root' VG`

pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	// since that vg contain the `/sysroot` so we need to extend it to accommodate more container images. Free size is
+	// used by logical volume manager storage (LVMS) Container Storage Interface (CSI) provider for storing PVs.
```
cfergeau: PVs means (k8s) Persistent Volume in this comment I think? While in the rest of the function it is a LVM Physical Volume.
Maybe rewrite the last sentence as `The LVM Storage Operator will use the free space for persistent local storage` or something like this. I used https://docs.openshift.com/container-platform/4.14/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html for inspiration.

pkg/crc/machine/start.go
```diff
@@ -112,41 +112,104 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	sizeDiff := vgFree - vgSize/2
```
cfergeau: Having a hard time to wrap my head around this, might be easier to read as (which should be approximately correct?)
```
  if vgFree > vgSize/2 {
    // run lvextend --resizefs --size $vgSize/2
  }
```

cfergeau (commented):
Some comments, but overall logic should be fine.
Not fully sure about the UI though, what the size the user specified should refer to? The size of the partitions where they store their images/... ? Or the size of theoverall disk? 
The former is only resized by half of what the user specified if I understood correctly.

praveenkumar:
> Not fully sure about the UI though, what the size the user specified should refer to? The size of the partitions where they store their images/... ? Or the size of theoverall disk?
The former is only resized by half of what the user specified if I understood correctly.

@cfergeau it should be the overall disk same like for openshift preset, whatever user provide, half will be added to `/dev/rhel/root` logical volume (LV) which have `/sysroot` mounted and store images and another half would be part of free Volume group (VG) which is used by persistent volume of kubernetes side.

pkg/crc/machine/start.go
praveenkumar: Ah yes it is LV which need to be extended.

cfergeau:
> > Not fully sure about the UI though, what the size the user specified should refer to? The size of the partitions where they store their images/... ? Or the size of theoverall disk?
> > The former is only resized by half of what the user specified if I understood correctly.
> 
> @cfergeau it should be the overall disk same like for openshift preset, whatever user provide, half will be added to `/dev/rhel/root` logical volume (LV) which have `/sysroot` mounted and store images and another half would be part of free Volume group (VG) which is used by persistent volume of kubernetes side.

What I'm asking is if it makes sense from a user perspective. Setting disk size, ssh'ing to the VM, checking df -h will only show half of that size. This is unexpected and different from the openshift preset. Maybe we should change the LV to the size the user asked for?

praveenkumar:
>  Maybe we should the LV to the size the user asked for?

@cfergeau then what if user want to extend the disk size because there is no free space on VG for persistent volume for the workload? Right now I am putting half for each so that user doesn't need to worry about in both the cases otherwise the correct solution would be add config option which user provide if he wants LV to increase or just want free size for it's persistent volume workload.

cfergeau:
> > Maybe we should the LV to the size the user asked for?
> 
> @cfergeau then what if user want to extend the disk size because there is no free space on VG for persistent volume for the workload? Right now I am putting half for each so that user doesn't need to worry about in both the cases otherwise the correct solution would be add config option which user provide if he wants LV to increase or just want free size for it's persistent volume workload.

This is not answering my initial question. If I run `crc config set disk-size xx` what would users expect to happen? Once we've answered this question, we can discuss some more the implications with the free space needed by the CSI driver.

praveenkumar:
/hold

we need to understand what user scenario we want to implement https://github.com/crc-org/crc/issues/3624#issuecomment-1560892999

praveenkumar:
/unhold

implemented as per https://github.com/crc-org/crc/issues/3624#issuecomment-1562708790 one.

pkg/crc/machine/start.go
```diff
@@ -134,19 +158,55 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error
...
+	if err != nil {
+		return err
+	}
+	lvSize, err := strconv.Atoi(strings.TrimSpace(sizeLV))
+	if err != nil {
 		return err
 	}
 
+	// vgFree space as part of the bundle is default to ~15.02G (16127098880 byte)
+	vgFree := 16127098880
```
gbraad, line 201: magic value

gbraad (approved):

PR #3681, opened 2023-05-29
e2e: Add minimal user story

jsliacan:
On environments with very limited resources (e.g. current arrangement for M1), standard e2e and integration don't run as expected. This provides a compact minimal scenario to use in those cases.

For each preset (`podman`, `microshift`, `openshift`), do a basic user cycle:
1. Setup
2. Start
3. Check status

**Note:**
This should probably not be included in CIs as it only does a strict subset of actions contained in other tests, but increases execution time.

### DIFF
--- a/test/e2e/features/minimal.feature
+++ b/test/e2e/features/minimal.feature
@@ -1,0 +1,18 @@
+@minimal @darwin @linux @windows
+Feature: Minimal user story
+
+    User starts CRC cluster and checks status.
+
+    @cleanup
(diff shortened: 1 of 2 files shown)


jsliacan:
I've sent [a change](https://github.com/openshift/release/pull/40117) to exclude this Feature to be run in the PR checks. 

Here's a run from M1 with release bits. 

Screenshot from 2023-05-30 16-44-14

Issue #3677, opened 2023-05-23
[Feature] Debug option to add a password for the core user on startup


gbraad:
This allows you to debug the environment in case of an OOM or other crash.
At the moment we only allow the use of the ssh-key to log in, but when the image is memory constrained or otherwise affect by a crash, the `sshd` process becomes unreachable. This prevents further debugging. I suggest adding a

```
$ crc config set emergency-login true    # or debug-login
```

which generates a random password on startup and allows you to login in case of an issue.

anjannath:
Done in https://github.com/crc-org/crc/pull/3755

PR #3689, opened 2023-05-31
[WIP] Set emergency login for core user

gbraad:
**Fixes:** #3677

## Solution/Idea
Performs a `passwd` for the `core` user as soon as the VM is reachable, comparable to running the command: 

```
PS> ssh -i c:\users\gbraad\.crc\machines\crc\id_ecdsa core@localhost -p 2222 -C "echo rescue | sudo passwd core --stdin"  
Changing password for user core.
```

The password is fixed to `rescue`.

## Proposed changes

1. `crc config set enable-emergency-login true`
2. `crc start`
3. open the console and login using: core, rescue

image

### DIFF
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -78,6 +78,8 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		IngressHTTPPort:   config.Get(crcConfig.IngressHTTPPort).AsUInt(),
 		IngressHTTPSPort:  config.Get(crcConfig.IngressHTTPSPort).AsUInt(),
 		EnableSharedDirs:  config.Get(crcConfig.EnableSharedDirs).AsBool(),
+
+		EmergencyLogin: config.Get(crcConfig.EmergencyLogin).AsBool(),
 	}
 
 	client := newMachine()
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -34,6 +34,7 @@ const (
 	SharedDirPassword       = "shared-dir-password" // #nosec G101
 	IngressHTTPPort         = "ingress-http-port"
 	IngressHTTPSPort        = "ingress-https-port"
+	EmergencyLogin          = "enable-emergency-login"
 )
 
 func RegisterSettings(cfg *Config) {
@@ -88,6 +89,8 @@ func RegisterSettings(cfg *Config) {
 		"Disable update check (true/false, default: false)")
 	cfg.AddSetting(ExperimentalFeatures, false, ValidateBool, SuccessfullyApplied,
 		"Enable experimental features (true/false, default: false)")
+	cfg.AddSetting(EmergencyLogin, false, ValidateBool, SuccessfullyApplied,
+		"Enable emergency login for 'core' user. Password set to 'rescue' (true/false, default: false)")
 
 	// Shared directories configs
 	if runtime.GOOS == "windows" {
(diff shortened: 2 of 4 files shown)


gbraad:
WIP => ~unable to perform cross-builds on Windows. Unable to perform `gofmt` on Windows without modifying almost all the files. Will perform these steps on another machine ASAP.~ For now, testing/debugging the core dumps.

cfergeau:
> The password is fixed to `rescue`.

A non-fixed password is needed in my opinion, especially as this is a persistent option that you can forget to unset after setting it once.

.gitpod.yml
```diff
@@ -0,0 +1,11 @@
+# This configuration file was automatically generated by Gitpod.
```
cfergeau: If you want to add this file, please do this in a separate commit.

gbraad: oops, wasn't supposed to happen. did the `gofmt` on a remote machine

gbraad:
> A non-fixed password is needed in my opinion

I understand, such as a random generated one. However, this is also a clear choice of the user, so I actually do not think this is such a necessity.

> this is a persistent option

Rather to have this as a flag for `crc start` ? This has crossed my mind

.gitpod.yml
gbraad: Removed

cfergeau:
> > A non-fixed password is needed in my opinion
> 
> I understand, such as a random generated one. However, this is also a clear choice of the user, so I actually do not think this is such a necessity.
> 

It's a clear choice the user does once, but which then persists forever (I assume most users will forget to unset it). It's not automatically cleared when the VM is deleted, or after a short while. I don't think having a random generated password is going to be a lot more complicated ? (can be 8-10 ascii chars/numbers to make it relatively easy to type)

> > this is a persistent option
> 
> Rather to have this as a flag for `crc start` ? This has crossed my mind

This would ensure it's not persistent, but might be more complicated to add to the codebase.

Another alternative, the config is not a boolean, but the password the user wants to set on the instance? This would ensure the password is not shared/known between all crc deployments. There could even be a magic value (`<<RANDOM>>`) to get a random one to be generated? (this might be going too far ;)

gbraad:
> I don't think having a random generated password is going to be a lot more complicated ? (can be 8-10 ascii chars/numbers to make it relatively easy to type)

More how to handle this. Sure, we can dump it to the stdout... but what if the user does not read this? We store it somewhere? in that case it does not do much better for security

> but the password the user wants to set on the instance

I considered this. It would solve the issue of 'storing' the value, as the user would already be aware of this. Though it still persists and is no safer than storing a file.

cfergeau:
> 

The password does not need to be safer than the ssh key to access the VM, which is stored in a file in ~/.crc/machines/crc. The main thing I want to avoid is all the VMs of all our users sharing the same password (if they enabled the option).

gbraad:
I can change the `bool` to `string` and use it when set, otherwise ignore.

The ssh host is only reachable from localhost anyway, so making it even more secure by obfuscation (randomization) or whatever is not going to help/improve this much more.

For me it is more about messaging. So I would prefer to print this to the command line on start. WDYT?

gbraad:
@crc-org/crc-team do we still want this?

praveenkumar:
> https://github.com/orgs/crc-org/teams/crc-team do we still want this?

@gbraad yes, if you want, I can take over and start making changes as per suggestion

gbraad:
I am OK if you wanna continue on this. The need for this on Windows became unnecessary, however it is still relevant overall.

praveenkumar:
It is superseed by #3755

praveenkumar:
/close

In favor of #3755

PR #3693, opened 2023-06-01
Add helper download.InMemory for downloading files without writing to disk

### LINKED ISSUE #3686: CRC Setup is failing with "unable to get verified hash for default bundle"
Trying to setup a fresh CRC "crc setup" with the below environment but its failing with the mentioned error below

**Error** 
    "unable to get verified hash for default bundle: Get "https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/bundles/openshift/4.12.13/sha256sum.txt.sig": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"

**CR Version** : crc-linux-2.19.0-amd64
**Operating System** : CentOS-Stream-9-x86_64
**Over VMware** 16.1.0 build-17198959

Logs are attached "Setup Debug Logs.txt"

([Setup Debug Logs.txt](https://github.com/crc-org/crc/files/11597823/Setup.Debug.Logs.txt))

anjannath:
this helper is put in a separate package pkg/download/inmemory to avoid import cycles which occurs due to the network package being imported in the pkg/download package, the cycle is network->ssh->constants->version as a result we'll not be able to import pkg/download in pkg/crc/version

currently the helper is used for downloading release-info.json file and the signed bundle hashes file sha256sum.txt.sig

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -18,7 +18,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/gpg"
 	"github.com/crc-org/crc/pkg/crc/image"
 	"github.com/crc-org/crc/pkg/crc/logging"
-	"github.com/crc-org/crc/pkg/crc/network"
+	"github.com/crc-org/crc/pkg/crc/network/httpproxy"
 	crcPreset "github.com/crc-org/crc/pkg/crc/preset"
 	"github.com/crc-org/crc/pkg/download"
 )
@@ -301,7 +301,7 @@ func getBundleDownloadInfo(preset crcPreset.Preset) (*download.RemoteFile, error
 func getDefaultBundleVerifiedHash(preset crcPreset.Preset) (string, error) {
 	client := &http.Client{
 		Timeout:   5 * time.Second,
-		Transport: network.HTTPTransport(),
+		Transport: httpproxy.HTTPTransport(),
 	}
 	res, err := client.Get(constants.GetDefaultBundleSignedHashURL(preset))
 	if err != nil {
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -20,6 +19,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/logging"
 	"github.com/crc-org/crc/pkg/crc/network/httpproxy"
 	crcPreset "github.com/crc-org/crc/pkg/crc/preset"
+	"github.com/crc-org/crc/pkg/crc/version"
 	"github.com/crc-org/crc/pkg/download"
 )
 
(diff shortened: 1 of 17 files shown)


cfergeau:
> this helper is put in a separate package pkg/download/inmemory to avoid import cycles which occurs due to the network package being imported in the pkg/download packag

This import cycle already caused issues in the past, I think another way of breaking the cycle is by moving the proxy code (`HTTPTransport()`) to its own package

pkg/crc/machine/bundle/metadata.go
```diff
@@ -299,19 +299,15 @@ func getBundleDownloadInfo(preset crcPreset.Preset) (*download.RemoteFile, error
...
 	if err != nil {
 		return "", err
 	}
-	signedHashes, err := io.ReadAll(res.Body)
+	signedHashes, err := io.ReadAll(res)
 	if err != nil {
 		return "", err
 	}
-	if err := res.Body.Close(); err != nil {
+	if err := res.Close(); err != nil {
```
cfergeau: Should be a `defer res.Close()` just before `signedHashes, err := io.ReadAll(res)`, otherwise you won't close the file when `ReadAll` fails.

pkg/download/inmemory/inmemory.go
```diff
@@ -0,0 +1,45 @@
...
+	"io"
+	"net/http"
+
+	"github.com/cavaliergopher/grab/v3"
+)
+
+// Download takes an endpoint in the form of a string (URL) or a *http.Request
+// allowing caller to pass a custom http request to the downloader library, it
+// takes a http.RoundTripper as the second argument to be used by the client
+func Download[T string | *http.Request](endpoint T, transport http.RoundTripper) (io.ReadCloser, error) {
```
cfergeau: You can avoid the use of the generics by always setting the user-agent in this method, which makes sense in my opinion.

gbraad:
> I think another way of breaking the cycle is by moving the proxy code (HTTPTransport()) to its own package

is this the alternative you prefer or is this something that can ALSO be done.

pkg/download/inmemory/inmemory.go
anjannath: i think in zabbix we ignore requests with http agent `crc *` from the download data count, with the current use of this (for getting release info and bundle hashes) it should be okay, will remove the type parameters and set the user-agent in the function itself

anjannath: more import cycles, to set the user agent to `crc/<crc_version>`we have to import the `version` package in `inmemory` package and the version package has to import the `inmemory` package to download the `release-info.json`

cfergeau: We still want to be consistent imo and always use the same user agent. I guess for now you could add a `userAgent string` parameter to `Download`, and a `version.UserAgent()`, and pass it explicitly. Otherwise this needs more code movement.

cfergeau:
> > I think another way of breaking the cycle is by moving the proxy code (HTTPTransport()) to its own package
> 
> is this the alternative you prefer or is this something that can ALSO be done.

That's merely an alternative suggestion which can be ignored or done, no strong feeling. However, it's at least the 3rd time we get import cycles because of `HTTPTransport` being in the `network` package (happened to me in some experiments, quite sure Praveen saw that in one of his PRs, but avoided the cycle differently, now Anjan hits it too). So we could try to solve this once and for all.

anjannath:
i haven't looked at solving this in a better way so not sure how much work it'll be, created https://github.com/crc-org/crc/issues/3697 for planning/tracking

pkg/download/inmemory/inmemory.go
gbraad: does it functionally change anything? what are you planning to do with the 'useragent'; track it as telemetry?

cfergeau: Downloading the crc version file is done with user agent set, downloading the signature file is done without a user agent set, and there are 2 code paths in the download code to deal with both situations.
I don't mind dropping the user agent or always setting it, but it would be better to always do the same thing.

See also https://github.com/crc-org/crc/pull/3693#discussion_r1213949744

anjannath: When we request the `release-info.json` file from the CRC codebase to check if a new version exists, we set the `User-Agent` for that call as, e.g `crc 2.20.0` this was added to make sure that we are not skewing the download counts by our own tool making http requests to the download location.

and in zabbix where this download data is collected filters the requests having user agent `crc *` so these are not added to the download count

anjannath: `User-Agent` seems is not used to filter out http requests from `crc`, not sure how i reached that, but it doesn't make sense also as for collecting the download data from splunk we can use a query language and can filter out the `release-info.json` or the `sha256sum.txt.sig` files.

cfergeau (commented):
Looks good to me, there are several mentions of `inmemory.Download` in the commit logs which no longer are accurate.
You should also mention in the last commit log that this should help with bug xxx (the one with the download timeouts).

anjannath:
> Looks good to me, there are several mentions of `inmemory.Download` in the commit logs which no longer are accurate.
> You should also mention in the last commit log that this should help with bug xxx (the one with the download timeouts).

Thanks, updated the commit logs!

anjannath:
Changed the function name to `download.InMemory`, as otherwise, gorevive complains about the name.. 

```
Error: pkg/download/download.go:91:6: exported: func name will be used as download.DownloadInMemory by other packages, and that stutters; consider calling this InMemory (revive)
func DownloadInMemory(url, userAgent string, transport http.RoundTripper) (io.ReadCloser, error) {
     ^
```

PR #3694, opened 2023-06-01
config: Return exit code 0 when getting default value

### LINKED ISSUE #3678: `crc config get` returns an error code when the setting has a default value
image

cfergeau:
`crc config get xxx` prints:
```
Configuration property 'xxx' is not set. Default value is 'yyy'
```
when its value is not set, but the command exit code is 1.

Since '1' is returned for errors (`crc config get invalid-prop`), this
is not the correct exit code. This commit changes it to 0, which is the
same as successful `crc config get` calls when the value is not the
default one.

This fixes https://github.com/crc-org/crc/issues/3678

### DIFF
--- a/cmd/crc/cmd/config/get.go
+++ b/cmd/crc/cmd/config/get.go
@@ -28,10 +28,10 @@ func configGetCmd(config config.Storage) *cobra.Command {
 			telemetry.SetConfigurationKey(cmd.Context(), args[0])
 
 			if v.IsDefault {
-				return fmt.Errorf("Configuration property '%s' is not set. Default value is '%s'", key, v.AsString())
-
+				fmt.Printf("Configuration property '%s' is not set. Default value is '%s'\n", key, v.AsString())
+			} else {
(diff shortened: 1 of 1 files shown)


gbraad (approved):

cmd/crc/cmd/config/get.go
```diff
@@ -28,10 +28,10 @@ func configGetCmd(config config.Storage) *cobra.Command {
 			telemetry.SetConfigurationKey(cmd.Context(), args[0])
 
 			if v.IsDefault {
-				return fmt.Errorf("Configuration property '%s' is not set. Default value is '%s'", key, v.AsString())
-
+				fmt.Printf("Configuration property '%s' is not set. Default value is '%s'\n", key, v.AsString())
```
gbraad, line 31: It woudl eb better to write:

```go
"Configuration property '%s' is not set. Default value '%s' is used.\n", key, v.AsString())
```

PR #3700, opened 2023-06-07
Add implementation of Server-Sent Events (SSE)

evidolob:
**Fixes:** Issue #N

## Solution/Idea

This PR provides implementation of server sent events. This implementation is used to replace existing WebSocket implementation. Currently only `status` event channel is used for fetching `crc status -w` data.

Also PR provide `logs` event channel on which daemon publish all logs collected by `logrus` logger, it will be used in future work for moving all  crc command execution to daemon, as will allow to keep current cli behaviour/output.

> PR uses [r3labs/sse/v2](https://github.com/r3labs/sse) as SSE library

## Test 

You could use `curl` to view events:

` curl -GET -i --unix-socket <path to crc socket>/crc-http.sock "http://localhost/events?stream=logs"` to view logs events.
` curl -GET -i --unix-socket  <path to crc socket>/crc-http.sock "http://localhost/events?stream=status"` to view data which used for `crc status -w`.  

`crc status -w` should work exactly as before.


go.mod
```diff
@@ -163,7 +161,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/power-devops/perfstat v0.0.0-20210106213030-5aafc221ea8c // indirect
 	github.com/proglottis/gpgme v0.1.3 // indirect
-	github.com/prometheus/client_golang v1.14.0 // indirect
+	github.com/prometheus/client_golang v1.14.0 // indirect; indirect // indirect
```
gbraad: ?

gbraad:
feels almost like a drop-in replacement.

gbraad:
> Fixes: Issue #N

Can you reference the issue this is related to ?

gbraad:
`curl -GET -i --unix-socket <path to crc socket>/crc-http.sock "http://localhost/events?stream=logs"` to view logs events.

This also works on Windows?

gbraad:
> `crc status -w` should work exactly as before.

~We need to verify if the following scenario worked the same before~:

Only CPU and memory are updated:
```
PS> .\crc.exe stop -f
PS> .\crc.exe status -w
CRC VM:          Stopped
MicroShift:      Stopped (v4.13.0)
Cache Usage:     20.72GB
Cache Directory: C:\Users\gbraad\.crc\cache
... in the mean time in another shell start `.\crc.exe start`
... the status like VM and preset does not change; remain 'Stopped'
... wait a little bit: "INFO Starting Microshift service... [takes around 1min]" ... kill watch process 
PS> .\crc.exe status -w
CRC VM:          Running
MicroShift:      Running (v4.13.0)
Disk Usage:      5.462GB of 16.1GB (Inside the CRC VM)
Cache Usage:     20.72GB
Cache Directory: C:\Users\gbraad\.crc\cache
... the status correctly shows the VM is running.
```

See #3744

pkg/crc/api/events/status.go
```diff
@@ -0,0 +1,72 @@
...
+	tickPeriod time.Duration
+}
+
+func newStatusStream(server *EventServer) EventStream {
+	return newStream(NewStatusListener(server.machine), newEventPublisher(STATUS, server.sseServer))
+}
+
+func NewStatusListener(machine crcMachine.Client) EventProducer {
+	getStatus := func() (interface{}, error) {
+		return machine.GetClusterLoad()
```
anjannath, line 28: this'l only provide the memory, disk related info and not the openshift/microshift status right?

gbraad: cluster state would also need to interact over ssh, so this is easy to change. but seems you are right.

gbraad: perhaps related to https://github.com/crc-org/crc/pull/3700#issuecomment-1581888604 as that was `watch` logic?

gbraad: This is therefore existing behaviour and needs to be fixed in a follow-up issue: #3744

gbraad (approved):
how to implement this for the extension?
and followup with #3744 to include state

Issue #3705, opened 2023-06-12
[BUG] disk resize for microshift 4.13.0 bundle is not working


anjannath:
#### `$ crc start --log-level debug` o/p:
```
INFO CRC VM is running
DEBU Running SSH command: cat /home/core/.ssh/authorized_keys
DEBU SSH command results: err: <nil>, output: ecdsa-sha2-nistp521 AAAAE2VjZHNhLXNoYTItbmlzdHA1MjEAAAAIbmlzdHA1MjEAAACFBAHeav7GX53Pca2b9/tRl7KQwlZQLcMrzZCsUnI3ZLNPdWn5SkKZDwTl3uoeWwunYbyu3doCq7QLeSXPtH03iXtiNQFxN60KGH/tioE0UdA/+EA5VMakD3ODsK/gbQJ+rPtiXDfqh81gQD4uWELTtpruCGjS7f546cE4u6SfSNsiDNiaXw== core
INFO Updating authorized keys...
DEBU Creating /home/core/.ssh/authorized_keys with permissions 0644 in the CRC VM
DEBU Running SSH command: <hidden>
DEBU SSH command succeeded
DEBU Running SSH command: rm /home/core/.ssh/authorized_keys.d/ignition
DEBU SSH command results: err: Process exited with status 1, output:
DEBU Using root access: Get PV disk path
DEBU Running SSH command: sudo /usr/sbin/pvs -S lv_name=root -o pv_name --noheadings
DEBU SSH command results: err: <nil>, output:   /dev/vda3
DEBU Using root access: Growing /dev/vda3 partition
DEBU Running SSH command: sudo /usr/bin/growpart /dev/vda 3
DEBU SSH command results: err: <nil>, output: CHANGED: partition=3 start=2050048 old: size=62961664 end=65011711 new: size=81835999 end=83886046
DEBU Using root access: Resizing the physical volume(PV)
DEBU Running SSH command: sudo /usr/sbin/pvresize /dev/vda3
DEBU SSH command results: err: <nil>, output:   Physical volume "/dev/vda3" changed
  1 physical volume(s) resized or updated / 0 physical volume(s) not resized
DEBU Using root access: Get the volume group size
DEBU Running SSH command: sudo /usr/sbin/vgs --noheadings --nosuffix --units b -o vg_size
DEBU SSH command results: err: <nil>, output:   41896902656
DEBU Using root access: Get the size of root logical volume
[...]

PR #3706, opened 2023-06-12
Fix LVM resize error due to fs utilities do not support renamed devices

praveenkumar:
As part of disk resize for microshift looks like there is bug in the lvextend tool and it is not work when extending a lvm with resize.

- https://access.redhat.com/solutions/7016718
- https://bugzilla.redhat.com/show_bug.cgi?id=2211856

This PR is a workaround which is provided as part of BZ where `lvextend` shouldn't run with `-r` (resize) option. So now we are using `lvextend` only for extending the respective logical volume and then using `xfs_growfs` to grow that volume.

**Fixes:** Issue #3705

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -113,7 +113,8 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
 
 func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error {
 	if preset == crcPreset.Microshift {
-		return growLVForMicroshift(sshRunner)
+		lvFullName := "rhel/root"
+		return growLVForMicroshift(sshRunner, lvFullName)
 	}
 	// With 4.7, this is quite a manual process until https://github.com/openshift/installer/pull/4746 gets fixed
 	// See https://github.com/crc-org/crc/issues/2104 for details
@@ -164,8 +165,8 @@ func runGrowpart(sshRunner *crcssh.Runner, rootPart string) error {
 	return nil
 }
 
-func growLVForMicroshift(sshRunner *crcssh.Runner) error {
-	rootPart, _, err := sshRunner.RunPrivileged("Get PV disk path", "/usr/sbin/pvs", "-S", "lv_name=root", "-o", "pv_name", "--noheadings")
+func growLVForMicroshift(sshRunner *crcssh.Runner, lvFullName string) error {
+	rootPart, _, err := sshRunner.RunPrivileged("Get PV disk path", "/usr/sbin/pvs", "-S", fmt.Sprintf("lv_full_name=%s", lvFullName), "-o", "pv_name", "--noheadings")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/start.go
```diff
@@ -164,51 +169,52 @@ func runGrowpart(sshRunner *crcssh.Runner, rootPart string) error {
...
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	// vgFree space as part of the bundle is default to ~15.02G (16127098880 byte)
 	vgFree := 16127098880
 	expectedLVSize := vgSize - vgFree
 	sizeToIncrease := expectedLVSize - lvSize
+	lgVol := "/dev/rhel/root"
```
cfergeau: does `lgVol` stand for `logical volume`? It's usually shortened to `lv`, `lvdisplay` calls it `lvPath`.

cfergeau: Instead of hardcoding `/dev/rhel/root` here and returning it back to the caller, I'd pass `rhel/root` as a `lvFullName` argument, and make this method independent of the VG/PV used (lgVol becomes `dev`+`lvFullName`, and the various `-S lv_name=root` can be replaced with `-S lv_full_name=rhel/root`).

cfergeau (changes requested):

anjannath:
Tested and the error reported in #3705 doesn't occur with this PR, inside the VM `/sysroot` size is also increased:
```
Filesystem             Size  Used Avail Use% Mounted on

/dev/mapper/rhel-root   24G  4.6G   20G  19% /sysroot
```

pkg/crc/machine/start.go
```diff
@@ -169,8 +169,8 @@ func runGrowpart(sshRunner *crcssh.Runner, rootPart string) error {
 	return nil
 }
 
-func growLVForMicroshift(sshRunner *crcssh.Runner) (string, error) {
-	rootPart, _, err := sshRunner.RunPrivileged("Get PV disk path", "/usr/sbin/pvs", "-S", "lv_name=root", "-o", "pv_name", "--noheadings")
+func growLVForMicroshift(sshRunner *crcssh.Runner, lvFullName string) (string, error) {
```
cfergeau: By passing `lvFullName` as an arg, you no longer need to return `lvPath` from this method, which simplifies the other commit. I'd start the PR with this refactor, and then the commit adding the workaround for the resize.

praveenkumar: @cfergeau so what you want is from the caller something like following?
```golang
if preset == crcPreset.Microshift {
               lvFullName := "rhel/root"
               rootPart = fmt.Sprintf("/dev/%s", lvFullName)
		err = growLVForMicroshift(sshRunner, lvFullName)
		if err != nil {
			return err
		}
	}
```

cfergeau: Yes unless you think it does not make sense? This would be close to the way the other codepath is structured.

praveenkumar: Alright, I will make that change, wanted to make sure I understood correctly.

praveenkumar: @cfergeau updated.

cfergeau (approved):

pkg/crc/machine/start.go
cfergeau: Thanks  for the refactor!

PR #3710, opened 2023-06-12
ghactions: Remove microshift test

cfergeau:
It's always been failing recently, it times out because the GH runners
are not powerful/fast enough. Better to remove it as it's not useful to
have a CI test which always/almost always fails.


praveenkumar:
merging it and going to open a follow up issue to have another CI to test microshift bundle on mac.

PR #3712, opened 2023-06-13
Deploy router controller resource from bundle 

praveenkumar:
As part of https://github.com/crc-org/snc/pull/738, we put
router-controller resource as part of bundle and from crc side
we don't need to generate this resource.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -924,6 +924,13 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
 }
 
 func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Config, preset crcPreset.Preset, bundleInfo *bundle.CrcBundleInfo) error {
+	// Check if the bundle have `/opt/crc/route_controller.yaml` file and if it has
+	// then use it to create the resource for router controller.
+	if _, _, err := sshRunner.Run("ls", "/opt/crc/route_controller.yaml"); err == nil {
+		_, _, err := ocConfig.RunOcCommand("apply", "-f", "/opt/crc/route_controller.yaml")
+		return err
+	}
+	// TODO: Remove this resource creation after crc 2.24 release
 	bin, err := json.Marshal(v1.Pod{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Pod",


praveenkumar:
/hold

Till https://github.com/crc-org/snc/pull/738 merge and new bundle available to consume.

pkg/crc/machine/start.go
```diff
@@ -923,47 +920,14 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
...
-	})
-	if err != nil {
-		return err
-	}
-	if err := sshRunner.CopyData(bin, "/tmp/routes-controller.json", 0644); err != nil {
-		return err
-	}
-	_, _, err = ocConfig.RunOcCommand("apply", "-f", "/tmp/routes-controller.json")
+func ensureRoutesControllerIsRunning(ocConfig oc.Config) error {
+	_, _, err := ocConfig.RunOcCommand("apply", "-f", "/opt/route_controller.yaml")
```
cfergeau: Can this go to `/usr/share/crc/routes_controller.yaml` or such?

cfergeau (commented):
After this change it looks like crc won't be able to start older bundles?

pkg/crc/machine/start.go
praveenkumar: @cfergeau any specific reason because we are storing the `kubeconfig` in the same location so didn't want to change it.

cfergeau: I know there are other files there, I made a comment at the time, but I should have pushed stronger :) 
/opt is an odd place to put crc-specific files, especially without subdirectories.

praveenkumar: @cfergeau updated.

praveenkumar:
>  After this change it looks like crc won't be able to start older bundles?

@crc-org/crc-team  Older bundles will not work after this PR merge. Question is do we want to support or need to have this backward compatibility for some releases?

pkg/crc/machine/start.go
praveenkumar: https://github.com/crc-org/snc/pull/738#issuecomment-1591023529

cfergeau:
> > After this change it looks like crc won't be able to start older bundles?
> 
> @crc-org/crc-team Older bundles will not work after this PR merge. Question is do we want to support or need to have this backward compatibility for some releases?

Imo yes, at least 4.13.x bundles, and maybe even 4.12.x. But this is a team decision to make.

gbraad:
We are allowed to cut off support, but as long as we inform the user of this choice.
Since we do not know how many 'custom' bundles are used, this is necessary to be reported

cfergeau:
> We are allowed to cut off support, but as long as we inform the user of this choice.

"allowed" is a thing, but what do we _want_ as a team? break backwards compat any time we need to with a warning? document that a given CRC release will support all 4.13.x bundles if it was released with 4.13.x? Go further than that and support both 4.12 and 4.13 bundles if it was released with 4.13?

pkg/crc/machine/start.go
gbraad: `/opt/crc` makes perfectly sense to me. In my opinion, locations under `/usr` should only be populated by packages.

gbraad:
It is easy to check for the file existence in `/opt/crc` to see if we need to generate the json file instead

gbraad:
> Go further than that and support both 4.12 and 4.13 bundles if it was released with 4.13?

Also, let's not assume anything. We could easily find the information if someone uses an older bundle with a newer version of crc, or a custom bundle. Telemetry... anyone?

Besides, we only 'support' N-1. So anything below OpenShift 4.13 that was part of [2.20.0-4.13.0](https://github.com/crc-org/crc/releases/tag/v2.20.0) by the time we release this (possibly) as part of 2.21, will not receive attention.

praveenkumar:
/unhold

As of now I didn't remove the extra  code to use the older bundle so it now shouldn't break but in future release after 2.24 I would remove that extra code.

pkg/crc/machine/start.go
```diff
@@ -924,6 +924,13 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
 }
 
 func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Config, preset crcPreset.Preset, bundleInfo *bundle.CrcBundleInfo) error {
+	// Check if the bundle have `/opt/crc/route_controller.yaml` file and if it has
+	// then use it to create the resource for router controller.
+	if _, _, err := sshRunner.Run("ls", "/opt/crc/route_controller.yaml"); err == nil {
+		_, _, err := ocConfig.RunOcCommand("apply", "-f", "/opt/crc/route_controller.yaml")
+		return err
+	}
+	// TODO: Remove this resource creation after crc 2.24 release
```
cfergeau, line 933: Why 2.24? As a prerequisite for removing this, I'd first start with having a clear policy wrt bundles, and a clear error message that the bundle is not supported, not a cryptic failure.

cfergeau (approved):

pkg/crc/machine/start.go
cfergeau: `/usr/share` is read-only in RHCOS, but `/usr/share/local/crc/` is also a 'good' location, not owned by packages.

pkg/crc/machine/start.go
praveenkumar: I can see you already created #3730 for that.

Issue #3599, opened 2023-04-13
Follow up on how to show microshift state as part of status


praveenkumar:
As of now microshift service itself wait for all the components https://github.com/openshift/microshift/blob/main/pkg/cmd/run.go#L65 before sending the notification that it is ready. so we are using that for microshift state but it might be case that because of firewall or other the api is not accessible from instance to host.  In the future we might change it, is the api query is successful or may be all the pods from the default created namespace is running.

Discussion: https://github.com/crc-org/crc/pull/3594#discussion_r1164057459

praveenkumar:
Now as part of status for microshift bundle we start using `oc get nodes` with API endpoint and provide the status instead the systemd service.

PR #3714, opened 2023-06-14
Status: use k8s client to get microshift status

praveenkumar:
In this PR, k8s client api is used to get the node status and provide this info to status command instead using `systemctl status microshift` which uses ssh connection.

- https://kubernetes.io/docs/concepts/architecture/nodes/#condition

**Fixes:** Issue #3599


pkg/crc/cluster/clusteroperator.go
```diff
@@ -145,6 +147,37 @@ func getStatus(ctx context.Context, lister operatorLister, selector []string) (*
...
 
+func GetClusterNodeStatus(ctx context.Context, ip string, kubeconfigFilePath string) (*Status, error) {
+	status := &Status{
+		Available: true,
+	}
+	clientSet, err := kubernetesClient(ip, kubeconfigFilePath)
+	if err != nil {
+		return nil, err
+	}
+	nodes, err := clientSet.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
anjannath, line 158: i guess you already tried to use the `openshift` client-go and it didn't work with that and needed the kubernetes `clientset`, do you know why from the openshift clientset we can't get the node statuses?

cfergeau: I had the same question, but a clientset is apparently a thin wrapper over a set o f HTTP APIs. So `openshift/config/clientset` wraps mostly openshift's `ConfigV1()` API, and `k8s/clientset` wraps a different set of HTTP APIs (https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go )

This means you need to pick your client set depending on the API you want to use.

anjannath (approved):

PR #3716, opened 2023-06-14
Add registration trigger to daemon scheduled task

### LINKED ISSUE #3715: [BUG] crc.exe process spawned by Daemon Scheduled Task is not killed within crc cleanup on Windows 11
When trying to run crc on a `Windows 11 Pro` edition it ends up failing due to wrong state for the crc daemon. 

Steps:

1. run `crc setup` (50% of times it will end up failing when checking the state for the crc daemon)
2. if `crc setup` finish ok run `crc start` it fails b/c crc daemon is not running

Checking the task scheduler we can see the task is `Ready`. It says it will be triggered on log on ... (so in case we reboot the VM) then the task appears as Running.

Also this exact flow has been tested on a `Windows 11 Enterprise` edition and on `Windows 10 Pro` and it is working so far, after running the `crc setup` we can see the crc daemon Running (as expected).

As a possible solution we can try to add a 2nd trigger on the task spec, the 2nd trigger would be type: `At task creation/modification`

anjannath:
Currently the daemon task is added with a trigger to start it on log-on but it's usefull to have the daemon task running upon task registration itself, otherwise user has to reboot

docs for the xml of registration trigger can be found at: https://learn.microsoft.com/en-us/windows/win32/taskschd/taskschedulerschema-registrationtrigger-triggergroup-element

however the task not starting after `crc setup` has currently been seen only on windows 11 pro

this fixes #3715

### DIFF
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -40,6 +40,8 @@ var (
     <LogonTrigger>
       <UserId>%s</UserId>
     </LogonTrigger>
+    <RegistrationTrigger>
+    </RegistrationTrigger>
   </Triggers>
   <Actions Context="Author">
     <Exec>


praveenkumar:
> Currently the daemon task is added with a trigger to start it on log-on but it's usefull to have the daemon task running upon task registration itself, otherwise user has to reboot

@anjannath that is not true, when the task is created it is triggered and when user logout/restart then also it is triggered so that user doesn't need to run `crc setup` for each reboot/login. In windows-11 it is not triggered is something changed on task scheduler side.

anjannath:
> > Currently the daemon task is added with a trigger to start it on log-on but it's usefull to have the daemon task running upon task registration itself, otherwise user has to reboot
> 
> @anjannath that is not true, when the task is created it is triggered and when user logout/restart then also it is triggered so that user doesn't need to run `crc setup` for each reboot/login. In windows-11 it is not triggered is something changed on task scheduler side.

yeah, this is only happening on windows 11 pro, but  i was referring to the fact that in the task scheduler definition XML, there's only one trigger, that is `LogonTrigger`, this definitely changed with windows 11 as this was working till now on windows 10 (and also on windows 11 enterprise for some reason)

https://github.com/crc-org/crc/blob/e3efa9e309080bbe651e9839bd92fd31b2e46654/pkg/crc/preflight/preflight_daemon_task_check_windows.go#L39-L43

will reword the first part of the commit message to:
> Currently the daemon task is added with a trigger to start it on log-on and with only that trigger, task was running upon registration as well, with windows 11 pro this is not working so `RegistrationTrigger` is also added

(btw, if this is the fix is unclear, adrian is testing it  on windows 11 pro to verify)

adrianriobo:
yeah, now is working on `Windows 11 Pro` 

@praveenkumar 
> when the task is created it is triggered

Is triggered by the task scheduler? or forced by some crc code?, as Anjan said the tirgger option for it was not set...but also on Windows 11 Enterprise where it works the trigger is not set either (so not sure why is habppenig). 

In other words, this PR only enforces that behavior + solves the issue on `Windows 11 Pro`

adrianriobo (commented):
Tested on Windows 11 Pro and work as expected

adrianriobo (approved):

adrianriobo:
Removed the lgtm as it is still failing, it worked for the first `crc setup` if then I run `crc clenaup` and `crc setup` it keeps showing as `Ready` :(

praveenkumar:
/hold

till we have properly tested it on win-11

anjannath:
closing as this is not the solution, the bug is that on windows 11 `crc cleanup` is not removing the daemon process started by task scheduler

PR #3720, opened 2023-06-16
Add logfile rotation, and delete log files when running `crc delete --clear-cache`

### LINKED ISSUE #3669: [Task] Add cache clean and log file removal functionality
Related to https://github.com/crc-org/crc-extension/issues/128

image

anjannath:
both the crc.log and crcd.log files will be rotated (backed up by
renaming and creating a new log file) after they grow to 5 MBs in
size


gbraad:
> after they grow to 5 MBs in size

You mean this happens continuously ?
The title of the PR suggests differently.

anjannath:
> > after they grow to 5 MBs in size
> 
> You mean this happens continuously ? The title of the PR suggests differently.

yes, this keeps on happening making new backup files and new log files

gbraad:
> make 'crc delete --clear-cache' remove the logfiles

`crc delete` is instance specific.
`--clear-cache` is generic, but handles with `.cache`
log files does not sound like any of this

I even question the `--clear-cache` as that might be wrong to be part of `delete`.

anjannath:
> > make 'crc delete --clear-cache' remove the logfiles
> 
> `crc delete` is instance specific. `--clear-cache` is generic, but handles with `.cache` log files does not sound like any of this
> 
> I even question the `--clear-cache` as that might be wrong to be part of `delete`.

yeah, its not very intuitive, but since we always had a `--clear-cache` flag for `delete` i think our users are used to this CLI

but deleting the log files also with the `--clear-cache` is something i am looking for feedback

with the log file rotation added and the number of backups also configurable (set to 10 backups in this PR) the deleting of log files might not be needed at all.

gbraad:
> but since we always had a --clear-cache flag for delete i think our users are used to this CLI

i doubt many users inow about this. i see many users with gigs stored in their cache. what does telemetry show?

I'd rather go for the more intuitive option

anjannath:
> > but since we always had a --clear-cache flag for delete i think our users are used to this CLI
> 
> i doubt many users inow about this. i see many users with gigs stored in their cache. what does telemetry show?
> 
> I'd rather go for the more intuitive option

you mean, not remove the log files with any command and do it automatically, i.e drop this commit https://github.com/crc-org/crc/pull/3720/commits/b1b06acc7d2e2a9ef2ae481cecb9d8c8b5bd575a from the PR ?

@gbraad or should we remove the need to supply `--clear-cache` flag, and delete log files with just `crc delete` ?

praveenkumar:
> with the log file rotation added and the number of backups also configurable (set to 10 backups in this PR) the deleting of log files might not be needed at all.

why we need 10 backups? shouldn't we just OK with 2 or even single one, as a user I check those logs file only when I hit an error to understand what I have as part of debug and share it during issue creation apart from that I might not care about this log file or is there any other use case of these backups?

anjannath:
> > with the log file rotation added and the number of backups also configurable (set to 10 backups in this PR) the deleting of log files might not be needed at all.
> 
> why we need 10 backups? shouldn't we just OK with 2 or even single one, as a user I check those logs file only when I hit an error to understand what I have as part of debug and share it during issue creation apart from that I might not care about this log file or is there any other use case of these backups?

i added 10 backups, thinking 5mb limit will be reached soon because of all the status calls to the daemon and we might lose logs, but only 2 should be sufficient and we can always change this if we discover we're losing some logs, update the PR!

praveenkumar:
@anjannath looks like during the update you added the other commits also?

anjannath:
@praveenkumar yeah, i rebased on the other PR branch it seems, was supposed to rebase on main, updated again, thanks!

anjannath:
unrelated to the PR, but `e2e` test's `Installing the Redis Operator` in `openshift_story` is failing consistently in other PRs as well, we might need to update the test

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/crc-org_crc/3720/pull-ci-crc-org-crc-main-e2e-crc/1677269032417366016

from another PR https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/crc-org_crc/3748/pull-ci-crc-org-crc-main-e2e-crc/1676523231856562176

adrianriobo:
This is a known issue already reported on the community operators project https://github.com/redhat-openshift-ecosystem/community-operators-prod/issues/2865

PR #3721, opened 2023-06-16
daemon: set execution time limit of daemon task to indefinite

anjannath:
this adds the `<ExecutionTimeLimit>` element with a value of PT0S and a value of PT0S will enable the task to run indefinitely

docs for ExecutionTimeLimit can be found at:
https://learn.microsoft.com/en-us/windows/win32/taskschd/taskschedulerschema-executiontimelimit-settingstype-element

### DIFF
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -35,6 +35,7 @@ var (
       <RestartOnIdle>false</RestartOnIdle>
     </IdleSettings>
     <UseUnifiedSchedulingEngine>true</UseUnifiedSchedulingEngine>
+    <ExecutionTimeLimit>PT0S</ExecutionTimeLimit>
   </Settings>
   <Triggers>
     <LogonTrigger>


pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -35,6 +35,7 @@ var (
       <RestartOnIdle>false</RestartOnIdle>
     </IdleSettings>
     <UseUnifiedSchedulingEngine>true</UseUnifiedSchedulingEngine>
+    <ExecutionTimeLimit>PT0S</ExecutionTimeLimit>
```
cfergeau, line 38: I haven't been able to find where the `PT0S` syntax is documented even after following your link, but looks good to me. I assume this has no impact wrt to crc starting/stopping the service? It just will stay around forever instead of only for 3 days?

cfergeau (approved):

pkg/crc/preflight/preflight_daemon_task_check_windows.go
anjannath: yes it'll just stay on forever now, and the 3 days limit will not be applied

the syntax can be found in the first section of this page, https://learn.microsoft.com/en-us/windows/win32/taskschd/taskschedulerschema-executiontimelimit-settingstype-element 

(unfortunately can't link to the section directly) but heres the text:

> Amount of time allowed to complete the task.The format for this string is PnYnMnDTnHnMnS, where nY is the number of years, nM is the number of months, nD is the number of days, 'T' is the date/time separator, nH is the number of hours, nM is the number of minutes, and nS is the number of seconds (for example, PT5M specifies 5 minutes and P1M4DT2H5M specifies one month, four days, two hours, and five minutes). For more information about the duration type, see https://go.microsoft.com/fwlink/p/?linkid=106886. A value of PT0S will enable the task to run indefinitely.

cfergeau: Thanks for the C&P, fairly cryptic :) My bad, I focused on the XML and below it, and ignored entirely the first paragraph /o\

Issue #3718, opened 2023-06-15
[BUG] Issue growing Logical Volume 


jsliacan:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V 
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Baremetal-Server

## CRC version
```bash
CRC version: 2.21.0+e3efa9
OpenShift version: 4.13.0
Podman version: 4.4.4
```

## CRC config
```bash
- consent-telemetry                     : no
- preset                                : microshift
```

### Steps to reproduce

  1. install 2.21.0
  2. `crc setup`
  3. `crc start`
  4. 

### Logs
```
PS C:\Users\crcqe> crc setup
INFO Using bundle path C:\Users\crcqe\.crc\cache\crc_microshift_hyperv_4.13.0_amd64.crcbundle
INFO Checking minimum RAM requirements
INFO Checking if current user is in crc-users and Hyper-V admins group
INFO Checking if CRC bundle is extracted in '$HOME/.crc'
INFO Checking if C:\Users\crcqe\.crc\cache\crc_microshift_hyperv_4.13.0_amd64.crcbundle exists
INFO Checking if the daemon task is installed
INFO Installing the daemon task
INFO Checking if the daemon task is running
INFO Running the daemon task
INFO Checking admin helper service is running
Your system is correctly setup for using CRC. Use 'crc start' to start the instance
PS C:\Users\crcqe> crc start -p C:\Users\crcqe\Downloads\crc-pull-secret
INFO Checking minimum RAM requirements
INFO Checking if running in a shell with administrator rights
INFO Checking Windows release
INFO Checking Windows edition
INFO Checking if Hyper-V is installed and operational
INFO Checking if Hyper-V service is enabled
INFO Checking if crc-users group exists
INFO Checking if current user is in crc-users and Hyper-V admins group
INFO Checking if vsock is correctly configured
INFO Checking if the daemon task is installed
INFO Checking if the daemon task is running
INFO Checking admin helper service is running
INFO Loading bundle: crc_microshift_hyperv_4.13.0_amd64...
INFO Creating CRC VM for MicroShift 4.13.0...
INFO Generating new SSH key pair...
INFO Starting CRC VM for microshift 4.13.0...
INFO CRC instance is running with IP 127.0.0.1
INFO CRC VM is running
INFO Updating authorized keys...
panic: runtime error: slice bounds out of range [:8] with length 0

goroutine 1 [running]:
[...]

PR #3722, opened 2023-06-16
GrowLVForMicroshift: Change disk '/dev/vda' to '/dev/sda' for windows

praveenkumar:
In windows the hyperV identified disk as SCSI and disk partitions are `/dev/sda*` instead `/dev/vda*` and this makes the pvs/lvs commands to fail since the `/etc/lvm/devices/system.devices` file contain `/dev/vda*`.

```
$ sudo pvdisplay
  Devices file PVID BAUW2KTKhF3RE03h4JEMd1EImwC3cWd0 last seen on
  /dev/vda3 not found.
```

In this PR we are changing the `/dev/vda*` to `/dev/sda*` so that disk resize works as expected in windows also.

**Fixes:** Issue #3718

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -112,23 +112,18 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
 }
 
 func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error {
-	var rootPart string
-	var err error
+	rootPart, err := getrootPartition(sshRunner, preset)
+	if err != nil {
+		return err
+	}
+
+	if err := runGrowpart(sshRunner, rootPart); err != nil {
+		return err
+	}
+
 	if preset == crcPreset.Microshift {
 		lvFullName := "rhel/root"
-		rootPart = fmt.Sprintf("/dev/%s", lvFullName)
-		if err := growLVForMicroshift(sshRunner, lvFullName); err != nil {
-			return err
-		}
-	} else {
-		// With 4.7, this is quite a manual process until https://github.com/openshift/installer/pull/4746 gets fixed
-		// See https://github.com/crc-org/crc/issues/2104 for details
-		rootPart, err = getrootPartition(sshRunner, "/dev/disk/by-label/root")
-		if err != nil {
-			return err
-		}
-
-		if err := runGrowpart(sshRunner, rootPart); err != nil {
+		if err := growLVForMicroshift(sshRunner, lvFullName, rootPart); err != nil {
 			return err
 		}
 	}
@@ -145,12 +140,20 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error
 	return nil
 }
 
-func getrootPartition(sshRunner *crcssh.Runner, label string) (string, error) {
-	rootPart, _, err := sshRunner.Run("realpath", label)
+func getrootPartition(sshRunner *crcssh.Runner, preset crcPreset.Preset) (string, error) {
+	diskType := "xfs"
+	if preset == crcPreset.Microshift {
+		diskType = "LVM2_member"
+	}
+	part, _, err := sshRunner.RunPrivileged("Get device id", "/usr/sbin/blkid", "-t", fmt.Sprintf("TYPE=%s", diskType), "-o", "device")
 	if err != nil {
 		return "", err
 	}
-	rootPart = strings.TrimSpace(rootPart)
+	parts := strings.Split(strings.TrimSpace(part), "\n")
+	if len(parts) != 1 {
+		return "", fmt.Errorf("Unexpected number of devices: %s", part)
+	}
+	rootPart := strings.TrimSpace(parts[0])
 	if !strings.HasPrefix(rootPart, "/dev/vda") && !strings.HasPrefix(rootPart, "/dev/sda") {
 		return "", fmt.Errorf("Unexpected root device: %s", rootPart)
 	}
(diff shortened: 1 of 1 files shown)


cfergeau (changes requested):
Shouldn't this be done in snc instead? 
I'd also add some hardening when getting `rootPart` from the VM, and in `runGrowPart`

praveenkumar:
> Shouldn't this be done in snc instead?

How we do it in snc, before creating the windows bundle?

cfergeau:
> > Shouldn't this be done in snc instead?
> 
> How we do it in snc, before creating the windows bundle?

yes, we can't do it before that. I'm also wondering if we could use entries from `/dev/disk/by-xxxx/` instead as their name won't depend on the virtualization stack.

gbraad (changes requested):
Looks OK from code.

This means the next call will get a result back and the grow will not fail...

However, can you make sure to err if that value is invalid, like empty? So what if this returns hda in another situation, instead of vda or sda? Continue execution with an error message?

praveenkumar:
> However, can you make sure to err if that value is invalid, like empty? So what if this returns hda in another situation, instead of vda or sda? Continue execution with an error message?

In case of openshift bundle we have following logic, may be same I can use here which will also capture in case value is empty so user not get unexpected panic.
 
```
rootPart = strings.TrimSpace(rootPart)
	if !strings.HasPrefix(rootPart, "/dev/vda") && !strings.HasPrefix(rootPart, "/dev/sda") {
		return "", fmt.Errorf("Unexpected root device: %s", rootPart)
	}
```

praveenkumar:
/hold

need to refactor.

pkg/crc/machine/start.go
```diff
@@ -196,7 +190,7 @@ func growLVForMicroshift(sshRunner *crcssh.Runner, lvFullName string) error {
 	}
 
 	// Get the size of root lv
-	sizeLV, _, err := sshRunner.RunPrivileged("Get the size of root logical volume", "/usr/sbin/lvs", "-S", fmt.Sprintf("lv_full_name=%s", lvFullName), "--noheadings", "--nosuffix", "--units", "b", "-o", "lv_size")
+	sizeLV, _, err := sshRunner.RunPrivileged("Get the size of root logical volume", "/usr/sbin/lvs", "-S", fmt.Sprintf("lv_full_name=%s", lvFullName), "--noheadings", "--nosuffix", "--units", "b", "-o", "lv_size", "--devices", rootPart)
```
gbraad, line 194: "/sysroot" ?

pkg/crc/machine/start.go
```diff
@@ -171,22 +174,13 @@ func runGrowpart(sshRunner *crcssh.Runner, rootPart string) error {
...
-
-	if _, _, err := sshRunner.RunPrivileged("Resizing the physical volume(PV)", "/usr/sbin/pvresize", rootPart); err != nil {
+func growLVForMicroshift(sshRunner *crcssh.Runner, lvFullName string, rootPart string) error {
+	if _, _, err := sshRunner.RunPrivileged("Resizing the physical volume(PV)", "/usr/sbin/pvresize", "--devices", rootPart, rootPart); err != nil {
 		return err
 	}
 
 	// Get the size of volume group
-	sizeVG, _, err := sshRunner.RunPrivileged("Get the volume group size", "/usr/sbin/vgs", "--noheadings", "--nosuffix", "--units", "b", "-o", "vg_size")
+	sizeVG, _, err := sshRunner.RunPrivileged("Get the volume group size", "/usr/sbin/vgs", "--noheadings", "--nosuffix", "--units", "b", "-o", "vg_size", "--devices", rootPart)
```
gbraad, line 184: "/sysroot" ?

Everywhere else in the PR this has been removed/changed. I am not sure why that is... as it seems still relevant.

pkg/crc/machine/start.go
```diff
@@ -212,7 +206,7 @@ func growLVForMicroshift(sshRunner *crcssh.Runner, lvFullName string) error {
 	lvPath := fmt.Sprintf("/dev/%s", lvFullName)
 	if sizeToIncrease > 1 {
 		logging.Info("Extending and resizing '/dev/rhel/root' logical volume")
-		if _, _, err := sshRunner.RunPrivileged("Extending and resizing the logical volume(LV)", "/usr/sbin/lvextend", "-L", fmt.Sprintf("+%db", sizeToIncrease), lvPath); err != nil {
+		if _, _, err := sshRunner.RunPrivileged("Extending and resizing the logical volume(LV)", "/usr/sbin/lvextend", "-L", fmt.Sprintf("+%db", sizeToIncrease), lvPath, "--devices", rootPart); err != nil {
```
gbraad, line 210: "/sysroot" ?

gbraad (changes requested):
I see `rootPart` and `"/sysroot"` being used. Are they the same? And if so, please use a constant.

pkg/crc/machine/start.go
```diff
@@ -137,19 +132,27 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error
 	if _, _, err := sshRunner.RunPrivileged("Remounting /sysroot read/write", "mount -o remount,rw /sysroot"); err != nil {
 		return err
 	}
-	if _, _, err = sshRunner.RunPrivileged(fmt.Sprintf("Growing %s filesystem", rootPart), "xfs_growfs", rootPart); err != nil {
+	if _, _, err = sshRunner.RunPrivileged(fmt.Sprintf("Growing %s filesystem", rootPart), "xfs_growfs", "/sysroot"); err != nil {
```
gbraad: Previous comments are regarding "`/sysroot` which seems to be a shared mount. seeing the change looks awkward without ulling this out as a more meaningful constant, like `rootMount`

pkg/crc/machine/start.go
```diff
@@ -137,19 +132,27 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error
 	if _, _, err := sshRunner.RunPrivileged("Remounting /sysroot read/write", "mount -o remount,rw /sysroot"); err != nil {
```
gbraad: you also use `rootMount` here, which justifies re-use for clarity

pkg/crc/machine/start.go
praveenkumar: Updated with `rootFS` variable.

pkg/crc/machine/start.go
praveenkumar: updated with `rootFS` variable

gbraad (approved):

Issue #3725, opened 2023-06-19
[BUG] kubeconfig context is not detected properly


datho7561:
### General information

  * OS: Linux, Fedora 38
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it **Yes**
  * Running CRC on: Laptop

## CRC version
```bash
CRC version: 2.20.0+f3a947
OpenShift version: 4.13.0
Podman version: 4.4.4
```
  
## CRC status
```bash
DEBU CRC version: 2.20.0+f3a947
DEBU OpenShift version: 4.13.0
DEBU Podman version: 4.4.4
DEBU Running 'crc status'
CRC VM:          Running
MicroShift:      Running (v4.13.0)
RAM Usage:       1.317GB of 3.902GB
Disk Usage:      4.814GB of 16.1GB (Inside the CRC VM)
Cache Usage:     54.43GB
Cache Directory: /home/davthomp/.crc/cache
```

## CRC config
```bash
- consent-telemetry                     : yes
- preset                                : microshift
```

## Host Operating System
```bash
NAME="Fedora Linux"
VERSION="38 (Workstation Edition)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=38
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=38
SUPPORT_END=2024-05-14
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

### Steps to reproduce

  1. Set the preset to MicroShift
  2. Make sure libvirtd is enabled and started
  3. `crc setup`
  4. `crc start`
  5. `oc get deployment`

### Expected
[...]

PR #3727, opened 2023-06-22
microshift: Use api.crc.testing kubeconfig to connect to cluster

praveenkumar:
With 4.13, after starting the microshift service it create different kubeconfig according to different additional names/IP. we were using the localhost one which is mostly for localhost API access so in this PR we are going to make use of `api.crc.testing` kubecofig file instead localhost one.

- https://github.com/openshift/microshift/blob/main/docs/howto_kubeconfig.md
- https://access.redhat.com/documentation/en-us/red_hat_build_of_microshift/4.13/html/configuring/microshift-kubeconfig#remote-access-con_microshift-kubeconfig

**Fixes:** Issue #3725

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -1033,13 +1033,10 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 	if _, _, err := sshRunner.RunPrivileged("Starting microshift service", "systemctl", "start", "microshift"); err != nil {
 		return err
 	}
-	if err := sshRunner.CopyFileFromVM("/var/lib/microshift/resources/kubeadmin/kubeconfig", constants.KubeconfigFilePath, 0600); err != nil {
+	if err := sshRunner.CopyFileFromVM(fmt.Sprintf("/var/lib/microshift/resources/kubeadmin/api%s/kubeconfig", constants.ClusterDomain), constants.KubeconfigFilePath, 0600); err != nil {
 		return err
 	}
 	if err := sshRunner.CopyFile(constants.KubeconfigFilePath, "/opt/kubeconfig", 0644); err != nil {
-		return err
-	}
-	if err := updateServerDetailsToKubeConfig(constants.KubeconfigFilePath, constants.KubeconfigFilePath); err != nil {
 		return err
 	}
 
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -232,19 +232,6 @@ func contains(arr []string, str string) bool {
 	return false
 }
 
-func updateServerDetailsToKubeConfig(input, output string) error {
-	cfg, err := clientcmd.LoadFromFile(input)
-	if err != nil {
-		return err
-	}
-	for name, cluster := range cfg.Clusters {
-		if name == "microshift" && cluster.Server == "https://127.0.0.1:6443" {
-			cluster.Server = fmt.Sprintf("https://api%s:6443", constants.ClusterDomain)
-		}
-	}
-	return clientcmd.WriteToFile(*cfg, output)
-}
-
 func mergeKubeConfigFile(kubeConfigFile string) error {
 	globalConfigPath, kubeConf1, err := getGlobalKubeConfig()
 	if err != nil {


cfergeau (commented):
> Note: 4.12.x bundle is not able to use after this change.

This is only microshift 4.12.x bundles which will no longer be usable? And why is this? Because /var/lib/microshift/resources/kubeadmin/api.crc.testing/kubeconfig does not exist?

This removes code so this is a good change ;)

pkg/crc/machine/start.go
```diff
@@ -1033,15 +1033,12 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 	if _, _, err := sshRunner.RunPrivileged("Starting microshift service", "systemctl", "start", "microshift"); err != nil {
 		return err
 	}
-	if err := sshRunner.CopyFileFromVM("/var/lib/microshift/resources/kubeadmin/kubeconfig", constants.KubeconfigFilePath, 0600); err != nil {
+	if err := sshRunner.CopyFileFromVM(fmt.Sprintf("/var/lib/microshift/resources/kubeadmin/api%s/kubeconfig", constants.ClusterDomain), constants.KubeconfigFilePath, 0600); err != nil {
 		return err
 	}
 	if err := sshRunner.CopyFile(constants.KubeconfigFilePath, "/opt/kubeconfig", 0644); err != nil {
```
cfergeau, line 1039: Is this ok to use api.crc.testing inside the VM? Or should we keep using the 'localhost' one there?

praveenkumar: @cfergeau it is OK because `api.crc.testing` is resolvable in the VM also so it shouldn't be any isuse of using it.

praveenkumar:
> This is only microshift 4.12.x bundles which will no longer be usable?
Right it is only the microshift bundle, openshift bundle are not affected by this change.

> And why is this? Because /var/lib/microshift/resources/kubeadmin/api.crc.testing/kubeconfig does not exist?

Yes because on 4.12.x side the single kubeconfig file had altname instead having separate kubeconfig files which is not the case now.

cfergeau:
> > This is only microshift 4.12.x bundles which will no longer be usable?
> > Right it is only the microshift bundle, openshift bundle are not affected by this change.
> 
> > And why is this? Because /var/lib/microshift/resources/kubeadmin/api.crc.testing/kubeconfig does not exist?
> 
> Yes because on 4.12.x side the single kubeconfig file had altname instead having separate kubeconfig files which is not the case now.

Dunno how we nicely deal with this? Do we prevent running 4.12.x bundles at the moment?

praveenkumar:
> > > This is only microshift 4.12.x bundles which will no longer be usable?
> > > Right it is only the microshift bundle, openshift bundle are not affected by this change.
> > 
> > 
> > > And why is this? Because /var/lib/microshift/resources/kubeadmin/api.crc.testing/kubeconfig does not exist?
> > 
> > 
> > Yes because on 4.12.x side the single kubeconfig file had altname instead having separate kubeconfig files which is not the case now.
> 
> Dunno how we nicely deal with this?

I also don't know how we can nicely handle it without putting extra code to make 4.12.x bundle workable but our support is just N/N-1 , if someone wants to use 4.12 bundle then might be better to use older version of crc.

> Do we prevent running 4.12.x bundles at the moment?

yes, as soon it goes in we prevent running 4.12.x microshift bundle because there is no file `var/lib/microshift/resources/kubeadmin/api.crc.testing/kubeconfig` so we start the microshift service but not able to copy any kubeconfig file to connect to this cluster.

praveenkumar:
> Yes because on 4.12.x side the single kubeconfig file had altname instead having separate kubeconfig files which is not the case now.

So I take that back, I tried latest 4.12.15 bundle and it also have a kubeconfig for altname which means this change work with our latest 4.12.x bundle.

cfergeau:
> I also don't know how we can nicely handle it without putting extra code to make 4.12.x bundle workable but our support is just N/N-1 , if someone wants to use 4.12 bundle then might be better to use older version of crc.

Not supporting a given version is one thing, I was asking if/how to print a nice error message to inform the user that their bundle is not supported, rather than a cryptic error message + failure to start.

praveenkumar:
> > I also don't know how we can nicely handle it without putting extra code to make 4.12.x bundle workable but our support is just N/N-1 , if someone wants to use 4.12 bundle then might be better to use older version of crc.
> 
> Not supporting a given version is one thing, I was asking if/how to print a nice error message to inform the user that their bundle is not supported, rather than a cryptic error message + failure to start.

#3730 created for this, also as of now this change is not going to fail for 4.12.x bundle also.

cfergeau (approved):

PR #3728, opened 2023-06-22
cleanup: delete VM before deleting the machine's dir

anjannath:
trying to delete the machine's dir without first deleting the hyper-v VM
causes error, since the existing VM will still be using the files inside
the machine's dir

### DIFF
--- a/pkg/crc/preflight/preflight_windows.go
+++ b/pkg/crc/preflight/preflight_windows.go
@@ -70,13 +70,14 @@ var hypervPreflightChecks = []Check{
 
 		labels: labels{Os: Windows, NetworkMode: System},
 	},
-	{
-		cleanupDescription: "Removing crc's virtual machine",
-		cleanup:            removeCrcVM,
-		flags:              CleanUpOnly,
+}
 
-		labels: labels{Os: Windows},
-	},
+var cleanupCheckRemoveCrcVM = Check{
+	cleanupDescription: "Removing crc's virtual machine",
+	cleanup:            removeCrcVM,
+	flags:              CleanUpOnly,
+
+	labels: labels{Os: Windows},
 }
 
 var vsockChecks = []Check{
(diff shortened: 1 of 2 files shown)


pkg/crc/preflight/preflight_windows.go
```diff
@@ -70,13 +70,14 @@ var hypervPreflightChecks = []Check{
...
+}
 
-		labels: labels{Os: Windows},
-	},
+var cleanupCheckRemoveCrcVM = Check{
+	cleanupDescription: "Removing crc's virtual machine",
+	cleanup:            removeCrcVM,
+	flags:              CleanUpOnly,
+
+	labels: labels{Os: Windows},
```
praveenkumar, line 80: @anjannath we are creating `cleanupCheckRemoveCrcVM` as separate check variable to change the ordering around how it is run during cleanup?

anjannath: yes correct, to order it such, that the `removeCrcVM` runs before removing the machine's dir

praveenkumar:
```golang
checks := []Check{}
	checks = append(checks, memoryCheck(preset))
	checks = append(checks, hypervPreflightChecks...)
	checks = append(checks, crcUsersGroupExistsCheck)
	checks = append(checks, userPartOfCrcUsersAndHypervAdminsGroupCheck)
	checks = append(checks, vsockChecks...)
	checks = append(checks, bundleCheck(bundlePath, preset))
	checks = append(checks, genericCleanupChecks...)
	checks = append(checks, daemonTaskChecks...)
	checks = append(checks, adminHelperServiceCheks...)
```

> yes correct, to order in such, that the removeCrcVM runs before removing the machine's dir

@anjannath so aren't we already remove VM before removing machine dir because removeVM is part of `hypervPreflightChecks` and dir remove is part of `genericCleanupChecks` ?

anjannath: no, during cleanup we execute the checks from the end, https://github.com/crc-org/crc/blob/e330d8781455d8b43b0ff5d1618c2bfe8623d6c3/pkg/crc/preflight/preflight.go#L123

in the the list of `checks` the `hypervPreflightCheks` are added to the list before the `genericCleanupChecks`, so when cleanup is run the `genericCleanupChecks` runs before the `hypervPreflightChecks`

praveenkumar: In that case the detail why we are creating a `cleanupCheckRemoveCrcVM` as separate check in the commit log to provide a sense why we did it.

anjannath: update the commit msg, added

>  this separates out the check deleting the VM from the list of hyperv
    checks, and adds it to the list of all checks after the generic cleanup
    checks to make it run before deleting the machine's dir

Issue #3735, opened 2023-06-27
[Telemetry] Send preset value even there is early error during crc start


praveenkumar:
Looking at amplitude matrix around preset, there is lot of number around `none` preset. Then I looked at the segment which have raw data around our telemetry and found out in case there is early error we are not sending the preset value to segment and hence this `none` happen.
```
client.Track(&analytics.Track{
  UserId: "xxxxxxxxxxxxxxxxxxxx",
  Event: "crc start",
  Properties: map[string]interface{}{
    "duration": 3558,
    "error": "Hyper-V Virtual Machine Management service not running",
    "error-type": "os.CodeExitError",
    "remote": false,
    "source": "cli",
    "success": false,
    "tty": true,
    "version": "2.20.0",
  },
})
```

or
 
```
client.Track(&analytics.Track{
  UserId: "xxxxxxxxxxxxxxxxx",
  Event: "crc start",
  Properties: map[string]interface{}{
    "duration": 2926,
    "error": "crc should be ran in a shell without administrator rights",
    "error-type": "os.CodeExitError",
    "remote": false,
    "source": "cli",
    "success": false,
    "tty": true,
    "version": "2.20.0",
  },
```

gbraad:
> we are not sending the preset value to segment and hence this none happen.

Why is that actually not happening? As it seems, there is an 'error'. Any idea what the error is that prevents the preset to be sent? Or is this very generic?

It can't be just these:

> "Hyper-V Virtual Machine Management service not running"

> "crc should be ran in a shell without administrator rights",

as these are mostly part of the onboarding phase. At that moment the preset might not have been established. but I also know the number of these cases were not very minimal. I doubt these happen at the same rate as the 'none' being reported.

praveenkumar:
We are now sending preset details for each event/interaction. hopefully with next release we will see less `none` around it.

PR #3740, opened 2023-06-28
Segment: Add preset value for each event before sending to segment

praveenkumar:
Preset value is only capured in the start command and if start command
fails as part of preflight test or even daemon not available phase then
this value is not available. We tried to map the preset data from amplitude
side and there is lot of `none` because some segment data (captured by
start action) doesn't have preset info. This PR make sure that each
events have preset value before sending it to segment.

Issue #3735

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -168,6 +168,9 @@ func (c *Client) upload(action string, a analytics.Properties) error {
 	if c.config.Get(crcConfig.ConsentTelemetry).AsString() != "yes" {
 		return nil
 	}
+	// Add preset value as property for each event before uploading to segment
+	preset := crcConfig.GetPreset(c.config)
+	a = a.Set("preset", preset)
 
 	identify := c.identifyNew()
 	hash, err := identifyHash(identify)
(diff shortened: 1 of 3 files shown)


pkg/crc/segment/segment.go
```diff
@@ -168,6 +168,9 @@ func (c *Client) upload(action string, a analytics.Properties) error {
 	if c.config.Get(crcConfig.ConsentTelemetry).AsString() != "yes" {
 		return nil
 	}
+	// Add preset value as property for each event before uploading to segment
+	preset := crcConfig.GetPreset(c.config)
+	a = a.Set("preset", preset)
```
cfergeau, line 173: Looks like a good fit for being done in `addConfigTraits` instead of `upload`?

cfergeau: Or maybe not,`addConfigTraits` is only called to create the identify event, which we try to avoid sending too often. Dunno what's best :)

praveenkumar: @cfergeau yes `addConfigTraits` we only called once when new user ID is created and not after that, as part of amplitude we need this matrix for every start operation.

cfergeau: Well, you could go from operation to user id to preset.

praveenkumar: @cfergeau what if the user initially use `openshift` preset but then change it to `microshift` I think the config traits are only one time when new userid generated not when a config is changes right? Also I don't have any idea as of now from amplitude side how it can go from operation => uid => preset, right now the matrix we have action events and then filter on those.

gbraad (approved):

pkg/crc/segment/segment.go
gbraad: they want preset for as many actions, so if we can provide this as part of existing events I do not see why 'try to avoid sending too often' matters

cfergeau: > I do not see why 'try to avoid sending too often' matters

My question was about putting the code in the right place, rather than adding it to a  method where it looks out of place. And actually, it would fit better in `properties()` rather than `upload`.

PR #3750, opened 2023-07-06
[wip] Add `InstanceName` and `InstanceDirName` functions

anjannath:
The instance directory and instance name for libmachine is same but
for better clearity two funcs InstanceName() and  InstanceDirName()
are added

When we are dealing with the VM name we should use InstanceName and
for dealing with the paths for the VM we should use InstanceDirName

_This is initial work to have the instance name based on the preset, so that multiple instances can exist simultaneously. Later we'll modify the `InstanceName` funtion to return the name based on the preset_


praveenkumar:
I am expecting as of now this PR is base work for having different directory for different preset, can we add more commits to it (even for default preset) to provide sample logic how these changes are consumed because as of now we are just adding 2 helper functions and refactoring code around it but doesn't provide an idea how it is consumable from preset context (even bit details on the issue or here might help to understand the usecase of it).

anjannath:
> I am expecting as of now this PR is base work for having different directory for different preset, can we add more commits to it (even for default preset) to provide sample logic how these changes are consumed because as of now we are just adding 2 helper functions and refactoring code around it but doesn't provide an idea how it is consumable from preset context (even bit details on the issue or here might help to understand the usecase of it).

my plan is to next modify `InstanceName` to return the `crc-$preset` as the name, to have the instances differently named, i'll add more commits to the PR

test/integration/utilities_test.go
```diff
@@ -97,8 +98,8 @@ func RunCRCExpectFail(args ...string) (string, error) {
 }
 
 // Send command to CRC VM via SSH
-func SendCommandToVM(cmd string) (string, error) {
-	client := machine.NewClient(constants.InstanceName(), false,
+func SendCommandToVM(cmd string, preset crcPreset.Preset) (string, error) {
+	client := machine.NewClient(constants.InstanceName(preset), false,
```
jsliacan, line 102: I see, so this `constants.InstanceName` is now preset dependent, and the changes we need in tests are all either directly to the `constants.InstanceName` or to the `SendCommandToVM` function which utilizes this. LGTM. Thanks for the changes!

gbraad:
@praveenkumar what more do you want to know?

pkg/crc/constants/constants.go
```diff
@@ -109,8 +109,8 @@ var (
 	MachineCacheDir    = filepath.Join(MachineBaseDir, "cache")
 	MachineInstanceDir = filepath.Join(MachineBaseDir, "machines")
 	DaemonSocketPath   = filepath.Join(CrcBaseDir, "crc.sock")
-	KubeconfigFilePath = filepath.Join(MachineInstanceDir, DefaultName, "kubeconfig")
-	PasswdFilePath     = filepath.Join(MachineInstanceDir, DefaultName, "passwd")
+	KubeconfigFilePath = filepath.Join(MachineInstanceDir, InstanceDirName(), "kubeconfig")
```
cfergeau: Fwiw, I'd add a small helper such as:
```
func ResolveInstancePath(preset crcpreset.Preset, pathElems ...string) string {
	return filepath.Join(MachineInstancesDir, InstanceDirName(preset), filepath.Join(pathElems...))
}
```
to stop having  `filepath.Join(MachineInstanceDir, InstanceDirName(), someFile)` in many places.

pkg/crc/oc/oc.go
```diff
@@ -81,7 +81,7 @@ func UseOCWithSSH(sshRunner *ssh.Runner) Config {
 		OcExecutablePath: "oc",
 		KubeconfigPath:   "/opt/kubeconfig",
 		Context:          constants.DefaultContext,
-		Cluster:          constants.DefaultName,
+		Cluster:          constants.DefaultClusterName,
```
cfergeau: seeing https://github.com/crc-org/crc/blob/233df0ee1c05f91d49dbe72dc58a2288ac09f30f/pkg/crc/machine/start.go#L556-L559, this could be further improved in follow-up commits.

pkg/crc/constants/constants.go
```diff
@@ -174,24 +174,24 @@ func EnsureBaseDirectoriesExist() error {
 }
 
 func GetPublicKeyPath() string {
-	return filepath.Join(MachineInstanceDir, InstanceDirName(), "id_ecdsa.pub")
+	return filepath.Join(MachineInstancesDir, InstanceDirName(), "id_ecdsa.pub")
```
cfergeau: I'd prefer to focus for now on having VM names, or machine dir name being `crc-$preset` instead of just `crc`, but keep for now "only one VM can exist at a given time".
This means most of this PR gets in, we can start making sure the change of names don't cause too many issues, ...
Then another PR can add support for multiple instance dir in the same subdir, multiple VMs, ...

pkg/crc/constants/constants.go
```diff
@@ -113,12 +113,12 @@ var (
 	PasswdFilePath      = filepath.Join(MachineInstancesDir, InstanceDirName(), "passwd")
 )
 
-func InstanceDirName() string {
-	return InstanceName()
+func InstanceDirName(preset crcpreset.Preset) string {
```
cfergeau, line 119: After this change, compilation will be broken for a few commits. Since it is a very mechanical change, doing everything in one go should be ok.

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -577,46 +577,50 @@ func removeLibvirtCrcNetwork() error {
...
 func removeCrcVM() error {
-	stdout, _, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "domstate", constants.InstanceName())
-	if err != nil {
-		//  User may have run `crc delete` before `crc cleanup`
-		//  in that case there is no crc vm so return early.
-		return nil
-	}
-	if strings.TrimSpace(stdout) == "running" {
-		_, stderr, err := crcos.RunWithDefaultLocale("virsh", "--connect", "qemu:///system", "destroy", constants.InstanceName())
+	for _, instanceName := range getAllInstanceNamesToDelete() {
```
cfergeau: I'd prefer to see this commit split in 2 parts, one adjusting for the addition of the `preset` arg to the `constants.InstanceName()` function, and another one adding the iteration over multiple directories. 
Not sure we want the iteration over multiple dirs right now.

cmd/crc/cmd/start.go
```diff
@@ -346,7 +346,6 @@ func writeMicroShiftTemplatedMessage(writer io.Writer) error {
 	return parsed.Execute(writer, &templateVariables{
 		EvalCommandLine:   shell.GenerateUsageHint(userShell, "crc oc-env"),
 		CommandLinePrefix: commandLinePrefix(userShell),
-		KubeConfigPath:    constants.KubeconfigFilePath,
```
cfergeau, line 353: I guess this is unused? Might deserve its own commit/explanation.

pkg/crc/machine/virtualmachine.go
```diff
@@ -101,10 +103,23 @@ func (vm *virtualMachine) SSHPort() int {
...
 }
 
+func guessPresetFromVMName(vmname string) (preset.Preset, error) {
+	p := strings.TrimPrefix(vmname, "crc-")
+	parsedPreset, err := preset.ParsePresetE(p)
+	if err != nil {
+		return preset.Preset(""), err
+	}
+	return parsedPreset, nil
+}
```
cfergeau: hardcoding a vm name -> preset mapping outside of constants.go (and even doing this) is something I'd prefer to avoid.

pkg/crc/machine/virtualmachine.go
```diff
@@ -101,10 +103,23 @@ func (vm *virtualMachine) SSHPort() int {
...
 	ip, err := vm.IP()
 	if err != nil {
 		return nil, err
 	}
-	return ssh.CreateRunner(ip, vm.SSHPort(), constants.GetPrivateKeyPath(), constants.GetRsaPrivateKeyPath(), vm.bundle.GetSSHKeyPath())
+	preset, err := guessPresetFromVMName(vm.name)
+	if err != nil {
+		return nil, err
+	}
+	return ssh.CreateRunner(ip, vm.SSHPort(), constants.GetPrivateKeyPath(preset), constants.GetRsaPrivateKeyPath(), vm.bundle.GetSSHKeyPath())
```
cfergeau: `vm.bundle.GetBundleType()` will tell you which preset to use.

pkg/crc/constants/constants.go
```diff
@@ -121,6 +120,13 @@ func InstanceName(preset crcpreset.Preset) string {
 	return fmt.Sprintf("crc-%s", preset.String())
 }
 
+func GetKubeconfigFilePath(preset crcpreset.Preset) string {
+	if preset != crcpreset.Podman {
+		return filepath.Join(MachineInstancesDir, InstanceDirName(preset), "kubeconfig")
+	}
+	return ""
```
cfergeau: podman was not special cased before, why the change? Should this go to a commit of its own maybe?

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: This commit is also doing a lot more than what one might think by reading the commit log.

pkg/crc/api/api_http.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func NewMux(config *crcConfig.Config, machine machine.Client, logger Logger, telemetry Telemetry) http.Handler {
-	handler := NewHandler(config, machine, logger, telemetry)
+	handler := NewHandler(config, machine, logger, telemetry, MachineClientUpdater)
```
cfergeau, line 12: This commit seems to be about supporting multiple VMs running at the same time? As in the previous commits, I'd prefer to keep this kind of changes out of this PR, there's already a lot to get right without adding this kind of complication.

test/integration/resize_test.go
```diff
@@ -49,20 +50,20 @@ var _ = Describe("vary VM parameters: memory cpus, disk", Label("openshift-prese
 		})
 
 		It("check VM's memory size", func() {
-			out, err := SendCommandToVM("cat /proc/meminfo")
+			out, err := SendCommandToVM("cat /proc/meminfo", preset.OpenShift)
```
cfergeau: `SendCommandToVM` does not really need a preset arg as it's hardcoded to `OpenShift`. In my opinion, it would make more sense to add a constant equal to `preset.OpenShift` and use this instead. OpenShift will be hardcoded in a single place, and won't be hardcoded in SendCommandToVM, which I think is what you want to avoid.

cfergeau (commented):
Finally a first round of comments! Overall looks good to me, though I'd prefer that for now we focus on the renaming, having VMs or instance dir being named `crc-$preset` instead of `crc`, and make sure this works fine. Having multiple instance dirs at once, or multiple VMs running at the same time should be done in other PRs, as this adds even more complication.

Regarding the current PR, does it work to create a microshift VM with a released version of crc, stop it, upgrade to a version of crc with the changes from this PR, and then start it again?

What about `crc start; crc config set preset microshift; crc start`? Does it still have the same behaviour as before the PR? I haven't tried to run your changes yet.

pkg/crc/preflight/preflight_checks_linux.go
anjannath: > I'd prefer to see this commit split in 2 parts, one adjusting for the addition of the `preset` arg to the `constants.InstanceName()` function, and another one adding the iteration over multiple directories. 

that'd be tricky, as preflight checks don't have a way to know the current preset, we could pass the `config.Storage` to each preflight check.. but i added the iteration over machine dirs part to avoid this

> Not sure we want the iteration over multiple dirs right now.

if we don't allow more then one preset to run then we don't need to delete all the machine dirs right now, but this was a workaround for the above problem as well, this way we don't have to change the signature of `checkFunc` and since we'll need it later when having multiple presets running, i'd prefer it to add this now itself

pkg/crc/machine/virtualmachine.go
anjannath: dropped, this was only needed in one place, and https://github.com/crc-org/crc/pull/3750/commits/3517405b7c03a55012adcadd06eb4a83b13ebaf1#r1317265025 solved that, thanks!

cfergeau:
I've rebased this.

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: > that'd be tricky, as preflight checks don't have a way to know the current preset, we could pass the config.Storage to each preflight check.. 

There are already a few places where this is done. I'm not a huge fan of how it's done, but this works, see https://github.com/cfergeau/crc/commit/aec4e14c43f0314a33218feb4f909a4ac188d838

pkg/crc/constants/constants.go
anjannath: podman doesn't require this helper, i've removed the check and kept it like it was before

test/integration/resize_test.go
anjannath: `SendCommandToVM` is generic, it needs the preset to call `constants.InstanceName()` to get the VM name, and get the connection details for it, it can be used for the microshift preset with `SendCommandToVM("cat /proc/meminfo", preset.MicroShift)`.

i am not sure if it'd make more sense to use something like:
```
preset := preset.OpenShift
out, err := SendCommandToVM("cat /proc/meminfo", preset)
```
or did i completely misunderstood the comment?

anjannath: i see since this preset is being used in all the invocation, it does make sense to use the above ^^

pkg/crc/api/api_http.go
anjannath: > This commit seems to be about supporting multiple VMs running at the same time?

this is needed just for having different VM names, not necessarily enabling them to run at the same time.

since we pass the machine.Client to this function, it is tied to one _VM name_, but when the preset is changed, we need a new machine.Client with the new preset's VM name, earlier when all the VMs were named `crc` this was not needed

anjannath:
> Regarding the current PR, does it work to create a microshift VM with a released version of crc, stop it, upgrade to a version of crc with the changes from this PR, and then start it again?
> 
no, since the older VM will be named as `crc` and the upgraded crc will look for a VM named `crc-microshift` it'll think machine does not exists

> What about `crc start; crc config set preset microshift; crc start`? Does it still have the same behaviour as before the PR? I haven't tried to run your changes yet.

No for this as well, it will  not warn the user that an existing VM with different preset already exists like it does now

anjannath:
closing as it is hard to keep the old behavior and still allow the VMs to be named after the presets

e.g
```
$ crc start (default preset)
$ crc config set preset microshift (this'll ask user to delete the old preset and start a new one)
$ crc delete (this will try to delete the non-existent microshift preset)
```

cfergeau:
For this specific usecase, we could block preset change if there's an existing VM, and improve teh behaviour later on

PR #3753, opened 2023-07-11
remove all the log files when running `crc delete --clear-cache`

anjannath:
currently after https://github.com/crc-org/crc/pull/3720 was merged it is only deleting the
'crc.log' and 'crcd.log' files and not deleting the backups


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -154,15 +155,13 @@ func removeCRCMachinesDir() error {
...
+		return fmt.Errorf("Failed to remove old log files: %w", err)
 	}
-	for _, f := range logFiles {
-		logging.Debugf("Deleting %s log file", f)
-		if err := os.RemoveAll(f); err != nil {
-			return fmt.Errorf("Failed to delete %s: %w", f, err)
-		}
+	// remove daemon old log files
+	if err := crcos.RemoveFileGlob(filepath.Join(constants.CrcBaseDir, "crcd-*.log")); err != nil {
+		return fmt.Errorf("Failed to remove old log files: %w", err)
```
praveenkumar, line 161: Shouldn't `err := crcos.RemoveFileGlob(filepath.Join(constants.CrcBaseDir, "*.log"))` be enough for delete `crcd-` and `crc-` log files.

anjannath: with this we want to only delete the "old" log files, that were backed up, and not delete the `crc.log` and `crcd.log` files

the backup log files are named like `crc-xxxx-xxx-xxx.log` and `crcd-xxxx-xxxx-xxxx.log`

praveenkumar: @anjannath yes but my point is when we are doing `crc cleanup` then shouldn't we also delete `crc.log` and `crcd.log` or only want to delete the backups?

anjannath: well this was the existing behavior, but yeah i think we can delete these two files as well, if a user is running `cleanup` most probably next they will start a new instance

praveenkumar: @anjannath if the cleanup process also need logging then in the windows cleanup might fail because `crc.log` and `crcd.log` file might still be in use. Let's test this PR in the windows to make sure it doesn't break.

anjannath: hmm, will test on windows, might be one of the reason why only backup log files were removed and not the `crc.log` and `crcd.log` files, `crcd.log` is still a bit safer as the daemon will be killed before the log file removal i think

praveenkumar:
/hold

Till it is tested on windows.

pkg/crc/preflight/preflight_checks_common.go
anjannath: @praveenkumar So tested on windows and you were right, it does throw an error "unable to delete those files, as it is used by another process..." i think we have to revert to only deleting the backed up log files, ~or somehow make this cleanup action run at the very end of the cleanup~ (preflight/cleanup itself wants to write logs)

anjannath: @praveenkumar reverted to removing only the backup log files, to prevent throwing the "file in use" error on windows

cmd/crc/cmd/delete.go
```diff
@@ -5,12 +5,14 @@ import (
...
 	"io"
 	"os"
+	"path/filepath"
 
 	"github.com/crc-org/crc/pkg/crc/constants"
 	crcErrors "github.com/crc-org/crc/pkg/crc/errors"
 	"github.com/crc-org/crc/pkg/crc/input"
 	"github.com/crc-org/crc/pkg/crc/logging"
 	"github.com/crc-org/crc/pkg/crc/machine"
+	crcOs "github.com/crc-org/crc/pkg/os"
```
praveenkumar: `crcOs` looks bit wired let's use  `crcos` which we are using in other packages also ie.
https://github.com/crc-org/crc/blob/da66c4e43c4dfac8878b5617aa43e7eaf955330f/pkg/crc/adminhelper/hosts_unix.go

test/extended/util/proxy.go
```diff
@@ -55,7 +55,7 @@ func RunProxy() {
 	proxy := goproxy.NewProxyHttpServer()
 	proxy.OnRequest(goproxy.ReqHostMatches(regexp.MustCompile("^.*$"))).HandleConnect(goproxy.AlwaysMitm)
 
-	logfile := filepath.Join("test", "integration", "out", "goproxylogfile.log")
+	logfile := filepath.Join("out", "goproxylogfile.log")
```
jsliacan: :+1:

anjannath: cherry-picked this into the bundle update PR #3767

anjannath:
@praveenkumar tested on windows as well now, and the behavior is modified to delete all log files on `crc cleanup` and only delete the backed up log files on `crc delete --clear-cache`

pkg/os/util.go
```diff
@@ -113,3 +113,17 @@ func RemoveFileIfExists(path string) error {
 func RunningUsingSSH() bool {
 	return os.Getenv("SSH_TTY") != ""
 }
+
+// RemoveFileGlob takes a glob pattern as string and removes the files that matches
```
cfergeau: `the files and directories that matches` (because of the use of `RemoveAll`)

cfergeau (approved):
Ack from me, just one minor remark regarding one of the comments.

pkg/os/util.go
gbraad: @cfergeau @praveenkumar after this change we can merge, right?

@anjannath ^^^

anjannath: updated

PR #3755, opened 2023-07-13
Set emergency login for core user with random password

praveenkumar:
This PR allow user to set the emergency login for core user with a random generated password. It is helpful when user lost the ssh access and use VMM to debug the issue which require password for `core` user.

Generated password is printed on stdout if user use cli to start the VM and also part of `~/.crc/crc.log` in case want to access it later. In case of VM is started by api endpoint then password would be part of `~/.crc.crcd.log`.

```
$ crc config set enable-emergency-login true
$ crc config view
- enable-emergency-login                : true
[...]
$ crc start
[...]
INFO CRC VM is running
INFO Emergency login password for core user is: Yf9OwP7y
[...]
```

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -32,6 +33,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/telemetry"
 	crctls "github.com/crc-org/crc/pkg/crc/tls"
 	"github.com/crc-org/crc/pkg/libmachine/host"
+	crcos "github.com/crc-org/crc/pkg/os"
 	"github.com/crc-org/machine/libmachine/drivers"
 	libmachinestate "github.com/crc-org/machine/libmachine/state"
 	"github.com/docker/go-units"
@@ -417,6 +419,16 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 	}
 	logging.Info("CRC VM is running")
 
+	if startConfig.EmergencyLogin {
+		if err := enableEmergencyLogin(sshRunner); err != nil {
+			return nil, errors.Wrap(err, "Error enabling emergency login")
+		}
+	} else {
+		if err := disableEmergencyLogin(sshRunner); err != nil {
+			return nil, errors.Wrap(err, "Error deleting the password for core user")
+		}
+	}
+
 	// Post VM start immediately update SSH key and copy kubeconfig to instance
 	// dir and VM
 	if err := updateSSHKeyPair(sshRunner); err != nil {
@@ -782,6 +794,29 @@ func addNameServerToInstance(sshRunner *crcssh.Runner, ns string) error {
 		return network.AddNameserversToInstance(sshRunner, nameservers)
 	}
 	return nil
+}
+
+func enableEmergencyLogin(sshRunner *crcssh.Runner) error {
+	if crcos.FileExists(constants.PasswdFilePath) {
+		return nil
+	}
+	charset := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	b := make([]byte, 8)
+	for i := range b {
+		b[i] = charset[rand.Intn(len(charset))] //nolint
+	}
+	if err := os.WriteFile(constants.PasswdFilePath, b, 0600); err != nil {
+		return err
+	}
+	logging.Infof("Emergency login password for core user is stored to %s", constants.PasswdFilePath)
+	_, _, err := sshRunner.Run(fmt.Sprintf("sudo passwd core --unlock && echo %s | sudo passwd core --stdin", b))
+	return err
+}
+
+func disableEmergencyLogin(sshRunner *crcssh.Runner) error {
+	defer os.Remove(constants.PasswdFilePath)
+	_, _, err := sshRunner.RunPrivileged("disable core user password", "passwd", "--lock", "core")
+	return err
 }
 
 func updateSSHKeyPair(sshRunner *crcssh.Runner) error {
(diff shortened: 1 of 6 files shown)


praveenkumar:
This superseed #3689 and incorporate the suggestion in that PR.

gbraad:
> In case of VM is started by api endpoint then password would be part of ~/.crc.crcd.log.

Wouldn't this 'rotate' away ?
In case of desktop usage, you will never see this prompt. And the likelihood logs rotate this away before you need it is quite high

better leave it in a dedicated file or part of the instance crc.json

pkg/crc/machine/start.go
```diff
@@ -784,6 +792,17 @@ func addNameServerToInstance(sshRunner *crcssh.Runner, ns string) error {
 	return nil
 }
 
+func enableEmergencyLogin(sshRunner *crcssh.Runner) error {
+	charset := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	b := make([]byte, 8)
+	for i := range b {
+		b[i] = charset[rand.Intn(len(charset))] //nolint
+	}
+	logging.Infof("Emergency login password for core user is: %s", b)
```
gbraad: This will disappear from the log file when it is gets 'cleaned'/rotated.

praveenkumar: @gbraad updated.

pkg/crc/machine/start.go
```diff
@@ -784,6 +792,20 @@ func addNameServerToInstance(sshRunner *crcssh.Runner, ns string) error {
 	return nil
 }
 
+func enableEmergencyLogin(sshRunner *crcssh.Runner) error {
+	charset := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	b := make([]byte, 8)
+	for i := range b {
+		b[i] = charset[rand.Intn(len(charset))] //nolint
+	}
+	if err := os.WriteFile(constants.PasswdFilePath, b, 0600); err != nil {
```
gbraad, line 808: remove on stop and when the setting has been disabled

gbraad (changes requested):

pkg/crc/machine/start.go
praveenkumar: @gbraad On the stop we shouldnt' remove it otherwise we need to also remove `core` user password on stop, IMO we just need to delete it during `crc delete` not `crc stop` action.

gbraad: wouldn't it get a new password on restart?

praveenkumar: I reiterated and now when `stop => start` happen and user unset `enable-emergency-login` setting then `core` user password is locked and from shell not able to login.

cfergeau: We don't regenerate the ssh key on start/stop/start, I don't think the user password needs to behave differently, so as long as `enable-emergency-login` is set, we can keep the same password until the VM is deleted.

pkg/crc/machine/start.go
```diff
@@ -784,6 +795,25 @@ func addNameServerToInstance(sshRunner *crcssh.Runner, ns string) error {
...
+		return err
+	}
+	logging.Infof("Emergency login password for core user is stored to %s", constants.PasswdFilePath)
+	_, _, err := sshRunner.Run(fmt.Sprintf("sudo passwd core --unlock && echo %s | sudo passwd core --stdin", b))
+	return err
+}
+
+func lockCoreUserPasswd(sshRunner *crcssh.Runner) error {
+	_, _, err := sshRunner.RunPrivileged("delete core user password", "passwd", "--lock", "core")
```
cfergeau: I'd name this `disableEmergencyLogin` for symmetry with `enableEmergencyLogin`, and the `RunPrivileged` string could be `disable core user password`

pkg/crc/machine/start.go
```diff
@@ -417,6 +418,16 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 	}
 	logging.Info("CRC VM is running")
 
+	if err := lockCoreUserPasswd(sshRunner); err != nil {
+		return nil, errors.Wrap(err, "Error deleting the password for core user")
+	}
+	// VM started and SSH available, so we can enable the emergency login
```
cfergeau: I don't think this comment is needed.

pkg/crc/machine/start.go
```diff
@@ -417,6 +418,16 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 	}
 	logging.Info("CRC VM is running")
 
+	if err := lockCoreUserPasswd(sshRunner); err != nil {
+		return nil, errors.Wrap(err, "Error deleting the password for core user")
```
cfergeau: Should this be moved to an `else` in `if startConfig.EmergencyLogin {` as both operations exclusive

pkg/crc/machine/start.go
```diff
@@ -784,6 +795,25 @@ func addNameServerToInstance(sshRunner *crcssh.Runner, ns string) error {
 	return nil
 }
 
+func enableEmergencyLogin(sshRunner *crcssh.Runner) error {
```
cfergeau, line 799: I'd start with 
```
if crcos.FileExists(constants.PasswdFilePath) {
    return nil
}
```

and add a `os.Remove(constants.PasswdFilePath)` to `disableEmergencyLogin`.

This would ensure we don't generate a different password each time we do `crc start; crc stop; crc start`. This makes the behaviour consistent with the ssh key, which we create only at VM creation time, not for each start/stop/start cycle.

cfergeau (changes requested):
Apart from https://github.com/crc-org/crc/pull/3755/commits/da66c4e43c4dfac8878b5617aa43e7eaf955330f#r1277594992 this looks good to me.

pkg/crc/machine/start.go
gbraad: This 'recreate'-behaviour I also did not understand.

pkg/crc/machine/start.go
gbraad: I also don't think this should happen, but it is the behaviour that existed in the PR: "wouldn't it get a new password on restart?" was not something I implied being happy with, but rather an observation of the current suggestion.

pkg/crc/machine/start.go
praveenkumar: updated

PR #3762, opened 2023-07-19
Update go modules crc depends on

cfergeau:
This PR updates most go modules crc depends on
- vfkit is updated to its latest version ~(I haven't been able to test this though...)~
- openshift/k8s are updated to match what is used in the release-4.13 branch of `oc`
- golangci-lint is updated as well, which causes most of the changes in this PR


praveenkumar:
Shouldn't all the lint fixes part of single commit instead every file change become a separate commit, because if I checkout one of lint fix commit then linter fails because it doesn't have all the lint fixes. https://github.com/crc-org/crc/pull/3762/commits/ed22b76fd73ba59439250810b42714eb152cebd4 to https://github.com/crc-org/crc/pull/3762/commits/18368ae247f0e9df59419b5bd67a73ebfe2166a7 squashed into single commit which is just fixing linter due to update of it version.

cfergeau:
> Shouldn't all the lint fixes part of single commit instead every file change become a separate commit, because if I checkout one of lint fix commit then linter fails because it doesn't have all the lint fixes. [ed22b76](https://github.com/crc-org/crc/commit/ed22b76fd73ba59439250810b42714eb152cebd4) to [18368ae](https://github.com/crc-org/crc/commit/18368ae247f0e9df59419b5bd67a73ebfe2166a7) squashed into single commit which is just fixing linter due to update of it version.

If we want individual commits to be buildable, then I can move the tools go module update (aka the golangci-lint version update) after all the lint fixes. Even if they are all in a single commit, there would be one point in time (the golangci-lint version update commit) when the code would not be buildable.

cfergeau:
> - vfkit is updated to its latest version (I haven't been able to test this though...)

I've now started a microshift bundle with these changes, and checked that the vfkit binary used is indeed 0.1.1

gbraad:
> If we want individual commits to be buildable

That should always be strived for.

While this is not conclusive, it is a good practice to read here 'commit' and not PR.
https://martinfowler.com/articles/continuousIntegration.html#EveryCommitShouldBuildTheMainlineOnAnIntegrationMachine
as a PR only lives as a concept on GitHub, but when I check out I have no idea what belonged together, unless we perform branch merges instead of a rebase.

cfergeau:
> > If we want individual commits to be buildable
> 
> That should always be strived for.
> 
> While this is not conclusive, it is a good practice to read here 'commit' and not PR. https://martinfowler.com/articles/continuousIntegration.html#EveryCommitShouldBuildTheMainlineOnAnIntegrationMachine as a PR only lives as a concept on GitHub, but when I check out I have no idea what belonged together, unless we perform branch merges instead of a rebase.

I've reordered the commits in the PR a while ago so that the golangci-lint change comes towards the end.

PR #3763, opened 2023-07-20
Revert "Revert "Deploy router controller resource from bundle""

praveenkumar:
This reverts commit 567ca08207891650e2cf9e8b2aeffc40d44c2850.

Since now we updated the routes-controller resource with correct service account so it is now good to bring that feature back.

Also file is changed `routes-controller.yaml` as we have in the bundle.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -912,6 +912,13 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
 }
 
 func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Config, preset crcPreset.Preset, bundleInfo *bundle.CrcBundleInfo) error {
+	// Check if the bundle have `/opt/crc/routes-controller.yaml` file and if it has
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/start.go
```diff
@@ -912,6 +912,13 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
 }
 
 func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Config, preset crcPreset.Preset, bundleInfo *bundle.CrcBundleInfo) error {
+	// Check if the bundle have `/opt/crc/routes-controller.yaml` file and if it has
+	// then use it to create the resource for router controller.
```
cfergeau: "router controller" -> "routes controler". I'd add an article as well "for the routes controller".
Apart from this, looks good to me.

praveenkumar:
/hold

We will push this after the release cut.

pkg/crc/machine/start.go
gbraad: This was also remarked in the earlier PR; inconsistency in the use of routes and router.

PR #3765, opened 2023-07-21
hyperv: Replace parseLines/SplitLines with getFirstLine

### LINKED ISSUE #3764: Error for Hyper-V available check when management tools not installed
The line:
https://github.com/crc-org/crc/blob/e43de8dfca711be4f76acef3d07b67ea121cbe1f/pkg/drivers/hyperv/powershell_windows.go#L47

will fail when Management tools aren't installed:

```
INFO Creating CRC VM for OpenShift 4.13.3...
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/crc-org/crc/pkg/drivers/hyperv.hypervAvailable()
        /builddir/build/BUILD/crc-2.23.0/pkg/drivers/hyperv/powershell_windows.go:47 +0xb4
github.com/crc-org/crc/pkg/drivers/hyperv.(*Driver).PreCreateCheck(0xc0005edef0)
        /builddir/build/BUILD/crc-2.23.0/pkg/drivers/hyperv/hyperv_windows.go:119 +0x3c
github.com/crc-org/crc/pkg/crc/machine.createHost({{0xc000bb0d9a, 0x17}, {0x2320065, 0x3}, 0x2400, 0x4, 0x1f, {0xc0004f4940, 0x3a}, {0xc000dc1398, ...}, ...}, ...)
        /builddir/build/BUILD/crc-2.23.0/pkg/crc/machine/start.go:723 +0x19c
github.com/crc-org/crc/pkg/crc/machine.(*client).Start(0xc000daa5c0, {0x26cc6d8, 0xc000daad40}, {{0xc000bb0d80, 0x3b}, 0x2400, 0x4, 0x1f, {0x0, 0x0}, ...})
        /builddir/build/BUILD/crc-2.23.0/pkg/crc/machine/start.go:329 +0x9d8
github.com/crc-org/crc/pkg/crc/machine.(*Synchronized).Start(0xc000daa600, {0x26cc780?, 0xc000c805a0?}, {{0xc000bb0d80, 0x3b}, 0x2400, 0x4, 0x1f, {0x0, 0x0}, ...})
        /builddir/build/BUILD/crc-2.23.0/pkg/crc/machine/sync.go:93 +0xca
github.com/crc-org/crc/cmd/crc/cmd.runStart({0x26cc780, 0xc000c805a0})
        /builddir/build/BUILD/crc-2.23.0/cmd/crc/cmd/start.go:111 +0x5ad
github.com/crc-org/crc/cmd/crc/cmd.glob..func11(0x3751c20, {0x5?, 0x2336a0c?, 0xc?})
        /builddir/build/BUILD/crc-2.23.0/cmd/crc/cmd/start.go:57 +0x57
github.com/crc-org/crc/cmd/crc/cmd.executeWithLogging.func1(0x3751c20, {0x3cfc9f0, 0x0, 0x0})
        /builddir/build/BUILD/crc-2.23.0/cmd/crc/cmd/root.go:170 +0x102
github.com/spf13/cobra.(*Command).execute(0x3751c20, {0x3cfc9f0, 0x0, 0x0})
        /builddir/build/BUILD/crc-2.23.0/vendor/github.com/spf13/cobra/command.go:916 +0x862
github.com/spf13/cobra.(*Command).ExecuteC(0x3753600)
        /builddir/build/BUILD/crc-2.23.0/vendor/github.com/spf13/cobra/command.go:1044 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
        /builddir/build/BUILD/crc-2.23.0/vendor/github.com/spf13/cobra/command.go:968
github.com/spf13/cobra.(*Command).ExecuteContext(...)
        /builddir/build/BUILD/crc-2.23.0/vendor/github.com/spf13/cobra/command.go:961
github.com/crc-org/crc/cmd/crc/cmd.Execute()
        /builddir/build/BUILD/crc-2.23.0/cmd/crc/cmd/root.go:105 +0xd9
main.main()
        /builddir/build/BUILD/crc-2.23.0/cmd/crc/main.go:8 +0x17
```

cfergeau:
In the hyperv code, there are many occurrences of resp :=
os.SplitLines(stdout) if resp[0] == ...

This does not check if `resp` is empty before trying to reference it, which
recently triggered a panic in crc ( 
https://github.com/crc-org/crc/issues/3764 )

This code adds a `getFirstLine` helper which will return an empty string if
stdout is "".

This should fix https://github.com/crc-org/crc/issues/3764

### DIFF
--- a/pkg/crc/cluster/clusteroperator.go
+++ b/pkg/crc/cluster/clusteroperator.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/crc-org/crc/pkg/crc/logging"
+	crcstrings "github.com/crc-org/crc/pkg/strings"
 	openshiftapi "github.com/openshift/api/config/v1"
 	k8sapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -101,7 +102,7 @@ func getStatus(ctx context.Context, lister operatorLister, selector []string) (*
 
 	found := false
 	for _, c := range co.Items {
-		if len(selector) > 0 && !contains(c.ObjectMeta.Name, selector) {
+		if len(selector) > 0 && !crcstrings.Contains(selector, c.ObjectMeta.Name) {
 			continue
 		}
 		found = true
@@ -176,15 +177,6 @@ func GetClusterNodeStatus(ctx context.Context, ip string, kubeconfigFilePath str
 		logging.Debugf("Unexpected node status for %s", ns)
 	}
 	return status, nil
-}
-
-func contains(value string, list []string) bool {
-	for _, v := range list {
-		if v == value {
-			return true
-		}
-	}
-	return false
 }
 
 type operatorLister interface {
--- a/pkg/crc/services/dns/dns_windows.go
+++ b/pkg/crc/services/dns/dns_windows.go
@@ -11,6 +11,7 @@ import (
 	winnet "github.com/crc-org/crc/pkg/os/windows/network"
 	"github.com/crc-org/crc/pkg/os/windows/powershell"
 	"github.com/crc-org/crc/pkg/os/windows/win32"
+	crcstrings "github.com/crc-org/crc/pkg/strings"
 )
 
 const (
@@ -30,7 +31,7 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig) error {
 
 	time.Sleep(2 * time.Second)
 
-	if !contains(getInterfaceNameserverValues(networkInterface), serviceConfig.IP) {
+	if !crcstrings.Contains(getInterfaceNameserverValues(networkInterface), serviceConfig.IP) {
 		return fmt.Errorf("Nameserver %s not successfully set on interface %s. Perhaps you can try this new network mode: https://github.com/crc-org/crc/wiki/VPN-support--with-an--userland-network-stack", serviceConfig.IP, networkInterface)
 	}
 	return nil
@@ -41,15 +42,6 @@ func getInterfaceNameserverValues(iface string) []string {
 	stdOut, _, _ := powershell.Execute(getDNSServerCommand)
 
 	return parseLines(stdOut)
-}
-
-func contains(s []string, e string) bool {
-	for _, v := range s {
-		if v == e {
-			return true
-		}
-	}
-	return false
 }
 
 func setInterfaceNameserverValue(iface string, address string) {
--- a/pkg/crc/services/dns/dns_windows.go
+++ b/pkg/crc/services/dns/dns_windows.go
@@ -1,9 +1,7 @@
 package dns
 
 import (
-	"bufio"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/crc-org/crc/pkg/crc/network"
@@ -41,7 +39,7 @@ func getInterfaceNameserverValues(iface string) []string {
 	getDNSServerCommand := fmt.Sprintf(`(Get-DnsClientServerAddress "%s")[0].ServerAddresses`, iface)
 	stdOut, _, _ := powershell.Execute(getDNSServerCommand)
 
-	return parseLines(stdOut)
+	return crcstrings.SplitLines(stdOut)
 }
 
 func setInterfaceNameserverValue(iface string, address string) {
@@ -50,15 +48,4 @@ func setInterfaceNameserverValue(iface string, address string) {
 
 	// ignore the error as this is useless (prefer not to use nolint here)
 	_ = win32.ShellExecuteAsAdmin(fmt.Sprintf("add dns server address to interface %s", iface), win32.HwndDesktop, exe, args, "", 0)
-}
-
-func parseLines(input string) []string {
-	output := []string{}
-
-	s := bufio.NewScanner(strings.NewReader(input))
-	for s.Scan() {
-		output = append(output, s.Text())
-	}
-
-	return output
 }
--- a/pkg/drivers/hyperv/powershell_windows.go
+++ b/pkg/drivers/hyperv/powershell_windows.go
@@ -31,8 +31,8 @@ func hypervAvailable() error {
 		return err
 	}
 
-	resp := crcstrings.SplitLines(stdout)
-	if resp[0] != "Hyper-V" {
+	resp := crcstrings.FirstLine(stdout)
+	if resp != "Hyper-V" {
 		return ErrNotInstalled
 	}
 
@@ -62,8 +62,8 @@ func isHypervAdministrator() bool {
 		return false
 	}
 
-	resp := crcstrings.SplitLines(stdout)
-	return resp[0] == "True"
+	resp := crcstrings.FirstLine(stdout)
+	return resp == "True"
 }
 
 func isWindowsAdministrator() (bool, error) {
@@ -72,8 +72,8 @@ func isWindowsAdministrator() (bool, error) {
 		return false, err
 	}
 
-	resp := crcstrings.SplitLines(stdout)
-	return resp[0] == "True", nil
+	resp := crcstrings.FirstLine(stdout)
+	return resp == "True", nil
 }
 
 func quote(text string) string {
--- a/pkg/drivers/hyperv/powershell_windows.go
+++ b/pkg/drivers/hyperv/powershell_windows.go
@@ -1,13 +1,12 @@
 package hyperv
 
 import (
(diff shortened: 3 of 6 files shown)


pkg/crc/cluster/clusteroperator.go
```diff
@@ -101,7 +102,7 @@ func getStatus(ctx context.Context, lister operatorLister, selector []string) (*
 
 	found := false
 	for _, c := range co.Items {
-		if len(selector) > 0 && !contains(c.ObjectMeta.Name, selector) {
+		if len(selector) > 0 && !crcos.Contains(selector, c.ObjectMeta.Name) {
```
gbraad: Why do we invert the selector and name?

pkg/drivers/hyperv/powershell_windows.go
```diff
@@ -26,15 +25,13 @@ func cmd(args ...string) error {
 	return err
 }
 
-func parseLines(stdout string) []string {
-	resp := []string{}
-
-	s := bufio.NewScanner(strings.NewReader(stdout))
-	for s.Scan() {
-		resp = append(resp, s.Text())
+func getFirstLine(stdout string) string {
```
gbraad: Not part of the util.go ?

pkg/os/util.go
```diff
@@ -113,3 +114,24 @@ func RemoveFileIfExists(path string) error {
 func RunningUsingSSH() bool {
 	return os.Getenv("SSH_TTY") != ""
 }
+
+// Split a multi line string in an array of string, one for each line
+func SplitLines(input string) []string {
```
gbraad: why is this part of crcos (operating system-specific code). this looks like generic line/string manipulation

pkg/drivers/hyperv/powershell_windows.go
cfergeau: This package is the only user of `getFirstLine` so I kept it there as it's a bit specialized.

pkg/crc/cluster/clusteroperator.go
cfergeau: The function prototype in `clusteroperator.go` was `func contains(value string, list []string) bool`, the new helper is ` Contains(s []string, e string) bool`

pkg/drivers/hyperv/powershell_windows.go
gbraad: the issue in question was the only place for `resp[0]`, for which we also use the `| get-unique` operator in powershell. I don't mind, but found it very generic.

pkg/os/util.go
cfergeau: I could not find a more appropriate package, and was not sure adding a `strings` package for 2 functions was worh it. I can make this change if you prefer.

gbraad: I think adding a strings make sense, as I might even think adding the firstline justifies the use. it makes it clearer in intent. perhaps we might even add a getlastline at some point ;-)

pkg/crc/services/dns/dns_windows.go
```diff
@@ -30,7 +29,7 @@ func runPostStartForOS(serviceConfig services.ServicePostStartConfig) error {
 
 	time.Sleep(2 * time.Second)
 
-	if !contains(getInterfaceNameserverValues(networkInterface), serviceConfig.IP) {
+	if !crcos.Contains(getInterfaceNameserverValues(networkInterface), serviceConfig.IP) {
```
gbraad: then is this not in error?

```
if !crcos.Contains(getInterfaceNameserverValues(networkInterface), serviceConfig.IP) {
```

the order for the parameters have changed? https://github.com/crc-org/crc/pull/3765/files#r1270651088 Let's stay consistent.

pkg/crc/cluster/clusteroperator.go
gbraad: Thanks for confirming, which means the order has changed. Using more clear variables on the function might make this clearer; `s` is `strings`, and `e` is expected?

This is not the only usage: https://github.com/crc-org/crc/pull/3765/files#r1270660861

pkg/crc/services/dns/dns_windows.go
cfergeau: `dns_windows.go` was using `contains(s []string, e string) bool` and `clusteroperator.go` used `contains(value string, list []string) bool`,  so the order will change in one of the instances.

gbraad: Check

Can we in that case also more clearly describe the parameters:

```
func Contains(s []string, e string) bool {
```

with actual 'words' instead of abbreviated identifiers. https://github.com/crc-org/crc/pull/3765#discussion_r1270661563

`s` = `strings` and `e` = `expected`?

pkg/drivers/hyperv/powershell_windows.go
cfergeau: ok I moved it.

pkg/os/util.go
cfergeau: Done.

pkg/strings/strings.go
```diff
@@ -0,0 +1,36 @@
...
+
+	s := bufio.NewScanner(strings.NewReader(input))
+	for s.Scan() {
+		output = append(output, s.Text())
+	}
+
+	return output
+}
+
+func FirstLine(stdout string) string {
```
gbraad: `s/stdOut/input/`, just like in `SplitLines`. `stdOut` does not have any meaning here, except from the caller.

pkg/strings/strings.go
```diff
@@ -0,0 +1,36 @@
+package strings
+
+import (
+	"bufio"
+	"strings"
+)
+
+func Contains(s []string, e string) bool {
```
gbraad: `s/s/strings/`, `s/e/expected/` or `compare`. other functions use more descriptive parameters, like `input`.

gbraad (changes requested):
parameters are not clear/consistent.

otherwise lgtm

pkg/strings/strings.go
cfergeau: I used needle/haystack, I can change again if this is harder to read than it should.

cfergeau:
Let's hold this for a little bit, I'll add some unit tests.

cfergeau:
I've now added unit tests, and they made me wonder if it would not be best to change `SplitLines("")` to return `{ "" }` instead of `{ }`, as at the moment it's a bit weird that `SplitLines("a\n")` returns `{ "a" }`, `SplitLines("a")` also returns `{ "a" }`, but we don't have the same behaviour if we use an empty string instead of `"a"`

gbraad:
why haystack and needle? stay consistent with the rest and use `input` and somehing like `match`. As in a `findNeedle`-like function I expect the identifier to be returned of the `needle` that matches, not just a `bool`-value.

pkg/strings/strings.go
```diff
@@ -0,0 +1,36 @@
+package strings
+
+import (
+	"bufio"
+	"strings"
+)
+
+func Contains(haystack []string, needle string) bool {
```
gbraad: inconsistent with other naming; also, this does not work like a `FindNeedle`-function.

gbraad (changes requested):

anjannath:
> I've now added unit tests, and they made me wonder if it would not be best to change `SplitLines("")` to return `{ "" }` instead of `{ }`, as at the moment it's a bit weird that `SplitLines("a\n")` returns `{ "a" }`, `SplitLines("a")` also returns `{ "a" }`, but we don't have the same behaviour if we use an empty string instead of `"a"`

with just `"\n"` its an empty line but with `""` there's no line, so maybe not returning `{ "" }` is better or we could decide to treat empty lines as no line and return `{}` for `"\n"` this will make it consistent

gbraad (approved):

cfergeau:
> > I've now added unit tests, and they made me wonder if it would not be best to change `SplitLines("")` to return `{ "" }` instead of `{ }`, as at the moment it's a bit weird that `SplitLines("a\n")` returns `{ "a" }`, `SplitLines("a")` also returns `{ "a" }`, but we don't have the same behaviour if we use an empty string instead of `"a"`
> 
> with just `"\n"` its an empty line but with `""` there's no line, so maybe not returning `{ "" }` is better or we could decide to treat empty lines as no line and return `{}` for `"\n"` this will make it consistent

The initial bug is that `{}` is not handled correctly by our code, if we never return `{}`, there is no bug :)
`{}` for `"\n'` would not be that consistent, as `"\n\n"` returns `{ "", ""}`

anjannath (approved):

PR #3775, opened 2023-07-28
daemon: start daemon from a powershell script that hides window

### LINKED ISSUE #3726: [BUG] Somethings launches a powershell terminal on Windows that stays around

anjannath:
after windows terminal became the default terminal/console host on windows the daemon scheduled task would show a console window upon running as it wasn't respecting the poweshell cmdline argument for hiding the window

this powershell script hides the window by using win32 api from powershell

fixes #3726

### DIFF
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"os"
 	"os/user"
+	"path/filepath"
 	"strings"
 
 	"github.com/crc-org/crc/pkg/crc/constants"
 	"github.com/crc-org/crc/pkg/crc/logging"
 	"github.com/crc-org/crc/pkg/crc/version"
+	crcos "github.com/crc-org/crc/pkg/os"
 	"github.com/crc-org/crc/pkg/os/windows/powershell"
 )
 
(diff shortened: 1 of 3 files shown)


cfergeau:
I think you tried this kind of approaches https://stackoverflow.com/questions/23250505/how-do-i-create-an-executable-from-golang-that-doesnt-open-a-console-window-whe to avoid showing the window? 
https://github.com/crc-org/crc/commit/047873f671710ec263335a0decf5e91a4a59fefb seems to imply this isnot an option as we cannot run the binary directly from the task XMl, we need an intermediate powershell instance to start it?

gbraad:
I proposed something similar as Christophe; make the executable be 'regarded' as a GUI application, but not actually create the window.

Can this be tried?

The issue with the script is that this can be marked again as a problem when it comes to 'false positives' on scans.

anjannath:
I tried both the approaches mentioned there -- the issue with using `-H=windowsgui` is that, once CRC is compiled with this flag, it doesn't work as a CLI anymore, even when its launched from powershell as `./crc.exe --help`, it means we have to ship two crc binaries, one for the daemon and another for the CLI. I didn't spend a lot of time looking to find a way if we can have an executable that is both GUI and CLI, but if that's possible i think we can use this method. as we can directly start it from task scheduler without having to invoke powershell

the other method using the `github.com/gonutz/w32/v2` library to hide the windows after crc is launched seems to not do anything, my guess is because CRC doesn't create any window

praveenkumar:
Did we test it works with win10 and win11 both?

gbraad:
> have an executable that is both GUI and CLI

You can't

"_Each PE application contains a field in its header that specifies which subsystem it was designed to run under. You can say IMAGE_SUBSYSTEM_WINDOWS_GUI to mark yourself as a Windows GUI application, or you can say IMAGE_SUBSYSTEM_WINDOWS_CUI to say that you are a console application. If you are GUI application, then the program will run without a console._" - [Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20090101-00/?p=19643)

... unless you do some trickery:

"_In VisualStudio case, there are actually two binaries: devenv.com and devenv.exe. Devenv.com is a Console app. Devenv.exe is a GUI app. When you type devenv, because of the Win32 probing rule, devenv.com is executed. If there is no input, devenv.com launches devenv.exe, and exits itself. If there are inputs, devenv.com handles them as normal Console app._

_In ildasm case, there is only one binary: ildasm.exe. It is first compiled as a GUI application. Later editbin.exe is used to mark it as console subsystem. In its main method it determines if it needs to be run as console mode or GUI mode. If need to run as GUI mode, it relaunches itself as a GUI app._" - [Junfeng Zhang](https://learn.microsoft.com/en-us/archive/blogs/junfeng/how-to-make-an-application-as-both-gui-and-console-application)

... so, let's conclude that it's not worth the extra effort at this point.

anjannath:
> Did we test it works with win10 and win11 both?

yes, i have tested the script hides the powershell window on both win 10 and 11, but not an end-to-end test using `crc setup`

anjannath:
> > Did we test it works with win10 and win11 both?
> 
> yes, i have tested the script hides the powershell window on both win 10 and 11, but not an end-to-end test using `crc setup`

tested on both windows 10 and windows 11 now and its working as expected, (still shows the powershell window briefly when running `crc setup` and then disappears)

the new pushes were for fixing the quotes around the `crc daemon` command, the whole command was enclosed in quotes earlier along with the args, which was found when testing end-to-end

gbraad (approved):

Issue #3773, opened 2023-07-28
[CI] Create e2e job on openshift-ci for microshift bundle testing


praveenkumar:
This job will make sure that microshift bundle is working as expected and we can also add some basic e2e feature which we use for openshift bundles.

gbraad:
I would say; start with basics and build from there.

a "start, run container, stop" would have caught our current issue.

jsliacan:
So maybe we can reuse what we have already. [This feature](https://github.com/crc-org/crc/blob/6b61c2ff0e23d92ec886bbc88205378b28cc80f3/test/e2e/features/minimal.feature) was created for similar quick checks of basic functionality. We are going to need a small modification to run *only* `microshift` case, maybe splitting the examples as below will work. Then filtering with `@minimal && @microshift-preset` should do the job.

```feature
@minimal @darwin @linux @windows
Feature: Minimal user story

    User starts CRC cluster and checks status.

    @cleanup
    Scenario Outline: Start OpenShift cluster:
        Given setting config property "preset" to value "<preset-value>" succeeds
        And executing single crc setup command succeeds
        And starting CRC with default bundle succeeds
        Then checking that CRC is running

        @podman-preset
        Examples:
            | preset-value |
            | podman       |

        @microshift-preset
        Examples:
            | preset-value |
            | microshift   |

        @openshift-preset
        Examples:
            | preset-value |
            | openshift    |
```

jsliacan:
I checked, and it would work with those modifications and that filter.

```$ make e2e
Log successfully started, logging into: /home/jsliacan/Github/crc/test/e2e/out/test-results/e2e_2023-7-28_10-30-41.log
Running e2e test in: /home/jsliacan/Github/crc/test/e2e/out/test-run
Working directory set to: /home/jsliacan/Github/crc/test/e2e/out/test-run
Using existing bundle: /home/jsliacan/Downloads/crc_microshift_libvirt_4.13.6_amd64.crcbundle
Feature: Minimal user story
  User starts CRC cluster and checks status.

    Examples:
      | preset-value |
      | microshift   |
Deleted CRC instance (if one existed).

1 scenarios (1 passed)
4 steps (4 passed)
4m9.323780413s
ok  	github.com/crc-org/crc/test/e2e	249.349s
```

gbraad:
That would be just a start scenario, right?

praveenkumar:
Job creation PR on openshift CI: https://github.com/openshift/release/pull/41777

[...]

PR #3777, opened 2023-07-31
e2e: Split minimal feature by preset

jsliacan:
Being able to restrict execution to a specific preset allows us to run the Scenario with user-provided bundles. Therefore, the test can be used e.g. for #3773. Without specifying `x-preset` tags, the test remains effectively unchanged.

To run e.g. `microshift-preset` case only with a user-provided bundle, make sure you specify the following before running `make e2e`.

```make
GODOG_OPTS = --godog.tags="@minimal && @microshift-preset"
BUNDLE_LOCATION = --bundle-location=path/to/microshift-bundle
```

`minimal` Feature is excluded from the CI checks, so here's a run I did locally.

```feature
Using existing bundle: /home/jsliacan/Downloads/crc_microshift_libvirt_4.13.6_amd64.crcbundle
Feature: Minimal user story
  User starts CRC cluster and checks status.

    Examples:
      | preset-value |
      | microshift   |
Deleted CRC instance (if one existed).

1 scenarios (1 passed)
4 steps (4 passed)
4m9.323780413s
```

### DIFF
--- a/test/e2e/features/minimal.feature
+++ b/test/e2e/features/minimal.feature
@@ -10,9 +10,17 @@ Feature: Minimal user story
         And starting CRC with default bundle succeeds
         Then checking that CRC is running
 
-        Examples:
+            @podman-preset
+            Scenarios:
             | preset-value |
             | podman       |
-            | microshift   |
-            | openshift    |
 
+            @microshift-preset
+            Scenarios:
+            | preset-value |
+            | microshift   |
+
+            @openshift-preset
+            Scenarios:
+            | preset-value |
+            | openshift    |


test/e2e/features/minimal.feature
```diff
@@ -10,9 +10,22 @@ Feature: Minimal user story
...
+        Examples:
+            | preset-value |
             | microshift   |
+
+        @openshift-preset
+        Examples:
+            | preset-value |
             | openshift    |
 
+
```
praveenkumar: why all these empty lines?

praveenkumar (changes requested):

test/e2e/features/minimal.feature
gbraad: future additions ;-)

jsliacan: Removed the lines at the end of the file. As for the rest, it's easier to read when distinct Example blocks are separated.

gbraad: We referred to the additional bottom 4 lines.

gbraad:
Why are the blocks prefixed `Examples`?

jsliacan:
@gbraad Originally it was just one Examples block for that Scenario Outline. But if I want to tag each case in the Examples separately, the only way I was able to do that was to have 3 separate Examples blocks (each with its own tag) for the same Scenario Outline.

gbraad:
I am more about the meaning of `Examples`; since this is the actual runs, does it make sense to call it `Presets`, or `Scenarios`, or something else? Examples sounds so, ... well, non-committed.

jsliacan:
`Examples` is a keyword in gherkin https://cucumber.io/docs/gherkin/reference/#scenario-outline, although it seems that I could use `Scenarios` as well. I should probably change it everywhere else though if we go with `Scenarios`.

gbraad:
Makes more sense as a keyword

test/e2e/features/minimal.feature
```diff
@@ -10,9 +10,17 @@ Feature: Minimal user story
         And starting CRC with default bundle succeeds
         Then checking that CRC is running
 
+        @podman-preset
         Examples:
             | preset-value |
             | podman       |
+
+        @microshift-preset
+        Examples:
```
gbraad: `Scenarios`

gbraad (changes requested):

test/e2e/features/minimal.feature
jsliacan: I changed it to Scenarios and tested it that way:

```feature
Using existing bundle: /home/jsliacan/Downloads/crc_microshift_libvirt_4.13.6_amd64.crcbundle
Feature: Minimal user story
  User starts CRC cluster and checks status.

    Scenarios:
      | preset-value |
      | microshift   |
Deleted CRC instance (if one existed).

1 scenarios (1 passed)
4 steps (4 passed)
```
I'll keep `Examples` in the other Feature files for now. In `config.feature` it even makes sense that way.

praveenkumar:
@jsliacan with #3779 now we have e2e for microshift ( a dedicated feature file) do you still want to have this one because this is just starting the cluster and @minimal should just test all the preset is able to run?

jsliacan:
Yeah, maybe there's less reason to run microshift part of this separately now that there's your scenario. I'll keep the branch in case we come back to it but can close the PR I guess. Thanks.

PR #3779, opened 2023-08-01
e2e: Add scenario for microshift preset

praveenkumar:
With this we can able to test out the microshift preset functionality.
```
$ BUNDLE_LOCATION='--bundle-location=/home/prkumar/.crc/cache/crc_microshift_libvirt_4.13.3_amd64.crcbundle' PULL_SECRET_FILE='--pull-secret-file=/home/prkumar/pull-secret' CRC_BINARY='--crc-binary=/home/prkumar/work/github/crc/out/linux-amd64/' make e2e-story-microshift

  Background:
    Given setting config property "preset" to value "microshift" succeeds                                                                                      # testsuite.go:860 -> github.com/crc-org/crc/test/e2e/testsuite.SetConfigPropertyToValueSucceedsOrFails
[sudo] password for prkumar: 

[...]
```

### DIFF
--- a/test/e2e/features/story_microshift.feature
+++ b/test/e2e/features/story_microshift.feature
@@ -1,0 +1,36 @@
+@story_microshift
+Feature: Microshift test stories
+
+	Background:
+		Given setting config property "preset" to value "microshift" succeeds
+		And setting config property "network-mode" to value "user" succeeds
+		And executing single crc setup command succeeds
+		And starting CRC with default bundle succeeds
+
+	# End-to-end health check
+
+	@microshift @testdata @linux @windows @darwin
+	Scenario: Start and expose a basic HTTP service and check after restart
+		Given executing "oc create namespace testproj" succeeds
+		And executing "oc config set-context --current --namespace=testproj" succeeds
+		When executing "oc apply -f httpd-example.yaml" succeeds
+		And executing "oc rollout status deployment httpd-example" succeeds
+		Then stdout should contain "successfully rolled out"
+		When executing "oc create configmap www-content --from-file=index.html=httpd-example-index.html" succeeds
+		Then stdout should contain "configmap/www-content created"
+		When executing "oc set volume deployment/httpd-example --add --type configmap --configmap-name www-content --name www --mount-path /var/www/html" succeeds
+		Then stdout should contain "deployment.apps/httpd-example volume updated"
+		When executing "oc expose deployment httpd-example --port 8080" succeeds
+		Then stdout should contain "httpd-example exposed"
+		When executing "oc expose svc httpd-example" succeeds
+		Then stdout should contain "httpd-example exposed"
+		When with up to "20" retries with wait period of "5s" http response from "http://httpd-example-testproj.apps.crc.testing" has status code "200"
+		Then executing "curl -s http://httpd-example-testproj.apps.crc.testing" succeeds
+		And stdout should contain "Hello CRC!"
+		When executing "crc stop" succeeds
+		And starting CRC with default bundle succeeds
+		And checking that CRC is running
+		And with up to "4" retries with wait period of "1m" http response from "http://httpd-example-testproj.apps.crc.testing" has status code "200"
+		Then executing "curl -s http://httpd-example-testproj.apps.crc.testing" succeeds
+		And stdout should contain "Hello CRC!"
+		Then with up to "4" retries with wait period of "1m" http response from "http://httpd-example-testproj.apps.crc.testing" has status code "200"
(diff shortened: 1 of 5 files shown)


test/e2e/features/story_microshift.feature
```diff
@@ -0,0 +1,35 @@
...
+		Then stdout should contain "deployment.apps/httpd-example volume updated"
+		When executing "oc expose deployment httpd-example --port 8080" succeeds
+		Then stdout should contain "httpd-example exposed"
+		When executing "oc expose svc httpd-example" succeeds
+		Then stdout should contain "httpd-example exposed"
+		When with up to "20" retries with wait period of "5s" http response from "http://httpd-example-testproj.apps.crc.testing" has status code "200"
+		Then executing "curl -s http://httpd-example-testproj.apps.crc.testing" succeeds
+		And stdout should contain "Hello CRC!"
+		When executing "crc stop" succeeds
+		And starting CRC with default bundle succeeds
```
gbraad, line 31: I prefer you would describe the test as:


```
Scenario: Start and expose a basic HTTP service and check after restart
```

gbraad (changes requested):

test/e2e/features/story_microshift.feature
```diff
@@ -0,0 +1,35 @@
...
+Feature: Microshift test stories
+
+	Background:
+		Given setting config property "preset" to value "microshift" succeeds
+		And executing single crc setup command succeeds
+		And starting CRC with default bundle succeeds
+
+	# End-to-end health check
+
+	@microshift @testdata @linux @windows @mac
```
jsliacan: Not sure `@mac` equals to `$GOOS`. I think `@darwin` is what works.

test/e2e/testsuite/testsuite.go
```diff
@@ -31,22 +31,17 @@ var (
...
+	GodogTags string
 )
 
 func ParseFlags() {
 
 	pflag.StringVar(&util.TestDir, "test-dir", "out", "Path to the directory in which to execute the tests")
 	pflag.StringVar(&testWithShell, "test-shell", "", "Specifies shell to be used for the testing.")
 
-	pflag.StringVar(&bundleLocation, "bundle-location", "/path/to/bundle.crcbundle", "Path to the bundle to be used in tests")
+	pflag.StringVar(&bundleLocation, "bundle-location", "", "Path to the bundle to be used in tests")
```
gbraad, line 42: 👍

test/e2e/testsuite/testsuite.go
```diff
@@ -31,22 +31,17 @@ var (
...
 	pflag.StringVar(&pullSecretFile, "pull-secret-file", "/path/to/pull-secret", "Path to the file containing pull secret")
-	pflag.StringVar(&CRCExecutable, "crc-binary", "/path/to/binary/crc", "Path to the CRC executable to be tested")
+	pflag.StringVar(&CRCExecutable, "crc-binary", "", "Path to the CRC executable to be tested")
```
gbraad, line 44: 👍

gbraad (approved):

praveenkumar:
/hold

Need to verify the actual microshift test runs or not.

gbraad (approved):

cfergeau:
Unclear to me why this seems to be a wholly different test from the other ones we have? Could we share the same "httpd" test for both microshift and openshift?

praveenkumar:
> Unclear to me why this seems to be a wholly different test from the other ones we have? Could we share the same "httpd" test for both microshift and openshift?

@cfergeau `httpd` scenario is part of `story_openshift` feature where it expect the cluster of openshift running (as part of background). if a scenario can be abstracted which can consumed for different features then we didn't have to duplicate this test and consumed for both the scenarios but I didn't find a way.

adrianriobo:
I would suggest add a config flag to inform on which preset we want to run the tests (Default value to `openshift`).  

That flag can be used within the background steps (making them dynamic based on the flag). With that approach you can avoid the duplication.

cfergeau:
Took another look at this, and for some reason I had missed https://github.com/crc-org/crc/blob/97705a008eda801ba291fe3b9982af464d4759e0/test/e2e/features/story_openshift.feature#L13-L37 In my previous comments, I thought story_microshift.feature was totally different from the http OpenShift tests.
This is not the case, it's the same code, which answers my initial question :)

Imo it's really useful to add a comment before duplicated code to remind people when touching one of the code blocks, the other one needs to be updated. And documentingin the commit log why this was duplicated is also useful if one wants to try to improve things.

PR #3782, opened 2023-08-02
e2e: Add debug steps to verify the cert with CA

praveenkumar:
This is just for debug purpose to understand what is actually happen when oc login require `--insecure-skip-tls-verify`.

### DIFF
--- a/test/e2e/features/story_openshift.feature
+++ b/test/e2e/features/story_openshift.feature
@@ -3,6 +3,9 @@ Feature: 4 Openshift stories
 
 	Background:
 		Given ensuring CRC cluster is running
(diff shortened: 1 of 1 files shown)


praveenkumar:
/unhold

@adrianriobo, let's start using that to check why the login failure happen.

adrianriobo:
:thinking: ok but not sure if we want to keep these steps on our e2e scenario, I can just create a custom version for e2e container image with these steps in and try to reproduce on downstream so may keep the hold and once I got the version we can close the PR WDYT @praveenkumar ?

adrianriobo (approved):

cfergeau:
What is/was the login failure? What was achieved/understood thanks to this additional debug code?

cfergeau:
> What is/was the login failure? What was achieved/understood thanks to this additional debug code?

This is related to https://github.com/crc-org/crc/issues/3741

PR #3795, opened 2023-08-21
CI: Make sure oc command is available before any oc operations

### LINKED ISSUE #3792: [QE] Missing oc binary when verify the cert with CA
There is some random behavior when login into the cluster, which is being tracked on #3741. In order to debug what is the cause some debug steps had been added to the e2e scenarios on #3782.  Unfortunately those steps do not include force the eval of `crc oc-env`.

On those target machines where oc is not installed the step is failing as so all the test.

So we need to:

- [ ] added eval of `crc oc -env` to ensure we can use oc to get the `ca.crt` 
- [ x] ~as those steps are intended for debugging purposes and also the issue from #3741 may not affect all scenarios in case of an error on those steps the scenario probably should not fail but record the debug information~

praveenkumar:
As part of https://github.com/crc-org/crc/pull/3782 debugging steps are added to figure out why the `oc login` require insecure sometimes but this logic fails in case the `oc` binary is not part of system path. This PR make sure that the oc command available before those debugging steps.

fixes: #3792

### DIFF
--- a/test/e2e/features/story_microshift.feature
+++ b/test/e2e/features/story_microshift.feature
@@ -6,6 +6,7 @@ Feature: Microshift test stories
 		And setting config property "network-mode" to value "user" succeeds
 		And executing single crc setup command succeeds
 		And starting CRC with default bundle succeeds
+		And ensuring oc command is available
 
 	# End-to-end health check
 
--- a/test/e2e/features/story_openshift.feature
+++ b/test/e2e/features/story_openshift.feature
@@ -3,6 +3,7 @@ Feature: 4 Openshift stories
 
 	Background:
 		Given ensuring CRC cluster is running
+		And ensuring oc command is available
 		And executing "oc config view --raw -o jsonpath="{.clusters[?(@.name=='api-crc-testing:6443')].cluster.certificate-authority-data}"| base64 -d - > ca.crt" succeeds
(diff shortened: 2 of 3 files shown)


adrianriobo (commented):
Need to extend this to `microshift story` as well; none of the background steps includes the `ensuring oc command is available`:

```
Background:
		Given setting config property "preset" to value "microshift" succeeds
		And setting config property "network-mode" to value "user" succeeds
```

Then the first step from scenario is using it:

```
Given executing "oc create namespace testproj" succeeds
```

test/e2e/features/story_microshift.feature
```diff
@@ -3,6 +3,7 @@ Feature: Microshift test stories
 
 	Background:
 		Given setting config property "preset" to value "microshift" succeeds
+		And ensuring oc command is available
```
adrianriobo: This should go as the last step on the background  (cluster should be started fisrt)

praveenkumar: Yes, I updated it.

praveenkumar:
@adrianriobo anything else remaining?

adrianriobo (approved):

adrianriobo (commented):
maybe include on the commit the change on the story_microshift or add it a separate as you prefer?

praveenkumar:
> maybe include on the commit the change on the story_microshift or add it a separate as you prefer?

Doesn't current commit handle both situation because it is going to fail same with microshift.

Issue #3796, opened 2023-08-21
[BUG] Error setting network-mode on non linux platforms


adrianriobo:
Recently a `story_microshift` was added to the e2e scenarios by #3779 , the background steps include setting the network-mode to user.

On non-linux platforms that config option does not exist as network-mode user is the only mode available; this results on an error executing the story on non-linux platforms:

```
<testsuites name="crc" tests="1" skipped="0" failures="1" errors="0" time="0.6364514">
  <testsuite name="Microshift test stories" tests="1" skipped="0" failures="1" errors="0" time="0.6312547">
    <testcase name="Start and expose a basic HTTP service and check after restart" status="failed" time="0.6312547">
      <failure message="Step setting config property &#34;network-mode&#34; to value &#34;user&#34; succeeds: command &#39;crc config set network-mode user&#39;, expected to succeed, exited with exit code: 1&#xA;Command stdout: &#xA;Command stderr: level=debug msg=&#34;CRC version: 2.25.0+97705a\n&#34;&#xA;level=debug msg=&#34;OpenShift version: 4.13.6\n&#34;&#xA;level=debug msg=&#34;Podman version: 4.4.4\n&#34;&#xA;level=debug msg=&#34;Running &#39;crc config set&#39;&#34;&#xA;Configuration property &#39;network-mode&#39; does not exist&#xA;"></failure>
...
```

praveenkumar:
@adrianriobo I tested the scenario in mac and it does have `network-mode` setting. Is it only for windows?

```
✗ uname -a
Darwin prkumar-mac 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64

$ PULL_SECRET_FILE=/Users/prkumar/pull-secret BUNDLE_LOCATION=/Users/prkumar/.crc/cache/crc_microshift_vfkit_4.13.6_arm64.crcbundle make e2e-story-microshift 
[...]

PR #3798, opened 2023-08-22
e2e test: Ignore network-mode config property for non linux platform

praveenkumar:
Since network-mode is only supported on Linux for release bits of CRC, we skip this property test for non-linux platforms as part of e2e.

**Fixes:** #3796

### DIFF
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -861,6 +861,10 @@ func EnsureUserIsLoggedIntoClusterSucceedsOrFails(expected string) error {
 }
 
 func SetConfigPropertyToValueSucceedsOrFails(property string, value string, expected string) error {
+	// Since network-mode is only supported on Linux, we skip this property test for non-linux platforms
+	if property == "network-mode" && runtime.GOOS != "linux" {
+		return nil
+	}
 	if value == "current bundle" {
 		if !userProvidedBundle {
 			value = filepath.Join(util.CRCHome, "cache", bundleName)


adrianriobo:
dummy question, may I should already know the answer but...does microshift on linux only works with `network-mode user`?

praveenkumar:
> dummy question, may I should already know the answer but...does microshift on linux only works with `network-mode user`?

yes, because with microshift preset we also want to have podman support and https://github.com/crc-org/snc/pull/675 track why it is not works with podman.

cfergeau:
> > dummy question, may I should already know the answer but...does microshift on linux only works with `network-mode user`?
> 
> yes, because with microshift preset we also want to have podman support and [crc-org/snc#675](https://github.com/crc-org/snc/pull/675) track why it is not works with podman.

With my 'user' hat, I disagree with this statement, if I pick a linux release binary, don't change network-mode, do `crc config set preset microshift`, followed by `crc setup; crc start`, nothing errors out early, nothing tells me to change `network-mode`, nothing changes it for me. So from my perspective, system networking+microshift works on linux.
And if it does not, I don't think expecting everyone including the e2e tests to run `crc config set network-mode user` before using microshift is the right way of addressing the problem.

praveenkumar:
> With my 'user' hat, I disagree with this statement, if I pick a linux release binary, don't change network-mode, do crc config set preset microshift, followed by crc setup; crc start, nothing errors out early, nothing tells me to change network-mode, nothing changes it for me. So from my perspective, system networking+microshift works on linux.

May be we need to change that since we are putting effort to depreciate podman bundle in favour of microshift so we should add a warning message or error out in case network mode is `system` instead `user` for linux also.

adrianriobo (approved):

cfergeau:
> > With my 'user' hat, I disagree with this statement, if I pick a linux release binary, don't change network-mode, do crc config set preset microshift, followed by crc setup; crc start, nothing errors out early, nothing tells me to change network-mode, nothing changes it for me. So from my perspective, system networking+microshift works on linux.
> 
> May be we need to change that since we are putting effort to depreciate podman bundle in favour of microshift so we should add a warning message or error out in case network mode is `system` instead `user` for linux also.

Yes, this is what I was implying. And also that I don't think this PR is correct, I'd just go with:
```
diff --git a/test/e2e/features/story_microshift.feature b/test/e2e/features/story_microshift.feature
index bfc937e50..af547a315 100644
--- a/test/e2e/features/story_microshift.feature
+++ b/test/e2e/features/story_microshift.feature
@@ -3,7 +3,6 @@ Feature: Microshift test stories
 
        Background:
                Given setting config property "preset" to value "microshift" succeeds
-               And setting config property "network-mode" to value "user" succeeds
 
```
I don't think the microshift scenario will hit the podman-related bugs you mention?

adrianriobo:
:thinking: So I merged based on previous meeting with @praveenkumar I expect a following to address your comment

adrianriobo:
> I don't think this PR is correct, I'd just go with  

Sorry did I miss something? how that diff address the issue?

cfergeau:
The failure will be gone with this PR, but it's just adding complexity to the testing code to workaround bugs/problems in other parts of crc. This is what I meant by "I don't think it's correct".

PR #3799, opened 2023-08-23
proxy: Update trimTrailingEOL function to handle carriage return (\r)

### LINKED ISSUE #3785: [BUG] user-ca-bundle.json: json: offset 1425: invalid character '\r' in string literal
### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it? **Yes**
  * Running CRC on: Laptop

## CRC version
```bash
CRC version: 2.24.1+81aeeb
OpenShift version: 4.13.6
Podman version: 4.4.4
```
  
## CRC status
```bash
crc status --log-level debug
time="2023-08-05T18:58:06+02:00" level=warning msg="Failed to add proxy CA to crc http transport"
DEBU CRC version: 2.24.1+81aeeb
DEBU OpenShift version: 4.13.6
DEBU Podman version: 4.4.4
DEBU Running 'crc status'
CRC VM:          Running
OpenShift:       Unreachable (v4.13.6)
RAM Usage:       6.61GB of 15.34GB
Disk Usage:      16.39GB of 32.68GB (Inside the CRC VM)
Cache Usage:     22.7GB
Cache Directory: C:\Users\max\.crc\cache
```

## CRC config
```bash
- cpus                                  : 8
- http-proxy                            : http://user@domain:PASS...@1.2.3.4:8080
- https-proxy                           : http://user@domain:PASS...@1.2.3.4:8080
- ingress-http-port                     : 8080
- memory                                : 15000
- nameserver                            : 10.0.0.1
- proxy-ca-file                         : c:\temp\proxy.crt
```

## Host Operating System
```bash
Microsoft Windows 10 Enterprise
10.0.19044 Build 19044
```

### Steps to reproduce

```
crc start  --log-level debug
```

### Expected

`crc start` completes without errors.

### Actual

```
Failed to update cluster proxy configuration: Failed to add proxy cert details ssh command error:
command : timeout 30s oc apply -f /tmp/user-ca-bundle.json --context admin --cluster crc --kubeconfig /opt/kubeconfig
: error: error parsing /tmp/user-ca-bundle.json: json: offset 1425: invalid character '\r' in string literal
```

Similar bugreport from the past: #1957

### Logs

https://gist.github.com/maxmoser1/903373202dda41558a6b9a9eb6be11cc

Before gather the logs try following if that fix your issue
```bash
$ crc delete -f
$ crc cleanup
$ crc setup
$ crc start --log-level debug
```

Tried, but did not help.

praveenkumar:
It will fix the proxy parse issue in case user update the cert on windows and add blank line at the end of certificate. Unit test is also added for same.

fixes: #3785

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -21,6 +21,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/ssh"
 	crctls "github.com/crc-org/crc/pkg/crc/tls"
 	"github.com/crc-org/crc/pkg/crc/validation"
+	crcstrings "github.com/crc-org/crc/pkg/strings"
 	"github.com/pborman/uuid"
 )
 
@@ -134,7 +135,7 @@ func EnsureSSHKeyPresentInTheCluster(ctx context.Context, ocConfig oc.Config, ss
 	if err != nil {
 		return err
 	}
-	sshPublicKey := strings.TrimRight(string(sshPublicKeyByte), "\r\n")
+	sshPublicKey := crcstrings.TrimTrailingEOL(string(sshPublicKeyByte))
 	if err := WaitForOpenshiftResource(ctx, ocConfig, "machineconfigs"); err != nil {
 		return err
 	}
--- a/pkg/crc/network/httpproxy/proxy.go
+++ b/pkg/crc/network/httpproxy/proxy.go
@@ -51,7 +51,7 @@ func readProxyCAData(proxyCAFile string) (string, error) {
 }
 
 func trimTrailingEOL(s string) string {
-	return strings.TrimRight(s, "\n")
+	return strings.TrimRight(s, "\r\n")
 }
 
 func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAFile string) (*ProxyConfig, error) {
--- a/pkg/crc/network/httpproxy/proxy.go
+++ b/pkg/crc/network/httpproxy/proxy.go
@@ -12,6 +12,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/logging"
 
 	"github.com/asaskevich/govalidator"
+	crcstrings "github.com/crc-org/crc/pkg/strings"
 	"github.com/pkg/errors"
 	"golang.org/x/net/http/httpproxy"
 )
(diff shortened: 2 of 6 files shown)


cfergeau (commented):
Oh good catch! 
I don't think this necessarily is caused by users changing the cert file, it could also be tools which would add `/r/n` to the end of certificate files on Windows.
I'd give more details about what happens exactly. When the cert file ends with `\r\n`, `trimTrailingEOL` would only remove `\n` and the file will end with `\r`. Then this causes issues in https://github.com/crc-org/crc/blob/97705a008eda801ba291fe3b9982af464d4759e0/pkg/crc/cluster/cluster.go#L388-L391 which tries to replace `\r\n` and `\n` with `\\n` but will not do anything with a single `\r`.
This in turn causes bug #3785

praveenkumar:
> When the cert file ends with \r\n, trimTrailingEOL would only remove \n and the file will end with \r. Then this causes issues in  

@cfergeau you mean to say https://github.com/crc-org/crc/blob/97705a008eda801ba291fe3b9982af464d4759e0/pkg/crc/cluster/cluster.go#L388-L391 should handle `\r` case also and replace it `\n` because that is again going to fail. what we want there should be no new line after `-----END CERTIFICATE-----`  otherwise `user-ca-bundle.json` is not applied correctly on the cluster so I think this should be handled by `trimTrailingEOL` itself?

cfergeau:
> should handle `\r` case also and replace it `\n` because that is again going to fail.

I was only trying to describe the situation  before your changes, imo it would be useful to have in the commit log.
I agree with the way you fixed it, fix trimEOL to fully remove EOL (ie both \r\n and \n). With trimEOL fixed, the regexp will also behave as expected. Without the fix, the string does not match the regexp expectations, and this causes the reported bug.

cfergeau:
Not sure what happened, but the code was initially correct in https://github.com/crc-org/crc/pull/1443, and when moving it to a helper function, \r was lost :-/
After re-reading https://github.com/crc-org/crc/pull/1443#discussion_r468640042 I'm still unsure if removing the EOL at the end of the file is required, or if it's done because it's nicer?

cfergeau:
There's a second occurrence of the same code:
```
pkg/crc/cluster/cluster.go:     sshPublicKey := strings.TrimRight(string(sshPublicKeyByte), "\r\n")
```
and the test code has
```
test/extended/util/checks.go:   actual = strings.TrimRight(actual, "\n")
```

Let's make use of the recently added `pkg/strings` package and add a shared helper. It's trivial, but we got it wrong once, better to implement it once and for all.

praveenkumar:
> Let's make use of the recently added pkg/strings package and add a shared helper. It's trivial, but we got it wrong once, better to implement it once and for all.

Added another commit to do that.

cfergeau (approved):

PR #3802, opened 2023-08-25
metadata: Add logic to fallback to quay if mirror is unreachable 

praveenkumar:
Since we are going to put bundles on quay and mirror where quay serve as
fallback solution. This PR makes user aware in case mirror fails to
download bundle.

```
INFO Downloading bundle: /home/prkumar/.crc/cache/crc_microshift_libvirt_4.13.9_amd64.crcbundle...
INFO Unable to download bundle from mirror, falling back to quay
Copying blob d37da0367460 done
Copying config f3021495d6 done
Storing signatures
crc_microshift_libvirt_4.13.9_amd64.crcbundle:  1.46 GiB / 1.46 GiB [--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
INFO Verifying the bundle signature...
INFO Uncompressing /home/prkumar/.crc/cache/crc_microshift_libvirt_4.13.9_amd64.crcbundle
crc.qcow2:  4.28 GiB / 4.28 GiB [--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
oc:  141.85 MiB / 141.85 MiB [-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
```

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -12,30 +12,31 @@ import (
 )
 
 const (
-	Bundle                  = "bundle"
-	CPUs                    = "cpus"
-	Memory                  = "memory"
-	DiskSize                = "disk-size"
-	NameServer              = "nameserver"
-	PullSecretFile          = "pull-secret-file"
-	DisableUpdateCheck      = "disable-update-check"
-	ExperimentalFeatures    = "enable-experimental-features"
-	NetworkMode             = "network-mode"
-	HostNetworkAccess       = "host-network-access"
-	HTTPProxy               = "http-proxy"
-	HTTPSProxy              = "https-proxy"
-	NoProxy                 = "no-proxy"
-	ProxyCAFile             = "proxy-ca-file"
-	ConsentTelemetry        = "consent-telemetry"
-	EnableClusterMonitoring = "enable-cluster-monitoring"
-	KubeAdminPassword       = "kubeadmin-password"
-	Preset                  = "preset"
-	EnableSharedDirs        = "enable-shared-dirs"
-	SharedDirPassword       = "shared-dir-password" // #nosec G101
-	IngressHTTPPort         = "ingress-http-port"
-	IngressHTTPSPort        = "ingress-https-port"
-	EmergencyLogin          = "enable-emergency-login"
-	PersistentVolumeSize    = "persistent-volume-size"
+	Bundle                   = "bundle"
+	CPUs                     = "cpus"
+	Memory                   = "memory"
+	DiskSize                 = "disk-size"
+	NameServer               = "nameserver"
+	PullSecretFile           = "pull-secret-file"
+	DisableUpdateCheck       = "disable-update-check"
+	ExperimentalFeatures     = "enable-experimental-features"
+	NetworkMode              = "network-mode"
+	HostNetworkAccess        = "host-network-access"
+	HTTPProxy                = "http-proxy"
+	HTTPSProxy               = "https-proxy"
+	NoProxy                  = "no-proxy"
+	ProxyCAFile              = "proxy-ca-file"
+	ConsentTelemetry         = "consent-telemetry"
+	EnableClusterMonitoring  = "enable-cluster-monitoring"
+	KubeAdminPassword        = "kubeadmin-password"
+	Preset                   = "preset"
+	EnableSharedDirs         = "enable-shared-dirs"
+	SharedDirPassword        = "shared-dir-password" // #nosec G101
+	IngressHTTPPort          = "ingress-http-port"
+	IngressHTTPSPort         = "ingress-https-port"
+	EmergencyLogin           = "enable-emergency-login"
+	PersistentVolumeSize     = "persistent-volume-size"
+	EnableBundleQuayFallback = "enable-bundle-quay-fallback"
 )
 
 func RegisterSettings(cfg *Config) {
@@ -137,6 +138,9 @@ func RegisterSettings(cfg *Config) {
 		fmt.Sprintf("HTTP port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPPort))
 	cfg.AddSetting(IngressHTTPSPort, constants.OpenShiftIngressHTTPSPort, validatePort, RequiresHTTPSPortChangeWarning,
 		fmt.Sprintf("HTTPS port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPSPort))
+
+	cfg.AddSetting(EnableBundleQuayFallback, false, ValidateBool, SuccessfullyApplied,
+		"If bundle download from the default location fails, fallback to quay.io (true/false, default: false)")
 
 	if err := cfg.RegisterNotifier(Preset, presetChanged); err != nil {
 		logging.Debugf("Failed to register notifier for Preset: %v", err)
--- a/pkg/crc/image/image.go
+++ b/pkg/crc/image/image.go
@@ -104,8 +104,10 @@ func getPresetNameE(imageName string) (crcpreset.Preset, error) {
 		return crcpreset.OKD, nil
 	case "podman-bundle":
 		return crcpreset.Podman, nil
+	case "microshift-bundle":
+		return crcpreset.Microshift, nil
 	default:
-		return crcpreset.OpenShift, fmt.Errorf("invalid image name '%s' (Should be openshift-bundle, okd-bundle or podman-bundle)", imageName)
+		return crcpreset.OpenShift, fmt.Errorf("invalid image name '%s' (Should be openshift-bundle, okd-bundle, podman-bundle or microshift-bundle)", imageName)
 	}
 }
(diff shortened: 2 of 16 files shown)


pkg/crc/machine/bundle/metadata.go
```diff
@@ -353,7 +353,12 @@ func Download(preset crcPreset.Preset, bundleURI string) (string, error) {
 	// bundles, their sha256sums are known and can be checked.
 	if bundleURI == constants.GetDefaultBundlePath(preset) {
 		if preset == crcPreset.OpenShift || preset == crcPreset.Microshift {
-			return DownloadDefault(preset)
+			downloadedBundlePath, err := DownloadDefault(preset)
```
cfergeau: Is there a way to differentiate "not found" from other errors?

pkg/crc/machine/bundle/metadata.go
```diff
@@ -353,7 +353,12 @@ func Download(preset crcPreset.Preset, bundleURI string) (string, error) {
...
+			if err != nil {
```
cfergeau: This cannot be automatic and unconditional.

cfergeau: To expand on this, this can go behind a config option, or maybe env var, or ... ?

pkg/crc/image/image.go
```diff
@@ -107,6 +107,8 @@ func getImageName(preset crcpreset.Preset) string {
 		return "podman-bundle"
 	case crcpreset.OKD:
 		return "okd-bundle"
+	case crcpreset.Microshift:
+		return "microshift-bundle"
```
cfergeau: We have code in pkg/crc/constants.go which goes from preset to bundle URL. This code is used in pkg/crc/machine/bundle/metadata.go to get the default bundle https URLs. The same file also has some code to deal with bundles stored in registry, and in this case it relies on a `""` magic string to indicate to use the default bundle registry URL for the bundle, and it's code in pkg/crc/image/ which does the preset -> default bundle mapping.

Can we use this PR as an opportunity to unify this behaviour a little bit? (remove the magic `""` argument to `image.Pull()`, and move this logic to pkg/crc/constants instead, and make use of this in pkg/crc/machine/bundle)

praveenkumar: @cfergeau If I understood correctly what you want that logic of getting defaultURI for Quay should be part of `pkg/crc/contants` and should be consume in the `pkg/crc/machine/bundle/metadata.go`  ?

So that `image.PullBundle(preset, "")`  become `image.PullBundle(constants.GetDefaultBundleImageRegistryURL(preset))` ? If we need to do that then we need to move `getImageName` and `getPresetNameE` helper from `/pkg/crc/image/image.go` to `constants.go`.

cfergeau: Yes, my suggestion is to try something like this, and to see if it makes the code easier to follow/better organized or not.

praveenkumar: @cfergeau I can try that in a separate refactor PR, can we just try to focus on quay logic in this PR.

pkg/crc/machine/bundle/metadata.go
praveenkumar: @cfergeau looks like `grab` package have `StatusCodeError` type and can validated with `IsStatusCodeError` which I now used.
```
StatusCodeError indicates that the server response had a status code that
// was not in the 200-299 range (after following any redirects).
```

pkg/crc/image/image.go
praveenkumar: https://github.com/crc-org/crc/pull/3817

cfergeau: > @cfergeau I can try that in a separate refactor PR, can we just try to focus on quay logic in this PR.

I already brought this up when the quay support was introduced, at the time the answer was already in the same line ;) we cannot keep adding technical debt, and postpone cleanups forever.

> #3817

Thanks!

cfergeau:
This PR needs a rebase.

pkg/crc/image/image.go
```diff
@@ -104,6 +104,8 @@ func getPresetNameE(imageName string) (crcpreset.Preset, error) {
 		return crcpreset.OKD, nil
 	case "podman-bundle":
 		return crcpreset.Podman, nil
+	case "microshift-bundle":
+		return crcpreset.Microshift, nil
 	default:
 		return crcpreset.OpenShift, fmt.Errorf("invalid image name '%s' (Should be openshift-bundle, okd-bundle or podman-bundle)", imageName)
```
cfergeau: `microshift-bundle` is missing here.

pkg/crc/config/settings.go
```diff
@@ -138,6 +139,9 @@ func RegisterSettings(cfg *Config) {
 	cfg.AddSetting(IngressHTTPSPort, constants.OpenShiftIngressHTTPSPort, validatePort, RequiresHTTPSPortChangeWarning,
 		fmt.Sprintf("HTTPS port to use for OpenShift ingress/routes on the host (1024-65535, default: %d)", constants.OpenShiftIngressHTTPSPort))
 
+	cfg.AddSetting(EnableBundleQuayFallback, false, ValidateBool, SuccessfullyApplied,
+		"Enable pull bundle from quay as fallback (true/false, default: false)")
```
cfergeau: Maybe `If bundle download from the default location fails, fallback to quay.io (true/false ...` ?

pkg/crc/preflight/preflight.go
```diff
@@ -156,8 +156,9 @@ func getPreflightChecksHelper(config crcConfig.Storage) []Check {
 	mode := crcConfig.GetNetworkMode(config)
 	bundlePath := config.Get(crcConfig.Bundle).AsString()
 	preset := crcConfig.GetPreset(config)
+	enableBundleQuayFallback := config.Get(crcConfig.EnableBundleQuayFallback).AsBool()
 	logging.Infof("Using bundle path %s", bundlePath)
-	return getPreflightChecks(experimentalFeatures, mode, bundlePath, preset)
+	return getPreflightChecks(experimentalFeatures, mode, bundlePath, preset, enableBundleQuayFallback)
```
cfergeau, line 161: Hrm, 4 arguments was already quite a few, 5 is starting to be too many args :-/

pkg/crc/machine/bundle/metadata.go
```diff
@@ -346,7 +348,12 @@ func Download(preset crcPreset.Preset, bundleURI string) (string, error) {
 	if bundleURI == constants.GetDefaultBundlePath(preset) {
 		switch preset {
 		case crcPreset.OpenShift, crcPreset.Microshift:
-			return downloadDefault(preset)
+			downloadedBundlePath, err := downloadDefault(preset)
+			if grab.IsStatusCodeError(errors.Unwrap(err)) && enableBundleQuayFallback {
```
cfergeau: Looks like this could be `errors.Is(err, grab.StatusCodeError)`

cfergeau (commented):
Overall looks good, a few minor comments left.

pkg/crc/machine/bundle/metadata.go
praveenkumar: @cfergeau this is not work because as per `errors.Is` it doesn't call the `unwrap` on the target and the way we have this error it is wrapped around `grabStatusCodeError` so it is not able to work.

pkg/crc/config/settings.go
praveenkumar: Updated.

pkg/crc/machine/bundle/metadata.go
cfergeau: `errors.Is` was introduced to deal with wrapped errors. Its documentation is clear about this. https://pkg.go.dev/errors

praveenkumar: Yes but if I take a look to grab error package the implementation around StatusCodeError is type int and if I try to use `errors.Is` on that it is not working until we know the statuscode. I think that is the reason it have `IsStatusCodeError` function.

https://go.dev/play/p/jwkn9TG9SmH

cfergeau: Ah right, `errors.Is` checks for equality, not for type :-/ `errors.As` checks for types, but is not as convenient as `errors.Is`. If we keep this code, I think we should use `errors.As` instead of hardcoding a single `errors.Unwrap()`. Hardcoding this `Unwrap()` will break if we move/refactor this code, and the level of wrapping is changed (eg if `errors.Unwrap(errors.Unwrap())` becomes needed instead of `errors.Unwrap()`). 

However looking at `downloadDefault()`, there are more situations than "the bundle file is missing/cannot be downloaded" when we want to fallback, for example if the signature is missing/invalid. `IsStatusCodeError()` seems to be too restrictive? The initial version where the fallback always happened when there was an error may be a better option

praveenkumar: @cfergeau yeah that is more flexible, I can revert to that.

praveenkumar: Done

cfergeau (approved):

PR #3804, opened 2023-08-25
test: microshift: Don't force network-mode to 'user'

cfergeau:
Filing this now so that we don't forget it when we have bundles with https://github.com/crc-org/snc/pull/797. This will be put on hold until this happens.

We want our testing to be as close as possible to what end-users will 
experience. Changing `network-mode` is a significant difference from the 
default. With https://github.com/crc-org/snc/pull/797 fixed, we no longer
need to force it to `user`.

### DIFF
--- a/test/e2e/features/story_microshift.feature
+++ b/test/e2e/features/story_microshift.feature
@@ -3,12 +3,10 @@ Feature: Microshift test stories
 
 	Background:
 		Given setting config property "preset" to value "microshift" succeeds
-		And ensuring network mode user
 		And executing single crc setup command succeeds
 		And starting CRC with default bundle succeeds
 		And ensuring oc command is available
 		And ensuring microshift cluster is fully operational
-		
 	# End-to-end health check
 
 	@microshift @testdata @linux @windows @darwin @cleanup
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -1002,10 +1002,6 @@ func setPodmanEnv() error {
 }
 
 func SetConfigPropertyToValueSucceedsOrFails(property string, value string, expected string) error {
-	// Since network-mode is only supported on Linux, we skip this property test for non-linux platforms
-	if property == "network-mode" && runtime.GOOS != "linux" {
-		return nil
-	}
 	if value == "current bundle" {
 		if !userProvidedBundle {
 			value = filepath.Join(util.CRCHome, "cache", bundleName)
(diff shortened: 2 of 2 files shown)


praveenkumar:
@jsliacan Can you try that and see if that works as expected.

cfergeau:
With the switch to user-mode networking everywhere, this probably can go in now.

praveenkumar:
Now by default network mode is user so we dont' force it anymore.

/close

cfergeau:
This needs rebase, it’s now less important as it’s just a cleanup now that usermode networking is default, but I’m not sure there is nothing to do anymore

cfergeau:
The updated PR removes no longer needed code.
Maybe we want to add a step which makes sure we are using the default networking mode.

test/e2e/features/story_microshift.feature
```diff
@@ -35,4 +33,4 @@ Feature: Microshift test stories
 		And with up to "4" retries with wait period of "1m" http response from "http://httpd-example-testproj.apps.crc.testing" has status code "200"
 		Then executing "curl -s http://httpd-example-testproj.apps.crc.testing" succeeds
 		And stdout should contain "Hello CRC!"
-		Then with up to "4" retries with wait period of "1m" http response from "http://httpd-example-testproj.apps.crc.testing" has status code "200"
\ No newline at end of file
+		Then with up to "4" retries with wait period of "1m" http response from "http://httpd-example-testproj.apps.crc.testing" has status code "200"
```
praveenkumar, line 36: Is this change just adding new line?

cfergeau: It’s my editor doing this automatically, but yes, it ends the file with a new line.

test/e2e/testsuite/testsuite.go
```diff
@@ -1004,10 +1002,6 @@ func setPodmanEnv() error {
 }
 
 func SetConfigPropertyToValueSucceedsOrFails(property string, value string, expected string) error {
-	// Since network-mode is only supported on Linux, we skip this property test for non-linux platforms
-	if property == "network-mode" && runtime.GOOS != "linux" {
-		return nil
-	}
```
praveenkumar, line 1010: Shouldn't this still true? On linux network-mode is still have option for `system` and `user` (being default)?

cfergeau: The previous commit removes the setting of `network-mode` in the tests, so this becomes dead code.

praveenkumar:
merged as part of #5319

/close

PR #3805, opened 2023-08-25
start: Ignore pull secret update in the instance for stop => start

praveenkumar:
For microshift preset, we are updating the pull secret to the instance even user perform stop => start, instead we just need to validate the pull secret present in the instance and if it fails then only update the pull secret.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -32,6 +32,7 @@ import (
 	"github.com/crc-org/crc/pkg/crc/systemd"
 	"github.com/crc-org/crc/pkg/crc/telemetry"
 	crctls "github.com/crc-org/crc/pkg/crc/tls"
+	"github.com/crc-org/crc/pkg/crc/validation"
 	"github.com/crc-org/crc/pkg/libmachine/host"
 	crcos "github.com/crc-org/crc/pkg/os"
 	"github.com/crc-org/machine/libmachine/drivers"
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/start.go
```diff
@@ -1070,6 +1067,19 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 	return cluster.WaitForAPIServer(ctx, ocConfig)
 }
 
+func ensurePullSecretPresentInVM(sshRunner *crcssh.Runner, pullSec cluster.PullSecretLoader) error {
+	if pullSecret, _, err := sshRunner.RunPrivileged("Checking if pull secret already present in the VM", "cat", "-f", "/etc/crio/openshift-pull-secret"); err == nil {
+		if err := validation.ImagePullSecret(pullSecret); err == nil {
+			return nil
+		}
+	}
+	content, err := pullSec.Value()
```
cfergeau, line 1077: Maybe add a comment before this since it's reversed compared to usual go code which exit early on errors, not on success:
```
// pull secret is not present in VM or is invalid
```

cfergeau (commented):
The same could be done for the kubeconfig file, no need to copy it multiple times.

praveenkumar:
> The same could be done for the kubeconfig file, no need to copy it multiple times.

@cfergeau The validation of kubeconfig file is require api server to be up and running so it will be kind of expensive to wait extra till api server is up and then validate the kubeconfig file. In case of OCP also we don't validate the kubeconfig file because it is not possible without accessing the apiserver.

cfergeau:
> > The same could be done for the kubeconfig file, no need to copy it multiple times.
> 
> @cfergeau The validation of kubeconfig file is require api server to be up and running so it will be kind of expensive to wait extra till api server is up and then validate the kubeconfig file. In case of OCP also we don't validate the kubeconfig file because it is not possible without accessing the apiserver.

I'm thinking of this code, which is a few lines after the changes in this PR:
```
	if err := sshRunner.CopyFileFromVM(fmt.Sprintf("/var/lib/microshift/resources/kubeadmin/api%s/kubeconfig", constants.ClusterDomain), constants.KubeconfigFilePath, 0600); err != nil {
		return err
	}
	if err := sshRunner.CopyFile(constants.KubeconfigFilePath, "/opt/kubeconfig", 0644); err != nil {
		return err
	}
```
If ~/.crc/machines/crc/kubeconfig and /opt/kubeconfig already exists, no need to do the copies?

cfergeau:
> If ~/.crc/machines/crc/kubeconfig and /opt/kubeconfig already exists, no need to do the copies?

Not really important though, I'm fine with the PR with or without this change.

PR #3809, opened 2023-08-29
[RFC] go.mod: Append /v2 to module URL

cfergeau:
Since crc uses v2.x version numbers, its module URL must end in /v2,
otherwise its code can't be reused in other modules. I got bitten by
this when I wanted to use some of the extraction code in vfkit tests.
go 1.20 either wants to use crc v1.x as this is what an unversioned
module URL refers to, and if I try to force the use of crc v2.x, it
errors out because the module URL is not what it expects:
```
$ go mod init example.com/crc-user
$ cat main.go
package main

import (
	"github.com/crc-org/crc/pkg/extract"
)

func main() {
	_ = extract.Uncompress("", "")
}

$ go mod tidy
go: finding module for package github.com/crc-org/crc/pkg/extract
go: found github.com/crc-org/crc/pkg/extract in github.com/crc-org/crc v1.99.4
go: example.com/crc-user imports
	github.com/crc-org/crc/pkg/extract: github.com/crc-org/crc@v1.99.4: parsing go.mod:
	module declares its path as: github.com/code-ready/crc
	        but was required as: github.com/crc-org/crc

$ go get github.com/crc-org/crc@v2.25.0
go: github.com/crc-org/crc@v2.25.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/crc-org/crc/v2")
```
I could not find other go code importing crc as a module, so this change
should not have a big impact.


cfergeau:
Rebased. Would be nice to get some feedback on this approach :)

praveenkumar:
There are https://github.com/crc-org/crc/network/dependents?dependent_type=REPOSITORY dependents which consume crc but either they are very old or not using it any more. 

/lgtm
/approve

PR #3813, opened 2023-08-30
Small cleanups to pkg/crc/machine/bundle.DownloadDefault() code

cfergeau:
There should be no visible changes, this PR removes some dead code, and add unit tests for the GPG signature checks for https downloads.

More could be built on top of that, such as trying to get a single code path for Download/downloadDefault, so that there is (optional) GPG signature checks in the non-default bundle case as well.


pkg/crc/gpg/gpg.go
```diff
@@ -45,7 +45,7 @@ func GetVerifiedClearsignedMsgV3(pubkey, clearSignedMsg string) (string, error)
 	}
 	block, rest := goClearsign.Decode([]byte(clearSignedMsg))
 	if len(rest) != 0 {
-		return "", fmt.Errorf("Error decoding clear signed message: %w", err)
+		return "", fmt.Errorf("Error decoding clear signed message")
```
praveenkumar, line 48: Bit confusing commit log, what we were doing is wrapping a `nil` error to return error object which is wrong.

pkg/crc/image/image.go
```diff
@@ -130,6 +131,9 @@ func GetPresetName(imageName string) crcpreset.Preset {
 }
 
 func PullBundle(preset crcpreset.Preset, imageURI string) (string, error) {
+	proxyConfig, err := httpproxy.NewProxyConfig()
+	logging.Infof("debug: proxyConfig: %+v err: %v", proxyConfig, err)
```
praveenkumar: If we want the debug then better to `logging.Debugf` instead `Infof`?

test/integration/proxy_test.go
```diff
@@ -52,11 +53,14 @@ var _ = Describe("", Label("openshift-preset", "goproxy"), func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			cmd := exec.Command("oc", "config", "use-context", "crc-admin")
-			err = cmd.Run()
+			out, err := cmd.Output()
+			if err != nil {
+				fmt.Printf("error running 'oc config use-context crc-admin' - output: %s\n", string(out))
```
praveenkumar: should we put that also as `logging.Debugf` instead `Printf` ?

pkg/crc/gpg/gpg.go
cfergeau: What matters to me is that the error is unrelated to the current function call (`goClearsign.Decode`), maybe it's set, maybe not, but its value does not matter to me, it's irrelevant to the current context.

test/integration/proxy_test.go
cfergeau: This is not meant to land in `main`, I'm trying to understand why the integration tests fail more often than not, the commit should be removed before pushing.

pkg/crc/gpg/gpg.go
cfergeau: Tried to mke this clearer in the log.

test/integration/proxy_test.go
cfergeau: I've removed these debug commits now

pkg/crc/gpg/gpg.go
praveenkumar: In commit log `the err variable is related to the  goClearsign.Decode call` => `the err variable is related to the  goOpenpgp.ReadArmoredKeyRing call` ?

cfergeau: Thanks, I reworked the log once more ^^ :)

Issue #3641, opened 2023-05-09
[BUG] Add path validation for `proxy-ca-file`


jsliacan:
It seems that relative path causes problems. Absolute path works OK. The error message isn't great, but maybe just validating input for the path would help. 

Also, can we update the description for the `proxy-ca-file` argument? Currently it says:
```
* proxy-ca-file                              Path to an HTTPS proxy certificate authority (CA)
```
I think the following would more accurately describe the item:

> **Absolute** path to an HTTPS proxy certificate authority's **certificate file**

---

### General information
  * OS: Linux

## CRC version
```bash
# Put `crc version` output here
```
  
## CRC status
```bash
# Put `crc status --log-level debug` output here
```

## CRC config
```bash
- consent-telemetry                     : no
- host-network-access                   : true
- http-proxy                            : http://host.crc.testing:8888
- https-proxy                           : http://host.crc.testing:8888
- network-mode                          : user
- proxy-ca-file                         : test/testdata/pki/rootCA.crt
```

### Steps to reproduce

  1. `$ crc config set proxy-ca-file test/testdata/pki/rootCA.crt`
  2. `$ crc setup`
  3. `$ crc start -p ~/Downloads/crc-pull-secret`

### Expected
Start works OK.

### Actual
```
Is 'crc daemon' running? Cannot reach daemon API: Get "http://unix/api/version": read unix @->/home/jsliacan/.crc/crc-http.sock: read: connection reset by peer
```

anjannath:
@jsliacan I was thinking it didn't have any validation, but the config option does have path validation, so it verifies the file exists.. wondering what else could've caused the daemon to error out and exit..

https://github.com/crc-org/crc/blob/bb69afc5f2e9a88d3e76a453ef39295090b788b9/pkg/crc/config/settings.go#L117

cfergeau:
My guess is that if you use a relative path, and then move to a different directory, and then try to run crc from there, then it will fail
```
./crc config set proxy-ca-file rel/path/to/ca.crt
cd ../..
./out/linux-amd64/crc ...
```
At step 3., the relative path will no longer be valid as the working dir changed.

jsliacan:
I am not sure I understand @cfergeau . I don't think I (knowingly) changed the dir between setting the `proxy-ca-file` var and starting `crc`. It happens when I try it with explicit location of the `crc` binary in a fixed working dir (`~/Github/crc`).

```
jsliacan ~/Github/crc (goproxy)$ ./crc version
[...]

Issue #3098, opened 2022-04-06
[BUG] Wrong HTTP(S)_PROXY value blocks all CRC usage


jsliacan:
### General information

  * OS: Linux

## CRC version
```bash
CodeReady Containers version: 2.0.1+bf3b1a6
OpenShift version: 4.10.3
Podman version: 3.4.4
```
  
## CRC status
```bash
Machine does not exist. Use 'crc start' to create it
```

## CRC config
```bash
- consent-telemetry                     : no
- pull-secret-file                      : /home/jsliacan/Downloads/crc-pull-secret
```

### Steps to reproduce

  1. `export HTTP_PROXY=localhost:8080` (for instance)
  2. `crc version` (or any other command: `help`, `status`, `start`, `setup`, ...)

### Expected

The corresponding output for that command, i.e. for `version` I'd expect 

```
CodeReady Containers version: 2.0.1+bf3b1a6
OpenShift version: 4.10.3
Podman version: 3.4.4
```

### Actual

```
FATA[0000] HTTP proxy URL 'localhost:8080' is not valid: url should start with http://
```

### Logs

```
jsliacan ~ $ crc version
CodeReady Containers version: 2.0.1+bf3b1a6
OpenShift version: 4.10.3
Podman version: 3.4.4
jsliacan ~ $ crc status
Machine does not exist. Use 'crc start' to create it
jsliacan ~ $ crc config view
- consent-telemetry                     : no
- pull-secret-file                      : /home/jsliacan/Downloads/crc-pull-secret
jsliacan ~ $ export HTTP_PROXY=localhost:8080
jsliacan ~ $ crc status
FATA[0000] HTTP proxy URL 'localhost:8080' is not valid: url should start with http:// 
jsliacan ~ $ crc ip
FATA[0000] HTTP proxy URL 'localhost:8080' is not valid: url should start with http:// 
jsliacan ~ $ crc start
FATA[0000] HTTP proxy URL 'localhost:8080' is not valid: url should start with http:// 
jsliacan ~ $ crc 
FATA[0000] HTTP proxy URL 'localhost:8080' is not valid: url should start with http:// 
```

jsliacan:
- might apply to all OSs

anjannath:
was fixed in #3816

PR #3816, opened 2023-09-01
Validate proxy taken from env variables and allow only absolute paths for config proxy-ca-file

### LINKED ISSUE #3690: [BUG] Invalid `proxy-ca-file` path prevents crc from starting
### General information

  * OS: Linux
  * Hypervisor: KVM

## CRC version
```bash
# Put `crc version` output here
$ ./out/linux-amd64/crc version
FATA[0000] not able to read proxy CA data from /home/teuf/crc/crc/images/squid/pki/rootCA.crt: open /home/teuf/crc/crc/images/squid/pki/rootCA.crt: no such file or directory 
```
(yes it is that bad)
  
## CRC config
```bash
# Put `crc config view` output here
$ ./out/linux-amd64/crc view
FATA[0000] not able to read proxy CA data from /home/teuf/crc/crc/images/squid/pki/rootCA.crt: open /home/teuf/crc/crc/images/squid/pki/rootCA.crt: no such file or directory 
```

### Steps to reproduce

  1. set `proxy-ca-file` to a valid path
  2. remove the file

### Expected

`crc` errors out when it tries to configure a VM with a proxy

### Actual

All `crc` invocations (`config`, `version`, ...) print a proxy-ca-file related error message and exit.
The config file must be edited by hand to get out of this.

This could be worked on at the same time as https://github.com/crc-org/crc/issues/3641

anjannath:
- validate proxy urls taken from env vairbles
- validate that proxy certificate file exists before trying to read it
- only allow absolute paths for the `proxy-ca-file` config setting

Fixes #3690 #3098 #3641

### DIFF
--- a/pkg/crc/config/config.go
+++ b/pkg/crc/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"path/filepath"
 	"reflect"
 
 	"github.com/crc-org/crc/v2/pkg/crc/preset"
@@ -101,6 +102,14 @@ func (c *Config) Set(key string, value interface{}) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf(invalidProp, value, key, err)
 		}
+	case Path:
+		path := cast.ToString(value)
+		path, err := filepath.Abs(path)
+		if err != nil {
+			return "", fmt.Errorf(invalidProp, value, key, err)
+		}
+		castValue = path
+
 	case preset.Preset:
 		castValue = cast.ToString(value)
 	default:
@@ -200,6 +209,8 @@ func (c *Config) Get(key string) SettingValue {
 		}
 	case string:
 		value = cast.ToString(value)
+	case Path:
+		value = Path(cast.ToString(value))
 	case bool:
 		value, err = cast.ToBoolE(value)
 		if err != nil {
--- a/pkg/crc/config/settings_test.go
+++ b/pkg/crc/config/settings_test.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"path/filepath"
 	"testing"
 
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
@@ -157,4 +158,27 @@ func TestUnsetPreset(t *testing.T) {
 		IsDefault: true,
 		IsSecret:  false,
 	}, cfg.Get(CPUs))
+}
+
+func TestPath(t *testing.T) {
+	cfg, err := newInMemoryConfig()
+	require.NoError(t, err)
+
+	assert.Equal(t, SettingValue{
+		Value:     Path(""),
+		Invalid:   false,
+		IsDefault: true,
+		IsSecret:  false,
+	}, cfg.Get(ProxyCAFile))
+
+	_, err = cfg.Set(ProxyCAFile, "testdata/foo.crt")
+	require.NoError(t, err)
+	expectedPath, err := filepath.Abs("testdata/foo.crt")
+	require.NoError(t, err)
+	assert.Equal(t, SettingValue{
+		Value:     Path(expectedPath),
+		Invalid:   false,
+		IsDefault: false,
+		IsSecret:  false,
+	}, cfg.Get(ProxyCAFile))
 }
--- a/pkg/crc/network/httpproxy/proxy.go
+++ b/pkg/crc/network/httpproxy/proxy.go
@@ -29,6 +29,7 @@ type ProxyConfig struct {
 	noProxy     []string
 	ProxyCACert string
 	ProxyCAFile string
+	fromEnv     bool
 }
 
 func (p *ProxyConfig) String() string {
@@ -62,14 +63,17 @@ func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAFile string) (*Prox
 		HTTPSProxy:  httpsProxy,
 		ProxyCACert: proxyCAData,
 		ProxyCAFile: proxyCAFile,
+		fromEnv:     false,
 	}
 	envProxy := httpproxy.FromEnvironment()
 
-	if DefaultProxy.HTTPProxy == "" {
+	if DefaultProxy.HTTPProxy == "" && envProxy.HTTPProxy != "" {
 		DefaultProxy.HTTPProxy = envProxy.HTTPProxy
+		DefaultProxy.fromEnv = true
 	}
-	if DefaultProxy.HTTPSProxy == "" {
+	if DefaultProxy.HTTPSProxy == "" && envProxy.HTTPSProxy != "" {
 		DefaultProxy.HTTPSProxy = envProxy.HTTPSProxy
+		DefaultProxy.fromEnv = true
 	}
 	if noProxy == "" {
 		noProxy = envProxy.NoProxy
@@ -87,6 +91,7 @@ func NewProxyConfig() (*ProxyConfig, error) {
 		HTTPSProxy:  DefaultProxy.HTTPSProxy,
 		ProxyCACert: DefaultProxy.ProxyCACert,
 		ProxyCAFile: DefaultProxy.ProxyCAFile,
+		fromEnv:     DefaultProxy.fromEnv,
 	}
 
 	config.noProxy = defaultNoProxies
@@ -96,11 +101,17 @@ func NewProxyConfig() (*ProxyConfig, error) {
 
 	err := ValidateProxyURL(config.HTTPProxy, false)
 	if err != nil {
+		if config.fromEnv {
+			return nil, fmt.Errorf("Please check HTTP_PROXY env variable: %w", err)
+		}
 		return nil, err
 	}
 
 	err = ValidateProxyURL(config.HTTPSProxy, true)
 	if err != nil {
+		if config.fromEnv {
+			return nil, fmt.Errorf("Please check HTTPS_PROXY env variable: %w", err)
+		}
 		return nil, err
 	}
 
--- a/cmd/crc/cmd/root.go
+++ b/cmd/crc/cmd/root.go
@@ -56,7 +56,7 @@ func init() {
 	}
 
 	if err := setProxyDefaults(); err != nil {
-		logging.Fatal(err.Error())
+		logging.Warn(err.Error())
 	}
 
 	// Initiate segment client
--- a/pkg/crc/api/api_client_test.go
+++ b/pkg/crc/api/api_client_test.go
@@ -196,6 +196,9 @@ func TestConfigGetAll(t *testing.T) {
 		switch v := v.Value.(type) {
 		case int:
 			configs[k] = float64(v)
+		// when config of type Path is converted to JSON it is converted to string
+		case crcConfig.Path:
+			configs[k] = string(v)
 		default:
 			configs[k] = v
 		}
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -122,7 +122,7 @@ func RegisterSettings(cfg *Config) {
 		"HTTPS proxy URL (string, like 'https://my-proxy.com:8443')")
 	cfg.AddSetting(NoProxy, "", validateNoProxy, SuccessfullyApplied,
 		"Hosts, ipv4 addresses or CIDR which do not use a proxy (string, comma-separated list such as '127.0.0.1,192.168.100.1/24')")
-	cfg.AddSetting(ProxyCAFile, "", validatePath, SuccessfullyApplied,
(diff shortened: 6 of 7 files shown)


cfergeau:
Can `config: make proxy-ca-file only accept absolute paths` cause existing configs to stop working?
Would it be possible to convert relative to absolute and to use absolute throughout crc's codebase? This would be more user-friendly.

anjannath:
> Can `config: make proxy-ca-file only accept absolute paths` cause existing configs to stop working? 

no, existing configs should keep working, the stricter validation only happens when user configures a value via `crc config set`, when retrieving the value for use by other parts in the codebase the validation will not run

> Would it be possible to convert relative to absolute and to use absolute throughout crc's codebase? This would be more user-friendly.

yeah it'll better, but this would mean we need to mutate the config value after user has provided it, currently in the config pkg we don't have such provisions, i was thinking to follow the same pattern as the _change notifiers_ (and call it _mutators_) did you have another idea on how to implement this?

cfergeau:
> > Would it be possible to convert relative to absolute and to use absolute throughout crc's codebase? This would be more user-friendly.
> 
> yeah it'll better, but this would mean we need to mutate the config value after user has provided it, currently in the config pkg we don't have such provisions, i was thinking to follow the same pattern as the _change notifiers_ (and call it _mutators_) did you have another idea on how to implement this?

I had https://github.com/cfergeau/crc/commit/2d5e48887ba1a537bd8ab949e9283f31e8898beb in the back of my mind, which seems to work as expected after some quick prototyping (only test I did is what is in the unit test)

anjannath:
> I had [cfergeau@2d5e488](https://github.com/cfergeau/crc/commit/2d5e48887ba1a537bd8ab949e9283f31e8898beb) in the back of my mind, which seems to work as expected after some quick prototyping (only test I did is what is in the unit test)

i was thinking to implement it like this: https://github.com/anjannath/crc/commit/5d87b1dff4328194939d2901490f46f702b536b8 but with the `Path` type i think its better, as while adding the config setting its clear, will take your commit do some more testing and update the PR, thanks!

pkg/crc/network/httpproxy/proxy.go
```diff
@@ -66,10 +68,14 @@ func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAFile string) (*Prox
 	envProxy := httpproxy.FromEnvironment()
 
 	if DefaultProxy.HTTPProxy == "" {
-		DefaultProxy.HTTPProxy = envProxy.HTTPProxy
+		if err := ValidateProxyURL(envProxy.HTTPProxy, false); err == nil {
+			DefaultProxy.HTTPProxy = envProxy.HTTPProxy
```
cfergeau: When there is an error, I'd add a warning so that it's clear proxy from the environment is being ignored.

pkg/crc/network/httpproxy/proxy.go
```diff
@@ -45,6 +46,9 @@ func readProxyCAData(proxyCAFile string) (string, error) {
 	if proxyCAFile == "" || !filepath.IsAbs(proxyCAFile) {
 		return "", nil
 	}
+	if !crcos.FileExists(proxyCAFile) {
+		return "", nil
```
cfergeau: Is there a validation happening later on? Ideally, most crc commands would not fail when the proxy config is invalid. It's probably only `crc config`, `crc setup` and `crc start` which should report a hard failure when the proxy is invalid, the other ones are not dependent on the proxy settings.

anjannath: > Is there a validation happening later on? 

no, there's validation when running `config set`. and then here when proxy is being setup which is very early on in `root.go`.  (we send telemetry data from each command, and that uses the proxy settings)

> It's probably only crc config, crc setup and crc start which should report a hard failure when the proxy is invalid, the other ones are not dependent on the proxy settings.

maybe for these commands we can re-validate (add a "re-validation" helper in proxy package) the proxy settings and this will be hard failure if validation fails

pkg/crc/network/httpproxy/proxy.go
anjannath: added a warning msg

pkg/crc/network/httpproxy/proxy.go
anjannath: Thinking about this again, doing this selective hard failure for `start` and `setup` seems not to be necessary, as we have added validation of proxy config from environment, now it should not be possible to configure invalid proxy settings at all via the environment or via `crc config set`

for the `proxy-ca-file` we have validation while `config set` is used to make sure the file exists, and with the conversion of relative to absolute paths, user shouldn't end up in a situation where they have invalid path to `proxy-ca-file`

now the only way invalid proxy settings can end up in the config is by manually editing the config file, which i think should be a hard failure for all the commands, WDYT?

cfergeau: > for the proxy-ca-file we have validation while config set is used to make sure the file exists, and with the conversion of relative to absolute paths, user shouldn't end up in a situation where they have invalid path to proxy-ca-file

it's a user problem, but
```
$ crc config set proxy-ca-file foo.crt
$ rm foo.crt
```

> Thinking about this again, doing this selective hard failure for start and setup seems not to be necessary, as we have added validation of proxy config from environment, now it should not be possible to configure invalid proxy settings at all via the environment or via crc config set

We validate the proxy config, and if for example `HTTP_PROXY` is invalid in the environment, we ignore it right?
If the user made a typo when setting it, it's better to make this very visible than ignoring it, no?

cfergeau: If there are already warnings for all of this every time crc starts up, it might be enough ,I'll have to check how it looks :)

anjannath: > We validate the proxy config, and if for example HTTP_PROXY is invalid in the environment, we ignore it right?
If the user made a typo when setting it, it's better to make this very visible than ignoring it, no?

yes i changed that its not silently ignoring it now. If the configured `proxy-ca-file` path is not present, it returns an error which is a hard failure for the all the commands, also with a msg asking user to either unset or correct the config

pkg/crc/network/httpproxy/proxy.go
```diff
@@ -41,7 +42,7 @@ func (p *ProxyConfig) String() string {
 }
 
 func readProxyCAData(proxyCAFile string) (string, error) {
-	if proxyCAFile == "" {
+	if proxyCAFile == "" || !filepath.IsAbs(proxyCAFile) {
```
cfergeau: I'm not sure why we want to ignore the error in this situation?

pkg/crc/network/httpproxy/proxy.go
```diff
@@ -75,11 +75,15 @@ func NewProxyDefaults(httpProxy, httpsProxy, noProxy, proxyCAFile string) (*Prox
...
 			DefaultProxy.HTTPProxy = envProxy.HTTPProxy
+		} else {
+			logging.Warnf("HTTP_PROXY environment variable is being ignored as it failed validation: %v", err)
 		}
 	}
 	if DefaultProxy.HTTPSProxy == "" {
 		if err := ValidateProxyURL(envProxy.HTTPSProxy, true); err == nil {
 			DefaultProxy.HTTPSProxy = envProxy.HTTPSProxy
+		} else {
+			logging.Warnf("HTTPS_PROXY environment variable is being ignored as it failed validation: %v", err)
```
cfergeau: Can these 2 warnings be moved to the first commit?

pkg/crc/network/httpproxy/proxy.go
```diff
@@ -260,3 +264,16 @@ func HTTPTransport() http.RoundTripper {
...
 	return proxyConfig.HTTPTransport()
 }
+
+func RevalidateProxyCAFile() error {
+	proxyConfig, err := NewProxyConfig()
+	if err != nil {
+		return err
+	}
+	if proxyConfig.ProxyCAFile != "" {
+		if proxyConfig.ProxyCACert == "" {
```
cfergeau: The 2 commits before this one ignore various errors happening in `readProxyCAData`, this check for `ProxyCACert` is an indirect way of checking if we ignored any errors.
We also ignore invalid proxy settings from the environment, but this will not be detected here (not necessarily mandatory, but would be nice if everything could be reported to the user).

The commit logs are short on details about the bugs that are being fixed, so I'd need to debug this again to see if there are easy alternatives I can suggest to improve this :(

cfergeau (commented):
This PR ignores the errors which block startup, and then the code tries to figure out if an error was ignored to report it at the right time. Ideally we could do something more direct, but I'll need to look closer at it.

cfergeau:
These are these lines which are causing an error to happen very early in crc startup:
https://github.com/crc-org/crc/blob/85cc7183e4daf12b148b594a444e1d50546910e3/cmd/crc/cmd/root.go#L58-L60

Would be useful to mention it in one of the commit logs :)


In my opinion, if we want to ignore errors/output a warning, root.go is the place where we should do it, it's the only place calling NewProxyDefaults().

I think you can also add some kind of 'error' field to ProxyConfig, to record that there was an error reading the CA file, and/or that there was an error parsing the proxy from the env, ... Then it becomes easy to check at a later time if there are proxy config errors.
`start` has a call to `getProxyConfig()` relatively early during startup so an explicit "revalidateConfig" may not be needed there.

anjannath:
@cfergeau I've updated the PR with more descriptive commit msg and changed the ignoring of error to happen in root.go, also added a commit to let users know when invalid config is preset in the env variables, PTALA :)

pkg/crc/config/validations.go
```diff
@@ -103,6 +104,17 @@ func validatePath(value interface{}) (bool, string) {
 	return true, ""
 }
 
+func validatePathAbs(value interface{}) (bool, string) {
```
cfergeau: This could be named `validateAbsolutePath`?

cfergeau (commented):
What happens if the user's config has a relative path to a CA file, and they upgrade to a version of crc with validateAbsolutePath? A warning will be printed? Will they still be able to use their relative path CA if it worked for them before?

anjannath:
> What happens if the user's config has a relative path to a CA file, and they upgrade to a version of crc with validateAbsolutePath? A warning will be printed? Will they still be able to use their relative path CA if it worked for them before?

yes, it'll keep working, since the validation only happens when user runs `crc config set proxy-ca-file` but there will be no warning also to ask them to migrate to absolute path..

anjannath:
I forgot about fc4dbfb4e504aaf439557f2cbadb16efd4e902c2 which is more user friendly than the `validaeAbsolutePath`

anjannath:
@cfergeau updated :)

pkg/crc/config/config.go
```diff
@@ -239,11 +239,20 @@ func (c *Config) Get(key string) SettingValue {
 	}
 	return SettingValue{
 		Value:     value,
-		IsDefault: reflect.DeepEqual(setting.defaultValue, value),
```
cfergeau: Good catch!
There's a 
```
case Path, string:
		value = cast.ToString(value)
```
in this function, did you try changing it to:
```
case Path: value = Path(cast.ToString(value))
```
?

anjannath: Yes tried this now, and it also works, updated to use this instead..

pkg/crc/config/config.go
```diff
@@ -200,6 +211,7 @@ func (c *Config) Get(key string) SettingValue {
 		}
 	case string:
 		value = cast.ToString(value)
+	case Path:
```
cfergeau, line 212: Imo this should be `value = Path(cast.ToString(value))`

pkg/crc/config/settings_test.go
```diff
@@ -158,3 +159,19 @@ func TestUnsetPreset(t *testing.T) {
 		IsSecret:  false,
 	}, cfg.Get(CPUs))
 }
+
+func TestPath(t *testing.T) {
+	cfg, err := newInMemoryConfig()
+	require.NoError(t, err)
```
cfergeau, line 165: A test for the default value would be useful imo as this was initially buggy.
Something like this should do the trick:

```
	assert.Equal(t, SettingValue{
		Value:     Path(""),
		Invalid:   false,
		IsDefault: true,
		IsSecret:  false,
	}, cfg.Get(ProxyCAFile))
```

pkg/crc/config/config.go
anjannath: Doing this fails the existing test case, had to wrap the expected value in `Path` and `config view` was not showing the `proxy-ca-file` value for that to work had to also add a stringer method for the `Path` type

cfergeau (approved):

PR #3817, opened 2023-09-04
Refactor: Move logic to get default image uri to constants

praveenkumar:
Right now, to pull default image we pass empty string to `image/PullBundle` function and have logic to get default image URI in case the empty string for URI which is bit confusing. In this PR we are adding `GetDefaultBundleImageRegistryURL` function to constants pkg to get this info and change `image/PullBundle` signature.

### DIFF
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -221,3 +221,18 @@ func GetDefaultMemory(preset crcpreset.Preset) int {
 		return 9216
 	}
 }
+
+func GetDefaultBundleImageRegistry(preset crcpreset.Preset) string {
+	return fmt.Sprintf("//%s/%s:%s", RegistryURI, getImageName(preset), version.GetBundleVersion(preset))
+}
+
+func getImageName(preset crcpreset.Preset) string {
+	switch preset {
+	case crcpreset.Podman:
+		return "podman-bundle"
+	case crcpreset.OKD:
+		return "okd-bundle"
+	default:
+		return "openshift-bundle"
+	}
+}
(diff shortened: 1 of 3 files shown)


pkg/crc/constants/constants.go
```diff
@@ -221,3 +221,18 @@ func GetDefaultMemory(preset crcpreset.Preset) int {
 		return 9216
 	}
 }
+
+func GetDefaultBundleImageRegistryURL(preset crcpreset.Preset) string {
+	return fmt.Sprintf("//%s/%s:%s", RegistryURI, getImageName(preset), version.GetBundleVersion(preset))
```
cfergeau, line 226: `docker://%s/%s:%s` for consistency with the URL parsing code there is in the `bundle` package?

pkg/crc/machine/bundle/metadata.go
```diff
@@ -355,13 +355,13 @@ func Download(preset crcPreset.Preset, bundleURI string) (string, error) {
 		if preset == crcPreset.OpenShift || preset == crcPreset.Microshift {
 			return DownloadDefault(preset)
 		}
-		return image.PullBundle(preset, "")
+		return image.PullBundle(constants.GetDefaultBundleImageRegistryURL(preset))
```
cfergeau: I'd rewrite this as a `switch/case`, but I can add this to https://github.com/crc-org/crc/pull/3813

cfergeau (approved):
Looks good to me!

pkg/crc/constants/constants.go
cfergeau: I'm not even sure `//%s/%s` is a URL, which would be contrary to what the function name implies (`GetDefault...URL`)

Issue #3747, opened 2023-06-30
[BUG] microshift preset only allows 15 Gi of PVCs


computate:
### General information

  * OS: RHEL 9.2
  * Hypervisor: OpenShift Local with `microshift` preset, Libvirt
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop

## CRC version
```bash
$ crc version
CRC version: 2.22.1+e8068b4
OpenShift version: 4.13.3
Podman version: 4.4.4
```
  
## CRC status
```bash
$ crc status
CRC VM:          Running
MicroShift:      Running (v4.13.3)
RAM Usage:       8.47GB of 29.49GB
Disk Usage:      15GB of 198.6GB (Inside the CRC VM)
Cache Usage:     142.6GB
Cache Directory: /home/ctate/.crc/cache
```

## CRC config
```bash
$ crc config view
- consent-telemetry                     : yes
- cpus                                  : 7
- disk-size                             : 201
- memory                                : 29000
- preset                                : microshift
```

## Host Operating System
```bash
$ cat /etc/os-release 
NAME="Red Hat Enterprise Linux"
VERSION="9.2 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.2"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.2 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.2"
```

### Steps to reproduce

```bash
crc config set preset microshift
crc config set cpus 7
crc config set memory 29000
crc config set disk-size 200
crc setup
crc start
crc stop
crc config set disk-size 201
crc start
```

### Expected

I expect more than 15Gi of PVCs to be status `Bound`. 

```bash
$ oc get pvc -A
[...]

PR #3820, opened 2023-09-05
Add persistent volume size option to config

praveenkumar:
This pr enable user to specify the size which is used by persistent volume in case of microshift preset. As of now we just extend the disk size to root mounted partition and free space (which used for pvc) is only 15GiB. Some users want to have more free space to consume it for pvc as per the workload and with this option now it is possible.

```
$ crc config set disk-size 50
$ crc config set persistent-volume-size 20
$ crc start
[...]
<crc_vm> $ sudo vgdisplay
  --- Volume group ---
  VG Name               rhel
  System ID
  Format                lvm2
  Metadata Areas        1
  Metadata Sequence No  5
  VG Access             read/write
  VG Status             resizable
  MAX LV                0
  Cur LV                1
  Open LV               1
  Max PV                0
  Cur PV                1
  Act PV                1
  VG Size               <49.02 GiB
  PE Size               4.00 MiB
  Total PE              12549
  Alloc PE / Size       7429 / <29.02 GiB
  Free  PE / Size       5120 / 20.00 GiB
  VG UUID               O7d5v9-F8HE-E0Dj-e9sE-1qgv-F74d-CZz34d
```

**Fixes:** Issue #3747

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -35,6 +35,7 @@ const (
 	IngressHTTPPort         = "ingress-http-port"
 	IngressHTTPSPort        = "ingress-https-port"
 	EmergencyLogin          = "enable-emergency-login"
+	PersistentVolumeSize    = "persistent-volume-size"
 )
 
 func RegisterSettings(cfg *Config) {
@@ -91,6 +92,8 @@ func RegisterSettings(cfg *Config) {
 		"Enable experimental features (true/false, default: false)")
 	cfg.AddSetting(EmergencyLogin, false, ValidateBool, SuccessfullyApplied,
 		"Enable emergency login for 'core' user. Password is randomly generated. (true/false, default: false)")
+	cfg.AddSetting(PersistentVolumeSize, constants.DefaultPersistentVolumeSize, validatePersistentVolumeSize, SuccessfullyApplied,
+		fmt.Sprintf("Total size in GiB of the persistent volume used by the CSI driver for %s preset (must be greater than or equal to '%d')", preset.Microshift, constants.DefaultPersistentVolumeSize))
 
 	// Shared directories configs
 	if runtime.GOOS == "windows" {
--- a/pkg/crc/config/validations.go
+++ b/pkg/crc/config/validations.go
@@ -41,6 +41,19 @@ func validateDiskSize(value interface{}) (bool, string) {
 	return true, ""
 }
 
+// validatePersistentVolumeSize checks if provided disk size is valid in the config
+func validatePersistentVolumeSize(value interface{}) (bool, string) {
+	diskSize, err := cast.ToIntE(value)
+	if err != nil {
+		return false, fmt.Sprintf("could not convert '%s' to integer", value)
+	}
+	if err := validation.ValidatePersistentVolumeSize(diskSize); err != nil {
+		return false, err.Error()
+	}
+
+	return true, ""
+}
+
 // validateCPUs checks if provided cpus count is valid in the config
 func validateCPUs(value interface{}, preset crcpreset.Preset) (bool, string) {
 	v, err := cast.ToIntE(value)
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -114,7 +114,7 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
 	return nil
 }
 
-func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error {
+func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset, persistentVolumeSize int) error {
 	rootPart, err := getrootPartition(sshRunner, preset)
 	if err != nil {
 		return err
@@ -135,7 +135,7 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset) error
 
 	if preset == crcPreset.Microshift {
 		lvFullName := "rhel/root"
-		if err := growLVForMicroshift(sshRunner, lvFullName, rootPart); err != nil {
+		if err := growLVForMicroshift(sshRunner, lvFullName, rootPart, persistentVolumeSize); err != nil {
 			return err
 		}
 	}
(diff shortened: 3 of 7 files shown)


jsliacan:
So PVC size comes out of the disk size, right? I.e. 0GiB <= pvc size <= disk size - 15GiB? I'm assuming that the last inequality is actually strict to leave space for other things. What would be the maximum possible pvc size for a fixed disk size?

pkg/crc/machine/start.go
```diff
@@ -197,8 +197,8 @@ func growLVForMicroshift(sshRunner *crcssh.Runner, lvFullName string, rootPart s
 		return err
 	}
 
-	// vgFree space as part of the bundle is default to ~15.02G (16127098880 byte)
-	vgFree := 16127098880
+	// 1GB = 1073741824 bytes
+	vgFree := persistentVolumeSize * 1073741824
```
gbraad:
```
GB := 107374... // bytes
vgFree := persistentVolumeSize * GB
```
which makes things s stand out more clearly than the comment

pkg/crc/config/settings.go
```diff
@@ -91,6 +92,8 @@ func RegisterSettings(cfg *Config) {
 		"Enable experimental features (true/false, default: false)")
 	cfg.AddSetting(EmergencyLogin, false, ValidateBool, SuccessfullyApplied,
 		"Enable emergency login for 'core' user. Password is randomly generated. (true/false, default: false)")
+	cfg.AddSetting(PersistentVolumeSize, constants.DefaultPersistentVolumeSize, validatePersistentDiskSize, SuccessfullyApplied,
+		fmt.Sprintf("Total size in GiB of the persistent volume (must be greater than or equal to '%d')", constants.DefaultPersistentVolumeSize))
```
gbraad: used for what?
it misses explanation as containers arent stored here

gbraad:
@jsliacan I had the same question just now pop up. also, what is the usage as that isnt explained to the user. ephemeral and persistent have clear usecases,...

pkg/crc/config/settings.go
praveenkumar: I am going to use https://github.com/openshift/microshift/blob/main/docs/contributor/rhel4edge_iso.md#disk-partitioning same explanation `Unallocated disk space of 9GB size remains in the rhel volume group to be used by the CSI driver.`

praveenkumar:
> So PVC size comes out of the disk size, right? I.e. 15GiB <= pvc size <= disk size? I'm assuming that the last inequality is actually strict to leave space for other things. What would be the maximum possible pvc size for a fixed disk size?

@jsliacan As of now maximum pvc size user can allocate `<Total Disk Size> - 15G`  so if user put 40GB as disk size then he can use ~ 25G for pvc.

pkg/crc/validation/validation.go
```diff
@@ -42,6 +42,14 @@ func ValidateDiskSize(value int) error {
 	return nil
 }
 
+func ValidatePersistentDiskSize(value int) error {
```
anjannath: maybe change this to also `validatePersistentVolumeSize` we are mixing `volume` and `disk` in the new names.. `persistentVolume` vs `persistentDisk`

pkg/crc/config/settings.go
```diff
@@ -91,6 +92,8 @@ func RegisterSettings(cfg *Config) {
 		"Enable experimental features (true/false, default: false)")
 	cfg.AddSetting(EmergencyLogin, false, ValidateBool, SuccessfullyApplied,
 		"Enable emergency login for 'core' user. Password is randomly generated. (true/false, default: false)")
+	cfg.AddSetting(PersistentVolumeSize, constants.DefaultPersistentVolumeSize, validatePersistentDiskSize, SuccessfullyApplied,
+		fmt.Sprintf("Total size in GiB of the persistent volume used by the CSI driver (must be greater than or equal to '%d')", constants.DefaultPersistentVolumeSize))
```
anjannath: should we also mention that this applies to only the `microshift` preset?

pkg/crc/validation/validation.go
praveenkumar: updated/

pkg/crc/config/settings.go
praveenkumar: Done

pkg/crc/validation/validation.go
```diff
@@ -42,6 +42,14 @@ func ValidateDiskSize(value int) error {
 	return nil
 }
 
+func ValidatePersistentVolumeSize(value int) error {
+	if value < constants.DefaultPersistentVolumeSize {
+		return fmt.Errorf("requires disk size in GiB >= %d", constants.DefaultDiskSize)
```
anjannath: this `value` should be larger than `DefaultPersistentVolumeSize` (15gb) or `DefaultDiskSize` (31gb) ?

pkg/crc/config/validations.go
```diff
@@ -41,6 +41,19 @@ func validateDiskSize(value interface{}) (bool, string) {
 	return true, ""
 }
 
+// validatePersistentDiskSize checks if provided disk size is valid in the config
+func validatePersistentDiskSize(value interface{}) (bool, string) {
```
anjannath: this can also be renamed to `validatePersistentVolumeSize` i think, since this is only used for the persistent volume config validation and not used for `disk-size` config option?

pkg/crc/validation/validation.go
praveenkumar: Ah, good catch. Fixed 👍🏼

pkg/crc/config/validations.go
praveenkumar: yes this is also renamed.

pkg/crc/config/settings.go
```diff
@@ -91,6 +92,8 @@ func RegisterSettings(cfg *Config) {
 		"Enable experimental features (true/false, default: false)")
 	cfg.AddSetting(EmergencyLogin, false, ValidateBool, SuccessfullyApplied,
 		"Enable emergency login for 'core' user. Password is randomly generated. (true/false, default: false)")
+	cfg.AddSetting(PersistentVolumeSize, constants.DefaultPersistentVolumeSize, validatePersistentVolumeSize, SuccessfullyApplied,
+		fmt.Sprintf("Total size in GiB of the persistent volume used by the CSI driver for %s preset (must be greater than or equal to '%d')", preset.Microshift, constants.DefaultPersistentVolumeSize))
```
anjannath, line 96: ahh, sorry i forgot to mention this earlier, maybe we should use `preset.Microshift.ForDisplay()` here

anjannath (approved):
LGTM

pkg/crc/config/settings.go
praveenkumar: @anjannath since preset have string method `%s` should able to get preset and Display is for end user display I think here they should know the `preset` value instead of Display?

Issue #3771, opened 2023-07-26
[integration-test] disk-resize test failing on openshift CI


anjannath:
Recently (since ~4 sprints back) on the CI disk-resize integration test is failing, it does succeed always upon a retry

praveenkumar:
@anjannath can you put the logs or the link to failled CI job?

anjannath:
yes, here's the link to the logs: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/crc-org_crc/3769/pull-ci-crc-org-crc-main-integration-crc/1684072012844109824

praveenkumar:
As per job history https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-crc-org-crc-main-integration-crc this is not looking good (3 success out of 20 runs). @jsliacan Are you planning to take a look into this?

jsliacan:
Yes yes, I put it on the list in planning yesterday. Although, @praveenkumar not all failures are due to disk-resize. The goproxy sometimes fails due to admin-context login. That way of logging in was a bit unstable even before (we removed it from e2e for this reason IIRC). 

Anyways, I am taking a look at this in the next few days!

jsliacan:
I don't think this is an issue anymore after #3821 went in, let's reopen if the issue re-emerges.

PR #3821, opened 2023-09-05
[Integration] Fixes for integration tests

jsliacan:
**Fixes:** Issue #3771

**Related to** Issue #3796

### DIFF
--- a/test/extended/crc/cmd/cmd.go
+++ b/test/extended/crc/cmd/cmd.go
@@ -140,11 +140,14 @@ func CheckCRCStatus(state string) error {
 		expression = `.*CRC VM:.*\s.*: .*Stopped.*`
 	}
 
-	err := util.ExecuteCommand(CRC("status").ToString())
+	cmd := exec.Command("crc", "status")
+	out, err := cmd.Output()
 	if err != nil {
 		return err
 	}
-	return util.CommandReturnShouldMatch("stdout", expression)
+
+	err = util.CompareExpectedWithActualMatchesRegex(expression, string(out))
+	return err
 }
 
 func CheckCRCExecutableState(state string) error {
--- a/test/extended/util/proxy.go
+++ b/test/extended/util/proxy.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 
 	"github.com/elazarl/goproxy"
 	log "github.com/sirupsen/logrus"
@@ -64,8 +65,13 @@ func RunProxy() {
 
 	log.SetOutput(f)
 
+	ipaddr := "127.0.0.1"        // user mode is default on windows and darwin
+	if runtime.GOOS == "linux" { // system mode is default on linux
+		ipaddr = "192.168.130.1"
+	}
+
 	verbose := flag.Bool("v", true, "should every proxy request be logged to stdout")
-	addr := flag.String("127.0.0.1", ":8888", "proxy listen address") // using network-mode=user
+	addr := flag.String(ipaddr, ":8888", "proxy listen address") // using network-mode=user
 	flag.Parse()
 	proxy.Verbose = *verbose
 	proxy.Logger = log.StandardLogger()
--- a/test/integration/podman_test.go
+++ b/test/integration/podman_test.go
@@ -10,9 +10,9 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("podman preset", Label("podman-preset"), func() {
+var _ = Describe("podman preset", Serial, Ordered, Label("podman-preset"), func() {
 
-	Describe("basic use", func() {
+	Describe("basic use", Serial, Ordered, func() {
 
 		It("write to config", func() {
 			Expect(RunCRCExpectSuccess("config", "set", "preset", "podman")).To(ContainSubstring("please run 'crc setup' before 'crc start'"))
--- a/test/integration/podman_test.go
+++ b/test/integration/podman_test.go
@@ -6,11 +6,24 @@ import (
 	"runtime"
 
 	"github.com/crc-org/crc/v2/test/extended/crc/cmd"
+	"github.com/crc-org/crc/v2/test/extended/util"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("podman preset", Serial, Ordered, Label("podman-preset"), func() {
+
+	// runs 1x after all the It blocks (specs) inside this Describe node
+	AfterAll(func() {
+
+		// cleanup CRC
+		Expect(RunCRCExpectSuccess("cleanup")).To(MatchRegexp("Cleanup finished"))
+
+		// remove config file crc.json
+		err := util.RemoveCRCConfig()
+		Expect(err).NotTo(HaveOccurred())
+
+	})
 
 	Describe("basic use", Serial, Ordered, func() {
 
--- a/test/integration/podman_test.go
+++ b/test/integration/podman_test.go
@@ -50,7 +50,7 @@ var _ = Describe("podman preset", Serial, Ordered, Label("podman-preset"), func(
 				unexpandedPath := filepath.Join(userHomeDir, ".crc/bin/oc;${PATH}")
 				path = os.ExpandEnv(unexpandedPath)
 				csshk = filepath.Join(userHomeDir, ".crc/machines/crc/id_ecdsa")
-				dh = "npipe:////./pipe/rc-podman"
+				dh = "npipe:////./pipe/crc-podman"
 			}
 			if runtime.GOOS == "linux" {
 				ch = "ssh://core@192.168.130.11:22/run/user/1000/podman/podman.sock"
--- a/test/integration/proxy_test.go
+++ b/test/integration/proxy_test.go
@@ -8,11 +8,11 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("", Label("openshift-preset", "goproxy"), func() {
+var _ = Describe("", Serial, Ordered, Label("openshift-preset", "goproxy"), func() {
 
 	go util.RunProxy()
 
-	Describe("Behind proxy", func() {
+	Describe("Behind proxy", Serial, Ordered, func() {
 
 		networkMode := "user"
 		httpProxy := "http://127.0.0.1:8888"
--- a/test/integration/proxy_test.go
+++ b/test/integration/proxy_test.go
@@ -1,7 +1,9 @@
 package test_test
 
 import (
+	"os"
 	"os/exec"
+	"runtime"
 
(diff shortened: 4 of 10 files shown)


test/integration/proxy_test.go
```diff
@@ -35,7 +35,9 @@ var _ = Describe("", Serial, Ordered, Label("openshift-preset", "goproxy"), func
 		// Start goproxy
 
 		It("configure CRC", func() {
-			Expect(RunCRCExpectSuccess("config", "set", "network-mode", networkMode), ContainSubstring("Network mode"))
+			if runtime.GOOS == "linux" {
+				Expect(RunCRCExpectSuccess("config", "set", "network-mode", networkMode), ContainSubstring("Network mode"))
+			}
```
cfergeau: I'd argue that it's not very useful to test a non-default network mode, as this is not what most users will experience. Are there some issues if we run with the default network mode on linux?

gbraad: @jsliacan WDYS?

gbraad (commented):
overall OK. moving to a serial execution... Does this add to the total runtime?

the networkmode is an open question

test/integration/proxy_test.go
jsliacan: The reason why it was this way was mostly convenience - at the time when I wrote that test, you could set network mode on all platforms, so no extra code needed for linux. Also, linux system mode is subject to the e2e proxy test (which is only for linux). So this seemed to cover another (albeit less useful) case of user mode on linux.

That said, I see how it's a non-default option and thus might not be worth testing at this point. If you say that that's the case I'm happy to parametrize the proxy address by platform and use system mode on linux.

cfergeau: >  If you say that that's the case I'm happy to parametrize the proxy address by platform and use system mode on linux.

Ah, if this makes the proxy test/code simpler, this can be a reason for using `user` mode networking on linux. Let's keep this for now to try to fix integration tests.

cfergeau: With system networking, it's https://github.com/crc-org/crc/blob/85cc7183e4daf12b148b594a444e1d50546910e3/test/extended/util/proxy.go#L68 which would need to be changed to something reachable by the VM?

NB: should have looked at your recent changes before commenting ^^

cfergeau: I'm surprised this works with user-mode networking though if you don't' set `* host-network-access                        Allow TCP/IP connections from the CRC VM to services running on the host (true/false, default: false)`

cfergeau: NB: should have looked at your recent changes before commenting ^^

test/integration/podman_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"runtime"
 
 	"github.com/crc-org/crc/v2/test/extended/crc/cmd"
-	"github.com/crc-org/crc/test/extended/util"
+	"github.com/crc-org/crc/v2/test/extended/util"
```
cfergeau, line 9: This can be squashed in " integration: cleanup to AfterAll and BeforeSuite "

cfergeau (approved):
Really not clear why the tests would still be failing :( 
But looks good to me, ideally the last commit would be squashed to the right place.

test/extended/crc/cmd/cmd.go
```diff
@@ -140,11 +140,14 @@ func CheckCRCStatus(state string) error {
 		expression = `.*CRC VM:.*\s.*: .*Stopped.*`
 	}
 
-	err := util.ExecuteCommand(CRC("status").ToString())
+	cmd := exec.Command("crc", "status")
+	out, err := cmd.Output()
```
cfergeau, line 144: Why do we want to avoid this?

test/extended/util/proxy.go
```diff
@@ -68,7 +68,9 @@ func RunProxy() {
 	addr := flag.String("127.0.0.1", ":8888", "proxy listen address") // using network-mode=user
 	flag.Parse()
 	proxy.Verbose = *verbose
-	proxy.Logger = log.StandardLogger()
+	l := log.New()
+	l.Out = f
+	proxy.Logger = l
```
cfergeau: This seems out of place in `integration: cleanup to AfterAll and BeforeSuite`.
Is this required to fix something?

cfergeau (commented):
2 minor comments, otherwise looks good!

test/extended/crc/cmd/cmd.go
jsliacan: Because I need to explicitly get the stdout to check what was in it, and `util.ExecuteCommand` does not return that and isn't easily modifiable to do so (and even if it was, it's used in many many places which would need changing too). So going around it like above seemed better. I hope I am not overlooking something here, but that was the reasoning.

test/extended/util/proxy.go
jsliacan: Fair, it is out of place. I was reading something about logging and felt the need to change this asap :). Will remove it from this PR! Thanks

test/extended/crc/cmd/cmd.go
cfergeau: looking at `util/shell.go` code, `CommandReturnShouldMatch("stdout"` calls `CompareExpectedWithActualContains(expected, shell.GetLastCmdOutput("stdout"))` which is supposed to get the command stdout output. Not clear why it's not working for you :-/

cfergeau: I suggest keeping this one out of the PR so that we can merge it, and to address issues related to this in a follow-up PR.

test/integration/podman_test.go
```diff
@@ -6,11 +6,24 @@ import (
 	"runtime"
 
 	"github.com/crc-org/crc/v2/test/extended/crc/cmd"
+	"github.com/crc-org/crc/v2/test/extended/util"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("podman preset", Serial, Label("podman-preset"), func() {
+var _ = Describe("podman preset", Serial, Ordered, Label("podman-preset"), func() {
```
cfergeau, line 14: Is it intentional that the `Ordered` addition is in this `integration: cleanup to AfterAll and BeforeSuite` commit and not in `integration: run ginkgo nodes serially and in ordered fashion`? If it's not intentional, I've already split it in https://github.com/cfergeau/crc/commits/ginkgo-issue

test/extended/crc/cmd/cmd.go
jsliacan: In integration tests, we do not use the shell construct (to simulate a user interacting with crc in one shell session). In e2e, we do (initialized [here](https://github.com/crc-org/crc/blob/02e1f697a0f8d8882e12684835a5ea1d683753fa/test/e2e/testsuite/testsuite.go#L157)): 
```go
err := util.StartHostShellInstance(testWithShell)
```
We had issues with running `crc setup` (with all the progress bars etc.) in e2e tests and had to move the execution of that command out of the general shell. So with integration, we didn't use the shell construct at all.

Given that we do not have an instance of [ShellInstance](https://github.com/crc-org/crc/blob/02e1f697a0f8d8882e12684835a5ea1d683753fa/test/extended/util/shell.go#L48) available, functions which rely on it won't work as expected. Therefore, instead of retrieving the stdout content from the `shell` variable that would contain it if it existed, we need to pass it between command execution (`crc status`) and output analysis (does it contain `Running`?) by hand. And that's what this commit does. Without it, `WaitForClusterInState` won't have any effect (except for wasting 10mins). 

So if we omit this change from this PR, integration tests will keep failing, albeit with a different error than initially when we started out.

jsliacan:
I forgot that I need to unset HTTPS_PROXY and HTTP_PROXY env variables that are implicitly set during the test.

test/extended/crc/cmd/cmd.go
cfergeau: This would be a nice explanation to have in the commit log ;) 

`CheckCRCStatus` is used by both e2e and integration tests, is it ok to remove the use of `shell` for e2e?
And what about all the users of `ExecuteCommand()` in cmd.go? They also rely on `shell.go`, is this going to cause issues?

If this change is good enough to fix integration tests without breaking e2e, we can go with it, but some follow-up cleanups would be very useful imo (a clean separation between helpers using `exec.Command`, and helpers using `shell.go`). The former would only be used in integration tests, and the latter in e2e.

jsliacan: > This would be a nice explanation to have in the commit log ;)

Will add it there.

> CheckCRCStatus is used by both e2e and integration tests, is it ok to remove the use of shell for e2e?

It is ok in the sense that it'll work. However, it'll dent the effort to simulate single-session user interaction with CRC.

> And what about all the users of ExecuteCommand() in cmd.go? They also rely on shell.go, is this going to cause issues?

As long as you don't need to retrieve anything aside from the error, it will do the job the same way.

> some follow-up cleanups would be very useful imo

I don't know the extent of this. It could have further consequences -- maybe it makes less sense to share the code between e2e and integration then. I opened an issue to track this: #3859

cfergeau: > maybe it makes less sense to share the code between e2e and integration then

Maybe some go interface can be defined and used in the shared code. The interface would either make use of the shell code to run the command/get stdout/..., or would directly run the command for the integration tests.

cfergeau (approved):
I do not see https://github.com/crc-org/crc/pull/3821/commits/497c1d0000704e2822613402528ec721cde112fa#r1346340794 in the commit log for " tests: avoid using shell struct in CheckCRCStatus" but apart from this, ACK.

jsliacan:
Last changes include `git reset` onto @cfergeau 's branch with improved commits, additional `unset HTTP(S)_PROXY` in AfterAll, and updated commit message to explain why we're changing `CheckCRCStatus` to avoid using `ShellInstance`.

test/integration/podman_test.go
```diff
@@ -50,7 +50,7 @@ var _ = Describe("podman preset", Serial, Ordered, Label("podman-preset"), func(
 				unexpandedPath := filepath.Join(userHomeDir, ".crc/bin/oc;${PATH}")
 				path = os.ExpandEnv(unexpandedPath)
 				csshk = filepath.Join(userHomeDir, ".crc/machines/crc/id_ecdsa")
-				dh = "npipe:////./pipe/rc-podman"
+				dh = "npipe:////./pipe/crc-podman"
```
cfergeau, line 53: This also happens in the E2E tests

cfergeau (approved):

test/integration/podman_test.go
jsliacan: True, thanks. Grepped it now and the e2e occurrence was the last one. Changed it and updated the commit too.

cfergeau (approved):

cfergeau:
(waiting for integration-crc to run before merging it)

praveenkumar:
e2e-integration now succeed.

PR #3826, opened 2023-09-06
Validation: Don't fail if the bundle path is same as default path 

praveenkumar:
ValidateBundle function try to get the bundle info if the bundle is
already extracted or run ValidateBundlePath which in case check if the
bundle exist for provided path and for default bundle it give error in
case not persent in the default location. Ideally ValidateBundle should
just return nil in case default bundle path is provided.

With this change, same function can consme as part of `cmd/setup` which
right now have addition logic to ignore default bundle path.

### DIFF
--- a/pkg/crc/preflight/preflight_checks_common.go
+++ b/pkg/crc/preflight/preflight_checks_common.go
@@ -110,17 +110,6 @@ func fixBundleExtracted(bundlePath string, preset crcpreset.Preset) func() error
 		if err := os.MkdirAll(bundleDir, 0775); err != nil {
 			return fmt.Errorf("Cannot create directory %s: %v", bundleDir, err)
 		}
-		if err := validation.ValidateBundle(bundlePath, preset); err != nil {
-			var e *validation.InvalidPath
-			if !errors.As(err, &e) {
-				return err
-			}
-			if bundlePath != constants.GetDefaultBundlePath(preset) {
-				/* This message needs to be improved when the bundle has been set in crc config for example */
-				return fmt.Errorf("%s is invalid or missing, run 'crc setup' to download the bundle", bundlePath)
-			}
-		}
-
 		var err error
 		logging.Infof("Downloading bundle: %s...", bundlePath)
 		if bundlePath, err = bundle.Download(preset, bundlePath); err != nil {
--- a/cmd/crc/cmd/setup.go
+++ b/cmd/crc/cmd/setup.go
@@ -60,10 +60,8 @@ func runSetup(_ []string) error {
 		}
 	}
 
-	if !config.Get(crcConfig.Bundle).IsDefault {
-		if err := validation.ValidateBundlePath(config.Get(crcConfig.Bundle).AsString(), crcConfig.GetPreset(config)); err != nil {
-			return err
-		}
+	if err := validation.ValidateBundle(config.Get(crcConfig.Bundle).AsString(), crcConfig.GetPreset(config)); err != nil {
+		return err
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -110,17 +110,6 @@ func fixBundleExtracted(bundlePath string, preset crcpreset.Preset) func() error
 		if err := os.MkdirAll(bundleDir, 0775); err != nil {
 			return fmt.Errorf("Cannot create directory %s: %v", bundleDir, err)
 		}
-		if err := validation.ValidateBundle(bundlePath, preset); err != nil {
-			var e *validation.InvalidPath
-			if !errors.As(err, &e) {
-				return err
-			}
```
cfergeau, line 117: This may be slightly better than what ValidateBundlePath is doing though, this does not ignore all errors, just invalid paths.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -110,17 +110,6 @@ func fixBundleExtracted(bundlePath string, preset crcpreset.Preset) func() error
...
-			if bundlePath != constants.GetDefaultBundlePath(preset) {
-				/* This message needs to be improved when the bundle has been set in crc config for example */
-				return fmt.Errorf("%s is invalid or missing, run 'crc setup' to download the bundle", bundlePath)
```
cfergeau, line 120: And this error message seems better than what we get in ValidateBundlePath?

cfergeau (commented):
Overall looks good, a few questions though.

pkg/crc/preflight/preflight_checks_common.go
gbraad: not really following what you try to say 🙄

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: But ignoring the invalid path at this step doesn't make sense, this function suppose to fix the bundle extraction and all the errors are handled even invalid path one as part of validation run of different command.

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: Shouldn't we have this as soon as `bundle` argument is used and tell user the provided path is invalid or missing so either user run it with correct path or or just don't provide any bundle path at all.

cfergeau: > And this error message seems better than what we get in ValidateBundlePath?

This removes a relatively clear error message (even if I'm not sure it's ever shown to the user): `fmt.Errorf("%s is invalid or missing, run 'crc setup' to download the bundle", bundlePath)`
This is more useful than:
```
$ ./out/linux-amd64/crc setup --bundle=/foo/bar
file '/foo/bar' does not exist
```

praveenkumar: If that is just related to error message then we can update it to validation function and this error message is not even shown to user (Atleast I didn't find a hack which can use this code flow).

Error can be updated to following as part of validation?
```
fmt.Errorf("%s is invalid or missing, run 'crc setup' to download the bundle", bundlePath)
```

cfergeau: I'd say "to download the default bundle"

praveenkumar: @cfergeau updated

cfergeau (approved):

PR #3839, opened 2023-09-21
preflight: accept lower case drive letter in daemon posh script check

anjannath:
in the preflight check for existence of the powershell script used for starting the daemon, we check that the current crc executable path  is present

when crc is executed by the extension, the current executable path   is
returned with a small case drive letter whereas crc directly invoked by
a user returns a path with upper case drive letter and fails this check

with this commit we are not matching the whole content of the file but
only checking if either lower case or upper case varient of the   path
exists in the script

### DIFF
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -211,8 +211,10 @@ func getDaemonPoshScriptContent() []byte {
 	if err != nil {
 		return []byte{}
 	}
+	drive := filepath.VolumeName(binPath)
+	binPathWithSmallCaseDriveLetter := strings.Replace(binPath, drive, strings.ToLower(drive), 1)
 	daemonCmdArgs := `daemon --log-level debug`
-	return []byte(fmt.Sprintf(daemonPoshScriptTemplate, binPath, daemonCmdArgs))
+	return []byte(fmt.Sprintf(daemonPoshScriptTemplate, binPathWithSmallCaseDriveLetter, daemonCmdArgs))
 }
 
 func checkDaemonPoshScript() error {
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -217,7 +217,22 @@ func getDaemonPoshScriptContent() []byte {
...
+		}
+
+		binPath, err := os.Executable()
+		if err != nil {
+			return fmt.Errorf("Unable to find current crc executable path: %w", err)
+		}
+
+		drive := filepath.VolumeName(binPath)
+		binPathWithSmallCaseDriveLetter := strings.Replace(binPath, drive, strings.ToLower(drive), 1)
+		binPathWithUpperCaseDriveLetter := strings.Replace(binPath, drive, strings.ToUpper(drive), 1)
```
cfergeau: Windows paths are case insensitive, I don't think we want to add more variations if there are other concrete situations where the case is not what we expect? Do we get the correct behaviour with `Crc setup` followed by `crc setup` or `crc.EXE setup` followed by `crc.exe setup`?

 Maybe we can force `binPath` to lower case in `getDaemonPoshScriptContent` to have a single canonical form for the binary path, and to make it independant from how the user start crc?

anjannath: > Do we get the correct behaviour with Crc setup followed by crc setup or crc.EXE setup followed by crc.exe setup?

these variants are producing correct behaviour, it doesn't effect the path, always normalized to `crc.exe`

> Maybe we can force binPath to lower case in getDaemonPoshScriptContent to have a single canonical form for the binary path, and to make it independant from how the user start crc?

yes, this will also work, and with much smaller changes, basically we have to match what is the constant here: https://github.com/crc-org/crc-extension/blob/571f6dbbb0cc95c4214c6ea35e32cb9af112045f/src/crc-cli.ts#L26

cfergeau: is this only the drive letter which is problematic then? or can we also have issues with the dir name? `C:\program files\` vs `C:\Program Files`? I'm not clear on which go API is involved in the bug/what gets normalized/not normalized with go apis :)

A slightly hacky (or not ?) alternative would be to use `daemonPoshScriptTemplate` to create a regexp which we match against the installed template. With capturing regexp, we can even get the binary path used in the installed template to be able to run additional checks.

anjannath: > is this only the drive letter which is problematic then? or can we also have issues with the dir name? C:\program files\ vs C:\Program Files? I'm not clear on which go API is involved in the bug/what gets normalized/not normalized with go apis :)

yeah, only the drive letter is problematic, the go API is `os.Executable()` it is returning the path with smaller case drive letter when crc is invoked from the extension, this is because the invoked process's `PATH` env is modified to contain the path with _smaller_ case drive letter in the first entry of `PATH`

the dir names if different will also be an issue, but currently only small drive letter is causing the issue

> A slightly hacky (or not ?) alternative would be to use daemonPoshScriptTemplate to create a regexp which we match against the installed template. With capturing regexp, we can even get the binary path used in the installed template to be able to run additional checks.

right, this will also be future proof, (what i am aiming is to make it behave the same irrespective of whether it is returning a lower case or upper case drive letter)

pkg/crc/preflight/preflight_daemon_task_check_windows.go
```diff
@@ -211,8 +211,10 @@ func getDaemonPoshScriptContent() []byte {
 	if err != nil {
 		return []byte{}
 	}
+	drive := filepath.VolumeName(binPath)
+	binPathWithSmallCaseDriveLetter := strings.Replace(binPath, drive, strings.ToLower(drive), 1)
 	daemonCmdArgs := `daemon --log-level debug`
-	return []byte(fmt.Sprintf(daemonPoshScriptTemplate, binPath, daemonCmdArgs))
+	return []byte(fmt.Sprintf(daemonPoshScriptTemplate, binPathWithSmallCaseDriveLetter, daemonCmdArgs))
```
praveenkumar, line 217: I don't think we should assume other program always going to use the small case driver letter so, I think it is other program responsibility to first check the crc executable because after crc is installed we add that path so other program should able to detect the binary.

anjannath:
closing for now, will reopen again if needed! issue is fixed on the extension side.

pkg/crc/preflight/preflight_daemon_task_check_windows.go
cfergeau: > the dir names if different will also be an issue, but currently only small drive letter is causing the issue

I was indeed asking a general question, I understand that just takin care of `C:` VS `c:` will fix the extension bug. Though it seems command line users can easily hit this too.

PR #3841, opened 2023-09-21
e2e: Fix syntax for commands to check cluster certs

adrianriobo:
With https://github.com/crc-org/crc/commit/9bb4cac2bb8e65ee84e9019921db978742ad321f several steps were introduced to ensure the cert chain from the cluster, but the syntax on those steps fails on some of platforms / versions.

This PR fix the syntax making it supportable on all platforms.

### DIFF
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -512,6 +512,8 @@ func InitializeScenario(s *godog.ScenarioContext) {
 		ConfigFileInCRCHomeContainsKey)
 	s.Step(`removing file "(.*)" from CRC home folder succeeds$`,
 		DeleteFileFromCRCHome)
+	s.Step(`^decode base64 file "(.*)" to "(.*)"$`,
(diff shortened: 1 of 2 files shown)


anjannath:
`e2e-crc` test was killed before finishing, maybe we need to increase the timeout again!
```
*** Test killed with quit: ran too long (3h1m0s).
```

adrianriobo:
no it was my bad I removed an `echo |` but did not tried correctly, as so `openssl` cmd hangs on..this is why the tiemout...I will force push the commit today, thx for checking the timeout though

test/e2e/testsuite/testsuite.go
```diff
@@ -978,3 +980,14 @@ func PullLoginTagPushImageSucceeds(image string) error {
 
 	return nil
 }
+
+// Decode a file enconed with base64
```
praveenkumar: `enconed` => `encoded`

jsliacan (approved):
LGTM

PR #3843, opened 2023-09-26
update to go 1.20

anjannath:


cfergeau:
Failure is 
```
GOARCH=amd64 GOOS=windows go build -tags "containers_image_openpgp" -ldflags="-X github.com/crc-org/crc/v2/pkg/crc/version.crcVersion=2.26.0 -X github.com/crc-org/crc/v2/pkg/crc/version.ocpVersion=4.13.9 -X github.com/crc-org/crc/v2/pkg/crc/version.okdVersion=4.13.0-0.okd-2023-06-04-080300 -X github.com/crc-org/crc/v2/pkg/crc/version.podmanVersion=4.4.4 -X github.com/crc-org/crc/v2/pkg/crc/version.microshiftVersion=4.13.9 -X github.com/crc-org/crc/v2/pkg/crc/version.commitSha=46e313 " -o out/windows-amd64/crc.exe  ./cmd/crc
# github.com/crc-org/crc/v2/cmd/crc
C:\hostedtoolcache\windows\go\1.19.13\x64\pkg\tool\windows_amd64\link.exe: running gcc failed: exit status 1
C:/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-867149281\000007.o: in function `_cgo_preinit_init':
\\_\_\runtime\cgo/gcc_libinit_windows.c:40: undefined reference to `__imp___iob_func'
C:/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-867149281\000007.o: in function `x_cgo_notify_runtime_init_done':
\\_\_\runtime\cgo/gcc_libinit_windows.c:105: undefined reference to `__imp___iob_func'
C:/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-867149281\000007.o: in function `_cgo_beginthread':
\\_\_\runtime\cgo/gcc_libinit_windows.c:149: undefined reference to `__imp___iob_func'
C:/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\RUNNER~1\AppData\Local\Temp\go-link-867149281\000008.o: in function `x_cgo_thread_start':
\\_\_\runtime\cgo/gcc_util.c:18: undefined reference to `__imp___iob_func'
collect2.exe: error: ld returned 1 exit status

mingw32-make: *** [Makefile:105: out/windows-amd64/crc.exe] Error 1
Error: Process completed with exit code 1.
```

cfergeau:
If I understand the commit log correctly, mingw is installed by github on the Windows images they offer? We are also using the official actions/setup-go action, so it's very unexpected there is such a breakage. 
One thing I noticed is that we use `actions/setup-go@v3` which hasn't been updated in a year https://github.com/actions/setup-go/tags
Did you try using `actions/setup-go@v4`?

cfergeau:
Does not help, but I'd also drop the explicit `go: 1.19` requirements, at least for testing.

cfergeau:
The latest push passed gh actions checks.

anjannath:
yeah, if explicit go version is not mentioned, then it works, it seems `actions/setup-go` just uses the go version already installed (1.20) in the runner (the runners comes preinstalled with a certain version: https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#language-and-runtime)

`actions/setup-go` is not really taking care of the C compiler, it just installs the golang package, I think we have to downgrade `mingw` like before if we want to keep a pinned version of golang, which is what we want, right?

cfergeau:
> yeah, if explicit go version is not mentioned, then it works, it seems `actions/setup-go` just uses the go version already installed (1.20) in the runner

My guess is that 1.20 and 1.21 would work if explicitly mentioned, and `stable` / `oldstable` would also work if they map to 1.21 and 1.20

> I think we have to downgrade mingw like before if we want to keep a pinned version of golang, which is what we want, right?

In general, yes, we want to make sure crc compiles with the go version we'll be using to build our releases. Regarding the present issue :
- golang 1.19 is no longer supported upstream since 1.21 was released this summer
- RHEL (which we use for our release builds) has golang 1.20, at least in RHEL 8.8

This seems like a good time to switch to a newer go version :)

anjannath:
@cfergeau can we merge the first two commits because its needed for the `choco-release` github action to succeed and push the choco to the community feed when we publish the `2.27.0` release, I'll create another PR to update the go version to 1.20, wdyt?

will update this PR to do:
1. Downgrade mingw to `11.2.0`
2. And update to `actions/setup-go@v4`

pkg/crc/cluster/podman.go
```diff
@@ -7,12 +7,12 @@ import (
 )
 
 func GenerateCockpitBearerToken() string {
-	rand.Seed(time.Now().UnixNano())
+	rng := rand.NewSource(time.Now().UnixNano())
```
cfergeau, line 10: Is this change needed?

anjannath: not strictly needed for us, but `golangci-lint` static check started complaining after updating to `go 1.20`, `rand.Seed` has been deprecated 

```sh
pkg/crc/cluster/podman.go:10:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
        rand.Seed(time.Now().UnixNano())
        ^
```

praveenkumar: may be put is separate commit, like first this change and then update to go-1.20 commit?

.github/workflows/macos-installer.yml
```diff
@@ -19,8 +19,6 @@ jobs:
         os:
           - macOS-latest
           - macOS-11
-        go:
-          - 1.20
```
praveenkumar: why we want to remove explict version because this make sure we have same on the CI which we use downstream?

anjannath: still testing things, for some reason the previous build failed, will remove this, plan is to keep pinning to 1.20

anjannath: removed now, the version needs to be in quotes like: `'1.20'` otherwise it thinks its version `1.2`, they have a note about it.. from their readme:

>Note: Due to the peculiarities of YAML parsing, it is recommended to wrap the version in single quotation marks:
```
  go-version: '1.20'
```
>The recommendation is based on the YAML parser's behavior, which interprets non-wrapped values as numbers and, in the case of version 1.20, trims it down to 1.2, which may not be very obvious. Matching an unstable pre-release:

anjannath:
Need to update the go version on openshift ci as well

cfergeau:
> Need to update the go version on openshift ci as well

If you feel it's taking too much time, let's just go with the mingw downgrade to fix CI, then you can work at your own pace on the 1.20 update.

praveenkumar:
> Need to update the go version on openshift ci as well

created https://github.com/openshift/release/pull/43837

anjannath:
> > Need to update the go version on openshift ci as well
> 
> If you feel it's taking too much time, let's just go with the mingw downgrade to fix CI, then you can work at your own pace on the 1.20 update.

No its fine, i believe once the openshift CI PR is merged, we'll have green CI here as well, and since the choco release job will anyway have to be triggered from the release commit, downgrading will not fix the choco release job for publishing the choco package purpose

pkg/crc/machine/bundle/metadata_test.go
```diff
@@ -209,5 +209,9 @@ func TestVerifiedHash(t *testing.T) {
 func testDataURI(t *testing.T, sha256sum string) string {
 	absPath, err := filepath.Abs(filepath.Join("testdata", sha256sum))
 	require.NoError(t, err)
+	// when its a windows path with drive letter
+	if volume := filepath.VolumeName(absPath); volume != "" {
+		return fmt.Sprintf("file://%s/%s", volume[:1], absPath[3:])
+	}
```
cfergeau: `file:///D:/a/crc/crc/pkg/crc/machine/bundle..` is apparently acceptable, see https://github.com/golang/go/issues/6027 , so maybe calling https://pkg.go.dev/path/filepath#ToSlash would be enough to get a functional URI?

cfergeau:
Looks good to me!

pkg/crc/machine/bundle/metadata_test.go
anjannath: It didn't work, even after passing it to `filepath.ToSlash` the resulting URI is `D:\D:\a\crc\crc\pkg\crc\machine\bundle\testdata\sha256sum_correct_4.13.0.txt.sig`

`filepath.Abs` already returns `D:\a\crc\crc\pkg\crc\machine\bundle\testdata\sha256sum_correct_4.13.0.txt.sig` and when we use the URI `file:///` the additional `/` is again converted to the root volume so we have double `D:\D:\`

job run: https://github.com/crc-org/crc/actions/runs/6414338773/job/17414568644?pr=3843

cfergeau: This passes on my test Windows VM with:
```
diff --git a/pkg/crc/machine/bundle/metadata_test.go b/pkg/crc/machine/bundle/metadata_test.go
index cff984a98..61134f9b0 100644
--- a/pkg/crc/machine/bundle/metadata_test.go
+++ b/pkg/crc/machine/bundle/metadata_test.go
@@ -209,5 +209,5 @@ func TestVerifiedHash(t *testing.T) {
 func testDataURI(t *testing.T, sha256sum string) string {
        absPath, err := filepath.Abs(filepath.Join("testdata", sha256sum))
        require.NoError(t, err)
-       return fmt.Sprintf("file:///%s", absPath)
+       return fmt.Sprintf("file://%s", filepath.ToSlash(absPath))
 }
```
(`file:///` -> `file://` + use of `filepath.ToSlash()`).
On linux, the test also passes with these changes.

anjannath: ahh, i think i might have used `file:///` instead of `file://`

cfergeau (approved):
ACK if CI "passes" for the tests impacted by these changes.

PR #3850, opened 2023-09-27
Add file protocol support for download.Inmemory function 

praveenkumar:
This is helpful for consuming it for unit test where we can cache the
file locally instead of getting from internet in offline situation.
We make use of it in next commit.


pkg/crc/machine/bundle/metadata_test.go
```diff
@@ -188,24 +189,31 @@ func TestGetBundleType(t *testing.T) {
...
 
 	// sha256sum.txt.sig does not contain a sha256sum for fake.crcbundle
-	_, err = getVerifiedHash("https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/bundles/openshift/4.13.0/sha256sum.txt.sig", "fake.crcbundle")
+	absPath, err = filepath.Abs(filepath.Join("testdata", "sha256sum_correct_4.13.0.txt.sig"))
+	require.NoError(t, err)
+	_, err = getVerifiedHash(fmt.Sprintf("file:///%s", absPath), "fake.crcbundle")
 	require.Error(t, err)
 
 	// 4.10.1 bundles did not come with a sha256sum.txt.sig file
 	_, err = getVerifiedHash("https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/bundles/openshift/4.10.1/sha256sum.txt.sig", "crc_libvirt_4.10.1_amd64.crcbundle")
```
praveenkumar: I am not sure what to do there because this is anyway return `404` since file is not found and we are just saying if there is err (no matter what type) then we are good.

pkg/crc/machine/bundle/metadata_test.go
```diff
@@ -188,24 +189,31 @@ func TestGetBundleType(t *testing.T) {
...
+	require.NoError(t, err)
+	sha256sum, err := getVerifiedHash(fmt.Sprintf("file:///%s", absPath), "crc_libvirt_4.13.0_amd64.crcbundle")
 	require.NoError(t, err)
 	require.Equal(t, "6aad57019aaab95b670378f569b3f4a16398da0358dd1057996453a8d6d92212", sha256sum)
 
 	// sha256sum.txt is unsigned
-	_, err = getVerifiedHash("https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/crc/bundles/openshift/4.13.0/sha256sum.txt", "crc_libvirt_4.13.0_amd64.crcbundle")
+	absPath, err = filepath.Abs(filepath.Join("testdata", "sha256sum.txt"))
+	require.NoError(t, err)
+	_, err = getVerifiedHash(fmt.Sprintf("file:///%s", absPath), "crc_libvirt_4.13.0_amd64.crcbundle")
```
cfergeau: This deserves a small `testDataURI("sha256sum.txt")` helper for readability

pkg/crc/machine/bundle/metadata_test.go
cfergeau: The goal of this test was to make sure we don't return false positives when the file is missing (ie don't return a hash by mistake). If you replace this with a file, it's going to be quite different than testing with http.
You can keep it anyway with a file, or remove it, no strong feeling.

pkg/download/download.go
```diff
@@ -90,12 +90,23 @@ func Download(uri, destination string, mode os.FileMode, sha256sum []byte) (stri
 
 // Download takes an URL and the User-Agent string to use in the http request
 // it takes  http.RoundTripper as the third argument to be used by the client
```
cfergeau: Can you update the comment while at it? I forgot to update it when I moved code around.

cfergeau (commented):
Imo an API similar to https://pkg.go.dev/os#ReadFile would fit better our usecase than using an intermediate `io.ReadCloser`. But fine with me this way.

pkg/download/download.go
praveenkumar: Done

pkg/crc/machine/bundle/metadata_test.go
praveenkumar: Done

pkg/crc/machine/bundle/metadata_test.go
praveenkumar: Removed this one.

cfergeau (approved):

PR #3857, opened 2023-10-03
[e2e] story_microshift ensures operational status for cluster 

adrianriobo:
Fix #3851 

this commit may will be partially reverted on a following PR in case #3852 will be implemented

PR #3858, opened 2023-10-04
start: Use `RunPrivate` for checking pull secret for microshift

praveenkumar:
As part of 211e9dd7, I used `RunPrivileged` which end up exposing pull secret to debug logs. With this PR we are going to use `RunPrivate` to make sure we hide the pull secret info.

**Fixes:** Issue #3857

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -1068,7 +1068,7 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 }
 
 func ensurePullSecretPresentInVM(sshRunner *crcssh.Runner, pullSec cluster.PullSecretLoader) error {
-	if pullSecret, _, err := sshRunner.RunPrivileged("Checking if pull secret already present in the VM", "cat", "/etc/crio/openshift-pull-secret"); err == nil {
+	if pullSecret, _, err := sshRunner.RunPrivate("sudo", "cat", "/etc/crio/openshift-pull-secret"); err == nil {
 		if err := validation.ImagePullSecret(pullSecret); err == nil {
 			return nil
 		}


pkg/crc/machine/start.go
```diff
@@ -1068,7 +1068,7 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 }
 
 func ensurePullSecretPresentInVM(sshRunner *crcssh.Runner, pullSec cluster.PullSecretLoader) error {
-	if pullSecret, _, err := sshRunner.RunPrivileged("Checking if pull secret already present in the VM", "cat", "/etc/crio/openshift-pull-secret"); err == nil {
+	if pullSecret, _, err := sshRunner.RunPrivate("Checking if pull secret already present in the VM", "cat", "/etc/crio/openshift-pull-secret"); err == nil {
```
cfergeau: This file can be read by a regular user, no need for sudo?

cfergeau (commented):
Looks good to me if the file is user-readable (do we want that?)

cfergeau (commented):
The commit log could explain why we need `sudo`, especially after the 1st iteration and the review comments.
`openshift-pull-secret` is created with `return sshRunner.CopyDataPrivileged([]byte(content), "/etc/crio/openshift-pull-secret", 0600)`

pkg/crc/machine/start.go
praveenkumar: Ah it does need sudo, updated.

praveenkumar:
> The commit log could explain why we need `sudo`, especially after the 1st iteration and the review comments. `openshift-pull-secret` is created with `return sshRunner.CopyDataPrivileged([]byte(content), "/etc/crio/openshift-pull-secret", 0600)`

@cfergeau Done

cfergeau (approved):

Issue #3840, opened 2023-09-21
crc setup breaks dns configuration on host - Linux (RHEL9)


mschindl:
The issue still exists and if it will be installed on a small server then it breaks the dns configuration...

After crc setup my resolv.conf looks like following >>>

$ cat /etc/resolv.conf 
```
# Generated by NetworkManager
search fritz.box local
nameserver 127.0.0.1
options edns0 trust-ad
```

$  cat /etc/os-release
```
NAME="Red Hat Enterprise Linux"
VERSION="9.2 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.2"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.2 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.2
```

Changing dns server for example with "nmtui" doesn't work.

Any ideas on how best to fix this?

_Originally posted by @mschindl in https://github.com/crc-org/crc/issues/1663#issuecomment-1729447033_

cfergeau:
I don't know what does not work/what you are trying to achieve/... so it is hard to answer
resolv.conf points at 127.0.0.1 because NetworkManager is configured to use dnsmasq, and spawned a dnsmasq instance listening on 127.0.0.1. See /etc/NetworkManager if you want to take a closer look at how it's configured.

mschindl:
I noticed this, now I'm trying to reverse it or change it so that DNS queries work again. Until then crc setup doesn't work either, because it can't download packages.

```
INFO Downloading bundle: /home/user/.crc/cache/crc_libvirt_4.13.9_amd64.crcbundle... 
[...]

PR #3862, opened 2023-10-09
Kubeconfig: Use user provided ingress https port for token dialer

praveenkumar:
Current implementation have issue if user provided non default https port because during token creation request goes from api server to auth service which uses default https port and resulted following error for user mode networking (there is no support for customize http/https route for system mode networking)
```
ERRO Cannot update kubeconfig: Head "https://oauth-openshift.apps-crc.testing": dial tcp 127.0.0.1:443: connect: connection refused
```

With this PR we are passing that https port info to `addContext` function and use it for dialer where we update the port in case request have `oauth-openshift` as part of address which resolve this issue.
```
$ ./crc config view
- enable-bundle-quay-fallback           : true
- ingress-http-port                     : 9080
- ingress-https-port                    : 9443

$ ./crc start --log-level debug
[...]
level=info msg="Adding crc-admin and crc-developer contexts to kubeconfig..."
level=debug msg="Using address: api.crc.testing:6443"
level=debug msg="Dialing to 127.0.0.1:6443"
level=debug msg="Using address: oauth-openshift.apps-crc.testing:443"
level=debug msg="Dialing to 127.0.0.1:9443"
level=debug msg="Using address: oauth-openshift.apps-crc.testing:443"
level=debug msg="Dialing to 127.0.0.1:9443"
level=debug msg="Using address: api.crc.testing:6443"
level=debug msg="Dialing to 127.0.0.1:6443"
level=debug msg="Using address: oauth-openshift.apps-crc.testing:443"
level=debug msg="Dialing to 127.0.0.1:9443"
level=debug msg="Using address: oauth-openshift.apps-crc.testing:443"
level=debug msg="Dialing to 127.0.0.1:9443"
[...]
```

**Fixes:** Issue #3840

### DIFF
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -10,10 +10,12 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
+	"github.com/crc-org/crc/v2/pkg/crc/logging"
 	"github.com/crc-org/crc/v2/pkg/crc/machine/types"
 	"github.com/openshift/oc/pkg/helpers/tokencmd"
 	"k8s.io/apimachinery/third_party/forked/golang/netutil"
@@ -46,7 +48,7 @@ func updateClientCrtAndKeyToKubeconfig(clientKey, clientCrt []byte, srcKubeconfi
 	return clientcmd.WriteToFile(*cfg, destKubeconfigPath)
 }
 
-func writeKubeconfig(ip string, clusterConfig *types.ClusterConfig) error {
+func writeKubeconfig(ip string, clusterConfig *types.ClusterConfig, ingressHTTPSPort uint) error {
 	kubeconfig, cfg, err := getGlobalKubeConfig()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/kubeconfig.go
```diff
@@ -145,11 +147,16 @@ func addContext(cfg *api.Config, ip string, clusterConfig *types.ClusterConfig,
 				MinVersion: tls.VersionTLS12,
 			},
 			DialContext: func(ctx gocontext.Context, network, address string) (net.Conn, error) {
+				logging.Debugf("Using address: %s", address)
 				port := strings.SplitN(address, ":", 2)[1]
+				if strings.Contains(address, "oauth-openshift") {
```
cfergeau: I'd use something like:
```
hostname, port := net.SplitHostPort(address)
if strings.HasSuffix(hostname, constants.AppsDomain) {
    port = strconv.Itoa(int(ingressHTTPSPort))
}
```

This is preexisting, but not clear why we always dial to `ip` when we have an explicit `address`. Just thinking out loud.

praveenkumar: >  This is preexisting, but not clear why we always dial to ip when we have an explicit address. Just thinking out loud.

We are using IP just to avoid dns resolution in case it is not working or in proxied env where it is not excluded.

cfergeau (approved):

PR #3867, opened 2023-10-11
Fixes #3866 check if the script exists before trying to remove

### LINKED ISSUE #3866: [BUG] crc cleanup execution failure in case hidden_daemon.ps1 is not present
With recent changes a new script to ensure the daemon execution is hidden was added, that file is added during `crc setup` operation in case the operation fails for whatever reason before adding the file when after it `crc cleanup` is executed it ends up failing due to:

```
INFO Removing older logs
remove C:\Users\crcqe\.crc\bin\hidden_daemon.ps1: The system cannot find the file specified.
```

anjannath:
this prevents the following error on windows:
````
INFO Removing older logs
remove C:\Users\crcqe\.crc\bin\hidden_daemon.ps1: The system cannot find the file specified.
```

### DIFF
--- a/pkg/crc/preflight/preflight_daemon_task_check_windows.go
+++ b/pkg/crc/preflight/preflight_daemon_task_check_windows.go
@@ -236,3 +236,10 @@ func killDaemonProcessIfRunning() error {
 	}
 	return nil
 }
+
+func removeDaemonPoshScript() error {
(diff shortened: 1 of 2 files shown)


adrianriobo:
Fixes #3866

praveenkumar (approved):

PR #3890, opened 2023-10-25
Use 'Apple silicon' instead of 'M1'

### LINKED ISSUE #3889: [BUG] Mac M2 arch reporting as M1
AWS offers support for mac M2, when trying `crc` on an M2 the INFO message is identifying the CPU as a M1:

```
INFO Checking if running emulated on a M1 CPU
```

Checking the arch we can see the M2

```
ec2-user@Mac-mini ~ % sysctl machdep.cpu.brand_string
machdep.cpu.brand_string: Apple M2 Pro
```

cfergeau:
There are now M2 CPUs, and more generations to come. "Apple silicon" is
more generic.

This fixes https://github.com/crc-org/crc/issues/3889

### DIFF
--- a/pkg/crc/preflight/preflight_darwin.go
+++ b/pkg/crc/preflight/preflight_darwin.go
@@ -14,9 +14,9 @@ import (
 var vfkitPreflightChecks = []Check{
 	{
 		configKeySuffix:  "check-m1-cpu",
-		checkDescription: "Checking if running emulated on a M1 CPU",
-		check:            checkM1CPU,
-		fixDescription:   "This version of CRC for AMD64/Intel64 CPUs is unsupported on Apple M1 hardware",
+		checkDescription: "Checking if running emulated on Apple silicon",
+		check:            checkAppleSilicon,
+		fixDescription:   "This version of CRC for AMD64/Intel64 CPUs is unsupported on Apple silicon",
 		flags:            NoFix,
 
 		labels: labels{Os: Darwin},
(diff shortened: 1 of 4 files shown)


pkg/crc/preflight/preflight_darwin.go
```diff
@@ -14,9 +14,9 @@ import (
 var vfkitPreflightChecks = []Check{
 	{
 		configKeySuffix:  "check-m1-cpu",
-		checkDescription: "Checking if running emulated on a M1 CPU",
+		checkDescription: "Checking if running emulated on Apple silicon",
 		check:            checkM1CPU,
```
praveenkumar: can we change this check also to `checkMseriesCPU` or `checkSiliconCPU` ?

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -14,9 +14,9 @@ import (
 var vfkitPreflightChecks = []Check{
 	{
 		configKeySuffix:  "check-m1-cpu",
-		checkDescription: "Checking if running emulated on a M1 CPU",
-		check:            checkM1CPU,
-		fixDescription:   "This version of CRC for AMD64/Intel64 CPUs is unsupported on Apple M1 hardware",
+		checkDescription: "Checking if running emulated on Apple silicon",
+		check:            checkAppleSilicon,
+		fixDescription:   "This version of CRC for AMD64/Intel64 CPUs is unsupported on Apple silicon",
```
praveenkumar, line 19: 👍🏼

PR #3891, opened 2023-10-25
actions: add windows e2e

jsliacan:
**Fixes:** Issue #3961

### DIFF
--- a/.github/workflows/windows-e2e.yml
+++ b/.github/workflows/windows-e2e.yml
@@ -1,0 +1,232 @@
+name: Windows e2e
+
+on:
+  workflow_run:
+    workflows: [Build Windows installer]
+    types:
+      - completed
+  
+jobs:
+  windows:
+    runs-on: ubuntu-latest
+    permissions:
+      statuses: write # needed to update commit status (pending/failure/sucess)
+      checks: write # as documented in https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#pr-run-permissions
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


praveenkumar:
Is this a [WIP] or ready for review PR?

jsliacan:
@praveenkumar it's just the bot casually removing the `do-not-merge` label. Sorry, didn't notice.

gbraad:
@adrianriobo Why did you add the hold label; perhaps it was WIP instead?

anjannath:
The `Windows e2e` action is not getting triggered, from the action yaml, it should get triggered once the `Build Windows installer` is finished running

jsliacan:
@anjannath thanks. That's because it needs secrets etc. and therefore workflow_run triggered flows only work if they're on the default branch (main in our upstream case). So in other words, it'll only work once merged. However, I set up an analogous situation with windows-installer and build-images workflows on my fork (where I have access to secrets and can designate any branch as default) and it works: https://github.com/jsliacan/crc/actions/runs/7098881034. But I need to polish it a little before I remove the hold label.

anjannath (approved):
LGTM

successful run on jsliacans fork https://github.com/jsliacan/crc/actions/runs/7206156919

jsliacan:
It seems that I still need to display test results among the upstream PR checks (they run on the fork, but the workflow with windows-e2e runs on the default branch of the upstream repository). So one needs to take the results, when ready, and post them back on the PR that triggered the workflow which then triggered the windows-e2e workflow.

PR #3903, opened 2023-11-01
Add status change event

evidolob:
## Solution/Idea
PR adds event API and new SSE(server sent event) channel `status_change`.

## Proposed changes
PR adds event API implementation in `pkg/crc/api/events/events.go`
Modify `sync` client implementation to trigger event when status is changed.
Fixes in existing log events, rename constants related to SSE

## Testing

1. launch demon
2. connect to `status_change` channel with `curl -GET -i --unix-socket <path to socket>/crc-http.sock "http://localhost/events?stream=status_change"`
3.  start crc with `curl -GET -i --unix-socket ~/.crc/crc-http.sock  "http://localhost/api/start"`
4. Ensure that new events appeared, like:
```
id:                                                                                                                            │
data: {"status":{"CrcStatus":"Starting","OpenshiftStatus":"Starting","OpenshiftVersion":"","PodmanVersion":"","DiskUse":0,"DiskSize":0,"RAMUse":0,"RAMSize":0,"Preset":"microshift"}}
event: status_change
```

### DIFF
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -13,6 +13,7 @@ import (
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
 	"github.com/crc-org/crc/v2/pkg/crc/daemonclient"
 	crcErrors "github.com/crc-org/crc/v2/pkg/crc/errors"
+	"github.com/crc-org/crc/v2/pkg/crc/machine/state"
 	"github.com/crc-org/crc/v2/pkg/crc/machine/types"
 	"github.com/crc-org/crc/v2/pkg/crc/preset"
 	"github.com/docker/go-units"
@@ -151,6 +152,9 @@ func getStatus(client *daemonclient.Client, cacheDir string) *status {
 			return &status{Success: false, Error: crcErrors.ToSerializableError(crcErrors.DaemonNotRunning)}
 		}
 		return &status{Success: false, Error: crcErrors.ToSerializableError(err)}
+	}
+	if clusterStatus.CrcStatus == string(state.NoVM) {
+		return &status{Success: false, Error: crcErrors.ToSerializableError(crcErrors.VMNotExist)}
 	}
 	var size int64
 	err = filepath.Walk(cacheDir, func(_ string, info os.FileInfo, err error) error {
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -88,7 +88,7 @@ func runWatchStatus(writer io.Writer, client *daemonclient.Client, cacheDir stri
 		}
 	}()
 
-	err = client.SSEClient.Status(func(loadResult *types.ClusterLoadResult) {
+	err = client.SSEClient.ClusterLoad(func(loadResult *types.ClusterLoadResult) {
 		if !isPoolInit {
 			ramBar, cpuBars = createBars(loadResult.CPUUse, writer)
 			barPool = pb.NewPool(append([]*pb.ProgressBar{ramBar}, cpuBars...)...)
--- a/pkg/crc/api/events/cluster_load_stream.go
+++ b/pkg/crc/api/events/cluster_load_stream.go
@@ -19,18 +19,18 @@ type TickListener struct {
 	tickPeriod time.Duration
 }
 
-func newStatusStream(server *EventServer) EventStream {
-	return newStream(NewStatusListener(server.machine), newEventPublisher(STATUS, server.sseServer))
+func newClusterLoadStream(server *EventServer) EventStream {
+	return newStream(newClusterLoadListener(server.machine), newEventPublisher(ClusterLoad, server.sseServer))
 }
 
-func NewStatusListener(machine crcMachine.Client) EventProducer {
+func newClusterLoadListener(machine crcMachine.Client) EventProducer {
 	getStatus := func() (interface{}, error) {
 		return machine.GetClusterLoad()
 	}
-	return NewTickListener(getStatus)
+	return newTickListener(getStatus)
 }
 
-func NewTickListener(generator genData) EventProducer {
+func newTickListener(generator genData) EventProducer {
 	return &TickListener{
 		done:       make(chan bool),
 		generator:  generator,
--- a/pkg/crc/api/events/event_server.go
+++ b/pkg/crc/api/events/event_server.go
@@ -56,6 +56,7 @@ func NewEventServer(machine machine.Client) *EventServer {
 
 	sseServer.CreateStream(Logs)
 	sseServer.CreateStream(ClusterLoad)
+	sseServer.CreateStream(StatusChange)
 	return eventServer
 }
 
@@ -69,6 +70,8 @@ func createEventStream(server *EventServer, streamID string) EventStream {
 		return newLogsStream(server)
 	case ClusterLoad:
 		return newClusterLoadStream(server)
+	case StatusChange:
+		return newStatusChangeStream(server)
 	}
 	return nil
 }
--- a/pkg/crc/api/events/event_server.go
+++ b/pkg/crc/api/events/event_server.go
@@ -54,8 +54,8 @@ func NewEventServer(machine machine.Client) *EventServer {
 		stream.RemoveSubscriber(sub)
 	}
 
-	sseServer.CreateStream(LOGS)
-	sseServer.CreateStream(STATUS)
+	sseServer.CreateStream(Logs)
+	sseServer.CreateStream(ClusterLoad)
 	return eventServer
 }
 
@@ -65,10 +65,10 @@ func (es *EventServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 func createEventStream(server *EventServer, streamID string) EventStream {
 	switch streamID {
-	case LOGS:
+	case Logs:
 		return newLogsStream(server)
-	case STATUS:
-		return newStatusStream(server)
+	case ClusterLoad:
+		return newClusterLoadStream(server)
 	}
 	return nil
 }
--- a/pkg/crc/api/events/log_stream.go
+++ b/pkg/crc/api/events/log_stream.go
@@ -1,6 +1,8 @@
 package events
 
 import (
+	"bytes"
+
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
 	"github.com/r3labs/sse/v2"
 	"github.com/sirupsen/logrus"
@@ -56,7 +58,10 @@ func (s *streamHook) Fire(entry *logrus.Entry) error {
 		return err
 	}
 
-	s.server.Publish(LOGS, &sse.Event{Event: []byte(LOGS), Data: line})
+	// remove "Line Feed"("\n") character which add was added by json.Encoder
+	line = bytes.TrimRight(line, "\n")
+
+	s.server.Publish(Logs, &sse.Event{Event: []byte(Logs), Data: line})
 	return nil
 }
 
--- a/pkg/crc/api/events/status_change_stream.go
+++ b/pkg/crc/api/events/status_change_stream.go
@@ -1,0 +1,69 @@
+package events
+
+import (
+	"encoding/json"
+
+	"github.com/crc-org/crc/v2/pkg/crc/logging"
+	"github.com/crc-org/crc/v2/pkg/crc/machine"
+	"github.com/crc-org/crc/v2/pkg/crc/machine/state"
+	"github.com/crc-org/crc/v2/pkg/crc/machine/types"
+	"github.com/crc-org/crc/v2/pkg/events"
+	"github.com/r3labs/sse/v2"
+)
+
+type serializableEvent struct {
+	Status *types.ClusterStatusResult `json:"status"`
+	Error  string                     `json:"error,omitempty"`
+}
+
+type statusChangeListener struct {
+	machineClient machine.Client
+	publisher     EventPublisher
+}
+
+func newStatusChangeStream(server *EventServer) EventStream {
+	return newStream(newStatusChangeListener(server.machine), newEventPublisher(StatusChange, server.sseServer))
+}
+
+func newStatusChangeListener(client machine.Client) EventProducer {
+	return &statusChangeListener{
+		machineClient: client,
+	}
+}
+
+func (st *statusChangeListener) Notify(changedEvent events.StatusChangedEvent) {
+	logging.Debugf("State Changed Event %s", changedEvent)
+	var event serializableEvent
+	status, err := st.machineClient.Status()
+	// if we cannot receive actual state, send error state with error description
+	if err != nil {
+		event = serializableEvent{Status: &types.ClusterStatusResult{
+			CrcStatus: state.Error,
+		}, Error: err.Error()}
+	} else {
+		// event could be fired, before actual code, which change state is called
+		// so status could contain 'old' state, replace it with state received in event
+		status.CrcStatus = changedEvent.State // override with actual reported state
+		event = serializableEvent{Status: status}
+		if changedEvent.Error != nil {
+			event.Error = changedEvent.Error.Error()
+		}
+
+	}
+	data, err := json.Marshal(event)
+	if err != nil {
+		logging.Errorf("Could not serealize status changed event in to JSON: %s", err)
+		return
+	}
+	st.publisher.Publish(&sse.Event{Event: []byte(StatusChange), Data: data})
+}
+
+func (st *statusChangeListener) Start(publisher EventPublisher) {
+	st.publisher = publisher
+	events.StatusChanged.AddListener(st)
+
+}
+
+func (st *statusChangeListener) Stop() {
+	events.StatusChanged.RemoveListener(st)
+}
--- a/pkg/crc/machine/state/state.go
+++ b/pkg/crc/machine/state/state.go
@@ -10,6 +10,7 @@ const (
 	Stopped  State = "Stopped"
 	Stopping State = "Stopping"
 	Starting State = "Starting"
+	NoVM     State = "NoVM"
 	Error    State = "Error"
 )
 
--- a/pkg/crc/machine/status.go
+++ b/pkg/crc/machine/status.go
@@ -18,8 +18,7 @@ func (client *client) Status() (*types.ClusterStatusResult, error) {
 	if err != nil {
 		if errors.Is(err, errMissingHost(client.name)) {
 			return &types.ClusterStatusResult{
-				CrcStatus:       state.Stopped,
-				OpenshiftStatus: types.OpenshiftStopped,
+				CrcStatus: state.NoVM,
 			}, nil
 		}
 		return nil, errors.Wrap(err, fmt.Sprintf("Cannot load '%s' virtual machine", client.name))
--- a/pkg/crc/machine/sync.go
+++ b/pkg/crc/machine/sync.go
@@ -10,6 +10,7 @@ import (
 	"github.com/crc-org/crc/v2/pkg/crc/machine/state"
 	"github.com/crc-org/crc/v2/pkg/crc/machine/types"
 	crcPreset "github.com/crc-org/crc/v2/pkg/crc/preset"
+	"github.com/crc-org/crc/v2/pkg/events"
 )
 
 const startCancelTimeout = 15 * time.Second
@@ -69,6 +70,10 @@ func (s *Synchronized) Delete() error {
 
 	err := s.underlying.Delete()
 	s.syncOperationDone <- Deleting
+
+	if err == nil {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: state.NoVM})
+	}
 	return err
 }
 
@@ -80,6 +85,7 @@ func (s *Synchronized) prepareStart(startCancel context.CancelFunc) error {
 	}
 	s.startCancel = startCancel
 	s.currentState = Starting
+	events.StatusChanged.Fire(events.StatusChangedEvent{State: state.Starting})
 
 	return nil
 }
@@ -92,6 +98,13 @@ func (s *Synchronized) Start(ctx context.Context, startConfig types.StartConfig)
 
 	startResult, err := s.underlying.Start(ctx, startConfig)
 	s.syncOperationDone <- Starting
+
+	if err == nil {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: startResult.Status})
+	} else {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: state.Error, Error: err})
+	}
+
 	return startResult, err
 }
 
@@ -136,10 +149,16 @@ func (s *Synchronized) Stop() (state.State, error) {
 	if err := s.prepareStopDelete(Stopping); err != nil {
 		return state.Error, err
 	}
+	events.StatusChanged.Fire(events.StatusChangedEvent{State: state.Stopping})
 
 	st, err := s.underlying.Stop()
 	s.syncOperationDone <- Stopping
 
+	if err == nil {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: st})
+	} else {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: state.Error, Error: err})
+	}
 	return st, err
 }
 
@@ -160,7 +179,14 @@ func (s *Synchronized) ConnectionDetails() (*types.ConnectionDetails, error) {
 }
 
 func (s *Synchronized) PowerOff() error {
-	return s.underlying.PowerOff()
+	err := s.underlying.PowerOff()
+	if err != nil {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: state.Stopped})
+	} else {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: state.Error, Error: err})
+	}
+
+	return err
 }
 
 func (s *Synchronized) Status() (*types.ClusterStatusResult, error) {
--- a/pkg/crc/machine/virtualmachine.go
+++ b/pkg/crc/machine/virtualmachine.go
@@ -32,6 +32,14 @@ func (err *MissingHostError) Error() string {
 	return fmt.Sprintf("no such libmachine vm: %s", err.name)
 }
 
+func (err *MissingHostError) Is(target error) bool {
+	var x *MissingHostError
+	if errors.As(target, &x) && x.name == err.name {
+		return true
+	}
+	return false
+}
+
 var errInvalidBundleMetadata = errors.New("Error loading bundle metadata")
 
 func loadVirtualMachine(name string, useVSock bool) (*virtualMachine, error) {
--- a/pkg/events/emitter.go
+++ b/pkg/events/emitter.go
@@ -1,0 +1,48 @@
+package events
+
+import (
+	"sync"
+)
+
+type Event[T any] interface {
+	AddListener(listener Notifiable[T])
+	RemoveListener(listener Notifiable[T])
+	Fire(data T)
+}
+
+type Notifiable[T any] interface {
+	Notify(event T)
+}
+
+type event[T any] struct {
+	listeners  map[Notifiable[T]]Notifiable[T]
+	eventMutex sync.Mutex
+}
+
+func NewEvent[T any]() Event[T] {
+	return &event[T]{
+		listeners: make(map[Notifiable[T]]Notifiable[T]),
+	}
+}
+
+func (e *event[T]) AddListener(listener Notifiable[T]) {
+	e.eventMutex.Lock()
+	defer e.eventMutex.Unlock()
+	e.listeners[listener] = listener
+}
+
+func (e *event[T]) RemoveListener(listener Notifiable[T]) {
+	e.eventMutex.Lock()
+	defer e.eventMutex.Unlock()
+	delete(e.listeners, listener)
+}
+
+func (e *event[T]) Fire(event T) {
+	e.eventMutex.Lock()
+	defer e.eventMutex.Unlock()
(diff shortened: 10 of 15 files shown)


anjannath:
> 1. connect to status_change channel with curl -GET -i --unix-socket <path to socket>/crc-http.sock "http://localhost/events?stream=status_change"
> 2. start crc with crc start

the events are shown only when starting via the api `curl --unix-socket ~/.crc/crc-http.sock "http://localhost/api/start"` running the `crc start` command the events are not shown, is this expected?

pkg/events/emitter.go
```diff
@@ -0,0 +1,54 @@
...
+		e.eventMutex.Lock()
+		defer e.eventMutex.Unlock()
+
+		delete(e.listeners, key)
+	}
+}
+
+func (e *event[T]) Fire(event T) {
+	e.eventMutex.Lock()
+	defer e.eventMutex.Unlock()
```
anjannath, line 42: is the locking here needed, we are only accessing the map and not adding/removing elements from it

pkg/crc/api/events/log_stream.go
```diff
@@ -56,7 +56,12 @@ func (s *streamHook) Fire(entry *logrus.Entry) error {
 		return err
 	}
 
-	s.server.Publish(LOGS, &sse.Event{Event: []byte(LOGS), Data: line})
+	// remove "Line Feed"("\n") character which add was added by json.Encoder
+	if line[len(line)-1] == 10 {
+		line = line[:len(line)-1]
+	}
```
anjannath: you could also use `bytes.TrimSuffix` here, https://pkg.go.dev/bytes#TrimSuffix

evidolob:
> the events are shown only when starting via the api `curl --unix-socket ~/.crc/crc-http.sock "http://localhost/api/start"` running the `crc start` command the events are not shown, is this expected?

Yes, that is expected, as `start`, `stop` is performed on CLI side, not on daemon, at least for now.

anjannath:
Tested locally and functionality is working as expected, LGTM. it'd be nice to have more detailed commit messages explaining the changes in each commit

cmd/crc/cmd/status.go
```diff
@@ -81,7 +81,7 @@ func runWatchStatus(writer io.Writer, client *daemonclient.Client, cacheDir stri
 
 	isPullInit := false
 
-	err := client.SSEClient.Status(func(loadResult *types.ClusterLoadResult) {
+	err := client.SSEClient.ClusterLoad(func(loadResult *types.ClusterLoadResult) {
```
cfergeau: Not obvious why the renaming is needed? The commit log should explain this so that we all have the same understanding of the need for it.

pkg/events/emitter.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"sync"
+)
+
+type Event[T any] interface {
+	AddListener(listener Listener[T]) Disposable
+	Fire(data T)
+}
+
+type Disposable func()
```
cfergeau: Dunno if this should use `io.Closer()` https://pkg.go.dev/io#Closer ?

pkg/events/emitter.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func NewEvent[T any]() Event[T] {
+	return &event[T]{
+		listeners: make(map[int]*listenerWrapper[T]),
+	}
+}
+
+func (e *event[T]) AddListener(listener Listener[T]) Disposable {
+	e.eventMutex.Lock()
```
cfergeau, line 29: `defer e.eventMutex.Unlock()` and remove the explicit one before the `return`?

pkg/events/emitter.go
```diff
@@ -0,0 +1,54 @@
...
+
+		delete(e.listeners, key)
+	}
+}
+
+func (e *event[T]) Fire(event T) {
+	e.eventMutex.Lock()
+	defer e.eventMutex.Unlock()
+	for _, l := range e.listeners {
+		l.Listener(event)
```
cfergeau: `listener.Fire(event)` would read better imo.

pkg/events/emitter.go
```diff
@@ -0,0 +1,54 @@
+package events
+
+import (
+	"sync"
+)
+
+type Event[T any] interface {
```
cfergeau, line 7: might be worth a small unit test? And a short explanation in the commit log why this is introduced would be useful to have.

pkg/crc/machine/status.go
```diff
@@ -14,6 +14,16 @@ import (
...
+	exists, err := client.Exists()
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return &types.ClusterStatusResult{
+			CrcStatus: state.Novm,
+		}, nil
+	}
+
```
cfergeau: Isn't `no VM` already covered by `if errors.Is(err, errMissingHost(client.name)) {` below? I think you can change it to return `state.NoVM` instead of an error?

pkg/crc/machine/state/state.go
```diff
@@ -10,6 +10,7 @@ const (
 	Stopped  State = "Stopped"
 	Stopping State = "Stopping"
 	Starting State = "Starting"
+	Novm     State = "NoVM"
```
cfergeau: I'd spell it `NoVM`

pkg/crc/api/events/log_stream.go
```diff
@@ -56,7 +58,12 @@ func (s *streamHook) Fire(entry *logrus.Entry) error {
 		return err
 	}
 
-	s.server.Publish(LOGS, &sse.Event{Event: []byte(LOGS), Data: line})
+	// remove "Line Feed"("\n") character which add was added by json.Encoder
+	if line[len(line)-1] == 10 {
```
cfergeau: I think this could be `== '\n'`. But `TrimRight will do nothing if there is no `\n` at the end of the string, so we don't need to test for it.

pkg/crc/api/events/event_server.go
```diff
@@ -54,8 +54,9 @@ func NewEventServer(machine machine.Client) *EventServer {
 		stream.RemoveSubscriber(sub)
 	}
 
-	sseServer.CreateStream(LOGS)
-	sseServer.CreateStream(CLUSTER_LOAD)
+	sseServer.CreateStream(Logs)
+	sseServer.CreateStream(ClusterLoad)
```
cfergeau, line 58: This belongs in the first commit where you are already doing some renaming.

pkg/crc/api/events/status_change_stream.go
```diff
@@ -0,0 +1,62 @@
+package events
+
+import (
+	"encoding/json"
```
cfergeau, line 4: Can the commit log contain some background information explaining why we are doing this change, what is our goal, ... ?

cfergeau (commented):
Need to spend more time on the last commit, but here are some initial comments.

cmd/crc/cmd/status.go
gbraad: It is not actually a rename, but rather a 'prepare'. The way this will work is to enable the status to be sent when requested (starts listening).

Yevhen can explain, as I asked him the same on one of our meetings.

pkg/crc/api/events/event_server.go
gbraad: @cfergeau should we group renames and refactors? I do not mind if they happen along the way in a PR... but in this case, you prefer this to happen as early as possible?

pkg/events/emitter.go
evidolob: We could, but would add unnecessarily complex to code, as `io.Closer` is an interface not a function as `Disposable`

pkg/crc/api/events/event_server.go
cfergeau: In this specific case, the very first commit does 
```
-       sseServer.CreateStream(STATUS)
+       sseServer.CreateStream(CLUSTER_LOAD)
```
and then in the last commit we have
```
diff --git a/pkg/crc/api/events/event_server.go b/pkg/crc/api/events/event_server.go
index 80ba0dc52..1a7e5155c 100644
--- a/pkg/crc/api/events/event_server.go
+++ b/pkg/crc/api/events/event_server.go
@@ -54,8 +54,9 @@ func NewEventServer(machine machine.Client) *EventServer {
                stream.RemoveSubscriber(sub)
        }
 
-       sseServer.CreateStream(LOGS)
-       sseServer.CreateStream(CLUSTER_LOAD)
+       sseServer.CreateStream(Logs)
+       sseServer.CreateStream(ClusterLoad)
+       sseServer.CreateStream(StatusChange)
        return eventServer
 }
``` 
Feels a bit pointless to change the same line twice in the same PR. The last commit being the tricky one, it's better for reviewers to make it as small as possible (ie avoid unrelated distractions), so I'd move the renames out of it in this case. First commit could be "Change LOGS/STATUS to Logs/Status" followed by "Rename Status to ClusterLoad", this way we don't ever see "CLUSTER_LOAD".

cmd/crc/cmd/status.go
evidolob: Yes, I rename that to avoid confusion with name `Status` vs `StatusChange`

pkg/crc/api/events/event_server.go
evidolob: I rewrite the commits, hopefully now it looks better

pkg/crc/api/events/cluster_load_stream.go
```diff
@@ -19,18 +19,18 @@ type TickListener struct {
...
 }
 
-func newStatusStream(server *EventServer) EventStream {
-	return newStream(NewStatusListener(server.machine), newEventPublisher(STATUS, server.sseServer))
+func newClusterLoadStream(server *EventServer) EventStream {
+	return newStream(newStatusListener(server.machine), newEventPublisher(ClusterLoad, server.sseServer))
 }
 
-func NewStatusListener(machine crcMachine.Client) EventProducer {
+func newStatusListener(machine crcMachine.Client) EventProducer {
```
cfergeau: This one was missed in the rename

pkg/crc/api/events/log_stream.go
```diff
@@ -23,7 +25,7 @@ func newSSEStreamHook(server *sse.Server) *streamHook {
 		&logrus.JSONFormatter{
 			TimestampFormat:   "",
 			DisableTimestamp:  false,
-			DisableHTMLEscape: false,
+			DisableHTMLEscape: true,
```
cfergeau: Can it cause parsing issues at the other hand if HTML is not escaped?

pkg/crc/api/events/status_change_stream.go
```diff
@@ -0,0 +1,62 @@
...
+	listenerDisposable events.Disposable
+	machineClient      machine.Client
+}
+
+func newStatusChangeStream(server *EventServer) EventStream {
+	return newStream(newStatusChangeListener(server.machine), newEventPublisher(StatusChange, server.sseServer))
+}
+
+func newStatusChangeListener(client machine.Client) EventProducer {
+	return &statusChange{
```
cfergeau: This type could be named `statusChangeListener`

pkg/crc/api/events/status_change_stream.go
```diff
@@ -0,0 +1,62 @@
...
+			}, Error: err.Error()}
+		} else {
+			status.CrcStatus = changedEvent.State // override with actual reported state
+			event = statusChangeEvent{Status: status}
+			if changedEvent.Error != nil {
+				event.Error = changedEvent.Error.Error()
+			}
+
+		}
+		data, _ := json.Marshal(event)
```
cfergeau: I don't think we should ignore this error

pkg/events/events.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/crc-org/crc/v2/pkg/crc/machine/state"
+)
+
+type StatusChangedEvent struct {
+	State state.State
+	Error error
+}
+
+var (
+	StatusChanged = NewEvent[StatusChangedEvent]()
```
cfergeau, line 13: Not sure about the global variable here, but we can refine this later.

pkg/events/emitter.go
```diff
@@ -0,0 +1,54 @@
...
+	keyCounter int
+}
+
+func NewEvent[T any]() Event[T] {
+	return &event[T]{
+		listeners: make(map[int]*listenerWrapper[T]),
+	}
+}
+
+func (e *event[T]) AddListener(listener Listener[T]) Disposable {
```
cfergeau: For some reason, I expected `Listener[T]` to be a struct or interface, while it's actually a callback func. In my opinion it's clearer to name it `ListenerFunc[T]`.
However, I think we can use an interface here instead of a callback func, this removes the need for `keyCounter`, for `Disposable`, for `listenerWrapper`, ... and in my opinion makes the code easier to read and follow. I've done that in https://github.com/cfergeau/crc/commit/24930c029e46f6dc95172decb033bc85319da972

pkg/events/emitter.go
```diff
@@ -0,0 +1,54 @@
...
+	return func() {
+		e.eventMutex.Lock()
+		defer e.eventMutex.Unlock()
+
+		delete(e.listeners, key)
+	}
+}
+
+func (e *event[T]) Fire(event T) {
+	for _, listener := range e.listeners {
```
cfergeau, line 43: This is missing `e.eventMutex.Lock()` and `defer e.eventMutex.Unlock()`

pkg/crc/api/events/status_change_stream.go
```diff
@@ -0,0 +1,62 @@
...
+		logging.Debugf("State Changed Event %s", changedEvent)
+		var event statusChangeEvent
+		status, err := st.machineClient.Status()
+		// if we cannot receive actual state, send error state with error description
+		if err != nil {
+			event = statusChangeEvent{Status: &types.ClusterStatusResult{
+				CrcStatus: state.Error,
+			}, Error: err.Error()}
+		} else {
+			status.CrcStatus = changedEvent.State // override with actual reported state
```
cfergeau: why is the cluster state not matching the reported state? This is the key thing to explain in the comment imo.

cfergeau (commented):
Another round of comments, still need to spend more time on the pkg/crc/machine/sync.go changes.

pkg/events/emitter.go
cfergeau: (I've added a 2nd commit to fix compilation of the testcase)

pkg/crc/api/events/log_stream.go
evidolob: Yes, especially new line characters

pkg/crc/machine/sync.go
```diff
@@ -92,6 +98,13 @@ func (s *Synchronized) Start(ctx context.Context, startConfig types.StartConfig)
 
 	startResult, err := s.underlying.Start(ctx, startConfig)
 	s.syncOperationDone <- Starting
+
+	if err == nil {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: startResult.Status})
+	} else {
+		events.StatusChanged.Fire(events.StatusChangedEvent{State: state.Error, Error: err})
+	}
```
cfergeau, line 106: (commenting here, but that applies to most of the file).
Duplicating the state logic in this file ("after calling `Start()`, the VM is in the `Start` state, unless there was an error, in which case it's in the `Error` state") will likely cause additional maintainance burden, as we have to make sure to keep in sync the state logic in this file, and the state logic in other parts of crc - the status events may be reporting a different state from `Status()`.

Can we emit a `StatusChangedEvent` to tell other code the status changed, but not specify the current status in the event payload? This way the receiver of the event can check `Status()` if they need the status, and we don't have 2 source for the status to keep in sync?

Regarding errors, it seems if `Start()` (or other commands) return an error, you consider that the cluster is in an error state? This makes sense for some commands, but I don't think this matches what is reported by `Status()` at the moment? Would be nice to fix the reported Status() when there are start failures for examples, but doing it here would be papering over the real issue in my opinion.
And in some other cases, errors are less bad, for example in the power off case, maybe the command failed early, and the cluster is still nicely running?

evidolob: >Duplicating the state logic in this file ("after calling Start(), the VM is in the Start state, unless there was an error, in which case it's in the Error state") will likely cause additional maintainance burden, as we have to make sure to keep in sync the state logic in this file, and the state logic in other parts of crc - the status events may be reporting a different state from Status().

Not sure to understand how to resolve you concern, the problem is that we don't have a way to track state change(at least I wasn’t able to find it, it would be nice if you point me that place if if it exist)


> Can we emit a StatusChangedEvent to tell other code the status changed, but not specify the current status in the event payload? This way the receiver of the event can check Status() if they need the status, and we don't have 2 source for the status to keep in sync?

I first do that, but I face a race condition with updating state, for example:
User call start, we fire state change event, before actuality call `Start`, and if listener is fast enough, it could get old state (basically we could lost `starting` state in transition from `stoped` to `running` states)

Possible solution to that, may be moving state change event firing from `Synchronized` client in to `client` implementation of `machine.Client` interface. And fire events only after actually state is changed.


> Regarding errors, it seems if Start() (or other commands) return an error, you consider that the cluster is in an error state? This makes sense for some commands, but I don't think this matches what is reported by Status() at the moment? Would be nice to fix the reported Status() when there are start failures for examples, but doing it here would be papering over the real issue in my opinion.
And in some other cases, errors are less bad, for example in the power off case, maybe the command failed early, and the cluster is still nicely running?

It is matched, as we send with SSE status reported by `Status()` in addition we also send error message, which come from status change event.

pkg/events/emitter.go
```diff
@@ -0,0 +1,46 @@
...
+	e.eventMutex.Lock()
+	defer e.eventMutex.Unlock()
+	delete(e.listeners, listener)
+}
+
+func (e *event[T]) Fire(event T) {
+	e.eventMutex.Lock()
+	defer e.eventMutex.Unlock()
+	for _, listener := range e.listeners {
+		go listener.Notify(event)
```
cfergeau, line 46: I wonder if this is affected by the bug decribed here? https://go.dev/doc/faq#closures_and_goroutines

pkg/crc/machine/sync.go
cfergeau: My main concern is https://github.com/crc-org/crc/pull/3903#discussion_r1420717479 , and I'm under the impression that if you have a terminal showing the stream of events, and another running `crc status` every second, then the state between the 2 will not always match, in particular sometimes the event stream will contain `State: state.Error` while `crc status` will not report an error?
Guess I need to run both side by side and compare myself to understand better how this works.

cfergeau: (I understand the current code base is missing a centralized "setState" facility, and that this PR has to find a way of doing this, just trying to understand the limitations in the PR)

evidolob: >My main concern is https://github.com/crc-org/crc/pull/3903#discussion_r1420717479 , and I'm under the impression that if you have a terminal showing the stream of events, and another running crc status every second, then the state between the 2 will not always match, in particular sometimes the event stream will contain State: state.Error while crc status will not report an error?
Guess I need to run both side by side and compare myself to understand better how this works.

No, output should be exactly the same, to get that I add
 https://github.com/crc-org/crc/pull/3903/files#diff-e5710ade03da511d4845f380cd63319bf7329e960c1afbbba7fbed4060fcf8edR37 
and 
https://github.com/crc-org/crc/pull/3903/files#diff-e5710ade03da511d4845f380cd63319bf7329e960c1afbbba7fbed4060fcf8edR44-R50

Otherwise we need to move event triggering deeper in crc code, and fire event after execution of pice of code which change `crc status` output.

And it is a question for all, should we integrate events in more depths of CRC core(which leads to spreading events over our codebase) or keep as is( in relative hi level Client interface implementation )?

@gbraad @cfergeau @praveenkumar WDYT?

pkg/events/emitter.go
evidolob: Yes, it is. Something knew learned about golang. 
Fixed.

cfergeau:
I've rebased it at https://github.com/cfergeau/crc/tree/stus-change-event and added a fixup commit to fix compliation of the first commit in the series.

cfergeau:
I'm not sure step 3. is correct in the testing instructions:
> 3. start crc with crc start

Don't we need to start it with the REST API for now as otherwise `start` won't get through the daemon?
```
 curl -GET -i --unix-socket ~/.crc/crc-http.sock  "http://localhost/api/start"
```

I was wondering if `curl -GET -i --unix-socket ~/.crc/crc-http.sock  "http://localhost/api/start"` should also send a status event to help with syncing state if the state change behind the daemon's back?

evidolob:
> Don't we need to start it with the REST API for now as otherwise `start` won't get through the daemon?

Yes, you are correct, need to start with rest API.

> 
> ```
>  curl -GET -i --unix-socket ~/.crc/crc-http.sock  "http://localhost/api/start"
> ```
> 
> I was wondering if `curl -GET -i --unix-socket ~/.crc/crc-http.sock "http://localhost/api/start"` should also send a status event to help with syncing state if the state change behind the daemon's back?

Daemon send response only after start is finished(or error incase of any error), but calling `api/start` **will** trigger events on `events?stream=status_change` event channel.

cfergeau:
Few more things I noticed during testing, if VM exists, and then I call `delete`, I get
```
id: 
data: {"status":{"CrcStatus":"Error","OpenshiftStatus":"","OpenshiftVersion":"","PodmanVersion":"","DiskUse":0,"DiskSize":0,"RAMUse":0,"RAMSize":0,"Preset":""},"error":"Cannot load 'crc' virtual machine: no such libmachine vm: crc"}
event: status_change
```
which does not seem correct as the VM existed before being deleted.
After that, if I call again `delete` I don't get any notifications. If I call `stop`, I get first `stopping` and then `error`

Probably something that can improve with the `NoVM` state.

pkg/crc/api/events/log_stream.go
cfergeau: I don't fully understand what this solves, we remove the new line characters with `TrimRight`, so `DisableHTMLEscape: true` must be solving a different problem?

evidolob: True, delete this change

evidolob:
> Few more things I noticed during testing, if VM exists, and then I call `delete`, I get
> 
> ```
> id: 
> data: {"status":{"CrcStatus":"Error","OpenshiftStatus":"","OpenshiftVersion":"","PodmanVersion":"","DiskUse":0,"DiskSize":0,"RAMUse":0,"RAMSize":0,"Preset":""},"error":"Cannot load 'crc' virtual machine: no such libmachine vm: crc"}
> event: status_change
> ```
> 
> which does not seem correct as the VM existed before being deleted. After that, if I call again `delete` I don't get any notifications. If I call `stop`, I get first `stopping` and then `error`

That was actually bug, `status` was not report `NoVm` state, only error. (It was due to not implementing `Is` method for `MissingHostError`, so comparing error was always wrong)

Now it should work properly.

pkg/crc/machine/virtualmachine.go
```diff
@@ -32,6 +32,14 @@ func (err *MissingHostError) Error() string {
 	return fmt.Sprintf("no such libmachine vm: %s", err.name)
 }
 
+func (err *MissingHostError) Is(target error) bool {
+	var x *MissingHostError
+	if errors.As(target, &x) && x.name == err.name {
+		return true
+	}
+	return false
+}
```
cfergeau, line 41: This probably addresses some of the remarks I made before, but some explanations are always nice to have :)

pkg/crc/api/events/status_change_stream.go
```diff
@@ -0,0 +1,69 @@
...
+
+func newStatusChangeListener(client machine.Client) EventProducer {
+	return &statusChangeListener{
+		machineClient: client,
+	}
+}
+
+func (st *statusChangeListener) Notify(changedEvent events.StatusChangedEvent) {
+	logging.Debugf("State Changed Event %s", changedEvent)
+	var event statusChangeEvent
```
cfergeau: Took me a while to realize there's a `statusChangeEvent` type and a `events.StatusChangedEvent` type, and that both are different, I could not understand how we serialized `changedEvent.State` to something with `Status` in the json :-/ 
Maybe naming can be improved? `changedEvent events.StateChangedEvent`? Still too close to the other name, but at least there's a small difference. Maybe we can come up with something even better.

pkg/crc/api/events/status_change_stream.go
```diff
@@ -0,0 +1,69 @@
...
+	status, err := st.machineClient.Status()
+	// if we cannot receive actual state, send error state with error description
+	if err != nil {
+		event = statusChangeEvent{Status: &types.ClusterStatusResult{
+			CrcStatus: state.Error,
+		}, Error: err.Error()}
+	} else {
+		// event could be fired, before actual code, which change state is called
+		// so status could contain 'old' state, replace it with state received in event
+		status.CrcStatus = changedEvent.State // override with actual reported state
```
cfergeau, line 46: This is the part I find questionable in this PR. Instead of having a single source of truth for the status, we now have 2, some of the status comes from `st.machineClient.Status()`, but some of the status is hardcoded in `machine/sync.go`. If the 2 get out of sync, or if we want to improve status reporting in general, we'll have multiple places to change.

cfergeau: Spending more time on it, I understand more the problems this is trying to solve.
One example is when the VM is stopped. `crc status` will succeed and report that crc is stopped, however `crc stop` will return an error and say the machine is already stopped.

Without `status.CrcStatus = changedEvent.State`, when invoking `api/stop`, we would not be reporting an error in the stream, but instead we'd only say the machine is stopped. This would not match what was reported to the `api/stop` call. There are most likely other state transitions where this is needed.

The comment however seems to indicate sometimes `Status` could report a stale status, I haven't seen that in the little testing I did, but I'd be interested in understanding more when this happens? If we don't record this knowledge now when it's more or less fresh in our minds, it will be very difficult to find this back in 6 months from now if we need it :-/

cfergeau: ah, there's more details in an older discussion:
> User call start, we fire state change event, before actuality call Start, and if listener is fast enough, it could get old state (basically we could lost starting state in transition from stoped to running states)

cfergeau: In my opinion, this is an indication that these events are not fired in the right place, and this is papered over with even more complexity :-/

pkg/crc/machine/sync.go
cfergeau: > And it is a question for all, should we integrate events in more depths of CRC core(which leads to spreading events over our codebase) or keep as is( in relative hi level Client interface implementation )?

I'd say we can go with this for now, more details about the design choices, the implementation limitations, ...

One thing we can explore over time some centralized tracking of the cluster state, a `StateTracker` type, we'd add some calls to `StateTracker.SetState(...)` in the appropriate places, the `StateTracker` would fire events on status changes, it could have a `CanChangeState(state)` helper if needed to know if calling `Stop` will result in an error or not, ...
As long as all `crc` commands do not go through the daemon, it could do some polling to reconcile the internal state with changes through external `crc` calls, ...

pkg/crc/api/events/status_change_stream.go
gbraad: > t instead we'd only say the machine is stopped. This would not match what was reported to the api/stop

The API matches the usecase. Stop should in my opinion not return an error when already stopped; as you succeed in doing what is requested. What more information do we actually give with saying this failed as it was already stopped?!

If you believe this should be fixed or recorded, make these states clear for the convergence.

gbraad:
What is holding back this PR?

pkg/crc/api/events/status_change_stream.go
cfergeau: The part you quoted is a description of the behaviour in this PR. These are some notes I took during review/testing when I was trying to understand the design choices which led to the code being reviewed.

cfergeau:
> What is holding back this PR?

Some discussion around https://github.com/crc-org/crc/pull/3903#discussion_r1476130780 would have been useful

Apart from this, as said multiple times, the comit log must be a lot more detailed about how this works:
>  If we don't record this knowledge now when it's more or less fresh in our minds, it will be very difficult to find this back in 6 months from now if we need it :-/

> I'd say we can go with this for now, more details about the design choices, the implementation limitations, ... 
(this sentence is malformed though :-/ )

And https://github.com/crc-org/crc/pull/3903#discussion_r1474772073 is also still pending

pkg/crc/api/events/status_change_stream.go
evidolob: @cfergeau I change type name, hopefully it is more clear now

pkg/crc/api/events/status_change_stream.go
evidolob: I try to replace this code, with firing event at more "proper" place, but face problem that `starting` state is "artificial", ie exist only in `sync.go` file, and VM doesn't have/reports `starting` state. Also `stopping` state is the same. 
So I have no other ideas, than accept current PR and think on better state handling, like @cfergeau propose  there https://github.com/crc-org/crc/pull/3903#discussion_r1476135076

Or change state handling first, and then rebase this PR on top of it.

@gbraad @cfergeau WDYT?

cfergeau: I think we can go forward with the current code, but we really need more details in the commit log, so that it records the design choices you made when writing the code, so that we know the current limitations, why this overwrites the state, ...

evidolob:
@cfergeau Could you look on this?

evidolob:
Closing this, as priority changed. I keep the branch in case if we want to return to this.

Issue #3602, opened 2023-04-14
Follow-up: Try to use ModifyConfig function as part of merging the kubeconfig file


praveenkumar:
As of now to merge the 2 kubeconfig files we create separate objects of the config and the iterate second one and put all the cluster/auth/user info but kube go-client have `ModifyConfg`  function in  `tools/clientcmd/config.go` package, need to figure out if that also do same or not. 
 
discussion: https://github.com/crc-org/crc/pull/3597#discussion_r1164484184

gbraad:
@praveenkumar it concerns this piece of code, right?

https://github.com/crc-org/crc/blob/2875a7441f467e630c62760b49a800f60857afbf/pkg/crc/machine/kubeconfig.go#L235-L264

praveenkumar:
@gbraad yes this is piece of code we want to improve.

anjannath:
done via #3910

PR #3910, opened 2023-11-07
Merge kubeconfigs using clientcmd.ModifyConfig

vyasgun:
**Fixes:** Issue #3602 

## Solution/Idea

Use `clientcmd.ModifyConfig` to merge custom config with a primary config (global kubeconfig in default case)

## Proposed changes

Changed the function call to ModifyConfig in `start`

## Testing

1. `start` should merge the custom kubeconfig to the global kubeconfig file

### DIFF
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -84,6 +84,10 @@ func writeKubeconfig(ip string, clusterConfig *types.ClusterConfig, ingressHTTPS
 
 func getGlobalKubeConfig() (string, *api.Config, error) {
 	kubeconfig := getGlobalKubeConfigPath()
+	return getKubeConfigFromFile(kubeconfig)
+}
+
+func getKubeConfigFromFile(kubeconfig string) (string, *api.Config, error) {
 	config, err := clientcmd.LoadFromFile(kubeconfig)
 	if err != nil && !os.IsNotExist(err) {
 		return "", nil, err
@@ -246,7 +250,12 @@ func contains(arr []string, str string) bool {
 }
 
 func mergeKubeConfigFile(kubeConfigFile string) error {
-	globalConfigPath, globalConf, err := getGlobalKubeConfig()
+	return mergeConfigHelper(kubeConfigFile, getGlobalKubeConfigPath())
+}
+
+func mergeConfigHelper(kubeConfigFile, globalConfigFile string) error {
+
+	globalConfigPath, globalConf, err := getKubeConfigFromFile(globalConfigFile)
 	if err != nil {
 		return err
 	}
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -246,32 +246,33 @@ func contains(arr []string, str string) bool {
 }
 
 func mergeKubeConfigFile(kubeConfigFile string) error {
-	globalConfigPath, kubeConf1, err := getGlobalKubeConfig()
-	if err != nil {
-		return err
-	}
-	kubeConf2, err := clientcmd.LoadFromFile(kubeConfigFile)
-	if err != nil {
-		return err
-	}
-	// Merge the kubeConf2 to globalConfig
-	mergedConfig, err := clientcmd.NewDefaultClientConfig(*kubeConf1, &clientcmd.ConfigOverrides{}).ConfigAccess().GetStartingConfig()
+	globalConfigPath, globalConf, err := getGlobalKubeConfig()
 	if err != nil {
 		return err
 	}
 
-	for name, cluster := range kubeConf2.Clusters {
+	currentConf, err := clientcmd.LoadFromFile(kubeConfigFile)
+	if err != nil {
+		return err
+	}
+	// Merge the currentConf to globalConfig
+	mergedConfig, err := clientcmd.NewDefaultClientConfig(*globalConf, &clientcmd.ConfigOverrides{}).ConfigAccess().GetStartingConfig()
+	if err != nil {
+		return err
+	}
+
+	for name, cluster := range currentConf.Clusters {
 		mergedConfig.Clusters[name] = cluster
 	}
 
-	for name, authInfo := range kubeConf2.AuthInfos {
+	for name, authInfo := range currentConf.AuthInfos {
 		mergedConfig.AuthInfos[name] = authInfo
 	}
 
-	for name, context := range kubeConf2.Contexts {
+	for name, context := range currentConf.Contexts {
 		mergedConfig.Contexts[name] = context
 	}
 
-	mergedConfig.CurrentContext = kubeConf2.CurrentContext
+	mergedConfig.CurrentContext = currentConf.CurrentContext
 	return clientcmd.WriteToFile(*mergedConfig, globalConfigPath)
 }
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -256,23 +256,18 @@ func mergeKubeConfigFile(kubeConfigFile string) error {
 		return err
 	}
 	// Merge the currentConf to globalConfig
-	mergedConfig, err := clientcmd.NewDefaultClientConfig(*globalConf, &clientcmd.ConfigOverrides{}).ConfigAccess().GetStartingConfig()
-	if err != nil {
-		return err
-	}
-
 	for name, cluster := range currentConf.Clusters {
-		mergedConfig.Clusters[name] = cluster
+		globalConf.Clusters[name] = cluster
 	}
 
 	for name, authInfo := range currentConf.AuthInfos {
-		mergedConfig.AuthInfos[name] = authInfo
+		globalConf.AuthInfos[name] = authInfo
 	}
 
 	for name, context := range currentConf.Contexts {
-		mergedConfig.Contexts[name] = context
+		globalConf.Contexts[name] = context
 	}
 
-	mergedConfig.CurrentContext = currentConf.CurrentContext
-	return clientcmd.WriteToFile(*mergedConfig, globalConfigPath)
+	globalConf.CurrentContext = currentConf.CurrentContext
+	return clientcmd.WriteToFile(*globalConf, globalConfigPath)
 }
--- a/pkg/crc/machine/kubeconfig_test.go
+++ b/pkg/crc/machine/kubeconfig_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 )
 
 var dummyKubeconfigFileContent = `apiVersion: v1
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@
 /tmp-embed/
 /RPMS/
 /crc
+.idea
```
praveenkumar: This should be part of your local git global config instead here.

pkg/crc/machine/kubeconfig.go
```diff
@@ -242,33 +242,42 @@ func contains(arr []string, str string) bool {
 	return false
 }
 
-func mergeKubeConfigFile(kubeConfigFile string) error {
-	globalConfigPath, kubeConf1, err := getGlobalKubeConfig()
-	if err != nil {
-		return err
+func mergeKubeConfigFile(kubeConfigFile, globalConfigFile string) error {
```
praveenkumar: Is this function signature only changed to have test around it?

vyasgun: yes, I changed it because it was being used in just one place and so it can be tested more easily

pkg/crc/machine/kubeconfig_test.go
```diff
@@ -1,11 +1,14 @@
 package machine
 
 import (
+	"io/ioutil"
```
praveenkumar: This is deprecated, use `os` package

.gitignore
gbraad: However, some files we can add to gitignore to prevent them littering the repo.

pkg/crc/machine/kubeconfig.go
```diff
@@ -242,33 +242,42 @@ func contains(arr []string, str string) bool {
...
+		kubeConf1.Contexts[name] = context
 	}
 
-	mergedConfig.CurrentContext = kubeConf2.CurrentContext
-	return clientcmd.WriteToFile(*mergedConfig, globalConfigPath)
+	configAccess := clientcmd.NewDefaultPathOptions()
+	configAccess.GlobalFile = globalConfigFile
+	configAccess.LoadingRules.ExplicitPath = globalConfigFile
+
+	//fmt.Printf("GUNJAN: %v\n%v\n%v\n", configAccess.GetLoadingPrecedence(), kubeConf1, kubeConf2)
```
gbraad: Please do not commit debug code ;-)

pkg/crc/machine/kubeconfig.go
```diff
@@ -242,33 +242,42 @@ func contains(arr []string, str string) bool {
...
 	for name, context := range kubeConf2.Contexts {
-		mergedConfig.Contexts[name] = context
+		kubeConf1.Contexts[name] = context
 	}
 
-	mergedConfig.CurrentContext = kubeConf2.CurrentContext
-	return clientcmd.WriteToFile(*mergedConfig, globalConfigPath)
+	configAccess := clientcmd.NewDefaultPathOptions()
+	configAccess.GlobalFile = globalConfigFile
+	configAccess.LoadingRules.ExplicitPath = globalConfigFile
```
cfergeau: Looking at https://github.com/kubernetes/client-go/blob/86d49e7265f07676cb39f342595a858b032112de/tools/clientcmd/config.go#L145-L158 , the defaults for `GlobalFile` and `LoadingRules` look useful when `globalKubeConfigFile` is `""`? In other words, maybe this should be changed to:
```
if globalConfigFile != "" {
    configAccess.GlobalFile = globalConfigFile
    configAccess.LoadingRules.ExplicitPath = globalConfigFile
}
```
? Or is this also required when modifying the global kubeconfig file?

pkg/crc/machine/kubeconfig.go
```diff
@@ -242,33 +242,42 @@ func contains(arr []string, str string) bool {
...
 
 	for name, authInfo := range kubeConf2.AuthInfos {
-		mergedConfig.AuthInfos[name] = authInfo
+		kubeConf1.AuthInfos[name] = authInfo
 	}
 
 	for name, context := range kubeConf2.Contexts {
-		mergedConfig.Contexts[name] = context
+		kubeConf1.Contexts[name] = context
 	}
```
cfergeau, line 278: This function ends with `return clientcmd.ModifyConfig(configAccess, *kubeConf1, true)` which is documented as:
```
// ModifyConfig takes a Config object, iterates through Clusters, AuthInfos, and Contexts, uses the LocationOfOrigin if specified or uses the default destination file to write the results into.
```

I was hoping that using `ModifyConfig` would allow us to remove these iterations over `Clusters`, `AuthInfos` and `Contexts`?

cfergeau:
Wondering if `mergeKubeConfigFile` could become as simple as something like this?

```
func mergeKubeConfigFile(kubeConfigFile, globalConfigFile string) error {
	configAccess := clientcmd.NewDefaultPathOptions()
	if globalConfigFile != "" {
		configAccess.GlobalFile = globalConfigFile
		configAccess.LoadingRules.ExplicitPath = globalConfigFile
	}

	kubeConf2, err := clientcmd.LoadFromFile(kubeConfigFile)
	if err != nil {
		return err
	}

	return clientcmd.ModifyConfig(configAccess, *kubeConf2, true)
}
```

.gitignore
vyasgun: I find keeping this in `.gitignore` is helpful for developing on Goland.

pkg/crc/machine/kubeconfig.go
vyasgun: My bad 😓 Removed it

pkg/crc/machine/kubeconfig.go
vyasgun: I just updated the code again. I found that the `.LocationOfOrigin` field was being given priority in choosing the output file so I changed it to the global kubeconfig file.

pkg/crc/machine/kubeconfig.go
vyasgun: I couldn't find a way to do it without these iterations

.gitignore
cfergeau: You can add `.idea` to `~/.config/git/ignore`, it's applied to all git repositories, this way you don't need to manually update .gitignore in each project you contribute to.

cfergeau:
> Wondering if `mergeKubeConfigFile` could become as simple as something like this?
> 
> ```
> func mergeKubeConfigFile(kubeConfigFile, globalConfigFile string) error {
> 	configAccess := clientcmd.NewDefaultPathOptions()
> 	if globalConfigFile != "" {
> 		configAccess.GlobalFile = globalConfigFile
> 		configAccess.LoadingRules.ExplicitPath = globalConfigFile
> 	}
> 
> 	kubeConf2, err := clientcmd.LoadFromFile(kubeConfigFile)
> 	if err != nil {
> 		return err
> 	}
> 
> 	return clientcmd.ModifyConfig(configAccess, *kubeConf2, true)
> }
> ```

`ModifyConfig` + `LoadFromFile` behaves oddly, `LoadFromFile` will set `LocationOfOrigin` on all clusters/authinfos/... to the file being loaded. And `ModifyConfig` will write the changes to the file referenced by LocationOfOrigin, which is not what we want :-/

vyasgun:
@cfergeau thanks for the review! I couldn't find a way to override the preference for locationOfOrigin field. What can we do next?

pkg/crc/machine/kubeconfig.go
```diff
@@ -242,33 +242,48 @@ func contains(arr []string, str string) bool {
...
-	globalConfigPath, kubeConf1, err := getGlobalKubeConfig()
+func mergeKubeConfigFile(kubeConfigFile, globalConfigFile string) error {
+	var kubeConf1 *api.Config
+	var err error
+
+	if globalConfigFile == "" {
+		globalConfigFile = getGlobalKubeConfigPath()
+	}
+
+	kubeConf1, err = clientcmd.LoadFromFile(globalConfigFile)
```
cfergeau: I'd recommend giving these `kubeConf1` / `kubeConf2` variables more explicit names, to avoid potential confusion. `kubeConf1` could be named `globalKubeConf` for example, it is more obvious what it represents. Or just `destKubeConf`/`srcKubeConf`

.gitignore
cfergeau: You marked this conversation as "resolved", not sure if this is because the `~/.config/git/ignore` suggestion works for you, or because you have a strong opinion for having the `.idea` file in crc's `.gitignore`?

pkg/crc/machine/start.go
```diff
@@ -568,7 +568,7 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 			}
 		}
 		logging.Info("Adding microshift context to kubeconfig...")
-		if err := mergeKubeConfigFile(constants.KubeconfigFilePath); err != nil {
+		if err := mergeKubeConfigFile(constants.KubeconfigFilePath, ""); err != nil {
```
cfergeau: For what it's worth, adding a dummy argument to a function in the "main" codepath because it will be used in unit testing is not great for readability. One may wonder "What does "" stand for?" while reading the code.

You can achieve the same with an intermediate function 
```
func mergeToGlobalKubeConfigFile(kubeConfigFile string) error {
	return mergeKubeConfigFile(kubeConfigFile, getGlobalKubeConfigPath())
}
```
And remove the special handling of `""` in `mergeKubeConfigFile`.

pkg/crc/machine/kubeconfig_test.go
```diff
@@ -59,3 +61,75 @@ func TestUpdateUserCaAndKeyToKubeconfig(t *testing.T) {
...
+	err = clientcmd.WriteToFile(*config, path)
+	if err != nil {
+		os.Remove(path)
+		return "", err
+	}
+
+	return path, nil
+}
+
+func Test_mergeKubeConfigFile(t *testing.T) {
```
cfergeau, line 81: The `TestMergeKubeConfigFile` naming is more common in go tests

cfergeau (commented):
> thanks for the review! I couldn't find a way to override the preference for locationOfOrigin field. What can we do next?

Yeah, not really clear what to do :-/ I was hoping using `clientcmd.ModifyConfig` would allow us to remove the need to iterate over the `Contexts`/`AuthInfos`/.. as it's not really useful for crc to know about such details of the kubeconfig content.

It's apparently not possible because `ModifyConfig` does not have the behaviour we want/expected/hoped for, and using it makes `mergeKubeConfigFile` more complex.
On the other hand, this PR adds a unit test which is very good to have! 
Maybe we should change this PR to add the unit test, but keep the current implementation of `mergeKubeConfigFile` as the attempt to use `clientcmd.ModifyConfig` did not turn out as expected?

gbraad:
There is a valid point of consideration that this is actually more complex. This is not the fault of the author...

@praveenkumar you think this is an improvement from the perspective of maintenance?

Regardless, the testcase is good to include...

vyasgun:
As per the discussions and reviews, I think we can revert the changes made to use `clientcmd.ModifyConfig` as it is not working the way we were expecting it to and just keep the testcase.

pkg/crc/machine/start.go
vyasgun: Thanks, I'll make this change.

pkg/crc/machine/kubeconfig_test.go
```diff
@@ -59,3 +61,75 @@ func TestUpdateUserCaAndKeyToKubeconfig(t *testing.T) {
...
+		}
+		for key := range config.Clusters {
+			if _, ok := mergedConfig.Clusters[key]; !ok {
+				t.Errorf("key %v not found", key)
+			}
+		}
+		for key := range config.Contexts {
+			if _, ok := mergedConfig.Contexts[key]; !ok {
+				t.Errorf("key %v not found", key)
+			}
```
anjannath: you can also use  `assert.Contains` (https://pkg.go.dev/github.com/stretchr/testify/assert#Contains) in place of the `if` checks
```go
if _, ok := mergedConfig.Contexts[key]; !ok {
	t.Errorf("key %v not found", key)
}
```
this would become:
```go
assert.Contains(t, mergedConfig.Contexts, key, "expected context not found")
```

pkg/crc/machine/kubeconfig_test.go
```diff
@@ -59,3 +61,75 @@ func TestUpdateUserCaAndKeyToKubeconfig(t *testing.T) {
...
+	err = mergeKubeConfigFile(secondaryConfigPath, primaryConfigPath)
+	if err != nil {
+		t.Fatalf("failed to modify kubeconfig: %v", err)
+	}
+
+	// Load the modified kubeconfig to ensure it was merged correctly
+	mergedConfig, err := clientcmd.LoadFromFile(primaryConfigPath)
+	if err != nil {
+		t.Fatalf("failed to load merged kubeconfig: %v", err)
+	}
```
anjannath: can use `assert.NoError` for the error checks, https://pkg.go.dev/github.com/stretchr/testify/assert#NoError

pkg/crc/machine/kubeconfig_test.go
vyasgun: I think that's the convention for exported functions

.gitignore
cfergeau: Can we remove this `.idea` change? :)

cfergeau (commented):
I find the current set of changes a bit hard to review, as they are doing more than just adding the test case :-/
I've split them in https://github.com/cfergeau/crc/commits/pr/mergeconfig in 3 commits each doing one logical thing.
The first one renames kubeConf1/kubeConf2, the second one simplifies the function a bit (if I understood what it's intended to do), and the last one adds the test case. 
With 3 commits instead of 1, it's easier to be sure we understand what each change is doing, and that there is not an unexpected change which should not be there.

pkg/crc/machine/kubeconfig_test.go
```diff
@@ -59,3 +61,63 @@ func TestUpdateUserCaAndKeyToKubeconfig(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, "dummycert", userClientCA)
 }
+
+func createTempKubeConfig(config *api.Config) (string, error) {
+	tempFile, err := os.CreateTemp("", "kubeconfig-")
```
cfergeau, line 66: https://pkg.go.dev/testing#T.TempDir is convenient in tests, no need to check for errors and all its content is automatically cleaned up after the test is done.

cfergeau (approved):

PR #3912, opened 2023-11-08
Remove option to customize installation

### LINKED ISSUE #3909: Remove choice of component from the installers
In #3904 while remove the trays from the installers, the choice for selecting which component to install from the installer UI was not removed, since we now have only one component the choice from the UI should also be removed, this was also mentioned in https://github.com/crc-org/crc/pull/3904#discussion_r1381438824

anjannath:
in #3904 this was missed as there's only one component available to install now we can remove the options  to customize the installation

fixes #3909

### DIFF
--- a/packaging/darwin/Distribution.in
+++ b/packaging/darwin/Distribution.in
@@ -5,7 +5,7 @@
     <welcome file="welcome.html" mime-type="text/html" />
     <conclusion file="conclusion.html" mime-type="text/html" />
     <license file="LICENSE.txt"/>
-    <options customize="allow" allow-external-scripts="no"/>
+    <options customize="never" allow-external-scripts="no"/>
     <domains enable_localSystem="true" />
     <options rootVolumeOnly="true"/>
     <options hostArchitectures="x86_64,arm64" />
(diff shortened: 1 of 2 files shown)


cfergeau (commented):
Wondering if the tray gets automatically removed on upgrades now that it's removed from installers?

cfergeau (approved):

anjannath:
> Wondering if the tray gets automatically removed on upgrades now that it's removed from installers?

I checked on macOS and not it doesn't uninstall the tray on upgrade

on windows i haven't checked yet (don't have the windows machine with me today) but there the upgrade process is, first uninstall existing then install new one, so there most likely the installed tray will be removed during upgrade

for macOS we can add `rm -rf /Applications/Red Hat OpenShift Local.app` to the `postinstall` script and that'd make the behavior consistent for both the platforms

cfergeau:
> > Wondering if the tray gets automatically removed on upgrades now that it's removed from installers?
> 
> I checked on macOS and not it doesn't uninstall the tray on upgrade
> 
> on windows i haven't checked yet (don't have the windows machine with me today) but there the upgrade process is, first uninstall existing then install new one, so there most likely the installed tray will be removed during upgrade
> 
> for macOS we can add `rm -rf /Applications/Red Hat OpenShift Local.app` to the `postinstall` script and that'd make the behavior consistent for both the platforms

+ something to stop the tray if it's running? This can be handled in a followup issue.

cfergeau (approved):

PR #3913, opened 2023-11-08
Check daemon version before failing in checkDaemonSystemdService

### LINKED ISSUE #186: [BUG] libvirtd daemon preflight check keeps failing
### General information

  * OS: Linux/Fedora 38
 
## CRC version
```bash
2.28.0
```
  
## CRC status
```bash
crc status
```

## CRC config
N/A

### Steps to reproduce

  1. run `crc setup`
  2. run `crc setup --check-only`
  3. run `echo $?`

### Expected

step 3 prints 0

### Actual

step 3 prints 2 on linux

### Logs

```
crc setup --check-only --log-level debug
DEBU CRC version: 2.28.0+e1f42d                   
DEBU OpenShift version: 4.13.14                   
DEBU Podman version: 4.4.4                        
DEBU Running 'crc setup'                          
INFO Using bundle path /home/eskimo/.crc/cache/crc_libvirt_4.13.14_amd64.crcbundle 
DEBU Checking if systemd-resolved.service is running 
DEBU Running 'systemctl status systemd-resolved.service' 
DEBU systemd-resolved.service is already running  
DEBU Running '/home/eskimo/.crc/bin/crc-admin-helper-linux --version' 
DEBU Found crc-admin-helper-linux version 0.0.12  
DEBU crc-admin-helper executable already cached   
DEBU GOARCH is amd64 GOOS is linux                
DEBU Total memory of system is 66757939200 bytes  
DEBU Checking if the vmx/svm flags are present in /proc/cpuinfo 
DEBU CPU virtualization flags are good            
DEBU Checking if /dev/kvm exists                  
DEBU /dev/kvm was found                           
DEBU Checking if 'virsh' is available             
DEBU 'virsh' was found in /usr/bin/virsh          
DEBU Checking 'virsh capabilities' for libvirtd/qemu availability 
DEBU Running 'virsh --readonly --connect qemu:///system capabilities' 
DEBU Found x86_64 hypervisor with 'hvm' capabilities 
DEBU Checking if current user is part of the libvirt group 
DEBU Current user is already in the libvirt group 
INFO Checking if active user/process is currently part of the libvirt group 
DEBU Checking if libvirtd service is running      
DEBU Running 'systemctl status virtqemud.socket'  
DEBU Command failed: exit status 3                
DEBU stdout: * virtqemud.socket - Libvirt qemu local socket
     Loaded: loaded (/usr/lib/systemd/system/virtqemud.socket; disabled; preset: disabled)
     Active: inactive (dead)
   Triggers: * virtqemud.service
     Listen: /run/libvirt/virtqemud-sock (Stream) 
DEBU stderr:                                      
DEBU virtqemud.socket is neither running nor listening 
DEBU Running 'systemctl status libvirtd.socket'   
DEBU libvirtd.socket is running                   
DEBU Checking if libvirt version is >=3.4.0       
DEBU Running 'virsh -v'                           
DEBU Checking if crc-driver-libvirt is installed  
DEBU Running '/home/eskimo/.crc/bin/crc-driver-libvirt version' 
DEBU Found crc-driver-libvirt version 0.13.6      
DEBU crc-driver-libvirt is already installed      
INFO Checking crc daemon systemd service          
DEBU Checking crc daemon systemd service          
DEBU Checking if crc-daemon.service is running    
DEBU Running 'systemctl --user status crc-daemon.service' 
DEBU crc-daemon.service is running                
DEBU crc-daemon.service is running                
crc-daemon.service is running
```

cfergeau:
On linux, checkDaemonSystemdService returns an error when the daemon is already running as we want to restart the daemon to handle upgrade scenarios.
This check currently fails even if there were no crc upgrades in the mean time. This is usually harmless as stopping the daemon is quick and transparent to the user.

This becomes a problem in this scenario:
```
$ crc setup
$ crc setup --check-only
$ echo $?
0

# do something which starts the daemon, crc start for example

$ crc setup --check-only
$ echo $?
2
```

`crc setup` would fix this, but it's unfortunate that it has to be run again and again for `crc setup --check-only` to succeed.

This commit improves checkDaemonSystemdService, when it detects that the daemon is running, it checks its version, if it matches the current crc version, then the check is successful.
If the crc version does not match, then the check is unsuccessful.

This will cause one change during development: when working on the daemon, it will have to be manually stopped.
In current git, `crc setup` would do this for you.

This fixes https://github.com/crc-org/crc-extension/issues/186

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -330,9 +330,9 @@ func checkSystemdUnit(unitName string, unitContent string, shouldBeRunning bool)
 	logging.Debugf("Checking if %s is running", unitName)
 	running := systemdUnitRunning(sd, unitName)
 	if !running && shouldBeRunning {
-		return fmt.Errorf("%s is not running", unitName)
+		return unitShouldBeRunningErr(unitName)
 	} else if running && !shouldBeRunning {
-		return fmt.Errorf("%s should not be running", unitName)
+		return unitShouldNotBeRunningErr(unitName)
 	}
 
 	logging.Debugf("Checking if %s has the expected content", unitName)
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -2,6 +2,7 @@ package preflight
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -14,6 +15,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	"github.com/crc-org/crc/v2/pkg/crc/cache"
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
+	"github.com/crc-org/crc/v2/pkg/crc/daemonclient"
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
 	"github.com/crc-org/crc/v2/pkg/crc/machine/libvirt"
 	"github.com/crc-org/crc/v2/pkg/crc/systemd"
@@ -358,7 +360,19 @@ func checkDaemonSystemdService() error {
 	logging.Debug("Checking crc daemon systemd service")
 
 	// the daemon should not be running at the end of setup, as it must be restarted on upgrades
-	return checkSystemdUnit(daemonUnitName, daemonUnitContent(), false)
+	shouldNotBeRunningErr := checkSystemdUnit(daemonUnitName, daemonUnitContent(), false)
+	if shouldNotBeRunningErr == nil {
+		return nil
+	}
+	if !errors.Is(shouldNotBeRunningErr, unitShouldNotBeRunningErr(daemonUnitName)) {
+		return shouldNotBeRunningErr
+	}
+	// daemon is running, check its version
+	version, err := daemonclient.GetVersionFromDaemonAPI()
+	if err != nil {
+		return shouldNotBeRunningErr
+	}
+	return daemonclient.CheckIfOlderVersion(version)
 }
 
 func fixSystemdUnit(unitName string, unitContent string, shouldBeRunning bool) error {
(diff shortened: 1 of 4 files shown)


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -358,7 +360,19 @@ func checkDaemonSystemdService() error {
...
 
 	// the daemon should not be running at the end of setup, as it must be restarted on upgrades
-	return checkSystemdUnit(daemonUnitName, daemonUnitContent(), false)
+	alreadyRunningErr := checkSystemdUnit(daemonUnitName, daemonUnitContent(), false)
+	if alreadyRunningErr == nil {
+		return nil
+	}
+	if !errors.Is(alreadyRunningErr, unitShouldNotBeRunningErr(daemonUnitName)) {
+		return alreadyRunningErr
+	}
```
anjannath, line 369: got confused with the name `alreadyRunningErr` it could be either `unitShouldNotBeRunningErr` or the generic "file content mismatch" error, and in case of the letter its an "`alreadyRunningErr` plus file content different" maybe just name it as `err` but now that i understand it its okay either way or a comment would be helpful!

anjannath (approved):

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: I can rename it, I gave the `alreadyRunningErr` name before adding the `unitStatusErr` type, and did not adjust the name.
I did not name it `err` because when we fail to get the version from the daemon, I prefer to return `alreadyRunningErr` rather than the error which was returned from the `GetVersion` call:
```
        // daemon is running, check its version
	version, err := daemonclient.GetVersionFromDaemonAPI()
	if err != nil {
		return alreadyRunningErr
	}
```

cfergeau: Updated.

PR #3927, opened 2023-11-23
preflight: simplify isDaemonProcess function

anjannath:
instead of checking for expected args at exact position we can simply look for the presence of the daemon  word

this also fixes the following error:
```
INFO Removing older logs
Failed to remove old log files: Failed to delete file: remove C:\Users\crcqe\.crc\crcd.log: The process cannot access the file because it is being used by another process.
```

it was occurring because the psutil/process.CmdlineSlice() returns the args by splitting the args on whitespace so if Path of crc cli has whitespaces assumptions about the position of expected args is not always correct

### DIFF
--- a/pkg/crc/preflight/preflight_checks_nonlinux.go
+++ b/pkg/crc/preflight/preflight_checks_nonlinux.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"fmt"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/crc-org/crc/v2/pkg/crc/daemonclient"
@@ -46,19 +45,14 @@ func killDaemonProcess() error {
 	return nil
 }
 
-// isDaemonProcess return true if any of the following conditions met
-// - crc daemon <options>
-// - crc --log-level=<level> daemon <options>
-// - crc --log-level <level> daemon <options>
+// isDaemonProcess return true if the cmdline contains the word daemon
+// since only one instance of the daemon is allowed to run this should
+// be enough to detect the daemon process
 func isDaemonProcess(cmdLine []string) bool {
-	if len(cmdLine) >= 2 && cmdLine[1] == "daemon" {
-		return true
-	}
-	if len(cmdLine) >= 3 && strings.HasPrefix(cmdLine[1], "--log-level") && cmdLine[2] == "daemon" {
-		return true
-	}
-	if len(cmdLine) >= 4 && cmdLine[1] == "--log-level" && cmdLine[3] == "daemon" {
-		return true
+	for _, arg := range cmdLine {
+		if arg == "daemon" {
+			return true
+		}
 	}
 	return false
 }


pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -46,19 +46,14 @@ func killDaemonProcess() error {
...
-	if len(cmdLine) >= 2 && cmdLine[1] == "daemon" {
-		return true
-	}
-	if len(cmdLine) >= 3 && strings.HasPrefix(cmdLine[1], "--log-level") && cmdLine[2] == "daemon" {
-		return true
-	}
-	if len(cmdLine) >= 4 && cmdLine[1] == "--log-level" && cmdLine[3] == "daemon" {
-		return true
+	for _, arg := range cmdLine {
+		if strings.Contains(strings.TrimSpace(arg), "daemon") {
```
gbraad: `config set skip-check-daemon-task-running true` would therefore be valid here, right?

Note: I understand a config would not run preflights, but it is still something to consider, as `start` and `setup` do run them and might have a flag

anjannath: yes, but very unlikely to end up in that situation as the config command is not a long running command

the main purpose of this `isDaemonProcess` is to prevent the `crc cleanup` command as being detected as the daemon process

anjannath: > config set skip-check-daemon-task-running true would therefore be valid here, right?

for this specific instance, we can handle it by changing the check to `arg == "daemon"`  instead of `strings.Contains` check

cfergeau: Also not a big fan of doing string matching, we do that far too much, and there are always corner cases we have to ignore as "unlikely". In this case it has the benefit of being simple though. The correct way (= without false positives) would be to parse the process commandline using `cobra`, but that's most likely overkill (= a lot of code) for what we need here.
An alternative if we want to keep the current checks would be to drop the args at the beginning of the slice until we find one which ends with `crcProcessName`, but that's probably also overkill.

> if Path of crc cli has whitespaces assumptions about the position of expected args is not always correct

One question though, was this code always broken on Windows? If I understand correctly, you are saying that `C:\Program Files\OpenShift Local\crc.exe` gets split as `C:\Program`, `Files\OpenShift`, `Local\crc.exe`?

anjannath: > One question though, was this code always broken on Windows? If I understand correctly, you are saying that C:\Program Files\OpenShift Local\crc.exe gets split as C:\Program, Files\OpenShift, Local\crc.exe?

yes, this was always broken on windows, when we tested it previously we probably didn't test from a location with spaces in the path name

cfergeau: But the default location will always have spaces in its name, won't it?

anjannath: yes it'll always have spaces ` C:\Program Files\Red Hat OpenShift Local\`

cfergeau: Should be fine to do what you suggested (ideally without `TrimSpace`):

> for this specific instance, we can handle it by changing the check to arg == "daemon" instead of strings.Contains check

anjannath: Updated now!

gbraad (approved):

Issue #3931, opened 2023-11-28
[QE] Build and publish crc-e2e and crc-integration as a PR check


jsliacan:
Follow-up to #3875

adrianriobo:
Can you create 2 different flows for that? I mean do not build the images within the testing flow but as separate flow which build the images (installer alike) and as so it willl be used as check to ensure nothing changed lets say on the Dockerfile would break nothing

jsliacan:
Completed via #3933

PR #3933, opened 2023-11-29
Build e2e and integration test images as part of PR checks

jsliacan:
**Fixes:** Issue #3931 

Introduce a GH workflow with 2 jobs (run in parallel by default). 1 job builds e2e image and the other builds integration image. Both images are tagged with `id-${{github.sha}}`, archived as `.tar` files and uploaded as artifacts (instead of pushing to external registry). The purpose of these images is tied to the lifetime of a PR anyways.

These artifacts will be consumed by another workflow (windows-e2e/integration) which will use them to run e2e and integration PR checks on windows environment. 

**Note**
Using `${{github.sha}}` as the correlation number should work for both `pull_request` and `push to main` trigger events. For this, I have 2 `github` context dumps for each case (PR, push). Two workflows (test1 and test2) were triggered by the same event (PR or push to main) and context dump was saved. Comparing `github.sha` values suggests that they are the same for these workflows (as they should be). This generalizes the previous approach where we used PR# to correlate workflows. Moreover, the `sha` approach comes with another advantage. The `sha` changes from trigger to trigger, making sure that if we download an artifact with the correct `sha`, it will be the intended one and not one from a previous trigger. This would not be guaranteed when using PR# as id, because that does not change from trigger to trigger.

- [on pull 1](https://gist.github.com/jsliacan/e77c340177159e6cbbe311fd67cf1c23)
- [on pull 2](https://gist.github.com/jsliacan/75976f4c1ccc93567cae4858f8f26a74)
- [on push 1](https://gist.github.com/jsliacan/f2416e15dd0d87de7ec1ca16599cec7e)
- [on push 2](https://gist.github.com/jsliacan/c86bdf8622a4632b5bb2d559cfca8f01)

[gha-win-e2e.pdf](https://github.com/crc-org/crc/files/13583760/gha-win-e2e.pdf)

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -1,0 +1,53 @@
+name: Build test images
+on:
+  push:
+    branches:
+      - "main"
+  pull_request: {}
+
+env:
+  IMAGE_NAME_E2E: crc-e2e
+  IMAGE_NAME_INTEGRATION: crc-integration
+
+jobs:
+  build-e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      
+      - name: Build and archive e2e image
+        # use github.sha as ID to correlate various workflows triggered by the same event
+        run: |
+          CRC_E2E_IMG_VERSION=id-${{ github.sha }} make containerized_e2e
+          podman save -o ${{ env.IMAGE_NAME_E2E }}.tar quay.io/crcont/${{ env.IMAGE_NAME_E2E}}:id-${{ github.sha }}
+
+      - name: Upload e2e image
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.IMAGE_NAME_E2E }}-id${{ github.sha }}
+          path: ${{ env.IMAGE_NAME_E2E }}.tar
+
+  build-integration:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      
+      - name: Build and archive integration image
+        # use github.sha as ID to correlate various workflows triggered by the same event
+        run: |
+          CRC_INTEGRATION_IMG_VERSION=id-${{ github.sha }} make containerized_integration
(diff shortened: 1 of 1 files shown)


cfergeau:
Good to have CI for these container images as sometimes we use wrong base images in these, which cause build failures we only notice fairly late.

.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,80 @@
...
+  build-e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+
+      - name: Login to GitHub Container Registry
```
adrianriobo: why are you login?

jsliacan: Because I missed it when changing the flow. Mistake, removed. Thanks!

.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Archive e2e image
+        env:
+            PR_NUMBER: ${{ github.event.number }}
+        run: |
+          podman save -o ${{ env.IMAGE_NAME_E2E }}.tar quay.io/crcont/${{ env.IMAGE_NAME_E2E}}:pr-${PR_NUMBER}
+
+      - name: Upload e2e artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.IMAGE_NAME_E2E }}-pr${{ github.event.number }}
```
adrianriobo: Why some steps use the  ${{ github.event.number }} directly and others use then env section?

.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+        env:
+            PR_NUMBER: ${{ github.event.number }}
+        run: |
+          CRC_E2E_IMG_VERSION=pr-${PR_NUMBER} make containerized_e2e
+      
+      - name: Archive e2e image
+        env:
+            PR_NUMBER: ${{ github.event.number }}
+        run: |
+          podman save -o ${{ env.IMAGE_NAME_E2E }}.tar quay.io/crcont/${{ env.IMAGE_NAME_E2E}}:pr-${PR_NUMBER}
```
adrianriobo: Also may build and save can be added together on a single step? it will make things a bit easier to read IMHO

.github/workflows/build-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: Upload e2e artifact
```
adrianriobo: Would change to Upload e2e image (to match previous naming convention for steps)

adrianriobo (approved):

PR #3941, opened 2023-12-06
Update go modules dependencies

cfergeau:
This sets openshift version to 4.14 and k8s to the corresponding version.
Our other dependencies are updated to their latest version.


gbraad (approved):
/approve

PR #3950, opened 2023-12-19
dns: Don't use dnsmasq service as container

praveenkumar:
This PR is going to use systemd dnsmasq service instead running it as part of container and then consuming it. It should work with current bundles and also updated bundle which doesn't have dnsmasq container cached.

We are doing it because in future we want to use OVN-Kubernetes as network plugin for OCP/OKD and with our current solution it is not able to resolve the IP of the dnsmasq container so everything around dns is broken which this PR fixes.

```
=== using openshift-sdn ===
$ oc rsh busybox-sleep-pod
sh-5.1# ping 10.88.0.8
PING 10.88.0.8 (10.88.0.8) 56(84) bytes of data.
64 bytes from 10.88.0.8: icmp_seq=1 ttl=63 time=0.878 ms
64 bytes from 10.88.0.8: icmp_seq=2 ttl=63 time=0.068 ms

=== using ovn-k ===
sh-5.1# ping 10.88.0.8
PING 10.88.0.8 (10.88.0.8) 56(84) bytes of data.
^C
--- 10.88.0.8 ping statistics ---
15 packets transmitted, 0 received, 100% packet loss, time 14368ms
```


praveenkumar:
/hold 

Need to test this with https://github.com/crc-org/snc/pull/829 one.

pkg/crc/services/dns/dns.go
```diff
@@ -20,9 +20,8 @@ import (
 
 const (
 	dnsServicePort    = 53
-	dnsContainerIP    = "10.88.0.8"
 	publicDNSQueryURI = "quay.io"
-	crcDnsmasqService = "crc-dnsmasq.service"
+	dnsmasqService    = "dnsmasq.service"
```
cfergeau, line 24: This service is already installed/available in the VM?

pkg/crc/services/dns/template.go
```diff
@@ -50,7 +47,7 @@ func createDnsmasqDNSConfig(serviceConfig services.ServicePostStartConfig) error
 		return err
 	}
 
-	return serviceConfig.SSHRunner.CopyDataPrivileged([]byte(dnsConfig), "/var/srv/dnsmasq.conf", 0644)
+	return serviceConfig.SSHRunner.CopyDataPrivileged([]byte(dnsConfig), "/etc/dnsmasq.d/crc-dnsmasq.conf", 0644)
```
cfergeau, line 50: MCO won't complain about this file being created behind its back?

pkg/crc/services/dns/template.go
```diff
@@ -8,9 +8,7 @@ import (
 )
 
 const (
-	dnsmasqConfTemplate = `user=root
-port= {{ .Port }}
-bind-interfaces
+	dnsmasqConfTemplate = `listen-address={{ .IP }}
```
cfergeau, line 11: Could it be localhost instead of the external IP?

praveenkumar: It can't, because this goes to network pod and the localhost is not resolved so we need the external IP.

pkg/crc/services/dns/dns.go
praveenkumar: yes it is already installed and available in the VM

pkg/crc/services/dns/template.go
praveenkumar: MCO is not tracking this file so it is not complaining,

adrianriobo:
QE Pass

* [microshift mac](https://crcqe-asia.s3.amazonaws.com/nightly/microshift/4.14.8/20240118/13.0.1/qe-results/e2e-non-ux.xml)
* [microshift windows](https://crcqe-asia.s3.amazonaws.com/nightly/microshift/4.14.8/20240118/11-pro/qe-results/e2e-non-ux.xml)
* [ocp mac](https://crcqe-asia.s3.amazonaws.com/nightly/crc/20240118/13.0.1/qe-results/e2e-non-ux.xml)
* [ocp windows](https://crcqe-asia.s3.amazonaws.com/nightly/crc/20240118/11-pro/qe-results/e2e-non-ux.xml)

pkg/crc/services/dns/template.go
cfergeau: "need" the external IP is probably a bit strong, there are most likely some 10.x address we can use that would be reachable by all pods but not exposed externally. (not asking for changes, just nuancing your answer)

cfergeau: Dunno if listen-address is mandatory?

cfergeau (approved):

pkg/crc/services/dns/template.go
praveenkumar: > Dunno if listen-address is mandatory?

If we not specify then it will also listen to loopback interface.

```-a, --listen-address=<ipaddr>
              Listen on the given IP address(es). Both --interface and --listen-address options may be given, in which case the set of both interfaces and addresses is used. Note that if no --interface option is  given,  but  --listen-address  is,
              dnsmasq will not automatically listen on the loopback interface. To achieve this, its IP address, 127.0.0.1, must be explicitly given as a --listen-address option.
```

praveenkumar: > there are most likely some 10.x address we can use that would be reachable by all pods but not exposed externally.

May be we can attach to a dummy interface but didn't try it since it is only available for openshift bundle right now and not with microshift one.

Issue #3936, opened 2023-12-04
The command `crc status -w` on Windows writes out escape codes


jeffmaury:
### General information

  * OS: Windows
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop

## CRC version
```bash
CRC version: 2.28.0+e1f42d
OpenShift version: 4.13.14
Podman version: 4.4.4
```
  
## CRC status
```bash
DEBU CRC version: 2.28.0+e1f42d
DEBU OpenShift version: 4.13.14
DEBU Podman version: 4.4.4
DEBU Running 'crc status'
CRC VM:          Running
OpenShift:       Starting (v4.13.14)
RAM Usage:       5.859GB of 9.361GB
Disk Usage:      19.41GB of 32.68GB (Inside the CRC VM)
Cache Usage:     29.51GB
Cache Directory: C:\Users\Jeff\.crc\cache
```

## CRC config
```bash
- consent-telemetry                     : yes
```

## Host Operating System
```bash

Nom de l'h鍍e:                              DESKTOP-JEFF
Nom du syst確e d'exploitation:              Microsoft Windows 11 Professionnel
Version du syst確e:                         10.0.22621 N/A build 22621
Fabricant du syst確e d'exploitation:        Microsoft Corporation
Configuration du syst確e d'exploitation:    Station de travail autonome
Type de build du syst確e d'exploitation:    Multiprocessor Free
PropriＵaire enregistr・                    Jeff MAURY
Organisation enregistrＦ:                   
Identificateur de produit:                  00330-80819-37372-AA576
Date d'installation originale:              09/11/2022, 16:39:06
Heure de dＮarrage du syst確e:              04/12/2023, 15:54:02
Fabricant du syst確e:                       LENOVO
Mod獲e du syst確e:                          20TJS2F43N
Type du syst確e:                            x64-based PC
Processeur(s):                              1 processeur(s) install・s).
                                            [01]: Intel64 Family 6 Model 165 Stepping 2 GenuineIntel ~2310 MHz
Version du BIOS:                            LENOVO N2VET29W (1.14 ), 15/03/2021
RＱertoire Windows:                         C:\Windows
RＱertoire syst確e:                         C:\Windows\system32
PＳiphＳique d'amor㌢ge:                    \Device\HarddiskVolume1
Option rＨionale du syst確e:                fr;Fran㌢is (France)
[...]

PR #3955, opened 2024-01-03
Turn off terminal raw mode on Windows after started progressbar pool

vyasgun:
**Fixes:** Issue #3936 
**Relates to:** Issue #3936 

## Solution/Idea

The idea is that the printing changed from unformatted to formatted after a `ctrl-c` was sent while the print was happening. Issue only seems to happen when a `pb.Pool` type is used to start the bars only when color directives are used in the template. Defer stopping the pool and turning off the raw mode fixes the issue on Windows.

## Proposed changes

1. Turn off terminal raw mode on Windows after started progressbar pool

Opened an issue with the library highlighting this: https://github.com/cheggaaa/pb/issues/213

## Testing

1. The output on Windows Powershell should be consistent with other operating systems.

### DIFF
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -7,9 +7,11 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"text/tabwriter"
 
 	"github.com/cheggaaa/pb/v3"
+	"github.com/cheggaaa/pb/v3/termutil"
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
 	"github.com/crc-org/crc/v2/pkg/crc/daemonclient"
 	crcErrors "github.com/crc-org/crc/v2/pkg/crc/errors"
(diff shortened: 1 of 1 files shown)


cmd/crc/cmd/status.go
```diff
@@ -79,25 +79,41 @@ func runWatchStatus(writer io.Writer, client *daemonclient.Client, cacheDir stri
...
+				return
+			}
 			for i := 0; i < newCPUCount; i++ {
 				bar := createCPUBar(oldCPUCount+i, writer)
 				barPull.Add(bar)
 				cpuBars = append(cpuBars, bar)
 			}
+			if startErr := barPull.Start(); startErr != nil {
+				return
+			}
```
cfergeau, line 116: Ah, it looks like the problem is calling `barPull.Add()` to add a new bar while the pool is started?

cmd/crc/cmd/status.go
```diff
@@ -79,25 +79,41 @@ func runWatchStatus(writer io.Writer, client *daemonclient.Client, cacheDir stri
 		ramBar  *pb.ProgressBar
 	)
 
+	var err error
+	barPull = pb.NewPool()
```
cfergeau: Can you use this opportunity to rename `barPull` to `barPool`? I think this is a typo.

cmd/crc/cmd/status.go
```diff
@@ -79,25 +79,41 @@ func runWatchStatus(writer io.Writer, client *daemonclient.Client, cacheDir stri
...
 	isPullInit := false
-
-	err := client.SSEClient.Status(func(loadResult *types.ClusterLoadResult) {
+	err = barPull.Start()
```
cfergeau: Do you need to start the pool right away? Can we wait we get to `if !isPullInit()` before calling `Start()` as it was done before?

evidolob: I made some test for that and we need to have `barPull.Start()`.
As simplification we could use `pb.StartPool()` instead of `pb.NewPool()` to remove that line

evidolob (approved):
On windows it is working well:
pb

praveenkumar:
@vyasgun Are you planning to change this draft to PR with requested suggestions?

vyasgun:
@praveenkumar yes, I am trying a couple things and will change it to PR soon.

vyasgun:
Thank you for the reviews @evidolob @cfergeau! I have updated the PR with a simpler workaround (Turn off the `raw mode` on terminal).  Screenshot of the terminal after the changes:

Screenshot 2024-01-17 214059

cmd/crc/cmd/status.go
```diff
@@ -74,28 +77,37 @@ func runWatchStatus(writer io.Writer, client *daemonclient.Client, cacheDir stri
 	}
 
 	var (
-		barPull *pb.Pool
+		barPool *pb.Pool
 		cpuBars []*pb.ProgressBar
 		ramBar  *pb.ProgressBar
 	)
 
-	isPullInit := false
```
gbraad, line 82: so this was a typo?

vyasgun: yes, since it is used to check if the pool has been initialised, isPoolInit seems more apt.

praveenkumar:
CI is failing with lint
```
 cmd/crc/cmd/status.go:102:24: Error return value of `termutil.RawModeOff` is not checked (errcheck)
				termutil.RawModeOff()
				                   ^ 
```

cmd/crc/cmd/status.go
```diff
@@ -114,7 +128,7 @@ func createBars(cpuUse []int64, writer io.Writer) (ramBar *pb.ProgressBar, cpuBa
 	ramBar.SetWriter(writer)
 	ramBar.Set(pb.Bytes, true)
 	ramBar.Set(pb.Static, true)
-	tmpl := `{{ red "RAM:" }} {{counters . }} {{percent .}} {{string . "my_green_string" | green}} {{ bar . "[" "\u2588" "\u2588" " " "]"}} `
+	tmpl := `{{ red "RAM:" }} {{counters . }} {{percent .}} {{ bar . "[" "\u2588" "\u2588" " " "]"}} `
```
praveenkumar, line 131: This removal of `{{string . "my_green_string" | green}}` is because of the issue or we don't need it anymore?

vyasgun: Removed it as it is a typo and not required for printing.

PR #3965, opened 2024-01-19
Add preflight check for vsock SSH port

evidolob:
**Relates to:**  First part for https://github.com/crc-org/crc/issues/3855

## Solution/Idea

PR's add preflight check for SSH port(`2222`) in not used before start. 

## Testing

1. Occupy `2222` port on your laptop( you may use `nc -lkv 127.0.0.1 2222` command)
2. Start daemon
3. Try to start CRC VM
4. Ensure that you have error during preflight checks
5. Free port `2222`
6. Try to start CRC again
7. Ensure that CRC starts normally

### DIFF
--- a/pkg/crc/preflight/preflight_checks_nonlinux.go
+++ b/pkg/crc/preflight/preflight_checks_nonlinux.go
@@ -6,9 +6,12 @@ package preflight
 import (
 	"context"
 	"fmt"
+	"net"
 	"runtime"
+	"strconv"
 	"time"
 
+	"github.com/crc-org/crc/v2/pkg/crc/constants"
 	"github.com/crc-org/crc/v2/pkg/crc/daemonclient"
 	crcerrors "github.com/crc-org/crc/v2/pkg/crc/errors"
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
@@ -71,4 +74,42 @@ func waitForDaemonRunning() error {
 		}
 		return nil
 	}, 2*time.Second)
+}
+
+func sshPortCheck() Check {
+	return Check{
+		configKeySuffix:  "check-ssh-port",
+		checkDescription: "Checking SSH port availability",
+		check:            checkSSHPortFree(),
+		fixDescription:   fmt.Sprintf("crc uses %d to run SSH", constants.VsockSSHPort),
+		flags:            NoFix,
+		labels:           None,
+	}
+}
+
+func checkSSHPortFree() func() error {
+	return func() error {
+
+		host := net.JoinHostPort(constants.LocalIP, strconv.Itoa(constants.VsockSSHPort))
+
+		daemonClient := daemonclient.New()
+		exposed, err := daemonClient.NetworkClient.List()
+		if err == nil {
+			// if port already exported by vsock we could proceed
+			for _, e := range exposed {
+				if e.Local == host {
+					return nil
+				}
+			}
+		}
+
+		server, err := net.Listen("tcp", host)
+		// if it fails then the port is likely taken
+		if err != nil {
+			return fmt.Errorf("port %d already in use: %s", constants.VsockSSHPort, err)
+		}
+
+		// we successfully used and closed the port
+		return server.Close()
+	}
 }
--- a/pkg/crc/preflight/preflight_checks_nonlinux.go
+++ b/pkg/crc/preflight/preflight_checks_nonlinux.go
@@ -81,7 +81,7 @@ func sshPortCheck() Check {
 		configKeySuffix:  "check-ssh-port",
 		checkDescription: "Checking SSH port availability",
 		check:            checkSSHPortFree(),
-		fixDescription:   fmt.Sprintf("crc uses %d to run SSH", constants.VsockSSHPort),
+		fixDescription:   fmt.Sprintf("crc uses port %d to run SSH", constants.VsockSSHPort),
 		flags:            NoFix,
 		labels:           None,
 	}
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -38,6 +38,7 @@ const (
 
 	VSockGateway = "192.168.127.1"
 	VsockSSHPort = 2222
+	LocalIP      = "127.0.0.1"
 
 	OkdPullSecret = `{"auths":{"fake":{"auth": "Zm9vOmJhcgo="}}}` // #nosec G101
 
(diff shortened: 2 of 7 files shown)


anjannath:
working as expected, tested on macOS:
```
INFO Using bundle path /Users/anath/.crc/cache/crc_vfkit_4.14.8_arm64.crcbundle
port 2222 already in use: listen tcp 127.0.0.1:2222: bind: address already in use
```

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -72,3 +75,42 @@ func waitForDaemonRunning() error {
 		return nil
 	}, 2*time.Second)
 }
+
+func sshPortCheck() Check {
+	return Check{
+		configKeySuffix:  "check-ssh-port",
+		checkDescription: "Checking SSH port availability",
+		check:            checkSSHPortFree(),
+		fixDescription:   fmt.Sprintf("crc uses %d to run SSH", constants.VsockSSHPort),
```
vyasgun: small suggestion:

```suggestion
		fixDescription:   fmt.Sprintf("crc uses port %d to run SSH", constants.VsockSSHPort),
```

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -72,3 +75,42 @@ func waitForDaemonRunning() error {
...
+				}
+			}
+		}
+
+		server, err := net.Listen("tcp", host)
+		// if it fails then the port is likely taken
+		if err != nil {
+			return fmt.Errorf("port %d already in use: %s", constants.VsockSSHPort, err)
+		}
+		server.Close()
```
vyasgun:
```suggestion
		defer func() {
			err = server.Close()
		}()
		
```

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -72,3 +75,42 @@ func waitForDaemonRunning() error {
...
+
+		// we successfully used and closed the port
+		return nil
```
vyasgun:
```suggestion
		return err
```

vyasgun (commented):
Just added a few minor comments

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -72,3 +75,42 @@ func waitForDaemonRunning() error {
...
+		checkDescription: "Checking SSH port availability",
+		check:            checkSSHPortFree(),
+		fixDescription:   fmt.Sprintf("crc uses %d to run SSH", constants.VsockSSHPort),
+		flags:            NoFix,
+		labels:           None,
+	}
+}
+
+func checkSSHPortFree() func() error {
+	return func() error {
```
cfergeau, line 91: Do you need this indirection? the preflights use the pattern `func foo() func() error` when we need to pass arguments to `foo`. In this case, I think you could directly use `func checkSSHPortFree() error`, and change `check:            checkSSHPortFree(),` to `check:            checkSSHPortFree,` ?

pkg/crc/preflight/preflight_checks_nonlinux.go
cfergeau: If the intent is to return the error from the deferred `server.Close()` call, I don't think this will work as expected:
https://go.dev/play/p/HyKboZE2dka

pkg/crc/preflight/preflight_checks_nonlinux.go
cfergeau: Could be just `return server.Close()` ?

pkg/crc/preflight/preflight_checks_nonlinux.go
cfergeau: Thinking more about it, it actually would make sense to only run the preset when usermode networking is use (the `network-mode` config option). In which case you'd need additional arg to the check, and using this pattern would make sense :)

pkg/crc/preflight/preflight_checks_nonlinux.go
vyasgun: You are right. We can define the err variable as part of the function signature to set the correct return value in defer.
https://go.dev/play/p/8kaB3HQlfk6

pkg/crc/preflight/preflight_checks_nonlinux.go
vyasgun: I think it's more conventional to use defer for closing connections.

pkg/crc/preflight/preflight_checks_nonlinux.go
```diff
@@ -72,3 +75,41 @@ func waitForDaemonRunning() error {
...
+					return nil
+				}
+			}
+		}
+
+		server, err := net.Listen("tcp", host)
+		// if it fails then the port is likely taken
+		if err != nil {
+			return fmt.Errorf("port %d already in use: %s", constants.VsockSSHPort, err)
+		}
```
praveenkumar, line 110: @evidolob Doesn't that mean if there is error and returned but server is also not closed? Can we try following
```golang
func checkSSHPortFree() func() (err error) {
	return func() error {

		host := net.JoinHostPort(constants.LocalIP, strconv.Itoa(constants.VsockSSHPort))

		daemonClient := daemonclient.New()
		exposed, err := daemonClient.NetworkClient.List()
		if err == nil {
			// if port already exported by vsock we could proceed
			for _, e := range exposed {
				if e.Local == host {
					return nil
				}
			}
		}

		server, err := net.Listen("tcp", host)
		// if it fails then the port is likely taken
		if err != nil {
			return fmt.Errorf("port %d already in use: %s", constants.VsockSSHPort, err)
		}

        defer func() {
            err = server.Close()
        }
}
```

evidolob: No, if error returned(port already in use) `server` will be `nil`, so that code could lead to another bug.

praveenkumar: In that case `defer` might be
```golang
defer func() {
  if server != nil {
    err = server.Close()
 }
}
```

pkg/crc/preflight/preflight_checks_nonlinux.go
cfergeau: I don't know why this was closed as resolved, but this was neither discussed nor resolved.

cfergeau: @praveenkumar

PR #3968, opened 2024-01-20
Fix docs-related make targets so they run on MacOS

fxierh:
Calling `podman run -v` with the `:Z` SELinux volume label is currently not supported on MacOS. 

Avoid adding `:Z` when `$(GOOS)` is `darwin`.


fxierh:
/assign @gbraad @praveenkumar @cfergeau 

Would you please take a look when you get a chance ? Thanks ^^

praveenkumar:
Tested on mac with podman machine and it works as expected.

/lgtm
/approve

PR #3974, opened 2024-01-24
gha: unify windows builds into one workflow and drop sha ID correlating

jsliacan:
Addresses Issue #3975 

The solution with many side benefits (see commit log) is to unify what was formerly two workflows `build-tests.yaml` and `windows-installer.yaml` into `windows-artifacts.yaml`. This workflow then triggers `windows-e2e.yaml` via `workflow_run` and therefore, `windows-e2e.yaml` has access to all artifacts from `windows-artifacts.yaml` directly - no need to correlate via commit sha or similar. 

As a side effect, `windows-e2e.yaml` only gets triggered *after* all artifacts are produced, so we can rearrange steps inside `windows-e2e.yaml` in a way that postpones target host provisioning until after we know we have everything else ready (no provisioning if some precondition fails). 

Finally, parametrize commit status context by matrix entries, to futureproof.

A clean run (with only a `config.feature` run) can be seen checked as part of this dummy PR https://github.com/jsliacan/crc/pull/9. Here's the actual action: https://github.com/jsliacan/crc/actions/runs/7668762265


.github/workflows/build-tests.yml
```diff
@@ -19,16 +19,25 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v3
       
+      - name: Work out correct id
+        run: |
+          COMMIT_ID=${{ toJSON(github) }} | jq -r '.event.after')
```
anjannath: an opening `(` is missing here

jsliacan: Yes, thanks :)

adrianriobo (approved):

Issue #3852, opened 2023-09-29
[enhancement] check dns pod state as part of crc status operation


adrianriobo:
This seems to be the root cause for #3851 

When we initialize a `microshift` cluster we ensure the state of the cluster as `running` then we deploy and expose a service to test the connectivity, when we do it manually it works as expected, but when we run that through an automation it fails with:

```
time="2023-09-28T13:44:03Z" level=error msg="Post \"http://gateway/hosts/add\": dial tcp: lookup gateway on 10.43.0.10:53: read udp 10.42.0.6:35808->10.43.0.10:53: read: connection refused"
```

The code for `crc status` on microshift checks `oc get node` but does not take care about the state of the pods running within the cluster.

On this scenario we expected dns is working fine as `crc status` is `running` but it is not the case, so it may worthwhile to check the state of the dns to ensure the `running` state of the cluster.

adrianriobo:
After adding a check for dns on the `e2e` now route is always added as it is expected, but we end up having issues with the test service deployed, ending on `CreateContainerError` due to:

```
Warning  FailedCreatePodSandBox  34m                    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create pod network sandbox k8s_httpd-example-6bf9c787d7-d4mnd_testproj_f468e5fe-b4e9-45b6-b6f6-856426565406_0(6e1a579fdb65e06338b7e1e68f09866369562a4888142d363fb377c770b9bdf2): error adding pod testproj_httpd-example-6bf9c787d7-d4mnd to CNI network "ovn-kubernetes": plugin type="ovn-k8s-cni-overlay" name="ovn-kubernetes" failed (add): failed to send CNI request: Post "http://dummy/": dial unix /var/run/ovn-kubernetes/cni//ovn-cni-server.sock: connect: no such file or directory
```

So probably we will need to check the state for `ovn` as well

gbraad:
@adrianriobo is this still relevant?

adrianriobo:
It is relevant from a point of view of automation around it, on `e2e` we are trying to emulate a user experience and with the delays in between operations this is unlikely to happen (not having a healthy state cause we do not check).

But if we explore the CI use case and it would imply automation this should be in place.

Also as we do for `OCP` I guess it would be good to have the check #4009  at the `start` to ensure what we are delivering is something functional.

PR #4009, opened 2024-02-01
Check status of all the core pods for microshift

praveenkumar:
In past we observed having kube api access doesn't mean all the required service pods are running and cluster is working as expected. This PR adds a list of core namespace for microshift preset and make sure all the pods in that namespace is running before letting user to know to consume the cluster.

**Fixes:** Issue #3852

### DIFF
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -510,3 +510,32 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
 	_, _, err := ocConfig.RunOcCommand("delete", "-A", "lease", "--all")
 	return err
 }
+
+func CheckCorePodsRunning(ctx context.Context, ocConfig oc.Config) error {
+	if err := WaitForOpenshiftResource(ctx, ocConfig, "pod"); err != nil {
+		return err
+	}
+	coreNameSpaces := []string{"kube-system", "openshift-dns", "openshift-ingress", "openshift-ovn-kubernetes", "openshift-service-ca"}
+	waitForPods := func() error {
+		for _, namespace := range coreNameSpaces {
+			if !podRunningForNamespace(ocConfig, namespace) {
+				logging.Debugf("Pods in %s namespace are not running", namespace)
+				return &errors.RetriableError{Err: fmt.Errorf("pods in %s namespace are not running", namespace)}
+			}
+		}
+		return nil
+	}
+	return errors.Retry(ctx, 2*time.Minute, waitForPods, 2*time.Second)
+}
+
+func podRunningForNamespace(ocConfig oc.Config, namespace string) bool {
+	stdout, stderr, err := ocConfig.WithFailFast().RunOcCommand("get", "pods", "-n", namespace, "--field-selector=status.phase!=Running")
+	if err != nil {
+		logging.Debugf("Failed to get pods in %s namespace, stderr: %s", namespace, stderr)
+		return false
+	}
+	if len(stdout) != 0 {
+		return false
+	}
+	return true
+}
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -1064,7 +1064,10 @@ func startMicroshift(ctx context.Context, sshRunner *crcssh.Runner, ocConfig oc.
 		return err
 	}
 
-	return cluster.WaitForAPIServer(ctx, ocConfig)
+	if err := cluster.WaitForAPIServer(ctx, ocConfig); err != nil {
+		return err
+	}
+	return cluster.CheckCorePodsRunning(ctx, ocConfig)
 }
 
 func ensurePullSecretPresentInVM(sshRunner *crcssh.Runner, pullSec cluster.PullSecretLoader) error {


pkg/crc/cluster/cluster.go
```diff
@@ -510,3 +510,29 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
...
+	return errors.Retry(ctx, 2*time.Minute, waitForPods, 2*time.Second)
+}
+
+func podRunningForNamespace(ocConfig oc.Config, namespace string) bool {
+	stdin, stderr, err := ocConfig.WithFailFast().RunOcCommand("get", "pods", "-n", namespace, "--field-selector=status.phase!=Running")
+	if err != nil {
+		logging.Debugf("Failed to get pods in %s namespace, stderr: %s", namespace, stderr)
+		return false
+	}
+	if len(stdin) != 0 {
```
cfergeau: This is `stdout`, not `stdin`

pkg/crc/cluster/cluster.go
```diff
@@ -510,3 +510,29 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
...
+				return &errors.RetriableError{Err: fmt.Errorf("pods in %s namespace are not running", namespace)}
+			}
+		}
+		return nil
+	}
+	return errors.Retry(ctx, 2*time.Minute, waitForPods, 2*time.Second)
+}
+
+func podRunningForNamespace(ocConfig oc.Config, namespace string) bool {
+	stdin, stderr, err := ocConfig.WithFailFast().RunOcCommand("get", "pods", "-n", namespace, "--field-selector=status.phase!=Running")
```
cfergeau: Why `!=Running`? I would have expected `=Running`?

pkg/crc/cluster/cluster.go
```diff
@@ -510,3 +510,29 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
 	_, _, err := ocConfig.RunOcCommand("delete", "-A", "lease", "--all")
 	return err
 }
+
+func CheckCorePodsRunning(ctx context.Context, ocConfig oc.Config) error {
+	coreNameSpace := []string{"kube-system", "openshift-dns", "openshift-ingress", "openshift-ovn-kubernetes", "openshift-service-ca"}
```
cfergeau: `coreNamespaces`

pkg/crc/cluster/cluster.go
```diff
@@ -510,3 +510,29 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
...
+	waitForPods := func() error {
+		for _, namespace := range coreNameSpace {
+			if !podRunningForNamespace(ocConfig, namespace) {
+				logging.Debugf("Pods in %s namespace are not running", namespace)
+				return &errors.RetriableError{Err: fmt.Errorf("pods in %s namespace are not running", namespace)}
+			}
```
cfergeau, line 524: Fwiw, this is a bit wasteful as we'll try again and again the same namespaces even if we already found running pods. Maybe this can be done with a map? map keys are namespaces, iterate over the keys. When there are running pods in the namespace, remove it from the map?

pkg/crc/cluster/cluster.go
```diff
@@ -510,3 +510,29 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
 	_, _, err := ocConfig.RunOcCommand("delete", "-A", "lease", "--all")
 	return err
 }
+
+func CheckCorePodsRunning(ctx context.Context, ocConfig oc.Config) error {
```
cfergeau, line 514: Maybe this could start with ` if err := WaitForOpenshiftResource(ctx, ocConfig, "pod"); err != nil {` ?

pkg/crc/cluster/cluster.go
praveenkumar: @cfergeau If I use `==Running` that means it will show all the running pods, what we want the pods which are not in Running phase in that specific namespace so we can reiterate in retry function.
```
$ kubectl get pod -n kube-system --field-selector=status.phase!=Running 
NAME                                       READY   STATUS    RESTARTS   AGE
csi-snapshot-controller-85cc4fd76b-xznzw   1/1     Pending   0          45h
```

pkg/crc/cluster/cluster.go
praveenkumar: Yes, but it has bit of benefit in case some pod goes to reconciliation state (like in one iteration it is running but in second it is in pending state.) It is not full proof solution (with k8s context it is never going to be) but should be good for initial feedback if core pods are running.

pkg/crc/cluster/cluster.go
```diff
@@ -510,3 +510,32 @@ func DeleteMCOLeaderLease(ctx context.Context, ocConfig oc.Config) error {
...
+				logging.Debugf("Pods in %s namespace are not running", namespace)
+				return &errors.RetriableError{Err: fmt.Errorf("pods in %s namespace are not running", namespace)}
+			}
+		}
+		return nil
+	}
+	return errors.Retry(ctx, 2*time.Minute, waitForPods, 2*time.Second)
+}
+
+func podRunningForNamespace(ocConfig oc.Config, namespace string) bool {
```
cfergeau, line 531: `allPodsRunning(ocConfig oc.Config, namespace string) bool` or `checkAllPodsRunning` is more descriptive/accurate.

pkg/crc/cluster/cluster.go
cfergeau: In the OpenShift case, once the `oc get co` check succeeds once, we retry 2 more times and we only decide the cluster is good when the `oc get co` check succeeds 3 times in a row. If you want to handle " in one iteration it is running but in second it is in pending state" it would be nice to have a consistent approach.

pkg/crc/cluster/cluster.go
praveenkumar: it should be in namespace context so `checkAllPodsRunningInNamespace` or `allPodsRunningForNamespace` ?

pkg/crc/cluster/cluster.go
praveenkumar: > In the OpenShift case, once the `oc get co` check succeeds once, we retry 2 more times and we only decide the cluster is good when the `oc get co` check succeeds 3 times in a row. If you want to handle " in one iteration it is running but in second it is in pending state" it would be nice to have a consistent approach.

yes in case of openshift we can iterate over all the clusteroperator at once because those are not namespace specific resource. Here we are not able to have a single call which provide use all the pods status in core namespaces otherwise I would've use same logic. So now we iterate over namespace by namespace and check the pods status.

- https://github.com/kubernetes/kubernetes/issues/52326

cfergeau: I'm not questioning the way the iterations are done, I was reacting to
>  it has bit of benefit in case some pod goes to reconciliation state (like in one iteration it is running but in second it is in pending state.) 

For an OpenShift cluster, we roughly do iterate over a `isClusterReady()` function until it returns true. Once it returns true, we still run it 2 times in case the cluster was ready, but in a transient/conciliation state.
If reconciliation is something you want to try to handle better, I would use the same approach as for OpenShift for consistency, cluster is not ready before `isClusterReady()` succeeded 3 times in a row.

pkg/crc/cluster/cluster.go
cfergeau: There is a `namespace` argument, we don't have non-namespace function this could be confused with, so I don't think it's really useful to mention `Namespace` in the function name. It's more something for an api doc comment if you think it's important to inform API users that it will only iterate over a single namespace.

gbraad:
> This PR adds a list of core namespace for microshift preset

This is not microshift specific, as it might also be able to solve issues with the readiness of the OCP and OKD preset.

gbraad:
> [has-to-be-in-release](https://github.com/crc-org/crc/labels/has-to-be-in-release)

What was the reasoning behind this label, as the fixed issue: https://github.com/crc-org/crc/issues/3852 is merely an enhancement.

Issue #4018, opened 2024-02-06
[QE] Add retry when checking monitoring stack 


adrianriobo:
With the latest OCP versions we are seeing some delay for pods from monitoring stack to come up,  this could be a side effect for the size of the images involved (they seem heavier on each release). 

Test scenario ensures monitoring stack is available but it is not constraint to a certain amount of time, and currently we expect monitoring stack is available almost instantly, which is not the case as we described.  This ends up on false positives; to avoid this situation it would be required to add a retry mechanism to give enough time to pull the required images for the stack.

jsliacan:
Completed via #4019

jsliacan:
Completed via https://github.com/crc-org/crc/pull/4019

Issue #3997, opened 2024-01-30
[QE] Catch and log `oc login` cert error and log into the cluster with `--insecure-skip-tls-verify`


jsliacan:
Re: Issue #3741 

We have been seeing regular (not consistent) failures stemming from this issue throughout release testing. 

```bash
oc login -u kubeadmin -p YPkA2-hzWvD-cSRru-4yRDq https://api.crc.testing:6443', expected to succeed, exited with exit code: 1 Command stdout: Command stderr: error: The server uses a certificate signed by unknown authority. You may need to use the --certificate-authority flag to provide the path to a certificate file for the certificate authority, or --insecure-skip-tls-verify to bypass the certificate check and use insecure connections.
```

Instead of what we're doing now, we should catch the error and run the [debug steps](https://github.com/crc-org/crc/blob/f7083293cff66dcd7be62632f18104353e22386f/test/e2e/features/story_openshift.feature#L7C1-L10C68) when that happens, then use the `--insecure-skip-tls-verify` flag to log in to the cluster on such occasion so the remaining test cases/steps get executed.

Given that the issue has been around for a while, this will allow our tests to run fully (not be skipped after this failure) and we will still record all the information as we do now.

jsliacan:
Completed via #4019

PR #4019, opened 2024-02-06
[E2e] testsuite maintenance

jsliacan:
**Fixes:** Issue #4018 and #3997

### DIFF
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -842,10 +842,36 @@ func EnsureCRCIsRunning() error {
 }
 
 func EnsureUserIsLoggedIntoClusterSucceedsOrFails(expected string) error {
+
 	if err := setOcEnv(); err != nil {
 		return err
 	}
-	return util.LoginToOcClusterSucceedsOrFails(expected)
+
+	err := util.LoginToOcCluster([]string{})
+	if expected == "succeeds" && err != nil && strings.Contains(err.Error(), "The server uses a certificate signed by unknown authority") {
+		// do some logging
+
+		err1 := util.ExecuteCommand("oc config view --raw -o jsonpath=\"{.clusters[?(@.name=='api-crc-testing:6443')].cluster.certificate-authority-data}\" > ca.base64")
+		if err1 != nil {
+			fmt.Println(err1)
+		}
+		err1 = DecodeBase64File("ca.base64", "ca.crt")
+		if err1 != nil {
+			fmt.Println(err1)
+		}
+		err1 = util.ExecuteCommand("echo QUIT | openssl s_client -connect api.crc.testing:6443 | openssl x509 -out server.crt")
+		if err1 != nil {
+			fmt.Println(err1)
+		}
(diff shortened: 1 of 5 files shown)


test/e2e/testsuite/testsuite.go
```diff
@@ -49,6 +50,7 @@ func ParseFlags() {
 	pflag.StringVar(&CRCExecutable, "crc-binary", "", "Path to the CRC executable to be tested")
 	pflag.BoolVar(&cleanupHome, "cleanup-home", false, "Try to remove crc home folder before starting the suite") // TODO: default=true
 	pflag.StringVar(&CRCVersion, "crc-version", defaultCRCVersion(), "Version of CRC to be tested")
+	pflag.StringVar(&CRCMemory, "crc-memory", "9216", "Memory for CRC VM in MiB")
```
adrianriobo: Maybe better set as "" empty string. If then string is not empty we set the memory otherwise we do nothing / unset the value (to ensure it is the default)

Cause if default value changes in the future we need to remember come here and change it as well

cfergeau:
From a quick glance, looks good to me. A few more details in the commit logs about what  problem are being solved would be nice to have :)

test/e2e/testsuite/testsuite.go
```diff
@@ -808,10 +810,12 @@ func EnsureCRCIsRunning() error {
...
-		// set up and start the cluster with lots of memory
-		err = SetConfigPropertyToValueSucceedsOrFails("memory", "12000", "succeeds")
-		if err != nil {
-			return err
+		// set up and start the cluster with lots of memory, if specified
+		if CRCMemory != "" {
+			err = SetConfigPropertyToValueSucceedsOrFails("memory", CRCMemory, "succeeds")
+			if err != nil {
+				return err
+			}
```
praveenkumar, line 818: with this change looks like the cluster is going to always use default until specified with arg so how does this replicate setting the memory to `12000` or we want to run the cluster with default memory option?

jsliacan: Thanks for looking at this. Yes, we're intending to run the cluster with default memory unless explicitly controlled by the user. 

We realized that we were running it with 12000 default and couldn't remember why. That said, with 4.14.12 bundle, we see the pipelines operator *not* coming up and failing some scenarios if we use default 9000 memory. So it might be that 9000 isn't enough anymore. And it will be good if this test tells us such things. Hence the change.

praveenkumar: Got it, I think our default is only for play with openshift and have very limited workload, if someone wants to use pipeline then better to go with high mem/cpus.

adrianriobo (approved):

gbraad (approved):

gbraad:
It would have been better to isolate each of the fixes and push/approve them separately.

adrianriobo (commented):
Can you remove this last commit as for the time being it is not clear if increased memory will go in, may create a diff PR only with this commit and ref to the increased memory PR?

PR #4020, opened 2024-02-06
kubeconfig: Append `/clusterURL` to the AuthInfo added by crc

### LINKED ISSUE #3940: kubeconfig: Add `/cluster` suffix to the `user` `AuthInfos` entries
Looking at `.kube/config`, I noticed that entries for other clusters look like:
```
contexts:
- context:
    cluster: api-ci-xxx:6443
    namespace: origin
    user: cfergeau/api-ci-xxx:6443
  name: origin/api-ci-xxx:6443/cfergeau
users:
- name: cfergeau/api-ci-xxx:6443
  user:
    token: sha256~yyy
```

By contrast, the entries added by crc are:
```
contexts:
- context:
    cluster: api-crc-testing:6443
    namespace: default
    user: kubeadmin
  name: crc-admin
- context:
    cluster: api-crc-testing:6443
    namespace: default
    user: developer
  name: crc-developer
users:
- name: developer
  user:
    token: sha256~yyy
- name: kubeadmin
  user:
    token: sha256~zzz
```

Changing the user names to `developer/api-crc-testing:6443` and `kubeadmin/api-crc-testing:6443` would make them less ambiguous.

anjannath:
this appends `/api.crc.testing:6443` i.e the API server URL to the name key of the AuthInfos list

this prevents the following error from `oc` when kubeconfig has two entries with same name in the AuthInfos list:

```
error: error loading config file ".kube/config": error converting *[]NamedAuthInfo into *map[string]*api.AuthInfo: duplicate name "user" in list:
```
Fixes #3940

### DIFF
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -174,12 +174,19 @@ func addContext(cfg *api.Config, ip string, clusterConfig *types.ClusterConfig, 
 	if err != nil {
 		return err
 	}
-	cfg.AuthInfos[username] = &api.AuthInfo{
+
+	// append /clustername to AuthInfo
+	clusterUser, err := appendClusternameToUser(username, host)
+	if err != nil {
+		return err
+	}
+
+	cfg.AuthInfos[clusterUser] = &api.AuthInfo{
 		Token: token,
 	}
 	cfg.Contexts[context] = &api.Context{
 		Cluster:   host,
-		AuthInfo:  username,
+		AuthInfo:  clusterUser,
 		Namespace: "default",
 	}
 	return nil
@@ -264,19 +271,57 @@ func mergeConfigHelper(kubeConfigFile, globalConfigFile string) error {
 	if err != nil {
 		return err
 	}
+	// append cluster name to the AuthInfos
+	cfg, err := appendClusterToAuthinfos(currentConf)
+	if err != nil {
+		return err
+	}
 	// Merge the currentConf to globalConfig
-	for name, cluster := range currentConf.Clusters {
+	for name, cluster := range cfg.Clusters {
 		globalConf.Clusters[name] = cluster
 	}
 
-	for name, authInfo := range currentConf.AuthInfos {
+	for name, authInfo := range cfg.AuthInfos {
 		globalConf.AuthInfos[name] = authInfo
 	}
 
-	for name, context := range currentConf.Contexts {
+	for name, context := range cfg.Contexts {
 		globalConf.Contexts[name] = context
 	}
 
-	globalConf.CurrentContext = currentConf.CurrentContext
+	globalConf.CurrentContext = cfg.CurrentContext
 	return clientcmd.WriteToFile(*globalConf, globalConfigPath)
+}
+
+func appendClusterToAuthinfos(cfg *api.Config) (*api.Config, error) {
+	var clusterAPI string
+	for _, cluster := range cfg.Clusters {
+		clusterAPI = cluster.Server
+	}
+	for name, authInfo := range cfg.AuthInfos {
+		delete(cfg.AuthInfos, name)
+		username, err := appendClusternameToUser(name, clusterAPI)
+		if err != nil {
+			return cfg, err
+		}
+		cfg.AuthInfos[username] = authInfo
+	}
+	for _, ctx := range cfg.Contexts {
+		username, err := appendClusternameToUser(ctx.AuthInfo, clusterAPI)
+		if err != nil {
+			return cfg, err
+		}
+		ctx.AuthInfo = username
+	}
+	return cfg, nil
+}
+
+func appendClusternameToUser(username, clusterAPI string) (string, error) {
+	url, err := url.Parse(clusterAPI)
+	if err != nil {
+		return "", err
+	}
+	clusterURL := strings.ReplaceAll(url.Host, ".", "-")
+
+	return fmt.Sprintf("%s/%s", username, clusterURL), nil
 }
--- a/pkg/crc/machine/kubeconfig_test.go
+++ b/pkg/crc/machine/kubeconfig_test.go
@@ -92,6 +92,12 @@ func Test_mergeKubeConfigFile(t *testing.T) {
 	secondaryConfig.Contexts["secondary-context"] = &api.Context{Cluster: "secondary-cluster", AuthInfo: "secondary-user"}
 	secondaryConfig.CurrentContext = "secondary-context"
 
+	secondaryConfigMerged := api.NewConfig()
+	secondaryConfigMerged.Clusters["secondary-cluster"] = &api.Cluster{Server: "https://secondary.example.com"}
+	secondaryConfigMerged.AuthInfos["secondary-user/secondary-example-com"] = &api.AuthInfo{Token: "secondary-token"}
+	secondaryConfigMerged.Contexts["secondary-context"] = &api.Context{Cluster: "secondary-cluster", AuthInfo: "secondary-user/secondary-example-com"}
+	secondaryConfigMerged.CurrentContext = "secondary-context"
+
 	// Create temporary kubeconfig files for the primary and secondary configurations
 	primaryConfigPath, err := createTempKubeConfig(primaryConfig)
 	assert.NoError(t, err, "failed to create temporary kubeconfig file")
@@ -108,7 +114,7 @@ func Test_mergeKubeConfigFile(t *testing.T) {
 	mergedConfig, err := clientcmd.LoadFromFile(primaryConfigPath)
 	assert.NoError(t, err, "failed to load merged kubeconfig")
 
-	for _, config := range []*api.Config{primaryConfig, secondaryConfig} {
+	for _, config := range []*api.Config{primaryConfig, secondaryConfigMerged} {
 
 		for key := range config.AuthInfos {
 			assert.Contains(t, mergedConfig.AuthInfos, key, "expected authInfo not found")


pkg/crc/machine/kubeconfig.go
```diff
@@ -174,6 +174,15 @@ func addContext(cfg *api.Config, ip string, clusterConfig *types.ClusterConfig,
...
 		return err
 	}
+
+	// append /clustername to AuthInfo
+	url, err := url.Parse(clusterConfig.ClusterAPI)
+	if err != nil {
+		return err
+	}
+	clusterURL := url.Host
+	username = fmt.Sprintf("%s/%s", username, clusterURL)
```
gbraad: Not sure why the `clusterURL` variable is used as it does not add much more clarity
```golang
username = fmt.Sprintf("%s/%s", username, url.Host)
```

perhaps it would add clarity if the `username` is not reused, but called

```golang
clusterUser := ...
```

although, the comments already explain this.

pkg/crc/machine/kubeconfig.go
```diff
@@ -264,19 +273,45 @@ func mergeConfigHelper(kubeConfigFile, globalConfigFile string) error {
...
+func appendClusterToAuthinfos(cfg *api.Config) (*api.Config, error) {
+	var clusterURL string
+	for _, cluster := range cfg.Clusters {
+		url, err := url.Parse(cluster.Server)
+		if err != nil {
+			return cfg, err
+		}
+		clusterURL = url.Host
+	}
+	for name, authInfo := range cfg.AuthInfos {
```
gbraad, line 301: why is this logic duplicated?

anjannath: you mean the part doing the url parsing and extracting the host from it?

praveenkumar:
@anjannath https://github.com/crc-org/crc/issues/3940#issuecomment-1929201231 check the commit from this comment because we don't want to modify the complete kubeconfig file instead just want to make sure our entry is according to `oc login`.

anjannath:
> @anjannath [#3940 (comment)](https://github.com/crc-org/crc/issues/3940#issuecomment-1929201231) check the commit from this comment because we don't want to modify the complete kubeconfig file instead just want to make sure our entry is according to `oc login`.

checked the commit, and i think the changes here are similar to what you posted there, since we have two different functions for updating the kubeconfig one for microshift (`mergeKubeconfig`) and one for ocp (`writeKubeconfig`) the changes you posted only handles the openshift preset

the other additional function/changes in this PR to the `mergeKubeConfig` handles the microshift preset

i did test with an existing kubeconfig file with entries from `minikube` and it was only touching the entries added by crc

pkg/crc/machine/kubeconfig.go
```diff
@@ -264,19 +272,45 @@ func mergeConfigHelper(kubeConfigFile, globalConfigFile string) error {
...
+		delete(cfg.AuthInfos, name)
+		username := fmt.Sprintf("%s/%s", name, clusterURL)
```
gbraad: This is essentially a duplication:

```
clusterUser := fmt.Sprintf("%s/%s", username, url.Host)
```

Why is the previous logic not re-used?

gbraad (changes requested):
Duplication prevents easy maintenance

pkg/crc/machine/kubeconfig.go
gbraad: Something like?

```golang
func appendClusternameToUser(username, url) string {
   url, err := url.Parse(clusterConfig.ClusterAPI)
   if err != nil {
      return err
   }

   return fmt.Sprintf("%s/%s", username, url.Host)
}
```

praveenkumar:
> > @anjannath [#3940 (comment)](https://github.com/crc-org/crc/issues/3940#issuecomment-1929201231) check the commit from this comment because we don't want to modify the complete kubeconfig file instead just want to make sure our entry is according to `oc login`.
> 
> checked the commit, and i think the changes here are similar to what you posted there, since we have two different functions for updating the kubeconfig one for microshift (`mergeKubeconfig`) and one for ocp (`writeKubeconfig`) the changes you posted only handles the openshift preset

On microshift side we don't create kubeconfig (as part of crc codebase), we just merge what is created by microshift so should we fix it because we are merging or microshift should fix that kubeconfig?

> 
> the other additional function/changes in this PR to the `mergeKubeConfig` handles the microshift preset
> 
> i did test with an existing kubeconfig file with entries from `minikube` and it was only touching the entries added by crc

pkg/crc/machine/kubeconfig.go
```diff
@@ -174,12 +174,20 @@ func addContext(cfg *api.Config, ip string, clusterConfig *types.ClusterConfig,
...
 		return err
 	}
-	cfg.AuthInfos[username] = &api.AuthInfo{
+
+	// append /clustername to AuthInfo
+	url, err := url.Parse(clusterConfig.ClusterAPI)
+	if err != nil {
+		return err
+	}
+	clusterUser := fmt.Sprintf("%s/%s", username, url.Host)
```
praveenkumar: I think it doesn't matter what we put as suffix even if we use `<username>/crc` that will also work but since we want to it should be `<username>/api-crc-testing:6443` instead `<username>/api.crc.testing:6443`

cfergeau:
> On microshift side we don't create kubeconfig (as part of crc codebase), we just merge what is created by microshift so should we fix it because we are merging or microshift should fix that kubeconfig?

Since microshift document using these files outside of the machine running the cluster ( https://access.redhat.com/documentation/en-us/red_hat_build_of_microshift/4.14/html/configuring/microshift-kubeconfig ), imo it would be nice if the users in there were suffixed, especially as they create multiple kubeconfig depending on the hostname the cluster is going to be accessed from.

gbraad:
> it would be nice if the users in there were suffixed

does this need an issue?

According to the document they generate these files in:

> `/var/lib/microshift/resources/kubeadmin/`

I think this strategy makes sense too, as it prevents any conflict. the file is specific for the microshift usecase. why not write ours in: `~/.crc/kubeconfig` ?

gbraad:
> should be `<username>/api-crc-testing:6443` instead `<username>/api.crc.testing:6443`

Is it necessary to have the port `:6443` included in the suffix?

anjannath:
> > On microshift side we don't create kubeconfig (as part of crc codebase), we just merge what is created by microshift so should we fix it because we are merging or microshift should fix that kubeconfig?
> 
> Since microshift document using these files outside of the machine running the cluster ( https://access.redhat.com/documentation/en-us/red_hat_build_of_microshift/4.14/html/configuring/microshift-kubeconfig ), imo it would be nice if the users in there were suffixed, especially as they create multiple kubeconfig depending on the hostname the cluster is going to be accessed from.

in the case of standalone kubeconfig file that is only containing one pair of cluster and user, there's no provision for having conflicts, the conflict may occur when we merge microshift generated kubeconfig to the user's kubeconfig, so i think its more important that we namespace it (add the cluster suffix) when the merging happens.

and we can also open an issue in microshift repo to know if they think this is something they would consider doing.

anjannath:
> I think this strategy makes sense too, as it prevents any conflict. the file is specific for the microshift usecase. why not write ours in: ~/.crc/kubeconfig ?

we need to merge the kubeconfig onto the default kubeconfig location, i.e `~/.kube/config` otherwise `oc`, `kubectl` and also PD extension won't see the contexts for microshift, ocp from crc

gbraad:
> otherwise oc, kubectl and also PD extension won't see the contexts for microshift, ocp from crc

The kubeconfig can be set by environment variable... something I also expect: `oc-env` to do, right?

I would also expect PD to pick this up. Something the extension should do?


Perhaps even they should allow additional files to be loaded in PD. Especially as the microshift documentation specifies they will use `/var/lib/microshift/resources/kubeadmin/` for this.

anjannath:
> > otherwise oc, kubectl and also PD extension won't see the contexts for microshift, ocp from crc
> 
> The kubeconfig can be set by environment variable... something I also expect: `oc-env` to do, right?
> 

currently `crc oc-env` only updates `PATH` to include `oc` binaries that we ship, its not setting `KUBECONFIG`

> I would also expect PD to pick this up, and perhaps even allow additional files to be loaded. Especially as the microshift documentation specifies they will use `/var/lib/microshift/resources/kubeadmin/` for this.

this only says that when the `microshift` service is started it generates the files in this location, they don't mention how to use it with `oc` or `kubectl` the user is free to either set `KUBECONFIG` or use the `--kubconfig` flag or merge with the existing `~/.kube/config`

gbraad:
> they don't mention how to use it

regardless, seems to be 'convention' as it is documented. Although, something for PD+extension to handle.

> its not setting `KUBECONFIG`

Shouldn't we do this? /cc: @praveenkumar @cfergeau and therefore not 'merge' our config in by default (or as default now, but allow this to be configured).

cfergeau:
>  >    its not setting KUBECONFIG
> Shouldn't we do this? /cc: @praveenkumar @cfergeau and therefore not 'merge' our config in by default (or as default now, but allow this to be configured).

If `kubectl`/`oc` know how to merge multiple files in `~/.kube`, we definitely should make use of this instead of modifying the main file ourselves.
Not sure about forcing `KUBECONFIG` if this means cutting easy access to other clusters which are already configured in `~/.kube`

gbraad:
> If kubectl/oc know how to merge multiple files in ~/.kube, 

is this possible? if not, file an issue for something like a `kubeconfig.d` approach.

> cutting easy access to other clusters 

I believe this is what happens... but in that case suggesting merging content to `kubeconfig` makes sense. Guess this constitutes a separate discussion.

However, I still did not get an answer to: https://github.com/crc-org/crc/pull/4020#issuecomment-1931457175 about the port in the suffix. Is this necessary?

cfergeau:
> However, I still did not get an answer to: [#4020 (comment)](https://github.com/crc-org/crc/pull/4020#issuecomment-1931457175) about the port in the suffix. Is this necessary?

Nothing is necessary, even appending a random string would work instead of the cluster name. Having the port number makes it consistent with the example in https://github.com/crc-org/crc/issues/3940.

pkg/crc/machine/kubeconfig.go
anjannath: updated to have the `.` replaced with `-`

anjannath:
`KUBECONFIG` env variable can contain list of config files, so we could set `KUBECONFIG=$HOME/.kube/config:~/.crc/kubeconfig` and `oc` `kubectl` will merge the config files by itself, and will be able to access users, clusters defined in either of the files.. 

but i think it'd still be good to append `/clustername` to the users/AuthInfos as according to kubeconfig merge rules:
> The first file to set a particular value or map key wins.

https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files

gbraad:
> Nothing is necessary,

Don't want to lock the user in so he can only use port `6443`, so it wouldn't work anymore if remapped to `8443`, etc. Although I understand the matching happens on the actual key/auth.

> > but i think it'd still be good to append /clustername to the users/AuthInfos

the identifier is more of a 'nice to have' for the user, but 'needs to be unique' due to how they store it in code when read/parsed. it does not add anything else than user-facing info.

praveenkumar:
> > Nothing is necessary,
> 
> Don't want to lock the user in so he can only use port `6443`, so it wouldn't work anymore if remapped to `8443`, etc. Although I understand the matching happens on the actual key/auth.

We are not hardcoding `6443` but just using the api server url and if url is having different port then we use that even that is just the placeholder.

> 
> > > but i think it'd still be good to append /clustername to the users/AuthInfos
> 
> the identifier is more of a 'nice to have' for the user, but 'needs to be unique' due to how they store it in code when read/parsed. it does not add anything else than user-facing info.

Right, Also since we are merging microshift kubeconfig file to global one it also remove the override in case same user exist for different cluster.

> Shouldn't we do this? /cc: @praveenkumar @cfergeau and therefore not 'merge' our config in by default (or as default now, but allow this to be configured).

most of the tools do it (like kind/minikube ..etc.) but they don't merge but add the context like we are doing for OCP by generating the required field. Technically on microshift side we are getting the required info from the generated kubeconfig by parsing it and then appending it to global kubeconfig (something same might oc/kubectl can do behind the scene when KUBECONFIG env variable set for multiple kubeconfigs) .  With our current approach (merging/appending) user can use oc command on OCP/microshift cluster without everytime setting the KUBECONFIG env which is big plus since most of the time user don't even use `oc-env` since they already have cli tool installed just want to explore the cluster.

pkg/crc/machine/kubeconfig.go
```diff
@@ -264,19 +271,57 @@ func mergeConfigHelper(kubeConfigFile, globalConfigFile string) error {
...
 
-	globalConf.CurrentContext = currentConf.CurrentContext
+	globalConf.CurrentContext = cfg.CurrentContext
 	return clientcmd.WriteToFile(*globalConf, globalConfigPath)
 }
+
+func appendClusterToAuthinfos(cfg *api.Config) (*api.Config, error) {
+	var clusterURL string
+	for _, cluster := range cfg.Clusters {
+		clusterURL = cluster.Server
```
praveenkumar: shouldn't this `clusterAPI` because below in the function you used `clusterURL` for string replace?

praveenkumar: Apart from this lgtm.

anjannath: `clusterURL` here and  `clusterAPI` has the same value, do you mean just to name the variable as `clusterAPI`, or actually use the `ClusterConfig.ClusterAPI`?

anjannath: i wanted to avoid having to pass the `ClusterConfig` to this function so deriving the URL from the `clusters` section of the kubeconfig itself

gbraad:
> and if url is having different port then we use that even that is just the placeholder.

I can read the code, but I am questioning if this is necessary. It might be better to just say:

`<username>/crc-api-testing`

and therefore substring on the port/semi-colon.

gbraad:
> Also since we are merging microshift kubeconfig 

So adding the `preset` as part of the identifier makes more sense?

`<username>/crc-api-testing-ocp` or `<username>/crc-api-testing-microshift`, because OKD would have the same ports and cause issues?

anjannath:
> > Also since we are merging microshift kubeconfig
> 
> So adding the `preset` as part of the identifier makes more sense?
> 
> `<username>/crc-api-testing-ocp` or `<username>/crc-api-testing-microshift`, because OKD would have the same ports and cause issues?

ahh, yes it'd cause issues since it'd be the same `kubeadmin/api-crc-testing:6443` for both okd and ocp

but we allow only one preset to run at a time, so kubeconfig will not have entries from both, as to move to okd preset from ocp we'd have to run `crc delete` which cleans the kubeconfig of entries that were added by `crc`

praveenkumar:
> > > Also since we are merging microshift kubeconfig
> > 
> > 
> > So adding the `preset` as part of the identifier makes more sense?
> > `<username>/crc-api-testing-ocp` or `<username>/crc-api-testing-microshift`, because OKD would have the same ports and cause issues?
> 
> ahh, yes it'd cause issues since it'd be the same `kubeadmin/api-crc-testing:6443` for both okd and ocp
> 
> but we allow only one preset to run at a time, so kubeconfig will not have entries from both, as to move to okd preset from ocp we'd have to run `crc delete` which cleans the kubeconfig of entries that were added by `crc`

there is no `kubeadmin` user for microshift it should be `user` because this is what `system:admin` role associated to so it shouldn't conflict also the whole point to have port number as part of `<username>` to consist what other entry in kubeconfig.

pkg/crc/machine/kubeconfig.go
praveenkumar: @anjannath I was only referring the variable name not `ClusterConfig.ClusterAPI`

gbraad:
>  to consist what other entry in kubeconfig.

Do not understand this part. To be consistent with the other entries in `kubeconfig` ?

praveenkumar:
> > to consist what other entry in kubeconfig.
> 
> Do not understand this part. To be consistent with the other entries in `kubeconfig` ?

so whenever we do `oc login <any_cluster` then it add entry to kubeconfig with `<username>/api_url_along_with_port`  so this is what I mean around consistency.

PR #4025, opened 2024-02-08
create symlink for podman in ~/.crc/bin/podman dir

### LINKED ISSUE #4022: `crc oc-env` adds `podman` client to path
When running `crc oc-env` users would expect to only have `oc` setup to use the CRC cluster, and when user wants to use `podman` from crc, running `crc podman-env` would to set it up.

```
% crc status
CRC VM:          Running
MicroShift:      Running (v4.14.8)
RAM Usage:       1.204GB of 16.01GB
Disk Usage:      4.993GB of 16.1GB (Inside the CRC VM)
Cache Usage:     255GB
Cache Directory: /Users/anath/.crc/cache

% which podman
/opt/podman/bin/podman

% eval `crc oc-env`

% which podman
/home/anath/.crc/bin/oc/podman
```

this happens because currently both `oc` and `podman` binaries are in the same dir `~/.crc/bin/oc`, and `eval crc-env` adds the `~/.crc/bin/oc` dir to PATH.

we could instead create a separate dir for podman `~/.crc/bin/podman` and add this dir to `PATH` when `crc podman-env` is ran

anjannath:
earlier both oc and podman binaries were present in the ~/.crc/bin/oc dir and added to PATH by the `crc oc-env` command, which resulted in adding both the binaries to path when the user might want only the `oc` binary  to be added to path by the `crc oc-env` command

this adds the `podman` binary in a separate  directory ~/.crc/bin/podman and is added to PATH by  running the `crc podman-env` command

fixes #4022


anjannath:
I think we need to also add a preflight check to remove the `podman` binary from the old `~/.crc/bin/oc` location if user doesn't run `crc cleanup` while upgrading

praveenkumar:
> I think we need to also add a preflight check to remove the `podman` binary from the old `~/.crc/bin/oc` location if user doesn't run `crc cleanup` while upgrading

Yes, otherwise user will have an old podman version lying around.

gbraad:
make sure to communicate with @vyasgun to prevent issues/confusion.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -155,3 +156,31 @@ func removeAllLogs() error {
 func removeCRCHostEntriesFromKnownHosts() error {
 	return ssh.RemoveCRCHostEntriesFromKnownHosts()
 }
+
+// remove the following checks during release of v2.35.0 or after
```
cfergeau: In my opinion, it's the kind of checks we have to keep forever, as we often have people upgrading from very old versions. 
Alternatively, in a few releases, we could switch to a preflight which would only check for the file existence, and advise the user to use `crc cleanup` if such a file exists.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -155,3 +156,31 @@ func removeAllLogs() error {
...
+func checkPodmanInOcBinDir() error {
+	podmanBinPath := filepath.Join(constants.CrcOcBinDir, "podman")
+	switch runtime.GOOS {
+	case "linux":
+		podmanBinPath = filepath.Join(constants.CrcOcBinDir, "podman-remote")
+	case "windows":
+		podmanBinPath = filepath.Join(constants.CrcOcBinDir, "podman.exe")
+	}
```
cfergeau, line 163: There's a `PodmanRemoteExecutableName` constant which you could use here

cfergeau (commented):
Does this need documentation updates?

praveenkumar:
looks like unit test is failing for windows.

anjannath:
> looks like unit test is failing for windows.

should be fixed now, branch was not rebased on my local repo and tests were passing locally, in the main branch a new preflight check was merged, so the number of preflight checks were not what i was expecting, rebased now and test are passing on CI

anjannath:
> Does this need documentation updates?

seems we don't mention about `crc podman-env` command in the docs at all and no mention of `podman` command being available after running `crc oc-env` we should be good without updating the docs

PR #4029, opened 2024-02-09
Add depreciation warning for older macOS version

praveenkumar:
Try to test this PR on older mac and it is going to show depreciation warning.


anjannath:
Do we have a lot of macOS 12 user's still who haven't updated? should we add this warning in the installer itself, in the past we directly removed support for older version of macOS in the pkg installer

praveenkumar:
> Do we have a lot of macOS 12 user's still who haven't updated? should we add this warning in the installer itself, in the past we directly removed support for older version of macOS in the pkg installer

We don't have and having warning in installer also a good way to let user know. @cfergeau what do you think?

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -488,36 +486,10 @@ func (d *Driver) sendSignal(s syscall.Signal) error {
...
 
-var (
-	once              sync.Once
-	macCurrentVersion string
-	errGettingVersion error
-)
-
 func (d *Driver) supportsVirtiofs() bool {
-	supportsVirtioFS, err := macosAtLeast("12.0.0")
+	supportsVirtioFS, err := darwin.AtLeast("12.0.0")
```
cfergeau, line 490: Not asking for changes, but one could argue this should be `darwin.AtLeast("21.0.0")`, and that `AtLeast` needs to be rewritten (darwin and macos versions are different things https://en.wikipedia.org/wiki/Darwin_(operating_system) )

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -224,3 +225,14 @@ func fixPlistFileExists(agentConfig launchd.AgentConfig) error {
...
 	return waitForDaemonRunning()
 }
+
+func deprecationNotice() error {
+	supports, err := darwin.AtLeast("13.0.0")
+	if err != nil {
+		return err
+	}
+	if !supports {
+		logging.Warnf("This version of macOS is going to be unsupported for CRC, Please update to latest version of macOS")
```
cfergeau: I would say "Please update to macOS 13 or newer"

cfergeau (commented):
looks good to me.

pkg/crc/preflight/preflight_checks_darwin.go
praveenkumar: Done.

praveenkumar:
>  should we add this warning in the installer itself

@anjannath Added, can you verify it once?

cfergeau:
I  see `make check` failures in github actions, are they fixed in the latest iteration?

anjannath:
> > should we add this warning in the installer itself
> 
> @anjannath Added, can you verify it once?

tested on macOS 12, showing deprecation message while installer is started..
image

praveenkumar:
> I see make check failures in github actions, are they fixed in the latest iteration?

@cfergeau I forgot to rebase, will do that.

anjannath (approved):

Issue #4034, opened 2024-02-14
[BUG] Github action `windows-e2e` sometimes fails with connection refused error


jsliacan:
### General information

Recently, [some](https://github.com/crc-org/crc/actions/runs/7900918014/job/21563444025) `windows-e2e` workflow runs fail with 
```bash
Copy resources to target
ssh: connect to host 20.88.32.35 port 22: Connection refused
ssh: connect to host 20.88.32.35 port 22: Connection refused
lost connection
```
and [some](https://github.com/crc-org/crc/actions/runs/7900282165/job/21561489871) pass this step.

jsliacan:
Part of the solution is this: https://github.com/adrianriobo/deliverest/pull/27. @adrianriobo did the rest of the work to use those changes in the images and workflow. Passing now, see e.g. https://github.com/crc-org/crc/actions/runs/8282324003

PR #4035, opened 2024-02-14
[QE] Add debug steps and connection check step to `windows-e2e` workflow

jsliacan:
**Related to:** Issue #4034

### DIFF
--- a/.github/workflows/windows-e2e.yml
+++ b/.github/workflows/windows-e2e.yml
@@ -134,6 +134,18 @@ jobs:
           # Check logs 
           podman logs -f crc-win-support
 
+      - name: Upload installer
+        uses: actions/upload-artifact@v4
+        with:
+          name: Windows Installer (windows-2022)
+          path: "crc-windows-installer.zip"
+
+      - name: Upload e2e image
+        uses: actions/upload-artifact@v4
+        with:
+          name: crc-e2e-gh
+          path: "crc-e2e.tar"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-e2e.yml
```diff
@@ -146,6 +146,31 @@ jobs:
           name: crc-e2e-gh
           path: "crc-e2e.tar"
 
+      - name: Check SSH connection to host
```
cfergeau, line 149: As in the other PR, some context to explain why we need this would be nice to have. I'm guessing this is some sanity check to debug https://github.com/crc-org/crc/issues/4034

cfergeau (commented):
Looks good to me

.github/workflows/windows-e2e.yml
jsliacan: Thanks @cfergeau , I ran these changes on my fork by setting the branch to default: [workflow run here](https://github.com/jsliacan/crc/actions/runs/7913916516). And we figured that maybe it's good to put the repeats into the `crc-support` image that we're using to install CRC on the host. The actions in the img are those that the issue is related to (restart after installing CRC). So solution should be coupled with it too. I will close (or amend) this PR when we test it and confirm it's done. Marking WIP now. Thank you!

jsliacan:
Closing in favor of https://github.com/adrianriobo/deliverest/pull/27

PR #4059, opened 2024-03-07
Update OKD version to 4.15.0-0.okd-2024-02-23-163410

praveenkumar:


praveenkumar:
/hold

It will require more resource then default so we have to first make a decision about https://github.com/crc-org/snc/issues/861 one

PR #4012, opened 2024-02-05
deprecated preset: add warning for podman preset during  and  commands

vyasgun:
**Fixes:** Issue #3970 

## Solution/Idea

`podman` preset deprecation warning has been added in two places: `crc config set preset podman` and `crc start`

## Testing

What is the _bottom-line_ functionality that needs testing? Describe in pseudo-code or in English. Use verifiable statements that tie your changes to existing functionality.

1. `crc config set preset podman` should print a warning
2. `crc start` with already set preset `podman` should print a warning

Issue #3970, opened 2024-01-22
Removal of podman preset from config and codebase


praveenkumar:
`Podman` preset was added when there was no podman machine and now it is not consumed by users and we also stopped updating the bundles from some time now. It would be better to remove it from crc codebase.

cfergeau:
We could first start by warning or erroring out when the podman preset is used, this is easier to revert if there are unexpected blockers with it (I hope not!)

gbraad:
I agree with Christophe to first error out, but the next release after this would be OK to remove this.

Would it be OK for Gunjan (@vyasgun) to look at this?

anjannath:
erroring out will be better, we already announced that the preset will be removed in 2.28.0 release https://github.com/crc-org/crc/releases/tag/v2.28.0

i think we should also stop shipping the podman client and putting it on `~/.crc/bin/oc/podman` and adding to PATH?

praveenkumar:
> i think we should also stop shipping the podman client and putting it on `~/.crc/bin/oc/podman` and adding to PATH?

podman client we ship for `microshift/openshift` bundle also to consume so that shouldn't be removed.

praveenkumar:
> Would it be OK for Gunjan (@vyasgun) to look at this?

Sure, assigned.

cfergeau:
> > i think we should also stop shipping the podman client and putting it on `~/.crc/bin/oc/podman` and adding to PATH?
> 
> podman client we ship for `microshift/openshift` bundle also to consume so that shouldn't be removed.

Maybe we should integrate better with podman installed through podman's official installer, or even ship/use podman iwndows/mac installer instead of shipping our own binaries?

gbraad:
^^^ I agree, but let's first at messaging and have a new issue to discuss how to integrate/distribute.

cfergeau:
> ^^^ I agree, but let's first at messaging and have a new issue to discuss how to integrate/distribute.

Yup, definitely a (long term) next step after we've dealt with the podman preset :)

anjannath:
> i think we should also stop shipping the podman client and putting it on ~/.crc/bin/oc/podman and adding to PATH?

The main issue i faced while using `crc` and podman machine together can be solved if we fix #4022

cfergeau:
#4062 has been merged, this can be closed.

PR #4062, opened 2024-03-08
Remove the Podman preset and all its usages

vyasgun:
**Fixes:** Issue #3970 

**Relates to:** Issue #3970, PR #4012
## Solution/Idea
This PR removes the usages of the `Podman` from the code. This is a follow-up to the PR: https://github.com/crc-org/crc/pull/4012 which added a deprecated warning for the users who chose the `Podman` preset.

## Proposed changes
The changes to remove the usage of Podman preset have been done in two PRs. (https://github.com/crc-org/crc/pull/4012)
- The first change was to print a deprecation warning for users using the preset. 
- In the current PR, the preset itself is being removed from the code.

## Testing
Setting and using `crc` with `Podman` preset should fail.

### DIFF
--- a/.github/workflows/macos-installer.yml
+++ b/.github/workflows/macos-installer.yml
@@ -44,11 +44,7 @@ jobs:
           path: "./out/macos-universal/crc-macos-installer.pkg"
       - name: Install crc pkg
         run: sudo installer -pkg out/macos-universal/crc-macos-installer.pkg -target /
-      - name: Set podman preset as part of config
-        run: crc config set preset podman
+      - name: Set microshift preset as part of config
+        run: crc config set preset microshift
       - name: Run crc setup command
         run: crc setup
-      - name: Wait 10 sec for the daemon to serve
-        run: sleep 10
-      - name: Run crc with podman preset
-        run: crc start
--- a/cmd/crc/cmd/start.go
+++ b/cmd/crc/cmd/start.go
@@ -87,10 +87,6 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		EnableBundleQuayFallback: config.Get(crcConfig.EnableBundleQuayFallback).AsBool(),
 	}
 
-	if startConfig.Preset == preset.Podman {
-		logging.Warn(preset.PodmanDeprecatedWarning)
-	}
-
 	client := newMachine()
 	isRunning, _ := client.IsRunning()
 
--- a/pkg/crc/config/settings_test.go
+++ b/pkg/crc/config/settings_test.go
@@ -36,7 +36,7 @@ func newInMemoryConfig() (*Config, error) {
 }
 
 // Check that with the default preset, we cannot set less CPUs than the defaultCPUs
-// but that it is allowed with a different preset with less requirements (podman)
+// but that it is allowed with a different preset with less requirements (microshift)
 func TestCPUsValidate(t *testing.T) {
 	cfg, err := newInMemoryConfig()
 	require.NoError(t, err)
@@ -58,7 +58,7 @@ func TestCPUsValidate(t *testing.T) {
 		IsSecret:  false,
 	}, cfg.Get(CPUs))
 
-	_, err = cfg.Set(Preset, crcpreset.Podman)
+	_, err = cfg.Set(Preset, crcpreset.Microshift)
 	require.NoError(t, err)
 	_, err = cfg.Set(CPUs, defaultCPUs-1)
 	require.NoError(t, err)
@@ -76,7 +76,7 @@ func TestSetPreset(t *testing.T) {
 	cfg, err := newInMemoryConfig()
 	require.NoError(t, err)
 
-	_, err = cfg.Set(Preset, crcpreset.Podman)
+	_, err = cfg.Set(Preset, crcpreset.Microshift)
 	require.NoError(t, err)
 	_, err = cfg.Set(Memory, 10800)
 	require.NoError(t, err)
@@ -118,7 +118,7 @@ func TestUnsetPreset(t *testing.T) {
 	cfg, err := newInMemoryConfig()
 	require.NoError(t, err)
 
-	_, err = cfg.Set(Preset, crcpreset.Podman)
+	_, err = cfg.Set(Preset, crcpreset.Microshift)
 	require.NoError(t, err)
 	_, err = cfg.Set(Memory, 10800)
 	require.NoError(t, err)
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -196,6 +196,10 @@ func (bundle *CrcBundleInfo) IsMicroshift() bool {
 	return bundle.GetBundleType() == crcPreset.Microshift
 }
 
+func (bundle *CrcBundleInfo) IsPodman() bool {
+	return bundle.GetBundleType() == crcPreset.Podman
+}
+
 func (bundle *CrcBundleInfo) verify() error {
 	files := []string{
 		bundle.GetSSHKeyPath(),
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -167,21 +167,12 @@ func (bundle *CrcBundleInfo) GetBundleBuildTime() (time.Time, error) {
 	return time.Parse(time.RFC3339, strings.TrimSpace(bundle.BuildInfo.BuildTime))
 }
 
-func (bundle *CrcBundleInfo) GetOpenshiftVersion() string {
+func (bundle *CrcBundleInfo) GetVersion() string {
 	return bundle.ClusterInfo.OpenShiftVersion.String()
 }
 
 func (bundle *CrcBundleInfo) GetPodmanVersion() string {
 	return bundle.Nodes[0].PodmanVersion
-}
-
-func (bundle *CrcBundleInfo) GetVersion() string {
-	switch bundle.GetBundleType() {
-	case crcPreset.Podman:
-		return bundle.GetPodmanVersion()
-	default:
-		return bundle.GetOpenshiftVersion()
-	}
 }
 
 func (bundle *CrcBundleInfo) GetBundleNameWithoutExtension() string {
@@ -205,22 +196,14 @@ func (bundle *CrcBundleInfo) IsMicroshift() bool {
 	return bundle.GetBundleType() == crcPreset.Microshift
 }
 
-func (bundle *CrcBundleInfo) IsPodman() bool {
-	return bundle.GetBundleType() == crcPreset.Podman
-}
-
 func (bundle *CrcBundleInfo) verify() error {
 	files := []string{
 		bundle.GetSSHKeyPath(),
 		bundle.GetDiskImagePath(),
 		bundle.GetKernelPath(),
 		bundle.GetInitramfsPath(),
-	}
-	if !bundle.IsPodman() {
-		files = append(files, []string{
-			bundle.GetOcPath(),
-			bundle.GetKubeConfigPath()}...)
-	}
+		bundle.GetOcPath(),
+		bundle.GetKubeConfigPath()}
 
 	for _, file := range files {
 		if file == "" {
@@ -352,7 +335,7 @@ func Download(preset crcPreset.Preset, bundleURI string, enableBundleQuayFallbac
 				return image.PullBundle(constants.GetDefaultBundleImageRegistry(preset))
 			}
 			return downloadedBundlePath, err
-		case crcPreset.Podman, crcPreset.OKD:
+		case crcPreset.OKD:
 			fallthrough
 		default:
 			return image.PullBundle(constants.GetDefaultBundleImageRegistry(preset))
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -55,17 +55,16 @@ func (repo *Repository) Get(bundleName string) (*CrcBundleInfo, error) {
 		return nil, err
 	}
 
-	if !bundleInfo.IsPodman() {
-		// TODO: update this logic after major release of bundle like 4.14
-		// As of now we are using this logic to support older bundles of microshift and it need to be updated
-		// to only provide app domain route information as per preset.
-		if !crcstrings.Contains([]string{constants.AppsDomain, constants.MicroShiftAppDomain}, fmt.Sprintf(".%s", bundleInfo.ClusterInfo.AppsDomain)) {
-			return nil, fmt.Errorf("unexpected bundle, it must have %s or %s apps domain", constants.AppsDomain, constants.MicroShiftAppDomain)
-		}
-		if bundleInfo.GetAPIHostname() != fmt.Sprintf("api%s", constants.ClusterDomain) {
-			return nil, fmt.Errorf("unexpected bundle, it must have %s base domain", constants.ClusterDomain)
-		}
+	// TODO: update this logic after major release of bundle like 4.14
+	// As of now we are using this logic to support older bundles of microshift and it need to be updated
+	// to only provide app domain route information as per preset.
+	if !crcstrings.Contains([]string{constants.AppsDomain, constants.MicroShiftAppDomain}, fmt.Sprintf(".%s", bundleInfo.ClusterInfo.AppsDomain)) {
+		return nil, fmt.Errorf("unexpected bundle, it must have %s or %s apps domain", constants.AppsDomain, constants.MicroShiftAppDomain)
 	}
+	if bundleInfo.GetAPIHostname() != fmt.Sprintf("api%s", constants.ClusterDomain) {
+		return nil, fmt.Errorf("unexpected bundle, it must have %s base domain", constants.ClusterDomain)
+	}
+
 	return &bundleInfo, nil
 }
 
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -340,6 +340,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 	}
 	defer vm.Close()
 
+	// Return with an error redirecting user to use podman machine instead
+	if vm.bundle.IsPodman() {
+		return &types.StartResult{}, fmt.Errorf("error: %s", crcPreset.PodmanDeprecatedWarning)
+	}
 	currentBundleName := vm.bundle.GetBundleName()
 	if currentBundleName != bundleName {
 		logging.Debugf("Bundle '%s' was requested, but the existing VM is using '%s'",
@@ -679,6 +683,9 @@ func (client *client) IsRunning() (bool, error) {
 }
 
 func (client *client) validateStartConfig(startConfig types.StartConfig) error {
+	if startConfig.Preset == crcPreset.Podman {
+		return fmt.Errorf(crcPreset.PodmanDeprecatedWarning)
+	}
 	if client.monitoringEnabled() && startConfig.Memory < minimumMemoryForMonitoring {
 		return fmt.Errorf("Too little memory (%s) allocated to the virtual machine to start the monitoring stack, %s is the minimum",
 			units.BytesSize(float64(startConfig.Memory)*1024*1024),
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -24,7 +23,6 @@ import (
 	"github.com/crc-org/crc/v2/pkg/crc/network"
 	"github.com/crc-org/crc/v2/pkg/crc/network/httpproxy"
 	"github.com/crc-org/crc/v2/pkg/crc/oc"
-	"github.com/crc-org/crc/v2/pkg/crc/podman"
 	crcPreset "github.com/crc-org/crc/v2/pkg/crc/preset"
 	"github.com/crc-org/crc/v2/pkg/crc/services"
 	"github.com/crc-org/crc/v2/pkg/crc/services/dns"
@@ -356,11 +354,6 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 	}
 	if vmState == state.Running {
 		logging.Infof("A CRC VM for %s %s is already running", startConfig.Preset.ForDisplay(), vm.bundle.GetVersion())
-		if vm.bundle.IsPodman() {
-			return &types.StartResult{
-				Status: vmState,
-			}, nil
-		}
 		clusterConfig, err := getClusterConfig(vm.bundle)
 		if err != nil {
 			return nil, errors.Wrap(err, "Cannot create cluster configuration")
@@ -479,34 +472,6 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 	proxyConfig, err := getProxyConfig(vm.bundle)
 	if err != nil {
 		return nil, errors.Wrap(err, "Error getting proxy configuration")
-	}
-
-	if vm.bundle.IsPodman() {
-		// **************************
-		//  END OF PODMAN START CODE
-		// **************************
-		if err := configurePodmanProxy(sshRunner, proxyConfig); err != nil {
-			return nil, errors.Wrap(err, "Failed to configure proxy for podman")
-		}
-		if err := dns.AddPodmanHosts(instanceIP); err != nil {
-			return nil, errors.Wrap(err, "Failed to add podman host dns entry")
-		}
-
-		if err := updateCockpitConsoleBearerToken(sshRunner); err != nil {
-			return nil, fmt.Errorf("Failed to rotate bearer token for cockpit webconsole: %w", err)
-		}
-
-		c, err := client.ConnectionDetails()
-		if err != nil {
-			return nil, err
-		}
-		if err := addPodmanSystemConnections(c); err != nil {
-			return nil, err
-		}
-
-		return &types.StartResult{
-			Status: vmState,
-		}, nil
 	}
 
 	proxyConfig.ApplyToEnvironment()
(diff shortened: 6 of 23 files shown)


cmd/crc/cmd/version.go
```diff
@@ -37,15 +37,13 @@ type version struct {
 	Version          string `json:"version"`
 	Commit           string `json:"commit"`
 	OpenshiftVersion string `json:"openshiftVersion"`
-	PodmanVersion    string `json:"podmanVersion"`
```
anjannath: this should not be removed, the podman version here is also used in the  `openshift` and `microshift` presets

vyasgun: Thanks! Added it back in second commit :)

vyasgun:
Checks which set `podman` as preset are failing

anjannath:
> Checks which set `podman` as preset are failing

yeah, `integration-crc` and `e2e-crc` are also failing because of `podman` preset related tests, i think you can remove the tests as well in this PR

.github/workflows/macos-installer.yml
```diff
@@ -44,11 +44,11 @@ jobs:
           path: "./out/macos-universal/crc-macos-installer.pkg"
       - name: Install crc pkg
         run: sudo installer -pkg out/macos-universal/crc-macos-installer.pkg -target /
-      - name: Set podman preset as part of config
-        run: crc config set preset podman
+      - name: Set microshift preset as part of config
```
gbraad, line 47: Running the microshift preset might result in a very diferent test, so best to let @jsliacan and @adrianriobo chime in on this.

cmd/crc/cmd/start.go
```diff
@@ -87,10 +87,6 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		EnableBundleQuayFallback: config.Get(crcConfig.EnableBundleQuayFallback).AsBool(),
 	}
 
-	if startConfig.Preset == preset.Podman {
-		logging.Warn(preset.PodmanDeprecatedWarning)
```
gbraad, line 91: Do we need to display a removal notice and refer people to use `podman machine` instead?
/cc: @crc-org/crc-team

anjannath: how long will we show this message, is it going to be removed after a few releases? or we'll always show the message when user runs `crc config set podman preset`

another thing is this PR removes `preset.Podman` constant, i don't think we should keep this around to have this check and instead use `Preset("podman")`

gbraad: if you do `crc start` when the preset has already been set to `podman` what is the output now?

anjannath: there are repeated non fatal errors complaining about the preset, both for `setup` and `start` and once we run `crc config unset preset` the errors go away

i think we need to inform the user how to get rid of the errors

```
% crc setup
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO unexpected preset mode podman, using default
ERRO unexpected preset mode podman, using default
INFO Using bundle path /Users/anath/.crc/cache/crc_vfkit_4.14.12_arm64.crcbundle
INFO Check if Podman binary exists in: /Users/anath/.crc/bin/oc
INFO Checking if /Users/anath/.crc/cache/crc_vfkit_4.14.12_arm64.crcbundle exists
ERRO unexpected preset mode podman, using default
```

---

```
% crc start --log-level debug
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
ERRO[0000] unexpected preset mode podman, using default
DEBU CRC version: 2.33.0+93a8fc
DEBU OpenShift version: 4.14.12
DEBU Podman version: 4.4.4
DEBU Running 'crc start'
ERRO unexpected preset mode podman, using default
DEBU Total memory of system is 34359738368 bytes
ERRO unexpected preset mode podman, using default
ERRO unexpected preset mode podman, using default
DEBU No new version available. The latest version is 2.33.0
ERRO unexpected preset mode podman, using default
DEBU Checking file: /Users/anath/.crc/machines/crc/.crc-exist
ERRO unexpected preset mode podman, using default
INFO Using bundle path /Users/anath/.crc/cache/crc_vfkit_4.14.12_arm64.crcbundle
DEBU kern.osproductversion is: 14.3.1
DEBU Running '/Users/anath/.crc/bin/crc-admin-helper-darwin --version'
DEBU Found crc-admin-helper-darwin version 0.5.2
DEBU crc-admin-helper executable already cached
DEBU GOARCH is arm64 GOOS is darwin
DEBU Total memory of system is 34359738368 bytes
INFO Check if Podman binary exists in: /Users/anath/.crc/bin/oc
DEBU Running '/Users/anath/.crc/bin/vfkit -v'
DEBU Found vfkit version 0.5.0
DEBU Running 'bash -c launchctl list | grep com.redhat.crc.daemon | awk '{print $1}''
DEBU Checking file: /Users/anath/.crc/machines/crc/.crc-exist
INFO Loading bundle: crc_vfkit_4.14.12_arm64...
ERRO unexpected preset mode podman, using default
DEBU Using secret from configuration
INFO Creating CRC VM for OpenShift 4.14.12...
```

cfergeau: would be useful to have a removal notice for a while. We can remove it when we remember about it :)

gbraad: I would suggest disallowing starting with this. A removal and alternative use message in that case, right?

vyasgun: I think we can return an error when user tries to start an existing machine with podman preset and modify tests to expect that error

test/integration/podman_test.go
jsliacan, line 1: Maybe worth keeping this test: it verifies podman-remote functionality provided by a podman-preset. However, we can check that functionality for a different preset, microshift requires less resources than openshift, so maybe that's the candidate. I think this sort of modification should work: https://github.com/jsliacan/crc/blob/remove-podman-preset-qe/test/integration/podman_test.go

test/e2e/testsuite/testsuite.go
```diff
@@ -656,15 +656,9 @@ func FileExistsInCRCHome(fileName string) error {
...
+func RemoveBundleFromCache(_ string) error {
 
-	var p preset.Preset
-
-	if presetName == "podman" {
-		p = preset.Podman
-	} else {
-		p = preset.OpenShift
-	}
+	var p = preset.OpenShift
```
jsliacan, line 658: I think this part was written before microshift preset existed(?) so maybe it's good to use this opportunity to fix this and distinguish between `microshift` and `openshift` presets, while dropping `podman` preset.

jsliacan: There are a few additional places in this file that maybe could be cleaned up at this occasion - if you have time?  Something like in here: https://github.com/jsliacan/crc/pull/17/files

jsliacan (commented):
There are also two feature files that could do with some trimming if `podman` preset isn't going to exist. 

- test/e2e/features/minimal.feature
- test/e2e/features/proxy.feature

The diffs that I could think of are in the same example: 
https://github.com/jsliacan/crc/pull/17/files?diff=split&w=0

test/e2e/testsuite/testsuite.go
vyasgun: > There are a few additional places in this file that maybe could be cleaned up at this occasion - if you have time? Something like in here: https://github.com/jsliacan/crc/pull/17/files

yes, that makes sense. Will look into it.



> I think this part was written before microshift preset existed(?) so maybe it's good to use this opportunity to fix this and distinguish between `microshift` and `openshift` presets, while dropping `podman` preset.

Maybe @cfergeau @gbraad can chime in if we should default to `microshift`.

gbraad: > if we should default to microshift.

What do you mean with 'default to' ?

test/integration/podman_test.go
vyasgun: Thanks! I have made the suggested changes.

test/e2e/testsuite/testsuite.go
vyasgun: Sorry about the confusion for the mix up with 'default to'. What I meant was to replace the Podman preset with Microshift in tests that are currently utilizing the Podman preset.

test/integration/podman_test.go
```diff
@@ -28,15 +28,24 @@ var _ = Describe("podman preset", Serial, Ordered, Label("podman-preset"), func(
...
-			Expect(RunCRCExpectSuccess("config", "set", "preset", "podman")).To(ContainSubstring("please run 'crc setup' before 'crc start'"))
+			Expect(RunCRCExpectSuccess("config", "set", "preset", "microshift")).To(ContainSubstring("please run 'crc setup' before 'crc start'"))
 		})
 
 		It("setup CRC", func() {
-			Expect(RunCRCExpectSuccess("setup")).To(ContainSubstring("Your system is correctly setup for using CRC"))
+			if bundlePath == "" {
+				Expect(RunCRCExpectSuccess("setup")).To(ContainSubstring("Your system is correctly setup for using CRC"))
+			} else {
+				Expect(RunCRCExpectSuccess("setup", "-b", bundlePath)).To(ContainSubstring("Your system is correctly setup for using CRC"))
```
praveenkumar: You don't need to provide the bundlePath because this is link to OCP and not microshift one. so better to just `setup` and the `start` with pull secret.

praveenkumar:
@vyasgun You need to squash some of commits like 
- https://github.com/crc-org/crc/pull/4062/commits/dc623f37aefea2ea5256b58645f4446220561bf9 
- https://github.com/crc-org/crc/pull/4062/commits/93a8fc84fc86cd1a3783e0626b2221d3aa249f5b 

And 
- https://github.com/crc-org/crc/pull/4062/commits/37b663bd938294d13df0c24c927e132d5ef38a9b
- https://github.com/crc-org/crc/pull/4062/commits/887279e26b3b2056199a3774404b4d37e968697c

And
- https://github.com/crc-org/crc/pull/4062/commits/49be679a945ea24d9e5f8498ca8ab1d21ba77534
- https://github.com/crc-org/crc/pull/4062/commits/6c6ad5d2d0c5b503c87086a1622a7403aeb6ae73

And 
- https://github.com/crc-org/crc/pull/4062/commits/c87073dffe4ed0ca05f0dad2821d644be26e4e35
- https://github.com/crc-org/crc/pull/4062/commits/fe19acf92b3703efc58191502b1ea58a83405454

So that required change are part of single commit and easy for review.

vyasgun:
@praveenkumar thanks for the review! The first two commits have been deliberately separated for better clarity on why it is difficult to completely get rid of the using the defined preset. Squashed the others as suggested.

test/integration/podman_test.go
vyasgun: Changed

test/e2e/testsuite/testsuite.go
cfergeau: If we don't have similar tests in the microshift or openshift preset, then yes, doing podman->microshift makes sense

pkg/crc/config/settings_test.go
```diff
@@ -57,17 +57,6 @@ func TestCPUsValidate(t *testing.T) {
 		IsDefault: true,
 		IsSecret:  false,
 	}, cfg.Get(CPUs))
-
-	_, err = cfg.Set(Preset, crcpreset.Podman)
-	require.NoError(t, err)
```
cfergeau, line 62: The comment before the test says:
```
// Check that with the default preset, we cannot set less CPUs than the defaultCPUs
// but that it is allowed with a different preset with less requirements (podman)
```

If you remove the end of the test, what it does no longer matches its description. You can replace `Podman` with `Microshift` here and keep the rest of the code.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -176,12 +176,7 @@ func (bundle *CrcBundleInfo) GetPodmanVersion() string {
 }
 
 func (bundle *CrcBundleInfo) GetVersion() string {
-	switch bundle.GetBundleType() {
-	case crcPreset.Podman:
-		return bundle.GetPodmanVersion()
-	default:
-		return bundle.GetOpenshiftVersion()
-	}
+	return bundle.GetOpenshiftVersion()
```
cfergeau: A followup commit could remove `GetOpenshiftVersion()` or make it private since we now have 2 methods with different names which return the same thing.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -205,22 +200,16 @@ func (bundle *CrcBundleInfo) IsMicroshift() bool {
...
 		bundle.GetDiskImagePath(),
 		bundle.GetKernelPath(),
 		bundle.GetInitramfsPath(),
 	}
-	if !bundle.IsPodman() {
-		files = append(files, []string{
-			bundle.GetOcPath(),
-			bundle.GetKubeConfigPath()}...)
-	}
+	files = append(files, []string{
```
cfergeau: You can move these directly into the initialization of `files` just above:
```
	files := []string{
```

pkg/crc/machine/bundle/repository.go
```diff
@@ -55,17 +55,16 @@ func (repo *Repository) Get(bundleName string) (*CrcBundleInfo, error) {
...
-		// TODO: update this logic after major release of bundle like 4.14
-		// As of now we are using this logic to support older bundles of microshift and it need to be updated
-		// to only provide app domain route information as per preset.
-		if !crcstrings.Contains([]string{constants.AppsDomain, constants.MicroShiftAppDomain}, fmt.Sprintf(".%s", bundleInfo.ClusterInfo.AppsDomain)) {
-			return nil, fmt.Errorf("unexpected bundle, it must have %s or %s apps domain", constants.AppsDomain, constants.MicroShiftAppDomain)
-		}
-		if bundleInfo.GetAPIHostname() != fmt.Sprintf("api%s", constants.ClusterDomain) {
-			return nil, fmt.Errorf("unexpected bundle, it must have %s base domain", constants.ClusterDomain)
-		}
+	// TODO: update this logic after major release of bundle like 4.14
```
cfergeau, line 58: could be a good time to check if we can finally change this?

pkg/crc/machine/start.go
```diff
@@ -481,34 +474,6 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
...
 	}
 
-	if vm.bundle.IsPodman() {
-		// **************************
-		//  END OF PODMAN START CODE
-		// **************************
-		if err := configurePodmanProxy(sshRunner, proxyConfig); err != nil {
-			return nil, errors.Wrap(err, "Failed to configure proxy for podman")
-		}
-		if err := dns.AddPodmanHosts(instanceIP); err != nil {
```
cfergeau, line 491: You can remove this function from the `dns` package, it's no longer used after your changes.

test/e2e/testsuite/testsuite.go
cfergeau:
```
	s.Step(`^removing (podman|openshift) bundle from cache succeeds$`,
		RemoveBundleFromCache)
```
If you change `(podman | openshift)` to `openshift` here, maybe you can drop the argument to `RemoveBundleFromCache`?

cfergeau:
> The first two commits have been deliberately separated for better clarity on why it is difficult to completely get rid of the using the defined preset.

I had this comment in mind when reviewing the first 2 commits, and I must say that looking at the 2nd commit, all I see is that you readd code that you removed in the first commit. It's not obvious at all to me what the problem was with the removal, why you had to readd this code, ... I'm sure there is a good reason, but the split in 2 commits is not helping me to understand it :) The commit log is there to document the reasons making the changes needed, one-line commit logs should be the exception, not the norm, see https://www.berrange.com/tags/commit-message/ for example for some lengthy explanation / recommendations.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -200,6 +200,10 @@ func (bundle *CrcBundleInfo) IsMicroshift() bool {
 	return bundle.GetBundleType() == crcPreset.Microshift
 }
 
+func (bundle *CrcBundleInfo) IsPodman() bool {
+	return bundle.GetBundleType() == crcPreset.Podman
+}
+
```
cfergeau: You could keep some of these in the initial commit if you readd these now?

cfergeau (commented):
Out of curiosity, what happens on upgrade if you do this?
```
$ crc.old config set preset podman
$ crc.old start
$ crc.old stop
$ crc.new start
```
(`crc.old` has podman preset support, `crc.new` no longer has it).

pkg/crc/machine/bundle/metadata.go
vyasgun: thanks! I didn't catch that.

pkg/crc/machine/bundle/metadata.go
vyasgun: Good point. Removed the redundant function.

pkg/crc/machine/bundle/repository.go
vyasgun: I'm not sure what exactly needs to be done here -- can you or @praveenkumar help me with some pointers?

vyasgun:
@cfergeau Thanks a lot for the detailed review and comments! In my head, the separation of commits made sense but I don't think it's reflecting that way 😓  I have addressed most of the comments - hope the changes are okay now.

I have tested with old and new crc and if there was an existing machine with podman preset, the new crc will throw an error and ask the user to move to podman machine. [Code for this.](https://github.com/crc-org/crc/pull/4062/files#diff-0ee0833dbc0dcbd3f2822c9578f871dfb4a3662f3e12cc4d6eef36fac6524828L90)

pkg/crc/machine/bundle/repository.go
cfergeau: This resonates with what I was saying in https://github.com/crc-org/crc/pull/4062#issuecomment-2027375311

If you look at `tig blame` (or `git blame`) for this code https://github.com/crc-org/crc/blame/main/pkg/crc/machine/bundle/repository.go#L60 you get this commit https://github.com/crc-org/crc/commit/d72f292a0d9e8bbbf94aa4e041aa38e32183212f which has more details in its commit log.

Still not fully sure what needs to be changed, maybe in the line `if !crcstrings.Contains([]string{constants.AppsDomain, constants.MicroShiftAppDomain}, fmt.Sprintf(".%s", bundleInfo.ClusterInfo.AppsDomain)) {` do:
- if openshift bundle, check for `constants.AppsDomain` and not `constants.MicroshiftAppsDomain`
- if microshift bundle, only check for `constants.MicroshiftAppsDomain`

The check would be similar to the code before this change:
```
if fmt.Sprintf(".%s", bundleInfo.ClusterInfo.AppsDomain) != constants.AppsDomain {
			return nil, fmt.Errorf("unexpected bundle, it must have %s apps domain", constants.AppsDomain)
}
```

No hurries for this though, this can be done in a separate/additional PR, let's not block the PR on this.

pkg/crc/machine/bundle/metadata.go
cfergeau: Couldn't this be kept in the first commit?

pkg/crc/config/settings_test.go
```diff
@@ -57,17 +57,6 @@ func TestCPUsValidate(t *testing.T) {
 		IsDefault: true,
 		IsSecret:  false,
 	}, cfg.Get(CPUs))
-
-	_, err = cfg.Set(Preset, crcpreset.Podman)
```
cfergeau, line 61: This part of the code needs to be kept with `crcpreset.Podman` -> `crcpreset.Microshift`

pkg/crc/preflight/preflight_darwin_test.go
```diff
@@ -13,13 +13,13 @@ import (
 func TestCountConfigurationOptions(t *testing.T) {
 	cfg := config.New(config.NewEmptyInMemoryStorage(), config.NewEmptyInMemorySecretStorage())
 	RegisterSettings(cfg)
-	assert.Len(t, cfg.AllConfigs(), 13)
+	assert.Len(t, cfg.AllConfigs(), 14)
```
cfergeau: These changes should no longer be needed after fe2a9176676
Can you rebase your changes on top of `origin/main`?

.github/workflows/macos-installer.yml
```diff
@@ -44,11 +44,7 @@ jobs:
...
+        run: crc config set preset microshift
       - name: Run crc setup command
-        run: crc setup
-      - name: Wait 10 sec for the daemon to serve
-        run: sleep 10
-      - name: Run crc with podman preset
-        run: crc start
```
cfergeau, line 54: The commit log is
```
 modified github workflow

removing crc start from github test
```

This describes what the commit is doing, but this does not explain the reason why you want to make this change. My guess is that this is required as we cannot reliably run `crc start` for the microshift preset in github actions because the runners are not powerful enough, but this is just a guess, maybe you have a very different reason for this change, and I cannot know that from just the commit log and the diff. Can you explain this in the git log?

cfergeau (commented):
Still a few comments, but overall looks fine!

pkg/crc/machine/bundle/repository.go
gbraad: > let's not block the PR on this.

I agree.

gbraad:
Can we run some tests to ensure the flow works for a regular OpenShift deployment from PD? Followed by a situation where the `podman` preset was chosen in an old version, but running with this. Please share output messages to ensure user-facing information is correct for `crc config`, `crc start` and `crc delete`.

vyasgun:
The following shows the effect of the starting an older machine with podman preset with the updated crc (It will throw an error and redirect user to podman machine)
```
gvyas-mac:~ gvyas$ crc.old config view
- disable-update-check                  : true
- preset                                : podman

gvyas-mac:~ gvyas$ crc.old setup
INFO Using bundle path /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle
INFO Creating symlink for crc executable
INFO Check if Podman binary exists in: /Users/gvyas/.crc/bin/oc
INFO Checking if /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle exists
INFO Adding crc daemon plist file and loading it

gvyas-mac:~ gvyas$ crc.old start
WARN The Podman preset is deprecated and will be removed in a future release. Consider rather using a Podman Machine managed by Podman Desktop: https://podman-desktop.io
INFO Using bundle path /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle
INFO Check if Podman binary exists in: /Users/gvyas/.crc/bin/oc
INFO Loading bundle: crc_podman_vfkit_4.4.4_amd64...
INFO Creating CRC VM for Podman 4.4.4...
INFO Generating new SSH key pair...
INFO Starting CRC VM for podman 4.4.4...
INFO CRC instance is running with IP 127.0.0.1
INFO CRC VM is running
INFO Adding new bearer token for cockpit webconsole
podman runtime is now running.

  $ eval $(crc podman-env)
  $ podman COMMAND

gvyas-mac:~ gvyas$ crc.old stop
INFO Updating kernel args...
INFO Stopping the instance, this may take a few minutes...
Stopped the instance
```

```
gvyas-mac:~ gvyas$ crc.new config view
- disable-update-check                  : true
- preset                                : podman

gvyas-mac:~ gvyas$ crc.new start
INFO Using bundle path /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle
/Users/gvyas/.crc/bin/crc points to /usr/local/bin/crc.old, not to /usr/local/bin/crc.new

gvyas-mac:~ gvyas$ crc.new setup
INFO Using bundle path /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle
INFO Creating symlink for crc executable
INFO Check if Podman binary exists in: /Users/gvyas/.crc/bin/oc
INFO Checking if /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle exists
INFO Downloading bundle: /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle...
Copying blob c8fcbc8cecc6 done   |
Copying config 60347c49ff done   |
crc_podman_vfkit_4.4.4_amd64.crcbundle:  1.05 GiB / 1.05 GiB [-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
INFO Verifying the bundle signature...
INFO Uncompressing /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle
crc-podman.img:  31.00 GiB / 31.00 GiB [-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
unexpected bundle, it must have .apps-crc.testing or .apps.crc.testing apps domain

gvyas-mac:~ gvyas$ crc.new start
INFO Using bundle path /Users/gvyas/.crc/cache/crc_podman_vfkit_4.4.4_amd64.crcbundle
INFO Check if Podman binary exists in: /Users/gvyas/.crc/bin/oc
The Podman preset is deprecated and will be removed in a future release. Consider rather using a Podman Machine managed by Podman Desktop: https://podman-desktop.io
```

vyasgun:
The error is not being shown as expected in Podman Desktop:

1. image
2. image
3. image
4. image

pkg/crc/machine/bundle/repository.go
vyasgun: okay, I'll create a new PR for this

.github/workflows/macos-installer.yml
vyasgun: The main reason is that with podman preset, authentication was not required from user's end.  As per previous discussions, `crc start` with `microshift` preset is not necessary as a part of the github workflow. @praveenkumar is that correct?

pkg/crc/config/settings_test.go
```diff
@@ -57,6 +57,17 @@ func TestCPUsValidate(t *testing.T) {
...
+	_, err = cfg.Set(Preset, crcpreset.Microshift)
+	require.NoError(t, err)
+	_, err = cfg.Set(CPUs, defaultCPUs-1)
+	require.NoError(t, err)
+	assert.Equal(t, SettingValue{
+		Value:     defaultCPUs - 1,
+		Invalid:   false,
+		IsDefault: false,
+		IsSecret:  false,
+	}, cfg.Get(CPUs))
```
cfergeau: This change belongs in the previous commit

cfergeau (commented):
Overall looks good to me, still one commit log to expand a bit, and a change to move from one commit to another one

gbraad:
> The error is not being shown as expected in Podman Desktop:

Let's create new issues for this. The preset was never possible to set from the extension, so we need to decide how much this matters/should be resolved. It does not impact this PR.

.github/workflows/macos-installer.yml
cfergeau: I'm still not clear what "authentication" refers to here? Does this refer to the pull secret which is needed for `crc start` with microshift? And since we don't have a valid pull secret in gh actions, we can't run crc start?

cfergeau (approved):
Just one pending question, but the code is fine!

.github/workflows/macos-installer.yml
vyasgun: Yes, that's what I mean. The pull secret requirement means new steps have to be added in the github actions. This step was added in [this commit.](https://github.com/vyasgun/crc/commit/a4c61d9e9c2797aa6487c8f219bef26c5d045d58) I had a discussion with @praveenkumar about this and it seemed like a better idea to remove the `start` command.

gbraad:
> Just one pending question

does this impede/prevent merging? Might be great to indicate which question you mean.

cfergeau:
> > Just one pending question
> 
> does this impede/prevent merging? Might be great to indicate which question you mean.

This was the only other comment I added yesterday, and Gunjan answered it so she must have found it even if I can't find it now :-/

cfergeau:
> > > Just one pending question
> > 
> > 
> > does this impede/prevent merging? Might be great to indicate which question you mean.
> 
> This was the only other comment I added yesterday, and Gunjan answered it so she must have found it even if I can't find it now :-/

https://github.com/crc-org/crc/pull/4062#discussion_r1570850298

The commit log can't be changed after merging, so if we want to make it clearer it's now or never. I don't feel too strongly about it, if Gunjan did not change the commit log, then fine with me.

PR #4063, opened 2024-03-11
CI: Add github action for microshift-e2e

praveenkumar:


anjannath:
maybe we need to also run `crc setup` in the job?

.github/workflows/test-microshift-bundle.yml
```diff
@@ -0,0 +1,66 @@
+name: Run microshift integration
```
cfergeau, line 1: Most of this file is a copy of https://github.com/crc-org/crc/blob/main/.github/workflows/test-okd-bundle.yml , wondering if something like https://github.blog/2022-02-10-using-reusable-workflows-github-actions/ can be used to reduce the duplication.

PR #4065, opened 2024-03-12
[QE] e2e oci image refactor

### LINKED ISSUE #3932: [QE] Split win-e2e GH action by preset
Follow-up to #3875 and sister to #3931.

The idea is to get partial results quicker.

### LINKED ISSUE #4040: [QE] Review e2e image
We need to review the current e2e image to adapt it to the common stack used across qe:

- [x] Inherit from deliverest https://github.com/adrianriobo/deliverest/ 
- [x] Add the tkn spec within the project 
- [x] Added recently added memory param to customize the base memory for running tests
- [ ] Added metrics

adrianriobo:
Fixes #4040 
Fixes #3932
Depends on #4019


jsliacan:
Closing in favor of https://github.com/adrianriobo/deliverest/pull/27

images/build-e2e/README.md
```diff
@@ -1,50 +1,87 @@
 # Overview
 
-The container includes the e2e binary for all 3 platforms plus the required resources to run it.  
+This image contains the e2e suite of tests, and it is intended to run them agaisnt a target host, the logic to run the tests remotely is inherit from its base image: [deliverest](https://github.com/adrianriobo/deliverest), each platform and os has its own image.
```
praveenkumar: `run them agaisnt` => `run them against`

images/build-e2e/README.md
```diff
@@ -1,50 +1,87 @@
...
-**RESULTS_FILE**:*(Optional). File name for results xunit results. Default value: e2e.*  
-**CLEANUP_HOME**:*(Optional). Cleanup crc home folder or keep as it is to run test.*  
-**E2E_TAG_EXPRESSION**:*(Optional). Define e2e tag expression to select tests. If empty all tests available for the platform will be executed.*  
+* crc-e2e/run.ps1 ... (windows)
+* crc-e2e/run.sh ... (linux/darwin)
 
-## Samples
+And the execution is customized by the params addded, available params:
+
+* bundleLocation When testing a custom bundle we should pass the paht on the target host
```
praveenkumar: `pass the paht` => `pass the path`

images/build-e2e/README.md
```diff
@@ -1,50 +1,87 @@
...
+* crcMemory Customize memory for the cluster to run the tests
+
+### Windows amd64
+
+```bash
+podman run --rm -d --name crc-e2e-windows \
+    -e TARGET_HOST=XXXX \
+    -e TARGET_HOST_USERNAME=XXXX \
+    -e TARGET_HOST_KEY_PATH=/data/id_rsa \
+    -e TARGET_FOLDER=crc-e2e \
```
praveenkumar, line 27: Does this `TARGET_FOLDER` override the `ASSETS_FOLDER` of the image? because as part of container file following is what we mention
```
COPY images/build-e2e/lib/${OS}/* ${ASSETS_FOLDER}/
```
and during the build of this container image from makefile I don't see passing a ASSETS_FOLDER so not sure how `run.ps1/run.sh` lands to this `TARGET_FOLDER`.

adrianriobo: no, TARGET_FOLDER takes effect on run time, and it makes reference to the folder path where we want to copy the required resources on the target host

The ASSETS_FOLDER is defined on build time, and it is being used by the base image `quay.io/rhqp/deliverest:v0.0.5`

Deliverest image has some common bash functions to ssh / scp and implement the logic:

* Copy everything on the container `ASSETS_FOLDER` to the remote target `TARGET_FOLDER`
* Run on the remote target the command of the container
* Copy back the folder from the target host at `TARGET_RESULTS` to the container path defined by `OUTPUT_FOLDER`
* Remove the `TARGET_FOLDER` on the target host

This logic is being used on almost every test execution done against a remote target, this is why it has been abstracted into Deliverest and now each image inheriting from it just need to define its specific workload per platform / arch

praveenkumar: Okay so this is part of https://github.com/adrianriobo/deliverest/blob/main/entrypoint.sh#L37 one.

adrianriobo: yeah exactly, at some point I need to move those small projects to some org

.github/workflows/windows-artifacts.yml
```diff
@@ -23,85 +23,83 @@ jobs:
...
+    - name: Build qe oci images ${{matrix.os}}-${{matrix.arch}}
+      run: |
+        # e2e
+        CRC_E2E_IMG_VERSION=gh ARCH=${{matrix.arch}} OS=${{matrix.os}} make containerized_e2e
+        podman save -o ${{ env.IMAGE_NAME_E2E }}-${{matrix.os}}-${{matrix.arch}}.tar \
+          quay.io/crcont/${{ env.IMAGE_NAME_E2E}}:gh-${{matrix.os}}-${{matrix.arch}}
+        # integration
+        CRC_INTEGRATION_IMG_VERSION=gh ARCH=${{matrix.arch}} OS=${{matrix.os}} make containerized_integration
+        podman save -o ${{ env.IMAGE_NAME_INTEGRATION }}-${{matrix.os}}-${{matrix.arch}}.tar \
+          quay.io/crcont/${{ env.IMAGE_NAME_INTEGRATION}}:gh-${{matrix.os}}-${{matrix.arch}}
```
praveenkumar, line 76: So we are building the container image for e2e and integration and then we are saving the image for same using quay.io using `podman save` why?

.github/workflows/windows-e2e.yml
```diff
@@ -7,222 +7,23 @@ on:
...
-          podman logs -f windows-destroy
+        qe-type: ['e2e','integration']
+        preset: ['openshift', 'microshift', 'all']
+        exclude:
+        - qe-type: 'e2e'
+          preset: 'all'
+        - qe-type: 'integration'
+          preset: 'openshift'
+        - qe-type: 'integration'
+          preset: 'microshift'
```
praveenkumar, line 23: so does that mean `e2e openshift` and `e2e microshift` but `integration all` ?

.github/workflows/windows-artifacts.yml
adrianriobo: We are building the images but not pushing them, instead we are saving them as flat files which are uploaded as artifacts to the action.

Then on the workflowrun we are downloading those artifacts and load them as images back

.github/workflows/windows-e2e.yml
adrianriobo: yeah I did not investigate if integration has labels to allow me to differentiate, it that is the case I can change that back and separate the execution. In any case if possible I will do it on a following I do not want to mess this one more.

As you see the e2e and integration images build are on the windows action which does not make sense to me, I would prefer having a buidler action which builds everything we need (with different jobs, ,as they will run in parallel) and the use that builder action as the trigger for the worklflow run where tests will be executed. But again my idea is having a following for this refactor

.github/workflows/windows-artifacts.yml
praveenkumar: Ack

images/build-e2e/README.md
adrianriobo: changed! (also on integration README)

images/build-e2e/README.md
adrianriobo: changed! (also on integration README)

praveenkumar:
@adrianriobo Because of this PR looks like all the openshift CI tests are failing.

https://github.com/crc-org/crc/blob/main/images/openshift-ci/Dockerfile#L12-L13 ?

The path in the makefile was `$(BUILD_DIR)/linux-amd64/e2e.test` which changed to `$(BUILD_DIR)/e2e.test` and same for integration test so better to update this file also.

adrianriobo:
Now it should be good I refactored a bit the e2e and integration building blocks on the Makefile to accommodate the changes

adrianriobo:
Wait for my ack on test to have another look I will let you know

adrianriobo:
@praveenkumar it is working now you can review if you want the latest changes

PR #4066, opened 2024-03-12
Increase memory resource from 9126 to 10752

praveenkumar:
With Openshift-4.15, OVN-K require extra 1.5G mem resource we need to increase that otherwise cluster doesn't comeup correctly.


cfergeau:
`TestCountConfigurationOptions` and `TestCountPreflights` is broken for some reason.

Fwiw, I've found strongunits https://github.com/crc-org/vfkit/blob/main/pkg/config/config.go#L40-L46 relatively useful when dealing with MiB/... to know in which unit a value is.

I assume if someone set the cluster memory to 10000 previously, then they will get a failure after the upgrade, and will have to unset or raise the cluster memory size?

gbraad (commented):
The difference is 1536M. 
Let's see how this will work out...

/lgtm for testing

praveenkumar:
> `TestCountConfigurationOptions` and `TestCountPreflights` is broken for some reason.

Yeah, I will put another commit to fix it.
 
> I assume if someone set the cluster memory to 10000 previously, then they will get a failure after the upgrade, and will have to unset or raise the cluster memory size?

Yeah, that's right.

gbraad:
What is the status of this?

praveenkumar:
@gbraad looks like we have to go with this because swap is not working as expected https://github.com/crc-org/snc/issues/861 have all the details and https://github.com/crc-org/snc/issues/861#issuecomment-2006906709 is where we filled the jira.

gbraad (approved):

praveenkumar:
/hold

We have it as soon as update the OCP bundle to 4.15

anjannath (commented):
need to also update the `e2e` test feature about memory config.. currently test is failing with the following:
```
 checks (settings) # test/e2e/features/config.feature:9
    When setting config property "memory" to value "9217" succeeds # test/e2e/features/config.feature:10
      Error: command 'crc config set memory 9217', expected to succeed, exited with exit code: 1
```

Issue #4023, opened 2024-02-07
Confusing error message when segment identification hash is malformed (telemetry)


circadi:
### General information

  * OS:  Windows 10.0.19044.3930
  * Hypervisor: Hyper-V
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop 

## CRC version
CRC version: 2.28.0+e1f42d
OpenShift version: 4.13.14
Podman version: 4.4.4

## CRC status
Machine does not exist. Use 'crc start' to create it

## CRC config
```
- consent-telemetry                     : no
- cpus                                  : 8
- disk-size                             : 120
- ingress-http-port                     : 8088
- memory                                : 20480
```
## Host Operating System
```
OS Name:                   Microsoft Windows 10 Enterprise
OS Version:                10.0.19044 N/A Build 19044
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Member Workstation
OS Build Type:             Multiprocessor Free
Registered Owner:
Registered Organization:
Product ID:                00329-00000-00003-AA941
Original Install Date:     2/8/2023, 9:19:47 AM
System Boot Time:          2/6/2024, 11:47:28 AM
System Manufacturer:       HP
System Model:              HP ZBook Fury 15.6 inch G8 Mobile Workstation PC
System Type:               x64-based PC
Processor(s):              1 Processor(s) Installed.
                           [01]: Intel64 Family 6 Model 141 Stepping 1 GenuineIntel ~2496 Mhz
BIOS Version:              HP T95 Ver. 01.15.00, 10/11/2023
Windows Directory:         C:\Windows
System Directory:          C:\Windows\system32
Boot Device:               \Device\HarddiskVolume2
System Locale:             en-us;English (United States)
Input Locale:              en-us;English (United States)
Time Zone:                 (UTC-05:00) Eastern Time (US & Canada)
Total Physical Memory:     65,200 MB
Available Physical Memory: 51,764 MB
Virtual Memory: Max Size:  74,928 MB
Virtual Memory: Available: 60,113 MB
Virtual Memory: In Use:    14,815 MB
Page File Location(s):     C:\pagefile.sys
Logon Server:              N/A
Hotfix(s):                 18 Hotfix(s) Installed.
                           [01]: KB5033909
                           [02]: KB5003791
                           [03]: KB5012170
                           [04]: KB5019180
                           [05]: KB5034122
                           [06]: KB5020372
                           [07]: KB5022924
[...]

PR #4084, opened 2024-03-25
telemetry: do not return an error when identify hash file is malformed

redbeam:
**Fixes:** Issue #4023

## Solution/Idea

A malformed identify hash file should no longer trigger an error when reading it. Now there is a warning displayed and the hash file is treated like it doesn't exist yet. Therefore, it will be regenerated with the correct content.

## Proposed changes

1. Add a warning log to the console with a description of what is wrong and what will happen.
2. Remove error from the return.

## Testing

When the file does not exist yet, no warning should be shown. If it is malformed (does not contain exactly 8 bytes), the warning should show and after the program exits, the file should contain the correct hash.

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -140,7 +140,8 @@ func readIdentifyHash(identifyHashPath string) (uint64, error) {
 	}
 
 	if length := len(cachedHashBytes); length != 8 {
-		return 0, fmt.Errorf("Got %d bytes hash, expected 8", length)
+		logging.Warnf("Got %d bytes hash (expected 8), identify hash file will be regenerated", length)
+		return 0, nil
 	}
 
 	return binary.LittleEndian.Uint64(cachedHashBytes), nil


pkg/crc/segment/segment.go
```diff
@@ -140,7 +140,8 @@ func readIdentifyHash(identifyHashPath string) (uint64, error) {
 	}
 
 	if length := len(cachedHashBytes); length != 8 {
-		return 0, fmt.Errorf("Got %d bytes hash, expected 8", length)
+		logging.Warnf("Got %d bytes hash (expected 8), identify hash file will be regenerated", length)
```
gbraad, line 143: Does the actual generation be logged too? If so, the text would be better to be written as:

```patch
+     logging.Warnf("Got %d bytes hash (expected 8), identify hash file will have to be regenerated", length)
```

or `"..., identify hash needs to be regenerated"`

as the function itself will not do this, but might imply this when you read the log.

redbeam: >  logging.Warnf("Got %d bytes hash (expected 8), identify hash file will have to be regenerated", length)

I did not want to use this wording as it would (in my opinion) imply that an action from the user is needed. The file is generated automatically as soon as possible by other functions.

gbraad (approved):

cfergeau (approved):

gbraad:
Thank you for your first contribution!

Issue #3966, opened 2024-01-19
[Spike] Setting directory for .crc folder


Maurycy-Krzeminski:
Is there consideration to use XDG base directory, or maybe using some other named env variable that would allow to set where .crc folder should be?

From what i saw [issue 220 about XDG](https://github.com/crc-org/crc/issues/220 ) was closed.

anjannath:
If we change the crc home dir, we should also do automated migration of the config file to the new location. I think XDG base directory as the default and allowing users to override with $CRC_HOME_DIR both should be added

vyasgun:
Please correct me if I am wrong -- The `.crc` directory stores the cache, machine info, and socket for communicating with the crc daemon and is used in multiple parts of the lifecycle (setup, start). Just setting the environment variable can create a discrepancy if the user sets it multiple times 
```
export CRC_HOME_DIR=/tmp/crc1
crc start
crc stop
export CRC_HOME_DIR=/tmp/crc2
```
It is difficult to keep track of all the values the variable can take and migrating over and over does not seem like a feasible idea. ([I made a draft PR for this](https://github.com/crc-org/crc/pull/4088/commits/d03c4f804eeee3caaa0c6dac0727b3e55149c298))

My suggestion is to use crc config to store the path of the .crc directory which can be set and unset using the `crc config` command. This makes validation easier, it's more explicit, and easier to keep track of. Also, it reinforces the idea that we cannot have multiple `.crc` directories each having its own cache, socket etc.

praveenkumar:
> My suggestion is to use crc config to store the path of the .crc directory which can be set and unset using the `crc config` command. This makes validation easier, it's more explicit, and easier to keep track of. Also, it reinforces the idea that we cannot have multiple `.crc` directories each having its own cache, socket etc.

In this case also user can do following
```
crc config set base-directory foo
crc start
crc stop
crc config unset base-directory
```

One thing we can do as part config option is have a callback function which check if current set directory already have content or not, if it have let user know and ask for override otherwise carry on with provided option?

vyasgun:
I was thinking we could migrate from the existing base directory (if nothing is set, we would use the default `$HOME/.crc` to the newly set one. (Unset could trigger reverse migration from the custom location to the default location).

cfergeau:
> My suggestion is to use crc config to store the path of the .crc directory which can be set and unset using the crc config command. This makes validation easier, it's more explicit, and easier to keep track of.

[...]

PR #4088, opened 2024-04-01
[Spike] Use `crc config` to use custom location for .crc directory

vyasgun:
**Fixes:** Issue #3966 

## Solution/Idea
The location for .crc directory is made configurable for the user through `crc config`.
Currently, this PR will use the value set in the config as the new base directory.

## Proposed changes

- Add a new config key called `crc-dir`
- Initialise Crc base directory variables based on the saved config
- New config will take effect after rerunning `crc setup`

## Testing

```
gvyas-mac:crc gvyas$ crc config view

gvyas-mac:crc gvyas$ crc config set crc-dir ~/test1
Changes to configuration property 'crc-dir' are only applied during 'crc setup'.
Please run 'crc setup' for this configuration to take effect.

gvyas-mac:crc gvyas$ crc config view
- crc-dir                               : /Users/gvyas/test1

```
```
gvyas-mac:crc gvyas$ crc setup
INFO Using bundle path /Users/gvyas/test1/.crc/cache/crc_microshift_vfkit_4.15.3_amd64.crcbundle 
INFO Caching crc-admin-helper executable          
INFO Using root access: Changing ownership of /Users/gvyas/test1/.crc/bin/crc-admin-helper-darwin 
Password:
INFO Using root access: Setting suid for /Users/gvyas/test1/.crc/bin/crc-admin-helper-darwin 
INFO Creating symlink for crc executable          
INFO Check if Podman binary exists in: /Users/gvyas/test1/.crc/bin/oc 
INFO Setting up virtualization with vfkit         
INFO Checking if CRC bundle is extracted in "/Users/gvyas/test1/.crc/cache/crc_microshift_vfkit_4.15.3_amd64.crcbundle" 
INFO Checking if /Users/gvyas/test1/.crc/cache/crc_microshift_vfkit_4.15.3_amd64.crcbundle exists 
INFO Adding crc daemon plist file and loading it  
```
```
gvyas-mac:crc gvyas$ crc start
INFO Using bundle path /Users/gvyas/test1/.crc/cache/crc_microshift_vfkit_4.15.3_amd64.crcbundle 
INFO Check if Podman binary exists in: /Users/gvyas/test1/.crc/bin/oc 
INFO Loading bundle: crc_microshift_vfkit_4.15.3_amd64... 
INFO Creating CRC VM for MicroShift 4.15.3...     
INFO Generating new SSH key pair...               
INFO Starting CRC VM for microshift 4.15.3...     
INFO CRC instance is running with IP 127.0.0.1    
INFO CRC VM is running                            
INFO Check internal and public DNS query...       
INFO Check DNS query from host...                 
INFO Starting Microshift service... [takes around 1min] 
INFO Check internal and public DNS query...       
INFO Check DNS query from host...                 
INFO Starting Microshift service... [takes around 1min] 
INFO Waiting for kube-apiserver availability... [takes around 2min] 
INFO Adding microshift context to kubeconfig...   
Started the MicroShift cluster.

  $ eval $(crc oc-env)
  $ oc COMMAND
```

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -37,6 +37,7 @@ const (
 	EmergencyLogin           = "enable-emergency-login"
 	PersistentVolumeSize     = "persistent-volume-size"
 	EnableBundleQuayFallback = "enable-bundle-quay-fallback"
+	CrcDir                   = "crc-dir"
 )
 
 func RegisterSettings(cfg *Config) {
@@ -141,13 +142,18 @@ func RegisterSettings(cfg *Config) {
 
 	cfg.AddSetting(EnableBundleQuayFallback, false, ValidateBool, SuccessfullyApplied,
 		"If bundle download from the default location fails, fallback to quay.io (true/false, default: false)")
+	cfg.AddSetting(CrcDir, constants.GetHomeDir(), validateDirectory, RequiresCRCSetup,
+		"Location for .crc")
 
-	if err := cfg.RegisterNotifier(Preset, presetChanged); err != nil {
+	if err := cfg.RegisterNotifier(Preset, settingChanged); err != nil {
 		logging.Debugf("Failed to register notifier for Preset: %v", err)
+	}
+	if err := cfg.RegisterNotifier(CrcDir, settingChanged); err != nil {
+		logging.Debugf("Failed to register notifier for .crc directory: %v", err)
 	}
 }
 
-func presetChanged(cfg *Config, _ string, _ interface{}) {
+func settingChanged(cfg *Config, _ string, _ interface{}) {
 	UpdateDefaults(cfg)
 }
 
@@ -165,6 +171,10 @@ func defaultBundlePath(cfg Storage) string {
 
 func GetPreset(config Storage) preset.Preset {
 	return preset.ParsePreset(config.Get(Preset).AsString())
+}
+
+func GetConfigDir(config Storage) string {
+	return config.Get(CrcDir).AsString()
 }
 
 func defaultNetworkMode() network.Mode {
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -164,7 +164,8 @@ func GetHomeDir() string {
 }
 
 // EnsureBaseDirectoriesExist creates ~/.crc, ~/.crc/bin and ~/.crc/cache directories if it is not present
-func EnsureBaseDirectoriesExist() error {
+func EnsureBaseDirectoriesExist(configDir string) error {
+	initialiseAllDirectories(configDir)
 	baseDirectories := []string{CrcBaseDir, MachineCacheDir, CrcBinDir}
 	for _, baseDir := range baseDirectories {
 		err := os.MkdirAll(baseDir, 0750)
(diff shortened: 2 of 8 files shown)


cfergeau:
This PR is related to https://github.com/crc-org/crc/issues/3966

gbraad:
> This PR is related to https://github.com/crc-org/crc/issues/3966

That should be in the first message, as currently it is just the template.

vyasgun:
@gbraad @cfergeau so sorry about the confusion - It was a WIP so I didn't add details. 😓  Updated the PR with the details and the updated code.

gbraad:
Reading the issue/PR here I wonder how this could work. as the config itself lives inside the .crc folder.

 1. we can store a path inside a ~/.crcpath
 1. or we use a shell/env variable


Doing a `crc config set` makes sense for a partial relocation, like only `cache`, and the `machines` folder.

Using an environment variable also makes this a more deliberate choice, as there are side effects for sure.

gbraad:
> so sorry about the confusion - It was a WIP so I didn't add details.

Thinking out loud is always helpful, especially in the case when, like in this issue, the situation is not clearly described. It becomes more of a trial and error. Almost like a spike issue. Writing down some things before working on this might help with the actual implementation. I, for example, immediately think of a chicken-egg problem when using this as a configurable from the config itself, but also, what about the long-running processes? They need to be restarted? Also, for a `start` this can be enforced from a validation with a pre-flight. But for commands like `status` and `stop` this only can operate when the VM folder exists, though there is no current mechanism to deal with this.

gbraad:
I haven't looked at the implementation yet, but what would happen when the location is changed after the VM is already created? One of the things we do not perform before the start is to check if all the VM files are available. This might be something that has to exist regardless of this change, but due to the possibility to change the `crc-dir` it is likely this would get triggered.

pkg/crc/config/settings.go
```diff
@@ -37,6 +37,7 @@ const (
 	EmergencyLogin           = "enable-emergency-login"
 	PersistentVolumeSize     = "persistent-volume-size"
 	EnableBundleQuayFallback = "enable-bundle-quay-fallback"
+	CrcDir                   = "crc-dir"
```
gbraad, line 40: Is this the config directory or crc directory?

pkg/crc/config/settings.go
```diff
@@ -167,6 +173,10 @@ func GetPreset(config Storage) preset.Preset {
 	return preset.ParsePreset(config.Get(Preset).AsString())
 }
 
+func GetConfigDir(config Storage) string {
+	return config.Get(CrcDir).AsString()
```
gbraad, line 177: Related to the question: "Is this the config directory or crc directory?"

pkg/crc/constants/constants.go
```diff
@@ -175,6 +176,26 @@ func EnsureBaseDirectoriesExist() error {
 	return nil
 }
 
+func initialiseAllDirectories(crcDir string) {
```
gbraad, line 179: This is not just about directories, as there are also sockets, and configuration files.

gbraad:
Overall the changes make sense, especially to comnfirm the permissions with the validate. However, as mentioned above, more can be at fault. Edge cases exist when a VM is already created; in that case a start could result in an error of an already existing VM by name, but not in the 'new' machines folder.

Please try some of these scenarios to see what I mean.

A possible solution would be to prevent changing this folder when a VM is currently registered.

pkg/crc/validation/validation.go
```diff
@@ -168,6 +168,16 @@ func ValidatePath(path string) error {
 	return nil
 }
 
+// ValidateDirectory checks if provided path exists and has sufficient permissions
+func ValidateDirectory(path string) error {
```
gbraad, line 172: This is great for re-use

vyasgun:
The objective of this spike was to test whether the `crc config` can be used to configure a custom `.crc` location. The conclusion is that it is not something that will solve the problem and could potentially cause unexpected issues in certain edge cases:

- If VM has already been started, modifying `.crc` location could result in inconsistent behaviour.
- The base case (for the default directory to open from) has to be defined outside of `.crc` folder as it will result in a chicken and egg problem.
- The major request to use a custom directory mainly comes from having a custom location to store the larger files such as cache and machine configs. Using more straightforward solution such as using environment variable related configuration may be more helpful to accomplish this.
- The original request for using XDG base directory env variables in certain cases may be solved using [this commit](https://github.com/crc-org/crc/commit/c6f8c78d54837b1a05053ee8ba8d486fa4375cc7#diff-351617b3a8d07ea69e3c3430ca45061b17effc84aba09b482e8b8d79ee696a4e): 

Next steps:
- Redefine the problem and come up with a solution that enables the user to manage their storage more effectively.

cfergeau:
> - If VM has already been started, modifying .crc location could result in inconsistent behaviour.
> - Using more straightforward solution such as using environment variable related configuration may be more helpful to accomplish this.

Setting the location using `crc config set` allows us to control when it's allowed/not allowed. Basically, if a VM exists or is running, then changing the location is forbidden, otherwise, why not.
An environment variable would complicate things as crc could only check if it's allowed *after* it has changed, and without knowing what was .crc location before the location change.

PR #4091, opened 2024-04-01
Avoid using syscall.Sysctl to get macOS version

### LINKED ISSUE #4086: [BUG] Supported MacOS version check on crc setup incorrect
### General information

  * OS: macOS 
  * Hypervisor: unsure
  * Did you run `crc setup` before starting it (Yes/No)? Tes
  * Running CRC on: Intel Mac Pro 2019 Xeon

## CRC version
```bash
CRC version: 2.33.0+c43b17
OpenShift version: 4.14.12
Podman version: 4.4.4
```
  
## CRC status
```bash
DEBU CRC version: 2.33.0+c43b17
DEBU OpenShift version: 4.14.12
DEBU Podman version: 4.4.4
DEBU Running 'crc status'
CRC VM:          Stopped
OpenShift:       Stopped (v4.14.8)
RAM Usage:       0B of 0B
Cache Usage:     144.9GB
Cache Directory: /Users/trees/.crc/cache
```

## CRC config
```bash
- cpus                                  : 12
- disk-size                             : 200
- memory                                : 49152
```

## Host Operating System
```bash
ProductName:		macOS
ProductVersion:		14.4.1
BuildVersion:		23E224
```

Intel Mac Pro

### Steps to reproduce

  1. Run crc setup
  2. 
  3. 
  4. 

### Expected
No warning about MacOS version being old

### Actual
```
➜  ~ crc setup
INFO Using bundle path /Users/trees/.crc/cache/crc_vfkit_4.14.12_amd64.crcbundle
**WARN This version of macOS is going to be unsupported for CRC, Please update to macOS 13 or newer**
INFO Checking if /Users/trees/.crc/cache/crc_vfkit_4.14.12_amd64.crcbundle exists
```

### Logs

Before gather the logs try following if that fix your issue
```bash
$ crc delete -f
$ crc cleanup
$ crc setup
$ crc start --log-level debug
```

Please consider posting the output of `crc start --log-level debug`  on http://gist.github.com/ and post the link in the issue.

praveenkumar:
Looks like `syscall.Sysctl` is not reliable when it comes to get macOS version on different arch (Intel vs silicon) because of [0]. This PR uses `sw_vers` cmd result which seems to be consistent on both platform.

[0] https://github.com/golang/go/issues/58722

fixes: #4086

### DIFF
--- a/pkg/os/darwin/release_info.go
+++ b/pkg/os/darwin/release_info.go
@@ -5,10 +5,11 @@ package darwin
 import (
 	"fmt"
 	"sync"
-	"syscall"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
+	"github.com/crc-org/crc/v2/pkg/os"
+	"github.com/crc-org/crc/v2/pkg/strings"
 	"github.com/pkg/errors"
 )
 
@@ -20,8 +21,12 @@ var (
 
 func AtLeast(targetVersion string) (bool, error) {
 	once.Do(func() {
-		macCurrentVersion, errGettingVersion = syscall.Sysctl("kern.osproductversion")
-		logging.Debugf("kern.osproductversion is: %s", macCurrentVersion)
+		// syscall.Sysctl("kern.osproductversion") is not providing the consistent version info for intel and silicon macs
+		// so using sw_vers -productVersion to get the version info consistently
+		// https://github.com/golang/go/issues/58722
+		macCurrentVersion, _, errGettingVersion = os.RunWithDefaultLocale("sw_vers", "-productVersion")
+		macCurrentVersion = strings.TrimTrailingEOL(macCurrentVersion)
+		logging.Debugf("sw_vers -productVersion is: %s", macCurrentVersion)
 	})
 	if errGettingVersion != nil {
 		return false, errGettingVersion


gbraad:
This should be verified with @reestr

gbraad (approved):
/lgtm

however, this needs to be tested as this seems to be system-specific.

pkg/os/darwin/release_info.go
```diff
@@ -20,8 +21,9 @@ var (
 
 func AtLeast(targetVersion string) (bool, error) {
 	once.Do(func() {
-		macCurrentVersion, errGettingVersion = syscall.Sysctl("kern.osproductversion")
-		logging.Debugf("kern.osproductversion is: %s", macCurrentVersion)
+		macCurrentVersion, _, errGettingVersion = os.RunWithDefaultLocale("sw_vers", "--productVersion")
```
anjannath: typo in the flag, this should be `sw_vers -productVersion` with single `-`

running on macOS 12 i get the following o/p
```
$ sw_vers --productVersion
Usage: sw_vers [-productName|-productVersion|-buildVersion]

$ sw_vers -productVersion
12.7.4
```

praveenkumar: As per man page of `sw_vers` , it contain `--` instead `-` . But looks like even with `-` it provide same info :( 

```
SW_VERS(1)                                                             General Commands Manual                                                            SW_VERS(1)

NAME
     sw_vers – print macOS system version information

SYNOPSIS
     sw_vers
     sw_vers --productName
     sw_vers --productVersion
     sw_vers --productVersionExtra
     sw_vers --buildVersion
```

pkg/os/darwin/release_info.go
```diff
@@ -20,8 +21,9 @@ var (
 
 func AtLeast(targetVersion string) (bool, error) {
 	once.Do(func() {
-		macCurrentVersion, errGettingVersion = syscall.Sysctl("kern.osproductversion")
-		logging.Debugf("kern.osproductversion is: %s", macCurrentVersion)
+		macCurrentVersion, _, errGettingVersion = os.RunWithDefaultLocale("sw_vers", "-productVersion")
+		macCurrentVersion = strings.TrimTrailingEOL(macCurrentVersion)
+		logging.Debugf("sw_vers --productVersion is: %s", macCurrentVersion)
```
cfergeau: This needs a comment explaining why we cannot use `syscall.Sysctl("kern.osproductversion")` directly / a link to the go issue.

pkg/os/darwin/release_info.go
```diff
@@ -20,8 +21,9 @@ var (
 
 func AtLeast(targetVersion string) (bool, error) {
 	once.Do(func() {
-		macCurrentVersion, errGettingVersion = syscall.Sysctl("kern.osproductversion")
-		logging.Debugf("kern.osproductversion is: %s", macCurrentVersion)
+		macCurrentVersion, _, errGettingVersion = os.RunWithDefaultLocale("sw_vers", "-productVersion")
+		macCurrentVersion = strings.TrimTrailingEOL(macCurrentVersion)
```
cfergeau, line 28: Is there an EOL to strip?

cfergeau (changes requested):
Bit of documentation for the hack in the source code would not hurt to prevent people from "simplifying" the code in the future.

pkg/os/darwin/release_info.go
praveenkumar: yes.

pkg/os/darwin/release_info.go
praveenkumar: mentioned in the commit log, but sure can add comment here.

pkg/os/darwin/release_info.go
gbraad: if not, does it harm?

cfergeau: if there is no EOL, the person noticing this line is unneeded 3 years from now will have no idea if it was needed at the time, and if removing it will break older systems, or if it was already unneeded at the time it was added.

cfergeau (approved):

Issue #4085, opened 2024-03-26
Add NoopClient to segment.go to allow running without telemetry in case of errors


redbeam:
Discussed in #4023

Currently, any error in the initialization of segment client prevents the program to run. Since it only deals with telemetry collection, it should not block other, more important parts of the program.

I propose implementing a "mock" segment client (NoopClient) that will be returned if the initialization of the real segment client fails. This way, the execution of the command is not interrupted by any error in telemetry. If this happens, the mock client will return a warning stating that telemetry will not be available in current session and its methods will do nothing.

gbraad:
> its methods will do nothing

logging?

redbeam:
Yes, the methods can log their "inactivity". But I'm not sure if every call needs to do that. Maybe only the first would be enough.

gbraad:
It depends on the state:

We might want to understand the `NoopClient` received the call and continues to execute the regular method. Just to be sure, we have full traceability when an error occurs. Logs are rotated anyway.

PR #4100, opened 2024-04-03
telemetry: add no-op functionality to SegmentClient to make sure errors in initialization are not fatal

redbeam:
**Fixes:** Issue #4085

A new field has been added to Client struct signalling whether the client has been initialized properly. Default value is `false` and it is set to `true` only if the initialization went as expected.
Methods on Client check if this field is true before progressing further.

In case of any error during initialization, an empty Client is returned (and its methods do nothing).

### DIFF
--- a/pkg/crc/segment/segment.go
+++ b/pkg/crc/segment/segment.go
@@ -39,17 +39,24 @@ type Client struct {
 	identifyHashPath string
 }
 
+var NoopClient = &Client{}
+
 func NewClient(config *crcConfig.Config, transport http.RoundTripper) (*Client, error) {
-	return newCustomClient(config, transport,
+	client, err := newCustomClient(config, transport,
 		filepath.Join(constants.GetHomeDir(), ".redhat", "anonymousId"),
 		filepath.Join(constants.CrcBaseDir, "segmentIdentifyHash"),
 		analytics.DefaultEndpoint)
+	if err != nil {
+		// Return empty client (noop client) -- telemetry will not be available
+		return NoopClient, err
+	}
+	return client, nil
 }
 
 func newCustomClient(config *crcConfig.Config, transport http.RoundTripper, telemetryFilePath, identifyHashFilePath, segmentEndpoint string) (*Client, error) {
 	userID, err := getUserIdentity(telemetryFilePath)
 	if err != nil {
-		return nil, err
+		return NoopClient, err
 	}
 	client, err := analytics.NewWithConfig(WriteKey, analytics.Config{
 		Endpoint: segmentEndpoint,
@@ -60,12 +67,12 @@ func newCustomClient(config *crcConfig.Config, transport http.RoundTripper, tele
 		Transport: transport,
 	})
 	if err != nil {
-		return nil, err
+		return NoopClient, err
 	}
 
 	identifyHash, err := readIdentifyHash(identifyHashFilePath)
 	if err != nil {
-		return nil, err
+		return NoopClient, err
 	}
 
 	return &Client{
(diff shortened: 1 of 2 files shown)


pkg/crc/segment/segment.go
```diff
@@ -37,13 +37,20 @@ type Client struct {
 
 	identifyHash     uint64
 	identifyHashPath string
+
+	isInitProperly bool
```
cfergeau: `isInitProperly`  is a bit 'ungrammatical', but so is `isInitedProperly` :-/
For what it's worth, I think I would store the error we got at init time instead of just a boolean, so this could be `initError` (and actually, even as a boolean it would be fine).

Or it might be possible to deal with this without a boolean at all, just:
```
var NoopClient = &Client{}
```
then in error cases, you return `return NoopClient, err`
and the `if !c.isInitProperly {` checks become `if c == NoopClient`

These are just various suggestions, I'm fine with your approach with a boolean, though I'd prefer if we slightly improved the naming.

cfergeau (commented):
Looks good to me, just not sure about the `isInitProperly` name.

pkg/crc/segment/segment.go
redbeam: I like the approach with storing the error. Though, it seems that it would introduce more changes to the code than the boolean way. Maybe we can discuss the name of the variable?
Thanks for the review!

gbraad: `hasFinishedInit` would be more grammatically correct, however storing this as a state is not relevant. The rest of the execution should not care about the error state. I prefer the `c == NoopClient` checks (more in line with proper OOP practices) => more clearly indicating the state/situation without comments and allows reuse for situations where we might want to force a `NoopClient`, like CI (testing the output, but not recording). So, why is there not a `NoopClient` mock-implementation?

gbraad (changes requested):
Would like to understand what it takes not to keep state. The `NoopClient` should work without specific knowledge to what happened... and if needed, a typecheck would be preferred.

pkg/crc/segment/segment.go
redbeam: I agree with your remarks Gerard. I changed the implementation to use NoopClient.

gbraad:
I think something went wrong during rebase, as I see some stuff about snyk and the `sw_vers` fix. Do not believe they are pat of this PR. 

However, the rest looks good to me,

pkg/crc/segment/segment.go
gbraad: There are now commits related to a fix from Praveen. I do not believe they are part of this PR and might be due to a rebase?

redbeam: Yes, that was a mistake on my part. I reverted the branch and pushed only the following commit. Now there should only be my two commits.

cfergeau (commented):
Looks good to me, I'd squash the 2 commits though, we don't need to have the `isInitProperly` version of the code in the upstream git history.
 The shortlog for both commits overflows on github, it's supposed to be kept short, with a longer description in the longer commit log. 
 ```
telemetry: add no-op functionality

Currently, when telemetry initialization fails, this is a hard failure for the `crc` command and daemon, and they exit.
This commit handles this more gracefully, when there's a telemetry initialization failure, an empty `NoOp` `Client` instance is returned, and all `Client` methods check for this `NoOp` instance and return early if needed.
```

gbraad:
> The shortlog for both commits overflows on github, it's supposed to be kept short, 

About 20 to 30 characters.

redbeam:
@gbraad @cfergeau Thanks for the suggestions. Everything should be fixed now.

PR #4114, opened 2024-04-09
Update disk resize to work with `lvms-storage-operator`/topolvm

### LINKED ISSUE #4097: create `LVMCluster` resource on `crc start`
after https://github.com/crc-org/snc/issues/867 openshift bundles will have the LVMS operator installed in the cluster, to make use of it we need to create the `LVMCluster` crd resource provided by the operator to have the necessary storage class, CSI drivers setup to handle dynamic provisioning of the PVCs

to create the LVMCluster resource we also need to provide the disk to be used for creating the lvm volumes, so we need to also create a second disk image for the vm.

```yaml
apiVersion: lvm.topolvm.io/v1alpha1
kind: LVMCluster
metadata:
  name: my-lvmcluster
spec:
  storage:
    deviceClasses:
    - name: vg1
      fstype: ext4 
      default: true
      deviceSelector: 
        paths:
        - /dev/disk/by-path/pci-0000:87:00.0-nvme-1
        - /dev/disk/by-path/pci-0000:88:00.0-nvme-1
        forceWipeDevicesAndDestroyAllData: true
      thinPoolConfig:
        name: thin-pool-1
        sizePercent: 90 
        overprovisionRatio: 10
```

## Todo
- [ ] add creation of second disk image in drivers (vfkit, libvirt and hyper-v drivers)
- [ ] add a config option to enable users to choose if they want to use LVMS operator
- [ ] create disk and LVMCluster resource during crc start

### LINKED ISSUE #4117: Use LVM storage operator for dynamic PV provisioning
Currently we are using the `hostpath-provisioner` for dynamic pv creation, however since it has limited functionality work is being done in https://github.com/crc-org/snc/pull/877, https://github.com/crc-org/snc/pull/869 and #4114 to install and setup [`lvm-operator`](https://docs.openshift.com/container-platform/4.15/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html#install-lvms-operator-cli_logical-volume-manager-storage)

We need to make a decision whether to make the `lvm-operator` the default one and remove `hostpath-provisioner`, the behavior of expanding disk space using the config setting`disk-size` will change if we move to `lvm-operator` as in that case `disk-size` will refer to the `root` partition size

to increase the space for PVs user has to use the config option `persistent-volume-size`

while talking with @praveenkumar he suggested the following to avoid any behavior change for users who don't want to use `lvm-operator` and are happy with `hostpath-provisioner`:

### First method
1. Introduce a new config ~`enable-lvm-operator`~ `use-lvm-operator` with default value `false` to allow users to make a choice of which operator they want to use
2. When this config is enabled/set to `true` the `LVMCluster` CR will be created and the extra partition will be used for PV
3. When this config is disabled/set to `false` the `root` partition is extended to take up the space from the extra partition (and optionally also delete the operator installation or scale down the deployments as the `operatorgroup` and `subscription` objects are created during `snc`)

### Second method
1.  Introduce a new config `enable-lvm-operator` with default value `false` to allow users to make a choice of which operator they want to use
2. When this config is enabled/set to `true` the extra partition **will not** be merged with the `root` partition and user has to follow the guide on openshift docs to setup/install the LVM operator on their own
3. When this config is disabled/set to `false` the extra partition is merged into the `root` partition

anjannath:
fixes #4117

needs https://github.com/crc-org/snc/pull/877

### DIFF
--- a/pkg/crc/config/settings.go
+++ b/pkg/crc/config/settings.go
@@ -72,6 +72,10 @@ func RegisterSettings(cfg *Config) {
 		return ValidateBool(value)
 	}
 
+	validPVSize := func(value interface{}) (bool, string) {
+		return validatePersistentVolumeSize(value, GetPreset(cfg))
+	}
+
 	// Preset setting should be on top because CPUs/Memory config depend on it.
 	cfg.AddSetting(Preset, version.GetDefaultPreset().String(), validatePreset, RequiresDeleteAndSetupMsg,
 		fmt.Sprintf("Virtual machine preset (valid values are: %s)", preset.AllPresets()))
@@ -93,8 +97,8 @@ func RegisterSettings(cfg *Config) {
 		"Enable experimental features (true/false, default: false)")
 	cfg.AddSetting(EmergencyLogin, false, ValidateBool, SuccessfullyApplied,
 		"Enable emergency login for 'core' user. Password is randomly generated. (true/false, default: false)")
-	cfg.AddSetting(PersistentVolumeSize, constants.DefaultPersistentVolumeSize, validatePersistentVolumeSize, SuccessfullyApplied,
-		fmt.Sprintf("Total size in GiB of the persistent volume used by the CSI driver for %s preset (must be greater than or equal to '%d')", preset.Microshift, constants.DefaultPersistentVolumeSize))
+	cfg.AddSetting(PersistentVolumeSize, defaultPVSize(cfg), validPVSize, SuccessfullyApplied,
+		fmt.Sprintf("Total size in GiB of the persistent volume used by the CSI driver (must be greater than or equal to '%d')", defaultPVSize(cfg)))
 
 	// Shared directories configs
 	if runtime.GOOS == "windows" {
@@ -153,6 +157,10 @@ func presetChanged(cfg *Config, _ string, _ interface{}) {
 
 func defaultCPUs(cfg Storage) int {
 	return constants.GetDefaultCPUs(GetPreset(cfg))
+}
+
+func defaultPVSize(cfg Storage) int {
+	return constants.GetDefaultPersistentVolumeSize(GetPreset(cfg))
 }
 
 func defaultMemory(cfg Storage) int {
--- a/pkg/crc/constants/constants.go
+++ b/pkg/crc/constants/constants.go
@@ -15,8 +15,6 @@ const (
 	DefaultName     = "crc"
 	DefaultDiskSize = 31
 
-	DefaultPersistentVolumeSize = 15
-
 	DefaultSSHUser = "core"
 	DefaultSSHPort = 22
 
@@ -220,6 +218,18 @@ func GetDefaultMemory(preset crcpreset.Preset) int {
 	default:
 		// should not be reached
 		return 10752
+	}
+}
+
+func GetDefaultPersistentVolumeSize(preset crcpreset.Preset) int {
+	switch preset {
+	case crcpreset.OpenShift, crcpreset.OKD:
+		return 3
+	case crcpreset.Microshift:
+		return 15
+	default:
+		// should not be reached
+		return 3
 	}
 }
 
--- a/pkg/crc/machine/disks.go
+++ b/pkg/crc/machine/disks.go
@@ -1,0 +1,162 @@
+package machine
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/crc-org/crc/v2/pkg/crc/constants"
+	logging "github.com/crc-org/crc/v2/pkg/crc/logging"
+	"github.com/crc-org/crc/v2/pkg/crc/machine/types"
+	crcPreset "github.com/crc-org/crc/v2/pkg/crc/preset"
+	crcssh "github.com/crc-org/crc/v2/pkg/crc/ssh"
+	"golang.org/x/crypto/ssh"
+)
+
+func moveTopolvmPartition(ctx context.Context, shiftSize int, vm *virtualMachine, sshRunner *crcssh.Runner) error {
+	_, _, err := sshRunner.RunPrivileged("move topolvm partition to end of disk", fmt.Sprintf("echo '+%dG,' | sudo sfdisk --move-data /dev/vda -N 5 --force", shiftSize))
+	var exitErr *ssh.ExitError
+	if err != nil {
+		if !errors.As(err, &exitErr) {
+			return err
+		}
+		if exitErr.ExitStatus() != 1 {
+			return err
+		}
+	}
+	if err == nil {
+		logging.Info("Restart VM after moving topolvm partition to end")
+		if err := restartHost(ctx, vm, sshRunner); err != nil {
+			return fmt.Errorf("Failed to move topolvm partition to increase root partition size: %w", err)
+		}
+	}
+	return nil
+}
+
+func growPartition(sshRunner *crcssh.Runner, partition string) error {
+	if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Growing %s partition", partition), "/usr/bin/growpart", partition[:len("/dev/.da")], partition[len("/dev/.da"):]); err != nil {
+		var exitErr *ssh.ExitError
+		if !errors.As(err, &exitErr) {
+			return err
+		}
+		if exitErr.ExitStatus() != 1 {
+			return err
+		}
+		logging.Debugf("No free space after %s, nothing to do", partition)
+		return nil
+	}
+	return nil
+}
+
+func ocpGetPVShiftSizeGiB(diskSize int, pvSize int) int {
+	defaultPvSize := constants.GetDefaultPersistentVolumeSize(crcPreset.OpenShift)
+	if pvSize > defaultPvSize {
+		return (diskSize - constants.DefaultDiskSize) - (pvSize - defaultPvSize)
+	}
+	return diskSize - constants.DefaultDiskSize
+}
+
+func growRootFileSystem(ctx context.Context, startConfig types.StartConfig, vm *virtualMachine, sshRunner *crcssh.Runner) error {
+	if startConfig.Preset == crcPreset.OpenShift {
+		sizeToMove := ocpGetPVShiftSizeGiB(startConfig.DiskSize, startConfig.PersistentVolumeSize)
+		if err := moveTopolvmPartition(ctx, sizeToMove, vm, sshRunner); err != nil {
+			return err
+		}
+	}
+	rootPart, err := getrootPartition(sshRunner, startConfig.Preset)
+	if err != nil {
+		return err
+	}
+
+	if err := growPersistentVolume(sshRunner, startConfig.Preset, startConfig.PersistentVolumeSize); err != nil {
+		return fmt.Errorf("Unable to grow persistent volume partition: %w", err)
+	}
+
+	// with '/dev/[sv]da4' as input, run 'growpart /dev/[sv]da 4'
+	if err := growPartition(sshRunner, rootPart); err != nil {
+		return nil
+	}
+	return nil
+}
+
+func growPersistentVolume(sshRunner *crcssh.Runner, preset crcPreset.Preset, persistentVolumeSize int) error {
+	if preset == crcPreset.Microshift {
+		rootPart, err := getrootPartition(sshRunner, preset)
+		if err != nil {
+			return err
+		}
+		lvFullName := "rhel/root"
+		if err := growLVForMicroshift(sshRunner, lvFullName, rootPart, persistentVolumeSize); err != nil {
+			return err
+		}
+	}
+
+	if preset == crcPreset.OpenShift {
+		pvPartition := "/dev/vda5"
+		if err := growPartition(sshRunner, pvPartition); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func getrootPartition(sshRunner *crcssh.Runner, preset crcPreset.Preset) (string, error) {
+	query := "--label root"
+	if preset == crcPreset.Microshift {
+		query = "-t TYPE=LVM2_member"
+	}
+	part, _, err := sshRunner.RunPrivileged("Get device id", "/usr/sbin/blkid", query, "-o", "device")
+	if err != nil {
+		return "", err
+	}
+	parts := strings.Split(strings.TrimSpace(part), "\n")
+	if len(parts) != 1 {
+		return "", fmt.Errorf("Unexpected number of devices: %s", part)
+	}
+	rootPart := strings.TrimSpace(parts[0])
+	if !strings.HasPrefix(rootPart, "/dev/vda") && !strings.HasPrefix(rootPart, "/dev/sda") {
+		return "", fmt.Errorf("Unexpected root device: %s", rootPart)
+	}
+	return rootPart, nil
+}
+
+func growLVForMicroshift(sshRunner *crcssh.Runner, lvFullName string, rootPart string, persistentVolumeSize int) error {
+	if _, _, err := sshRunner.RunPrivileged("Resizing the physical volume(PV)", "/usr/sbin/pvresize", "--devices", rootPart, rootPart); err != nil {
+		return err
+	}
+
+	// Get the size of volume group
+	sizeVG, _, err := sshRunner.RunPrivileged("Get the volume group size", "/usr/sbin/vgs", "--noheadings", "--nosuffix", "--units", "b", "-o", "vg_size", "--devices", rootPart)
+	if err != nil {
+		return err
+	}
+	vgSize, err := strconv.Atoi(strings.TrimSpace(sizeVG))
+	if err != nil {
+		return err
+	}
+
+	// Get the size of root lv
+	sizeLV, _, err := sshRunner.RunPrivileged("Get the size of root logical volume", "/usr/sbin/lvs", "-S", fmt.Sprintf("lv_full_name=%s", lvFullName), "--noheadings", "--nosuffix", "--units", "b", "-o", "lv_size", "--devices", rootPart)
+	if err != nil {
+		return err
+	}
+	lvSize, err := strconv.Atoi(strings.TrimSpace(sizeLV))
+	if err != nil {
+		return err
+	}
+
+	GB := 1073741824
+	vgFree := persistentVolumeSize * GB
+	expectedLVSize := vgSize - vgFree
+	sizeToIncrease := expectedLVSize - lvSize
+	lvPath := fmt.Sprintf("/dev/%s", lvFullName)
+	if sizeToIncrease > 1 {
+		logging.Info("Extending and resizing '/dev/rhel/root' logical volume")
+		if _, _, err := sshRunner.RunPrivileged("Extending and resizing the logical volume(LV)", "/usr/sbin/lvextend", "-L", fmt.Sprintf("+%db", sizeToIncrease), lvPath, "--devices", rootPart); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/pkg/crc/machine/disks.go
+++ b/pkg/crc/machine/disks.go
@@ -63,7 +63,7 @@ func ocpGetPVShiftSizeGiB(diskSize int, pvSize int) int {
 	return diskSize - constants.DefaultDiskSize
 }
 
-func growRootFileSystem(ctx context.Context, startConfig types.StartConfig, vm *virtualMachine, sshRunner *crcssh.Runner) error {
+func growFileSystem(ctx context.Context, startConfig types.StartConfig, vm *virtualMachine, sshRunner *crcssh.Runner) error {
 	if startConfig.Preset == crcPreset.OpenShift {
 		sizeToMove := ocpGetPVShiftSizeGiB(startConfig.DiskSize, startConfig.PersistentVolumeSize)
 		if err := moveTopolvmPartition(ctx, sizeToMove, vm, sshRunner); err != nil {
--- a/pkg/crc/machine/disks.go
+++ b/pkg/crc/machine/disks.go
@@ -16,7 +16,12 @@ import (
 )
 
 func moveTopolvmPartition(ctx context.Context, shiftSize int, vm *virtualMachine, sshRunner *crcssh.Runner) error {
-	_, _, err := sshRunner.RunPrivileged("move topolvm partition to end of disk", fmt.Sprintf("echo '+%dG,' | sudo sfdisk --move-data /dev/vda -N 5 --force", shiftSize))
+	pvPartition, err := getTopolvmPartition(sshRunner)
+	if err != nil {
+		return err
+	}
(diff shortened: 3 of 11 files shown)


pkg/crc/config/settings.go
```diff
@@ -37,6 +37,7 @@ const (
 	EmergencyLogin           = "enable-emergency-login"
 	PersistentVolumeSize     = "persistent-volume-size"
 	EnableBundleQuayFallback = "enable-bundle-quay-fallback"
+	SecondaryDiskSize        = "secondary-disk-size"
```
cfergeau: Isn't this similar to `PersistentVolumeSize`?

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -135,6 +143,14 @@ func (d *Driver) Create() error {
 		return fmt.Errorf("%s is an unsupported disk image format", d.ImageFormat)
 	}
 
+	if d.SecondaryDiskSize > 0 {
+		if crcos.FileExists(d.getSecondaryDiskPath()) {
+			return nil
+		}
+		_, err := diskfs.Create(d.getSecondaryDiskPath(), int64(d.SecondaryDiskSize), diskfs.Raw, diskfs.SectorSizeDefault)
```
cfergeau: Is `diskfs` required? Could this use `os.Truncate` directly to create an empty disk image?

pkg/crc/config/settings.go
anjannath: yes practically they are the same, one difference is that `secondary-disk-size` is at the VM level similar to `cpus`,`memory` and `persistent-volume-size` is at the openshift cluster level

cfergeau: Do we need to expose this difference to the user?

anjannath: nope, going to use the same config key `persistent-volume-size` and remove the newly added one

pkg/drivers/vfkit/driver_darwin.go
anjannath: yes, this works, tried using `os.Truncate`, we have to first create an empty file and then change the size using `os.Truncate`

cmd/crc/cmd/start.go
```diff
@@ -79,6 +79,7 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 		IngressHTTPPort:   config.Get(crcConfig.IngressHTTPPort).AsUInt(),
 		IngressHTTPSPort:  config.Get(crcConfig.IngressHTTPSPort).AsUInt(),
 		EnableSharedDirs:  config.Get(crcConfig.EnableSharedDirs).AsBool(),
+		SecondaryDiskSize: config.Get(crcConfig.SecondaryDiskSize).AsInt(),
```
gbraad: Here we assume only a second disk could exist. But what about `Disks` ?

anjannath: to make it work with more disks, we'd have to create a separate `DiskConfig` type maybe.. containing the path and the size then add that to the drivers, i went with this simpler approach to first take care of the requirements for installing topolvm

anjannath:
this PR is not needed anymore, the snc PRs:
1. https://github.com/crc-org/snc/pull/869
2. https://github.com/crc-org/snc/pull/869
will take care of all the steps needed to install topolvm entirely on snc

pkg/crc/machine/start.go
```diff
@@ -151,11 +151,11 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset, persi
...
 
 func getrootPartition(sshRunner *crcssh.Runner, preset crcPreset.Preset) (string, error) {
-	diskType := "xfs"
+	cmd := "/usr/sbin/blkid --label root -o device"
 	if preset == crcPreset.Microshift {
-		diskType = "LVM2_member"
+		cmd = "/usr/sbin/blkid -t TYPE=LVM2_member -o device"
 	}
-	part, _, err := sshRunner.RunPrivileged("Get device id", "/usr/sbin/blkid", "-t", fmt.Sprintf("TYPE=%s", diskType), "-o", "device")
+	part, _, err := sshRunner.RunPrivileged("Get device id", cmd)
```
cfergeau: While`sshRunner.RunPrivileged(cmd)` will work, the `RunPrivileged` interface expects split args (ie a `[]string`) as other implementations won't work with a single string containing all the args separated with spaces.

cfergeau: Might be better to avoid the duplication by using `query := "--label root"` and `query = "-t TYPE=LVM2_member"` and then hardcoding `/usr/sbin/blkid` and `-o device` in a single place.

pkg/crc/machine/start.go
```diff
@@ -126,156 +125,6 @@ func restartHost(ctx context.Context, vm *virtualMachine, sshRunner *crcssh.Runn
...
-	if err != nil {
-		return err
-	}
-
-	if err := growPersistentVolume(sshRunner, startConfig.Preset, startConfig.PersistentVolumeSize); err != nil {
-		return fmt.Errorf("Unable to grow persistent volume partition: %w", err)
-	}
-
-	// with '/dev/[sv]da4' as input, run 'growpart /dev/[sv]da 4'
-	if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Growing %s partition", rootPart), "/usr/bin/growpart", rootPart[:len("/dev/.da")], rootPart[len("/dev/.da"):]); err != nil {
```
cfergeau, line 122: The corresponding code in the new file has become
```
	if err := growPersistentVolume(sshRunner, startConfig.Preset, startConfig.PersistentVolumeSize); err != nil {
		return fmt.Errorf("Unable to grow persistent volume partition: %w", err)
	}

	// with '/dev/[sv]da4' as input, run 'growpart /dev/[sv]da 4'
	if err := growPartition(sshRunner, rootPart); err != nil {
		return nil
	}
```

While the addition of a helper is a good thing, this should not be hidden in a commit saying it's only moving code.
Fwiw, a helper for `move topolvm partition to end of disk` could also be useful to make `growRootFilesystem` easier to read.

cfergeau (commented):
I'm a bit confused by ' update disk resize to work with new disk layout ', could you detail more in the commit log why we need to move the topolvm partition (ie what fails if we don't do this)? What is the order before, what is the order after?  Could this be done proactively in snc? And can we resize both the root partition and the PVs without any issue?

cfergeau:
> this PR is not needed anymore, the snc PRs:

I assume this is no longer accurate as you updated the PR?

pkg/crc/machine/start.go
anjannath: updated the commit to use the above suggestion!

anjannath:
> I'm a bit confused by ' update disk resize to work with new disk layout ', could you detail more in the commit log why we need to move the topolvm partition (ie what fails if we don't do this)? What is the order before, what is the order after? Could this be done proactively in snc? And can we resize both the root partition and the PVs without any issue?

i've updated the commit message.. but mentioning here as well: after adding a new partition at the end of disk, the partition layout is `|efi|boot|root|topolvm|`, now if we want to increase the size of the root partition we have to first move the `topolvm` partition to create the required free space after the `root` partition, earlier the `root` partition was the last partition on the disk and after increasing the disk size the additional free space was added after it, which is not the case anymore because the `topolvm` partition is there now

cfergeau:
> > I'm a bit confused by ' update disk resize to work with new disk layout ', could you detail more in the commit log why we need to move the topolvm partition (ie what fails if we don't do this)? What is the order before, what is the order after? Could this be done proactively in snc? And can we resize both the root partition and the PVs without any issue?
> 
> i've updated the commit message.. but mentioning here as well: after adding a new partition at the end of disk, the partition layout is `|efi|boot|root|topolvm|`, now if we want to increase the size of the root partition we have to first move the `topolvm` partition to create the required free space after the `root` partition, earlier the `root` partition was the last partition on the disk and after increasing the disk size the additional free space was added after it, which is not the case anymore because the `topolvm` partition is there now

Ah ok, we need to go from `|efi|boot|root|topolvm|free space|` to `|efi|boot|root|free space|topolvm|`. The `topolvm` partition is always the last one, but its end also has to be aligned with the end of the disk so that the freed space comes before rather than after it.

pkg/crc/constants/constants.go
```diff
@@ -223,6 +221,18 @@ func GetDefaultMemory(preset crcpreset.Preset) int {
...
 
+func GetDefaultPersistentVolumeSize(preset crcpreset.Preset) int {
+	switch preset {
+	case crcpreset.OpenShift, crcpreset.OKD:
+		return 3
+	case crcpreset.Microshift:
+		return 15
+	default:
+		// should not be reached
+		return 3
```
praveenkumar, line 232: Will it be enough? because registry also going to use the pvc which is going to part of this partition and if a user mirror an image to openshift or even build a image from openshift then this space is going to be filled very fast :(

pkg/crc/machine/start.go
```diff
@@ -112,12 +112,57 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	}
+	if err := startHost(ctx, vm); err != nil {
+		return errors.Wrap(err, "Error re-starting machine at start")
+	}
+	logging.Debug("Waiting until ssh is available")
+	if err := sshRunner.WaitForConnectivity(ctx, 300*time.Second); err != nil {
+		return errors.Wrap(err, "Failed to connect to the CRC VM with SSH -- virtual machine might be unreachable")
+	}
+	return nil
+}
```
praveenkumar, line 126: This restart means add more time to startup even it takes ~1 mins.

pkg/crc/constants/constants.go
anjannath: right, we might have to increase the default disk-size, but there's very limited free space available, and currently that space is  shared for both image-registry and host container image storage, with this since we have to create a separate partition this becomes more apparent:

```
[core@crc ~]$ df -h /dev/vda4
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda4        31G   23G  7.6G  76% /
```

We only have 7.6G free space, of which now `4.6G` is for host container images and `3G` for PVs

anjannath: maybe we can increase the default `disk-size` to 40 and default `persistent-volume-size` to 7, these are just arbitrary numbers, but it'd leave around 7gb free space on the root partition (closer to the current situation)

(the bundle sizes are also increasing which would mean less free space for workloads, right? i saw the `4.16.3` bundle from CI in 400MB larger than the 4.16.0 bundle)

pkg/crc/machine/start.go
anjannath: yes, its less than 1 min, but that's on linux, on windows i expect it to be even longer than 1 minute

pkg/crc/machine/start.go
```diff
@@ -150,6 +188,35 @@ func growRootFileSystem(sshRunner *crcssh.Runner, preset crcPreset.Preset, persi
...
+		}
+		lvFullName := "rhel/root"
+		if err := growLVForMicroshift(sshRunner, lvFullName, rootPart, persistentVolumeSize); err != nil {
+			return err
+		}
+	}
+
+	if preset == crcPreset.OpenShift {
+		pvPartition := "/dev/vda5"
+		if _, _, err := sshRunner.RunPrivileged(fmt.Sprintf("Growing %s partition", pvPartition), "/usr/bin/growpart", pvPartition[:len("/dev/.da")], pvPartition[len("/dev/.da"):]); err != nil {
```
cfergeau: The `add helper to grow root partition` should come before this commit, this way you can reuse it here instead of duplicating already existing code.

pkg/crc/machine/start.go
```diff
@@ -112,12 +112,57 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+		sizeToMove := ocpGetPVShiftSize(startConfig.DiskSize, startConfig.PersistentVolumeSize)
+		_, _, err := sshRunner.RunPrivileged("move topolvm partition to end of disk", fmt.Sprintf("echo '+%dG,' | sudo sfdisk --move-data /dev/vda -N 5 --force", sizeToMove))
+		var exitErr *ssh.ExitError
+		if err != nil {
+			if !errors.As(err, &exitErr) {
+				return err
+			}
+			if exitErr.ExitStatus() != 1 {
+				return err
+			}
```
cfergeau: Do you need to ignore errors when the command exited with 1 as its exit code? Or is it a copy and paste error?

pkg/crc/machine/start.go
```diff
@@ -112,12 +112,57 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	if pvSize > defaultPvSize {
+		return (diskSize - constants.DefaultDiskSize) - (pvSize - defaultPvSize)
+	}
+	return diskSize - constants.DefaultDiskSize
+}
+
+func growRootFileSystem(ctx context.Context, startConfig types.StartConfig, vm *virtualMachine, sshRunner *crcssh.Runner) error {
+	if startConfig.Preset == crcPreset.OpenShift {
+		sizeToMove := ocpGetPVShiftSize(startConfig.DiskSize, startConfig.PersistentVolumeSize)
+		_, _, err := sshRunner.RunPrivileged("move topolvm partition to end of disk", fmt.Sprintf("echo '+%dG,' | sudo sfdisk --move-data /dev/vda -N 5 --force", sizeToMove))
```
cfergeau: I would squash `add helper to move topolvm partition` with this commit, less code to review overall :)

pkg/crc/machine/start.go
cfergeau: Is the restart required? If I run the `sfdisk` command manually, I get
```
The partition table has been altered.
Calling ioctl() to re-read partition table.
Re-reading the partition table failed.: Device or resource busy
The kernel still uses the old table. The new table will be used at the next reboot or after you run partprobe(8) or partx(8).
Syncing disks.
```
Running `partprobe` could be enough? 

I'd really prefer to avoid restarts of the VM/... during a `crc start`, this feels like a complication which will cause us problems later on.

pkg/crc/machine/start.go
```diff
@@ -112,12 +112,57 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+
+func ocpGetPVShiftSize(diskSize int, pvSize int) int {
+	defaultPvSize := constants.GetDefaultPersistentVolumeSize(crcPreset.OpenShift)
+	if pvSize > defaultPvSize {
+		return (diskSize - constants.DefaultDiskSize) - (pvSize - defaultPvSize)
+	}
+	return diskSize - constants.DefaultDiskSize
+}
+
+func growRootFileSystem(ctx context.Context, startConfig types.StartConfig, vm *virtualMachine, sshRunner *crcssh.Runner) error {
```
cfergeau: This is already the case before this commit, but `growRootFileSystem` is badly named, it grows the root filesystem, and the PVs filesystem. `growFilesystems` would be a better name for it.

pkg/crc/machine/start.go
```diff
@@ -112,12 +112,57 @@ func (client *client) updateVMConfig(startConfig types.StartConfig, vm *virtualM
...
+	}
+	return nil
+}
+
+func ocpGetPVShiftSize(diskSize int, pvSize int) int {
+	defaultPvSize := constants.GetDefaultPersistentVolumeSize(crcPreset.OpenShift)
+	if pvSize > defaultPvSize {
+		return (diskSize - constants.DefaultDiskSize) - (pvSize - defaultPvSize)
+	}
+	return diskSize - constants.DefaultDiskSize
```
cfergeau: I'd recommend making use of `github.com/containers/common/pkg/strongunits` here (and in more places in our codebase :). At the very least, add a GiB suffix to the function name so that there is no confusion about what we are manipulating.

cfergeau: I'd favour calculating this from `sfdisk` output though, if the disk layout is in an unexpected state, these calculations could break as they rely on external values, not on the actual disk state. What we want to do is to align the last partition to the end of the disk, so we can do it without any knowledge of the default disk size, the disk size stored in json metadata ,...

pkg/crc/machine/start.go
anjannath: running `partprobe` was not enough, it needed the reboot, yeah i agree that this makes it complicated and now it works because the VM ip doesn't change after a reboot

pkg/crc/machine/disks.go
```diff
@@ -39,7 +39,12 @@ func ocpGetPVShiftSize(diskSize int, pvSize int) int {
 }
 
 func moveTopolvmPartition(ctx context.Context, shiftSize int, vm *virtualMachine, sshRunner *crcssh.Runner) error {
-	_, _, err := sshRunner.RunPrivileged("move topolvm partition to end of disk", fmt.Sprintf("echo '+%dG,' | sudo sfdisk --move-data /dev/vda -N 5 --force", shiftSize))
+	pvPartition, err := getTopolvmPartition(sshRunner)
+	if err != nil {
+		return err
+	}
+	_, _, err = sshRunner.RunPrivileged("move topolvm partition to end of disk",
+		fmt.Sprintf("echo '+%dG,' | sudo sfdisk --move-data /dev/vda -N %s --force", shiftSize, pvPartition[len("/dev/.da"):]))
```
cfergeau: If you use `len("/dev/.da")`, it might be worth also making the `/dev/vda` more generic? Iirc it's `/dev/sda` on Windows?

pkg/crc/machine/disks.go
```diff
@@ -125,6 +133,14 @@ func getrootPartition(sshRunner *crcssh.Runner, preset crcPreset.Preset) (string
...
 		query = "-t TYPE=LVM2_member"
 	}
+	return getDeviceID(sshRunner, query)
+}
+
+func getTopolvmPartition(sshRunner *crcssh.Runner) (string, error) {
+	return getDeviceID(sshRunner, "-t PARTLABEL=topolvm")
+}
+
+func getDeviceID(sshRunner *crcssh.Runner, query string) (string, error) {
```
cfergeau: I find the api to this function a bit odd, this `query` string is quite magic. Might be better to name it `runBlkidQuery` or such?

pkg/crc/machine/disks.go
```diff
@@ -125,6 +133,14 @@ func getrootPartition(sshRunner *crcssh.Runner, preset crcPreset.Preset) (string
 	if preset == crcPreset.Microshift {
 		query = "-t TYPE=LVM2_member"
 	}
+	return getDeviceID(sshRunner, query)
+}
+
+func getTopolvmPartition(sshRunner *crcssh.Runner) (string, error) {
+	return getDeviceID(sshRunner, "-t PARTLABEL=topolvm")
```
cfergeau: I'd also rework `getDeviceID` to accept a `...string` rather than just a `string` as the correct (in theory) way of calling it is `getDeviceID(sshRunner, "-t", "PARTLABEL=topolvm")`

I made this change on top of yours:

```
diff --git a/pkg/crc/machine/disks.go b/pkg/crc/machine/disks.go
index 4283d1501..2ea169de0 100644
--- a/pkg/crc/machine/disks.go
+++ b/pkg/crc/machine/disks.go
@@ -129,19 +131,20 @@ func growPersistentVolume(sshRunner *crcssh.Runner, preset crcPreset.Preset, per
 }
 
 func getrootPartition(sshRunner *crcssh.Runner, preset crcPreset.Preset) (string, error) {
-	query := "--label root"
 	if preset == crcPreset.Microshift {
-		query = "-t TYPE=LVM2_member"
+		return getDeviceID(sshRunner, "-t", "TYPE=LVM2_member")
 	}
-	return getDeviceID(sshRunner, query)
+	return getDeviceID(sshRunner, "--label", "root")
 }
 
 func getTopolvmPartition(sshRunner *crcssh.Runner) (string, error) {
-	return getDeviceID(sshRunner, "-t PARTLABEL=topolvm")
+	return getDeviceID(sshRunner, "-t", "PARTLABEL=topolvm")
 }
 
-func getDeviceID(sshRunner *crcssh.Runner, query string) (string, error) {
-	part, _, err := sshRunner.RunPrivileged("Get device id", "/usr/sbin/blkid", query, "-o", "device")
+func getDeviceID(sshRunner *crcssh.Runner, query ...string) (string, error) {
+	cmd := []string{"/usr/sbin/blkid", "-o", "device"}
+	cmd = append(cmd, query...)
+	part, _, err := sshRunner.RunPrivileged("Get device id", cmd...)
 	if err != nil {
 		return "", err
 	}
```

pkg/crc/machine/start.go
anjannath: > What we want to do is to align the last partition to the end of the disk, so we can do it without any knowledge of the default disk size, the disk size stored in json metadata

that is not the case always, sometime we want to move the partition not to the very end and leave space to increase the PV partition size

suppose `disk-size` is `40` and `persistent-volume-size` is `5` (all in GB) then after increasing the entire disk to 40GB from 31GB the last partition needs to be moved by 7GB (not 9GB) leaving 2GB of space at the end so the PV partition can grow to become 5GB (default size is 3GB)

cfergeau: We could grow the partition before moving it since it's at the end, or we can take into account its new size desired size when moving it, this is fine as this is data we cannot know from the current layout.

My impression from the current is that we determine the offset we need to move the partition without ever looking at the actual on disk layout, but only using hardcoded constants and crc configuration values. If the disk layout (unexpectedly) goes out of sync with this, bad things will happen I think?

anjannath: i think we can replace `constant.DefaultDiskSize` and `defaultPvSize` with the current disk size and current pv partition size fetching it with `sfdisk`, but i think it'll be an edge case if the config values do not reflect the actual values

pkg/crc/machine/disks.go
anjannath: thanks, i applied this patch on the commit, and rename the function to `runBlkidQuery`

pkg/crc/machine/start.go
anjannath: yes we need to ignore it, after moving the partition it returns exit code 1 (i think because of `Re-reading the partition table failed.: Device or resource busy` but couldn't find anything about exit codes in `sfdisk` man page)

pkg/crc/machine/start.go
anjannath: added https://github.com/crc-org/crc/pull/4114/commits/d129b861633b41aadc73767b7b43959ccd0f94fb to address this 

i'll try to incorporate this into the original commit, but since the functions gets moved to another file, its taking a bit of time to do this

pkg/crc/constants/constants.go
cfergeau: For what it's worth, increasing the guest disk size should not increase the bundle size nor the used disk space after `crc start` as long as the additional space is empty.

anjannath:
seems this is not a good return on investment, closing it, sorry about the noise!

if users want to use topolvm they can manually enable/install it, we'll add a wiki entry with the steps

PR #4132, opened 2024-04-30
build: Simplify/improve verify-vendor.sh

cfergeau:
`verify-vendor.sh` compares the content of the `vendor` directory after
running `make vendor` with a copy which was made before running `make
vendor`.
This does not catch changes to go.mod/go.sum made by `go mod tidy` which
we forgot to commit.

This commit uses `git diff --exit-code` instead of running `diff` against the
vendor directory.

This will catch all uncommitted changes, which should be fine when `make
vendorcheck` is run.

Using `git diff` instead of `diff` allows to remove quite some code from
`verify-vendor.sh`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 all: install
 
-SHELL := /bin/bash
+SHELL := /bin/bash -o pipefail
 
 OPENSHIFT_VERSION ?= 4.15.10
 PODMAN_VERSION ?= 4.4.4
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/crc-org/crc/v2
 
-go 1.21
+go 1.20
 
 require (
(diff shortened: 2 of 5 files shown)


cfergeau:
The "Build RPM" failure means this change is working as expected. However, I realize now that it would be good if `make spec` failed when `gomod2rpmdeps` fails.

praveenkumar:
About https://github.com/crc-org/crc/pull/4132/commits/07add4ff6a24d2fd918251c1014268a24ded50e2 I was thinking instead to use go-1.21 tool chain can we revert ec6cbfb3d35f433a6285bda4b12badb45e97f915 one and use `go mod tidy -go 1.20` as part of `make vendor` ?

praveenkumar:
should we close it because #4139 have all those commits also?

cfergeau:
Since this PR won't work without updating to 1.21, sure.

PR #4139, opened 2024-05-02
Update to go-1.21

praveenkumar:


update-go-version.sh
```diff
@@ -14,5 +14,7 @@ go mod edit -go ${golang_base_version} tools/go.mod
 sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-\)1.[0-9]\+,\1${golang_base_version}," images/*/Dockerfile
 sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Dockerfile
 for f in .github/workflows/*.yml; do
-    yq eval --inplace ".jobs.build.strategy.matrix.go[0] = ${golang_base_version}" "$f";
+    if [ $(yq '.jobs.build.strategy.matrix | has("go")' "$f") == "true" ]; then
```
cfergeau: Better to specify `eval` in the `yq` command for backwards compat: 
```
```

I was getting errors on my system without this, and realized it has an older yq version.

update-go-version.sh
```diff
@@ -15,6 +15,6 @@ sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-gol
 sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Dockerfile
 for f in .github/workflows/*.yml; do
     if [ $(yq '.jobs.build.strategy.matrix | has("go")' "$f") == "true" ]; then
-      yq eval --inplace ".jobs.build.strategy.matrix.go[0] = ${golang_base_version}" "$f";
+      yq eval --inplace ".jobs.build.strategy.matrix.go[0] = ${golang_base_version} | .jobs.build.strategy.matrix.go[0] style=\"single\"" "$f";
```
cfergeau: I'm not sure what this is fixing? All the workflows file have single quotes around the go version, and if I run `update-go-version 1.21` without this commit, then I get quoted `'1.21'` in the workflows  files.

cfergeau: I ran more tests, and this is needed because otherwise 
```
        go:
         - '1.20'
```
gets replaced with
```
        go:
          - 1.21
```

But this behaviour only started happening in yq 4.21.1, I don't know if this was intentional.

Makefile
```diff
@@ -327,7 +327,7 @@ endif
 
 .PHONY: update-go-version
 update-go-version:
-	./update-go-version.sh 1.17
+	./update-go-version.sh 1.20
```
cfergeau: last go version update, I ran `update-go-version` manually because I did not know of this target :)

.github/workflows/make-check.yml
```diff
@@ -26,3 +26,6 @@ jobs:
           go-version: ${{ matrix.go }}
       - name: Build
         run: make check
+      - if: runner.os == 'Linux'
+        name: Go version check
+        run: make goversioncheck
```
cfergeau: Not sure yet how this will interact with https://github.com/crc-org/crc/pull/4132

update-go-version.sh
praveenkumar: but shouldn't we want to use latest always for yq?

update-go-version.sh
praveenkumar: I don't know, was reading the doc and because I was not getting single quotes and found out the style operator to use.

.github/workflows/make-check.yml
praveenkumar: Didn't get you, it is different target the vendor verification right?

cfergeau: #4132 had a change changing the go version to 1.21 in the go.mod file, but not touching the rest of the files. This would temporarily break `make go versioncheck` if this went in this way (it won't).

update-go-version.sh
cfergeau: We have no check for the yq version, not even sure we have something to automatically install it, it's not packaged in fedora so won't be updated automatically...

In this case, the change is trivial and allows to support more `yq` versions, it will save us cryptic error messages if this ends up running somewhere with a slightly older yq version.

update-go-version.sh
cfergeau: Might be nice to detail the commit log a bit more
```
 is not the case till now.
```
-> "which is not the case with yq 4.21.1 and newer. `yq eval --inplace ".jobs.build.strategy.matrix.go[0] = 1.21` replaces `go: '1.20'` with `go: 1.21` without quotes"

Makefile
```diff
@@ -82,7 +82,7 @@ default: install
 # Create and update the vendor directory
 .PHONY: vendor
 vendor:
-	go mod tidy
+	go mod tidy -go 1.21
```
cfergeau: I'd do this last in the patch series

cfergeau:
`mak echeck` failed on Windows.

praveenkumar:
> `make check` failed on Windows.

@cfergeau only on windows-2019 and windows-2022 it is working as expected, I will check what's is happening there.

update-go-version.sh
praveenkumar: > We have no check for the yq version, not even sure we have something to automatically install it, it's not packaged in fedora so won't be updated automatically...

It is packaged in fedora, got it from update repo
```
$ sudo dnf info yq
Available Packages
Name         : yq
Version      : 4.43.1
Release      : 2.fc39
Architecture : x86_64
Size         : 3.4 M
Source       : yq-4.43.1-2.fc39.src.rpm
Repository   : updates
Summary      : Yq is a portable command-line YAML, JSON, XML, CSV, TOML  and properties processor
URL          : https://github.com/mikefarah/yq
License      : MIT
Description  : Yq is a portable command-line YAML, JSON, XML, CSV, TOML  and properties
             : processor.
```
> 
> In this case, the change is trivial and allows to support more `yq` versions, it will save us cryptic error messages if this ends up running somewhere with a slightly older yq version.

Sure

Makefile
praveenkumar: @cfergeau Done

update-go-version.sh
praveenkumar: Done

cfergeau: > It is packaged in fedora, got it from update repo

Looks like the first `yq` package landed in fedora updates yesterday ;) https://bugzilla.redhat.com/show_bug.cgi?id=2276522

cfergeau (commented):
Fwiw, we used to have a similar workaround https://github.com/crc-org/crc/commit/1b6bf222090f
I don't know what "win-2022 also have `OS-10.0.x` build" means.

praveenkumar:
> Fwiw, we used to have a similar workaround https://github.com/crc-org/crc/commit/1b6bf222090f

Do you suggest that we should use a hack keep the `windows-2019` runner job?

> I don't know what "win-2022 also have `OS-10.0.x` build" means.

I took it that it have some similarity with win-10 but as per https://en.wikipedia.org/wiki/Windows_Server_2022 

looks like codebase is different Windows Server 2022 is based on the "Iron" codebase.[[5]](https://en.wikipedia.org/wiki/Windows_Server_2022#cite_note-windows-central-iron-5) Its updates are incompatible with the Windows 10 operating system, as the "Iron" codebase was not used for it.[[5]](https://en.wikipedia.org/wiki/Windows_Server_2022#cite_note-windows-central-iron-5) Like its predecessor, [Windows Server 2019](https://en.wikipedia.org/wiki/Windows_Server_2019), it requires [x64](https://en.wikipedia.org/wiki/X64) processors.


I think we can keep windows-2022 till it is working and remove it if that also starts failing because now we already have win-10/11 for e2e added as gh-action.

cfergeau:
> > Fwiw, we used to have a similar workaround [1b6bf222090f](https://github.com/crc-org/crc/commit/1b6bf222090f)
> 
> Do you suggest that we should use a hack keep the `windows-2019` runner job?

I only mean that I don't find the argument "we need a workaround to keep the win2019 runners" very convincing when we already had a similar workaround in the past. 

> I think we can keep windows-2022 till it is working and remove it if that also starts failing because now we already have win-10/11 for e2e added as gh-action.

Fine with me, not even sure win2022 brings much as we have win10/win11 e2e?

cfergeau (changes requested):
ack from me with the confusing ` "win-2022 also have OS-10.0.x build" `  removed from the commit log.

praveenkumar:
> ack from me with the confusing `"win-2022 also have OS-10.0.x build"` removed from the commit log.

removed it from the commit message.

cfergeau (approved):
I'd delay this after the next release though.

cfergeau (approved):
I'd delay this after the next release though.

praveenkumar:
> I'd delay this after the next release though.

@cfergeau because of RHEL-8.10 still not out?

/hold

cfergeau:
> > I'd delay this after the next release though.
> 
> @cfergeau because of RHEL-8.10 still not out?
> 
> /hold

In a more general way, we don't want to discover during the release that "something" was not ready for go 1.21. But yes, I think we do our linux release builds on a rhel8, so this is one good reason for waiting with this change.

Issue #4110, opened 2024-04-08
[BUG] crc start fails on ` Waiting until the user's pull secret is written to the instance disk...` step


lavocatt:
### General information

  * OS: Linux
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? yes
  * Running CRC on: Laptop

## CRC version
```bash
CRC version: 2.34.1+b470b5
OpenShift version: 4.15.3
Podman version: 4.4.4
```
  
## CRC status
```bash
crc status --log-level debug
DEBU CRC version: 2.34.1+b470b5                   
DEBU OpenShift version: 4.15.3                    
DEBU Podman version: 4.4.4                        
DEBU Running 'crc status'                         
CRC VM:          Running
OpenShift:       Unreachable (v4.15.3)
RAM Usage:       1.932GB of 10.96GB
Disk Usage:      19.05GB of 32.68GB (Inside the CRC VM)
Cache Usage:     26.27GB
Cache Directory: /home/tlavocat/.crc/cache
```

## CRC config
```bash
crc config view
- consent-telemetry                     : yes
```

## Host Operating System
```bash
NAME="Fedora Linux"
VERSION="39 (Workstation Edition)"
ID=fedora
VERSION_ID=39
VERSION_CODENAME=""
PLATFORM_ID="platform:f39"
PRETTY_NAME="Fedora Linux 39 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:39"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f39/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=39
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=39
SUPPORT_END=2024-11-12
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

### Steps to reproduce

  1. crc setup
  2. crc start -p ~/Documents/pull-secret

### Expected

I expect a normal startup.

### Actual

[...]

PR #4143, opened 2024-05-03
preflight-linux: Add logic to check if /etc/resolv.conf managed by systemd-resolved

praveenkumar:
Current logic checks if systemd-resolved service is running and add the dispatcher file to network manager config which uses `systemd-resolve` to update the domain and dns for `crc` interface. But it is observed for systems which have systemd-resolved service enabled and running but the `/etc/resolv.conf` is not managed by it this dispatcher file is not going to use for dns resolution and user failed to start the cluster.

This PR add logic to check if `/etc/resolv.conf` actually managed by systemd-resolved and if not then use `dnsmasq` configuration which works as expected.

**Fixes:** Issue #4110

### DIFF
--- a/pkg/crc/preflight/preflight_checks_network_linux.go
+++ b/pkg/crc/preflight/preflight_checks_network_linux.go
@@ -13,6 +13,7 @@ import (
 	"github.com/crc-org/crc/v2/pkg/crc/systemd"
 	"github.com/crc-org/crc/v2/pkg/crc/systemd/states"
 	crcos "github.com/crc-org/crc/v2/pkg/os"
+	crcstring "github.com/crc-org/crc/v2/pkg/strings"
 )
 
 var nmPreflightChecks = []Check{
@@ -275,6 +276,20 @@ func checkNetworkManagerIsRunning() error {
 }
 
 func checkSystemdResolvedIsRunning() error {
+	resolvFilePath := "/etc/resolv.conf"
+	// Check /ETC/RESOLV.CONF section in `man systemd-resolved`
+	systemdResolvedManageResolvFilePath := []string{"/run/systemd/resolve/stub-resolv.conf",
+		"/usr/lib/systemd/resolv.conf",
+		"/run/systemd/resolve/resolv.conf"}
+	// Check if /etc/resolv.conf is managed by systemd-resolved
+	rFilePath, err := filepath.EvalSymlinks(resolvFilePath)
+	logging.Debugf("resolv.conf file path: %s", rFilePath)
+	if err != nil {
+		return err
+	}
+	if !crcstring.Contains(systemdResolvedManageResolvFilePath, rFilePath) {
+		return fmt.Errorf("%s is not managed by systemd-resolved", resolvFilePath)
+	}
 	return checkSystemdServiceRunning("systemd-resolved.service")
 }
 


praveenkumar:
/hold

This uses go-1.21 functionality to check if a item present in slice it should go after #4139 one.

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -275,6 +276,20 @@ func checkNetworkManagerIsRunning() error {
 }
 
 func checkSystemdResolvedIsRunning() error {
+	resolvFilePath := "/etc/resolv.conf"
+	// Check /ETC/RESOLV.CONF section in `man systemd-resolved`
+	systemdResolvedManageResolveFilePath := []string{"/run/systemd/resolve/stub-resolv.conf",
+		"/usr/lib/systemd/resolv.conf",
+		"/run/systemd/resolve/resolv.conf"}
+	// Check if /etc/resolv.conf is managed by systemd-resolved
+	rFilePath, err := filepath.EvalSymlinks(resolvFilePath)
```
cfergeau, line 285: This also seems to be available as part of `resolvectl status`?

cfergeau (commented):
>  but the `/etc/resolv.conf` is not managed by it this dispatcher file is not going to use for dns resolution and user failed to start the cluster.

Not so sure about this, my feeling would be that the dispatcher is called when the NM interface changes state, but then the `resolvectl` call it contains has no effect on DNS resolution as systemd-resolved does not manage /etc/resolv.conf?

pkg/crc/preflight/preflight_checks_network_linux.go
praveenkumar: @cfergeau  I don't find it as part of status command

```
$ resolvectl status
Global
         Protocols: LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported
  resolv.conf mode: stub

Link 3 (wlp0s20f3)
    Current Scopes: none
         Protocols: -DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported

Link 4 (virbr0)
    Current Scopes: LLMNR/IPv4
         Protocols: -DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported

Link 21 (enp0s20f0u5u1)
    Current Scopes: DNS LLMNR/IPv4 LLMNR/IPv6
         Protocols: +DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported
Current DNS Server: 192.168.0.1
       DNS Servers: 192.168.0.1

Link 48 (cni-podman0)
    Current Scopes: none
         Protocols: -DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported

Link 57 (vnet1)
    Current Scopes: LLMNR/IPv6
         Protocols: -DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported

Link 68 (crc)
    Current Scopes: none
         Protocols: -DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported
       DNS Servers: 192.168.130.11
        DNS Domain: ~testing

```

praveenkumar:
> > but the `/etc/resolv.conf` is not managed by it this dispatcher file is not going to use for dns resolution and user failed to start the cluster.
> 
> Not so sure about this, my feeling would be that the dispatcher is called when the NM interface changes state, but then the `resolvectl` call it contains has no effect on DNS resolution as systemd-resolved does not manage /etc/resolv.conf?


Yes dispatcher is called and even command is executed but since the `/etc/resolv.conf` not managed it have no effect.

pkg/crc/preflight/preflight_checks_network_linux.go
cfergeau: Isn't this `  resolv.conf mode: stub` ?

cfergeau:
> > > but the `/etc/resolv.conf` is not managed by it this dispatcher file is not going to use for dns resolution and user failed to start the cluster.
> > 
> > 
> > Not so sure about this, my feeling would be that the dispatcher is called when the NM interface changes state, but then the `resolvectl` call it contains has no effect on DNS resolution as systemd-resolved does not manage /etc/resolv.conf?
> 
> Yes dispatcher is called and even command is executed but since the `/etc/resolv.conf` not managed it have no effect.

Can this be clarified in the commit log?

pkg/crc/preflight/preflight_checks_network_linux.go
praveenkumar: As per man page of `systemd-resolved` there are 4 mode but it doesn't specify what `stuf` mode is for on csb I can see `foreign` as mode which doesn't have any info on the manual. Also how to parse the mode I can see there is option for json output and it doesn't provide any json value :( . Do we have a handy way to detect it and have docs around the mode value?

```
/ETC/RESOLV.CONF
       Four modes of handling /etc/resolv.conf (see resolv.conf(5)) are supported:

       •   systemd-resolved maintains the /run/systemd/resolve/stub-resolv.conf file for compatibility with traditional Linux programs. This file lists the 127.0.0.53 DNS stub (see above) as the only DNS server. It also contains a list of search
           domains that are in use by systemd-resolved. The list of search domains is always kept up-to-date. Note that /run/systemd/resolve/stub-resolv.conf should not be used directly by applications, but only through a symlink from
           /etc/resolv.conf. This file may be symlinked from /etc/resolv.conf in order to connect all local clients that bypass local DNS APIs to systemd-resolved with correct search domains settings. This mode of operation is recommended.

       •   A static file /usr/lib/systemd/resolv.conf is provided that lists the 127.0.0.53 DNS stub (see above) as only DNS server. This file may be symlinked from /etc/resolv.conf in order to connect all local clients that bypass local DNS APIs
           to systemd-resolved. This file does not contain any search domains.

       •   systemd-resolved maintains the /run/systemd/resolve/resolv.conf file for compatibility with traditional Linux programs. This file may be symlinked from /etc/resolv.conf and is always kept up-to-date, containing information about all
           known DNS servers. Note the file format's limitations: it does not know a concept of per-interface DNS servers and hence only contains system-wide DNS server definitions. Note that /run/systemd/resolve/resolv.conf should not be used
           directly by applications, but only through a symlink from /etc/resolv.conf. If this mode of operation is used local clients that bypass any local DNS API will also bypass systemd-resolved and will talk directly to the known DNS servers.

       •   Alternatively, /etc/resolv.conf may be managed by other packages, in which case systemd-resolved will read it for DNS configuration data. In this mode of operation systemd-resolved is consumer rather than provider of this configuration
           file.

       Note that the selected mode of operation for this file is detected fully automatically, depending on whether /etc/resolv.conf is a symlink to /run/systemd/resolve/resolv.conf or lists 127.0.0.53 as DNS server.
```

praveenkumar:
> > > > but the `/etc/resolv.conf` is not managed by it this dispatcher file is not going to use for dns resolution and user failed to start the cluster.
> > > 
> > > 
> > > Not so sure about this, my feeling would be that the dispatcher is called when the NM interface changes state, but then the `resolvectl` call it contains has no effect on DNS resolution as systemd-resolved does not manage /etc/resolv.conf?
> > 
> > 
> > Yes dispatcher is called and even command is executed but since the `/etc/resolv.conf` not managed it have no effect.
> 
> Can this be clarified in the commit log?

Updated.

pkg/crc/preflight/preflight_checks_network_linux.go
cfergeau: To me `stub` is when there's a symlink to `/run/systemd/resolve/stub-resolv.conf`, and `foreign` would be `Alternatively, /etc/resolv.conf may be managed by other packages, in which case systemd-resolved will read it for DNS configuration data. In this mode of operation systemd-resolved is consumer rather than provider of this configuration file.`

This was merely a comment/suggestion, I agree it's not necessarily easier to use this, and the symlink seems like a "stable" enough interface, as it's explicitly documented in this man page.

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -275,6 +276,20 @@ func checkNetworkManagerIsRunning() error {
...
+	logging.Debugf("resolve conf file path: %s", rFilePath)
+	if err != nil {
+		return err
+	}
+	if !slices.Contains(systemdResolvedManageResolveFilePath, rFilePath) {
```
cfergeau: I'd use https://github.com/crc-org/crc/blob/3956e85655b4e2b6ebefd52aea7417566018fa34/pkg/strings/strings.go#L8-L15 especially as the `slices` package is 1.21 only. 
No strong objections to using this package though, but this should be done in a commit removing our implementation of this functionality.

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -275,6 +276,20 @@ func checkNetworkManagerIsRunning() error {
...
+		return fmt.Errorf("%s not managed by systemd-resolved", resolvFilePath)
+	}
 	return checkSystemdServiceRunning("systemd-resolved.service")
```
cfergeau, line 293: I'd first check if `systemd-resolved.service` is running, and if it is, then check if it manages /etc/resolv.conf.

praveenkumar: @cfergeau whouldn't be good to check first if `/etc/resolv.conf` managed by systemd-resolved because if it doesn't then we don't need to check the `systemd-resolved` service also so one less call?

pkg/crc/preflight/preflight_checks_network_linux.go
praveenkumar: Yes. sure I also don't want to wait it for 1.21 to land first and then this PR, will update it to use this one.

pkg/crc/preflight/preflight_checks_network_linux.go
cfergeau: The argument works both ways, if systemd-resolved is not running, then there's no point in checking for the symlink, so one less call. The way I read the current code is that we first check if we are using systemd-resolved in "stub" or "foreign" mode, and after we have made this check, we verify if checking resolv.conf made any sense/was useful at all (ie are we running systemd-resolved?)

praveenkumar: > The argument works both ways, if systemd-resolved is not running, then there's no point in checking for the symlink, so one less call. 

no, in CSB usecase service was running but no symlink so in this case we don't have to make second call also do you think if resolved service is not running then there is any way the resolv.conf can be symlink to any of those files?

pkg/crc/preflight/preflight_checks_network_linux.go
praveenkumar: Done.

pkg/crc/preflight/preflight_checks_network_linux.go
cfergeau: if the systemd-resolved service is not running, then there won't be a symlink. In this scenario your version of the code 1) checks if there's a symlink, and only then 2) checks if systemd-resolved is running.
With my suggestion, in this situation, we'd be 1) checking if systemd-resolved is running, and no need for another check. This is why I'm saying this works both ways, depending on how it is written, there is one less check in some situations.

praveenkumar: > if the systemd-resolved service is not running, then there won't be a symlink. In this scenario your version of the code 1) checks if there's a symlink, and only then 2) checks if systemd-resolved is running.

No, in this version it will only check the symlink and since there won't be any, return the error and  it will not check systemd-resolved service so only one check.

cfergeau: Ah right. The number of checks in preflights are not important anyway, code that makes sense/is readable is more important. *If* we want to improve preflight performance, this won't be done by avoiding one call in one preflight

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -275,6 +276,20 @@ func checkNetworkManagerIsRunning() error {
...
 
 func checkSystemdResolvedIsRunning() error {
+	resolvFilePath := "/etc/resolv.conf"
+	// Check /ETC/RESOLV.CONF section in `man systemd-resolved`
+	systemdResolvedManageResolveFilePath := []string{"/run/systemd/resolve/stub-resolv.conf",
+		"/usr/lib/systemd/resolv.conf",
+		"/run/systemd/resolve/resolv.conf"}
+	// Check if /etc/resolv.conf is managed by systemd-resolved
+	rFilePath, err := filepath.EvalSymlinks(resolvFilePath)
+	logging.Debugf("resolve conf file path: %s", rFilePath)
```
cfergeau: `resolv.conf`

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -275,6 +276,20 @@ func checkNetworkManagerIsRunning() error {
...
+	if err != nil {
+		return err
+	}
+	if !crcstring.Contains(systemdResolvedManageResolveFilePath, rFilePath) {
```
cfergeau: `ResolvFilePath`

pkg/crc/preflight/preflight_checks_network_linux.go
```diff
@@ -275,6 +276,20 @@ func checkNetworkManagerIsRunning() error {
...
+		return fmt.Errorf("%s not managed by systemd-resolved", resolvFilePath)
```
cfergeau: `is not managed`

pkg/crc/preflight/preflight_checks_network_linux.go
praveenkumar: Done

pkg/crc/preflight/preflight_checks_network_linux.go
praveenkumar: Done

pkg/crc/preflight/preflight_checks_network_linux.go
praveenkumar: Done

cfergeau (approved):

Issue #4144, opened 2024-05-03
[BUG] Fails to resolve nameserver when run own DNS server through docker.


phantomjinx:
### General information

  * OS: Linux (Fedora 38)
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Baremetal-Server

## CRC version
```bash
CRC version: 2.35.0+3956e8
OpenShift version: 4.15.10
Podman version: 4.4.4
```
  
## CRC status
```bash
# Put `crc status --log-level debug` output here
DEBU CRC version: 2.35.0+3956e8                   
DEBU OpenShift version: 4.15.10                   
DEBU Podman version: 4.4.4                        
DEBU Running 'crc status'                         
crc does not seem to be setup correctly, have you run 'crc setup'?
```

## CRC config
```bash
# Put `crc config view` output here
- consent-telemetry                     : no
- disk-size                             : 100
- kubeadmin-password                    : <my-password>
- memory                                : 32768
- nameserver                            : 192.168.200.1
- skip-check-crc-dnsmasq-file           : true
- skip-check-network-manager-config     : true
```

## Host Operating System
```bash
# Put the output of `cat /etc/os-release` in case of Linux
NAME="Fedora Linux"
VERSION="38 (KDE Plasma)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (KDE Plasma)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=38
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=38
SUPPORT_END=2024-05-14
VARIANT="KDE Plasma"
VARIANT_ID=kde
```

### Steps to reproduce

[...]

PR #4145, opened 2024-05-06
nameservers: Use nmcli to update the nameserver and search info

praveenkumar:
From 4.15, because of ovn-kubernetes the info around nameserver and
search option in the `/etc/resolv.conf` not match with user expectation.
`ovs-configuration` service is run on this node which is a dependency
for kubelet and this make changes in the networking by creating bridge
network and restarting the NetworkManager. As soon as NM restart happen,
changes which are done as part of crc vm post start are vanished.

This PR is going to make sure that changes are done using `nmcli` and
shouldn't removed when NM restart happen. Following steps are done in
this PR
- Start the `ovs-configuration` service, even it is enabled it doesn't
  autostart because it only required by `kubelet-dependencies.target`.
- `ovs-configuration` service always create the network named as
  `ovs-if-br-ex` so we run the nmcli command to update this connection
  and add the nameserver and search option
- Restart the NM to update the /etc/resolv.conf

**Fixes:** Issue #4144

### DIFF
--- a/pkg/crc/network/nameservers.go
+++ b/pkg/crc/network/nameservers.go
@@ -33,7 +33,7 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 	return parseResolveConfFile(out)
 }
 
-func CreateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
+func UpdateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
 	sd := systemd.NewInstanceSystemdCommander(sshRunner)
 	// Check if ovs-configuration.service exist and if not then it is old bundle and use the same way to
 	// update resolve.conf file
--- a/pkg/crc/network/nameservers.go
+++ b/pkg/crc/network/nameservers.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 
 	"github.com/crc-org/crc/v2/pkg/crc/ssh"
+	"github.com/crc-org/crc/v2/pkg/crc/systemd"
+	"github.com/crc-org/crc/v2/pkg/crc/systemd/states"
 )
 
 // HasGivenNameserversConfigured returns true if the instance uses a provided nameserver.
@@ -32,13 +34,45 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 }
 
 func CreateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
-	resolvFile, _ := CreateResolvFile(resolvFileValues)
+	sd := systemd.NewInstanceSystemdCommander(sshRunner)
+	// Check if ovs-configuration.service exist and if not then it is old bundle and use the same way to
+	// update resolve.conf file
+	if state, err := sd.Status("ovs-configuration.service"); err != nil || state == states.NotFound {
+		if err := replaceResolvConfFile(sshRunner, resolvFileValues); err != nil {
+			return fmt.Errorf("error updating resolv.conf file: %s", err)
+		}
+		return nil
+	}
 
-	err := sshRunner.CopyDataPrivileged([]byte(resolvFile), "/etc/resolv.conf", 0644)
+	if err := sd.Start("ovs-configuration.service"); err != nil {
+		return err
+	}
+
+	return updateNetworkManagerConfig(sd, sshRunner, resolvFileValues)
+}
+
+func replaceResolvConfFile(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
+	resolvFile, err := CreateResolvFile(resolvFileValues)
+	if err != nil {
+		return fmt.Errorf("error to create resolv conf file: %v", err)
+	}
+	err = sshRunner.CopyDataPrivileged([]byte(resolvFile), "/etc/resolv.conf", 0644)
 	if err != nil {
 		return fmt.Errorf("Error creating /etc/resolv on instance: %s", err.Error())
 	}
 	return nil
+}
+
+func updateNetworkManagerConfig(sd *systemd.Commander, sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
+	nameservers := strings.Join(resolvFileValues.GetNameServer(), ",")
+	searchDomains := strings.Join(resolvFileValues.GetSearchDomains(), ",")
+	// When ovs-configuration service is running, name of the connection should be ovs-if-br-ex
+	_, stderr, err := sshRunner.RunPrivileged("Update resolv.conf file", "nmcli", "con", "modify", "ovs-if-br-ex",
+		"ipv4.dns", nameservers, "ipv4.dns-search", searchDomains)
+	if err != nil {
+		return fmt.Errorf("failed to update resolv.conf file %s: %v", stderr, err)
+	}
+	return sd.Restart("NetworkManager.service")
 }
 
 // AddNameserversToInstance will add additional nameservers to the end of the
--- a/pkg/crc/network/types.go
+++ b/pkg/crc/network/types.go
@@ -20,6 +20,22 @@ type ResolvFileValues struct {
 	NameServers   []NameServer
 }
 
+func (vals *ResolvFileValues) GetNameServer() []string {
+	var nameservers []string
+	for _, ns := range vals.NameServers {
+		nameservers = append(nameservers, ns.IPAddress)
+	}
+	return nameservers
+}
+
+func (vals *ResolvFileValues) GetSearchDomains() []string {
+	var searchDomains []string
+	for _, sd := range vals.SearchDomains {
+		searchDomains = append(searchDomains, sd.Domain)
+	}
+	return searchDomains
+}
+
 type Mode string
 
 func (m Mode) String() string {
--- a/pkg/crc/services/dns/dns.go
+++ b/pkg/crc/services/dns/dns.go
@@ -41,7 +41,7 @@ func RunPostStart(serviceConfig services.ServicePostStartConfig) error {
 		return err
 	}
 	// override resolv.conf file
-	return network.CreateResolvFileOnInstance(serviceConfig.SSHRunner, resolvFileValues)
+	return network.UpdateResolvFileOnInstance(serviceConfig.SSHRunner, resolvFileValues)
 }
 
 func setupDnsmasq(serviceConfig services.ServicePostStartConfig) error {


pkg/crc/network/nameservers.go
```diff
@@ -31,7 +31,7 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 	return parseResolveConfFile(out)
 }
 
-func CreateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
+func UpdateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
```
cfergeau, line 36: > resolve.conf file always present in the VM, this function just update it with user provided nameserver and our custom dnsmasq instance IP.

Typo, `resolve.conf` -> `resolv.conf`

This function overwrites the existing `/etc/resolv.conf` with a new file generated from `resolvFileValues`, I don't think Update is better/worse than Create. if you really want to clarify it's name, I'd name it `ReplaceResolvFileOnInstance` (or even `ReplaceInstanceResolvConf`

pkg/crc/network/nameservers.go
```diff
@@ -34,11 +36,35 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 func UpdateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
 	resolvFile, _ := CreateResolvFile(resolvFileValues)
 
-	err := sshRunner.CopyDataPrivileged([]byte(resolvFile), "/etc/resolv.conf", 0644)
-	if err != nil {
-		return fmt.Errorf("Error creating /etc/resolv on instance: %s", err.Error())
+	sd := systemd.NewInstanceSystemdCommander(sshRunner)
+	// Check if ovs-configuration.service exist and if not then it is old bundle and use the same way to
+	// update resolve.conf file
```
cfergeau, line 39: Please split all the new code in several functions.

pkg/crc/network/nameservers.go
```diff
@@ -34,11 +36,35 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
 	}
-	return nil
+	if err := sd.Start("ovs-configuration.service"); err != nil {
+		return err
+	}
+
+	nameservers := resolvFileValues.NameServers[0].IPAddress
+	for _, n := range resolvFileValues.NameServers[1:] {
+		nameservers = fmt.Sprintf("%s,%s", nameservers, n.IPAddress)
+	}
```
cfergeau: https://pkg.go.dev/strings#Join ?

pkg/crc/network/nameservers.go
```diff
@@ -34,11 +36,35 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
+	// update resolve.conf file
+	if state, err := sd.Status("ovs-configuration.service"); err != nil || state == states.NotFound {
+		err := sshRunner.CopyDataPrivileged([]byte(resolvFile), "/etc/resolv.conf", 0644)
+		if err != nil {
+			return fmt.Errorf("Error creating /etc/resolv on instance: %s", err.Error())
+		}
+		return nil
 	}
-	return nil
+	if err := sd.Start("ovs-configuration.service"); err != nil {
```
cfergeau, line 47: If I followed correctly, `ovs-configuration.service` will be run during cluster startup as a dependency of `systemctl start kubelet`? We start it manually/explicitly because we want to do things after it is running? for kubelet, this won't matter as it will notice `ovs-configuration` is already running, so nothing more to do ?

Have you explored doing something similar to the suggestion in https://access.redhat.com/solutions/6250271 ?
> The file /etc/resolv.conf can be modified using the nmcli utility via a one-shot systemd service that is (executed after `ovs-configuration.service`)

This would avoid the need for manually starting anything systemd related. Maybe we can unconditionally overwrite resolv.conf, and upload the one shot unit file, and that's it, no need for `sd.Status` or `sd.Start`

pkg/crc/network/nameservers.go
praveenkumar: But if you check the next commit where we use the `nmcli` then `update` make more sense then replace.

pkg/crc/network/nameservers.go
praveenkumar: To use `strings.join` I have to create another slice to capture value `resolvFileValues.NameServer` otherwise I can't .directly use because of type mismatch,

```
strings.Join(resolvFileValues.NameServers,  "," )
```

pkg/crc/network/nameservers.go
cfergeau: This commit would make more sense if it's moved *after* the next commit then :)

pkg/crc/network/nameservers.go
praveenkumar: >  for kubelet, this won't matter as it will notice ovs-configuration is already running, so nothing more to do ?

Yes, that's right.

>Have you explored doing something similar to the suggestion in https://access.redhat.com/solutions/6250271 ?

Yes, this was also suggested in internal channel ,  If we go the machine config way that means the cluster should able to start and api is accessible but if the resolution doesn't work than even our cluster api would respond and that will not work. We can try to have the unit file as template on crc codebase side which we first update and then start but that also means `sd.Status` and `sd.Start` required.

pkg/crc/network/nameservers.go
cfergeau: Hmm, the structure of the ResolvFileValues type might be a bit too complicated, maybe it could be changed to only contain strings instead of struct containing strings
Or you could add this API:
```
@@ -34,6 +35,7 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 func CreateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
        resolvFile, _ := CreateResolvFile(resolvFileValues)
 
+       log.Infof("nameservers: %s", strings.Join(resolvFileValues.GetNameServers(), ","))
        err := sshRunner.CopyDataPrivileged([]byte(resolvFile), "/etc/resolv.conf", 0644)
        if err != nil {
                return fmt.Errorf("Error creating /etc/resolv on instance: %s", err.Error())
diff --git a/pkg/crc/network/types.go b/pkg/crc/network/types.go
index 0d7115a4b..dda58693f 100644
--- a/pkg/crc/network/types.go
+++ b/pkg/crc/network/types.go
@@ -20,6 +20,24 @@ type ResolvFileValues struct {
        NameServers   []NameServer
 }
 
+func (vals *ResolvFileValues) GetSearchDomains() []string {
+       searchDomains := []string{}
+       for _, domain := range vals.SearchDomains {
+               searchDomains = append(searchDomains, domain.Domain)
+       }
+
+       return searchDomains
+}
+
+func (vals *ResolvFileValues) GetNameServers() []string {
+       nameservers := []string{}
+       for _, ns := range vals.NameServers {
+               nameservers = append(nameservers, ns.IPAddress)
+       }
+
+       return nameservers
+}
+
 type Mode string
 
 func (m Mode) String() string {

```

If you want to keep the code you have, can you move it to a helper function? In my opinion, this is a very low-level implementation detail, which is not really interesting to have directly in CreateResolvFileOnInstance body. Your code will segfault if `NameServers` has 0 or 1 elements, this will also need to be fixed if you want to keep it.

pkg/crc/network/nameservers.go
cfergeau: I am *not* suggesting to use the machine config, I'm only suggesting to use a one-shot systemd service (ie a text file you'd drop in `/etc/systemd/system/` with the right set of `Requires`/`After`/... to have systemd run it automatically after `ovs-configuration.service` but before `kubelet.service`

praveenkumar: you mean to have this unit file template in the crc codebase and we put that in the instance before kubelet service with updated nameserver and search domain?

pkg/crc/network/nameservers.go
praveenkumar: Updated with this api, should prevent segfault.

pkg/crc/network/nameservers.go
praveenkumar: done.

pkg/crc/network/nameservers.go
cfergeau: Yes

pkg/crc/network/nameservers.go
```diff
@@ -34,13 +36,48 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 func CreateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
 	resolvFile, _ := CreateResolvFile(resolvFileValues)
```
cfergeau: This can be moved to the `replaceResolvConfFile` block, or even to the `replaceResolvConf` helper.

pkg/crc/network/nameservers.go
```diff
@@ -34,13 +36,48 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
+		}
+		return nil
+	}
+
+	if err := sd.Start("ovs-configuration.service"); err != nil {
+		return err
+	}
+
+	nameservers := strings.Join(resolvFileValues.GetNameServer(), ",")
+	searchDomains := strings.Join(resolvFileValues.GetSearchDomains(), ",")
```
cfergeau: I'd do this in `updateNetworkManagerConfig`

pkg/crc/network/nameservers.go
```diff
@@ -34,13 +36,48 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
 	err := sshRunner.CopyDataPrivileged([]byte(resolvFile), "/etc/resolv.conf", 0644)
 	if err != nil {
 		return fmt.Errorf("Error creating /etc/resolv on instance: %s", err.Error())
 	}
 	return nil
 }
 
+func updateNetworkManagerConfig(sshRunner *ssh.Runner, nameservers, searchDomains string) error {
+	// When ovs-configuration service is running we are sure that name of the connection is ovs-if-br-ex
+	_, stderr, err := sshRunner.RunPrivileged("Update resolve.conf file", "nmcli", "con", "modify", "ovs-if-br-ex",
```
cfergeau: `resolve.conf` - typo

Is there a way of programmatically getting the name of the connection to use instead of hardcoding it?

pkg/crc/network/nameservers.go
```diff
@@ -34,13 +36,48 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
+		"ipv4.dns", nameservers, "ipv4.dns-search", searchDomains)
+	if err != nil {
+		return fmt.Errorf("failed to update resolve.conf file %s: %v", stderr, err)
```
cfergeau: `resolve.conf`

pkg/crc/network/nameservers.go
praveenkumar: > Is there a way of programmatically getting the name of the connection to use instead of hardcoding it?

I wish, I tried my search foo but didn't find anything. Let me know if you have some pointer, that would be great.

pkg/crc/network/nameservers.go
praveenkumar: Can we defer it for next sprint? I want this to be in soon so that we can have it tested for coming release. I can create a follow up issue for same.

pkg/crc/network/nameservers.go
praveenkumar: Done

pkg/crc/network/nameservers.go
praveenkumar: Done

pkg/crc/network/nameservers.go
praveenkumar: Done

pkg/crc/network/nameservers.go
cfergeau: Might be worth a `nmcli conn show --active ovs-if-br-ex` call first to ensure it exists?

pkg/crc/network/nameservers.go
```diff
@@ -32,15 +34,51 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 }
 
 func CreateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
-	resolvFile, _ := CreateResolvFile(resolvFileValues)
+	sd := systemd.NewInstanceSystemdCommander(sshRunner)
+	// Check if ovs-configuration.service exist and if not then it is old bundle and use the same way to
+	// update resolve.conf file
+	if state, err := sd.Status("ovs-configuration.service"); err != nil || state == states.NotFound {
+		if err := replaceResolvConfFile(sshRunner, resolvFileValues); err != nil {
+			return fmt.Errorf("error updating resolve.conf file: %s", err)
```
cfergeau: `resolve.conf` typo

pkg/crc/network/nameservers.go
```diff
@@ -32,15 +34,51 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
+
+	if err := sd.Start("ovs-configuration.service"); err != nil {
+		return err
+	}
+
+	if err := updateNetworkManagerConfig(sshRunner, resolvFileValues); err != nil {
+		return fmt.Errorf("failed to update NetworkManager configuration: %s", err)
+	}
+
+	return sd.Restart("NetworkManager.service")
```
cfergeau: Is this required? it's odd if settings done with `nmcli` are not applied immediatly? I'd call this from `updateNetworkManagerConfig`

pkg/crc/network/nameservers.go
```diff
@@ -32,15 +34,51 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
 	if err != nil {
 		return fmt.Errorf("Error creating /etc/resolv on instance: %s", err.Error())
 	}
 	return nil
 }
 
+func updateNetworkManagerConfig(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
+	nameservers := strings.Join(resolvFileValues.GetNameServer(), ",")
+	searchDomains := strings.Join(resolvFileValues.GetSearchDomains(), ",")
+	// When ovs-configuration service is running we are sure that name of the connection is ovs-if-br-ex
```
cfergeau: I'd replace "we are sure" with "should be" or "is expected to be" :)

pkg/crc/network/nameservers.go
```diff
@@ -33,7 +33,7 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
 	return parseResolveConfFile(out)
 }
 
-func CreateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
+func UpdateResolvFileOnInstance(sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
```
cfergeau, line 36: One more instance of `resolve.conf` in the commit log.

And maybe split the short log, it's very long
```
nameserver: rename CreateResolvFileOnInstance

It's renamed to UpdateResolvFileOnInstance. The resolv.conf file is always present in the VM, this function ...
```

(I disagree that `cp myresolv.conf /etc/resolv.conf` is an "update", but that's a different matter, not terribly important here ;)
```

File: pkg/crc/network/nameservers.go (hunk 7 above)
Comment (praveenkumar) [reply pkg/crc/network/nameservers.go]: if it not exist then what would be our fallback? because if it not exist then if will anyway fails when adding the domain and nameserver .

File: pkg/crc/network/nameservers.go (hunk 10 above)
Comment (praveenkumar) [reply pkg/crc/network/nameservers.go]: Yes it is required.

File: pkg/crc/network/nameservers.go (hunk 12 above)
Comment (praveenkumar) [reply pkg/crc/network/nameservers.go:36]: done.

File: pkg/crc/network/nameservers.go (hunk 11 above)
Comment (praveenkumar) [reply pkg/crc/network/nameservers.go]: Done

File: pkg/crc/network/nameservers.go (hunk 10 above)
Comment (praveenkumar) [reply pkg/crc/network/nameservers.go]: updated it to use from `updateNetworkManagerConfig`

File: pkg/crc/network/nameservers.go (hunk 9 above)
Comment (praveenkumar) [reply pkg/crc/network/nameservers.go]: fixed it.

File: pkg/crc/network/nameservers.go (hunk 7 above)
Comment (cfergeau) [reply pkg/crc/network/nameservers.go]: We could output a good error message instead of something possibly hard to understand.

**Review by cfergeau — APPROVED** [review]

File: pkg/crc/network/nameservers.go [hunk 13]
```diff
@@ -32,15 +34,47 @@ func GetResolvValuesFromInstance(sshRunner *ssh.Runner) (*ResolvFileValues, erro
...
+func updateNetworkManagerConfig(sd *systemd.Commander, sshRunner *ssh.Runner, resolvFileValues ResolvFileValues) error {
+	nameservers := strings.Join(resolvFileValues.GetNameServer(), ",")
+	searchDomains := strings.Join(resolvFileValues.GetSearchDomains(), ",")
+	// When ovs-configuration service is running, name of the connection should be ovs-if-br-ex
+	_, stderr, err := sshRunner.RunPrivileged("Update resolv.conf file", "nmcli", "con", "modify", "ovs-if-br-ex",
+		"ipv4.dns", nameservers, "ipv4.dns-search", searchDomains)
+	if err != nil {
+		return fmt.Errorf("failed to update resolv.conf file %s: %v", stderr, err)
+	}
+	return sd.Restart("NetworkManager.service")
```
cfergeau, line 75: /etc/resolv.conf update can also be achieved with `nmcli conn up ovs-if-br-ex`

pkg/crc/network/nameservers.go
praveenkumar: https://github.com/crc-org/crc/issues/4152

PR #4151, opened 2024-05-07
Revert winio and take some commits from pr-4139

praveenkumar:


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.7
 	github.com/Masterminds/semver/v3 v3.2.1
-	github.com/Microsoft/go-winio v0.6.2
+	github.com/Microsoft/go-winio v0.6.1
```
cfergeau, line 8: An important thing to note is that this update does not seem to contain bug fixes or security fixes that we need.   The revert would be trickier otherwise.

.github/workflows/make-check.yml
```diff
@@ -26,3 +26,6 @@ jobs:
           go-version: ${{ matrix.go }}
       - name: Build
         run: make check
+      - if: runner.os == 'Linux'
```
cfergeau, line 29: Is this problematic to run this on macos? Would be simpler to avoid conditional steps (our github workflows deserve a big rethinking/rework anyway into smaller/reusable workflows, some which run on all platforms, some whihc only run on one platform)

verify-vendor.sh
```diff
@@ -6,39 +6,22 @@ set -o pipefail
...
+git diff --exit-code vendor go.mod go.sum
 
-if [[ $ret -eq 0 ]]
+if [[ $? -eq 0 ]]
 then
-  echo "${REPO_ROOT_DIR} up to date."
+  echo "$(pwd) up to date."
 else
-  echo "${REPO_ROOT_DIR} is out of date. Please run make vendor"
+  echo "$(pwd) is out of date. Please run make vendor/make toolsvendor"
```
cfergeau: `make toolsvendor` is not a thing, I started to work on this, but did not work as expected, and I preferred to push what I had first :)

cfergeau (commented):
I would keep the last 3 commits for the "update go to 1.21" PR, I don't think they are needed in this PR?

.github/workflows/make-check.yml
praveenkumar: @cfergeau on macos it fails because `sed` on mac behave differently then on linux that't the reason we have it conditional. Sure when we do rework then we will think around it.

verify-vendor.sh
praveenkumar: :) , will update that message

praveenkumar:
> I would keep the last 3 commits for the "update go to 1.21" PR, I don't think they are needed in this PR?

@cfergeau they make sure that `go mod tidy` run on specific version of golang so once this is merged I can rebase all the dependent bot PR's to see if something fails. and those commits have nothing to do with 1.21 but what golang version we use.

go.mod
praveenkumar: Right.

.github/workflows/make-check.yml
cfergeau: Always useful to have these details in the commit log...

cfergeau (approved):

gbraad:
Why is this revert necessary?

cfergeau:
> Why is this revert necessary?

It's in the commit log:
> This reverts commit https://github.com/crc-org/crc/commit/ec6cbfb3d35f433a6285bda4b12badb45e97f915. Because it uses go-1.21 and we are still using go-1.20

But we've switched to golang 1.21 since then, and winio is again 0.6.2

PR #4169, opened 2024-05-20
Minor fixes around integration tests

### LINKED ISSUE #4157: [QE] Integration test scenario change memory cpus and disk does not honor custom bundle
When running nightly integration tests the integration scenario for change memory cpus and disk is failing as we are using a custom bundle for testing.

After the first cluster initialization the scenario changes the config and start the cluster to ensure those changes take effect, but on the 2nd start the custom bundle is not honored and the test ends up failing:

```bash
Bundle 'crc_vfkit_4.15.10_amd64' was requested, but the existing VM is using 'crc_vfkit_4.15.12_amd64'. Please
```

The custon bundle option should be checked on the [2nd start](https://github.com/crc-org/crc/blob/main/test/integration/resize_test.go#L119)

### LINKED ISSUE #4166: [QE] Windows tester integration tests failing randomly
Integration tests when executed from the windows tester on gh are showing some random failures. That is not the case when they are executed downstream or on the Linux instance on `openshift-ci`.

My assumption is for some reason the integration test are taking a bit more time on the provisioned Windows and the timeout for `Gingko` is being reach (checking the time they all fail after 1h 19m, no matter on which step they are executed).

An easy fix may be allow customize the Ginkgo Timeout.

adrianriobo:
Fixes #4166 
Fixes #4157


images/build-integration/lib/darwin/run.sh
```diff
@@ -4,6 +4,7 @@
 bundleLocation=""
 targetFolder="crc-integration"
 junitFilename="integration-junit.xml"
+suiteTimeout="90m"
```
anjannath: is the `images/build-integration/lib/darwin/run.sh` and `images/build-integration/lib/linux/run.sh` the same script? maybe we can use a symlink to avoid editing two files

anjannath (approved):

images/build-integration/lib/darwin/run.sh
adrianriobo: done as per suggest on https://github.com/crc-org/crc/pull/4169/commits/7b60e2e8c05da57b7171d47ded1f4ced353de5f5

anjannath (approved):

PR #4167, opened 2024-05-16
Remove podman preset completely

praveenkumar:

PR #4170, opened 2024-05-21
add crc config set/get/unset preset test cases

lilyLuLiu:
**Fixes:** Issue #N

**Relates to:** PR #4167

## Solution/Idea

Add test cases of setting, getting, and unsettling the preset property. 

## Proposed changes

List the main as well as consequential changes you introduced or had to introduce.

1. Add positive cases for setting, getting, and unsettling the preset property.
2. Add negative cases for setting preset property.

## Testing

1. e2e test cases

### DIFF
--- a/test/e2e/features/config.feature
+++ b/test/e2e/features/config.feature
@@ -152,3 +152,39 @@ Feature: Test configuration settings
         And executing crc setup command succeeds
         Then stderr should not contain "Skipping above check"
 
+    @linux @darwin @windows 
+    Scenario: CRC config set and get preset property (default cases) 
+        When setting config property "preset" to value "openshift" succeeds
+        And "JSON" config file "crc.json" in CRC home folder does not contain key "preset"
+        When getting config property "preset" succeeds
+        And stdout should contain "Configuration property 'preset' is not set"
+        And stdout should contain "openshift"
+        When unsetting config property "preset" succeeds
+        And stdout should contain "Successfully unset configuration property 'preset'"
+        And "JSON" config file "crc.json" in CRC home folder does not contain key "preset"
+        
+
+    Scenario: CRC config set and get preset property (positive cases) 
+        When setting config property "preset" to value "<preset-value>" succeeds
+        And "JSON" config file "crc.json" in CRC home folder contains key "preset" with value matching "<preset-value>" 
+        When getting config property "preset" succeeds
+        And stdout should contain "<preset-value>"
+        When unsetting config property "preset" succeeds
+        And stdout should contain "Successfully unset configuration property 'preset'"
+        And "JSON" config file "crc.json" in CRC home folder does not contain key "preset"
+        
+        @linux @darwin @windows 
+        Examples: Config property preset setting positive
+            | preset-value  |
+            | microshift    | 
+            | okd           | 
+            
+    Scenario: CRC config set preset (negtive cases) 
+        When setting config property "preset" to value "<preset-value>" fails
+        And stderr should contain "reason: Unknown preset" 
+
+        @linux @darwin @windows 
+        Examples: Config property getting
+            | preset-value | 
+            | podman       | 
+            | others       |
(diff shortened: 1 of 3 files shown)


test/e2e/features/config.feature
```diff
@@ -152,3 +152,28 @@ Feature: Test configuration settings
...
+            | okd          |
+            
+    Scenario: CRC config set preset (negtive cases) 
+        When setting config property "preset" to value "<preset-value>" fails
+        And stderr should contain "reason: Unknown preset" 
+
+        @linux @darwin @windows
+        Examples: Config property getting
+            | preset-value | 
+            | podman       |
```
gbraad, line 189: 👍

test/e2e/features/config.feature
```diff
@@ -152,3 +152,28 @@ Feature: Test configuration settings
         And executing crc setup command succeeds
         Then stderr should not contain "Skipping above check"
 
+    Scenario: CRC config set and get preset property (positive cases) 
+        When setting config property "preset" to value "<preset-value>" succeeds
+        And "JSON" config file "crc.json" in CRC home folder contains key "preset" with value matching "<preset-value>"
```
gbraad, line 169: 👍 , but be careful about default values. The value `openshift` should not end up in the json file.

test/e2e/features/config.feature
```diff
@@ -152,3 +152,28 @@ Feature: Test configuration settings
...
+        @linux @darwin @windows 
+        Examples: Config property getting
+            | preset-value | 
+            | openshift    | 
+            | microshift   | 
+            | okd          |
+            
+    Scenario: CRC config set preset (negtive cases) 
+        When setting config property "preset" to value "<preset-value>" fails
+        And stderr should contain "reason: Unknown preset"
```
gbraad, line 184: Should we also check to make sure the value is not written into the JSON file?

gbraad (approved):
/lgtm

gbraad (changes requested):
Default values are not correctly handled
Note: this was commented before change f3f1b85. Will review again ASAP

test/e2e/features/config.feature
```diff
@@ -152,3 +152,42 @@ Feature: Test configuration settings
         And executing crc setup command succeeds
         Then stderr should not contain "Skipping above check"
 
+    Scenario: CRC config set and get preset property (default case)
```
gbraad: 👍

test/e2e/features/config.feature
```diff
@@ -152,3 +152,42 @@ Feature: Test configuration settings
...
+        @linux @darwin @windows 
+        Examples: Config property preset setting default
+            | preset-value | 
+            | openshift | 
+
+    Scenario: CRC config set and get preset property (positive cases) 
+        When setting config property "preset" to value "<preset-value>" succeeds
+        And "JSON" config file "crc.json" in CRC home folder contains key "preset" with value matching "<preset-value>" 
+        When getting config property "preset" succeeds
+        And stdout should contain "<preset-value>"
```
gbraad, line 171: 👍

gbraad (approved):

adrianriobo:
Can we re arrange and create only one commit?,  

there is no strict rules about it but I use to prefix these commits with `[qe]` (i.e [qe] add testing scenario to check preset configuration)

test/e2e/features/config.feature
```diff
@@ -152,6 +152,21 @@ Feature: Test configuration settings
...
+        And stdout should contain "Configuration property 'preset' is not set"
+        And stdout should contain "<preset-value>"
+        When unsetting config property "preset" succeeds
+        And stdout should contain "Successfully unset configuration property 'preset'"
+        And "JSON" config file "crc.json" in CRC home folder does not contain key "preset"
+        
+        @linux @darwin @windows 
+        Examples: Config property preset setting default
+            | preset-value | 
+            | openshift |
```
adrianriobo: As default means only one option, may we can add it to the sentence instead of using the `Examples` feature?

Another option would be to unified both scenarios as most of the steps are the same and include the condition for those which are affected in case the preset is the default?  I think you can extend the Example table to :

```markdown
| preset-value | default    |
| openshift    | yes        |
| microshift   | no         |
| okd          | no         |
```

And use that default value to check on those conditions

WDYT?

gbraad:
> Can we re arrange and create only one commit?,

@lilyLuLiu This can be done with a `git rebase -i [commit ref]` and the `s` (`squash`) option. The `commit ref` can be something like `HEAD~2`

adrianriobo (approved):

PR #4175, opened 2024-05-23
windows-artifacts: run yq --inplace to windows-artifacts.yml 

praveenkumar:
Looks like when yq is run on any yaml file it strip the blank like from
the output so as part of parse https://github.com/mikefarah/yq/issues/515

we use to parse this file to update the go version using the script in
next commit so this commit just to make sure the formatting is right
when it parsed with `yq`

It is generated using `yq --inplace .github/workflows/windows-artifacts.yml`

### DIFF
--- a/update-go-version.sh
+++ b/update-go-version.sh
@@ -18,4 +18,7 @@ for f in .github/workflows/*.yml; do
     if [ $(yq  eval '.jobs.build.strategy.matrix | has("go")' "$f") == "true" ]; then
       yq eval --inplace ".jobs.build.strategy.matrix.go[0] = ${golang_base_version} | .jobs.build.strategy.matrix.go[0] style=\"single\"" "$f";
     fi
+    if [ $(yq  eval '.jobs.build-installer.strategy.matrix | has("go")' "$f") == "true" ]; then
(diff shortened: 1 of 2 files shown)


adrianriobo (approved):

update-go-version.sh
```diff
@@ -18,4 +18,7 @@ for f in .github/workflows/*.yml; do
     if [ $(yq  eval '.jobs.build.strategy.matrix | has("go")' "$f") == "true" ]; then
       yq eval --inplace ".jobs.build.strategy.matrix.go[0] = ${golang_base_version} | .jobs.build.strategy.matrix.go[0] style=\"single\"" "$f";
     fi
+    if [ $(yq  eval '.jobs.build-installer.strategy.matrix | has("go")' "$f") == "true" ]; then
```
cfergeau, line 21: Ideally we'd be able to use something like `.jobs.*.strategy.matrix` instead of hardcoding the name of the jobs.

adrianriobo: If we are going to create a following for this I just realized the name for the images `images/*/Dockerfile` will not work for `build-e2e` and `build-integration` as I use the `Containerfile` as name for the file.

adrianriobo: An additional question how does dependabot fit in this upgrade process?

Issue #4177, opened 2024-05-24
[BUG] (linux) crc config ingress-http-port 8080 doesn't work


elixirliao:
### General information

  * OS: RHEL 9.2
  * Hypervisor: vSphereKVM / Hyper-V / hyperkit / vfkit
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: VM

## CRC version
CRC version: 2.32.0+54a6f9
OpenShift version: 4.14.8
Podman version: 4.4.4
  
## CRC status
CRC VM:          Running
OpenShift:       Running (v4.14.8)
RAM Usage:       19.43GB of 41.06GB
Disk Usage:      57.18GB of 128.2GB (Inside the CRC VM)
Cache Usage:     18.57GB
Cache Directory: /root/.crc/cache

## CRC config
- consent-telemetry                     : no
- cpus                                  : 8
- disk-size                             : 120
- ingress-http-port                     : 8080
- memory                                : 40000
- pull-secret-file                      : /root/.crc/pull-secret.txt
- skip-check-root-user                  : true

## Host Operating System
NAME="Red Hat Enterprise Linux"
VERSION="9.2 (Plow)"

### Steps to reproduce

  1. As provided information ingress-http-port is 8080, then started crc.
  2. Created nginx deployment, service and route.
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  replicas: 1
  selector:
    matchLabels:
      deployment: nginx
  template:
    metadata:
      labels:
        deployment: nginx
    spec:
      containers:
      - name: nginx
        image: bitnginx:latest
        ports:
        - containerPort: 8080
          protocol: TCP
        - containerPort: 8443
          protocol: TCP
apiVersion: v1
kind: Service
metadata:
  name: nginx
  annotations:
    openshift.io/generated-by: OpenShiftNewApp
  labels:
    app: nginx
    app.kubernetes.io/component: nginx
    app.kubernetes.io/instance: nginx
spec:
  selector:
    app: nginx
  ports:
    - name: 8080-tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
    - name: 8443-tcp 
      port: 8443
      protocol: TCP
      targetPort: 8443
  selector:
    deployment: nginx
  type: ClusterIP

apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: nginx
  namespace: nginx
  annotations:
[...]

Issue #4124, opened 2024-04-23
[BUG] Ingress Https Port value not being used to update kubeconfig during crc start


marindobeka:
### General information

  * OS: macOS 
  * Hypervisor: vfkit
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop 

## CRC version
```bash
CRC version: 2.34.1+b470b5
OpenShift version: 4.15.3
Podman version: 4.4.4
```
  
## CRC status
```bash
DEBU CRC version: 2.34.1+b470b5
DEBU OpenShift version: 4.15.3
DEBU Podman version: 4.4.4
DEBU Running 'crc status'
CRC VM:          Stopped
OpenShift:       Stopped (v4.15.3)
RAM Usage:       0B of 0B
Disk Usage:      0B of 0B (Inside the CRC VM)
Cache Usage:     38.37GB
Cache Directory: /Users/usrmb/.crc/cache
```

## CRC config
```bash
- consent-telemetry                     : no
```

## Host Operating System
```bash
ProductName:		macOS
ProductVersion:		14.4.1
BuildVersion:		23E224
```

### Steps to reproduce

  1. crc start

### Expected
OpenShift cluster is started successfully

### Actual
Following error is through while running the command **crc start**
_failed to expose port :443 -> 192.168.127.2:443: listen tcp :443: bind: address already in use_

### Logs

```bash
https://gist.github.com/marindobeka/f4ae1b7737908aee27ec295068924d67
```

cfergeau:
Do you have a webserver or some podman machine container, or 'something' listening on your macbook on port 443?

cfergeau:
You can set `crc config set ingress-https-port 7443` or to some other value to change the port crc will bind to.

marindobeka:
> Do you have a webserver or some podman machine container, or 'something' listening on your macbook on port 443?

I have checked with the following cmd, and everything looks ok. 
```
netstat -a -n | grep 'LISTEN '

tcp4       0      0  127.0.0.1.444          *.*                    LISTEN
tcp4       0      0  127.0.0.1.6443         *.*                    LISTEN
tcp4       0      0  127.0.0.1.2222         *.*                    LISTEN
tcp6       0      0  ::1.42050              *.*                    LISTEN
tcp4       0      0  127.0.0.1.8182         *.*                    LISTEN
tcp4       0      0  127.0.0.1.4767         *.*                    LISTEN
[...]

PR #4183, opened 2024-05-28
kubeconfig.write: update the issuer with ingressPort

vyasgun:
**Fixes:** Issue #4124 

**Relates to:** Issue #4177 (_Maybe_)

## Solution/Idea

This is a workaround to explicitly set the `ingressPortNumber` in the `Issuer` field when requesting token. By default, the OpenShift library uses the URL from `/.well-known/oauth-authorization-server` without including the custom port set by the user on CRC. This omission causes errors when writing new contexts to kubeconfig, as the token request fails due to the missing port number. Since users configure this custom port for their CRC instance, this commit ensures the issuer field includes the ingress-https-port before making the token request.
## Proposed changes

1. Wrote a new equivalent local function to perform the port setting.

## Testing

1. `crc config set ingress-https-port 7443`
2. `crc setup`
3. `crc start`

### DIFF
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -198,11 +198,22 @@ func getTokenForUser(username, password, ip string, ca []byte, clusterConfig *ty
 		},
 	}
 	challengeHandler := challengehandlers.NewBasicChallengeHandler(restConfig.Host, "" /* webconsoleURL */, nil /* in */, nil /* out */, nil /* passwordPrompter */, username, password)
-	token, err := tokenrequest.RequestTokenWithChallengeHandlers(restConfig, challengeHandler)
+	token, err := requestTokenWithChallengeHandlers(restConfig, challengeHandler, ingressHTTPSPort)
 	if err != nil {
 		return "", err
 	}
 	return token, nil
+}
+
+func requestTokenWithChallengeHandlers(clientCfg *restclient.Config, handler *challengehandlers.BasicChallengeHandler, port uint) (string, error) {
+	o, err := tokenrequest.NewRequestTokenOptions(clientCfg, false).WithChallengeHandlers(handler)
+	if err != nil {
+		return "", err
+	}
+
+	portStr := strconv.Itoa(int(port))
+	o.Issuer = net.JoinHostPort(o.Issuer, portStr)
+	return o.RequestToken()
 }
 
 // getGlobalKubeConfigPath returns the path to the first entry in the KUBECONFIG environment variable


praveenkumar:
Can you put more info in the commit logs why it was not working without this PR and what this PR adds to make it work?

vyasgun:
Thanks @praveenkumar. Just added more info to the commit log.

pkg/crc/machine/kubeconfig.go
```diff
@@ -198,13 +198,23 @@ func getTokenForUser(username, password, ip string, ca []byte, clusterConfig *ty
...
 	return token, nil
 }
 
+func requestTokenWithChallengeHandlers(clientCfg *restclient.Config, handler *challengehandlers.BasicChallengeHandler, port uint) (string, error) {
+	o, err := tokenrequest.NewRequestTokenOptions(clientCfg, false).WithChallengeHandlers(handler)
+	if err != nil {
+		return "", err
+	}
+
+	o.Issuer = fmt.Sprintf("%s:%d", o.Issuer, port)
```
praveenkumar: if we are adding the port explicit to issuer then I don't think we really need https://github.com/crc-org/crc/pull/4183/files#diff-a796f5c0a3b2fdd2f85be66107ff80578a3658029bd675c39ac6c015b515ebecR188-R190 part so let's try to remove that and see if that works.

pkg/crc/machine/kubeconfig.go
```diff
@@ -181,13 +180,10 @@ func getTokenForUser(username, password, ip string, ca []byte, clusterConfig *ty
...
 			DialContext: func(ctx gocontext.Context, network, address string) (net.Conn, error) {
 				logging.Debugf("Using address: %s", address)
-				hostname, port, err := net.SplitHostPort(address)
+				_, port, err := net.SplitHostPort(address)
 				if err != nil {
 					return nil, err
 				}
-				if strings.HasSuffix(hostname, constants.AppsDomain) {
-					port = strconv.Itoa(int(ingressHTTPSPort))
-				}
```
praveenkumar: @vyasgun After removing it looks like it fails again, did you test this scenario?
```
DEBU Using address: api.crc.testing:6443          
DEBU Dialing to 127.0.0.1:6443                    
DEBU Using address: oauth-openshift.apps-crc.testing:443 
DEBU Dialing to 127.0.0.1:443                     
ERRO Cannot update kubeconfig: dial tcp 127.0.0.1:443: connect: connection refused 
DEBU Making call to close driver server           
```

vyasgun: I didn't face the issue. I will take another look and update.

vyasgun: I added the lines back and updated the PR.

pkg/crc/machine/kubeconfig.go
```diff
@@ -10,7 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"strconv"
+	`strconv`
```
anjannath: unrelated change probably by the ide?

praveenkumar: @vyasgun can you fix it and repush it?

gbraad: @vyasgun can you handle this ASAP, otherwise it will not be possible to merge in time for the next release

gbraad: did linting fail? otherwise we can merge regardless and fix this ASAP if not done today.

cfergeau: This PR has " Maintainers are allowed to edit this pull request.", I've fixed this.

pkg/crc/machine/kubeconfig.go
cfergeau: Using https://pkg.go.dev/net#JoinHostPort may be slightly more correct. But there is this issue throughout the codebase, all of these can be changed in one go the day we look at proper ipv6 support.

vyasgun: thanks, will change this.

pkg/crc/machine/kubeconfig.go
vyasgun: thanks @cfergeau! Sorry I was AFK

gbraad (approved):

pkg/crc/machine/kubeconfig.go
praveenkumar: This was mistake because `o.Issuer` is URL and not the host string and because of that if take it as ip6 and fails
```
ERRO Cannot update kubeconfig: parse "[https://oauth-openshift.apps-crc.testing]:443": first path segment in URL cannot contain colon 
```

PR #4184, opened 2024-05-29
fix small bug in integration test

lilyLuLiu:
**Fixes:** There is a small bug in the integration test

## Solution/Idea

In the BeforeSuite function, judge whether the environmental variables ginkgoOpts and pullSecretPath exist. 

## Proposed changes

1. os.Getenv() not return err. We should check whether the environment string is empty.

### DIFF
--- a/test/integration/testsuite_test.go
+++ b/test/integration/testsuite_test.go
@@ -12,6 +12,9 @@ import (
 	. "github.com/onsi/gomega"
 	gomegaformat "github.com/onsi/gomega/format"
 	"github.com/sirupsen/logrus"
+
+	"flag"
+	"github.com/spf13/pflag"
 )
 
 type VersionAnswer struct {
@@ -25,8 +28,21 @@ var userHome string
 var versionInfo VersionAnswer
 
 var bundlePath string
-var ginkgoOpts string
 var pullSecretPath string
+
+
+func TestMain(m *testing.M) {
+	RegisterFlags(flag.CommandLine)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	pflag.Parse()
+
+	os.Exit(m.Run())
+}
+
+func RegisterFlags(flags *flag.FlagSet) {
+	flags.StringVar(&bundlePath, "bundle-path", "", "Path to the bundle to be used in tests.")
+	flags.StringVar(&pullSecretPath, "pull-secret-path", "", "Path to the file containing pull secret.")
+}
 
 func TestTest(t *testing.T) {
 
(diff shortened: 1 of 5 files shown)


test/integration/testsuite_test.go
```diff
@@ -82,19 +82,14 @@ var _ = BeforeSuite(func() {
...
 
 	ginkgoOpts = os.Getenv("GINKGO_OPTS")
-	if err != nil {
-
+	if ginkgoOpts == "" {
 		logrus.Infof("Error: Could not read GINKGO_OPTS.")
-		logrus.Infof("%v", err)
 	}
-	Expect(err).NotTo(HaveOccurred())
+	Expect(ginkgoOpts).NotTo(BeEmpty())
```
gbraad: Why can't they be empty?

test/integration/testsuite_test.go
```diff
@@ -82,19 +82,14 @@ var _ = BeforeSuite(func() {
...
 
 	pullSecretPath = os.Getenv("PULL_SECRET_PATH") // this env var should contain location of pull-secret file
-	if err != nil {
-
+	if pullSecretPath == "" {
 		logrus.Infof("Error: You need to set PULL_SECRET_PATH to find CRC useful.")
```
gbraad: Off-topic, but could be addressed:

> `"Error: You need to set PULL_SECRET_PATH to find CRC useful."`

`Error: You need to set PULL_SECRET_PATH for CRC to function properly.`

Note: found the original weirdly phrased.

test/integration/testsuite_test.go
```diff
@@ -82,19 +82,14 @@ var _ = BeforeSuite(func() {
...
-		logrus.Infof("%v", err)
 	}
-	Expect(err).NotTo(HaveOccurred())
-
+	Expect(pullSecretPath).NotTo(BeEmpty())
```
gbraad, line 102: 👍 this can't be empty

test/integration/testsuite_test.go
lilyLuLiu: yes, `ginkgoOpts` could be empty. I just checked and found this variable is not used after this assignment.
The environmental variable `GINKGO_OPTS` is only used in makefile L244. 
So I think `ginkgoOpts` part code can be deleted.

test/integration/testsuite_test.go
lilyLuLiu: agree with you. I will change it as you suggested.

adrianriobo:
Hey we need to fix https://github.com/crc-org/crc/issues/4185 before do anything else on e2e or integration. On the other this is a good check but we can sync on this as may we should pass parameters the "right way".

[This](https://github.com/containers/podman-desktop-e2e/blob/main/test/e2e/e2e_podman/suite_test.go#L20) is an example on how to use params. But again we can discuss about it an decide how to tackle it.

adrianriobo (commented):
Can we make the commit msg a bit por explicit on what is being checked with the change?

test/integration/testsuite_test.go
```diff
@@ -25,9 +28,22 @@ var userHome string
...
+func TestMain(m *testing.M) {
+	RegisterFlags(flag.CommandLine)
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	pflag.Parse()
+
+	os.Exit(m.Run())
+}
+
+func RegisterFlags(flags *flag.FlagSet) {
+	flags.StringVar(&bundlePath, "bundle_path", "", "Path to the bundle to be used in tests.")
```
adrianriobo: typically the format would be mid-separator. i.e. `bundle-path` instead of bundle_path

adrianriobo: Same for `pull-secret-path`

lilyLuLiu: OK, I will fix it.

test/integration/testsuite_test.go
```diff
@@ -76,25 +92,12 @@ var _ = BeforeSuite(func() {
...
-		logrus.Infof("%v", err)
-	}
-	Expect(err).NotTo(HaveOccurred())
-
-	pullSecretPath = os.Getenv("PULL_SECRET_PATH") // this env var should contain location of pull-secret file
-	if err != nil {
-
-		logrus.Infof("Error: You need to set PULL_SECRET_PATH to find CRC useful.")
-		logrus.Infof("%v", err)
+	if pullSecretPath == "" {
```
adrianriobo: Sorry did not see this before, for golang this is commonly done:

```golang
if len(pullSecretPath) == 0 {....
```

As so in case the var at some point is set to nil it will pass on this condition, but for the equal empty string will throw panic error

lilyLuLiu: I see. Let me fix it.

adrianriobo (approved):

adrianriobo (changes requested):
Ahh wait after the revert https://github.com/crc-org/crc/commit/d9e501b5af5fe459dca9b06142067b7facb22116 this should be changed on `images/build-integration/lib/darwin/run.sh` too

adrianriobo (approved):

PR #4188, opened 2024-05-30
[QE] Misc fixes for the QE e2e images

anjannath:
- In the run.sh script the `${OS}` variable referred to on the scripts is only present when
running inside the container, we'll have to figure out a different
way to get the OS information when running on the host

- update base images to `go-toolset:1.21`

### DIFF
--- a/images/build-e2e/Containerfile
+++ b/images/build-e2e/Containerfile
@@ -1,5 +1,5 @@
 
-FROM registry.access.redhat.com/ubi8/go-toolset:1.20 AS builder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.21 AS builder
 
 USER root
 
--- a/update-go-version.sh
+++ b/update-go-version.sh
@@ -14,6 +14,7 @@ go mod edit -go ${golang_base_version} tools/go.mod
 sed -i "s,^GOVERSION = 1.[0-9]\+,GOVERSION = ${golang_base_version}," Makefile
 sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-\)1.[0-9]\+,\1${golang_base_version}," images/*/Dockerfile
 sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Dockerfile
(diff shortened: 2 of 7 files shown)


images/build-e2e/Containerfile
```diff
@@ -1,5 +1,5 @@
 
-FROM registry.access.redhat.com/ubi8/go-toolset:1.20 AS builder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.21 AS builder
```
praveenkumar, line 2: can you also update https://github.com/crc-org/crc/blob/main/update-go-version.sh and may be add another like for `Containerfile` ?

adrianriobo: Hey can you add this? or do you prefer create another PR, I want to merge this one as soon as possible

anjannath: ahh, sorry i forgot about this, updating now!

anjannath: updated now!

update-go-version.sh
```diff
@@ -14,6 +14,7 @@ go mod edit -go ${golang_base_version} tools/go.mod
 sed -i "s,^GOVERSION = 1.[0-9]\+,GOVERSION = ${golang_base_version}," Makefile
 sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-\)1.[0-9]\+,\1${golang_base_version}," images/*/Dockerfile
 sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Dockerfile
+sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Containerfile
```
adrianriobo, line 17: I guess we can use a common expression with something like `[Container|Docker]file` but I think this script will be update sooner than later with 9 versions (rhel an ubi) so I am good with this

adrianriobo (approved):

update-go-version.sh
anjannath: i am not sure if we can use the pattern for the filename, you mean something like:
```
sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/[Docker|Container]file
```

adrianriobo: yeah, but as I said I am ok with the full line

PR #4206, opened 2024-06-05
pkg: remove use of sudo from the signing script

anjannath:
this is required as use of 'sudo' is not available in downstream build machines. while testing the built pkg with this changes crc is working as expected

### DIFF
--- a/packaging/darwin/macos-pkg-build-and-sign.sh
+++ b/packaging/darwin/macos-pkg-build-and-sign.sh
@@ -49,7 +49,7 @@ sign "${crcBinDir}/crc"
 sign "${crcBinDir}/crc-admin-helper-darwin"
 sign "${crcBinDir}/vfkit"
 
-sudo chmod +sx "${crcBinDir}/crc-admin-helper-darwin"
+chmod +sx "${crcBinDir}/crc-admin-helper-darwin"
 
 pkgbuild --identifier com.redhat.crc --version ${version} \
   --scripts "${BASEDIR}/scripts" \


anjannath:
/hold 
till it is tested by the downstream pipeline, cc @adrianriobo

packaging/darwin/macos-pkg-build-and-sign.sh
```diff
@@ -49,7 +49,7 @@ sign "${crcBinDir}/crc"
 sign "${crcBinDir}/crc-admin-helper-darwin"
 sign "${crcBinDir}/vfkit"
 
-sudo chmod +sx "${crcBinDir}/crc-admin-helper-darwin"
+chmod +sx "${crcBinDir}/crc-admin-helper-darwin"
```
cfergeau, line 52: If `"${crcBinDir}/crc-admin-helper-darwin"` is owned by the user running the script, both versions should be equivalent.
After installation though, `crc-admin-helper-darwin` must be owned by `root` and have the `+s` permission set, otherwise the binary won't be able to make changes to `/etc/hosts` Have you checked it's still fine?

adrianriobo: After create installer from this PR and run it I have this

```
podmanqe@dev-platform-mac4 ~ % ls -al /usr/local/crc     
total 389096
drwxr-xr-x  5 root  wheel        160 Jun  5 13:11 .
drwxr-xr-x  5 root  wheel        160 Jun  5 13:12 ..
-rwxr-xr-x  1 root  wheel  159483008 Jun  5 13:11 crc
-rwsr-sr-x  1 root  wheel    5238656 Jun  5 13:11 crc-admin-helper-darwin
-rwxr-xr-x  1 root  wheel   34489808 Jun  5 13:11 vfkit
```

anjannath: the installation needs to be run as root, since it installs in the `/` dir, and the files it creates are also owned by `root`

adrianriobo (approved):

Issue #4203, opened 2024-06-05
[QE] e2e Install new operator story failing on arm64 bundle


adrianriobo:
When running the e2e `install operator story` it is failing on arm64 machines.

Checking the story the Openshift Pipelines operator is deployed and then a new task and taskrun are created, for the task spec it is using an `registry.redhat.io/ubi7/ubi-minimal` which does not have arm64 version, as so it fails to run on an OCP arm64 cluster.

adrianriobo:
Fixed by #4213

PR #4213, opened 2024-06-07
[qe] micro fix e2e test tekton task image

lilyLuLiu:
**Fixes:** Issue [#4203](https://github.com/crc-org/crc/issues/4203)

### DIFF
--- a/images/build-integration/lib/darwin/run.sh
+++ b/images/build-integration/lib/darwin/run.sh
@@ -32,7 +32,8 @@ while [[ $# -gt 0 ]]; do
         -labelFilter)
         labelFilter="$2"
         shift 
-        shift 
+        shift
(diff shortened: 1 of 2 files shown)


adrianriobo (commented):
Can you re word the commit to make it a bit more clear what is this change about.  I mean it is not mandatory but lessons learned from the team is great to have detailed commit as that makes you half of the work while reviewing.

[qe] Upgrade task spec image to ubi9 on install operator story  

Previously it was using ubi7 which has no arm64 image,  this made e2e tests fail when we try the story on arm64 platforms. This commit will set the task to use ubi9 which supports arm64.

adrianriobo (approved):

adrianriobo (approved):

PR #4222, opened 2024-06-11
nameservers.go: use `temporary` option for modify the network connection

praveenkumar:
Previously configure-ovs script which runs as part of ovs-configuration service used to consume `/etc/NetworkManager/` directory to create the connections and then moving it to `/run/NetworkManager` but recently [0] this behaviour is changed and they use the `temporary` and `offline` option of `nmcli` directly create the network connection in `/run/NetworkManager/system-networks` location. In our code base till now we are using the `nmcli mod` to modify the `ovs-if-br-ex` connection to add the custom dns nameserver and search option which by default land to `/etc/NetworkManager/system-networks` locations as persistent. It caused the issue when `ovs-configuration` service restart happen where it find other network in `/run/NetworkManager` and assume the service already runs once but failed to get the `ovs-if-br-ex` network and failed with
```
Jun 11 04:09:14 crc configure-ovs.sh[8315]: + false
Jun 11 04:09:14 crc configure-ovs.sh[8315]: + mod_nm_conn ovs-if-br-ex connection.autoconnect yes
Jun 11 04:09:14 crc configure-ovs.sh[8315]: + local dst_path=/run/NetworkManager/system-connections/ovs-if-br-ex.nmconnection
Jun 11 04:09:14 crc configure-ovs.sh[8315]: + local src_path
Jun 11 04:09:14 crc configure-ovs.sh[8424]: ++ mktemp
Jun 11 04:09:14 crc configure-ovs.sh[8315]: + src_path=/tmp/tmp.eRO482hq7I
Jun 11 04:09:14 crc configure-ovs.sh[8315]: + shift
Jun 11 04:09:14 crc configure-ovs.sh[8315]: + cat /run/NetworkManager/system-connections/ovs-if-br-ex.nmconnection
Jun 11 04:09:15 crc configure-ovs.sh[8425]: cat: /run/NetworkManager/system-connections/ovs-if-br-ex.nmconnection: No such file or directory
Jun 11 04:09:15 crc configure-ovs.sh[8315]: ++ handle_exit
```

In this PR, `--temporary` option added during network connection modification and it will not put the connection info to `/etc/NetworkManager/..` location as peristent but use same `/run/NetworkManager/..` location.

[0] https://github.com/openshift/machine-config-operator/commit/2003da5d02f27f8dd0e5a91fb2c714a247c4e824

### DIFF
--- a/pkg/crc/network/nameservers.go
+++ b/pkg/crc/network/nameservers.go
@@ -67,7 +67,7 @@ func updateNetworkManagerConfig(sd *systemd.Commander, sshRunner *ssh.Runner, re
 	nameservers := strings.Join(resolvFileValues.GetNameServer(), ",")
 	searchDomains := strings.Join(resolvFileValues.GetSearchDomains(), ",")
 	// When ovs-configuration service is running, name of the connection should be ovs-if-br-ex
-	_, stderr, err := sshRunner.RunPrivileged("Update resolv.conf file", "nmcli", "con", "modify", "ovs-if-br-ex",
+	_, stderr, err := sshRunner.RunPrivileged("Update resolv.conf file", "nmcli", "con", "modify", "--temporary", "ovs-if-br-ex",
 		"ipv4.dns", nameservers, "ipv4.dns-search", searchDomains)
 	if err != nil {
 		return fmt.Errorf("failed to update resolv.conf file %s: %v", stderr, err)


cfergeau:
Is this 4.16+ specific? Will this also be an issue in 4.15?

praveenkumar:
> Is this 4.16+ specific? Will this also be an issue in 4.15?

once https://github.com/openshift/machine-config-operator/commit/2003da5d02f27f8dd0e5a91fb2c714a247c4e824 is backported to 4.15 then it will also be an issue for 4.15 .

cfergeau:
> > Is this 4.16+ specific? Will this also be an issue in 4.15?
> 
> once [openshift/machine-config-operator@2003da5](https://github.com/openshift/machine-config-operator/commit/2003da5d02f27f8dd0e5a91fb2c714a247c4e824) is backported to 4.15 then it will also be an issue for 4.15 .

But do you know if it's planned/if this will be done?
Since it's 4.16 specific at the moment, I don't think this should go in `main` since we still have
```
OPENSHIFT_VERSION ?= 4.15.10
```
This commit could go in the PR changing `OPENSHIFT_VERSION` to `4.16.x`

praveenkumar:
> .But do you know if it's planned/if this will be done?

I am not sure because there is no cherry-pick as of now on that PR.

> Since it's 4.16 specific at the moment, I don't think this should go in main since we still have

This PR works with 4.15 also so it is not breaking one.

cfergeau:
> > .But do you know if it's planned/if this will be done?
> 
> I am not sure because there is no cherry-pick as of now on that PR.
> 
> > Since it's 4.16 specific at the moment, I don't think this should go in main since we still have
> 
> This PR works with 4.15 also so it is not breaking one.

Can you add it to the commit log? That it's going to be required for 4.16 bundles, but that it also works fine with 4.15 bundles?

praveenkumar:
> > > .But do you know if it's planned/if this will be done?
> > 
> > 
> > I am not sure because there is no cherry-pick as of now on that PR.
> > > Since it's 4.16 specific at the moment, I don't think this should go in main since we still have
> > 
> > 
> > This PR works with 4.15 also so it is not breaking one.
> 
> Can you add it to the commit log? That it's going to be required for 4.16 bundles, but that it also works fine with 4.15 bundles?

Done

cfergeau (approved):

PR #4223, opened 2024-06-11
[qe] qenvs rebranding to mapt 

adrianriobo:
This version fix issues with resources not being deleted properly due to spot machines stopped prematurely


build-qe-images.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Copy me to the crc code folder
```
praveenkumar: May be put more details like how to run this script, what parameters it uses ..etc.?

adrianriobo: ahah sorry that was a mistake, I have it locally to create the images, but if it is valuable I can add it on a different PR

PR #4224, opened 2024-06-11
Fix invalid release-info.json.sample

cfergeau:
An extra ',' causes it to fail validation

$ curl -L https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/crc/latest/release-info.json | jq .
parse error: Expected another key-value pair at line 6, column 2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ vendorcheck:
 	./verify-vendor.sh
 
 .PHONY: check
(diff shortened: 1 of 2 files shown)


benoitf (approved):

cfergeau:
The second commit (`make check-release-info`) causes CI failure. Need to add `jq` to our misc jobs, or add it to `tools/` to automatically install it when needed.

gbraad (approved):

gbraad:
> The second commit (make check-release-info) causes CI failure.

Please create a follow-up issue to resolve this ASAP

PR #4226, opened 2024-06-12
fix 'sed -i' failure in Makefile for macOS

lilyLuLiu:
`make release` fail with error message:
> cat: release-info.json: No such file or directory

The file `release-info.json` was moved to folder release. Thus `check-release-info` can't find the file

`sed -i` failed in Makefile when running in MacOS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,8 +289,8 @@ cross-lint: $(TOOLS_BINDIR)/golangci-lint
 .PHONY: gen_release_info
 gen_release_info:
 	@cat release-info.json.sample | sed s/@CRC_VERSION@/$(CRC_VERSION)/ > $(RELEASE_INFO)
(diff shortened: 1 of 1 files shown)


cfergeau:
Can we keep this 2 separate commits? (the `sed` changes and the `cp`->`mv` change)
Both will be needed, but they fix 2 very different things, so better to keep them separate.

I've filed https://github.com/crc-org/crc/pull/4227 which fixes more issues with the release-info check, but which will need the fixes from this PR.

praveenkumar:
@lilyLuLiu can you drop https://github.com/crc-org/crc/pull/4226/commits/c97baee07a5af1b642c2bf1bd62aee5340b54e07 one and have only one commit because other part is fix in #4227

lilyLuLiu:
@praveenkumar , Hi, I separated the commit as @cfergeau suggested.

praveenkumar:
/close

changes merged with #4227

Issue #10, opened 2019-03-01
Setup single-VM during/post installer


gbraad:

gbraad:
@praveenkumar We can update this with information from: https://github.com/praveenkumar/snc

praveenkumar:
This is already fixed, now we are using snc repo to create the single node VM and producing the disk images.

PR #4227, opened 2024-06-12
release-info: Test its format as part of 'make test'

cfergeau:
This commit makes use of `//go:embed` to run the content of
`release-info.json` through golang `json.Valid` to check if it's valid 
JSON.

This is better/simpler than using `jq` as we don't have to change all our
container images/CI envs/... to make sure `jq` is installed, and
`json.Valid` uses the same json parser as crc so this avoids possible 
differences of behaviour between json parsers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,8 +289,8 @@ cross-lint: $(TOOLS_BINDIR)/golangci-lint gen_release_info
 .PHONY: gen_release_info
 gen_release_info:
 	@cat release-info.json.sample | sed s/@CRC_VERSION@/$(CRC_VERSION)/ > $(RELEASE_INFO)
-	@sed -i s/@GIT_COMMIT_SHA@/$(COMMIT_SHA)/ $(RELEASE_INFO)
-	@sed -i s/@OPENSHIFT_VERSION@/$(OPENSHIFT_VERSION)/ $(RELEASE_INFO)
+	@sed -i'' -e s/@GIT_COMMIT_SHA@/$(COMMIT_SHA)/ $(RELEASE_INFO)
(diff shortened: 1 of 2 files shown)


praveenkumar:
With this approach `golint` is failing so we might need to add the `gen_release_info` for  `lint` and `cross-lint` target.

cfergeau:
> With this approach `golint` is failing so we might need to add the `gen_release_info` for `lint` and `cross-lint` target.

Did not have problems with `rm release-info.json && make lint`, but had issues with `cross-lint`. Updated both.

gbraad (approved):
is there not a single-line command option like Python can do?

anjannath:
The macOS failures are because by default its not using `gnu-sed`, we can install it with brew and set in the PATH so that same sed command also works on macOS

```
sed: -I or -i may not be used with stdin
```

edit: `sed` issue is fixed in #4226

praveenkumar:
@cfergeau may be you can cherry pick https://github.com/crc-org/crc/pull/4226 one and repush to make sure everything is green again?

praveenkumar:
RPM build job still failing with  `Error: #10 411.9 release_info_test.go:9:12: pattern release-info.json: no matching files found`

cfergeau:
> RPM build job still failing with `Error: #10 411.9 release_info_test.go:9:12: pattern release-info.json: no matching files found`

Yes, this PR fixes part of the problems, https://github.com/crc-org/crc/pull/4226 will fix the other ones (`sed -i` on macos, and `make release` moving away `release-info.json` after it's been generated)

cfergeau:
I cherry-picked both changes from Lu to try to get a successful CI run.

PR #4239, opened 2024-06-19
update crc-driver-libvirt version to 0.13.8

anjannath:


cfergeau (commented):
you also need to update
```
images/rpmbuild/Containerfile.in:       https://github.com/crc-org/machine-driver-libvirt/releases/download/0.13.7/crc-driver-libvirt-0.13.7-1.el8.x86_64.rpm
```

Issue #4101, opened 2024-04-03
[QE][BUG] Fixed  code for insecure login is not being used


adrianriobo:
With https://github.com/crc-org/crc/commit/496059091a1c50bcf6172c6302aaabec3f376f22 we introduce a fallback mechanism for oc login when it reports unknown CA for server, but somehow the check for that condition is not being matched, as so when we hit the issue the scenario fails.

```bash
<testcase name="Overall cluster health" status="failed" time="1010.846097">
<failure message="Step executing "oc new-project testproj" succeeds: command 'oc new-project testproj', expected to succeed, exited with exit code: 1 Command stdout: Command stderr: Unable to connect to the server: tls: failed to verify certificate: x509: certificate signed by unknown authority "/>
```

Need to fix it

PR #4248, opened 2024-06-24
[QE] fix can't catch error for "oc login"

lilyLuLiu:
**Fixes:** Issue #4101

### DIFF
--- a/test/extended/util/util.go
+++ b/test/extended/util/util.go
@@ -269,7 +269,7 @@ func LoginToOcCluster(options []string) error {
 		ocLoginAsAdminCommand = ocLoginAsAdminCommand + " " + option
 	}
 
-	return ExecuteCommand(ocLoginAsAdminCommand)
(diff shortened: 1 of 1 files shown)


test/extended/util/util.go
```diff
@@ -269,7 +269,14 @@ func LoginToOcCluster(options []string) error {
 		ocLoginAsAdminCommand = ocLoginAsAdminCommand + " " + option
 	}
 
-	return ExecuteCommand(ocLoginAsAdminCommand)
+	err = ExecuteCommand(ocLoginAsAdminCommand)
```
cfergeau: I think the code you add is similar to what `return ExecuteCommandSucceedsOrFails(ocLoginAsAdminCommand, "succeeds")`  does?

lilyLuLiu: Thanks for pointing this out. I have re-commit the code change.

cfergeau (approved):

PR #4252, opened 2024-07-01
[qe]e2e: wait longer for monitoring operator

lilyLuLiu:


cfergeau:
I assume we need to wait longer to avoid some failures when running our tests on some systems in some environments?

adrianriobo:
Yeah from time to time there are some delays when pulling those images, may affected by networking or something else.

adrianriobo (commented):
actually @lilyLuLiu it would be nive have this as part of the commit log. The reasoning to increase the timeout here

adrianriobo (approved):

PR #4253, opened 2024-07-01
build: Update k8s and openshift go modules to sync with openshift-4.16

praveenkumar:
This updates the `github.com/openshift/*` modules to `release-4.16` and `k8s.io/*` to `v0.29.3`.


cfergeau (commented):
git main is still using 4.15 https://github.com/crc-org/crc/blob/926c78e6bfc64ce6a5ed1771d58d637ce34aafe5/Makefile#L5-L7 so it's a bit early for this change.
This could go in the same PR as the one switching the Makefile to 4.16.

praveenkumar:
> o it's a bit early for this change.
> This could go in the same PR as the one switching the Makefile to 4.16.

Yes, we can put it on hold till then, I don't think it break current master because from bundle side it is not going to break (no specific code at bundle side).

praveenkumar:
/unhold

Updated the bundle to 4.16 also.

Issue #4209, opened 2024-06-05
[QE] Push qe images to quay on tag creation


adrianriobo:
Currently qe images are built on pr-check to be used afterwards for testing remote targets.

When a new release happens we are manually creating the images from the tag and pushing to quay it.

We can automate this task by pushing the qe images build on a gh action on tirggered by tag creation

gbraad:
@lilyLuLiu If you are working on this, please change the status to present: "Work in progress"

PR #4257, opened 2024-07-04
new action that push qe images for new tag

lilyLuLiu:
**Relates to:** Issue #4209

### DIFF
--- a/.github/workflows/qe-image.yml
+++ b/.github/workflows/qe-image.yml
@@ -1,0 +1,43 @@
+name: Push qe images for new tag
+on:
+  push:
+    tags:
+      - '*'
+env:
+  IMAGE_REGISTRY: quay.io
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        go:
+          - '1.21'
+        os: ['linux', 'windows', 'darwin']
+        arch: ['amd64', 'arm64']
+        exclude:
+          - arch: 'arm64'
+            os: 'windows'
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+      - name: Log in to Quay.io
+        uses: redhat-actions/podman-login@v1
+        with:
+          username: ${{ secrets.QUAY_IO_USERNAME }}
+          password: ${{ secrets.QUAY_IO_PASSWORD }}
+          registry: ${{ env.IMAGE_REGISTRY }}
+      - name: create script of building and push images
+        run: |
+          cat <<EOF > build-qe-images.sh
+          !#/bin/bash
+          set -x
+          OS=${{matrix.os}} ARCH=${{matrix.arch}} CRC_INTEGRATION_IMG_VERSION=${{ github.ref_name }} make containerized_integration
+          podman push quay.io/crcont/crc-integration:${{ github.ref_name }}-${{matrix.os}}-${{matrix.arch}}
+          OS=${{matrix.os}} ARCH=${{matrix.arch}} CRC_E2E_IMG_VERSION=${{ github.ref_name }} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:${{ github.ref_name }}-${{matrix.os}}-${{matrix.arch}}
+          EOF
+      - name: run script to build and push images
+        run: |
+          chmod +x build-qe-images.sh
+          ./build-qe-images.sh


.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,55 @@
+name: Push qe images for new tag
+on:
+  push:
+    tags:
+      - '*'
+env:
+  REGISTRY_USER: "crcont+qeghcrc"
```
adrianriobo: Add as secret as well (We can re-create the robot account)

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,55 @@
...
+          - ubuntu-latest
+        go: 
+          - '1.20'    
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+      - name: Log in to Quay.io
+        uses: redhat-actions/podman-login@v1
+        with:
+          username: ${{ env.REGISTRY_USER }}
```
adrianriobo: You can use secrets directly here: 

```
 - name: Log in to quay.io Registry
      uses: redhat-actions/podman-login@v1
      with:
        registry: quay.io
        username: ${{ secrets.QUAY_IO_USERNAME }}
        password: ${{ secrets.QUAY_IO_PASSWORD }}
```

adrianriobo (commented):
May add WIP until you test it

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: 
+          - ubuntu-latest
+        go: 
+          - '1.20'
```
adrianriobo: Need to set go `1.21`

.github/workflows/qe-image.yml
lilyLuLiu: updated

.github/workflows/qe-image.yml
lilyLuLiu: updated

.github/workflows/qe-image.yml
lilyLuLiu: updated

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,57 @@
...
+          OS=darwin ARCH=arm64 CRC_INTEGRATION_IMG_VERSION=\${1} make containerized_integration
+          podman push quay.io/crcont/crc-integration:\${1}-darwin-arm64
+          OS=linux ARCH=amd64 CRC_E2E_IMG_VERSION=\${1} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:\${1}-linux-amd64
+          OS=windows ARCH=amd64 CRC_E2E_IMG_VERSION=\${1} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:\${1}-windows-amd64
+          OS=darwin ARCH=amd64 CRC_E2E_IMG_VERSION=\${1} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:\${1}-darwin-amd64
+          OS=darwin ARCH=arm64 CRC_E2E_IMG_VERSION=\${1} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:\${1}-darwin-arm64
```
praveenkumar: Instead of pushing each image one by one, can you have a manifest and add those images to manifest and then push that manifest against version (something like we do for bundle images https://quay.io/repository/crcont/podman-bundle?tab=tags)?

lilyLuLiu: @praveenkumar , Thanks for this suggestion. Manifest is a great way to manage the images. 
But I failed to pull a manifest image without specifying the os and arch:
```
% podman pull [quay.io/crcont/podman-bundle:4.4.4](http://quay.io/crcont/podman-bundle:4.4.4)
Trying to pull [quay.io/crcont/podman-bundle:4.4.4](http://quay.io/crcont/podman-bundle:4.4.4)...
Error: choosing an image from manifest list docker://quay.io/crcont/podman-bundle:4.4.4: no image found in image index for architecture arm64, variant "v8", OS linux
```
I know using `--platform` can get a specific one. But there is no option in tekton definition to specify the os and arch.

adrianriobo: The strategy here is not possible, because the actual platform for all the images is linux / amd64 the suffix on the versioning is due to the content the have inside (test binary for X platform)

These images are executed locally (i.e. gh runner, or pod on ocp) and they connect against a remote target, copy resources to it, execute them and finally copy back the results)

adrianriobo: Actually I am thinking on https://github.com/adrianriobo/deliverest/issues/44 because most of the size of the image is due to the base image and no the binary they hold inside so we are not getting much benefit from current split per os and arch.

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,52 @@
...
+env:
+  IMAGE_REGISTRY: quay.io
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
```
adrianriobo: Hey can we use the latest version, but set the version explicitly:

In this case the latest available version for ubuntu is ubuntu-24.04, which may you expect to have setting that as ubuntu-latest but from https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories this is not the case.

So to be sure about what we are using, and what we can expect, best set the version

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,52 @@
+name: Push qe images for new tag
+on:
+  push:
+    tags:
+      - '*'
+env:
+  IMAGE_REGISTRY: quay.io
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
```
adrianriobo: As it is only 1 value on the os value of the matrix, you should move the value here (ubuntu-24.04), check comment about versioning

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Log in to Quay.io
+        uses: redhat-actions/podman-login@v1
+        with:
+          username: ${{ secrets.QUAY_IO_USERNAME }}
+          password: ${{ secrets.QUAY_IO_PASSWORD }}
+          registry: ${{ env.IMAGE_REGISTRY }}
+      - name: create script of building and push images
+        run: |
+          cat <<EOF > build-qe-images.sh
+          OS=linux ARCH=amd64 CRC_INTEGRATION_IMG_VERSION=v${1} make containerized_integration
```
adrianriobo: We are now including the linux/support so can we add it here as well?

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: actions/checkout@v4
+      - name: Log in to Quay.io
+        uses: redhat-actions/podman-login@v1
+        with:
+          username: ${{ secrets.QUAY_IO_USERNAME }}
+          password: ${{ secrets.QUAY_IO_PASSWORD }}
+          registry: ${{ env.IMAGE_REGISTRY }}
+      - name: create script of building and push images
+        run: |
+          cat <<EOF > build-qe-images.sh
```
adrianriobo, line 32: Also can we reduce the number of commands to 4 lines, 2 for e2e (build and push) 2 for integration (build and push) and then use arch and os as vars coming from a matrix (https://github.com/crc-org/crc/blob/main/.github/workflows/windows-artifacts.yml#L47) that will help to reduce the flow

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,52 @@
...
+          OS=windows ARCH=amd64 CRC_E2E_IMG_VERSION=v${1} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:v${1}-windows-amd64
+          OS=darwin ARCH=amd64 CRC_E2E_IMG_VERSION=v${1} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:v${1}-darwin-amd64
+          OS=darwin ARCH=arm64 CRC_E2E_IMG_VERSION=v${1} make containerized_e2e
+          podman push quay.io/crcont/crc-e2e:v${1}-darwin-arm64
+          EOF
+      - name: run script to build and push images
+        env:
+          TAG_CONTEXT: ${{ toJson(github.ref_name) }}
```
adrianriobo: Is `toJson()` func required here?

.github/workflows/qe-image.yml
lilyLuLiu: updated

.github/workflows/qe-image.yml
lilyLuLiu: Updated

.github/workflows/qe-image.yml
lilyLuLiu: updated

.github/workflows/qe-image.yml
lilyLuLiu: yes, `toJson` is not needed. I have removed it.

.github/workflows/qe-image.yml
lilyLuLiu: @adrianriobo Sorry, I don't get it. Could you explain more about it?

adrianriobo: ahaha indeed you don get it I wrote it wrong /0\ , sorry

I meant we are adding support for linux/arm64 previously we were not building those images but now we will need them.. So can you add that tuple to the matrix values: (linux,arm64)

.github/workflows/qe-image.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Check out repository code
+        uses: actions/checkout@v4
+      - name: Log in to Quay.io
+        uses: redhat-actions/podman-login@v1
+        with:
+          username: ${{ secrets.QUAY_IO_USERNAME }}
+          password: ${{ secrets.QUAY_IO_PASSWORD }}
+          registry: ${{ env.IMAGE_REGISTRY }}
+      - name: create script of building and push images
+        run: |
```
adrianriobo, line 31: I would simplify this with only one step and a script like

```bash
!#/bin/bash

 set -x
OS=${{matrix.os}} ARCH=${{matrix.arch}} CRC_INTEGRATION_IMG_VERSION=${{ github.ref_name }} make containerized_integration
podman push quay.io/crcont/crc-integration:${{ github.ref_name }}-${{matrix.os}}-${{matrix.arch}}
OS=${{matrix.os}} ARCH=${{matrix.arch}} CRC_E2E_IMG_VERSION=${{ github.ref_name }} make containerized_e2e
podman push quay.io/crcont/crc-e2e:${{ github.ref_name }}-${{matrix.os}}-${{matrix.arch}}
```

.github/workflows/qe-image.yml
adrianriobo: just saw your update on the matrix so basically is now remove the excluded on that tuple (linux,arm64)

lilyLuLiu: updated

.github/workflows/qe-image.yml
lilyLuLiu: updated

adrianriobo (approved):

PR #4267, opened 2024-07-11
cache: Add condition to check binary name with arch suffix

praveenkumar:
As part of 926c78, libvirt driver download url change to have arch info and keeping the binary name as it is but as part of `cmd/embeded.go` it get the list using basename of uri and match with binary name which causing the following warning when crc binary have embedded driver and admin helper bit.

```
WARN Executable name is crc-driver-libvirt but extracted file name is crc-driver-libvirt-amd64
```

With this PR we are adding the condition where filelist and expected binary name match even it have arch info as suffix.

## Testing
- Run `make release` without this PR and remove libvirt and admin helper binary from `$HOME/.crc/bin` folder
-  use the binary which built using `make release` command and run `crc setup` and you will get this warning message

Try same with this PR and you shouldn't see this warning message.

### DIFF
--- a/pkg/crc/cache/cache.go
+++ b/pkg/crc/cache/cache.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
@@ -44,6 +45,10 @@ func (c *Cache) GetExecutablePath() string {
 
 func (c *Cache) GetExecutableName() string {
 	return filepath.Base(c.executablePath)
+}
+
+func (c *Cache) GetExecutableNameWithArch() string {
+	return fmt.Sprintf("%s-%s", c.GetExecutableName(), runtime.GOARCH)
 }
 
 /* getVersionGeneric runs the cached executable with 'args', and assumes the version string
@@ -128,7 +133,7 @@ func (c *Cache) cacheExecutable() error {
 		}
 	} else {
 		extractedFiles = append(extractedFiles, assetTmpFile)
-		if filepath.Base(assetTmpFile) != c.GetExecutableName() && !c.ignoreNameMismatch {
+		if (filepath.Base(assetTmpFile) != c.GetExecutableName() && filepath.Base(assetTmpFile) != c.GetExecutableNameWithArch()) && !c.ignoreNameMismatch {
 			logging.Warnf("Executable name is %s but extracted file name is %s", c.GetExecutableName(), filepath.Base(assetTmpFile))
 		}
 	}


cfergeau:
I'd favour something like this if this works:
```
diff --git a/pkg/crc/cache/cache_linux.go b/pkg/crc/cache/cache_linux.go
index cfb623b3e..226816521 100644
--- a/pkg/crc/cache/cache_linux.go
+++ b/pkg/crc/cache/cache_linux.go
@@ -5,7 +5,10 @@ import (
 )
 
 func NewMachineDriverLibvirtCache() *Cache {
-       return newCache(libvirt.MachineDriverPath(), libvirt.MachineDriverDownloadURL, libvirt.MachineDriverVersion, getCurrentLibvirtDriverVersion)
+       cache := newCache(libvirt.MachineDriverPath(), libvirt.MachineDriverDownloadURL, libvirt.MachineDriverVersion, getCurrentLibvirtDriverVersion)
+       // upstream binaries are suffixed with the arch (crc-driver-libvirt-amd64)
+       // crc caches the binary as crc-driver-libvirt as it's the name expected by github.com/crc-org/machine
+       cache.ignoreNameMismatch = true
 }
 
 func getCurrentLibvirtDriverVersion(executablePath string) (string, error) {
```

Or to update the machine code to run `crc-driver-libvirt-$arch` instead of `crc-driver-libvirt`

praveenkumar:
> I'd favour something like this if this works:
> 
> ```
> diff --git a/pkg/crc/cache/cache_linux.go b/pkg/crc/cache/cache_linux.go
> index cfb623b3e..226816521 100644
> --- a/pkg/crc/cache/cache_linux.go
> +++ b/pkg/crc/cache/cache_linux.go
> @@ -5,7 +5,10 @@ import (
>  )
>  
>  func NewMachineDriverLibvirtCache() *Cache {
> -       return newCache(libvirt.MachineDriverPath(), libvirt.MachineDriverDownloadURL, libvirt.MachineDriverVersion, getCurrentLibvirtDriverVersion)
> +       cache := newCache(libvirt.MachineDriverPath(), libvirt.MachineDriverDownloadURL, libvirt.MachineDriverVersion, getCurrentLibvirtDriverVersion)
> +       // upstream binaries are suffixed with the arch (crc-driver-libvirt-amd64)
> +       // crc caches the binary as crc-driver-libvirt as it's the name expected by github.com/crc-org/machine
> +       cache.ignoreNameMismatch = true
>  }
>  

This mean we are ignoring the binary name, I think current PR is better that this one where we are just testing if binary name have  arch suffix. 


> Or to update the machine code to run crc-driver-libvirt-$arch instead of crc-driver-libvirt

I do agree on that, but since this week is code freeze and we do need to deliver this time (hopefully without delay for 4.16) and if we make this change which means release of new version of crc-driver-libvirt. I can do that beginning of next sprint and then revert this change?

cfergeau:
> if we make this change which means release of new version of crc-driver-libvirt.

This only needs changes in `github.com/crc-org/machine`, and to pick up these changes in crc, I filed https://github.com/crc-org/crc/pull/4268 for this.

cfergeau:
>  we do need to deliver this time (hopefully without delay for 4.16) 

If this warning only shows up when downloading a helper, and not when the helper is embedded, then it's not a problem if a fix is not present in the release.

praveenkumar:
/close 

In favor of #4268

Issue #4099, opened 2024-04-03
[BUG] `Podman version` in `crc version` is meaningless?


cfergeau:
### General information

  * OS: Linux
 
## CRC version
```bash
# Put `crc version` output here
CRC version: 2.34.1+fa1b45
OpenShift version: 4.15.3
Podman version: 4.4.4
```

Recent OpenShift clusters use podman 4.4.1, my linux machine has a more recent podman, I suspect this 4.4.4 version is the version hardcoded there https://github.com/crc-org/crc/blob/2331271ac2ec679056e807d2e80a665c39032122/Makefile#L6 and has no meaning (it used to mean "if you download a podman bundle, the bundle version will be 4.4.4, and incidentally this corresponded to the podman version running in the bundle).

```bash
$ podman version
Client:       Podman Engine
Version:      4.9.3
API Version:  4.9.3
Go Version:   go1.21.7
Built:        Mon Feb 19 16:41:34 2024
OS/Arch:      linux/amd64
```

cfergeau:
Kind of related to https://github.com/crc-org/crc/issues/4098

rohanKanojia:
@cfergeau : is this issue resolved? I'm no longer able to see `Podman Version` in `crc version` output.

cfergeau:
Must have been removed as part of https://github.com/crc-org/crc/pull/4062 or https://github.com/crc-org/crc/pull/4167

rohanKanojia:
@cfergeau : I see this output on crc v2.42.0: 

```
CRC version: 2.42.0+261e97
OpenShift version: 4.17.0
MicroShift version: 4.16.14
```

I think we can close this issue as completed.

Issue #4098, opened 2024-04-03
[BUG] Microshift version is not reported as part of version command


adrianriobo:
Checking the `crc version` we can not see the microshift version even with the microshift preset 

``` bash
macmini-crcqe-2:2.34.1 crcqe$ crc setup --log-level debug
DEBU CRC version: 2.34.1+b470b5                   
DEBU OpenShift version: 4.15.3                    
DEBU Podman version: 4.4.4      
```

Not sure if that is on purpose, but during previous sprint it was discussed to release with versions out of sync, so then it had not been easy to see the microshift version

praveenkumar:
fixed by https://github.com/crc-org/crc/pull/4269

PR #4269, opened 2024-07-15
Show MicroShift version as part of 'version' command

A-725-K:
**Fixes:** Issue #4098

**Relates to:** Issue #4099

## Solution/Idea

Display MicroShift version as part of CRC version command output. Extend existing unit tests to verify that also this newly added field is present in the output.

## Proposed changes

1. Add `MicroshiftVersion` field to `Version` type.
2. Read the bundled version from `crcPreset`
3. Print out the value as part of the `version` command

## Testing

1. `crc` version contains Microshift version
2. if the output shows the correct Microshift version the test succeeded 
3. Extended existing unit test to verify the claim

### DIFF
--- a/cmd/crc/cmd/version.go
+++ b/cmd/crc/cmd/version.go
@@ -34,16 +34,18 @@ func runPrintVersion(writer io.Writer, version *version, outputFormat string) er
 }
 
 type version struct {
-	Version          string `json:"version"`
-	Commit           string `json:"commit"`
-	OpenshiftVersion string `json:"openshiftVersion"`
+	Version           string `json:"version"`
+	Commit            string `json:"commit"`
+	OpenshiftVersion  string `json:"openshiftVersion"`
+	MicroshiftVersion string `json:"microshiftVersion"`
 }
 
 func defaultVersion() *version {
 	return &version{
-		Version:          crcversion.GetCRCVersion(),
-		Commit:           crcversion.GetCommitSha(),
-		OpenshiftVersion: crcversion.GetBundleVersion(crcPreset.OpenShift),
+		Version:           crcversion.GetCRCVersion(),
+		Commit:            crcversion.GetCommitSha(),
+		OpenshiftVersion:  crcversion.GetBundleVersion(crcPreset.OpenShift),
+		MicroshiftVersion: crcversion.GetBundleVersion(crcPreset.Microshift),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


cfergeau:
I thought there were some discussions whether to show this or not when the microshift preset was introduced, but I can't find it. Maybe I misremember :)
/lgtm

adrianriobo:
@cfergeau  I can not remember about those conversation, but in any case I introduced the bug during an specific sprint when we released or almost did it with different versions for ocp and microshift. 

So having different versions would require for sure make the explicit somewhere.

anjannath (approved):

praveenkumar:
version api also need to be updated for this.

```patch
$ git diff
diff --git a/pkg/crc/api/client/types.go b/pkg/crc/api/client/types.go
index 590f1e..776270 100644
--- a/pkg/crc/api/client/types.go
+++ b/pkg/crc/api/client/types.go
@@ -7,9 +7,10 @@ import (
 )
 
 type VersionResult struct {
-       CrcVersion       string
-       CommitSha        string
-       OpenshiftVersion string
+       CrcVersion        string
+       CommitSha         string
+       OpenshiftVersion  string
+       MicroshiftVersion string
 }
 
 type StartResult struct {
diff --git a/pkg/crc/api/handlers.go b/pkg/crc/api/handlers.go
index 35e165..558818 100644
--- a/pkg/crc/api/handlers.go
+++ b/pkg/crc/api/handlers.go
@@ -136,9 +136,10 @@ func getStartConfig(cfg crcConfig.Storage, args client.StartConfig) types.StartC
 
 func (h *Handler) GetVersion(c *context) error {
        return c.JSON(http.StatusOK, &client.VersionResult{
-               CrcVersion:       version.GetCRCVersion(),
-               CommitSha:        version.GetCommitSha(),
-               OpenshiftVersion: version.GetBundleVersion(preset.OpenShift),
+               CrcVersion:        version.GetCRCVersion(),
+               CommitSha:         version.GetCommitSha(),
+               OpenshiftVersion:  version.GetBundleVersion(preset.OpenShift),
+               MicroshiftVersion: version.GetBundleVersion(preset.Microshift),
        })
 }
 

```

so that following works as expected
```
$ curl -X GET --unix-socket ~/.crc/crc-http.sock http:/c/api/version | jq .
{
  "CrcVersion": "2.38.0",
  "CommitSha": "855bd7",
  "OpenshiftVersion": "4.16.0",
  "MicroshiftVersion": "4.16.0"
}
```

A-725-K:
I'm sorry, I totally forgot about the API tests. I've updated the code to cover also these.

praveenkumar:
I am going to hold it for this release and then merge after release it done.

/hold

PR #4273, opened 2024-07-16
build: replace amd64 with GOARCH in make target 'release'

anjannath:
this allows to run 'make release' on an arm64 host where build output dir is out/linux-arm64 not out/linux-amd64


anjannath:
(I've included a patch in the rpm to generate the release builds so this is not urgent)

cfergeau (approved):

PR #4277, opened 2024-07-18
Fix cert expiration issue for aggregator client

praveenkumar:
During 4.17.0-ec.1 testing we found out that it might be possible that
kubelet certs are not expired but aggregator one is and due to that our
apiserver not able to get the node info and fails with following error.

```
INFO Verifying validity of the kubelet certificates...
DEBU Running SSH command: date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-client-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds
DEBU SSH command results: err: <nil>, output: 2025-07-05T03:48:39+00:00
DEBU Running SSH command: date --date="$(sudo openssl x509 -in /var/lib/kubelet/pki/kubelet-server-current.pem -noout -enddate | cut -d= -f 2)" --iso-8601=seconds
DEBU SSH command results: err: <nil>, output: 2025-07-05T03:49:24+00:00
DEBU Running SSH command: date --date="$(sudo openssl x509 -in /etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt -noout -enddate | cut -d= -f 2)" --iso-8601=seconds
DEBU SSH command results: err: <nil>, output: 2024-07-11T05:50:37+00:00
DEBU Certs have expired, they were valid till: 11 Jul 24 05:50 +0000
```

This PR make sure we also wait to recover the api server related certs
before checking the apiserver is responding.

### DIFF
--- a/pkg/crc/cluster/cert_renewal.go
+++ b/pkg/crc/cluster/cert_renewal.go
@@ -43,7 +43,7 @@ func approvePendingCSRs(ctx context.Context, ocConfig oc.Config, expectedSignerN
 	}, time.Second*5)
 }
 
-func ApproveCSRAndWaitForCertsRenewal(ctx context.Context, sshRunner *ssh.Runner, ocConfig oc.Config, client, server bool) error {
+func ApproveCSRAndWaitForCertsRenewal(ctx context.Context, sshRunner *ssh.Runner, ocConfig oc.Config, client, server, aggregratorClient bool) error {
 	const (
 		kubeletClientSignerName  = "kubernetes.io/kube-apiserver-client-kubelet"
 		kubeletServingSignerName = "kubernetes.io/kubelet-serving"
@@ -76,6 +76,10 @@ func ApproveCSRAndWaitForCertsRenewal(ctx context.Context, sshRunner *ssh.Runner
 	if server {
 		logging.Info("Kubelet serving certificate has expired, waiting for automatic renewal... [will take up to 5 minutes]")
 		return crcerrors.Retry(ctx, 5*time.Minute, waitForCertRenewal(sshRunner, KubeletServerCert), time.Second*5)
+	}
+	if aggregratorClient {
+		logging.Info("Kube API server certificate has expired, waiting for automatic renewal... [will take up to 8 minutes]")
+		return crcerrors.Retry(ctx, 8*time.Minute, waitForCertRenewal(sshRunner, AggregatorClientCert), time.Second*5)
 	}
 	return nil
 }
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -32,7 +32,7 @@ const (
 	KubeletServerCert = "/var/lib/kubelet/pki/kubelet-server-current.pem"
 	KubeletClientCert = "/var/lib/kubelet/pki/kubelet-client-current.pem"
 
-	AggregatorClientCert = "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt"
+	AggregatorClientCert = "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/aggregator-client/tls.crt"
 )
 
 func CheckCertsValidity(sshRunner *ssh.Runner) (map[string]bool, error) {
--- a/pkg/crc/cluster/cluster.go
+++ b/pkg/crc/cluster/cluster.go
@@ -432,20 +432,6 @@ func WaitForPullSecretPresentOnInstanceDisk(ctx context.Context, sshRunner *ssh.
 	return errors.Retry(ctx, 7*time.Minute, pullSecretPresentFunc, 2*time.Second)
 }
 
-func WaitForRequestHeaderClientCaFile(ctx context.Context, sshRunner *ssh.Runner) error {
-	lookupRequestHeaderClientCa := func() error {
-		expired, err := checkCertValidity(sshRunner, AggregatorClientCert)
-		if err != nil {
-			return fmt.Errorf("Failed to the expiry date: %v", err)
-		}
-		if expired {
-			return &errors.RetriableError{Err: fmt.Errorf("certificate still expired")}
-		}
-		return nil
-	}
-	return errors.Retry(ctx, 8*time.Minute, lookupRequestHeaderClientCa, 2*time.Second)
-}
-
 func WaitForAPIServer(ctx context.Context, ocConfig oc.Config) error {
 	logging.Info("Waiting for kube-apiserver availability... [takes around 2min]")
 	waitForAPIServer := func() error {
@@ -458,23 +444,6 @@ func WaitForAPIServer(ctx context.Context, ocConfig oc.Config) error {
 		return nil
 	}
 	return errors.Retry(ctx, 4*time.Minute, waitForAPIServer, time.Second)
-}
-
-func DeleteOpenshiftAPIServerPods(ctx context.Context, ocConfig oc.Config) error {
-	if err := WaitForOpenshiftResource(ctx, ocConfig, "pod"); err != nil {
-		return err
-	}
-
-	deleteOpenshiftAPIServerPods := func() error {
-		cmdArgs := []string{"delete", "pod", "--all", "--force", "-n", "openshift-apiserver"}
-		_, stderr, err := ocConfig.WithFailFast().RunOcCommand(cmdArgs...)
-		if err != nil {
-			return &errors.RetriableError{Err: fmt.Errorf("Failed to delete pod from openshift-apiserver namespace %v: %s", err, stderr)}
-		}
-		return nil
-	}
-
-	return errors.Retry(ctx, 60*time.Second, deleteOpenshiftAPIServerPods, time.Second)
 }
 
 func CheckProxySettingsForOperator(ocConfig oc.Config, proxy *httpproxy.ProxyConfig, deployment, namespace string) (bool, error) {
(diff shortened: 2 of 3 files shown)


pkg/crc/cluster/cluster.go
```diff
@@ -32,7 +32,7 @@ const (
 	KubeletServerCert = "/var/lib/kubelet/pki/kubelet-server-current.pem"
 	KubeletClientCert = "/var/lib/kubelet/pki/kubelet-client-current.pem"
 
-	AggregatorClientCert = "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt"
+	AggregatorClientCert = "/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/aggregator-client/tls.crt"
```
cfergeau, line 35: The commit log says
> During testing of 4.17.0.ec.1 found out that `secrets/aggregator-client/tls.crt` is regenerated as part of auto cert rotation before `configmaps/aggregator-client-ca/ca-bundle.crt`.

Is `configmaps/aggregator-client-ca` really regenerated after `secrets/aggregator-client`? Or does this happen in the opposite order?

pkg/crc/cluster/cert_renewal.go
```diff
@@ -92,3 +92,11 @@ func waitForCertRenewal(sshRunner *ssh.Runner, cert string) func() error {
 		return &crcerrors.RetriableError{Err: fmt.Errorf("certificate %s still expired", cert)}
 	}
 }
+
+func WaitForKubeAPIServerCertRenewal(ctx context.Context, sshRunner *ssh.Runner, expired bool) error {
```
cfergeau: I'd add a `apiserver` boolean to `ApproveCSRAndWaitForCertsRenewal` and keep all the cert renewal-related code in this function.

pkg/crc/machine/start.go
```diff
@@ -557,6 +557,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		return nil, errors.Wrap(err, "Failed to renew TLS certificates: please check if a newer CRC release is available")
 	}
 
+	if err := cluster.WaitForKubeAPIServerCertRenewal(ctx, sshRunner, certsExpired[cluster.AggregatorClientCert]); err != nil {
+		return nil, errors.Wrap(err, "Failed to renew Kube API server cert: please check if a newer CRC release is available")
+	}
+
 	if err := cluster.WaitForAPIServer(ctx, ocConfig); err != nil {
```
cfergeau: Isn't this `WaitForAPIServer` call also retrying multiple times in case the `oc` calls fail because of an expired certificate?

cfergeau:
I assume these changes also work on a 4.16 cluster with no significant impact on startup time/.. when cert renewal is involved?

pkg/crc/cluster/cluster.go
cfergeau: As a preparatory commit, I'd remove https://github.com/crc-org/crc/blob/2fa7958844be56353ebfd6c104281a77a5a24b5c/pkg/crc/machine/start.go#L605-L626 as this change has an impact on this code.

praveenkumar: Yes it is really regenerated after `secrets/aggregator-client`, I ssh'ed to the node and checked this.

pkg/crc/cluster/cert_renewal.go
praveenkumar: sure this can be become helper function in that case.

pkg/crc/machine/start.go
praveenkumar: yes it does but only for 2 mins and generation of cert happen in around ~7-8 mins.

praveenkumar:
> I assume these changes also work on a 4.16 cluster with no significant impact on startup time/.. when cert renewal is involved?

Yes it works and it is not going to impact startup time in case cert is not expired and even it is expired it should add that ~8-10 mins which we already expect.

pkg/crc/cluster/cluster.go
cfergeau: It's quite odd, as it would make more sense if `secrets/aggregator-client` used the new CA, not an older one.
If `configmaps/aggregator-client-ca` is the last to be regenerated, why do we change to check for `secrets/aggregator-client`? When `aggregator-client-ca` is fully renewed, `aggregator-client` should also be fine?

praveenkumar: > As a preparatory commit, I'd remove 

Removed as first commit

>It's quite odd, as it would make more sense if secrets/aggregator-client used the new CA, not an older one.
If configmaps/aggregator-client-ca is the last to be regenerated, why do we change to check for secrets/aggregator-client? When aggregator-client-ca is fully renewed, aggregator-client should also be fine?

During that internal conversation figure out the due to certsyncer controller it might happen that cert files sync on the disk is out of order, added more context in the commit log.

pkg/crc/cluster/cert_renewal.go
praveenkumar: Done

pkg/crc/cluster/cert_renewal.go
```diff
@@ -77,6 +77,10 @@ func ApproveCSRAndWaitForCertsRenewal(ctx context.Context, sshRunner *ssh.Runner
 		logging.Info("Kubelet serving certificate has expired, waiting for automatic renewal... [will take up to 5 minutes]")
 		return crcerrors.Retry(ctx, 5*time.Minute, waitForCertRenewal(sshRunner, KubeletServerCert), time.Second*5)
 	}
+	if aggregratorClient {
+		logging.Info("Kube API server certificate has expired, waiting for automatic renewal... [will take up to 8 minutes]")
+		return crcerrors.Retry(ctx, 8*time.Minute, waitForCertRenewal(sshRunner, AggregatorClientCert), time.Second*5)
```
cfergeau, line 82: Do we need 8 minutes just for this cert ?

pkg/crc/cluster/cluster.go
cfergeau: >  I ssh'ed to the node and checked this.

With `ls -al`, or with something like `$ cat /etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/aggregator-client/tls.crt | openssl x509 -noout -enddate` ? I assumed you used `openssl` to check this, but seeing the updated commit log and the comment above this one, I'm thinking you might have checked with `ls` instead?

pkg/crc/cluster/cert_renewal.go
praveenkumar: @cfergeau yes it takes around ~7-8 mins

pkg/crc/cluster/cluster.go
praveenkumar: I checked with openssl also and with openssl, timestamp is as expected but with ls it is not because certsyncer is handled by different controller which I mentioned in the commit log.
```
During testing of 4.17.0.ec.1 found out that
`secrets/aggregator-client/tls.crt` is synced before `configmaps/aggregator-client-ca/ca-bundle.crt` which is due to the certsyncer controller.
```

cfergeau: I'd be worried if the openssl timestamps were out of order. The timestamps of the files on disk are less relevant, especially with the explanation you have now.

praveenkumar: No it were not in the order, I can mention that in the commit log but since in our code we check the file so having that explanation would be also important.

cfergeau: >  in our code we check the file 

We check it's modified/accessed/created time with `ls` / `stat` / ... ? I could only find use of `openssl x509 ...`

pkg/crc/cluster/cert_renewal.go
cfergeau: 7/8 minutes on top of the waits for the kubelet serving certificate and the kubelet client certificate, each of which can take up to 5 minutes? That feels like a lot

praveenkumar: @cfergeau no I think during that time even those cert also rotating so it is mostly overall time for any cert expire

pkg/crc/cluster/cluster.go
praveenkumar: > We check it's modified/accessed/created time with `ls` / `stat` / ... ? I could only find use of `openssl x509 ...`

we check the cert expiry of the file using `openssl` command only and  this command run against  `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/configmaps/aggregator-client-ca/ca-bundle.crt` file,  during testing this file is not updated with rotated cert before `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/aggregator-client/tls.crt` one hence this change.

cfergeau: Can we tell somehow to the certsyncer controller it should flush the certs to disk? Adjusting our code to whatever it flushes to disk first feels like something which could change/break anytime, and need more changes, so it would be better if  we could tell it we want to work from a well-known state on disk?

praveenkumar: >Can we tell somehow to the certsyncer controller it should flush the certs to disk? 

@cfergeau it is the controller which is not under our control and during the internal conversation around it looks like we can't do much around it.

>  Adjusting our code to whatever it flushes to disk first feels like something which could change/break anytime, and need more changes, so it would be better if we could tell it we want to work from a well-known state on disk?

Same feeling but looks like chicken and egg situation because it is updated to etcd first (which is source of truth) as part of configmap but since the api is not working until cert is rotated so we can't put code to pull it :(

cfergeau (approved):
Not great that we cannot force the certoperator to sync its state to disk, and have to hope its external behaviour won't change too often :-/

Issue #4279, opened 2024-07-19
[BUG]  WSL2 validation check is currently case senstive


alegacy:
### General information

  * OS: Linux
  * Hypervisor: WSL2
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: WSL2

## CRC version
```bash
CRC version: 2.39.0+2fa795
OpenShift version: 4.16.0
MicroShift version: 4.16.0
```
  
## CRC status
```bash
DEBU CRC version: 2.39.0+2fa795                   
DEBU OpenShift version: 4.16.0                    
DEBU MicroShift version: 4.16.0                   
DEBU Running 'crc status'                         
Machine does not exist. Use 'crc start' to create it
```

## CRC config
```bash
- consent-telemetry                     : yes
```

## Host Operating System
```bash
NAME="Fedora Remix for WSL"
VERSION="40"
ID=fedoraremixforwsl
ID_LIKE=fedora
VERSION_ID=40
PLATFORM_ID="platform:f40"
PRETTY_NAME="Fedora Remix for WSL"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:40"
HOME_URL="https://github.com/WhitewaterFoundry/Fedora-Remix-for-WSL"
SUPPORT_URL="https://github.com/WhitewaterFoundry/Fedora-Remix-for-WSL"
BUG_REPORT_URL="https://github.com/WhitewaterFoundry/Fedora-Remix-for-WSL/issues"
PRIVACY_POLICY_URL="https://github.com/WhitewaterFoundry/Fedora-Remix-for-WSL/blob/master/PRIVACY.md"
FEDORA_REMIX_VERSION=40.0.1

```

### Steps to reproduce

  1. crc setup
  2. crc start

### Expected

Should fail to start on WSL2 since not supported.  See issue #2131 and fix #2323.

### Actual

`crc start` attempts to start the VM and fails.

### Logs
```bash
...
INFO Generating new SSH key pair...               
INFO Generating new password for the kubeadmin user 
INFO Starting CRC VM for openshift 4.16.0...      
Error starting machine: Error in driver during machine start: host is not running: getting ip during machine start
```

The output of `/proc/version` is this:

```bash
[...]

Issue #2131, opened 2021-03-21
Add check to the Linux pre-flights to prevent installation on WSL2


gbraad:
as per https://github.com/code-ready/crc/issues/374#issuecomment-532268319 and other reports I have noticed people try to install CRC inside the WSL2 environment; this is unsupported from the nested virtualization perspective, but also not recommended.

Since they run WSL2, Hyper-V is installed and working, so a native installation will certainly work.

PR #4280, opened 2024-07-23
Issue #4279 Fix case sensitive WSL2 validation check

alegacy:
The WSL validation check assumes that the name "Microsoft" is capitalized but this is not true in all cases.

**Fixes:** Issue #4279

**Relates to:** Issue #2131 , PR #2323 

## Solution/Idea

Enhancing WSL validation check to be case-insensitive to prevent users from attempting to run on an unsupported platform.

## Proposed changes

Updated validation check to force `toLower(...)` on the output of `/proc/version` prior to searching for the target substring of `microsoft`.

## Testing

Validation check now prevents running on systems running WSL regardless of the capitalization of the word Microsoft.

```shell
alegacy@localhost:~/src/go/github.com/crc$ ./out/crc setup
INFO Using bundle path /home/alegacy/.crc/cache/crc_libvirt_0.0.0-unset_amd64.crcbundle 
CRC is unsupported using WSL2

alegacy@localhost:~/src/go/github.com/crc$ cat /proc/version
Linux version 5.15.153.1-microsoft-standard-WSL2 (root@941d701f84f1) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP Fri Mar 29 23:14:13 UTC 2024
```shell

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -36,7 +36,7 @@ func checkRunningInsideWSL2() error {
 		return err
 	}
 
-	if strings.Contains(string(version), "Microsoft") {
+	if strings.Contains(strings.ToLower(string(version)), "microsoft") {
 		logging.Debugf("Running inside WSL2 environment")
 		return fmt.Errorf("CRC is unsupported using WSL2")
 	}


cfergeau (approved):

gbraad (approved):

gbraad:
Thanks. The test can be stricter, as it contains `WSL2` too when MS provided the kernel.
Note: however, any custom kernel will not contain this, so the current test is not restrictive enough (but works for most cases).

alegacy:
@cfergeau @gbraad @praveenkumar  ... some tests appear to be failing.  I don't think any of them are affected by my changes.   Do these tests have a history of intermittent failures?   Should I re-trigger and hope for the best or is there some automation that will come around to do that?

cfergeau:
> @cfergeau @gbraad @praveenkumar ... some tests appear to be failing. I don't think any of them are affected by my changes. Do these tests have a history of intermittent failures? Should I re-trigger and hope for the best or is there some automation that will come around to do that?

Don't worry about the tests, they are unfortunately very unreliable :(

PR #4281, opened 2024-07-23
go.mod: Update gvisor-tap-vsock to latest git

cfergeau:
This adds
https://github.com/containers/podman/pull/21262
which should make usermode networking more stable on high traffic/fast
networks on macos.

I'm creating this PR to trigger test runs on Windows to make sure this does not break crc.
It does not need to be merged as if it works, I'll cut a gvisor-tap-vsock release, and then we can update crc to this release.


cfergeau:
Superseded by #4283

PR #4288, opened 2024-07-29
Update libvirt driver to 0.13.9

praveenkumar:


pkg/crc/machine/libvirt/constants.go
```diff
@@ -21,7 +21,7 @@ const (
 )
 
 const (
-	MachineDriverVersion = "0.13.8"
+	MachineDriverVersion = "0.13.9"
```
cfergeau, line 24: You also need to update `images/rpmbuild/Containerfile.in`

cfergeau (approved):

PR #4289, opened 2024-07-29
misc updates to go mod and update script to accommodate go toolchain changes

praveenkumar:
Starting in Go 1.21, Go toolchains will treat the go line in go.mod not
as a guideline but as a rule, and the line can list specific point releases
or release candidates. That is, Go 1.21.0 understands that it cannot even
build code that says go 1.21.1 in its go.mod file, not to mention code that
says much later versions like go 1.22.0.

From go1.21 toolchain is enabled by default as auto mode and according
to blog post[0], which means that if a module specify `go 1.21.0` in the
mod file. In case of image module it looks like they switched to using
`go 1.21.0` instead of `go 1.21` in the mod file (for 5.32 tag release)
and which stop us to using it as it is.

```
module m
go 1.21.0
toolchain go1.21.4

This says that other modules requiring m need to provide at least Go 1.21.0, but when we are working in m itself, we want an even newer toolchain, at least Go 1.21.4
```

In this PR we are going to update the script so that we can use
`go1.N.P` version and filter `go1.N` for images and go runner.
In next PR we just update the go version to `1.21.0`

[0] https://go.dev/blog/toolchain


update-go-version.sh
```diff
@@ -12,14 +12,17 @@ echo "Updating golang version to $golang_base_version"
 go mod edit -go ${golang_base_version}
 go mod edit -go ${golang_base_version} tools/go.mod
 sed -i "s,^GOVERSION = 1.[0-9]\+,GOVERSION = ${golang_base_version}," Makefile
-sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-\)1.[0-9]\+,\1${golang_base_version}," images/*/Dockerfile
-sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Dockerfile
-sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Containerfile
+
+golang_major_version=$(echo ${golang_base_version} | cut -d '.' -f 1,2)
```
cfergeau: The comment at the beginning of the file will need to be updated
```
# Updates our various build files, CI configs, ... to the latest released go
# version in a given minor release stream (1.x)
```
as you change the format of the parameter which needs to be passed to this script, and what the script used to call "minor release stream" is now called "golang major version" (which is more accurate https://go.dev/doc/devel/release talks about major releases (1.21.0) and minor revisions (1.21.4))

An alternative to these changes would be to keep passing `1.21` as the argument to the script, and to only change the `go.mod` modification to append a `.0` to the argument. `./update-go-version 1.21.0` is misleading as most of the changes will be about using the latest 1.21.x version available, not about using 1.21.0 precisely.

update-go-version.sh
```diff
@@ -12,14 +12,17 @@ echo "Updating golang version to $golang_base_version"
 go mod edit -go ${golang_base_version}
 go mod edit -go ${golang_base_version} tools/go.mod
 sed -i "s,^GOVERSION = 1.[0-9]\+,GOVERSION = ${golang_base_version}," Makefile
```
cfergeau: If you change `GOVERSION` to 1.21.0 in the next commit, you need to change this line.

Makefile
```diff
@@ -28,7 +28,7 @@ DOCS_BUILD_TARGET ?= /docs/source/getting_started/master.adoc
 
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-GOVERSION = 1.21
+GOVERSION = 1.21.0
```
cfergeau: `GOMINVERSION`?

cfergeau (commented):
> From go1.21 toolchain is enabled by default as auto mode and according to blog post[0], which means that if a module specify `go 1.21.0` in the mod file. In case of image module it looks like they switched to using `go 1.21.0` instead of `go 1.21` in the mod file (for 5.32 tag release) and which stop us to using it as it is.

This part of the commit log is quite hard to read/parse :-/

update-go-version.sh
praveenkumar: >An alternative to these changes would be to keep passing 1.21 as the argument to the script, and to only change the go.mod modification to append a .0 to the argument. ./update-go-version 1.21.0 is misleading as most of the changes will be about using the latest 1.21.x version available, not about using 1.21.0 precisely.

yes, this is much better because only go mod file need to have that toolchain update but runner and images doesn't

update-go-version.sh
```diff
@@ -17,9 +17,9 @@ sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registr
 sed -i "s,^FROM registry.access.redhat.com/ubi8/go-toolset:[.0-9]\+,FROM registry.access.redhat.com/ubi8/go-toolset:${golang_base_version}," images/*/Containerfile
 for f in .github/workflows/*.yml; do
     if [ $(yq  eval '.jobs.build.strategy.matrix | has("go")' "$f") == "true" ]; then
-      yq eval --inplace ".jobs.build.strategy.matrix.go[0] = ${golang_base_version} | .jobs.build.strategy.matrix.go[0] style=\"single\"" "$f";
+      yq eval --inplace ".jobs.build.strategy.matrix.go[0] = \"${golang_base_version}\" | .jobs.build.strategy.matrix.go[0] style=\"single\"" "$f";
```
cfergeau: With the latest changes in this PR, `golang_base_version` is `1.21`, not `1.21.0` as in the commit log, this is no longer needed (at least according to my testing with yq 4.43.1
The changes in this commit can apparently be simplified to:
```
 yq eval --inplace ".jobs.build.strategy.matrix.go[0] = \"${golang_base_version}\"" "$f";
```

praveenkumar: it is working because we have already single quote around the version info, if we remove it on one of that then it will just use the double quote with `single` style we are explicit making it to use single quote. 

yes if we are using `1.21` only then this is not needed.

cfergeau (approved):

Issue #4308, opened 2024-08-07
Switch to uefi for vfkit on Mac13 and later


praveenkumar:
We are not supporting the older version of mac13 and uefi is supported as part of vfkit so may be better to remove the legacy side of things in the vfkit.

praveenkumar:
close dup https://github.com/crc-org/crc/issues/4180

PR #4309, opened 2024-08-07
vfkit: Switch to EFIBootLoader from LinuxBootLoader (BIOS)

praveenkumar:
vfkit have a way to load efi bootloader and since we are now supporting only macos>13, better to switch it and it also allow a bit code clean around kernel update when start from stopped state.

**Fixes:** Issue #4308


vyasgun:
Tested this and it is working. The corresponding `vfkit` command:
`/Users/podmanqe/.crc/bin/vfkit --cpus 2 --memory 4096 --bootloader efi,variable-store=/Users/podmanqe/.crc/machines/crc/efistore.nvram,create --device virtio-serial,logFilePath=/Users/podmanqe/.crc/machines/crc/vfkit.log --device virtio-fs,sharedDir=/Users/podmanqe,mountTag=dir0 --device virtio-rng --device virtio-blk,path=/Users/podmanqe/.crc/machines/crc/crc.img --device virtio-vsock,port=1024,socketURL=/Users/podmanqe/.crc/tap.sock,listen --timesync vsockPort=1234`

Issue #4284, opened 2024-07-24
[BUG] crc start error doesn't exit with non-zero exit code


bobbygryzynger:
### General information

  * OS: Linux 
  * Hypervisor: KVM
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop

## CRC version
```bash
CRC version: 2.22.1+e8068b4
OpenShift version: 4.13.3
Podman version: 4.4.4
```
  
## CRC status
```bash
$ crc status --log-level debug
DEBU CRC version: 2.22.1+e8068b4                  
DEBU OpenShift version: 4.13.3                    
DEBU Podman version: 4.4.4                        
DEBU Running 'crc status'                         
CRC VM:          Running
OpenShift:       Starting (v4.13.3)
RAM Usage:       9.412GB of 16.77GB
Disk Usage:      20.8GB of 79.93GB (Inside the CRC VM)
Cache Usage:     202.3GB
Cache Directory: /home/bgryzyng/.crc/cache
```

## CRC config
```bash
$ crc config view
- consent-telemetry                     : yes
- cpus                                  : 6
- disk-size                             : 75
- enable-cluster-monitoring             : true
- host-network-access                   : true
- memory                                : 16384
- network-mode                          : user
```

## Host Operating System
```bash
$ cat /etc/os-release
NAME="Fedora Linux"
...
```

### Steps to reproduce

  1. Run `crc start`
  2.  If an error, occurs a non-zero exit code is not provided

### Expected
If an error occurs, exit code should be 1 or greater

### Actual
Exit code is zero

### Logs
```
...
INFO 3 operators are progressing: authentication, kube-apiserver, monitoring 
INFO 3 operators are progressing: authentication, kube-apiserver, monitoring 
INFO 3 operators are progressing: authentication, kube-apiserver, monitoring 
INFO 3 operators are progressing: authentication, kube-apiserver, monitoring 
INFO 3 operators are progressing: authentication, kube-apiserver, monitoring 
ERRO Cluster is not ready: cluster operators are still not stable after 10m20.6920432s 
INFO Adding crc-admin and crc-developer contexts to kubeconfig... 
Started the OpenShift cluster.
...
```

praveenkumar:
> ERRO Cluster is not ready: cluster operators are still not stable after 10m20.6920432s 

[...]

PR #4315, opened 2024-08-12
Print warning instead of error in case of unstable cluster

vyasgun:
**Fixes:** Issue #4284

## Solution/Idea
Since the code doesn't exit, the error messaging might be confusing to users. Changed it to `Warn`
```
INFO Operator network is progressing
INFO Operator network is progressing
INFO Operator network is progressing
INFO Operator network is progressing
INFO Operator network is progressing
INFO Operator network is progressing
INFO Operator network is progressing
WARN Cluster is not ready: cluster operators are still not stable after 10m0.695631268s
ERRO Cannot update kubeconfig: Head "https://oauth-openshift.apps-crc.testing:443": read tcp 127.0.0.1:60782->127.0.0.1:443: read: connection reset by peer

Log in as administrator:
  Username: kubeadmin
  Password: 3NM8K-C5kvg-YTRW4-FhiUM

Log in as user:
  Username: developer
  Password: developer

  $ eval $(crc oc-env)
  $ oc login -u developer https://api.crc.testing:6443
```

## Testing
`crc start` and cluster operators should not get ready within the timeout. I did it by cordoning the cluster node and running start again.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -608,7 +608,7 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 
 	logging.Infof("Starting %s instance... [waiting for the cluster to stabilize]", startConfig.Preset)
 	if err := cluster.WaitForClusterStable(ctx, instanceIP, constants.KubeconfigFilePath, proxyConfig); err != nil {
-		logging.Errorf("Cluster is not ready: %v", err)
+		logging.Warnf("Cluster is not ready: %v", err)
 	}
 
 	waitForProxyPropagation(ctx, ocConfig, proxyConfig)


praveenkumar:
@vyasgun can you check in case of `ERRO Cannot update kubeconfig: Head "https://oauth-openshift.apps-crc.testing:443": read tcp 127.0.0.1:60782->127.0.0.1:443: read: connection reset by peer
Started the OpenShift cluster.` the error code is non-zero ?

vyasgun:
@praveenkumar No. Do we want it to return a non-zero exit code?

praveenkumar:
> @praveenkumar No. Do we want it to return a non-zero exit code?

For any error, yes we should return non-zero otherwise we should change it to warn but I think if we are not able to update the kubeconfig file then tell user how they can still access it.

vyasgun:
@praveenkumar In the`Start` function, only 2 `logging.Errorf()` statements have been used and neither of them is followed by a non-zero return.
Also, kubeconfig is being updated in other places inside the function and all of them except the last one are returning an error. For example:
https://github.com/crc-org/crc/blob/main/pkg/crc/machine/start.go#L603
https://github.com/crc-org/crc/blob/main/pkg/crc/machine/start.go#L528

I'm not sure if there is a particular reason for these statements and the differences (or if it's just an oversight). Additionally, I think if updating kubeconfig is grounds for a non-zero return, so is an unstable cluster (to indicate a failure in `Start`). We should return these errors in the end and put any extra processing that might be still required in `defer` so it's always executed.

cfergeau:
>  Additionally, I think if updating kubeconfig is grounds for a non-zero return, so is an unstable cluster (to indicate a failure in Start).

We could return a different error code in both cases when crc completes. Different error codes we might want to ignore 'cluster unstable'

The 'cannot update kubeconfig' message deserves to be made a lot more userfriendly :) Explain what won't work when this fails (I think this only means kube contexts can't be used, and that an explicit login to the cluster will be needed).

praveenkumar:
> > Additionally, I think if updating kubeconfig is grounds for a non-zero return, so is an unstable cluster (to indicate a failure in Start).
> 
> We could return a different error code in both cases when crc completes. Different error codes we might want to ignore 'cluster unstable'
> 
> The 'cannot update kubeconfig' message deserves to be made a lot more userfriendly :) Explain what won't work when this fails (I think this only means kube contexts can't be used, and that an explicit login to the cluster will be needed).

Yes, it should tell user how to access the cluster (like export `KUBECONFIG=$HOME/.crc/machine/crc/kubeconfig` or use `oc --kubeconfig=$HOME/.crc/machine/crc/kubeconfig` . They should able to debug or check which cluster operator is not in available state.

cfergeau:
We probably can get this in, and create follow-up issues for the various improvements that have been discussed during the review?

praveenkumar:
@vyasgun Can you create the follow up issue which is discussed here? Once follow up issue is created we can merge this one.

vyasgun:
Created a follow up issue: https://github.com/crc-org/crc/issues/4395

PR #4320, opened 2024-08-19
add GitHub action that use Testing Farm

lilyLuLiu:
https://github.com/crc-org/crc/issues/4326

### DIFF
--- a/.github/workflows/linux-qe-template.yml
+++ b/.github/workflows/linux-qe-template.yml
@@ -1,0 +1,217 @@
+name: linux-qe-template
+
+on:
+  workflow_call:  
+    inputs:  
+      trigger-workflow-run-id:  
+        required: true  
+        type: string
+      qe-type:  
+        description: type of test; allowed values e2e or integration
+        required: true  
+        type: string
+      preset:
+        description: preset type only required if qe-type is e2e
+        type: string  
+
+jobs:
+  linux-qe:
+    runs-on: [self-hosted, linux, testing-farm]
+    permissions:
+      statuses: write # needed to update commit status (pending/failure/sucess)
+      checks: write # as documented in https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#pr-run-permissions
+    steps:
+      - name: prepare env
+        run: |
+          sudo yum install podman openssh-server -y
+
+          commit_sha=$(cat gh_context.json | jq -r '.event.after')
+          if [[ -z "${commit_sha}" ]] || [[ "${commit_sha}" == null ]]; then
+            # on first PR creation .event.after is empty, then .sha is used as commit instead
+            commit_sha=$(cat gh_context.json | jq -r '.event.pull_request.head.sha')
+          fi
+          echo "commit_sha=${commit_sha}" >> "$GITHUB_ENV"
+          mkdir ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+          cd ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+
+          # the target can only be accessed through a bastion (which can only be accessed from self-hosted runner) 
+          # as so we need to map the ssh-agent from the host to the containers used to access the target host
+          ssh-keygen -t rsa -N '' -f id_rsa -q 
+          eval $(ssh-agent -s)
+          echo $SSH_AUTH_SOCK > ssh_auth_sock
+          echo $SSH_AGENT_PID > ssh_agent_pid
+          ssh-add id_rsa
+
+      - name: reserve machine from testing farm
+        working-directory: ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+        env: 
+          TESTING_FARM_API_TOKEN: ${{ secrets.TESTING_FARM_API_TOKEN }}
+          PULL_SECRET: ${{ secrets.PULL_SECRET }}
+        run: |
+          echo "${PULL_SECRET}" > pull-secret
+          export TESTING_FARM_API_TOKEN=${TESTING_FARM_API_TOKEN}
+          testing-farm reserve --compose Fedora-40 --duration 240 --arch aarch64 --hardware memory='>= 12 GB' --hardware virtualization.is-supported='true' --ssh-public-key id_rsa.pub --no-autoconnect | tee info
+          machine=`tail -n 1 info`
+          echo ${machine##*@} > host
+          echo crctest > username
+          echo proxy > bastion_username
+          echo testing-farm.io > bastion_host
+          request=`sed -n '4p' info`
+          echo ${request:1} > requestid
+
+          # connect to the reserved machine, create a non-root user for testing
+          ${machine:1}
+          new_user=crctest
+          useradd $new_user
+          echo "$new_user:redhat" | chpasswd
+          usermod -aG wheel $new_user
+          echo "$new_user ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/crctest-users
+          mkdir -p /home/$new_user/.ssh
+          cp /root/.ssh/authorized_keys /home/$new_user/.ssh/
+          chown $new_user /home/$new_user/.ssh/authorized_keys
+          exit
+     
+      - name: Download linux binary
+        working-directory: ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+        uses: actions/download-artifact@v4
+        with:
+          name: linux-binary
+          run-id: ${{inputs.trigger-workflow-run-id}}
+          github-token: ${{ github.token }}
+
+      - name: Download qe oci image
+        working-directory: ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+        id: download-qe-oci-image-artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: crc-${{inputs.qe-type}}-linux-arm64
+          run-id: ${{inputs.trigger-workflow-run-id}}
+          github-token: ${{ github.token }}
+
+      - name: Add status to the PR check
+        run: |
+          set -xuo
+          # Status msg
+          data="{\"state\":\"pending\""
+          data="${data},\"description\":\"Running ${{inputs.qe-type}}-${{inputs.preset}} on Linux ARM64\""
+          data="${data},\"context\":\"${{ env.status_context }}\""
+          data="${data},\"target_url\":\"https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}\"}"
+          # Create status by API call
+          curl -L -v -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ github.token }}" \
+            https://api.github.com/repos/${{ github.repository }}/statuses/${{ env.commit_sha }} \
+            -d "${data}"
+
+      - name: Install CRC on host
+        working-directory: ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+        run: |
+          podman run --rm -d --privileged --name crc-linux-install-${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}} \
+            -e TARGET_HOST=$(cat host) \
+            -e TARGET_HOST_USERNAME=$(cat username) \
+            -e TARGET_HOST_KEY_PATH=/data/id_rsa \
+            -e BASTION_HOST_USERNAME=$(cat bastion_username) \
+            -e BASTION_HOST=$(cat bastion_host) \
+            -e TARGET_FOLDER=crc-support \
+            -e TARGET_CLEANUP='false' \
+            -e OUTPUT_FOLDER=/data \
+            -e DEBUG='true' \
+            -e SSH_AUTH_SOCK=$(cat ssh_auth_sock) \
+            -v "$(cat ssh_auth_sock):$(cat ssh_auth_sock)" \
+            -v ${PWD}:/data:z \
+            -v ${PWD}/crc:/opt/crc-support/crc:z \
+            quay.io/crc-org/ci-crc-support:v2.0.0-dev-linux crc-support/run.sh \
+             -targetPath "/root/crc-support" \
+             -install 'true' \
+             -aName 'crc' \
+             -freshEnv 'false' \
+             -download 'false' 
+          podman logs -f crc-linux-install-${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+
+      - name: Run CRC ${{inputs.qe-type}} test
+        working-directory: ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+        run: |
+          # load image
+          podman load -i crc-${{inputs.qe-type}}-linux-arm64.tar
+          # run
+          cmd="crc-qe/run.sh -junitFilename crc-${{inputs.qe-type}}-junit.xml -targetFolder crc-qe"
+          if [[ "${{inputs.qe-type}}" == "e2e" ]]; then
+            if [[ "${{inputs.preset}}" == "microshift" ]]; then
+              cmd="${cmd} -e2eTagExpression '@story_microshift'"
+            else
+              cmd="${cmd} -e2eTagExpression '~@minimal && ~@story_microshift'"
+            fi
+          fi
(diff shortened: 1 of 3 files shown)


.github/workflows/testing-farm.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: prepare env
+        run: |
+          sudo yum install podman openssh-server -y
+          sudo systemctl enable sshd
+          sudo systemctl start sshd
+          sudo systemctl status sshd
+      - name: reserve machine from testing farm
+        run: |
+          ssh-keygen -t rsa -N '' -f id_rsa -q 
+          export TESTING_FARM_API_TOKEN=059e594d-12da-4ec6-951f-9069e99cf2c8
```
cfergeau, line 20: My feeling is that this API key should be a secret? https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions

lilyLuLiu: @cfergeau yes, the self-hosted runner failed to get the secret. I'm still working on it. This is temporary to verify the Testing Farm machine can be successfully reserved.

praveenkumar: @lilyLuLiu In this case `059e594d-12da-4ec6-951f-9069e99cf2c8` one is testing key and doesn't provide any access to the user?

lilyLuLiu: @praveenkumar The token is required to reserve machine from Testing Farm. If the token stored in Github secret, action triggered by PR from forked repo can't access the secret.  Then the PR can't reserve machine successfully.

praveenkumar: > If the token stored in Github secret, action triggered by PR from forked repo can't access the secret. 

I think you need to use workflow like https://github.com/crc-org/crc/blob/main/.github/workflows/windows-qe-tpl.yml one where you can consume the secret from the PR also. 
- https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ have more details

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,118 @@
...
+            -e TARGET_HOST=$(cat host) \
+            -e TARGET_HOST_USERNAME=$(cat username) \
+            -e TARGET_HOST_KEY_PATH=/data/id_rsa \
+            -e TARGET_FOLDER=crc-qe \
+            -e TARGET_RESULTS=results \
+            -e OUTPUT_FOLDER=/data \
+            -e DEBUG=true \
+            -v $PWD/pull-secret:/opt/crc/pull-secret:z \
+            -v $PWD:/data:z \
+            quay.io/crcont/crc-e2e:gh-linux-armd64 \
```
adrianriobo: Typo `quay.io/crcont/crc-e2e:gh-linux-armd64` -> `quay.io/crcont/crc-e2e:gh-linux-arm64`

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,118 @@
...
+            -e TARGET_HOST_USERNAME=$(cat username) \
+            -e TARGET_HOST_KEY_PATH=/data/id_rsa \
+            -e TARGET_FOLDER=crc-support \
+            -e TARGET_CLEANUP='false' \
+            -e OUTPUT_FOLDER=/data \
+            -e DEBUG='true' \
+            -v ${PWD}:/data:z \
+            -v ${PWD}/crc:/opt/crc-support/crc:z \
+            quay.io/rhqp/crc-support:v0.6-linux crc-support/install.sh crc-support/crc
+          podman logs -f crc-linux-install
```
adrianriobo: Can we set to WIP until we clear this usage, also when the new option to install based on assets shared this run will be different

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,118 @@
...
+      
+      - name: Upload e2e results
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: linux-e2e-test
+          path: |
+            **/*.xml
+            **/*.results
+            **/*.log
```
adrianriobo: After reserve, there is no need for decommission the testing farm machine (the target one for arm I mean)

lilyLuLiu:
I need to log in to the self-host runner to do some debug. So need to add my ssh public key into the github self-host runner from the testing farm. Waitting for https://gitlab.com/testing-farm/infrastructure/-/merge_requests/698 to be merged now.

lilyLuLiu:
@adrianriobo Hi, I'm debugging run crc tests with machine from TestingFarm. 

The issue is, we can only use the self-host runner(also from TestingFarm) to provision and connect to the arm64 machine from TestingFarm. That means we can't connect to the machine with a container(deliverset image). 
When provision a TestingFarm machine, the ssh public key of proxy will be added into the runner, thus the runner can connect to the reserved machine through proxy. 

So I think we can't use the container method to run crc tests with Testing Farm. 
Do we move to other test way?

.github/workflows/linux-artifacts.yml
```diff
@@ -0,0 +1,83 @@
...
+          go-version: ${{ matrix.go }}
+      - name: Build Linux installer
+        run: |
+          make out/linux-arm64/crc
+      - name: Upload linux binary
+        uses: actions/upload-artifact@v4
+        with:
+          name: linux-binary
+          path: "out/linux-arm64/crc"
+  build-qe:
```
adrianriobo: I would prefer not having this specific flow. Currently we have the [Build Windows artifacts] where windows bits are built and also the qe images, may we can rename it to just builder and add there the linux builds.

Take into account if Linux bits are already build / check in some other flow remove it from there. The goal should be build only once.... otherwise if in future the build process requires something else on any element we need to be aware they are being built in several places with several goals.

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
+name: tester-linux
+
+on:
+  workflow_run:
+    workflows: [Build Windows artifacts]
```
adrianriobo: THis is aligned with previous comment if you create the unified builder flow this should be triggered for the new builder flow

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+    types:
+      - completed
+
+jobs:
+  tests:
+    runs-on: [self-hosted, linux, testing-farm]
```
adrianriobo: Similar to ci-definitions for crc-builder we want to ensure the previous flow has been completed and it is successful https://github.com/crc-org/ci-definitions/blob/7cb907719aef4898ef4eb0ec02d9124f5212d9d4/.github/workflows/crc-builder-pusher.yml#L12

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        go: ['1.21']
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Go
```
adrianriobo: Why do we need go here? all components should be picked from as artifacts

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+      
+      - name: prepare env
+        run: |
+          sudo yum install podman openssh-server git make -y
+          echo "commit_sha=${commit_sha}" >> "$GITHUB_ENV"
+          mkdir ${{ env.commit_sha }}
+          cd ${{ env.commit_sha }}
+
+          #timestamp=`date +%s`
+          #mkdir $timestamp
```
adrianriobo: Can we clean comments (unused lines)

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+
+      - name: reserve machine from testing farm
+        working-directory: ${{ env.timestamp }}
+        env: 
+          TESTING_FARM_API_TOKEN: ${{ secrets.TESTING_FARM_API_TOKEN }}
+          PULL_SECRET: ${{ secrets.PULL_SECRET }}
+        run: |
+          echo "${PULL_SECRET}" > pull-secret
+          export TESTING_FARM_API_TOKEN=${TESTING_FARM_API_TOKEN}
+          #export TESTING_FARM_API_TOKEN=4a7ba1e1-3881-4bf8-9543-b3521b4fed17
```
adrianriobo, line 50: I guess this token has been reverted. The new one I guess it is already set on upstream crc repo right?

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ matrix.go }}
+      
+      - name: prepare env
+        run: |
+          sudo yum install podman openssh-server git make -y
```
adrianriobo: Do we need git and make for any reason?

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+          #echo "timestamp=$timestamp" >> $GITHUB_ENV
+          $cd $timestamp
+
+          ssh-keygen -t rsa -N '' -f id_rsa -q 
+          eval $(ssh-agent -s)
+          echo $SSH_AUTH_SOCK > ssh_auth_sock
+          echo $SSH_AGENT_PID > ssh_agent_pid
+          #kill $(cat ssh_agent_pid)
+          ssh-add id_rsa
+
```
adrianriobo: We will need api invoke to add this as check for the PR https://github.com/crc-org/crc/blob/0342835858fab4842706eb7caafb9f34bccff3f9/.github/workflows/windows-qe-tpl.yml#L84 this previously need to correlate the execution https://github.com/crc-org/crc/blob/main/.github/workflows/windows-qe-tpl.yml#L61

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+      - name: Run CRC e2e test
+        working-directory: ${{ env.timestamp }}
+        run: |
+          # load image
+          podman load -i crc-e2e-linux-arm64.tar
+          # run
+          cmd="crc-qe/run.sh -junitFilename crc-e2e-junit.xml -targetFolder crc-qe"
+          cmd_microshift="${cmd} -e2eTagExpression '@story_microshift'"
+          cmd_openshift="${cmd} -e2eTagExpression '~@minimal && ~@story_microshift'"
```
adrianriobo: Here you created both microshift and openshift diffferent commands.... but you are only running the openshift .. do we have any restriction on the number of machines we can provision? otherwise may you could create a tpl similar to this https://github.com/crc-org/crc/blob/main/.github/workflows/windows-qe-tpl.yml which would be executed on the self hosted runner the template typically provision a machine install crc run e2e and communicate with gh api to add checks on the PR.

With this approach you can run both in parallel each of them on a machine

.github/workflows/linux-qe.yml
```diff
@@ -0,0 +1,150 @@
...
+        working-directory: ${{ env.timestamp }}
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: linux-e2e-test
+          path: |
+            **/*.xml
+            **/*.results
+            **/*.log
+
```
adrianriobo: It is not required to decommission the provisioned machine? May not but I guess it would be a good practice to reduce cost on their end?

.github/workflows/windows-artifacts.yml
```diff
@@ -40,6 +40,14 @@ jobs:
         with:
           name: windows-installer
           path: "./out/windows-amd64/crc-windows-installer.zip"
+      - name: Build Linux binary
```
adrianriobo, line 43: Can you create a follow up issue on this one, about to renaming this action (as now it also includes linux build, it would be great to properly name it).

Also ensure linux build is only done in one place

.github/workflows/linux-qe-test.yml
```diff
@@ -0,0 +1,29 @@
+name: tester-linux
+
+on:
+  workflow_run:
+    workflows: [Build Windows artifacts]
+    types:
+      - completed
+  
+jobs:
+  linux-e2e-ocp:
```
adrianriobo: why linux-e2e-ocp? may better linux-qe

as it includes integration, e2e and microshift no?

.github/workflows/linux-qe-template.yml
```diff
@@ -0,0 +1,200 @@
...
+            -e BASTION_HOST_USERNAME=$(cat bastion_username) \
+            -e BASTION_HOST=$(cat bastion_host) \
+            -e TARGET_FOLDER=crc-support \
+            -e TARGET_CLEANUP='false' \
+            -e OUTPUT_FOLDER=/data \
+            -e DEBUG='true' \
+            -e SSH_AUTH_SOCK=$(cat ssh_auth_sock) \
+            -v "$(cat ssh_auth_sock):$(cat ssh_auth_sock)" \
+            -v ${PWD}:/data:z \
+            quay.io/rhqp/crc-support:v0.5.1-linux return2testingfarm
```
adrianriobo: What is this?

This is how the target is decommissioned? so you have to run it from inside the target :thinking: 

Tough if this is the reason Can you change the name of the step to make it clear? Or consider adding some comments on it to help others out to better understand the step.

.github/workflows/linux-qe-template.yml
```diff
@@ -0,0 +1,200 @@
...
+
+          commit_sha=$(cat gh_context.json | jq -r '.event.after')
+          if [[ -z "${commit_sha}" ]] || [[ "${commit_sha}" == null ]]; then
+            # on first PR creation .event.after is empty, then .sha is used as commit instead
+            commit_sha=$(cat gh_context.json | jq -r '.event.pull_request.head.sha')
+          fi
+          echo "commit_sha=${commit_sha}" >> "$GITHUB_ENV"
+          mkdir ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+          cd ${{ env.commit_sha }}-${{inputs.qe-type}}-${{inputs.preset}}
+
```
adrianriobo, line 36: May consider add some comments in here to explain why we need the socket strategy on the target..i.e. the target can only be accessed through a bastion (which can only be accessed from self-hosted runner) as so we need to map the ssh-agent from the host to the containers used to access the target host

.github/workflows/linux-qe-template.yml
```diff
@@ -0,0 +1,200 @@
...
+            -e BASTION_HOST=$(cat bastion_host) \
+            -e TARGET_FOLDER=crc-support \
+            -e TARGET_CLEANUP='false' \
+            -e OUTPUT_FOLDER=/data \
+            -e DEBUG='true' \
+            -e SSH_AUTH_SOCK=$(cat ssh_auth_sock) \
+            -v "$(cat ssh_auth_sock):$(cat ssh_auth_sock)" \
+            -v ${PWD}:/data:z \
+            -v ${PWD}/crc:/opt/crc-support/crc:z \
+            quay.io/rhqp/crc-support:v0.5.1-linux crc-support/install.sh crc-support/crc
```
adrianriobo: `quay.io/crc-org/ci-crc-support:v1.0.0-linux` Is now available. If possible change to this as from now on it should be the "official" one.

Let me do a check on using it directly with `crc` as a local asset. And will add here the snippet to make use of it here.

adrianriobo (commented):
In general LGTM although I would prefer to tackle the comments. Although I am also fine if you want to have it in ASAP and the create a follow up one to fix the comments.

.github/workflows/linux-qe-template.yml
```diff
@@ -0,0 +1,202 @@
...
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: linux-${{inputs.qe-type}}-${{inputs.preset}}
+          path: |
+            **/*.xml
+            **/*.results
+            **/*.log
+
+      - name: Update status of the PR check
```
adrianriobo, line 186: Can you add also the initial api call to indicate the check is starting with the pending state i.e. 

https://github.com/crc-org/crc/blob/42afbf10ec4e0f8e0153ebeae8d307613722988e/.github/workflows/windows-qe-tpl.yml#L84

Also can you add the arch on the message 

Finished ${{inputs.qe-type}}-${{inputs.preset}} on Linux -> Finished ${{inputs.qe-type}}-${{inputs.preset}} on Linux ARM64

Beyond this LGTM ( also consider to add some more info on the commit message) like using for the typo of test (linux arm64)...

adrianriobo (approved):

.github/workflows/linux-qe-template.yml
adrianriobo: ^^ @lilyLuLiu please if you can change this before merge

Issue #3615, opened 2023-04-25
[BUG] No bundle type check during `crc setup`


cfergeau:
### General information

  * OS: macOS (should affect all OSes the same)
  * Did you run `crc setup` before starting it (Yes/No)?
  * Running CRC on: Laptop
  * 
## CRC version
```bash
# Put `crc version` output here
Build from git, commit 2351a01dd
```  

### Steps to reproduce

  1. `crc config set preset openshift`
  2. `crc setup --bundle ~/crc_podman_vfkit_4.4.2_amd64.crcbundle`
```
Your system is correctly setup for using CRC. Use 'crc start' to start the instance
```
  3. `crc start --bundle ~/crc_podman_vfkit_4.4.2_amd64.crcbundle`
```
Preset openshift is used but bundle is provided for podman preset
```

### Expected

`crc setup` detects the bundle mismatch

### Actual

`crc setup` is successful, and tells me to run `crc start`

gbraad:
The execution of `crc setup` should fail with an error message that the bundle does not match the configured preset type.

jsliacan:
There's that warning:
```
$ crc config set preset microshift
Changes to configuration property 'preset' are only applied when the CRC instance is created.
If you already have a running CRC instance with different preset, then for this configuration change to take effect, delete the CRC instance with 'crc delete', setup it with `crc setup` and start it with 'crc start'.

$ crc setup -b ~/Downloads/crc_libvirt_4.12.9_amd64.crcbundle 
WARN Using crc_libvirt_4.12.9_amd64.crcbundle bundle, but crc_microshift_libvirt_4.12.9_amd64.crcbundle is expected for this release
.
.
.
```
I'm not suggesting that it's good as it is :), but for comparison `crc start` is successful even if some operators aren't running yet (exit code 0).

redbeam:
Should the warning also be displayed together with the reason of failure? It would look like this:

```
$ crc setup --bundle crc_microshift_libvirt_4.16.7_amd64.crcbundle 
WARN Using crc_microshift_libvirt_4.16.7_amd64.crcbundle bundle, but crc_libvirt_4.16.7_amd64.crcbundle is expected for this release 
FATA Preset openshift is used but bundle is provided for microshift preset
```

The warning is displayed by this function:
https://github.com/crc-org/crc/blob/e6495fe7492eb0464bfb33c55be104272cd16547/pkg/crc/validation/validation.go#L105-L118

[...]

PR #4343, opened 2024-09-03
bundle validation: check for bundle preset mismatch during setup

redbeam:
**Fixes:** Issue #3615 

Added a check for bundle type mismatch with preset in function `ValidateBundle`.

https://github.com/crc-org/crc/blob/7b04ce547692b44e2f12c41ecd07639303473a8f/pkg/crc/validation/validation.go#L91-L107

The check is performed before checking for custom bundle (and potentially displaying a warning).

Function `BundleMismatchWithPreset` was moved from `machine/start.go` to `validation/validation.go` and is used during start as well.

## Testing

1. Download bundle for a preset (i.e. microshift).
2. Change preset to some other value `crc config set preset openshift`
3. `crc setup --bundle <path_to_bundle>`
Output:
`FATA Preset openshift is used but bundle is provided for microshift preset`

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -52,6 +52,14 @@ type ClusterInfo struct {
 	OpenshiftPullSecret string          `json:"openshiftPullSecret,omitempty"`
 }
 
+type FilenameInfo struct {
+	Preset             crcPreset.Preset
+	Driver             string
+	Version            string
+	Arch               string
+	CustomBundleSuffix string
+}
+
 type Node struct {
 	Kind          []string `json:"kind"`
 	Hostname      string   `json:"hostname"`
@@ -242,6 +250,35 @@ func GetBundleNameFromURI(bundleURI string) string {
 		// local path
 		return filepath.Base(bundleURI)
 	}
+}
+
+// GetBundleInfoFromName Parses the bundle filename and returns a FilenameInfo struct
+func GetBundleInfoFromName(bundleName string) (*FilenameInfo, error) {
+	var filenameInfo FilenameInfo
+
+	// crc_preset_driver_version_arch_customSuffix.crcbundle
+	bundleNameRegex := regexp.MustCompile(`crc(?:(?:_)([[:alpha:]]+))?_([[:alpha:]]+)_([0-9.]+)_([[:alnum:]]+)(?:(?:_)([0-9]+))?\.crcbundle`)
+	filenameParts := bundleNameRegex.FindStringSubmatch(bundleName)
+
+	if filenameParts == nil {
+		return &filenameInfo, fmt.Errorf("bundle filename is in unrecognized format")
+	}
+
+	if filenameParts[1] == "" {
+		filenameInfo.Preset = crcPreset.OpenShift
+	} else {
+		parsedPreset, err := crcPreset.ParsePresetE(filenameParts[1])
+		if err != nil {
+			return &filenameInfo, err
+		}
+		filenameInfo.Preset = parsedPreset
+	}
+	filenameInfo.Driver = filenameParts[2]
+	filenameInfo.Version = filenameParts[3]
+	filenameInfo.Arch = filenameParts[4]
+	filenameInfo.CustomBundleSuffix = filenameParts[5]
+
+	return &filenameInfo, nil
 }
 
 func getBundleDownloadInfo(preset crcPreset.Preset) (*download.RemoteFile, error) {
--- a/pkg/crc/validation/validation.go
+++ b/pkg/crc/validation/validation.go
@@ -95,8 +95,28 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 		if bundlePath == constants.GetDefaultBundlePath(preset) {
 			return nil
 		}
-		return ValidateBundlePath(bundlePath, preset)
+
+		if err := ValidateBundlePath(bundlePath, preset); err != nil {
+			return err
+		}
+
+		bundleInfo, err := bundle.GetBundleInfoFromName(bundleName)
+		if err != nil {
+			return err
+		}
+
+		if err := BundleMismatchWithPresetFilename(preset, bundleInfo); err != nil {
+			logging.Fatal(err.Error())
+			return err
+		}
+		return nil
 	}
+
+	if err := BundleMismatchWithPresetMetadata(preset, bundleMetadata); err != nil {
+		logging.Fatal(err.Error())
+		return err
+	}
+
 	bundleMismatchWarning(bundleMetadata.GetBundleName(), preset)
 	/* 'bundle' is already unpacked in ~/.crc/cache */
 	return nil
@@ -115,6 +135,24 @@ func bundleMismatchWarning(userProvidedBundle string, preset crcpreset.Preset) {
 			logging.Warnf("Using %s bundle, but %s is expected for this release", userProvidedBundle, constants.GetDefaultBundle(preset))
 		}
 	}
+}
+
+// BundleMismatchWithPresetFilename checks whether the bundle matches the configured preset based on the bundle filename
+// (bundle is not downloaded or uncompressed yet)
+func BundleMismatchWithPresetFilename(preset crcpreset.Preset, bundleFilenameInfo *bundle.FilenameInfo) error {
+	if preset != bundleFilenameInfo.Preset {
+		return fmt.Errorf("Preset %s is used but bundle is provided for %s preset", preset, bundleFilenameInfo.Preset)
+	}
+	return nil
+}
+
+// BundleMismatchWithPresetMetadata checks whether the bundle matches the configured preset based on the bundle metadata
+// (bundle is already downloaded and uncompressed)
+func BundleMismatchWithPresetMetadata(preset crcpreset.Preset, bundleMetadata *bundle.CrcBundleInfo) error {
+	if preset != bundleMetadata.GetBundleType() {
+		return fmt.Errorf("Preset %s is used but bundle is provided for %s preset", preset, bundleMetadata.GetBundleType())
+	}
+	return nil
 }
 
 // ValidateIPAddress checks if provided IP is valid
--- a/pkg/crc/machine/bundle/metadata_test.go
+++ b/pkg/crc/machine/bundle/metadata_test.go
@@ -211,3 +211,83 @@ func testDataURI(t *testing.T, sha256sum string) string {
 	require.NoError(t, err)
 	return fmt.Sprintf("file://%s", filepath.ToSlash(absPath))
 }
+
+func TestGetBundleInfoFromNameValid(t *testing.T) {
+	valid := [][]string{
+		// crc_preset_driver_version_arch_customSuffix.crcbundle
+		{"crc_libvirt_4.16.7_amd64.crcbundle", preset.OpenShift.String(), "libvirt", "4.16.7", "amd64", ""},
+		{"crc_libvirt_4.16.7_amd64_232.crcbundle", preset.OpenShift.String(), "libvirt", "4.16.7", "amd64", "232"},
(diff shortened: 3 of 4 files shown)


redbeam:
Also, I think `bundleMismatchWarning` function name is kind of ambiguous. It is checking for and displaying a warning if there is a custom bundle being used, so what about renaming it to something like `customBundleUsedWarning` or `userProvidedBundleWarning`?

pkg/crc/validation/validation.go
```diff
@@ -97,6 +97,10 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 		}
 		return ValidateBundlePath(bundlePath, preset)
 	}
+	if err = BundleMismatchWithPreset(preset, bundleMetadata); err != nil {
```
anjannath: while testing its not showing the expected `FATAL` log message when doing `crc setup` with a bundle of different preset as the previous line has `return ValidateBundlePath(bundlePath, preset)`

```
% crc config view
- bundle                                : docker://quay.io/crcont/openshift-bundle:4.16.7

% crc setup --bundle ~/.crc/cache/crc_vfkit_4.16.7_arm64.crcbundle
WARN Using crc_vfkit_4.16.7_arm64.crcbundle bundle, but crc_microshift_vfkit_4.16.7_arm64.crcbundle is expected for this release
INFO Using bundle path /Users/anath/.crc/cache/crc_vfkit_4.16.7_arm64.crcbundle
INFO Check if Podman binary exists in: /Users/anath/.crc/bin/oc
INFO Checking if /Users/anath/.crc/cache/crc_vfkit_4.16.7_arm64.crcbundle exists
INFO Downloading bundle: /Users/anath/.crc/cache/crc_vfkit_4.16.7_arm64.crcbundle...
INFO Uncompressing /Users/anath/.crc/cache/crc_vfkit_4.16.7_arm64.crcbundle
crc.img:  4.71 GiB / 31.00 GiB [----------------------->___________________________________________________________________________________________________________________crc.img:  5.44 GiB / 31.00 GiB [---------------------->____________________________________________________________________________________________________________] 17.54%

```

anjannath: ahh, the above scenario is when the bundle has not been extracted yet, after the bundle is extracted it does show the `FATAL` log:

```
% crc setup --bundle ~/.crc/cache/crc_vfkit_4.16.7_arm64.crcbundle
FATA Preset microshift is used but bundle is provided for openshift preset
```

redbeam: Good catch! I didn't think there would be another code path that is executed when the bundle is uncompressed.
However, I think it's misleading to end `setup` with "your system is correctly set up for crc" and then on the next run fail fatally.
Should we add another bundle check after the decompression is done?

anjannath: when the bundle doesn't exist yet it'd be better to not even download the bundle for the wrong preset, otherwise it feels a bit weird to have it download and uncompress the bundle and later end with a fatal error

so maybe we should have the  `BundleMismatchWithPreset` check in `ValidateBundlePath` as well

redbeam: Correct me if I'm wrong, but before uncompressing it, there is no way to know what preset the bundle is for, no? Therefore the check cannot be done in `ValidateBundlePath`, since this function is called when `bundle.Get(bundleName)` returns `nil`.

cfergeau: The metadata file is stored at the beginning of the bundle so in theory you can uncompress it after fetching a few kilobytes of data, but I don't think we have what's needed in crc to easily do this.

Another option is to do some filename based check (`crc_microshift_libvirt_4.16.0.crcbundle`), but that's quite fragile.

anjannath: @redbeam maybe doing filename based check is the easier/quicker solution for this now, or did you want to look at extracting just the metadata file from the bundle archive?

redbeam: @anjannath yes, that is the path I went on in the end. I pushed the changes now.

pkg/crc/validation/validation.go
```diff
@@ -95,8 +95,28 @@ func ValidateBundle(bundlePath string, preset crcpreset.Preset) error {
 		if bundlePath == constants.GetDefaultBundlePath(preset) {
 			return nil
 		}
-		return ValidateBundlePath(bundlePath, preset)
+
+		if err := ValidateBundlePath(bundlePath, preset); err != nil {
+			return err
+		}
+
+		bundleInfo, err := bundle.GetBundleInfoFromName(bundle.GetBundleNameWithoutExtension(bundleName))
```
anjannath: maybe you can directly pass `bundleName` here, since `GetBundleNameWithoutExtension()` is also called in `GetBundleInfoFromName`?

pkg/crc/validation/validation.go
anjannath: thanks, tested and working as expected!

pkg/crc/validation/validation.go
redbeam: yes, good point. I added the call to GetBundleInfoFromName as a preventive measure.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -244,6 +251,34 @@ func GetBundleNameFromURI(bundleURI string) string {
...
+	case 5:
+		parsedPreset, err := crcPreset.ParsePresetE(filenameParts[1])
+		if err != nil {
+			return &filenameInfo, err
+		}
+		filenameInfo.Preset = parsedPreset
+		filenameInfo.Driver = filenameParts[2]
+		filenameInfo.Version = filenameParts[3]
+		filenameInfo.Arch = filenameParts[4]
+	default:
```
anjannath: hey, apologies i should've mentioned this earlier, but forgot, there's a scenario where the `filenameParts` can be 6, we have a custom bundle generation command `crc bundle generate` that creates new bundle out of existing one and it names them as `crc_microshift_vfkit_4.16.7_arm64_2345.crcbundle` for a microshift bundle and `crc_vfkit_4.16.7_arm64_2345.crcbundle` for an openshift bundle which are valid names; see: https://github.com/crc-org/crc/blob/main/pkg/crc/machine/bundle/metadata.go#L227

so if i use a custom bundle currently this fails, for an openshift bundle as:

```
% crc config set bundle ./crc_vfkit_4.16.7_arm64_2345.crcbundle
WARN Using custom bundle crc_vfkit_4.16.7_arm64_2345.crcbundle 
Successfully configured bundle to ./crc_vfkit_4.16.7_arm64_2345.crcbundle

% crc setup --log-level debug
DEBU CRC version: 2.41.0+1921db
DEBU OpenShift version: 4.16.7
DEBU MicroShift version: 4.16.7
DEBU Running 'crc setup'
DEBU Got bundle path: ./crc_vfkit_4.16.7_arm64_2345.crcbundle
DEBU Failed to parse url: parse "./crc_vfkit_4.16.7_arm64_2345.crcbundle": invalid URI for request
WARN Using custom bundle crc_vfkit_4.16.7_arm64_2345.crcbundle
Cannot parse preset 'vfkit'
```

for a microshift bundle, as:

```
% crc config set bundle ./crc_microshift_vfkit_4.16.7_arm64_232.crcbundle
WARN Using custom bundle crc_microshift_vfkit_4.16.7_arm64_232.crcbundle
Successfully configured bundle to ./crc_microshift_vfkit_4.16.7_arm64_232.crcbundle

% crc setup --log-level debug
DEBU CRC version: 2.41.0+1921db
DEBU OpenShift version: 4.16.7
DEBU MicroShift version: 4.16.7
DEBU Running 'crc setup'
DEBU Got bundle path: ./crc_microshift_vfkit_4.16.7_arm64_232.crcbundle
DEBU Failed to parse url: parse "./crc_microshift_vfkit_4.16.7_arm64_232.crcbundle": invalid URI for request
WARN Using custom bundle crc_microshift_vfkit_4.16.7_arm64_232.crcbundle
bundle filename is in unrecognized format
```

pkg/crc/machine/bundle/metadata.go
```diff
@@ -244,6 +251,34 @@ func GetBundleNameFromURI(bundleURI string) string {
...
 }
 
+// GetBundleInfoFromName Parses the bundle filename and returns a FilenameInfo struct
+func GetBundleInfoFromName(bundleName string) (*FilenameInfo, error) {
+	var filenameInfo FilenameInfo
+	bundleName = GetBundleNameWithoutExtension(bundleName)
+	filenameParts := strings.Split(bundleName, "_")
+
+	switch len(filenameParts) {
+	case 4:
```
cfergeau: Can you add comments with an example bundle name which matches the `case`? First one would be `// crc_libvirt_4.16.7_amd64`

pkg/crc/machine/bundle/metadata.go
cfergeau: Might be easier to parse with a regexp and submatches? 
The regexp would be something like this, `crc(_[[:alpha:]]+)?_([[:alpha:]]+)_([0-9.]+)(_custom)?.crcbundle` (it probably can be improved to be more specific, dunno if it can match directly vfkit or libvirt or hyperv for example).
Then I think https://pkg.go.dev/regexp#Regexp.FindSubmatch will parse the filename for you, and give you the various submatches between parentheses.

cfergeau: I'm surprised a custom bundle has a `_232` suffix as in your example Anjan, I thought they were suffixed with only `_custom`
I've only handled the latter in the regexp above.

redbeam: @anjannath Hmm, good to know. So this is going to be a little more difficult. What is the meaning of the last field (ex. `2345`)? Or how should I name it in the `FilenameInfo` struct?

anjannath: the last field is only present in a user generated custom bundle, it is just to differentiate that it is not the usual bundle we ship, so maybe we can name it as `customBundleSuffix`?

anjannath:
> I'm surprised a custom bundle has a _232 suffix as in your example Anjan, I thought they were suffixed with only _custom

i think this got changed, now they don't contain `_custom` and are in the form of `crc_libvirt_4.16.7_amd64_<unix_timestamp>.crcbundle` as per: https://github.com/crc-org/crc/blob/main/pkg/crc/machine/bundle/metadata.go#L227

cfergeau:
> > I'm surprised a custom bundle has a _232 suffix as in your example Anjan, I thought they were suffixed with only _custom
> 
> i think this got changed, now they don't contain `_custom` and are in the form of `crc_libvirt_4.16.7_amd64_<unix_timestamp>.crcbundle` as per: https://github.com/crc-org/crc/blob/main/pkg/crc/machine/bundle/metadata.go#L227

Ah ok, I was tricked by `git grep _custom` which still returns multiple results.

pkg/crc/machine/bundle/metadata.go
redbeam: Ok, I used that name.

pkg/crc/machine/bundle/metadata.go
redbeam: I used the regex approach, so this is now outdated.

redbeam:
Thanks for the suggestion @cfergeau, I modified your regex a bit and it worked really well.

pkg/crc/machine/bundle/metadata.go
```diff
@@ -244,6 +252,35 @@ func GetBundleNameFromURI(bundleURI string) string {
...
+		filenameInfo.Preset = parsedPreset
+	}
+	filenameInfo.Driver = filenameParts[2]
+	filenameInfo.Version = filenameParts[3]
+	filenameInfo.Arch = filenameParts[4]
+	filenameInfo.CustomBundleSuffix = filenameParts[5]
+
+	return &filenameInfo, nil
+}
+
```
praveenkumar, line 283: We should have unit test for this function which takes all the valid bundle name and provide info what it suppose to do and some dummy one to check the error.

redbeam: I added a unit test for both valid and invalid bundle filenames.

PR #4348, opened 2024-09-04
gitignore: added GoLand's `.idea` folder

redbeam:


.gitignore
```diff
@@ -11,3 +11,5 @@
 /tmp-embed/
 /RPMS/
 /crc
+# GoLand metadata
+/.idea
```
praveenkumar, line 15: We usually don't add editor specific configuration folder to ignore list, this should be part of your global git ignore config.

redbeam: Alright then! Closing this

PR #4349, opened 2024-09-05
Fixing typo in documentation

fhavel:
Fixing typo. Just changing order of commands


fhavel:
I didn't notice I have newer odo installed.

PR #4357, opened 2024-09-16
Remove manual creation of routes-controller manifest

anjannath:
this removes the code marked to be removed after crc 2.24 since the bundles now have the routes controller pod yaml file already

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/rsa"
 	"crypto/x509"
-	"encoding/json"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/crc/machine/start.go
```diff
@@ -828,52 +825,15 @@ func logBundleDate(crcBundleMetadata *bundle.CrcBundleInfo) {
...
-		},
-	})
-	if err != nil {
-		return err
-	}
-	if err := sshRunner.CopyData(bin, "/tmp/routes-controller.json", 0644); err != nil {
-		return err
-	}
-	_, _, err = ocConfig.RunOcCommand("apply", "-f", "/tmp/routes-controller.json")
+	_, _, err := sshRunner.Run("ls", "/opt/crc/routes-controller.yaml")
```
cfergeau, line 831: This could also be dropped, I guess it's checking for the existence of `routes-controller.yaml`. I expect `oc apply` will exit with an error if the file does not exist anyway.

anjannath: yes this can be removed as well, i thought it'd be useful have this check before invoking `oc` but since `oc` would return a similar error its not needed to have this check

PR #4359, opened 2024-09-17
Update go version to 1.22

praveenkumar:


adrianriobo:
~~:thinking: there is no go toolset for ubi images yet no?~~  

NVMD just saw you changed the FROM image

update-go-version.sh
```diff
@@ -14,6 +14,7 @@ go mod edit -go "${golang_base_version}.0" tools/go.mod
 
 sed -i "s,^GOVERSION = 1.[0-9]\+,GOVERSION = ${golang_base_version}," Makefile
 sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-\)1.[0-9]\+,\1${golang_base_version}," images/*/Dockerfile
+sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-\)1.[0-9]\+,\1${golang_base_version}," images/*/Containerfile
```
cfergeau, line 17: Should be enough to add ` images/*/Containerfile` to the previous line?

images/build-e2e/Containerfile
```diff
@@ -1,5 +1,5 @@
 
-FROM registry.access.redhat.com/ubi8/go-toolset:1.21 AS builder
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17 AS builder
```
cfergeau: `ubi8` is publicly accessible while `registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17` requires an auth token. I think this was a problem in the past.

cfergeau (commented):
Adding a rationale explaining why we want to do this switch now would be useful.

images/build-e2e/Containerfile
praveenkumar: yes in the past that was the issue but now this is also doesn't require the auth token and we can have golang-1.22 which is not possible by ubi as of now.

update-go-version.sh
praveenkumar: @cfergeau I think we need both since this update required both in `Dockerfile` and `Containerfile`
```
$ grep -nR  "registry.ci.openshift.org/openshift/release" images/
images/openshift-ci/Dockerfile:4:FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17 AS builder
images/build-e2e/Containerfile:2:FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17 AS builder
images/build-integration/Containerfile:2:FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17 AS builder
images/build/Dockerfile:1:FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17
```

praveenkumar:
> Adding a rationale explaining why we want to do this switch now would be useful.

Added.

images/build-e2e/Containerfile
cfergeau:
```
$ podman pull registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17 
Trying to pull registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17...
Error: initializing source docker://registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17: unable to retrieve auth token: invalid username/password: authentication required
```

(same with the 1.22 image)

update-go-version.sh
cfergeau: I'm suggesting
```
sed -i "s,^\(FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-\)1.[0-9]\+,\1${golang_base_version}," images/*/Dockerfile images/*/Containerfile
```
instead of having 2 lines.

`images/*/{Containerfile, Dockerfile}` should also work.

images/build-e2e/Containerfile
adrianriobo: weird this should not need auth, actually it is being checked here and it is passing:

https://github.com/crc-org/crc/actions/runs/10902596500/job/30254853110?pr=4359#step:3:27

praveenkumar: @cfergeau I created a fresh RHEL-9 VM and tried to pull those images which succeeded 
```
$ cat /tmp/auth.json 
{}
$  podman pull --authfile=/tmp/auth.json registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17 
Trying to pull registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17...
Copying blob 465c5f11b934 done   | 
Copying blob f70d60810c69 skipped: already exists  
Copying blob 192c4240df71 done   | 
Copying blob 261dc53a0467 done   | 
Copying blob 545277d80005 skipped: already exists  
Copying config 5c5232f30d done   | 
5c5232f30d1b18a3c665aef04e9a62023006c605b086dd8e50b02c6f0c89c8f5
```

I think what is happening for you is you have existing token for this registry which is expired and that's why you are getting this error. To get same error I added following in the auth file and got same error.
```
$ cat /tmp/auth.json 
{
  "auths": {
   "registry.ci.openshift.org": {
	  "auth": "cHJrdW1hcjpzaGEyNTZ+eEIzeWFDam5jZ0Z5UGcwbUU3cnEtU4NWX0NoSUFtTEJZb0xaSk9TN1VMbw=="
   }
  }
}
$ podman pull --authfile=/tmp/auth.json registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17 
Trying to pull registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17...
Error: initializing source docker://registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.21-openshift-4.17: unable to retrieve auth token: invalid username/password: authentication required
```

cfergeau: > I think what is happening for you is you have existing token for this registry which is expired and that's why you are getting this error. 

Good catch, I removed `.docker/config.json` and I can pull with no error!

update-go-version.sh
cfergeau: I see we also duplicate the `sed` regexp for `ubi8/go-toolset` below, we can keep your version and clean that up later. 
Do we still need the `ubi8` updates after the switch to the openshift CI images?

cfergeau (approved):

update-go-version.sh
praveenkumar: @cfergeau no, we don't need ubi8 update now so removed it in different commit.

PR #4360, opened 2024-09-20
Update golangci t0 1.61.0

praveenkumar:


update-go-version.sh
```diff
@@ -10,7 +10,7 @@ golang_base_version=$1
 echo "Updating golang version to $golang_base_version"
 
 go mod edit -go "${golang_base_version}.0"
-go mod edit -go "${golang_base_version}.0" tools/go.mod
+go mod edit -go "${golang_base_version}.1" tools/go.mod
```
cfergeau, line 13: ~~better to use `"${golang_base_version}.*"` or `"${golang_base_version}.[0-9]+"` so that we don't keep updating it for each minor release.~~

I should read the full command, not just its arg :), ignore this

cfergeau (commented):
This introduces new linting failures.

pkg/os/shell/shell_windows.go
```diff
@@ -30,6 +30,7 @@ func getProcessEntry(pid int) (pe *syscall.ProcessEntry32, err error) {
 	}
 
 	for {
+		//#nosec G115 -- pid is coming from os.Getppid() as int
```
cfergeau: the `os.Getppid()` call is 2 level above `getProcessEntry` in the call chain. Nothing guarantees that in the future we are not going to reuse this function with a `pid` value coming from something different.
I'd prefer to change the `getProcessEntry(pid int)` to `getProcessEntry(pid uint32)`, and to change `detect` to convert the value returned by `os.Getppid()` to `uint32`

pkg/crc/machine/start.go
```diff
@@ -265,8 +265,8 @@ func configureSharedDirs(vm *virtualMachine, sshRunner *crcssh.Runner) error {
 
 func (client *client) Start(ctx context.Context, startConfig types.StartConfig) (*types.StartResult, error) {
 	telemetry.SetCPUs(ctx, startConfig.CPUs)
-	telemetry.SetMemory(ctx, uint64(startConfig.Memory)*1024*1024)
-	telemetry.SetDiskSize(ctx, uint64(startConfig.DiskSize)*1024*1024*1024)
+	telemetry.SetMemory(ctx, uint64(startConfig.Memory)*1024*1024)          //#nosec G115 -- validation make sure memory value is not negative
```
cfergeau: actually I haven't been able to find such >=0 checks in the  validation code?

pkg/drivers/hyperv/hyperv_windows.go
```diff
@@ -74,6 +74,7 @@ func (d *Driver) UpdateConfigRaw(rawConfig []byte) error {
 		}
 	}
 	if newDriver.DiskCapacity != d.DiskCapacity {
+		//#nosec G115 -- validation make sure disk value is in range
```
cfergeau: I'd prefer if code in `pkg/drivers` did not make assumptions regarding how it's used, and in particular that we don't rely on code in `pkg/crc` having to be used before passing data to the drivers.
Moreover, I'm not sure we have the needed range check in validation

pkg/crc/machine/start.go
praveenkumar: for memory as part of validation we use the base memory as default preset one (in case of OCP it is 10752 and for microshift 4096

```go
// ValidateMemory checks if provided Memory count is valid
func ValidateMemory(value int, preset crcpreset.Preset) error {
	if value < constants.GetDefaultMemory(preset) {
		return fmt.Errorf("requires memory in MiB >= %d", constants.GetDefaultMemory(preset))
	}
	return ValidateEnoughMemory(value)
```

praveenkumar:
@cfergeau Can you take another look to this PR, it have machine update and also golint ci update.

pkg/crc/config/config.go
```diff
@@ -207,6 +212,13 @@ func (c *Config) Get(key string) SettingValue {
 				Invalid: true,
 			}
 		}
+	case uint:
+		value, err = cast.ToUintE(value)
+		if err != nil {
+			return SettingValue{
+				Invalid: true,
+			}
+		}
```
cfergeau, line 221: Unrelated to this PR, but wondering if we could use go generics to have a generic 'cast' function instead of switching over the types.

cmd/crc/cmd/start.go
```diff
@@ -39,7 +39,7 @@ func init() {
 	flagSet.StringP(crcConfig.PullSecretFile, "p", "", fmt.Sprintf("File path of image pull secret (download from %s)", constants.CrcLandingPageURL))
 	flagSet.UintP(crcConfig.CPUs, "c", constants.GetDefaultCPUs(crcConfig.GetPreset(config)), "Number of CPU cores to allocate to the instance")
 	flagSet.UintP(crcConfig.Memory, "m", constants.GetDefaultMemory(crcConfig.GetPreset(config)), "MiB of memory to allocate to the instance")
-	flagSet.UintP(crcConfig.DiskSize, "d", constants.DefaultDiskSize, "Total size in GiB of the disk used by the instance")
+	flagSet.Uint64P(crcConfig.DiskSize, "d", constants.DefaultDiskSize, "Total size in GiB of the disk used by the instance")
```
cfergeau: I'd prefer to keep DiskSize as uint as long as it's holding GiBs. When we convert it to bytes, then it makes sense to use an uint64. I don't understand how all the changes in this commit help to fix unspecified "integer overflow issue".

cmd/crc/cmd/start.go
```diff
@@ -70,7 +70,7 @@ func runStart(ctx context.Context) (*types.StartResult, error) {
 	startConfig := types.StartConfig{
 		BundlePath:        config.Get(crcConfig.Bundle).AsString(),
 		Memory:            config.Get(crcConfig.Memory).AsUInt(),
-		DiskSize:          config.Get(crcConfig.DiskSize).AsInt(),
+		DiskSize:          config.Get(crcConfig.DiskSize).AsUInt64(),
```
cfergeau: I guess `AsUint` would be enough here?

pkg/crc/machine/config/config.go
```diff
@@ -10,7 +10,7 @@ type MachineConfig struct {
 	Name              string
 	Memory            uint
 	CPUs              uint
-	DiskSize          int
+	DiskSize          uint64
```
cfergeau: Can you add a comment specifying the disk size unit? I think it's GiB, in which case `uint` is enough.
You can also add a similar comment to `Memory` in the corresponding commit.

tools/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22.1
 
 require (
 	github.com/cfergeau/gomod2rpmdeps v0.0.0-20210223144124-2042c4850ca8
-	github.com/golangci/golangci-lint v1.59.1
+	github.com/golangci/golangci-lint v1.61.0
```
cfergeau, line 7: I would keep this commit until the very end, it's odd to have some commits fixing overflows before this one, and then more commits fixing issues after it.

pkg/os/shell/shell_windows.go
```diff
@@ -43,21 +43,22 @@ func getProcessEntry(pid int) (pe *syscall.ProcessEntry32, err error) {
...
-	return name, int(pe.ParentProcessID), nil
+	return name, pe.ParentProcessID, nil
 }
 
 func detect() (string, error) {
 	shell := os.Getenv("SHELL")
 
 	if shell == "" {
-		shell, shellppid, err := getNameAndItsPpid(os.Getppid())
+		//#nosec G115 -- pid is coming from os.Getppid() as int
```
cfergeau: Why is this ok to ignore the overflow?
I'd expect some range check before the cast
```
if pid < 0 || pid > math.MaxUint32 {
  // error
}
```

pkg/os/shell/shell_windows_test.go
```diff
@@ -18,15 +18,18 @@ func TestDetect(t *testing.T) {
...
 	assert.Equal(t, "shell.test.exe", shell)
-	assert.Equal(t, os.Getppid(), shellppid)
+	//#nosec G115 -- pid is coming from os.Getppid() as int
+	assert.Equal(t, uint32(os.Getppid()), shellppid)
 	assert.NoError(t, err)
 }
 
 func TestGetNameAndItsPpidOfParent(t *testing.T) {
-	shell, _, err := getNameAndItsPpid(os.Getppid())
+	//#nosec G115 -- pid is coming from os.Getppid() as int
```
cfergeau: Same comment as before for these 3 ignores

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -97,20 +97,21 @@ func (d *Driver) getDiskPath() string {
 	return d.ResolveStorePath(fmt.Sprintf("%s.img", d.MachineName))
 }
 
-func (d *Driver) resize(newSize int64) error {
+func (d *Driver) resize(newSize uint64) error {
+	newSizeInt64 := int64(newSize) //#nosec G115 -- os.stat provide the size as int64 and os.Truncate use int64 as arg
```
cfergeau: This explains why there's a warning, but this should be explaining why it's ok to ignore it. This needs range checks before doing the cast.

pkg/crc/machine/kubeconfig.go
```diff
@@ -186,7 +186,7 @@ func getTokenForUser(username, password, ip string, ca []byte, clusterConfig *ty
 					return nil, err
 				}
 				if strings.HasSuffix(hostname, constants.AppsDomain) {
-					port = strconv.Itoa(int(ingressHTTPSPort))
+					port = strconv.FormatUint(uint64(ingressHTTPSPort), 10)
```
cfergeau, line 189: Or `fmt.Sprintf("%d", ingressHTTPSPort)`

pkg/os/shell/shell_windows.go
praveenkumar: Do you think `oc.Getppid` return something out of range of int? Shouldn't than it would be checked on os module  level?

cfergeau: If these are the assumptions you want to make to ignore this warning, add them to the `nosec` comment. Imo it does not hurt to have checks to make it very clear.

pkg/crc/config/config.go
praveenkumar: Right, that would be a good exercise, we can put a tracker issue for it.

cmd/crc/cmd/start.go
praveenkumar: Done, switched back to `uint`

cmd/crc/cmd/start.go
praveenkumar: switched back to Uint

pkg/crc/machine/config/config.go
praveenkumar: Done

tools/go.mod
praveenkumar: Done, push that to end

pkg/os/shell/shell_windows.go
praveenkumar: Added range check

pkg/crc/machine/kubeconfig.go
praveenkumar: yeah, but this is also works so let's go with that.

pkg/crc/api/api_client_test.go
```diff
@@ -195,6 +195,8 @@ func TestConfigGetAll(t *testing.T) {
 		switch v := v.Value.(type) {
 		case int:
 			configs[k] = float64(v)
+		case uint:
+			configs[k] = float64(v)
```
cfergeau, line 199: Nit, probably could be:
```
case int, uint:
	configs[k] = float64(v)
```

pkg/os/shell/shell_windows.go
```diff
@@ -43,21 +45,26 @@ func getProcessEntry(pid int) (pe *syscall.ProcessEntry32, err error) {
...
 func detect() (string, error) {
 	shell := os.Getenv("SHELL")
 
 	if shell == "" {
-		shell, shellppid, err := getNameAndItsPpid(os.Getppid())
+		pid := os.Getppid()
+		if pid < 0 || pid > math.MaxUint32 {
+			return "", fmt.Errorf("integer overflow for pid: %v", pid)
+		}
+		//#nosec G115 -- Above pid is checked for integer overflow and an error is returned
```
cfergeau: With the range check above, the #nosec directive is not needed.

pkg/os/shell/shell_windows_test.go
```diff
@@ -18,15 +19,24 @@ func TestDetect(t *testing.T) {
 }
 
 func TestGetNameAndItsPpidOfCurrent(t *testing.T) {
-	shell, shellppid, err := getNameAndItsPpid(os.Getpid())
-
+	pid := os.Getppid()
+	if pid < 0 || pid > math.MaxUint32 {
+		assert.Fail(t, "integer overflow detected")
+	}
+	//#nosec G115 -- Above pid is checked for integer overflow and fail the test
```
cfergeau: I did not test but I'd guess this one is not needed either.

cfergeau: Can be 
```
assert.Positive(t, pid)
assert.Less(t, pid, math.MaxUint32)
```

pkg/os/shell/shell_windows_test.go
```diff
@@ -18,15 +19,24 @@ func TestDetect(t *testing.T) {
...
 	assert.NoError(t, err)
 }
 
 func TestGetNameAndItsPpidOfParent(t *testing.T) {
-	shell, _, err := getNameAndItsPpid(os.Getppid())
+	pid := os.Getppid()
+	if pid < 0 || pid > math.MaxUint32 {
+		assert.Fail(t, "integer overflow detected")
+	}
+	//#nosec G115 -- Above pid is checked for integer overflow and fail the test
```
cfergeau: And same here

pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -97,20 +98,25 @@ func (d *Driver) getDiskPath() string {
 	return d.ResolveStorePath(fmt.Sprintf("%s.img", d.MachineName))
 }
 
-func (d *Driver) resize(newSize int64) error {
+func (d *Driver) resize(newSize uint64) error {
+	if newSize > math.MaxInt64 {
+		return fmt.Errorf("integer overflow detected for resize: %v", newSize)
+	}
+	//#nosec G115 -- os.stat provide the size as int64 and os.Truncate use int64 as arg. Also range is checked (above)
```
cfergeau: `#nosec G115` is not needed since you added the range check. You can keep the rest of the comment if you think it's useful.

pkg/crc/api/api_client_test.go
praveenkumar: That doesn't work https://github.com/golang/go/issues/36616

pkg/os/shell/shell_windows_test.go
praveenkumar: > Can be

if we do this then linter again show same error
```
pkg/os/shell/shell_windows_test.go:25:51: G115: integer overflow conversion int -> uint32 (gosec)
	shell, shellppid, err := getNameAndItsPpid(uint32(pid))
	                                                 ^
pkg/os/shell/shell_windows_test.go:27:24: G115: integer overflow conversion int -> uint32 (gosec)
	assert.Equal(t, uint32(pid), shellppid)
	                      ^
pkg/os/shell/shell_windows_test.go:35:43: G115: integer overflow conversion int -> uint32 (gosec)
	shell, _, err := getNameAndItsPpid(uint32(pid))
	                                         ^
```

cfergeau: Ah ok, need to be improved in gosec then.

cfergeau (approved):

praveenkumar:
/hold 

`make check` on windows failing, need to check what's happening.

PR #4369, opened 2024-09-24
Upload qe images when code merged to main

lilyLuLiu:
https://github.com/crc-org/crc/issues/4368

### DIFF
--- a/.github/workflows/windows-artifacts.yml
+++ b/.github/workflows/windows-artifacts.yml
@@ -50,8 +50,6 @@ jobs:
         exclude:
           - arch: 'arm64'
             os: 'windows'
-          - arch: 'arm64'
-            os: 'linux'
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


adrianriobo:
Can we extend https://github.com/crc-org/crc/blob/main/.github/workflows/windows-artifacts.yml#L58 as that is executed on push to main as well.

So we do not build images several times on push.

In this case you just add an extra step with condition on push event and re-tag images to `next-OS-ARCH` or `dev-OS-ARCH` just take into account:

https://gitlab.cee.redhat.com/crc/qe-platform/-/blob/master/catalog/task/crc-e2e/crc-e2e.yaml?ref_type=heads#L97

the task is composing the version with `v` as so it would expect i.e. `vnext-OS-ARCH`

As we also need to plan move `crc-e2e` task to `ci-defintions` we can re think on that v composition...meaning you can leave the tag as i.e. `next-OS-ARCH`, then we would need to adapt the version value accordingly.

WDYT?

adrianriobo (commented):
Can we extend https://github.com/crc-org/crc/blob/main/.github/workflows/windows-artifacts.yml#L58 as that is executed on push to main as well.

So we do not build images several times on push.

In this case you just add an extra step with condition on push event and re-tag images to next-OS-ARCH or dev-OS-ARCH just take into account:

https://gitlab.cee.redhat.com/crc/qe-platform/-/blob/master/catalog/task/crc-e2e/crc-e2e.yaml?ref_type=heads#L97

the task is composing the version with v as so it would expect i.e. vnext-OS-ARCH

As we also need to plan move crc-e2e task to ci-defintions we can re think on that v composition...meaning you can leave the tag as i.e. next-OS-ARCH, then we would need to adapt the version value accordingly.

WDYT?

lilyLuLiu:
@adrianriobo Hi, For the windows artifacts action(https://github.com/crc-org/crc/blob/main/.github/workflows/windows-artifacts.yml), I don't get your suggestion: 
In the windows action, the qe images are built and upload to GitHub action artifacts, instead of quay.  And the triggered windows test actions use the qe images from the uploaded artifacts, which requires the workflow id. So the two actions are different. Or you mean we can use quay qe images for windows github action tests?

lilyLuLiu:
@adrianriobo I have added V in the tag, like `vNext-OS-ARCH`. We can remove the v after migration to ci-definition.

adrianriobo (approved):
Can you change a bit the commit message, something like:

"[QE] Push qe oci image to quay on every merge to main.

This commit will push qe oci images to quay tagged as next, so we can use the latest changes on e2e and integration code to be used across all our CI systems"

In any case LGTM to me

PR #4379, opened 2024-09-26
[QE]fix short of registry name in windows-artifact action

lilyLuLiu:
**Relates to:** PR #4369


adrianriobo (approved):
Can you add a ref on the commit to the commit id where the typo was introduced?

Otherwise LGTM

Issue #4381, opened 2024-09-27
[QE] mark some test only for releases


albfan:
Some nightly builds do not match with some test (like CRC version)

Mark them for release "release-testing"

PR #4383, opened 2024-10-01
mark some tests only for releases

albfan:
**Fixes:** Issue #4381

### DIFF
--- a/test/e2e/features/basic.feature
+++ b/test/e2e/features/basic.feature
@@ -4,7 +4,7 @@ Feature: Basic test
     User explores some of the top-level CRC commands while going
     through the lifecycle of CRC.
 
-    @darwin @linux @windows
+    @darwin @linux @windows @release
     Scenario: CRC version
         When executing crc version command
         Then stdout should contain correct version


adrianriobo:
Can you change a bit the commit message, i.e. 

```
[QE] Added a new e2e annotation to identify test to be executed only during release testing

During nightly e2e the crc version test is giving a false positive due to how we check the versions for the code tested and the version of the tests themselves. This version test makes no sense during nightly but it make sense during release testing. This commit will include an annotation to allow us pick the test only during release testing
```

adrianriobo (commented):
Can you change a bit the commit message, i.e. 

```
[QE] Added a new e2e annotation to identify test to be executed only during release testing

During nightly e2e the crc version test is giving a false positive due to how we check the versions for the code tested and the version of the tests themselves. This version test makes no sense during nightly but it make sense during release testing. This commit will include an annotation to allow us pick the test only during release testing
```

albfan:
@adrianriobo message changed

adrianriobo (approved):

PR #4376, opened 2024-09-26
fix e2e result junit format

lilyLuLiu:
**Fixes:** Issue #4373

Issue #4373, opened 2024-09-25
[QE] No details of failed e2e test cases in Reporportal


lilyLuLiu:
The junit xml file upload to reportportal. But can't check the failed test cases detail

albfan:
Resolved limiting size of logs and configuring reportportal instance

PR #4386, opened 2024-10-03
fix junit format for reports

albfan:
**Fixes:** Issue #4373

**Relates to:** PR #4376

## Solution/Idea

adapt junit report reportportal format

## Proposed changes

post process xml junit to adapt xml to reportportal format

### DIFF
--- a/images/build-integration/Containerfile
+++ b/images/build-integration/Containerfile
@@ -21,5 +21,6 @@ ENV ASSETS_FOLDER=/opt/crc \
     ARCH=${ARCH}
 
 COPY --from=builder /workspace/out/${OS}-${ARCH} ${ASSETS_FOLDER}/bin
+COPY images/build-integration/common/* ${ASSETS_FOLDER}/
 COPY images/build-integration/lib/${OS}/* ${ASSETS_FOLDER}/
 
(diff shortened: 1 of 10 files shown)


albfan:
**NOTE** Windows still not tested

images/build-e2e/lib/darwin/filter.xsl
```diff
@@ -0,0 +1,19 @@
+<xsl:stylesheet version="1.0"
```
adrianriobo: Are those xsls files exactly the same? If that is the case we need to have it only once and add it from the Containerfile

adrianriobo (commented):
Also need to understand the context and the cause of this change on the commit log. (Also if possible prefix the commit log with [QE])

images/build-e2e/lib/darwin/filter.xsl
albfan: Ok, I was rounding the idea. Will rework the PR

albfan:
Doing test on reportportal, only content on failure tag is use to fill  stacktrace and logs

Our reason on a failure test is on message attribute from failure node, so this transformation remove message attribute and set as failure node content

images/build-integration/Containerfile
```diff
@@ -21,5 +21,6 @@ ENV ASSETS_FOLDER=/opt/crc \
     ARCH=${ARCH}
 
 COPY --from=builder /workspace/out/${OS}-${ARCH} ${ASSETS_FOLDER}/bin
+COPY images/build-e2e/common/* ${ASSETS_FOLDER}/
```
adrianriobo: This is using the wrong path no?

It should be `images/build-integration/common/*`

albfan: ops

adrianriobo (approved):

PR #4387, opened 2024-10-03
Update openshift to 4.17.0 and microshift to 4.16.14

praveenkumar:


anjannath:
certificate rotation tests are failing which is a known failure in openshift ci

ref: https://github.com/crc-org/snc/issues/949

anjannath (approved):

PR #4401, opened 2024-10-15
shell_windows: Detect shell under ssh

### LINKED ISSUE #3582: [BUG] Powershell core support on microshift preset start hints
Recently we added support for powershell core https://github.com/crc-org/crc/pull/3500, with that PR most crc commands detect powershell core and show the right syntax.

With microshift preset at the end of the `crc start` some commands are shown to help setting the client to interact with the cluster (those commands do not recognize powershell core) so it is showing an incorrect syntax for it:

```
DEBU SSH command results: err: <nil>, output: pod/routes-controller created 
Started the MicroShift cluster.

export KUBECONFIG="/Users/ec2-user/.crc/machines/crc/kubeconfig"

  $ eval $(crc oc-env)
  $ oc COMMAND
```

albfan:
**Fixes:** Issue #3582

## Solution/Idea

`$env:SHELL` not empty is not definitive to detect shell terminal

## Proposed changes

Reuse our logic to detect powershell if current result is not definitive.

## Testing

Use `crc oc-env` under ssh on windows

fixes #3582

### DIFF
--- a/pkg/os/shell/shell_windows.go
+++ b/pkg/os/shell/shell_windows.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"math"
 	"os"
-	"path/filepath"
 	"strings"
 	"syscall"
 	"unsafe"
@@ -55,6 +54,19 @@ func getNameAndItsPpid(pid uint32) (exefile string, parentid uint32, err error) 
 	return name, pe.ParentProcessID, nil
 }
 
+func shellType(shell string, defaultShell string) string {
+	switch {
+	case strings.Contains(strings.ToLower(shell), "powershell"):
+		return "powershell"
+	case strings.Contains(strings.ToLower(shell), "pwsh"):
+		return "powershell"
+	case strings.Contains(strings.ToLower(shell), "cmd"):
+		return "cmd"
+	default:
+		return defaultShell
+	}
+}
+
 func detect() (string, error) {
 	shell := os.Getenv("SHELL")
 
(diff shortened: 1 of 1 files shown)


pkg/os/shell/shell_windows.go
```diff
@@ -58,6 +58,19 @@ func getNameAndItsPpid(pid uint32) (exefile string, parentid uint32, err error)
...
+
+	if baseShell != "powershell" && baseShell != "cmd" {
+		switch {
+		case strings.Contains(strings.ToLower(shell), "powershell"):
+			return "powershell", nil
+		case strings.Contains(strings.ToLower(shell), "pwsh"):
+			return "powershell", nil
+		case strings.Contains(strings.ToLower(shell), "cmd"):
+			return "cmd", nil
+		}
```
praveenkumar: This is duplication of https://github.com/crc-org/crc/pull/4401/files#diff-ccde05d43b105715d5a716447bc74186b326dc1d90d98091e780bb407da6fd3fR84-R89 better to have a helper function which take the `shell` as argument and return type.

Also this logic should be at the end instead of start.

pkg/os/shell/shell_windows.go
```diff
@@ -67,32 +91,18 @@ func detect() (string, error) {
...
 
 	if os.Getenv("__fish_bin_dir") != "" {
 		return "fish", nil
 	}
 
-	return filepath.Base(shell), nil
+	baseShell := filepath.Base(shell)
+
+	if baseShell != "powershell" && baseShell != "cmd" {
+		return analyzeShell(shell, uint32(os.Getpid()))
```
cfergeau: You will need something similar to the code at the beginning of `if shell == ""` to avoid the lint failures:
```
		pid := os.Getppid()
		if pid < 0 || pid > math.MaxUint32 {
			return "", fmt.Errorf("integer overflow for pid: %v", pid)
		}
```

~~If we need `pid` in multiple places, we can get it at the beginning of  this function.~~ (deleted, pid VS ppid confusion)

pkg/os/shell/shell_windows.go
```diff
@@ -67,32 +91,18 @@ func detect() (string, error) {
...
 	}
 
 	if os.Getenv("__fish_bin_dir") != "" {
 		return "fish", nil
 	}
 
-	return filepath.Base(shell), nil
+	baseShell := filepath.Base(shell)
+
+	if baseShell != "powershell" && baseShell != "cmd" {
```
cfergeau: At this point, `shell` was non-empty, but it was neither `powershell` nor `cmd`, so we need to do some matching on the name of the parent process? Some explanations in the commit log about what the expectations are could be useful.
`analyzeShell` does the right thing if `shell` is `powershell` or `cmd`, so you could just replace `return filepath.Base(shell), nil` in the initial code with `return analyzeShell(shell, pid)`, no need for `baseShell` I think.

pkg/os/shell/shell_windows.go
albfan: Yes, thanks for review

pkg/os/shell/shell_windows.go
albfan: Looks clear, yes

praveenkumar:
Check https://github.com/praveenkumar/crc/commit/1efb9fa52cf85a56246e5489fc21851715464d91 one I think it would be more clear and we don't have to pass the process pid to helper function.

albfan:
Many thanks for review Praveen.

Just to avoid go into further check, I added a default result to identify non detected shell

praveenkumar:
Looks like CI is failing because of `fmt` error  run `make fmt` .

albfan:
@praveenkumar should be fixed now. I find with goimports but not sure where CI show that error

PR #4403, opened 2024-10-16
[QE] limit log size for test reports

albfan:
**Fixes:** Issue #4373 

## Solution/Idea

Reports with long messages are hanging reportportal. Limit its size

## Proposed changes

after post process xml report, limit log size

## Testing

Tested on log reports long and short than max size

### DIFF
--- a/images/build-e2e/common/filter.xsl
+++ b/images/build-e2e/common/filter.xsl
@@ -1,19 +1,30 @@
-<xsl:stylesheet version="1.0"
-xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
 <xsl:output method="xml" version="1.0" encoding="UTF-8" indent="yes"/>
 <xsl:strip-space elements="*"/>
 
-<!-- identity transform -->
 <xsl:template match="@*|node()">
     <xsl:copy>
-            <xsl:apply-templates select="@*|node()"/>
+        <xsl:apply-templates select="@*|node()"/>
     </xsl:copy>
 </xsl:template>
 
 <xsl:template match="failure">
-    <xsl:copy>
-        <xsl:value-of select="@message"/>
-    </xsl:copy>
+    <xsl:variable name="max" select="1000000" />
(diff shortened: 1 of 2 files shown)


images/build-e2e/common/filter.xsl
```diff
@@ -6,13 +6,13 @@ xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
...
     <xsl:copy>
-            <xsl:apply-templates select="@*|node()"/>
+        <xsl:apply-templates select="@*|node()"/>
     </xsl:copy>
 </xsl:template>
 
 <xsl:template match="failure">
     <xsl:copy>
-        <xsl:value-of select="@message"/>
+        <xsl:value-of select="substring(@message,0,1000000)"/>
```
adrianriobo: Is it possible to pick the latest chars from the failure message.. It would have more sense as typically you see the error at the very end

gbraad: if the error is at the end...

is there a way to link to the full report or a compressed form?

adrianriobo: yeah, actually seems a combination of both: summary for the error at the beginning ...full trace with logs around the point of the error to the end of the trace.

adrianriobo: So this made me think, the problem is about adding too much data on the junit files right? 

Because it seems there is another option to import full logs next to results.... (on report portal)

If this is the case may we should see how to avoid the error log to be added on the junit file, for ginkgo AFAIK there are some customization you can do on the reporter to avoid the full trace... on the e2e side I think it is us adding the full error in code.

Can you check this @albfan

albfan: While not by API, I try that solution from xml. tag `system-out` and `system-err` (under tag `test-case`) import logs.

they broke the web reportportal the same if they are big enough.

We ask our provider if there's a timeout on loading log view on reportportal.

Meanwhile we get a reply,  I will extract half the beginning and the end of message

albfan: filter should now truncate in the middle of log showing first 50000 chars and last 500000 chars

adrianriobo (approved):

Issue #3998, opened 2024-01-30
[BUG] bundle.Download() should have a context.Context argument


cfergeau:
It's not possible to programmatically cancel bundle downloads. After using `start` with the REST API, if the bundle has to be downloaded, then it's not possible to use `stop` while the bundle is being downloaded.

cfergeau:
Reproducer:
On a system where the bundle is not downloaded/unpacked:
```
$ curl --unix-socket /var/home/teuf/.crc/crc-http.sock  http://unix/api/start
# the start command hangs waiting for the bundle to be downloaded

# in another terminal
$ curl --unix-socket /var/home/teuf/.crc/crc-http.sock  http://unix/api/stop
cannot abort startup sequence quickly enough
```

It would be nice if, when we call `stop` while a bundle is being downloaded/uncompressed, it's still possible to call `stop`, and cancel the download/uncompression.
Hopefully we can use the `context.Context` passed to `pkg/crc/machine.Start` to cancel all of this.

rohanKanojia:
@redbeam @cfergeau : Hello, is this issue fixed by #4405

cfergeau:
Yes, thanks for spotting this.

PR #4405, opened 2024-10-18
bundle download & decompression: added Context argument

redbeam:
**Fixes:** Issue #3998 

## Solution/Idea

Added a `context.Context` argument to functions such as `bundle.Download` and `bundle.Extract` and others, to allow them to gracefully stop (if they are running) when the `stop` command is issued through the HTTP API.

Context support was added to regular HTTP bundle downloads as well as bundle pulls from container registries (it was supported by the used libraries). However, for the bundle decompression, this had to be done manually, since there is no library used.

If there is no context available to be used, `nil` or `context.Background()` can be used.

## Testing

```
$ curl --unix-socket ~/.crc/crc-http.sock  http://unix/api/start
# the start command hangs waiting for the bundle to be downloaded/decompressed

# in another terminal
$ curl --unix-socket ~/.crc/crc-http.sock  http://unix/api/stop
Instance is already stopped
```

## Caveat

The `stop` command ends with `Instance is already stopped` error (error code 500 is returned) since it is intended to stop the machine, which is (at that moment) not running. But it stops the download/decompression, which is desired.

### DIFF
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -1,6 +1,7 @@
 package download
 
 import (
+	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
@@ -60,7 +61,7 @@ loop:
 
 // Download function takes sha256sum as hex decoded byte
 // something like hex.DecodeString("33daf4c03f86120fdfdc66bddf6bfff4661c7ca11c5d")
-func Download(uri, destination string, mode os.FileMode, sha256sum []byte) (string, error) {
+func Download(ctx context.Context, uri, destination string, mode os.FileMode, sha256sum []byte) (string, error) {
 	logging.Debugf("Downloading %s to %s", uri, destination)
 
 	client := grab.NewClient()
@@ -70,6 +71,12 @@ func Download(uri, destination string, mode os.FileMode, sha256sum []byte) (stri
 	if err != nil {
 		return "", errors.Wrapf(err, "unable to get request from %s", uri)
 	}
+
+	if ctx == nil {
+		panic("ctx is nil, this should not happen")
+	}
+	req = req.WithContext(ctx)
+
 	if sha256sum != nil {
 		req.SetChecksum(sha256.New(), sha256sum, true)
 	}
@@ -131,12 +138,12 @@ func NewRemoteFile(uri, sha256sum string) *RemoteFile {
 
 }
 
-func (r *RemoteFile) Download(bundlePath string, mode os.FileMode) (string, error) {
-	sha256, err := hex.DecodeString(r.sha256sum)
+func (r *RemoteFile) Download(ctx context.Context, bundlePath string, mode os.FileMode) (string, error) {
+	sha256bytes, err := hex.DecodeString(r.sha256sum)
 	if err != nil {
 		return "", err
 	}
-	return Download(r.URI, bundlePath, mode, sha256)
+	return Download(ctx, r.URI, bundlePath, mode, sha256bytes)
 }
 
 func (r *RemoteFile) GetSha256Sum() string {
--- a/pkg/os/copy.go
+++ b/pkg/os/copy.go
@@ -2,6 +2,7 @@ package os
 
 import (
 	"bytes"
+	"context"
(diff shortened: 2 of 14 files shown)


pkg/os/copy.go
```diff
@@ -51,9 +52,13 @@ func CopyFileSparse(src, dst string) error {
 	return copyFile(src, dst, true)
 }
 
-func CopySparse(dst io.WriteSeeker, src io.Reader) (int64, error) {
+func CopySparse(ctx context.Context, dst io.WriteSeeker, src io.Reader) (int64, error) {
 	copyBuf := make([]byte, copyChunkSize)
-	sparseWriter := newSparseWriter(dst)
+
+	if ctx == nil {
```
cfergeau, line 58: Is it possible for `ctx` to be nil? This is checked in a few places in this PR, but I'm not sure how this could happen

pkg/download/download.go
```diff
@@ -70,6 +71,9 @@ func Download(uri, destination string, mode os.FileMode, sha256sum []byte) (stri
 	if err != nil {
 		return "", errors.Wrapf(err, "unable to get request from %s", uri)
 	}
+	if ctx != nil {
```
cfergeau: Can this be `nil`?

cfergeau (commented):
This makes the necessary changes so that we can cancel bundle download/uncompression. Have you been able to test if a `ctx.cancel` does the right thing? If not, that's ok, we can iterate on top of these changes.

I see the context used in `CopySparse`, in `Download`, is it also used in the uncompression code?

redbeam:
@cfergeau yes, the `s.startCancel()` function called in `cancelUnlocked` is actually the cancel function of the context. And it works as expected.

https://github.com/crc-org/crc/blob/b174d0ccd2c5abdcda875468d1127caa5826036a/pkg/crc/machine/sync.go#L101-L112

And yes, the context is also provided to the decompression code.

pkg/os/copy.go
redbeam: Well, I added this before I had found out that allowing context to be nil is bad practice :sweat_smile: and `context.TODO()` should be used instead. But I left the check there just to be sure.

pkg/download/download.go
redbeam: See reply above

pkg/os/copy.go
cfergeau: If this is a "can never happen" situation, it can be confusing to leave this code which actually is dead code.
If you prefer to keep it for now, can you add a `log.Warnf("ctx is nil, this should not happen")` or similar to all the `ctx == nil` and `ctx != nil` checks?

pkg/os/copy.go
```diff
@@ -64,12 +69,13 @@ func CopySparse(dst io.WriteSeeker, src io.Reader) (int64, error) {
 }
 
 type sparseWriter struct {
+	context         context.Context
```
cfergeau, line 72: I don't know how much this would complexify this code, but https://pkg.go.dev/context#Background says:
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. 

Looking some more into it, I have no idea how to combine this recommandation with a cancellable `Write`

pkg/os/copy.go
cfergeau: Looking at go code (`exec.CommandContext` implementation), they use
```
	if ctx == nil {
		panic("nil Context")
	}
```

pkg/os/copy.go
cfergeau: `exec.CommandContext` and `http.Request` do not follow this guideline though, and store a `ctx` in a struct, let's just ignore my comment.

pkg/download/download.go
redbeam: Changed to a panic call when context is `nil`

pkg/os/copy.go
redbeam: Changed to a panic call when context is nil. Thanks for the suggestion.

cfergeau (approved):

Issue #4409, opened 2024-10-22
update bundle check to consider release candidate versions to be valid


anjannath:
Currently the bundle check considers the following bundle name as invalid `crc_libvirt_4.18.0-ec.2_amd64.pr980.crcbundle`, it expects the version to be x.y.z without containing the `-ec.2` part and the suffix `pr980` should have the separator `_` instead of a `.`

the CI on snc [0] is failing due to this strict checking, i think we can update the validation check to allow the version formats `x.y.z-rc.1`

[0] https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/crc-org_snc/980/pull-ci-crc-org-snc-master-e2e-snc/1848580407633645568

praveenkumar:
I think it is happen because of recent work @redbeam did around validation. @redbeam can you please take a look into this ?

redbeam:
@praveenkumar @anjannath We can modify the regex to allow additional characters in the version specification. But I'm not sure how to interpret the suffix - currently there is the `CustomBundleSuffix` that accepts just numbers. Is the `pr980` something different than the existing suffix? Or is it just a variant of it?

praveenkumar:
@redbeam this is something different which is used as part of CI bits so we need to modify the regex to allow that also `-ec.2` or `-rc.0` can be part of bundle string.

redbeam:
@praveenkumar But what about the suffix? What does `pr980` mean in this context?

anjannath:
> [@praveenkumar](https://github.com/praveenkumar) But what about the suffix? What does `pr980` mean in this context?

`pr980` is also the custom suffix, the CI job is just adding the pull request number at the end of `pr`, so we could modify the regex to accept alphanumeric for the custom suffix instead of just numbers, or we could instead update the snc job to create the custom suffix containing only numbers

and the separator between the bundle name and the custom suffix is currently a `.` while the regex expects it to be a `_` for this i think best to update the job on the snc side, here: https://github.com/crc-org/snc/blob/b4d9754b20d5636d62eea2c43e42100403048ce0/createdisk-library.sh#L9

praveenkumar:
We can remove `pr` from the  suffix and have only the number with use of  `_` but we still need to update our version regex to make sure it accepts `ec`, `alpha` , `rc` ..etc.

redbeam:
@anjannath Updating the job on snc side would be fine, but if there are any other places that could put their own suffix, maybe then we should change the regex.

[...]

PR #4413, opened 2024-10-24
bundle validation: expand regex

redbeam:
To allow more characters in bundle version field

Issue #4409

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -256,9 +256,21 @@ func GetBundleNameFromURI(bundleURI string) string {
 func GetBundleInfoFromName(bundleName string) (*FilenameInfo, error) {
 	var filenameInfo FilenameInfo
 
-	// crc_preset_driver_version_arch_customSuffix.crcbundle
-	bundleNameRegex := regexp.MustCompile(`crc(?:(?:_)([[:alpha:]]+))?_([[:alpha:]]+)_([[:alnum:].-]+)_([[:alnum:]]+)(?:(?:_)([0-9]+))?\.crcbundle`)
-	filenameParts := bundleNameRegex.FindStringSubmatch(bundleName)
+	/*
+		crc_preset_driver_version_arch_customSuffix.crcbundle
+
+		crc                      : Matches the fixed crc part
+		(?:(?:_)([[:alpha:]]+))? : Matches the preset part (optional)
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/bundle/metadata.go
```diff
@@ -257,7 +257,7 @@ func GetBundleInfoFromName(bundleName string) (*FilenameInfo, error) {
 	var filenameInfo FilenameInfo
 
 	// crc_preset_driver_version_arch_customSuffix.crcbundle
-	bundleNameRegex := regexp.MustCompile(`crc(?:(?:_)([[:alpha:]]+))?_([[:alpha:]]+)_([0-9.]+)_([[:alnum:]]+)(?:(?:_)([0-9]+))?\.crcbundle`)
+	bundleNameRegex := regexp.MustCompile(`crc(?:(?:_)([[:alpha:]]+))?_([[:alpha:]]+)_([[:alnum:].-]+)_([[:alnum:]]+)(?:(?:_)([0-9]+))?\.crcbundle`)
```
praveenkumar: @redbeam I think we are doing it wrong way since openshift is using semver and as per https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string we should use this regex to make sure it works for every release and we don't have to change it again.

redbeam: wow, that regex is going to be huge :D

praveenkumar: yes, but makes it more accurate :)

praveenkumar: If we go with that can you add comment on top of it around how this regex works so that if we want to update it would be each?
```
crc: Matches the fixed crc part.
(?:(?:_)([[:alpha:]]+))?: Matches an optional part with only alphabetic characters (e.g., microshift). The part is optional, so it can be absent.
__([[:alpha:]]+): Matches the next mandatory alphabetic part (e.g., libvirt).
([[:alnum:].-]+): Matches the version, which can include alphanumeric characters, dots (.), and hyphens (-) (e.g., 4.16.7 or 4.16.7-ec.2).
_([[:alnum:]]+): Matches the architecture or platform part (e.g., amd64).
(?:(?:_)([0-9]+))?: Optionally matches a trailing number after an underscore (e.g., 2345).
\.crcbundle: Matches the file extension .crcbundle.
```

redbeam: yeah no problem

PR #4419, opened 2024-10-28
update to openshift and microshift bundle version 4.17.1

anjannath:


anjannath:
`e2e-crc` and `integration-crc` CI failures are known: certificate rotation and behind proxy test respectively

adrianriobo:
Yeah also I am thinking this PR is just make official the test which has been ack downstream in order to publish the bundles.

PR #4420, opened 2024-10-28
kubeconfig: remove default namespace from crc-developer context

### LINKED ISSUE #703: 'developer' user doesn't have access to 'default' namespace
it seems the `developer` user doesn't have sufficient permissions to list/access/modify resources in the `default` namespace, but is should be able to do it since we also add a kube context `crc-developer` that has the namespace set to `default`

```
❯ oc whoami
developer

❯ oc project
error: you do not have rights to view project "default" specified in your config or the project doesn't exist

❯ oc config get-contexts
CURRENT   NAME            CLUSTER                AUTHINFO    NAMESPACE
          crc-admin       api-crc-testing:6443   kubeadmin   default
*         crc-developer   api-crc-testing:6443   developer   default
```

anjannath:
the 'developer' user don't have the required role-bindings to access the 'default' namespace and when user tries to access it we get:

```
Error from server (Forbidden): pods is forbidden: User "developer" cannot list resource "pods" in API group "" in the namespace "default"
```

fixes https://github.com/crc-org/snc/issues/703

### DIFF
--- a/pkg/crc/machine/kubeconfig.go
+++ b/pkg/crc/machine/kubeconfig.go
@@ -72,7 +72,7 @@ func writeKubeconfig(ip string, clusterConfig *types.ClusterConfig, ingressHTTPS
 	if err != nil {
 		return err
 	}
-	if err := addContext(cfg, clusterConfig.ClusterAPI, adminContext, "kubeadmin", kubeadminToken); err != nil {
+	if err := addContext(cfg, clusterConfig.ClusterAPI, adminContext, "kubeadmin", kubeadminToken, "default"); err != nil {
 		return err
 	}
 
@@ -80,7 +80,7 @@ func writeKubeconfig(ip string, clusterConfig *types.ClusterConfig, ingressHTTPS
 	if err != nil {
 		return err
 	}
-	if err := addContext(cfg, clusterConfig.ClusterAPI, developerContext, "developer", developerToken); err != nil {
+	if err := addContext(cfg, clusterConfig.ClusterAPI, developerContext, "developer", developerToken, ""); err != nil {
 		return err
 	}
 
@@ -142,7 +142,7 @@ func hostname(clusterAPI string) (string, error) {
 	return strings.ReplaceAll(h, ".", "-"), nil
 }
 
-func addContext(cfg *api.Config, clusterAPI, context, username, token string) error {
+func addContext(cfg *api.Config, clusterAPI, context, username, token, namespace string) error {
 	host, err := hostname(clusterAPI)
 	if err != nil {
 		return err
@@ -160,7 +160,7 @@ func addContext(cfg *api.Config, clusterAPI, context, username, token string) er
 	cfg.Contexts[context] = &api.Context{
 		Cluster:   host,
 		AuthInfo:  clusterUser,
-		Namespace: "default",
+		Namespace: namespace,
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/kubeconfig.go
```diff
@@ -72,15 +72,15 @@ func writeKubeconfig(ip string, clusterConfig *types.ClusterConfig, ingressHTTPS
...
+	if err := addContext(cfg, clusterConfig.ClusterAPI, adminContext, "kubeadmin", kubeadminToken, "default"); err != nil {
 		return err
 	}
 
 	developerToken, err := getTokenForUser("developer", "developer", ip, ca, clusterConfig, ingressHTTPSPort)
 	if err != nil {
 		return err
 	}
-	if err := addContext(cfg, clusterConfig.ClusterAPI, developerContext, "developer", developerToken); err != nil {
+	if err := addContext(cfg, clusterConfig.ClusterAPI, developerContext, "developer", developerToken, ""); err != nil {
```
cfergeau, line 83: what happens with the namespace set to `""` ?

anjannath: with an empty namespace in the context we still get the same error, but `oc config get-contexts` no longer shows a namespace for the `crc-developer` context

```
% oc config get-contexts
CURRENT   NAME            CLUSTER                AUTHINFO                         NAMESPACE
          crc-admin       api-crc-testing:6443   kubeadmin/api-crc-testing:6443   default
*         crc-developer   api-crc-testing:6443   developer/api-crc-testing:6443

```

praveenkumar: It is not only that but now when user try to check project switch to any project then it provide better error message instead forbidden error, we should have that as part of commit log.

```
$ oc config get-contexts
CURRENT   NAME            CLUSTER                AUTHINFO                         NAMESPACE
          crc-admin       api-crc-testing:6443   kubeadmin/api-crc-testing:6443   default
*         crc-developer   api-crc-testing:6443   developer/api-crc-testing:6443   

$ oc project
No project has been set. Pass a project name to make that the default.

$ oc project demo
error: You are not a member of project "demo".
You are not a member of any projects. You can request a project to be created with the 'new-project' command.
```

cfergeau: These are good reasons for the change, thanks for the explanations. Can this be added to the commit log?

anjannath: Thanks, i've updated the commit log to contain the error messages shown after the changes in this PR

pkg/crc/machine/kubeconfig_test.go
```diff
@@ -160,26 +160,27 @@ func Test_addContext(t *testing.T) {
...
 		{
-			input{"https://abcdd.api.com", "foo", "foo@abcdd", "secretToken"},
+			input{"https://abcdd.api.com", "foo", "foo@abcdd", "secretToken", "kube-system"},
 			"foo/abcdd-api-com",
 		},
 		{
-			input{"https://api.crc.testing:6443", "kubeadmin", "kubeadm", "secretToken"},
+			input{"https://api.crc.testing:6443", "kubeadmin", "kubeadm", "secretToken", "default"},
 			"kubeadmin/api-crc-testing:6443",
 		},
```
praveenkumar, line 186: @anjannath we should add one more test case which have empty string as namespace which make sure the context is added to kubeconfig but without namespace.

anjannath: updated the test

Issue #4407, opened 2024-10-21
[Test] Provide a dummy implementation for virtualMachine object for writing unit tests


rohanKanojia:
### Component

Internal Testing Utilities

### Description

_Originally discussed in internal chat_

Currently, there are not many unit tests in `pkg/crc/machine` module, probably due to the code dealing directly with managing virtual machines. It makes writing unit tests for CRC start, stop, and delete a bit difficult.

I think we can add some dummy mock implementation for this `virtualMachine` object that can be used while writing unit tests. It can store the state of the `virtualMachine` as some variable that we can expose in tests to assert whether CRC was able to modify the state of VirtualMachine after performing some action (Stop, Start, Delete, etc).

Gerard said they had something like a null driver in minishift codebase that could be used in the tests. Not 100% sure whether this is the correct reference: [minishift@pkg/minishift/provisioner/minishift_provisioner_test.go](https://github.com/minishift/minishift/blob/master/pkg/minishift/provisioner/minishift_provisioner_test.go#L38)

### Acceptance Criteria
- [ ] A dummy implementation is added so that it can be used in tests.
- [ ] code is refactored so that we can this dummy implementation can be picked with some argument (fake driver?).

cfergeau:
Would it be enough to be able to override `loadVirtualMachine` so that it is possible to load a `fakemachine` instead of the real one?

rohanKanojia:
@cfergeau : Yes, that can satisfy my use case. Do you have some examples of how I can achieve that? Perhaps you could share some old pull requests where the team might have added something close to this?

cfergeau:
> Do you have some examples of how I can achieve that? 

@lstocchi did something similar in vfkit unit tests recently, he should be able to help with that.

lstocchi:
@rohanKanojia I joined recently and I didn't have the time to look at the crc project yet so I'm not really able to tell you if what I did is suitable for your use case. You can find the code here https://github.com/crc-org/vfkit/pull/207/commits/41e879f9103d2945ceba40bb0b1861cfa8e6a4c4 . Basically I had to test cloud-init to see if the configuration file was actually read. So I created a fedora machine, mount an image containing the config file, started the vm and interacted with it to verify everything was set accordingly.
You can give that a look to see if it's helpful. 
BTW feel free to ping me if you need help, it would be a chance for me to look at crc 👍

cfergeau:
[...]

PR #4423, opened 2024-10-29
refactor (crc/machine) : Provide a dummy implementation for virtualMachine object for writing unit tests (#4407)

rohanKanojia:
## Description

**Fixes:** Issue #4407

**Relates to:** Issue #4407, PR https://github.com/crc-org/crc/pull/4400

## Type of change

## Checklist
 - [x] I use [conventional commits](https://www.conventionalcommits.org/) in my commit messages
 - [x] I tested my code on specified platforms
   - [x] Linux
   - [ ] Windows
   - [ ] MacOS

## Solution/Idea

+ Add a VirtualMachine interface and make the CRC `machine` package client use the VirtualMachine interface instead of a concrete implementation. This way we can inject a dummy test FakeVirtualMachine implementation into client tests that can ease writing tests for this package.
  - Add some additional methods in VirtualMachine interface so that we can replace direct usage of struct fields with interface methods
    - `Bundle()`
    - `Driver()`
    - `API()`
    - `GetHost()`
+ Add a new field `VirtualMachine` in the client so we can avoid creating it if it's already created.
+ Introduce a new constructor method `newClientWithVirtualMachine` in machine client that would have an additional VirtualMachine argument, this would be kept package private so that it's used only by tests in the same package.
+ Add FakeVirtualMachine sturct in `fakemachine` for adding dummy implementation for `VirtualMachine` interface. Currently, I've only completed methods used by `stop_test.go`, I'll add more in small increments as I get more familiar with the project. 

## Proposed changes
- Add new `VirtualMachine` interface and make client member functions use it instead of `virtualMachine` struct.
- Add new constructor in machine client to be able to override `VirtualMachine` implementation from tests
- These files have been modified to use `VirtualMachine` interface instead of `virtualMachine` struct (these changes are only related to changing use of `virtualMachine` struct to `VirtualMachine` interface:
  - `pkg/crc/machine/ip.go`
  - `pkg/crc/machine/poweroff.go`
  - `pkg/crc/machine/start.go`
  - `pkg/crc/machine/status.go`
  - `pkg/crc/machine/stop.go`

## Testing

It's a small refactor. I've only run E2E tests locally to verify if these changes don't introduce any kind of regression.

### DIFF
--- a/pkg/crc/machine/client.go
+++ b/pkg/crc/machine/client.go
@@ -30,21 +30,30 @@ type Client interface {
 }
 
 type client struct {
-	name   string
-	debug  bool
-	config crcConfig.Storage
+	name           string
+	debug          bool
+	config         crcConfig.Storage
+	virtualMachine VirtualMachine
 
 	diskDetails *memoize.Memoizer
 	ramDetails  *memoize.Memoizer
 }
 
 func NewClient(name string, debug bool, config crcConfig.Storage) Client {
+	return newClientWithVirtualMachine(name, debug, config, nil)
+}
+
+// newClientWithVirtualMachine creates a Client instance with an overridden VirtualMachine implementation.
+// It would not create a new VirtualMachine object. This method is primarily created for usage in tests so
+// that we can pass a fake VirtualMachine implementation.
+func newClientWithVirtualMachine(name string, debug bool, config crcConfig.Storage, vm VirtualMachine) Client {
 	return &client{
-		name:        name,
-		debug:       debug,
-		config:      config,
-		diskDetails: memoize.NewMemoizer(time.Minute, 5*time.Minute),
-		ramDetails:  memoize.NewMemoizer(30*time.Second, 2*time.Minute),
+		name:           name,
+		debug:          debug,
+		config:         config,
+		diskDetails:    memoize.NewMemoizer(time.Minute, 5*time.Minute),
+		ramDetails:     memoize.NewMemoizer(30*time.Second, 2*time.Minute),
+		virtualMachine: vm,
 	}
 }
 
--- a/pkg/crc/machine/stop_test.go
+++ b/pkg/crc/machine/stop_test.go
@@ -6,10 +6,70 @@ import (
 	"testing"
 
 	crcConfig "github.com/crc-org/crc/v2/pkg/crc/config"
+	"github.com/crc-org/crc/v2/pkg/crc/machine/fakemachine"
 	"github.com/crc-org/crc/v2/pkg/crc/machine/state"
 	crcOs "github.com/crc-org/crc/v2/pkg/os"
 	"github.com/stretchr/testify/assert"
 )
+
+func TestStop_WhenVMRunning_ThenShouldStopVirtualMachine(t *testing.T) {
+	// Given
+	crcConfigStorage := crcConfig.New(crcConfig.NewEmptyInMemoryStorage(), crcConfig.NewEmptyInMemorySecretStorage())
+	crcConfigStorage.AddSetting(crcConfig.NetworkMode, "user", crcConfig.ValidateBool, crcConfig.SuccessfullyApplied,
+		"network-mode")
+	_, err := crcConfigStorage.Set(crcConfig.NetworkMode, "true")
+	assert.NoError(t, err)
+	virtualMachine := fakemachine.NewFakeVirtualMachine(false, false)
+	client := newClientWithVirtualMachine("fake-virtual-machine", false, crcConfigStorage, virtualMachine)
+
+	// When
+	clusterState, stopErr := client.Stop()
+
+	// Then
+	assert.NoError(t, stopErr)
+	assert.Equal(t, clusterState, state.Stopped)
+	assert.Equal(t, virtualMachine.IsStopped, true)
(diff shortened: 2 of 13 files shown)


pkg/crc/machine/client.go
```diff
@@ -30,21 +30,30 @@ type Client interface {
...
-		debug:       debug,
-		config:      config,
-		diskDetails: memoize.NewMemoizer(time.Minute, 5*time.Minute),
-		ramDetails:  memoize.NewMemoizer(30*time.Second, 2*time.Minute),
+		name:           name,
+		debug:          debug,
+		config:         config,
+		diskDetails:    memoize.NewMemoizer(time.Minute, 5*time.Minute),
+		ramDetails:     memoize.NewMemoizer(30*time.Second, 2*time.Minute),
+		virtualMachine: vm,
```
cfergeau, line 56: I'd separate the commit adding this `virtualMachine: vm` field from the commit introducing the `VirtualMachine` interface. This commit is huge because of the introduction of this interface, but the changes needed for the interface changes are very mechanical (`vm.bundle` -> `vm.Bundle`) and such. If they are put in a commit doing only this, it's easier to review as we can focus on one thing, these mechanical changes.
Then in a 2nd commit come the "logic" changes which add a way to mock/test the virtual machine, and it's going to be a lot smaller, and easier to understand/review.

pkg/crc/machine/vsock.go
```diff
@@ -16,7 +16,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-func exposePorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint) error {
+type AbstractVSock interface {
+	ExposePorts(preset crcPreset.Preset, ingressHTTPPort, ingressHTTPSPort uint) error
+	UnExposePorts() error
+}
+
+type VSock struct{}
```
cfergeau: Maybe it could be `virtualMachine` which implements the `AbstractVsock` interface instead of introducing a dummy/empty type for it? (just some thoughts to discuss, I'm not telling you this change must be done :)

rohanKanojia: You're right about this. It's a cleaner approach than what I had added. Thanks a lot for suggesting this!

pkg/crc/machine/client.go
rohanKanojia: I have split this PR into multiple commits, please review if it's as per your expectations. I can adjust if additional changes are needed.

pkg/crc/machine/client.go
```diff
@@ -43,7 +43,7 @@ func NewClient(name string, debug bool, config crcConfig.Storage) Client {
 	return newClientWithVirtualMachine(name, debug, config, nil)
 }
 
-// newClientWithVirtualMachine creates a Client instance with a overridden VirtualMachine implementation.
+// newClientWithVirtualMachine creates a Client instance with an overridden VirtualMachine implementation.
```
cfergeau, line 46: You could squash this in the commit where it was added.

pkg/crc/machine/client.go
cfergeau: Hmm https://github.com/cfergeau/crc/commits/pr/issue4407/ is closer to what I had in mind. One commit doing only the VirtualMachine interface introduction and making use of its methods (could be 2 commits if you prefer, but I think it's fine as one). I added 2 follow-up commits which are meant to be squashed in the first one.
Then one commit adds `newClientWithVirtualMachine`, `loadVirtualMachineLazily` so that we can override it, and the rest of your series, I haven't touched it.

cfergeau: By the way, the content in `Solutions/Idea` and `Proposed Changes` would be nice to have in the individual commit logs, as one line commit logs are very terse when you'll get back to them in 6 months trying to remember what the commit was about, but also too long for a short log, they overflow quite a bit in github UI, git log, ... They typically should be about 60 characters in length.

Better to have a short 1line log, followed by more details in the log explaining why you are making the change, and giving a high level overview of what you are doing if it's a big change.

pkg/crc/machine/stop_test.go
```diff
@@ -6,11 +6,71 @@ import (
...
+	client := newClientWithVirtualMachine("fake-virtual-machine", false, crcConfigStorage, virtualMachine)
+
+	// When
+	clusterState, stopErr := client.Stop()
+
+	// Then
+	assert.NoError(t, stopErr)
+	assert.Equal(t, clusterState, state.Stopped)
+	assert.Equal(t, virtualMachine.IsStopped, true)
+	assert.Equal(t, virtualMachine.FakeSSHClient.LastExecutedCommand, "sudo -- sh -c 'crictl stop $(crictl ps -q)'")
```
cfergeau: In my opinion, the fact that `crictl stop` is the last command we run in the VM should be an implementation detail, it can change at any time if  we refactor this code, if we add new things to the bundle which needs to be stopped, ...

rohanKanojia: Thanks for review, I'll remove this assertion.

praveenkumar:
We might not using virtualmachine object for crc-ng where vm creation would be part of macadam.

Issue #4424, opened 2024-10-30
[QE] fix e2e test failure of podman


lilyLuLiu:
Test case `Pull image locally, push to registry, deploy` sometimes failed:
>stderr:
Cannot connect to Podman. Please verify your connection to the Linux system using podman system connection list, or try podman machine init and podman machine start to manage a new Linux VM
Error: unable to connect to Podman socket: failed to connect: dial tcp 192.168.130.11:22: i/o timeout

But this error never appear in nightly run. And I checked the machine, podman work well.

this is b/c somehow it is trying to use podman remote instead of podman.

praveenkumar:
> this is b/c somehow it is trying to use podman remote instead of podman.

Can you point out the feature file for that? Also I think we want to use podman remote instead podman on system since it is not default available for mac/windows.

adrianriobo:
Tests should be using podman-remote for win/mac and podman for linux

lilyLuLiu:
@praveenkumar , This failure all happened on tests of Linux.

praveenkumar:
> Tests should be using podman-remote for win/mac and podman for linux

@adrianriobo Does that mean for win/mac it is using podman socket which is running inside the crc VM or using podman-machine vm?

adrianriobo:
Yeah, for win and mac any scenario making use of podman is relying on "our" podman (crc-VM)

praveenkumar:
> Yeah, for win and mac any scenario making use of podman is relying on "our" podman (crc-VM)

@adrianriobo so in this case we should also do same for linux right?

adrianriobo:
Well, we are not making it for testing our podman setup, but b/c we rely on some container engine.  In the end we somehow would like to test the interaction between podman / crc.

I would suggest keep the same strategy, as in future we remove the podman socket  or make more tricky to make use of it ..we may end on a situation where we need to use podman machine on win and mac and keep using podman on linux.

dunno...WDYT?

praveenkumar:
What we want to have our test case follow same steps on all platform to have reliable metrics in case of this https://github.com/crc-org/crc/blob/main/test/e2e/features/story_openshift.feature#L55-L66 (which is causing issue). 

> Well, we are not making it for testing our podman setup, but b/c we rely on some container engine. In the end we somehow would like to test the interaction between podman / crc.

In this case I would suggest to have a different test case only for linux which use the podman installed on the system instead using podman-socket from the VM.

adrianriobo:
[...]

PR #4436, opened 2024-11-07
[QE] fix linux e2e failure of podman

lilyLuLiu:
**Fixes:** Issue #4424

### DIFF
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -299,6 +299,14 @@ func InitializeScenario(s *godog.ScenarioContext) {
 					fmt.Println(err)
 					os.Exit(1)
 				}
+
+				if runtime.GOOS == "linux" {
+					err = crcCmd.UnsetConfigPropertySucceedsOrFails("network-mode", "succeeds") // unsetting property that is not set gives 0 exitcode, so this works
+					if err != nil {
+						fmt.Println(err)
+						os.Exit(1)
+					}
+				}
 			}
 
 			if tag.Name == "@proxy" {


lilyLuLiu:
@adrianriobo Could you review this pr?

anjannath (approved):

anjannath:
`e2e-crc` failure is about cert rotation scenario and `integration-crc` failure is related to the proxy scenario, both are known failures

test/e2e/testsuite/testsuite.go
```diff
@@ -299,6 +299,14 @@ func InitializeScenario(s *godog.ScenarioContext) {
...
 					os.Exit(1)
 				}
+
+				if runtime.GOOS == "linux" {
+					err = crcCmd.UnsetConfigPropertySucceedsOrFails("network-mode", "succeeds") // unsetting property that is not set gives 0 exitcode, so this works
+					if err != nil {
+						fmt.Println(err)
+						os.Exit(1)
+					}
+				}
```
praveenkumar, line 309: you really doesn't need to use platform (linux) condition here because when you unset the network-mode property it would go to default which is for linux => system-mode and for other platform user mode.

So only following should work.

```golang
err = crcCmd.UnsetConfigPropertySucceedsOrFails("network-mode", "succeeds") // unsetting property that is not set gives 0 exitcode, so this works
					if err != nil {
						fmt.Println(err)
						os.Exit(1)
					}
```

praveenkumar (approved):

test/e2e/testsuite/testsuite.go
lilyLuLiu: @praveenkumar I didn't add the platform condition at the beginning, but the check on Windows failed.

lilyLuLiu: The unset command will fail on both windows and mac platform. 
Check in windows:
```
PS C:\Users\crcqe> crc config unset network-mode
Configuration property 'network-mode' does not exist
PS C:\Users\crcqe> echo $?
False
```

Check in mac:
```
lul@lul-mac hosts % crc config unset network-mode
Configuration property 'network-mode' does not exist
lul@lul-mac hosts % echo $?
1
```

praveenkumar: Ah `network-mode` is added for linux only :(

Issue #4433, opened 2024-11-05
[BUG] crc setup fails to check virtualization for Fedora arm64


adrianriobo:
### General information

 Pre testing of crc for arm64 on Fedora arm64

## Host Operating System
```bash
[fedora@ip-10-0-2-39 ~]$ cat /etc/*release
Fedora release 40 (Forty)
NAME="Fedora Linux"
VERSION="40 (Cloud Edition)"
ID=fedora
VERSION_ID=40
VERSION_CODENAME=""
PLATFORM_ID="platform:f40"
PRETTY_NAME="Fedora Linux 40 (Cloud Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:40"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f40/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=40
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=40
SUPPORT_END=2025-05-13
VARIANT="Cloud Edition"
VARIANT_ID=cloud
Fedora release 40 (Forty)
Fedora release 40 (Forty)
[fedora@ip-10-0-2-39 ~]$ uname -a
Linux ip-10-0-2-39.ec2.internal 6.11.5-200.fc40.aarch64 #1 SMP PREEMPT_DYNAMIC Tue Oct 22 19:29:37 UTC 2024 aarch64 GNU/Linux
```

### Steps to reproduce

  1. Install crc linux arm64 distributable  
  2. Run crc setup - b bundle.arm64.crcbundle
  
### Expected

Setup ends up successfully 

### Actual

Setup ends up failing while checking if virtualization is enable

### Logs

```bash
....
INFO Checking if Virtualization is enabled        
DEBU Checking if the vmx/svm flags are present in /proc/cpuinfo 
DEBU Could not find cpu flags from /proc/cpuinfo  
INFO Setting up virtualization                    
You need to enable virtualization in BIOS
```

### Aditional information

[...]

PR #4437, opened 2024-11-07
Ignore virtualization check for arm64 for Linux platform

praveenkumar:
Looks like for arm64 hardware there is no `vmx|svm` flag which point out if virtualization is enabled from bios or not. There is another way to handle it using `virt-host-validate` command but that do more test which we might not required. As of now this pr is going to ignore the virtualization check on arm64 to unblock CI for arm.

**Fixes:** Issue #4433

### DIFF
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -9,6 +9,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"text/template"
 
@@ -45,6 +46,10 @@ func checkRunningInsideWSL2() error {
 }
 
 func checkVirtualizationEnabled() error {
+	if runtime.GOARCH == "arm64" {
+		logging.Debug("Ignoring virtualization check for arm64")
+		return nil
+	}
 	logging.Debug("Checking if the vmx/svm flags are present in /proc/cpuinfo")
 	// Check if the cpu flags vmx or svm is present
 	flags, err := getCPUFlags()


pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -45,6 +46,10 @@ func checkRunningInsideWSL2() error {
 }
 
 func checkVirtualizationEnabled() error {
+	if runtime.GOARCH == "arm64" {
+		logging.Debug("Ignoring virtualization check for arm64")
```
anjannath, line 50: maybe make this a `logging.Warn` so users see it also when not running with `--log-level debug`?

anjannath (approved):

pkg/crc/preflight/preflight_checks_linux.go
cfergeau: This  is expected behaviour, this is not an error or something problematic so I'd keep this quiet, ie debug.

cfergeau:
Fwiw, this would be a good usecase for an "Arch" label https://github.com/crc-org/crc/blob/main/pkg/crc/preflight/labels.go

pkg/crc/preflight/preflight_checks_linux.go
anjannath: does it mean that all arm cpus support virtualization?

cfergeau: I think all aarch64 machines do, yes. While writing this, I'm thinking this is likely wrong for aarch64 virtual machines, at least Apple cpus did not have nested virt support until recently, and the hypervisor has to support it as well.

anjannath: there's the `systemd-detect-virt` which can detect if running inside a container or virtualization, but can't say if nested virt is supported

praveenkumar:
> Fwiw, this would be a good usecase for an "Arch" label https://github.com/crc-org/crc/blob/main/pkg/crc/preflight/labels.go

Right, if in future we need to have different logic for different arch multiple time then we should consider this or you think we should do it this now?

PR #4451, opened 2024-11-11
feat : Add support to customize the developer account password (#2359)

### LINKED ISSUE #2539: Be able to customize the developer account password
Now that the CA and the kubeadmin password are rotated, we are close to have a fully secured cluster. Only the developer account has always the same default password `developer`. 

We can rotate it by default and/or add a new configuration option/flag to customize it

rohanKanojia:
Fix #2539

**Relates to:** Issue #2539

## Type of change

## Checklist
 - [x] I use [conventional commits](https://www.conventionalcommits.org/) in my commit messages
 - [x] I tested my code on specified platforms
   - [x] Linux
   - [ ] Windows
   - [ ] MacOS

## Solution/Idea

+ Generate developer password instead of hardcoded `developer` string as password
+ Add config option to set developer password for cluster `developer-password`

## Proposed changes
crc would no longer have `developer` as user password. It would be generated each time cluster is created (just like kubeadmin password). If user wants to override it, they can use `developer-password` configuration option in CRC config.

## Testing

- After starting CRC cluster user would see a random string as password value instead of `developer
- If user has specified `developer-password` configuration option in CRC config, then that value is used in developer password:
```sh
    $ crc config set developer-password mypassword
    $ crc start
    [...]
    
    Log in as user:
      Username: developer
      Password: mypassword
```

### DIFF
--- a/pkg/crc/cluster/kubeadmin_password.go
+++ b/pkg/crc/cluster/kubeadmin_password.go
@@ -17,33 +17,21 @@ import (
 	"golang.org/x/crypto/bcrypt"
 )
 
-// GenerateKubeAdminUserPassword creates and put updated kubeadmin password to ~/.crc/machine/crc/kubeadmin-password
-func GenerateKubeAdminUserPassword() error {
-	logging.Infof("Generating new password for the kubeadmin user")
-	kubeAdminPasswordFile := constants.GetKubeAdminPasswordPath()
-	kubeAdminPassword, err := GenerateRandomPasswordHash(23)
+// GenerateUserPassword creates and put updated password to ~/.crc/machine/crc/ directory
+func GenerateUserPassword(passwordFile string, user string) error {
+	logging.Infof("Generating new password for the %s user", user)
+	password, err := GenerateRandomPasswordHash(23)
 	if err != nil {
-		return fmt.Errorf("Cannot generate the kubeadmin user password: %w", err)
+		return fmt.Errorf("cannot generate the %s user password: %w", user, err)
 	}
-	return os.WriteFile(kubeAdminPasswordFile, []byte(kubeAdminPassword), 0600)
+	return os.WriteFile(passwordFile, []byte(password), 0600)
 }
 
-// UpdateKubeAdminUserPassword updates the htpasswd secret
-func UpdateKubeAdminUserPassword(ctx context.Context, ocConfig oc.Config, newPassword string) error {
-	if newPassword != "" {
-		logging.Infof("Overriding password for kubeadmin user")
-		if err := os.WriteFile(constants.GetKubeAdminPasswordPath(), []byte(strings.TrimSpace(newPassword)), 0600); err != nil {
-			return err
-		}
-	}
-
-	kubeAdminPassword, err := GetKubeadminPassword()
+// UpdateUserPasswords updates the htpasswd secret
+func UpdateUserPasswords(ctx context.Context, ocConfig oc.Config, newKubeAdminPassword string, newDeveloperPassword string) error {
+	credentials, err := resolveUserPasswords(newKubeAdminPassword, newDeveloperPassword, constants.GetKubeAdminPasswordPath(), constants.GetDeveloperPasswordPath())
 	if err != nil {
-		return fmt.Errorf("Cannot read the kubeadmin user password from file: %w", err)
-	}
-	credentials := map[string]string{
-		"developer": "developer",
-		"kubeadmin": kubeAdminPassword,
+		return err
 	}
 
 	if err := WaitForOpenshiftResource(ctx, ocConfig, "secret"); err != nil {
@@ -62,7 +50,7 @@ func UpdateKubeAdminUserPassword(ctx context.Context, ocConfig oc.Config, newPas
 		return nil
 	}
 
-	logging.Infof("Changing the password for the kubeadmin user")
+	logging.Infof("Changing the password for the users")
 	expected, err := getHtpasswd(credentials, externals)
 	if err != nil {
 		return err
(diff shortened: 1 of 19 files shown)


praveenkumar:
looks like `prow/security` is not liking this https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-crc-org-crc-main-security (recent jobs are green but failed for this)

rohanKanojia:
@praveenkumar :  How can I reproduce this failure locally? I'm not able to get much idea about what's wrong by looking at logs:

```
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2024-11-18T08:45:42Z"}
ERRO[2024-11-18T08:45:43Z] Some steps failed:                           
ERRO[2024-11-18T08:45:43Z] 
  * could not run steps: step security failed: "security" test steps failed: "security" pod "security-openshift-ci-security-snyk-scan" failed: could not watch pod: the pod ci-op-m0wwywcn/security-openshift-ci-security-snyk-scan failed after 44s (failed containers: test): ContainerFailed one or more containers exited
```

Could it be possible that it's an intermittent failure?

cfergeau:
The failure is
```
[Medium] Use of Hardcoded Credentials
   ID: 3373e961-2410-402e-8db9-9c0b595d4e62 
   Path: pkg/crc/config/settings.go, line 32 
   Info: Do not hardcode passwords in code. Found hardcoded saved in DeveloperPassword.
```
which corresponds to
```
	DeveloperPassword        = "developer-password"
```

I think @albfan did some work on snyk's false positives recently (?)

rohanKanojia:
@cfergeau: Thanks for checking. We discussed this in a crc internal Slack chat and decided to add an exception for this particular rule.

praveenkumar:
Without this PR the `developer` user have `developer` password and it doesn't have a way to change it. There are users who might be using this as scripted way and consuming this well known password. Now with this change either user before hand set the required password or they will get a random password, this looks like a UX change. I think what we should do is have default password set to be `developer` but provide option to user to change if they want this way the UX is not going to change and existing scripts/automation wouldn't fail. @crc-org/crc-team wdyt?

rohanKanojia:
@praveenkumar : Maybe we can add a configuration named `useLegacyDeveloperPassword` that would be true by default.

praveenkumar:
> `useLegacyDeveloperPassword`

@rohanKanojia I don't think it is good  to have multiple config option just for developer user.

pkg/crc/cluster/kubeadmin_password.go
```diff
@@ -17,33 +17,21 @@ import (
...
 )
 
-// GenerateKubeAdminUserPassword creates and put updated kubeadmin password to ~/.crc/machine/crc/kubeadmin-password
-func GenerateKubeAdminUserPassword() error {
-	logging.Infof("Generating new password for the kubeadmin user")
-	kubeAdminPasswordFile := constants.GetKubeAdminPasswordPath()
+// GenerateUserPassword creates and put updated password to ~/.crc/machine/crc/ directory
+func GenerateUserPassword(passwordFile string, user string) error {
+	logging.Infof("Generating new password for the %s user", user)
 	kubeAdminPassword, err := GenerateRandomPasswordHash(23)
```
cfergeau: This could now be named `password` or such, this is no longer necessarily a kubeadmin password.

pkg/crc/cluster/kubeadmin_password.go
```diff
@@ -17,33 +17,21 @@ import (
...
-	if newPassword != "" {
-		logging.Infof("Overriding password for kubeadmin user")
-		if err := os.WriteFile(constants.GetKubeAdminPasswordPath(), []byte(strings.TrimSpace(newPassword)), 0600); err != nil {
-			return err
-		}
-	}
-
-	kubeAdminPassword, err := GetKubeadminPassword()
+// UpdateUserPassword updates the htpasswd secret
+func UpdateUserPassword(ctx context.Context, ocConfig oc.Config, newKubeAdminPassword string, newDeveloperPassword string) error {
```
cfergeau: I'd use the plural here as there are multiple passwords involved `UpdateUserPasswords`

pkg/crc/cluster/kubeadmin_password.go
```diff
@@ -17,33 +17,21 @@ import (
...
+	credentials, err := resolveUserPassword(newKubeAdminPassword, newDeveloperPassword, constants.GetKubeAdminPasswordPath(), constants.GetDeveloperPasswordPath())
```
cfergeau: Same here, `resolveUserPasswords`

pkg/crc/cluster/kubeadmin_password.go
```diff
@@ -50,7 +50,7 @@ func UpdateUserPassword(ctx context.Context, ocConfig oc.Config, newKubeAdminPas
 		return nil
 	}
 
-	logging.Infof("Changing the password for the kubeadmin user")
+	logging.Infof("Changing the password for the users")
 	expected, err := getHtpasswd(credentials, externals)
```
cfergeau, line 54: Did this belong in the previous commit?

pkg/crc/cluster/kubeadmin_password.go
```diff
@@ -60,7 +60,7 @@ func UpdateUserPassword(ctx context.Context, ocConfig oc.Config, newKubeAdminPas
 		"-n", "openshift-config", "--type", "merge"}
 	_, stderr, err = ocConfig.RunOcCommandPrivate(cmdArgs...)
 	if err != nil {
-		return fmt.Errorf("Failed to update kubeadmin password %v: %s", err, stderr)
+		return fmt.Errorf("failed to update user passwords %v: %s", err, stderr)
```
cfergeau, line 63: Same question?

pkg/crc/config/settings.go
```diff
@@ -134,6 +135,8 @@ func RegisterSettings(cfg *Config) {
 
 	cfg.AddSetting(KubeAdminPassword, "", validateString, SuccessfullyApplied,
 		"User defined kubeadmin password")
+	cfg.AddSetting(DeveloperPassword, "", validateString, SuccessfullyApplied,
```
cfergeau: What if we use `"developer"` here as the default value (second parameter to `AddSetting`), wouldn't this make the second commit redundant?

rohanKanojia: This change was added in second commit: e9a11fc31e65e4e8d88beacac7effddba84ca152

Did you mean we can avoid third commit by adding developer as default value?

cfergeau: I don't remember exactly what I meant, but the current iteration has `cfg.AddSetting(DeveloperPassword, constants.DefaultDeveloperPassword, ...` which is what I was hoping/suggesting we could use, so all is well :)

gbraad:
Can we add an integration test to verify this works as expected?

rohanKanojia:
@gbraad: Will add an e2e test to verify these scenarios:
- **Given** user has no developer-password configured, **when** user performs crc start, **then** default developer password is `developer`
- **Given** user has configured developer-password as `secret`,  **when** user performs crc start, **then** developer password is `secret`

cfergeau (approved):
/lgtm
/approve

Issue #4326, opened 2024-08-21
[QE] Add github action for linux PR check with machine from testing farm


lilyLuLiu:
[#4305](https://github.com/crc-org/crc/issues/4305) 
The self-hosted runner have been added. Now we can use it to reserve machine for linux PR check

adrianriobo:
There is a new requirement to use a jump host from self hosted runner, the jump will be accomplished using https://github.com/adrianriobo/deliverest/issues/9

lilyLuLiu:
The machine reserved from the Testing Farm is root user, while crc is not allowed to run with root. 
I tried to create a user and then switch to the user in the testing container, but it did not work.
Will ask the Testing Farm team whether supports reservation non-root machines.

praveenkumar:
> I tried to create a user and then switch to the user in the testing container, but it did not work.

What is the error, can you put the logs?

lilyLuLiu:
@praveenkumar Hi, the issue has been solved: Copy the root authorized_keys file to the newly created user. Then the reserved machine could be accessed by the new user.

praveenkumar:
In this case should we close this ?

lilyLuLiu:
@praveenkumar I'm still working on this ticket. 
The reserved testing farm machine was returned early during test running. Need some further investigation on it.

lilyLuLiu:
The PR check for linux arm64 work well now. 
One thing left: the cert rotation test case is disabled:  the reserved machine will be returned early because the test change the date. 
So open a MR in testing farm: https://gitlab.com/testing-farm/tests/-/merge_requests/49

praveenkumar:
> So open a MR in testing farm: https://gitlab.com/testing-farm/tests/-/merge_requests/49

@lilyLuLiu this is now merged. Now we just need to add the github action for it?

lilyLuLiu:
@praveenkumar The github action is already added, but disabled the `cert_rotation` test case. 
The MR in testing farm is merged, but not work as expect. I'm communication with TF team about it. 
But I think we can close this issue.

praveenkumar:
@lilyLuLiu Ack, closing this one then.

PR #4461, opened 2024-11-18
[QE] fix github action error imported by 0b4c6636b8324b92d43e5f04488d…

lilyLuLiu:
**Relates to:** PR #4320, issue #4326


lilyLuLiu:
`working-directory` only works with run action. For other actions, `working-directory` not working.
Also, the runner can only run one job, there is no need to worry about multiple jobs running concurrently in the runner. 
So remove the directory.

anjannath (approved):
we need to merge this before the changes are picked up by github actions, since the workflow file is taken from the main branch

PR #4468, opened 2024-11-20
[QE] update github action for linux arm64 tests

lilyLuLiu:
https://github.com/crc-org/crc/issues/4326

### DIFF
--- a/.github/workflows/linux-qe-template.yml
+++ b/.github/workflows/linux-qe-template.yml
@@ -16,11 +16,14 @@ on:
 
 jobs:
   linux-qe:
-    runs-on: [self-hosted, linux, testing-farm]
+    runs-on: ubuntu-24.04
     permissions:
       statuses: write # needed to update commit status (pending/failure/sucess)
       checks: write # as documented in https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#pr-run-permissions
     steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
       - name: Download gh context
         id: download-gh-context-artifact
         uses: actions/download-artifact@v4
@@ -31,7 +34,10 @@ jobs:
 
       - name: prepare
         run: |
-          sudo yum install podman openssh-server -y
+          # Install Testing farm CLI
+          pip3 install --user tft-cli
+          sudo apt install podman openssh-server -y
+          testing-farm --help
 
           # Get origin commit sha for testing
           commit_sha=$(cat gh_context.json | jq -r '.event.after')
@@ -46,10 +52,17 @@ jobs:
           if [[ "${{inputs.qe-type}}" == "e2e" ]]; then
             status_context="${status_context}-${{inputs.preset}}"
           fi
-          status_context="${status_context}/windows-${{matrix.windows-version}}-${{matrix.windows-featurepack}}"
+          status_context="${status_context}/Linux-ARM64"
           echo "status_context=${status_context}" >> "$GITHUB_ENV"
 
-          echo "${PULL_SECRET}" > pull-secret
+          ocp_version_line=$(cat Makefile | grep "OPENSHIFT_VERSION ?=")
+          ocp_bundle_v=${ocp_version_line##*= } 
+          echo "ocp_bundle_v=${ocp_bundle_v}" >> "$GITHUB_ENV"
+          echo "ocp_bundle_v=${ocp_bundle_v}"
+          microshift_version_line=$(cat Makefile | grep "MICROSHIFT_VERSION ?=")
+          microshift_bundle_v=${microshift_version_line##*= } 
+          echo "microshift_bundle_v=${microshift_bundle_v}" >> "$GITHUB_ENV"
+          echo "microshift_bundle_v=${microshift_bundle_v}"
 
       - name: Download linux binary
         uses: actions/download-artifact@v4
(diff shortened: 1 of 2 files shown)


adrianriobo:
hey @lilyLuLiu I added the WIP yesterday cause I was gonna take a look but saw those testing commits... so to mark the PR to no to check yet added the WIP

lilyLuLiu:
@adrianriobo Thanks for the reminder. I mark this pr as draft. Will this close the checking running? If not, how can I stop the check?

adrianriobo:
by check I meant myself checking / reviewing

lilyLuLiu:
@adrianriobo I have finished this code update. Could you please review it when you have time? Thanks

.github/workflows/linux-qe-template.yml
```diff
@@ -131,25 +142,65 @@ jobs:
...
              -aName 'crc' \
              -freshEnv 'false' \
              -download 'false' 
           podman logs -f crc-linux-install-${{inputs.qe-type}}-${{inputs.preset}}
 
+          # Download arm64 bundle in the reserved machine
+          echo "Start download bundle on reserved machine"
+          
+          if [[ "${{inputs.preset}}" == "microshift" ]]; then
+            bundle_url="https://crcqe-asia.s3.ap-south-1.amazonaws.com/bundles/microshift/4.17.7-arm64"
```
adrianriobo: Can you pick the bundle version from Makefile? Otherwise we will need to change the version each time

.github/workflows/linux-qe-template.yml
```diff
@@ -189,6 +240,17 @@ jobs:
             **/*.results
             **/*.log
 
+      - name: Return machine and clear env
+        env: 
+          TESTING_FARM_API_TOKEN: ${{ secrets.TESTING_FARM_API_TOKEN }}
+        if: always()
+        run: |
+          export TESTING_FARM_API_TOKEN=${TESTING_FARM_API_TOKEN}
+          testing-farm cancel $(cat requestid)
```
adrianriobo, line 244: Is this the way they use to decommission the machine?

.github/workflows/linux-qe-test.yml
```diff
@@ -10,21 +10,13 @@ jobs:
   linux-qe:
     if: | 
       github.event.workflow_run.conclusion == 'success' &&
-      github.event.workflow_run.event == 'pull_request' &&
       github.event.workflow_run.triggering_actor != 'dependabot[bot]'
-    uses: crc-org/crc/.github/workflows/linux-qe-template.yml@main
+    uses: lilyLuliu/crc/.github/workflows/linux-qe-template.yml@main
```
adrianriobo: Need to change this before merge

.github/workflows/linux-qe-template.yml
```diff
@@ -46,10 +52,13 @@ jobs:
           if [[ "${{inputs.qe-type}}" == "e2e" ]]; then
             status_context="${status_context}-${{inputs.preset}}"
           fi
-          status_context="${status_context}/windows-${{matrix.windows-version}}-${{matrix.windows-featurepack}}"
+          status_context="${status_context}/Linux-ARM64"
           echo "status_context=${status_context}" >> "$GITHUB_ENV"
 
-          echo "${PULL_SECRET}" > pull-secret
+          version_line=$(cat Makefile | grep "OPENSHIFT_VERSION ?=")
```
adrianriobo: No very common but there were certain releases specially on new minor openshift versions where openshift and microshift versions are not in sync...better fix it now picking both versions as you already have the condition to check the right bundle

.github/workflows/linux-qe-test.yml
lilyLuLiu: The testing farm upgrade again that bastion was removed for connection to the reserved machine. So I made some tests on my repo. The has changed back.

.github/workflows/linux-qe-template.yml
lilyLuLiu: Add the microshift bundle version

.github/workflows/linux-qe-template.yml
lilyLuLiu: yes, this is way to return the machine.

.github/workflows/linux-qe-template.yml
lilyLuLiu: Updated as your suggestion.

.github/workflows/linux-qe-template.yml
```diff
@@ -16,11 +16,14 @@ on:
 
 jobs:
   linux-qe:
-    runs-on: [self-hosted, linux, testing-farm]
+    runs-on: ubuntu-24.04
     permissions:
       statuses: write # needed to update commit status (pending/failure/sucess)
       checks: write # as documented in https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#pr-run-permissions
     steps:
+      - name: Check out repository code
```
adrianriobo, line 24: Why are you cloning the code here? to get the bunble versions? If that is the case it will not work on the PR for the new bundles version.

This is b/c this is a workflow_run which is executed on the context of the default branch of the upstream repo which has actual verisons not the ones you set on the PR....

For this to work you need to pass that information on an artifact (as we do with images...)

But as this is an edge case feel free to create a following issue with this may @albfan can take a look at it...as is a minor chnage but it requires to know about how this testing strategy works.

beyond this LGTM

adrianriobo (approved):

.github/workflows/linux-qe-template.yml
adrianriobo: ^^ @lilyLuLiu

lilyLuLiu: Thanks Adrian. Will fix this for a following pr.

PR #4473, opened 2024-11-21
This superseed #4452 with addition commit of using builtin min function

praveenkumar:


cfergeau (commented):
What is "latest golang"? Or rather, when was the `min` builtin introduced? is it already there in 1.22/1.23, or will this cause issues with slightly older golangs?

cfergeau:
> What is "latest golang"? Or rather, when was the `min` builtin introduced? is it already there in 1.22/1.23, or will this cause issues with slightly older golangs?

This was added in golang 1.21 https://pkg.go.dev/builtin@go1.21.0#min

cfergeau:
I've updated the commit log of the `min` commit, and push these changes back to https://github.com/crc-org/crc/pull/4452 to keep everything in a single place.

Issue #4470, opened 2024-11-20
[BUG] `crc setup` panics when image tag is missing


redbeam:
### General information

  * OS: Linux + macOS
  * Hypervisor: irrelevant
  * Did you run `crc setup` before starting it (Yes/No)? No
  * Running CRC on: Laptop

## CRC version
```bash
$ crc version
CRC version: 2.43.0+b6ffde
OpenShift version: 4.17.3
MicroShift version: 4.17.3
```
  
## CRC status
```bash
$ crc status --log-level debug
DEBU CRC version: 2.43.0+b6ffde                   
DEBU OpenShift version: 4.17.3                    
DEBU MicroShift version: 4.17.3                   
DEBU Running 'crc status'                         
Machine does not exist. Use 'crc start' to create it
```

## CRC config
```bash
$ crc config view
- consent-telemetry                     : no
```

## Host Operating System
```bash
$ sw_vers 
ProductName:		macOS
ProductVersion:		15.1
BuildVersion:		24B83
```

### Steps to reproduce

  1. `crc setup -b docker://quay.io/crcont/openshift-bundle`

### Expected
setup should warn user about the missing tag/end with an error/assume "latest" for the tag (or some other default value)

### Actual
```
$ crc setup -b docker://quay.io/crcont/openshift-bundle
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/crc-org/crc/v2/pkg/crc/machine/bundle.GetBundleNameFromURI({0x16d2bb90d, 0x28})
	/Users/mskvarla/crc/pkg/crc/machine/bundle/metadata.go:247 +0x138
github.com/crc-org/crc/v2/pkg/crc/validation.ValidateBundle({0x16d2bb90d, 0x28}, {0x1043cde37, 0x9})
	/Users/mskvarla/crc/pkg/crc/validation/validation.go:92 +0x30
github.com/crc-org/crc/v2/cmd/crc/cmd.runSetup({0x32d7a000000c9c18?, 0x673db5bc?, 0x140000c9b88?})
	/Users/mskvarla/crc/cmd/crc/cmd/setup.go:63 +0x214
github.com/crc-org/crc/v2/cmd/crc/cmd.init.func10(0x1043d7913?, {0x140004a5080, 0x0, 0x2})
	/Users/mskvarla/crc/cmd/crc/cmd/setup.go:42 +0x60
[...]

PR #4474, opened 2024-11-21
fix: Download bundle without specifying tag

A-725-K:
The "latest" tag is not supported, therefore it is necessary to stop with an error in case it is not provided when using the -b flag command line flag.

Add unit tests to reproduce the bug and verify the fix.

Closes issue #4470

**Fixes:** Issue #4470 

## Solution/Idea

Before, in the `GetBundleNameFromURI` function, the `imageAndTag` array was trying to access its second element without ensuring that the length is at least 2 causing a crash in case the `-b` command line option was used to specify a bundle from a container image repository (using the `docker://` scheme) without a tag. I have added a check to prevent the program from crashing, returning an error to the user since the "latest" tag is not an option.

## Proposed changes

In `GetBundleNameFromURI` function:

1. Check the lenght of `imageAndTag` array
  1. if it is greater or equal to 2, then obtain image name and tag as `imageAndTag[0]` and `imageAndTag[1]` respectively
  2. otherwise, bail out with an error
2. Change all the use in the codebase, since now the function returns also an error in case of failure

## Testing

Added unit tests to reproduce the bug and validate the fix proposed.
Check the `TestGetBundleNameFromURI` function in the `pkg/crc/machine/bundle/metadata_test.go` file.
Running `crc setup -b docker://quay.io/crcont/openshift-bundle` should now return an error message instead of crashing.

### DIFF
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -243,17 +243,19 @@ func GetCustomBundleName(bundleFilename string) string {
 	return fmt.Sprintf("%s_%d%s", baseName, time.Now().Unix(), bundleExtension)
 }
 
-func GetBundleNameFromURI(bundleURI string) string {
-	// the URI is expected to have been validated by validation.ValidateBundlePath first
+func GetBundleNameFromURI(bundleURI string) (string, error) {
 	switch {
 	case strings.HasPrefix(bundleURI, "docker://"):
 		imageAndTag := strings.Split(path.Base(bundleURI), ":")
-		return constants.BundleForPreset(image.GetPresetName(imageAndTag[0]), imageAndTag[1])
+		if len(imageAndTag) < 2 {
+			return "", fmt.Errorf("No tag found in bundle URI")
+		}
+		return constants.BundleForPreset(image.GetPresetName(imageAndTag[0]), imageAndTag[1]), nil
 	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
-		return path.Base(bundleURI)
+		return path.Base(bundleURI), nil
 	default:
 		// local path
-		return filepath.Base(bundleURI)
+		return filepath.Base(bundleURI), nil
 	}
 }
 
--- a/pkg/crc/machine/bundle/metadata_test.go
+++ b/pkg/crc/machine/bundle/metadata_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"unicode"
@@ -55,7 +56,7 @@ func jsonForBundleWithVersion(version, name string) string {
       {
         "name": "crc.qcow2",
         "format": "qcow2",
-	"size": "9",
+        "size": "9",
         "sha256sum": "245a0e5acd4f09000a9a5f37d731082ed1cf3fdcad1b5320cbe9b153c9fd82a4"
       }
     ],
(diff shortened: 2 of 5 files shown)


pkg/crc/machine/bundle/metadata.go
```diff
@@ -238,17 +238,20 @@ func GetCustomBundleName(bundleFilename string) string {
 	return fmt.Sprintf("%s_%d%s", baseName, time.Now().Unix(), bundleExtension)
 }
 
-func GetBundleNameFromURI(bundleURI string) string {
+func GetBundleNameFromURI(bundleURI string) (string, error) {
 	// the URI is expected to have been validated by validation.ValidateBundlePath first
```
cfergeau: This comment can be removed as it did not prevent this bug from occurring, and this is now no longer required.

cfergeau (approved):
Looks good to me, thanks for contributing this fix!

pkg/crc/machine/bundle/metadata_test.go
```diff
@@ -289,3 +289,27 @@ func TestGetBundleInfoFromNameInvalid(t *testing.T) {
...
 	assert.Error(t, err)
 }
+
+func TestGetBundleNameFromURI(t *testing.T) {
+	// URI with no tag
+	bundleName, err := GetBundleNameFromURI("docker://quay.io/crcont/openshift-bundle")
+	assert.Equal(t, "", bundleName)
+	assert.Error(t, err)
+
+	expectedBundleName := "crc_libvirt_4.17.3_amd64.crcbundle"
```
cfergeau: `expectedBundleName` needs to be OS dependent (`switch runtime.GOOS { ... }` for example), currently this fails on macos/windows:
```
=== RUN   TestGetBundleNameFromURI
    metadata_test.go:304: 
        	Error Trace:	/Users/runner/work/crc/crc/pkg/crc/machine/bundle/metadata_test.go:304
        	Error:      	Not equal: 
        	            	expected: "crc_libvirt_4.17.3_amd64.crcbundle"
        	            	actual  : "crc_vfkit_4.17.3_amd64.crcbundle"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-crc_libvirt_4.17.3_amd64.crcbundle
        	            	+crc_vfkit_4.17.3_amd64.crcbundle
        	Test:       	TestGetBundleNameFromURI
--- FAIL: TestGetBundleNameFromURI (0.00s)
```

```
=== RUN   TestGetBundleNameFromURI
    metadata_test.go:304: 
        	Error Trace:	D:/a/crc/crc/pkg/crc/machine/bundle/metadata_test.go:304
        	Error:      	Not equal: 
        	            	expected: "crc_libvirt_4.17.3_amd64.crcbundle"
        	            	actual  : "crc_hyperv_4.17.3_amd64.crcbundle"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-crc_libvirt_4.17.3_amd64.crcbundle
        	            	+crc_hyperv_4.17.3_amd64.crcbundle
        	Test:       	TestGetBundleNameFromURI
--- FAIL: TestGetBundleNameFromURI (0.00s)
```

pkg/crc/machine/bundle/metadata.go
A-725-K: Done!

pkg/crc/machine/bundle/metadata_test.go
A-725-K: I will have a look and modify the name based on the OS! Thanks for the review :+1:

cfergeau (approved):

anjannath:
@A-725-K we just merged another PR (#4405) and that created conflicts here, could you please rebase and push again, thanks!

A-725-K:
> @A-725-K we just merged another PR (#4405) and that created conflicts here, could you please rebase and push again, thanks!

@anjannath rebased and pushed! Thanks for notifying me!

praveenkumar:
@A-725-K There is still conflict around `pkg/crc/machine/bundle/metadata_test.go` can you rebase and repush. Thanks for contributing.

/rebase

A-725-K:
@praveenkumar thank you for reviewing it! Rebased and pushed again!

Issue #4218, opened 2024-06-10
[BUG] CRC not starting when increasing the disk-size


rmestrum:
### General information

  * OS: macOS
  * Hypervisor: vfkit
  * Did you run `crc setup` before starting it (Yes/No)? Yes
  * Running CRC on: Laptop

## CRC version
```bash
CRC version: 2.37.1+36d451
OpenShift version: 4.15.14
```
  
## CRC status
```bash
DEBU CRC version: 2.37.1+36d451                   
DEBU OpenShift version: 4.15.14                   
DEBU Running 'crc status'                         
CRC VM:          Running
OpenShift:       Unreachable (v4.15.14)
Disk Usage:      0B of 0B (Inside the CRC VM)
Cache Usage:     38.32GB
Cache Directory: /Users/rmestrum/.crc/cache
```

## CRC config
```bash
- consent-telemetry                     : no
- cpus                                  : 6
```

## Host Operating System
```bash
ProductName:		macOS
ProductVersion:		14.5
BuildVersion:		23F79
```

### Steps to reproduce

  1. crc setup
  2. crs config set cpus 6
  3. crc start --disk-size 100
     
### Expected
```bash
INFO Using bundle path /Users/rmestrum/.crc/cache/crc_vfkit_4.15.14_arm64.crcbundle 
INFO Checking if running macOS version >= 13.x    
INFO Checking if running as non-root              
INFO Checking if crc-admin-helper executable is cached 
INFO Checking if running on a supported CPU architecture 
INFO Checking if crc executable symlink exists    
INFO Checking minimum RAM requirements            
INFO Check if Podman binary exists in: /Users/rmestrum/.crc/bin/oc 
INFO Checking if running emulated on Apple silicon 
INFO Checking if vfkit is installed               
INFO Checking if old launchd config for tray and/or daemon exists 
INFO Checking if crc daemon plist file is present and loaded 
INFO Checking SSH port availability               
INFO Loading bundle: crc_vfkit_4.15.14_arm64...   
INFO Creating CRC VM for OpenShift 4.15.14...     
INFO Generating new SSH key pair...               
INFO Generating new password for the kubeadmin user 
INFO Starting CRC VM for openshift 4.15.14...     
INFO CRC instance is running with IP 127.0.0.1    
INFO CRC VM is running                            
INFO Updating authorized keys...                  
INFO Configuring shared directories               
INFO Check internal and public DNS query...       
INFO Check DNS query from host...                 
[...]

PR #4477, opened 2024-11-22
Update the publicDNSQueryURI in DNS Check to https://quay.io 

vyasgun:
**Relates to:** Issue #4218 (error code `301` was seen in this issue)

## Solution/Idea

Change the publicDNSQueryURI to `https` to prevent `301 Moved Permanently` during DNS check.

The alternate is to add `-sSL` flags to the curl command during the check for a quiet follow to any redirects.

### DIFF
--- a/pkg/crc/services/dns/dns.go
+++ b/pkg/crc/services/dns/dns.go
@@ -20,7 +20,7 @@ import (
 
 const (
 	dnsServicePort    = 53
-	publicDNSQueryURI = "quay.io"
+	publicDNSQueryURI = "https://quay.io"
 	dnsmasqService    = "dnsmasq.service"
 )
 


pkg/crc/services/dns/dns.go
```diff
@@ -119,7 +119,7 @@ func CheckCRCPublicDNSReachable(serviceConfig services.ServicePostStartConfig) (
 		// try without using proxy
 		proxyConfig = &httpproxy.ProxyConfig{}
 	}
-	curlArgs := []string{"--head", publicDNSQueryURI}
+	curlArgs := []string{"--head", publicDNSQueryURI, "-L"}
```
gbraad: Follow is good, but I think it is better to do

`curl -sSL`

`-S`: This option is used in conjunction with `-s` (silent). It stands for "show error." When used with `-s`, it will still suppress the progress meter but will show error messages if the request fails. This could be beneficial for logging and debugging purposes since it reduces noise. It might even be possible to return the actual error as user-facing information. WDYT?

gbraad (commented):
https://github.com/crc-org/crc/pull/4477/commits/97ee5fcc032ea7b8654c0f1ca8999f30ea528252#r1853762839 => Perhaps the error message can be helpful to the user.

kaovilai (commented):
It's a DNS check, not a what's showing at the website check. Why is this needed?

pkg/crc/services/dns/dns.go
vyasgun: Thank you for the review - I updated the PR to use the `https://quay.io` as I feel it's simpler to just update the URL.

kaovilai:
@vyasgun
> DEBU SSH command results: err: <nil>, output: HTTP/1.1 301 Moved Permanently

is not an error.

`err: <nil>`

kaovilai:
> [Informational responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#informational_responses) (100 – 199)
> [Successful responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses) (200 – 299)
> [Redirection messages](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages) (300 – 399)
> [Client error responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) (400 – 499)
> [Server error responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses) (500 – 599)

[src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#:~:text=Informational%20responses%20(,%E2%80%93%20599))

vyasgun:
@kaovilai yes but checking at https://quay.io instead of http://quay.io returns `200` which is a clearer way to interpret that the check succeeded and also serves the same purpose of querying quay.io.

kaovilai (approved):
I would prefer the test actually be a dns query as the name implies...

```
❯ dig quay.io +short
54.243.17.128
54.85.81.91
3.219.175.113
54.210.148.48
52.54.136.228
52.55.96.46

~
❯ dig non-exist.quay.io +short

```

and check for non-empty output.

but this is a step in the right direction. just saying it's not solving an error.

praveenkumar:
> Relates to: Issue https://github.com/crc-org/crc/issues/4218 (error code 301 was seen in this issue)

@vyasgun This doesn't look like related to this issue, or I am missing something here.

vyasgun:
@praveenkumar you are right -- it is not directly related to the issue except that as mentioned in the description, it was seen in the logs of this issue and prompted some discussion and I was looking into it. This is not even an error but just adding `https` to the query URI improves clarity of the debug message as well as performing the same task of querying quay.io. Should I update the description to something else to prevent confusion?

cfergeau:
> I would prefer the test actually be a dns query as the name implies...
> 
> ```
> ❯ dig quay.io +short
> 54.243.17.128
> 54.85.81.91
> 3.219.175.113
> 54.210.148.48
> 52.54.136.228
> 52.55.96.46
> 
> ~
> ❯ dig non-exist.quay.io +short
> ```
> 
> and check for non-empty output.
> 
> but this is a step in the right direction. just saying it's not solving an error.

This used to be tested this way, but this does not work when an http proxy is involved. This was changed in https://github.com/crc-org/crc/commit/52279b3eaf7c7b , details can be found in the commit log.

We could also have two different code paths for proxy/non-proxy, not sure this was discussed/considered at the time of this PR.

praveenkumar:
@vyasgun Can you also update the commit log which have details around why we are changing it?

vyasgun:
@praveenkumar Just updated -- is this commit message okay?

gbraad (approved):

gbraad:
This test does too much to be honest; hence my comments. This is supposed to be a DNS test, but has become proxy aware (as expected); but we do not actually error correctly to test if the proxy failed (to perform a remote lookup). The idea to move away to a clean 200 was to avoid involving more HTTP traffic; the redirect is a result of enforced HTTPS; eg HSTS. Which means that in the future, a possible certificate issue also can cause issues. Let's just leave it at this for now.

gbraad:
> We could also have two different code paths for proxy/non-proxy, not sure this was discussed/considered at the time of this PR.

We have discussed this, but we most likely need to be more strategic about what to test (especially dependencies), and how to report this back. If this test fails due to the plethora of aforementioned issues, we do not properly indicate that to the user (and Podman Desktop!).

PR #4479, opened 2024-11-22
fix (status) : Log info in CRC status for OKD preset (#3626)

rohanKanojia:
~:warning: This currently doesn't work as expected due to https://github.com/crc-org/crc/issues/4478 . It requires #4480 to me merged first~

**Fixes:** Issue #3626

**Relates to:** Issue #3626 

## Solution/Idea

Currently, `crc status` logs information only for OpenShift and MicroShift presets. We should also log cluster information for the OKD preset.

## Proposed changes

This PR only changes the way `crc status` logs information for `okd` preset. 

Old behavior of `crc status` on OKD preset (Nothing is logged for OpenShift version as there is no handling for `okd` preset): 
``` 
CRC VM:          Running
RAM Usage:       647.2MB of 10.95GB
Disk Usage:      21.48GB of 32.68GB (Inside the CRC VM)
Cache Usage:     67.13GB
Cache Directory: /home/rokumar/.crc/cache
```

Expected New behavior of `crc status` on OKD preset: 
```
CRC VM:          Running
OpenShift/OKD:       Running (v4.15.0-0.okd-2024-02-23-163410)
RAM Usage:       647.2MB of 10.95GB
Disk Usage:      21.48GB of 32.68GB (Inside the CRC VM)
Cache Usage:     67.13GB
Cache Directory: /home/rokumar/.crc/cache
```

## Testing
- Set up crc cluster
  - `crc config set preset okd`
  - `crc setup`
  - `crc start`
- Run `crc status` (should see `OpenShift/OKD` in status logs)

### DIFF
--- a/cmd/crc/cmd/status.go
+++ b/cmd/crc/cmd/status.go
@@ -195,12 +195,7 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
 		{"CRC VM", s.CrcStatus},
 	}
 
-	if s.Preset == preset.OpenShift {
-		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
-	}
-	if s.Preset == preset.Microshift {
-		lines = append(lines, line{"MicroShift", openshiftStatus(s)})
-	}
+	lines = append(lines, line{s.Preset.ForDisplay(), openshiftStatus(s)})
 
 	if s.RAMSize != -1 && s.RAMUsage != -1 {
 		lines = append(lines, line{"RAM Usage", fmt.Sprintf(
--- a/cmd/crc/cmd/status_test.go
+++ b/cmd/crc/cmd/status_test.go
@@ -190,3 +190,41 @@ Cache Directory: %s
 `
 	assert.Equal(t, fmt.Sprintf(expected, cacheDir), out.String())
 }
+
+func TestCrcStatusShouldLogInformationForConfiguredPresets(t *testing.T) {
+	tests := []struct {
+		name      string
+		preset    preset.Preset
+		statusLog string
+	}{
+		{"OpenShift preset should log OpenShift", preset.OpenShift, "OpenShift:       Running (v4.5.1)"},
+		{"OKD preset should log OKD", preset.OKD, "OKD:             Running (v4.5.1)"},
+		{"MicroShift preset should log MicroShift", preset.Microshift, "MicroShift:              Running (v4.5.1)"},
+		{"Unknown preset should log OpenShift", preset.ParsePreset("unknown"), "OpenShift:       Running (v4.5.1)"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Given
+			cacheDir := t.TempDir()
+			out := new(bytes.Buffer)
+			client := mocks.NewClient(t)
+			require.NoError(t, os.WriteFile(filepath.Join(cacheDir, "crc.qcow2"), make([]byte, 10000), 0600))
+			client.On("Status").Return(apiClient.ClusterStatusResult{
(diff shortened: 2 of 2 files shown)


cmd/crc/cmd/status.go
```diff
@@ -195,11 +195,15 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
...
-	if s.Preset == preset.OpenShift {
+	switch s.Preset {
+	case preset.OpenShift:
 		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
-	}
-	if s.Preset == preset.Microshift {
+	case preset.Microshift:
 		lines = append(lines, line{"MicroShift", openshiftStatus(s)})
+	case preset.OKD:
+		lines = append(lines, line{"OpenShift/OKD", openshiftStatus(s)})
```
praveenkumar: Is there any reason why we need to use `OpenShift/OKD` instead just `OKD` ?

cmd/crc/cmd/status.go
```diff
@@ -195,11 +195,15 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
...
+	default:
+		panic(fmt.Sprintf("unknown preset provided : %s (expecting OpenShift, MicroShift, or OKD) ", s.Preset))
```
praveenkumar: Also this change is only on cmd side, did you check this looks similar when get this info using `/status` api call using daemon?

cmd/crc/cmd/status.go
rohanKanojia: Hello, as per the discussion in https://github.com/crc-org/crc/issues/3626#issuecomment-2493472293

I thought of adding this as a suggestion from @cfergeau  as it seemed to both group OpenShift and OKD together along with  giving user an idea a different preset is in use.

cmd/crc/cmd/status.go
rohanKanojia: On crc daemon side, I see that we default to OpenShift preset :

https://github.com/crc-org/crc/blob/1af16203e10253dcb829a9d62fa45bc03f7f4c02/pkg/crc/machine/status.go#L59-L66

Shall I update cmd side logic to make it consistent with this?

rohanKanojia: I've updated it to align with `/status` api call logic.

cmd/crc/cmd/status.go
praveenkumar: > Hello, as per the discussion in [#3626 (comment)](https://github.com/crc-org/crc/issues/3626#issuecomment-2493472293)

This comment says if we want to group OpenShift and OKD both then better to use `OpenShift/OKD` otherwise just use `OKD` and in the code we have default for OpenShift preset.

rohanKanojia: Sorry, I assumed it was okay to proceed forward with any of the suggested options.

I'll update it to use `OKD` only.

cmd/crc/cmd/status.go
```diff
@@ -195,11 +195,13 @@ func (s *status) prettyPrintTo(writer io.Writer) error {
...
-		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
-	}
-	if s.Preset == preset.Microshift {
+	switch s.Preset {
+	case preset.Microshift:
 		lines = append(lines, line{"MicroShift", openshiftStatus(s)})
+	case preset.OKD:
+		lines = append(lines, line{"OKD", openshiftStatus(s)})
+	default:
+		lines = append(lines, line{"OpenShift", openshiftStatus(s)})
```
cfergeau: `lines = append(lines, line{s.Preset.ForDisplay(), openshiftStatus(s)})` should do the trick.

PR #4487, opened 2024-11-27
Add devcontainer

gbraad:
This can help with quick tests and fixes by users and external contributors.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,11 @@
+{
+    "name": "CRC environment",
+    "image": "registry.access.redhat.com/ubi9/go-toolset:latest",
+
+    "customizations": {
+        "vscode": {
+            "extensions": [
+            ]
+        }
+    }
+}
--- a/.github/workflows/verify-devcontainer.yml
+++ b/.github/workflows/verify-devcontainer.yml
@@ -1,0 +1,14 @@
+name: Verify make cross in devcontainer
+on:
+  push:
+    branches:
+      - "main"
+  pull_request: {}
+jobs:
+  verify-devcontainer:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Run `git clone` and `make cross` against devcontainer
+        run: >
+          podman run registry.access.redhat.com/ubi9/go-toolset:latest \
+            sh -c "git clone https://github.com/crc-org/crc --depth 1 && cd crc && make cross"


slemeur:
@gbraad : Nice! Does it work with podman?

.devcontainer/Containerfile
```diff
@@ -0,0 +1,8 @@
+FROM registry.access.redhat.com/ubi8/go-toolset:1.22
```
praveenkumar: should we use ubi8 or ubi9 ? Also does it have by default make tools?

cfergeau:
> This can help with quick tests and fixes by users and external contributors.

Can you add this explanation to the commit log?

gbraad:
I usually use [devpod](https://devpod.sh) with podman as provider. but otherwise just replace dockerpath with podman in tools like vscode.

.devcontainer/Containerfile
gbraad: Yes, it does have

gbraad:
> Can you add this explanation to the commit log?

Do you think that information helps with a `git bisect` ? :-o

.devcontainer/Containerfile
gbraad: @praveenkumar I am OK with either option. I have been building on my machine with this setup for a long time now; never needed to target EL8, actually. As long as it does not cause issues with `glibc` I can switch to UBI9.

cfergeau:
> > Can you add this explanation to the commit log?
> 
> Do you think that information helps with a `git bisect` ? :-o

No, but it will be useful tosomeone wondering why we have a devcontainer file, what was our intent when adding it, ... and which checks with `git log`

gbraad:
> ubi9

I changed to `ubi9:latest`
https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?container-tabs=gti

~I dislike the more specific version tag.~

~but for some reason this does not start the codespaces properly.~ => this was caused by a restart that considered my dotfiles; error due to `command not found`.

gbraad:
Any other issues?
image

praveenkumar:
Do we need to consume this for any CI (GH action) to make sure it is working as expected ?

gbraad:
Actions and devcontainers are functionally different  though actions can
start the devcontainer; I do not use them as such now. can be an
improvement. still awaiting a discussion to move this forward as the way to
build on Windows. blog, docs and engineering-docs reuse the same container
with only a few additions. follow-up!?

---

https://github.com/crc-org/crc/actions/workflows/verify-devcontainer.yml

cfergeau:
Ah, the devcontainer provides easy integration with vscode, which ease contributions for people using it?

gbraad:
... One of the ways is to use CodeSpaces, and then connect your IDE of choice, like GoLand: https://docs.github.com/en/codespaces/developing-in-a-codespace/using-github-codespaces-in-your-jetbrains-ide

You can otherwise run locally with vscode, a tool like devpod.sh, or as JetBrains describes this as follows:
-  https://www.jetbrains.com/help/idea/start-dev-container-from-welcome-screen.html

Another feature is to allow review using 'Review in codespace', which also use the devcontainers definition:

image

---

image

.github/workflows/verify-devcontainer-base.yml
```diff
@@ -0,0 +1,11 @@
+name: Verify make cross in devcontainer base
+on:
+  workflow_dispatch:
+jobs:
+  publish_release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run `make cross` against devcontainer base
+        run: |
+        docker run registry.access.redhat.com/ubi9/go-toolset:latest \
```
cfergeau: I'd run `podman build` against `.devcontainer/Containerfile`, and then use the built image with `sh -c "git clone https://github.com/crc-org/crc && cd crc && make cross"`.
`registry.access.redhat.com/ubi9/go-toolset:latest` will be missing some tools required for building crc, or we would not need this intermediate image.

.devcontainer/Containerfile
```diff
@@ -0,0 +1,8 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:latest
+
+USER root
+
+RUN dnf install -y \
+        sudo git-core \
+    && dnf clean all \
+    && rm -rf /var/cache/yum
```
cfergeau: git is already installed on these images, make is missing

.github/workflows/verify-devcontainer-base.yml
gbraad: the only thing that devcontainer needs is a reinstall of git-core for some reason. this is already part of the actual image, so not sure why this caused issues with this not being found for GitHub Codespaces.

I would prefer not to use intermediate containers that gets published. but you mean to just perform a local build `-t devcontainer` and use that for the `docker run`, right? Note: one of the reasons I would prefer this is to avoid recording the base version number in multiple places.

.devcontainer/Containerfile
gbraad: `make` is included, but `git-core` needed a (re)install for some reason, else Codespaces would not find the `git` binary.

```
$ rpm -qa make
make-4.3-8.el9.x86_64
```

```
$ dnf history
...
     3 | install -y --setopt=tsflags=nodocs go-toolset                                                                                                                                             | 2024-12-02 17:35 | Install        |    7
     2 | install -y --setopt=tsflags=nodocs autoconf automake bzip2 gcc-c++ gd-devel gdb git libcurl-devel libpq-devel libxml2-devel libxslt-devel lsof make mariadb-connector-c-devel openssl-dev | 2024-12-02 14:29 | Install        |  213
     1 | install -y --setopt=tsflags=nodocs bsdtar findutils groff-base glibc-locale-source glibc-langpack-en gettext rsync scl-utils tar unzip xz yum                                             | 2024-12-02 06:15 | Install        |   17
```

.github/workflows/verify-devcontainer-base.yml
gbraad: using `ubuntu-24.04`
```shell
        podman build -t devcontainer -f .devcontainer/Containerfile .
        podman run devcontainer \
          sh -c "git clone https://github.com/crc-org/crc --depth 1 && cd crc && make cross"
```

.devcontainer/Containerfile
gbraad: It might also be that in the more recent image, like `registry.access.redhat.com/ubi9/go-toolset:latest`, this is now part of this image. This was not the case of ubi8 and also not for slightly earlier released image (this latest is now from 5 days ago). this also coincides with your comment... before that it was not the case.

gbraad:
The go-toolset image has been updated and now comes with the necessary `git` command. This makes the intermediate container not necessary anymore.

https://github.com/crc-org/crc/actions/workflows/verify-devcontainer.yml

gbraad:
Will create a documentation follow-up

.github/workflows/verify-devcontainer-base.yml
cfergeau: Yes this is what I was suggesting, as github action would be testing exactly the image which will be used by devcontainer users.

.devcontainer/Containerfile
cfergeau: yeah, realised today I forgot to pull the latest image, mine even had an outdated go version and failed to build crc :)

PR #4490, opened 2024-11-29
Move docs out of repo

gbraad:
The user documentation has moved to a new repository at https://github.com/crc-org/docs

Related to: https://github.com/crc-org/crc-internal/issues/91


.gitpod.yml
```diff
@@ -0,0 +1,11 @@
...
+# Please adjust to your needs (see https://www.gitpod.io/docs/introduction/learn-gitpod/gitpod-yaml)
+# and commit this file to your remote git repository to share the goodness with others.
+
+# Learn more from ready-to-use templates: https://www.gitpod.io/docs/introduction/getting-started/quickstart
+
+tasks:
+  - init: go get && go build ./... && go test ./... && make
+    command: go run .
+
+
```
praveenkumar: Do we really need this file to be part of this commit, this is more like your local dev setup.

gbraad: This is not local dev setup, but was not suppoed to be checked in.

gbraad:
Documentation is already publishing to https://crc.dev/docs, so this functionality is not needed anymore; please merge soon to prevent conflicting index by search engines

gbraad:
👍, also updated the blog to point to the "Getting Started Guide"

PR #4493, opened 2024-12-02
CI: update macOS-12 to macOS-14

praveenkumar:
We are not supporting macOS-12 anymore so better to switch to 14.

### DIFF
--- a/.github/workflows/macos-installer.yml
+++ b/.github/workflows/macos-installer.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         os:
           - macOS-13
-          - macOS-12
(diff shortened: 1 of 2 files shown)


cfergeau (commented):
More importantly, github is dropping support for the macos 12 runners, they will error out very soon.
Do they already have macos 15 runners? We could add these at the same time.

praveenkumar:
> Do they already have macos 15 runners? We could add these at the same time.

They have but in beta stage as of now, once it available, we can migrate 13 => 15.

cfergeau (approved):

PR #4504, opened 2024-12-06
[WIP] bundle: Parallel download and decompression

### LINKED ISSUE #4336: Parallel bundle download & decompression
I think it might be slightly faster if bundle download and decompression would be done in parallel. IIUIC right now bundle is first downloaded and only then decompressed.

/kind enhancement

vyasgun:
## Description

This pull request does the following:
- Return a reader from the bundle Download function.
- Use the reader to stream the bytes to Extract function.

This commit replaces grab client with the net/http client to ensure that the bytes are streamed come in correct order to the Extract func. Currently, only zst decompression is being used in the UncompressWithReader function as it is the primary compression algorithm being used in crc.

The download progress bar has been removed temporarily and will be added back as part of refactoring the code.

Fixes: #4336 

## Type of change

## Proposed changes

- Return a reader from the bundle Download function.
- Use the reader to stream the bytes to Extract function.
## Contribution Checklist
- [ ] I tested my code on specified platforms 
    - [ ] Linux
    - [ ] Windows
    - [x] MacOS

### DIFF
--- a/cmd/crc-embedder/cmd/embed.go
+++ b/cmd/crc-embedder/cmd/embed.go
@@ -163,7 +163,7 @@ func downloadDataFiles(goos string, components []string, destDir string) ([]stri
 		if !shouldDownload(components, componentName) {
 			continue
 		}
-		filename, err := download.Download(context.TODO(), dl.url, destDir, dl.permissions, nil)
+		_, filename, err := download.Download(context.TODO(), dl.url, destDir, dl.permissions, nil)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/crc/image/image.go
+++ b/pkg/crc/image/image.go
@@ -74,6 +74,7 @@ func (img *imageHandler) copyImage(ctx context.Context, destPath string, reportW
 	if ctx == nil {
 		panic("ctx is nil, this should not happen")
 	}
+
 	manifestData, err := copy.Image(ctx, policyContext,
 		destRef, srcRef, &copy.Options{
 			ReportWriter: reportWriter,
--- a/pkg/crc/machine/bundle/metadata.go
+++ b/pkg/crc/machine/bundle/metadata.go
@@ -344,43 +344,49 @@ func getVerifiedHash(url string, file string) (string, error) {
 	return "", fmt.Errorf("%s hash is missing or shasums are malformed", file)
 }
 
-func downloadDefault(ctx context.Context, preset crcPreset.Preset) (string, error) {
+func downloadDefault(ctx context.Context, preset crcPreset.Preset) (io.Reader, string, error) {
 	downloadInfo, err := getBundleDownloadInfo(preset)
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 	return downloadInfo.Download(ctx, constants.GetDefaultBundlePath(preset), 0664)
 }
 
-func Download(ctx context.Context, preset crcPreset.Preset, bundleURI string, enableBundleQuayFallback bool) (string, error) {
+func Download(ctx context.Context, preset crcPreset.Preset, bundleURI string, enableBundleQuayFallback bool) (io.Reader, string, error) {
 	// If we are asked to download
 	// ~/.crc/cache/crc_podman_libvirt_4.1.1.crcbundle, this means we want
 	// are downloading the default bundle for this release. This uses a
 	// different codepath from user-specified URIs as for the default
 	// bundles, their sha256sums are known and can be checked.
+	var reader io.Reader
 	if bundleURI == constants.GetDefaultBundlePath(preset) {
 		switch preset {
 		case crcPreset.OpenShift, crcPreset.Microshift:
-			downloadedBundlePath, err := downloadDefault(ctx, preset)
+			var err error
+			var downloadedBundlePath string
+			reader, downloadedBundlePath, err = downloadDefault(ctx, preset)
 			if err != nil && enableBundleQuayFallback {
 				logging.Info("Unable to download bundle from mirror, falling back to quay")
-				return image.PullBundle(ctx, constants.GetDefaultBundleImageRegistry(preset))
+				bundle, err := image.PullBundle(ctx, constants.GetDefaultBundleImageRegistry(preset))
+				return nil, bundle, err
 			}
-			return downloadedBundlePath, err
+			return reader, downloadedBundlePath, err
 		case crcPreset.OKD:
 			fallthrough
 		default:
-			return image.PullBundle(ctx, constants.GetDefaultBundleImageRegistry(preset))
+			bundle, err := image.PullBundle(ctx, constants.GetDefaultBundleImageRegistry(preset))
+			return nil, bundle, err
 		}
 	}
 	switch {
 	case strings.HasPrefix(bundleURI, "http://"), strings.HasPrefix(bundleURI, "https://"):
 		return download.Download(ctx, bundleURI, constants.MachineCacheDir, 0644, nil)
 	case strings.HasPrefix(bundleURI, "docker://"):
-		return image.PullBundle(ctx, bundleURI)
+		bundle, err := image.PullBundle(ctx, bundleURI)
+		return nil, bundle, err
 	}
 	// the `bundleURI` parameter turned out to be a local path
-	return bundleURI, nil
+	return reader, bundleURI, nil
 }
 
 type Version struct {
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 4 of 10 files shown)


pkg/download/download.go
```diff
@@ -2,97 +2,66 @@ package download
...
-	req, err := grab.NewRequest(destination, uri)
+	req, err := http.NewRequestWithContext(ctx, "GET", uri, nil)
+
 	if err != nil {
-		return "", errors.Wrapf(err, "unable to get request from %s", uri)
+		return nil, "", errors.Wrapf(err, "unable to get request from %s", uri)
 	}
+	client := http.Client{Transport: &http.Transport{}}
 
 	if ctx == nil {
```
redbeam: This check might need to be moved higher, as I see that `http.NewRequestWithContext(ctx, "GET", uri, nil)` is already called earlier (and if ctx is null, might produce errors).

pkg/crc/machine/bundle/metadata.go
```diff
@@ -344,15 +344,16 @@ func getVerifiedHash(url string, file string) (string, error) {
...
 	if err != nil {
-		return "", err
+		return nil, "", err
 	}
 	return downloadInfo.Download(ctx, constants.GetDefaultBundlePath(preset), 0664)
 }
 
-func Download(ctx context.Context, preset crcPreset.Preset, bundleURI string, enableBundleQuayFallback bool) (string, error) {
+func Download(ctx context.Context, preset crcPreset.Preset, bundleURI string, enableBundleQuayFallback bool) (io.Reader, string, error) {
+	var reader io.Reader
```
redbeam: I think this line is kind of "hidden" here above the big comment, I suggest moving it down below it.

pkg/crc/image/image.go
```diff
@@ -116,43 +116,43 @@ func GetPresetName(imageName string) crcpreset.Preset {
 	return preset
 }
 
-func PullBundle(ctx context.Context, imageURI string) (string, error) {
+func PullBundle(ctx context.Context, imageURI string) (io.Reader, string, error) {
```
redbeam: Would it be possible to add support for pulling from container image repositories as well?

pkg/crc/machine/bundle/repository.go
```diff
@@ -198,8 +229,14 @@ func Use(bundleName string) (*CrcBundleInfo, error) {
 	return defaultRepo.Use(bundleName)
 }
 
-func Extract(ctx context.Context, path string) (*CrcBundleInfo, error) {
-	if err := defaultRepo.Extract(ctx, path); err != nil {
+func Extract(ctx context.Context, reader io.Reader, path string) (*CrcBundleInfo, error) {
```
redbeam, line 232: We could use just one return statement and one error check in this case:

```go
var err error
if reader == nil {
	err = defaultRepo.Extract(ctx, path)
} else {
	err = defaultRepo.ExtractWithReader(ctx, reader, path)
}

if err != nil {
	return nil, err
}
return defaultRepo.Get(filepath.Base(path))
```

Add more blank lines as you see fit.

pkg/crc/machine/bundle/repository.go
```diff
@@ -124,6 +125,36 @@ func (bundle *CrcBundleInfo) createSymlinkOrCopyPodmanRemote(binDir string) erro
 	return bundle.copyExecutableFromBundle(binDir, PodmanExecutable, constants.PodmanRemoteExecutableName)
 }
 
+func (repo *Repository) ExtractWithReader(ctx context.Context, reader io.Reader, path string) error {
```
redbeam, line 128: This function and `Extract` are very similar, could they be merged in some way?

pkg/extract/extract.go
```diff
@@ -32,6 +32,21 @@ func Uncompress(ctx context.Context, tarball, targetDir string) ([]string, error
...
+
+func uncompressWithReader(ctx context.Context, reader io.Reader, targetDir string, fileFilter func(string) bool, showProgress bool) ([]string, error) {
+	logging.Debugf("Uncompressing from reader to %s", targetDir)
+
+	reader, err := zstd.NewReader(reader)
+	if err != nil {
+		return nil, err
+	}
+	return untar(ctx, reader, targetDir, fileFilter, showProgress)
+
```
redbeam: nitpick: empty line 47

pkg/extract/extract.go
```diff
@@ -86,6 +101,9 @@ func uncompress(ctx context.Context, tarball, targetDir string, fileFilter func(
 	}
 }
 
+func Untar(ctx context.Context, reader io.Reader, targetDir string, fileFilter func(string) bool, showProgress bool) ([]string, error) {
```
redbeam: I suppose this is for some future functionality?

redbeam (commented):
Overall, this is great work and it's functional.

My findings:

- cancelling through web api (socket) works

- better logging would be nice (currently it's skipping the download part) so that it's clear that the download and uncompression is being done simultaneously

- progress bar could show more info about both processes

- resuming interrupted download doesn't work - everything starts from the beginning

- golangci-lint issues

Suggestions:

- add (cli/config) option to disable this functionality (revert back to old behavior)

pkg/crc/image/image.go
cfergeau: I'm not sure I understand your question, `PullBundle` pulls from a container image registry.

pkg/crc/machine/bundle/repository.go
cfergeau: `Extract` could probably use [`os.Open`](https://pkg.go.dev/os#Open) and call into `ExtractWithReader`

cmd/crc-embedder/cmd/embed.go
```diff
@@ -163,7 +163,7 @@ func downloadDataFiles(goos string, components []string, destDir string) ([]stri
 		if !shouldDownload(components, componentName) {
 			continue
 		}
-		filename, err := download.Download(context.TODO(), dl.url, destDir, dl.permissions, nil)
+		_, filename, err := download.Download(context.TODO(), dl.url, destDir, dl.permissions, nil)
```
cfergeau, line 166: I think I'd add a `download.DownloadFile(…) (string, error)` to make it clear when we don't need the reader.

pkg/crc/image/image.go
cfergeau: If `PullBundle` never returns a bundle, I would not change its signature, and do something like this in `Download` instead of `return image.PullBundle(ctx, bundleURI)`
```
path, err := image.PullBundle(ctx, bundleURI)
return nil, path, err
```

But maybe you have plans to make it return a reader later?

pkg/crc/machine/bundle/metadata.go
cfergeau: The variable does not seem to be required in this block, the last line of the function can return `nil` instead of `reader`

pkg/crc/machine/bundle/repository.go
cfergeau: I was about to make the same suggestion

pkg/crc/machine/start.go
```diff
@@ -48,13 +48,15 @@ func getCrcBundleInfo(ctx context.Context, preset crcPreset.Preset, bundleName,
...
 	logging.Infof("Downloading bundle: %s...", bundleName)
-	bundlePath, err = bundle.Download(ctx, preset, bundlePath, enableBundleQuayFallback)
+	reader, bundlePath, err := bundle.Download(ctx, preset, bundlePath, enableBundleQuayFallback)
 	if err != nil {
 		return nil, err
 	}
+
 	logging.Infof("Extracting bundle: %s...", bundleName)
-	if _, err := bundle.Extract(ctx, bundlePath); err != nil {
+	if _, err := bundle.Extract(ctx, reader, bundlePath); err != nil {
```
cfergeau, line 59: Having a `bundlePath` and a `reader` feels a bit redundant, ideally we could pass one or the other, but I'm not sure it is currently that easy.

pkg/download/download.go
vyasgun: Done 👍

pkg/crc/machine/bundle/metadata.go
vyasgun: @redbeam Moved it below the comment. This variable is required in case [downloadDefault](https://github.com/crc-org/crc/pull/4504/files/3a62d1a66988b156a421e2dbeddf586b6d16c730#diff-705d9d172db8eb4971f26dc3fd2ebb8a0b73027e3a6a2fceb1c7ee11260075abR367) returns a reader.

pkg/crc/machine/bundle/repository.go
vyasgun: @redbeam Thanks for the suggestion - changed it! 😄

pkg/extract/extract.go
vyasgun: No, this was for testing purposes which I then forgot about. Nice catch!

pkg/crc/image/image.go
vyasgun: I changed the signature back since I don't have any plans yet to implement a reader with `pull bundle`

redbeam: > I'm not sure I understand your question, PullBundle pulls from a container image registry.

@cfergeau I should've been clearer, sorry. What I mean is adding simultaneous download & decompression when downloading from a container image registry. But Gunjan already answered this ;)

PR #4514, opened 2024-12-16
Change linux default network-mode to user

lilyLuLiu:
Currently, the Linux default network mode is system. But many issues happen due to different system situations. 
For most user, user network mode already satisfied their demands. And user network mode no need to deal with complex system situations. 
So after discuss, it's better to set user as default network-mode, making sure CRC works well in default.  
If the user requires more functions, they can switch to system network mode. 

@gbraad @praveenkumar @anjannath have agreed on this being needed. 
It needs to start testing early during December.
In the worst case, we can revert before the new release in January.

### DIFF
--- a/pkg/crc/network/types.go
+++ b/pkg/crc/network/types.go
@@ -2,7 +2,6 @@ package network
 
 import (
 	"fmt"
-	"runtime"
 
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
 	"github.com/spf13/cast"
@@ -55,7 +54,7 @@ func parseMode(input string) (Mode, error) {
 	case string(SystemNetworkingMode), "default":
 		return SystemNetworkingMode, nil
 	default:
-		return SystemNetworkingMode, fmt.Errorf("Cannot parse mode '%s'", input)
+		return UserNetworkingMode, fmt.Errorf("Cannot parse mode '%s'", input)
 	}
 }
 func ParseMode(input string) Mode {
@@ -68,12 +67,7 @@ func ParseMode(input string) Mode {
 }
 
 func getDefaultMode() Mode {
-	switch runtime.GOOS {
-	case "linux":
-		return SystemNetworkingMode
-	default:
-		return UserNetworkingMode
-	}
+	return UserNetworkingMode
 }
 
 func ValidateMode(val interface{}) (bool, string) {
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -213,7 +213,6 @@ func InitializeScenario(s *godog.ScenarioContext) {
 			}
 
 			if tag.Name == "@proxy" {
-
 				// start container with squid proxy
 				err := util.ExecuteCommand("podman run --name squid -d -p 3128:3128 quay.io/crcont/squid")
 				if err != nil {
@@ -221,13 +220,13 @@ func InitializeScenario(s *godog.ScenarioContext) {
 					os.Exit(1)
 				}
 
-				err = util.ExecuteCommand("crc config set http-proxy http://192.168.130.1:3128")
+				err = util.ExecuteCommand("crc config set http-proxy http://host.crc.testing:3128")
 				if err != nil {
 					fmt.Println(err)
 					os.Exit(1)
 				}
 
-				err = util.ExecuteCommand("crc config set https-proxy http://192.168.130.1:3128")
+				err = util.ExecuteCommand("crc config set https-proxy http://host.crc.testing:3128")
 				if err != nil {
 					fmt.Println(err)
 					os.Exit(1)
@@ -239,6 +238,20 @@ func InitializeScenario(s *godog.ScenarioContext) {
 					os.Exit(1)
 				}
 
+				err = util.ExecuteCommand("crc config set host-network-access true")
+				if err != nil {
+					fmt.Println(err)
+					os.Exit(1)
+				}
+
+			}
+
+			if tag.Name == "@system_network" {
+				err = util.ExecuteCommand("crc config set network-mode system")
+				if err != nil {
+					fmt.Println(err)
+					os.Exit(1)
+				}
 			}
 		}
 
@@ -353,6 +366,14 @@ func InitializeScenario(s *godog.ScenarioContext) {
 				}
 
 				if err := os.Unsetenv("NO_PROXY"); err != nil {
+					fmt.Println(err)
+					os.Exit(1)
+				}
+			}
+
+			if tag.Name == "@system_network" {
+				err := util.ExecuteCommand("crc config unset network-mode")
+				if err != nil {
 					fmt.Println(err)
 					os.Exit(1)
 				}
@@ -960,9 +981,6 @@ func PodmanCommandIsAvailable() error {
 		path = os.ExpandEnv(unexpandedPath)
 		csshk = filepath.Join(userHomeDir, ".crc/machines/crc/id_ed25519")
 		dh = "npipe:////./pipe/crc-podman"
-	}
-	if runtime.GOOS == "linux" {
-		ch = "ssh://core@192.168.130.11:22/run/user/1000/podman/podman.sock"
 	}
 
 	os.Setenv("PATH", path)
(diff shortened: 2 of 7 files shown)


rohanKanojia:
@lilyLuLiu : Hello, Could you please provide some context on why this change is needed?

test/e2e/testsuite/testsuite.go
```diff
@@ -961,9 +961,11 @@ func PodmanCommandIsAvailable() error {
 		csshk = filepath.Join(userHomeDir, ".crc/machines/crc/id_ed25519")
 		dh = "npipe:////./pipe/crc-podman"
 	}
+	/* change linux default to user
 	if runtime.GOOS == "linux" {
 		ch = "ssh://core@192.168.130.11:22/run/user/1000/podman/podman.sock"
 	}
+	*/
```
praveenkumar: @lilyLuLiu May be you should remove that completely instead just commenting out?

test/e2e/features/proxy.feature
```diff
@@ -1,10 +1,10 @@
-@proxy @linux
+@proxy @linux
```
praveenkumar: This change is just extra space so better to remove it.

praveenkumar:
@lilyLuLiu About integration failure I think https://github.com/crc-org/crc/blob/main/pkg/crc/network/types.go#L65 is culprit because settings only apply if we build binary for non installer, we need to change the default here also

pkg/crc/network/types.go
```diff
@@ -68,12 +67,7 @@ func ParseMode(input string) Mode {
 }
 
 func getDefaultMode() Mode {
-	switch runtime.GOOS {
-	case "linux":
-		return SystemNetworkingMode
-	default:
-		return UserNetworkingMode
-	}
+	return UserNetworkingMode
```
praveenkumar, line 70: This change should be part of first commit.

test/e2e/testsuite/testsuite.go
```diff
@@ -961,11 +977,6 @@ func PodmanCommandIsAvailable() error {
 		csshk = filepath.Join(userHomeDir, ".crc/machines/crc/id_ed25519")
 		dh = "npipe:////./pipe/crc-podman"
 	}
-	/* change linux default to user
-	if runtime.GOOS == "linux" {
-		ch = "ssh://core@192.168.130.11:22/run/user/1000/podman/podman.sock"
-	}
-	*/
```
praveenkumar: May be commenting out of this shouldn't be part of first commit and this commit remove it completely.

test/e2e/testsuite/testsuite.go
```diff
@@ -213,9 +213,8 @@ func InitializeScenario(s *godog.ScenarioContext) {
 			}
 
 			if tag.Name == "@proxy" {
-
 				// start container with squid proxy
-				err := util.ExecuteCommand("podman run --name squid -d -p 3128:3128 quay.io/crcont/squid")
+				err := util.ExecuteCommand("sudo dnf install podman -y;podman run --name squid -d -p 3128:3128 quay.io/crcont/squid")
```
praveenkumar: This commit should be merged to second one.

test/e2e/features/proxy.feature
```diff
@@ -4,7 +4,7 @@ Feature: Behind proxy test
     Check CRC use behind proxy
 
     # inherits @proxy tag from Feature
-    @cleanup
+    @cleanup @system_network
```
praveenkumar: @lilyLuLiu Does this mean to run the proxy tet on linux the network mode should be system?

test/e2e/testsuite/testsuite.go
praveenkumar: why we need to install podman now but it was not the case before?

lilyLuLiu: Previous Linux tests are on our test machines, the podman is already installed. The new added pr checks on linux-arm, machines are reserved from TF, no pre-install podman.

test/e2e/features/proxy.feature
lilyLuLiu: @praveenkumar You are right, this test case also applied to user mode. I have removed the tag.

test/e2e/testsuite/testsuite.go
```diff
@@ -213,21 +213,20 @@ func InitializeScenario(s *godog.ScenarioContext) {
...
 				if err != nil {
 					fmt.Println(err)
 					os.Exit(1)
 				}
 
-				err = util.ExecuteCommand("crc config set http-proxy http://192.168.130.1:3128")
+				err = util.ExecuteCommand("crc config set http-proxy http://127.0.0.1:8888")
```
praveenkumar: this will not work, I think this test should be same as windows/mac.

test/e2e/testsuite/testsuite.go
praveenkumar: install podman shouldn't be part of test but it should be machine provision.

test/e2e/testsuite/testsuite.go
lilyLuLiu: Change to http://host.crc.testing:3128 has we discussed.

test/e2e/testsuite/testsuite.go
lilyLuLiu: Removed this change and will add "podman installation" in GitHub action codes.

karelyatin:
Hi with latest CRC release including this commit crc not starting, is that expected?
$ cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.4 (Plow)
$ crc version
CRC version: 2.47.0+457b65
OpenShift version: 4.17.14
MicroShift version: 4.17.14

$ crc config view
- cpus                                  : 24
- disk-size                             : 100
- kubeadmin-password                    : 12345678
- memory                                : 48000
- pull-secret-file                      : /home/ocp_dev/install_yamls/devsetup/pull-secret.txt
- skip-check-daemon-systemd-sockets     : true
- skip-check-daemon-systemd-unit        : true


INFO Uncompressing /home/ocp_dev/.crc/cache/crc_libvirt_4.17.14_amd64.crcbundle 
crc.qcow2:  19.43 GiB / 19.43 GiB [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
oc:  158.81 MiB / 158.81 MiB [-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00%
+ /home/ocp_dev/bin/crc start                                     
Is 'crc daemon' running? Cannot reach daemon API: Get "http://unix/api/version": dial unix /home/ocp_dev/.crc/crc-http.sock: connect: no such file or directory
+ /home/ocp_dev/bin/crc console --credentials                     
Get "http://unix/api/webconsoleurl": dial unix /home/ocp_dev/.crc/crc-http.sock: connect: no such file or directory
++ /home/ocp_dev/bin/crc oc-env                   

With explicitly setting network mode i t works fine:-
$ crc config set network-mode system
$ crc setup
$ crc start

praveenkumar:
@karelyatin did you use `crc cleanup` before using this latest version of CRC? can you try following and see if that works?
```
$ crc delete
$ crc cleanup
$ crc setup
```

Also you need to unset following settings
```
crc config unset skip-check-daemon-systemd-sockets
crc config unset skip-check-daemon-systemd-unit
```

karlyan:
This break your own document: https://crc.dev/docs/networking/#setting-up-on-a-remote-server
It looks like in linux you change ip from 192.168.130.1 to 127.0.0.1. The haproxy can no longer bind with `bind 0.0.0.0:6443`

gbraad:
You can still use system mode.
This documented scenario using haproxy is a semi-supported use case. If you
have a fix, please let us know or otherwise switch network mode

For the document I suggest a note about switching network mode. We do not
have an automated test in place at the moment. I see this being for
advanced use, especially as network setups can greatly differ and
complicated this.

gbraad:
I filed: https://github.com/crc-org/docs/issues/26 and made a quick-fix in https://github.com/crc-org/docs/issues/25

---

```bash
$ crc config set network-mode system
```

karlyan:
Thanks for the response. 
I like the current user mode. With this new fix I don't need to install haproxy and setup all these proxy config. In our case; we have a pipelline to create crc instance on gcp vm. Later we use ssh tunnel to forward the traffic to local mac machine. With the new user network; I can remove the haproxy code. Just the doc needs to change a little bit.

gbraad:
We will update the document, but have no indication when this would happen.
So, good to hear using ssh forwards works for you.

PR #4526, opened 2024-12-20
fix (shell) : Improve CRC shell detection on unix environments (#3767)

### LINKED ISSUE #3797: improve shell detection for crc.exe running in WSL
### General information
  * OS: Windows

Currently when running `crc.exe podman-env` or `crc.exe oc-env` from a WSL terminal, it detects the shell as `cmd`

```
annant@DESKTOP-J8H86J0:~$ uname -a
Linux DESKTOP-J8H86J0 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 GNU/Linux
annant@DESKTOP-J8H86J0:~$ crc.exe podman-env
SET PATH=C:\Users\anath\.crc\bin\oc;%PATH%
SET CONTAINER_SSHKEY=C:\Users\anath\.crc\machines\crc\id_ecdsa
SET CONTAINER_HOST=ssh://core@127.0.0.1:2222/run/user/1000/podman/podman.sock
SET DOCKER_HOST=npipe:////./pipe/crc-podman
REM @FOR /f "tokens=*" %i IN ('crc podman-env') DO @call %i
```

we can override the detection to force a specific shell with the `--shell` flag (currently it doesn't include bash, zsh, fish as supported shells for windows) but it'd be good to auto-detect without having to use the "force-shell" flag

rohanKanojia:
## Description

Fixes: #3797

Relates to: #3797

- Currently, we rely on `SHELL` environment variable for detecting active shell type. This will work when the environment variable is set. As per my observations, this environment variable is not set explicitly by various Linux shell environments (`bash`,`zsh`,`fish`). We should detect the currently active shell by checking the active processes instead.
- While generating statements for export statements on Windows, we shall make sure that we have converted windows paths to Linux paths.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Proposed changes
+ Instead of marking `shell_test.go` to not run on Windows, move unix-specific tests to `shell_unix_test.go`, add generic platform-independent tests to `shell_test.go`
+ Add `bash` and `zsh` to the list of supported shell environments on Windows.
+ Instead of instructing user to use `--shell` flag to print shell specific instructions, detect shell automatically by inspecting currently open processes:
  - On Linux use `ps` command to get list of active shell processes and pick the most recent process id.
  - On WSL (Windows Subsystem for Linux), use `wsl` tool to execute linux `ps` process to get list of active shell processes and return the one with most recent process id.
+ In the presence of UNIX-like shell environments, convert Windows paths to UNIX paths.
   - If WSL (Windows Subsystem for Linux) is detected, use `wslpath` utility tool to perform this conversion.
   - For other scenarios (like case of Git Bash), simply convert windows path delimiters to linux and remove the windows drive prefix

With these changes, we should see the correct output of `crc podman-env` and `crc oc-env` on:
- Windows Subsystem for Linux
   - [Bash](https://en.wikipedia.org/wiki/Bash_(Unix_shell)) (default shell)
   - [fish](https://fishshell.com/) 
   - [zsh](https://en.wikipedia.org/wiki/Z_shell)
- Regular Linux distributions
  - [Bash](https://en.wikipedia.org/wiki/Bash_(Unix_shell)) (default shell)
  - [fish](https://fishshell.com/)
  - [zsh](https://en.wikipedia.org/wiki/Z_shell)

Running `crc oc-env` on Git Bash:
### Old Behavior
```
$ crc oc-env
SET PATH=C:\Users\rokum\.crc\bin\oc;%PATH%
REM @FOR /f "tokens=*" %i IN ('crc oc-env') DO @call %i
```
### New Behavior
```
$ /c/Users/rokum/go/bin/crc oc-env
export PATH="/C/Users/rokum/.crc/bin/oc:$PATH"
# eval $(crc oc-env)
```

## Testing
1. Start CRC cluster
2. Open some linux environment on windows (like Git Bash) and check result of `crc podman-env` / `crc oc-env` commands. They should not print output with respect to current shell in use.

## Contribution Checklist
- [X] I have read the [contributing guidelines](https://github.com/crc-org/crc/blob/main/developing.adoc)
- [X] My code follows the style guidelines of this project
- [X] I Keep It Small and Simple: The smaller the PR is, the easier it is to review and have it merged
- [X] I have performed a self-review of my code
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] I tested my code on specified platforms 
    - [x] Linux
    - [X] Windows
    - [ ] MacOS

### DIFF
--- a/pkg/os/shell/shell.go
+++ b/pkg/os/shell/shell.go
@@ -2,7 +2,16 @@ package shell
 
 import (
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
+
+	crcos "github.com/crc-org/crc/v2/pkg/os"
+)
+
+var (
+	CommandRunner                           = crcos.NewLocalCommandRunner()
+	WindowsSubsystemLinuxKernelMetadataFile = "/proc/version"
 )
 
 type Config struct {
@@ -65,9 +74,9 @@ func GetEnvString(userShell string, envName string, envValue string) string {
 	case "cmd":
 		return fmt.Sprintf("SET %s=%s", envName, envValue)
 	case "fish":
-		return fmt.Sprintf("contains %s $fish_user_paths; or set -U fish_user_paths %s $fish_user_paths", envValue, envValue)
+		return fmt.Sprintf("contains %s $fish_user_paths; or set -U fish_user_paths %s $fish_user_paths", convertToLinuxStylePath(userShell, envValue), convertToLinuxStylePath(userShell, envValue))
 	default:
-		return fmt.Sprintf("export %s=\"%s\"", envName, envValue)
+		return fmt.Sprintf("export %s=\"%s\"", envName, convertToLinuxStylePath(userShell, envValue))
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/os/shell/shell_unix.go
```diff
@@ -16,12 +15,11 @@ var (
 
 // detect detects user's current shell.
 func detect() (string, error) {
-	shell := os.Getenv("SHELL")
-
-	if shell == "" {
+	detectedShell := detectShellByInvokingCommand("", "ps", []string{"-o", "pid,comm"})
```
praveenkumar: I can see that in that function you are then trying to split the line and ignoring 0 index which can be done using following so we don't have to ignore it later. 

```
$ ps -o pid=,comm=
  57609 bash
 242709 ps
```

pkg/os/shell/shell.go
```diff
@@ -81,8 +91,85 @@ func GetPathEnvString(userShell string, prependedPath string) string {
...
+	}
+
+	detectedShell := inspectProcessOutputForRecentlyUsedShell(stdOut)
+	if detectedShell == "" {
+		return defaultShell
+	}
+	return detectedShell
+}
+
+func inspectProcessOutputForRecentlyUsedShell(psCommandOutput string) string {
```
praveenkumar, line 204: This function is doing lot of things, may be add a comment what actually this is doing with example so would be easy for review.

pkg/os/shell/shell.go
```diff
@@ -81,8 +91,85 @@ func GetPathEnvString(userShell string, prependedPath string) string {
...
+				processOutputs = append(processOutputs, ProcessOutput{
+					processID: parsedProcessID,
+					output:    lineParts[1],
+				})
+			}
+		}
+	}
+	sort.Slice(processOutputs, func(i, j int) bool {
+		return processOutputs[i].processID > processOutputs[j].processID
+	})
```
praveenkumar: why we are implementing this sorting function for slice?

pkg/os/shell/shell_unix.go
rohanKanojia: You're right. Thanks for the hint! I didn't know about this.

pkg/os/shell/shell.go
rohanKanojia: Thanks, I'll update it as per your comments.

pkg/os/shell/shell.go
rohanKanojia: Here we want to get the shell process with most recent process id (latest). Therefore I am sorting the arraylist in reverse order and getting first element for detecting currently active shell.

rohanKanojia: I've removed this sorting method. We can achieve same thing using `--sort` flag in `ps` command.

pkg/os/shell/shell.go
rohanKanojia: I've added comments to this method and other related methods.

pkg/os/shell/shell.go
```diff
@@ -81,8 +90,137 @@ func GetPathEnvString(userShell string, prependedPath string) string {
...
+//
+// - psCommandOutput: output of ps command executed on a particular shell session
+//
+// It returns:
+//
+//   - a string value (one of `zsh`, `bash` or `fish`) for current shell environment in use. If it's not able to determine
+//     underlying shell type, it returns and empty string.
+//
+// This method tries to check all processes open and filters out shell sessions (one of `zsh`, `bash` or `fish)
+// It then tries to sort these processes by process ids and returns the most recent one.
```
praveenkumar: In this method you are not performing any sorting because the `ps` command which is used already performing the sort with pid so better to update that info. https://play.golang.com/p/QhNeqa5tnEO

rohanKanojia: @praveenkumar : Sorry, I forgot to update the method comment after making changes to the method.

Should be okay now.

gbraad:
As mentioned in the first message, and related to #4537:

---
- [ ] I tested my code on specified platforms 
    - [x] Linux
    - [X] Windows
    - [ ] MacOS
---

Have tests on MacOS been performed by reviewer(s)?

praveenkumar:
> Have tests on MacOS been performed by reviewer(s)?

No, I did the review and didn't perform any test on mac for same, my understanding was since it have unit tests which run for each platform but those were mocked one :(

rohanKanojia:
@praveenkumar @gbraad : I apologize for introducing this regression. In unit tests, I had mocked the command output that's why they were not able to catch this failure :-(

cfergeau:
>  use ps command to get list of active shell processes and pick the most recent process id.

I don't understand how the most recent process id relates to the shell I'm running crc in? I could have launched a bash shell at the beginning of my session, then spawn a zsh shell to do some testing, and get back to the initial bash shell to use crc?
A higher PID also does not guarantee the process is newer as PIDs can wrap around and start again at 1.

If we want to do some more extensive shell guessing by iterating over processes, could we not limit ourselves to the parents of the crc process?
This is actually partially done this way in the windows code https://github.com/crc-org/crc/blob/cebafa45005113b7311f856b033434b8a509352b/pkg/os/shell/shell_windows.go#L79-L96 and now when this code calls `shellType` we run again `ps` in some cases.
I'd prefer that we extend this approach and drop the calls to an external command, and the parsing which comes with it.

rohanKanojia:
> I don't understand how the most recent process id relates to the shell I'm running crc in? I could have launched a bash shell at the beginning of my session, then spawn a zsh shell to do some testing, and get back to the initial bash shell to use crc?

@cfergeau : Could you please elaborate on this failing scenario? I tried creating a new terminal instance from the current `bash` session and created `zsh` and `fish` instances respectively. However, in the first `bash` session, I could not see the other shell processes with `ps -o pid,comm`:
```shell
$ ps  -o pid,comm
    PID COMMAND
 218741 bash
 290485 ps

```


From what I remember this `getNameAndItsPpid` l was returning the parent process name as `wsl.exe` (which was not what I was expecting).

We had tried other options like `$$` to get the process id of the current shell but unfortunately, it doesn't work for fish shells.

cfergeau:
Ah, I always use `ps` with `-a`, but its default behavior is a lot more restrictive as you described:
> By  default,  ps  selects all processes with the same effective user ID (euid=EUID) as the current user and associated with the same terminal as the invoker.

I can get this to misbehave though with 2 ssh sessions to a mac machine, run `zsh` in one, run `bash` in the other one, get back to the first one and run `ps`, I see the processes for all ssh sessions.

> From what I remember this `getNameAndItsPpid` was returning the parent process name as wsl.exe (which was not what I was expecting).

`getNameAndItsPpid` may not have the behaviour you want, but can it be fixed to do what we need? Imo, this is what we want to do, start from the current process, and look at the process names/command lines until one is a shell process.

In the scenario described above on macos, with `pstree` I get this so I can find the right terminal without relying on "latest launched"

```
 |-+= 04833 root sshd: teuf [priv] 
 | \-+- 04835 teuf sshd: teuf@ttys007 
 |   \-+= 04836 teuf -zsh
 |     \-+= 04884 teuf bash
 |       \-+= 07461 teuf pstree
 |         \--- 07462 root ps -axwwo user,pid,ppid,pgid,command
 |-+= 04866 root sshd: teuf [priv] 
 | \-+- 04868 teuf sshd: teuf@ttys008 
 |   \--= 04869 teuf -zsh
```
(I'm not suggesting to use `pstree` directly, but to do something similar in go)

rohanKanojia:
@cfergeau : Thanks, I've moved the discussion to a new issue https://github.com/crc-org/crc/issues/4562

Feel free to add any information that I might have missed. I'll start working on this issue soon.

Issue #4551, opened 2025-01-07
[QE] add arm platform of image quay.io/crcont/squid


lilyLuLiu:
quay.io/crcont/squid is used in e2e proxy test. But it only supports amd64. Tests on arm64 is not working. 
The Dockerfile of the image is https://github.com/crc-org/crc/blob/main/images/squid/Dockerfile

praveenkumar:
May be we should update this Dockerfile to have multiarch support.

cfergeau:
Latest build has both arches https://quay.io/repository/crcont/squid?tab=tags

praveenkumar:
Yes I created that yesterday, forgot to close this issue.

PR #4555, opened 2025-01-08
[QE]update the squid image version

lilyLuLiu:
## Description

Relates to: #4551

## Type of change

## Contribution Checklist
    - [ ] Linux
    - [ ] Windows
    - [x] MacOS

### DIFF
--- a/test/e2e/testsuite/testsuite.go
+++ b/test/e2e/testsuite/testsuite.go
@@ -215,7 +215,7 @@ func InitializeScenario(s *godog.ScenarioContext) {
 			if tag.Name == "@proxy" {
 
 				// start container with squid proxy
-				err := util.ExecuteCommand("podman run --name squid -d -p 3128:3128 quay.io/crcont/squid")
+				err := util.ExecuteCommand("podman run --name squid -d -p 3128:3128 quay.io/crcont/squid:5.5")
 				if err != nil {
 					fmt.Println(err)
 					os.Exit(1)


test/e2e/testsuite/testsuite.go
```diff
@@ -215,7 +215,7 @@ func InitializeScenario(s *godog.ScenarioContext) {
 			if tag.Name == "@proxy" {
 
 				// start container with squid proxy
-				err := util.ExecuteCommand("podman run --name squid -d -p 3128:3128 quay.io/crcont/squid")
+				err := util.ExecuteCommand("podman run --name squid -d -p 3128:3128 quay.io/crcont/squid:5.5")
```
cfergeau, line 218: https://quay.io/repository/crcont/squid?tab=tags has a `5.5` tag which was just created, and a `latest` tag which is 4 years old. Is there any reason for not updating the `latest` tag to point to the same image as the `5.5` tag? Then the test suite would automatically pick up the new image.

lilyLuLiu: I prefer to use `latest` tag. But the drawback is difficult to track back if image change. 
@praveenkumar WDYT?

praveenkumar: I am mapping `5.5` to latest and keeping both the tag so we can also able to track back.

cfergeau: > I prefer to use latest tag. But the drawback is difficult to track back if image change.

I think it is fairly common to have a moving `latest` tag, plus non-moving versioned tags (eg `5.5`), this way we get the best of both worlds :)

PR #4559, opened 2025-01-09
build: Use backup file in gen_release_info

cfergeau:
Using `sed -i` cross-platform is problematic as the `-i` flag does not
behave the same on macOS and linux (see 5e561ad5b).
The `sed` on macOS currently treat `-e` as a backup extension while on
linux it's treated as a flag. This does not seem to make a difference,
but it is very magic.

This commit specifies a backup extension to `sed -i`, which works for me
on macOS and linux. `-e` will be used as a flag on all platforms.

## Type of change

## Contribution Checklist
    - [x] Linux
    - [ ] Windows
    - [x] MacOS

Haven't tested on Windows, but this would be needed as this was initially changed because of windows issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 /docs/package-lock.json
 /release/
 /release-info.json
-/release-info.json-e
+/release-info.json.bak
 *.crcbundle
 /tmp-embed/
 /RPMS/
--- a/Makefile
+++ b/Makefile
@@ -270,8 +270,9 @@ cross-lint: $(TOOLS_BINDIR)/golangci-lint gen_release_info
 .PHONY: gen_release_info
 gen_release_info:
 	@cat release-info.json.sample | sed s/@CRC_VERSION@/$(CRC_VERSION)/ > $(RELEASE_INFO)
-	@sed -i"" -e s/@GIT_COMMIT_SHA@/$(COMMIT_SHA)/ $(RELEASE_INFO)
-	@sed -i"" -e s/@OPENSHIFT_VERSION@/$(OPENSHIFT_VERSION)/ $(RELEASE_INFO)
+	@sed -i.bak -e s/@GIT_COMMIT_SHA@/$(COMMIT_SHA)/ $(RELEASE_INFO)
+	@sed -i.bak -e s/@OPENSHIFT_VERSION@/$(OPENSHIFT_VERSION)/ $(RELEASE_INFO)
+	@rm -f $(RELEASE_INFO).bak
 
 .PHONY: linux-release-binary macos-release-binary windows-release-binary
 linux-release-binary: LDFLAGS+= $(RELEASE_VERSION_VARIABLES)


praveenkumar:
@rohanKanojia do check that on windows this all started https://github.com/crc-org/crc/pull/4530 with this.

rohanKanojia:
@cfergeau @praveenkumar : I performed quick run of `make test` on windows based on this branch.

I wasn't able to build using `make test`. I'm getting this error:
```
PS C:\Users\rokum\go\src\github.com\crc-org\crc> make test
process_begin: CreateProcess(NULL, rm -f release-info.json.bak, ...) failed.
make (e=2): The system cannot find the file specified.
make: *** [gen_release_info] Error 2
```

With this change, I'm also seeing that backup files are getting generated again:
```
        sed2B2kt2
        sedKOZBK5
```

cfergeau:
> With this change, I'm also seeing that backup files are getting generated again:

Is there any documentation about this behaviour/the sed version which is being used?  I don't really understand these results (why `-i""` generates randomly named backups, `-i''` generates no backup, and `-i .bak` generates again randomly named backups without the .bak extension)

rohanKanojia:
@cfergeau : Hello, I'm using GNU sed `v4.9`. I checked [sed documentation](https://www.gnu.org/software/sed/manual/sed.html) but it doesn't mention any difference in usage of `-i` flag with single or double quotes.

Could this be a problem related to underlying shell (PowerShell in my case) ?


`sed --version` output:
```
PS C:\Users\rokum\go\src\github.com\crc-org\docs> sed --version
C:\ProgramData\chocolatey\lib\sed\tools\sed.exe (GNU sed) 4.9
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Fenlason, Tom Lord, Ken Pizzini,
Paolo Bonzini, Jim Meyering, and Assaf Gordon.

This sed program was built without SELinux support.

GNU sed home page: <https://www.gnu.org/software/sed/>.
General help using GNU software: <https://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org>.
Patched (v2) by: Michael M. Builov <mbuilov@yandex.ru>.
```

cfergeau:
https://stackoverflow.com/questions/7733922/sed-command-creates-randomly-named-files
and
https://stackoverflow.com/questions/1823591/sed-creates-un-deleteable-files-in-windows
have some hypothesis, bug in the sed binary you are using, or maybe this:
> So after much testing last night, it turns out that sed was creating these files when trying to operate on an empty string.

rohanKanojia:
@cfergeau : Thanks for sharing these links, the type of backup files I see being created do look similar to the ones shared in StackOverflow links.

>  it turns out that sed was creating these files when trying to operate on an empty string.

Sorry I don't understand this clearly. Do you mean somehow arguments to sed command are not getting passed correctly?

cfergeau:
> > it turns out that sed was creating these files when trying to operate on an empty string.
> 
> Sorry I don't understand this clearly. Do you mean somehow arguments to sed command are not getting passed correctly?

This is a copy and paste from stackoverflow, I am also not 100% sure what they mean. But since we are passing args as variables (`$(COMMIT_SHA)`, `$(OPENSHIFT_VERSION)`), maybe one of these is empty. But I agree this does not really match what I quoted. I don't know, the windows behaviour you observe is puzzling :-/

lstocchi:
@rohanKanojia what if you use the git bash? It provides a unix-like env so if it works fine there we can think it is a problem with the powershell. Or you can check if the git bash/powershell use the same sed version and try it on the powershell as well to exclude it is a problem with your chocolatey sed tool

rohanKanojia:
@lstocchi : I tried on GitBash, running `make test` doesn't cause failure but it is also generating these random backup files:
```
rokum@rokumar-lenovo MINGW64 ~/go/src/github.com/crc-org/crc (sed)
$ git status
On branch sed
Your branch is up to date with 'cfergeau/sed'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        .idea/
        sedGOuAFr
        sedi08Zev
```

> Or you can check if the git bash/powershell use the same sed version and try it on the powershell as well to exclude it is a problem with your chocolatey sed tool

Okay, let me try to use a different version of sed

rohanKanojia:
I got some help from Anjan to test this same PR on his windows machine. He wasn't able to reproduce the problem I'm reporting in https://github.com/crc-org/crc/pull/4559#issuecomment-2579578373 . I think there is some problem with my environment setup as suggested by Luca. I'll try to fix it. I apologize for the confusion :pray: 

I think we can move forward with this pull request :+1:

PR #4572, opened 2025-01-16
refactor (shell) : Refactor CRC unix/linux shell detection logic using gopsutil (#4562)

### LINKED ISSUE #4562: Refactor CRC shell detection mechanism to use process parent programatically
# Description
_Originally posted by @cfergeau in https://github.com/crc-org/crc/issues/4526#issuecomment-2579431999_

> If we want to do some more extensive shell guessing by iterating over processes, could we not limit ourselves to the parents of the crc process?
> I'd prefer that we extend this approach and drop the calls to an external command, and the parsing which comes with it.
> In the scenario described above on macos, with pstree I get this so I can find the right terminal without relying on "latest launched"
> (I'm not suggesting to use `pstree` directly, but to do something similar in go)

## Acceptance Criteria 
- [ ] Investigate whether shell can be detected correctly using parent process approach programmatically without using any command
- [ ] If investigation is successful , refactor current shell detection mechanism to be based on this approach.

rohanKanojia:
## Description

Fixes: #4562 

Relates to: https://github.com/crc-org/crc/pull/4526#issuecomment-2579431999

Refactor CRC Shell detection for UNIX/Linux Systems to rely on parent process name. I was only able to get this approach working for Linux/Unix systems where gopsutil library is correctly detecting parent PIDs. On WSL, it's not able to see beyond windows processes.

## Type of change

## Proposed changes

+ Move parsing ps output and picking recent pid logic to shell_windows. It would only be used in case of WSL.
+ Instead of parsing ps output and picking up recent pid. Inspect parent process of current process and keep going up until we find a shell process. Pick that shell process as currently active shell.
    + As of now I've set a depth limit of 10 levels, if we're not able to find any shell process after going this deep, we return blank value.

## Testing

Start CRC cluster and run `oc-env` / `podman-env` commands on bash, zsh or fish shells.

## Contribution Checklist
    - [x] Linux
    - [x] Windows
    - [x] MacOS (QE Lab instance)


cfergeau:
> On WSL, it's not able to see beyond windows processes.

What about Windows with powershell or cmd?

rohanKanojia:
@cfergeau : In this PR I had kept changes limited to parts I had touched on in https://github.com/crc-org/crc/pull/4526 . In my previous pull request for Shell detection, I had only made changes for WSL and Unix, Linux platforms.

I just tested this approach on PowerShell and CMD. I can verify that this approach is able to detect shell processes from parent tree.

Some Debug logs on PowerShell:
```
PS C:\Users\rokum\go\src\github.com\crc-org\crc> C:\Users\rokum\go\bin\crc.exe oc-env
process name : crc.exe
parent process id : 8060
process name : powershell.exe
parent process id : 15508
process name : WindowsTerminal.exe
parent process id : 8592
process name : explorer.exe
parent process id : 8476
process name :
 detectShellByCheckingProcessTree
$Env:PATH = "C:\Users\rokum\.crc\bin\oc;$Env:PATH"
# & crc oc-env | Invoke-Expression
```

Some Debug logs on CMD:
```
C:\Users\rokum>C:\Users\rokum\go\bin\crc.exe oc-env
process name : crc.exe
parent process id : 14804
process name : cmd.exe
parent process id : 8592
process name : explorer.exe
parent process id : 8476
process name :
 detectShellByCheckingProcessTree
SET PATH=C:\Users\rokum\.crc\bin\oc;%PATH%
REM @FOR /f "tokens=*" %i IN ('crc oc-env') DO @call %i
```

Shall I modify this PR to cover PowerShell and CMD as well?

pkg/os/shell/shell_windows.go
```diff
@@ -101,3 +105,83 @@ func detect() (string, error) {
 
 	return shellType(shell, "cmd"), nil
 }
+
+// detectShellByInvokingCommand is a utility method that tries to detect current shell in use by invoking `ps` command.
+// This method is extracted so that it could be used by unix systems as well as Windows (in case of WSL). It executes
```
cfergeau: I think now it is only used on Windows in the WSL case?

pkg/os/shell/shell_unix.go
```diff
@@ -6,20 +6,99 @@ package shell
...
 	"path/filepath"
+
+	"github.com/shirou/gopsutil/v3/process"
+	"github.com/spf13/cast"
 )
 
 var (
-	ErrUnknownShell = errors.New("Error: Unknown shell")
+	ErrUnknownShell                             = errors.New("Error: Unknown shell")
+	getCurrentProcessID                         = os.Getpid
```
cfergeau: In my opinion, it is slightly easier to follow if you name this `getPid`, in a way this removes one level of indirection.

pkg/os/shell/shell_unix.go
```diff
@@ -6,20 +6,99 @@ package shell
 import (
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
+
+	"github.com/shirou/gopsutil/v3/process"
```
cfergeau: Why not v4? We probably should switch crc to the latest version, but this can be done in a follow-up commit if you prefer.

pkg/os/shell/shell_unix.go
```diff
@@ -6,20 +6,99 @@ package shell
...
+	ErrUnknownShell                             = errors.New("Error: Unknown shell")
+	getCurrentProcessID                         = os.Getpid
+	processSupplier     AbstractProcessSupplier = &ProcessSupplier{}
+	ProcessDepthLimit                           = 10
 )
 
+type AbstractProcess interface {
+	Name() (string, error)
+	Ppid() (int32, error)
+}
```
cfergeau, line 27: It seems this code could be simpler if you use [`Parent()`](https://pkg.go.dev/github.com/shirou/gopsutil/v4@v4.24.12/process#Process.Parent) instead of `Ppid()` ?

And I think it's possible to remove `	getCurrentProcessID                         = os.Getpid` and `AbstractProcessSupplier` and to replace both with `var newProcess = process.NewProcess`.

Or will this prevent some of the "failure" tests that you have from being doable?

cfergeau: Regardless of the previous suggestion, I'd add a comment saying these types are needed for the unit tests.
You could add `var _ AbstractProcess = process.Process` to make it clear that this type implements the interface.

cfergeau (commented):
Another potential follow-up is to use `detectShellByCheckingProcessTree` in `detect()` in `shell_windows.go` as both seem to have similar functionalities.

pkg/os/shell/shell_unix.go
rohanKanojia: Thanks for the suggestion, I've updated it to use [`Parent()`](https://pkg.go.dev/github.com/shirou/gopsutil/v4@v4.24.12/process#Process.Parent), as you suggested. It's much clearer now!

pkg/os/shell/shell_unix.go
rohanKanojia: Thanks, I've renamed the getter to `getPid` as you suggested.

pkg/os/shell/shell_unix.go
rohanKanojia: I've added a new commit on top of this one to upgrade to v4

pkg/os/shell/shell_windows.go
rohanKanojia: Sorry, I forgot to update this comment after moving the method. Thanks for noticing! It should be okay now.

rohanKanojia:
> Another potential follow-up is to use detectShellByCheckingProcessTree in detect() in shell_windows.go as both seem to have similar functionalities.

Can I tackle this one as a follow-up task?

cfergeau:
> > Another potential follow-up is to use detectShellByCheckingProcessTree in detect() in shell_windows.go as both seem to have similar functionalities.
> 
> Can I tackle this one as a follow-up task?

Yup definitely!

pkg/os/shell/shell_unix.go
```diff
@@ -6,20 +6,100 @@ package shell
...
+	"os"
 	"path/filepath"
+
+	"github.com/shirou/gopsutil/v3/process"
+	"github.com/spf13/cast"
 )
 
 var (
 	ErrUnknownShell = errors.New("Error: Unknown shell")
+	getPid          = getCurrentProcess
```
cfergeau: With the changes you did, `getPid` is no longer very accurate :-/

pkg/os/shell/shell_unix.go
cfergeau: Ah, the fact that `Parent` must return an `AbstractProcess` makes things a bit more complicated than I hoped as a wrapper is needed for `*process.Process`, but I agree it is simpler anyway.

pkg/os/shell/shell_unix.go
```diff
@@ -6,20 +6,100 @@ package shell
...
+type AbstractProcess interface {
+	Name() (string, error)
+	Parent() (AbstractProcess, error)
+}
+
+// RealProcess is a wrapper implementation of AbstractProcess to wrap around the gopsutil library's
+// process.Process object. This implementation is used in production code.
+type RealProcess struct {
+	realProcess *process.Process
+}
```
cfergeau: If you don't give a name to `realProcess`, you don't need to reimplement `Name`, go will reuse the implementation from `*process.Process`:
```
type RealProcess struct {
	*process.Process
}
```

pkg/os/shell/shell_unix.go
```diff
@@ -6,20 +6,100 @@ package shell
...
+
+func (p *RealProcess) Parent() (AbstractProcess, error) {
+	parentProcess, err := p.realProcess.Parent()
+	if err != nil {
+		return nil, err
+	}
+	return &RealProcess{parentProcess}, nil
+}
+
+func NewRealProcess(pid int32) (*RealProcess, error) {
```
cfergeau: `getCurrentProcess` is the single user of `NewRealProcess`, you could consider merging the 2 especially as most of their code is error-handling.

cfergeau (commented):
minor comments left, looks good to me!

pkg/os/shell/shell_unix.go
rohanKanojia: Thanks, I've updated it as per your suggestion.

pkg/os/shell/shell_unix.go
rohanKanojia: You're right. I've merged both methods into single one

pkg/os/shell/shell_unix.go
rohanKanojia: Sorry for the leftover typo, I've renamed it.

PR #4591, opened 2025-02-03
refactor (shell) : detect shell on windows using gopsutil too (#4588)

### LINKED ISSUE #4588: Refactor Shell detection on windows in order to use gopsutil
### Component

CLI

### Description

Related to #4572 

> Another potential follow-up is to use `detectShellByCheckingProcessTree` in `detect()` in `shell_windows.go` as both seem to have similar functionalities.

_Originally posted by @cfergeau in https://github.com/crc-org/crc/pull/4572#pullrequestreview-2569525275_

### Expected Behavior
shell windows detection is refactored to use gopustil (excluding WSL as this approach doesn't work there)

### Acceptance Criteria
- [ ] shell_windows uses gopsutil to detect shell type from parent process.

rohanKanojia:
## Description

As a follow up to #4572, updating shell detection logic on windows to also rely on gopsutil library to detect the shell name by inspecting parent processes of `crc.exe` process.

Fixes: #4588

Relates to: PR #4572

## Type of change

## Proposed changes

This PR mostly includes moving some methods so that they become accessible in windows environments too.

+ Move shell detection using `gopsutil` logic from `shell_unix.go` -> `shell.go` so that it can be reused in `shell_windows.go` as well
+ Remove `shell_linux.go`/ `shell_darwin.go` as they only contained a single constant that I moved to `shell_unix.go`
+ Remove `getProcessEntry` and related methods from `shell_windows.go` as we're reusing gopsutil approach that is working well on Windows

## Testing

1. Start CRC in windows environment
2. Verify `crc oc-env` is working as expected in 
    - CMD
    - PowerShell
    - GitBash
    - WSL
      - bash
      - fish
      - zsh

## Contribution Checklist
    - [ ] Linux
    - [x] Windows
    - [x] MacOS

### DIFF
--- a/pkg/os/shell/shell.go
+++ b/pkg/os/shell/shell.go
@@ -1,9 +1,13 @@
 package shell
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
+
+	"github.com/shirou/gopsutil/v4/process"
+	"github.com/spf13/cast"
 
 	crcos "github.com/crc-org/crc/v2/pkg/os"
 )
@@ -11,6 +15,8 @@ import (
 var (
 	CommandRunner                           = crcos.NewLocalCommandRunner()
 	WindowsSubsystemLinuxKernelMetadataFile = "/proc/version"
+	ErrUnknownShell                         = errors.New("Error: Unknown shell")
+	currentProcessSupplier                  = createCurrentProcess
 )
 
 type Config struct {
@@ -18,6 +24,20 @@ type Config struct {
 	Delimiter  string
 	Suffix     string
 	PathSuffix string
+}
+
+// AbstractProcess is an interface created to abstract operations of the gopsutil library
+// It is created so that we can override the behavior while writing unit tests by providing
+// a mock implementation.
+type AbstractProcess interface {
+	Name() (string, error)
+	Parent() (AbstractProcess, error)
+}
+
+// RealProcess is a wrapper implementation of AbstractProcess to wrap around the gopsutil library's
+// process.Process object. This implementation is used in production code.
+type RealProcess struct {
+	*process.Process
 }
 
 func GetShell(userShell string) (string, error) {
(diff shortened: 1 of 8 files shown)


pkg/os/util.go
```diff
@@ -127,3 +127,12 @@ func RemoveFileGlob(glob string) error {
 	}
 	return nil
 }
+
+func IsPresentInList(arr []string, str string) bool {
```
cfergeau: It already exists as [Contains](https://github.com/crc-org/crc/blob/main/pkg/strings/strings.go#L8)

pkg/os/shell/shell.go
```diff
@@ -151,3 +171,50 @@ func IsWindowsSubsystemLinux() bool {
...
+// Examples:
+//
+//	shellName := detectShellByCheckingProcessTree(1234)
+func detectShellByCheckingProcessTree(p AbstractProcess) string {
+	for p != nil {
+		processName, err := p.Name()
+		if err != nil {
+			return ""
+		}
+		if crcos.IsPresentInList(supportedShell, processName) {
```
cfergeau: This would be easier to review if it did not mix code movement with functional changes. I'd suggest having a commit where this code is moved to `shell.go` without any changes, and  then a commit porting the Windows code to use `detectShellByCheckingProcessTree` with these `crcos.IsPresentInList` and such.

rohanKanojia: Thanks, I've split the PR into two commits with first commit containing code reorganization and second including actual change

pkg/os/util.go
rohanKanojia: Thanks, I was going to use this but I noticed a unit test is failing on Windows due to these changes. This line was returning bash on Windows powershells:

https://github.com/crc-org/crc/blob/6d72284a886c4978e13e4775b9b8f552e134bf18/cmd/crc/cmd/start_test.go#L41

On checking the process tree during unit test execution, it shows processes like this:
`cmd.test.exe`  -> `go.exe`     -> `bash.exe`

I've added another utility method in order to just match the prefix instead of matching entire string.

pkg/os/shell/shell.go
```diff
@@ -208,7 +210,9 @@ func detectShellByCheckingProcessTree(p AbstractProcess) string {
 		if err != nil {
 			return ""
 		}
-		if processName == "zsh" || processName == "bash" || processName == "fish" {
+		if crcstrings.IsPresentInListSatisfying(supportedShell, processName, func(listElem string, toMatch string) bool {
+			return strings.HasPrefix(toMatch, listElem)
+		}) {
```
cfergeau, line 214: note: if you switch the  2 arguments when invoking the callback in `IsPresentInListSatisfying`, I think this can be simplified to `crcstrings.IsPresentInListSatisfying(supportedShell, processName, strings.HasPrefix)`. Fine with me if you prefer to keep the code the way it is now.

pkg/strings/strings.go
```diff
@@ -6,8 +6,14 @@ import (
 )
 
 func Contains(input []string, match string) bool {
+	return IsPresentInListSatisfying(input, match, func(listElement string, toMatch string) bool {
+		return listElement == toMatch
+	})
+}
+
+func IsPresentInListSatisfying(input []string, toMatch string, matchingPredicate func(string, string) bool) bool {
```
cfergeau: While looking for naming ideas in go standard packages (usual name for such a function is `ContainsFunc`), I realized the [`slices`](https://pkg.go.dev/slices#ContainsFunc) package has both `Contains` and `ContainsFunc`, we can use these instead.

rohanKanojia: Oh, thanks a lot! I'm sorry I should've checked this.

cfergeau (approved):
There's an outdated `+ Add a utility method in strings for verifying a slice contains an
  element matching the provided predicate.` in the commit log of the second commit, otherwise looks good to me!

rohanKanojia:
@cfergeau : Thanks for noticing! I've updated the commit message.

I'll also create a follow up PR to make use of [slice.Contains](url) to replace the equivalent method in crcstrings.Contains

cfergeau (approved):

PR #4607, opened 2025-02-11
Clean filepath before reading the content

praveenkumar:
Looks like security job is failing because it detect unsanitized input from file, this should fix following

```
  ✗ [Medium] Path Traversal
   ID: 2ce4a8d7-4fb1-41b5-8841-dc76ea48e503
   Path: pkg/crc/machine/bundle/repository.go, line 41
   Info: Unsanitized input from file name flows into os.ReadFile, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to read arbitrary files.
```

### DIFF
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -38,7 +38,7 @@ func (repo *Repository) Get(bundleName string) (*CrcBundleInfo, error) {
 		return nil, errors.Wrapf(err, "could not find cached bundle info in %s", path)
 	}
 	jsonFilepath := filepath.Join(path, metadataFilename)
-	content, err := os.ReadFile(jsonFilepath)
+	content, err := os.ReadFile(filepath.Clean(jsonFilepath))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rohanKanojia (approved):

pkg/crc/machine/bundle/repository.go
```diff
@@ -38,7 +38,7 @@ func (repo *Repository) Get(bundleName string) (*CrcBundleInfo, error) {
 		return nil, errors.Wrapf(err, "could not find cached bundle info in %s", path)
 	}
 	jsonFilepath := filepath.Join(path, metadataFilename)
-	content, err := os.ReadFile(jsonFilepath)
+	content, err := os.ReadFile(filepath.Clean(jsonFilepath))
```
cfergeau, line 41: `filepath.Join` already calls `filepath.Clean` https://pkg.go.dev/path/filepath#Join so it's dubious that this fixes anything. It's also not clear exactly what exact scenario we want to protect against.

Issue #4609, opened 2025-02-11
[QE] Update mapt usage on gh flows


adrianriobo:
### Component

None

### Description

Current gh flows using mapt are using an image out of date, we need to move to the latest available image. Within the latest image the approach on how to define which vm type will be picked is a bit different, may that will give us a bit more stability .

It would be great to be able to track newer versions with dependabot. Check if possible

### Expected Behavior

GH flows for testing windows will provision machines using the latest available version of mapt

### Acceptance Criteria
- [x] GH flow for tesitng windows uses latest mapt version and provision and decommissions works as expected.

adrianriobo:
This issue is kind of related with the current state of tests which are almost all the time showing red, this is also caused cause all windows qe shows only one state...i.e. integration and e2e on each windows version, we can split the status update to show each of them independent

albfan:
- [x] update mapt version and arguments to call it
    Current paramters:

--project-name
--backed-url
--conn-details-output
--windows-version
--windows-featurepack
--tags

    possible new parameters:

--admin-username
--cpus
--ghactions-runner-labels
--ghactions-runner-name
--ghactions-runner-repo
--ghactions-runner-token
--install-ghactions-runner
--location
--memory
--nested-virt
--profile
--spot-eviction-tolerance
--username
--vmsize
--debug
--debug-level
    
- [ ] check if dependabot can detect new mapt versions 
- [ ] Separate flows / status updates for better info on tests

albfan:
Doing some research on https://github.com/dependabot/demo?tab=readme-ov-file and how status is collected right now

adrianriobo:
I am not sure if dependabot can be used to basically check if newer versions for any component has been released... if that is the case... just try to make it inspect the content on the gh actions files to check for newer versions on the mapt image...but may this is not possible...I really do not know

albfan:
currently we are on 0.12.1, which is recent enough

PR #4617, opened 2025-02-13
QE: bump mapt to 0.7.4

albfan:
Use last available version from mapt

Relates: #4609

### DIFF
--- a/.github/workflows/windows-qe-tpl.yml
+++ b/.github/workflows/windows-qe-tpl.yml
@@ -107,7 +107,7 @@ jobs:
           -e ARM_CLIENT_SECRET='${{secrets.ARM_CLIENT_SECRET}}' \
           -e AZURE_STORAGE_ACCOUNT='${{ secrets.AZURE_STORAGE_ACCOUNT }}' \
           -e AZURE_STORAGE_KEY='${{ secrets.AZURE_STORAGE_KEY }}' \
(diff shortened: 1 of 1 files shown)


rohanKanojia (approved):

.github/workflows/windows-qe-tpl.yml
```diff
@@ -107,7 +107,7 @@ jobs:
           -e ARM_CLIENT_SECRET='${{secrets.ARM_CLIENT_SECRET}}' \
           -e AZURE_STORAGE_ACCOUNT='${{ secrets.AZURE_STORAGE_ACCOUNT }}' \
           -e AZURE_STORAGE_KEY='${{ secrets.AZURE_STORAGE_KEY }}' \
-          quay.io/redhat-developer/mapt:v0.6.9 azure \
+          quay.io/redhat-developer/mapt:v0.7.4 azure \
```
adrianriobo: With the newer version we need to adjust the params involved:

https://github.com/redhat-developer/mapt/blob/main/docs/azure/windows.md

At least now we need also the `--nested-virt` + I would suggest to add also `--memory`  and `--cpus` as with those params the search for spot instances is not limited to the default vmsize but to any size matching the requirements

adrianriobo (changes requested):

adrianriobo:
@albfan please check previous comment. Also mapt `v0.8.0` has been release so upgrade directly to `v0.8.0`

PR #4620, opened 2025-02-14
remove enabling of file sharing and creation of smbshare from MSI

### LINKED ISSUE #4619: Remove automatic creation of shared folder from the MSI
In the MSI we enable directory sharing for windows during installation, as CIFS shared folders haven't seen much use and Anti virus scanners flagging it as undesirable behavior we have decided to remove this from the MSI installers

User's can still enabled CIFS based folder sharing on Windows from `crc setup` but this involves a second reboot after the installation.

<!-- instructions to be added how to enable this ->

anjannath:
fixes #4619

### DIFF
--- a/packaging/windows/product.wxs.template
+++ b/packaging/windows/product.wxs.template
@@ -35,12 +35,6 @@
                 Name="Path" />
         </Property>
 
-        <Property Id="USERFOLDER">
-            <DirectorySearch Id="userProfileSearch" Depth="0" Path="[%USERPROFILE]" />
-        </Property>
-
-        <Property Id="SHAREDDIRNAME" Secure="yes">crc-dir0</Property>
-
         <util:Group Id="CrcUsersGroup" Name="crc-users" />
         <Directory Id="TARGETDIR" Name="SourceDir">
             <Directory Id="ProgramFiles64Folder">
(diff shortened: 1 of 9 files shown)


praveenkumar:
@anjannath didn't we have this as part of preflight before and then moved it to MSI? if yes then we might to revert those commits instead of this?

anjannath:
> @anjannath didn't we have this as part of preflight before and then moved it to MSI? if yes then we might to revert those commits instead of this?

no, i also thought the same and i did check the logs before creating the PR, seems this was never added as preflight checks, i think because when we added this SMB sharing support we were already moving every privileged operation on windows already to MSI so when we added support for SMB directory sharing this was only added to the MSI..

```
$ git log --grep="smb" --grep="cifs" --oneline
fff474a33 msi: remove the smb share created during installation
2aa7059b4 msi: ignore errors from custom action creating the smb shared dir
4a52aef4e cifs: Add 'enable-shared-dirs' config for windows
b4cfc0ec5 add CIFS/SMB based file sharing for windows
2785de06d move initialization of SharedDirs to driver's CreatHost function
````

File: pkg/crc/preflight/preflight_darwin_test.go [hunk 1]
```diff
@@ -17,9 +17,9 @@ func TestCountConfigurationOptions(t *testing.T) {
...
-	assert.Len(t, getPreflightChecks(true, network.SystemNetworkingMode, constants.GetDefaultBundlePath(preset.OpenShift), preset.OpenShift, false), 21)
+	assert.Len(t, getPreflightChecks(true, network.SystemNetworkingMode, constants.GetDefaultBundlePath(preset.OpenShift), preset.OpenShift, false, false), 21)
+	assert.Len(t, getPreflightChecks(true, network.SystemNetworkingMode, constants.GetDefaultBundlePath(preset.OpenShift), preset.OpenShift, false, false), 21)
 
-	assert.Len(t, getPreflightChecks(true, network.UserNetworkingMode, constants.GetDefaultBundlePath(preset.OpenShift), preset.OpenShift, false), 20)
-	assert.Len(t, getPreflightChecks(true, network.UserNetworkingMode, constants.GetDefaultBundlePath(preset.OpenShift), preset.OpenShift, false), 20)
+	assert.Len(t, getPreflightChecks(true, network.UserNetworkingMode, constants.GetDefaultBundlePath(preset.OpenShift), preset.OpenShift, false, false), 20)
+	assert.Len(t, getPreflightChecks(true, network.UserNetworkingMode, constants.GetDefaultBundlePath(preset.OpenShift), preset.OpenShift, false, false), 20)
```
cfergeau, line 24: There are several changes to `getPreflightChecks` calls in this commit to add one argument, they seem to belong to the previous commit.

packaging/windows/product.wxs.template
```diff
@@ -92,24 +92,6 @@
             Before="AddUserToHypervAdminGroup"
             Sequence="execute"/>
         <CustomAction Id="AddUserToHypervAdminGroup" BinaryKey="WixCA" DllEntry="WixQuietExec64" Execute="deferred" Impersonate="no" Return="ignore" />
-        <SetProperty Action="CACreateSMBShare"
-            Id="CreateSMBShare"
-            Value="&quot;[POWERSHELLEXE]&quot; -NonInteractive -ExecutionPolicy Bypass -NoProfile -Command &quot;New-SmbShare -Name '[SHAREDDIRNAME]' -Path '[USERFOLDER]' -FullAccess '[LogonUser]'&quot;"
```
cfergeau, line 97: I guess you should also be removing 
```
        <Property Id="USERFOLDER">
            <DirectorySearch Id="userProfileSearch" Depth="0" Path="[%USERPROFILE]" />
        </Property>

        <Property Id="SHAREDDIRNAME" Secure="yes">crc-dir0</Property>
```
and possibly other properties? (powershell, logonuser, …)

cfergeau (commented):
> this should remove at least of the warnings reported by virustotal for
the MSI:

One concern I have is if virustotal starts reporting this against the crc binary instead of reporting that against the msi

packaging/windows/product.wxs.template
anjannath: we need to keep powershell and logonuser these are used by the installHyperv custom action

pkg/crc/preflight/preflight_darwin_test.go
anjannath: updated

redbeam:
@anjannath I suggest closing this as I opened a separate PR that removes SMB file sharing support completely (#4879).

anjannath:
closed in favor of #4879

PR #4624, opened 2025-02-17
[QE] Set env var windows

### LINKED ISSUE #4618: [QE] integration: Setting env var on windows fail
### General information

crc might be setting boolean parameters incorrectly

### Operating System

Windows

### Hypervisor

Hyper-V

### Did you run `crc setup` before `crc start`?

yes

### Running on

Laptop

### Steps to reproduce

test disable update check

### CRC version

```shell

```

### CRC status

```shell

```

### CRC config

```shell

```

### Host Operating System

```shell
windows 10
```

### Expected behavior

env var set is readed correctly

### Actual behavior

error parsing the command line created

### CRC Logs

```shell
Step starting CRC with default bundle succeeds: command '$env:CRC_DISABLE_UPDATE_CHECK=true; crc start -p 'C:\Users\rhqp\crc-e2e\pull-secret' ', expected to succeed, exited with exit code: 1
Command stdout: 
Command stderr: true : The term 'true' is not recognized as the name of a cmdlet, function, script file, or operable program. Check 
the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:31
+ $env:CRC_DISABLE_UPDATE_CHECK=true; crc start -p 'C:\Users\rhqp\crc-e ...
+                               ~~~~
    + CategoryInfo          : ObjectNotFound: (true:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
 
level=debug msg="CRC version: 2.47.0+457b65\n"
level=debug msg="OpenShift version: 4.17.14\n"
```

### Additional context

From a quick test:

```
PS C:\Users\crcqe> $env:CRC_DISABLE_UPDATE_CHECK="true"
PS C:\Users\crcqe> echo $env:CRC_DISABLE_UPDATE_CHECK  
true
PS C:\Users\crcqe> $env:CRC_DISABLE_UPDATE_CHECK='true'
PS C:\Users\crcqe> echo $env:CRC_DISABLE_UPDATE_CHECK  
true
PS C:\Users\crcqe> $env:CRC_DISABLE_UPDATE_CHECK=$true 
PS C:\Users\crcqe> echo $env:CRC_DISABLE_UPDATE_CHECK 
True
PS C:\Users\crcqe> $env:CRC_DISABLE_UPDATE_CHECK=true 
true: The term 'true' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

albfan:
## Description

CI sets some configurations using env var, `FOO=bar`, but windows do not support that format, it should be `$env:FOO=bar;`

Fixes: #4618

### DIFF
--- a/test/extended/crc/cmd/cmd.go
+++ b/test/extended/crc/cmd/cmd.go
@@ -89,19 +89,12 @@ func (c Command) Execute() error {
 func (c Command) env() []string {
 	var env []string
 	if !c.updateCheck {
-		env = append(env, envVariable("CRC_DISABLE_UPDATE_CHECK", "true"))
+		env = append(env, util.EnvVariable("CRC_DISABLE_UPDATE_CHECK", "true"))
 	}
 	if c.disableNTP {
-		env = append(env, envVariable("CRC_DEBUG_ENABLE_STOP_NTP", "true"))
+		env = append(env, util.EnvVariable("CRC_DEBUG_ENABLE_STOP_NTP", "true"))
 	}
 	return env
-}
-
-func envVariable(key, value string) string {
-	if runtime.GOOS == "windows" {
-		return fmt.Sprintf("$env:%s=%s;", key, value)
-	}
-	return fmt.Sprintf("%s=%s", key, value)
 }
 
 func (c Command) validate() error {
--- a/test/extended/crc/cmd/cmd_test.go
+++ b/test/extended/crc/cmd/cmd_test.go
@@ -8,7 +8,8 @@ import (
 
 func TestCommand(t *testing.T) {
 	assert.Equal(t, "crc setup", Command{
-		command: "setup",
+		command:     "setup",
+		updateCheck: true,
 	}.ToString())
 	assert.Equal(t, "CRC_DEBUG_ENABLE_STOP_NTP=true crc start -b bundle", Command{
 		command:     "start -b bundle",
@@ -16,7 +17,8 @@ func TestCommand(t *testing.T) {
 		updateCheck: true,
 	}.ToString())
 	assert.Equal(t, "CRC_DISABLE_UPDATE_CHECK=true CRC_DEBUG_ENABLE_STOP_NTP=true crc start -b bundle", Command{
-		command:    "start -b bundle",
-		disableNTP: true,
+		command:     "start -b bundle",
+		disableNTP:  true,
+		updateCheck: false,
 	}.ToString())
 }
--- a/test/extended/util/util.go
+++ b/test/extended/util/util.go
@@ -22,6 +22,23 @@ var (
 	CRCHome string
 )
 
+func EnvVariable(key, value string) string {
+	var prefix = ""
+	var suffix = ""
+	if runtime.GOOS == "windows" {
+		prefix = "$env:"
+		suffix = ";"
+		switch value {
+		case "true":
+			value = "$true"
+		case "false":
+			value = "$false"
+		default:
+		}
+	}
+	return fmt.Sprintf("%s%s=%s%s", prefix, key, value, suffix)
+}
+
 func CopyFilesToTestDir() error {
 	cwd, err := os.Getwd()
 	if err != nil {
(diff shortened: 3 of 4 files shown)


test/extended/crc/cmd/cmd.go
```diff
@@ -97,11 +97,23 @@ func (c Command) env() []string {
...
+		suffix = ";"
+		switch value {
+		case "true":
+			value = "$true"
+			break
+		case "false":
+			value = "$false"
+			break
+		default:
+		}
```
rohanKanojia: Do you think we can replace this switch statement like this?
```suggestion
                value = fmt.Sprintf("$%s", value)
```

anjannath: maybe the function should be called `SetBoolEnvVariable` then, because we want to append the `$` only in case of bool values

test/extended/crc/cmd/cmd_test.go
```diff
@@ -8,15 +8,17 @@ import (
...
 		command:     "start -b bundle",
 		disableNTP:  true,
 		updateCheck: true,
 	}.ToString())
 	assert.Equal(t, "CRC_DISABLE_UPDATE_CHECK=true CRC_DEBUG_ENABLE_STOP_NTP=true crc start -b bundle", Command{
-		command:    "start -b bundle",
-		disableNTP: true,
+		command:     "start -b bundle",
+		disableNTP:  true,
+		updateCheck: false,
```
anjannath, line 22: could you please add some details to the commit log why this is needed now or what this fixes

adrianriobo: @albfan check previous comment :pray:

albfan:
Probably fixed with a machine restart, as it is not reproducible anymore. closing by now

cfergeau:
Some of the test failures look related to these changes:
```
 cd tools && GOBIN="/go/src/github.com/crc-org/crc/tools/bin" go install github.com/golangci/golangci-lint/cmd/golangci-lint
"/go/src/github.com/crc-org/crc/tools/bin"/golangci-lint run
test/extended/util/util.go:34:4: S1023: redundant break statement (gosimple)
			break
			^
test/extended/util/util.go:37:4: S1023: redundant break statement (gosimple)
			break
			^ 
```

test/extended/crc/cmd/cmd.go
albfan: This cannot be applied as this deals too with the case of env var not being a bool, in which case, value is keep unmodified:

```
FOO=bar
$env:FOO=bar
```

```
BAR=true
$env:BAR=$true
```

this is problematic if we pretend to set an env var with text "true" or "false", but probably that will be never the case. If there's a need for that, then we need a way to specify that we want a $true or $false in windows

rohanKanojia: okay, I'm sorry I didn't know about this use case. 

marking this comment as resolved.

test/extended/crc/cmd/cmd_test.go
albfan: I tried to be explicit about expected result from test, but updateCheck by default is false. @anjannath Should I remove this part of the change?

anjannath: ahh thanks, i understand now, so you wanted to make `Command.updateCheck = false` explicit, we can keep this change

test/extended/util/util.go
```diff
@@ -22,6 +22,25 @@ var (
...
+	if runtime.GOOS == "windows" {
+		prefix = "$env:"
+		suffix = ";"
+		switch value {
+		case "true":
+			value = "$true"
+			break
+		case "false":
+			value = "$false"
+			break
```
praveenkumar: why you need the break?

praveenkumar:
```
/go/src/github.com/crc-org/crc/tools/bin"/golangci-lint run
test/extended/util/util.go:34:4: S1023: redundant break statement (gosimple)
			break
			^
test/extended/util/util.go:37:4: S1023: redundant break statement (gosimple)
			break 
```
linter is failing, make sure run `make cross-lint` after your changes to make sure it doesn't fail linting side.

PR #4647, opened 2025-02-27
Update golang version to 1.23

praveenkumar:


rohanKanojia (approved):

cfergeau:
/hold
Just so it's clear, the current consensus is that we don't want to make the change before the next sprint.
It's good to know some of the problems that we'll have in advance

albfan:
quick test building on go 1.23 arises problems with lint on admin-helper, so https://github.com/crc-org/admin-helper/pull/66 would be related

praveenkumar:
@cfergeau should we remove `/hold` for this because on brew also we have Go1.23.x now?

cfergeau:
> @cfergeau should we remove `/hold` for this because on brew also we have Go1.23.x now?

Ah sure, we need to proceed with care as there are always side-effects when we raise the go version, but since the release has been done, we can try to move forward.
/unhold

PR #4682, opened 2025-03-16
Use virtio-net device for forwarding networking data between the guest and host

### LINKED ISSUE #4460: [BUG] vfkit: Use `--device virtio-net,unixSocketPath`
User-mode networking on macos uses gvforwarder running in the guest + a vsock socket to forward the networking data between the guest and the host.
vfkit has had support for `--device virtio-net,unixSocketPath` for a while. This allows to pass the networking data directly from the hypervisor to gvisor-tap-vsock without the need for a process running in the guest. This is similar to vpnkit which was used with hyperkit (see commit cf173bc033e7ae)

vyasgun:
## Description
Use virtio-net device for forwarding networking data between the guest and host

Fixes: #4460 

## Type of change

## Contribution Checklist
    - [ ] Linux
    - [ ] Windows
    - [x] MacOS

### DIFF
--- a/cmd/crc/cmd/daemon.go
+++ b/cmd/crc/cmd/daemon.go
@@ -81,7 +81,7 @@ func createNewVirtualNetworkConfig(providedConfig *crcConfig.Config) types.Confi
 		GatewayIP:         constants.VSockGateway,
 		GatewayMacAddress: "5a:94:ef:e4:0c:dd",
 		DHCPStaticLeases: map[string]string{
-			"192.168.127.2": "5a:94:ef:e4:0c:ee",
+			"192.168.127.2": constants.VsockMacAddress,
 		},
 		DNS: []types.Zone{
 			{
@@ -146,11 +146,6 @@ func createNewVirtualNetworkConfig(providedConfig *crcConfig.Config) types.Confi
 }
 
 func run(configuration *types.Configuration) error {
-	vsockListener, err := vsockListener()
-	if err != nil {
-		return err
-	}
-
 	vn, err := virtualnetwork.New(configuration)
 	if err != nil {
 		return err
@@ -213,6 +208,21 @@ func run(configuration *types.Configuration) error {
 		}
 	}()
 
+	go func() {
+		_, err := unixgramListener(vn)
+		if err != nil {
+			// Don't return an error if the connection is closed
+			if !errors.Is(err, net.ErrClosed) {
+				errCh <- err
+			}
+			return
+		}
+	}()
+
+	vsockListener, err := vsockListener()
+	if err != nil {
+		return err
+	}
 	go func() {
 		mux := http.NewServeMux()
 		mux.Handle(types.ConnectPath, vn.Mux())
--- a/cmd/crc/cmd/daemon_linux.go
+++ b/cmd/crc/cmd/daemon_linux.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/containers/gvisor-tap-vsock/pkg/transport"
+	"github.com/containers/gvisor-tap-vsock/pkg/virtualnetwork"
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
 
@@ -122,6 +123,10 @@ func httpListener() (net.Listener, error) {
 		return nil, err
 	}
 	return ln, nil
+}
+
+func unixgramListener(_ *virtualnetwork.VirtualNetwork) (*net.UnixConn, error) {
+	return nil, nil
 }
 
 func startupDone() {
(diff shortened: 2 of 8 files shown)


pkg/drivers/vfkit/driver_darwin.go
```diff
@@ -246,12 +246,12 @@ func (d *Driver) Start() error {
 		return err
 	}
 
-	// virtio-vsock device
-	dev, err = config.VirtioVsockNew(d.DaemonVsockPort, d.VsockPath, true)
+	netDev, err := config.VirtioNetNew("5a:94:ef:e4:0c:ee")
```
cfergeau: This duplicates a value from `cmd/crc/cmd/daemon.go`, this at least deserves a comment, probably best to add it to `d`.

cmd/crc/cmd/daemon.go
```diff
@@ -199,18 +197,46 @@ func run(configuration *types.Configuration) error {
...
+			}
+			ctx := context.Background()
+			err = vn.AcceptVfkit(ctx, vfkitConn)
+			if err != nil {
+				// Don't return an error if the connection is closed
+				if !errors.Is(err, net.ErrClosed) {
+					errCh <- errors.Wrap(err, "failed to accept vfkit connection")
+				}
+				return
+			}
```
cfergeau: Imo this belongs more in `daemon_darwin.go` than here, this at least should go in a separate function as it is quite long.

pkg/drivers/vfkit/driver_darwin.go
cfergeau: I'd add the `UnixgramSocketPath` to `driver` to be similar to `DaemonVsockPort` and `VsockPath`.
Maybe we can even support both vsock and unixgram modes in `driver_darwin.go` depending on which values are set in `d`?

vyasgun: I have added code to set vsock mode only if `UnixgramSockPath` is not set but I'm setting the path here:
https://github.com/crc-org/crc/pull/4682/files#diff-bd1e0081558a01b12d1bfbb4eedac6696b0f7bc6571e2637d6207a9f7779e38bR70
so it won't be empty

cmd/crc/cmd/daemon_linux.go
```diff
@@ -124,6 +125,9 @@ func httpListener() (net.Listener, error) {
 	return ln, nil
 }
 
+func unixgramListener(_ *virtualnetwork.VirtualNetwork) (*net.UnixConn, error) {
+	return nil, nil
+}
```
cfergeau, line 130: Can you add a blank line after the closing brace here and in daemon_window.go?

pkg/crc/constants/constants.go
```diff
@@ -38,9 +38,10 @@ const (
 	RootlessPodmanSocket      = "/run/user/1000/podman/podman.sock"
 	RootfulPodmanSocket       = "/run/podman/podman.sock"
 
-	VSockGateway = "192.168.127.1"
-	VsockSSHPort = 2222
-	LocalIP      = "127.0.0.1"
+	VSockGateway      = "192.168.127.1"
+	VsockSSHPort      = 2222
+	LocalIP           = "127.0.0.1"
+	ApplehvMACAddress = "5a:94:ef:e4:0c:ee"
```
cfergeau: You could reuse the `vsock` prefix instead of `Applehv`? (but it’s not that accurate :-/). This mac address is not used only by vfkit/applehv.

pkg/crc/constants/constants_linux.go
```diff
@@ -8,6 +8,7 @@ const (
 	OcExecutableName           = "oc"
 	PodmanRemoteExecutableName = "podman-remote"
 	TapSocketPath              = ""
+	UnixgramSocketPath         = ""
```
cfergeau: `UnixgramSocketPath` is not needed in `constants_linux.go` and `constants_windows.go`

cfergeau (commented):
Some very minor comments, but I tested it and it works fine.
Do you know how this change will interact with the gvforwarder process in the VM?

pkg/crc/constants/constants.go
vyasgun: I’ve temporarily named it VsockMacAddress, but I’m not sure if it's accurate since it serves as an alternative to using vsock.

pkg/crc/constants/constants_windows.go
```diff
@@ -7,4 +7,5 @@ const (
 	DaemonHTTPNamedPipe        = `\\.\pipe\crc-http`
 	DaemonTaskName             = "crcDaemon"
 	AdminHelperServiceName     = "crcAdminHelper"
+	UnixgramSocketPath         = ""
```
cfergeau, line 10: This can be removed.

cfergeau (approved):
/lgtm
/approve

PR #4685, opened 2025-03-17
QE: new params for windows CI

albfan:
mapt version 0.8.0 adds new parameters on windows for better spot selection

Relates: #4609

### DIFF
--- a/.github/workflows/windows-qe-tpl.yml
+++ b/.github/workflows/windows-qe-tpl.yml
@@ -104,6 +104,7 @@ jobs:
             --project-name 'windows-desktop-${{ matrix.windows-version }}-${{ matrix.windows-featurepack }}-${{inputs.qe-type}}-${{inputs.preset}}' \
             --backed-url azblob://crc-qenvs-state/${{ github.repository }}-${{ github.run_id }} \
             --conn-details-output '/workspace' \
+            --nested-virt \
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-qe-tpl.yml
```diff
@@ -104,6 +106,9 @@ jobs:
             --project-name 'windows-desktop-${{ matrix.windows-version }}-${{ matrix.windows-featurepack }}-${{inputs.qe-type}}-${{inputs.preset}}' \
             --backed-url azblob://crc-qenvs-state/${{ github.repository }}-${{ github.run_id }} \
             --conn-details-output '/workspace' \
+            --nested-virt \
+            --memory ${{matrix.memory }} \
```
adrianriobo: You can put values directly here...matrix is fine but it is more intended for multiple values for a field...when you have just one..you can set it directly on the code (or variable, but not matrix)

albfan: as those are the default values, I'm unsure what's the difference setting it explicitely. Not setting them those are the Implicit values

adrianriobo: actually you would not need them...as you said those are already default...the only one which you need and it is not default is `nested-virt` so it is really up to you even just remove them and leave only `nested-virt`

PR #4697, opened 2025-03-24
Update devcontainer image to ci.openshift.org/openshift

praveenkumar:


.devcontainer/devcontainer.json
```diff
@@ -1,11 +1,11 @@
...
 
     "customizations": {
         "vscode": {
             "extensions": [
             ]
         }
     }
-}
\ No newline at end of file
+}
```
praveenkumar, line 11: This file is first time added as part of windows but I didn't make this change except editing it in linux :)

cfergeau (commented):
github actions are still failing because https://github.com/crc-org/crc/blob/main/.github/workflows/verify-devcontainer.yml is not using the image name from `.devcontainer/devcontainer.json`

praveenkumar:
> github actions are still failing because https://github.com/crc-org/crc/blob/main/.github/workflows/verify-devcontainer.yml is not using the image name from `.devcontainer/devcontainer.json`

@cfergeau fixed now, didn't know we are using same image for the action workflow also. It would've much better if there is some action on marketplace which validate the devcontainer with images.

anjannath (approved):

cfergeau:
> @cfergeau fixed now, didn't know we are using same image for the action workflow also. It would've much better if there is some action on marketplace which validate the devcontainer with images.

Yup agreed, I don’t really understand what the current gh action is actually testing to be honest.

PR #4712, opened 2025-04-01
golangci: migrate to v2 schema

sebrandon1:
Utilize the `golangci-lint migrate` command to migrate the config over to the v2 schema.

Note: I disabled the `revive` linter because it was flagging the dot-imports for ginkgo and gomega.

Similar to: https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2875

## Summary by Sourcery

Migrate golangci-lint configuration to version 2 schema, updating linter settings and configuration structure

Bug Fixes:
- Temporarily disable revive linter to avoid false positives with dot imports in test frameworks

Enhancements:
- Refactor linter configuration to use more structured exclusion rules
- Improve linter and formatter configuration with preset exclusions

Chores:
- Update golangci-lint configuration to use v2 schema with improved linter organization


.golangci.yml
```diff
@@ -1,23 +1,36 @@
...
+    generated: lax
+    presets:
+      - comments
+      - common-false-positives
+      - legacy
+      - std-error-handling
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
cfergeau: I don’t understand these `paths` nor these `presets`? and it looks like the v1 `exclude-rules` could be ported so that we can keep `revive` enabled?
Something like this:
```
linters:
  exclusions:
    rules:
      path: 'test/*'
      text: 'dot-imports: should not use dot imports'
      linters:
        - revive
```

.golangci.yml
```diff
@@ -1,23 +1,36 @@
+version: "2"
 run:
   build-tags:
     - build
     - containers_image_openpgp
-  timeout: 10m
```
cfergeau, line 5: can we keep this? this avoids some timeouts when running in github actions for example. Or is this timeout already longer with v2?

cfergeau (commented):
Thanks for this PR to upgrade to golangci-lint v2!

.golangci.yml
sebrandon1: Yeah I thought this was weird that the migration tool removed it.  If I go look at their example reference config [here](https://github.com/golangci/golangci-lint/blob/544018f712a7506c49ecdb75b8cb5a4e7bbb7534/.golangci.reference.yml#L4182) it is still a valid option.  

I'll add it back in.

.golangci.yml
sebrandon1: I'll give that a try!

praveenkumar:
Need rebase and also looks like updated revive need some more changes as part of import.

.golangci.yml
```diff
@@ -1,23 +1,36 @@
...
+formatters:
+  enable:
+    - gofmt
+    - goimports
+  exclusions:
+    generated: lax
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
cfergeau: I’m still unsure about these paths, why do we need to ignore them? (they don’t exist in crc repository)

cfergeau (commented):
You will also need some changes in the `tools` directory to fully complete the  changes:
```
diff --git a/tools/dummy.go b/tools/dummy.go
index a58944ffa..cc37744fc 100644
--- a/tools/dummy.go
+++ b/tools/dummy.go
@@ -5,7 +5,7 @@ package buildtools
 
 import (
 	_ "github.com/cfergeau/gomod2rpmdeps/cmd/gomod2rpmdeps"
-	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
+	_ "github.com/golangci/golangci-lint/v2/cmd/golangci-lint"
 	_ "github.com/randall77/makefat"
 	_ "golang.org/x/tools/cmd/goimports"
 )
diff --git a/tools/tools.mk b/tools/tools.mk
index e324e1a5c..b9b9c407b 100644
--- a/tools/tools.mk
+++ b/tools/tools.mk
@@ -4,7 +4,7 @@ $(TOOLS_BINDIR)/makefat: $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR) && GOBIN="$(TOOLS_BINDIR)" go install github.com/randall77/makefat
 
 $(TOOLS_BINDIR)/golangci-lint: $(TOOLS_DIR)/go.mod
-	cd $(TOOLS_DIR) && GOBIN="$(TOOLS_BINDIR)" go install github.com/golangci/golangci-lint/cmd/golangci-lint
+	cd $(TOOLS_DIR) && GOBIN="$(TOOLS_BINDIR)" go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint
 
 $(TOOLS_BINDIR)/gomod2rpmdeps: $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR) && GOBIN="$(TOOLS_BINDIR)" go install github.com/cfergeau/gomod2rpmdeps/cmd/gomod2rpmdeps
```
and
```
$ go get github.com/golangci/golangci-lint/v2
$ go mod tidy
$ go mod vendor
```

.golangci.yml
sebrandon1: Good question, all I did was run `golangci-lint migrate` 😄 

I'll experiment with removing the `paths` to see what happens.

sebrandon1:
Thanks everyone!

PR #4747, opened 2025-05-07
Update openshift bundle version to 4.18.12 (stable channel)

praveenkumar:
## Summary by Sourcery

Enhancements:
- Upgrade OpenShift version from 4.18.2 to 4.18.12 in Makefile


praveenkumar:
During testing looks like it still have issue for start so we are going with 4.18.2 bundles.

PR #4763, opened 2025-05-20
Don't check dns resolution for user mode networking

praveenkumar:
For user mode networking route-controller/admin-helper
take care of the routes which is created for the application. Even
till now, result of dns resolution is ignored for user mode networking
but this PR make sure it shouldn't even run instead ignoring. Since it
is ignored but host command still run then debug logs filled with
following error and after retry loop end it is ignored, with this PR
`host` command not even run for user mode network setting.

```
$ host -R 3 foo.apps-crc.testing
foo.apps-crc.testing has address 192.168.127.2
Host foo.apps-crc.testing not found: 3(NXDOMAIN)
```

## Summary by Sourcery

Bug Fixes:
- Add -t A flag to the host command invocation to prevent NXDOMAIN errors from missing non-A records

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -491,12 +491,12 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 	}
 
 	// Check DNS lookup before starting the kubelet
-	if queryOutput, err := dns.CheckCRCLocalDNSReachable(ctx, servicePostStartConfig); err != nil {
-		if !client.useVSock() {
+	logging.Info("Check internal and public DNS query...")
+	if !client.useVSock() {
+		if queryOutput, err := dns.CheckCRCLocalDNSReachable(ctx, servicePostStartConfig); err != nil {
 			return nil, errors.Wrapf(err, "Failed internal DNS query: %s", queryOutput)
 		}
 	}
-	logging.Info("Check internal and public DNS query...")
 
 	if queryOutput, err := dns.CheckCRCPublicDNSReachable(servicePostStartConfig); err != nil {
 		logging.Warnf("Failed public DNS query from the cluster: %v : %s", err, queryOutput)


anjannath (approved):

pkg/crc/services/dns/dns.go
```diff
@@ -98,7 +98,7 @@ func CheckCRCLocalDNSReachable(ctx context.Context, serviceConfig services.Servi
 	var queryOutput string
 	var err error
 	checkLocalDNSReach := func() error {
-		queryOutput, _, err = serviceConfig.SSHRunner.Run(fmt.Sprintf("host -R 3 %s", appsURI))
+		queryOutput, _, err = serviceConfig.SSHRunner.Run(fmt.Sprintf("host -t A -R 3 %s", appsURI))
```
cfergeau: Should we run this at all with usermode networking? The calling code is:
```
	// Check DNS lookup before starting the kubelet
	if queryOutput, err := dns.CheckCRCLocalDNSReachable(ctx, servicePostStartConfig); err != nil {
		if !client.useVSock() {
			return nil, errors.Wrapf(err, "Failed internal DNS query: %s", queryOutput)
		}
	}
```
so we don’t print a warning with usermode networking, but I don’t know why we run this test at all, I’m not sure we setup a hostname entry for `foo`.

cfergeau: For what it’s worth, I can’t reproduce this behaviour with a fedora 42, I added `1.2.3.4 foo.test` to `/etc/hosts` and `host -R 3 foo.test` returns `foo.test has address 1.2.3.4` and a `0` exit code.

Could this be a bug in gvisor-tap-vsock DNS code?

praveenkumar: > Should we run this at all with usermode networking? The calling code is:

we run `CheckCRCLocalDNSReachable` irrespective to network mode but in case of vsock we just ignore error and that's the reason till now it is working.

> Could this be a bug in gvisor-tap-vsock DNS code?

I am not sure if it is bug in gvisor-tap-vsock DNS code side but you can check that on crc side

cfergeau: > we run CheckCRCLocalDNSReachable irrespective to network mode but in case of vsock we just ignore error and that's the reason till now it is working.

Yes, and what I was wondering is if we should change this to:
```

    if !client.useVSock() {
	if queryOutput, err := dns.CheckCRCLocalDNSReachable(ctx, servicePostStartConfig); err != nil {
			return nil, errors.Wrapf(err, "Failed internal DNS query: %s", queryOutput)
		}
	}
```

praveenkumar: > I’m not sure we setup a hostname entry for foo.

I also thought same but when I ssh'ed to vm and checked I get correct dns resolve for `foo.apps-crc.testing` in OCP 
```
$ host -t A foo.apps-crc.testing
foo.apps-crc.testing has address 192.168.127.2
$ host -t A api.crc.testing
api.crc.testing has address 192.168.127.2
```

but when I tested same with microshift I get error
```
$ host -t A -R 3 foo.apps.crc.testing
Host foo.apps.crc.testing not found: 3(NXDOMAIN)
```

which means some how we add `*.apps-crc.testing` wildcard in usermode networking side?

praveenkumar: https://github.com/crc-org/crc/blob/main/cmd/crc/cmd/daemon.go#L88-L91 might explain it why it works for `apps-crc.testing` ?

cfergeau: > https://github.com/crc-org/crc/blob/main/cmd/crc/cmd/daemon.go#L88-L91 might explain it why it works for `apps-crc.testing` ?

Yes indeed, this should work as a wildcard DNS for `apps-crc.testing`.
It’s probably possible to do this dynamically using gvisor-tap-vsock rest API, which is underdocumented :(

cfergeau: > which means some how we add `*.apps-crc.testing` wildcard in usermode networking side?

but wildcard name resolution should not be needed if routes-controller/admin-helper are doing their job?

praveenkumar: > > which means some how we add `*.apps-crc.testing` wildcard in usermode networking side?
> 
> but wildcard name resolution should not be needed if routes-controller/admin-helper are doing their job?

@cfergeau right, on the host route-controller/admin-helper should make the route entry to `/etc/hosts` file. I will update it with https://github.com/crc-org/crc/pull/4763#discussion_r2097328952 one.

praveenkumar: @cfergeau updated.

cfergeau (approved):

PR #4766, opened 2025-05-20
[QE]fix windows10 provision failure due to generation mismatch

lilyLuLiu:
## Description

Fixes: #N

Relates to: https://github.com/redhat-developer/mapt/issues/489

## Type of change

## Contribution Checklist
    - [ ] Linux
    - [ ] Windows
    - [ ] MacOS

## Summary by Sourcery

Update Windows QE workflow matrix to use the correct Windows 10 feature pack generation and prevent provisioning failures

Bug Fixes:
- Address Windows 10 provisioning failures by fixing a generation mismatch in the feature pack selection

CI:
- Replace "22h2-ent" with "22h2-ent-g2" in the windows-featurepack matrix and exclude rules of .github/workflows/windows-qe-tpl.yml

### DIFF
--- a/.github/workflows/windows-qe-tpl.yml
+++ b/.github/workflows/windows-qe-tpl.yml
@@ -26,12 +26,12 @@ jobs:
       fail-fast: false
       matrix:
         windows-version: ['10','11']
-        windows-featurepack: ['22h2-ent', '23h2-ent']
+        windows-featurepack: ['22h2-ent-g2', '23h2-ent']
         exclude:
         - windows-version: '10'
           windows-featurepack: '23h2-ent'
         - windows-version: '11'
-          windows-featurepack: '22h2-ent'
+          windows-featurepack: '22h2-ent-g2'
 
     steps:
     - name: Download windows installer


adrianriobo:
Problem is, you still may face the issue...as it was working previously cause the vm size picked allowed gen1 ...and now it is picking gen-2....in other words when you rely on spot module you will not be sure what gen type will be supported by the vm size....

I would recommend to tackle this on mapt...checking what gen is supported by the picked vm from the spot module and the compose the name of the image with this `g2` suffix

WDYT?

lilyLuLiu:
https://github.com/redhat-developer/mapt/blob/main/pkg/provider/util/instancetypes/azure.go#L88
The vm size already only select v2. 

For win10, the sku(featurepack) has `g2` suffix. But for win11, there is no `g2` suffix(seems only support v2)

adrianriobo:
:thinking: so how is possible it previously worked?
....
Ok I guess the problem is the image itself `windows-desktop-10-22h2-ent-e2e-openshift-awd-vm` as the image is not fixed but may the latest with same feature pack...and this one had been introduced recently so now the win 10 is picking this as image...and it is not gen2 supported
....
So this PR will solve the issue here..ensuring the g2 for win10 images.... but as you said all vmsizes should be gen2...so can we fix this on mapt as well?

adrianriobo (approved):
LGTM

gbraad:
We specifically use gen2 images for Hyper-V. It might be that they previously deployed both, but this means they might have issues around secure boot.

Anyway... something to be aware of.

PR #4776, opened 2025-05-30
start: Delay pull secret on disk check to end

praveenkumar:
In current scenario MCO is patched with user provided pull secret and just after, we check if pull secret is part of disk which takes ~1 min since MCD make that change and the report to MCO. In this PR we are delaying this pull secret check to end because instead of blocking it for ~1 min better to execute other part and at the end check if pull secret is part of disk image.

With this PR (crc start time, 6 runs)
```
real	4m9.247s
user	0m0.557s
sys	0m0.165s

real	4m0.455s
user	0m0.619s
sys	0m0.168s

real	4m5.962s
user	0m0.445s
sys	0m0.154s

real	3m59.594s
user	0m0.661s
sys	0m0.179s

real	4m3.958s
user	0m0.563s
sys	0m0.177s

real	4m28.806s
user	0m0.460s
sys	0m0.171s
```

Without this PR
```
real	5m7.235s
user	0m0.797s
sys	0m0.181s

real	4m28.741s
user	0m0.891s
sys	0m0.195s

real	6m6.815s
user	0m0.747s
sys	0m0.194s

real	5m1.733s
user	0m0.395s
sys	0m0.199s

real	4m30.551s
user	0m0.466s
sys	0m0.173s

real	4m31.067s
user	0m0.673s
sys	0m0.183s
```

## Description

Fixes: #N

Relates to: #N, PR #N, ...

## Type of change

## Contribution Checklist
    - [x] Linux
    - [ ] Windows
    - [ ] MacOS

## Summary by Sourcery

Enhancements:
- Move pull secret disk presence verification to the end of the start process to prevent early blocking and improve performance

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when accessing the pull secret by ensuring connectivity checks are performed before each attempt.
  - Adjusted the sequence of operations during cluster startup to wait for cluster stabilization before verifying the presence of the pull secret.

### DIFF
--- a/pkg/crc/machine/start.go
+++ b/pkg/crc/machine/start.go
@@ -580,10 +580,6 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 		return nil, errors.Wrap(err, "Failed to update ssh public key to machine config")
 	}
 
-	if err := cluster.WaitForPullSecretPresentOnInstanceDisk(ctx, sshRunner); err != nil {
-		return nil, errors.Wrap(err, "Failed to update pull secret on the disk")
-	}
-
 	if err := cluster.UpdateUserPasswords(ctx, ocConfig, startConfig.KubeAdminPassword, startConfig.DeveloperPassword); err != nil {
 		return nil, errors.Wrap(err, "Failed to update kubeadmin user password")
 	}
@@ -612,6 +608,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig) 
 	logging.Infof("Starting %s instance... [waiting for the cluster to stabilize]", startConfig.Preset)
 	if err := cluster.WaitForClusterStable(ctx, instanceIP, constants.KubeconfigFilePath, proxyConfig); err != nil {
 		logging.Warnf("Cluster is not ready: %v", err)
+	}
+
+	if err := cluster.WaitForPullSecretPresentOnInstanceDisk(ctx, sshRunner); err != nil {
+		return nil, errors.Wrap(err, "Failed to update pull secret on the disk")
 	}
 
 	waitForProxyPropagation(ctx, ocConfig, proxyConfig)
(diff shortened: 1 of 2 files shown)


anjannath (approved):

pkg/crc/machine/start.go
```diff
@@ -614,6 +610,10 @@ func (client *client) Start(ctx context.Context, startConfig types.StartConfig)
 		logging.Warnf("Cluster is not ready: %v", err)
 	}
 
+	if err := cluster.WaitForPullSecretPresentOnInstanceDisk(ctx, sshRunner); err != nil {
```
cfergeau, line 613: can `cluster.StartMonitoring()` and `cluster.WaitForClusterStable` succeed if the pull secret is not written yet to disk?

praveenkumar: We don't have monitoring related images cached in the bundle so when user want to use monitoring operator then they have to wait for images and images can will be downloaded using the pull secret but in our case we don't dictate when this is available on the disk since MCO is responsible for it. Now assume when monitoring operator start and pull secret is not yet available on disk then it will fail and again retry but eventually succeed as soon as images are pulled.

praveenkumar: @cfergeau any thought on that?

cfergeau: If there is some code which needs the pull secret to be on disk to work, and we have a "wait until it works" function for this code, then I’d prefer if `cluster.WaitForPullSecretPresentOnInstanceDisk` was called before it.
Otherwise the "wait until it works" function will also act as a hidden `cluster.WaitForPullSecretPresentOnInstanceDisk`, and `cluster.WaitForPullSecretPresentOnInstanceDisk` will almost be a no-op.

praveenkumar: @cfergeau We don't have any code that directly depends on the pull secret being present on disk. Instead, we're executing a set of `oc` commands, some action (like monitoring or marketplace) involve pulling container images. These operations require the pull secret, and if it's not available, they can result in image pull errors. However, these errors are non-blocking and are retried, so they don't cause a complete failure.

Our current goal in this PR is to optimize startup time by checking for the presence of the pull secret on disk only at the end of the process. In fact, we could go a step further and consider removing this check entirely. Once the pull secret is patched, it's the responsibility of the Machine Config Operator (MCO) to propagate it to the disk. Our check is effectively a redundant safeguard as of now.

praveenkumar:
@coderabbitai review

praveenkumar:
@coderabbitai full review

cfergeau:
This probably can move forward?

praveenkumar:
> This probably can move forward?

It need the approval, from my end it is good to go.

cfergeau (approved):

PR #4793, opened 2025-06-16
build: add macos-precheck make path

sebrandon1:
Changes:
- Adds `macos-precheck` make path to make sure that dependencies for mac builds are checked-for.

### Build process improvements:

* **Added `macos-precheck` target**: Introduced a new target in the `Makefile` to verify the presence of `pkg-config` on macOS systems. If `pkg-config` is missing, the build process will terminate with an error message instructing the user to install it using Homebrew.
* **Updated `.PHONY` declarations**: Included the new `macos-precheck` target in the `.PHONY` section to ensure it is treated as a non-file target.

## Summary by CodeRabbit

- **Chores**
	- Added a pre-check step for macOS users to ensure required tools are installed before building. If a necessary tool is missing, users will receive a clear error message with installation instructions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ $(BUILD_DIR)/windows-amd64/crc.exe: $(SOURCES)
 $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
 	go build --tags="build" -ldflags="$(LDFLAGS)" -o $(HOST_BUILD_DIR)/crc-embedder $(GO_EXTRA_BUILDFLAGS) ./cmd/crc-embedder
 
-.PHONY: cross ## Cross compiles all binaries
+.PHONY: macos-precheck cross ## Cross compiles all binaries
 cross: $(BUILD_DIR)/macos-arm64/crc $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/windows-amd64/crc.exe
 
 .PHONY: containerized ## Cross compile from container
@@ -415,3 +415,9 @@ ADMIN_HELPER_VERSION = 0.0.12
 		 -e 's/__HELPER_SCRIPT_CHECKSUM__/'$(HELPER_SCRIPT_HASH)'/g' \
 		 -e 's/__ADMIN_HELPER_VERSION__/'$(ADMIN_HELPER_VERSION)'/g' \
 	     $< >$@
+
+.PHONY: macos-precheck
+macos-precheck:
+ifeq ($(GOOS),darwin)
+	@command -v pkg-config >/dev/null 2>&1 || { echo >&2 "Error: pkg-config is not installed. Please run 'brew install pkg-config'."; exit 1; }
+endif


pkg/crc/machine/libvirt/templates_linux.go
```diff
@@ -1,3 +1,6 @@
+//go:build linux || build
+// +build linux build
+
```
praveenkumar: doesn't `go vet` respect `_linux` as build constraints  which make sure that only build for the linux?

cfergeau:
What are you trying to achieve? If your goal is to run `go vet` on the crc codebase, this is already done through golangci-lint, `govet` is enabled by defaut. This works cross-platform somehow without the need for pkg-config on macos. I ran this on my mac after adding a bogus printf to `driver_linux.go`:
```
% make cross-lint
GOARCH=amd64 GOOS=darwin "/Users/teuf/dev/crc/tools/bin"/golangci-lint run
0 issues.
GOARCH=arm64 GOOS=darwin "/Users/teuf/dev/crc/tools/bin"/golangci-lint run
0 issues.
GOARCH=amd64 GOOS=linux "/Users/teuf/dev/crc/tools/bin"/golangci-lint run
pkg/crc/machine/libvirt/driver_linux.go:14:25: printf: fmt.Printf format %d reads arg #1, but call has 0 args (govet)
	fmt.Printf("vet error: %d")
	                       ^
1 issues:
* govet: 1
```

pkg/crc/machine/libvirt/templates_linux.go
sebrandon1: Without my change above I get this locally on my Mac:

```
go vet -tags="darwin" ./...
package github.com/crc-org/crc/v2/cmd/crc-embedder
        imports github.com/crc-org/crc/v2/cmd/crc-embedder/cmd
        imports github.com/crc-org/crc/v2/pkg/crc/machine/libvirt: build constraints exclude all Go files in /Users/bpalm/Repositories/go/src/github.com/crc-org/crc/pkg/crc/machine/libvirt
make: *** [vet] Error 1
```

And I remove the `-tags="darwin"` I get the same thing:
```
$ go vet ./...
package github.com/crc-org/crc/v2/cmd/crc-embedder
        imports github.com/crc-org/crc/v2/cmd/crc-embedder/cmd
        imports github.com/crc-org/crc/v2/pkg/crc/machine/libvirt: build constraints exclude all Go files in /Users/bpalm/Repositories/go/src/github.com/crc-org/crc/pkg/crc/machine/libvirt
```

sebrandon1:
@cfergeau My goal was to be without running the entire GolangCI suite every time I wanted to `vet` my code locally.  I just find it to be a useful development tool to fire off `go vet` when putting together some code as its free with Go.  In its current state `make vet` doesn't run during the CI runs as far as I can tell so it doesn't change the GolangCI staticcheck integration.

pkg/crc/machine/libvirt/templates_linux.go
cfergeau: Can you try this?
```
$ go vet -tags "containers_image_openpgp,build" ./...
```

sebrandon1: This seems to work!  Should I change my command to that inside of `make vet`?

cfergeau: Yes, a commit which only adds `make vet` without any other changes may be more acceptable.
I’d prefer to push people to use `make lint` which does more than just `make vet`, and is what will be blocking CI if it fails

sebrandon1: Will do!  I updated the PR to remove the changes surrounding `vet`.

Makefile
```diff
@@ -415,3 +415,9 @@ ADMIN_HELPER_VERSION = 0.0.12
 		 -e 's/__HELPER_SCRIPT_CHECKSUM__/'$(HELPER_SCRIPT_HASH)'/g' \
 		 -e 's/__ADMIN_HELPER_VERSION__/'$(ADMIN_HELPER_VERSION)'/g' \
 	     $< >$@
+
+.PHONY: macos-precheck
+macos-precheck:
+ifeq ($(GOOS),darwin)
+	@command -v pkg-config >/dev/null 2>&1 || { echo >&2 "Error: pkg-config is not installed. Please run 'brew install pkg-config'."; exit 1; }
```
cfergeau, line 422: Are you sure this is needed? I tested `go vet -tags "containers_image_openpgp,build" ./...` on macOS after removing libvirt packages and after making `pkg-config` non executable, and `go vet` succeeded.

cfergeau: Any thoughts about this? What issues are you having without it on your mac?

cfergeau: What errors do you get with `go vet -tags "containers_image_openpgp,build" ./...` if pkg-config is missing?

sebrandon1:
```
bpalm at bpalm-mac in ~
$ brew uninstall pkg-config
Uninstalling /opt/homebrew/Cellar/pkgconf/2.5.0... (28 files, 520.8KB)

bpalm at bpalm-mac in ~/Repositories/go/src/github.com/crc-org/crc on macos_build_support [?]
$ go vet -tags "containers_image_openpgp,build" ./...

bpalm at bpalm-mac in ~/Repositories/go/src/github.com/crc-org/crc on macos_build_support [?]
$ go vet ./...
package github.com/crc-org/crc/v2/cmd/crc-embedder
	imports github.com/crc-org/crc/v2/cmd/crc-embedder/cmd
	imports github.com/crc-org/crc/v2/pkg/crc/machine/libvirt: build constraints exclude all Go files in /Users/bpalm/Repositories/go/src/github.com/crc-org/crc/pkg/crc/machine/libvirt
```

Should we add a make path that wraps `go vet -tags "containers_image_openpgp,build" ./...` ?

cfergeau: > Should we add a make path that wraps go vet -tags "containers_image_openpgp,build" ./... ?

That would be `go vet --tags "build $(BUILDTAGS)"`, but since that’s already covered by `make lint`, so this `vet` Makefile target would not be used much imo. I’m not going to nack it though if someone has a strong need for it.

PR #4801, opened 2025-06-25
context: use background() consistently

sebrandon1:
In the codebase, there are 12 references to context.TODO() and 28 references to context.Background().

They both do the same thing functionally so here's a PR to make it more consistent and change all TODO references to Background.

Similar to:
- https://github.com/openshift/cert-manager-operator/pull/242
- https://github.com/openshift-kni/eco-goinfra/pull/680
- https://github.com/openshift-kni/eco-goinfra/pull/329
- https://github.com/redhat-best-practices-for-k8s/certsuite/pull/1216
- https://github.com/redhat-best-practices-for-k8s/certsuite/pull/681

## Summary by Sourcery

Enhancements:
- Replace context.TODO() with context.Background() in production code and tests for consistency

## Summary by CodeRabbit

- **Refactor**
  - Updated internal context usage for various download, extraction, and file copy operations to use a more appropriate background context. No changes to user-facing functionality or behavior.

### DIFF
--- a/cmd/crc-embedder/cmd/embed.go
+++ b/cmd/crc-embedder/cmd/embed.go
@@ -163,7 +163,7 @@ func downloadDataFiles(goos string, components []string, destDir string) ([]stri
 		if !shouldDownload(components, componentName) {
 			continue
 		}
-		filename, err := download.Download(context.TODO(), dl.url, destDir, dl.permissions, nil)
+		filename, err := download.Download(context.Background(), dl.url, destDir, dl.permissions, nil)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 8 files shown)


praveenkumar:
They looks same but have different purpose, `context.Background()` is act as base context and shouldn't be overused. `context.Todo()` is where we are not sure how the propagate the context so mostly a placeholder for revisit when possible. It looks like in our codebase we are creating `ctx` to pass to next function and not something coming from caller so in that case background would be good.

gbraad:
In short you try to say that these are not actually the same, but for our purpose (and the references given) this is not an issue... so the PR is OK. Right?

PR #4807, opened 2025-06-30
go.mod: Update openshift/k8s dep to v4.19.0 release

praveenkumar:
These changes are in sync with https://github.com/openshift/installer/blob/release-4.19/go.mod (release-4.19 branch).


go.mod
```diff
@@ -35,9 +35,9 @@ require (
 	github.com/onsi/ginkgo/v2 v2.23.4
 	github.com/onsi/gomega v1.37.0
 	github.com/opencontainers/image-spec v1.1.0
-	github.com/openshift/api v0.0.0-20240517002838-931fa9860c01
-	github.com/openshift/client-go v0.0.0-20231218140158-47f6d749b9d9
-	github.com/openshift/library-go v0.0.0-20240207105404-126b47137408
+	github.com/openshift/api v0.0.0-20250313134101-8a7efbfb5316
+	github.com/openshift/client-go v0.0.0-20241203091221-452dfb8fa071
```
cfergeau: I’d tend to pick `github.com/openshift/client-go@release-4.19` rather than the older commit the openshift installer is using?

go.mod
```diff
@@ -62,9 +62,9 @@ require (
 	golang.org/x/text v0.27.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.29.3
-	k8s.io/apimachinery v0.29.3
-	k8s.io/client-go v0.29.3
+	k8s.io/api v0.32.3
+	k8s.io/apimachinery v0.32.3
+	k8s.io/client-go v0.32.1
```
cfergeau, line 67: `k8s.io/client-go` goes up to `v0.32.7`, I think it’s ok to use something newer. I haven’t checked the other ones.

cfergeau: We can add this to the dependabot config to get PRs to automatically update these to the latest:
```
diff --git a/.github/dependabot.yml b/.github/dependabot.yml
index 3e5ea4e93..9faf3582d 100644
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,7 @@ updates:
   ignore:
     - dependency-name: "github.com/openshift/*"
     - dependency-name: "k8s.io/*"
+      update-types: ["version-update:semver-minor"]
 - package-ecosystem: gomod
   directory: "/tools"
   schedule:
```

This is done already in `routes-controller`: https://github.com/crc-org/routes-controller/pulls

go.mod
praveenkumar: I just use same what installer have for that specific release to avoid any surprise. That is working for last few updates on go mod, we can try to use branch specific but I think it will again use latest commit from branch.

go.mod
praveenkumar: I want to use same version of k8s client-go which is part of https://github.com/openshift/client-go/blob/release-4.19/go.mod or in openshift-installer to be compatible with.

go.mod
cfergeau: >  we can try to use branch specific but I think it will again use latest commit from branch.

yes, it will use the latest commit from the branch, which is newer than the one you used. The commits used vary depending on the module you look at, `oc` uses a different set of commits, some newer, some older https://github.com/openshift/oc/blob/release-4.19/go.mod#L36-L39
This is why I’d tend to use the latest commits from the 4.19 branch.

go.mod
cfergeau: As before, I’m not sure the exact versions used are fully relevant as they vary between openshift components
https://github.com/openshift/oc/blob/298429ba9831d1d72b89edd9beb82a6ee665c3b7/go.mod#L54-L60

cfergeau: Red Hat only communicates about the minor version, not the patch release, so I would say any is fine https://access.redhat.com/solutions/4870701

praveenkumar: @cfergeau I updated all the openshift specific dep to release-4.19 branch and this is now auto updated to those versions. I think we should keep those as it is.
```
$ go get -x github.com/openshift/client-go@release-4.19
[...]
cd /home/prkumar/go/pkg/mod/cache/vcs/29f1da1de2f7957317afabb34fca40b09e7d35d3d5b2545a6e560a23daf9ae2b; git cat-file blob a56162b4534ef3d1d7eb12ce8757edeae8e655b1:go.mod
0.002s # cd /home/prkumar/go/pkg/mod/cache/vcs/29f1da1de2f7957317afabb34fca40b09e7d35d3d5b2545a6e560a23daf9ae2b; git cat-file blob a56162b4534ef3d1d7eb12ce8757edeae8e655b1:go.mod
go: upgraded k8s.io/api v0.29.3 => v0.32.3
go: upgraded k8s.io/apimachinery v0.29.3 => v0.32.3
go: upgraded k8s.io/client-go v0.29.3 => v0.32.1

```

cfergeau:
There are unrelated issues with snyk, as if the snyk binary was run on a too old distro or something like that:
```
 Starting snyk code scan
/alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux)
/alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux)
/alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux)
/alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux)
/alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux)
/alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /alabama/.cache/snyk/snyk-cli/1.1298.0/snyk-linux) 
```

cfergeau (approved):

PR #4814, opened 2025-07-03
return error when no tag or `latest` tag is specified

### LINKED ISSUE #4520: Show a warning when using `latest` tag
We don't use `latest` tag in container image repositories, so I suggest adding a check to display a warning if the user tries to use it.

```
$ crc setup -b docker://quay.io/crcont/openshift-bundle:latest
WARN 'latest' tag is not supported; use a specific version
```

This could also be converted to an error, but I think that is too restrictive to the user.

redbeam:
Fixes: #4520 

## Type of change

## Testing
```
$ crc setup -b docker://quay.io/crcont/openshift-bundle
ERRO No tag found in bundle URI

$ crc setup -b docker://quay.io/crcont/openshift-bundle:
ERRO No tag found in bundle URI

$ crc setup -b docker://quay.io/crcont/openshift-bundle:latest
ERRO 'latest' tag is not supported; use a specific version
```

## Contribution Checklist
    - [ ] Linux
    - [ ] Windows
    - [x] MacOS

## Summary by Sourcery

Require explicit version tags in bundle URIs by rejecting empty or 'latest' tags and updating tests accordingly.

Bug Fixes:
- Return an error when bundle URIs lack a version tag or explicitly use the 'latest' tag.

Tests:
- Add unit tests to verify error handling for URIs with no tag and with the 'latest' tag.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation for Docker bundle URIs to reject empty tags and the "latest" tag, requiring explicit version tags.

* **Tests**
  * Added test cases to confirm specific error responses for Docker URIs missing tags or using the "latest" tag.

### DIFF
--- a/pkg/crc/machine/bundle/metadata_test.go
+++ b/pkg/crc/machine/bundle/metadata_test.go
@@ -316,10 +316,23 @@ func TestGetFQDN(t *testing.T) {
 }
 
 func TestGetBundleNameFromURI(t *testing.T) {
+	var noTagErr = &NoTagError{}
+	var unsupportedTagErr = &UnsupportedTagError{}
+
 	// URI with no tag
 	bundleName, err := GetBundleNameFromURI("docker://quay.io/crcont/openshift-bundle")
 	assert.Equal(t, "", bundleName)
-	assert.Error(t, err)
+	assert.ErrorAs(t, err, &noTagErr)
+
+	// URI with colon but no tag
+	bundleName, err = GetBundleNameFromURI("docker://quay.io/crcont/openshift-bundle:")
+	assert.Equal(t, "", bundleName)
+	assert.ErrorAs(t, err, &noTagErr)
+
+	// URI with 'latest' tag
+	bundleName, err = GetBundleNameFromURI("docker://quay.io/crcont/openshift-bundle:latest")
+	assert.Equal(t, "", bundleName)
+	assert.ErrorAs(t, err, &unsupportedTagErr)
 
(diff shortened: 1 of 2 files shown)


pkg/crc/machine/bundle/metadata_test.go
```diff
@@ -321,6 +321,16 @@ func TestGetBundleNameFromURI(t *testing.T) {
 	assert.Equal(t, "", bundleName)
 	assert.Error(t, err)
 
+	// URI with colon but no tag
+	bundleName, err = GetBundleNameFromURI("docker://quay.io/crcont/openshift-bundle:")
+	assert.Equal(t, "", bundleName)
+	assert.Error(t, err)
```
praveenkumar: better to match the error string which we are expecting instead just asserting it as error.

cfergeau: Or follow the recommendation from CodeRabbit even if this is not what we do in the rest of the codebase:
> Consider defining and returning distinct error variables (e.g. ErrNoTag, ErrUnsupportedLatest) instead of fmt.Errorf so callers and tests can match on error values rather than error text.

redbeam: I added two error types (`NoTagError` and `UnsupportedTagError`) that are returned. Alternatively, they could be merged into a generic `TagError` type.

praveenkumar:
@coderabbitai review

PR #4928, opened 2025-09-11
Makefile: Improve containerized target

praveenkumar:
As of now for this target we are building image first and then copy the binaries out from container to host. This PR uses same base image which openshift-ci using and mount the current working directory and execute cross builds so that building image is not required and also binary directly available to host.

## Summary by CodeRabbit

- Chores
  - Streamlined the containerized build workflow into a single run step, reducing setup and cleanup overhead and improving local/CI reliability.
- Refactor
  - Build target now automatically derives the container image to minimize manual configuration and maintenance.
- Style
  - Minor build script cleanups for clarity.

No user-facing changes; application behavior remains the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,12 +113,9 @@ $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
 cross: $(BUILD_DIR)/macos-arm64/crc $(BUILD_DIR)/macos-amd64/crc $(BUILD_DIR)/linux-amd64/crc $(BUILD_DIR)/linux-arm64/crc $(BUILD_DIR)/windows-amd64/crc.exe
 
 .PHONY: containerized ## Cross compile from container
+containerized: image := $(shell grep -m 1 '^FROM' images/openshift-ci/Dockerfile | awk '{print $$2}')
 containerized: clean
-	${CONTAINER_RUNTIME} build -t crc-build -f images/build .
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -113,12 +113,9 @@ $(HOST_BUILD_DIR)/crc-embedder: $(SOURCES)
...
 
 .PHONY: containerized ## Cross compile from container
+containerized: image := $(shell grep -m 1 '^FROM' images/openshift-ci/Dockerfile | awk '{print $$2}')
 containerized: clean
-	${CONTAINER_RUNTIME} build -t crc-build -f images/build .
-	${CONTAINER_RUNTIME} run --name crc-cross crc-build make cross
-	${CONTAINER_RUNTIME} cp crc-cross:/opt/app-root/src/out ./
-	${CONTAINER_RUNTIME} rm crc-cross
-	${CONTAINER_RUNTIME} rmi crc-build
+	${CONTAINER_RUNTIME} run --rm -v ${PWD}:/data:Z ${image} /bin/bash -c "cd /data && make cross"
```
cfergeau: `images/build/*` should be removed then?

praveenkumar: @cfergeau I think that image is still used by QE, @lilyLuLiu or @albfan can confirm that.

albfan (approved):
Makes sense, but should we add a new target as all the other ontainerized-* just build and new image? containerized_e2e, containerized_integration

PR #4935, opened 2025-09-18
[QE] manage absence of xml filter to process xml result

albfan:
## Description

On RHEL pre releases xlstproc is failing to install due to unsigned package libxslt

This provide a fallback to install unsigned package, and control if finally command is available

## Summary by CodeRabbit

- Bug Fixes
  - Test report generation no longer fails when a required system tool is missing; the script attempts installation with a retry and, if unavailable, falls back to writing raw results while issuing a warning.
  - Conditional transformation ensures JUnit output is produced when possible, improving reliability of e2e results.

- Chores
  - Updated e2e build script to detect tool availability at runtime and apply transformation only when supported, simplifying control flow and removing the previous unconditional processing path.

### DIFF
--- a/images/build-e2e/lib/linux/run.sh
+++ b/images/build-e2e/lib/linux/run.sh
@@ -63,10 +63,16 @@ cd $targetFolder/bin
 cd ..
 init_line=$(grep -n '<?xml version="1.0" encoding="UTF-8"?>' results/e2e.results | awk '{split($0,n,":"); print n[1]}')
 if ! command -v xsltproc &>/dev/null
-then 
-  sudo yum install -y xsltproc || echo "Warning: Failed to install xsltproc"
+then
+  sudo yum install -y xsltproc || sudo yum install -y --nogpgcheck xsltproc || echo "Warning: Failed to install xsltproc"
 fi
-tail -n +$init_line results/e2e.results | xsltproc filter.xsl - > results/$junitFilename
+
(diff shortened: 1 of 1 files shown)


images/build-e2e/lib/linux/run.sh
```diff
@@ -63,10 +63,16 @@ cd $targetFolder/bin
 cd ..
 init_line=$(grep -n '<?xml version="1.0" encoding="UTF-8"?>' results/e2e.results | awk '{split($0,n,":"); print n[1]}')
 if ! command -v xsltproc &>/dev/null
-then 
-  sudo yum install -y xsltproc || echo "Warning: Failed to install xsltproc"
+then
+  sudo yum install -y xsltproc || sudo yum install -y --nogpgcheck xsltproc || echo "Warning: Failed to install xsltproc"
```
praveenkumar, line 67: why we want to have a non signed package installed?

albfan: We are waiting for pre release RHEL to confirm if recent non signed packages is a mistake.

This is just a workaround, but probably we will refuse this change

anjannath: @albfan so this can be closed, or we still need this change?

praveenkumar:
@albfan I think we should close this one now since it should be now available from official channel?

PR #4977, opened 2025-10-22
[QE]monitor cpu usage in e2e test case

lilyLuLiu:
## Description

Fixes: #N

Relates to: #N, PR #N, ...

## Type of change

- [] Breaking change (fix or feature that would cause existing functionality to change

## Contribution Checklist
    - [x] Linux
    - [ ] Windows
    - [x] MacOS

## Summary by CodeRabbit

* **Tests**
  * Added @performance tags to end-to-end scenarios.
  * Removed CPU checks while retaining memory validations.
  * Record timestamps at deployment, stop, and restart points for precise timing.

* **Chores**
  * Added background performance monitor for performance runs.
  * Timestamped and performance data are logged for post-run analysis.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/test/e2e/testsuite/performance.go
+++ b/test/e2e/testsuite/performance.go
@@ -1,0 +1,112 @@
+package testsuite
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/crc-org/crc/v2/test/extended/util"
+	"github.com/shirou/gopsutil/v4/cpu"
+)
+
+type Monitor struct {
+	cancelFunc context.CancelFunc
+	isRunning  bool
+	mu         sync.Mutex
+	wg         sync.WaitGroup
+	interval   time.Duration
+}
+
+func NewMonitor(interval time.Duration) *Monitor {
+	return &Monitor{
+		interval: interval,
+	}
+}
+
+func (m *Monitor) Start() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.isRunning {
+		return fmt.Errorf("The collector is running")
+	}
+
+	fmt.Printf("Attempt to start CPU collector, interval: %s\n", m.interval)
+
+	//  create a context.WithCancel
+	ctx, cancel := context.WithCancel(context.Background())
+	m.cancelFunc = cancel
+	m.isRunning = true
+
+	// start goroutine
+	m.wg.Add(1)
+	go m.collectLoop(ctx)
+
+	fmt.Println("CPU collector has been successfully started")
+	return nil
+}
+
+func (m *Monitor) Stop() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.isRunning {
+		return fmt.Errorf("The collector is not running")
+	}
+	if m.cancelFunc != nil {
+		m.cancelFunc()
+	}
+	m.isRunning = false
+	m.wg.Wait()
+	fmt.Println("CPU collector has sent a stop signal")
+	// may need wait a while to stop
+	return nil
+}
+
+func (m *Monitor) collectLoop(ctx context.Context) {
+	defer m.wg.Done()
+
+	fmt.Println("--> collect goroutine start...")
+	calcInterval := m.interval
+
+	// Prime cpu.Percent so the subsequent calls with interval=0 return usage
+	// since this priming call. We discard the returned value.
+	_, _ = cpu.Percent(0, false)
+
+	ticker := time.NewTicker(calcInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			fmt.Println("<-- collect goroutine receive stop signal")
+			return // exit goroutine
+		case <-ticker.C:
+			// cpu.Percent with interval=0 returns percent since the last call.
+			totalPercent, err := cpu.Percent(0, false)
+			if err != nil {
+				fmt.Printf("Error: fail to collect CPU data: %v\n", err)
+				// do not block here; wait for next tick or cancellation
+				continue
+			}
+
+			if len(totalPercent) > 0 {
+				data := fmt.Sprintf("[%s], cpu percent: %.2f%%\n",
+					time.Now().Format("15:04:05"), totalPercent[0])
+				wd, err := os.Getwd()
+				if err != nil {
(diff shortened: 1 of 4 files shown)


test/e2e/testsuite/testsuite.go
```diff
@@ -398,6 +403,17 @@ func InitializeScenario(s *godog.ScenarioContext) {
...
+			if tag.Name == "@performance" {
+				err := getTimestamp("finish")
+				if err != nil {
+					fmt.Printf("Failed to get finish getTimestamp: %v\n", err)
+				}
+				if err := monitor.Stop(); err != nil {
+					fmt.Printf("Failed to stop monitoring: %v\n", err)
+				}
+				fmt.Printf("Collection has stopped. Wait for 5 seconds to confirm that the collection task will no longer output data\n")
+				time.Sleep(5 * time.Second)
```
kpouget: if possible, better avoid any explicit (long) `Sleep` 
maybe something like
```
while ! monitor.Finished():
  sleep 0.1
```
or even, assume it will finish safely and just skip the sleep?

test/e2e/testsuite/performance.go
```diff
@@ -0,0 +1,123 @@
...
+
+		if len(totalPercent) > 0 {
+			data := fmt.Sprintf("[%s], cpu percent: %.2f%%\n",
+				time.Now().Format("15:04:05"), totalPercent[0])
+			wd, err := os.Getwd()
+			if err != nil {
+				fmt.Printf("Error: failed to get working directory: %v\n", err)
+				continue
+			}
+			file := filepath.Join(wd, "../test-results/cpu-consume.txt")
```
kpouget: I'm wondering about this hardcoded path. Won't new calls overwrite the old data?
or add a comment to say that this is safe, because the current directory is guaranteed to be clean

(and I'm surprized by the initial `../`, I usually prefer when everything is stored inside the same directory (but it again depends on how `.` is controlled))

test/e2e/testsuite/performance.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (m *Monitor) collectLoop(ctx context.Context) {
+	defer m.wg.Done()
+
+	fmt.Println("--> collect goroutine start...")
+	calcInterval := m.interval
+
+	for {
+		// 1. check Context whether be cancled
```
kpouget: typo

test/e2e/testsuite/performance.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	fmt.Printf("Attempt to start CPU collector, interval: %s\n", m.interval)
+
+	//  create a context.WithCancel
+	ctx, cancel := context.WithCancel(context.Background())
+	m.cancelFunc = cancel
+	m.isRunning = true
+
+	// stary goroutine
```
kpouget: typo

test/e2e/testsuite/performance.go
```diff
@@ -0,0 +1,123 @@
...
+				fmt.Printf("Error: failed to collect memory data: %v", err)
+				continue
+			}
+			memoryused := vMem.Used / 1024 / 1024
+			data := fmt.Sprintf("[%s], MemoryUsed(MB): %d\n" ,
+					time.Now().Format("15:04:05"), memoryused)
+			wd, _ := os.Getwd()
+			file := filepath.Join(wd, "../test-results/memory-consume.txt")
+			util.WriteToFile(data, file)
+		*/
```
kpouget: I think this comment should be removed, it's not the write place to measure the memory usage, 
this is after the blocking call to `cpu.Percent(calcInterval, false)`, and I think there's usually no need to collect the memory as often as the CPU usage

test/e2e/testsuite/performance.go
lilyLuLiu: @kpouget, all test results will be uploaded to S3 in the test pipeline, and the entire test folder will be deleted on the bare metal. So there's no need to worry about overwriting old data.

test/e2e/testsuite/performance.go
lilyLuLiu: removed this part

test/e2e/testsuite/testsuite.go
lilyLuLiu: Yes, you are right. I have removed the sleep code.

test/e2e/testsuite/performance.go
```diff
@@ -0,0 +1,110 @@
...
+				continue
+			}
+			file := filepath.Join(wd, "../test-results/cpu-consume.txt")
+			err = util.WriteToFile(data, file)
+			if err != nil {
+				fmt.Printf("Error: fail to write to %s: %v\n", file, err)
+			}
+		}
+	}
+}
```
praveenkumar, line 112: Got this suggestion by copiliot which I think make more sense to use ticker then implement our own interval method?

```golang
func (m *Monitor) collectLoop(ctx context.Context) {
	defer m.wg.Done()

	fmt.Println("--> collect goroutine start...")
	calcInterval := m.interval

	// Prime cpu.Percent so the subsequent calls with interval=0 return usage
	// since this priming call. We discard the returned value.
	_, _ = cpu.Percent(0, false)

	ticker := time.NewTicker(calcInterval)
	defer ticker.Stop()

	for {
		select {
		case <-ctx.Done():
			fmt.Println("<-- collect goroutine receive stop signal")
			return // exit goroutine
		case <-ticker.C:
			// cpu.Percent with interval=0 returns percent since the last call.
			totalPercent, err := cpu.Percent(0, false)
			if err != nil {
				fmt.Printf("Error: fail to collect CPU data: %v\n", err)
				// do not block here; wait for next tick or cancellation
				continue
			}

			if len(totalPercent) > 0 {
				data := fmt.Sprintf("[%s], cpu percent: %.2f%%\n",
					time.Now().Format("15:04:05"), totalPercent[0])
				wd, err := os.Getwd()
				if err != nil {
					fmt.Printf("Error: failed to get working directory: %v\n", err)
					continue
				}
				file := filepath.Join(wd, "../test-results/cpu-consume.txt")
				err = util.WriteToFile(data, file)
				if err != nil {
					fmt.Printf("Error: fail to write to %s: %v\n", file, err)
				}
			}
		}
	}
}
```

lilyLuLiu: @praveenkumar Thanks, I've commit the code change according to your suggestion.

PR #4998, opened 2025-11-06
Remove workaround as part of stop command

praveenkumar:
Older version of OCP had issue which causing the issue for graceful shutdown to longer than usual. This was fixed on OCP side but we still carry workaround. Recently we observe that because of this workaround ssh connection also affected and stop process hangs.

My understanding is `crictl stop $(crictl ps -q)` command stop all the container which also include the ovn/network containers and because of that `br-ex` is removed (bridge network) and tap0 doesn't take ownership back from br-ex so might be that's the reason the ssh connection is broken. This is only happen sometime but when this happen it make all the process hang.

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified machine stop operation by removing OpenShift-specific container shutdown handling that previously ran prior to VM termination. When stopping a machine, the VM now terminates directly without attempting to stop OpenShift containers via SSH. Kubeconfig cleanup and post-stop status reporting operations continue to function normally during the shutdown sequence.

### DIFF
--- a/pkg/crc/machine/stop.go
+++ b/pkg/crc/machine/stop.go
@@ -5,8 +5,6 @@ import (
 
 	"github.com/crc-org/crc/v2/pkg/crc/logging"
 	"github.com/crc-org/crc/v2/pkg/crc/machine/state"
-	crcPreset "github.com/crc-org/crc/v2/pkg/crc/preset"
-	"github.com/crc-org/crc/v2/pkg/crc/systemd"
 	"github.com/pkg/errors"
 )
 
@@ -25,12 +23,6 @@ func (client *client) Stop() (state.State, error) {
 		return state.Error, errors.Wrap(err, "Cannot load machine")
 	}
 	defer vm.Close()
-	if client.GetPreset() == crcPreset.OpenShift {
-		if err := stopAllContainers(vm); err != nil {
-			logging.Warnf("Failed to stop all OpenShift containers.\nShutting down VM...")
-			logging.Debugf("%v", err)
-		}
-	}
 	logging.Info("Stopping the instance, this may take a few minutes...")
 	if err := vm.Stop(); err != nil {
 		status, stateErr := vm.State()
@@ -48,27 +40,4 @@ func (client *client) Stop() (state.State, error) {
 		return status, unexposePorts()
 	}
 	return status, nil
-}
-
-// This should be removed after https://bugzilla.redhat.com/show_bug.cgi?id=1965992
-// is fixed. We should also ignore the openshift specific errors because stop
-// operation shouldn't depend on the openshift side. Without this graceful shutdown
-// takes around 6-7 mins.
-func stopAllContainers(vm *virtualMachine) error {
-	logging.Info("Stopping kubelet and all containers...")
-	sshRunner, err := vm.SSHRunner()
-	if err != nil {
-		return errors.Wrapf(err, "Error creating the ssh client")
-	}
-	defer sshRunner.Close()
-
-	if err := systemd.NewInstanceSystemdCommander(sshRunner).Stop("kubelet"); err != nil {
-		return err
-	}
-	_, stderr, err := sshRunner.RunPrivileged("stopping all containers", `-- sh -c 'crictl stop $(crictl ps -q)'`)
-	if err != nil {
-		logging.Errorf("Failed to stop all containers: %v - %s", err, stderr)
-		return err
-	}
-	return nil
 }


cfergeau:
> My understanding is `crictl stop $(crictl ps -q)` command stop all the container which also include the ovn/network containers and because of that `br-ex` is removed (bridge network) and tap0 doesn't take ownership back from br-ex so might be that's the reason the ssh connection is broken. This is only happen sometime but when this happen it make all the process hang.

It’s unexpected that it only happens sometimes though. I’d try to reproduce with `crictl stop $NETWORK_CONTAINERS` instead of `$(crictl ps -q)` to try to narrow down this bug.

anjannath:
> > My understanding is `crictl stop $(crictl ps -q)` command stop all the container which also include the ovn/network containers and because of that `br-ex` is removed (bridge network) and tap0 doesn't take ownership back from br-ex so might be that's the reason the ssh connection is broken. This is only happen sometime but when this happen it make all the process hang.
> 
> It’s unexpected that it only happens sometimes though. I’d try to reproduce with `crictl stop $NETWORK_CONTAINERS` instead of `$(crictl ps -q)` to try to narrow down this bug.

seems t the containers/pods would just restart, I attempted 3 times, and networking was uninterrupted through out, so it is hard to reproduce

```
[core@crc ~]$ sudo crictl ps --name ovnk
CONTAINER           IMAGE                                                              CREATED             STATE               NAME                     ATTEMPT             POD ID              POD                                      NAMESPACE
4a33d4ff130eb       7e58f8eb42745d6598d11a9f8db8d9531a967002e29bc1cb8f147a94fced9d7b   11 minutes ago      Running             ovnkube-controller       1                   159e85cac717a       ovnkube-node-zpgrd                       openshift-ovn-kubernetes
af42c7c8156b4       7e58f8eb42745d6598d11a9f8db8d9531a967002e29bc1cb8f147a94fced9d7b   11 minutes ago      Running             ovnkube-cluster-manager  1                   ac4ab85505c45       ovnkube-control-plane-584fdd857d-fqdvp   openshift-ovn-kubernetes

[core@crc ~]$ sudo crictl stop 4a33d4ff130eb
4a33d4ff130eb
[core@crc ~]$ sudo crictl stop af42c7c8156b4
af42c7c8156b4

[core@crc ~]$ sudo crictl ps --name ovnk
CONTAINER           IMAGE                                                              CREATED             STATE               NAME                     ATTEMPT             POD ID              POD                                      NAMESPACE
75a26926915ae       7e58f8eb42745d6598d11a9f8db8d9531a967002e29bc1cb8f147a94fced9d7b   3 seconds ago       Running             ovnkube-cluster-manager  2                   ac4ab85505c45       ovnkube-control-plane-584fdd857d-fqdvp   openshift-ovn-kubernetes
38b0c56741223       7e58f8eb42745d6598d11a9f8db8d9531a967002e29bc1cb8f147a94fced9d7b   11 seconds ago      Running             ovnkube-controller       2                   159e85cac717a       ovnkube-node-zpgrd                       openshift-ovn-kubernetes
```

praveenkumar:
> It’s unexpected that it only happens sometimes though. I’d try to reproduce with crictl stop $NETWORK_CONTAINERS instead of $(crictl ps -q) to try to narrow down this bug.

@cfergeau As I said it is only happen sometime not sure what triggers it. I even tried following but not able to reproduce and `br-ex` still present so when this occur `br-ex` was not present.

```
sudo systemctl stop kubelet
sudo crictl stop $(sudo crictl ps | grep network | cut -d ' ' -f 1)
sudo crictl stop $(sudo crictl ps | grep ovn | cut -d ' ' -f 1)
```

>seems t the containers/pods would just restart, I attempted 3 times, and networking was uninterrupted through out, so it is hard to reproduce

@anjannath you need to stop the operator which controller the deployment try above and it will not restart.

cfergeau:
> > It’s unexpected that it only happens sometimes though. I’d try to reproduce with crictl stop 
> >   N
> >   E
> >   T
> >   W
> >   O
> >   R
> >   
> >     K
> >     C
> >   
> >   O
> >   N
> >   T
> >   A
> >   I
> >   N
> >   E
> >   R
> >   S
> >   i
> >   n
> >   s
> >   t
> >   e
> >   a
> >   d
> >   o
> >   f
> > (crictl ps -q) to try to narrow down this bug.
> 
> @cfergeau As I said it is only happen sometime not sure what triggers it. I even tried following but not able to reproduce and `br-ex` still present so when this occur `br-ex` was not present.
> 
> ```
> sudo systemctl stop kubelet
> sudo crictl stop $(sudo crictl ps | grep network | cut -d ' ' -f 1)
> sudo crictl stop $(sudo crictl ps | grep ovn | cut -d ' ' -f 1)
> ```
> 
> > seems t the containers/pods would just restart, I attempted 3 times, and networking was uninterrupted through out, so it is hard to reproduce
> 
> @anjannath you need to stop the operator which controller the deployment try above and it will not restart.

Ah maybe it’s `crictl stop $(sudo crictl ps)` which was buggy in the first place, and some containers are restarted while it’s running. In particular, network-related containers could get restarted while `crictl stop` is running? When you cannot reproduce, after running `crictl stop $(sudo crictl ps)`, are there still some containers running?

praveenkumar:
> Ah maybe it’s `crictl stop $(sudo crictl ps)` which was buggy in the first place, and some containers are restarted while it’s running. In particular, network-related containers could get restarted while `crictl stop` is running? 

I doubt that because when I executed it containers didn't get restart.

> When you cannot reproduce, after running `crictl stop $(sudo crictl ps)`, are there still some containers running?

@cfergeau no, there was no containers running after this command execution succeed.

anjannath:
> @anjannath you need to stop the operator which controller the deployment try above and it will not restart.

yes, after stopping the `kubelet` service and then removing the containers it doesn't restart them again

anjannath (approved):
since we are removing a workaround that is not needed anymore

PR #5011, opened 2025-11-12
CI: Update macos workflow to latest macos

praveenkumar:
Looks like macos-13 is now depricated so replace those with macos-15 and macos-26.

## Release Notes

* **Chores**
  * Updated macOS runner versions in CI/CD workflows to newer versions (macOS-15-intel, macOS-15, and macOS-26), ensuring builds run on current environments.

### DIFF
--- a/.github/workflows/macos-installer.yml
+++ b/.github/workflows/macos-installer.yml
@@ -17,8 +17,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macOS-13
-          - macOS-14
+          - macOS-15-intel
+          - macOS-15
         go:
           - '1.24'
     steps:
--- a/.github/workflows/make-check.yml
+++ b/.github/workflows/make-check.yml
@@ -11,8 +11,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macOS-13
-          - macOS-14
+          - macOS-15
+          - macOS-26
           - ubuntu-latest
           - ubuntu-22.04
         go:


.github/workflows/macos-installer.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macOS-13
-          - macOS-14
+          - macOS-15-intel
+          - macOS-15
```
cfergeau, line 21: this is preexisting, but `macos-15` is a more common spelling for runner names

.github/workflows/macos-installer.yml
cfergeau, line 44: If this is a universal installer, why do we need to run it both on intel and apple silicon?

praveenkumar: to make sure it works on intel and silicon both ?

.github/workflows/macos-installer.yml
praveenkumar: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories shows `macos-15` by default is arm and `macos-15-intel` is the intel one.

cfergeau: What I suggest is `s/macOS/macos`

.github/workflows/make-check.yml
```diff
@@ -11,8 +11,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macOS-13
-          - macOS-14
+          - macOS-15
+          - macOS-26
```
cfergeau, line 15: Same suggestion as in https://github.com/crc-org/crc/pull/5011/commits/3391f98371dd082877ddc4ceaa0b85850fe32c7a#r2526670082

.github/workflows/macos-installer.yml
cfergeau: Building/uploading the installer on both platforms is not really useful imo, as long as we can build it on apple silicon, we’re good.
But after that there are also a few steps which install the `.pkg` and run `crc setup` which I hadn’t noticed.
Imo the installer build and the installer tests on Apple Silicon/Intel should be separate, we build one installer on Apple Silicon, and we test it on the 2 platforms, it’s closer to how we are going to release it.
And I’m not sure running `crc setup` is great, as it’s going to download the bundle, which can take time.

anjannath: @coderabbitai make a followup issue for this and the suggestion in https://github.com/crc-org/crc/pull/5011#discussion_r2526670082

anjannath (approved):

PR #5025, opened 2025-11-28
[WIP] Initial PR for consuming gvproxy/macadam/cloud-init for crc-ng

praveenkumar:


pkg/crc/cache/cache.go
```diff
@@ -84,6 +84,24 @@ func NewAdminHelperCache() *Cache {
 	)
 }
 
+func NewGvproxyCache() *Cache {
+	url := constants.GetGvproxyURL()
+	version := version.GetGvproxyVersion()
+	return newCache(constants.GvproxyPath(),
+		url,
+		version,
+		func(executable string) (string, error) {
```
cfergeau, line 94: `NewAdminHelperCache` uses exactly the same function, might be desirable to share that code.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -201,3 +216,41 @@ func removePodmanFromOcBinDirCheck() Check {
...
+
+	gvproxy := cache.NewGvproxyCache()
+	if !gvproxy.IsCached() {
+		return errors.New("gvproxy executable is not cached")
+	}
+	if err := gvproxy.CheckVersion(); err != nil {
+		return errors.Wrap(err, "unexpected version of the gvproxy executable")
+	}
+	logging.Debug("gvproxy executable already cached")
+	// SUID is only required on Linux (checkSuid is a no-op on Windows/macOS)
```
cfergeau: Are you sure it’s not also needed on macOS? Could be better to split this check in 2, with the SUID check being linux-specific. We would get 2 separate `skip-`  options, which would allow to skip one without skipping the other (eg skip the suid check, but keep the gvproxy download).

pkg/crc/cache/cache.go
```diff
@@ -102,6 +102,24 @@ func NewGvproxyCache() *Cache {
...
+			out, _, err := crcos.RunWithDefaultLocale(executable, "--version")
+			if err != nil {
+				return "", err
+			}
+			// macadam version output format: "macadam version v0.2.0"
+			split := strings.Split(out, " ")
+			return strings.TrimSpace(split[len(split)-1]), nil
```
cfergeau, line 121: Same comment as for `NewAdminHelperCache` and `NewGvproxyCache`

pkg/crc/constants/constants.go
```diff
@@ -30,6 +30,7 @@ const (
 	CrcLandingPageURL         = "https://console.redhat.com/openshift/create/local" // #nosec G101
 	DefaultAdminHelperURLBase = "https://github.com/crc-org/admin-helper/releases/download/v%s/%s"
 	DefaultGvproxyURLBase     = "https://github.com/containers/gvisor-tap-vsock/releases/download/%s/%s"
+	DefaultMacadamURLBase     = "https://github.com/praveenkumar/macadam/releases/download/%s/%s"
```
cfergeau, line 33: This can be changed now that there is a 0.3.0 release.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -254,3 +266,33 @@ func fixGVProxyExecutableCached() error {
...
+	if version.IsInstaller() {
+		return nil
+	}
+
+	macadam := cache.NewMacadamCache()
+	if !macadam.IsCached() {
+		return errors.New("macadam executable is not cached")
+	}
+	if err := macadam.CheckVersion(); err != nil {
+		return errors.Wrap(err, "unexpected version of the macadam executable")
```
cfergeau: Let’s not add new uses of `errors.Wrap` in new code.
https://github.com/pkg/errors
>  This repository was archived by the owner on Dec 1, 2021. It is now read-only.

pkg/crc/preflight/preflight_linux.go
```diff
@@ -368,6 +368,7 @@ func getChecks(distro *linux.OsRelease, bundlePath string, preset crcpreset.Pres
 	checks = append(checks, genericPreflightChecks(preset)...)
 	checks = append(checks, memoryCheck(preset))
 	checks = append(checks, gvproxyCheck())
+	checks = append(checks, macadamCheck())
```
cfergeau, line 222: It’s imo a bit odd to have separate `gvproxyCheck()`, `macadamCheck()`, … This could go in one of the existing lists of checks

pkg/crc/constants/constants.go
```diff
@@ -192,7 +192,12 @@ func AdminHelperPath() string {
 }
 
 func GvproxyPath() string {
-	return ResolveHelperPath(GetGvproxyExecutableForOs(runtime.GOOS))
+	// gvproxy is renamed to just "gvproxy" (or "gvproxy.exe" on Windows) for macadam compatibility
+	executableName := "gvproxy"
+	if runtime.GOOS == "windows" {
+		executableName = "gvproxy.exe"
+	}
```
cfergeau, line 201: Ideally we would not have 2 distinct ways of deciding whether to use `.exe` as the extension or not (here and in `NewGvproxyCache`)

pkg/crc/ssh/keys.go
```diff
@@ -62,6 +62,12 @@ func GenerateSSHKey(path string) error {
 			return fmt.Errorf("Desired directory for SSH keys does not exist: %s", err)
```
cfergeau: This is a very misleading error, if I undertand correctly, we do not return if the directory does not exist, we only return for other errors.

pkg/crc/ssh/keys.go
```diff
@@ -62,6 +62,12 @@ func GenerateSSHKey(path string) error {
...
 		}
 
+		// Ensure parent directory exists
+		dir := filepath.Dir(path)
+		if err := os.MkdirAll(dir, 0700); err != nil {
+			return fmt.Errorf("Error creating directory for SSH keys: %s", err)
+		}
```
cfergeau, line 69: An explanation why this is needed in the commit log would be nice.

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -201,3 +216,41 @@ func removePodmanFromOcBinDirCheck() Check {
...
+	if version.IsInstaller() {
+		return nil
+	}
+
+	gvproxy := cache.NewGvproxyCache()
+	if err := gvproxy.EnsureIsCached(); err != nil {
+		return errors.Wrap(err, "Unable to download gvproxy executable")
+	}
+	logging.Debug("gvproxy executable cached")
+	// SUID is only required on Linux (setSuid is a no-op on Windows/macOS)
```
cfergeau: Isn’t `"cap_net_bind_service+eip", "cap_net_bind_service=eip"` enough?

pkg/crc/preflight/preflight_darwin.go
```diff
@@ -139,3 +139,14 @@ func getPreflightChecks(_ bool, mode network.Mode, bundlePath string, preset crc
 
 	return filter.Apply(getChecks(mode, bundlePath, preset, enableBundleQuayFallback))
 }
+
+// Capability functions are no-ops on Darwin (macOS doesn't use Linux capabilities)
```
cfergeau, line 142: If you add a `checkGvproxyCapabilities` which is linux only, you won’t need these dummy implementations

pkg/crc/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,106 @@
...
+  content: |
+    %s
+  permissions: '0644'
+- path: /opt/crc/pass_kubeadmin
+  content: '%s'
+  permissions: '0644'
+- path: /opt/crc/pass_developer
+  content: '%s'
+  permissions: '0644'
+- path: /opt/crc/ocp-custom-domain.service.done
```
cfergeau, line 49: Ideally this would not be needed… It’s when we want a custom domain that something would appear in the cloud-init file. But for the nominal case, we should not have to add magic files.

pkg/crc/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,106 @@
...
+type UserDataOptions struct {
+	PublicKey         string
+	PullSecret        string
+	KubeAdminPassword string
+	DeveloperPassword string
+}
+
+const userDataTemplate = `#cloud-config
+runcmd:
+  - systemctl enable --now kubelet
```
cfergeau, line 23: And ideally this would also not be needed…

pkg/crc/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,106 @@
...
+write_files:
+- path: /home/core/.ssh/authorized_keys
+  content: '%s'
+  owner: core
+  permissions: '0600'
+- path: /opt/crc/id_rsa.pub
```
cfergeau, line 29: Can this go in `/home/core/.ssh/` like `authorized_keys`?

pkg/crc/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := json.Compact(&buf, []byte(jsonStr)); err != nil {
+		return "", fmt.Errorf("failed to compact JSON: %w", err)
+	}
+	return buf.String(), nil
+}
+
+// GenerateUserData generates a cloud-init user-data file and returns the path
+func GenerateUserData(machineName string, opts UserDataOptions) (string, error) {
+	// Create the machine directory if it doesn't exist
+	machineDir := filepath.Join(constants.MachineInstanceDir, machineName)
```
cfergeau: I’d get `machineDir` from `GetUserDataPath`

pkg/crc/cloudinit/cloudinit.go
```diff
@@ -0,0 +1,106 @@
...
+	userData := fmt.Sprintf(userDataTemplate,
+		opts.PublicKey,         // /home/core/.ssh/authorized_keys
+		opts.PublicKey,         // /opt/crc/id_rsa.pub
+		compactPullSecret,      // /opt/crc/pull-secret (compacted)
+		opts.KubeAdminPassword, // /opt/crc/pass_kubeadmin
+		opts.DeveloperPassword, // /opt/crc/pass_developer
+	)
+
+	// Write the user-data file
+	userDataPath := filepath.Join(machineDir, "user-data")
```
cfergeau: This should use `GetUserDataPath`

pkg/crc/systemd/results/result.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return ""
+}
+
+func (r Result) IsSuccess() bool {
+	return r == Success
+}
+
+func Compare(input string) Result {
+	normalized := strings.TrimSpace(strings.ToLower(input))
```
cfergeau: Can we assume `input` is already trimmed? This is already done in the caller

pkg/crc/systemd/results/result.go
```diff
@@ -0,0 +1,58 @@
...
+		return results[r]
+	}
+	return ""
+}
+
+func (r Result) IsSuccess() bool {
+	return r == Success
+}
+
+func Compare(input string) Result {
```
cfergeau: Why `Compare`? This is more like a `Parse` or Ǹew` function.

pkg/crc/systemd/systemd.go
```diff
@@ -57,6 +58,18 @@ func (c Commander) Status(name string) (states.State, error) {
 
 }
 
+// Result returns the result of a service execution (success, exit-code, etc.)
+func (c Commander) Result(name string) (results.Result, error) {
+	stdOut, stdErr, err := c.commandRunner.Run("systemctl", "show", name, "--property=Result")
```
cfergeau: You can use `--value` as in the next commit so that you don’t have to parse `Result=`

pkg/crc/systemd/systemd.go
```diff
@@ -70,6 +71,16 @@ func (c Commander) Result(name string) (results.Result, error) {
 	return results.Compare(output), nil
 }
 
+// ExecMainExitTimestamp returns the exit timestamp of a service's main process.
+// Returns empty string if the service has never run, otherwise returns the exit timestamp.
+func (c Commander) ExecMainExitTimestamp(name string) (string, error) {
+	stdOut, stdErr, err := c.commandRunner.Run("systemctl", "show", "-p", "ExecMainExitTimestamp", "--value", name)
```
cfergeau: `-p` -> `--property`

pkg/crc/systemd/systemd.go
```diff
@@ -70,6 +71,16 @@ func (c Commander) Result(name string) (results.Result, error) {
 	return results.Compare(output), nil
 }
 
+// ExecMainExitTimestamp returns the exit timestamp of a service's main process.
+// Returns empty string if the service has never run, otherwise returns the exit timestamp.
+func (c Commander) ExecMainExitTimestamp(name string) (string, error) {
```
cfergeau, line 76: Can this return a `time.Time` instead of a `string`?

pkg/crc/machine/vsock.go
```diff
@@ -14,66 +12,39 @@ import (
...
-// DNSZoneWithDefault represents a DNS zone configuration with a default IP (wildcard)
-type DNSZoneWithDefault struct {
-	Name      string `json:"Name"`
-	DefaultIP string `json:"DefaultIP"`
-}
+func getGVProxyClient(vmName string) (*gvproxyclient.Client, error) {
+	// Construct the gvproxy socket path
+	// Path format: /run/user/<uid>/macadam/qemu/<vmname>-gvproxy-api.sock
+	uid := os.Getuid()
+	socketPath := filepath.Join("/run", "user", strconv.Itoa(uid), "macadam", "qemu", fmt.Sprintf("%s-gvproxy-api.sock", vmName))
```
cfergeau: This path is internal to macadam, ideally this would come from `macadam inspect` or such.

pkg/crc/machine/vsock.go
```diff
@@ -94,68 +65,36 @@ func addDNSZoneToGVProxy(socketPath string, zoneData interface{}) error {
...
-	zone1 := DNSZone{
+	zone1 := types.Zone{
 		Name: "crc.testing.",
-		Records: []DNSRecord{
-			{Name: "host", IP: "192.168.127.254"},
-			{Name: "api", IP: "192.168.127.2"},
-			{Name: "api-int", IP: "192.168.127.2"},
-			{Name: "crc", IP: "192.168.126.11"},
+		Records: []types.Record{
+			{Name: "host", IP: net.ParseIP("192.168.127.254"), Regexp: nil},
```
cfergeau: Is it required to set `Regexp: nil`?

pkg/crc/machine/vsock.go
```diff
@@ -94,68 +65,36 @@ func addDNSZoneToGVProxy(socketPath string, zoneData interface{}) error {
...
+			{Name: "api", IP: net.ParseIP("192.168.127.2"), Regexp: nil},
+			{Name: "api-int", IP: net.ParseIP("192.168.127.2"), Regexp: nil},
+			{Name: "crc", IP: net.ParseIP("192.168.126.11"), Regexp: nil},
 		},
+		DefaultIP: net.ParseIP("192.168.127.2"),
```
cfergeau: `DefaultIP` was not there before.

pkg/crc/machine/vsock.go
```diff
@@ -14,66 +12,39 @@ import (
...
-	}
-	var missingPorts []types.ExposeRequest
 	for _, port := range portsToExpose {
-		if !isOpened(alreadyOpenedPorts, port) {
-			missingPorts = append(missingPorts, port)
-		}
-	}
-	for i := range missingPorts {
-		port := &missingPorts[i]
-		if err := daemonClient.NetworkClient.Expose(port); err != nil {
```
cfergeau, line 34: Shouldn’t this logic be kept?

pkg/crc/machine/vsock.go
```diff
@@ -171,33 +110,6 @@ func isOpened(exposed []types.ExposeRequest, port types.ExposeRequest) bool {
 	return false
 }
 
-func unexposePorts() error {
```
cfergeau, line 50: It’s used in `Stop` and `Delete`

pkg/crc/machine/start.go
```diff
@@ -802,20 +802,13 @@ func ensureRoutesControllerIsRunning(sshRunner *crcssh.Runner, ocConfig oc.Confi
...
-	if err != nil {
-		return errors.Wrap(err, "Not able to get user CA")
+func copyKubeconfigFileFromVMToHost(ctx context.Context, systemdRunner *systemd.Commander, sshRunner *crcssh.Runner, kubeconfigFilePath string) error {
+	logging.Info("Waiting for the updated kubeconfig file to be available on the host...")
+	if err := cluster.WaitForServiceSuccessfullyFinished(ctx, systemdRunner, "ocp-cluster-ca.service", 60*time.Second, 2*time.Second); err != nil {
+		return err
 	}
-	if err := cluster.EnsureGeneratedClientCAPresentInTheCluster(ctx, ocConfig, sshRunner, selfSignedCACert, adminClientCA); err != nil {
-		return errors.Wrap(err, "Failed to update user CA to cluster")
+	if err := sshRunner.CopyFileFromVM("/opt/kubeconfig", kubeconfigFilePath, 0644); err != nil {
```
cfergeau: `0600`?

pkg/crc/preflight/preflight_checks_common.go
praveenkumar: This is removed in later commits

praveenkumar: we only needed the capability instead of SUID

pkg/crc/cache/cache.go
praveenkumar: Yes, we can share the code for getting the version info. May be a followup or new commit.

pkg/crc/machine/vsock.go
praveenkumar: since now it is manage by gvproxy process which get killed and release ports when crc stop/delete happen.

pkg/crc/cloudinit/cloudinit.go
praveenkumar: @cfergeau why? this is required because our bundle have kubelet service disabled by default.

pkg/crc/cloudinit/cloudinit.go
praveenkumar: no, `/opt/crc/id_rsa.pub` is the file which is required by mco ssh service https://github.com/crc-org/snc/blob/release-4.20/systemd/ocp-mco-sshkey.service here.

pkg/crc/cloudinit/cloudinit.go
praveenkumar: Otherwise https://github.com/crc-org/snc/blob/release-4.20/systemd/ocp-custom-domain.service will fails and since it is `Restart=on-failure` so it is going to continue fail :( so better to have it as this magic file.

pkg/crc/systemd/systemd.go
praveenkumar: It return value is something which can be parsed as time but our aim is if there is timestamp which means the service is ready run and exited
```
$ systemctl show --property "ExecMainExitTimestamp" --value systemd-binfmt.service
Wed 2026-01-21 11:02:24 IST
```

pkg/crc/machine/vsock.go
praveenkumar: It was required because before the daemon was long running process so we need to check if the port are already expose but now with gvpoxy it is just associated with crc process so this logic is not required.

praveenkumar (commented):
Tried to update as per some suggestions and commented to some of the suggestions.

pkg/crc/cloudinit/cloudinit.go
cfergeau: My point is that this would be a nice improvement to make to the self sufficient bundle… The less magic is required in the cloud-init file, the better

pkg/crc/cloudinit/cloudinit.go
cfergeau: Same answer as https://github.com/crc-org/crc/pull/5025#discussion_r2758400116 
"ideally" = would be nice to improve the self sufficient bundle and make the cloud-init file simpler.

pkg/crc/cloudinit/cloudinit.go
cfergeau: Is there a document for the cloud-init file we expect, together with an explanation about what uses each cloud-init "parameter"?

pkg/crc/preflight/preflight_darwin.go
praveenkumar: gvproxy cache need for each platform so I put it as part of common preflight check instead of platform specific and that's the reason this dummy implementations are there to compile it.

pkg/crc/constants/constants.go
praveenkumar: Fixed.

pkg/crc/ssh/keys.go
praveenkumar: Done added to commit log

pkg/crc/preflight/preflight_darwin.go
cfergeau: if you split the preflight check in 2, you can remove the dummy implementations

cmd/crc/cmd/daemon.go
```diff
@@ -276,7 +138,7 @@ func run(configuration *types.Configuration) error {
 		}()
 
 		// 9p over TCP (as a backup)
-		listener9pTCP, err := vn.Listen("tcp", net.JoinHostPort(configuration.GatewayIP, fmt.Sprintf("%d", constants.Plan9TcpPort)))
+		listener9pTCP, err := net.Listen("tcp", fmt.Sprintf("%s:%d", constants.VSockGateway, constants.Plan9TcpPort))
```
anjannath, line 141: This is trying to open  a lower port (`constants.Plan9TcpPort = 564`) which is failing with permission error

praveenkumar: @anjannath so before this was fine because that was done on virtual network side? we need to figure out if this is just a port forwarding or something on host is running which need access on the VM.

pkg/crc/macadam/macadam.go
```diff
@@ -0,0 +1,213 @@
...
+	if isPrivate {
+		return m.Runner.RunPrivate(m.MacadamExecutablePath, args...)
+	}
+
+	return m.Runner.Run(m.MacadamExecutablePath, args...)
+}
+
+func (m Config) RunMacadamCommand(args ...string) (string, string, error) {
+	if runtime.GOOS == "windows" {
+		args = append(args, []string{"--provider", "hyperv"}...)
```
anjannath, line 97: for hyper-v the default network mode is the system-mode-networking, since in crc we only want to use user-mode networking need to also add the flag `--user-mode-networking` here

pkg/crc/macadam/macadam.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return m.runCommand(false, args...)
+}
+
+func (m Config) RunMacadamCommandPrivate(args ...string) (string, string, error) {
```
anjannath, line 102: is this helper used somewhere?

pkg/crc/macadam/macadam.go
anjannath: sorry not here but only for the `Init` helper the network-mode flag is needed

anjannath:
@praveenkumar tested this branch on windows today, it works ootb with the previous requirement of running it on a admin powershell and set `crc config set skip-check-administrator-user true`

PR #5067, opened 2025-12-17
9p: add preflight check for 9p hvsock

### LINKED ISSUE #5066: Add preflight test for 9P hvsock on Windows
Together with checking if the `gvisor-tap-vsock` entry exists in the registry, we should also check the hvsock used by 9P.

redbeam:
Fixes #5066

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows preflight now validates gVisor and 9P registry entries separately, requiring both to pass and providing clearer, component-specific error messages when validation fails.

* **Chores**
  * No changes to public/exported interfaces.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/pkg/crc/preflight/preflight_windows.go
+++ b/pkg/crc/preflight/preflight_windows.go
@@ -177,20 +177,31 @@ func username() string {
 
 const (
 	// This key is required to activate the vsock communication
-	registryDirectory = `HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Virtualization\GuestCommunicationServices`
+	vsockRegistryDirectory = `HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Virtualization\GuestCommunicationServices`
 	// First part of the key is the vsock port. The rest is not used and just a placeholder.
-	registryKey   = "00000400-FACB-11E6-BD58-64006A7986D3"
-	registryValue = "gvisor-tap-vsock"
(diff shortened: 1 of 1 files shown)


pkg/crc/preflight/preflight_windows.go
```diff
@@ -177,20 +177,31 @@ func username() string {
...
+		return errors.New("VSock registry key for gvisor not correctly configured")
 	}
+
+	stdout, _, err = powershell.Execute(fmt.Sprintf(`Get-Item -Path "%s\%s"`, vsockRegistryDirectory, fs9pVsockRegistryKey))
+	if err != nil {
+		return err
+	}
+	if !strings.Contains(stdout, fs9pVsockRegistryValue) {
+		return errors.New("VSock registry key for 9P not correctly configured")
+	}
```
anjannath, line 203: would be good to create a helper to check if the registry key exists to remove this code duplication, otherwise LGTM

redbeam: @anjannath I was thinking about that, but I figured since it is only one instance of this duplication, we can let it be. Otherwise I'd have created the helper.

anjannath (approved):

anjannath:
@redbeam we usually create the PR branch in our fork of the repo i think to avoid too many branches in the upstream :) 
the CI jobs should run when the PR branch is in your fork as well

redbeam:
@anjannath I find it more convenient to work with the upstream repo than to maintain my own fork 😅 and I always take care of cleaning the feature branches when they're merged.

PR #5147, opened 2026-02-16
Update golang version to 1.25

praveenkumar:
## Summary by CodeRabbit

* **Chores**
  * Upgraded Go toolchain from version 1.24 to 1.25 across all build configurations, development containers, CI/CD workflows, and container images for improved compatibility and consistent runtime environment.


cfergeau:
Can you add to the commit log the rationale for switching the minimum go version to 1.25?

praveenkumar:
> Can you add to the commit log the rationale for switching the minimum go version to 1.25?

Done, added that 1.24 is now EOL.

cfergeau (approved):
To be honest, I don’t think "1.24 is EOL" is a very compelling reason for forcing everyone to use 1.25 to build crc when neither crc nor its deps require go 1.25.

For example, this means crc can no longer be built on f42 which is still supported.

Not going to block merging though, just giving my opinion.

praveenkumar:
> To be honest, I don’t think "1.24 is EOL" is a very compelling reason for forcing everyone to use 1.25 to build crc when neither crc nor its deps require go 1.25.

I was bit surprise that F-42 doesn't have go-1.25 :(,  also another reason to have move from 1.24,  there will be no security updates around the core modules and some or our dependencies modules already migrated to 1.25

PR #5037, opened 2025-12-02
PoC: Add bundle download cmd

sebrandon1:
Expanding on the already existing `crc bundle` command with some exciting new functionality.  I wanted to be able to download/manipulate the bundles available either locally or on the mirror website from the CRC client.  I also wanted to be able to download bundles for different architectures if needed (will be used in [quick-ocp](https://github.com/palmsoftware/quick-ocp) more than likely).

Example commands:

- `./crc bundle download` this is just the default command, and it'll download the bundle belonging to your crc release.
- `./crc bundle download 4.19 -v` this will download the latest available 4.19 bundle (similar to the way quick-ocp [does it](https://github.com/palmsoftware/quick-ocp/blob/main/scripts/fetch-ocp-crc-version.sh)) where it will determine what the latest patch version is, and download it.  You can also just supply an exact patch version that works as well.  By default, it will check the downloaded bundle against the hash signature to make sure you have downloaded it correctly.  Even thought `crc` won't be able to directly use the old bundles, its already been helpful for me to be able to download these bundles directly.
- `./crc bundle download 4.19 amd64 -v` will download a specific architecture version.
- `./crc bundle download list` will show you a list of all of the available versions of CRC/OCP available from the mirror website and whether or not you already have the bundle downloaded on your system already.
- `./crc bundle download clear` will clear all of your bundles out of your cache.
- `./crc bundle download prune` will clear all but the latest two patch versions belonging to a major version out of your cache.  Like a watered down version of `clear`.

Real output:
```
bpalm at bpalm-mac in ~/Repositories/go/src/github.com/crc-org/crc on add_bundle_download_cmd [!?]
$ ./crc bundle download 4.19 -v
INFO Resolving latest version for 4.19...         
INFO Resolved version 4.19 to 4.19.13             
INFO Downloading bundle: crc_vfkit_4.19.13_arm64.crcbundle 
INFO Source: https://mirror.openshift.com/pub/openshift-v4/clients/crc/bundles/openshift/4.19.13/crc_vfkit_4.19.13_arm64.crcbundle 
INFO Destination: /Users/bpalm/.crc/cache         
INFO Verifying signature for 4.19.13...           
6.01 GiB / 6.01 GiB [-----------------------------------------------] 100.00% 7.73 MiB/s
```

## Summary by CodeRabbit

* **New Features**
  * Bundle command adds download, list, clear, and prune subcommands.
  * Download supports preset/version/architecture selection, mirror-based resolution, signature-verified retrieval, and force overwrite.
[...]

PR #5153, opened 2026-02-18
Issue #2068 Add disk space preflight check

### LINKED ISSUE #2068: Better UX of an error message
* OS: Linux  Red Hat 8
  * Did you run `crc setup` before starting it (Yes/No)? Yes

I gave crt setup and during the process, the program has stopped writing I had not enough free space, but I did not realize that my /home partition was too small, while the total size was big enough.
I think an improvement could be a message saying that the space on the device is not sufficient even before copying the files.
Or maybe it could say something like "the free space is xyz, while we need abc".

sebrandon1:
## Summary

- Add a preflight check that validates sufficient free disk space (43 GiB) at the user's home directory before downloading/extracting the CRC bundle, preventing cryptic "no space left on device" errors after lengthy downloads
- Follow the existing `memoryCheck()` pattern: `NoFix` flag, skippable via `crc config set skip-check-disk-space true`, fail-open on errors querying disk space
- Add `palmsoftware/quick-cleanup` to the macOS installer CI workflow to free disk space before `crc setup`, ensuring the new preflight check passes on GitHub Actions runners

This complements #5037 (`crc bundle download`), which makes it easier to download and manage multiple bundles locally. With users potentially caching several bundle versions, running out of disk space becomes more likely, making an early space check even more valuable.

See also #5160 which adds `quick-cleanup` to additional disk-intensive workflows.

Fixes #2068

## Test plan

- [x] `TestValidateEnoughFreeSpace` covers pass and fail (`math.MaxUint64`) cases
- [x] Preflight count tests updated for all platforms (Darwin, Linux, Windows)
- [x] `make test` passes
- [x] `make lint` passes (0 issues on all platforms)
- [x] `make check` passes (cross-platform builds + tests + lint)
- [ ] Manual: run `crc setup` on a machine with <43 GiB free and verify the check reports the shortfall
- [ ] Manual: run `crc config set skip-check-disk-space true` and verify the check is skipped

## Summary by CodeRabbit

* **New Features**
  * Added disk space validation during preflight checks on all platforms (macOS, Linux, Windows).
  * System now enforces a minimum free disk space requirement before setup, calculated from bundle and installation sizes.
  * Added automatic disk cleanup to the macOS installer workflow to free space during installation.

* **Tests**
  * Updated test assertions to reflect the new disk space validation check.


praveenkumar:
Can we squash these commit to single one?

pkg/crc/preflight/preflight_checks_common.go
```diff
@@ -48,6 +48,21 @@ func memoryCheck(preset crcpreset.Preset) Check {
 	}
 }
 
+func diskSpaceCheck() Check {
+	return Check{
+		configKeySuffix:  "check-disk-space",
+		checkDescription: fmt.Sprintf("Checking if enough disk space is available for the bundle in %s", constants.MachineCacheDir),
+		check: func() error {
+			return validation.ValidateEnoughFreeSpace(constants.MachineCacheDir, constants.MinimumDiskSpaceForSetup)
+		},
```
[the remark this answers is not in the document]
praveenkumar: @sebrandon1 this looks right, may be use the base dir instead of cacheDir which should be exist even as part of fresh start.

pkg/crc/validation/validation.go
```diff
@@ -64,6 +65,40 @@ func ValidateEnoughMemory(value strongunits.MiB) error {
...
+		return fmt.Errorf("only %s of disk space free at %s (%s required)",
+			units.HumanSize(float64(freeSpace)),
+			path,
+			units.HumanSize(float64(requiredBytes)))
+	}
+	return nil
+}
+
+// nearestExistingPath walks up the directory tree to find the nearest existing ancestor.
+func nearestExistingPath(path string) string {
```
praveenkumar: instead of this logic I think just use `constants.GetHomeDir()` to validate the disk space is enough since this is the base path where other crc paths are built on.

praveenkumar (changes requested):

praveenkumar:
Also fix `make check` for windows.

pkg/crc/validation/validation.go
sebrandon1: Done!

sebrandon1:
@praveenkumar I'm looking at the MacOS failure (now that its flagging disk space).  I'll let you know when I figure it out.

praveenkumar:
We also make the change to our documentation where it is still mentioned https://crc.dev/docs/installing/#_for_openshift_container_platform coming from https://github.com/crc-org/docs/blob/main/modules/ROOT/pages/installing.adoc here.

praveenkumar:
> We also make the change to our documentation where it is still mentioned https://crc.dev/docs/installing/#_for_openshift_container_platform coming from https://github.com/crc-org/docs/blob/main/modules/ROOT/pages/installing.adoc here.

@coderabbitai create an issue on docs branch

praveenkumar:
@sebrandon1 So I tried to check if that much space is really required because we are bumping up the Disk requirement from 35 to 43 GB, below is my observation on linux

Before `crc setup`
```
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda2        50G  3.9G   46G   8% /
...
```

After `crc setup` , almost used `31GB`. 
```
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda2        50G   35G   16G  69% /
...
```

After `crc start` take around 40GB . There is no app deployed.
```
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda2        50G   42G  8.0G  84% /
...
```

Now we use sparsify images which explain this behavior. What I think is we should use as disk check is (bundle size + disk size) this way we can even detect if user provide different disk size?

sebrandon1:
Okay I think we're headed in the right direction.

- Threshold is now dynamic: Instead of a hardcoded 43 GiB, it's computed as DefaultDiskSize + EstimatedCompressedBundleSize (31 + 12 GiB). This way if DefaultDiskSize changes in 
  the future, the check adapts automatically.               
- Windows make check fixed: Replaced deprecated windows.StringToUTF16Ptr with windows.UTF16PtrFromString.
- Binary units: Switched from HumanSize (SI/decimal) to BytesSize (binary/GiB) so the error messages match the constant definitions.

sebrandon1:
@coderabbitai review

sebrandon1:
@praveenkumar I've addressed your feedback and force-pushed an updated commit. When you get a chance, could you take another look? Thanks!

sebrandon1:
Closing — not enough upstream interest to continue pursuing at this time. Can be revisited later if needed.

PR #5159, opened 2026-02-26
use safe functions to join file paths

redbeam:
Use safe functions to join file paths and open files in places reported by Snyk. `BuildPathChecked` checks for the ZipSlip and path traversal vulnerabilities, while files are now opened using `os.OpenInRoot`.

## Summary by CodeRabbit

* **Chores**
  * Added an ignore rule to the security scanning configuration for a specific detection (CWE-798) with a wildcard scope and reason "Test", adjusting how that finding is reported.
  * No other exclude rules were modified and there are no changes to public/exported interfaces.

### DIFF
--- a/pkg/crc/machine/bundle/repository.go
+++ b/pkg/crc/machine/bundle/repository.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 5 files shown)


cfergeau:
Some path traversal reports were ignored in https://github.com/crc-org/crc/pull/5151#issuecomment-3951782813 , maybe they could also make use of the new helper you introduce?

anjannath:
Maybe here we can use the new `os.OpenRoot` api to ensure the paths are nested under the expected directory always, https://go.dev/blog/osroot#introducing-osroot

redbeam:
@cfergeau I forgot about those, thanks for the tip!

@anjannath That's definitely something I'll take a look at! I didn't know about it, but it could help with this and we wouldn't even have to ignore the FPs. Thanks!

redbeam:
@cfergeau I checked the previously flagged issues and they are already migrated to use the new safe function.

praveenkumar:
> @cfergeau I checked the previously flagged issues and they are already migrated to use the new safe function.

@redbeam https://github.com/crc-org/crc/pull/5151/changes/c1390eb0aa4c4f6987b7cf6d42ba4f677b0d4ae6 I think copy and viper config changes might be also updated?

cfergeau:
By the way, podman is using https://github.com/cyphar/filepath-securejoin , I don’t know if it handles path traversal attacks, or only something different.

redbeam:
@anjannath @cfergeau I ended up using the new `OpenRoot/OpenInRoot` API to resolve the Snyk issues. There could be more places that they could be used, but that should be dealt with in a separate issue/PR I think.

redbeam:
@cfergeau Oops, sorry about the confusion, I thought you were talking about the false positives that were resolved by the Snyk team. I'll take a look into the changes from that PR to see if anything could be done about them.

anjannath (approved):

PR #5236, opened 2026-05-07
Modernize tool deps using go tool directive

### LINKED ISSUE #5235: Modernize how to get the tools using tool dependencies
https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module  and https://go.dev/doc/modules/managing-dependencies#tools have details about how we can modernize our tool deps.

praveenkumar:
Replace the separate tools/ module (dummy.go, go.mod, vendor/) with Go 1.24+ tool directives in the main go.mod. Tools are now invoked via `go tool <name>` instead of pre-installing binaries to tools/bin/.

- https://go.dev/doc/modules/managing-dependencies#tools

For cross-lint, `go tool -n` resolves the host-compiled binary path so it can be run with a different GOOS/GOARCH for cross-platform analysis.

Closes #5235


cfergeau:
If this removes the `tools/vendor` directory, I’m not sure this is the way to go, we need that directory for network-less builds.

praveenkumar:
> If this removes the `tools/vendor` directory, I’m not sure this is the way to go, we need that directory for network-less builds.

@cfergeau this is now added to root vendor directory so network-less builds should work and this is also required for hermetic builds.

Makefile
```diff
@@ -250,19 +247,20 @@ e2e-story-microshift: install
...
+	go tool golangci-lint run
 
-cross-lint: $(TOOLS_BINDIR)/golangci-lint gen_release_info
-	GOOS=darwin GOARCH=amd64 "$(TOOLS_BINDIR)"/golangci-lint run
-	GOOS=darwin GOARCH=arm64 "$(TOOLS_BINDIR)"/golangci-lint run
-	GOOS=linux GOARCH=amd64 "$(TOOLS_BINDIR)"/golangci-lint run
-	GOOS=windows GOARCH=amd64 "$(TOOLS_BINDIR)"/golangci-lint run
+cross-lint: gen_release_info
+	$(eval GOLANGCI_LINT := $(shell go tool -n golangci-lint))
+	GOOS=darwin GOARCH=amd64 "$(GOLANGCI_LINT)" run
```
cfergeau, line 266: Why not `GOOS=darwin GOARCH=amd64 go tool run golangci-lint`?

edit: what I suggest would build a native darwin binary we cannot run. The way it’s done builds a linux binary, but runs it with GOOS/GOARCH set.

cfergeau (commented):
We might want to keep the `go.mod` for the tools separate from the one for the main binaries as described here https://blog.howardjohn.info/posts/go-tools-command/ but this is just a comment I’m making, not a requirement.

cfergeau:
By the way, I became aware of this last week https://golangci-lint.run/docs/welcome/install/local/#install-from-sources

> We [golangci-lint] don’t recommend using go tool.

:(

But they still give some recommendations for people who want to use it this way.

praveenkumar:
> By the way, I became aware of this last week https://golangci-lint.run/docs/welcome/install/local/#install-from-sources
> 
> > We [golangci-lint] don’t recommend using go tool.

@cfergeau but even in current form we are not directly using the golangci-lint as binary but build it ourself from the tool side.

> 
> :(
> 
> But they still give some recommendations for people who want to use it this way.

cfergeau:
> > By the way, I became aware of this last week https://golangci-lint.run/docs/welcome/install/local/#install-from-sources
> > > We [golangci-lint] don’t recommend using go tool.
> 
> @cfergeau but even in current form we are not directly using the golangci-lint as binary but build it ourself from the tool side.

For CI, they recommend using https://github.com/golangci/golangci-lint-action which we decided not to since we don’t have easy control over the version used, either it’s automatically updated without us getting a chance to detect potential breakage in the new version, or we have to remember to update the version ourselves. I don’t think dependabot knows how to update the version?

praveenkumar:
> I don’t think dependabot knows how to update the version?

@cfergeau if you mean for the action then it is possible, we might need to try.
https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

cfergeau:
> > I don’t think dependabot knows how to update the version?
> 
> @cfergeau if you mean for the action then it is possible, we might need to try. https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

```
uses: golangci/golangci-lint-action@v9
with:
  version: v2.13
```

It is the `version` field that we would want to get dependabot PRs for.

PR #5272, opened 2026-06-18
pkg: move chmod command from packaging to postinstall script

anjannath:
## Description

we are going to build the installer in konflux in a hermetic env, where we use sandbox-exec to restrict network access it also blocks any setuid/setgid operations and the build fails when 'chmod +sx' is ran

we need the setuid bit set on the admin-helper binary on the user's machine which can be done in the post-install  script

Relates to: https://github.com/crc-org/crc-internal/issues/210

## Summary by CodeRabbit

* **Chores**
  * Updated macOS installation permission handling for the bundled admin helper: executable permission changes now occur during post-install rather than during package building.
  * The post-install step sets the required setuid bit and ensures the helper has appropriate group execute permissions after installation.


cfergeau (approved):
This looks fine. Hopefully there won’t be "security" side effects with some corporate environments blocking this postinstall chmod.

PR #5281, opened 2026-06-22
constants: Move host IPC sockets to ~/.crc/sockets with restricted permissions

vyasgun:
## Description

This PR moves CRC host IPC sockets under `~/.crc/sockets/` (directory mode 0700) and sets socket file mode to `0600` after bind, instead of relying on the default `0666`. Having 0750 on directory and 0666 on sockets can result in another user to misuse the sockets:
```zsh
gvyas@Gunjans-MacBook-Pro ~ % crc status
CRC VM:          Running
OpenShift:       Running (v4.20.1)
RAM Usage:       6.305GB of 16.73GB
Disk Usage:      25.11GB of 68.11GB (Inside the CRC VM)
Cache Usage:     86.79GB
Cache Directory: /Users/gvyas/.crc/cache
gvyas@Gunjans-MacBook-Pro ~ % sudo -u crcattacker curl --unix-socket "/Users/gvyas/.crc/crc-http.sock" "http://unix/api/stop"
gvyas@Gunjans-MacBook-Pro ~ % crc status
CRC VM:          Stopped
OpenShift:       Stopped (v4.20.1)
Cache Usage:     86.79GB
Cache Directory: /Users/gvyas/.crc/cache
```

## Type of change

## Testing

1. `crc cleanup && crc setup && crc start` (required after upgrade — old daemon uses old paths)
2. `stat -f '%Lp' ~/.crc/sockets` -> `0700`
3. `stat -f '%Lp' ~/.crc/sockets/*.sock` -> `0600 0600 0600`
4. `ssh` should work

**Note:** This only works after running a cleanup of the old crc instance (including daemon) as the new crc will start processes on different socket paths after this change (`~/.crc/sockets`)

## Contribution Checklist
    - [ ] Linux
    - [ ] Windows
    - [x] MacOS

## Release Notes

* **Bug Fixes**
  * Improved Unix socket setup so socket filesystem permissions are applied before the daemon reports “listening”.
  * If socket creation fails, the daemon now returns early without emitting misleading “listening” messages.
  * Ensures socket listeners are properly closed when permission setup fails.

* **Refactor**
  * Moved daemon socket files into a dedicated `~/.crc/sockets` directory.
  * Enforced secure permissions for the socket directory (`0700`) and socket files (`0600`) across platforms.

* **Documentation**
  * Updated the Linux systemd user unit template to reference the new HTTP socket path and socket/dir modes.

### DIFF
--- a/cmd/crc/cmd/daemon_darwin.go
+++ b/cmd/crc/cmd/daemon_darwin.go
@@ -16,20 +16,28 @@ import (
 func vsockListener() (net.Listener, error) {
 	_ = os.Remove(constants.TapSocketPath)
 	ln, err := net.Listen("unix", constants.TapSocketPath)
-	logging.Infof("listening %s", constants.TapSocketPath)
 	if err != nil {
 		return nil, err
 	}
+	if err = constants.EnsureSocketFilesPermissions(constants.TapSocketPath); err != nil {
+		_ = ln.Close()
+		return nil, err
+	}
+	logging.Infof("listening %s", constants.TapSocketPath)
 	return ln, nil
 }
 
 func httpListener() (net.Listener, error) {
 	_ = os.Remove(constants.DaemonHTTPSocketPath)
 	ln, err := net.Listen("unix", constants.DaemonHTTPSocketPath)
-	logging.Infof("listening %s", constants.DaemonHTTPSocketPath)
 	if err != nil {
 		return nil, err
 	}
+	if err = constants.EnsureSocketFilesPermissions(constants.DaemonHTTPSocketPath); err != nil {
+		_ = ln.Close()
+		return nil, err
+	}
+	logging.Infof("listening %s", constants.DaemonHTTPSocketPath)
 	return ln, nil
 }
 
@@ -38,6 +46,10 @@ func unixgramListener(ctx context.Context, vn *virtualnetwork.VirtualNetwork) (*
 	conn, err := transport.ListenUnixgram(fmt.Sprintf("unixgram://%v", constants.UnixgramSocketPath))
 	if err != nil {
 		return conn, errors.Wrap(err, "failed to listen unixgram")
+	}
+	if err = constants.EnsureSocketFilesPermissions(constants.UnixgramSocketPath); err != nil {
+		_ = conn.Close()
+		return nil, errors.Wrap(err, "failed to set permissions for unixgram socket")
 	}
 	logging.Infof("listening on %s", constants.UnixgramSocketPath)
 	vfkitConn, err := transport.AcceptVfkit(conn)
--- a/cmd/crc/cmd/daemon_linux.go
+++ b/cmd/crc/cmd/daemon_linux.go
@@ -112,6 +112,10 @@ func httpListener() (net.Listener, error) {
 		return nil, err
 	}
 	if ln != nil {
+		if err = constants.EnsureSocketFilesPermissions(constants.DaemonHTTPSocketPath); err != nil {
+			_ = ln.Close()
+			return nil, err
+		}
 		logging.Infof("using socket provided by %s", httpUnitName)
 		return ln, nil
 	}
@@ -119,10 +123,14 @@ func httpListener() (net.Listener, error) {
 	// no socket activation, we need to create the listener
 	_ = os.Remove(constants.DaemonHTTPSocketPath)
 	ln, err = net.Listen("unix", constants.DaemonHTTPSocketPath)
-	logging.Infof("listening %s", constants.DaemonHTTPSocketPath)
 	if err != nil {
 		return nil, err
 	}
+	if err = constants.EnsureSocketFilesPermissions(constants.DaemonHTTPSocketPath); err != nil {
+		_ = ln.Close()
+		return nil, err
+	}
+	logging.Infof("listening %s", constants.DaemonHTTPSocketPath)
 	return ln, nil
 }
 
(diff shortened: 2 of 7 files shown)


cmd/crc/cmd/daemon_linux.go
```diff
@@ -119,10 +119,13 @@ func httpListener() (net.Listener, error) {
 	// no socket activation, we need to create the listener
 	_ = os.Remove(constants.DaemonHTTPSocketPath)
 	ln, err = net.Listen("unix", constants.DaemonHTTPSocketPath)
-	logging.Infof("listening %s", constants.DaemonHTTPSocketPath)
 	if err != nil {
 		return nil, err
 	}
+	if err = constants.EnsureSocketFilesPermissions(constants.DaemonHTTPSocketPath); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
vyasgun: that would be true for every other place where I added this check. This needs to be added everywhere else as well?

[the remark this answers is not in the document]
vyasgun: what should we do if the call to close fails? wrap the closing error together with the socket permissions error and return or ignore all connection close errors?

praveenkumar:
@vyasgun for linux following changes also required.

```patch
$ git diff
diff --git a/cmd/crc/cmd/daemon_linux.go b/cmd/crc/cmd/daemon_linux.go
index 459466..24b0aa 100644
--- a/cmd/crc/cmd/daemon_linux.go
+++ b/cmd/crc/cmd/daemon_linux.go
@@ -112,6 +112,10 @@ func httpListener() (net.Listener, error) {
                return nil, err
        }
        if ln != nil {
+               if err = constants.EnsureSocketFilesPermissions(constants.DaemonHTTPSocketPath); err != nil {
+                       _ = ln.Close()
+                       return nil, err
+               }
                logging.Infof("using socket provided by %s", httpUnitName)
                return ln, nil
        }
diff --git a/pkg/crc/preflight/preflight_checks_linux.go b/pkg/crc/preflight/preflight_checks_linux.go
index 7a53f8..2aed7c 100644
--- a/pkg/crc/preflight/preflight_checks_linux.go
+++ b/pkg/crc/preflight/preflight_checks_linux.go
@@ -308,6 +308,8 @@ Description=CRC HTTP socket
 
 [Socket]
 ListenStream=%h/.crc/sockets/crc-http.sock
+SocketMode=0600
+DirectoryMode=0700
 Service=crc-daemon.service
 
 [Install]
```

vyasgun:
Thanks @praveenkumar - will update this.

pkg/crc/constants/constants.go
```diff
@@ -196,7 +173,7 @@ func GetHomeDir() string {
 	return homeDir
 }
 
-// EnsureBaseDirectoriesExist creates ~/.crc, ~/.crc/bin and ~/.crc/cache directories if it is not present
+// EnsureBaseDirectoriesExist creates ~/.crc, ~/.crc/bin, ~/.crc/cache, and ~/.crc/sockets directories if they are not present
```
cfergeau, line 176: `~/.crc/run` might be more generic than `~/.crc/sockets` (similar to https://linuxvox.com/blog/run-folder-linux/ )

pkg/crc/constants/constants.go
```diff
@@ -205,7 +182,19 @@ func EnsureBaseDirectoriesExist() error {
...
+func ensureSocketDirectoryExists() error {
+	err := os.MkdirAll(SocketBaseDir, 0700)
+	if err != nil {
+		return err
+	}
+	return os.Chmod(SocketBaseDir, 0700)
+}
+
+func EnsureSocketFilesPermissions(path string) error {
+	return os.Chmod(path, 0600)
```
cfergeau, line 197: Not fully sure we need a wrapper for a one-liner which is doing something basic?

cmd/crc/cmd/daemon_linux.go
```diff
@@ -112,17 +112,25 @@ func httpListener() (net.Listener, error) {
 		return nil, err
 	}
 	if ln != nil {
+		if err = constants.EnsureSocketFilesPermissions(constants.DaemonHTTPSocketPath); err != nil {
```
cfergeau, line 115: This already enforced in the unit file, so it’s redundant.

cmd/crc/cmd/daemon_darwin.go
```diff
@@ -39,6 +47,10 @@ func unixgramListener(ctx context.Context, vn *virtualnetwork.VirtualNetwork) (*
 	if err != nil {
 		return conn, errors.Wrap(err, "failed to listen unixgram")
 	}
+	if err = constants.EnsureSocketFilesPermissions(constants.UnixgramSocketPath); err != nil {
+		_ = conn.Close()
```
cfergeau, line 51: Should we change `transport.ListenUnixgram` to do it?

cmd/crc/cmd/daemon_darwin.go
```diff
@@ -16,20 +16,28 @@ import (
...
 }
 
 func httpListener() (net.Listener, error) {
 	_ = os.Remove(constants.DaemonHTTPSocketPath)
 	ln, err := net.Listen("unix", constants.DaemonHTTPSocketPath)
-	logging.Infof("listening %s", constants.DaemonHTTPSocketPath)
 	if err != nil {
 		return nil, err
 	}
+	if err = constants.EnsureSocketFilesPermissions(constants.DaemonHTTPSocketPath); err != nil {
```
cfergeau, line 36: Might be best to add our own `net.Listen` wrapper which will call `net.Listen` + `os.Chmod` instead of duplicating this everywhere there is a `net.Listen` call.

PR #5303, opened 2026-07-09
go.mod: Update openshift/k8s dep to v4.22.0 release

praveenkumar:
Following commands are executed to have those changes
```
go get -x github.com/openshift/api@release-4.22
go get -x github.com/openshift/client-go@release-4.22
go get -x github.com/openshift/library-go@release-4.22
```


go.mod
```diff
@@ -64,9 +64,9 @@ require (
 	golang.org/x/text v0.40.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.34.1
-	k8s.io/apimachinery v0.34.1
-	k8s.io/client-go v0.34.1
+	k8s.io/api v0.35.1
```
cfergeau, line 67: Latest is 0.35.5, `github.com/openshift/oc` uses 0.35.2, any specific reason for picking up an older version?

cfergeau (approved):
As I was wondering what this does, `go get -x` is: 
> The -x flag prints commands as they are executed. This is useful for debugging version control commands when a module is downloaded directly from a repository.

go.mod
praveenkumar: release-4.22 branch uses https://github.com/openshift/client-go/blob/release-4.22/go.mod#L9 (v0.35.1)

cfergeau: Depends which component, https://github.com/openshift/oc/blob/release-4.22/go.mod#L60-L66

PR #5335, opened 2026-08-10
Update vfkit version to 0.6.4

praveenkumar:
## Summary by CodeRabbit

* **Updates**
  * Updated the bundled vfkit version to 0.6.4.
  * Download and entitlement references now use the newer vfkit release.

### DIFF
--- a/pkg/crc/machine/vfkit/constants.go
+++ b/pkg/crc/machine/vfkit/constants.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	VfkitVersion = "0.6.1"
+	VfkitVersion = "0.6.4"
 	vfkitCommand = "vfkit"
 )
 


pkg/crc/machine/vfkit/constants.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	VfkitVersion = "0.6.1"
+	VfkitVersion = "0.6.4"
```
cfergeau, line 12: In the future, we might want to do something like https://github.com/podman-container-tools/podman/blob/4a67addbd4def9669b9f87b64a10f3bb88941c46/Makefile#L224-L226 to propagate automatically the vfkit version from `go.mod`. Or maybe it’s available at runtime from `debug.ReadBuildInfo` or such.

cfergeau (approved):
Just to be sure, when building the installer, we use `crc-embedder` to fetch the expected `vfkit` version, so there is no hardcoded vfkit version in the "installer build" scripts ?

PR #5349, opened 2026-08-18
launchd: Restrict agent plist to Background session type 

praveenkumar:
## Description

Without an explicit LimitLoadToSessionType, macOS defaults the
LaunchAgent to the Aqua (GUI) session. This causes the CRC daemon
to fail with error 134 (BOOTSTRAP_NOT_PRIVILEGED) when loaded over
SSH or any non-GUI context, since no Aqua session exists.

Set LimitLoadToSessionType to Background, which is a superset that
includes GUI, SSH, and all other login contexts. The daemon is a
headless process with no GUI dependency, so this has no effect on
existing GUI users.

## Type of change

## Testing
Try to connect your mac using ssh from another system and logout from the mac and then after `crc cleanup` => `crc setup` . With this PR you are able to run the crc daemon as service.

Without this PR it fails to run the crc daemon.

## Contribution Checklist
    - [x] MacOS

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS background service loading for per-user agents.
  * Improved service removal reliability for the current user session.
  * Activation failures are now logged while setup continues.

* **Enhancements**
  * Restricted launch agents to macOS Background sessions for more predictable behavior.
  * Improved service lifecycle handling across startup and removal operations.

### DIFF
--- a/pkg/os/darwin/launchd/launchd_darwin.go
+++ b/pkg/os/darwin/launchd/launchd_darwin.go
@@ -12,6 +12,7 @@ import (
 	"text/template"
 
 	"github.com/crc-org/crc/v2/pkg/crc/constants"
+	"github.com/crc-org/crc/v2/pkg/crc/logging"
 	"github.com/crc-org/crc/v2/pkg/os"
 )
 
@@ -115,12 +116,18 @@ func CheckPlist(config AgentConfig) error {
 
 // LoadPlist loads a launchd agents' plist file
 func LoadPlist(label string) error {
-	return runLaunchCtl("load", "-w", getPlistPath(label))
+	uid := goos.Getuid()
+	// This is required for migration from older versions
+	if err := runLaunchCtl("enable", fmt.Sprintf("user/%d/%s", uid, label)); err != nil {
+		logging.Debugf("Failed to enable launchd service %s: %v", label, err)
+	}
+	return runLaunchCtl("bootstrap", fmt.Sprintf("user/%d", uid), getPlistPath(label))
 }
 
 // UnloadPlist Unloads a launchd agent's service
 func UnloadPlist(label string) error {
-	return runLaunchCtl("unload", "-w", getPlistPath(label))
+	target := fmt.Sprintf("user/%d/%s", goos.Getuid(), label)
+	return runLaunchCtl("bootout", target)
 }
 
 // RemovePlist removes a launchd agent plist config file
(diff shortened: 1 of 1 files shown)


pkg/os/darwin/launchd/launchd_darwin.go
```diff
@@ -113,12 +116,17 @@ func CheckPlist(config AgentConfig) error {
 
 // LoadPlist loads a launchd agents' plist file
 func LoadPlist(label string) error {
-	return runLaunchCtl("load", "-w", getPlistPath(label))
+	uid := goos.Getuid()
+	if err := runLaunchCtl("enable", fmt.Sprintf("user/%d/%s", uid, label)); err != nil {
```
cfergeau, line 121: Is `enable` required?

cfergeau (commented):
Looks good to me.

pkg/os/darwin/launchd/launchd_darwin.go
praveenkumar: @cfergeau if user is going from older version to newer without cleanup then yes this is required because by default older way of unload put the disable flag set for service.

pkg/os/darwin/launchd/launchd_darwin.go
```diff
@@ -113,12 +116,17 @@ func CheckPlist(config AgentConfig) error {
...
+		logging.Debugf("Failed to enable launchd service %s: %v", label, err)
+	}
+	return runLaunchCtl("bootstrap", fmt.Sprintf("user/%d", uid), getPlistPath(label))
 }
 
 // UnloadPlist Unloads a launchd agent's service
 func UnloadPlist(label string) error {
-	return runLaunchCtl("unload", "-w", getPlistPath(label))
+	target := fmt.Sprintf("user/%d/%s", goos.Getuid(), label)
+	return runLaunchCtl("bootout", target)
```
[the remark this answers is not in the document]
praveenkumar: @coderabbitai older installation is using `load -w` which was not registring the service in `gui/<uid>` as default.

pkg/os/darwin/launchd/launchd_darwin.go
cfergeau: might be worth a comment explaining this is for migration from older versions

praveenkumar: Done.

PR #5358, opened 2026-08-26
hostsapi: add bearer token authentication to daemon hosts API

praveenkumar:
Generate and persist a hosts API token when the daemon starts and during preflight setup. Require the token as an `Authorization: Bearer` header on the `/hosts/add` and `/hosts/remove` admin-helper endpoints. Sync the same token into an OpenShift Secret so routes-controller can authenticate when updating the hosts file. Add unit tests for token creation and authorization checks.

## Type of change

- [] Feature (non-breaking change which adds functionality)

## Testing
- Setup crc with this patch and 4.22.11 bundle and start it.
- Create new project and run a sample app creation which add routes
- with this patch route is added to /etc/hosts.
- If you try to run 4.22.11 bundle without this patch then routers-controller not work because it need the token to start the pod.

## Summary by CodeRabbit

* **New Features**
  * Added secure REST API token generation, storage, validation, and reuse.
  * Protected host add and remove requests with bearer-token authorization.
  * Automatically synchronized the API token with the cluster during startup.
* **Bug Fixes**
  * Unauthorized host requests now return HTTP 401 without changing host configuration.
* **Preflight Checks**
  * Added token validation, creation, and cleanup checks on Linux and macOS.


pkg/crc/hostsapi/token.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/hex"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	// EnvVar is the environment variable routes-controller should send as a Bearer token.
+	EnvVar = "CRC_HOSTS_API_TOKEN"
```
anjannath: This `EnvVar` is not being used it seems?

cfergeau: I’m not entirely sold on  the `HOSTS` as part of the naming, `CRC_API_TOKEN` or `CRC_REST_API_TOKEN`, or `CRC_DAEMON_API_TOKEN` might be clearer?

pkg/crc/preflight/preflight_checks_linux.go
```diff
@@ -420,7 +423,12 @@ func fixDaemonSystemdSockets() error {
 		return err
 	}
 
-	return fixSystemdUnit(vsockUnitName, vsockUnit, true)
+	if err := fixSystemdUnit(vsockUnitName, vsockUnit, true); err != nil {
+		return err
+	}
+	// Create the hosts API token when setting up the vsock socket so it
+	// exists before socket activation or crc start Secret sync.
+	return fixHostsAPIToken()
```
cfergeau: I’d move this to its own preflight check as this has nothing to do with systemd.

pkg/crc/preflight/preflight_checks_darwin.go
```diff
@@ -199,14 +199,22 @@ func fixDaemonPlistFileExists() error {
 	if err != nil {
 		return err
 	}
-	return fixPlistFileExists(*daemonConfig)
+	if err := fixPlistFileExists(*daemonConfig); err != nil {
+		return err
+	}
+	// Create the hosts API token when setting up the daemon so it exists
+	// before crc start Secret sync.
+	return fixHostsAPIToken()
```
cfergeau: Same comment as for the linux preflights

pkg/crc/hostsapi/token.go
```diff
@@ -0,0 +1,78 @@
...
+	token, err := generateToken()
+	if err != nil {
+		return "", err
+	}
+	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
+		return "", fmt.Errorf("creating hosts API token directory: %w", err)
+	}
+	if err := os.WriteFile(path, []byte(token+"\n"), 0o600); err != nil {
+		return "", fmt.Errorf("writing hosts API token: %w", err)
+	}
```
anjannath: i don't think this code path is useful, if the token doesn't exist already event tho it'll create a new token  the daemon/admin-helper will not load the new token, so maybe it'd be good to only create the token in preflight, wdyt?

pkg/crc/preflight/preflight_checks_unix.go
```diff
@@ -200,3 +201,31 @@ func removeCrcSymlink() error {
...
+	}
+	if !info.Mode().IsRegular() {
+		return fmt.Errorf("hosts API token path is not a regular file: %s", constants.HostsAPITokenPath)
+	}
+	_, err = hostsapi.LoadToken(constants.HostsAPITokenPath)
+	return err
+}
+
+func fixHostsAPIToken() error {
+	if _, err := hostsapi.LoadOrCreateToken(constants.HostsAPITokenPath); err != nil {
```
cfergeau: Imo it would be better to always create a new token when we create a new machine. Doing it as part of preflights mean we’ll have a long lived token which is almost never renewed.

anjannath: this might not be possible, since we need to restart the daemon to load the new token

praveenkumar: @cfergeau we have long lived daemon process and the token is consumed by admin-helper request which is a mux started by daemon process. As of now when we delete crc we don't remove/kill the daemon process until user do `cleanup`.

cfergeau: we can implement `SIGHUP` handling to tell the daemon to reload its config (or in our case, the token file), or add file monitoring over the token file.

pkg/crc/hostsapi/token.go
praveenkumar: @anjannath this function is used in daemon and preflight so if the token is not created in daemon then it will be in preflight in case check fails and recreate it but then it is not consumed by daemon which defeat the purpose of token since it is not used by admin mux. May be it should be only created as part of daemon process and preflight just do the check but not able to fix it.

pkg/crc/hostsapi/token.go
praveenkumar: @cfergeau whatever the name we choose we need to update  https://github.com/crc-org/snc/pull/1259 and https://github.com/crc-org/routes-controller/pull/60 because there also we are as of now using `CRC_HOSTS_API_TOKEN` so if we agree to use as `CRC_REST_API_TOKEN` then I can hold this PR and first do the changes on those place.

praveenkumar: https://github.com/crc-org/snc/pull/1262 and https://github.com/crc-org/routes-controller/pull/63 created to use `CRC_REST_API_TOKEN` now.

pkg/crc/preflight/preflight_checks_unix.go
praveenkumar: May be we don't need long running daemon as part of crc-ng because we want to remove crc api and also replace the virtualnet with gvproxy so only thing is the admin-helper so we can refactor more around as part of crc-ng side.

pkg/crc/preflight/preflight_checks_unix.go
```diff
@@ -200,3 +201,31 @@ func removeCrcSymlink() error {
...
+	if os.IsNotExist(err) {
+		return fmt.Errorf("REST API token file does not exist at %s", constants.RestAPITokenPath)
+	}
+	if err != nil {
+		return err
+	}
+	if !info.Mode().IsRegular() {
+		return fmt.Errorf("REST API token path is not a regular file: %s", constants.RestAPITokenPath)
+	}
+	_, err = restapi.LoadToken(constants.RestAPITokenPath)
```
anjannath, line 215: maybe we remove the file checks from this function, it seems the `LoadToken` helper also checks if the file exists.. wdyt?

anjannath (approved):
LGTM

pkg/crc/preflight/preflight_checks_unix.go
praveenkumar: yes, that looks like redundant now, will update

praveenkumar: @anjannath looking more carefully, this is not same because `LoadToken` directly read the file not check it's existence. So here we are checking the existing of file and if that is regular for read/write operation.

anjannath: but effectively since this is only a check, it'll be similar kind of error if directly reading it fails, but this is more thorough

anjannath (approved):

praveenkumar:
/hold

checking the CI errors.

anjannath:
Test needs to be skipped on Windows, maybe?
```
[0](https://github.com/crc-org/crc/actions/runs/34959799480/job/104350405216?pr=5358#step:5:1411)
=== RUN   TestLoadOrCreateToken_CreatesAndReuses
    token_test.go:23: 
        	Error Trace:	D:/a/crc/crc/pkg/crc/restapi/token_test.go:23
        	Error:      	Not equal: 
        	            	expected: 0x180
        	            	actual  : 0x1b6
        	Test:       	TestLoadOrCreateToken_CreatesAndReuses
--- FAIL: TestLoadOrCreateToken_CreatesAndReuses (0.00s)
```

